PR #932, opened 2015-08-17
Do not require extensions to be signed

joshbruning:
A pref to require installs be signed by a Mozilla-issued
add-on signing certificate is set to true by default in
Firefox 41, causing the WebDriver Firefox extension
to fail to install. Setting it to false allows WebDriver
to continue working on Firefox 41 in lieu of being
signed.


joshbruning:
FYI: I verified prior to this change, Firefox 41 Beta cannot be provisioned, and I verified after the change Firefox 41 Beta can be provisioned - and that there is no regression on Firefox 39 and 40.

lukeis:
thanks! merged with ca1e60e0e68c2eea323e267b7d09c7b2289314fa

PR #963, opened 2015-08-24
Add FluentWait.withMessage with string supplier

o-schneider:
Add a withMethod(Supplier<String> messgaeSupplier) to be able to evaluate the message when the wait is over, thus enabling developers to show more info on the context of failure through closures. Internally, the FluentWait class is using a supplier.
withMessage(String message) method kept for backward compatibility creates a supplier that return the given string.


o-schneider:
Okay, as I'm a little stupid, I wrote the patch using lambdas directly in the browser, but the project is still built with Java 7. I close this, and will be making a proper JDK7 PR with tests. :)

PR #964, opened 2015-08-25
Add FluentWait.withMessage with string supplier

o-schneider:
Add a withMethod(Supplier<String> messgaeSupplier) to be able to
evaluate the message when the wait is over, thus enabling developers to
show more info on the context of failure through closures. Internally,
the FluentWait class is using a supplier. withMessage(String message)
method kept for backward compatibility creates a supplier that return
the given string.


o-schneider:
I signed the CLA

barancev:
Merged as fef98c146cac5f089f537d230f46131990ceaeb8

PR #969, opened 2015-08-26
This fixes #498

### LINKED ISSUE #498: Send_keys Python 2.7 function does not handle long type variables
If I try to send a long number to edit box in Python 2.7, I get a type error. This is because function does not check for long type like it does for int. This happens only on Python 2.7 because Python 3 does not have long type.

>Type should be long: <type 'long'>
>Traceback (most recent call last):
 > File "sele.py", line 7, in <module>
 >   element.send_keys(longnumber)
 > File "c:\python27\lib\site-packages\selenium\webdriver\remote\webelement.py", line 320, in >send_keys
>    for i in range(len(val)):
>TypeError: object of type 'long' has no len()

Code to reproduce:
```python
driver = webdriver.Firefox()
driver.get('http://www.google.com')
element = driver.find_element_by_id('lst-ib')
longnumber = (65536*65536)
print ('Type should be long: %s' % type(longnumber))
element.send_keys(longnumber)
```

whorlwater:
Support for all Number types in WebElement.send_keys(), instead of only int.
Prevents TypeError from long type in Python 2.7 and is Python 3 compatible.

### DIFF
--- a/py/selenium/webdriver/remote/webelement.py
+++ b/py/selenium/webdriver/remote/webelement.py
@@ -16,6 +16,7 @@
 # under the License.
 
 import hashlib
+from numbers import Number
 import os
 import zipfile
 try:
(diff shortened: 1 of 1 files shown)


ddavison:
fixed in 2ac6a7d.  thanks :)

PR #971, opened 2015-08-26
java: fix log string for RemoteTargetLocator#defaultContent

alb-i986:
driver.switchTo().defaultContent() used to log the following:

   Executing: [switch to frame: null]

It turns out that the handler SwitchToFrame expects the parameter id to be set to null for switching to the default content.


ddavison:
fixed in a2404b8. thanks! :+1:

PR #1003, opened 2015-09-03
browser name for edge should be MicrosoftEdge

vveliev:
When browser_name='edge'  is used selenium cannot find driver,


lukeis:
looks like i did the same change at the same time :)  closing this as it's already there now

PR #1006, opened 2015-09-04
added Microsoft edge, and another way to pass in driver location for WebDriver

vveliev:


java/server/src/org/openqa/grid/common/defaults/DefaultNode.json
```diff
@@ -30,11 +30,20 @@
...
+          "browserName": "MicrosoftEdge",
+          "platform": "WINDOWS",
+          "maxInstances": 5,
+          "seleniumProtocol": "WebDriver"
         }
       ],
   "configuration":
   {
     "proxy": "org.openqa.grid.selenium.proxy.DefaultRemoteProxy",
+    "Dwebdriver.ie.driver=C://Selenium/Node/IEDriverServer32.exe": "",
```
lukeis: we should not be coding into the default config these paths. Normally it's recommended to just put these executables on the system PATH. Also, this assumes windows, which is not something we can assume at that point (even though it configures 'ie')

lukeis:
i'm not inclined to accept this patch, this file is mostly meant to be used as an example format to then be changed by the user for their own environment and not intended for actual consumption.

vveliev:
I agree on the part that this is example file, and people who know what they are doing may find it useful. 
It's personal choice to decide if you want where to include executables, and I spend lot's of time before i found this solution. And I also believe it makes my start script be cleaner. 

same option will apply for linux drivers: 
"Dwebdriver.chrome.driver=/opt/selenium/chromedriver":"",

lukeis:
normally those options should be passed in on the command line starting the node itself. it's a bit of a hack to put them into the node config like that... and I'm not even sure that works as expected.

vveliev:
Well I've been using it for about half a year or so , have not notice any issues with drivers , Windows 7 , 10(new box for edge) and Linux boxes . 

Anyway your call , my philosophy: if node will not work without this configs , then this configs should be inside this file. 

But as I said it's your call , does  not effect my setup. But it's nice alternative to know about.

cgoldberg:
@vveliev, why would you assume other users want *your* specific paths hardcoded?   You have no idea where they have browsers installed.

vveliev:
@cgoldberg , that is true , but from my knowledge this configs are only used as example. 

I think we are wasting more time on this then we should.

cgoldberg:
> I think we are wasting more time on this then we should.

How so?  I spent less than 10 seconds reviewing your patch.

vveliev:
@cgoldberg , I guess it's me spending more time then I should trying to backup my commit, when I have everything set-up and will not be using this file regardless. 

This is example config. 

When I set-up selenium grid initially, I spend  couple of hours to find solution when drivers information can be included in to the node configuration file.(My personal preference)

This config file does not support "\#" comments so I cannot include it as the comment. 
I thought it's common sense to remove drivers of replace the location when people using example config file for their needs.    

I have no interest to argue if we should have value hard-coded or not.
Then most generic example should only have Firefox and Crome since then can be installed on Linux , Windows and OS X.

PR #1007, opened 2015-09-04
added Microsoft edge, and another way to pass in driver location

vveliev:


lukeis:
same comments from #1006 apply to this. Also it would be nicer if the commits were together in one pull request.

PR #1030, opened 2015-09-13
Added newWindowAvailable until Condition

ghost:
Added a condition that waits for a new window to be available.
When successful, this returns the handle of the new window so
it is easily available to switch too.

CLA has been signed!

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/1030)

### DIFF
--- a/javascript/webdriver/until.js
+++ b/javascript/webdriver/until.js
@@ -129,6 +129,27 @@ until.ableToSwitchToFrame = function(frame) {
   }
 };
 
+/**
+ * Creates a condition that waits for a new window to be present.
+ * Once successful, a promise will be fulfilled with the handle
+ * for the new window.
+ *
+ * @return {!until.Condition.<!webdriver.Window>} New Condition
+ */
+until.newWindowAvailable = function() {
+  var numHandles;
+  return new until.Condition('for the number of handles to change', function(driver) {
+    return driver.getAllWindowHandles().then(function(handles) {
+      if(numHandles && handles.length > numHandles) {
+        return handles[handles.length - 1];
+      }
+      else {
+        numHandles = handles.length;
+        return false;
+      }
+    });
+  });
+};
 
 /**
  * Creates a condition that waits for an alert to be opened. Upon success, the


lukeis:
It would be useful to initialize numHandles when the condition is initially created, for cases when a 'click' will block and the new window already exists. 

Also, you should track the handles instead of trying to rely on order of them returned. The list returned by the drivers does not guarantee the newest is last in the list. You should do some 'set' logic to find the new handle.

ckuhn:
What would do you mean by initialize numHandles? What would it be initialized to?

lukeis:
The current set of window handles

ckuhn:
Ok. So call getAllWindowHandles() before returning the condition?

ddavison:
any update on this?  did you still need clarification on what @lukeis wanted, @ghost ?

javascript/webdriver/until.js
```diff
@@ -129,6 +129,27 @@ until.ableToSwitchToFrame = function(frame) {
...
+ * Once successful, a promise will be fulfilled with the handle
+ * for the new window.
+ *
+ * @return {!until.Condition.<!webdriver.Window>} New Condition
+ */
+until.newWindowAvailable = function() {
+  var numHandles;
+  return new until.Condition('for the number of handles to change', function(driver) {
+    return driver.getAllWindowHandles().then(function(handles) {
+      if(numHandles && handles.length > numHandles) {
```
jleyba: This will never be satisfied when going from 0 -> 1 window.

`if (typeof numHandles === 'number' && handles.length > numHandles) {`

javascript/webdriver/until.js
```diff
@@ -129,6 +129,27 @@ until.ableToSwitchToFrame = function(frame) {
   }
 };
 
+/**
+ * Creates a condition that waits for a new window to be present.
+ * Once successful, a promise will be fulfilled with the handle
+ * for the new window.
+ *
+ * @return {!until.Condition.<!webdriver.Window>} New Condition
```
jleyba: `{!until.Condition<string>}`

PR #1035, opened 2015-09-14
[java] new ExpectedCondition windowsToBe N

alb-i986:
**Use case**
- save number of current windows open (m)
- click link which opens a new window
- wait until windowsToBe m+1
- switch to the new window

About the name, another option is `numberOfWindowsToBe`, more verbose but also more precise.
What do you think?

Also, I see that `RemoteWebDriver#getWindowHandles` may throw WebDriverException in case of ClassCastException. Shall we catch it in windowsToBe? I'm asking that because the javadoc of `WebDriver#getWindowHandles` does not mention any `@throws`.

And how about adding an integration test (maybe in FirefoxDriverTest?) to have the use case above in the test suite (tests as documentation)?

Thanks

### DIFF
--- a/java/client/src/org/openqa/selenium/support/ui/ExpectedConditions.java
+++ b/java/client/src/org/openqa/selenium/support/ui/ExpectedConditions.java
@@ -830,6 +830,24 @@ public class ExpectedConditions {
     };
   }
 
+  public static ExpectedCondition<Boolean> numberOfwindowsToBe(final int expectedNumberOfWindows) {
+    return new ExpectedCondition<Boolean>() {
+      @Override
+      public Boolean apply(WebDriver driver) {
+        try {
+          return driver.getWindowHandles().size() == expectedNumberOfWindows;
+        } catch (WebDriverException e) {
+          return null;
+        }
+      }
+
+      @Override
+      public String toString() {
+        return "number of open windows to be " + expectedNumberOfWindows;
+      }
+    };
+  }
+
   /**
    * An expectation with the logical opposite condition of the given condition.
    *
--- a/java/client/test/org/openqa/selenium/support/ui/ExpectedConditionsTest.java
+++ b/java/client/test/org/openqa/selenium/support/ui/ExpectedConditionsTest.java
@@ -34,8 +34,10 @@ import static org.openqa.selenium.support.ui.ExpectedConditions.urlToBe;
 import static org.openqa.selenium.support.ui.ExpectedConditions.visibilityOf;
 import static org.openqa.selenium.support.ui.ExpectedConditions.visibilityOfAllElements;
 import static org.openqa.selenium.support.ui.ExpectedConditions.visibilityOfAllElementsLocatedBy;
+import static org.openqa.selenium.support.ui.ExpectedConditions.numberOfwindowsToBe;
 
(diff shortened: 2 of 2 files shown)


ddavison:
hey @alb-i986 , thanks for the PR.

so.. most of us committers will be hesitant to merge this sort of stuff because this class itself, is actually just a **sample** class.

it's so easy to create your own:

```java
public class Conditions {
  private Conditions() {}

  public static ExpectedCondition<Boolean> windowsToBe(final int expectedNumberOfWindows) {
    return new ExpectedCondition<Boolean>() {
      @Override
      public Boolean apply(WebDriver driver) {
        try {
          return driver.getWindowHandles().size() == expectedNumberOfWindows;
        } catch (WebDriverException e) {
          return null;
        }
      }

      @Override
      public String toString() {
        return "number of open windows to be " + expectedNumberOfWindows;
      }
    };
  } 
}
```

alb-i986:
Hey @ddavison ,
I see.
I do have my own CustomExpectedConditions, that's not a problem.
The idea here was to share a common ExpectedCondition with the community, so as to avoid reinventing the wheel.

But if you believe it's not interesting, I'll close it, as well as another PR I opened a while ago with another common ExpectedCondition.
In this case, it would be nice if we made `ExpectedConditions` extendable, by making the constructor `protected`, so as to allow clients to import all of the ExpectedCondition's, both official and custom, in one `import`, wouldn't it?

ddavison:
personally I mean I don't care. :) adding it is just fine, but other committers shrug and sigh at it....

* cough* @jimevans

jimevans:
Since I've been called out, let me state for the record that I'm only adamantly opposed to adding things to the .NET implementation. This is mostly because lambda functions are a very common structure in modern .NET programming, obviating the need for a specialized class whose methods merely return functions with the proper signature.

ddavison:
this is true. only recently has java supported lambda functions in java 8.

I wouldn't mind adding this tomorrow.  I'm all for providing more conditions for people, as long as they know they can implement their own expected conditions.

ddavison:
@alb-i986 i do like the idea of it being "numberOfWindowsToBe".  Go ahead and change that, and squash your commits into one, and I'll merge it in.

alb-i986:
@jimevans 
I think it's not only a matter of signature but also of body: even with lambda functions you still need a class collecting common behaviors. How would you achieve code reuse otherwise?

jimevans:
@alb-i986 Honestly? I'd have a companion project, not part of Selenium proper. But I don't get to make those decisions for non-.NET code.

alb-i986:
@jimevans 
Fair enough, that's an option. So you agree on having such a class.

Not sure if this is what you mean, but actually it turns out (I've just found out) that the package "support", of which ExpectedConditions is part, is already published as a separate maven artifact:

http://search.maven.org/#search%7Cga%7C1%7Ca%3A%22selenium-support%22

Which is included in selenium-java, which users usually depend on.

@ddavison 
When you're ready, I squashed the commits as requested. Please let me know if it's ok.

ddavison:
merged as 270fcf664da7fe0a58f59426e4be085b2ef2766e.  thanks :)  (sorry it took so long)

PR #1036, opened 2015-09-15
Add Ruby test suite to Travis

p0deje:


lukeis:
:+1: minus the testing bits of course ;)

p0deje:
@lukeis Sure, I'll clean everything up once I'm 100% sure everything works smoothly.

p0deje:
@titusfortner We now have some unit/phantomjs tests for Ruby on Travis.

titusfortner:
It's great to have something, finally! I guess I'm curious what we expect this to catch with this, though, since most of our unit tests seem somewhat trivial and our integration testing is a virtual browser using a deprecated version of Ruby. I suppose it works as a very basic regression check that we didn't do something spectacularly dumb in a code update. Especially as we start to work on w3c implementation, though, it seems like it would be good to have a more complete set of regression tests. Should the Ruby bindings try to duplicate everything that Java is doing? The supported Ruby versions with the supported browsers with different settings, on each of the main 3 platforms, and all sent to the java server as well (since there was that weird Edge bug where the name worked locally but not through the server)... or do we do the basic checks and rely on bug reports? :)

lukeis:
@titusfortner just wanted to say that this only runs against pull requests and like Python only proves to be a quick sanity that it compiles and basic functionality. We still need to add to CI all the combinations of test (CI being jenkins)

titusfortner:
@lukeis - is that something you are working on for Python as well? Can the Jenkins server we are using handle running browsers directly (not through Sauce), or do we need to get that part working in Travis?

bootstraponline:
It'd be nice to run on Sauce to get videos, selenium logs as json, screenshots, multiple browser support, etc. Does SeleniumHQ use both Travis and Jenkins?

titusfortner:
We're using Travis for sanity checks only for both Ruby and Python. The rest of the testing is done here: http://ci.seleniumhq.org:8080

titusfortner:
The problem with Sauce is that everything is remote, and I've gotten different test results manually running local and remote. So just remote probably isn't sufficient.

bootstraponline:
The issue with local runs on a service like Travis is when something goes wrong, there's almost no information to debug the problem sufficiently. Parallel local browsers (IE, Safari, Firefox, etc.) is even more fun.

lukeis:
for our purposes we have to be able to run on the machine, sauce only provides us a way to inject the selenium standalone server... not do things like run python or ruby or .NET without doing a 'remote' webdriver call.

To that end, as mentioned, travis isn't ideal and we will likely be strapping together some machines to hook up to our jenkins instance to get the additional coverage.

bootstraponline:
Makes sense.

lukeis:
I should say that, travis doesn't even support all the platforms we need... (windows!)

titusfortner:
Watir uses Travis with chrome and Firefox and multiple versions of ruby, but it looks like that account is grandfathered into a setup I couldn't replicate for selenium. Also, it's only Linux. Also I don't think it would work with the current rake file. I need to see if we can get this implementation to use a different ruby version at least.

PR #1037, opened 2015-09-15
Fix the problem with adding of plugin into firefox profile. We tried …

BlackSmith:
The fix  solves the problem with adding of plugin into firefox profile. We tried  create a folder, which was already created, because we create the folder during creating of file (if it's missing) and then we tried to create folder itself.

Error in my case:
`OSError: [Errno 17] File exists: '/tmp/tmpu0hQBg.firebug-2.0.12.xpi/content/firebug/net/`

### DIFF
--- a/py/selenium/webdriver/firefox/firefox_profile.py
+++ b/py/selenium/webdriver/firefox/firefox_profile.py
@@ -252,7 +252,7 @@ class FirefoxProfile(object):
         if addon.endswith('.xpi'):
             tmpdir = tempfile.mkdtemp(suffix='.' + os.path.split(addon)[-1])
             compressed_file = zipfile.ZipFile(addon, 'r')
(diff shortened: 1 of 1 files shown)


lukeis:
there's already guards in place if the directory doesn't exist, not sure how you're getting this issue... can you show me some example code + the xpi that caused the error?  The following ran successfully for me.

```
>>> from selenium import webdriver as w
>>> fp = w.FirefoxProfile()
>>> fp.add_extension("/Users/linman/Downloads/firebug-2.0.12-fx.xpi")
>>> d = w.Firefox(firefox_profile=fp)
```

BlackSmith:
Sorry, I can't  reproduce again. Now it looks like good. Maybe, the problem was solved in the some  python packages update.  :o(

lukeis:
Ok, going to close this for now then.

PR #1039, opened 2015-09-15
Fixes non-relative import in chrome webdriver

clavery:
When installing the latest version from pypi (2.47.2) an ImportError occurs (on python 3.4) in code like:

``` python
  File "/Users/clavery/.venv3/default/lib/python3.4/site-packages/selenium-2.47.2-py3.4.egg/selenium/webdriver/__init__.py", line 20, in <module>
    from .chrome.webdriver import WebDriver as Chrome
  File "/Users/clavery/.venv3/default/lib/python3.4/site-packages/selenium-2.47.2-py3.4.egg/selenium/webdriver/chrome/webdriver.py", line 19, in <module>
    from remote_connection import ChromeRemoteConnection
ImportError: No module named 'remote_connection'
```

The remote_connection module is not being imported via a relative import.

I have signed the CLA.

### DIFF
--- a/py/selenium/webdriver/chrome/webdriver.py
+++ b/py/selenium/webdriver/chrome/webdriver.py
@@ -16,10 +16,10 @@
 # under the License.
 
 import base64
-from remote_connection import ChromeRemoteConnection
 from selenium.webdriver.remote.command import Command
 from selenium.webdriver.remote.webdriver import WebDriver as RemoteWebDriver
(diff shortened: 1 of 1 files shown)


lukeis:
arg, forgot to test against py3 :)

Thanks for the patch, merged with e634a381412011f547d2b532fb1602dccb5e1c42

lukeis:
2.47.3 is now in pypi

PR #1040, opened 2015-09-15
{} vs [] for use with non-number indices

joshbruning:
Prepare for tightening of the Array definition which would
make this warning an error.


lukeis:
@joshbruning this change should be done in the wicked good xpath project first and then we should import the updated version -

https://github.com/google/wicked-good-xpath/blob/master/src/functionCall.js#L494

joshbruning:
For anyone watching this, see https://github.com/google/wicked-good-xpath/pull/24

PR #1042, opened 2015-09-16
rb - update file loading to allow tests to be run from relative directories

titusfortner:
This came up as an issue at the selenium conference workshop when the Ruby group couldn't find an easy way to run a single ruby test in the project.

With this code we can run specific spec files within the selenium/rb directory like this:
export WD_SPEC_DRIVER=chrome; rspec spec/integration/selenium/webdriver/driver_spec.rb
This also allows running from within Rubymine (importantly including using the debugger).

@p0deje - I changed spec_support.rb to require_relative instead of autoload for RackServer. I didn't dig into why RackServer needs to be lazy loaded; I'm guessing it's because we don't need it if we're just running unit tests. That doesn't seem to be much overhead, but if the "right answer" to this is easy and better, let me know, otherwise let's merge this.

(also, I ran ./go test_rb on my mac with this code and everything passed).

### DIFF
--- a/rb/spec/integration/selenium/webdriver/app_cache_spec.rb
+++ b/rb/spec/integration/selenium/webdriver/app_cache_spec.rb
@@ -17,7 +17,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-require File.expand_path("../spec_helper", __FILE__)
+require_relative 'spec_helper'
 
 
 module Selenium::WebDriver::DriverExtensions
--- a/rb/spec/integration/selenium/webdriver/driver_spec.rb
+++ b/rb/spec/integration/selenium/webdriver/driver_spec.rb
@@ -17,7 +17,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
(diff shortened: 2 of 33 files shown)


p0deje:
If it works, why not?

My only question is - does it work when spec is run outside `rb` directory? For example:

```
$ cd rb/spec/integration/
$ bundle exec rspec selenium/webdriver/app_cache_spec.rb
```

titusfortner:
Yes, that works as well. Thanks, I'll merge.

p0deje:
@titusfortner Ouch, ain't we rebase and push instead of merging PRs?

titusfortner:
I wanted a merged PR to link to. :) I thought if I didn't use the github UI that I would then have to manually close the PR and I thought that would be confusing.

p0deje:
@titusfortner If you just rebase and push to master, PR will be shown as merged too.

titusfortner:
Yeah, @lukeis educated me on this on IRC earlier. :)

PR #1049, opened 2015-09-17
rb: Switch to RSpec expect syntax

p0deje:
@titusfortner Please, take a look. Conversion has been made with [transpec](https://github.com/yujinakayama/transpec) so there is not much to review here. If you're fine with these changes, I'll rebase and push to master.

### DIFF
--- a/rb/lib/selenium/client/base.rb
+++ b/rb/lib/selenium/client/base.rb
@@ -20,7 +20,7 @@
 module Selenium
   module Client
 
-		# Driver constructor and session management commands
+    # Driver constructor and session management commands
     module Base
       include Selenium::Client::Protocol
       include Selenium::Client::GeneratedDriver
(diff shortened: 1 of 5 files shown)


titusfortner:
Yes, much needed, I'm tired of looking at the shoulds. :)
I'm assuming you did a full ./go test_rb with this?

p0deje:
Yeah, the Safari part was failing, but everything else passed. I'm merging this then.

titusfortner:
@p0deje  - Something in this commit caused it to start logging this on every single test:
"localhost - - [17/Sep/2015:21:43:26 PDT] "GET /xhtmlTest.html HTTP/1.1" 200 1923
- -> /xhtmlTest.html"
Can you take a look and see what might have caused it?

p0deje:
@titusfortner Not sure why, but it happens with RSpec >= 3. I've not dug much and just made WEBrick silent. See https://github.com/SeleniumHQ/selenium/commit/c7d9e989285ee60662186f66ece45c16ce118c04.

PR #1065, opened 2015-09-23
Python 3 compatibility for remote Authorization

lrowe:
Avoids `TypeError: a bytes-like object is required, not 'str'`.

### DIFF
--- a/py/selenium/webdriver/remote/remote_connection.py
+++ b/py/selenium/webdriver/remote/remote_connection.py
@@ -416,8 +416,8 @@ class RemoteConnection(object):
                        "Content-type": "application/json;charset=\"UTF-8\"",
                        "Accept": "application/json"}
             if parsed_url.username:
-                auth = base64.standard_b64encode('%s:%s' %
-                       (parsed_url.username, parsed_url.password)).replace('\n', '')
+                auth = base64.standard_b64encode(('%s:%s' %
+                       (parsed_url.username, parsed_url.password)).encode('ascii')).decode('ascii').replace('\n', '')
                 headers["Authorization"] = "Basic %s" % auth
             if body and method != 'POST' and method != 'PUT':
                 body = None


lukeis:
@lrowe this doesn't appear to work for me in python3, well, the whole thing doesn't work... since base64.standard_base64encode  requires a 'bytes' object and not a string:

```
>>> import base64
>>> base64.standard_b64encode('lukeis:!@#%@!!#'.encode('ascii').decode('ascii').replace('\n', ''))
  File "/usr/local/Cellar/python3/3.4.3/Frameworks/Python.framework/Versions/3.4/lib/python3.4/base64.py", line 98, in standard_b64encode
    return b64encode(s)
  File "/usr/local/Cellar/python3/3.4.3/Frameworks/Python.framework/Versions/3.4/lib/python3.4/base64.py", line 62, in b64encode
    encoded = binascii.b2a_base64(s)[:-1]
TypeError: 'str' does not support the buffer interface
```

so the line argument passed to the b64encode needs to be surrounded with bytes(<value>, 'UTF-8')... problem is that syntax is not python 2 compatible... so looks like we'll need to handle that here in some way.

lrowe:
@lukeis there is a new open parens on the previous line of my diff, equivalent to:
>>> base64.standard_b64encode(('lukeis:!@#%@!!#'.encode('ascii')).decode('ascii').replace('\n', ''))

lukeis:
@lrowe the point is it results in this error for python3:

  File "/usr/local/Cellar/python3/3.4.3/Frameworks/Python.framework/Versions/3.4/lib/python3.4/base64.py", line 98, in standard_b64encode
    return b64encode(s)
  File "/usr/local/Cellar/python3/3.4.3/Frameworks/Python.framework/Versions/3.4/lib/python3.4/base64.py", line 62, in b64encode
    encoded = binascii.b2a_base64(s)[:-1]
TypeError: 'str' does not support the buffer interface

lrowe:
@lukeis I messed up my example in my comment, but the patch has been run on Python3, here is the code in the patch:
```
$ python3
Python 3.5.0 (default, Sep 22 2015, 12:32:59) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.72)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import base64
>>> class parsed_url:
...     username = 'user'
...     password = 'pass'
... 
>>> auth = base64.standard_b64encode(('%s:%s' %
...        (parsed_url.username, parsed_url.password)).encode('ascii')).decode('ascii').replace('\n', '')
>>> auth
'dXNlcjpwYXNz'
```

lukeis:
ah, ha. Thanks. Also, have you signed the cla?

https://docs.google.com/forms/d/11Z8LoYpTGUIwCegifVH1YtL9smxVDNk-fOykUZTAWhE/viewform?hl=en_US&formkey=dFFjXzBzM1VwekFlOWFWMjFFRjJMRFE6MQ#gid=0

lrowe:
Signed the CLA just now.

lukeis:
thanks! merged as 01cbd65f7c9175c4fb0564b5b3b1d6e3258d7f2b

PR #1069, opened 2015-09-24
Fix WebDriver Promise tests for Microsoft Edge.

juangj:
Edge modifies the error's stack when the error is thrown, so we have to grab the "original" stack later, after the error is thrown.

### DIFF
--- a/javascript/webdriver/test/promise_test.js
+++ b/javascript/webdriver/test/promise_test.js
@@ -1862,7 +1862,7 @@ function testLongStackTraces_appendsInitialPromiseCreation_resolverThrows() {
   webdriver.promise.LONG_STACK_TRACES = true;
 
   var error = Error('hello');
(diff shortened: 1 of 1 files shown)


juangj:
FWIW, some evidence that Edge does this:

```
<script>
var error = Error('hello');
console.log(error.stack);

try {
  (function() {
    throw error;
  })();
} catch (e) {
  console.log(e.stack);
}
</script>
```

Output:

```
Error: hello
   at Global code (http://localhost/throw.html:2:1)
throw.html (3,1)
Error: hello
   at Anonymous function (http://localhost/throw.html:7:5)
   at Global code (http://localhost/throw.html:6:4)
throw.html (10,3)
```

lukeis:
@jleyba

jleyba:
Merged in e255009dcdcd836b5dba9b66ee431e31478ea3a9

PR #1081, opened 2015-09-26
Added a builder for creating DesiredCapabilities.

alexec:
Adding this builder will allow more concise ways of creating capabilities.

### DIFF
--- a/java/client/src/org/openqa/selenium/remote/BUCK
+++ b/java/client/src/org/openqa/selenium/remote/BUCK
@@ -16,6 +16,7 @@ java_library(
     'BrowserType.java',
     'CapabilityType.java',
     'DesiredCapabilities.java',
+    'DesiredCapabilitiesBuilder.java',
   ],
   deps = [
     '//java/client/src/org/openqa/selenium:webdriver-api',
--- a/java/client/src/org/openqa/selenium/remote/build.desc
+++ b/java/client/src/org/openqa/selenium/remote/build.desc
@@ -12,6 +12,7 @@ java_library(name = "base",
   srcs = [
     "CapabilityType.java",
     "DesiredCapabilities.java",
+    "DesiredCapabilitiesBuilder.java",
     "BrowserType.java",
     "UnreachableBrowserException.java",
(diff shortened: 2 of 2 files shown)


jimevans:
I oppose this pull request. We should be moving away from direct use of `DesiredCapabilities` and encouraging the use of type safe browser-specific options classes (for strongly-typed languages like Java and C#).

lukeis:
@alexec the build fails, because you'll need to add this file to the build rules (to both build.desc and BUCK in the same directory).

Also as Jim mentions, it would be more worthwhile to invest changes like this into a Options class for each of the browsers (that would in turn generate a DesiredCapability)

alexec:
Hi @lukeis , @jimevans , I have fixed the build (well, works on my machine, waiting for TravisCI to complete). I'm afraid I don't know much about long term/strategic goals, but I would hope this is useful when working with remote driver. I'll have a think about a solution to more strongly typed options.

alexec:
Driver Class | Options Class | Has `toCapabilities`?
---|---|---
ChromeDriver | ChromeOptions | Yes
EdgeDriver | EdgeOptions | Yes 
FirefoxDriver | - | -
HtmlUnitDriver | - | -
InternetExplorerDriver | - | -
OperaDriver | OperaOptions | Yes

alexec:
So... some looking into the source code has shown me that there does not appear to be a one-to-one mapping for `Capabilities` and `*Options` classes. For example, `ChromeOptions`, does not, contain `platform` or `version`. 

You can also do things which might be an oxymoron:

~~~
new ChromeDriver(DesiredCapabilities.firefox());
~~~

I guess I'm left scratching my head to what a `Capability` is? Is seems to be a composite of:

1. Which browser.
2. What platform the browser is on.
3. What version of the binary file we're using, but also where that binary is.
4. Configuration of the binary, e.g. extensions.
5. Options for the browser (which may be keyed on something like `ChromeOptions.CAPABILITIES`.
6. User profile.
7. Abilities of the browser, e.g. does it run JS.

PR #415, opened 2015-04-05
Java. Page Factory enhancement.

TikhomirovSergey:
- Ability to use not only Webdriver to populate page object fields via method

``` java
public static <T> T initElements(SearchContext context, Class<T> pageClassToProxy).
```

This change was done in order to simplify page object instantiation and population when there is the necessity to implement something that describes widget or repeatable element set nested in a root element. So if there is a class with constructors like follows:

``` java
public PageObject(Search context)

//or

public PageObject(WebElement rootElement)
```

and there is no necessity to implement customized decorating engines then it is possible to populete its WebElement-fields this way (only one step):

``` java
PageObject po = PageFactory.initElements(driver, PageObject,class) //then PageObject.class has 
//the default constructor or constructor with a SearchContext- or WebDriver- parameter


//or 
PageObject po = PageFactory.initElements(element, PageObject,class) //then PageObject.class has 
//the default constructor or constructor with a SearchContext- or WebElement- parameter

//and so on... 
```
- the additional checking of field modifiers before it will be populated. 

Field that is going to be filled should not  be static and final. I found these issues some time ago. I tried to implement page object with final fields that weren't annotated by @FindBy's. Also I found that it tries to set proxy-WebElement as a value of a static WebElement-field. In both cases I caught **IllegalModifierException**. Actually I think that both issues are minor but it would be good if whey were fixed.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/415)

PR #1082, opened 2015-09-26
Change PageFactory so that it support creating page object for part o…

alexec:
…f a page, but allowing us to just pass the SearchContext as a parameter


TikhomirovSergey:
This PR duplicates #415 that is waiting for the decision already.

alexec:
Correct. This is a duplicate -- great minds eh? :)

PR #1105, opened 2015-10-05
Fix log to file for Windows (Use platform-dependant separator)

sdimkov:
Use platform-dependant line separator instead of hardcoded '\n'.

Fixes logging to file for Windows.

### DIFF
--- a/java/server/src/org/openqa/selenium/remote/server/log/TerseFormatter.java
+++ b/java/server/src/org/openqa/selenium/remote/server/log/TerseFormatter.java
@@ -44,8 +44,7 @@ public class TerseFormatter extends Formatter {
    * Line separator string.  This is the value of the line.separator
    * property at the moment that the TerseFormatter was created.
    */
-  private final String lineSeparator = java.security.AccessController.doPrivileged(
-      new sun.security.action.GetPropertyAction("line.separator"));
+  private final String lineSeparator = System.getProperty("line.separator");
 
   /*
    * DGF - These have to be compile time constants to be used with switch


java/server/src/org/openqa/selenium/remote/server/log/TerseFormatter.java
```diff
@@ -40,6 +40,13 @@
    */
   private static final String SUFFIX = " - ";
 
+  /**
+   * Line separator string.  This is the value of the line.separator
+   * property at the moment that the TerseFormatter was created.
+   */
+  private final String lineSeparator = java.security.AccessController.doPrivileged(
+      new sun.security.action.GetPropertyAction("line.separator"));
```
jleyba: No dependencies on the sun JDK

sdimkov: Noted. I'll fix it in a bit.

I took this line from the reference implementation in the JDK itself. I didn't realize though that it uses internal API.

ddavison:
how about `String.format("%n")` instead? also, instead of declaring a variable, just pass it in considering theres onlt one usage of it. 

additionally, this change would make it platfom **in**dependent ;)

sdimkov:
I believe using %n conversion and line separator are equally portable and platform independent.

I guess the only real advantage of using the property approach is being able to override it while with format() conversion literal you can't. 

Also using local variable will perform a bit better than resolving formatting conversion but such micro-optimization is irrelevant.

I mostly wrote it this way because that's how this problem is resolved in the JDK itself:
http://hg.openjdk.java.net/jdk6/jdk6/jdk/file/814bf0775b52/src/share/classes/java/util/logging/SimpleFormatter.java

ddavison:
fair argument, thanks for replying

LGTM then :+1:

sdimkov:
Thanks!

I'll be glad to see it merged to master so that it comes with next release.

I have a grid setup on Windows and this bug makes my logs unreadable.

sdimkov:
@jleyba @ddavison any objections on merging this fix ?

ddavison:
done in d60bac12e49472745a41705f2c64b3c8436c890f

thanks! :)

PR #1115, opened 2015-10-07
Making ChromeDriver directly implement HasTouchScreen interface.

rovner:
To be able to use org.openqa.selenium.interactions.touch.TouchActions
with chrome mobile emulation
(see https://sites.google.com/a/chromium.org/chromedriver/mobile-emulation)
ChromeDriver should implement HasTouchScreen interface.
In other case, it would be impossible to use ChromeDriver to
simulate mobile evens such as flick or scroll in hub+node architecture
(see issue https://code.google.com/p/chromedriver/issues/detail?id=1029),
because ChromeDriver on hub doesn't't have touch screen capability.
With ChromeDriver implementing HasTouchScreen it is possible to
augment RemoteWebDriver on the client side and use touch actions.

### DIFF
--- a/java/client/src/org/openqa/selenium/chrome/ChromeDriver.java
+++ b/java/client/src/org/openqa/selenium/chrome/ChromeDriver.java
@@ -27,7 +27,10 @@ import org.openqa.selenium.html5.Location;
 import org.openqa.selenium.html5.LocationContext;
 import org.openqa.selenium.html5.SessionStorage;
 import org.openqa.selenium.html5.WebStorage;
+import org.openqa.selenium.interactions.HasTouchScreen;
+import org.openqa.selenium.interactions.TouchScreen;
 import org.openqa.selenium.remote.FileDetector;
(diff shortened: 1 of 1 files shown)


pazone:
This is really interesting functionality. Can anybody review it please?

jleyba:
The ChromeDriver only reports that it has a touch screen when mobile emulation is enabled, so marking it as always having a touch screen would not be appropriate. You can use the [org.openqa.selenium.remote.Augmenter](https://github.com/SeleniumHQ/selenium/blob/ab1e647d0fc8fc39e6b00ae94321ab228b6728f2/java/client/src/org/openqa/selenium/remote/Augmenter.java) to dynamically add the appropriate interfaces to a WebDriver instance based on the capabilities returned by the server.

Flaburgan:
I'm sorry to ask help here but can you be a but more specific about how to use that Augmenter class to add `HasTouchScreen` to my webdriver ?

PR #1121, opened 2015-10-07
java: fix camel case in method name numberOfWindowsToBe

alb-i986:
Sorry, I've just noticed a typo in the name of the method I added in #1035.
I see you've just released 2.48.0 so I've deprecated the old signature.

My apologies


ddavison:
that was my mistake too for not checking.  thanks for fixing that

merged as 5715c9

Issue #1137, opened 2015-10-11
README of PyPI package is broken and needs update


bittner:
The Python implementation of the bindings for Selenium on PyPI has a README with at least two issues:

1. README **[is outdated](https://github.com/SeleniumHQ/selenium/blob/master/py/README)** (e.g. supported Python version 3.4 not mentioned but [declared on PyPI](https://pypi.python.org/pypi/selenium#downloads))
1. README **[doesn't display properly on PyPI](https://pypi.python.org/pypi/selenium)** (the source code displays instead of the formatted text)

The latter is probably related to a syntax error introduced in [commit 3fc2db2](https://github.com/SeleniumHQ/selenium/commit/3fc2db275149748b21ea0370d3d3ace2df7c2676#commitcomment-13704294) (table markup, line 17). It may also be advisable to use the currently missing `.rst` file extension.

cgoldberg:
yikes... it looks like the PyPI page hasn't been rendering since the 2.45 release:

broken PyPI pages for last few releases:
* https://pypi.python.org/pypi/selenium/2.46.0
* https://pypi.python.org/pypi/selenium/2.46.1
* https://pypi.python.org/pypi/selenium/2.47.0
* https://pypi.python.org/pypi/selenium/2.48.0

I'm working on fixing these problems (and a few others I found).  I'll submit a PR shortly and link to it in this issue.

-Corey

cgoldberg:
sumbitte.
see the PR cover letter and diff for details: https://github.com/SeleniumHQ/selenium/pull/1138

barancev:
Fixed by commit 85741443926afdac0856c72b1fc099c0c448ebcb

cgoldberg:
8574144 did not fix this issue.  It just renamed "README" to "README.rst".  That change alone just makes it render nicer when viewing through GitHub's website.  An additional change needs to be made in order to fix the issue of rendering on PyPI:

> README doesn't display properly on PyPI (the source code displays instead of the formatted text)

When a release gets published to PyPI, it is done using info declared in `setup.py`.  https://github.com/SeleniumHQ/selenium/blob/master/setup.py#L37 still has the following line:

`'long_description': open(join(abspath(dirname(__file__)), "py", "README")).read(),`

which should be updated to reflect the new renamed README:

`'long_description': open(join(abspath(dirname(__file__)), "py", "README.rst")).read(),`

---

[...]

PR #1138, opened 2015-10-11
fixed README and index.rst for Python documentation

cgoldberg:
This PR fixes issues with the python README and index.rst docs.

issues fixed that were reported in #1137:
- fixed broken table that was making the document invalid reStructuredText.  In its current state, rendering of the document on PyPI is broken. (it shows raw source rather than rendered reStructuredText)  
  - see: https://pypi.python.org/pypi/selenium/2.48.0
  - this is a regression introduced by commit: https://github.com/SeleniumHQ/selenium/commit/3fc2db275149748b21ea0370d3d3ace2df7c2676
- changed supported Python versions to be more generic:
  - replaced "Python 2.6, 2.7" with "Python 2.6+"
  - replaced "Python 3.2, 3.3" with "Python 3.2+"

other fixes:
- renamed page from "Selenium Client Driver" to "Selenium WebDriver Client - Python Bindings", and removed redundant description.
- updated server download location and server run command to reflect version 2.48.0 (it wasn't updated since 2.46)
- tweaked link to GitHub for viewing source
- fixed a typo in page: s/uisng/using

Note: `py/docs/source/index.rst` is a copy of `py/README` and should be kept in sync.  This PR updates both docs.

### DIFF
--- a/py/README
+++ b/py/README
@@ -153,6 +153,7 @@ Use The Source Luke!
 ====================
 
 <<<<<<< HEAD
+<<<<<<< HEAD
 View source code online at:
 
 * https://github.com/SeleniumHQ/selenium/tree/master/py
@@ -163,3 +164,11 @@ View source code online at https://github.com/SeleniumHQ/selenium/tree/master/py
 >>>>>>> adding travis config to run python tests
 =======
 >>>>>>> Bump python to 2.47
+=======
+View source code online at https://github.com/SeleniumHQ/selenium/tree/master/py/
+=======
+View source code online at:
+
+* https://github.com/SeleniumHQ/selenium/tree/master/py
+>>>>>>> fixed README and index.rst for Python documentation
+>>>>>>> fixed README and index.rst for Python documentation


py/README
```diff
@@ -14,20 +12,20 @@ The `selenium` package is used automate web browser interaction from Python.
...
 
-Several browsers/drivers are supported (Firefox, Chrome, Internet Explorer, PhantomJS), as well as the Remote protocol.
+Several browsers/drivers are supported (Firefox, Chrome, Internet Explorer, PhantomJS, etc), as well as the Remote protocol.
 
 Supported Python Versions
 =========================
 
-* Python 2.6, 2.7
-* Python 3.2, 3.3
+* Python 2.6+
```
bittner: I think "2.6, 2.7" is fine. There will be no 2.8 ever according to our BDFL.

py/README
```diff
@@ -14,20 +12,20 @@ The `selenium` package is used automate web browser interaction from Python.
...
+* Python 3.2+
```
bittner: I think "3.2, 3.3, 3.4" is more accurate.

(Has it already been tested with 3.5? Will it automagically be with Python 3.6 once that is released?)

py/README
```diff
@@ -14,20 +12,20 @@ The `selenium` package is used automate web browser interaction from Python.
 +-----------+-----------------------------------------------------------------------------------+
 | **Docs**: | `selenium package API <http://selenium.googlecode.com/git/docs/api/py/api.html>`_ |
 +-----------+-----------------------------------------------------------------------------------+
-| **Dev**:  | https://github.com/SeleniumHQ/selenium                                               |
+| **Dev**:  | https://github.com/SeleniumHQ/selenium                                            |
```
bittner: I would recommend switching the whole table markup to the [simple table markup style](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#simple-tables). The grid tables are in general very fragile (read: error prone).

ddavison:
thanks for reviewing, @bittner .  @cgoldberg , your code does need changes as well as a rebase.  I'm not hugely familiar with the python bindings, so i can't speak for the versioning

bittner:
@cgoldberg Regarding the duplicate content, we should read content out from `README` in `index.rst`. See the [behave-django](https://raw.githubusercontent.com/mixxorz/behave-django/master/docs/index.rst) project docs for an example. The `.. include::` directive is described in the [Sphinx docs](http://sphinx-doc.org/rest.html#directives).

cgoldberg:
fucked up my branch and don't feel like fixing it.  deleting this PR

PR #1143, opened 2015-10-13
Improve firefox extension install.rdf parsing

jmif:
Firefox extension loading now supports em namespace (http://www.mozilla.org/2004/em-rdf#) as default namespace in install.rdf.  Some extensions set this namespace as default which breaks parsing in the current implementation as it results an <id> tag being present instead of em:id.

### DIFF
--- a/rb/lib/selenium/webdriver/firefox/extension.rb
+++ b/rb/lib/selenium/webdriver/firefox/extension.rb
@@ -62,12 +62,12 @@ module Selenium
           rdf_path = File.join(directory, "install.rdf")
           doc = REXML::Document.new(File.read(rdf_path))
 
-          id_node = REXML::XPath.first(doc, "//em:id")
+          id_node = REXML::XPath.first(doc, "//em:id|//id")
 
           if id_node
             id_node.text
           else
-            attr_node = REXML::XPath.first(doc, "//@em:id")
+            attr_node = REXML::XPath.first(doc, "//@em:id|//@id")
 
             if attr_node.nil?
               raise Error::WebDriverError, "cannot locate extension id in #{rdf_path}"


p0deje:
I've looked through your link and also checked [MDN](https://developer.mozilla.org/en-US/Add-ons/Install_Manifests) documentation for `install.rdf` and can't find an example of using `em` namespace - it says that you should use `<em:id>` for specifying addon indentifier.

Is there anything that I miss?

jmif:
uBlock and/or uBlock Origin FF extension uses (or did use at the time of this PR) an XML format that broke parsing.

p0deje:
Okay, but as [MDN](https://developer.mozilla.org/en-US/Add-ons/Install_Manifests) says, `<em:id>` must be present or addon may not install. I guess you should then report an issue to uBlock.

p0deje:
Closing this for now, please reopen if my understanding is wrong.

jmif:
em:id is in fact present in the xml file in uBlock rdf file and it's specified as valid XML.  The current code looks for an explicit em:id tag, not a "tag with a namespace of em and a name of id".  There are two ways to achieve em:id in valid XML:

1) Explicitly stating em:id as the tag name with a em namespace defined above.
2) Simpy using an id tag name with the default namespace set to em

The standard way most extensions do it looks like this (without default namespace):

`
<xml ns:em="em_namepsace">
  <em:id>ID_HERE</em:id>
</xml>
`

This is equivalent to using a default namespace and is perfectly valid XML/fits the MDN spec:

`
<xml ns="em_namepsaace">
   <id>ID_HERE</id>
</xml>
`

This may not be 100% syntactically correct (fixed this a while ago) but it illustrates the idea.  The current code base does not support the second, valid, way of specifying an ID in xml using default namespaces.

p0deje:
Okay, I see what you're saying. I'm looking through uBlock [install.rdf](https://github.com/gorhill/uBlock/blob/master/platform/firefox/install.rdf) and as far as I understand, we should actually look for `.//id`, not `.//em:id/id` because it's namespaced with `r`.

With that said, this PR doesn't solve the problem in general, so I'll keep it closed and push the proper fix shortly. Thanks for the issue and detailed explanations!

p0deje:
I've pushed commit 000621a386a058abd464c696ca4c72b7297687a9 which fixes the issue (though I referenced the wrong issue in commit).

PR #1145, opened 2015-10-13
Fix format string.

juangj:
Preconditions.checkArgument only accepts "%s", and not "%d" or any of the other placeholders.


ddavison:
fixed in 9c1b1e66c9326c8d7c38cc7fb8814877210a4388

PR #1164, opened 2015-10-16
 Add .NET bindings for WiniumDriver

tkurnosova:
WiniumDriver [https://github.com/2gis/Winium] is Selenium-based tools for testing of Windows apps.
This patch includes .NET bindings for WiniumDriver.

Signed-off-by: Tatyana Kurnosova tgkurnosova@gmail.com

### DIFF
--- a/dotnet/src/webdriver/WebDriver.csproj
+++ b/dotnet/src/webdriver/WebDriver.csproj
@@ -282,6 +282,12 @@
     <Compile Include="UnhandledAlertException.cs" />
     <Compile Include="WebDriverException.cs" />
     <Compile Include="WebDriverResult.cs" />
+    <Compile Include="Winium\IWiniumOptions.cs" />
+    <Compile Include="Winium\SilverlightOptions.cs" />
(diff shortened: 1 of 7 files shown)


jimevans:
Much like there are no Appium classes included in the Selenium project's .NET bindings, likewise, there should not be Winium classes included here as well. This sounds like code that should live in a companion project to Selenium (maybe the Winium project?), not in the Selenium project itself.

TikhomirovSergey:
@tkurnosova 
Maybe it makes sense to collaborate with the Appium. We've thought about WindowsMobile automation :)

tkurnosova:
When writing tests we use bindings selenium (OpenQA.Selenium for .net). This request allows to create instance of the driver not only as RemoteWebDriver.

jimevans:
And that's what I'm saying doesn't belong in the core Selenium bindings, anymore than an AppiumDriver belongs in the core Selenium .NET bindings.

tkurnosova:
Unfortunately, that's true. Thanks!

NickAb:
@TikhomirovSergey we are already working on establishing contact on this issue.

PR #1178, opened 2015-10-21
@FindBy(how = How.LINK_TEXT_PATTERN: useful when working with PFP

s2oBCN:
A new type of "findBy" that can be very useful almost when working
with page factory pattern and want to keep code as clean as possible.
And thinking that usually the links don't have an "id" and that maybe
one can't modify the source to add the 'id'.


jleyba:
We appreciate your efforts, but at this time we have no plans to extend the number of locator strategies provided in the core library.

PR #1034, opened 2015-09-14
Java. Event firing: improvements and new features

TikhomirovSergey:
## Problems
- Current design doesn't allow end user to log events optionally. E.g.
  if they want to log only searchings or searchings and navigation then
  they are forced to leave unimplemented methods in their code
- Current design doesn't allow to make any extension without backward
  compatibility problems because each new method should be implemented or
  left empty
- There are some issues with the logging of thrown exceptions
# What has been done
- WebDriverEvent listener was separated to few interfaces. It is still
  here and aggretes them (in order to support backward compatibility)
- New listeners were added
- EventFiringWebdraver was redesigned. The ability to use listeners
  optionally is provided now

Warning!!! There is an issue of backward compatility. Users will face it
if they directly use such classes as EventFiringKeyboard,
EventFiringMouse, EventFiringTouch. But I hope there not many cases and
it is not the critical issue.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/1034)

PR #1179, opened 2015-10-21
EventFiringWebDriver: added an event on "submit()"

s2oBCN:
When "recording" the interaction with a page,
there's the need to record the "submit" action.
For us, this is similar to a "click()",
so then we decided to dispatch the "click" event,
but maybe a new "submit" event could be dispatched...

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/1179)


java/client/src/org/openqa/selenium/support/events/EventFiringWebDriver.java
```diff
@@ -349,7 +349,9 @@ public void click() {
     }
 
     public void submit() {
+      dispatcher.beforeClickOn(element, driver);
```
lukeis: this shouldn't be beforeClickOn, but rather beforeSubmit? and added to WebDriverEventListener.

Since submit does not synthesize the 'click' events, this isn't accurate.

java/client/src/org/openqa/selenium/support/events/EventFiringWebDriver.java
```diff
@@ -349,7 +349,9 @@ public void click() {
...
       element.submit();
+      dispatcher.afterClickOn(element, driver);
```
lukeis: same here

TikhomirovSergey:
Guys
There is one more similar PR #1034 
Maybe does it make sense to combine these changes somehow?

beatngu13:
Anyone still interested in this? I would like to use this feature and I can also take care of necessary changes to this PR (new rebase, interface changes, more tests, etc.).

barancev:
Sorry for keeping this PR on hold for long time. We have implemented a new EventFiringDecorator that makes the proposed changes obsolete. See commits afdb348 and 47ee733

PR #1183, opened 2015-10-21
Selenium IDE - Webdriver Playback - Window per Test Case Fix

fredmon3:
While running a Selenese test suite using Webdriver Playback
(experimental) in the Selenium IDE a browser window per case was opened.
Webdriver Playback now keeps track of the lastSessionId used in the
previous case and continues to use that session for subsequent cases
preventing additional browser windows from opening.

P.S.  I realize this needs a refactor, I'm unsure of the best way to do it however.  My knee jerk reaction is to wrap the entire file in an iife to make the lastSessionId "private", export all of the "public" bits and potentially eliminate self.sessionId.

### DIFF
--- a/ide/main/src/content/webdriver-backed-selenium.js
+++ b/ide/main/src/content/webdriver-backed-selenium.js
@@ -10,6 +10,8 @@ function WebdriverBackedSelenium(baseUrl, webDriverBrowserString) {
   }
 }
 
+var lastSessionId;
+
 WebdriverBackedSelenium.prototype.reset = function() {
   //TODO destroy the current session and establish a new one?
 };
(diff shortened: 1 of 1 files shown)


ddavison:
each test run should ideally be in a new browser, to retain a fresh space for the test to run, unhindered by any potential cookies that may cause intermittent issues.

this code here, appears to open one browser, and reuse the same browser over and over again, with no configuration option *at least*.

samitbadle:
Hi Fred,

Thank you for your contribution. This issue has already been fixed and is undergoing some final testing. I will push the changes this week.

Cheers,
Samit Badle

fredmon3:
Sorry for any duplication of effort and I look forward to making additional contributions.  Thanks Samit!

PR #1208, opened 2015-10-31
Enable user to change Civetweb ACL from IEDriverServer command line

lsowen:


lsowen:
The most common use case that I imagine (and what I use it for) is to allow `IEDriverServer` to run in a local virtual machine and access it from the host machine without the need for the `RemoteWebDriver Server`.

In addition to making the usage somewhat easier, I had a weird corner case with a legacy application with a ton of iframes which only works in IE.  I'm able to automate testing when using `IEDriverServer` directly, but not when going through the `RemoteWebDriver Server`.  I tracked this down to the fact that going through `RemoteWebDriver` uses sizzle locators to find the iframes, whereas directly it doesn't.  (I wasn't able to find _why_ the sizzle locators were failing, because in easier test cases they worked perfectly.  I attribute it to some super weirdness with a poorly written legacy web application.)

lsowen:
Anything I can do to help this change move forward?

jimevans:
Having the patience to let me return from traveling and recover from illness would be a start. I'm not entirely sure I agree with the rationale for the change, so I'll need to thoroughly review it. Further (real-time) discussion on the matter would be best directed to the project IRC channel (#selenium on Freenode).

jimevans:
I'd like to discuss the rationale behind this pull request more thoroughly before merging it. As I stated before, I worry about the security implications of allowing this to be set in the server executable. Until I can discuss this further with you, I won't be merging the change in. Specifically, I'll need to know what wasn't working for you using the Java remote server, and what your WebDriver code in the language bindings of your choice looked like. I suspect I know what was happening, but I need to confirm it. Please feel free to connect with me via the project IRC channel to discuss it further.

lsowen:
Hi @jimevans,

I updated this PR per our conversation on IRC last week.  It now mimics the Chrome WebDriver behavior in the fact that the CLI argument is now a comma separated listed of IPs passed to `--whitelisted-ips`.

Please let me know if you see any more changes that need to be made.  Thanks for your help!

jimevans:
Merged in 008d9bf

Issue #1204, opened 2015-10-29
Selenium scrolls badly when trying to click on Firefox


germanp:
I've found that selenium scrolls badly when I try to click on a table's row and when to reach this row the browser must scroll down. When this happen, the browser scrolls one extra row past the expected one and it clicks in the wrong location.

I've found that this issue occurs after we upgraded to selenium version 2.4.80, using 2.47.3 this issue doesn't occurs, at least with firefox.

Greets, let me know if you need more info, tks

p0deje:
It might be related to #1202 

please provide a way to reproduce the issue, we can't do anything about it until that is provided. We can reopen the issue when it is provided.

Issue #1202, opened 2015-10-28
Click button does not work after upgrade to selenium 2.48.0


Tset-Noitamotua:
Doing test automation on a Ember.js web application:
I have a button which is covered by a `<div>` with a cooky notification. It looks something like this:

![image](https://cloud.githubusercontent.com/assets/5635462/10795869/fa412dfa-7d9b-11e5-9f84-2ecfc214e2e6.png)

With **Selenium 2.47.1** clicking on that button works as expected in Firefox.
With **Selenium 2.48.0** clicking this button does not work in Firefox, although there is no exception or error. It's like selenium thinks that the click worked properly but actually the click had no effect. Here the Robot Framework log:

![image](https://cloud.githubusercontent.com/assets/5635462/10796560/fd4b41ea-7d9e-11e5-89c3-d5e8309424fc.png)

The structure of HTML is something like this:

```html
<header>...</header>
<main>
  <div>
    ::before
    <form>
      <button>Last check</button>
    </form>
  ::after
  </div>
</main>
<div>
  <div class="cookie-notification">
    ::before
    <a href="http://...">Close</a>
    ::after
  </div>
</div>
```

Just for comparison - executing the same click with **Selenium 2.47.1** in Chrome results in a WebDriverException:

![image](https://cloud.githubusercontent.com/assets/5635462/10797758/6650ed48-7da4-11e5-80a6-82b7ac58de40.png)

```stacktrace
Traceback (most recent call last):
  File "<string>", line 2, in click_button
  File "C:\Python27\lib\site-packages\Selenium2Library\keywords\keywordgroup.py", line 15, in _run_on_failure_decorator
    return method(*args, **kwargs)
  File "C:\Python27\lib\site-packages\Selenium2Library\keywords\_formelement.py", line 316, in click_button
    element.click()
  File "C:\Python27\lib\site-packages\selenium\webdriver\remote\webelement.py", line 69, in click
    self._execute(Command.CLICK_ELEMENT)
  File "C:\Python27\lib\site-packages\selenium\webdriver\remote\webelement.py", line 448, in _execute
    return self._parent.execute(command, params)
  File "C:\Python27\lib\site-packages\Selenium2Library\webdrivermonkeypatches.py", line 11, in execute
    result = self._base_execute(driver_command, params)
  File "C:\Python27\lib\site-packages\selenium\webdriver\remote\webdriver.py", line 196, in execute
    self.error_handler.check_response(response)
[...]

Issue #1166, opened 2015-10-16
Click on link   (Selenium 2.48.2 , firefox 41.0.1 , Linuix)


vveliev:
I'm running to the issue that when I try to click on the link. It does not  do clicking.

We have single page application and after performing some action on the page,I try to click on link.

Not sure how to debug this problem.  will start with console output of selenium.  

```
12:09:37.261 INFO - Done: [find element: By.id: addButtonID]
12:09:37.261 DEBUG - Finished: POST /session/be5c8c0e-bc90-4db4-808d-8a26b63af13d/element
12:09:37.263 DEBUG - Handling: POST /session/be5c8c0e-bc90-4db4-808d-8a26b63af13d/element/346/click
12:09:37.264 INFO - Executing: [click: 346 [[FirefoxDriver: firefox on LINUX (062cf4ba-afa5-422b-98d2-e18cfbcb3f0b)] -> id: addButtonID]])
12:09:37.264 DEBUG - Executing: clickElement [062cf4ba-afa5-422b-98d2-e18cfbcb3f0b, clickElement {id={90a81d2c-2fff-4fa3-a4a2-8fbe5702b148}}]
12:09:37.264 DEBUG - CookieSpec selected: default
12:09:37.264 DEBUG - Auth cache not set in the context
12:09:37.265 DEBUG - Connection request: [route: {}->http://127.0.0.1:7055][total kept alive: 0; route allocated: 0 of 2000; total allocated: 0 of 2000]
12:09:37.265 DEBUG - Connection leased: [id: 1586][route: {}->http://127.0.0.1:7055][total kept alive: 0; route allocated: 1 of 2000; total allocated: 1 of 2000]
12:09:37.265 DEBUG - Opening connection {}->http://127.0.0.1:7055
12:09:37.265 DEBUG - Connecting to /127.0.0.1:7055
12:09:37.265 DEBUG - Connection established 127.0.0.1:58154<->127.0.0.1:7055
12:09:37.265 DEBUG - http-outgoing-1586: set socket timeout to 10800000
12:09:37.265 DEBUG - Executing request POST /hub/session/062cf4ba-afa5-422b-98d2-e18cfbcb3f0b/element/%7B90a81d2c-2fff-4fa3-a4a2-8fbe5702b148%7D/click HTTP/1.1
12:09:37.265 DEBUG - Target auth state: UNCHALLENGED
12:09:37.265 DEBUG - Proxy auth state: UNCHALLENGED
[...]

PR #1211, opened 2015-11-02
firefox: Throw error when element cannot be clicked

p0deje:
Current behavior is that when element is overlapped by another one,
trying to click it will silent fail because click is happening on
overlapping element. This commit changes the behavior so instead error
is thrown telling that overlapping element would receive click. It makes
the behavior mostly similar to ChromeDriver.

I've implemented this after it was [mentioned on IRC](https://botbot.me/freenode/selenium/2015-10-23/?msg=52586420&page=3) that this behavior might make sense and after a bunch of issues were reported (#1202, #1204, #1166).

That is my first time I deal with FirefoxDriver, so the code may not be the best. Also, there is now some duplication with `webdriver.chrome.isElementClickable` function so maybe it makes sense to DRY it up. Also, since I don't have much experience with Java/JavaScript testing of Selenium, it was easier for me to add regression test in Ruby.

@barancev @lukeis Please, take a look.

### DIFF
--- a/javascript/firefox-driver/js/syntheticMouse.js
+++ b/javascript/firefox-driver/js/syntheticMouse.js
@@ -77,11 +77,62 @@ SyntheticMouse.newResponse = function(status, message) {
 };
 
 
+SyntheticMouse.prototype.isElementShownAndClickable = function(element) {
+  var error = this.isElementShown(element);
+  if (error) {
+    return error;
+  }
+
+  var error = this.isElementClickable(element);
+  if (error) {
+    return error;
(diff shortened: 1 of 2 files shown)


barancev:
Merged in a2401418b8f73753f3fc9a7ef666be715177b6a5

vveliev:
Is exeption expected when all you need to do is scroll the page. such that element becomes visible ?!

barancev:
@vveliev Yes, Selenium scrolls the target element to the view port.

PR #1217, opened 2015-11-03
Treat a null response to getCookies as [].

juangj:
The Edge driver returns null (see
https://connect.microsoft.com/IE/Feedback/Details/1872177), which is
incorrect, but we might as well avoid the NPE in the client until that
gets fixed.


titusfortner:
Oh good, this saves me from having to do a better job explaining/reproducing the bug I filed yesterday - https://connect.microsoft.com/IE/feedback/details/1970246/ :)

PR #1246, opened 2015-11-12
Python 3.2+ instead of listing specific Python 3.x versions

nikolas:
Python 3.4 is listed as being supported by Selenium in setup.py:
https://github.com/SeleniumHQ/selenium/blob/master/setup.py

### DIFF
--- a/py/README
+++ b/py/README
@@ -27,7 +27,7 @@ Supported Python Versions
 =========================
 
 * Python 2.6, 2.7
-* Python 3.2, 3.3, 3.4
+* Python 3.2+
 
 Installing
 ==========


ivargas-handy:
Can you test Python 3.5 as well, and add that?

cgoldberg:
Perhaps it should just read: "Python 3.2+", rather than listing specific versions.  I don't believe the project's CI actually runs all those versions anyway.  

Also... I haven't tested it exhaustively, but CPython 3.5 should work fine.

nikolas:
@cgoldberg good idea. I've added a new commit to this branch.

lukeis:
merged in 4b14c304934f8559ef6c755ffa39caed1d685d5c

PR #1258, opened 2015-11-16
Add a check for null results from executing JavaScript

### LINKED ISSUE #1257: ExecuteJavaScript always throws when JavaScript returns null
*This is an issue in the WebDriver.Support library for .NET.*

When using the ```OpenQA.Selenium.Support.Extensions.WebDriverExtensions.ExecuteJavaScript``` extension method and the script invoked returns ```null```, then the method ```IsInstaceOfType``` (WebDriverExtensions.cs, line 91) will return ```false``` (```null``` isn't an instance of any type). This causes the method to throw a ```WebDriverException```, rather than returning ```null``` to the caller.

chrisblock:
The System.Type.IsInstanceOfType method will always return false when
handed a null value. This change checks explicitly for null to prevent an
exception from being thrown in this case. A check for a value type is also
made to avoid attemping to convert to a value type from a null value.

Fixes #1257

### DIFF
--- a/dotnet/src/support/Extensions/WebDriverExtensions.cs
+++ b/dotnet/src/support/Extensions/WebDriverExtensions.cs
@@ -16,11 +16,8 @@
 // limitations under the License.
 // </copyright>
 
-using System;
-using System.Collections.Generic;
-using System.Linq;
 using System.Reflection;
-using System.Text;
+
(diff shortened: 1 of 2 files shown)


jimevans:
Merged in 2a28c18

PR #1291, opened 2015-11-20
py: phantomjs: make sure the spawned phantomjs process is killed

dryice:
Currently with many

> wd = webdriver.PhantomJS()
> wd.quit()

calls, there will be many spawned phantomjs process remained, and
they'll use up the RAM.

This issue has been described at
https://code.google.com/p/selenium/issues/detail?id=5921

and has been fixed at
https://code.google.com/p/selenium/source/detail?r=885b62bbb7ee

However dbac7447d803a6efd33a031df841fa5f51e70ce5 and
1bc9c24d67b30ae1ebb7be7debdbfd7c8fd93f3e removed the fix.

This commit brings the fix back

### DIFF
--- a/py/selenium/webdriver/phantomjs/service.py
+++ b/py/selenium/webdriver/phantomjs/service.py
@@ -17,7 +17,6 @@
 import platform
 import subprocess
 import time
-import signal
 
 from selenium.common.exceptions import WebDriverException
 from selenium.webdriver.common import utils
@@ -105,11 +104,10 @@ class Service(object):
         try:
             if self.process:
                 self.process.stdin.close()
-                if platform.system() == 'Windows':
-                    self.process.kill()
-                else:
-                    self.process.send_signal(signal.SIGTERM)
+                self.process.terminate()
                 self.process.wait()
+                if platform.system() != 'Windows':
+                    self.process.kill()
                 self.process = None
         except OSError:
             # kill may not be available under windows environment


dryice:
PS: I've signed the CLA.

lukeis:
Did you test this on Windows? Because I think that's where the problem with doing this is.

dryice:
No. I don't have access to a Windows machine to test. But I've just made a commit that keep the original kill() call on Windows, and send SIGTERM on Posix OSs.

cgoldberg:
this has been changed many times, so let's get it right :)

first... you can call `process.terminate()`, so you don't have to import signal module.  It sends the same SIGTERM.

and the proper way to stop and kill the process:

on Posix, it should call these:
`terminate()`
`wait()`
`kill()`

on Windows, `kill` is a alias for `terminate()`

see: https://docs.python.org/2/library/subprocess.html#popen-objects

dryice:
Yes. just changing kill() to terminate() in the original code could get the result "send SIGTERM on Posix, and call TerminateProcess() on W32"

And on Posix, we can do an extra kill() to make sure.

lukeis:
i've refactored the code so that all the different drivers that require an executable use a common service, this fix. The use of terminate is included (I still need to do some testing with it). But I'm going to close this PR.

0ee1c15d7dd98344cd2e35e7a4bfa92561e4a686

dryice:
Thanks for resolving this once and for all!

BTW: I agree with @cgoldberg to do
terminate()
wait()
kill()

It looks the new code is doing
terminate()
kill()
wait()
that didn't give the executable enough time to terminate gracefully.

menesis:
The last comment is still valid in version 2.53.0 – `kill()` right after `terminate()` leaves a stray process (if phantomjs is invoked via node).

PR #1317, opened 2015-11-25
Enhance check for value types to exclude Nullable

### LINKED ISSUE #1316: ExecuteJavaScript does not handle Nullable<T> very well
*This is an issue in the WebDriver.Support library for .NET.*

Almost immediately after my pull request last time (#1258), I realized that the ```Nullable<T>``` type is a struct, and thus also a value type. One of the checks I added via the pull request will throw when any value type is used. This check should be modified to exclude the ```Nullable<T>``` type, since it does support null values.

chrisblock:
The System.Nullable type is a value type that can be assigned
the null value. This change checks for types wrapped with
Nullable and does not throw when they are encountered.

Fixes #1316


jimevans:
Merged in 5a04a82

PR #1319, opened 2015-11-26
RemoteWebDriver#getCookies() ignores httpOnly returned by driver

lqc:
Firefox, Chrome 2.19+ and IEDriverServer 2.47+ all return the "httpOnly" value, but it's ignored by Java bindings when constructing Cookie object.


barancev:
Landed. This required bit more work to implement httpOnly support in RemoteWebDriver and HtmlUnitDriver too.

PR #1323, opened 2015-11-28
IDEA config: mark common/src/web/ as a Test Resources Root

alb-i986:
Mark `common/src/web/` as a Test Resources Root.

My purpose was in particular to have search results grouped correctly, when the switch "Group by test/production" is active. Please see screenshot.

screen shot

Here I'm searching for a text that happens to be contained in a file under `common/src/web/`: `pageWithOnBeforeUnloadMessage.html`. As you can see, it looks like it is production code.


lukeis:
merged with e6f01da18c7561e6560c2d2a6ba6cd3d3c1e306a

PR #1327, opened 2015-11-30
Fix a NullPointerException in RemoteWebDriver.getCurrentUrl

ekuefler:
execute() is allowed to return null (see line 643), which will cause a
NullPointerException to be thrown in getCurrentUrl. This doesn't fix the
underlying problem, but it causes it to throw a WebDriverException
instead of a NullPointerException so the user can handle it in a
reasonable way.

### DIFF
--- a/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java
+++ b/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java
@@ -323,7 +323,12 @@ public class RemoteWebDriver implements WebDriver, JavascriptExecutor,
   }
 
   public String getCurrentUrl() {
-    return execute(DriverCommand.GET_CURRENT_URL).getValue().toString();
+    Response response = execute(DriverCommand.GET_CURRENT_URL);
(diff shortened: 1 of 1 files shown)


java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java
```diff
@@ -323,7 +323,12 @@ public String getTitle() {
   }
 
   public String getCurrentUrl() {
-    return execute(DriverCommand.GET_CURRENT_URL).getValue().toString();
+    Response response = execute(DriverCommand.GET_CURRENT_URL);
+    if (response == null) {
```
lukeis: can you also add  || response.getValue() != null

lukeis:
also can you confirm you've signed the [CLA](http://goo.gl/qC50R)

ekuefler:
Addressed comment. Yes, I've signed the CLA.

lukeis:
merged as d94c64f547dc09e0942506efddc394a728c11999

PR #1328, opened 2015-12-01
refactor concrete WebDriverException#getSupportUrl's

alb-i986:
extract urls to constants

### DIFF
--- a/java/client/src/org/openqa/selenium/InvalidSelectorException.java
+++ b/java/client/src/org/openqa/selenium/InvalidSelectorException.java
@@ -27,6 +27,9 @@ package org.openqa.selenium;
  */
 public class InvalidSelectorException extends NoSuchElementException {
 
+  private static final String SUPPORT_URL =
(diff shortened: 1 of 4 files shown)


ddavison:
I like this change, but i'm not sure I like the access modifier you put on there.  I think it could be handy to have the ability to call `InvalidSelectorException.SUPPORT_URL` from other places as well.  change this to public, unless you have a valid reason for your choice :)

ddavison:
disregard... forgot that there was a getter :)

alb-i986:
BASE_SUPPORT_URL could be split into "http://seleniumhq.org" + "/exceptions" and the first token could be made public, and moved to some other class: other code might need it. E.g. now I see that DefaultSelenium could benefit of that.

ddavison:
I was also thinking initially of an enumeration, but i think that the url's are fine contained in their respective classes

i'll merge soon

ddavison:
committed in 22a21b4f90f08be938d844a5e1496ef406c217fd.  thanks! :+1:

PR #1339, opened 2015-12-03
rb - add SessionNotCreatedError

bayandin:
I actually don't know why it was `nil` before. There were no comments about it.

But now instead of `Uncaught exception: status code 500` we can get clear error message from server:  `Uncaught exception: Could not determine your device from Appium arguments or desired capabilities. Please make sure to specify the 'deviceName' and 'platformName' capabilities` (this example comes from Appium)

### DIFF
--- a/rb/lib/selenium/webdriver/common/error.rb
+++ b/rb/lib/selenium/webdriver/common/error.rb
@@ -159,7 +159,12 @@ module Selenium
       #
 
       class InvalidSelectorError < WebDriverError; end           # 32
-                                                                 # 33
+
+      #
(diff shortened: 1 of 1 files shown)


titusfortner:
I think this was just never added. Looks like Jari [added a few that weren't implemented]( 555b622) 4 years ago, so I'm not sure why he didn't add this one as well. [The documentation for response status](https://code.google.com/p/selenium/wiki/JsonWireProtocol#Messages) indicates that it should be: 'SessionNotCreatedException.' Since Error is obviously more correct for Ruby, though, I agree with this implementation.

PR #1346, opened 2015-12-04
Allow to reuse sessions

a-chernykh:
It's often necessary to reuse already running session. For instance,
when you don't want to wait for browser to open. Example:

```
@browser = Selenium::WebDriver.for :remote, url: 'http://dockerhost.dev:4444/wd/hub', session_id: '46fa1494-3b69-4dd0-bfdf-915aa28a8d71'
```

### DIFF
--- a/rb/lib/selenium/webdriver/remote/bridge.rb
+++ b/rb/lib/selenium/webdriver/remote/bridge.rb
@@ -65,6 +65,7 @@ module Selenium
           http_client          = opts.delete(:http_client) { Http::Default.new }
           desired_capabilities = opts.delete(:desired_capabilities) { Capabilities.firefox }
           url                  = opts.delete(:url) { "http://#{Platform.localhost}:4444/wd/hub" }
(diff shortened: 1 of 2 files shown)


titusfortner:
What is the use case? I'm not sure I understand when you would have access to a session_id variable, without also having access to the browser variable that already contains it.

andreychernih:
The use case is to save some time on browser startup. I have docker containers which are running user's cucumber tests. A user is expecting to see the result of such tests almost immediately. Every new test which I am running does not have an access to browser instance of previous test (because it's a new process), that's why I need to be able to re-use session id.

titusfortner:
This is a bad idea in general. Your tests should be independent, and should always start in a known state. This feature goes against both principles, so we won't be merging it.

pannous:
Selenium is not just a test tool, rather an API to automate browsers;
Please reopen.

titusfortner:
@pannous - if you look at #3101 I've specified the solution that I'll be implementing for this.

PR #1369, opened 2015-12-11
Fix ChromeOptions passing

rmarx:
Just passing "this" will serialize the ChromeOptions object as-is. Since the var names have underscore suffixes, those are serialized as well, and none of the options are actually accepted by ChromeDriver (silently ignored). 

i.e. current code leads to serialization:
"chromeOptions":{"options_":{ "detach" : true },"extensions_":[],"logPrefs_":{"performance":"ALL"},"proxy_":null}

while we actually want this behaviour:
"chromeOptions":{ "detach" : true }

Since extensions, logPrefs and proxy are added to DesiredCapabilitites and not ChromeOptions separately.


jleyba:
Using `this` is correct usage. If it's not being serialized properly, there's a bug elsewhere in the code

PR #1388, opened 2015-12-17
Resolve inconsistency in Webdriver.__init__ desired_capabilities parameter

drobota:


lukeis:
This is a breaking API change. We should still accept the old 'capabilities' keyword arg. I would only be ok removing it in a major release (like 3.0)

drobota:
@lukeis  I am totally agree with that however this needs to be dome somewhen.

AutomatedTester:
And that can be done when we move to Selenium 3 (date to be determined)

David Burns
Email: david.burns@theautomatedtester.co.uk
URL: http://www.theautomatedtester.co.uk/

drobota:
I've added that in https://github.com/robotframework/Selenium2Library/pull/552

Dude-X:
@drobota would be nice to revisit this with the latest revision and see if your patch is needed.

drobota:
Inconsistency is still exist different drivers have different constructor variables 
f.e. firefox has - [capabilities](https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/firefox/webdriver.py#L52)
when safari  - [desired_capabilities](https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/safari/webdriver.py#L36)
In my view it has to be consistent with remote web driver class which has [desired_capabilities](https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/webdriver.py#L59)

lmtierney:
I know it's been a while... but would be good to make them consistent for the `4.0` release ;)

lmtierney:
This PR is outdated and obsolete as #5513 has been merged

PR #1427, opened 2015-12-28
Added CharSequence[] param in *ChangeValueOf event + test

Kuhtich:


Kuhtich:
Signed a Contributor License Agreement as Roman (kuhtich@gmail.com)

java/client/src/org/openqa/selenium/support/events/AbstractWebDriverEventListener.java
```diff
@@ -67,11 +67,11 @@ public void afterClickOn(WebElement element, WebDriver driver) {
     // Do nothing.
   }
 
-  public void beforeChangeValueOf(WebElement element, WebDriver driver) {
+  public void beforeChangeValueOf(WebElement element, WebDriver driver, CharSequence[] value) {
```
lukeis: change 'value' to keysToSend, here and in other references in this commit.  'value' implies the user is getting the value of the element, rather than just the keys we're sending to it.

Kuhtich: done

lukeis:
merged as 969f586385c84aeed99447634118b9e81fd82348

thanks :) and sorry it took me so long to come back to this one.

meeroslaph:
@Kuhtich, @lukeis Hey guys, what is this new parameter `CharSequence[] keysToSend` used for? Unfortunately no info present in this pull request, nor it is available in the Javadoc:
https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/support/events/WebDriverEventListener.java

From the quick overview, it looks like the value of `keysToSend` is going to be used for logging purposes. Is that correct?

PR #1442, opened 2016-01-05
fix mistake in description of grid JSON timeouts

azhurbilo:


malikdavid87:
I think its not a mistake.
If you look at GridHubConfiguration.java at this line:
"setBrowserTimeout(Integer.parseInt(helper.getParamValue("-browserTimeout")) * 1000);"

you can see that the browserTimeout value is indeed in seconds.

azhurbilo:
@malikdavid87 in your case it's parsing of console arguments, and yes it use seconds.
But in JSON timeouts in ms. 

```
if (o.has(RegistrationRequest.BROWSER_TIME_OUT)
          && !o.get(RegistrationRequest.BROWSER_TIME_OUT).isJsonNull()) {
        setBrowserTimeout(o.get(RegistrationRequest.BROWSER_TIME_OUT).getAsInt());
      }
```

barancev:
Command line options accept values in seconds (unlike JSON config file).

PR #1486, opened 2016-01-17
Allow GeckoDriverService to be extended.

joshbruning:
Allow GeckoDriverService to be extended.

Note that this is the case for other driver services.


lukeis:
merged as 38a174cba3821c483ef3044f5b0f9c0484678e1a

PR #1487, opened 2016-01-17
Allow use of JUnit filters.

joshbruning:
Allow use of JUnit filters.

This change allows the JUnit Filter API to work correctly, and
will fall back to it being broken if some version of JUnit4's
Suite/ParentRunner hierarchy is incompatible with it. I've
verified this should work with at least JUnit 4.10, 4.11, and
4.12.

### DIFF
--- a/java/client/test/org/openqa/selenium/testing/ParallelSuite.java
+++ b/java/client/test/org/openqa/selenium/testing/ParallelSuite.java
@@ -17,17 +17,21 @@
 
 package org.openqa.selenium.testing;
 
+import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.ConcurrentLinkedQueue;
 
 import org.junit.runner.Runner;
 import org.junit.runner.notification.RunNotifier;
+import org.junit.runners.ParentRunner;
 import org.junit.runners.Suite;
 import org.junit.runners.model.InitializationError;
 import org.junit.runners.model.RunnerBuilder;
 
 import com.google.common.base.Preconditions;
+import com.google.common.base.Throwables;
 
 public class ParallelSuite extends Suite {
   public static final String PARALLELIZATION_ENV_VAR_NAME = "PARALLEL_DRIVER_COUNT";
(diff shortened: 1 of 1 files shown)


barancev:
What's the purpose of this change? We don't use filtering. Well, actually we do filtering, but an alternative flavor of it: https://github.com/joshbruning/selenium/blob/patch-2/java/client/test/org/openqa/selenium/testing/SeleniumTestRunner.java#L80-L84

joshbruning:
This is a bug fix, not a new feature.

bazel defines --test_filter and --test_arg (see http://bazel.io/docs/bazel-user-manual.html)
(buck defines ----test-selectors and -filter). These allow developers (me) to filter (for example a single test method) easily per invocation.

I'm not familiar with buck's implementation; however, bazel's is here: https://github.com/bazelbuild/bazel/blob/master/src/java_tools/junitrunner/java/com/google/testing/junit/runner/junit4/JUnit4Runner.java

You can see how it uses org.junit.runner.manipulation.Filter for increased parallelization (sharding) as well as for including/excluding tests.

I would be quite surprised if buck is implemented using something other than the standard JUnit mechanism.

Suite and ParentRunner both support Filter correctly; however, Selenium's ParallelSuite does not. ParentRunner.run is not really intended to be overridden (without a super call) as it calls key private methods; it is public so that it can be invoked. ParentRunner implements Filterable (by extending Suite which extends ParentRunner), and it should honor any calls to its "public void filter(Filter filter)".

barancev:
An attempt to apply this PR caused an exception:
```
java.util.Collections$UnmodifiableCollection cannot be cast to java.util.List
java.lang.ClassCastException: java.util.Collections$UnmodifiableCollection cannot be cast to java.util.List
	at org.openqa.selenium.testing.ParallelSuite.getFilteredChildren(ParallelSuite.java:75)
	at org.openqa.selenium.testing.ParallelSuite.run(ParallelSuite.java:55)
```

joshbruning:
Thanks - I was wrong about 4.12. I'll update the code to support that version as well.

joshbruning:
This should be ready for review again.

joshbruning:
Hey, I squashed my commits and sync'd up to head. Would appreciate if someone would take a look so we can get this in, or I can just abandon it and update the Google-fork automation we have to ignore this change locally.

shs96c:
After discussion on the IRC channel, it sounds like we don't need the ParallelSuite at all. I've deleted it, rendering this PR unnecessary.

PR #1507, opened 2016-01-19
Parse message when error response is dictionary

### LINKED ISSUE #1505: [Python] WebDriverException message is not returned in Chrome after upgrade to 2.49.1
Environment: chromedriver 2.20 // Python 2.7.11 // Windows 7

Since upgrading Selenium to 2.49.1, we're seeing an issue when WebDriverExceptions (or items which inherit it) are raised, the exception message is being returned as an empty string. I'm also seeing the issue in 2.49.2

The behavior we're seeing is the response is being passed to check_response in errorhandler.py as a dictionary in Chrome, but (for example) in Firefox, it's being passed as a a string (so it hits the json parsing block). Chrome does not return a message, Firefox does.

Example to Reproduce

    driver = webdriver.Firefox()
    driver.get("http://www.python.org")
    elem = driver.find_element_by_name("NameDoesNotExist")

Exception on Firefox

    elenium.common.exceptions.NoSuchElementException: Message: Unable to locate element: {"method":"name","selector":"NameDoesNotExist"}

Repeat code, but with Chrome: Exception on Chrome:

    selenium.common.exceptions.NoSuchElementException: Message:

adamreyher:
Resolves issue in Python version 2.49.1 where the response error
message was being sent to the exception as an empty string if
the response was passed to the error handler as a dictionary.

Fixes #1505

### DIFF
--- a/py/selenium/webdriver/remote/errorhandler.py
+++ b/py/selenium/webdriver/remote/errorhandler.py
@@ -157,7 +157,7 @@ class ErrorHandler(object):
             if exception_class == ErrorInResponseException:
                 raise exception_class(response, value)
             raise exception_class(value)
-        if message != "" and 'message' in value:
+        if message == "" and 'message' in value:
             message = value['message']
 
         screen = None


barancev:
Fixed by commit ab99cb6

PR #1520, opened 2016-01-22
Check element corners and middle for clickability

twalpole:
There has been a lot of discussion in Issue #1202 about the recent changes to click behavior in FF.  IMHO the new behavior makes more sense from a user emulation POV, however it does lack the ability to click on partially covered elements which a user definitely could.  This PR checks the elements four corners as well as the middle for clickability and therefore allows a click on an element that is partially covered.

It also fixes some CSS in an existing ruby test

CLA has been signed

### DIFF
--- a/javascript/firefox-driver/js/utils.js
+++ b/javascript/firefox-driver/js/utils.js
@@ -653,19 +653,24 @@ Utils.getClickablePoint = function(element) {
   element = element.wrappedJSObject ? element.wrappedJSObject : element;
   var rect = bot.dom.getClientRect(element);
 
-  if (element.getClientRects().length > 1) {
-    for (var i = 0; i < element.getClientRects().length; i++) {
-      var candidate = element.getClientRects()[i];
-      if (candidate.width != 0 && candidate.height != 0) {
-        return findClickablePointOrMiddle(candidate);
+  var rects = goog.array.filter(element.getClientRects(), function(r) {
+    return r.width !=0 && r.height != 0;
+  });
+
+  if (rects.length > 0) {
+    for (var i = 0; i < rects.length; i++) {
+      var candidate = rects[i];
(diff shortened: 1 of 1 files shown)


barancev:
Merged (and fixed)

jochenberger:
This seems to cause issues with elements that have a `border-radius`. Trying to click the 0,0 corner of a button with a `border-radius` will click the parent element instead.

merike:
Could you please clarify which version this was released in or will be released in? I'm having an issue with inline two-line left-aligned links where the text on first line is very short and thus the center of the link block is actually outside link element and thus fails to click. This issue sounds like it might fix it without forcing a click on 1, 1 coordinates..

barancev:
The fix should be released in version 2.50

merike:
Ok, in this case it didn't fix my issue as 2.50.1 fails the same without 1, 1 coordinates.

barancev:
To click at an arbitrary point use Actions. We can't check every point of the element :)

twalpole:
@jochenberger This change still tries for a click in the middle of the element, and only if that isn't clickable does it try corners, so it's no worse than the behavior was previously but yes, if the element is not clickable in the middle and has border-radius on corners that would be clickable if the border-radius wasn't there it won't click in the corner - I guess it could be expanded to try the middles of each side of the client rects as well as the corners (or instead of) but there has to be a limit at some point of how many points can be tried

jochenberger:
Sure, i didn't mean to imply that the new behavior was worse. I think there would be much less of a hassle if #1533 were fixed (including scrolling to an element on a page with a fixed header/footer).

phoenix384:
I'm also still having problems with elements where the middle is not clickable. Clicking the corners sometimes seem to miss the element.

Issue #1437, opened 2015-12-30
Firefox (installed by the Mac's homebrew) has a different path


ph-hs:
~/Applications/Firefox.app/

That should be check in addition to /Applications/Firefox.app/ in my opinion. As a fallback obviously.

> # to install
> 
> brew cask install firefox

lukeis:
@ph-hs we could possibly add this fallback... we'd need to add it to each of the language bindings too, any in particular you cared about? (java, python, ruby, javascript, .NET)

ph-hs:
I checked out the source, and started working on the Java one (tried to break a test first), but CrazyFun was hanging as it went through the tests :-(  My intention was to work it though for all applicable clients. 

In answer to your Q, Java is most important to me.

lukeis:
added to java and python:  21bbdf51b8b1ed35926887f51933a5f6256beaa5

I'll leave this open for ruby, .NET, javascript.

lukeis:
ruby apparently by default explores all of someone's PATH by default... so it just needs to be set there for ruby people

ph-hs:
Ruby on Mac still needs to do ~/Applications/Firefox.app/Contents/MacOS/firefox and /Applications/Firefox.app/Contents/MacOS/firefox as not everyone sets their path:

    Pauls-MacBook-Pro:Foo phammant$ firefox 
    -bash: firefox: command not found

lukeis:
@titusfortner see previous comment :)

titusfortner:
Honestly, I'm reluctant to add code that is only used by an application that recognizes that [it is doing it wrong](https://github.com/caskroom/homebrew-cask/issues/13201) and plans to change. 
Symlinks should work, or ```Selenium::WebDriver::Firefox::Binary.path = '~/Applications/Firefox.app/Contents/MacOS/firefox'```
That being said, it is only one line of code, and [you are not the only one](http://stackoverflow.com/questions/28130038/selenium-webdriver-cant-find-firefox-when-installed-with-homebrew-cask).

titusfortner:
This should do it for Ruby: https://github.com/SeleniumHQ/selenium/commit/b76925dc98a86e885aac243e763a35aea846e593

jleyba:
By default, the JS bindings will check the default installation directory for each platform. Users that install Firefox elsewhere must specify that location through the API. This is an intentional design decision.

ph-hs:
Well if their going to change to /Applications, then fine.

[...]

PR #1523, opened 2016-01-25
Expand paths for correct `File.exist?` calls

nixme:
Related #1437

### DIFF
--- a/rb/lib/selenium/webdriver/firefox/binary.rb
+++ b/rb/lib/selenium/webdriver/firefox/binary.rb
@@ -170,8 +170,8 @@ module Selenium
 
           def macosx_path
             path = "/Applications/Firefox.app/Contents/MacOS/firefox-bin"
-            path = "~/Applications/Firefox.app/Contents/MacOS/firefox-bin" unless File.exist?(path)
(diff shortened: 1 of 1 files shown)


rb/lib/selenium/webdriver/firefox/binary.rb
```diff
@@ -170,8 +170,8 @@ def windows_path
 
           def macosx_path
             path = "/Applications/Firefox.app/Contents/MacOS/firefox-bin"
-            path = "~/Applications/Firefox.app/Contents/MacOS/firefox-bin" unless File.exist?(path)
-            path = Platform.find_binary("firefox-bin") unless File.exist?(path)
+            path = "~/Applications/Firefox.app/Contents/MacOS/firefox-bin" unless File.exist?(File.expand_path(path))
```
titusfortner: I don't think this one needs to be changed since the only previous path is already an absolute one.

titusfortner:
Merged, thanks - 709ddbcb183f4304477ae8e7fd922bc3207878d3
It ended up only being a one line, but please sign the CLA if you haven't already.

PR #1525, opened 2016-01-25
Display seconds in config page for timeouts

### LINKED ISSUE #1064: Timeout settings are confusing and inconsistent
Hi, I need to extend the default timeout (as my test needs to sleep for a while) and I run into these two confusing behaviors of the Selenium server:

1. The hub settings
====
The docs say, the `-timeout` and `-browserTimeout` values should be in seconds:
> -timeout <nnnn>: an integer number of seconds we should allow a client to be idle

If the hub is started with timeout value provided: `java -jar selenium-server-standalone-2.47.1.jar -role hub -timeout 900` the console prints: `timeout : 900000`. So it look like the internal representation is in ms. I think this (not the internal representation, but the config output and "config input" (= CLI params) should be consistent to avoid confusion about unit of the value.

2. The node settings
====
Also for the node, according to the help, the `-timeout` and `-browserTimeout` values should be in seconds.
So I start the node with command `java -jar selenium-server-standalone-2.47.1.jar -role node -timeout 900`. But not when I check in the grid console the Configuration tab for the node, there is printed `timeout:900`. So is this ms or secs now? 

Unlike the hub settings, it looks like the input value is actually in miliseconds here - when tested, the firefox browser window created by node run with this param (`-timeout 900`) is closed in less than few seconds (`WARN - session ext. key ... has TIMED OUT due to client inactivity and will be released.`], however, using miliseconds value (`-timeout 900000`) makes the node holding for 15 minutes. I also find few questions on Stackoverflow from people trying to use the seconds value on the node and reporting it not to be preventing the browser from timeout (which could simply be caused by the wrong unit used - secs/ms).

So the settings is inconsistent amongst hub, node and the documentation.

semonte:
Configuration page displays timeouts in milliseconds without
indicating units. Since -timeout and -browserTimeout are given
in seconds, it makes sense to display these values in seconds in
configuration page.

This pull requests includes code that displays timeout and
browserTimeout paramaters as seconds in the configuration page
in the following format:

timeout : 10s
browserTimeout : 10s

Fixes #1064 partially ("The hub settings" part).

### DIFF
--- a/java/server/src/org/openqa/grid/web/servlet/beta/ConfigPrinter.java
+++ b/java/server/src/org/openqa/grid/web/servlet/beta/ConfigPrinter.java
@@ -17,25 +17,36 @@
 
 package org.openqa.grid.web.servlet.beta;
 
+import com.google.common.collect.ImmutableSet;
+
 import org.openqa.grid.common.RegistrationRequest;
 
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 public class ConfigPrinter {
 
-  public static String printConfigValue(String configKey, String configValue) {
-    return configValue != null && configKeyShouldBePrintedInSeconds(configKey)
-           ? printSecondsFromMillis(Integer.valueOf(configValue))
-           : configValue;
+  private static final Set<String> CONFIG_KEYS_TO_BE_PRINTED_IN_SECONDS =
+    ImmutableSet.of(RegistrationRequest.BROWSER_TIME_OUT,
+                    RegistrationRequest.TIME_OUT);
+
+
+  public static String printConfigValue(String configKey, Object configValue) {
+    if (configValue == null) {
+      return null;
+    }
+
+    return configKeyShouldBePrintedInSeconds(configKey)
+           ? printSecondsFromMillis(Long.valueOf(String.valueOf(configValue)))
+           : String.valueOf(configValue);
   }
 
   private static boolean configKeyShouldBePrintedInSeconds(String configKey) {
-    return RegistrationRequest.BROWSER_TIME_OUT.equals(configKey)
-           || RegistrationRequest.TIME_OUT.equals(configKey);
+    return CONFIG_KEYS_TO_BE_PRINTED_IN_SECONDS.contains(configKey);
   }
 
-  private static String printSecondsFromMillis(int millis) {
-    return String.format("%ds", (int) TimeUnit.MILLISECONDS.toSeconds((long) millis));
+  private static String printSecondsFromMillis(long millis) {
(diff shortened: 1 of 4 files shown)


java/server/src/org/openqa/grid/web/servlet/beta/ConfigPrinter.java
```diff
@@ -0,0 +1,41 @@
...
+
+public class ConfigPrinter {
+
+  public static String printConfigValue(String configKey, String configValue) {
+    return configValue != null && configKeyShouldBePrintedInSeconds(configKey)
+           ? printSecondsFromMillis(Integer.valueOf(configValue))
+           : configValue;
+  }
+
+  private static boolean configKeyShouldBePrintedInSeconds(String configKey) {
```
lukeis: how about we keep a private Set and check if it contains the value instead? easier to add more keys later if need be.

java/server/src/org/openqa/grid/web/servlet/beta/ConfigPrinter.java
```diff
@@ -0,0 +1,41 @@
...
+
+package org.openqa.grid.web.servlet.beta;
+
+import org.openqa.grid.common.RegistrationRequest;
+
+import java.util.concurrent.TimeUnit;
+
+public class ConfigPrinter {
+
+  public static String printConfigValue(String configKey, String configValue) {
```
lukeis: instead of forcing everything to a string and then back to an int, let's change the method signature to be Object configValue and then check if the Object is an integer.

java/server/src/org/openqa/grid/web/servlet/beta/ConfigPrinter.java
semonte: Good idea, I will create a set for the keys.

java/server/src/org/openqa/grid/web/servlet/beta/ConfigPrinter.java
semonte: I have simplified the casting, caller can now pass in an object.

lukeis:
since i refactored how grid stores and loads config, you'll need to rebase your changes and update the code accordingly.

Also I see some Long.valueOf(String.valueOf(object)));  code... can you change that to check if the object is an instance of Long and then just cast? I dislike converting an object's type unnecessarily.

semonte:
This PR is no longer valid, since the config page currently prints the values as they were given in startup. For example, starting hub with "-browserTimeout 2 -timeout 5" will print the following:

browserTimeout : 2
timeout : 5

This is expected behaviour.

lukeis:
Ok, thanks for checking!

PR #1528, opened 2016-01-25
Fix keyCodes for some characters in the FirefoxDriver.

DrMarcII:
Original Google bug report:
"We started seeing failures on Firefox.  The code does sendKeys("# hello world"), but for some strange reason the hash seems to have been lost in the input."

On further investigation we identified issues with a variety of additional characters.


juangj:
Merged (83984c0e8e3fd59a96a0f9584d9a6cc6663be830)

Issue #1300, opened 2015-11-21
Python Firefox webdriver can fail on slow systems


jacekn:
When I try to load Firefox webdriver in python using:
`webdriver.Firefox()`

It occasionally fails after around 30 seconds with below traceback:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/firefox/webdriver.py", line 77, in __init__
    self.binary, timeout),
  File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/firefox/extension_connection.py", line 49, in __init__
    self.binary.launch_browser(self.profile)
  File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/firefox/firefox_binary.py", line 68, in launch_browser
    self._wait_until_connectable()
  File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/firefox/firefox_binary.py", line 103, in _wait_until_connectable
    raise WebDriverException("Can't load the profile. Profile "
selenium.common.exceptions.WebDriverException: Message: Can't load the profile. Profile Dir: %s If you specified a log_file in the FirefoxBinary constructor, check it for details.
```

I am almost sure it's to do with hardcoded 30 seconds timeout in firefox_binary.py:101. When I changed that value to 60 it started working.

I tried to load webdriver like this:
`webdriver.Firefox(timeout=60)`

but after looking at the code the timeout is not actually used so made no difference

How to replicate:
 * this issue can only be seen on slow systems (like RaspberryPi) where firefox takes over 30 seconds to start
 * the following python code can be used to replicate the problem:
```
from selenium import webdriver
browser = webdriver.Firefox()
```

jacekn:
I started working on possible fix here:
https://github.com/jacekn/selenium/commits/issue1300

Note that the above branch is untested and not ready for merge, posting here for reference to help anybody else interested in troubleshooting

mjonker-embed:
Seeing the same issue

willpz:
I am getting this error everytime I start FF with this command: "browser = webdriver.Firefox()" , even if I change this hardcoded timeout value to a bigger number (in my test was 100s).
Test environment:
- Raspberry Pi model B running Raspbian Wheezy
- python 2.7 or python 3.2
- Selenium 2.48
- Iceweasel 38.4.0

samsun387:
Seeing the exact same issue..

[...]

PR #1530, opened 2016-01-26
Firefox: Actually use timeout in launch_browser

francoisfreitag:
If I'm not mistaken, `_wait_until_connectable()` waited for 30 seconds, no matter the timeout value given in the Firefox WebDriver `__init__()` method.

_My CLA is signed._


cgoldberg:
i don't know the ramifications of this change, so I'll defer to other reviewers...
but your assertion that the wait is hardcoded to 30 is indeed correct.  good catch!

agude:
Just to link them together: this hardcoded timeout is the source of Issue #1300.

lukeis:
merged as e4b10e1adf8dbe810b4f26c1015697ef9da51a57

PR #1532, opened 2016-01-26
Detect the running platform and set ChromeDriverServiceFileName accor…

rahvee:
…dingly ("chromedriver" or "chromedriver.exe")

Although I couldn't build this project on mono, and a zillion unit tests failed in windows, the changes in this pull request are extremely simple and straightforward, and I was able to manually test in windows.

If you create a project and add Selenium via NuGet, I confirmed that I was able to use chrome in windows, but the new ChromeDriver() constructor threw exception in mono.

This pull request enables the new ChromeDriver() constructor to work in mono, and allows selenium to work with chrome on OSX (and hopefully other linux/unix variants)

### DIFF
--- a/dotnet/src/webdriver/Chrome/ChromeDriverService.cs
+++ b/dotnet/src/webdriver/Chrome/ChromeDriverService.cs
@@ -30,7 +30,7 @@ namespace OpenQA.Selenium.Chrome
     /// </summary>
     public sealed class ChromeDriverService : DriverService
     {
-        private const string ChromeDriverServiceFileName = "chromedriver.exe";
+        private static readonly string ChromeDriverServiceFileName;
         private static readonly Uri ChromeDriverDownloadUrl = new Uri("http://chromedriver.storage.googleapis.com/index.html");
         private string logPath = string.Empty;
         private string urlPathPrefix = string.Empty;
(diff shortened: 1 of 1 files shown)


jimevans:
Have you signed the [CLA](https://spreadsheets.google.com/spreadsheet/viewform?hl=en_US&formkey=dFFjXzBzM1VwekFlOWFWMjFFRjJMRFE6MQ#gid=0)?

Also, even though I don't necessarily agree with the rationale behind this pull request (the `CreateDefaultService` overload is there to allow for the uncommon[1], non-Windows case) I'll be adding some comments on how to better fit in with the coding style of the project. Alternatively, you can run code analysis on the WebDriver and WebDriver.Support projects within Visual Studio[2], which will flag the style differences for your code.

[1] And it _**is*_ uncommon, for users of the WebDriver .NET bindings to be running on a platform other than Windows. This is not up for debate.
[2] Unfortunately, yes, you'll need Visual Studio to do this. However, if you're on a non-Windows platform, you can easily get a Windows VM running, even if it's from [modern.ie](https://dev.windows.com/en-us/microsoft-edge/tools/vms/), and install [Visual Studio 2015 Community Edition](https://go.microsoft.com/fwlink/?LinkId=691978&clcid=0x409) into the VM so that you can use the developer platform.

rahvee:
Signed the CLA just now.

jimevans:
My previous comment on this PR contained content that was belittling of the effort and that's uncalled for. I was wrong, and I publicly apologize for that tone. Rather than change history and edit the previous comment to remove the offending bits, I'm posting a public apology here, and leaving the original comment as a reminder that such things have no place in a public forum.

rahvee:
No worries. I thought I deleted this pull request. Please see https://github.com/SeleniumHQ/selenium/pull/1534 instead.

PR #1534, opened 2016-01-27
Made ChromeDriverServiceFileName a method instead of a const, so it w…

rahvee:
…ill work on unix/linux


jimevans:
Merged as f048ab8

PR #1569, opened 2016-02-01
Fix type warning

joshbruning:
Fix type warning revealed by improved array-like object
type checking in the Closure Compiler.


lukeis:
merged as 9b11f1f9e766237e6bb28a8023d046a45501f96c

PR #1582, opened 2016-02-03
[java] optimize JsonToBeanConverter

alb-i986:
In case the input is a JsonNull, return null asap.


barancev:
Merged as 2021f288084990783dee85f8d2372cb4b569f484

PR #1623, opened 2016-02-10
[Java] Fix deselecting options in <select>

glibas:
1. Fix inconsistency between java doc and implementation:
   deselectByVisibleText, deselectByIndex and deselectByValue
   did not throw NoSuchElementException, while they should have to.
2. Remove check !isMultiple() while selecting/deselecting by index,
   since no need as index is unique.
3. Add appropriate tests to SelectElementTest
4. Include SelectElementTest in LargeTests suite, since it was
   not a part of any.

### DIFF
--- a/java/client/src/org/openqa/selenium/support/ui/Select.java
+++ b/java/client/src/org/openqa/selenium/support/ui/Select.java
@@ -233,16 +233,19 @@ public class Select {
    *
    * @param value The value to match against
    * @throws NoSuchElementException If no matching option elements are found
+   * @throws UnsupportedOperationException If the SELECT does not support multiple selections
    */
   public void deselectByValue(String value) {
+    if (!isMultiple()) {
+      throw new UnsupportedOperationException(
+          "You may only deselect options of a multi-select");
+    }
+    
     List<WebElement> options = element.findElements(By.xpath(
         ".//option[@value = " + Quotes.escape(value) + "]"));
     boolean matched = false;
     for (WebElement option : options) {
       setSelected(option, false);
-      if (!isMultiple()) {
-        return;
-      }
       matched = true;
     }
     if (!matched) {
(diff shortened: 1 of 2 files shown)


java/client/src/org/openqa/selenium/support/ui/Select.java
```diff
@@ -283,15 +282,31 @@ public void deselectByVisibleText(String text) {
...
+   * Select or deselect specified option
+   *
+   * @param option
+   *          The option which state needs to be changed
+   * @param select
+   *          Indicates whether the option needs to be selected (true) or
+   *          deselected (false)
+   */
+  private void setSelected(WebElement option, boolean select) {
+    if ((!option.isSelected() && select) || (option.isSelected() && !select)) {
```
lukeis: it's not efficient to make the remote call to isSelected multiple times here. Please either revert the change to this method or store the isSelected value in a local variable.

glib-briia: Hi @lukeis,

Thank you for your comment. It is indeed. Updated in commit f326050e1d611da34dbce9748de6d627efd46f3f.

java/client/src/org/openqa/selenium/support/ui/Select.java
```diff
@@ -245,11 +237,16 @@ public void deselectAll() {
   public void deselectByValue(String value) {
     List<WebElement> options = element.findElements(By.xpath(
         ".//option[@value = " + Quotes.escape(value) + "]"));
-
+    boolean matched = false;
     for (WebElement option : options) {
-      if (option.isSelected()) {
-        option.click();
+      setSelected(option, false);
+      if (!isMultiple()) {
```
lukeis: deselecting a non-multiple select doesn't make sense. What would it even mean to deselect something in a normal select? A user can't do it.

Rather than adding this check here (same goes for deselectByVisibleText), we could check at the start of the method and throw an exception if it is not multiple.

lukeis:
also can you confirm you've signed the CLA? http://goo.gl/qC50R

glib-briia:
Yes, you are right, thank you. Updated in commit bf60cb9909e8ddb84a8ece88e558c5a6c04a108d. 
Appropriate tests added. 

_CLI is signed._

lukeis:
Thanks!

merged as 5883af125a72d901b7fbe79664cce9fa4d7983a8

PR #1631, opened 2016-02-13
Set chromedriver to be a new process group leader

bsedat:
Allows for easier killing of the entire chromedriver process tree.

I have signed the CLA as "Benjamin Sedat".

### DIFF
--- a/rb/lib/selenium/webdriver/chrome/service.rb
+++ b/rb/lib/selenium/webdriver/chrome/service.rb
@@ -102,6 +102,7 @@ module Selenium
           @process       = ChildProcess.build(*server_command)
 
           @process.io.inherit! if $DEBUG == true
+          @process.leader = true
           @process.start
         end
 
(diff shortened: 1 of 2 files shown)


p0deje:
I'm not sure I follow your reasoning. What do you mean by "easier killing of the entire chromedrvier process tree"?

I may understand Unix process model wrong, but as far as I get it, chromedriver inherits runner process group ID (e.g. test runner, irb), so that when signal is sent to test runner, all processes in this process group receive signal. If chromedriver starts new process group, the signals will not be sent to it and its children process (Chrome).

Can you please elaborate on this a bit more?

bsedat:
I was running into an issue where chromedriver was getting orphaned and reparented to pid 1. I was hoping to use this as a way to facilitate killing the whole chromedriver process tree when this occurs. Any suggestions for workarounds? Or should this an option that could be set?

p0deje:
I would rather figure out the reason why chromedriver gets orphaned, it should not happen. Can you find a way to reproduce the issue?

p0deje:
Closing since there is no response for more than 3 months. Please re-open if necessary.

PR #1649, opened 2016-02-17
Add Travis CI badge to README.md

Dominator008:
Also remove trailing whitespaces.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,15 @@
 # Selenium
 
-Selenium is an umbrella project encapsulating a variety of tools and 
-libraries enabling web browser automation. Selenium specifically 
-provides infrastructure for the [W3C WebDriver specification](https://dvcs.w3.org/hg/webdriver/raw-file/tip/webdriver-spec.html) 
-— a platform and language-neutral coding interface compatible with all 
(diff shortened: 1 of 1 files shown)


lukeis:
Thanks for the offer, but Travis is not our main CI, we have another Jenkins instance and only pull requests are run through Travis for Python and ruby changes.

So for now we'd prefer to not have this.

johnjbarton:
FWIW, I think the Travis would be helpful to know that the build known to be failing. As it is, I opened an issue that was probably not useful.  The fact that another CI also runs is fine, but that does not reduce the benefit of the Travis signal as far as I can tell.

cgoldberg:
@lukeis ,
are the jenkins results publicly viewable somewhere?

lukeis:
@cgoldberg http://ci.seleniumhq.org:8080/

Dominator008:
@lukeis https://wiki.jenkins-ci.org/display/JENKINS/Embeddable+Build+Status+Plugin seems to provide a Jenkins CI badge.

cgoldberg:
@lukeis 
are the python tests also run by Jenkins?  I don't see them

lukeis:
@cgoldberg sadly no... they are not.

@Dominator008 will need to bring it up with @juangj to install.

PR #1655, opened 2016-02-17
Send keys optimization.

atanistra:
Change range(len(val) to string iteration.
Remove duplicate for loop.


lukeis:
I'd rather see a change similar to what #1666 is doing... since this code snippet appears to be duplicated and should be consolidated.

lukeis:
merged #1666 instead, thanks for the PR

PR #1666, opened 2016-02-18
[py] send_keys() cleanup

mfalesni:
- Made it possible to ignore files when doing send_keys(). This was bugging me for a long time but now it blocked some of our testing because we type file paths in the UI and they might exist on the testing machine.
- I saw the send_keys()' preprocessing logic is multiplied, so I took it out and separated it to a function.

### DIFF
--- a/py/selenium/webdriver/common/action_chains.py
+++ b/py/selenium/webdriver/common/action_chains.py
@@ -275,10 +275,6 @@ class ActionChains(object):
             element.send_keys(*keys_to_send))
         return self
 
-    def _keys_to_typing(self, value):
-        """DEPRECATED: Wrapper here just in case someone has been using this."""
-        return keys_to_typing(value)
-
     # Context manager so ActionChains can be used in a 'with .. as' statements.
     def __enter__(self):
         return self # Return created instance of self.


py/selenium/webdriver/common/__init__.py
```diff
@@ -14,3 +14,20 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+from .keys import Keys
```
lukeis: i don't believe it's good form to put functionality in the init file. It seems more appropriate to put it in common/keys.py

py/selenium/webdriver/remote/webelement.py
```diff
@@ -313,25 +317,15 @@ def send_keys(self, *value):
             # file_input.send_keys(os.path.abspath("path/to/profilepic.gif"))
 
         """
+        ignore_files = kwargs.get('ignore_files', False)
         # transfer file to another machine only if remote driver is used
         # the same behaviour as for java binding
-        if self.parent._is_remote:
+        if self.parent._is_remote and not ignore_files:
             local_file = self.parent.file_detector.is_local_file(*value)
```
lukeis: did you look at trying to set the 'file_detector'? this was provided to allow exactly what you're trying to do.

for example if you wanted to disable it:

    from selenium.webdriver.remote import file_detector
    driver.file_detector = file_detector.UselessFileDetector()

    #and then to set it back to default behavior
    driver.file_detector = file_detector.LocalFileDetector()

py/selenium/webdriver/common/__init__.py
mfalesni: Well, actually, I figured out this import is useless as there is no instance of Keys that could be passed to the function as the Keys is just a holder of the values. So I will delete it. Therefore this code is unrelated to the Keys class and if you oppose the package file, I propose putting it into the utils.py file, since it is a utility function breaking things into lists.

py/selenium/webdriver/remote/webelement.py
mfalesni: Hmm, but ... doesn't look very pythonic to me. I have 2 points to this approach:

1) A side effect, affecting the behaviour of a method from different class (driver/element).
2) From programmer's point of view, if a method is named 'send_keys', it should send keys. File upload field handling should get its own method so the files can be handled accordingly. I know this cannot be changed so that is why I added the additional parameter that can make switching of the behaviour more practical and straightforward.

mfalesni: @lukeis So, I solved it for our project, but I was thinking - since it made a pretty awful code IMO (try/finally/if): https://github.com/RedHatQE/cfme_tests/pull/2690/files#diff-54c758ec2117f433d2c05581244ceb74L756

What would you say if I implemented this behaviour as a context manager so one would just do:

```python
with webdriver.file_detector(UselessFileDetector):
    # do something
```

If it did not have such file detector already, it would create it, set it, run the code in the block and then set it back. If it had the same kind of detector already, it would do nothing. This would abstract the behaviour from the try/finally block.

mfalesni:
Weird, python tests passed but Java code exploded.

py/selenium/webdriver/common/action_chains.py
```diff
@@ -274,18 +276,8 @@ def send_keys_to_element(self, element, *keys_to_send):
...
-                typing.append(val)
-            elif isinstance(val, int):
-                val = str(val)
-                for i in range(len(val)):
-                    typing.append(val[i])
-            else:
-                for i in range(len(val)):
-                    typing.append(val[i])
-        return typing
+        """DEPRECATED: Wrapper here just in case someone has been using this."""
```
lukeis: This is a private method, don't need to deprecate, just remove. We'll make mention in the changelog.

lukeis:
Looking good, have you signed the CLA?

mfalesni:
@lukeis I did fill the Google Doc form.

lukeis:
merged as d095cd2a410cbf8ffd95cac9aed364221151f722

PR #1676, opened 2016-02-21
Added an overload of ExecuteJavaScript extension method that returns void

arnonax:

Sometimes I want to execute JavaScript through Selenium, but the script doesn't return a value. Before this change I either had to use the ((IJavaScriptExecutor)IWebDriver).ExecuteScript(myScript, args) or somehow change my script to return something and use the ExecuteJavaScript extension, with the generic argument. 
This change allows to execute JavaScript through the Selenium Extensions without casting and without having to return any value.


arnonax:
Does someone has an estimation for how long it should take to until the change is reviewed, and what are the next steps? (This is a very small change and I submitted it 23 days ago...)

jimevans:
Merged in c824d7b

PR #1677, opened 2016-02-22
fix mixed content in grid console

gregjhogan:
If you have a reverse proxy enforcing HTTPS you get a mixed content error in most modern browsers which blocks pulling down content over HTTP


lukeis:
merged as c46671f96e9be203c627ca1920c23cecfd69c8b1

PR #1683, opened 2016-02-22
Update DefaultWait{T}.cs

Johnsavadkuhi:
if (clock == null)
            {
                throw new ArgumentNullException("clock", "clock cannot be null");
            }


jimevans:
Merged in 0740288

PR #1684, opened 2016-02-23
Catch Net::ReadTimeout errors when stopping the browser

bsedat:

### DIFF
--- a/rb/lib/selenium/webdriver/chrome/service.rb
+++ b/rb/lib/selenium/webdriver/chrome/service.rb
@@ -83,7 +83,7 @@ module Selenium
 
           @process.poll_for_exit STOP_TIMEOUT
         rescue ChildProcess::TimeoutError
-          # ok, force quit
+        ensure
           @process.stop STOP_TIMEOUT
         end
 
--- a/rb/lib/selenium/webdriver/edge/service.rb
+++ b/rb/lib/selenium/webdriver/edge/service.rb
@@ -83,7 +83,7 @@ module Selenium
 
           @process.poll_for_exit STOP_TIMEOUT
         rescue ChildProcess::TimeoutError
-          # ok, force quit
+        ensure
           @process.stop STOP_TIMEOUT
         end
 
--- a/rb/lib/selenium/webdriver/firefox/service.rb
+++ b/rb/lib/selenium/webdriver/firefox/service.rb
@@ -82,7 +82,7 @@ module Selenium
 
           @process.poll_for_exit STOP_TIMEOUT
         rescue ChildProcess::TimeoutError
-          # ok, force quit
+        ensure
           @process.stop STOP_TIMEOUT
         end
 
--- a/rb/lib/selenium/webdriver/phantomjs/service.rb
+++ b/rb/lib/selenium/webdriver/phantomjs/service.rb
@@ -78,7 +78,7 @@ module Selenium
 
           @process.poll_for_exit STOP_TIMEOUT
         rescue ChildProcess::TimeoutError
-          # ok, force quit
+        ensure
           @process.stop STOP_TIMEOUT
 
           if Platform.jruby? && !$DEBUG


titusfortner:
I'm not sure I follow what condition your system has to be in for the server to still be running, but not responding. Do we also need to avoid `poll_for_exit` in that case? Do you have a reproducible test case that would make this easier for me to understand the desirability of this? @p0deje - do you see a reason to rescue or not rescue this here?

bsedat:
I've run into this before when the system is under heavy load (and probably fairly underpowered).

You bring up a good issue regarding the `poll_for_exit` call. It's still possible that the server has received the request, but the system is too hosed in order for it to respond in time. I think I'm okay moving the rescue to be tightly around the Net::HTTP calls if that is what is preferred.

p0deje:
Assuming something goes completely wrong in stopping chromedriver (i.e. it doesn't respond), it's perfectly fine to raise an exception stating this. The process is still killed since it's child process of test runner (or any other process starting selenium session), so I don't see a reason to suppress the exception. In this case, you'll have no idea if something is going wrong.

bsedat:
I was seeing the Process exit handler error out before killing the child process. In this case I believe the child process is still alive, just that the host is too taxed to respond to HTTP requests promptly. 

I think we should make a best-effort to notify the child process to shut down gracefully, and if we fail to do that then kill it forcefully. The Net::HTTP call was throwing the `Net::ReadTimeout` error and aborting the exit handler before it got a chance to call `@process.stop STOP_TIMEOUT` (which probably should bubble up any fatal errors at that point).

p0deje:
I see your point, though I'm still not sure if this brings any issues rather than raises `Net::ReadTimeout` - the chromedriver process is still killed, isn't it?

bsedat:
Seems like it, although I can't confirm because we were running into this while running a legacy test suite in over-taxed containers that get destroyed promptly either way. The ruby process would exit abnormally (exit code 1) because an exit handler raised an error - catching that and letting ChildProcess kill the browser once the graceful shutdown timed out seemed to follow what the code was already doing when catching the ChildProcess timeout.

I believe this is a separate problem from my other pull request https://github.com/SeleniumHQ/selenium/pull/1631. That was attempting to address a long-running process that starts and stops many browsers over hours/days leaking orphaned chromedriver/chrome processes.

I did a quick test and confirmed that one exit handler raising an error does not abort others that have been registered. I'm happy to surround the `Net:HTTP` call with a specific rescue block to allow ChildProcess to wait a little bit longer for the process to maybe end on its own (the call to `@process.poll_for_exit STOP_TIMEOUT`) if you would like.

p0deje:
> The ruby process would exit abnormally (exit code 1) because an exit handler raised an error

Are you saying that previous calls to ChromeDriver were working fine and only the last one (stop) is timing out?

> That was attempting to address a long-running process that starts and stops many browsers over hours/days leaking orphaned chromedriver/chrome processes.

I see the problem here, though maybe it makes sense to extract the part which starts/stops browser and fork it into its own process group from the long-running process?

Generally, there is no single approach of stopping drivers in different bindings. Java will throw exception if request to `/shutdown` times out but ensures process is killed, JavaScript simply kills the process, Python/Ruby make request to `/shutdown` without rescuing timeouts. Assuming there is no specific way to do that, I'm fine to make one of the following:

1. Change `#stop` so that it ensures call to `@process.stop` (which simply does nothing if process is already stopped). In this case, we'll still be getting exceptions and be able to see that something is wrong.
2. Rescue timeout error and force `@process.stop` (what this PR does). This may be a bit misleading since exception is suppressed, so maybe exceptions should be re-raised.

@titusfortner @lukeis Do you have any thoughts on this?

titusfortner:
I think what we want is:
```
rescue ChildProcess::TimeoutError
ensure
  @process.stop STOP_TIMEOUT
end
```
We don't need the ChildProcess::TimeoutError raised, but I think we would want the Net::ReadTimeout raised, but either way, we'll get the process stopped

bsedat:
Updated the pull request. Maybe in the future there can be a configuration option somewhere for configuring the behavior of the browser's exit handler?

titusfortner:
Great, thanks. What alternatives do you want to be able to configure?

titusfortner:
Merged: 0d942bcd55a897101aa921f45a9b1feeb3f761dc

PR #1695, opened 2016-02-25
Add documentation explaining stale WebElements

ERonayne:

Documentation was added to remote/webelement.py and
remote/errorhandler.py in order to make it clear why a
StaleElementReferenceException was thrown. An example was also provided
of how to handle the exception.

### DIFF
--- a/py/selenium/webdriver/remote/errorhandler.py
+++ b/py/selenium/webdriver/remote/errorhandler.py
@@ -81,6 +81,23 @@ class ErrorHandler(object):
         """
         Checks that a JSON response from the WebDriver does not have an error.
 
+        If an element is no longer part of the DOM a StaleElementReferenceException
+        will be thrown. An element is most often no longer part of the DOM because
+        the DOM has been destroyed. The DOM is destroyed when a new web page is loaded.
+        This even includes refreshing the same web page.
+
+        In order to handle a StaleElementReferenceException you must re-find the WebElement.
+
+        For Example:
+
+            try:
+                element = driver.find_element_by_css_selector(css_selector)
+                element.click()
+            except StaleElementReferenceException:
+                driver.get(url)
+                element = driver.find_element_by_css_selector(css_selector)
+                element.click()
+
         :Args:
          - response - The JSON response from the WebDriver server as a dictionary
            object.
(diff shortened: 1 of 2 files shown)


py/selenium/webdriver/remote/errorhandler.py
```diff
@@ -81,6 +81,23 @@ def check_response(self, response):
         """
         Checks that a JSON response from the WebDriver does not have an error.
 
+        If an element is no longer part of the DOM a StaleElementReferenceException
```
lukeis: not really a good reason to add this py doc here. Most users shouldn't be calling (or even looking at) the ErrorHandler class, it's mostly an internal class to handle responses from the server

py/selenium/webdriver/remote/webelement.py
```diff
@@ -47,7 +47,25 @@ class WebElement(object):
     reference is still valid.  This essentially determines whether or not the
     element is still attached to the DOM.  If this test fails, then an
     ``StaleElementReferenceException`` is thrown, and all future calls to this
-    instance will fail."""
+    instance will fail.
+
+    If an element is no longer part of the DOM a StaleElementReferenceException
```
lukeis: again this is not really an appropriate place to be putting this sort of documentation

py/selenium/webdriver/remote/webelement.py
```diff
@@ -47,7 +47,25 @@ class WebElement(object):
...
+
+    For Example:
+
+        try:
+            element = driver.find_element_by_css_selector(css_selector)
+            element.click()
+        except StaleElementReferenceException:
+            driver.get(url)
+            element = driver.find_element_by_css_selector(css_selector)
+            element.click()
```
lukeis: This example code is showing a very bad code practice. A more practical example would be using an expected_condition (ignoring stale element):

    from selenium.webdriver.support.wait import WebDriverWait
    from selenium.webdriver.support.expected_conditions import visibility_of_element_located
    from selenium.common.exceptions import StaleElementReferenceException

    WebDriverWait(driver, 10, ignored_exceptions=StaleElementReferenceException)
        .until(visibility_of_element_located((By.CSS_SELECTOR, selector))).click()

(You don't actually have to supply the ignored_exceptions for visibility_of_element_located, the code actually ignores StaleElementReference... but I'm just trying to be verbose in showing how one uses expected conditions and can ignore this one)

ERonayne: Thanks for the feedback. I was unaware of expected_conditions. I had never seen them used before but they definitely seem useful. I added the doc in these places mostly because this was where I first looked for information on StaleElementReferenceException. Generally, I prefer reading source code over the official documentation, but I understand most people probably don't. If you don't think it's appropriate then should the PR be deleted?

PR #1697, opened 2016-02-25
Added the new value of the element to the ElementValueChanged and Ele…

arnonax:

Added the new value of the element to the ElementValueChanged and ElementValueChanging event args.

Note: I created a new overload of OnElementChanging and OnElementChanged but also kept the original overloads in order to preserve backward compatibility in case someone overrides them. I've added an [Obsolete] attribute on the original overloads to let users know that they should use the new ones instead.


arnonax:
Hi @jimevans ,

Can you please tell me when can I expect this pull request to be pulled?
I'm using these events to write some logs automatically, and adding the value that is passed to SendKeys will be very helpful when reading the log to analyze test failures.

Thanks in advance,
Arnon.

jimevans:
Merged in 6dd04e6. Thank you for the contribution.

arnonax:
Thanks!

PR #1708, opened 2016-02-26
[atoms] Implement clear functionality for checkbox & radiobutton.Also added nontextual editable input types.

gmanikpure:

### DIFF
--- a/javascript/atoms/test/action_test.html
+++ b/javascript/atoms/test/action_test.html
@@ -183,10 +183,32 @@
 
     function testClearingNumberInputWithInvalidData() {
       var e = goog.dom.getElement('numberField');
-      bot.action.type(e, "e");
+      bot.action.type(e, 'e');
       bot.action.clear(e);
-      bot.action.type(e, "3");
-      assertEquals("3", goog.dom.forms.getValue(e));
+      bot.action.type(e, '3');
+      assertEquals('3', goog.dom.forms.getValue(e));
+    }
+
+    function testClearingAEnterableControlInput() {
+      var e = goog.dom.getElement('weekField');
+      e.value = '2016-W09';
+      assertNotEquals('', e.value);
+      bot.action.clear(e);
+      assertEquals('', e.value);
+    }
+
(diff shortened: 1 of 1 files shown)


gmanikpure:
@juangj 
Jason, could you please have a look?
Thanks,

javascript/atoms/action.js
```diff
@@ -1,3 +1,20 @@
+/**
+ * @license
```
lukeis: why are you adding another license here and in dom.js?

javascript/atoms/action.js
```diff
@@ -75,27 +93,41 @@ bot.action.checkInteractable_ = function(element) {
...
+
   if (element.value) {
     element.value = '';
     bot.events.fire(element, bot.events.EventType.CHANGE);
-  } else if (bot.dom.isElement(element, goog.dom.TagName.INPUT) &&
-             (element.getAttribute('type') && element.getAttribute('type').toLowerCase() == "number")) {
-    // number input fields that have invalid inputs
-    // report their value as empty string with no way to tell if there is a
-    // current value or not
+  } else if (bot.dom.isEditable(element) && !bot.dom.isTextual(element)) {
```
lukeis: why wouldn't we call bot.dom.isNonTextual (and not have to negate the value).

Also it seems like the type="number" case wouldn't evaluate to true here? (since according to bot.dom it is textual)

AutomatedTester:
this PR seems to be against the general spirit of WebDriver which is to emulate users.

If a user wants to clear a checkbox they need to click again and in the case of a radiobutton a user can not clear it so webdriver shouldn't magically clear it either.

For these reason I am closing this PR.

gmanikpure:
Sorry for the inconvenience. I tried to follow WebDriver specifications document 
https://www.w3.org/TR/webdriver/#element-clear

"If the element is editable then set the value of element to empty string. Otherwise, for checkbox / radio button attribute types, set the checkedness to false."
Please correct me if i am wrong, Thanks!

Issue #1679, opened 2016-02-22
Memory leak in org.openqa.selenium.io.TemporaryFilesystem


xmhubj:
We have a long run environment to playback selenium scripts. The command to start the remote server is as the following.

java -Xdump:java:events=user -jar -Djava.util.logging.config.file=/opt/my/logging.properties /opt/my/selenium-server-standalone-reduced-2.47.1.jar -timeout 3600 -browserTimeout 3600

Find that the real memory consumption of this process keeps growing over time.

## Expected Behavior -
The real memory consumption of the selenium-server-standalone process should not grow overtime with steady load.

## Actual Behavior -
The heap dump of the selenium server process shows that the class "java.lang.ApplicationShutdownHooks", loaded by "<system class loader>", occupies 97,222,648 (93.28%) bytes.

And there are large number of org.openqa.selenium.io.TemporaryFilesystem$1 objects (159724 in my case) dominated by class "java.lang.ApplicationShutdownHooks" and keeps growing over time.

![dominator_tree](https://cloud.githubusercontent.com/assets/15373664/13215020/95602dde-d98d-11e5-9905-77eefeac5d60.jpg)

## Steps to reproduce -
No special steps. 

PS: for each test, we stop the BMP proxy and run driver.quit.

Cloud you advise how to find the root cause of this issue? Thanks.

dr29bart:
Shutdown hooks are not deleted for some reason.

xmhubj:
I simply took a look at the source code of class org.openqa.selenium.io.TemporaryFilesystem, only found that shutdownHooks are added and no removeShutdownHook is called.
line 62: Runtime.getRuntime().addShutdownHook(shutdownHook), 

And looks like shutdownHooks are only removed when stopping the Jetty server.
...
/**
   * Stops the Jetty server
   */
  public void stop() {
    int numTries = 0;
    Exception shutDownException = null;

    // this may be called by a shutdown hook, or it may be called at any time
    // in case it was called as an ordinary method, try to clean up the shutdown
    // hook
    try {
      if (shutDownHook != null) {
        Runtime.getRuntime().removeShutdownHook(shutDownHook);
      }
    } catch (IllegalStateException ignored) {
    } // thrown if we're shutting down; that's OK
...

Can anyone help me out. Thanks!

krmahadevan:
@xmhubj 

This was interesting. :+1: 

Here's where I think the problem is.

[...]

PR #1709, opened 2016-02-27
Fixing memory leak in TemporaryFileSystem

krmahadevan:

Fixes issue : #1679

Added logic to remove the ShutdownHook that 
TemporaryFileSystem was injecting for cleaning up 
temporary files created by Selenium, after the cleanup
was completed.

### DIFF
--- a/java/client/src/org/openqa/selenium/io/TemporaryFilesystem.java
+++ b/java/client/src/org/openqa/selenium/io/TemporaryFilesystem.java
@@ -146,7 +146,7 @@ public class TemporaryFilesystem {
 
   public boolean deleteBaseDir() {
     boolean wasDeleted = baseDir.delete();
-    if (wasDeleted) {
+    if (! baseDir.exists()) {
       Runtime.getRuntime().removeShutdownHook(shutdownHook);
     }
     return wasDeleted;


krmahadevan:
@barancev 
Can you please help take a look at this and let me know if this PR is fine ?

java/client/src/org/openqa/selenium/io/TemporaryFilesystem.java
```diff
@@ -145,6 +145,10 @@ boolean shouldReap() {
   }
 
   public boolean deleteBaseDir() {
-    return baseDir.delete();
+    boolean wasDeleted = baseDir.delete();
+    if (wasDeleted) {
```
lukeis: I don't think we need to necessarily check if that returned true or not, we should just remove the shutdown hook. Since nothing would check or do anything with this flag anyways.

krmahadevan: @lukeis 
I also thought about it, but thought that maybe we should give the hook another chance to cleanup when the JVM exits assuming that the current attempt of deleting the `baseDir` failed. Please let me know if you still feel that this can be removed and i will update this PR with those changes as well.

lukeis: I was just trying to check on what java is doing. It appears that delete could return false if the path was already deleted. Which we would want to remove the shutdown hook at that point too.

Maybe change this line to if (!baseDir.exists())

krmahadevan:
@lukeis 
That makes sense. I have made the changes as per your recommendations.

java/client/src/org/openqa/selenium/io/TemporaryFilesystem.java
```diff
@@ -145,6 +145,9 @@ boolean shouldReap() {
   }
 
   public boolean deleteBaseDir() {
+    if (! baseDir.exists()) {
```
lukeis: heh :) shouldn't this be after the delete?

krmahadevan:
I looked at the Travis Job and couldn't make out why it failed though [ seems to be something related to the environment ] :(

java/client/src/org/openqa/selenium/io/TemporaryFilesystem.java
krmahadevan: Yikes... yeah you are correct... let me change that..

lukeis:
that's my fault... i need to fix that...

krmahadevan:
ok, I think I did it properly this time. Please bear with me.. the excitement is making me get carried away... Its my first PR into Selenium :)

lukeis:
merged as 8ca0979612e4b7a0605a77eb3a36b6c6913e53aa

thanks @krmahadevan :)

PR #1710, opened 2016-02-28
DON'T MERGE - Testing Travis CI failures

titusfortner:
Travis only kicks off with PRs, will delete after verifying test.

### DIFF
--- a/py/force_travis_test.txt
+++ b/py/force_travis_test.txt
@@ -1,0 +1,1 @@
+Python Change
--- a/rb/force_travis_test.txt
+++ b/rb/force_travis_test.txt
@@ -1,0 +1,1 @@
+Ruby Change


titusfortner:
@lukeis - [According to Travis](https://travis-ci.org/SeleniumHQ/selenium/builds/112463122
), Ruby isn't working with Phantomjs 1.9
and on Phantomjs 2.1, Python hangs here:
```
build/lib.linux-x86_64-2.7/selenium/test/selenium/webdriver/phantomjs/phantomjs_frame_switching_tests.py 
```
Do we have a policy or preference for which version of this we support? :)

lukeis:
@titusfortner i think i would prefer the latest, if there's a particular test that hangs we can skip it in python (will take a look later)

titusfortner:
There we go. @lukeis - I'll merge the first two commits later this weekend. The python firefox tests run just fine, but there are a couple that fail for some reason. For Chrome, the Python tests likely need to be able to respond to `ENV['CHROME_BINARY']` & `ENV['CHROMEDRIVER']` and chrome needs to be started with the arg: `--no-sandbox`

titusfortner:
I merged this: 0ca948224051dbed6fa2ae2f9396802086ab9150
Also of note, @lukeis, I removed the `build` command. I don't see a need to build everything in the app just to test Python & Ruby, but we can put it back if you think we need it.

lukeis:
@titusfortner the build command was there first to make sure changes to Java in pull requests don't break the build :) put it back please?

titusfortner:
@lukeis - isn't that what our Jenkins CI is for? The Ruby/Python tests should build the things it needs to test from, and even then, I think all that gets used is the Firefox extension if we are actually running firefox. Am I missing something?

lukeis:
This only gets run on pull requests, Jenkins only runs after it's committed / pushed to master. Think of the pull requests for random user, would you trust them to not break the build? ;)

titusfortner:
Oh, I didn't realize Jenkins didn't run for PRs. This completely makes sense now. Back in - 285ecb5996919f4a6ab3c8116be6c5d0c79173b2

PR #1726, opened 2016-03-02
Python: Page Object Implementation

joshuaduffy:

Page object implementation in Python. This allows you to follow the Page Object Model design pattern for test development and borrows its ideas from the C#/Java implementation of PageObjects.

Each page object inherits from `PageObject` and defines a number of `FindBy` or `FindAllBy` within.

An example the implementation in action:

```
        from selenium.webdriver import Firefox
        from selenium.webdriver.support.page_objects import PageObject, FindBy, FindAllBy

        class SeleniumHome(PageObject):
            search_box = FindBy(By.ID, 'q')
            submit_button = FindBy(By.ID, 'submit')

            def __init__(self, driver):
                super().__init__(driver) #  Python 3 __super__

            def search_for(self, text):
                self.search_box.send_keys(text)
                self.submit_button.click()

                return SearchPage(self.driver)


        class SearchPage(PageObject):
            search_results = FindAllBy(By.CLASS_NAME, 'gs-title')

        wd = Firefox()
        wd.get('http://www.seleniumhq.org/')
        page = SeleniumHome(wd)
        results = page.search_for('Hello')
        results.search_results[2].click()
```

The implementation is very simple, and only defines the `PageObject`, `FindBy` and `FindAllBy` objects.

The magic happens in the `FindBy`/`FindAllBy`, upon getting it, we return an instance (or list of instances) of the `WebElement` defined by the parameters in the `FindBy`/`FindAllBy` constructor

Tests are included. Let me know your thoughts on this. I'm completely open to anything.

Thanks.

### DIFF
--- a/py/docs/source/webdriver_support/selenium.webdriver.support.page_objects.rst
+++ b/py/docs/source/webdriver_support/selenium.webdriver.support.page_objects.rst
@@ -1,6 +1,28 @@
 selenium.webdriver.support.page_objects
 =======================================
 
+The Page Object Pattern is a useful design pattern to consider when building automated test suites using selenium.
+
+A more comprehensive overview of the pattern can be found here_.
+
+.. _here: http://www.seleniumhq.org/docs/06_test_design_considerations.jsp
+
+***********
+Page Object
+***********
+
+The base object that all pages must be based on. You implement the `Find By`_ and `Find All By`_ object on this as properties.
+
 .. automodule:: selenium.webdriver.support.page_objects.page_object
+
+*******
+Find By
+*******
+
 .. automodule:: selenium.webdriver.support.page_objects.find_by
+
+***********
+Find All By
+***********
+
 .. automodule:: selenium.webdriver.support.page_objects.find_all_by


joshuaduffy:
Fixed incorrect import in tests and rebased.

joshuaduffy:
It seems to be skipping my tests?

ivargas-handy:
Your FindAllBy class is a subclass of list. I don't consider a list of WebElements a variatio of list. Why not just make the class implement the magic methods that make it act like a list?

drobota:
There is good [library](https://holmiumcore.readthedocs.org/en/latest/) which represents page object pattern.

p0deje:
And also one more [good library](http://wgnet.github.io/webium/).

Dude-X:
I like this page object pattern but I wouldn't subclass the list or WebElement class. I would rather you implement the magic methods that make it list like; basically favor composition over inheritance. The webium project is a good example to emulate.

joshuaduffy:
Yeah a lot of great libraries exist already, I spent some time looking at the others before embarking on this pull request.

I think it would be really useful to have this out of the box, like the bindings for C# and Java for example.

With regards to the use of subclassing `list` and `WebElement`...

I'm in agreement, it feels awful and it's a bit hacky (I'm not even calling `super()`) The main reason I did this is so we get auto-completion in IDE's and to get rid of the duck typing warning that PyCharm will raise when using a `FindBy` or `FindAllBy` attribute.

I believe the `EventFiringWebDriver` uses empty methods to get around this, which is something I would be happy to do.

Just wanted to gauge peoples interest in this, and see if the contributors would be happy to include such functionality.

Any questions give me a shout. I'll be happy to discuss any design choices and I'm open to suggestions.

Thanks,
Josh

Dude-X:
If you want to have good support for PyCharm, I would recommend writing docstrings that documents the class and it's methods; PyCharm can use this info, as well as regular users. If you do this, I would lobby the committers to add your code. It would bring parity to the Python bindings compared to the other ones.

joshuaduffy:
It seems like the `rtypes` I've included, along with return value from WebDriverWait gives us auto-completion in IDE's/PyCharm. I've also included some documentation in the docstrings, and sorted the branch out.

Let me know what you think.

py/selenium/webdriver/support/page_objects/find_all_by.py
```diff
@@ -0,0 +1,61 @@
...
+            class Home(PageObject):
+                # Must be used on a PageObject
+                elements = FindAllBy(By.ID, 'qs')
+                more_elements = FindAllBy(By.CLASS_NAME, 'elementsList')
+
+        throws TimeoutException if nothing can be found via the specified locator values.
+
+        :rtype: list(WebElement)
+        """
+        if not By.is_valid(by) or not isinstance(value, str):
```
Dude-X: Let's support python 2 and 3 by adding a check for `basestring` , define basestring as str if python is version 3

py/selenium/webdriver/support/page_objects/find_all_by.py
```diff
@@ -0,0 +1,61 @@
...
+            raise InvalidSelectorException("Invalid locator values passed in")
+
+        self.by = by
+        self.value = value
+
+    def __get__(self, obj, *args):
+        if not isinstance(obj, PageObject):
+            raise WebDriverException("FindAllBy can only be used on a PageObjects")
+        self.elements = WebDriverWait(obj.driver, 100).until(
```
Dude-X: let's make the timeout for webdriver wait a configurable option, instead of defaulting to 100 seconds.
I guess a default class global is good, with a method to set it. A conservative default is 30 seconds.

py/selenium/webdriver/support/page_objects/find_all_by.py
```diff
@@ -0,0 +1,61 @@
...
+        """
+        if not By.is_valid(by) or not isinstance(value, str):
+            raise InvalidSelectorException("Invalid locator values passed in")
+
+        self.by = by
+        self.value = value
+
+    def __get__(self, obj, *args):
+        if not isinstance(obj, PageObject):
+            raise WebDriverException("FindAllBy can only be used on a PageObjects")
```
Dude-X: An error message like `FindAllBy can only be used on a PageObject instance. The instance passed was: %r`.

py/selenium/webdriver/support/page_objects/find_by.py
```diff
@@ -0,0 +1,62 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+from selenium.common.exceptions import InvalidSelectorException, WebDriverException
+from selenium.webdriver.common.by import By
+from selenium.webdriver.remote.webelement import WebElement
+from selenium.webdriver.support.page_objects.page_object import PageObject
+from selenium.webdriver.support.wait import WebDriverWait
+
+class FindBy(object):
```
Dude-X: Same comments, Python 3 compatibility, helpful error messages.

py/selenium/webdriver/support/page_objects/find_all_by.py
joshuaduffy: Great catch, cheers. This pattern is actually used throughout the Python bindings (`EventFiringWebDriver` is something that comes to mind)

I will change my implementation for this pull request but we should probably use `basestring` throughout the codebase?

py/selenium/webdriver/support/page_objects/find_all_by.py
joshuaduffy: Cheers, a property on the FindBy/FindAllBy classes should do the trick, this allows us to set timeout values per locator.

py/selenium/webdriver/support/page_objects/find_all_by.py
Dude-X: The simplest thing to do is:
```python
try:
    basestring
except NameError:
    basestring = str
```
It would be nice to use the six library but that would require a big update to the existing code.

py/selenium/webdriver/support/page_objects/find_all_by.py
Dude-X: I think a timeout exception is fine, so that the end user can write their own handling. Otherwise, we are making more decisions on behalf of the end user, which means you will have to put more custom code to change behavior.

py/selenium/webdriver/support/page_objects/find_all_by.py
joshuaduffy: Done.

py/selenium/webdriver/support/page_objects/find_all_by.py
joshuaduffy: What if a user uses a FindBy or FindAllBy outside the context of another class? We cannot display such an error.

py/selenium/webdriver/support/page_objects/find_by.py
joshuaduffy: Done.

joshuaduffy:
I've added documentation, better Python 3 support, a configurable time-out option and rebased.

py/selenium/webdriver/support/page_objects/page_object.py
```diff
@@ -0,0 +1,67 @@
...
+
+            class SeleniumHome(PageObject):
+                uri = 'http://www.seleniumhq.org/'
+
+                search_box = FindBy(By.ID, 'q')
+                submit_button = FindBy(By.ID, 'submit')
+
+                def __init__(self, driver):
+                    super().__init__(driver) #  Python 3 __super__
+                    self.driver.get(self.uri)
```
Dude-X: One small nit, but this is major. The init function of a page object should not navigate to the page upon creating the page object! That would be quite annoying if instantiating subclasses or even instance of this class causes the driver to jump around different pages. The help example should have a separate function to navigate to the page.

joshuaduffy: Good catch, done.

Dude-X:
Looks good to me. I'll see if the core committers will add this. Probably will happen next week if there's no objection.

py/selenium/webdriver/support/page_objects/page_object.py
```diff
@@ -0,0 +1,71 @@
...
+# specific language governing permissions and limitations
+# under the License.
+from selenium.common.exceptions import WebDriverException
+from selenium.webdriver.remote.webdriver import WebDriver
+
+class PageObject(object):
+    def __init__(self, driver):
+        """
+        Creates a new instance of a PageObject and takes a WebDriver instance upon initialisation.
+
```
cgoldberg: could you add a brief snippet explaining the PageObject pattern, or a link to view more info?  People reading the generated docs might not be familiar with it.

joshuaduffy: See new commit for a link in the documentation along with an overview of how the `PageObject` works.

joshuaduffy:
I can fix the commit history and rebase everything as and when.

I'm not too sure who to direct this too, but @p0deje @barancev @lukeis would you please be able to take a look at this?

As @Dude-X said it would bring parity against C#/Java for the core of selenium to support for the page object pattern out of the box for Python.

p0deje:
I'd say it's better not to support one more thing in core since there are multiple libraries providing similar functionality and it's better to support diversity in open source. However, I'm not a Python bindings maintainer, so even though I'd not want anything like this in Ruby part, it's not up to me to decide.

AutomatedTester:
Hi @joshuaduffy 

I have had a look over this and firstly want to thank you for the effort that you have put into this. Originally I didnt want to have something in like this because I felt, like @jarib and I am guessing @p0deje , that Python as a dynamic language didnt really need this.

I won't be accepting this PR because if we add it in, people will start being dogmatic about it and I don't think the Selenium project should be taking that responsibility. Having helped Mozilla's WebQA team get started on best practises and seen how they grown with it and never needed anything like this, it makes me more inclined to not accept this.

Thank you for the great amount of work and I hope to see more PRs from you in the future

joshuaduffy:
Hey @AutomatedTester

No worries I understand, at the end of the day its more functionality that would sit under code, which if I was yourselves, I would like to avoid as much as possible on a mature project.

I can definitely think of some compelling arguments as to why the page object pattern is useful when writing in a dynamic language after working on some very large projects.

I'm actively watching the issues tab, since I usually use selenium on a day to day basis, so don't be surprised if you see a PR from me again.

Thanks.

PR #1728, opened 2016-03-02
Remove unneccessary 'throws' annotations

ghost:

If you use the addExtension methods you have to handle an IOException although it cannot be thrown.
The constructors of FileExtension/ClasspathExtension do not throw IOExceptions. So the 'throws' annotation of the method addExtension is not neccessary.


lukeis:
merged as 40c203384ff58f1881431740eca6058221c7ba15

PR #1736, opened 2016-03-03
Clean the anonymous profiles timely

xmhubj:
Clean the anonymous profiles in Temp timely

Anonymous profiles generated by the method `FirefoxProfile.toJson()`
are not cleaned after the conversion. This causes the temporary files
keep growing overtime and consume disk space.

They also consume the system memory because the file names are kept
in the variable `temporaryFiles` of `TemporaryFilesystem` object,
which can only be cleaned by JVM application shutdown hooks when
the JVM terminated.

The fix proposed cleans the anonymous profile at each `toJson`
conversion. Thus, it can clean the anonymous profiles timely.


barancev:
Sorry, I could not apply this patch due to merge conflicts, but I've implemented this feature in a more robust manner (using try-finally) and added regression tests too, see 738b060b6081aa20035b1d5774564d5d0a3197cc

Thank you for pinning down this issue!

xmhubj:
like your patch!

PR #1794, opened 2016-03-13
Fix test case report text corruption if encoding differen from ISO_8859_1

swarmshine:
- [X] I verify that I have signed the [Contributor License Agreement](https://github.com/SeleniumHQ/selenium/blob/master/CONTRIBUTING.md#step-6-sign-the-cla)
# Summary

If test case contains non ISO_8859_1 characters html report could not display them correctly.
# How to reproduce:
## scenario:

`open http://ru.wikipedia.org/wiki/Selenium`
`assertText id=content *инструмент*`
## target page:

target-page
## Expected html report result:

report-good
## Actual html report result:

encoding-wrong
## Description

Problem:
If test case contains non ISO_8859_1 characters html report could not display them correctly.

How problem was solved:
1. By specifying file encoding for output and adding `<meta content="text/html;charset=ENCODING" http-equiv="content-type">` to html output header.
2. Current version of `selenium-testrunner.js` uses standard html POST method by `form.submit()`. It uses `application/x-www-form-urlencoded` content type. Unfortunately, `form.submit()` do not attach charset to content type. On server side jetty waits for charset and if it could not find it – uses default ISO_8859_1. 
`form.submit()` was replaced by ajax post with fixed content type header. XMLHttpRequest allows to modify content type header.
`Content-Type: application/x-www-form-urlencoded` 
was changed to
`Content-Type: application/x-www-form-urlencoded;charset=UTF-8`

We can change default UTF-8 report encoding by setting `org.openqa.selenium.server.htmlrunner.result.encoding` system property

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/1794)

### DIFF
--- a/java/server/src/org/openqa/selenium/server/htmlrunner/HTMLLauncher.java
+++ b/java/server/src/org/openqa/selenium/server/htmlrunner/HTMLLauncher.java
@@ -32,8 +32,9 @@ import org.openqa.selenium.server.browserlaunchers.BrowserOptions;
 import org.openqa.selenium.server.browserlaunchers.Sleeper;
 
 import java.io.File;
-import java.io.FileWriter;
+import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.OutputStreamWriter;
 import java.net.URLEncoder;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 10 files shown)


swarmshine:
@samitbadle , could you please review this pull request https://github.com/SeleniumHQ/selenium/pull/1794 or suggest other team mate who can help with that? We created patch 1 month ago. It fixes two issues. Guys from irc selenium chat sad you can probably help me with that.

barancev:
I'm sorry that we have not applied this PR in time, this would make Selenium HtmlRunner more user friendly. Fortunately we took encoding issues into account when we implemented the new (WebDriver based) version of the runner, and it generates reports in UTF encoding.

PR #1797, opened 2016-03-14
rb: Extract service into shared superclass

p0deje:
An attempt to share service implementation between Chrome/Edge/IE/Marionette/PhantomJS to simplify code and make it easier to maintain. Basically do the same thing that is done in Python bindings.
- [x] Refactor Chrome service to use superclass
- [x] Refactor Edge service to use superclass
- [x] Refactor Marionette service to use superclass
- [x] Refactor PhantomJS service to use superclass
- [x] Refactor IE server to use superclass (if possible)
- [x] Run Chrome tests to see if everything is fine
- [x] Run Edge tests to see if everything is fine
- [x] Run Marionette tests to see if everything is fine
- [x] Run PhantomJS tests to see if everything is fine
- [x] Run IE tests to see if everything is fine (ran IE9 on Windows 7)
- [x] Document all the API changes (and deprecate)
- [x] Document all the behaviour changes
- [x] Fix unit tests

@titusfortner There is still a lot to be done here, but it would be great to hear what you think about it (and should I proceed with it at all).

### DIFF
--- a/rb/lib/selenium/webdriver/edge/service.rb
+++ b/rb/lib/selenium/webdriver/edge/service.rb
@@ -25,7 +25,7 @@ module Selenium
       # @api private
       #
 
-      class Service
+      class Service < WebDriver::Service
         DEFAULT_PORT = 17556
 
         private


titusfortner:
Oh man, this is great. This was definitely on my list to do, since what we have now is kind of a mess and really doesn't need to be. All of the services should essentially be doing the same things, and I think we need to be more explicit about what parameters and capabilities are supported. I'm not going to get a chance to dig into this and test things until after our v3.0RC updates, but it would be great to get this into our initial release of 3.0.

p0deje:
@titusfortner Coo, I'll proceed with the refactoring and keep you in sync.

p0deje:
Well, if we can get it into 3.0, then I won't deprecate private API methods and just remove them.

p0deje:
@titusfortner I've finished this refactoring and all the specs (both unit and integration) are now passing. I'll push the changes tomorrow unless you spot some problem in the code.

p0deje:
For some reason PhantomJS fails to download on Travis, thus its specs are failing. I'll try to fix it in the different PR and merge this regardless of this issue.

PR #1832, opened 2016-03-16
fix umlauts encoding

computerlyrik:

### DIFF
--- a/java/server/src/org/openqa/selenium/server/htmlrunner/HTMLTestResults.java
+++ b/java/server/src/org/openqa/selenium/server/htmlrunner/HTMLTestResults.java
@@ -46,7 +46,9 @@ public class HTMLTestResults {
   private final HTMLSuiteResult suite;
 
   private static final String HEADER = "<html>\n" +
-      "<head><style type='text/css'>\n" +
+      "<head>\n"+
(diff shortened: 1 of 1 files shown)


java/server/src/org/openqa/selenium/server/htmlrunner/HTMLTestResults.java
```diff
@@ -46,7 +46,9 @@
   private final HTMLSuiteResult suite;
 
   private static final String HEADER = "<html>\n" +
-      "<head><style type='text/css'>\n" +
+      "<head>\n"+
+      "<meta charset=\"UTF-8\">\n"+
```
thiagofigueiro: Is there a recommended quote style for attributes?  I see single and double here, double further down.

swarmshine:
Fixing HEADER in report is not enough in common scenario.
Could you please check out Pull Request #1794
We should fix both report encoding and also HTTP Request parsing with correct encoding.

barancev:
I'm sorry that we have not applied this PR in time, this would make Selenium HtmlRunner more user friendly. Fortunately we took encoding issues into account when we implemented the new (WebDriver based) version of the runner, and it generates reports in UTF encoding.

PR #1848, opened 2016-03-19
Replace Throwables.propagate with equivalent.

joshbruning:
Where safe, replace Throwables.propagate with
equivalent behavior: either "throw e" or
"throw new RuntimeException(e)". This is in
preparation for Guava owners to delete this
method. Note that not all instances of
Throwables.propagate have been removed: those
which could possibly result in the change of
type of the resulting exception.


lukeis:
merged as e7c6b9129e0b3fbe9f616e4b59acb8c96dd0bd2d

PR #1863, opened 2016-03-22
Initialize Alert object by calling alert.text

### LINKED ISSUE #1822: [Python webdriver] switch_to.alert does not call alert.text
## Meta -

OS: any
Selenium Version: Webdriver
Browser: Any
Language: Python
## Expected Behavior -

driver.switch_to.alert should initialize the Alert object by calling alert.text for feature parity with Java.
## Actual Behavior -

Alert object with driver instance is created, but no Alert is actually received from server because no command is executed, so no exceptions are thrown.
## Steps to reproduce -

```
try:
    alert = driver.switch_to.alert
except NoAlertPresentException:
    print 'We never get here as alert is not requested from server'
```

rouke-broersma:

this is how it's done in the Java client, and makes sense as you would want to fail, or have your alert reference as early as possible.

Fixes #1822

### DIFF
--- a/py/selenium/webdriver/support/expected_conditions.py
+++ b/py/selenium/webdriver/support/expected_conditions.py
@@ -262,7 +262,6 @@ class alert_is_present(object):
     def __call__(self, driver):
         try:
             alert = driver.switch_to.alert
-            alert.text
             return alert
         except NoAlertPresentException:
             return False


lukeis:
personally i don't like the java client approach and prefer how python does it by not forcing the user to make an extra api call if they don't want to. Also, if you really want to be in line with the java client, then text should be a private variable in the Alert class that is cached, instead of fetched every time (since as is with your change, there would always be two api calls to get the alert text if you wanted to get the text)

Also, this change (if it is indeed accepted) would need to update the ExpectedConditions class and possibly related tests-
https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/support/expected_conditions.py#L264-L266


but in general I'm :-1: on this change

Mobrockers:
No api call is made at all though, currently, which is just plain weird. You're asking for the alert, and you're getting an empty object back. This does not make sense to me. If I ask for an alert, and there is no alert, I would expect an exception.

I cannot find where Java caches the alert text in the alert, I have based my approach on:
https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java#L1065-1096

I do agree that I should not have submitted the pr before updating the ExpectedConditions class and checking if any tests need updating, I apologize for that.

lukeis:
looks like that changed a while ago in java:
https://github.com/SeleniumHQ/selenium/commit/4d3f89c5937ce2f0ff4a85a9b39c17b3c12d8134

Mobrockers:
I would argue initializing alert is also more in line with the switch_to api in general. All other switch_to calls execute a server command.

Mobrockers:
Also I have checked all the tests and cannot find any test that needs a change for this pr. Please correct me if I'm wrong.

lmtierney:
Merging to align with other bindings, thank you!

PR #1881, opened 2016-03-25
Save the promise before dispatchEvent().

johnjbarton:
- [ x] By placing an `X` in the preceding checkbox, I verify that I have signed the [Contributor License Agreement](https://github.com/SeleniumHQ/selenium/blob/master/CONTRIBUTING.md#step-6-sign-the-cla)

The execute() API is async, it returns a promise. The FirefoxDomExecutor is synchronous,
it 'sends' the command with dispatchEvent() and receives the response in the onResponse_
handler.  Since we null the pendingCommand_ in the onResponse_ handler, we must save
the the promise needed by the API before calling dispatchEvent().

### DIFF
--- a/javascript/webdriver/firefoxdomexecutor.js
+++ b/javascript/webdriver/firefoxdomexecutor.js
@@ -129,9 +129,15 @@ webdriver.FirefoxDomExecutor.prototype.execute = function(command) {
   event.initEvent(webdriver.FirefoxDomExecutor.EventType_.COMMAND,
       /*canBubble=*/true, /*cancelable=*/true);
 
+  // The API for execute() is async, we need to promise the result.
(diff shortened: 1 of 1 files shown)


juangj:
Wow. I guess this was always broken?

johnjbarton:
I believe recent API change from callbacks to promises broke this.

johnjbarton:
See the problem was probably introduced here:
https://github.com/SeleniumHQ/selenium/commit/0fa587daec4605adc688e1f736fb9564b37fd044

juangj:
Merged in 9d438c2ae38c49cd136f3197e13e99eb441834ee

juangj:
Thanks for the fix! So the answer was, basically, yes this has been broken since the last commit. I wonder why we never noticed.

PR #1889, opened 2016-03-28
Use simple Request class name.

joshbruning:
Use simple Request class name rather than the
fully-specified class name (that is, including
the package name) as it is already imported.


ddavison:
merged in as 7bea19f.  thanks!

PR #1904, opened 2016-03-31
Adding a space when encoding ids starting with digits

rdbaron:

escapeCss() references the following:
https://drafts.csswg.org/cssom/#serialize-an-identifier

> To escape a character as code point means to create a string of "\"
> (U+005C), followed by the Unicode code point as the smallest possible
> number of hexadecimal digits in the range 0-9 a-f (U+0030 to U+0039 and
> U+0061 to U+0066) to represent the code point in base 16, _followed by a
> single SPACE (U+0020)._

Currently the digit is being encoded, but no SPACE is being added.

Adding the SPACE.

Signed-off-by: Bob Baron rdbaron@gmail.com


jleyba:
Merged in c554cf3b7ee1302a5ad4e830e6cdc4e418906287

PR #1912, opened 2016-04-04
Fail if an error is not thrown.

joshbruning:
Fix test that intends to check that an
error is thrown but doesn't. fail()
throws an Error which was previously
ignored under the same condition that
the expected Error was ignored.


lukeis:
merged as ae3e1bfebf16e2f8380636636a132823316d3d51

PR #1913, opened 2016-04-04
Emphasize predicate failure over timeout details.

joshbruning:
Rewording the FluentWait error message to emphasize
the predicate failure over timeout details reduces
the eagerness of engineers to immediately jump to
the conclusion that they just need to wait _longer_
(must be some random WebDriver flakiness/timeout
issue) which is not usually the appropriate fix -
rather than think about the actual cause of the
predicate failure and conceptually group similar
failures appropriately.

### DIFF
--- a/java/client/test/org/openqa/selenium/support/ui/FluentWaitTest.java
+++ b/java/client/test/org/openqa/selenium/support/ui/FluentWaitTest.java
@@ -172,7 +172,8 @@ public class FluentWaitTest {
   @Test
   public void timeoutMessageIncludesCustomMessage() {
     TimeoutException expected = new TimeoutException(
-      "Timed out after 0 seconds: Expected custom timeout message");
+        "Expected condition failed: Expected custom timeout message "
+        + "(tried for 0 second(s) with 500 MILLISECONDS interval)");
 
     when(mockClock.laterBy(0L)).thenReturn(2L);
     when(mockCondition.apply(mockDriver)).thenReturn(null);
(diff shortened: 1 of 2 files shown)


java/client/test/org/openqa/selenium/support/ui/FluentWaitTest.java
```diff
@@ -172,7 +172,8 @@ public void timeoutMessageIncludesLastIgnoredException() {
   @Test
   public void timeoutMessageIncludesCustomMessage() {
     TimeoutException expected = new TimeoutException(
-      "Timed out after 0 seconds: Expected custom timeout message");
+        "Could not match predicate: Expected custom timeout message "
```
lukeis: I think many users will be confused by the term 'predicate', since it's an internal detail of how expected conditions are constructed.

Can we change that bit back to be "Timed out, %s (tried for %d ..."

joshbruning: What do you think about "Expected condition failed: " ? Would like to keep as much focus on the condition as possible.

lukeis: that's better :)

joshbruning: Fixed - thanks!

lukeis:
did you meant to include the changes to the prebuilt files?

joshbruning:
No. :-( I'll be back in the office on Monday.

joshbruning:
Back - everything should be fine and dandy now.

lukeis:
merged as 1e39027d0c846baf3dbd977d162821782a621417

PR #1918, opened 2016-04-04
check service process exit code while waiting for opened port

rutsky:

If process immediately fails (e.g. due to missing shared library) previously Sevice waited for about 30 seconds before returning `Can not connect to the Service` error.

Now if process fails during waiting for opened port, this will be explicitly reported.

Also `time.sleep(1)` was moved to the end of cycle to remove unneeded last one second sleep without actual check.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/1918)

### DIFF
--- a/py/selenium/webdriver/common/service.py
+++ b/py/selenium/webdriver/common/service.py
@@ -82,10 +82,15 @@ class Service(object):
                 )
         count = 0
         while not self.is_connectable():
-            count += 1
-            time.sleep(1)
+            # Check is process finished with non-zero exit code.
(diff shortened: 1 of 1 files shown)


cgoldberg:
I do wish that this area of code was more pythonic... rather than incrementing the count each time through the loop and breaking when it equals 30, I'd rather see something based on `for i in range(30)`.  But that's unrelated to to this PR.

I'm :+1:  on merging this.

AutomatedTester:
A similar patch appears to have been added already

PR #1952, opened 2016-04-09
Driver command "[Get]/status" throw exception "Session id not found" with Java WebDriver Client

### LINKED ISSUE #1465: [Java][Client]Driver command "[Get]/status" required session id
Step to reproduce:
1. Start WebDriver server.
2. Execute command "[Get]/status" => throw exception "Session id not found".

I see this command in this document which doesn't require Session id ("https://github.com/seleniumhq/selenium/wiki/jsonwireprotocol")

Please help me to check this issue

Evit15:

This issue doesn't happen on C# WebDriver Client.
On document of Selenium dose not also require session id for this command (https://github.com/seleniumhq/selenium/wiki/jsonwireprotocol#status)

Fixes #1465

### DIFF
--- a/java/client/src/org/openqa/selenium/remote/HttpCommandExecutor.java
+++ b/java/client/src/org/openqa/selenium/remote/HttpCommandExecutor.java
@@ -21,6 +21,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static org.openqa.selenium.remote.DriverCommand.GET_ALL_SESSIONS;
 import static org.openqa.selenium.remote.DriverCommand.NEW_SESSION;
 import static org.openqa.selenium.remote.DriverCommand.QUIT;
+import static org.openqa.selenium.remote.DriverCommand.STATUS;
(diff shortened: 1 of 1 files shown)


lukeis:
"/status" may be in the JSONWP document, but it's not in the [W3C spec](https://w3c.github.io/webdriver/webdriver-spec.html).

It is purely an endpoint for a server and not for a driver session. It should be treated differently, similarly to how we do in other code:
https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/remote/service/DriverService.java#L172-L180

We don't expose any api that calls the execute method with that command and I would recommend not trying to extend your library to do so, since not all remote ends are required to implement '/status' (for example marionette does not). Rather your library can fetch the remote url from the driver instance if need be and then make your own http request (as shown in the previous link).

As such, I'm not inclined to accept this patch as we don't intend to support a driver session being able to make a call to '/status'.

PR #1958, opened 2016-04-11
Fix CoveringUpSauceErrorsRule: the original exception was lost

alb-i986:
When re-creating the driver fails, throw the original exception, and expose the exception occurred while re-creating the driver as a suppressed exception.

Plus, extract to local var `t.getMessage()`.


lukeis:
merged as f426cde5f40ae3687bb0eb1c47ef5d77ddcc9a53

PR #1983, opened 2016-04-18
Use process.poll() to check if the browser is terminated.

gunesacar:

FirefoxBinary's `kill()` method does the following check before
attempting to kill the browser process:

```
if self.process:
    self.process.kill()
    self.process.wait()
```

However, checking if self.process is True does not guarantee that the
process is actually running. We get the following exception if we call
quit() on FirefoxDriver when the Firefox process is already terminated:
`OSError: [Errno 3] No such process`

With this PR, the kill method first checks if process.poll() returns
None before attempting to kill the process.

Note that process.poll() is already used in _wait_until_connectable
method to check if the browser process has terminated.

Add a test to demonstrate the missing check and the fix. With the
current code we get the following error - which gets fixed with this PR:

```
    def test_that_we_can_quit_when_firefox_is_terminated(self):
        self.webdriver = webdriver.Firefox()
        self.webdriver.binary.kill()
>       self.webdriver.quit()

[snip]/selenium/test/selenium/webdriver/firefox/ff_launcher_tests.py:46:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
[snip]/selenium/webdriver/firefox/webdriver.py:123: in quit
    self.binary.kill()
[snip]/selenium/webdriver/firefox/firefox_binary.py:77: in kill
    self.process.kill()
/usr/lib/python2.7/subprocess.py:1556: in kill
    self.send_signal(signal.SIGKILL)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <subprocess.Popen object at 0x7f86918dfb50>, sig = 9

    def send_signal(self, sig):
        """Send a signal to the process
                """
>       os.kill(self.pid, sig)
E       OSError: [Errno 3] No such process
```

### DIFF
--- a/py/selenium/webdriver/firefox/firefox_binary.py
+++ b/py/selenium/webdriver/firefox/firefox_binary.py
@@ -72,7 +72,7 @@ class FirefoxBinary(object):
 
         This is useful when the browser is stuck.
         """
-        if self.process:
+        if self.process and self.process.poll() is None:
             self.process.kill()
             self.process.wait()
 
--- a/py/test/selenium/webdriver/firefox/ff_launcher_tests.py
+++ b/py/test/selenium/webdriver/firefox/ff_launcher_tests.py
@@ -40,6 +40,11 @@ class FirefoxLauncherTests (unittest.TestCase):
         self.webdriver2.quit()
         self.webdriver3.quit()
 
+    def test_that_we_can_quit_when_firefox_is_terminated(self):
+        self.webdriver = webdriver.Firefox()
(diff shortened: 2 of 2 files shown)


AutomatedTester:
This is handled by Geckodriver now so we don't need to check it

PR #1987, opened 2016-04-19
Ruby - `mouseMoveTo` should send only Fixnum coordinates

ShockwaveNN:

I'm not sure why, but if I send coordinates like this to `mouseMoveTo`:
`@driver.action.move_to(element, 330.1, 13).click.perform`
it click in one part of page, wrong part
But if i send
`@driver.action.move_to(element, 330, 13).click.perform`
it works fine.

Maybe it somewhat  depends on http call, that performs - they looks like

```
"{"element":"0.14091974799732143-9","xoffset":330.1,"yoffset":13}"
"{"element":"0.23202931091340706-9","xoffset":330,"yoffset":13}"
```

 and fraction part somehow messed up requests.

Anyhow, I don't see a reason to send coordinates with fraction part - coordinates could be only whole number.

### DIFF
--- a/rb/lib/selenium/webdriver/common/action_builder.rb
+++ b/rb/lib/selenium/webdriver/common/action_builder.rb
@@ -253,7 +253,7 @@ module Selenium
 
       def move_to(element, right_by = nil, down_by = nil)
         if right_by && down_by
-          @actions << [:mouse, :move_to, [element, right_by, down_by]]
+          @actions << [:mouse, :move_to, [element, Integer(right_by), Integer(down_by)]]
         else
           @actions << [:mouse, :move_to, [element]]
         end
@@ -281,7 +281,7 @@ module Selenium
       #
 
       def move_by(right_by, down_by)
-        @actions << [:mouse, :move_by, [right_by, down_by]]
+        @actions << [:mouse, :move_by, [Integer(right_by), Integer(down_by)]]
         self
       end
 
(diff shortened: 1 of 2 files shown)


p0deje:
Documentation for `ActionBuilder#move_to` and `ActionBuilder#move_by` accept integers, so I guess it's better to use coercion there:

```diff
diff --git a/rb/lib/selenium/webdriver/common/action_builder.rb b/rb/lib/selenium/webdriver/common/action_builder.rb
index 7f8ebae..80a911d 100644
--- a/rb/lib/selenium/webdriver/common/action_builder.rb
+++ b/rb/lib/selenium/webdriver/common/action_builder.rb
@@ -253,7 +253,7 @@ module Selenium

       def move_to(element, right_by = nil, down_by = nil)
         if right_by && down_by
-          @actions << [:mouse, :move_to, [element, right_by, down_by]]
+          @actions << [:mouse, :move_to, [element, Integer(right_by), Integer(down_by)]]
         else
           @actions << [:mouse, :move_to, [element]]
         end
@@ -281,7 +281,7 @@ module Selenium
       #

       def move_by(right_by, down_by)
-        @actions << [:mouse, :move_by, [right_by, down_by]]
+        @actions << [:mouse, :move_by, [Integer(right_by), Integer(down_by)]]
         self
       end
```

It will also be consistent with Python bindings which do `int()` on arguments.

Also, it would be great to add specs for this change.

ShockwaveNN:
@p0deje all done

p0deje:
It seems strange to me that we have changes in `#move_to` and `#move_by` but a spec for `#drag_and_drop`. Can we instead add two specs for modified methods.

And maybe it would be great to squash the commit into one.

ShockwaveNN:
@p0deje done

p0deje:
Thanks!

Pushed in 2bb9c0b9f2d9da3050f16fa7b7bbbdcfef23021a.

PR #2031, opened 2016-04-29
Fix SessionCleaner dying off due to uncaught exception.

zmokhtar:

Because the run() method does not catch any exceptions, throwing an unchecked exception in checkExpiry propagates up and causes the SessionCleaner thread to die off.  Once this happens, browser sessions will no longer be closed due to their timeout.

This issue is described here:
https://github.com/seleniumhq/selenium-google-code-issue-archive/issues/7248

### DIFF
--- a/java/server/src/org/openqa/selenium/remote/server/SessionCleaner.java
+++ b/java/server/src/org/openqa/selenium/remote/server/SessionCleaner.java
@@ -25,6 +25,7 @@ import org.openqa.selenium.remote.server.log.LoggingManager;
 import org.openqa.selenium.remote.server.log.PerSessionLogHandler;
 import org.openqa.selenium.support.events.EventFiringWebDriver;
 
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 class SessionCleaner extends Thread {   // Thread safety reviewed
@@ -115,7 +116,8 @@ class SessionCleaner extends Thread {   // Thread safety reviewed
           try {
             deleteSession.call();
           } catch (Exception e) {
-            throw new RuntimeException(e);
+            log.log(Level.WARNING, "Could not delete session " + session.getSessionId(), e);
+            continue;
           }
         }
 


java/server/src/org/openqa/selenium/remote/server/SessionCleaner.java
```diff
@@ -115,7 +116,8 @@ void checkExpiry() {
           try {
             deleteSession.call();
           } catch (Exception e) {
-            throw new RuntimeException(e);
+            log.log(Level.WARNING, "Could not delete session " + session.getSessionId(), e);
+            continue;
```
lukeis: I don't think it should 'continue' here, the Session will not be removed and it will perpetually try to delete this session. It should fall through, but the call to `driverSessions.deleteSession(sessionId);` will likely also need to be wrapped in a try/catch.

zmokhtar: In theory, that sounds like it might be a good idea, but in practice the exception is caused by the session already being closed and deleted. I have this running in production on a long running selenium server and I have never seen the log fire twice on the same session, meaning by the time the loop comes back around, the session has been cleaned out by some other part of the code.  I would rather repeatedly try to delete the session than assume that things are ok and remove this from the active sessions list. It is of utmost importance that we not leak browsers.

gouchaoer:
perfect！
i have the session clean problem too.
in my case, i use selenium to drive a lot of phantomjs.
but many phantonjs session just can not delete, which lead to memory leak.(if i quit the selenium-server-alone, the all phantomjs sessions exit...so obvously it is selenium's problem)

i clone repo and build master branch, the session clean problem got solved!

i suggest that selenium release a version for this big bug fix!

gouchaoer:
I think this commit should also cherry-pick to branch origin/leg-rc
you know this bug got fixed in 2016/5/2 and merged into master branch(but not 2.x branch). So in 2.53.1 release this bug still exists.

lukeis:
3.0 beta 2 has been released and contains this change. We won't be releasing any more 2.53.X, please use the latest 3.0 release

PR #2041, opened 2016-05-02
Find visible elements

remcowesterhoud:

Adds an expected conditions for finding visible elements as list. Right now this is only possible for the presence of elements.

### DIFF
--- a/py/test/selenium/webdriver/common/webdriverwait_tests.py
+++ b/py/test/selenium/webdriver/common/webdriverwait_tests.py
@@ -72,7 +72,7 @@ class WebDriverWaitTest(unittest.TestCase):
 
     def testShouldFailToFindElementsWhenExplicitWaiting(self):
         self._loadPage("dynamic")
-        with self.assertRaises(TimeOutException):
+        with self.assertRaises(TimeoutException):
             elements = WebDriverWait(self.driver, 0.7).until(EC.presence_of_all_elements_located((By.CLASS_NAME, "redbox")))
             
     def testShouldWaitUntilAtLeastOneVisibleElementsIsFoundWhenSearchingForMany(self):
@@ -89,7 +89,7 @@ class WebDriverWaitTest(unittest.TestCase):
 
     def testShouldFailToFindVisibleElementsWhenExplicitWaiting(self):
         self._loadPage("hidden_partially")
-        with self.assertRaises(TimeOutException):
+        with self.assertRaises(TimeoutException):
             elements = WebDriverWait(self.driver, 0.7).until(EC.visibility_of_all_elements_located((By.CLASS_NAME, "redbox")))
 
     def testShouldWaitOnlyAsLongAsTimeoutSpecifiedWhenImplicitWaitsAreSet(self):


lukeis:
Can you also add a test for this?

https://github.com/seleniumhq/selenium/blob/master/py/test/selenium/webdriver/common/webdriverwait_tests.py

The pages that these tests are loading are here:
https://github.com/seleniumhq/selenium/blob/master/common/src/web/

py/selenium/webdriver/support/expected_conditions.py
```diff
@@ -101,6 +101,22 @@ def __init__(self, locator):
...
+    locator is used to find the element
+    returns the list of WebElements once they are located
+    """
+    def __init__(self, locator):
+        self.locator = locator
+
+    def __call__(self, driver):
+        visible_elements = []
+        for element in _find_elements(driver, self.locator):
+            if _element_if_visible(element) is not False:
```
cgoldberg: I think this line would be better written as:
`if not _element_if_visible(element):`

py/selenium/webdriver/support/expected_conditions.py
```diff
@@ -101,6 +101,22 @@ def __init__(self, locator):
...
+class visibility_of_all_elements_located(object):
+    """ An expectation for checking that there is at least one element visible
+    on a web page.
+    locator is used to find the element
+    returns the list of WebElements once they are located
+    """
+    def __init__(self, locator):
+        self.locator = locator
+
+    def __call__(self, driver):
```
cgoldberg: can't you use a list comprehension to replace your entire `__call__` method?

```
def __call__(self, driver):
	return [element for element in _find_elements(driver, self.locator) if _element_if_visible(element)]
```

remcowesterhoud:
I've added the tests and replaced the call method with the list comprehension @cgoldberg mentioned.

py/test/selenium/webdriver/common/webdriverwait_tests.py
```diff
@@ -78,6 +78,27 @@ def testShouldFailToFindElementsWhenExplicitWaiting(self):
...
+        add_hidden.click()
+
+        elements = WebDriverWait(self.driver, 2).until(EC.visibility_of_all_elements_located((By.CLASS_NAME, "redbox")))
+        self.assertTrue(len(elements) == 2)
+
+    def testShouldFailToFindVisibleElementsWhenExplicitWaiting(self):
+        self._loadPage("hidden_partially")
+        try:
+            elements = WebDriverWait(self.driver, 0.7).until(EC.visibility_of_all_elements_located((By.CLASS_NAME, "redbox")))
+        except TimeoutException as e:
```
cgoldberg: you could assert the exception is raised:
https://docs.python.org/2.7/library/unittest.html#unittest.TestCase.assertRaises

remcowesterhoud:
@cgoldberg I've changed it. I did it this way because I saw it being done in other testcases the same way.

lukeis:
awesome, thanks for the patch! (and the reworks :) )

PR #2052, opened 2016-05-04
visibility_of_all implies it only returns elements if all visible

rouke-broersma:

This is not what the function does.
visibility_of_any in my opinion correctly conveys the usage of this expected condition.

### DIFF
--- a/py/test/selenium/webdriver/common/webdriverwait_tests.py
+++ b/py/test/selenium/webdriver/common/webdriverwait_tests.py
@@ -84,13 +84,13 @@ class WebDriverWaitTest(unittest.TestCase):
         add_visible.click()
         add_hidden.click()
 
-        elements = WebDriverWait(self.driver, 2).until(EC.visibility_of_all_elements_located((By.CLASS_NAME, "redbox")))
+        elements = WebDriverWait(self.driver, 2).until(EC.visibility_of_any_elements_located((By.CLASS_NAME, "redbox")))
         self.assertTrue(len(elements) == 2)
 
     def testShouldFailToFindVisibleElementsWhenExplicitWaiting(self):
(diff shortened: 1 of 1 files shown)


remcowesterhoud:
If you're gonna change the name of this class you should also change the tests that use it.
You can find those at https://github.com/SeleniumHQ/selenium/blob/master/py/test/selenium/webdriver/common/webdriverwait_tests.py

py/selenium/webdriver/support/expected_conditions.py
```diff
@@ -102,7 +102,7 @@ def __init__(self, locator):
     def __call__(self, driver):
         return _find_elements(driver, self.locator)
         
-class visibility_of_all_elements_located(object):
+class visibility_of_any_elements_located(object):
```
cgoldberg: should this return a boolean?

cgoldberg: shouldn't __call__ return a boolean?

Mobrockers: This is clearly not the author's intent and returning elements instead of a boolean is not out of line with the rest of the expected conditions.

See presence_of_element_located: https://github.com/Mobrockers/selenium/blob/f878b53d9dd66ef25444b82e50218acd36f6bc4e/py/selenium/webdriver/support/expected_conditions.py#L49
visibility_of_element_located:https://github.com/Mobrockers/selenium/blob/f878b53d9dd66ef25444b82e50218acd36f6bc4e/py/selenium/webdriver/support/expected_conditions.py#L61

etcetc

Or do you mean as a final condition if no visible elements van be found by the given locator?
In that case perhaps you're right yes.

PR #2055, opened 2016-05-04
Add filename to IOException if file copy was failed.

tirana:

I got the IOException "Could not transfer all bytes" and found that there was no glue what wrong.
filename may give a hint, at least.

### DIFF
--- a/java/client/src/org/openqa/selenium/io/FileHandler.java
+++ b/java/client/src/org/openqa/selenium/io/FileHandler.java
@@ -191,7 +191,7 @@ public class FileHandler {
       final long copied = Files.copy(from.toPath(), out);
       final long length = from.length();
       if (copied != length) {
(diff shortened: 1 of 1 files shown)


java/client/src/org/openqa/selenium/io/FileHandler.java
```diff
@@ -191,7 +191,7 @@ private static void copyFile(File from, File to, Filter onlyCopy) throws IOExcep
       final long copied = Files.copy(from.toPath(), out);
       final long length = from.length();
       if (copied != length) {
-        throw new IOException("Could not transfer all bytes.");
+        throw new IOException("Could not transfer all bytes of " + from.toPath());
```
lukeis: seems reasonable to also want to include the 'to' location?

tirana: done

PR #2060, opened 2016-05-05
[node] Fix 2059 - exec.js fails to set commandline options

### LINKED ISSUE #2059: io/exec.js fails to set commandline arguments
## Meta -
OS:  OSX 10.11.3
Selenium Version:   2.53.0 
Browser:  Firefox 46.0.1

## Expected Behavior -

Firefox should open listening with the debugger server and dhtml profile.

## Actual Behavior -

Firefox opens, but the command line options are not set. Interestingly, a `ps` shows that the command was executed with those options, but they were not applied.

## Steps to reproduce 

```js
var binary = new firefox.Binary('/Applications/Firefox.app/Contents/MacOS/firefox-bin');
binary.addArguments('--start-debugger-server 6080', '-P dhtml')
let options = (new firefox.Options()).setBinary(firefoxBinary());
const driver = new firefox.Driver(options);
```

## Suspected bug

I believe the issue is with node's child process spawn, when I switched that out for exec it worked fine.

```js
const child = require('child_process');

command = "/Applications/Firefox.app/Contents/MacOS/firefox-bin -P dhtml";
args = [ '-foreground', '--start-debugger-server 6080', '-P dhtml' ];

//fails
child.spawn(command, args)

// succeeds
child.exec(command + " " + args.join(" "))
```

jasonLaster:
This PR fixes #2059. I did not add tests because i was not sure how to test that firefox had the right profile and command line options set. Unfortunately, the failing case has the right command in `ps`.

### DIFF
--- a/javascript/node/selenium-webdriver/io/exec.js
+++ b/javascript/node/selenium-webdriver/io/exec.js
@@ -120,7 +120,7 @@ class Command {
 module.exports = function exec(command, opt_options) {
   var options = opt_options || {};
 
-  var proc = childProcess.spawn(command, options.args || [], {
+  var proc = childProcess.exec([command].concat(options.args || []).join(" "), {
     env: options.env || process.env,
     stdio: options.stdio || 'ignore'
   }).once('exit', onExit);


jleyba:
This changes the semantics of the function: instead of spawning a sub-process capable of streaming its output, it executes it in one go, returning buffered output.  See my comments on #2059 for more info.

PR #2084, opened 2016-05-11
Added ability to change wires path on Firefox driver

thaskell1:

executable_path in webdriver.py for Firefox web driver was hardcoded to use "wires" (environmental path variable). Added additional statements to allow for direct executable_path specification via "capabilities"


lmtierney:
`wires` is now `geckodriver` since this is so old... Regardless, the path will not be set through desired_capabilities

Issue #2069, opened 2016-05-09
Feature request: support phantomjs-prebuilt in selenium-webdriver


zixia:
## Meta -

OS:  Win32

Selenium Version:  2.53.2

Browser:  phantomjs, which installed by `npm install phantomjs-prebuilt`

Browser Version:  2.1.7

## Expected Behavior -

``` bash
npm install selenium-webdriver
npm install phantomjs-prebuilt
```

node js source

``` javascript
const driver = new WebDriver.Builder()
.withCapabilities(
  WebDriver.Capabilities.phantomjs()
).build()
```

should find phantomjs executable, and return success.
## Actual Behavior -

cant find phantomjs executable.

```
test\node_modules\selenium-webdriver\phantomjs.js:80
    throw Error(
    ^

Error: The PhantomJS executable could not be found on the current PATH. Please download the latest version from http://phantomjs.org/download.html and ensure it can be found on your PATH. For more information, see https://github.com/ariya/phantomjs/wiki
    at Error (native)
    ...
```
## Steps to reproduce -

step 1. open cmd window
step 2. run

``` bash
mkdir test
cd test
npm install selenium-webdriver
npm install phantomjs-prebuilt
```

step 3. then run this node js source  

```
const WebDriver = require('selenium-webdriver')
const driver = new WebDriver.Builder()
.withCapabilities(
  WebDriver.Capabilities.phantomjs()
).build()
```

step 4. webdriver said that cant find phantomjs executable.   
## Steps to quick fix -

step 1. find executable installed by phatomjs-prebuilt in node_modules.bin

```
  D:\test>dir node_modules\.bin\phantomjs*
   node_modules\.bin 的目录

  2016/05/09  11:59               335 phantomjs
  2016/05/09  11:59               212 phantomjs.cmd
```

step 2. reason: in selenium-webdriver/phantomjs.js:

``` javascript
  const PHANTOMJS_EXE =
      process.platform === 'win32' ? 'phantomjs.exe' : 'phantomjs';
```

  it only find .exe . but the phantomjs-prebuilt installed executable name is `.cmd`
step 3. fix: add `.cmd` to search `io.findPath`:

``` javascript
function findExecutable(opt_exe) {
  const PHANTOMJS_CMD = 'phantomjs.cmd'
  var exe = opt_exe || io.findInPath(PHANTOMJS_EXE, true) || io.findInPath(PHANTOMJS_CMD, true);
```

jleyba:
[...]

PR #2086, opened 2016-05-11
Add 'How to build driver with phantomjs-prebuilt'

vtange:

Added Instructions for something like this in the comments above "Find Executable" function of PhantomJS
In response to https://github.com/SeleniumHQ/selenium/issues/2069

### DIFF
--- a/javascript/node/selenium-webdriver/phantomjs.js
+++ b/javascript/node/selenium-webdriver/phantomjs.js
@@ -17,23 +17,23 @@
 
 /**
  * @fileoverview Defines a {@linkplain Driver WebDriver} client for the PhantomJS
- * web browser. By default, it is expected that PhantomJS exists in your working directory.
+ * web browser. By default, it is expected that the PhantomJS executable can be located on your [PATH](https://en.wikipedia.org/wiki/PATH_(variable))
  * 
  *  __Using a Custom PhantomJS Binary__
  * 
  * If you have PhantomJS.exe placed somewhere other than the root of your working directory,
  * you can build a custom Capability and attach the executable's location to the Capability
  * 
- * For example, if you're using the phantomjs-prebuilt module (https://www.npmjs.com/package/phantomjs-prebuilt) from npm:
+ * For example, if you're using the [phantomjs-prebuilt](https://www.npmjs.com/package/phantomjs-prebuilt) module from npm:
  * 
- * //setup custom phantomJS capability
- * var phantomjs_exe = require('phantomjs-prebuilt').path;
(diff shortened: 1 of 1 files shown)


jleyba:
You've updated the docs on a private function. If you want this notice to be included in the generated API docs, you should add a fileoverview comment to the top of the file. For example, see [selenium-webdriver/chrome.js](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/chrome.js#L18) (here are the [public docs generated](http://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/chrome.html) extracted from that file).

vtange:
OK. Moved the docs to the top of the file under @fileoverview. Ring me up if it still needs work

javascript/node/selenium-webdriver/phantomjs.js
```diff
@@ -15,6 +15,28 @@
 // specific language governing permissions and limitations
 // under the License.
 
+/**
+ * @fileoverview Defines a {@linkplain Driver WebDriver} client for the PhantomJS
+ * web browser. By default, it is expected that PhantomJS exists in your working directory.
```
jleyba: By default, it is expected that the PhantomJS executable can be located on your
    [PATH](https://en.wikipedia.org/wiki/PATH_(variable))

javascript/node/selenium-webdriver/phantomjs.js
```diff
@@ -15,6 +15,28 @@
...
+ * 
+ *  __Using a Custom PhantomJS Binary__
+ * 
+ * If you have PhantomJS.exe placed somewhere other than the root of your working directory,
+ * you can build a custom Capability and attach the executable's location to the Capability
+ * 
+ * For example, if you're using the phantomjs-prebuilt module (https://www.npmjs.com/package/phantomjs-prebuilt) from npm:
```
jleyba: For example, if you're using the [phantomjs-prebuilt](https://www.npmjs.com/package/phantomjs-prebuilt)
    module from npm:

javascript/node/selenium-webdriver/phantomjs.js
```diff
@@ -15,6 +15,28 @@
...
+ * 
+ * //setup custom phantomJS capability
```
jleyba: Indent each line in this example by at least 4 spaces so it renders as a code block:

    *    // setup a custom phantomJS capability
    *    var phantomjs_exe = require('phantomjs-prebuilt').path;

PR #2101, opened 2016-05-13
Fix capabilities created by Firefox options

davehunt:
Creating a pull request in order to get feedback from @AutomatedTester as he mentioned custom profiles were working for some people using Selenium & GeckoDriver (wires). I can't see how it could be working, but I certainly don't want to break it if I'm missing something. I'll add a couple of inline comments too to explain the reasons for the changes. I'm also not sure where the tests are for this particular feature.

### DIFF
--- a/py/selenium/webdriver/firefox/options.py
+++ b/py/selenium/webdriver/firefox/options.py
@@ -16,7 +16,6 @@
 # under the License.
 
 from selenium.common.exceptions import WebDriverException
-from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
 from selenium.webdriver.firefox.firefox_profile import FirefoxProfile
 
 
@@ -87,14 +86,10 @@ class Options(object):
 
             returns a dictionary with everything
         """
-        firefox = DesiredCapabilities.FIREFOX.copy()
-        firefox_options = self._firefox_options
+        capabilities = self._firefox_options
         if self.binary_location:
-            firefox_options["binary"] = self.binary_location
+            capabilities["binary"] = self.binary_location
         if self._profile:
-            firefox_options["profile"] = self._profile.encoded
-        firefox_options["args"] = self.arguments
-
-        firefox["requiredCapabilities"] = firefox_options
-
-        return firefox
+            capabilities["firefox_profile"] = self._profile.encoded
+        capabilities["args"] = self.arguments
+        return capabilities


py/selenium/webdriver/firefox/options.py
```diff
@@ -87,14 +86,10 @@ def to_capabilities(self):
 
             returns a dictionary with everything
         """
-        firefox = DesiredCapabilities.FIREFOX.copy()
```
davehunt: This causes existing capabilities to be overwritten when updating them from a FirefoxOptions object. It meant that I was unable to set the `marionette` capability to `True`.

py/selenium/webdriver/firefox/options.py
```diff
@@ -87,14 +86,10 @@ def to_capabilities(self):
...
-        firefox_options = self._firefox_options
+        capabilities = self._firefox_options
         if self.binary_location:
-            firefox_options["binary"] = self.binary_location
+            capabilities["binary"] = self.binary_location
         if self._profile:
-            firefox_options["profile"] = self._profile.encoded
-        firefox_options["args"] = self.arguments
-
-        firefox["requiredCapabilities"] = firefox_options
```
davehunt: This means all capabilities are within the `requiredCapabilities`, which wasn't working for me with GeckoDriver.

py/selenium/webdriver/firefox/options.py
```diff
@@ -87,14 +86,10 @@ def to_capabilities(self):
...
-
-        return firefox
+            capabilities["firefox_profile"] = self._profile.encoded
```
davehunt: GeckoDriver is expecting `firefox_profile`, but I'm not sure if this should be fixed here or there.. is there some sort of reference document for capabillties?

py/selenium/webdriver/firefox/options.py
AutomatedTester: This is a design decision. If this is not working in GeckoDriver it is a bug there or in Gecko itself. I need to pick `desiredCapabilties` or `requiredCapabilties`. Since a person is asking for stuff it needs to be required.

davehunt:
I landed a variation of this in https://github.com/SeleniumHQ/selenium/commit/e70ce5261641773e76ff239c1e0a8adbe61ab400

PR #2104, opened 2016-05-13
Interface for Select component extracted

YevheniiaSilbernagl:

This interface will allow to create custom realisation of like-select components, which actually do not have select tag inside.

### DIFF
--- a/java/client/src/org/openqa/selenium/support/ui/Select.java
+++ b/java/client/src/org/openqa/selenium/support/ui/Select.java
@@ -59,7 +59,6 @@ public class Select implements ISelect{
    * @return Whether this select element support selecting multiple options at the same time? This
    *         is done by checking the value of the "multiple" attribute.
    */
-  @Override
   public boolean isMultiple() {
     return isMulti;
   }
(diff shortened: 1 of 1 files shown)


jtymoschuk:
could somebody help me to understand what is the problem with tests?

lukeis:
It's not the tests, you have compilation issues. First you need to add an entry in the BUCK and build.desc in the same folder. Next you should make the methods on the new interface public.

jtymoschuk:
@lukeis Thank you for your support, I've fixed that issues, but is still fails. Sems like in a py module, but I didn't change py code, could you please take a look once more time?

java/client/src/org/openqa/selenium/support/ui/build.desc
```diff
@@ -31,6 +31,7 @@ java_library(name = "components",
 
 java_library(name = "elements",
   srcs = [
+    'ISelect.java',
```
lukeis: double quotes required

java/client/src/org/openqa/selenium/support/ui/Select.java
```diff
@@ -59,20 +59,23 @@ public Select(WebElement element) {
    * @return Whether this select element support selecting multiple options at the same time? This
    *         is done by checking the value of the "multiple" attribute.
    */
+  @Override
```
lukeis: can you revert all the Overrides added? it's not required for implementing interfaces, only overriding methods extended.

jtymoschuk:
@lukeis overrides removed, tests passed

PR #2121, opened 2016-05-18
Create a new suite of Python tests to be run using Tox

davehunt:
Some initial work to allow the Python tests to be run using Tox without needing to build them. Using a parameterised pytest fixture to provide the various drivers (currently just Firefox and PhantomJS) and following PEP8 and other Python good practices.

Existing tests are still run using:

``` bash
$ ./go test_py
```

New tests are run using:

``` bash
$ pip install tox
$ tox
```

To run on a specific Python version, pass `-e` and specify one or more environments. For example `tox -e py27` will run against Python 2.7.

Wiki pages would need to be updated of course to reflect how to run the new tests, and Travis CI should run the tests against as many drivers as possible for every pull request (I'm experimenting with doing so in this patch). We could also use http://www.appveyor.com/ for the Windows drivers (Internet Explorer and Edge), and our Jenkins instance. As far as I can tell these tests are not currently run anywhere, so they may not be in a passing state.

I'm okay with migrating all tests in one go so that we don't have two suites, but I wanted to get feedback from @lukeis, @AutomatedTester, and @Dude-X on my initial approach. I think it makes sense to get this running for a minimum of drivers initially, and then expand.

I have to admit that the CrazyFun/Rake stuff confuses me. We could try to get the various Python test targets to run the appropriate additional tests via `tox`, or wait until we've migrated to the new suite and change the targets as needed. To be honest I'm fine with avoiding CrazyFun for anything other than the dependencies.


Dude-X:
On my first pass, only nit is that you have a bare `except:` clause instead of using the safer `except Exception:` statement.

davehunt:
> On my first pass, only nit is that you have a bare except: clause instead of using the safer except Exception: statement.

@Dude-X that's not my code, I'm just bringing it across from the existing suite.

Dude-X:
@davehunt You're right that's the current code.

Another note, last time I tried using a yield fixture, It seemed to me that a test failure didn't execute the statements after the yield, so maybe we need a try/finally clause. I am going to verify this myself to see if that's the case.

lukeis:
:+1: from me. And yes, let's migrate the whole thing if possible, I don't want two test suites. I'll help with the crazy fun...

AutomatedTester:
👍🏼

py/tests/network.py
```diff
@@ -0,0 +1,46 @@
...
+                struct.pack('256s', ifname[:15])
+            )[20:24])
+
+def get_lan_ip():
+    try:
+        ip = socket.gethostbyname(socket.gethostname())
+    except:
+        return '0.0.0.0'
+    if ip.startswith("127.") and os.name != "nt":
+        interfaces = ["eth0","eth1","eth2","en0","en1","en2","en3","en4","wlan0","wlan1","wifi0","ath0","ath1","ppp0"]
```
cgoldberg: This will fail on most newer linux hosts.
You can't rely on a hardcoded list of possible interface names.  Now that most linux distros are using systemd, the traditional interface naming scheme isn't used anymore.

For example, on my machine (Ubuntu 16.04), the network interface for my ethernet adapter is assigned "enp4s0".  For the gory details, see: https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/

One possible solution would be to inspect `/sys/class/net` to get the local interface name (however, that will only work on Linux, not on other systems like mac osx).

.travis.yml
```diff
@@ -41,6 +41,8 @@ script:
       pip install --user virtualenv
       export PATH=$PATH:$HOME/.local/bin
       ./go py_prep_for_install_release //py:phantomjs_test:run
+      pip install tox
```
cgoldberg: I think CI is failing because this line needs to be changed to `pip install --user tox`
see: https://docs.travis-ci.com/user/languages/python#pip

py/tests/network.py
davehunt: This isn't my code, I'm just moving it from the existing location. Any improvements to this code should be separate to this patch.

.travis.yml
davehunt: Yeah, I will give that a try when I have time to return to this patch. Thanks for the tip! 👍

davehunt:
I landed an alternate version of this.

PR #2140, opened 2016-05-21
Adding Travis build badge

manoj9788:


lukeis:
I believe others have proposed this before, reasoning still stands that we don't want it on our README

#1649

travis-ci is not our main ci, it's only used for pull requests and thus this badge would never be truly accurate.

manoj9788:
I see, ok! thanks.

PR #2141, opened 2016-05-21
Remove redundant generics type

asashour:
Remove redundant generics type

### DIFF
--- a/java/client/src/org/openqa/selenium/interactions/CompositeAction.java
+++ b/java/client/src/org/openqa/selenium/interactions/CompositeAction.java
@@ -63,7 +63,7 @@ public class CompositeAction implements Action {
   }
 
   public List<Action> asList() {
-    ImmutableList.Builder<Action> builder = new ImmutableList.Builder<Action>();
+    ImmutableList.Builder<Action> builder = new ImmutableList.Builder<>();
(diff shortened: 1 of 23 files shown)


java/client/src/org/openqa/selenium/remote/Command.java
```diff
@@ -45,7 +45,7 @@ public String getName() {
   }
 
   public Map<String, ?> getParameters() {
-    return parameters == null ? new HashMap<String, Object>() : parameters;
+    return parameters == null ? new HashMap<>() : parameters;
```
joshbruning: From travis-ci:

1. ERROR in /home/travis/build/SeleniumHQ/selenium/java/client/src/org/openqa/selenium/remote/Command.java (at line 48)
	return parameters == null ? new HashMap<>() : parameters;
	       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Type mismatch: cannot convert from Map<capture#5-of ? extends Object,capture#6-of ? extends Object> to Map<String,?>

asashour: Thanks, updated

PR #2144, opened 2016-05-21
Methods which could be static.

asashour:
Methods which could be static.


lukeis:
not sure i like this style, if the method all of a sudden becomes public or we need to unit test the class, then it becomes more difficult to mock the method. In general (from bad experiences) I think things should tend to not be static.

asashour:
That's fine.

PR #2152, opened 2016-05-23
Detect early if logger will be used

asashour:
This spares unnecessary
StringBuilder creation when the logger
will not be used to log the message.


barancev:
Good catch! Applied as dcc855b227bd7f895baf6201a073188d11e77930

PR #2158, opened 2016-05-24
[Java] Better naming convention By class

manoj9788:

### DIFF
--- a/java/client/src/org/openqa/selenium/By.java
+++ b/java/client/src/org/openqa/selenium/By.java
@@ -438,7 +438,7 @@ public abstract class By {
     public List<WebElement> findElements(SearchContext context) {
       if (context instanceof FindsByCssSelector) {
         return ((FindsByCssSelector) context)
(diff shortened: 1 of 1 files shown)


jleyba:
Going to reject this PR since it introduces unnecessary churn into the code base (plus it's debatable whether the renamed variables add anything wrt clarity)

manoj9788:
@jleyba The reason  for changing is to make it same across the docs and and the parameter name. would you be wanting to show in docs as linkText for a partialLinkText ? Similarly for others ?

cgoldberg:
> The reason for changing is to make it same across 
> the docs and and the parameter name

@manoj9788 
I can't figure out you mean by that... I guess you are to trying to make the parameter names more consistent with method names, so the generated docs are more consistent?  FWIW, I think the existing names are fine, if not better

A good PR should include a clear summary of the change set, including the reason you are making the changes or a link to an issue that provides this.

PR #2189, opened 2016-05-31
[js] Turn on marionette by default when creating new Firefox Options

JohanLorenzo:
Firefox 47 will go [release on July 7th](https://wiki.mozilla.org/RapidRelease/Calendar). After that date, Marionette will be the only way for Selenium to connect. I'd propose to turn it on by default. What do you think @jleyba ? 

### DIFF
--- a/javascript/node/selenium-webdriver/README.md
+++ b/javascript/node/selenium-webdriver/README.md
@@ -10,17 +10,20 @@ Selenium may be installed via npm with
 
     npm install selenium-webdriver
 
-Out of the box, Selenium includes everything you need to work with Firefox. You
-will need to download additional components to work with the other major
-browsers. The drivers for Chrome, PhantomJS, Opera, and Microsoft's IE and Edge
-web browsers are all standalone executables that should be available on your
-[PATH](http://en.wikipedia.org/wiki/PATH_%28variable%29). The SafariDriver
+You will need to download additional components to work with the other major
+browsers. The drivers for Firefox, Chrome, PhantomJS, Opera, and Microsoft's IE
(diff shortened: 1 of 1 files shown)


jleyba:
I'm OK with this change, but I'd prefer to wait closer to the release before making the switch.

Also, the README and docs in the firefox module should be updated to mention that users need to download and extra component (you could just copy what's there for chrome and IE)

Mobrockers:
Can the marionette driver be called by selenium grid yet? What would be the browser name to start a firefox browser using marionette be? From what I saw, marionette is used by setting a desired capability flag, I don't see how that would work with starting a grid node.

jleyba:
@Mobrockers I believe the java code default to Marionette now, so you shouldn't need to do anything special. You'll need to build grid from master since Marionette was not the default in 2.53.x

Mobrockers:
So.. Does that mean using the grid for firefox will be broken for a couple months until selenium 3 comes out, unless I build from master?

titusfortner:
You can use Marionette with Selenium 2.53, and so long as your version of Firefox is 45+ then you can specify which protocol you want to use (Firefox Driver or Marionette) by toggling the desired capability.

Mobrockers:
Right, I see, by adding the driver to the path, grid will be able to use it. Will a system property, like chromedriver and iedriver have, become available at some point? I think that would be a bit more convenient/flexible in a distributed environment.

jleyba:
Can you please start a thread on selenium-developers@? This conversation has nothing to do with this pull request (a change for the JS client)

Mobrockers:
You're right, sorry!

PR #2203, opened 2016-06-02
Add test to check that elements not rendered are unclickable.

danielprejean:

I think that elements not rendered (e.g. outside the bounds of an element with overflow: hidden) should not be clickable.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/2203)


AutomatedTester:
This test is already covered by wpt and there are already similar tests in the selenium tree too

PR #2222, opened 2016-06-06
Redirect stdin in subprocess.Popen function call

ranjotsingh:

When calling the subprocess.Popen function, only stdout and stderr is redirected but not stdin.

You may encounter this issue when using pyinstaller to create an .exe (specifically in no console mode), where, for example, phantomjs will not run unless stdin is also redirected.

This modification solves the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/2222)


Dude-X:
You are redirecting stdin for all drivers; is this change necessary for all drivers? Or only specific to PhantomJS? Did you test it won't break functionality for every driver?

AutomatedTester:
Thanks for submitting this patch, I appreciate it. Unfortunately we are not going to move this patch into the tree.

PR #2237, opened 2016-06-08
W3c new session key

isonic1:

Clients that are implementing the current w3c standard cannot create new sessions for the current bindings. This PR will handle both old standard and new w3c for new sessions.

https://w3c.github.io/webdriver/webdriver-spec.html#new-session


p0deje:
Since latest selenium-webdriver implements protocol handshake, I guess we no longer need these changes.

Thank you for pull requests and sorry it took so long to take a look.

PR #2244, opened 2016-06-09
Need to shutdown a ghost driver mode phantomjs process.

TakesxiSximada:

A phantomjs of ghost driver mode of the child process was not able to shut down, because of  a send_remote_shutdown_command function of selenium.webdriver.phantomjs.service.Service does not call send_remote_shutdown function of superclass.

### DIFF
--- a/py/selenium/webdriver/phantomjs/service.py
+++ b/py/selenium/webdriver/phantomjs/service.py
@@ -64,6 +64,7 @@ class Service(service.Service):
         return "http://localhost:%d/wd/hub" % self.port
 
     def send_remote_shutdown_command(self):
+        super(Service, self).send_remote_shutdown_command()
         if self._cookie_temp_file:
             os.close(self._cookie_temp_file_handle)
             os.remove(self._cookie_temp_file)


lukeis:
phantomjs does not respond to the /shutdown url, when one tries you get

HTTP/1.1 404 Not Found
Date: Fri, 10 Jun 2016 19:45:25 GMT
Content-Type: application/json; charset=utf-8
Content-Length: 83

{"error":"unknown command","message":"GET /shutdown did not match a known command"}


Your processes are still running for another reason.

TakesxiSximada:
fmm....

This problem is not expected to occur on Windows. And this problem does not occur even posix os by the timing.

First phantomjs is about, 
When we run "npm install phantomjs", npm create a file "node_modules/.bin/phantomjs".
"node_modules/.bin/phantomjs" is the following files .

```
$ cat node_modules/.bin/phantomjs
#!/usr/bin/env node

/**
 * Script that will execute the downloaded phantomjs binary.  stdio are
 * forwarded to and from the child process.
 *
 * The following is for an ugly hack to avoid a problem where the installer
 * finds the bin script npm creates during global installation.
 *
 * {NPM_INSTALL_MARKER}
 */

var path = require('path')
var spawn = require('child_process').spawn

var binPath = require(path.join(__dirname, '..', 'lib', 'phantomjs')).path

var args = process.argv.slice(2)

// For Node 0.6 compatibility, pipe the streams manually, instead of using
// `{ stdio: 'inherit' }`.
var cp = spawn(binPath, args)
cp.stdout.pipe(process.stdout)
cp.stderr.pipe(process.stderr)
process.stdin.pipe(cp.stdin)

cp.on('error', function (err) {
  console.error('Error executing phantom at', binPath)
  console.error(err.stack)
})

cp.on('exit', function(code){
  // Wait few ms for error to be printed.
  setTimeout(function(){
    process.exit(code)
  }, 20)
});

process.on('SIGTERM', function() {
  cp.kill('SIGTERM')
  process.exit(1)
})
```
A real phantomjs process be born using `spawn` function by childprocess ("node_modules/.bin/phantomjs" proccess).
It is the grandchild process.

Structure at this time , is the following structure .

```
[current process]        [child process]                                  [grandchild]
selenium (python) ->  node_modules/.bin/phantomjs ->  node_modules/phantomjs/lib/phantomjs.js
```    

Implementation of the termination process is the following (this is selenium python binding code).

https://github.com/SeleniumHQ/selenium/blob/051c8b110a1aec35247cd45fa4db85c6e522cdcb/py/selenium/webdriver/common/service.py#L151-L153
```
                self.process.terminate()
                self.process.kill()
                self.process.wait()
```

In Posix OS process.terminate () is SIGTERM, process.kill () is SIGKILL.
If child process before issuing the SIGTERM to grandchild process, parent process issued SIGKILL to child process...
A child process will die, but a grandchild process will continue to live as a zombie.
Actually, Stops correctly, sleep between "self.process.ternimate()" and "self.process.kill()"

In Posix OS process.terminate () is Win32 API TerminateProcess(), and  process.kill() s Win32 API TerminateProcess() too.
So it completed successfully, it work correctly child process's SIGTERM Handler.

node_modules/.bin/phantomjs
```
process.on('SIGTERM', function() {
  cp.kill('SIGTERM')
  process.exit(1)
})
```

The cause of this problem is that it does not properly close their phantomjs.

To fix it will be considered in three ways.

(1) Correctly to end the phantomjs ( using the shutdown API).
(2) Sleep between process.terminate() and process.kill().
(3) Only if the child process has not been completed in certain time, parent process call self.process.kill().

This PR is closed. but I think this problem is "selenium python binding libyrary" of the problem.
therefore issue the issue.

thanks

TakesxiSximada:
i issued https://github.com/SeleniumHQ/selenium/issues/2272

PR #2254, opened 2016-06-10
Start the driver services detached

canonic-epicure:

I have a nodejs script that run some tests using selenium-webdriver nodejs module. In my script I'm listening on the `process.on('SIGINT', ...)` to perform some cleanup tasks. One of those tasks is to close the IE page, that has been opened for testing.

But, on Windows, the CTRL+C is propagated through all the processed attached to a shell, so, it is also processed by the IE server which quits and leaves the page opened. When my script tries to close the page (driver.quit()) I just have an exception, because IE server has been already terminated.

Solution is to launch the IE server detached from the console so it will survive CTRL+C, process the page close request and then closed by the `driver.quit()`

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/2254)


javascript/node/selenium-webdriver/io/exec.js
```diff
@@ -115,7 +115,8 @@ module.exports = function exec(command, opt_options) {
 
   var proc = childProcess.spawn(command, options.args || [], {
     env: options.env || process.env,
-    stdio: options.stdio || 'ignore'
+    stdio: options.stdio || 'ignore',
+    detached: true
```
jleyba: You're welcome to surface this option through the API, but I'm not going to make this the default behavior.

SamuraiJack:
Its not possible to configure the options, please provide such functionality. See for example "createServiceFromCapabilities" in the IE module, options for the DriverService are hardcoded

SamuraiJack:
And what's the reasoning for not running the service detached btw?

jleyba:
> Its not possible to configure the options, please provide such functionality. See for example "createServiceFromCapabilities" in the IE module, options for the DriverService are hardcoded

Yes, you'll have to pipe the option all the way up.

> And what's the reasoning for not running the service detached btw?

Users don't always properly clean-up and would be surprised by orphaned browsers/processes accumulating when their script finishes (or exits abnormally).

SamuraiJack:
> Yes, you'll have to pipe the option all the way up.

Can you explain? As I mentioned, in the [createServiceFromCapabilities](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/ie.js#L390), the IE driver service is instantiated with the fixed options, I don't understand how I can pass additional option to it. In fact, [the spawn](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/io/exec.js#L116) is also done with the fixed options. May be I'm missing something, can you please point me to the right direction?

>Users don't always properly clean-up and would be surprised by orphaned browsers/processes accumulating when their script finishes (or exits abnormally).

This is exactly the case I'm trying to solve and which requires the "detached : true" option (please re-read my original post). And there will be no orphaned processes because the "exec" command does all the cleanup on the main process exit: https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/io/exec.js#L124

SamuraiJack:
Any updates?

diemol:
@canonic-epicure, would you like to continue with this PR?

canonic-epicure:
I'm still using this change as a patch in my local installation. It solves the problem I've described for me. 

I'm not planning to continue right now (quite busy), however in few months I'll be revisiting this topic and at that point I'd gladly have this PR properly merged. Please feel free to pick it up as well.

diemol:
@canonic-epicure, sure. 
We do plan to implement this for now, so PRs are still welcome for this. However please note that we prefer to have it as an option, so it is configurable and not by default. 

I will close this for now as we are in the process of cleaning old PRs that went stale. Whenever you have the time, please feel free to open it again.

PR #2255, opened 2016-06-10
GeckoDriverService should wait for the geckodriver to be running

ngsankha:

### DIFF
--- a/java/client/src/org/openqa/selenium/net/PortProber.java
+++ b/java/client/src/org/openqa/selenium/net/PortProber.java
@@ -25,6 +25,7 @@ import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 import java.net.Socket;
 import java.net.UnknownHostException;
+import java.net.SocketTimeoutException;
 import java.util.Random;
 import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


AutomatedTester:
@lukeis could you review please? I have done a quick look and looks good to me.

java/client/src/org/openqa/selenium/net/PortProber.java
```diff
@@ -156,4 +157,22 @@ public static boolean pollPort(int port, int timeout, TimeUnit unit) {
 
     return false;
   }
+
+  public static void waitForPortUp(int port, int timeout, TimeUnit unit) {
+    long end = System.currentTimeMillis() + unit.toMillis(timeout);
+    while (System.currentTimeMillis() < end) {
+      try {
+        Socket socket = new Socket();
+        socket.connect(new InetSocketAddress("localhost", port), 1000);
```
lukeis: is this the only difference between pollPort?

sankha93: @lukeis yes, and we are catching a few other Exceptions as a result of this.

lukeis: so, why not modify the other pollPort? I'm not a fan of seemingly duplicate code that does the same thing

sankha93: @lukeis The `pollPort` function checks if a server can be created on that port, where as this checks if the server is running so that it can be connected to. Or are you referring to some other `pollPort` function that I missed?

lukeis: i think you're confusing checkPortIsFree?  pollPort, directly above this method does almost the same same logic.

lukeis: ah, you're right, it's trying to bind instead... i thought we had other logic... let me check

lukeis: looks like we don't... ok then :)

sankha93:
@lukeis When will the next standalone jar be released? This would be really important for making it work reliably with Geckodriver. :)

lukeis:
¯\_(ツ)_/¯  3.0-beta will be the next release and will be done "when it's ready"

By Christmas, definitely Christmas....

https://botbot.me/freenode/selenium/2016-04-29/?msg=65149334&page=4

PR #2286, opened 2016-06-14
assign marionette_port to each service instance

twalpole:

If a unique marionette_port option isn't passed to geckodriver it uses 2828 - This means that multiple instances of geckodriver all end up talking to the same instance of firefox even though they each launch their own.  The end result of this when using selenium is strange behavior in the first session created once multiple sessions are created (double actions, random errors, etc) and other sessions not actually being independent.

This change assigns a unique marionette_port option and keeps track of whats been assigned since the port isn't actually taken until geckodriver starts up its firefox instance.

### DIFF
--- a/rb/lib/selenium/webdriver/firefox/service.rb
+++ b/rb/lib/selenium/webdriver/firefox/service.rb
@@ -27,11 +27,11 @@ module Selenium
 
       class Service < WebDriver::Service
         DEFAULT_PORT = 4444
-
+        @@reserved_marionette_ports = []
         private
 
         def start_process
-          server_command = [@executable_path, "--binary=#{Firefox::Binary.path}", "--webdriver-port=#{@port}", *@extra_args]
+          server_command = [@executable_path, "--binary=#{Firefox::Binary.path}", "--webdriver-port=#{@port}", "--marionette-port=#{@marionette_port}", *@extra_args]
           @process       = ChildProcess.build(*server_command)
 
           if $DEBUG == true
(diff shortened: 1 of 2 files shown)


twalpole:
@lukeis Sure -- any idea how that is possible without modifying the .travis.yml ?

lukeis:
@twalpole you should disable it for your whole fork. You should be able to execute the same things locally and when you make pull requests they will run in travis here

twalpole:
@lukeis  ok -- just to point out there is a way for the selenium project to fix this in the travis.yml by encrypting the channel name - last comment of https://github.com/travis-ci/travis-ci/issues/1094

lukeis:
@twalpole that only works because mozilla is in charge of the irc server it seems... we are not (freenode)... also i can't find anything about that "secure" config they use.

titusfortner:
So, I'm pretty sure that this is a workaround for [this Marionette bug](https://github.com/mozilla/geckodriver/issues/58), so I'm not sure we want to implement it.

This does pass for local Ruby bindings, but it won't solve the issues with running it on a remote server.

@p0deje - what do you think?

twalpole:
@titusfortner that bug report seems to be about a session being shut down and not able to start another - this allows to run multiple sessions at once - i think they're different

twalpole:
@lukeis it has nothing to do with Mozilla running the irc server - it uses Travis encrypted variables - https://docs.travis-ci.com/user/environment-variables/#Encrypted-Variables - which makes them only accessible to the repo that encrypts them and not to forks

lukeis:
@twalpole doesn't look like that will help us:
"Encrypted environment variables are not available to pull requests from forks due to the security risk of exposing such information to unknown code."

we only use travis for pull requests

twalpole:
@lukeis  The notification server setting isn't an environment variable though -- I believe it will work, willing to be proven wrong though

p0deje:
I'm not sure we should address this issue in client bindings. I believe that geckodriver (just like chromedriver) should pick up some random port for Marionette when starting Firefox. It looks like internal property of geckodriver-Firefox, not part of client's responsibility.

I believe it would be better to create an issue for that in [geckodriver](https://github.com/mozilla/geckodriver/issues) bug tracker and wait for them to fix it. At the same time, there is no workaround available in Ruby but we might want to allow passing `--marionette-port` when starting Firefox driver.

twalpole:
geckodriver has implemented random port assignment in https://github.com/mozilla/geckodriver/commit/4d9e5b321e2665cb94f59fab16798a210020dc4b so this fix is not needed once the next geckodriver release happens. - Closing

PR #2310, opened 2016-06-17
Fix http proxying by including full url in path

### LINKED ISSUE #2309: WebDriver Proxy setting does not work
## Meta -
OS:  Mac OS X 10.11.5
Selenium Version: 2.53.2

## Expected Behavior -

Expect selenium-webdriver to connect to a remote Selenium server through a proxy when using `usingWebDriverProxy`.

## Actual Behavior -

selenium-webdriver does not proxy requests correctly to remote Selenium server, even when `usingWebDriverProxy` option is set correctly. Instead, it sends webdriver requests directly to the proxy server.

## Steps to reproduce -

1. On Mac, install and run [SquidMan](http://squidman.net/) proxy server on port 8080 (default settings)
2. Run the code below using NodeJS, which attempts to connect to a remote Selenium server.
3. When run, will output an error. Should instead get "successful connection using proxy" message.

```
var webdriver = require('selenium-webdriver');
new webdriver.Builder()
    .forBrowser('chrome')
    .usingWebDriverProxy('http://localhost:8080')
    .usingServer('http://ondemand.saucelabs.com/wd/hub')
    .buildAsync()
    .then(function(driver) {
        driver.quit();
    })
    .catch(function(ex) {
        var err = ex.stack.toString();
        if (err.indexOf('Sauce Labs Authentication Error') > -1) {
            return console.log('Successful connection using proxy!')
        }
        console.error(err);
    });
```

loyalchow:

To correctly proxy, both the host header AND the path need to be set to the host you want to connect to. More information can be found here: http://stackoverflow.com/questions/3862813/how-can-i-use-an-http-proxy-with-node-js-http-client

Fixes #2309

### DIFF
--- a/javascript/node/selenium-webdriver/package.json
+++ b/javascript/node/selenium-webdriver/package.json
@@ -35,7 +35,9 @@
     "multer": "^0.1.7",
     "promises-aplus-tests": "^2.1.0",
     "serve-index": "^1.6.1",
-    "sinon": "^1.17.2"
+    "sinon": "^1.17.2",
+    "bufferjs": "~1.0.0",
+    "filternet": "^0.0.2"
   },
   "scripts": {
     "test": "mocha --harmony -t 600000 --recursive test"
--- a/javascript/node/selenium-webdriver/test/http/http_test.js
+++ b/javascript/node/selenium-webdriver/test/http/http_test.js
@@ -19,6 +19,7 @@
 
 var assert = require('assert'),
     http = require('http'),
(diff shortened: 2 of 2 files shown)


jleyba:
You need to add tests demonstrating these changes are correct.

loyalchow:
@jleyba I've added a transparent http proxy server to tests, and updated proxy request tests

jleyba:
Thanks for the PR, but I ended up making more changes. #2309 should be fixed (at least the example you provided works)

Issue #2196, opened 2016-06-02
Misleading property name - Python


remcowesterhoud:
In the Python WebElement class there is a property call [parent](https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/webelement.py#L419).
This property returns the WebDriver that was used to find this element. I feel this name is misleading as when I have an element and I call parent on it I'd expect it to return another element, wich is the parent of the element it was called on.

``` html
<div id='parent'>
    <div id='child'>
        <span>Text!</span>
    </div>
</div>
```

``` py
driver = webdriver.Chrome()
child_element = driver.find_element_by_id('child')
parent_element = driver.find_element_by_id('parent')
parent = child_element.parent
```

In this case the value of parent will be equal to driver.
My expectation would be that parent would be equal to parent_element

I'd create PR that adds a new property for getting the webelement parent but I can't decide on a proper name for it as parent is the only thing that makes sense to me.

Mobrockers:
:+1:

cgoldberg:
> I can't decide on a proper name for it
`parent_element` ?

jleyba:
Wr have intentionally avoided adding DOM traversal APIs like parentNode etc. No such PRs will be accepted

remcowesterhoud:
@jleyba May I ask what the reason for this decision is?

andreastt:
@remcowesterhoud A DOM traversal API is something that can be implemented as an abstraction on top of WebDriver and thus considered out of scope. In my opinion it is something worthwhile to be worked on if you have the interest for it.

I can sympathise with the confusing naming for `parent` however.

barancev:
Closing the issue, see the related PR #2338 for explanation.

PR #2338, opened 2016-06-22
Renamed property 'parent' to 'driver'

remcowesterhoud:

Replaced the misleading name `parent` for `driver` in the `WebElement` class.
The `parent` property is now deprecated.
Related issue: #2196 

Also made the `get_root_parent()` function from `py/selenium/webdriver/remote/utils.py` deprecated as it is a useless function. It gets the `parent` of a `WebElement` (which is an instance of `WebDriver`). It then continues tries to find the `parent` of this `WebDriver` until it no longer has one. Since `WebDriver` doesn't contain a `parent` attribute it will always stop the loop after getting the `parent` of the passed `WebElement`. Hence this function does the exact same as the `parent` property (now renamed to `driver`) of the `WebElement` class.

### DIFF
--- a/py/selenium/webdriver/remote/utils.py
+++ b/py/selenium/webdriver/remote/utils.py
@@ -50,6 +50,19 @@ def return_value_if_exists(resp):
         return resp["value"]
 
 
+def get_root_parent(elem):
+    import warnings
+    warnings.warn("get_root_parent(WebElement) is a deprecated alias, use WebElement.parent",
+                  DeprecationWarning, 2)
+    parent = elem.parent
+    while True:
+        try:
+            parent.parent
+            parent = parent.parent
+        except AttributeError:
+            return parent
+
+
 def unzip_to_temp_dir(zip_file_name):
     """Unzip zipfile to a temporary directory.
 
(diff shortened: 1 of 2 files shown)


lukeis:
get_root_parent isn't a useless utility, if you find an element from another element, the parent will be set as the starting element. 

So it's a misnomer to call this variable a 'driver' since it could be another webelement. 

I don't agree with this change

remcowesterhoud:
@lukeis Then there seems to be a bug in the way `parent` is set.

If I run this script:
```py

driver = webdriver.Chrome()
driver.get('http://google.com')
body = driver.find_element_by_tag_name('body')
child = body.find_element_by_tag_name('div')
print(child.parent)
driver.quit()
```

The output I get is:
`<selenium.webdriver.chrome.webdriver.WebDriver (session="a06675bda40ec7056b77c3ec3e7df006")>`

So if I find an element from another element `parent` still returns the `WebDriver` that was used to find the element. Which makes sense because the doc method says:
`"""Internal reference to the WebDriver instance this element was found from."""`

AutomatedTester:
I would rather we removed the property but we can't unfortunately until we go through the selenium 3 cycle. Closing this as the documentation does warn you.

PR #2340, opened 2016-06-22
Optional wait for page to load for webelement.click (Python)

hjwp:

I expect that this might be the start of a discussion, rather than a straight merge, but I thought that having some code to look at might help said discussion.  Or y'all could just merge it of course, I think it's ready to go (although I haven't figured out how to run the test suite successfully yet).

Some context --

I'm the author of a book on web testing.  One of the things about selenium that I've found everyone (and I include myself here) struggles with are waits, explicit or implicit.  Essentially, everyone's naive expectation is that, when you ask selenium to "do something" like click on a link, it will wait until the next page has loaded before executing the next command.
That is, of course, the naive expectation, and for good reasons -- selenium can't tell the difference between a click that will cause a page load and one that won't, and in a world of increasingly dynamic sites, javascript and rich client UIs, more and more clicks don't cause page loads.

But.  That expectation is still there.  And lots and lots of websites _don't_ use fancy javascript, and even in the ones that do, many parts of that site still use normal hyperlinks, forms and so on.

One solution to this problem that's been tried is "implicit waits".  But anyone that's used them for a while knows they're unsatisfactory.  They will make selenium do a retry loop if it's ever asked for an element that it can't find, and that will cover some use cases, but there are many cases where it won't do what the user intuitively wants.  An implicit wait won't cover the case where the element exists both on the page before the click, and on the page after the click.  And it doesn't cover things like retrieving the page source or current url.  In short, implicit waits are so dissatisfactory that they're almost worth ignoring altogether, and moving onto explicit waits instead.

But explicit waits are complicated and ugly.  You either end up building your own wait_for helper functions, or you can use selenium's built-in WebDriverWait classes, and, for all that they're very powerful, they also aren't the most Pythonic, and they certainly end up causing a lot of boilerplate.

This PR proposes to address the problem with a relatively modest new piece of functionality:  an optional `wait_for_page_load` argument to `WebElement.click()`.  I'd love the default value to be True, but that wouldn't be very backward compatible, so a default of False would probably be a nice compromise.

Under the hood, the wait_for uses a technique which I've found to be fairly reliable, which is that, before clicking, we fetch the current page body, and then after clicking, we wait for that page body element to go stale.  This technique addresses several of the shortcomings of the implicit_wait approach, ie, conceptually, the wait is associated with the first action, not with the second action (if that makes sense).  I should thank @AutomatedTester for helping me to develop said solution, a couple of years back.  More info here: http://www.obeythetestinggoat.com/how-to-get-selenium-to-wait-for-page-load-after-a-click.html

As a nice bonus, we can make available the wait for page load as a context manager for general use, hanging off of WebDriver:

```
with self.driver.wait_for_page_load():
    # do stuff
```

Nice and Pythonic.

So that's what I have so far, but I'm sure there's lots of stuff I haven't thought of.  Testing is one (pointers on how to run the test suite appreciated!).  Choosing default values for the wait time is another, where to store them, or whether to even have them.  And lots of other issues I'm sure.

What does everyone think?

### DIFF
--- a/py/selenium/webdriver/remote/webdriver.py
+++ b/py/selenium/webdriver/remote/webdriver.py
@@ -31,6 +31,8 @@ from .file_detector import FileDetector, LocalFileDetector
 from selenium.common.exceptions import WebDriverException
 from selenium.webdriver.common.by import By
 from selenium.webdriver.common.html5.application_cache import ApplicationCache
+from selenium.webdriver.support.ui import WebDriverWait
+from selenium.webdriver.support.expected_conditions import staleness_of
 
 try:
     str = basestring
@@ -725,6 +727,18 @@ class WebDriver(object):
         self.execute(Command.SET_TIMEOUTS, {
             'ms': float(time_to_wait) * 1000,
             'type': 'page load'})
+
+    @contextmanager
+    def wait_for_page_load(self, time_to_wait=10):
+        """
+        Context manager that will wait for the page
+        to load after executing the wrapped action.
+        Implemented by waiting for the previous page body to
+        go stale
+        """
+        old_page_body = self.find_element_by_tag_name('body')
+        yield
+        WebDriverWait(self, time_to_wait).until(staleness_of(old_page_body))
 
     def find_element(self, by=By.ID, value=None):
         """
--- a/py/selenium/webdriver/remote/webelement.py
+++ b/py/selenium/webdriver/remote/webelement.py
@@ -29,7 +29,6 @@ from selenium.common.exceptions import WebDriverException
 from selenium.webdriver.common.by import By
 from selenium.webdriver.common.utils import keys_to_typing
 
-
 try:
     str = basestring
 except NameError:
@@ -67,9 +66,15 @@ class WebElement(object):
         """The text of the element."""
         return self._execute(Command.GET_ELEMENT_TEXT)['value']
 
-    def click(self):
-        """Clicks the element."""
-        self._execute(Command.CLICK_ELEMENT)
+
+    def click(self, wait_for_page_load=False):
+        """Clicks the element, optionally waiting until a new page loads"""
+        if wait_for_page_load:
+            with self.parent.wait_for_page_load():
+                self._execute(Command.CLICK_ELEMENT)
+        else:
+            self._execute(Command.CLICK_ELEMENT)
+
 
     def submit(self):
         """Submits a form."""


ivargas-handy:
Hi Harry, nice work! I like this a lot; it's up to the committers to accept this change, but I think it would be nice to add some basic tests, (once you figure out it out, the #selenium chatroom on freenode should help) and to add some way to change the default wait time. I don't have an opinion on how that option can be configured, as a config change in a method call, like `driver.page_load_wait(10)` or it should be default option in the `click` method. The current API doesn't set any defaults except for the polling interval.

cgoldberg:
+1000

jimevans:
I vehemently oppose this change. I appreciate that you're discussing the Python implementation specifically, but the Selenium project as a whole can't be so parochial.

I especially oppose introducing a `wait_for_page_load()` method of any sort. It implies semantics that are utterly meaningless. As you yourself point out, many pages use dynamic JavaScript processing, while others don't. There is no way to make such a method consistently meaningful for even the proverbial 80% case. It would cause a massive, nightmarish, and unmanageable influx of issues of the form "`wait_for_page_load` doesn't wait until page is loaded," because what's really meant by that is "`wait_for_page_load doesn't wait until _my personal definition for when my specific_ page is loaded."

Finally, and here's the piece that many people miss, Selenium WebDriver isn't the end-all, be-all of web automation libraries. Companion projects are awesome. Wrapper libraries are awesome. People should be encouraged to find and use those.

hjwp:
Jim, can you give me a couple of examples of pitfalls in my approach?  is it the word "page load" that is too vague?

jimevans:
@hjwp Not only the phrase "page load", but the fact that you've entirely obscured the intent of what's being waited for. Even if you change the phrase, the semantic purpose of your API change is to "wait for the page to load," there is no "one-size-fits-all" definition of what that means.

smccarthy:
Couldn't the body go stale, and still have more changes to be done? Wouldn't you want body to be non-stale for X amount of time? And even then, things can still be going on, right?

hjwp:
well, in the specific implementation i've suggested, it means "wait for the current page body to go stale".  My gut feeling is that that's easy to understand, and shouldn't cause too many surprises, but I appreciate that you've probably been through more wars than I have!   Any specific counterexamples?

hjwp:
@smccarthy good point.  Hadn't thought of that.

hjwp:
i have a feeling that i'm about to see what a can of worms it is that I'm opening here, but still, let's have a go!

does the combination of "old body stale" + "implicit wait" now work pretty well?  Or what about "old body stale" + "document.ready"?

hjwp:
(I meant document.readyState === 'complete')

hjwp:
No examples yet!   Perhaps it is because there are none?  Come now, selenium friends.  I know you've been bitten in the past.  I know implicitly_wait was a huge disappointment.   But do not let fear hold you back.  A better world is possible.  A glorious world in which the users' requirements are anticipated, and the 80% of cases are well served by code that uses less boilerplate.  Whilst still not making promises we cannot keep, and helping the 20% to know who they are.  A wonderful world, where the principle of least surprise is obeyed, and users are delighted and amazed.  Join me!

shs96c:
This would be better handled by having a custom function to pass to the "Wait" explicit wait that does what you need. That resolves the issue that @jimevans is pointing out (and which I agree with), and doesn't cause us to add hard-to-define-and-implement heuristics for waiting for a page to load.

Incidentally, the old RC bindings had a "waitForPageToLoad" method. It was horrible to use and debug issues with, and that experience is one of the reasons I'm opposing this diff as it currently stands.

AutomatedTester:
I am going to decline this PR.

Unfortunately "waiting for page load" is another case of us trying to solve the [halthing problem](https://en.wikipedia.org/wiki/Halting_problem).

## Background
To know that a page is loaded, which means the browser has stopped all network traffic from the initial access of the URL, the browser will fire a `DOMContentLoaded` event. 

## What Selenium WebDriver does does in the server side

Selenium tries to do it's best. We look at the URL before we click and if it "looks" like it will cause a page transition we add an observer. This observer then tells us that the page is loaded and we return to the client ending the `click`.

## Python implementation
I really dislike the cyclic dependency that has been created here between the support package and the driver. The support package is what it says on the tin, purely for support. The rest of the API is there to give you the primatives to be able to drive a web browser. From this you should be able to build further tools/APIs.

There is a massive footgun in this code you have given. Since people _may_ be using implicit waits, the hidden explicit wait will definitely hurt their test times.

Your code has shown that we have already given people the right tools to be able to do this themselves. 

Thank you for taking the time to do this

hjwp:
Thanks for taking the time to write a detailed reply David.

When you say "selenium tries to do its best", the story with the observer waiting for `DOMContentLoaded`, are you saying that .click() already "blocks" until page load, when it spots a url that looks like a link?

I was about to say - "that doesn't match my experience" - but now that I think of it, many of the problems I've seen are actually happening on form submit, so it's a send_keys('\n') that's actually triggering the page load.  I'm guessing selenium doesn't have similar block/wait/observer code in the case of form submission?

AutomatedTester:
No, and kind of for good reason.

When you type something in Selenium we do something like the following psuedo code

Test
```
element.send_keys("I <3 cheese")
```

Selenium Client
```
def send_keys(keysToSend):
   _execute(Command.SEND_KEYS, element, keysToSend.split(""))

**Comment by AutomatedTester** [comment]
No, and kind of for good reason.

When you type something in Selenium we do something like the following psuedo code

Test
```
element.send_keys("I <3 cheese")
```

Selenium Client
```
def send_keys(keysToSend):
   _execute(Command.SEND_KEYS, element, keysToSend.split(""))
```

Selenium driver
```
element.prototype.sendKeys = function(keys) {
   for (key in keys) {
     // perform down, press and up key events
   }
}
```

Since we split things to get a better emulation we would need to add the observer before, and then remove it after, each group of events we tell the browser to emit. Doing so has a reasonable high cost and if we do it wrong can create incredible leaks in the browser.

Since not all key strokes need this, we have given the Support libraries that people can use for when they are submitting a form and not in any of the other areas.

PR #2348, opened 2016-06-23
Added is_stale to WebElement

remcowesterhoud:

A bunch of times I am checking if a `WebElement` is still present on the `DOM`. It's easy to do this by using a `try` and `except`. But I feel `WebElement` itself should provide a way to check if it is `stale` or not.


AutomatedTester:
Thank you for taking the time to create this PR but unfortunately we will not be accepting it as it does not fill the model that we want for WebElements.

PR #2356, opened 2016-06-24
Created ExpectedCondition that checks if an element is clickable

remcowesterhoud:

This `expected condition` checks if a `WebElement` is clickable. Much like `element_to_be_clickable` except it takes a `WebElement` instead of a `locator`.
Made following the `Java` implementation of this `expected condition` which can be found [here](https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/support/ui/ExpectedConditions.java#L675)

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/2356)

### DIFF
--- a/py/selenium/webdriver/support/expected_conditions.py
+++ b/py/selenium/webdriver/support/expected_conditions.py
@@ -225,10 +225,13 @@ class clickability_of(object):
         self.element = element
 
     def __call__(self, driver):
-        element = visibility_of(self.element)(driver)
-        if element and element.is_enabled():
-            return element
-        else:
+        try:
+            element = visibility_of(self.element)(driver)
+            if element and element.is_enabled():
+                return element
+            else:
+                return False
+        except StaleElementReferenceException:
             return False
 
 
--- a/py/selenium/webdriver/support/expected_conditions.py
+++ b/py/selenium/webdriver/support/expected_conditions.py
@@ -217,6 +217,21 @@ class element_to_be_clickable(object):
             return False
 
 
+class clickability_of(object):
+    """ An Expectation for checking an element is visible and enabled such that
+    you can click it."""
+
+    def __init__(self, element):
+        self.element = element
+
+    def __call__(self, driver):
+        element = visibility_of(self.element)(driver)
+        if element and element.is_enabled():
+            return element
+        else:
+            return False
+
+
 class staleness_of(object):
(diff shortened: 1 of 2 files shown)


jleyba:
Instead of creating a duplicate method, change `element_to_be_clickable` to work with a `WebElement` as well as a locator

remcowesterhoud:
Woudl't it be pereferable to keep it in the same style as [`visibility_of_element_located`](https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/support/expected_conditions.py#L66) and [`visibility_of`](https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/support/expected_conditions.py#L83)?

Dude-X:
@remcowesterhoud How about modifying : `element_to_be_clickable`?

lukas-linhart:
@remcowesterhoud While your intention is good, the true clickability of an element cannot be determined until you actually try and attempt to click the element in question, because the element can be both displayed and enabled, but still overlaid by another element, in which case you will get an error like this: `selenium.common.exceptions.WebDriverException: Message: unknown error: Element is not clickable at point (697, 302). Other element would receive the click: <code>...</code>`

Please, correct me if I'm wrong on this. If the above is correct, the `clickability_of` method could be misleading in this case and I would probably mention it in the docstring.

remcowesterhoud:
@lukas-linhart There's nothing wrong with getting that error. It will just tell you that there is indeed a clickable element, however there is a different element covering it. It's exactly what currently happens when you use a locator to find it instead of a `WebElement`. [See here.](https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/support/expected_conditions.py#L206)

lukas-linhart:
@remcowesterhoud You are right, I have completely overlooked that. In that case I think the `element_to_be_clickable` expected condition have set an unfortunate precedent regarding what a clickability is. But that's a topic for another discussion. Thank you.

diemol:
Closing due to the lack of feedback.

codingpy:
The implementation had been given:

https://github.com/SeleniumHQ/selenium/blob/5fe1af712f3f83c5497b4e275fe831cad060a6d1/py/selenium/webdriver/support/expected_conditions.py#L293-L313

Issue #2390, opened 2016-06-28
Node.js client firefox tests do not pass


oddui:
## Meta -
OS: OSX 10.9.5
Selenium Version: master branch
Node Version: v4.2.2
Firefox Version: tried 46.0.1, 47, dev
GeckoDriver Version: v0.8.0

## Expected Behavior -
Tests pass.

## Actual Behavior -
27 firefox tests fail. [output.txt](https://github.com/SeleniumHQ/selenium/files/336409/output.txt)

## Steps to reproduce -
`npm test`

I'm trying to make some changes to the node client but cannot get the tests pass. Is there a specific version of firefox that i need to run the tests? Thanks.

jleyba:
Yeah, I haven't finished going through the failures with marionette.

If you have FF46 installed, try running with

```
SELENIUM_BROWSER=legacy-firefox npm test
```

You can set `SELENIUM_BROWSER` to a comma delimited list to run against multiple browsers:

```
SELENIUM_BROWSER=legacy-firefox,firefox,chrome,phantomjs npm test
```

oddui:
I tried the listed browsers and it seems only chrome pass the tests.

- `chrome` ✅ 

- `phantomjs` 1.9
10 fails on `Error: Server terminated early with status 1`

- `phantomjs` 2.1
6 fails on some `InvalidCookieDomainError`

- `firefox` 46
27 fails as in the issue

- `legacy-firefox` 46
31 fails  [output.txt](https://github.com/SeleniumHQ/selenium/files/338363/output.txt)

Are the fails known issues? Just want to make sure my dependencies are setup correctly.

jleyba:
```
SELENIUM_BROWSER=chrome,phantomjs npm test
```

The above passed for me. Well, phantomjs crashed a few times, but it was always a different point in the test suite each time I ran, so I consider it a phantomjs issue.

Looking at the Firefox tests on 47.0.1 and will spend time on getting the Marionette tests green today.

oddui:
@jleyba thanks for fixing the tests. Is is possible to have a 3.0 preview released on npm? So that people can start using the libs. Thanks.

PR #2391, opened 2016-06-28
Change proxy to not depend on node util module

oddui:
- `util.isArray` is [deprecated in node 4](https://nodejs.org/dist/latest-v4.x/docs/api/util.html#util_util_isarray_object), changed to `Array.isArray`.
- moved proxy.js into lib

Please consider accepting this, it makes it easier to use in browsers. Thanks.

Not sure why travis is failing, seems unrelated to the pr. Also need some help to get the node tests pass with firefox #2390. I'm happy to make adjustments to meet your standard.

### DIFF
--- a/javascript/node/selenium-webdriver/lib/proxy.js
+++ b/javascript/node/selenium-webdriver/lib/proxy.js
@@ -1,0 +1,99 @@
+// Licensed to the Software Freedom Conservancy (SFC) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The SFC licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


javascript/node/selenium-webdriver/lib/proxy.js
```diff
@@ -0,0 +1,102 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+/**
+ * @fileoverview Defines functions for configuring a webdriver proxy:
+ *
+ *     var webdriver = require('selenium-webdriver'),
```
jleyba: Everything in the lib directory is intended to be environment agnostic, so please don't reference node-specific modules.

oddui: @jleyba thanks, updated. Let me know if it needs further changes.

PR #2406, opened 2016-06-29
Enable use of Firefox profiles with Marionette in Node.js bindings

andreastt:


andreastt:
@jleyba

javascript/node/selenium-webdriver/firefox/index.js
```diff
@@ -367,14 +373,9 @@ class Driver extends webdriver.WebDriver {
       let service = createGeckoDriverService(binary);
       serverUrl = service.start();
       onQuit = () => service.kill();
+      caps.set(Capability.PROFILE, profile.encode());
```
jleyba: This seems unnecessarily expensive for local runs (especially if the profile includes an extension). Could we write the profile to disk and specify the path via capabilities?

andreastt: That is not currently possible.

However, @jgraham and I have been discussing the possibility of standardising certain common settings to share between browsers, such as binary location, flags, profile, preferences, and so on.

The argument to not ship profiles across the wire when they’re on the local file system is a strong contender, I think.

There is an [agenda item](https://www.w3.org/wiki/WebDriver/2016-July-F2F#Agenda) to discuss this at the next working group F2F in Seattle on 13-14 July.

jleyba: SGTM

I tweaked things a bit to avoid encoding empty profiles. See d59ccc169a7d0a23763d042badadc54241d8056e

andreastt: Wonderful, thanks!

jgraham: (fwiw if you are making a performance argument about sending profiles over the wire it would be interesting to see a profile that shows it's significant. My expectation would be that on the local machine it's much faster than starting Firefox, for example, so won't make much difference to the overall test time).

andreastt: I have seen profiles that are in the hundreds of megabyte in size, which might matter if you’re running tests, say, in a disposable Linux container with limited disk storage.

On the other hand it would add more complexity to the implementation, so it should be well-weighed decision against that.

marco-c:
@jleyba would you consider releasing a new version with this fix? We'd need it in https://github.com/web-push-libs/web-push/issues/179.

jleyba:
@marco-c I'll look into doing a point release for 2.53.x

marco-c:
Awesome, thanks!

marco-c:
@jleyba do you know when you're going to publish the point release? Thanks!

marco-c:
@jleyba, I filed #2514 for the point release. Let me know if you aren't planning to publish it anymore so I can stop asking 😄

PR #2407, opened 2016-06-29
Add Switch Context extension commands to Marionette for the Node.js bindings

andreastt:
The Switch Context command allows running WebDriver commands against
the chrome/XUL context in Firefox.

### DIFF
--- a/javascript/node/selenium-webdriver/firefox/index.js
+++ b/javascript/node/selenium-webdriver/firefox/index.js
@@ -117,9 +117,11 @@ const Binary = require('./binary').Binary,
     Profile = require('./profile').Profile,
     decodeProfile = require('./profile').decode,
     executors = require('../executors'),
+    http = require('../http'),
     httpUtil = require('../http/util'),
     io = require('../io'),
     capabilities = require('../lib/capabilities'),
+    command = require('../lib/command'),
     logging = require('../lib/logging'),
     promise = require('../lib/promise'),
     webdriver = require('../lib/webdriver'),
@@ -266,6 +268,23 @@ class Options {
   }
 }
 
+
+/**
+ * Enum of available command contexts.
+ *
+ * Command contexts are specific to Marionette, and may be used with the
+ * {@link #context=} method. Contexts allow you to direct all subsequent
+ * commands to either "content" (default) or "chrome". The latter gives
+ * you elevated security permissions.
+ *
+ * @enum {string}
+ */
+const Context = {
+  CONTENT: "content",
+  CHROME: "chrome",
+};
+
+
 const GECKO_DRIVER_EXE =
     process.platform === 'win32' ? 'geckodriver.exe' : 'geckodriver';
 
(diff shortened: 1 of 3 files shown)


andreastt:
@jleyba

javascript/node/selenium-webdriver/firefox/index.js
```diff
@@ -421,13 +470,46 @@ class Driver extends webdriver.WebDriver {
    */
   setFileDetector() {
   }
+
+  /**
+   * Get the context that is currently in effect.
+   *
+   * @return {Context} Current context.
+   */
+  get context() {
```
jleyba: style nit: I'd prefer actual `getContext()` and `setContext(ctx)` methods since they involve RPCs (this would also be consistent with the rest of the API)

javascript/node/selenium-webdriver/firefox/index.js
```diff
@@ -421,13 +470,46 @@ class Driver extends webdriver.WebDriver {
    */
   setFileDetector() {
   }
+
+  /**
+   * Get the context that is currently in effect.
+   *
+   * @return {Context} Current context.
```
jleyba: actually returns `{!promise.Promise<Context>}`

javascript/node/selenium-webdriver/firefox/index.js
```diff
@@ -421,13 +470,46 @@ class Driver extends webdriver.WebDriver {
...
+   * Changing the current context has a stateful impact on all subsequent
+   * commands. The {@link Context.CONTENT} context has normal web
+   * platform document permissions, as if you would evaluate arbitrary
+   * JavaScript. The {@link Context.CHROME} context gets elevated
+   * permissions that lets you manipulate the browser chrome itself,
+   * with full access to the XUL toolkit.
+   *
+   * Use your powers wisely.
+   *
+   * @param {Context} ctx The context to switch to.
```
jleyba: `@return {!promise.Promise<void>}`

jleyba:
Could you add a test demonstrating basic usage to

https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/test/firefox/firefox_test.js

javascript/node/selenium-webdriver/firefox/index.js
jleyba: In the constructor, should we store whether marionette is being used and fail fast here if not?

andreastt: I originally coded this in, but then realised it might be better to let it fail with an “unknown command” error instead. Though I guess this can be confusing to users because they might not realise it’s specific to Marionette.

This comes down to stylistic preferences. Would you like to check if we’re talking to a Marionette remote and throw or leave it as-is?

jleyba: Eh, let's just leave it as is with an unknown command

andreastt:
I will add a test soon.

javascript/node/selenium-webdriver/firefox/index.js
andreastt: To be clear, I’m very happy to change this 😄

andreastt:
@jleyba Added some tests, but it would be good if you provided some feedback on them as I’m not really sure about this setup.

jleyba:
LGTM!

PR #2410, opened 2016-06-30
Enable custom port lock for FirefoxDriver

kennyschuoler:

When using multiple firefox instances at the same moment the driver
tries to lock the default port - 1 without the possibility to lock a custom
port. It is however possible to define a custom port using
FirefoxProfile.PORT_PREFERENCE
This fix solves following WebDriverException:
Unable to bind to locking port 7054 within 45000 ms


lukeis:
can you check to make sure there's a test that uses the custom port preference, if not, please add one that launches firefox with the pref set?

barancev:
It's the purpose of the locking port to be global. If you set different locking ports for different processes it's breaking the main idea of system-wide synchronization. "Unable to bind to locking port" means you're starting too many Firefox instances simultaneously, and you have to resolve _this_ issue instead of attempts to disable locking.

PR #2425, opened 2016-07-05
py: break Mobile and SwitchTo reference cycle with weakref proxy.

orent:

Browser now closes immediately when driver object is deleted.

Not tested with full test suite, but it's a patch I have been using for a while.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/2425)

### DIFF
--- a/py/selenium/webdriver/remote/mobile.py
+++ b/py/selenium/webdriver/remote/mobile.py
@@ -43,7 +43,8 @@ class Mobile(object):
     AIRPLANE_MODE = ConnectionType(1)
 
     def __init__(self, driver):
-        self._driver = driver
+        from weakref import proxy
+        self._driver = proxy(driver)
 
     @property
     def network_connection(self):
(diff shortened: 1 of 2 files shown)


lukeis:
can you help better explain the circumstances where this fixes an issue you're seeing? Or show some sample code that caused the issue? Or even better add a test to the test suite in python that showed it's fixed :)

Dude-X:
@lukeis @AutomatedTester this seems like a good thing to merge in. It's a minor win for memory efficiency.

orent:
On Python implementations with a refcounting GC it helps the external browser window close immediately when the driver object is deleted. Helps avoid external browsers remaining open in interactive sessions and other situations.

ATM, this instant closing works on python 2 but on on 3. It has something to do with the way selenium exceptions are re-raised from lower level exceptions keeping references to the frame objects.

AutomatedTester:
landed in 62aa81a0fd7eb41b35ec7ed70dc75b08040fd7b7

PR #2446, opened 2016-07-11
fix javadocs for TakesScreenshot and OutputType

alb-i986:

### DIFF
--- a/java/client/src/org/openqa/selenium/OutputType.java
+++ b/java/client/src/org/openqa/selenium/OutputType.java
@@ -24,8 +24,7 @@ import java.io.OutputStream;
 import java.util.Base64;
 
 /**
- * Defines the output type for a screenshot. See org.openqa.selenium.Screenshot for usage and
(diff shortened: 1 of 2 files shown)


java/client/src/org/openqa/selenium/OutputType.java
```diff
@@ -24,8 +24,7 @@
 import java.util.Base64;
 
 /**
- * Defines the output type for a screenshot. See org.openqa.selenium.Screenshot for usage and
- * examples.
+ * Defines the output type for a screenshot.
```
lukeis: instead of removing can you reference org.openqa.selenium.TakesScreenshot ?

alb-i986: There's already a `@see TakesScreenshot` down below: I think it would be redundant.

lukeis: indeed there is :)

PR #2482, opened 2016-07-17
DeviceRotation implementation

rafe-g:


rafael-chavez:
@lukeis fixed my commits.

java/client/src/org/openqa/selenium/DeviceRotation.java
```diff
@@ -0,0 +1,81 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package org.openqa.selenium;
+
+import java.util.Map;
+import com.google.common.collect.ImmutableMap;
+
+public class DeviceRotation {
```
lukeis: This class will be consumed by users, please add JavaDoc to it and all public methods

java/client/src/org/openqa/selenium/BUCK
```diff
@@ -60,6 +61,7 @@ java_library(name = 'core',
     '//java/client/src/org/openqa/selenium/interactions:exceptions',
     '//java/client/src/org/openqa/selenium/logging:api',
     '//java/client/src/org/openqa/selenium/security:security',
+	'//third_party/java/guava:guava',
```
lukeis: nit: indentation (use spaces)

lukeis:
@rafael-chavez which email do you want associated to the commit?

Also, you're going to need to add an implementation of calling this endpoint in the AddRotatable class. I don't like the way it's implemented there, but you don't need to change that, maybe changing the if/else if to a switch statement instead.
https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/remote/AddRotatable.java#L36

rafael-chavez:
@lukeis updated, please review.

java/client/src/org/openqa/selenium/remote/AddRotatable.java
```diff
@@ -19,28 +19,61 @@
...
   public InterfaceImplementation getImplementation(Object value) {
     return new InterfaceImplementation() {
       public Object invoke(ExecuteMethod executeMethod, Object self, Method method, Object... args) {
-        if ("rotate".equals(method.getName())) {
-          return executeMethod.execute(DriverCommand.SET_SCREEN_ORIENTATION,
-              ImmutableMap.of("orientation", args[0]));
-        } else if ("getOrientation".equals(method.getName())) {
-          return ScreenOrientation.valueOf((String) executeMethod.execute(
-              DriverCommand.GET_SCREEN_ORIENTATION, null));
+        RotatableMethod m = RotatableMethod.valueOf(method.getName());
```
lukeis: you don't need to create a enum for this, java allows you to switch on a string ;) (since java 7?)

java/client/src/org/openqa/selenium/remote/AddRotatable.java
```diff
@@ -19,28 +19,61 @@
...
+        Object methodImp;
```
lukeis: call this 'response' instead

java/client/src/org/openqa/selenium/remote/AddRotatable.java
```diff
@@ -19,28 +19,61 @@
...
+          case ROTATE:
+            String cmd = "";
+            if (args[0] instanceof ScreenOrientation) {
+              cmd = DriverCommand.SET_SCREEN_ORIENTATION;
+            } else if (args[0] instanceof DeviceRotation) {
+              cmd = DriverCommand.SET_SCREEN_ROTATION;
+            } else {
+              throw new IllegalArgumentException("rotate parameter must be either of type 'ScreenOrientation' or 'DeviceRotation'");
+            }
+            methodImp = executeMethod.execute(cmd, ImmutableMap.of("orientation", args[0]));
```
lukeis: i didn't think that setting the Rotation with DeviceRotation should send the payload with "orientation" as the key in the json object.

https://github.com/SeleniumHQ/mobile-spec/blob/master/spec-draft.md#devicerotation

If we're strictly following what is said there, it should be "parameters" in that case... but that was written when we thought the w3c spec was also going to do that, but it's not. I'd recommend that you just do:

    executeMethod.execute(DriverCommand.SET_SCREEN_ROTATION, ((DeviceRotation)args[0]).parameters());

for the case of when it is DeviceRotation... (and then we should update the spec :) )

rafael-chavez: Ah i see. Ok ill update this.

rafael-chavez:
updated mobile spec [here](https://github.com/SeleniumHQ/mobile-spec/pull/5)

Issue #2464, opened 2016-07-14
Javascript client - Builder leaves out browser custom commands when using server url


oddui:
## Meta -
Selenium Version:  master branch
Browser:  Chrome

## Expected Behavior -
Driver has `launchApp` command when using server url in builder.

## Actual Behavior -
```
TypeError: driver.launchApp is not a function
```

## Steps to reproduce -

Start chromedriver on `http://127.0.0.1:9515` and run
```
var webdriver = require('selenium-webdriver'),
  chrome = require('selenium-webdriver/chrome');

var driver = new webdriver.Builder()
  .forBrowser('chrome')
  .setChromeOptions(new chrome.Options())
  .usingServer('http://127.0.0.1:9515')
  .build();

driver.launchApp();
```

PR #2557, opened 2016-08-02
Allow defining custom commands on a deferred executor in node

sjelin:
I work on the Protractor team and some of our users have asked for us to support commands like `setNetworkConnection` which are part of appium's extended JSON Wire Protocol.  Since we don't want to swich to `wd` or `webdriverio`, I have been working on a tool that patches these commands to `WebDriver` ([current implementation](https://github.com/angular/webdriver-js-extender)).  This was relatively straightforward to do with `HttpExecutor`'s [`defineCommand`](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/lib/http.js#L276), except the one snag was that the driver's executor is generally a `DeferredExecutor` and there is currently no way to pass the `defineCommand` call through the `DeferredExecutor` to the `HttpExecutor`.  This lead to some [ugly and imperfect hacks](https://github.com/angular/webdriver-js-extender/blob/master/lib/index.ts#L43) which I'd like to get rid of.

### DIFF
--- a/javascript/node/selenium-webdriver/lib/command.js
+++ b/javascript/node/selenium-webdriver/lib/command.js
@@ -229,6 +229,21 @@ class Executor {
    *     result.
    */
   execute(command) {}
+
+  /**
+   * Defines a new command for use with this executor. When a command is sent,
+   * the {@code path} will be preprocessed using the command's parameters; any
+   * path segments prefixed with ":" will be replaced by the parameter of the
+   * same name. For example, given "/person/:name" and the parameters
+   * "{name: 'Bob'}", the final command path will be "/person/Bob".
+   *
+   * @param {string} name The command name.
+   * @param {string} method The HTTP method to use when sending this command.
+   * @param {string} path The path to send the command to, relative to
+   *     the WebDriver server's command root and of the form
+   *     "/path/:variable/segment".
+   */
+  defineCommand(name, method, path) {}
 }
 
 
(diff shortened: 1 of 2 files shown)


javascript/node/selenium-webdriver/lib/command.js
```diff
@@ -229,6 +229,21 @@ class Executor {
...
+   * same name. For example, given "/person/:name" and the parameters
+   * "{name: 'Bob'}", the final command path will be "/person/Bob".
+   *
+   * @param {string} name The command name.
+   * @param {string} method The HTTP method to use when sending this command.
+   * @param {string} path The path to send the command to, relative to
+   *     the WebDriver server's command root and of the form
+   *     "/path/:variable/segment".
+   */
+  defineCommand(name, method, path) {}
```
jleyba: I want to keep the base Executor interface separate from HTTP.

I'm working (slowly, given real work commitments) on refactoring how deferred executors are handled (for #2464).
Defining additional commands will probably be easier as a result of that.

Also, one option is to expose defineCommand() on the `Builder`, and have it just ignore extra commands if the target browser doesn't use a HTTP executor.

sjelin: Alright.  Would you be interested in a CL which exposed `defineCommand()` on the `Builder`?

sjelin: Or of course I could just leave `defineCommand` on the `DeferredExecutor` but remove this change to the interface

jleyba:
This shouldn't be needed anymore with 83dfb2cd1e0fad723cda16ae41c6851c38fef623 and dcaf7b2f17b79687a46c328abf8db45aaf802fbf. Both commits are in 3.0.0-beta-1, which is up on npm.

PR #2582, opened 2016-08-08
[js] Fix typo in execute_script_test.js

vitalets:

### DIFF
--- a/javascript/node/selenium-webdriver/test/execute_script_test.js
+++ b/javascript/node/selenium-webdriver/test/execute_script_test.js
@@ -47,7 +47,7 @@ test.suite(function(env) {
 
     test.it('fails if script throws', function() {
       execute('throw new Error("boom")')
-          .then(function() { throw shoudlHaveFailed; })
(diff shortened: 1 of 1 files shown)


vitalets:
We did not catch it before as tests always passed and `.then` was not executed

jleyba:
In fairness, the test still would have failed with a ReferenceError if this code path triggered :)

Thanks for the patch!

vitalets:
Yes, a kind of tricky test :)

btw, @jleyba I noticed that count of browser tests of javascript binding is significantly less than for java/python. Are they needed or it's intentional?

jleyba:
Converting tests is tedious. I converted enough to be relatively happy with the coverage and called it a day. PRs for increased test coverage are always welcome.

vitalets:
Ok, I will have a look on it. 
Could you advice how to measure coverage for javascript tests?
I use `mocha --harmony -t 600000 --recursive test` that seems not to set any coverage tool.

PR #2597, opened 2016-08-10
goog.dom.TagName changed type from string to object.

johnjbarton:

### DIFF
--- a/javascript/atoms/dom.js
+++ b/javascript/atoms/dom.js
@@ -33,6 +33,7 @@ goog.require('goog.dom.TagName');
 goog.require('goog.math');
 goog.require('goog.math.Coordinate');
 goog.require('goog.math.Rect');
+goog.require('goog.object');
 goog.require('goog.string');
 goog.require('goog.style');
 goog.require('goog.userAgent');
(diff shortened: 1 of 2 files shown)


johnjbarton:
This is FYI more than a pull request: this change was made by folks who are changing goog.dom.TagName (in the next month they say). I'm passing it along.

javascript/atoms/dom.js
```diff
@@ -312,18 +313,18 @@ bot.dom.getAttribute = function(element, attributeName) {
...
  * List of elements that support the "disabled" attribute, as defined by the
  * HTML 4.01 specification.
- * @private {!Array.<goog.dom.TagName>}
+ * @private {!Object<string, boolean>}
  * @const
  * @see http://www.w3.org/TR/html401/interact/forms.html#h-17.12.1
  */
-bot.dom.DISABLED_ATTRIBUTE_SUPPORTED_ = [
+bot.dom.DISABLED_ATTRIBUTE_SUPPORTED_ = goog.object.createSet([
   goog.dom.TagName.BUTTON,
```
lukeis: Shouldn't these all be wrapped in String(goog.dom.TagName...) then?

Assuming they are adding a toString function to them?

johnjbarton: A toString() is added to the goog.dom.TagName class.  

Deep in createSet the argument array values are used in a property assignment:
https://github.com/google/closure-library/blob/0630ddc6697ee7921b5b125b2e882ebf5cabc028/closure/goog/object/object.js#L671
The property access converts to string:
http://www.ecma-international.org/ecma-262/6.0/#sec-topropertykey

So I think we don't need to convert toString explicitly.

lukeis: I feel like this is potentially relying on something too deep in the ecma definition... is there any browser quirk there? Also, when I dug deep in there, it seems it would actually attempt to call valueOf first and then toString...
http://www.ecma-international.org/ecma-262/6.0/#sec-toprimitive

Guess I'd prefer to be more explicit rather than potentially leaving ourselves open to a browser quirk

jleyba: There's nothing special here. The keys in an object are always coerced into a string, that's just how JS works (and one of the reasons the Map class was added in ES6)

diemol:
Closing as this went stale and has no reactions from the other committers.

PR #2633, opened 2016-08-19
Able to add ".page.com" cookie when page.com opened

a0s:

```
browser = Watir::Browser.new :firefox
browser.goto "http://domain.com"
browser.cookies.add('name', 'value', domain: '.domain.com')
```

https://github.com/seleniumhq/selenium-google-code-issue-archive/issues/1953

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/2633)

### DIFF
--- a/javascript/firefox-driver/js/firefoxDriver.js
+++ b/javascript/firefox-driver/js/firefoxDriver.js
@@ -728,7 +728,8 @@ FirefoxDriver.prototype.addCookie = function(respond, parameters) {
   } else {
     var currLocation = respond.session.getWindow().location;
     var currDomain = currLocation.host;
-    if (currDomain.indexOf(cookie.domain) == -1) {  // Not quite right, but close enough
+    if (currDomain.indexOf(cookie.domain) == -1 && 
+        ".".concat(currDomain).indexOf(cookie.domain) == -1) {  // A little better
       throw new WebDriverError(bot.ErrorCode.INVALID_COOKIE_DOMAIN,
           'You may only set cookies for the current domain');
     }


javascript/firefox-driver/js/firefoxDriver.js
```diff
@@ -728,7 +728,8 @@ FirefoxDriver.prototype.addCookie = function(respond, parameters) {
   } else {
     var currLocation = respond.session.getWindow().location;
     var currDomain = currLocation.host;
-    if (currDomain.indexOf(cookie.domain) == -1) {  // Not quite right, but close enough
+    if (currDomain.indexOf(cookie.domain) == -1 && 
+        ".".concat(currDomain).indexOf(cookie.domain) == -1) {  // A little better
```
tommywo: `currDomain.indexOf(cookie.domain) == -1 &&` is redundant

lukeis:
i don't really see the value in this patch, easy enough in the client side code to just pass in the domain without the leading "." ?

orangeudav:
Because cookie's domain may start and may not start with dot - two different record type (http://stackoverflow.com/questions/9618217/what-does-the-dot-prefix-in-the-cookie-domain-mean). And you not support start-from-dot form of domain

joesho112358:
@a0s I am not seeing this. Is it still an issue?

```
> browser.cookies.to_a.last
 => {:name=>"name", :value=>"value", :path=>"/", :domain=>".domain.com", :expires=>nil, :secure=>false} 
```

littleclay:
Any word on this? Our website uses the domain style that starts with a dot. We can add the cookie without the preceding dot, but that results in two cookies with different domains. Also, the server is looking for the domain including the dot and isn't matching the cookie added that doesn't start with a dot.

jrgp:
Hi,

I also have this use case.

I'm setting a cookie for a site, eg `www.domain.com`, and in some cases my site tries to delete that same cookie with the domain set to `.www.domain.com`. Because `.www.domain.com` does not match `www.domain.com`, the cookie does not actually get deleted, and this is causing problems with our automated selenium testing.

Any word on this decision? @lmtierney ? (sorry for the highlight but it looks like you're active on this project)

It looks like this file was refactored a little while ago so this PR won't be able to be merged. I can submit a new PR if that'd help.

p0deje:
This PR adds support for legacy Firefox driver. What about other drivers? Does it work in ChromeDriver or GeckoDriver?

WebDriver specification for [adding cookie](https://www.w3.org/TR/webdriver/#add-cookie) has the following section:

> If cookie name or cookie value is null, **cookie domain is not equal to the current browsing context’s active document’s domain**, cookie secure only or cookie HTTP only are not boolean types, or cookie expiry time is not an integer type, or it less than 0 or greater than the maximum safe integer, **return error with error code invalid argument.** 

The part of domain equality is a bit vague, so my guess is that none of spec-compliant drivers currently support adding cookie with leading dot. In that case, we probably need to fix the spec first.

jrgp:
ChromeDriver does not exhibit this problem. I haven't tried the other drivers.

p0deje:
I can also confirm it works fine in latest GeckoDriver. With that said, I see no reason not to merge this PR.

Sorry for it hanging around for 2 years.

PR #2661, opened 2016-08-24
Fix PhantomJsDriver maven reference

valfirst:
- [X] By placing an `X` in the preceding checkbox, I verify that I have signed the [Contributor License Agreement]


lukeis:
Looks like there's a newer version though? 1.3.0?

valfirst:
Right, that's going to be a next PR, just to be clear in changes: fix issue, then upgrade the library

PR #2673, opened 2016-08-26
Fix loading from -hubConfig and -nodeConfig files

mach6:

---

Changes to make -hubConfg and -nodeConfig work again.
- If/when a hub/node config file is passed, it is used to initialize
  the GridHubConfiguration or GridNodeConfiguration object.
- Initialization of a new configuration object from an other
  configuration object does not protect values which are
  otherwise protected by merge()
- Any passed args present from the command line (to GridLauncherV3)
  will override values supplied in the hub/node config file.
- Any value supplied for -hubConfig or -nodeConfig is checked
  for existence via new FileExistsValueValidator class
  Additional changes;
- Hide parameter "-avoidProxy" from usage()
- Remove whitespace in DefaultHub.json

### DIFF
--- a/java/server/src/org/openqa/grid/internal/utils/configuration/GridHubConfiguration.java
+++ b/java/server/src/org/openqa/grid/internal/utils/configuration/GridHubConfiguration.java
@@ -31,6 +31,7 @@ import org.openqa.grid.common.exception.GridConfigurationException;
 import org.openqa.grid.internal.listeners.Prioritizer;
 import org.openqa.grid.internal.utils.CapabilityMatcher;
 import org.openqa.grid.internal.utils.DefaultCapabilityMatcher;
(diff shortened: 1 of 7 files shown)


java/server/src/org/openqa/grid/internal/utils/configuration/GridHubConfiguration.java
```diff
@@ -121,6 +159,23 @@ public CapabilityMatcher convert(String capabilityMatcherClass) {
...
+    merge(other);
+  }
+
+  /**
+   * Initialize from another configuration's values when they are set.
+   * @param other
+   */
+  protected void initFrom(GridNodeConfiguration other) {
+    super.initFrom(other);
+    super.merge(other);
```
mach6: This line is duplicative. Will remove it

PR #2683, opened 2016-08-29
Fix/Add support for node servlets

mach6:

Support for node -servlets bound to `/extra` and
`/resources/` was broken when the pre-existing
(2.x/rc) `SeleniumServer` code was removed.
This change set adds support back.
- Updated interface `GridNodeServer`
- Updated (new) `SeleniumServer` class to accept/process
  extra servlets
- Updated `SelfRegisteringRemote` to process `nodeConfig.servlets`
  that are passed.
- Updated tests in this area

Misc Other;
- Unrelated fixes to `GridLauncherV3` (unused import, code path
  correction)

### DIFF
--- a/java/server/src/org/openqa/grid/internal/utils/SelfRegisteringRemote.java
+++ b/java/server/src/org/openqa/grid/internal/utils/SelfRegisteringRemote.java
@@ -32,6 +32,8 @@ import org.openqa.grid.common.exception.GridException;
 import org.openqa.grid.internal.utils.configuration.GridHubConfiguration;
 import org.openqa.grid.internal.utils.configuration.GridNodeConfiguration;
 import org.openqa.grid.shared.GridNodeServer;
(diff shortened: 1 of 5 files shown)


lukeis:
Looks good, will take another look when I'm on a real computer again. But, no tests? :)

mach6:
@lukeis .. Happy to add tests.. Also, fyi this was the path of least resistance. I'd _really_ like to change `SeleniumServer` such that it does not `implement GridNodeServer` -- The fact that it currently impelments `GridNodeSever` seems odd considering it only accepts a `StandaloneConfiguration` -- IMO we should have `SeleniumServer` not implement `GridNodeServer` and create a new `NodeServer extends SeleniumServer implements GridNodeServer`..  This way the separation makes more sense and the extra `-servlets` support is only in the new `NodeServer` class

PR #2686, opened 2016-08-30
Add Gecko XBL extension commands for anonymous nodes

andreastt:
See the commit messages on each of the patches.

### DIFF
--- a/py/selenium/webdriver/firefox/webdriver.py
+++ b/py/selenium/webdriver/firefox/webdriver.py
@@ -30,19 +30,22 @@ import socket
 import sys
 
 from .firefox_binary import FirefoxBinary
+from .options import Options
 from .remote_connection import FirefoxRemoteConnection
 from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
 from selenium.webdriver.firefox.extension_connection import ExtensionConnection
 from selenium.webdriver.firefox.firefox_profile import FirefoxProfile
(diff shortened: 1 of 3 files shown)


py/selenium/webdriver/firefox/webdriver.py
```diff
@@ -30,20 +30,40 @@
...
 from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
 from selenium.webdriver.firefox.extension_connection import ExtensionConnection
 from selenium.webdriver.firefox.firefox_profile import FirefoxProfile
 from selenium.webdriver.remote.webdriver import WebDriver as RemoteWebDriver
+from selenium.webdriver.remote.webelement import WebElement as RemoteWebElement
 from .service import Service
-from .options import Options
+
+
+class FirefoxWebElement(RemoteWebElement):
```
lukeis: i think this should get put into its own file. This could start to grow :)

lukeis:
the py docs added here don't make it clear when one would use this... maybe a link to some mozilla documentation around what xbl is? or examples?

andreastt:
@lukeis I addessed the documentation issue in f5445ed, but I really don’t know more about this than you do.

py/selenium/webdriver/firefox/webdriver.py
andreastt: Fixed in f5445ed.

lukeis:
Thanks :+1:

PR #2694, opened 2016-08-31
Adding touch screen funtionality to ChromeDriver

kdaimiel:
It seems that ChromeDriver supports touch actions so I have added this functionality to be able to run this kind of actions in a browser with Selenium. This has come because I had to test hybrid apps with a touch screen laptop.

### DIFF
--- a/java/client/src/org/openqa/selenium/chrome/ChromeDriver.java
+++ b/java/client/src/org/openqa/selenium/chrome/ChromeDriver.java
@@ -108,6 +108,7 @@ public class ChromeDriver extends RemoteWebDriver
 
   private RemoteLocationContext locationContext;
   private RemoteWebStorage webStorage;
+  private TouchScreen touchScreen;
 
   /**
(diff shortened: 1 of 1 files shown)


eduramiba:
+1

java/client/src/org/openqa/selenium/chrome/ChromeDriver.java
```diff
@@ -207,4 +210,9 @@ public void setLocation(Location location) {
   public void launchApp(String id) {
     execute(ChromeDriverCommand.LAUNCH_APP, ImmutableMap.of("id", id));
   }
+
+  @Override
+  public TouchScreen getTouch() {
+	return new RemoteTouchScreen(this.getExecuteMethod());
```
lukeis: would be better to store as a local private variable (like locationContext / webStorage) and only inialize it once on construction.
also fix the indentation here

kdaimiel: Changes have been made. Thank you

PR #2717, opened 2016-09-05
Changed all bare excepts to use base Exception class

isaulv:

In Python a bare except clause will catch any transient system errors. This is not recommended except in a very few exceptional cases, which none apply for this project. This PR fixes it so that uses the base exception class `Exception` so that some system errors do bubble up.


AutomatedTester:
thank you!

PR #2718, opened 2016-09-05
[js] switchTo().window() should use a 'handle' parameter rather than `name`

Standard8:

This is to match the W3C Webdriver spec, and fixes `switchTo().window()` for use with geckodriver.

Note that the python driver already uses `handle` rather than `name`

I've tried running the tests locally, but either they are taking a very long time, or there's something wrong with them on my machine.

### DIFF
--- a/javascript/node/selenium-webdriver/lib/webdriver.js
+++ b/javascript/node/selenium-webdriver/lib/webdriver.js
@@ -1654,10 +1654,12 @@ class TargetLocator {
    *     when the driver has changed focus to the specified window.
    */
   window(nameOrHandle) {
-    let paramName = this.driver_.getExecutor().w3c ? 'handle' : 'name';
     return this.driver_.schedule(
         new command.Command(command.Name.SWITCH_TO_WINDOW).
-            setParameter(paramName, nameOrHandle),
+            // "name" supports the legacy drivers. "handle" is the W3C
+            // compliant parameter.
+            setParameter('name', nameOrHandle).
+            setParameter('handle', nameOrHandle),
         'WebDriver.switchTo().window(' + nameOrHandle + ')');
   }
 
--- a/javascript/node/selenium-webdriver/test/lib/webdriver_test.js
+++ b/javascript/node/selenium-webdriver/test/lib/webdriver_test.js
@@ -199,10 +199,8 @@ describe('WebDriver', function() {
   }
 
   class FakeExecutor {
-    constructor(w3c) {
+    constructor() {
       this.commands_ = new Map;
-
-      this.w3c = w3c || false;
     }
 
     execute(command) {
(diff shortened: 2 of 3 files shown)


javascript/firefox-driver/js/nsCommandProcessor.js
```diff
@@ -533,13 +533,13 @@ nsCommandProcessor.prototype.execute = function(jsonCommandString,
...
  *     command response in.
- * @param {{name: string}} parameters The command parameters.
+ * @param {{handle: string}} parameters The command parameters.
  * @param {number} opt_searchAttempt Which attempt this is at finding the
  *     window to switch to.
  */
 nsCommandProcessor.prototype.switchToWindow = function(response, parameters,
                                                        opt_searchAttempt) {
-  var lookFor = parameters.name;
+  var lookFor = parameters.handle;
```
andreastt: /javascript/firefox-driver is the Selenium implementation of a WebDriver for Firefox. Since it generally isn’t W3C compatible, it shouldn’t change.

javascript/node/selenium-webdriver/lib/webdriver.js
```diff
@@ -1656,7 +1656,7 @@ class TargetLocator {
   window(nameOrHandle) {
     return this.driver_.schedule(
         new command.Command(command.Name.SWITCH_TO_WINDOW).
-            setParameter('name', nameOrHandle),
+            setParameter('handle', nameOrHandle),
```
andreastt: This should only be sent if the driver is speaking to a W3C conformant remote, so we need an if-condition check like we have in the Python bindings.

Standard8:
@andreastt Thank you for the hints, I've undone the webdriver for Firefox change, and added the W3C compliance detection.

I also worked out how to run the nodejs tests, and I've added a couple of test cases to test that the parameter and w3c detection works correctly.

For reference, cd into `javascript/node/selenium-webdriver`, `npm install` then `npm test`. Need geckodriver in $PATH.

javascript/node/selenium-webdriver/lib/webdriver.js
```diff
@@ -1654,9 +1654,10 @@ class TargetLocator {
    *     when the driver has changed focus to the specified window.
    */
   window(nameOrHandle) {
+    let paramName = this.driver_.getExecutor().w3c ? 'handle' : 'name';
```
jleyba: I'd rather just send the parameter twice than break encapsulation here. There's already precedence with webelement IDs

jleyba:
Looks like we're missing e2e coverage for window switching in

https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/test/window_test.js

Mind adding a test or 2?

Standard8:
> Mind adding a test or 2?

@jleyba It took a bit of work, but I've now added a test, it seems to work fine.

PR #2748, opened 2016-09-09
Better default check for canBubble

mathiasrw:
I found that your code was comparing the result of `typeof` with a variable named `undefined`.

As typeof returns a string it would compare to a string with the content `'undefined'`

Rewrote the line to a more simple one.


lukeis:
Thanks for catching this! Although the original intention is to check for undefined and then default to true, with your change if one passes in false it will change the value to true. Mind adjusting it accordingly? (Just changing undefined to "undefined" would do it)

mathiasrw:
Aww - you are right !

Will change it...

PR #2768, opened 2016-09-13
Clear lastException if it is not the cause.

joshbruning:
Clear lastException if it is not the cause.

If another retry or timeout exception would be caused by a false or null value, the last exception is not the cause of the timeout.


lukeis:
@joshbruning you know you have commit rights? ;)  Also, try to avoid creating a branch on the main repo itself please :)

PR #265, opened 2014-10-07
Adding POC for REST api on Selenium Hub

dimacus:
This is not necessarily a pull request, more of a request for information. 

I'm looking to close issue #5586

Here is the idea, make a REST style API that sits on top of the existing grid and allows users to interrogate the state from outside. After the basic API is implemented, it will be possible to close #1843 because the console could use JS to make async updates without reloading the whole console. By fixing above two defects we can also close #8002.

Here is a working POC, the following screenshots will describe individual endpoints. This is still in proposal stage, but I hope some for of this will be liked and accepted since I desperately need this and tired of writing new servlets that may break one day.
- /api 
  Self documenting HTML page that provides a brief description of the endpoints. It's impossible to add a new end point without having the page self update. This will allow the API documentation to always stay relavant.
  api
- /nodes
  Returns an array of current hosts, and the "proxies" and ports that the nodes attached with. The "proxy" ID is a unique key that describes individual proxy to allow further queries without name collisions. The proxy ID is a base64 encoded name of the proxie's getId(), which is roughly http://10.63.95.65:5555
  api
- /proxy/PROXY-ID
  Returns information about given proxy, such as registered capabilities and the config the proxy was launched with.
  api
-  /sessions
  A hash map of current session IDs (TestSession.getExternalKey()). Each session has a host and proxy ID attached, which gives ability to look up the host of all current tests easily.
  api
- /session/SESSION-ID
  Retrives some information about the current test session, wheter it has been orphaned or not, and what proxy it is connected to.
  api

Any feedback that will help me get this api into the grid is extremely welcome!

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/265)

PR #2771, opened 2016-09-14
HubStatusServlet - use GET with query string instead of GET with body

tommywo:

### DIFF
--- a/java/server/src/org/openqa/grid/web/servlet/HubStatusServlet.java
+++ b/java/server/src/org/openqa/grid/web/servlet/HubStatusServlet.java
@@ -33,6 +33,7 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
@@ -55,9 +56,9 @@ import javax.servlet.http.HttpServletResponse;
  *      ]
  * }
  *
- * if no param is specified, all params known to the hub are returned.
+ * alternatively you can use a query string ?configuration=timeout,servlets
  *
- * {"configuration": []  }
+ * if no param is specified, all params known to the hub are returned.
  *
  */
 public class HubStatusServlet extends RegistryBasedServlet {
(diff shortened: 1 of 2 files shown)


lukeis:
can you make it allow both POST data and url parameter? I'm a bit concerned here about breaking people unnecessarily.

@mach6

mach6:
I share the concerns of @lukeis -- It should continue support (although it is odd from an HTTP protocol point of view) `GET with body` too.

One additional suggestion I have -- Can you make the default behavior when the query string is empty be return an _application/json_ document with all params known by the servlet (as the javadoc for the servlet states, but currently does not do)?  
++ _See javadoc line_ 58  -- `if no param is specified, all params known to the hub are returned.`

All said, personally I'm in the opinion that the `HubStatusServlet` needs to be deprecated in `3.0.0`  and replaced with a REST api for getting "hub" status which includes the information this servlet provides and more.

tommywo:
It's now backward compatible with `GET with body`. 
I like the idea of REST API (i see there is pr open for it already -#265)

java/server/src/org/openqa/grid/web/servlet/HubStatusServlet.java
```diff
@@ -101,7 +102,10 @@ private JsonObject getResponse(HttpServletRequest request) throws IOException {
       if (request.getInputStream() != null) {
         JsonObject requestJSON = getRequestJSON(request);
         List<String> keysToReturn = null;
-        if (requestJSON != null && requestJSON.has("configuration")) {
+
+        if (request.getParameter("configuration")!=null && !"".equals(request.getParameter("configuration"))) {
```
mach6: inconsistent use of whitespace here

java/server/test/org/openqa/grid/internal/StatusServletTests.java
```diff
@@ -273,6 +274,34 @@ public void testHubGetSpecifiedConfig() throws IOException {
...
+  public void testHubGetSpecifiedConfigWithQueryString() throws IOException {
+
+    HttpClient client = httpClientFactory.getHttpClient();
+
+    ArrayList<String> keys = new ArrayList<>();
+    keys.add(URLEncoder.encode("timeout", "UTF-8"));
+    keys.add(URLEncoder.encode("I'm not a valid key", "UTF-8"));
+    keys.add(URLEncoder.encode("servlets", "UTF-8"));
+
+    String query = "?configuration="+String.join(",",keys);
```
mach6: inconsistent use of whitespace here

tommywo:
@mach6 done

tommywo:
@lukeis is it ok now? can you merge?

PR #2789, opened 2016-09-16
Fix #2721 and cleanup/refactor node registration request

### LINKED ISSUE #2721: Settings supplied via json are not applied when starting hub & node using selenium-server-standalone-3.0.0-beta2.jar
## Meta -
OS:  Windows 10

Selenium Version:  3.0.0-beta2

Browser:  N/A

Browser Version:  

## Expected Behavior -
Settings from hubConfig.json & nodeConfig.json are applied when starting selenium-server-standalone as hub or node (the attached hubConfig.json is set to run the hub on port 7777 with timeout:15 & browserTimeout:15, and nodeConfig.json is set to run the node on port 6666 with fairly standard settings)
## Actual Behavior -
The hub runs on port 4444 with default timeouts set and the node on 5555.
There is no mention of the hubConfig.json settings under the 'the final configuration comes from' heading when viewing the Grid Console page
## Steps to reproduce -

1. Unpack the attached example.zip to any location on your machine
2. Run grid-hub.bat
3. Run grid-node.bat
4. Open http://localhost:4444/grid/console?config=true&configDebug=true

## Additional Info -
When swapping out selenium-server-standalone-3.0.0-beta2.jar for the 2.53.1 version (and changing the SELENIUM_VERSION value to 2.53.1 in both batch files), all settings in the json files are applied as expected when starting the hub & node
The same behaviour is also seen when running with selenium-server-standalone-3.0.0-beta1.jar

[example.zip](https://github.com/SeleniumHQ/selenium/files/455622/example.zip)

mach6:

---
- Fix serialization/deserialization of `List<DesiredCapabilities>`
  for `GridNodeCofiguration`
- Change the format of node config files for Selenium V3 - There
  is no need to have the redundant `configuration: { ... }` section.
- Change the message in `RegistrationRequest` such that it does not
  contain dupicative information in separate json objects ..
  `"id"` and `"capabilities"` are part of `"configuration"` in Selenium V3
- Cleanup the api for `RegistrationRequest` -- will break compilation for
  people moving from 2.x -- removed the ability to change the
  GridNodeConfiguration reference via the `RegistrationRequest` object and
  removed all other setters.
- Add ability to handle booleans in `BrowserDesiredCapabilityConverter`
- Add ability to get a `GridRole` by String
- Update/Add tests

---

Note this change set makes the expected format of `nodeConfig.json` for V3 look like this;

``` json
{
  "capabilities": [
    {
      "browserName": "firefox",
      "maxInstances": 10
    }
  ],
  "proxy": "org.openqa.grid.selenium.proxy.DefaultRemoteProxy",
  "maxSession": 10,
  "port": 5557,
  "register": true,
  "registerCycle": 4000
}
```

instead of this;

``` json
{
  "capabilities": [
    {
      "browserName": "firefox",
      "maxInstances": 10
    }
  ],
  "configuration": {
      "proxy": "org.openqa.grid.selenium.proxy.DefaultRemoteProxy",
      "maxSession": 10,
      "port": 5557,
      "register": true,
      "registerCycle": 4000
  }
}
```

The same is true for the registration request that is sent from the node to the hub. Also note, this change set handles the older registration request from 2.x and 3.0.0-beta nodes

### DIFF
--- a/java/server/src/org/openqa/grid/common/GridRole.java
+++ b/java/server/src/org/openqa/grid/common/GridRole.java
@@ -20,24 +20,46 @@ package org.openqa.grid.common;
 public enum GridRole {
   NOT_GRID, HUB, NODE;
 
+  private static final String WD_S = "wd";
+  private static final String WEBDRIVER_S = "webdriver";
+  private static final String NODE_S = "node";
+  private static final String HUB_S = "hub";
+  private static final String STANDALONE_S = "standalone";
+
   public static GridRole get(String role) {
     if (role == null || role.equals("")) {
       return NOT_GRID;
     }
     switch (role) {
-      case "wd":
-      case "webdriver":
-      case "node":
+      case WD_S:
+      case WEBDRIVER_S:
+      case NODE_S:
         return NODE;
 
-      case "hub":
+      case HUB_S:
         return HUB;
 
-      case "standalone":
+      case STANDALONE_S:
         return NOT_GRID;
 
       default:
         return null;
+    }
+  }
+
+  public String toString() {
+    switch (this) {
+      case NODE:
+        return NODE_S;
+
+      case HUB:
+        return HUB_S;
+
+      case NOT_GRID:
+        return STANDALONE_S;
+
+      default:
+        throw new IllegalStateException("Unrecognized GridRole");
     }
   }
 }
--- a/java/server/src/org/openqa/grid/common/RegistrationRequest.java
+++ b/java/server/src/org/openqa/grid/common/RegistrationRequest.java
@@ -17,34 +17,27 @@
 
 package org.openqa.grid.common;
 
-import com.google.gson.Gson;
-import com.google.gson.JsonArray;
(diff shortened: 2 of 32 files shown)


java/server/src/org/openqa/grid/common/GridRole.java
```diff
@@ -40,4 +40,17 @@ public static GridRole get(String role) {
         return null;
     }
   }
+
+  public static String asString(GridRole role) {
```
mach6: @lukeis Ah yea,makes sense. Thanks!  I changed the method to `@Override public String toString()` ... It's not used anywhere in the current code base, but seemed like a good addition. I'm happy to remove it, if you disagree.

lukeis: happy to have it in there the way you have it now. Thanks!

lukeis:
everything looks pretty good to me... i'm  hoping you've done some of your own testing with this :)

Also... why are you reading this... you're supposed to be on vacation! :-P

PR #2815, opened 2016-09-22
Increase service shutdown timeout in ruby bindings

jbarbuto:

We've been getting intermittent `Net::ReadTimeout` exceptions with
builds on our hosted CI provider. Since increasing `STOP_TIMEOUT` to 20,
these exceptions have stopped (so far). This ends up increasing
`http.read_timeout` from 2 to 10 (`STOP_TIMEOUT / 2`).

This timeout appears to have been initially defined for Chrome on Apr 2011
(bb25ecfb6ece8edbc5034182d41a91ed2116b561), but increased Chrome memory
usage since then may cause shutdown to take longer, and the increased use
of hosted CI providers (with potential load issues due to noisy neighbors)
may also call for a longer timeout. Also note the python bindings appear to
have 30 seconds as a timeout:
https://github.com/SeleniumHQ/selenium/blob/a1b3b478/py/selenium/webdriver/common/service.py#L120

There's more background on the read timeout issue at
https://github.com/SeleniumHQ/selenium/pull/1684.

### DIFF
--- a/rb/lib/selenium/webdriver/common/service.rb
+++ b/rb/lib/selenium/webdriver/common/service.rb
@@ -34,7 +34,7 @@ module Selenium
     class Service
       START_TIMEOUT       = 20
       SOCKET_LOCK_TIMEOUT = 45
-      STOP_TIMEOUT        = 5
+      STOP_TIMEOUT        = 20
 
       attr_accessor :host
 


jbarbuto:
I increased the proposed timeout further from 10 to 20, since we saw one timeout with 10. This matches `START_TIMEOUT` as well.

titusfortner:
Merged - baf180f
Thanks for your contribution!

PR #2866, opened 2016-10-03
Element ... is not clickable at point

phoe:
The message "Element is not clickable at point..." gives no information about what element was meant to be clicked, which slows down debugging.

This pull request adds more information to the error message to the Firefox driver and the Chrome driver.

### DIFF
--- a/javascript/chrome-driver/atoms.js
+++ b/javascript/chrome-driver/atoms.js
@@ -240,14 +240,9 @@ webdriver.chrome.isElementClickable = function(elem, coord) {
     return makeResult(
         false, 'Element is not clickable at point ' + coordStr);
   }
-  var elemAtPointHTML = elemAtPoint.outerHTML;
-  if (elemAtPoint.hasChildNodes()) {
-    var inner = elemAtPoint.innerHTML;
-    var closingTag = '</' + elemAtPoint.tagName + '>';
-    var innerStart = elemAtPointHTML.length - inner.length - closingTag.length;
-    elemAtPointHTML = elemAtPointHTML.substring(0, innerStart) + '...' +
-        elemAtPointHTML.substring(innerStart + inner.length);
-  }
+  var elemAtPointHTML = elemAtPoint.outerHTML.replace(elemAtPoint.innerHTML, 
+                                                      elemAtPoint.hasChildNodes() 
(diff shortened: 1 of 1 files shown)


javascript/chrome-driver/atoms.js
```diff
@@ -259,9 +259,17 @@ webdriver.chrome.isElementClickable = function(elem, coord) {
     }
     parentElemIter = parentElemIter.parentNode;
   }
+  var elemHTML = elem.outerHTML;
+  if (elem.hasChildNodes()) {
+    var inner = elem.innerHTML;
```
lukeis: this if block can be reduced to:

    elemHTML = elem.outerHTML.replace(elem.innerHTML, '...');

lukeis: or remove the if block and just have:

    var elemHTML = elem.outerHTML.replace(elem.innerHTML, elem.hasChildNodes() ? '...' : '');

phoe: I copied the whole block from line 244. Does your reduction apply to that code block as well?

lukeis: ah, yes it does.

phoe: @lukeis Done. Awaiting further review.

phoe:
@lukeis Thanks! Which beta will this be released in and when can I expect that beta to appear?

lukeis:
likely not until 3.0.0 :) I don't think we're doing anymore betas...
https://seleniumhq.wordpress.com/2016/10/04/selenium-3-is-coming/

lukeis:
and for chrome, not until the next version of chromedriver

meet-sapu:
hi @phoe @lukeis   , This issue was headache to me while writing a web scraping script in python , I solution for this came out to be simple , I simply scrolled down(or it could be up in your case) the webpage and then clicking the button worked for , as it turns out it the webdriver is firefox or internet explorer this is not n issue but it's an issue for chrome webdriver .

phoe:
@meet-sapu We worked around this issue exactly the same say, but as you said, it's a workaround.

Also, I experienced the contrary - this issue manifested itself in Firefox for us.

PR #2870, opened 2016-10-04
add invisibility_of expected condition

aiudirog:

I wanted to wait for a progress bar to disappear after a file upload
and I already had the WebElement.  When I found there wasn't an
opposite to visibility_of, I wrote this one since it wasn't a hard
addition and hoped I could contribute it.

I have tested it in my personal code, but if there are any extra test
cases that need to be written for it, I can contribute those too.

### DIFF
--- a/py/selenium/webdriver/support/expected_conditions.py
+++ b/py/selenium/webdriver/support/expected_conditions.py
@@ -203,6 +203,29 @@ class invisibility_of_element_located(object):
             return True
 
 
+class invisibility_of(object):
+    """ An expectation for checking that an element that is either invisible
+    or not present in the DOM.
+
+    element is the WebElement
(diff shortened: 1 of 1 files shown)


BeyondEvil:
Would it make sense to add ElementNotVisibleException as well to the list of exceptions, @aiudirog ?

To handle the situation where the element is present (ie not triggering NoSuchElement) and not stale.

aiudirog:
Yes, I  think that could be pretty useful. Not necessarily in the context of this function, but maybe in places like attempting to click on an element that isn't visible. It is still there and should raise an exception reflecting that it is there but can't be seen.

lmtierney:
It appears this has already been added with #6163. I'm sorry we didn't get around to merging this one.

PR #2880, opened 2016-10-05
JRuby compatibility

joankaradimov:
I am using selenium 2.53.4 on JRuby 9.1.5.0 with Capybara 2.9.2. I get test failures for some tests. Although the failing tests are always the same, the exception is different each time. It's usually one of these:
- NoMethodError: undefined method `+' for nil:NilClass
- IOError: closed stream
- Net::HTTPBadResponse: wrong status line: "Content-Type: application/json; charset=UTF-8"

The stack trace is always:

```
     ....... [some jruby stack frames that rspec refuses to display] ........
     # ./features/build/tmp/jrubyExec/gems/selenium-webdriver-2.53.4/lib/selenium/webdriver/remote/http/default.rb:111:in `response_for'
     # ./features/build/tmp/jrubyExec/gems/selenium-webdriver-2.53.4/lib/selenium/webdriver/remote/http/default.rb:60:in `request'
     # ./features/build/tmp/jrubyExec/gems/selenium-webdriver-2.53.4/lib/selenium/webdriver/remote/http/common.rb:59:in `call'
     # ./features/build/tmp/jrubyExec/gems/selenium-webdriver-2.53.4/lib/selenium/webdriver/remote/bridge.rb:649:in `raw_execute'
     # ./features/build/tmp/jrubyExec/gems/selenium-webdriver-2.53.4/lib/selenium/webdriver/remote/bridge.rb:627:in `execute'
     # ./features/build/tmp/jrubyExec/gems/selenium-webdriver-2.53.4/lib/selenium/webdriver/remote/bridge.rb:604:in `find_elements_by'
     # ./features/build/tmp/jrubyExec/gems/selenium-webdriver-2.53.4/lib/selenium/webdriver/common/search_context.rb:84:in `find_elements'
     # ./features/build/tmp/jrubyExec/gems/capybara-2.9.2/lib/capybara/selenium/node.rb:155:in `find_css'
     # ./features/build/tmp/jrubyExec/gems/capybara-2.9.2/lib/capybara/node/base.rb:103:in `find_css'
     # ./features/build/tmp/jrubyExec/gems/capybara-2.9.2/lib/capybara/queries/selector_query.rb:116:in `block in resolve_for'
     # ./features/build/tmp/jrubyExec/gems/capybara-2.9.2/lib/capybara/node/base.rb:81:in `synchronize'
     # ./features/build/tmp/jrubyExec/gems/capybara-2.9.2/lib/capybara/queries/selector_query.rb:114:in `resolve_for'
     # ./features/build/tmp/jrubyExec/gems/capybara-2.9.2/lib/capybara/node/finders.rb:38:in `block in find'
     # ./features/build/tmp/jrubyExec/gems/capybara-2.9.2/lib/capybara/node/base.rb:85:in `synchronize'
     # ./features/build/tmp/jrubyExec/gems/capybara-2.9.2/lib/capybara/node/finders.rb:33:in `find'
     # ./features/build/tmp/jrubyExec/gems/capybara-2.9.2/lib/capybara/session.rb:735:in `block in find'
     # ./features/build/tmp/jrubyExec/gems/capybara-2.9.2/lib/capybara/dsl.rb:52:in `block in find'
     ......... [some irrelevant user stack frames] ......
```

Note that the line numbers in `default.rb` might not be correct since I had added some debug output. More specifically I changed this:

``` rb
def response_for(request)
  http.request request
end
```

... to this...

``` rb
def response_for(request)
  begin
    print 'doing... '
    result = http.request request
    puts 'done.'
    return result
  rescue Exception => e
    puts 'exception'
  end
end
```

Part of the output was:

```
doing... done.
doing... done.
doing... done.
doing... doing... done.
exception
```

I would assume this is something that does not happen in MRI. But being GIL-free JRuby is affected. The code in this PR resolves the problem for me.

I am not sure if this isn't something that should be fixed in Capybara. However, as far as I can see, the code in this class is not meant to be used with HTTP pipelining (I believe the `persistent.rb` class is meant for that). So it could make some sense to have the monitor synchronized block there.

### DIFF
--- a/rb/lib/selenium/webdriver/remote/http/default.rb
+++ b/rb/lib/selenium/webdriver/remote/http/default.rb
@@ -26,6 +26,8 @@ module Selenium
       module Http
         # @api private
         class Default < Common
+          include MonitorMixin
+
           attr_accessor :proxy
 
           private
@@ -100,7 +102,9 @@ module Selenium
           end
 
           def response_for(request)
-            http.request request
+            synchronize do
+              http.request request
+            end
           end
 
           def new_http_client


titusfortner:
@joankaradimov Thanks for this submission, I'll take a look at it soon. In the meantime, can you check the box to indicate you've signed the CLA? Thanks

joankaradimov:
Checked!

headius:
This does seem like thread unsafety somewhere inside `http.request`, and presumably `http` is being shared across parallel calls? Fixing this as you have seems quite reasonable...it's a small change and you're protecting Selenium from a downstream library's flaw. But it would be worth discussing with Capybara folks to see if this is *expected* to be thread-safe, so it can be fixed there as well.

titusfortner:
@twalpole - any input on this one / how it affects Capybara? I don't actually use or test either Selenium or Watir with JRuby; do you for Capybara?

twalpole:
@titusfortner We haven't been testing Capybara with JRuby because it hasn't worked correctly on Travis in quite some time.

p0deje:
I'm a bit worried by the fact there are nothing that looks thread-unsafe in Selenium code, meaning that it might be a race-condition in `net/http` itself. Is there any chance to get a test case that can be run against JRuby for debugging and narrowing bug?

> However, as far as I can see, the code in this class is not meant to be used with HTTP pipelining (I believe the persistent.rb class is meant for that).

AFAIK this is something that `net/http` handles itself. When we call `Net::HTTP.new` instead of `Net::HTTP.start`, it open connection and stores it for the following requests. Since we never explicitly call `#finish` on connection, it is not automatically closed.

I wonder if that is a real reason of this issue and we should instead close connection after making request (though I have no idea how this might affect performance/stability).

> This does seem like thread unsafety somewhere inside http.request, and presumably http is being shared across parallel calls? Fixing this as you have seems quite reasonable...it's a small change and you're protecting Selenium from a downstream library's flaw.

Agree on that, though if it's really JRuby issue, I'd try to figure it out along with applying hotfix.

> But it would be worth discussing with Capybara folks to see if this is expected to be thread-safe, so it can be fixed there as well.

Capybara itself does not make requests, does it? What kind of thread-unsafety this can be then?

joankaradimov:
This issue does not happen on Capybara 2.7.1 with Selenium 2.53.4. However, it does happen in Capybara 2.8.0 with Selenium 2.53.4.

Even if it is an issue in JRuby's `net/http` further bisection of Capybara might give an indication what the underlying problem is.

twalpole:
@headius It turned out the lack of entropy in /dev/random on Travis caused massive delays in either installing JRuby or running tests under JRuby and Travis would time out the test run due to lack of progress.

twalpole:
I can confirm that monkeypatching this change into selenium-webdriver 2.53.4 did fix the errors I was  seeing in the Capybara test suite once I got it running under JRuby again.

twalpole:
The change in Capybara 2.8 that makes this issue show up is lazily evaluating the results returned from selenium.  I can't see how it would be causing multiple requests to occur simultaneously though.

twalpole:
https://gist.github.com/twalpole/80c08b52fdc60326fafb93171d71d4c8  Is a tiny gist that produces the error on a pretty regular basis for me.

twalpole:
@headius  I think what I'm seeing with JRuby 9.1.5.0 is that when #first is called on a lazy enumerator the evaluation is done in a separate thread which returns the first result but then goes on to incorrectly evaluate a second element in the enumerator.  Is that possible?

headius:
@twalpole `Enumerator` use can trigger multiple threads to be executing, since `Enumerator#next` works like a Fiber and has to be driven by a native thread. However, you should normally not see the "nexter" thread and the main thread running at the same time, just like with Fiber.

What you describe sounds like it might be a bug in the way the threaded `next` logic works; it's greedily trying to get the next result before it has been requested.

headius:
Hmm...well a simple attempt to reproduce this didn't raise any flags:

```
$ jruby -e "i = 0; e = Enumerator.new {|y| p i; y.yield i; i+=1}; e.first; sleep"
0
```

I realized after my last comment that first does not use the `next` and its thread to do anything; in the above case, all it does is fire off an `each` call that terminates immediately after yielding the first result, all on the same thread.

@twalpole Can you show me the code you think is having problems under JRuby?

twalpole:
@headius I posted a gist above that will reproduce the issue.  The elements found by selenium-webdriver are stored in a Capybara::Result object - https://github.com/jnicklas/capybara/blob/master/lib/capybara/result.rb#L22 .  Under JRuby calling first on an instance of the Result object causes the lazy select block - https://github.com/jnicklas/capybara/blob/master/lib/capybara/result.rb#L29 - to be evaluated twice  (once after the result of first has been returned and the main thread has continued).  It is only evaluated once under MRI.

twalpole:
@headius  Here's a simplified example

```
class A
  include Enumerable

  def initialize()
    @result_enum = ["Text", "Text", "Text"].lazy.select do |t|
      puts "evaluating lazy select"
      t == "Text"
    end
  end

  def each(&block)
    loop do
      block.call(@result_enum.next)
    end
  end
end

A.new().first
```

Using MRI that outputs "evaluating lazy select" once -- with JRuby twice

headius:
Ahh this may be a bug in lazy enumeration, then, which we implement in Ruby (borrowed from a backport project).

Great reproduction! I'll file a JRuby issue for you :-)

twalpole:
@titusfortner, @joankaradimov I've fixed this in Capybara by disabling the lazy results evaluation when using JRuby - it will go out in 2.10.1 some time this weekend.

headius:
We should be able to have a fix for this in JRuby 9.1.6.0 (next release, some time before RubyConf).

p0deje:
It feels good to wake up and see issue completely resolved :) I think there is no longer need to synchronize request calls in selenium, so I guess we can close this PR.

Thank you everyone for reporting, nailing down and addressing the issue, you rock!

titusfortner:
Agreed, thanks @twalpole for your help on this!

PR #2882, opened 2016-10-06
New FirefoxDriver ctor precedence logic and moz:firefoxOptions su…

andreastt:
This change tries to make sense of the keyword arguments accepted by
FirefoxDriver's constructor in the transition phase between the Firefox
add-on driver and the Mozilla supported geckodriver HTTP service.

Prior to this patch, which arguments were consumed and respected varied
greatly on whether the `capabilities["marionette"]` key was set to true.
Following this change, the options passed on to geckodriver will
consistently come out of firefox.Options.

The patch implements precedence logic for `capabilities`,
`firefox_options`, and `firefox_binary`/`firefox_profile`.  They will
populate the underlying data model in that order, which means a binary-
or profile location set in `firefox_options` will override one set in the
`capabilities` dictionary, and that `firefox_binary` and `firefox_profile`
will override both of these.

The goal is to make it predictable what key is chosen, and to make the
underlying data model somewhat easier to understand.  The latter is
particularly important because the extension connection used for the old
Firefox add-on does not accept a binary, profile, or arguments through
capabilities since it is the Python bindings themselves that manage the
lifetime of the browser.

In the case of geckodriver, which is an external HTTP service, it
manages the lifetime and accepts a `moz:firefoxOptions` dictionary that
is not unlike `chromeOptions` as of version 0.11.0.  Versions prior to
geckodriver 0.11.0 accepted `firefox_binary`, `firefox_profile`, and
`firefox_args` as top-level entries on the capabilities JSON Object.
This is no longer supported behaviour by either the Python bindings
or geckodriver.

Fixes https://github.com/mozilla/geckodriver/issues/228.

### DIFF
--- a/py/selenium/webdriver/firefox/options.py
+++ b/py/selenium/webdriver/firefox/options.py
@@ -16,81 +16,98 @@
 # under the License.
 
 from selenium.common.exceptions import WebDriverException
+from selenium.webdriver.firefox.firefox_binary import FirefoxBinary
 from selenium.webdriver.firefox.firefox_profile import FirefoxProfile
 
 
+class Log(object):
+    def __init__(self):
+        self.level = None
+
+    def to_capabilities(self):
+        if self.level is not None:
+            return {"log": {"level": self.level}}
+        return {}
+
+
 class Options(object):
+    KEY = "moz:firefoxOptions"
 
     def __init__(self):
-        self._binary_location = ''
+        self._binary = None
         self._profile = None
         self._arguments = []
-        self._extension_files = []
-        self._extensions = []
-        self._firefox_options = {}
+        self.log = Log()
+
+    @property
+    def binary(self):
+        """Returns the location of the binary."""
+        return self._binary
+
+    @binary.setter
+    def binary(self, new_binary):
+        """Sets location of the browser binary, either by string or
+        ``FirefoxBinary`` instance.
+
+        """
+        if not isinstance(new_binary, FirefoxBinary):
+            new_binary = FirefoxBinary(new_binary)
+        self._binary = new_binary
 
     @property
     def binary_location(self):
-        """
-        Returns the location of the binary otherwise an empty string
-        """
-        return self._binary_location
+        return self.binary
 
-    @binary_location.setter
+    @binary.setter
     def binary_location(self, value):
-        """
-        Allows you to set where the firefox binary lives
-
-        :Args:
-         - value: path to the firefox binary
-        """
-        self._binary_location = value
-
-    @property
-    def arguments(self):
-        """
-        Returns a list of arguments needed for the browser
-        """
-        return self._arguments
-
-    def add_argument(self, argument):
-        """
-        Adds an argument to the list
-
-        :Args:
-         - Sets the arguments
-        """
-        if argument:
-            self._arguments.append(argument)
-        else:
-            raise ValueError("argument can not be null")
+        self.binary = value
 
     @property
     def profile(self):
-        """
-            Returns a FirefoxProfile object if one has been set before else None
-
-        """
+        """Returns the Firefox profile to use."""
         return self._profile
 
     @profile.setter
-    def profile(self, value):
-        if not isinstance(value, FirefoxProfile):
-            raise WebDriverException("When passing in a value to profile,"
-                                     " please pass in a FirefoxProfile object.")
-        self._profile = value
+    def profile(self, new_profile):
+        """Sets location of the browser profile to use, either by string
+        or ``FirefoxProfile``.
+
+        """
+        if not isinstance(new_profile, FirefoxProfile):
+            new_profile = FirefoxProfile(new_profile)
+        self._profile = new_profile
+
+    @property
+    def arguments(self):
+        """Returns a list of browser process arguments."""
+        return self._arguments
+
+    def add_argument(self, argument):
+        """Add argument to be used for the browser process."""
+        if argument is None:
+            raise ValueError()
+        self._arguments.append(argument)
 
     def to_capabilities(self):
-        """
-            Creates a capabilities with all the options that have been set and
+        """Marshals the Firefox options to a `moz:firefoxOptions`
+        object.
 
-            returns a dictionary with everything
         """
-        desired = {}
-        if self.binary_location:
-            desired["binary"] = self.binary_location
-        if self._profile:
-            desired["firefox_profile"] = self._profile.encoded
-        desired["args"] = self.arguments
-        capabilities = {"desiredCapabilities": desired}
-        return capabilities
+        # This intentionally looks at the internal properties
+        # so if a binary or profile has _not_ been set,
+        # it will defer to geckodriver to find the system Firefox
+        # and generate a fresh profile.
+        opts = {}
+
+        if self._binary is not None:
+            opts["binary"] = self._binary._start_cmd
+        if self._profile is not None:
+            opts["profile"] = self._profile.encoded
+        if len(self._arguments) > 0:
+            opts["args"] = self._arguments
+
+        opts.update(self.log.to_capabilities())
+
+        if len(opts) > 0:
+            return {Options.KEY: opts}
+        return {}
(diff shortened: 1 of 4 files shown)


andreastt:
I found the tests can be run like this:

     % tox -e py27-marionette py/test/selenium/webdriver/marionette/mn_options_tests.py -- -r=aPp -n0

py/selenium/webdriver/firefox/webdriver.py
```diff
@@ -47,35 +47,99 @@ class WebDriver(RemoteWebDriver):
...
+        instance will always come from the most specific variable.
+        In this case that would be `firefox_profile`.  This will result in
+        `firefox_options.profile` to be ignored because it is considered
+        a less specific setting than the top-level `firefox_profile`
+        keyword argument.  Similarily, if you had specified a
+        `capabilities["firefoxOptions"]["profile"]` Base64 string,
+        this would rank below `firefox_options.profile`.
+
+        :param firefox_profile: Instance of ``FirefoxProfile`` object.
+        :param firefox_binary: Full path to the Firefox binary to use.
```
davehunt: The current implementation supports either a string or a `FirefoxBinary` instance.

py/selenium/webdriver/firefox/webdriver.py
```diff
@@ -47,35 +47,99 @@ class WebDriver(RemoteWebDriver):
...
+        if capabilities is None:
+            capabilities = DesiredCapabilities.FIREFOX.copy()
+
+        if capabilities.get("binary"):
+            self.binary = capabilities["binary"]
+
+        # firefox_options overrides capabilities
+        if firefox_options is not None:
+            self.binary = firefox_options.binary or self.binary
+            self.profile = firefox_options.profile or self.profile
```
davehunt: `self.profile` will be `None`

py/selenium/webdriver/firefox/webdriver.py
```diff
@@ -47,35 +47,99 @@ class WebDriver(RemoteWebDriver):
...
+            self.binary = firefox_binary
+            self.options.binary = firefox_binary
+        if firefox_profile is not None:
+            self.profile = firefox_profile
+            self.options.profile = firefox_profile
+
+        # disable native events if globally disabled
+        if self.profile is not None:
+            self.profile.native_events_enabled = (
+                self.NATIVE_EVENTS_ALLOWED and self.profile.native_events_enabled)
```
davehunt: Previously if no profile was specified then we created a default one, which meant we were able to set the native events for the profile. By leaving GeckoDriver to create a default profile, we won't have the native events preferences set. I'm not overly familiar with these preferences, so I'm not sure if it matters.

py/selenium/webdriver/firefox/webdriver.py
```diff
@@ -47,35 +47,99 @@ class WebDriver(RemoteWebDriver):
...
+
+        # W3C remote
+        # TODO(ato): Perform conformance negotiation
```
davehunt: Does this mean as an alternative to checking the `marionette` capability? If so, +1.

py/test/selenium/webdriver/marionette/mn_options_tests.py
```diff
@@ -15,25 +15,114 @@
...
-from selenium.webdriver.firefox.options import Options
 
 
 @pytest.fixture
-def driver(capabilities):
-    options = Options()
-    driver = Firefox(
-        capabilities=capabilities,
-        firefox_options=options)
+def driver(options = None):
```
davehunt: What is the `options = None` doing? I'm not sure it's needed.

py/test/selenium/webdriver/marionette/mn_options_tests.py
```diff
@@ -15,25 +15,114 @@
...
+    def test_profile(self):
+        opts = Options()
+        assert opts.profile is None
+
+        other_profile = FirefoxProfile()
+        assert other_profile != opts.profile
+        opts.profile = other_profile
+        assert other_profile == opts.profile
+
+        with TemporaryDirectory() as path:
```
davehunt: You should be able to use the built-in pytest fixture `tmpdir` instead: http://docs.pytest.org/en/latest/tmpdir.html

shs96c:
FYI, the java bindings throw an exception if the user has set profiles in more than one place that aren't the same profile. Silently overwriting a user choice seemed less than idea.

py/selenium/webdriver/firefox/webdriver.py
andreastt: I think one possible approach is to look at what version the system Firefox is, and if it is less than or equal to 47.0.1 default to the old add-on approach, otherwise implicitly go for geckodriver.

I’ll file this under “possible future improvements” (-:

andreastt:
It is however the only backwards compatible solution since the Python bindings have supported that behaviour for years.

py/selenium/webdriver/firefox/webdriver.py
andreastt: I think you’re right this might matter when you’re _not_ using geckodriver. It is an intentional change not to pass a profile to geckodriver as it is able to create one itself.

py/test/selenium/webdriver/marionette/mn_options_tests.py
andreastt: Reverted.

py/test/selenium/webdriver/marionette/mn_options_tests.py
andreastt: Using `tmpdir_factory` now.

py/selenium/webdriver/firefox/webdriver.py
andreastt: Addressed this, in a few other bugs that were if `firefox_options` was not set.

py/selenium/webdriver/firefox/webdriver.py
andreastt: Addressed.

py/selenium/webdriver/firefox/options.py
```diff
@@ -16,81 +16,90 @@
...
-        self._firefox_options = {}
+        self.log = Log()
 
     @property
-    def binary_location(self):
-        """
-        Returns the location of the binary otherwise an empty string
-        """
-        return self._binary_location
+    def binary(self):
```
AutomatedTester: Can we have the original method names back with deprecation warnings in addition to this change. These are in Selenium 2 and Selenium 3 is supposed to be purely removing RC, as requested by @shs96c

py/selenium/webdriver/firefox/webdriver.py
```diff
@@ -47,35 +48,115 @@ class WebDriver(RemoteWebDriver):
...
+
             executor = FirefoxRemoteConnection(
                 remote_server_addr=self.service.service_url)
             RemoteWebDriver.__init__(
                 self,
                 command_executor=executor,
                 desired_capabilities=capabilities,
                 keep_alive=True)
+
+            self._is_remote = True
```
AutomatedTester: This needs to be false.

`self._is_remote` is used to know when we need to ship a file over to a selenium server on a different machine.

py/test/selenium/webdriver/marionette/mn_options_tests.py
```diff
@@ -15,25 +15,97 @@
 # specific language governing permissions and limitations
 # under the License.
 
-
+import errno
```
AutomatedTester: This import doesnt appear to be used.

andreastt: Fixed.

py/selenium/webdriver/firefox/webdriver.py
andreastt: Fixed.

py/selenium/webdriver/firefox/options.py
andreastt: Fixed.

PR #2886, opened 2016-10-06
Attempt to diagnose py27-remote issues in Travis CI

davehunt:


tommywo:
this could be the problem:

Xlib:  extension "RANDR" missing on display ":99.0". 
https://travis-ci.org/SeleniumHQ/selenium/jobs/165659976#L1055

davehunt:
@tommywo as far as I can tell, we can safely ignore that warning.

lmtierney:
Closing... unless you want to try something more @davehunt. @barancev refactored the suite to re-use driver instance when it can so we may be in better shape now

PR #2890, opened 2016-10-07
Update calls that return a pure object with keys to look for 'value' key

AutomatedTester:
This is to be landed when https://github.com/mozilla/geckodriver/pull/263 lands otherwise we are going to have incompatibility due to changes in the spec.


shs96c:
You mean landing at the same time, or any time before?

AutomatedTester:
> You mean landing at the same time, or any time before?

What I dont want is broken users because that just makes both projects look bad... It really should be coordinated.

lmtierney:
Closing as this is quite old and the code has been updated with the changes since

PR #2905, opened 2016-10-11
Introduce marks for xfailing tests against specific drivers

davehunt:
This is not intended for merging as is, but I'd like to get some feedback from pytest devs (maybe @RonnyPfannschmidt and @The-Compiler) as this is currently using pytest internals and I'm wondering if there's a better approach. Basically this allows us to mark tests as expected to fail based on the value of a parameter for a fixture. Rather than an imperative xfail, this allows tests to run to completion, and therefore also allows them to xpass. /cc @AutomatedTester

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -21,6 +21,7 @@ import time
 import urllib
 
 import pytest
+from _pytest.skipping import MarkEvaluator
 
 from selenium import webdriver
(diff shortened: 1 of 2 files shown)


lmtierney:
You've introduced this functionality, closing the PR

PR #2927, opened 2016-10-14
Fixes error when calling Javascript from WebDriver Playback

### LINKED ISSUE #2916: "self" instead of "this" in remote-selenium-commandhandlers.js, line 376
## Meta -

Selenium Version:  
 IDE 2.9.1.1-signed

There's a bug in selenium-ide\chrome\content\remote-selenium-commandhandlers.js: line 376 should be `this.result = arg;` instead of `self.result = arg;`. The latter generates an error whenever we use getEval with WebDriver Playback.

A diff file is attached.
[remote-selenium-commandhandlers.js.diff.txt](https://github.com/SeleniumHQ/selenium/files/528070/remote-selenium-commandhandlers.js.diff.txt)

shrodi:

A undefined variable "self" was used instead of "this" on line 376.

Fixes #2916

### DIFF
--- a/ide/main/src/content/remote-selenium-commandhandlers.js
+++ b/ide/main/src/content/remote-selenium-commandhandlers.js
@@ -364,16 +364,16 @@ AccessorHandler.prototype.execute = function(seleniumApi, command) {
 };
 
 function AccessorResult(result) {
+  var self = this;
   //TODO Samit: should we modify this?
   if (result.terminationCondition) {
-    var self = this;
     this.terminationCondition = function() {
       return result.terminationCondition.call(self);
     };
   } else if (Deferred.isPromise(result)) {
     //We have a deferred, we need to change the terminating condition to include the resolution of the deferred
     result.done(function(arg) {
-      this.result = arg;
+      self.result = arg;
     });
     this.terminationCondition = function() {
       if (result.isRejected()) {


ide/main/src/content/remote-selenium-commandhandlers.js
```diff
@@ -373,7 +373,7 @@ function AccessorResult(result) {
   } else if (Deferred.isPromise(result)) {
     //We have a deferred, we need to change the terminating condition to include the resolution of the deferred
     result.done(function(arg) {
-      self.result = arg;
+      this.result = arg;
```
lukeis: this does not fix the problem, you would need to define above this function var self =this; and revert this change. or change the line below to `}.bind(this));`

shrodi: Ok, I think I get it now... Will change.

shrodi:
Ok I made the change. I preferred to move the declaration+definition of self outside of the "if" block instead of creating a new one: that way its scope will be shared by all anonymous functions below.

ide/main/src/content/remote-selenium-commandhandlers.js
shrodi: Change made in commit 66532680d5a1728dc4689955f32c07b6e7987871.

barancev:
Selenium IDE development has been discontinued. No more new features and no bug fixes. From Firefox 55 onwards, Selenium IDE will no longer work [1]. The last supported version is Firefox ESR 52.

Yes, we've started a project to develop a new tool that should replace Selenium IDE. Follow #4406 for the progress.

Sorry this issue has not beed properly and timely addressed, and thank you for submission!

[1] https://seleniumhq.wordpress.com/2017/08/09/firefox-55-and-selenium-ide/

PR #2929, opened 2016-10-15
[Ruby client] remove freeze for Selenium::WebDriver::SearchContext::FINDERS for appium_lib

KazuCocoa:
Remove `.freeze` for `Selenium::WebDriver::SearchContext::FINDERS`

I used `selenium-webdriver 3.0`'s ruby client with Appium. Then I found failed case to start devices via appium server and `appium_lib` which is appium ruby client. The `appium_lib` extends `Selenium::WebDriver::SearchContext::FINDERS` to extend some search elements for mobile application. ( [search results](https://github.com/appium/ruby_lib/search?utf8=%E2%9C%93&q=searchContext) )

BTW, `Selenium::WebDriver::SearchContext::FINDERS` is freezed from `selenium-webdriver 3.0`.
Could you unfreeze `FINDERS` to works fine `appium_lib` ? 

---

I accepted CLA.

### DIFF
--- a/rb/lib/selenium/webdriver/common/search_context.rb
+++ b/rb/lib/selenium/webdriver/common/search_context.rb
@@ -32,7 +32,7 @@ module Selenium
         partial_link_text: 'partial link text',
         tag_name: 'tag name',
         xpath: 'xpath'
-      }.freeze
+      }
 
       #
       # Find the first element matching the given arguments.


titusfortner:
That's a private API that Appium shouldn't be using. It's also a constant which should be immutable. This should be addressed in Appium code not here.

KazuCocoa:
ideally, I thinks so, too. thanks.

PR #2934, opened 2016-10-16
Removed JavaFX Pair class which was causing a compile error, replaced…

alexkogon:

… with custom implementation

Tested on AWS and failing job from  http://ci.seleniumhq.org/ now passes
[ec2-user@ip-172-31-30-159 selenium]$ ./go //java/client/test/org/openqa/selenium/remote:client-tests:run offline=true haltonerror=false haltonfailure=false log=truests.xml
(in /home/ec2-user/selenium)
Not using buckd because watchman isn't installed.
[-] PARSING BUCK FILES...FINISHED 1.9s [100%]
BUILT 95/96 JOBS 3.5s //java/client/test/org/openqa/selenium/remote:client-tests#testsjar
BUILT 96/96 JOBS 0.0s //java/client/test/org/openqa/selenium/remote:client-tests
[-] BUILDING...FINISHED 8.5s [100%]
Not using buckd because watchman isn't installed.
[-] PARSING BUCK FILES...FINISHED 2.0s [100%]
[-] BUILDING...FINISHED 2.6s [100%]
TESTING //java/client/test/org/openqa/selenium/remote:client-tests
PASS    <100ms 16 Passed   0 Skipped   0 Failed   org.openqa.selenium.remote.AugmenterTest
NOTESTS <100ms  0 Passed   0 Skipped   0 Failed   org.openqa.selenium.remote.BaseAugmenterTest
PASS    <100ms 16 Passed   0 Skipped   0 Failed   org.openqa.selenium.remote.ErrorHandlerTest
PASS    <100ms 11 Passed   0 Skipped   0 Failed   org.openqa.selenium.remote.JdkAugmenterTest
PASS     808ms 45 Passed   0 Skipped   0 Failed   org.openqa.selenium.remote.RemoteClientTests
PASS     303ms  5 Passed   0 Skipped   0 Failed   org.openqa.selenium.remote.RemoteLogsTest
PASS    <100ms  2 Passed   0 Skipped   0 Failed   org.openqa.selenium.remote.RemoteWebDriverInitializationTest
PASS    <100ms  6 Passed   0 Skipped   0 Failed   org.openqa.selenium.remote.internal.CircularOutputStreamTest
PASS    <100ms 15 Passed   0 Skipped   0 Failed   org.openqa.selenium.remote.internal.WebElementToJsonConverterTest
TESTS PASSED

### DIFF
--- a/java/client/test/org/openqa/selenium/remote/ErrorHandlerTest.java
+++ b/java/client/test/org/openqa/selenium/remote/ErrorHandlerTest.java
@@ -58,8 +58,7 @@ import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.interactions.InvalidCoordinatesException;
 import org.openqa.selenium.interactions.MoveTargetOutOfBoundsException;
 
-import javafx.util.Pair;
-
+import java.io.Serializable;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
@@ -74,6 +73,26 @@ import java.util.Set;
 @RunWith(JUnit4.class)
 public class ErrorHandlerTest {
   private ErrorHandler handler;
+
+  class Pair<K,V>
+    extends Object
+    implements Serializable {
+    K theKey;
+    V theValue;
+    Pair(K key, V value){
+      theKey=key;
+      theValue=value;
+    }
+
+
+    public K getKey() {
+      return theKey;
+    }
+
+    public V getValue() {
+      return theValue;
+    }
+  }
 
   @Before
   public void setUp() {


alexkogon:
FAIL    185.6s 722 Passed  60 Skipped   1 Failed   org.openqa.selenium.chrome.ChromeDriverTests
FAILURE org.openqa.selenium.chrome.ChromeDriverTests testBasicKeyboardInput: 
Expected: is "abc def"
      but: was "abc de"
java.lang.AssertionError: 
Expected: is "abc def"

I'm not sure it is possible my change caused this?

lukeis:
We have some flaky tests, just another one to catch :)

I think it would be better to move the class somewhere else. Not looking at the code base right now, can recommend a good place later if you need one.

alexkogon:
Sure, just let me know where you would like it; since it was only being used in this single class I just put it in as an inner class.

If we want to make it a general reusable class it should probably go somewhere where it is both usable in functional and testing code not just for testing...

Let me know and I can run refactor it out with the IDE quickly

alexkogon:
By the way there is also an HTMLUnit test failure when I ran them on AWS, not sure if this is significant or just flaky (the jobs aren't running on Jenkins at all because of the missing Python library). Not sure if these tests are run in Travis...

FAILURE org.openqa.selenium.htmlunit.HtmlUnitDriverTests testShouldBeAbleToInputJapanese: The element's value should either remain in Romaji or be converted properly. It was: --

alexkogon:
It should also be noted that the class I wrote only matches the usage in the Test class; if we want to make a general purpose class we may want to add more methods.

lukeis:
I just changed to using a Map instead, has the same concept with an Entry. Avoiding the 'pair' entirely for now.

Also I just noticed you were modifying the file to be executable? Try to avoid doing that in the future, there's no reason for java files to be executable.

99a8acef4c62dfcf912f9479d68b44965830a6b5

alexkogon:
Modifying the file to be executable? With chmod? I certainly didn't do anything like that; I did however use IntelliJ on Windows to edit it, maybe it was a side effect of that.

alexkogon:
As for using a map instead of the Pair I was merely doing what Simon asked me to do to replace the Pair class being used from JavaFX with a simple implementation that did the same thing. But I guess you prefer to just waste my time?

Issue #2950, opened 2016-10-17
Using Curb client causes 'RuntimeError: can't modify frozen Array'


clarkenciel:
## Meta -

OS: Ubuntu
Selenium Version:  3.0.0
## Expected Behavior -

Users should be able to use the Curb remote client in place of the standard Http client without any trouble.
## Actual Behavior -

Requiring `selenium/webdriver/remote/http/curb.rb` in a file causes a `RuntimeError`.
This is a result of style changes made in commit [cccc9fe4a7b4549bb6aa830397f437286b260f21](https://github.com/SeleniumHQ/selenium/commit/cccc9fe4a7b4549bb6aa830397f437286b260f21). Specifically, the `QUIT_ERRORS` constant in the `Bridge` class was frozen to fit with Rubocop's ["Style/MutableConstants" ](http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/MutableConstant) setting. As a result, when `curb.rb` tries to push `Curl::Err::RecvError` into that constant, the `RuntimeError` is thrown.
## Steps to reproduce -

require `selenium/webdriver/remote/http/curb`.

titusfortner:
fixed with #2951 
released in Selenium gem 3.0.4

PR #2951, opened 2016-10-18
Fix issue 2950

clarkenciel:
This is a proposed fix to [issue 2950](https://github.com/SeleniumHQ/selenium/issues/2950) that tries to fix the issue while not making rubocop angry by mutating constants. It converts the QUIT_ERRORS constant into a class variable and provides a class-level "setter," `protect_against`, on the `Bridge` class to add errors to the list of errors that `Bridge`s ignore when quitting.

### DIFF
--- a/rb/lib/selenium/webdriver/remote/bridge.rb
+++ b/rb/lib/selenium/webdriver/remote/bridge.rb
@@ -48,6 +48,19 @@ module Selenium
           COMMANDS[name] = [verb, url.freeze]
         end
 
+        @@quit_errors = []
+
+        #
+        # Adds an error to the list of errors that Bridge protects against when quitting
+        #
+        # @param error [Exception]
+        #   Exception that `Bridge`s will now protect against when quitting
+        
+        def self.protect_against(error)
+          @@quit_errors.push(error)
+        end
+        
+
         attr_accessor :context, :http, :file_detector
         attr_reader :capabilities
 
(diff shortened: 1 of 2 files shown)


titusfortner:
Oh good catch, we don't have any tests around this.

1. Naming. `#protect_against` is vague, perhaps `#add_quit_error`
2. Multiple Classes. These changes also need to be made to `W3CRemote` class
3. Missing Default. Needs to include by default: `[IOError]` instead of just `[]`
4. Class Variable. Shouldn't be needed. What about something like this?

```
        @quit_errors = [IOError]

        class << self; attr_reader :quit_errors; end

        def self.add_quit_error(error)
          @quit_errors << error
        end
```

That could be accessed by:
`*self.class.quit_errors`

Though, the more I think about this, the more I don't like it. This isn't a "constant that needs to be modified." This is a value that changes per instance depending on the class of the @http value in the options. I'm not sure the use case for alternating between http clients, but that would show why the above approach isn't good. Plus the only place that is using this is in the instance method `#quit`.

There's a few ways to do this, one would be to put the constant in the `Curb` class itself:
```
QUIT_ERRORS = [Curl::Err::RecvError].freeze
```
Then keep it as an instance variable in `Bridge` by adding to `Bridge#initialize`
```
          @quit_errors = [IOError]
          @quit_errors += @http.class::QUIT_ERRORS if defined?(@http.class::QUIT_ERRORS)
```
Then we'd reference `*@quit_errors` instead of `**@quit_errors`

titusfortner:
Merged with 0b688fd7a79fa92d4de069b68d4e072500ed1f2f
released with Selenium gem 3.0.4

Thanks for your contribution

PR #2962, opened 2016-10-19
[ruby] Make service stopping more robust

p0deje:
On Windows we might face random errors when starting/stopping service
often. One example is an ephemeral port exhaustion which raises
`Errno::ECONNRESET`. The errors are already documented in `SocketPoller`
so we can simple rescue them and proceed without raising exception since
after that process is stopped anyways.

Similar behavior is implemented in `Selenium::Server#stop`.

The failures have been caught while adding AppVeyor CI to Watir project.
There, we have doctests which stop/start browser between every test: [build 6](https://ci.appveyor.com/project/p0deje/watir/build/6/job/04a6wc7uyt2p2887), [build 9](https://ci.appveyor.com/project/p0deje/watir/build/%239/job/9tpf5ed4hxru0xuw) and [build 10](https://ci.appveyor.com/project/p0deje/watir/build/%2310/job/dd984f493076osm6).

@titusfortner Please take a look. If you're ok with the changes, I'll merge it as soon as I test it thoroughly.

### DIFF
--- a/rb/lib/selenium/webdriver/common/socket_poller.rb
+++ b/rb/lib/selenium/webdriver/common/socket_poller.rb
@@ -23,6 +23,9 @@ require 'socket'
 module Selenium
   module WebDriver
     class SocketPoller
+      NOT_CONNECTED_ERRORS = [Errno::ECONNREFUSED, Errno::ENOTCONN, SocketError]
+      NOT_CONNECTED_ERRORS << Errno::EPERM if Platform.cygwin?
+
       def initialize(host, port, timeout = 0, interval = 0.25)
         @host     = host
         @port     = Integer(port)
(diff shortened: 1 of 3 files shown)


rb/lib/selenium/webdriver/common/socket_poller.rb
```diff
@@ -23,6 +23,9 @@
 module Selenium
   module WebDriver
     class SocketPoller
+      NOT_CONNECTED_ERRORS = [Errno::ECONNREFUSED, Errno::ENOTCONN, SocketError]
```
titusfortner: Doesn't this need to include `Errno::ECONNRESET` to fix the issue?

p0deje: Ugh, that's right.

p0deje:
This hasn't made things more stable for me, so closing (at least for now).

PR #3006, opened 2016-10-25
Include LICENSE in MANIFEST.in for sdist

pmlandwehr:

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -17,4 +17,5 @@ include py/selenium/selenium.py
 include py/selenium/__init__.py
 include py/CHANGES
 include py/README.rst
+include LICENSE
 recursive-include selenium.egg-info *


AutomatedTester:
The `setup.py` has what licence this is, why include it in the actual egg/wheel ?

pmlandwehr:
O, interesting point hadn't realized. cc'ing @jakirkham for some clarity). I'm building selenium for [conda-forge](http://conda-forge.github.io (the feedstock is [here.](https://github.com/conda-forge/selenium-feedstock))We'd like to explicitly link to a copy of the license file itself, but that requires it be in the `sdist` as its own file. On the other hand, _if_ `conda` is ok with linking to the setup.py instead, the additional bundle is unnecessary.

AutomatedTester:
https://github.com/SeleniumHQ/selenium/blob/master/setup.py#L38 is where we say we use Apache2 in `setup.py`

jakirkham:
It would be great if open sourcing code/licensing was that easy. 😄

Unfortunately, open source licenses normally requires the "Distribution of License" pointed out by the Open Source Initiative (OSI) in [item 7]( https://opensource.org/osd-annotated ). Just about every open source license (certainly all of the ones that OSI approves) has a few words along these lines. In the case of the Apache 2.0 license, it is stated as such in item 4.1 (or 4.a).

>
> 4\. Redistribution. You may reproduce and distribute copies of the Work or Derivative Works thereof in any medium, with or without modifications, and in Source or Object form, provided that You meet the following conditions:
> 
>4.1. You must give any other recipients of the Work or Derivative Works a copy of this License...

xref: https://opensource.org/licenses/Apache-2.0
xref: https://www.apache.org/licenses/LICENSE-2.0

So we would very much appreciate having the license file included and packaged. This helps us distribute this nice piece of software to others. 😄

jakirkham:
Also should note IANAL.

pmlandwehr:
@jakirkham but setup.py does actually include the text of the Apache 2 text - it's just not in separate file. Can we use `license_file: setup.py` in meta.yaml for the conda build?

AutomatedTester:
Actually, all files that make it into the wheel/egg have the licence in the header.

jakirkham:
Wasn't sure what you were referring too at first. You are meaning this [license header]( https://github.com/SeleniumHQ/selenium/blob/e07eaf0e9b86e999b503ca41edec4cfa03fe43e4/setup.py#L3-L18 )? That's really not the same thing as the license file. Apache 2.0 also clarifies what the license is in section 1.

> "License" shall mean the terms and conditions for use, reproduction, and distribution as defined by Sections 1 through 9 of this document.

xref: https://opensource.org/licenses/Apache-2.0
xref: https://www.apache.org/licenses/LICENSE-2.0

<hr>

TBH I think @pmlandwehr has proposed a simple and reasonable change here. There is no harm (that I know of) in including the license file and noted benefits. Sorry if I'm just confused, but is there a reason not to merge this?

PR #3032, opened 2016-10-30
[py] EdgeDriver with log file and verbose option

Herst:

I took inspiration for how it's done with Firefox and IE to add support for logging to a file and setting a parameter (in this case `--verbose` for verbose output) to the EdgeDriver. Quite useful since the Microsoft's driver for Edge ("Microsoft WebDriver") still has quite some issues.


Herst:
Considering that I didn't change anything related to Firefox I think it is safe to say that I'm not responsible for that one test failing as I see it.

PR #3033, opened 2016-10-30
Node.js: Dependency versions npm ignore

paazmaya:
- When using the at least `>=` to mark a minimum version number, use the full version number.
- Ignore directories that should not be in the npm package, so it does not increase the resulting file size


jleyba:
I updated dependencies in dc84c6dc3f5e52c1f6aa7f1cd9bdf5eb225ec23d

Also, at this point, I don't see a point in excluding test dependencies. I like that users can checkout out from npm and run the tests themselves.

PR #3049, opened 2016-11-02
Avoid resolving host

asolntsev:

### DIFF
--- a/java/client/src/org/openqa/selenium/WebDriverException.java
+++ b/java/client/src/org/openqa/selenium/WebDriverException.java
@@ -19,8 +19,6 @@ package org.openqa.selenium;
 
 import org.openqa.selenium.internal.BuildInfo;
 
-import java.net.InetAddress;
-import java.net.UnknownHostException;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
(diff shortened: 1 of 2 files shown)


lukeis:
we like having the host name or ip address in the stacktrace though... it's thoroughly helpful in debugging systemic issues on particular hosts.

asolntsev:
@lukeis I understand. But still, you should just configure your logging properly instead of hoping that Selenium does it for you. For example, configure `log4j` to append ip to every log.

What if your test fails with `NullPointerException` instead of `WebDriverException`? They you don't have host/ip in logs. How do you debug systemic issues in this case?

What I want to say is that it's not a universal solution. But it has a very bad effect on selenium itself.

lukeis:
My use case isn't for 'localhost'. Only when running selenium standalone server on remote nodes (like in a grid configuration). Only if one customizes the standalone server could they get this information.

I agree that it's not ideal for networks that have various issues. If there's a way to get the java code to not do a lookup and just grab the system host / ip information (without pinging external servers like dns). That would be better.

asolntsev:
@lukeis Ok, you run standalone selenium server on a remote node. So what? You start it as a java process, right? You redirect its output to some log file, right? You know where this file is located. When you read file on a remote node, you now what node is it. You **don't need** to see ip/hostname in logs.

lukeis:
No... the remote node is discarded (VM) and the test receives the exception message and logs it.

When particular nodes are causing issues, this is where this information becomes very helpful.

asolntsev:
@lukeis Stop. `WebDriverException` writes an IP of machine where it's thrown. It means: ip of **machine where test is executed**, not the remote node.

lukeis:
that's incorrect, it does both. On the remote machine where the exception originated and on the local machine when using Remote Web Driver.

I'm merely trying to point out, this information is very useful to users (me being one of them). And we won't just be removing it because some networks may have problematic DNS. I would suggest either attempting to change the java code so it does not trigger a DNS lookup or adding a configuration flag of some kind to the java code where it will then skip the IP / hostname lookup.

asolntsev:
@lukeis Ok, thank you for the clarification.

Yes, I understand that this information is very useful to users. But I claim that this is not universal solution. In the following cases IP/host resolving **does not help you**:
1. Two nodes run on the same IP/host
2. WebDriverException failed to resolve ip/host
3. Your tests are green. You haven't got any `WebDriverException`. How do you get to know on which node tests have been executed then?

So this solution
1. Is slow
2. Is not universal (it doesn't always work)

I suggest that we need a better solution. I believe that Selenium Grid should provide the "name" or "id" of selected node (or something like that) when it chooses some node. 

P.S. Anyway, WebDriverException **should not** resolve ip/host of the local machine. This information is never useful, and this makes life harder of all those users who don't use Selenium Grid at all.

barancev:
Fixed time ago.

PR #3052, opened 2016-11-03
Fix Selenium 3.0.1 java script api - passing arguments for jar

sptq:

This commit is fixing error when we want use the new version of selenium

Because now the command is build to this form:
```
java -jar selenium-server-standalone-3.0.1.jar -Dwebdriver.ie.driver=IEDriverServer_Win32_2.53.1.exe
```

and for old `selenium-server-standalone-2.53.1.jar` is working but for new version command should look like this: 

```
java -Dwebdriver.ie.driver=IEDriverServer_Win32_2.53.1.exe -jar selenium-server-standalone-3.0.1.jar 
```

### DIFF
--- a/javascript/node/selenium-webdriver/remote/index.js
+++ b/javascript/node/selenium-webdriver/remote/index.js
@@ -489,7 +489,7 @@ class SeleniumServer extends DriverService {
           let port = resolved[0];
           let jvmArgs = resolved[1];
           let args = resolved[2];
-          return jvmArgs.concat('-jar', jar, '-port', port).concat(args);
+          return jvmArgs.concat(args).concat('-jar', jar, '-port', port);
         });
 
     super('java', {


lukeis:
this won't work... and you should be using jvmArgs for passing in things like webdriver.ie.driver, not `args`

Other args do require to be after the jar in the command line

sptq:
Ok, so the protractor repo have the problem, because they are using the constructor passing all arguments for jvm. https://github.com/angular/protractor/blob/master/lib/driverProviders/local.ts#L101 


When I've changed this line everything started to work.

lukeis:
@sptq that code needs to change to use jvmArgs instead of args 

here: https://github.com/angular/protractor/blob/master/lib/driverProviders/local.ts#L98

and then allow you as a user to pass in jvmArgs, similar to args -
https://github.com/angular/protractor/blob/master/lib/driverProviders/local.ts#L89-L91

with how you were changing it, other arguments for the standalone server (like -timeout) would attempt to be before the -jar, and thus be invalid.

Please submit a PR to protractor

sptq:
Thx for help :)

PR #3094, opened 2016-11-12
rb: Add property method to element (w3c get element property)

lmtierney:

### DIFF
--- a/rb/lib/selenium/webdriver/remote/bridge.rb
+++ b/rb/lib/selenium/webdriver/remote/bridge.rb
@@ -541,6 +541,11 @@ module Selenium
           execute :getElementAttribute, id: element.ref, name: name
         end
 
+        # Backwards compatibility for w3c
+        def element_property(element, name)
+          execute_script 'return arguments[0][arguments[1]]', element, name
+        end
+
         def element_value(element)
           execute :getElementValue, id: element
         end
--- a/rb/spec/integration/selenium/webdriver/element_spec.rb
+++ b/rb/spec/integration/selenium/webdriver/element_spec.rb
@@ -94,6 +94,13 @@ module Selenium
         expect(driver.find_element(id: 'withText').attribute('nonexistent')).to be_nil
       end
 
+      not_compliant_on browser: :edge do
+        it 'should get property value' do
+          driver.navigate.to url_for('formPage.html')
+          expect(driver.find_element(id: 'withText').property('nodeName')).to eq('TEXTAREA')
+        end
+      end
+
       it 'should clear' do
         driver.navigate.to url_for('formPage.html')
         driver.find_element(id: 'withText').clear
--- a/rb/lib/selenium/webdriver/common/element.rb
+++ b/rb/lib/selenium/webdriver/common/element.rb
@@ -119,6 +119,20 @@ module Selenium
       end
 
       #
+      # Get the value of a the given property with the same name of the element. If the value is not
+      # set, nil is returned.
+      #
+      # @param [String]
+      #   property name
+      # @return [String,nil]
+      #   property value
+      #
+
+      def property(name)
+        bridge.element_property self, name
+      end
+
+      #
       # Get the text content of this element
       #
       # @return [String]
--- a/rb/lib/selenium/webdriver/remote/w3c_bridge.rb
+++ b/rb/lib/selenium/webdriver/remote/w3c_bridge.rb
@@ -525,7 +525,7 @@ module Selenium
         end
 
         def element_property(element, name)
-          execute :getElementProperty, id: element.values.first, name: name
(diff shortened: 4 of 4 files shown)


rb/lib/selenium/webdriver/remote/bridge.rb
```diff
@@ -541,6 +541,11 @@ def element_attribute(element, name)
           execute :getElementAttribute, id: element.ref, name: name
         end
 
+        # Backwards compatibility for w3c
+        def element_property(element, name)
+          execute :getElementAttribute, id: element, name: name
```
AutomatedTester: this should try use `getElementProperty` and if the driver doesnt understand that then fallback to `getElementAttribute` (not sure if this is already handled elsewhere)

lmtierney: This is actually the fallback, the getElementProperty is already implemented [here](https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/remote/w3c_bridge.rb#L528) but there was no path to calling it. Should this execute js in order to get the property value like you implemented it in python instead?

titusfortner: Right, Ruby implements W3C calls in a separate class. I don't think we should add methods to the JWP implemented Bridge to match new W3C calls.

rb/spec/integration/selenium/webdriver/element_spec.rb
```diff
@@ -94,6 +94,11 @@ module WebDriver
         expect(driver.find_element(id: 'withText').attribute('nonexistent')).to be_nil
       end
 
+      it 'should get property value' do
```
titusfortner: This needs to add a guard. Looks like [Edge hasn't implemented it, yet](https://developer.microsoft.com/en-us/microsoft-edge/platform/documentation/webdriver-commands/), so:
```
compliant_on browser: :firefox do 
```

lmtierney: Aha, I couldn't find that page, thanks!

Also, shouldn't this be

```
not_compliant_on browser: :edge do
```

Since we want to test the backwards compatibility as well?

lmtierney: Actually that's probably a good workshop one

rb/lib/selenium/webdriver/remote/bridge.rb
titusfortner: @AutomatedTester - how is Python doing this? Are you going to support calling `getElementProperty` on a non-w3c driver?

lukeis: Calling this atom will get an attribute instead of the property also. The intention of doing the divide of getting attribute / property is to allow users calling these to "do what I say" rather than our past approach of "do what I mean".

getting element property i don't believe is implemented in an atom. Here's what python does:

https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/webelement.py#L107-L110

It falls back to essentially `return driver.execute_script('return arguments[0][arguments[1]]', element, property_name)`

titusfortner: @lmtierney - ok, yes, let's do that as the fall-back rather than calling `getElementAttribute`.

rb/spec/integration/selenium/webdriver/element_spec.rb
titusfortner: As for spec guards, you're right, there shouldn't be any need for this with the backward compatible feature. Even for Edge... Since Edge is only half supporting W3C so far, I created [this workaround](https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/edge/legacy_support.rb). We can actually make this workaround less complicated now that as of 3.0.1 Ruby sends JWP commands to Remote always.

rb/lib/selenium/webdriver/remote/bridge.rb
lmtierney: Sounds good

rb/spec/integration/selenium/webdriver/element_spec.rb
lmtierney: So for now, just add the fallback method to the legacy workaround as well?

lmtierney:
@titusfortner updated, had to update the element.property command to pass the element itself back to the bridge in order to execute js on it

rb/spec/integration/selenium/webdriver/element_spec.rb
titusfortner: Oh, crap, you're right. We want: `not_compliant_on browser: :edge do`

lmtierney:
Updated and squashed

titusfortner:
merged: 8b47d0a

PR #3101, opened 2016-11-14
Allow a Ruby Driver class to be instantiated with an existing session id

robbiewareham:

Before I get hit for raising this PR, I will try and explain our use case which while
possibly unique, I still believe is valid.

We have an implementation in which we have a bespoke orchestration layer which creates a Selenium
 docker container (including Browsermob) and passes the dynamic url as a connection string to the
test. NB. For various reasons, the orchestration layer is NodeJS while the test is Ruby.

This is an alternative to Selenium Grid which we feel is more stable and maintainable especially when supporting HAR file generation.

Our uses cases are;

1) The orchestration layer provides the session id to the test rather than the current approach of
the orchestration layer querying Selenium API after the test has created the session

2) The orchestration layer can choose Selenium bindings it feels are must suitable during a test run;

e.g. Pass session id to Ruby, do some processing that Ruby is good at, pass session to Javascript binding to perform other processing.

My personal feeling is that this PR is within the remit of the the W3C standard as it is not breaking or abusing any of the defined methods. I don't believe therre is anything say that all request must be made from the same client?

I am more than happy to discuss  this further at anytime zone in case of clarification/discussion

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/3101)

### DIFF
--- a/rb/lib/selenium/webdriver/remote/bridge.rb
+++ b/rb/lib/selenium/webdriver/remote/bridge.rb
@@ -67,6 +67,9 @@ module Selenium
           desired_capabilities = opts.delete(:desired_capabilities) { Capabilities.firefox }
           url = opts.delete(:url) { "http://#{Platform.localhost}:#{port}/wd/hub" }
 
+          # Added to allow an existing session id to be attached to
+          existing_session = opts.delete(:existing_session) {nil}
+
           unless opts.empty?
             raise ArgumentError, "unknown option#{'s' if opts.size != 1}: #{opts.inspect}"
           end
@@ -85,7 +88,12 @@ module Selenium
           http_client.server_url = uri
 
           @http = http_client
-          @capabilities = create_session(desired_capabilities)
+          if existing_session
+            # Attach to session id rather than creating new session
+            @capabilities  = attach_to_session(existing_session)
+          else
+            @capabilities  = create_session(desired_capabilities)
+          end
 
           @file_detector = nil
         end
@@ -126,6 +134,21 @@ module Selenium
           return Capabilities.json_create resp['value'] if @session_id
 
           raise Error::WebDriverError, 'no sessionId in returned payload'
+        end
+
+        # Method to attach to an existing session
+        def attach_to_session(session_id)
+          # We need to set this instance variable as it used in the raw_execute method
+          @session_id = session_id
+
+          # Make a request to get the details of the session (ie. /wd/hub/session/{session_id})
+          # NB. The raw_execute method expects session_id to be from an instance variable rather than passed in
+          resp = raw_execute :getCapabilities, {}
+          raise Error::WebDriverError, "Expected session #{session_id} in response but #{resp['sessionId']} " \
+                                        'was returned in payload' unless @session_id == resp['sessionId']
+
+          # The Capabilities object in the GET /session is exactly the same the object returned from POST /session
+          Capabilities.json_create resp['value']
         end
 
         def status
--- a/rb/spec/integration/selenium/webdriver/remote/driver_spec.rb
+++ b/rb/spec/integration/selenium/webdriver/remote/driver_spec.rb
@@ -32,6 +32,18 @@ module Selenium
             expect(driver).to be_kind_of(DriverExtensions::HasRemoteStatus)
             expect(driver.remote_status).to be_kind_of(Hash)
           end
+
+          it 'should allow instantiation with an existing selenium session' do
+            new_driver = Selenium::WebDriver.for(
+                :remote,
+                existing_session: driver.session_id)
+
+            expect(new_driver.session_id).to be(driver.session_id)
+            expect(new_driver.capabilities).to eq(driver.capabilities)
+            expect(driver).to be_kind_of(DriverExtensions::HasRemoteStatus)
+            expect(driver.remote_status).to be_kind_of(Hash)
+          end
+
         end
       end
     end # Remote


robbiewareham:
NB. I am at Selenium Conf if anyone wants to discuss

titusfortner:
@robbiewareham Thanks for this PR! As per our conversation at the conference, I'm hesitant to add it because in general I don't think it is a good idea to use session_id this way. That being said your desire to be able to switch between binding is interesting. If maintainers of other bindings think this is something that could reasonably be added/supported across the project, then we can discuss this implementation and merge it.

I'll push a conversation about this on Slack/irc channel tomorrow. (You're welcome to [join in on slack](https://seleniumhq.herokuapp.com/) or irc)

robbiewareham:
@titusfortner Thanks, I'll keep any eye out on Slack

Quantumplation:
Even beyond switching between various bindings, this would be helpful for things like long-running tests, or resuming tests after a reboot (for example, to gracefully resume a test through docker-container reboots).

jlipps:
We've also seen examples of this being useful for the Appium bindings, for example in order to use multiple driver bindings for the same session (which was necessary in one case to work around some protractor issues). I'm not sure of why it should be disallowed on principle.

titusfortner:
Aha, I found the conversation from the irc channel on 2/16/2016 with @lukeis and @jimevans regarding the similar PR #1346

https://gist.github.com/titusfortner/ea71ebcfccac0deaf600cd2f29b8d06f

rb/lib/selenium/webdriver/remote/bridge.rb
```diff
@@ -85,7 +88,12 @@ def initialize(opts = {})
           http_client.server_url = uri
 
           @http = http_client
-          @capabilities = create_session(desired_capabilities)
+          if existing_session
```
esparta: How about a little bit of simplification?
```patch
-       @capabilities = create_session(desired_capabilities)
+       @capabilites = 
+          if existing_session
+            # Attach to session id rather than creating new session
+            attach_to_session(existing_session)
+          else
+            create_session(desired_capabilities)
+          end
```

robbiewareham:
The argument against seems to be based on good test practice which I cannot argue against.

However, I thought that Selenium is not a test tool, rather an API to automate browsers;

[Selenium automates browsers. That's it! What you do with that power is entirely up to you. Primarily, it is for automating web applications for testing purposes, but is certainly not limited to just that.](http://docs.seleniumhq.org/)

To restrict functionality of Selenium based on upon the best practice of one use case seems to confirm that Selenium is simply there for automated testing? 

The fact that;

a) There are 2 almost matching PRs from different people in less than 12 months
b) Appium bindings support it
c) Non-official JS support it

..shows that there must be a need for this functionality from the community?

titusfortner:
I don't disagree with anything you are saying here. That being said, I don't want to add all of this to the Ruby bindings unless @shs96c, @AutomatedTester @jleyba & @jimevans all agree to (eventually) explicitly support this additional functionality in their respective bindings...

Evaluating alternatives it appears that multiple hierarchy issues would make subclassing what you need more challenging unless you are only using the Remote class. Plus I don't see an easy way to add event listener support that would allow you to avoid having to constantly close extra sessions.
I'd probably be ok with adding an `||=` [here](https://github.com/SeleniumHQ/selenium/blob/e0a54f8/rb/lib/selenium/webdriver/remote/bridge.rb#L69). That would make subclassing much easier allowing custom code in the `#initalize` method before calling `super`. Would that be sufficient for your needs?

titusfortner:
Once #3490 gets merged, then all of the Bridges will be initialized the same way, which means I can change `@capabilities =` to `@capabilities ||=` in one place to support this functionality.

This will allow you to do this:
https://gist.github.com/titusfortner/2d8f0b0d78727bfbab165bccbd04efea

That example class can be made a subclass of any existing Bridge and still work. I expect to close this out next week and make it available in Selenium 3.1.

p0deje:
@robbiewareham 

First of all, sorry for this hanging such a long time. I've been trying to approach merging this PR several times, but have constantly been distracted by other things.

Once I finally got to it, I started realizing that the way Selenium works became more complicated than it used to when you opened this PR. First of all, now it implements protocol handshake, which sends a mixed payload with desired capabilities containing both OSS and W3C variants to a create session endpoint and then, based on the response format, figures out which dialect (OSS or W3C) we should use to communicate to this remote. 

For example, this is how the payload and response look like for a browser that implements old OSS dialect:

```ruby

driver = Selenium::WebDriver.for(:remote, desired_capabilities: :chrome)
# 2018-03-02 08:39:49 INFO Selenium -> POST session
# 2018-03-02 08:39:49 INFO Selenium    >>> http://127.0.0.1:4444/wd/hub/session | {"desiredCapabilities":{"browserName":"chrome","version":"","platform":"ANY","javascriptEnabled":true,"cssSelectorsEnabled":true,"takesScreenshot":false,"nativeEvents":false,"rotatable":false},"capabilities":{"firstMatch":[{"browserName":"chrome"}]}}
# 2018-03-02 08:39:49 DEBUG Selenium      > {"Accept"=>"application/json", "Content-Type"=>"application/json; charset=utf-8", "Content-Length"=>"250"}
# 2018-03-02 08:39:50 INFO Selenium <- {"status":0,"sessionId":"4cd777f1aa09806e01843b6e05f78241","value":{"applicationCacheEnabled":false,"rotatable":false,"mobileEmulationEnabled":false,"networkConnectionEnabled":false,"chrome":{"chromedriverVersion":"2.35.528157 (4429ca2590d6988c0745c24c8858745aaaec01ef)","userDataDir":"/var/folders/3s/yybq9c5n4pd0z9t4phtp6ml00000gn/T/.org.chromium.Chromium.wGEphs"},"takesHeapSnapshot":true,"pageLoadStrategy":"normal","databaseEnabled":false,"handlesAlerts":true,"hasTouchScreen":false,"version":"64.0.3282.186","platform":"Mac OS X","browserConnectionEnabled":false,"nativeEvents":true,"acceptSslCerts":false,"acceptInsecureCerts":false,"locationContextEnabled":true,"webStorageEnabled":true,"browserName":"chrome","takesScreenshot":true,"javascriptEnabled":true,"cssSelectorsEnabled":true,"setWindowRect":true,"unexpectedAlertBehaviour":"","webdriver.remote.sessionid":"4cd777f1aa09806e01843b6e05f78241"}}
# 2018-03-02 08:39:50 INFO Selenium Detected OSS dialect.
```

And this is how the payload and response look like for a browser implementing new W3C dialect:

```ruby

driver = Selenium::WebDriver.for(:remote, desired_capabilities: :firefox)
# 2018-03-02 08:39:22 INFO Selenium -> POST session
# 2018-03-02 08:39:22 INFO Selenium    >>> http://127.0.0.1:4444/wd/hub/session | {"desiredCapabilities":{"browserName":"firefox","version":"","platform":"ANY","javascriptEnabled":true,"cssSelectorsEnabled":true,"takesScreenshot":true,"nativeEvents":false,"rotatable":false},"capabilities":{"firstMatch":[{"browserName":"firefox"}]}}
# 2018-03-02 08:39:22 DEBUG Selenium      > {"Accept"=>"application/json", "Content-Type"=>"application/json; charset=utf-8", "Content-Length"=>"251"}
# 2018-03-02 08:39:24 INFO Selenium <- {"value":{"sessionId":"45b24981-2e48-b940-b4ae-06c5da3fd916","capabilities":{"moz:profile":"/var/folders/3s/yybq9c5n4pd0z9t4phtp6ml00000gn/T/rust_mozprofile.ZaMN2cLgI9PD","rotatable":false,"timeouts":{"implicit":0,"pageLoad":300000,"script":30000},"pageLoadStrategy":"normal","moz:headless":false,"moz:accessibilityChecks":false,"acceptInsecureCerts":false,"browserVersion":"58.0.1","platformVersion":"17.4.0","moz:processID":18665,"browserName":"firefox","platformName":"darwin","moz:webdriverClick":true,"webdriver.remote.sessionid":"45b24981-2e48-b940-b4ae-06c5da3fd916"}}}
# 2018-03-02 08:39:24 INFO Selenium Detected W3C dialect.
```

Protocol handshake allows us to know how we should talk to a remote. Without knowing this, we won't be able to properly craft requests.

Now let's take a look at your proposal. If we already know session ID, we also either need to know which dialect we should use or we need to figure this out somehow. With OSS dialect we can use `getCapabilities` like you did, but W3C dialect lacks [such an endpoint](https://w3c.github.io/webdriver/webdriver-spec.html#list-of-endpoints). The same problem hold for the capabilities itself - we somehow need to know which of them should be applied for the remote in use.

The problems I described above is actually what makes it hard for me now to reason about implementing attachment to an existing session. The best idea I have is to change Ruby bindings public API in the way that allows constructing the bridge you want completely manually. But I'm not sure it gives what this PR initially provided.

I'll be glad if you could take a look at this comment and may work with me to address those problems somehow.

p0deje:
With Selenium 4.0.0.alpha1, the handshake is finally removed so now Selenium only supports W3C dialect, which doesn't have any command to retrieve capabilities from a remote node. With that said, your custom Grid-like solution needs to provide [an extension command](https://w3c.github.io/webdriver/#dfn-extension-command) for this to work and Ruby client needs to change its publicly available API to allow for such extensions. Another option would be to subclass `Driver`/`Bridge` so that you can implement custom behavior, but I guess Ruby bindings still need to allow such extension play nicely. I'll be happy to move forward and make all the necessary changes, but it's kind of hard to guess how it's used, so I need assistance.

Since there was no response for several months, I'm closing this PR, but, as I said before, I'll be happy to continue the work on it, so please comment in this PR.

PR #3125, opened 2016-11-17
Fixed java client's implementation of NetworkConnection.

mcharsley:

Changed java client's RemoteNetworkConnection to send just the integer
bitmask for the network connection, not the full object.
This will bring the java client in line with
 - the python client (see https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/mobile.py#L60 )
 - what the spec expects (see https://github.com/SeleniumHQ/mobile-spec/blob/master/spec-draft.md#device-modes )
 - and what chromedriver expects. prior to this change I saw

    [18.564][INFO]: COMMAND SetNetworkConnection {
       "parameters": {
          "type": {
             "airplaneMode": true,
             "class": "org.openqa.selenium.mobile.NetworkConnection$ConnectionType",
             "dataEnabled": false,
             "hCode": 1,
             "wifiEnabled": false
          }
       }
    }
being sent, which resulted in a RESPONSE SetNetworkConnection unknown
error: invalid connection_type error.

### DIFF
--- a/java/client/src/org/openqa/selenium/remote/mobile/RemoteNetworkConnection.java
+++ b/java/client/src/org/openqa/selenium/remote/mobile/RemoteNetworkConnection.java
@@ -42,7 +42,7 @@ public class RemoteNetworkConnection implements NetworkConnection {
   @Override
   public ConnectionType setNetworkConnection(
       ConnectionType type) {
-    Map<String, Integer> mode = ImmutableMap.of("type", type.getBitMask());
+    Map<String, ConnectionType> mode = ImmutableMap.of("type", type);
     return new ConnectionType(((Number) executeMethod.execute(DriverCommand.SET_NETWORK_CONNECTION,
                                                             ImmutableMap
                                                                 .of("parameters", mode)))
--- a/java/client/src/org/openqa/selenium/mobile/NetworkConnection.java
+++ b/java/client/src/org/openqa/selenium/mobile/NetworkConnection.java
@@ -94,10 +94,6 @@ public interface NetworkConnection {
       return (mask / 4) > 0;
     }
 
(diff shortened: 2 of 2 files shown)


java/client/src/org/openqa/selenium/remote/mobile/RemoteNetworkConnection.java
```diff
@@ -42,7 +42,7 @@ public ConnectionType getNetworkConnection() {
   @Override
   public ConnectionType setNetworkConnection(
       ConnectionType type) {
-    Map<String, ConnectionType> mode = ImmutableMap.of("type", type);
+    Map<String, Integer> mode = ImmutableMap.of("type", type.getBitMask());
```
lukeis: can you change this instead to just `type.toString()` and then you wouldn't have to expose the getBitMask in the enum. (Alternatively you could have used `type.hashCode()` but that doesn't feel as nice)

lukeis: looking at it again, we can just add the `toJson()` method to ConnectionType and then the JsonToBeanConverter will do it's thing and serialize it.

    public String toJson(){ return toString(); }

mcharsley: Reflection!? Oh well, when in Rome... Have removed getBitMask, and added toJson method. Thanks.

mcharsley:
Lukeis: I've pushed some new changes corresponding to your comments
Being completely new to github, I'm not sure if you automatically get informed or not. Hopefully if you haven't then this will let you know...

Please take another look? thanks

PR #3134, opened 2016-11-18
Change log level for protocol debug messages to 'fine'

Artur-:
Logging two rows with low level debug info for each test makes it hard to see any relevant output.

### DIFF
--- a/java/client/src/org/openqa/selenium/remote/ProtocolHandshake.java
+++ b/java/client/src/org/openqa/selenium/remote/ProtocolHandshake.java
@@ -83,7 +83,7 @@ public class ProtocolHandshake {
 
     if (result.isPresent()) {
       Result toReturn = result.get();
-      LOG.fine(String.format("Detected dialect: %s", toReturn.dialect));
+      LOG.info(String.format("Detected dialect: %s", toReturn.dialect));
       return toReturn;
     }
 


java/client/src/org/openqa/selenium/remote/ProtocolHandshake.java
```diff
@@ -77,13 +77,13 @@ public Result createSession(HttpClient client, Command command)
...
       parameters.append("}");
-      LOG.info("Falling back to straight W3C remote end connection");
+      LOG.fine("Falling back to straight W3C remote end connection");
       result = createSession(client, parameters);
     }
 
     if (result.isPresent()) {
       Result toReturn = result.get();
-      LOG.info(String.format("Detected dialect: %s", toReturn.dialect));
+      LOG.fine(String.format("Detected dialect: %s", toReturn.dialect));
```
lukeis: can we leave this one as info?

One has the ability to set java log levels... although it seemingly is too complicated for many users. But those that log issues this is pretty important information to see by default in the console when we potentially need to troubleshoot things.

Artur-: I agree that setting log levels is way too complicated, but I do not think that should prevent reducing log spam when running hundreds of tests. How about logging only once per test suite, i.e. track logging using a static boolean?

lukeis: That is much more complicated and I wouldn't merge that. Only standard Java logging should be used.

Artur-: Alright, changed it back to `info`

PR #3143, opened 2016-11-21
Issue 2479/snake case

joesho112358:

### DIFF
--- a/rb/lib/selenium/webdriver/edge/legacy_support.rb
+++ b/rb/lib/selenium/webdriver/edge/legacy_support.rb
@@ -27,8 +27,7 @@ module Selenium
            send_keys_to_active_element get_window_handles get_current_window_handle
            get_window_size set_window_size get_window_position set_window_position
            maximize_window get_alert_text accept_alert dismiss_alert].each do |cmd|
-          jwp = Remote::Bridge::COMMANDS[cmd]
-          Remote::W3CBridge.command(cmd, jwp.first, jwp.last)
+          Remote::W3CBridge::COMMANDS[cmd] = Remote::Bridge::COMMANDS[cmd]
         end
 
         def execute_script(script, *args)
--- a/rb/lib/selenium/webdriver/remote/w3c_bridge.rb
+++ b/rb/lib/selenium/webdriver/remote/w3c_bridge.rb
@@ -118,8 +118,7 @@ module Selenium
         end
 
         def status
-          jwp = Selenium::WebDriver::Remote::Bridge::COMMANDS[:status]
-          self.class.command(:status, jwp.first, jwp.last)
+          COMMANDS[:status] = Selenium::WebDriver::Remote::Bridge::COMMANDS[:status]
           execute :status
         end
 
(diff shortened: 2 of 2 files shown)


rb/lib/selenium/webdriver/remote/commands.rb
```diff
@@ -25,191 +25,194 @@ class Bridge
...
+            touch_scroll: [:post, 'session/:session_id/touch/scroll'.freeze],
+            touch_flick: [:post, 'session/:session_id/touch/flick'.freeze],
+
+            #
+            # logs
+            #
+
+            get_available_log_types: [:get, 'session/:session_id/log/types'.freeze],
+            get_log: [:post, 'session/:session_id/log'.freeze]
+        }
```
titusfortner: Without my running specs for this yet, can we freeze this constant now, or does Edge still make this difficult?

rb/lib/selenium/webdriver/edge/legacy_support.rb
```diff
@@ -27,8 +27,7 @@ module LegacySupport
            send_keys_to_active_element get_window_handles get_current_window_handle
```
titusfortner: Now that I've changed Remote calls to all use `Remote::Bridge` instead of `Remote::W3CBridge` we should get rid of this file entirely and move the methods directly into `Edge::Bridge`
Can you help with that as part of this PR, or should I do that? I'd like for this to be done together.

titusfortner:
Thanks for this PR!

I changed over everything in the `Bridge` classes to using snake_case, but left the commands in the constant to camelCase to match with other bindings... That being said, I don't think this is necessary, and it is more "Ruby" to have our symbols be snake_case.

I also appreciate our not modifying the constant.

Have you run the tests on Edge? It needs to get done before this gets merged in, but I can do that.

rb/lib/selenium/webdriver/edge/legacy_support.rb
joesho112358: That seems like it should be a separate piece to me, but I wouldn't argue against it. If you have a game plan, let me know what I can help with.

joesho112358:
@titusfortner  I do not have a machine that runs edge; so I would appreciate the help there. I guess I was assuming the travis build covered it....

rb/lib/selenium/webdriver/edge/legacy_support.rb
joesho112358: From what I can see, all the methods in `legacy_support.rb` are overridden in the `Edge::Bridge` already. So I think if everything is pointed at `Remote::Bridge`, then it should be ok to delete.

titusfortner:
Yeah, we don't have automatic coverage for windows browsers yet, but I'm past due for a run through on everything, so I'll do that this weekend to make sure everything works.

ShirleyPerkins:
Introducing Robot Framework and answering the question- Robot Framework Vs. Selenium WebDriver?
http://blog.testproject.io/2016/11/22/robot-framework-introduction/

titusfortner:
I went ahead and removed the extra Edge code and got all the tests passing, thanks for your contribution!

Merged: ef2003cd6a61364a1a353481d906b3bfdbe00f97

PR #3151, opened 2016-11-21
Replace HashBiMap with thread-safe and lighter-weight ConcurrentHashMap

asmundak:

HashBiMap is not thread safe. And its ability to map back from value
to key is used in a single place and can be easily replaced.


jleyba:
Travis failures are due to flaky tests.

PR #3158, opened 2016-11-23
rb: add ability to pass a custom driver path

lmtierney:

This also deprecates a couple module methods in efforts to reduce duplication

### DIFF
--- a/rb/lib/selenium/webdriver/common/service.rb
+++ b/rb/lib/selenium/webdriver/common/service.rb
@@ -37,10 +37,10 @@ module Selenium
       STOP_TIMEOUT        = 20
 
       @executable = nil
-      @missing = ''
+      @missing_text = nil
 
       class << self
-        attr_accessor :executable, :missing_text
+        attr_reader :executable, :missing_text
       end
 
       attr_accessor :host
--- a/rb/lib/selenium/webdriver/chrome.rb
+++ b/rb/lib/selenium/webdriver/chrome.rb
@@ -27,7 +27,7 @@ module Selenium
   module WebDriver
     module Chrome
       def self.driver_path=(path)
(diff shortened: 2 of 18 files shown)


rb/lib/selenium/webdriver/common/service.rb
```diff
@@ -36,17 +36,31 @@ class Service
       SOCKET_LOCK_TIMEOUT = 45
       STOP_TIMEOUT        = 20
 
+      @executable = nil
+      @missing = ''
```
esparta: Is this `@missing_text` ?

rb/lib/selenium/webdriver/common/service.rb
```diff
@@ -36,17 +36,31 @@ class Service
...
+
+      class << self
+        attr_accessor :executable, :missing_text
```
esparta: Do we mean to let anyone outside the subclasses to set up the `missing_text`?
If the answer is no, then I'd suggest to have two different ones:
```ruby
class << self
  attr_reader :missing_text
  protected
  attr_writer :missing_text
end
```

rb/lib/selenium/webdriver/common/service.rb
lmtierney: Yep, thanks for catching that!

rb/lib/selenium/webdriver/common/service.rb
lmtierney: It doesn't hurt anything, but it really only needs to be `attr_reader` since it is never set (and isn't intended to be) in the code. Updated, thanks again

titusfortner:
merged 5ce7249dae677cc6117cd9b21e5a85aedd71ed5e

PR #3174, opened 2016-11-28
[rb] adds ability to use firefox profiles with geckodriver

### LINKED ISSUE #2933: Profiles for Firefox
## Meta -

OS: OSX el capitan
Ruby: 2.3.1
Selenium Version: 3.0.0
Browser:  Firefox 49.0.1
## Code

```
profile = Selenium::WebDriver::Firefox::Profile.new
profile['browser.download.dir'] = '/Users/af/Downloads/'
profile['browser.download.folderList'] = 2
profile['browser.download.manager.showWhenStarting'] = false
profile['browser.helperApps.neverAsk.saveToDisk'] = 'text/csv'
browser = Selenium::WebDriver.for :firefox, :profile => profile
```

Actually I use Watir, but error is reproduced with pure selenium without [watir](https://github.com/watir/watir)
## Try to create profile for firefox but get an error:

/Users/af/.rvm/gems/ruby-2.3.1/gems/selenium-webdriver-3.0.0/lib/selenium/webdriver/remote/w3c_bridge.rb:80:in **'initialize': unknown option:** {:profile=>#<Selenium::WebDriver::Firefox::Profile:0x007fc76b93e1a0 @model=nil, @native_events=false, @secure_ssl=false, @untrusted_issuer=true, @load_no_focus_lib=false, @additional_prefs={"browser.download.dir"=>"/Users/af/Sites/", "browser.download.folderList"=>2, "browser.download.manager.showWhenStarting"=>false, "browser.helperApps.neverAsk.saveToDisk"=>"text/csv"}, @extensions={}>} (ArgumentError)

from /Users/af/.rvm/gems/ruby-2.3.1/gems/selenium-webdriver-3.0.0/lib/selenium/webdriver/firefox/w3c_bridge.rb:34:in 'initialize'

from /Users/af/.rvm/gems/ruby-2.3.1/gems/selenium-webdriver-3.0.0/lib/selenium/webdriver/common/driver.rb:49:in 'new'

from /Users/af/.rvm/gems/ruby-2.3.1/gems/selenium-webdriver-3.0.0/lib/selenium/webdriver/common/driver.rb:49:in 'for'

from /Users/af/.rvm/gems/ruby-2.3.1/gems/selenium-webdriver-3.0.0/lib/selenium/webdriver.rb:82:in 'for'

 **What can be wrong? Thank you for response.**

lmtierney:

This closes #2933

### DIFF
--- a/rb/spec/integration/selenium/webdriver/firefox/profile_spec.rb
+++ b/rb/spec/integration/selenium/webdriver/firefox/profile_spec.rb
@@ -24,12 +24,25 @@ module Selenium
     module Firefox
       describe Profile do
         let(:profile) { Profile.new }
+        let(:browser) { GlobalTestEnv.driver == :remote ? :remote : :firefox }
 
         def read_generated_prefs(from = nil)
           prof = from || profile
           dir = prof.layout_on_disk
 
           File.read(File.join(dir, 'user.js'))
+        end
+
+        def profile_opts
+          if GlobalTestEnv.driver == :remote
+            opt = { desired_capabilities: GlobalTestEnv.remote_capabilities.dup }
+            opt[:desired_capabilities][:firefox_profile] = profile
+            opt[:url] = GlobalTestEnv.remote_server.webdriver_url if GlobalTestEnv.remote_server?
+            opt
+          else
+            { desired_capabilities: Remote::Capabilities.firefox(marionette: GlobalTestEnv.browser == :firefox),
+              profile: profile }
+          end
         end
 
         it 'should set additional preferences' do
@@ -140,24 +153,31 @@ module Selenium
           expect(string).to include('user_pref("network.proxy.type", 4)')
         end
 
-        it 'should be able to use the same profile more than once' do
-          profile['browser.startup.homepage'] = url_for('formPage.html')
+        describe 'with browser' do
+          before(:each) do
+            profile['browser.startup.homepage'] = url_for('simpleTest.html')
+            profile['browser.startup.page'] = 1
+          end
 
-          begin
-            if GlobalTestEnv.driver == :remote
-              opt = { desired_capabilities: GlobalTestEnv.remote_capabilities.dup }
-              opt[:desired_capabilities][:firefox_profile] = profile
-              opt[:url] = GlobalTestEnv.remote_server.webdriver_url if GlobalTestEnv.remote_server?
-            else
-              opt = { desired_capabilities: Remote::Capabilities.firefox(marionette: browser == :firefox),
-                      profile: profile }
+          it 'should instantiate the browser with the correct profile' do
+            begin
+              driver1 = WebDriver.for(browser, profile_opts.dup)
+              expect{ wait(5).until { driver1.find_element(id: 'oneline') } }.to_not raise_error
+            ensure
+              driver1.quit
             end
+          end
 
-            driver_one = WebDriver.for(GlobalTestEnv.driver, opt.dup)
-            driver_two = WebDriver.for(GlobalTestEnv.driver, opt.dup)
-          ensure
-            driver_one.quit if driver_one
-            driver_two.quit if driver_two
+          it 'should be able to use the same profile more than once' do
+            begin
+              driver1 = WebDriver.for(browser, profile_opts.dup)
+              expect{ wait(5).until { driver1.find_element(id: 'oneline') } }.to_not raise_error
+              driver2 = WebDriver.for(browser, profile_opts.dup)
+              expect{ wait(5).until { driver2.find_element(id: 'oneline') } }.to_not raise_error
+            ensure
+              driver1.quit if driver1
+              driver2.quit if driver2
+            end
           end
         end
       end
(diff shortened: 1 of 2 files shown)


lmtierney:
I think I'm done making changes

rb/build.desc
```diff
@@ -117,7 +117,8 @@ ruby_library(name = "firefox",
     "lib/selenium/webdriver/firefox.rb"
   ],
   resources = [
-    { "//javascript/webdriver/atoms:getAttribute": "rb/lib/selenium/webdriver/atoms/getAttribute.js"}
+    { "//javascript/webdriver/atoms:getAttribute": "rb/lib/selenium/webdriver/atoms/getAttribute.js"},
+    { "//javascript/firefox-driver:webdriver_prefs" : "rb/lib/selenium/webdriver/firefox/extension/prefs.json" }
```
titusfortner: Are the other bindings using things from firefox-driver build to do this? Maybe we can check on how much the prefs.json changes and see if we can just commit a hard coded copy? I don't think we're planning on changing that code much post-geckodriver.

rb/lib/selenium/webdriver/firefox/profiles_ini.rb
```diff
@@ -57,11 +57,10 @@ def parse
               is_relative = Regexp.last_match(1).strip == '1'
             when /^Path=(.+)$/
               path = Regexp.last_match(1).strip
+              p = path_for(name, is_relative, path)
+              @profile_paths[name] = p if p
```
titusfortner: Moving this code inside the last case statement doesn't seem right. Why are we defining `name` and `is_relative` variables there if we aren't using them anywhere?

rb/lib/selenium/webdriver/remote/w3c_capabilities.rb
```diff
@@ -79,7 +79,7 @@ def firefox(opts = {})
             opts[:browser_version] = opts.delete(:version) if opts.key?(:version)
             opts[:platform_name] = opts.delete(:platform) if opts.key?(:platform)
 
-            new({browser_name: 'firefox'}.merge(opts))
+            new({browser_name: 'firefox', marionette: true}.merge(opts))
```
titusfortner: For Se 3.x we shouldn't need to specify `marionette: true`? I'll see if I can figure out where this is getting used that it might be necessary here.

rb/spec/integration/selenium/webdriver/firefox/profile_spec.rb
```diff
@@ -22,137 +22,161 @@
 module Selenium
   module WebDriver
     module Firefox
-      compliant_on driver: :ff_legacy do
```
titusfortner: I think we still need to specify that this is only compliant on `:firefox` & `:ff_legacy`. We can use `browser:` instead of `driver:` if we can use it local and remote.

rb/build.desc
lmtierney: I know python still use [this here](https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/firefox/firefox_profile.py#L62-L65), and it is mentioned in other build.desc files (didn't check to see where it was used).  The source file itself [is here](https://github.com/SeleniumHQ/selenium/blob/master/javascript/firefox-driver/webdriver.json). I'm not sure what all default preferences are desired, but there's a lot in there.

rb/lib/selenium/webdriver/firefox/profiles_ini.rb
lmtierney: I'm confused by the question... They are being used? I'm not saying this parser is the best way to do it, I just moved 2 lines so that it would work correctly. We can re-write it if you think that's better

rb/lib/selenium/webdriver/remote/w3c_capabilities.rb
lmtierney: Yeah I believe that's right. I originally updated both capabilities.rb & w3c_capabilities.rb files so that they had the correct option (since they are used by remote). But in reality only capabilities.rb needs `marionette: false`. However, we can't add that yet since remote geckodriver currently uses capabilities.rb. For now, users will have to use `Selenium::WebDriver::Remote::Capabilities.firefox(marionette: false)` even though they **shouldn't** need to pass that once we get w3c & remote married

rb/spec/integration/selenium/webdriver/firefox/profile_spec.rb
lmtierney: Are these specs specified to be run outside firefox tests? I was under the impression that it was only included in `firefox-test`, `ff-legacy-test`, `remote-firefox-test`, and `remote-ff-legacy-test`.

The tests should work for remote as well

lmtierney:
@sjieg the prefs.json file is built with the gem, if you run ./go //rb:gem:build you will get the gem built with the prefs.json included which you can then install locally

titusfortner:
merged with d4f142f93f743c94ee855266d99b798603ff3523 & 19de192131dc23f10f4eaa5a53bf012e949d269d

Released in selenium 3.0.4 gem

Thanks for your contribution

PR #3182, opened 2016-11-29
Session cleaner and Default session

GxtPerfectoM:

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/3182)

### DIFF
--- a/java/server/src/org/openqa/selenium/remote/server/DefaultSession.java
+++ b/java/server/src/org/openqa/selenium/remote/server/DefaultSession.java
@@ -116,7 +116,13 @@ public class DefaultSession implements Session {
                                       new LinkedBlockingQueue<Runnable>());
 
     // Ensure that the browser is created on the single thread.
-    EventFiringWebDriver initialDriver = execute(webDriverFutureTask);
+    EventFiringWebDriver initialDriver;
+    try {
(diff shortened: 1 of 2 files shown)


lukeis:
can you help demonstrate when these error scenarios occur? (like adding them to a test)

Or show some of the stacktraces that you've encountered to get a better understanding of what issue you're hitting

GxtPerfectoM:
These issues happen in our environment in the past and until now we used our own implementation of the code in-order to prevent these issues - So i don't have concrete stack traces or tests.

The DefaultSession fix was done because in our system when a session was created and the device was not found an exception would be thrown and no cleanup for the session was made.

The SessionCleaner happen on some of our clouds, probably because of the code for the session deletion and then the thread for the SessionCleaner ended and created a memory leak of sessions and threads and temporary files.
As i wrote in the commit, it is a good idea in any case to protect the thread execution if you need to make sure it runs forever

diemol:
Apologies for this late reply. Most of the code where this PR was directed does not exist anymore.
Grid 3 is not under development and is being replaced with Grid 4. If you'd like to continue with this feature, feel free to have a look at the new codebase and create a GitHub issue with the "Feature Request" template to continue the conversation.

PR #3183, opened 2016-11-29
Updating getSlotCounts to use sessions allowed

mitchloudenbeck:

Switching getSlotCounts to use the maxSessions when counting the total
slots. Old method would count all browser options so the free/total counts
could be much higher than what was actually available.

### DIFF
--- a/java/server/src/org/openqa/grid/web/servlet/HubStatusServlet.java
+++ b/java/server/src/org/openqa/grid/web/servlet/HubStatusServlet.java
@@ -134,22 +134,16 @@ public class HubStatusServlet extends RegistryBasedServlet {
   }
 
   private JsonObject getSlotCounts() {
-    int freeSlots = 0;
     int totalSlots = 0;
     int usedSlots = 0;
 
     for (RemoteProxy proxy : getRegistry().getAllProxies()) {
-      totalSlots += proxy.getTestSlots().size() > proxy.getMaxNumberOfConcurrentTestSessions() ?
-        proxy.getMaxNumberOfConcurrentTestSessions() : proxy.getTestSlots().size();
+      totalSlots += Math.min(proxy.getMaxNumberOfConcurrentTestSessions(), proxy.getTestSlots().size());
       usedSlots += proxy.getTotalUsed();
     }
-    freeSlots = totalSlots - usedSlots;
-
     JsonObject result = new JsonObject();
-
-    result.addProperty("free", freeSlots);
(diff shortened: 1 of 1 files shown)


java/server/src/org/openqa/grid/web/servlet/HubStatusServlet.java
```diff
@@ -136,16 +136,14 @@ private JsonObject getResponse(HttpServletRequest request) throws IOException {
...
 
     for (RemoteProxy proxy : getRegistry().getAllProxies()) {
-      for (TestSlot slot : proxy.getTestSlots()) {
-        if (slot.getSession() == null) {
-          freeSlots += 1;
-        }
-
-        totalSlots += 1;
-      }
+      totalSlots += proxy.getTestSlots().size() > proxy.getMaxNumberOfConcurrentTestSessions() ?
```
lukeis: totalSlots += Math.min(proxy.getMaxNumberOfConcurrentTestSessions(), proxy.getTestSlots().size());

mach6: don't need `+=` here. `=` will do

java/server/src/org/openqa/grid/web/servlet/HubStatusServlet.java
```diff
@@ -136,16 +136,14 @@ private JsonObject getResponse(HttpServletRequest request) throws IOException {
...
+        proxy.getMaxNumberOfConcurrentTestSessions() : proxy.getTestSlots().size();
+      usedSlots += proxy.getTotalUsed();
```
mach6: same comment about `+=`

mach6:
LGTM, once comments addressed. :)

java/server/src/org/openqa/grid/web/servlet/HubStatusServlet.java
mitchloudenbeck: Don't we still want += here? It is summing all the slots/sessions of all the nodes connected to the hub.

lukeis: reducing this further... could just get rid of the two other variables and just have:

    freeSlots += Math.min(proxy.getMaxNumberOfConcurrentTestSessions(), proxy.getTestSlots().size()) - proxy.getTotalUsed();

and yeah, i agree still need the `+=` :)

PR #3197, opened 2016-12-02
Fix which makes possible to use /Application/Firefox.app as binaryNam…

vinogradoff:

…e for Mac OS

/Application/Firefox50.app is a valid MacOS File name, but not Firefox executable. So it was possible to use /Application/Firefox50 but not /Application/Firefox50.app, which is not logical. Now both versions are possible.

### DIFF
--- a/java/client/src/org/openqa/selenium/firefox/internal/Executable.java
+++ b/java/client/src/org/openqa/selenium/firefox/internal/Executable.java
@@ -129,8 +129,14 @@ public class Executable {
         binaryName += ".exe";
 
     } else if (current.is(MAC)) {
-      if (!binaryName.endsWith(".app"))
+     
+      if (!binaryName.endsWith(".app")){
+        File macBinary = new File(binaryName);
+        if (macBinary.exists()){
+          return macBinary;
+        }
         binaryName += ".app";
+      }
       binaryName += "/Contents/MacOS/firefox-bin";
     }
 


lmtierney:
This is failing a lot of the builds, did it build clean when you did it locally? I don't really see how it would have

davehunt:
The binary is actually `Firefox.app/Contents/MacOS/firefox-bin` and this is how it should be specified.

lmtierney:
@davehunt if you look at the java code right below the removal, if it's specified as Firefox.app or Firefox (where it appends .app) it appends `/Contents/MacOS/firefox-bin` to the path

davehunt:
Ah, I see. In that case this looks good to me. 😛

davehunt:
Actually, this will break if the full binary path is specified as it will append `/Contents/MacOS/firefox-bin` to it.

vinogradoff:
@lmtierney exactly. And it failed to do it for Firefox.app because it returned to early (removed code).

I think it is easier for non-techies to configure Firefox.app - it is self-explaining.

vinogradoff:
@davehunt oops, yes, you are right.

java/client/src/org/openqa/selenium/firefox/internal/Executable.java
```diff
@@ -123,22 +123,24 @@ private static File locateFirefoxBinaryFromSystemProperty() {
     if (binaryName == null)
       return null;
 
-    File binary = new File(binaryName);
-    if (binary.exists())
```
lukeis: this check should remain, but could be expanded to ensure 'binary' is also a File, not a directory

lukeis: if you did that, you wouldn't need to make the adjustments below to the mac things.

vinogradoff: Thanks, will give it a try

vinogradoff:
See #3199 instead.

PR #2269, opened 2016-06-12
.NET: Add support for .NET Core

qmfrederik:
- [X ] By placing an `X` in the preceding checkbox, I verify that I have signed the [Contributor License Agreement](https://github.com/SeleniumHQ/selenium/blob/master/CONTRIBUTING.md#step-6-sign-the-cla)

This PR makes the Selenium WebDriver project compile on .NET Core. The changes basically boil down to fixes for:
- The new reflection API
- The absence of Serialization
- Standardization on `.Dispose` instead of `.Close`
- Changes in how OS-related information is accessed
- Changes in the `HttpWebRequest` and `HttpWebResponse` classes

Happy to work with you on any feedback you may provide,

PR #3200, opened 2016-12-03
Core compat

Brar:

I've extended pull request #2269 in order to make WebDriver.Support work for my current .Net core projects.
This pull request is somewhat more invasive than #2269 as I had to get rid of System.Runtime.Remoting which is not (and probably will never be) supported by .NET core. I opted for Castle.DynamicProxy as replacement.
I also opted to target .NET Standard 1.3 instead of 1.5 as this was easily possible.
Besides that it is a pretty raw port and I didn'd bother to optimise existing code.
I guess there's still a lot of work left for someone who really knows what she's doing but it probably is a good start.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,4 @@ Gemfile.lock
 .idea/inspectionProfiles/Project_Default.xml
 .idea/misc.xml
 third_party/py/googlestorage/client_secrets.json
+TestResult.xml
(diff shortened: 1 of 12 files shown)


Jetski5822:
Do you have it as a nuget package somewhere?

Brar:
@Jetski5822 
Nope.
I've been doing this for an internal project where "works for me" is enough.
I've filed this pull request in order to prevent people from completely starting from the scratch when trying to make selenium work with .Net Core.
As there was little interest in this until now, I didn't plan to put much effort into maintaining/evolving it unless explicitly requested.
I could certainly try to do a nuget package if you really want me to.

Jetski5822:
That's cool, no worries man. I am having issues with some of the drivers when upgrading Coypu.... so maybe I'll clone this branch and use

PR #3217, opened 2016-12-07
Prepare for Closure TagName update.

joshbruning:
TagName items will change type requiring toString()
to be called in the future. This change makes domcore.js
and attribute.js forward-compatible with the new version.


joshbruning:
Seems like the old dom.js code may have covered this case. I think I can just update domcore.js to resolve all of these.

lukeis:
@joshbruning what about #2597 ?

PR #3219, opened 2016-12-08
Add support for net-http-persistent v3

johnsyweb:

Context
-------

When we did a routine `bundle update`, net-http-persistent jumped from
v2.9.4 to v3.0.0. This major version change comes with a new
initializer with keyword arguments:
https://github.com/drbrain/net-http-persistent/commit/5d4b76c22dd38d29b6fbc1ed700a2a1c78c5abc4#diff-9da42550715aa09be7174a0babadcad3R512

This results in errors like so:

```
     ArgumentError:
       wrong number of arguments (2 for 0)
     # ./vendor/bundle/ruby/2.2.0/gems/net-http-persistent-3.0.0/lib/net/http/persistent.rb:505:in `initialize'
     # ./vendor/bundle/ruby/2.2.0/gems/selenium-webdriver-3.0.3/lib/selenium/webdriver/remote/http/persistent.rb:45:in `new'
     # ./vendor/bundle/ruby/2.2.0/gems/selenium-webdriver-3.0.3/lib/selenium/webdriver/remote/http/persistent.rb:45:in `new_http_client'
     # ./spec/support/selenium_custom_http_persisted_client.rb:37:in `new_http_client'
     # ./vendor/bundle/ruby/2.2.0/gems/selenium-webdriver-3.0.3/lib/selenium/webdriver/remote/http/default.rb:35:in `http'
     # ./vendor/bundle/ruby/2.2.0/gems/selenium-webdriver-3.0.3/lib/selenium/webdriver/remote/http/persistent.rb:49:in `response_for'
     # ./vendor/bundle/ruby/2.2.0/gems/selenium-webdriver-3.0.3/lib/selenium/webdriver/remote/http/default.rb:57:in `request'
     # ./spec/support/selenium_custom_http_persisted_client.rb:30:in `block in request'
```

Change
------

Detect the library version and choose the required initializer.

### DIFF
--- a/rb/lib/selenium/webdriver/remote/http/persistent.rb
+++ b/rb/lib/selenium/webdriver/remote/http/persistent.rb
@@ -42,7 +42,12 @@ module Selenium
               proxy = URI.parse(url)
             end
 
-            Net::HTTP::Persistent.new 'webdriver', proxy
+            if Net::HTTP::Persistent::VERSION >= '3'
+              Net::HTTP::Persistent.new name: 'webdriver', proxy: proxy
+            else
+              warn 'Support for this version of net-http-persistent is deprecated. Please upgrade.'
+              Net::HTTP::Persistent.new 'webdriver', proxy
+            end
           end
 
           def response_for(request)


johnsyweb:
The [CI failures](https://travis-ci.org/SeleniumHQ/selenium/jobs/182228330#L945) seem to be a subset of [those](https://travis-ci.org/SeleniumHQ/selenium/jobs/182074732#L939) on the [current `master`](https://github.com/SeleniumHQ/selenium/commit/b990d61b4eb9606db2b2084bc3be81a2e68d9d40).

titusfortner:
@johnsyweb - Thanks for this. Rather than basing it on specific versions, I was thinking we should use [`#arity`](https://ruby-doc.org/core-2.3.1/Method.html#method-i-arity) to address these situations.

johnsyweb:
`#arity` was my first thought, @titusfortner; I should have put that in the description. Unfortunately, since all of the arguments are optional in both versions of the initializer, we can't use this:

```ruby
Net::HTTP::Persistent.method(:new).arity #=> -1
```

I went with this approach with the knowledge that versions pre-3.0.0 have not used keyword arguments (and should any bug fixes be backported then I would expect @drbrain to keep the existing syntax for compatibility with old Rubies), along with the expectation that all versions post-3.0.0 will use keyword arguments.

I'm happy to consider other approaches :-)

titusfortner:
Well shoot, I was hoping for a more generic solution. Thanks for checking it, though.

The other concern I have is maintaining version conditionals in code in general. I almost feel like it would be better to restrict it to 2.x in the gemfile (but that might not be a good long term decision).

Perhaps it would be enough to put a deprecation notice [similar to how I did it in Watir](https://github.com/watir/watir/blob/master/lib/watir/wait.rb#L38-L42) and give people time to update their code then remove the conditional a few versions down the road.

johnsyweb:
Gemfile restrictions may break applications that depend on net-http-persistent.

A depracation warning seems like a good approach. I'll do that.

titusfortner:
merged in 0d3853ebff56618892c72d67a6450bde6c11b9e8
released in selenium 3.0.4 gem

Thanks for your contribution!

johnsyweb:
Thanks.

Issue #3216, opened 2016-12-07
_cookie_temp_file being closed and removed twice


francisbergin:
## Meta -
OS: Windows 7 (Cygwin)
Selenium Version: 3.0.2
Browser: PhantomJS
Browser Version: 2.1.1
Language: Python
Language Version: 3.4.5

## Expected Behavior -
When quitting the driver with `WebDriver.quit()`, `Service.send_remote_shutdown_command()` which closes and removes `_cookie_temp_file` should be called only once.

## Actual Behavior -
Whenever the Python GC runs and executes `Service.__del__()`, `Service.send_remote_shutdown_command()` is called again and attempts to close and delete `_cookie_temp_file` a second time causing a `FileNotFoundError` and an `OSError: [Errno 9] Bad file descriptor`.
## Steps to reproduce -
```python
import gc
from selenium import webdriver
driver=webdriver.PhantomJS()
driver.get('https://google.com')
driver.quit()
driver=1
gc.collect()
```

## More info -
The problem seems to occur when the PhantomJS process is not terminated and the garbage collector runs. There is often a delay between sending the `quit` command to PhantomJS and the process actually ending and if the garbage collector happens to run during that time and `self.process` is [not `none`](https://github.com/SeleniumHQ/selenium/blob/cdebe0e55f8dfaa2294f18f73aeb513fa76e0aef/py/selenium/webdriver/common/service.py#L145), then `send_remote_shutdown_command()` is run a second time. Also, this issue will most likely happen to users experiencing #767.

cgoldberg:
#3263 should fix this

fbergin:
This seems to be fixed. Tested in version `3.3.1`.

jmbowman:
I'm still seeing this same problem with 3.3.1 on Python 2.7, Mac OS X 10.11.6, as well as Python 2.7 and 3.2 on Travis (Ubuntu 12.04.5):

```
Traceback (most recent call last):
  File "/Users/jeremybowman/Source/needle/.tox/py27/lib/python2.7/site-packages/nose/suite.py", line 228, in run
    self.tearDown()
  File "/Users/jeremybowman/Source/needle/.tox/py27/lib/python2.7/site-packages/nose/suite.py", line 351, in tearDown
    self.teardownContext(ancestor)
  File "/Users/jeremybowman/Source/needle/.tox/py27/lib/python2.7/site-packages/nose/suite.py", line 367, in teardownContext
    try_run(context, names)
[...]

PR #3263, opened 2016-12-18
[py] fix shutdown and process termination

cgoldberg:
This PR addresses 2 issues with how webdriver stops services from the python bindings.

1. Changes the order of events to stop a process.  Previously. the order was `terminate()`->`kill()`->`wait()`.  So... it would send a SIGTERM signal to the process, immediately followed by a SIGKILL signal.  Since `wait()` is called after the process has already been forcefully stopped, there is nothing to wait on, so it does nothing.  The proper sequence should be:  `terminate()`->`wait()`->`kill()`.  That will send a SIGINT, wait for graceful process termination, then send a SIGKILL to forcefully stop it as a last resort.

2. ~~Removes the `__del__()` method from the `Service()` class.  The existence of this method is causing #3216.  Using `__del__()` to call `stop()` is unreliable and not a good idea in Python.
The proper way to stop the service is to explicitly call the `stop()` method~~

2. Add a try/except inside `__del__()` so inconsequential exceptions arent displayed

----

### DIFF
--- a/py/selenium/webdriver/common/service.py
+++ b/py/selenium/webdriver/common/service.py
@@ -125,12 +125,13 @@ class Service(object):
             url_request.urlopen("%s/shutdown" % self.service_url)
         except URLError:
             return
-        count = 0
-        while self.is_connectable():
-            if count == 30:
+
+        for _ in 30:
+            if self.is_connectable():
                 break
-            count += 1
-            time.sleep(1)
+            else:
+                time.sleep(1)
+
 
     def stop(self):
         """
@@ -164,5 +165,13 @@ class Service(object):
                 self.process.kill()
                 self.process = None
         except OSError:
-            # kill may not be available under windows environment
+            pass
+
+    def __del__(self):
+        # `subprocess.Popen` doesn't send signal on `__del__`;
+        # so we attemt to close the launched process when `__del__`
+        # is triggered.
+        try:
+            self.stop()
+        except Exception:
             pass


py/selenium/webdriver/common/service.py
```diff
@@ -160,14 +160,9 @@ def stop(self):
...
+                self.process.kill()
                 self.process = None
         except OSError:
             # kill may not be available under windows environment
             pass
-
-    def __del__(self):
-        # subprocess.Popen doesn't send signal on __del__;
-        # we have to try to stop the launched process.
-        self.stop()
```
lukeis: can we add a try / except around this to make it more stable?

I like the idea of any mistakenly un-quit drivers closing down after the process is quit (the java server does this... also i forget to do driver.quit often when i use the command line repl :) )

cgoldberg: yea, i think that just swallowing exceptions from a try/except is ok here.  

I've updated my branch and re-added the __del__  

p.s. I'll get that dastardly `__del__` out of there eventually.... fear not! :)

p.p.s. sorry I also piggybacked the tiny style change to the connection test in `send_remote_shutdown_command`...  since it was so egresiusly un-pythonic. :)

cgoldberg:
oops.. this should be:
`for x in range(30)` !!!!!

cgoldberg:
@lukeis can you fix?

lukeis:
and this is why i tried to stay away from python :-D

lukeis:
3064a0e5b467d2da4414f402729913704ea46642

cgoldberg:
arg.. im all over the place :(

it should be: `for _ in range(3)`  I forgot range in the original

lukeis:
fa53697e71d4f7dc3173cb2ff2705fcbee457484

PR #3264, opened 2016-12-18
Split Open/Read Timeouts, defaulting Http::Default open_timeout to nil

richseviora:
Ruby 2.3 changed `Net::HTTP#open_timeout` default from `nil` to `60`. As a result, debuggers that freeze the process (like RubyMine) timeout when evaluating browser operations. [Details here](https://youtrack.jetbrains.com/issue/RUBY-18301).

The following changes were made:
- Split `Remote::HTTP::Common#timeout` accessor to `open_timeout` and `read_timeout`.
- Added backwards compatibility for `#timeout=`.
- Defaulted `Remote::HTTP::Default#http` open_timeout to nil (maintains behaviour w/ Ruby 2.2 and earlier).
- Added specs for above.

### DIFF
--- a/rb/lib/selenium/webdriver/remote/http/default.rb
+++ b/rb/lib/selenium/webdriver/remote/http/default.rb
@@ -45,6 +45,7 @@ module Selenium
           # @param [Numeric] value - Timeout in seconds to apply to both open timeout and read timeouts.
           # @deprecated Please set the specific desired timeout {#read_timeout} or {#open_timeout} directly.
           def timeout=(value)
+            Kernel.warn 'Selenium::WebDriver::Remote::Http::Default#timeout= is deprecated. Use #read_timeout= or #open_timeout= instead'
             self.open_timeout = value
             self.read_timeout = value
           end
@@ -60,7 +61,7 @@ module Selenium
             end
 
             # Defaulting open_timeout to nil to be consistent with Ruby 2.2 and earlier.
-            http.open_timeout = self.open_timeout || nil
+            http.open_timeout = self.open_timeout
             http.read_timeout = self.read_timeout if self.read_timeout
 
             http
--- a/rb/spec/unit/selenium/webdriver/remote/http/default_spec.rb
+++ b/rb/spec/unit/selenium/webdriver/remote/http/default_spec.rb
@@ -25,7 +25,7 @@ module Selenium
       module Http
         describe Default do
           let(:client) do
-            client = Default.new
+            client            = Default.new
             client.server_url = URI.parse('http://example.com')
 
             client
@@ -39,23 +39,25 @@ module Selenium
           end
 
           describe '#initialize' do
-            let(:client) {Default.new(read_timeout: 22, open_timeout: 23)}
+            let(:client) { Default.new(read_timeout: 22, open_timeout: 23) }
+
             it 'accepts read timeout options' do
               expect(client.open_timeout).to eq 23
             end
+
             it 'accepts open timeout options' do
               expect(client.read_timeout).to eq 22
             end
           end
 
           describe '#timeout=' do
-            let(:value_set) {22}
-            before {
-              client.timeout = value_set
-            }
+            let(:value_set) { 22 }
+            before { client.timeout = value_set }
+
             it 'assigns value to #read_timeout' do
               expect(client.read_timeout).to eq value_set
             end
+
             it 'assigns value to #open_timeout' do
               expect(client.open_timeout).to eq value_set
             end
@@ -63,7 +65,7 @@ module Selenium
 
           it 'uses the specified timeout' do
             client.timeout = 10
-            http = client.send :http
+            http           = client.send :http
 
             expect(http.open_timeout).to eq(10)
             expect(http.read_timeout).to eq(10)
@@ -71,7 +73,7 @@ module Selenium
 
           it 'uses the specified proxy' do
             client.proxy = Proxy.new(http: 'http://foo:bar@proxy.org:8080')
-            http = client.send :http
+            http         = client.send :http
 
             expect(http).to be_proxy
             expect(http.proxy_address).to eq('proxy.org')


titusfortner:
Digging into this code some more...

1. `@open_timeout` and `@read_timeout` only matter for Default class. `Curb` subclass only uses `@timeout` and `Persistent` subclass can already [set them directly](https://github.com/drbrain/net-http-persistent/blob/master/lib/net/http/persistent.rb#L372) on the returned client instance. I'm not sure if anyone does it, but theoretically anyone could subclass Common and use their own client, so this client implementation specific detail shouldn't go there... 

2. It gets a little weird because our code currently allows a timeout value to be set when using `Persistent` class, but it won't actually do anything...

3. This fixes the RubyMine problem for users who do not set a client timeout. I suspect that the actual desired behavior for users who increase or decrease the default timeout value in `Default` class is to adjust the read timeout not the open timeout. I'm wondering if when `@timeout` is set, if we should either throw a warning about the change and/or only change the read and force open to be nil unless explicitly set.

4. Is there any reason not to allow timeouts to be set during the initialize?
Instead of:
```
  client = Selenium::WebDriver::Remote::Http::Default.new
  client.timeout = 120 
  driver = Selenium::WebDriver.for(:remote, http_client: client)
```

what if we allowed:
```
  client = Selenium::WebDriver::Remote::Http::Default.new(open_timeout: 5, read_timeout: 120)
  driver = Selenium::WebDriver.for(:remote, http_client: client)
```

richseviora:
My thoughts:

1. Agreed. I'll push the split down to `::Default`.
2. Also agreed! Should we override `#timeout=` on `::Persistent` to include a deprecation warning and remove the accessor altogether next major version? Right now only the `Curb` and `Default` subclasses only use it, and I don't think it makes sense for that value to be accessible on the common interface if it's specific to certain implementations.
3. I'm vaguely leaning towards keeping existing behaviour (`timeout=` sets read and open), plus warning and deprecating `timeout=`. I imagine the difference between `open_timeout` and `read_timeout` is such that there's no conceptual relationship between the two values so moving away from it might be a good idea. That said, I don't have any experience with this project and so don't know what typical usecase scenarios are like.
4. I can't think of any, and will revise accordingly.

richseviora:
BTW, I'm noticing that the Chrome integration tests are failing. I don't have J1.8 on this machine but I can't imagine my changes are at fault given that `Net::HTTP#open_timeout=nil` was existing behaviour previous to 2.3.

rb/lib/selenium/webdriver/remote/http/default.rb
```diff
@@ -28,6 +28,25 @@ module Http
...
           attr_accessor :proxy
 
+          attr_accessor :open_timeout
+          attr_accessor :read_timeout
+
+          # Initializes object.
+          # Warning: Setting +open_timeout+ to non-nil values will cause a separate thread to spawn.
+          # Debuggers that freeze the process will not be able to evaluate any operations if that happens.
+          # @param [Hash] options - Options hash accepting :open_timeout and :read_timeout values.
+          def initialize(options = {})
```
richseviora: Good lord, I don't know how I missed reading about keyword args until now, but absolutely!

richseviora:
Thanks! I'm glad you approve. This is my first change to a major OS project, so this is all rather new to me!

Re: accessor vs instance variable setting, this is a habit I picked up from programming in C#. Their fields operate similarly to ruby instance variables, and there's a very strong best practice not to interact with fields directly for the reasons you've mentioned.

I honestly don't have an opinion on whether to use `self.cat = dog` vs `@cat = dog` in constructors. 
The argument for the former is that it preserves consistent handling of property settings regardless of whether it's set during initialization, or further down the road. However, the argument for the latter is that if the settor has outside effects, then those outside effects may cause problems during initialization when the object is not in a complete state. I've personally had that happen to me.

So maybe the answer is... depends? :smile:

I'll push the amended changes shortly, and wait to hear back from yourself and @p0deje on the final changes.

rb/lib/selenium/webdriver/remote/http/default.rb
```diff
@@ -28,6 +28,27 @@ module Http
...
+          # @param [Numeric] open_timeout - Open timeout to apply to HTTP client.
+          # @param [Numeric] read_timeout - Read timeout (seconds) to apply to HTTP client.
+          def initialize(open_timeout: nil, read_timeout: nil)
+            @open_timeout = open_timeout
+            @read_timeout = read_timeout
+          end
+
+          # Maintaining backward compatibility.
+          # @param [Numeric] value - Timeout in seconds to apply to both open timeout and read timeouts.
+          # @deprecated Please set the specific desired timeout {#read_timeout} or {#open_timeout} directly.
```
p0deje: We should probably not only add `@deprecated` tag but also notify people using this method that it's deprecated, for example using [`Gem::Deprecated`](http://ruby-doc.org/stdlib-2.3.0/libdoc/rubygems/rdoc/Gem/Deprecate.html) or simple [`Kernel#warn`](https://ruby-doc.org/core-2.3.3/Kernel.html#method-i-warn).

rb/lib/selenium/webdriver/remote/http/default.rb
```diff
@@ -38,10 +59,9 @@ def http
               http.verify_mode = OpenSSL::SSL::VERIFY_NONE
             end
 
-            if @timeout
-              http.open_timeout = @timeout
-              http.read_timeout = @timeout
-            end
+            # Defaulting open_timeout to nil to be consistent with Ruby 2.2 and earlier.
+            http.open_timeout = self.open_timeout || nil
```
p0deje: Won't this be `nil || nil` by default? Why not just do this:

```ruby
http.open_timeout = self.open_timeout
```

rb/spec/unit/selenium/webdriver/remote/http/default_spec.rb
```diff
@@ -31,6 +31,36 @@ module Http
...
 
+          it 'assigns default timeout to nil' do
+            http = client.send :http
+
+            expect(http.open_timeout).to eq nil
+            expect(http.read_timeout).to eq 60
+          end
+
+          describe '#initialize' do
+            let(:client) {Default.new(read_timeout: 22, open_timeout: 23)}
```
p0deje: Please, add space after `{` and before `}`.

p0deje: And empty line after.

rb/spec/unit/selenium/webdriver/remote/http/default_spec.rb
```diff
@@ -31,6 +31,36 @@ module Http
...
+            it 'accepts read timeout options' do
+              expect(client.open_timeout).to eq 23
+            end
```
p0deje: Please add empty line after.

rb/spec/unit/selenium/webdriver/remote/http/default_spec.rb
```diff
@@ -31,6 +31,36 @@ module Http
...
+            it 'accepts open timeout options' do
+              expect(client.read_timeout).to eq 22
+            end
+          end
+
+          describe '#timeout=' do
+            let(:value_set) {22}
+            before {
```
p0deje: Guess this can be one-liner:

```ruby
before { client.timeout = value_set }
```

or if you want to have it multiline, please use `do ... end`:

```ruby
before do
  client.timeout = value_set
end
```

rb/spec/unit/selenium/webdriver/remote/http/default_spec.rb
```diff
@@ -31,6 +31,36 @@ module Http
...
+            it 'accepts read timeout options' do
+              expect(client.open_timeout).to eq 23
+            end
+            it 'accepts open timeout options' do
+              expect(client.read_timeout).to eq 22
+            end
+          end
+
+          describe '#timeout=' do
+            let(:value_set) {22}
```
p0deje: Please, add space after `{` and before `}`.

rb/spec/unit/selenium/webdriver/remote/http/default_spec.rb
```diff
@@ -31,6 +31,36 @@ module Http
...
+            before {
+              client.timeout = value_set
+            }
```
p0deje: Please add empty line after.

rb/spec/unit/selenium/webdriver/remote/http/default_spec.rb
```diff
@@ -31,6 +31,36 @@ module Http
...
+            it 'assigns value to #read_timeout' do
+              expect(client.read_timeout).to eq value_set
+            end
```
p0deje: Please add empty line after

rb/lib/selenium/webdriver/remote/http/default.rb
richseviora: I don't know when the accessor would be removed, so I've opted to use `Kernel#warn` instead.

Along the same logic, I think we should also deprecate `Common#timeout` and `Common#timeout=` and push that logic down to `Curb` since it's the only subclass that's using it. Otherwise we'll have an abstract class that has an otherwise valid property that's been deprecated (eventually removed) on one of its concrete implementations.

richseviora:
@p0deje updated code style per your feedback as well.

titusfortner:
Merged: 1e60e34c54e48c27b55d0040d340f1ed4dfabf20
Released in selenium 3.0.4 gem

Thanks for your contribution

inouetakuya:
@p0deje @titusfortner 

Could you reflect the changes in this PR in the wiki?

https://github.com/SeleniumHQ/selenium/wiki/Ruby-Bindings#internal-timeouts

> Internal timeouts
>
> Internally, WebDriver uses HTTP to communicate with a lot of the drivers (the JsonWireProtocol). By default, Net::HTTP from Ruby's standard library is used, which has a default timeout of 60 seconds. If you call e.g. Driver#get, Driver#click on a page that takes more than 60 seconds to load, you'll see a Timeout::Error raised from Net::HTTP. You can configure this timeout (before launching a browser) by doing:
>
>   client = Selenium::WebDriver::Remote::Http::Default.new
>   client.timeout = 120 # seconds
>   driver = Selenium::WebDriver.for :remote, http_client: Client

p0deje:
@inouetakuya Thank you for pointing out - I've changed to `read_timeout`.

PR #3285, opened 2016-12-23
Set browser.process.leader to true

csolg:
By default, the child process does not create a new process group. This means there's no guarantee that the entire process tree will die when the child process is killed. To solve this:
process = ChildProcess.build(*args)
process.leader = true
process.start

### DIFF
--- a/rb/lib/selenium/webdriver/chrome/service.rb
+++ b/rb/lib/selenium/webdriver/chrome/service.rb
@@ -40,6 +40,7 @@ module Selenium
           @process       = ChildProcess.build(*server_command)
 
           @process.io.inherit! if $DEBUG
+          @process.leader = true
           @process.start
         end
 


titusfortner:
@csolg - You need to sign the CLA before we can accept this PR

csolg:
@titusfortner I've signed the License.

p0deje:
I guess it makes sense, but for the sake of clarity, can you give a concrete example when child process is killed but its grandchildren are not?

Also, maybe we should add same thing to all services.

p0deje:
Same as #1631.

csolg:
The problem was on my server. In just a few days accumulated Chrome processes. Now it fixed.

p0deje:
> The problem was on my server. In just a few days accumulated Chrome processes.

You mean there were multiple Chrome/ChromeDriver zombie processes?

> Now it fixed.

With this PR? Or why is it fixed?

csolg:
Yes, I mean there were multiple Chrome zombie processes. It fixed with this PR.

p0deje:
Thank you, merged in e7d4f5e21e793cc8fd5fb16278d49d07b4f7d660

PR #3343, opened 2017-01-09
grab the lastest geckodriver in the travis install script

lmtierney:


lukeis:
i like it... except since the json blob from the github api contains the url of the thing we want, why not get it instead of guessing at the name, let's assume "linux64" will remain constant.

    curl -s 'https://api.github.com/repos/mozilla/geckodriver/releases/latest' | python -c "import sys, json; r = json.load(sys.stdin); print [a for a in r['assets'] if 'linux64' in a['name']][0]['browser_download_url'];"

lmtierney:
updated

PR #3346, opened 2017-01-10
[py] fix inverted logic and flake8 failure

lmtierney:
The logic got inverted for determining if it the service has shutdown. It should break if not connectable. This was making each test take an additional 30 seconds (see Travis failures)


lukeis:
thanks @lmtierney

PR #3361, opened 2017-01-14
Added ExpectedCondition AnyTextToBePresentInElement

Rieth:
Added ExpectedCondition to check if AnyTextToBePresentInElement ignoring white spaces.

### DIFF
--- a/dotnet/src/support/UI/ExpectedConditions.cs
+++ b/dotnet/src/support/UI/ExpectedConditions.cs
@@ -310,6 +310,27 @@ namespace OpenQA.Selenium.Support.UI
                 }
             };
         }
+      
+        /// <summary>
+        /// An expectation for checking if any text is present in the specified element.
+        /// </summary>
+        /// <param name="element">The WebElement</param>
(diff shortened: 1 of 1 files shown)


jimevans:
Pull requests for `ExpectedConditions` in the .NET bindings are not going to be accepted.it is the opinion of the .NET bindings maintainers that the class should not have been created for .NET in the first place. Anonymous methods (so-called "lambdas") in C# are a pervasive feature of the language, and the syntax for using them is nowhere near as cumbersome as the same construct in Jav, which is why that language can make good use of a helper class. This is the kind of thing that an add-on or companion project would be ideal for.

deyshin:
I was interested in mirroring Java's `ExpectedConditions` as well. For the developers using multiple languages, I think it will be helpful to have a similar API. 

I understand the power of Lambda Functions in C#/.NET. But `ExpectedConditions` does make it easier to utilize IDE to write easy-to-read code. 

Thought it may be a mistake to add the class in the first place, I feel it's worse to leave class half-finished. 

If there has been a discussion about this matter, please refer me to the discussion; I'm interested to follow through others' thoughts.

Issue #2704, opened 2016-09-01
Selenium::WebDriver::Element#send_keys no longer works as expected


twalpole:
OS:  
OSX/linux
Selenium Version:  
Head - 3.0.0.beta2.1
Browser:  
Firefox
Browser Version:  
49 beta
## Expected Behavior -

Expect modifier keys to be released when passed in an array
## Actual Behavior -

Modifier keys are no longer released 
## Steps to reproduce -

Using the ruby bindings (selenium-webdriver gem)

```
element.send_keys([:shift, 'a'], 'bcdef'
```

In previous (2.x) versions of selenium-webdriver with Firefox < 48 this would cause element to be sent the keys 'Abcdef'.  When used with the current head version and geckodriver/marionette it ends up sending "A&lt;null>BCDEF" to the element.  This appears to be because the Webdriver spec has removed the previous requirement that a NULL being sent should be interpreted as releasing key modifiers.  This also prevents things like `element.send_keys([:ctrl, 'a'], :backspace, 'abcd') for selecting the current contents of a field and sending new content.   The NULL is currently inserted by selenium [here](https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/common/keys.rb#L114).  Splitting up the passed items into multiple send key requests may solve this issue, although the current webdriver spec for [Element Send Keys](https://www.w3.org/TR/webdriver/#element-send-keys) requires using set selection range to set the insertion caret which could break things the [:ctrl, 'a'] shown above.

lukeis:
Please log GeckoDriver issues with Mozilla.

https://github.com/mozilla/geckodriver/issues

Please be sure to include a completely reproducible test script for them, without a way to reproduce the issue you are seeing there is no good way for them to fix the problem. A reproducible test script includes a WebDriver script and a link to the page in question. If you can not provide a link, create one using jsfiddle.net

twalpole:
@lukeis This isn't a geckodriver issue -- geckodriver is adhering to the current webdriver spec.  It's selenium that's sticking extra nulls into the keys being sent

lmtierney:
@twalpole can you see if #3375 works for this? I tried it with the latest geckodriver & Nightly and it seems to be working for your examples

lmtierney:
If the spec updates to handle null characters then I we can just strip it out, but until then this should provide the functionality

twalpole:
[...]

PR #3375, opened 2017-01-18
[rb] split w3c send_keys strings on null to release modifiers

lmtierney:
This hopefully addresses #2704


twalpole:
This isn't going to have the same behavior as before either, since the selection and caret position get reset at the beginning of every send_keys call.  There is an issue on the webdriver spec to add back in the behavior that was removed - https://github.com/w3c/webdriver/issues/553 - so it may be premature to pull this change in.

lmtierney:
Will put this PR on hold for w3c/webdriver#553

lmtierney:
Canceling as @twalpole showed this would not be an equivalent fix

PR #3379, opened 2017-01-18
fix python HTTPS encoding for python driver

llaskin:

### DIFF
--- a/py/selenium/webdriver/remote/remote_connection.py
+++ b/py/selenium/webdriver/remote/remote_connection.py
@@ -471,7 +471,7 @@ class RemoteConnection(object):
             request.add_header('Accept', 'application/json')
             request.add_header('Content-Type', 'application/json;charset=UTF-8')
 
-            if parsed_url.username
+            if parsed_url.username:
                 base64string = base64.b64encode('%s:%s' % (parsed_url.username, parsed_url.password))
                 request.add_header("Authorization", "Basic %s" % base64string)
 


py/selenium/webdriver/remote/remote_connection.py
```diff
@@ -471,6 +471,9 @@ def _request(self, method, url, body=None):
             request.add_header('Accept', 'application/json')
             request.add_header('Content-Type', 'application/json;charset=UTF-8')
 
+            base64string = base64.b64encode('%s:%s' % (parsed_url.username, parsed_url.password))
```
Dude-X: This will always add the authorization header to the request object. Is this the right scope for these two lines? If username/password are not defined, it will encode 'Basic :'

llaskin: good point.

llaskin: I think this needs to be wrapped in an if statement, but otherwise, it should be here.  Do you have any thoughts where to put it?

Dude-X: I don't know. It seems like it should go in one of the else blocks one or two levels above. Probably in the request object that doesn't have a password manager.

Dude-X: It doesn't seem to break the driver, but do we always want to pass an auth header in all of the requests?

llaskin:
@AutomatedTester care to give me your thoughts here, as to whats causing the failures I am seeing?

AutomatedTester:
which failures in particular?

llaskin:
https://travis-ci.org/SeleniumHQ/selenium/jobs/193472318 and the 2 firefox failures? One seems to be marionette and I haven't made any changes around that

AutomatedTester:
That linked failure is grid issue...

llaskin:
So ignore it?

PR #3411, opened 2017-01-25
[rb] refrain from sending unnecessary default capabilities for w3c

lmtierney:
Previously, 'any' was sent for some capabilities if they weren't set by the user, which does not line up with the current state of the W3C spec. Instead, leave them as undefined.


lmtierney:
merged with ac31f00

PR #3416, opened 2017-01-26
[rb] initial ruby w3c interactions implementation

lmtierney:


rb/lib/selenium/webdriver/common/interactions/pointer_input.rb
```diff
@@ -0,0 +1,153 @@
...
+          @element = element
+          @origin = origin
+        end
+
+        def type
+          :move
+        end
+
+        def assert_direction(direction)
+          raise TypeError, "#{direction.inspect} is not a valid pointer type" unless DIRECTIONS.include? direction
```
twalpole: not a valid direction?

lmtierney: Indeed, I'll wait a bit for any other reviews and update it accordingly. Thanks!

titusfortner:
What would we think about getting rid of these extensions for W3C and just putting it in the W3CBridge? The spec is going to say what is or isn't allowed, so there shouldn't be any reason to extend for them based on the driver.

Looks like we'd be following Jari's thoughts from 7 years ago. :)
https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/common/driver.rb#L91

rb/spec/integration/selenium/webdriver/key_input_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    compliant_on driver: :firefox do
+      describe W3CActionBuilder do
+        describe 'KeyInput interactions' do
+          it 'sends keys to the active element' do
+            driver.navigate.to url_for('bodyTypingTest.html')
+            keylogger = driver.find_element(id: 'body_result')
+
+            driver.action.send_keys('ab').perform
+
+            wait.until { keylogger.text.split.size == 2 }
```
luke-hill: Wouldn't a `.length` work quicker (As it's 1method not 2)? Apologies if misunderstanding.

From irb:

```
irb(main):001:0> string1 = 'ab'
=> "ab"
irb(main):002:0> string1.length
=> 2
```

rb/spec/integration/selenium/webdriver/key_input_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+            event_input = driver.find_element(id: 'theworks')
+            keylogger = driver.find_element(id: 'result')
+
+            event_input.click
+
+            driver.action.key_down(:shift).send_keys('ab').key_up(:shift).perform
+            wait.until { event_input.attribute(:value).size == 2 }
+
+            expect(event_input.attribute(:value)).to eq('AB')
+            expect(keylogger.text.strip).to match(/^(focus )?keydown keydown keypress keyup keydown keypress keyup keyup$/)
```
luke-hill: if we're not capturing this regex, we can use `[]` as non-capturing items are "slightly" faster. But again it's probably negligible

rb/lib/selenium/webdriver/common/interactions/pointer_actions.rb
```diff
@@ -0,0 +1,359 @@
...
+
+      def move_to(element, right_by = nil, down_by = nil, device_name: nil)
+        pointer = get_device(device_name) || primary_pointer
+        # New actions offset is from center of element
+        if right_by || down_by
+          size = element.size
+          left_offset = (size[:width] / 2).to_i
+          top_offset = (size[:height] / 2).to_i
+          left = -left_offset + right_by
+          top = -top_offset + down_by
```
lmtierney: I put it this way to show the steps of calculating the final value. Center point - left/top offset in order to get 0(ish) location, then applying the user's offset.

rb/lib/selenium/webdriver/common/interactions/pointer_input.rb
```diff
@@ -0,0 +1,146 @@
...
+        def type
+          @direction
+        end
+
+        def assert_button(button)
+          if button.is_a? Symbol
+            raise ArgumentError, "#{button.inspect} is not a valid button!" unless BUTTONS.key? button
+            button = BUTTONS[button]
+          end
+          raise ArgumentError, 'Button number cannot be negative!' unless button >= 0
```
lmtierney: A pointer is not limited to 3 buttons. Per the spec, the button number must only be >= 0 (section with heading including **process a pointer up or pointer down action**)

luke-hill: Think I'm getting confused with this then: https://github.com/SeleniumHQ/selenium/pull/3416/files#diff-6c3855ad2db7f00a616f2441fef00f56R76

Sorry :+1:

rb/spec/integration/selenium/webdriver/key_input_spec.rb
lmtierney: `keylogger.text.split` returns an Array, that being said your argument is still valid. `length` vs `size` is very marginal in speed especially at small sizes so it's really just user preference. I can change it to length if desired for code style

    Benchmark.measure { 10000000.times {[1,2,3,4].length}}
    => #<Benchmark::Tms:0x007f95450df020 @label="", @real=2.8369630000088364, @cstime=0.0, @cutime=0.0, @stime=0.020000000000000004, @utime=2.81, @total=2.83>

    Benchmark.measure { 10000000.times {[1,2,3,4].size}}
    => #<Benchmark::Tms:0x007f9543113b38 @label="", @real=2.8466349999653175, @cstime=0.0, @cutime=0.0, @stime=0.010000000000000009, @utime=2.8200000000000003, @total=2.83>

rb/spec/integration/selenium/webdriver/key_input_spec.rb
lmtierney: I suppose if we were really worried about speed in the specs, this should use `=~`. These `expect` were all copied from the current keyboard_spec, but I'd be happy to update both accordingly

rb/lib/selenium/webdriver/common/interactions/pointer_input.rb
lmtierney: Ah yep so that hash is just so that users have an easier way of specifying a standard mouse button without specifying an integer itself (see the `button.is_a?` conditional above that). Buttons specified as integers are also accepted

luke-hill: Yep, so basically 0-2 have alias' and 3+ (Which I'm guessing are the ones you see on gaming mice e.t.c.) and non-aliased

lmtierney:
merged 2cd8007

PR #3430, opened 2017-01-28
Fix javadocs for @FindBy

fil7:
Fix javadocs for @FindBy ('@' was not rendered in code samples):
http://seleniumhq.github.io/selenium/docs/api/java/org/openqa/selenium/support/FindBy.html


shs96c:
Thank you!

PR #3431, opened 2017-01-28
Allow for customisation of TestSlot

krmahadevan:
The current implementation does not let downstream
consumers of the Grid provide a customised way of 
injecting a TestSlot. 
This is very useful when one needs to build an On-Demand Grid, 
wherein there are no real proxies, but a ghost proxy stands 
for the real proxy and based on new session requests, 
nodes are spun off.

Fixed this by defining a method for instantiating a 
TestSlot so that downstream RemoteProxy implementations
can choose to provide their own customised TestSlots.

Also enhanced the SeleniumProtocol enum to house all
logic related to forming the protocol and the path
from a desired capability into it.

### DIFF
--- a/java/server/src/org/openqa/grid/common/SeleniumProtocol.java
+++ b/java/server/src/org/openqa/grid/common/SeleniumProtocol.java
@@ -17,8 +17,47 @@
 
 package org.openqa.grid.common;
 
+import static org.openqa.grid.common.RegistrationRequest.PATH;
+import static org.openqa.grid.common.RegistrationRequest.SELENIUM_PROTOCOL;
+
+import org.openqa.grid.common.exception.GridException;
+
+import java.util.Arrays;
+import java.util.Map;
+
 public enum SeleniumProtocol {
-  Selenium, WebDriver;
+  Selenium("/selenium-server/driver"),
+  WebDriver("/wd/hub");
+  private String path;
+
+  SeleniumProtocol(String path) {
+    this.path = path;
+  }
+
+  public static SeleniumProtocol fromCapabilitiesMap(Map<String, ?> capabilities) {
+    String type = (String) capabilities.get(SELENIUM_PROTOCOL);
+    if (type == null || type.trim().isEmpty()) {
+      return WebDriver;
+    }
+    try {
+      return SeleniumProtocol.valueOf(type);
+    } catch (IllegalArgumentException e) {
+      throw new GridException(type + " isn't a valid protocol type for grid. Valid values :[" +
+                              Arrays.toString(values()) + "]", e);
+    }
+  }
+
+  public String getPathConsideringCapabilitiesMap(Map<String, ?> capabilities) {
+    String localPath = (String) capabilities.get(PATH);
+    if (localPath != null) {
+      return localPath;
+    }
+    return path;
+  }
+
+  public String getPath() {
+    return path;
+  }
 
   public boolean isSelenium() {
     return Selenium.equals(this);
--- a/java/server/src/org/openqa/grid/internal/BaseRemoteProxy.java
+++ b/java/server/src/org/openqa/grid/internal/BaseRemoteProxy.java
@@ -18,8 +18,6 @@
 package org.openqa.grid.internal;
 
 import static org.openqa.grid.common.RegistrationRequest.MAX_INSTANCES;
-import static org.openqa.grid.common.RegistrationRequest.PATH;
-import static org.openqa.grid.common.RegistrationRequest.SELENIUM_PROTOCOL;
 
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
@@ -144,8 +142,7 @@ public class BaseRemoteProxy implements RemoteProxy {
     for (DesiredCapabilities capability : capabilities) {
       Object maxInstance = capability.getCapability(MAX_INSTANCES);
 
-      SeleniumProtocol protocol = getProtocol(capability);
-      String path = getPath(capability);
+      SeleniumProtocol protocol = SeleniumProtocol.fromCapabilitiesMap(capability.asMap());
 
       if (maxInstance == null) {
         log.warning("Max instance not specified. Using default = 1 instance");
@@ -158,43 +155,11 @@ public class BaseRemoteProxy implements RemoteProxy {
         for (String k : capability.asMap().keySet()) {
           c.put(k, capability.getCapability(k));
         }
-        slots.add(new TestSlot(this, protocol, path, c));
+        slots.add(newTestSlot( protocol, c));
       }
     }
 
     this.testSlots = Collections.unmodifiableList(slots);
-  }
-
-  private SeleniumProtocol getProtocol(DesiredCapabilities capability) {
-    String type = (String) capability.getCapability(SELENIUM_PROTOCOL);
-
-    SeleniumProtocol protocol;
-    if (type == null) {
-      protocol = SeleniumProtocol.WebDriver;
-    } else {
-      try {
-        protocol = SeleniumProtocol.valueOf(type);
-      } catch (IllegalArgumentException e) {
-        throw new GridException(
-            type + " isn't a valid protocol type for grid. See SeleniumProtocol enum.", e);
-      }
-    }
-    return protocol;
-  }
-
-  private String getPath(DesiredCapabilities capability) {
-    String type = (String) capability.getCapability(PATH);
-    if (type == null) {
-      switch (getProtocol(capability)) {
-        case Selenium:
-          return "/selenium-server/driver";
-        case WebDriver:
-          return "/wd/hub";
-        default:
-          throw new GridException("Protocol not supported.");
-      }
-    }
-    return type;
   }
 
   public void setupTimeoutListener() {
@@ -270,7 +235,7 @@ public class BaseRemoteProxy implements RemoteProxy {
     }
 
     void cleanUpAllSlots() {
-      for (TestSlot slot : testSlots) {
+      for (TestSlot slot : getTestSlots()) {
         try {
           cleanUpSlot(slot);
         } catch (Throwable t) {
@@ -342,7 +307,7 @@ public class BaseRemoteProxy implements RemoteProxy {
       return null;
     }
     // any slot left for the given app ?
-    for (TestSlot testslot : testSlots) {
+    for (TestSlot testslot : getTestSlots()) {
       TestSession session = testslot.getNewSession(requestedCapability);
 
       if (session != null) {
@@ -355,7 +320,7 @@ public class BaseRemoteProxy implements RemoteProxy {
   public int getTotalUsed() {
     int totalUsed = 0;
 
-    for (TestSlot slot : testSlots) {
+    for (TestSlot slot : getTestSlots()) {
       if (slot.getSession() != null) {
         totalUsed++;
       }
@@ -365,7 +330,7 @@ public class BaseRemoteProxy implements RemoteProxy {
   }
 
   public boolean hasCapability(Map<String, Object> requestedCapability) {
-    for (TestSlot slot : testSlots) {
+    for (TestSlot slot : getTestSlots()) {
       if (slot.matches(requestedCapability)) {
         return true;
       }
@@ -544,7 +509,7 @@ public class BaseRemoteProxy implements RemoteProxy {
 
   public long getLastSessionStart() {
     long last = -1;
-    for (TestSlot slot : testSlots) {
+    for (TestSlot slot : getTestSlots()) {
       last = Math.max(last, slot.getLastSessionStart());
     }
     return last;
--- a/java/server/src/org/openqa/grid/internal/RemoteProxy.java
+++ b/java/server/src/org/openqa/grid/internal/RemoteProxy.java
@@ -20,7 +20,7 @@ package org.openqa.grid.internal;
 import com.google.gson.JsonObject;
 
 import org.openqa.grid.common.RegistrationRequest;
-import org.openqa.grid.common.exception.GridException;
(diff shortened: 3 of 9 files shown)


krmahadevan:
ping @lukeis - Can you please help vet this out ?

java/server/src/org/openqa/grid/common/SeleniumProtocol.java
```diff
@@ -17,8 +17,50 @@
...
+    } catch (IllegalArgumentException e) {
+      throw new GridException(type + " isn't a valid protocol type for grid. Valid values :[" +
+                              Arrays.toString(values()) + "]", e);
+    }
+  }
+
+  private void computePath(DesiredCapabilities capability) {
+    String localPath = (String) capability.getCapability(PATH);
+    if (localPath != null) {
+      this.path = localPath;
```
lukeis: i think this is pretty dangerous, you're changing the enum's value of path. enums are static, so you're changing the path for *every* instance of the protocol type.

I'm imagining a hub that has multiple nodes with different paths.... if one changes the path, then all subsequent will have that new path, rather than the default /wd/hub path that they might assume.

java/server/src/org/openqa/grid/internal/RemoteProxy.java
```diff
@@ -39,6 +39,16 @@
...
  */
 public interface RemoteProxy extends Comparable<RemoteProxy> {
+
+  /**
+   *
+   * @param protocol - A {@link SeleniumProtocol} object that identifies the request flavor.
+   * @param capabilities - the type of test the client is interested in performing.
+   * @return - The entity on a proxy that can host a test session.
+   */
+  default TestSlot newTestSlot(SeleniumProtocol protocol, Map<String, Object> capabilities) {
```
lukeis: it's like you're trying to make a TestSlot factory so you can override it. I'm happy to have selenium provide a mechanism for that, just not sure this is the way I'd approach it. If you don't want to make another Factory class, maybe getTestSlotClass? I'm not sure....

@mach6 have any suggestions?

java/server/src/org/openqa/grid/common/SeleniumProtocol.java
krmahadevan: @lukeis - Thanks for catching this. I completely missed that part. I will fix this.

java/server/src/org/openqa/grid/internal/RemoteProxy.java
krmahadevan: If we resort to a getTestSlotClass() wherein we are returning the actual test slot class, I thought that would bring in reflection (because now the proxy would have to instantiate the class) and we also would need to add additional checks which figures out if the class that was provided satisfies the `isAssignableFrom()` clause and all that. That was why I thought I would keep it simple by just defining a default method in the `RemoteProxy` interface, and any one specifically looking for customizing TestSlots can choose to explicitly override that method from their `DefaultRemoteProxy` extension. But please do let me know if there are other options that can be considered so that I can explore that and get this sorted out.

java/server/src/org/openqa/grid/common/SeleniumProtocol.java
krmahadevan: @lukeis - I have fixed this and also added a unit test as well.

java/server/src/org/openqa/grid/internal/RemoteProxy.java
krmahadevan: @mach6 - Please let me know your suggestions if any. I would like to wrap this PR up, so am trying to find out what else is pending from my side.

krmahadevan: ping @mach6 . Any updates on this ?

ping @lukeis - If there are no updates from @mach6 , please let me know what more is needed from my side to get this merged.

mach6: I'd like to see a test that exercise the changes to TestSlot, overriding the default behavior.

java/server/src/org/openqa/grid/common/SeleniumProtocol.java
```diff
@@ -17,8 +17,44 @@
...
+    }
+    try {
+      return SeleniumProtocol.valueOf(type);
+    } catch (IllegalArgumentException e) {
+      throw new GridException(type + " isn't a valid protocol type for grid. Valid values :[" +
+                              Arrays.toString(values()) + "]", e);
+    }
+  }
+
+  public String getPath(Map<String, Object> capabilities) {
```
mach6: It seems inconsistent to me that `getProtocol` takes `DesiredCapabilities` and `getPath` takes a `Map<String, Object>`.   Can they both take `DesiredCapabilities` as the original `private` code (in `BaseRemoteProxy`) that this was moved from did? 

Also, now that these methods are `public`, I'd like to suggest names of 
- `public static SeleniumProtocol  fromDesiredCapabilities()`
- `public String getPathConsideringDesiredCapabilities()`
- `public String getPath()` which returns ONLY the configured (via the enum's constructor) `this.path`

java/server/src/org/openqa/grid/internal/BaseRemoteProxy.java
```diff
@@ -342,7 +309,7 @@ public TestSession getNewSession(Map<String, Object> requestedCapability) {
       return null;
     }
     // any slot left for the given app ?
-    for (TestSlot testslot : testSlots) {
+    for (TestSlot testslot : getTestSlots()) {
```
mach6: getTotalUsed(), hasCapability() -- other places in this class still enumerate on the private immutable List `testSlots`. Will that be a problem per you intended change?

java/server/src/org/openqa/grid/internal/BaseRemoteProxy.java
```diff
@@ -158,45 +157,13 @@ public BaseRemoteProxy(RegistrationRequest request, Registry registry) {
...
-        slots.add(new TestSlot(this, protocol, path, c));
+        slots.add(newTestSlot( protocol, c));
       }
     }
 
     this.testSlots = Collections.unmodifiableList(slots);
   }
 
-  private SeleniumProtocol getProtocol(DesiredCapabilities capability) {
-    String type = (String) capability.getCapability(SELENIUM_PROTOCOL);
```
mach6: unused import left in this class, as a result of this change

java/server/src/org/openqa/grid/internal/BaseRemoteProxy.java
```diff
@@ -158,45 +157,13 @@ public BaseRemoteProxy(RegistrationRequest request, Registry registry) {
...
-      } catch (IllegalArgumentException e) {
-        throw new GridException(
-            type + " isn't a valid protocol type for grid. See SeleniumProtocol enum.", e);
-      }
-    }
-    return protocol;
-  }
-
-  private String getPath(DesiredCapabilities capability) {
-    String type = (String) capability.getCapability(PATH);
```
mach6: unused import left in this class, as a result of this change

java/server/src/org/openqa/grid/internal/RemoteProxy.java
krmahadevan: @mach6 

> I'd like to see a test that exercises the changes to TestSlot, overriding the default behaviour.

What exactly would we be asserting in these tests ? A downstream user is ideally speaking going to be sub-classing `TestSlot` and then work with that sub-class by accessing it via the `TestSession`. Can you please elaborate the unit test requirements here ?

java/server/src/org/openqa/grid/internal/BaseRemoteProxy.java
krmahadevan: fixed.

java/server/src/org/openqa/grid/internal/BaseRemoteProxy.java
krmahadevan: fixed.

java/server/src/org/openqa/grid/internal/BaseRemoteProxy.java
krmahadevan: It shouldn't be a problem because the only place wherein we would like to leverage the logic of `BaseRemoteProxy` is when we are iterating the `TestSlot` list for new session creation. But just to ensure that we are consistent across the board, I have fixed all the places to access the `testSlots` list only via the `getTestSlots()` method.

java/server/src/org/openqa/grid/common/SeleniumProtocol.java
krmahadevan: Fixed. We now work with only the `Map` since `DesiredCapabilities` is not available for access from within `TestSlot` (that's the place wherein `getPathConsideringDesiredCapabilities()` would be accessed).

java/server/src/org/openqa/grid/internal/RemoteProxy.java
krmahadevan: ping @mach6

krmahadevan:
ping @lukeis @mach6 - Gentle reminder :) Please let me know what else is needed from myside before we get this merged.

java/server/src/org/openqa/grid/common/SeleniumProtocol.java
```diff
@@ -17,8 +17,47 @@
...
+    }
+    try {
+      return SeleniumProtocol.valueOf(type);
+    } catch (IllegalArgumentException e) {
+      throw new GridException(type + " isn't a valid protocol type for grid. Valid values :[" +
+                              Arrays.toString(values()) + "]", e);
+    }
+  }
+
+  public String getPathConsideringDesiredCapabilities(Map<String, ?> capabilities) {
```
mach6: But it's not considering "desired" capabilities anymore. It's just "capabilities" so the method names needs to better reflect this. Same with `fromDesiredCapablities` above.

java/server/src/org/openqa/grid/internal/RemoteProxy.java
mach6: Shouldn't you be able to create a `DummyTestSlot` for the purposes of a functional test?

java/server/src/org/openqa/grid/common/SeleniumProtocol.java
krmahadevan: fixed again.

java/server/src/org/openqa/grid/internal/RemoteProxy.java
krmahadevan: @mach6 - Yes I very much can. But I would like to understand the intent of the functional test. It can at the max just do an `instanceof` check in the assertion. So can you please help me understand what would the test actually be doing as a value add. There is no change in code. The only thing that has been done in this PR is to move the instantiation of the `TestSlot` to an interface (which has been given a default behaviour as well)

krmahadevan:
@mach6 - I have added a test that leverages a custom slot via the proxy. Please let me know if there is anything you need from me additionally to facilitate this merge.

mach6:
@krmahadevan funny. I just pushed this -- https://github.com/mach6/selenium/commit/622abc5a690d2a78781a2d6d23dcf073f8d9945e

Seems we working this in parallel. 

Let's combine the salient points from both our updates to this PR. Will you make an attempt at this or should I?

krmahadevan:
@mach6 - I didn't realise that you were working on this as well. I looked at the commit you shared.
I did add a functional test (just as how you added), and I fixed the SeleniumProtocolTest (boneheaded mistake from my side). IMO there's nothing left to combine in both. You can very well proceed with my PR and if it satisfies all the pre-requisites of the review, I think its good to be merged. Please let me know if that makes sense.

mach6:
@krmahadevan yep. the test approach is basically the same. I did add javadocs to many of the methods, rename the method from `newTestSlot` to `createTestSlot`, and remove some unnecessary `Before` code in `BaseRemoteProxyTest`.. I can merge these changes later, I reckon. Either way it should be on top of this PR.

krmahadevan:
@mach6  - Thanks for helping with this and getting this merged.

mach6:
@krmahadevan please rebase this so we can see what happens with the CI result. `grid_tests` are broken for this PR. I want to ensure this is not a result of the changes in the PR. Upstream `grid_tests` are not currently broken. Also please squash.

krmahadevan:
@mach6 - Sure will do.Can you please tell me how do I run the `grid_tests` alone locally ?

mach6:
simply run the following from a terminal while in the base directory of the project
```shell 
$ ./go test_grid
```

krmahadevan:
@mach6 - I found the reason why the build is failing. Its due to ` java.lang.NoClassDefFoundError: Could not initialize class org.openqa.selenium.WebDriverException` that is arising out of `SeleniumProtocol` because my change includes this exception. But I am not able to figure out how do I get `BUCK` to include this. Any suggestsions on how do I get this included ?

I have rebased this off of master and squashed my commits into one and pushed them as well.

mach6:
fyi, the `test_grid` step passed -- https://travis-ci.org/SeleniumHQ/selenium/builds/201752827

krmahadevan:
Oh ok.  I still haven't figured out how Buck works.. 
Does it mean that we are good now and this PR can be merged ?

krmahadevan:
Ok I guess I now know what is the root cause of the test failures on my mac. For some weird reason `NetworkInterface.getByName("en0")` returns null. Currently there is no null check in [WebDriverException](https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/WebDriverException.java#L92-L97). That explains all the other errors that I faced on my machine. Now I would need to figure out why I don't get a valid `NetworkInterface` for the name `en0` on my MAC

mach6:
This stands merged. Thanks!

PR #3432, opened 2017-01-30
Add support for ISO 8601 date format for logs

artgo:

Add support for [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) date format `yyyy-MM-dd'T'HH:mm:ss.SSS'Z'` as well as already converted Date object.


artgo:
It is being passed by Android from Appium. See https://github.com/appium/appium/issues/6785

artgo:
Lint failures (in appveyor) are unrelated and are in different files.

artgo:
@lukeis FYI :)

shs96c:
This is a bug on the part of Appium. The expectation is that log entry timestamps are (admittedly rather oddly for JSON) numbers representing the UNIX Epoch timestamp at which the log entry was created.

Thank you, however, for taking the time to file this PR!

It should be noted that `SimpleDateFormat` is not thread-safe. Putting it on the `static` field will at some point lead to unusual error cases.

PR #3438, opened 2017-02-01
[rb] Fix element equality for elements returned from execute_script with marionette

twalpole:

Elements returned to the ruby binding (selenium-webdriver) from `execute_script` were failing equality when using marionette.

### DIFF
--- a/rb/lib/selenium/webdriver/common/bridge_helper.rb
+++ b/rb/lib/selenium/webdriver/common/bridge_helper.rb
@@ -32,7 +32,13 @@ module Selenium
           arg.map { |e| unwrap_script_result(e) }
         when Hash
           element_id = element_id_from(arg)
-          element_id = arg if self.is_a?(Remote::W3CBridge) && element_id
+          if element_id
+            if self.is_a?(Firefox::W3CBridge)
+              element_id = {'element-6066-11e4-a52e-4f735466cecf' => element_id}
+            elsif self.is_a?(Remote::W3CBrdige)
+              element_id = arg
+            end
+          end
           return Element.new(self, element_id) if element_id
           arg.each { |k, v| arg[k] = unwrap_script_result(v) }
         else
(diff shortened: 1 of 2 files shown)


twalpole:
@titusfortner This was the simplest solution to the issue.  I'm not really clear on why the elements ids are actually hashes with one constant key, but wasn't sure how many other places were dependant on it so didn't want to make a large change.

titusfortner:
Oh look, I definitely implemented this poorly. It think it makes sense though for this logic to go in the `#element_id_from`, though. What do you think of this?
```
def element_id_from(result)
        oss = result.delete 'ELEMENT'
        if self.is_a? Remote::Bridge
          oss
        elsif result.keys.all? { |k| k[/^element-/] }
          result
        end
end
```

Then we can just get rid of the line I wrote in the first place and everything should be good:
`element_id = arg if self.is_a?(Remote::W3CBridge) && element_id`

Also, I probably need to update `it 'should unwrap elements in deep objects' do` since it isn't verifying that the Element was properly created. Maybe I need to do that verification in the Element instantiation.

twalpole:
@titusfortner  I'm not sure it makes sense to use a regex to match the other keys- the webdriver spec specifies the 'element-6066-11e4-a52e-4f735466cecf' key so what does allowing others gain us?  Unless IE is using something different??

titusfortner:
Whoa. I completely didn't realize that was a hard-coded value in the spec.
This makes more sense now, except wtf. I'll have to ping the authors
tomorrow about why this is.

twalpole:
To make odds of a collision tiny

titusfortner:
merged here: f14424f7327d446424572a0e2c9261f85039e42a

Thanks for pointing out this issue. It turned out I was doing things in `Edge::Bridge` and `Remote::W3CBridge` that were completely unnecessary, so the code is much cleaner now.

PR #3443, opened 2017-02-01
Handle EPIPE errors along with the current ECONNRESET

matthewsanders:
There is an open issue with Node.js throwing EPIPE errors when you pipe to a closed stream.  This was causing me unpredictable errors while loading pages.

The errors are pretty similar to ECONNRESET in the sense that you could simply try again and shouldn't throw every time it happens.

More info here: https://github.com/nodejs/node/issues/947

### DIFF
--- a/javascript/node/selenium-webdriver/http/index.js
+++ b/javascript/node/selenium-webdriver/http/index.js
@@ -226,7 +226,7 @@ function sendRequest(options, onOk, onError, opt_data, opt_proxy) {
   });
 
   request.on('error', function(e) {
-    if (e.code === 'ECONNRESET') {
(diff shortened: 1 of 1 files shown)


wluu:
I've been running into this issue with Protractor: https://github.com/angular/protractor/issues/4294

After manually adding that fix, I have not seen the EPIPE errors in our test suites. Granted, I've been running our test suites with the fix only a couple of times so far.

jleyba:
Sorry we let this PR sit for a year without review :(

Retries for EPIPE errors was added in c41f9ef2435328b3cd05a892769db6697ec5e9e2

PR #2210, opened 2016-06-04
Update bundled JRuby to 9.1.2.0 fully working on Java 8

p0deje:
@titusfortner @shs96c I'm not sure that's what you guys want, but I sort of managed to upgrade bundle JRuby version to 9k.

``` bash
$ java -jar third_party/jruby/jruby-complete.jar -version
jruby 9.1.2.0 (2.3.0) 2016-05-26 7357c8f Java HotSpot(TM) 64-Bit Server VM 25.45-b02 on 1.8.0_45-b14 +jit [darwin-x86_64]

$ java -jar third_party/jruby/jruby-complete.jar -S gem list

*** LOCAL GEMS ***

albacore (1.0.0)
builder (3.2.2)
ci_reporter (2.0.0)
diff-lcs (1.1.3)
jar-dependencies (default: 0.3.2)
jruby-openssl (0.9.15 java)
json (1.8.3 java)
json_pure (1.8.3)
minitest (default: 5.4.1)
net-telnet (default: 0.1.1)
nokogiri (1.6.7.2 java)
power_assert (default: 0.2.3)
psych (2.0.17 java)
racc (1.4.13 java)
rake (10.5.0, default: 10.4.2)
rdoc (default: 4.2.0)
rspec (2.4.0)
rspec-core (2.4.0)
rspec-expectations (2.4.0)
rspec-mocks (2.4.0)
rubyzip (1.2.0)
test-unit (default: 3.1.1)
```

I'm not sure if tasks are actually working (only checked `//rb:unit-test`), but now it fails with circular dependency:

``` bash
$ ./go //rb:unit-test
...
go aborted!
Circular dependency detected: TOP => //rb:unit-test => //rb:android => //rb:common => //java/client/test/org/openqa/selenium/environment => //java/client/test/org/openqa/selenium/environment:environment => build/java/client/test/org/openqa/selenium/environment/environment.jar => //java/client/test/org/openqa/selenium/environment/webserver => //java/client/test/org/openqa/selenium/environment/webserver:webserver => build/java/client/test/org/openqa/selenium/environment/webserver/webserver.jar => //javascript:calcdeps => //javascript:calcdeps
```

I actually have no idea how to fix it and why it didn't fail before. It looks like javascript build depends on java build which depends on javascript build :)

Let me know if that is what you need (at least for the start) or if there is anything else I can do to help you.

PR #3444, opened 2017-02-02
Allow Java options when starting a standalone selenium server in Ruby

aliking:
Existing server allows additional args to be passed in, but they are only placed at the end of the constructed command line. For any options to the `java` command, they need to be placed before the `-jar` switch. 

This pulls out any additional args that start with `-D` and put them in the right place. 

Also includes a change to the gemspec checking that the __FILE__ location is the same as the pwd. This will probably be unnecessary form most people, but was failing on my jenkins server, due to mounted directories. 

### DIFF
--- a/rb/lib/selenium/server.rb
+++ b/rb/lib/selenium/server.rb
@@ -233,10 +233,8 @@ module Selenium
     def process
       @process ||= (
         # extract any additional_args that start with -D as options
-        options = if @additional_args
-          @additional_args.dup - @additional_args.delete_if{|arg| arg.match(/^-D.*$/)}
-        end
-        cp = ChildProcess.build('java', *options, '-jar', @jar, '-port', @port.to_s, *@additional_args)
+        properties = @additional_args.dup - @additional_args.delete_if { |arg| arg[/^-D/] }
+        cp = ChildProcess.build('java', *properties, '-jar', @jar, '-port', @port.to_s, *@additional_args)
         io = cp.io
 
         if @log.is_a?(String)
--- a/rb/selenium-webdriver.gemspec
+++ b/rb/selenium-webdriver.gemspec
@@ -6,7 +6,7 @@ raise "cwd(#{cwd}) must be #{root} when reading gemspec" if root != cwd
 
 Gem::Specification.new do |s|
   s.name = 'selenium-webdriver'
-  s.version = '3.0.6'
+  s.version = '3.0.5'
 
   s.authors = ['Alex Rodionov', 'Titus Fortner']
   s.email = ['p0deje@gmail.com', 'titusfortner@gmail.com']


lmtierney:
Just out of curiosity, why are you starting standalone with ruby?

rb/lib/selenium/server.rb
```diff
@@ -232,7 +232,11 @@ def stop_process
 
     def process
       @process ||= (
-        cp = ChildProcess.build('java', '-jar', @jar, '-port', @port.to_s, *@additional_args)
+        # extract any additional_args that start with -D as options
+        options = if @additional_args
+          @additional_args.dup - @additional_args.delete_if{|arg| arg.match(/^-D.*$/)}
```
titusfortner: `@additional_args` is always going to return `true` since it defaults to `[]`
let's name it `properties` or `system_properties`, and we should avoid using `#match` where possible. Otherwise this is good. :)
I think my preference would be:
```
properties = @additional_args.dup - @additional_args.delete_if { |arg| arg[/^-D/] }
```

aliking:
Addressed comments. Thanks for being so responsive.

These are both a bit unique to our setup. Our jenkins box has two levels of symlinking on the jenkins home directory - with `expand_path` I was getting two paths that were logically the same, but lexically different. 

The reason that we're starting Selenium server in this way is down to a framework that we're using which connects to the selenium server as though it is remote regardless of where it is. I've considered changing it, but it keeps other code consistent.

titusfortner:
Merged into master, will be available on next release.

@aliking - is this an open source framework, or something custom for your company? Thanks for this contribution.

aliking:
It's https://github.com/chemistrykit/chemistrykit and https://github.com/chemistrykit/selenium-connect. At this point our fork has diverged a lot from that, but I'm less concerned since it's not being updated.  I'm trying to unpick some of the dependencies on small customizations on old forks of other things, so I'm super glad to not have to do that with selenium. Thanks again!

titusfortner:
Ah Dave's framework. The different chemistry terms confused me too much, and it looked like it replicated a lot of test runner things that I didn't see use for. So your fork is not open source?

titusfortner:
@aliking - Bad news on the `File#realpath` usage. All of Selenium is built with an old customized version of JRuby. So far we haven't been able to get everything working with an updated JRuby, so we're stuck with its limitations. For the most part it doesn't matter for the rest of the Ruby code. Except it needs to read the gemspec file to build the gem, and it is too old to know about `#realpath`.

@p0deje - I think I need to revert this before the next release. We could theoretically work around the JRuby build process for this, but that way lies madness. I'm not sure how close you were to figuring out the updated JRuby, and now that we got things working in Buck with what is there it isn't any kind of a priority any longer. Any other thoughts?

You can see the error with: `./go //rb:gem:build`

p0deje:
@titusfortner I've spent a couple of hours trying to upgrade JRuby once again, but it's much more complicated than it was before because it now uses ChildProcess inside :( I can definitely figure it out but I don't want to do this again if it's not going to be merged like with #2210. Need your advice on this :)

titusfortner:
Thanks @p0deje, I went ahead and reverted this change for now.

I was the one who pointed @shs96c to use ChildProcess. The issue had to do with something he needed in a standard library that wasn't working with our legacy JRuby (I'd have to dig through conversations from last summer to refresh my memory). Let's come up with a plan for this when we're all face to face at the Selenium Conference in April.

p0deje:
> Let's come up with a plan for this when we're all face to face at the Selenium Conference in April.

Sounds reasonable.

PR #3445, opened 2017-02-02
[rb] prevent SocketError when default port not free

titusfortner:
By happenstance the default port for Chromedriver isn't working on my machine.

[This block](https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/common/service.rb#L71-L75) executes the `SocketLock#locked` method before it executes the `Service#find_free_port` method. But the `SocketLock#lock` method, when it calls `TCPServer#new` on a locked port, ends up throwing a `SocketError` which causes `SocketLock#lock` to timeout with `Error::WebDriverError`.

Removing the `lock` call here seems to fix it. I've spent the past hour trying to write a spec that replicates what I'm seeing without success.

@p0deje - can you take a look at this and let me know if this is the correct fix, and help me figure out how to write a spec for it?

### DIFF
--- a/rb/lib/selenium/webdriver/common/socket_lock.rb
+++ b/rb/lib/selenium/webdriver/common/socket_lock.rb
@@ -35,13 +35,9 @@ module Selenium
       #
 
       def locked
-        lock
-
-        begin
-          yield
-        ensure
-          release
-        end
+        yield
+      ensure
+        release
       end
 
       private


p0deje:
So now there is no lock at all :)

The idea to have socket lock is to make sure we start multiple processes sequentially to avoid race conditions. Lock happens not on default ChromeDriver port, but on `port + 1`.

1. We try to lock on `port + 1`.
2. If it is locked, we wait until it's released.
3. If it is not locked, we lock it and then search for port, start process, etc.

It's probably something about your environment, we can pair and try to figure it out, but I'm sure we shouldn't merge it.

titusfortner:
Oh, right, yes, this is definitely the wrong fix.

Is the only reason we are polling to allow for multiple chromedriver processes to be running on different ports? So it is outside the scope of the code to ensure that the port we are using for the socket lock is free?

The default chromedriver port is 9415
[This line](https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/common/service.rb#L138) tries to lock on 9414.
Right now on my machine that address is in use, so it raises / rescues `Errno::EADDRINUSE` exceptions until the `SOCKET_LOCK_TIMEOUT` is exceeded when it raises the aforementioned `SocketError`.

In my case I'm sure restarting my machine will get it to work, I just thought I'd see if the code was supposed to be able to accommodate this issue.

@p0deje - I'll close this, we can talk more later if you think we can do something more here.

twalpole:
So the socket locking is for use as an interprocess mutex?  Maybe using something like File#flock would be better (if it's supported by all the platforms selenium supports) rather than depending on a port not to be used by any other process.

p0deje:
@twalpole Yes, it's basically a mutex. As far as I remember, locking with socket was chosen since it's available for all the platforms (Linux, macOS, BSD, Windows, JRuby) (opposing to `flock` which is available for POSIX only IIRC).

twalpole:
Hmm, my understanding was that flock was supported on modern windows and posix (Linux, bad, macOS, etc) but I could be wrong about that. Also not sure what other platforms selenium-webdriver claims to support

PR #3452, opened 2017-02-03
Add AttributeError handling when stopping.

seanpianka:
Add AttributeError handling when calling Service.stop. I noticed that this issue occurs after attempting to initialize a webdriver.Chrome instance while 'chromedriver' is not in PATH.

Here is the aforementioned error:

`Message: 'chromedriver' executable needs to be in PATH. Please see https://sites.google.com/a/chromium.org/chromedriver/home`

And here is the aforementioned stacktrace:

```
Exception ignored in: <bound method Service.__del__ of <selenium.webdriver.chrome.service.Service object at 0x7fae11780a20>>
  File "/home/sean/.../lib/python3.6/site-packages/selenium/webdriver/common/service.py", line 176, in __del__
    self.stop()
  File "/home/sean/.../lib/python3.6/site-packages/selenium/webdriver/common/service.py", line 146, in stop
    if self.process is None:
AttributeError: 'Service' object has no attribute 'process'
```

The proposed changes should not introduce any adverse effects as it will handle similarly to a situation where `self.process` does exist.

### DIFF
--- a/py/selenium/webdriver/common/service.py
+++ b/py/selenium/webdriver/common/service.py
@@ -142,7 +142,10 @@ class Service(object):
             except Exception:
                 pass
 
-        if self.process is None:
+        try:
+            if self.process is None:
+                return
+        except AttributeError:  # self has no attribute 'process'
             return
 
         try:


cgoldberg:
+1

seanpianka:
Did handling an `AttributeError` honestly cause that many normally-passing tests to fail?

seanpianka:
Further errors, described in the following stacktrace, likely bear similar root causes (`chromedriver` path being supplied while it is not in `PATH`).
```
Exception ignored in: <bound method Service.__del__ of <selenium.webdriver.chrome.service.Service object at 0x7f5144f8da20>>
  File "/home/sean/Development/.virtualenvs/vindicta/lib/python3.6/site-packages/selenium/webdriver/common/service.py", line 173, in __del__
  File "/home/sean/Development/.virtualenvs/vindicta/lib/python3.6/site-packages/selenium/webdriver/common/service.py", line 149, in stop
  File "/home/sean/Development/.virtualenvs/vindicta/lib/python3.6/site-packages/selenium/webdriver/common/service.py", line 120, in send_remote_shutdown_command
ImportError: sys.meta_path is None, Python is likely shutting down
```

PR #3459, opened 2017-02-04
fix bug in proxy constructor that some properties are not proper set

zhuopeng:
in the __init__ method of ```Proxy``` class,  ```socksProxy```, ```socksUsername```, ```socksPassword``` are set without go though their property setter, thus could cause potential problem. 

The fix is simply set those properties with already defined setters :)


Dude-X:
Looks good to me +1

PR #3463, opened 2017-02-05
Build the `get-text` atom needed by the w3c spec.

djurassl:
We assume that all compliant browsers have support for XPath
and so no longer rely on wgxpath.


lmtierney:
?

shs96c:
This was branch was created to give a stable point for the W3C WebDriver spec to point to and should not be merged into the tree.

Issue #3465, opened 2017-02-06
Erro in class invisibility_of_element_located


reinaldorossetti:
OS:  
Windows 10
Selenium Version:  
3.0.2, IDE ATOM
Browser:  
Firefox and Chromium

Browser Version:  
FF 51.01 (64-bit)
Chromie Versão 55.0.2883.87 m

Script Test:

```
from selenium import webdriver
from selenium.webdriver.support.ui import WebDriverWait
from time import sleep
from selenium.webdriver.common.by import By
from selenium.webdriver.support import expected_conditions
from selenium.common.exceptions import ElementNotVisibleException
from selenium.common import exceptions

driver = webdriver.Firefox()
driver.get("http://www.jqueryrain.com/?eKV1bCMa")

def wait_element_present(driver, elem):
    present = WebDriverWait(driver, 60).until(
        expected_conditions.presence_of_element_located((By.CSS_SELECTOR, elem))
    )
    print(present)
    return present

def wait_element_not_present(driver, elem):
    present = WebDriverWait(driver, 15, ignored_exceptions=\
            exceptions.NoSuchElementException).until(
        expected_conditions.invisibility_of_element_located((By.CSS_SELECTOR, elem))
    )
    print(present)
    return present

driver.switch_to.frame(0)
button = wait_element_present(driver, "button#quick_demo")
button.click()

test = wait_element_present(driver, ".loadingoverlay")
test2 = wait_element_not_present(driver, ".loadingoverlay")
```

Show Error:

```
<selenium.webdriver.firefox.webelement.FirefoxWebElement (session="f2103a91-b8b9-46a0-bc0b-b8600151608e", element="a1573d62-794a-4fd8-8a06-6ee16b6046be")>
<selenium.webdriver.firefox.webelement.FirefoxWebElement (session="f2103a91-b8b9-46a0-bc0b-b8600151608e", element="977eaa87-b6ee-4404-84df-3776569a314b")>
<selenium.webdriver.firefox.webelement.FirefoxWebElement (session="f2103a91-b8b9-46a0-bc0b-b8600151608e", element="977eaa87-b6ee-4404-84df-3776569a314b")>
<selenium.webdriver.firefox.webelement.FirefoxWebElement (session="f2103a91-b8b9-46a0-bc0b-b8600151608e", element="977eaa87-b6ee-4404-84df-3776569a314b")>
[...]

PR #3473, opened 2017-02-07
fix the problem when the element is string

reinaldorossetti:
fix the problem when the element is string  or dict.

https://github.com/SeleniumHQ/selenium/issues/3465


reinaldorossetti:
fix the problem when the element is string or dict.

PR #3474, opened 2017-02-07
[rb] Initial implementation of Logger class

p0deje:
This is initial and dead simple implementation of Logger for Ruby.

Apart from obvious `logger.debug` calls, it takes all the output from child processes allowing to write them to log as well.

By default, it writes log to a tempfile (though doesn't unlink it after) which means it's silent unless you tell it to write to a known location or standard output:

```ruby
Selenium::WebDriver::Logger.io = STDOUT # or File.open
Selenium::WebDriver::Logger.level = Logger::DEBUG # or any other level
```

There is still a lot to wrap here, but I'd like to hear your thoughts on this.

### DIFF
--- a/rb/selenium-webdriver.gemspec
+++ b/rb/selenium-webdriver.gemspec
@@ -26,7 +26,7 @@ HTML of the application.'
   s.add_runtime_dependency 'childprocess', ['~> 0.5']
   s.add_runtime_dependency 'websocket', ['~> 1.0']
 
-  s.add_development_dependency 'rspec', ['< 3.5']
+  s.add_development_dependency 'rspec', ['~> 3.0']
   s.add_development_dependency 'rack', ['~> 1.0']
   s.add_development_dependency 'ci_reporter', ['~> 1.6', '>= 1.6.2']
   s.add_development_dependency 'webmock', ['~> 2.0']


p0deje:
> So, I didn't even realize $DEBUG was a default Ruby thing. Running test just with the Logger cleans everything up so nicely; this is going to be hugely useful.

So I actually left support for $DEBUG, if the script is run with `-d` then I set logger level to `:debug` which:

1. Enables all logger messages.
2. Redirect all output from childprocess to logger output (though without formatting).

For example:

```
I, [2017-02-17T10:05:56.548406 #40369]  INFO -- Selenium: polling for socket on ["127.0.0.1", 9515]
Starting ChromeDriver 2.27.440174 (e97a722caafc2d3a8b807ee115bfb307f7d2cfd9) on port 9515
I, [2017-02-17T10:05:56.800383 #40369]  INFO -- Selenium: -> POST session
```

> I don't think we should require $DEBUG = true for things to get logged since $DEBUG = true will report a lot more than we usually want.

It is still possible to enable debug logging for Selenium only:

```ruby
```

> I'm tentatively thinking that we have the default value be WARN, and wire calls to/from Ruby bindings recorded at INFO level, and then if DEBUG is set, we automatically turn on debugging for server & drivers. Would that make sense?

Done, though keep in mind that drivers usually has their own loggers (e.g. `--logpath` for ChromeDriver) so it's not super useful to redirect driver stdout to logger. Maybe we don't even need this, not sure.

> I think we want to default to STDOUT not to a file.

So now it's STDOUT, but you can use file:

```ruby
Selenium::WebDriver.logger.output = 'selenium.log'
```

> I'd like the ability to do something like this so I can focus on exactly on the extra detail I want from a larger run:
> 
> driver.navigate.to url_for('xhtmlTest.html')
> WebDriver.logger.level = 0
> driver.title
> WebDriver.logger.level = 2

You can do that now :)

> Also #at_exit call for logger can result in race conditions like this:

Yeah, this is fixed now.

p0deje:
Do you want to use custom formatting? Right now it's informative, but maybe excessive:

```
I, [2017-02-17T10:05:56.548406 #40369]  INFO -- Selenium: message
```

Explanation of format is [here](https://ruby-doc.org/stdlib-2.1.0/libdoc/logger/rdoc/Logger.html#class-Logger-label-Format).

luke-hill:
Just something I thought of yesterday (And probably not for this PR)

Because what is being communicated in $DEBUG can be quite different to physically logged messages, do you think there would be a way (Probably not easy), to pipe the `$DEBUG` messages to one logfile and all the other regular debug/info/warn/error/fatal/unknowns to another file

p0deje:
Only Selenium `$DEBUG` messages are used by Logger. If there is Grid or any other process that prints something to stdout on `$DEBUG` - there is not much we can do about it.

luke-hill:
So in essence having one consumption of the Logger class already running and an additional Selenium::Logger (I know they consume the same thing), in theory is possible?

That would be good, because then we can pipe them to different files.

p0deje:
Not sure I understand your question correctly. Different loggers are different objects, they don't consume the same thing, so yeah, you can pipe them to different files.

titusfortner:
Yes, this does the things I'm excited about! :)

I wrote [a few specs](https://gist.github.com/titusfortner/d432695ad62ed4810663ce26b238668f
), I probably need to come up with more to make sure I'm thinking through the use cases.

One departure from status quo, which I think I'm ok with...
Previously, you could set `$DEBUG = true` before the line you wanted. Now, unless it is set before the driver is instantiated, the Logger will swallow this output. So, we'll just have to specify that if all of the `$DEBUG` output is desired, it needs to go at the beginning of your code, and for more fine-tuned Selenium-specific data, you need to use Logger.

As for what I was saying about info vs debug levels...
For instance on this line[https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/server.rb#L181]
doing `@log = opts.fetch(:log) { WebDriver.logger.debug? }`
And as for the browsers, I was thinking about geckodriver's implementation specifically since that's the one I've been using most for logging errors, and I don't think there is currently an option for outputting to a file... This implementation probably needs to wait until we implement the `FirefoxOptions` class. (which I'm thinking we might try to do at the Selenium workshop since it should be relatively straightforward and show off the code base decently well).

In thinking through use cases for the logs... if someone tells me that their code isn't working and I ask them to turn the logger on and send me the result, most likely I just want to see what's actually sent to and from the server/driver. Socket & Port details probably don't matter to me... But, information that isn't currently getting logged like the details of what is getting started with `ChildProcess.build(*args)` might be useful in that context. Do you have any feelings on what should be in info vs debug? I'll ask in Slack channel as well.

As for passing IO object to ChildProcess, I can test this on Windows and see what it looks like. I'm going to need to do a lot of testing on all the platforms before merging #3490 anyway.

I think a shorter default format is good, and we can easily enough allow a user to override it.

I think I like how Mozilla does it with timestamps / app / level / message:
```
1487357647158	mozprofile::profile	INFO	Using profile path <...> 
1487357647162	geckodriver::marionette	INFO	Starting browser 
1487357647174	geckodriver::marionette	INFO	Connecting to Marionette on <...>
1487357649125	Marionette	INFO	Listening on port 49666
```

p0deje:
> I wrote a few specs, I probably need to come up with more to make sure I'm thinking through the use cases.

Thanks, I'll wrap them up.

> Previously, you could set $DEBUG = true before the line you wanted. Now, unless it is set before the driver is instantiated, the Logger will swallow this output. So, we'll just have to specify that if all of the $DEBUG output is desired, it needs to go at the beginning of your code, and for more fine-tuned Selenium-specific data, you need to use Logger.

I don't understand what changed here because if you're talking about logging for main process, you can put `Selenium::WebDriver.logger = :debug` before the line you want and it would work like before. If you're talking about logging childprocesses, then you should see `$DEBUG` before starting the process anyway, otherwise it won't work.

> In thinking through use cases for the logs... if someone tells me that their code isn't working and I ask them to turn the logger on and send me the result, most likely I just want to see what's actually sent to and from the server/driver. Socket & Port details probably don't matter to me... But, information that isn't currently getting logged like the details of what is getting started with ChildProcess.build(*args) might be useful in that context. Do you have any feelings on what should be in info vs debug? I'll ask in Slack channel as well.

I'd say we should get started with this and then add more logging in places we need.

> As for passing IO object to ChildProcess, I can test this on Windows and see what it looks like. I'm going to need to do a lot of testing on all the platforms before merging #3490 anyway.

I'd say it might introduce some flakiness (e.g. stale file descriptors), but we can give it a try.

> I think a shorter default format is good, and we can easily enough allow a user to override it.

What do you think about this format:

```ruby
Selenium::WebDriver.logger.warn "example message"
#=> 2017-02-20 12:07:58 WARN Selenium example message
```

titusfortner:
@lmtierney ran into an issue with rspec 3.5. That line should include both ~> 3.0 and <3.5

Also do we want to add formatter to list of delegated methods for users to override default?

Either way, this code is ready to merge, we can make tweaks later, like you said.

p0deje:
> @lmtierney ran into an issue with rspec 3.5. That line should include both ~> 3.0 and <3.5

What was the issue? RSpec 3.5 seems to work fine on Travis and locally.

lmtierney:
Unless the issue before was fixed now in a new RSpec release? @titusfortner do you remember what the issue was? I forget. @p0deje did all the specs run clean with 3.5?

lmtierney:
I think this was one of the test runs that had an issue with 3.5 https://travis-ci.org/SeleniumHQ/selenium/jobs/188438199

lmtierney:
Yeah seems fine now, not sure what was wrong before. I'd say the change is fine then

titusfortner:
Lgtm then. :-)

p0deje:
@titusfortner Waiting for you to approve the PR.

p0deje:
Merged.

p0deje:
I've also add section to wiki about logging: https://github.com/SeleniumHQ/selenium/wiki/Ruby-Bindings#logging

rosskevin:
I'm trying to capture the `console.debug` and `console.info` from chrome, but `Selenium::WebDriver.logger.level = :debug` seems to only affect selenium logging itself.  The docs section doesn't mention altering specific loggers.

How do adjust the `level` of logs from chrome (only)?  I tried the following with no luck:

```ruby 
Capybara.register_driver $default_driver do |app|
  Capybara::Selenium::Driver.new(app, {
    browser: :remote,
    url: 'http://127.0.0.1:4444/wd/hub',
    desired_capabilities: Selenium::WebDriver::Remote::Capabilities.chrome
  })
end

Selenium::WebDriver::Driver.for(:remote).logger.level = :debug
```

PR #3481, opened 2017-02-07
Fix erro in webelemet not created, verifing if is str or dict.

reinaldorossetti:

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/3481)


reinaldorossetti:
Fix erro in webelemet not created, verifing if is str or dict.

py/selenium/webdriver/support/expected_conditions.py
```diff
@@ -95,6 +95,8 @@ def __call__(self, ignored):
 
 
 def _element_if_visible(element, visibility=True):
+    if isinstance(element, str) or isinstance(element, dict):
+           raise StaleElementReferenceException("Invalid locator")
```
Dude-X: This is not the right exception class. There is an InvalidSelectorException class that covers bad locators.

Dude-X:
I don't think this is the right level to put a type check; what problem are you trying to solve?
Also I wouldn't check on particular types, I would check if the variable element is not of instance WebElement.

py/selenium/webdriver/support/expected_conditions.py
reinaldorossetti: I have problem because element not is webelement is str in firefox e dict in chrome.

reinaldorossetti: <selenium.webdriver.remote.webelement.WebElement (session="49274937103ea6cd11bc1c9a3a4784ee", element="0.9158218795761794-1")>
<selenium.webdriver.remote.webelement.WebElement (session="49274937103ea6cd11bc1c9a3a4784ee", element="0.9158218795761794-2")>
  File "D:\Dropbox\Dropbox\Meu Curso de Selenium\testes\loading_page.py", line 30, in <module>
    test = wait_element_not_present(driver, ".loadingoverlay")
  File "D:\Dropbox\Dropbox\Meu Curso de Selenium\testes\loading_page.py", line 20, in wait_element_not_present
    expected_conditions.invisibility_of_element_located((By.CSS_SELECTOR, elem))
  File "C:\Python35-32\lib\site-packages\selenium\webdriver\support\wait.py", line 71, in until
    value = method(self._driver)
  File "C:\Python35-32\lib\site-packages\selenium\webdriver\support\expected_conditions.py", line 198, in __call__
    return _element_if_visible(_find_element(driver, self.locator), False)
  File "C:\Python35-32\lib\site-packages\selenium\webdriver\support\expected_conditions.py", line 100, in _element_if_visible
    return element if element.is_displayed() == visibility else False
AttributeError: 'dict' object has no attribute 'is_displayed'

Dude-X: How is that happening? That function is a 'private' function that should only be called when there's an element to examine from the public expected condition functions.

reinaldorossetti: This InvalidSelectorException not work because function need return true, do you seem:

class invisibility_of_element_located(object):
    """ An Expectation for checking that an element is either invisible or not
    present on the DOM.

    locator used to find the element
    """
    def __init__(self, locator):
        self.locator = locator

    def __call__(self, driver):
        try:
            return _element_if_visible(_find_element(driver, self.locator), False)
        except (NoSuchElementException, StaleElementReferenceException):
            # In the case of NoSuchElement, returns true because the element is
            # not present in DOM. The try block checks if the element is present
            # but is invisible.
            # In the case of StaleElementReference, returns true because stale
            # element reference implies that element is no longer visible.
            return True

Dude-X: You wrote `expected_conditions.invisibility_of_element_located((By.CSS_SELECTOR, elem))`
Is `elem` a string? 
If it's an element you should use `visibility_of(element)` with `until_not` or `staleness_of`.

reinaldorossetti: I don't know because it's happening, but happen. This solution solve my problem.

reinaldorossetti: from selenium import webdriver
from selenium.webdriver.support.ui import WebDriverWait
from selenium.webdriver.support import expected_conditions
from selenium.common.exceptions import ElementNotVisibleException

driver = webdriver.Chrome()
driver.get("http://www.jqueryrain.com/?eKV1bCMa")

def wait_element_present(driver, elem):
    present = WebDriverWait(driver, 60).until(
        expected_conditions.presence_of_element_located((By.CSS_SELECTOR, elem))
    )
    print(present)
    return present

def wait_element_not_present(driver, elem):
    present = WebDriverWait(driver, 60).until(
        expected_conditions.invisibility_of_element_located((By.CSS_SELECTOR, elem))
    )
    print(present)
    return present

driver.switch_to.frame(0)
button = wait_element_present(driver, "button#quick_demo")
button.click()

test = wait_element_present(driver, ".loadingoverlay")
test = wait_element_not_present(driver, ".loadingoverlay")

Dude-X: I'm sorry, but this pull request will most likely won't be accepted because it doesn't solve the problem for everyone. I suspect that you may have a bug in your code where the `elem` is an actual element and not a `str`ing that represents a CSS locator. If there is a Dict or String being produced by Firefox and Chrome you will have to file a bug report, since WebDriver is only suppose to produce WebElements. The only exception to this is if you extract data via a javascript call (`execute_script`).

You should not ignore exceptional cases too, that is most likely a bug in your code.

reinaldorossetti: It is string.

reinaldorossetti: How do you test if I'm wrong or right? How should I make the above code work?

Dude-X: I've tested the code you wrote on a simple web page. What you wrote seems like it should work, except in the case of waiting for an element to go 'invisible' you should not care about the result (The result will be True for invisibility). Maybe you should go to the #selenium chat room on freenode IRC and ask for help there.

reinaldorossetti: This problem only happen in invisibility_of_element_located, no have problem that return true or false, but  the problem is that it throws an exception in the code and stop my test.

reinaldorossetti: How do I know if the problem came from chrome or firefox? How do I test this?

Dude-X:
@AutomatedTester I don't think this should have been merged. Clearly the function expects a web element type to have the is displayed function and it throws an exception. Either we add type checking for all expected condition functions for consistency and safety, or let python do its thing. 

That being said I have seen cases were Chrome or Firefox does not return a proper WebElement object but that's probably a bug in the driver.

Issue #2478, opened 2016-07-16
Ruby Capabilities


titusfortner:
Might split this up into multiple issues, but wanted to have a discussion with the Ruby community on this first:
1. Each Bridge class has duplicated code for creating capabilities that can be DRY'd up.
2. I'm not sure why Capabilities is name spaced under Remote, because we use them locally as well.
3. Ruby has a weird mix of how it handles capabilities. Some values are explicitly supported and are accessed by symbol, and some values are allowed by default and are accessed by strings, which leads to confusion. Perhaps we support these explicitly?
4. Ruby is supporting a number of different ChromeOption Capabilities that no longer exist
5. Some capabilities are allowed to be parameters when initializing the Bridge, and others are required to be passed in through the DesiredCapabilities object. One way or another, we should make this consistent
6. I also find it weird that we allow (some) values to be accessed by both `caps.foo` and `caps[:foo]` but there might not be a compelling reason to change that.

llaskin:
In minor relation to this bug I have made https://github.com/SeleniumHQ/selenium/pull/3279 which should take care of (3), I think?

luke-hill:
Note on this - From a Ruby Perspective. There is (I believe), issues with some of the ie capabilities which have a `.` in there name. such as `ie.ensureCleanSession`. This is hard to symbolise.

1) Is there a reason this is written as an ie parameter with ie in its name. Wouldn't it be better to pass a warning across along the lines of, capabilities passed in will not be used.

2) If you pass in wrong capabilities, there doesn't seem to be any catchers at the moment, I've tried passing in `caps['hellothere'] = true`, and this seems to work.

Wrt the Bridge class, I did notice some duplication, and mentioned this on Slack, but I would question why the js/css capabilities aren't default true for every class. In my eyes that would be better, still leaving the option to turn them off, but make them all default on.

titusfortner:
As I'm rewriting these, I'm starting with what is accepted as a parameter when initializing the driver in the first place.

When doing:
```
Selenium::WebDriver::Driver.for browser_name, opt
```

The allowed values for keys in the `opt` are:
Any value passed in that is not one of these is going to throw a deprecation notice:
1. `:url`
2. `:port`
3. `:service_args`
4. `:driver_path`
5. `:http_client`
6. `:required_capabilities`
7. `:desired_capabilities`
8. `:options`
9. `:profile`
10. `:marionette`

[...]

PR #3490, opened 2017-02-09
[rb] Update driver creation and capability processing

titusfortner:
Addresses #2478

Existing code does extra processing for desired capabilities when run locally that doesn't get applied when running via remote/server. This code standardizes what is allowed to be passed in with driver initialization and makes deprecation notices for things that will no longer be supported.

Much of existing code in the Bridge subclasses is duplicated in all of the subclasses, or in need of being moved to Capabilities class. This code initializes in the superclass using methods describing the 3 things it is responsible for: `#process_capabilities`,  `#create_http_client` and `#create_session`. The only things needing to be relegated to the subclasses are processing service arguments and referencing the applicable Service and Capabilities.

It is definitely more DRY, hopefully it is also easier to understand what it is doing.

@lmtierney & @twalpole let me know if you have any feedback on this approach or implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/3490)


rb/lib/selenium/webdriver/firefox/bridge.rb
```diff
@@ -21,49 +21,36 @@ module Selenium
...
+        def start_service(opts)
+          if opts[:desired_capabilities] && opts[:desired_capabilities][:firefox_binary]
+            Binary.path = opts[:desired_capabilities][:firefox_binary]
+          end
 
-        def create_launcher(port, profile)
-          Launcher.new Binary.new, port, profile
+          profile = opts.delete(:profile) || opts[:desired_capabilities].firefox_profile
+          port = opts.delete(:port) || DEFAULT_PORT
+          Launcher.new(Binary.new, port, profile).tap { |l| l.launch }
```
p0deje: `.tap(&:launch)`

rb/lib/selenium/webdriver/remote/bridge.rb
```diff
@@ -605,11 +552,58 @@ def find_elements_by(how, what, parent = nil)
...
+
+          http_client.server_url = uri
+          http_client
+        end
+
+        def start_service(opts)
+          path = opts.delete(:driver_path) || bridge_module.driver_path
+          port = opts.delete(:port) || bridge_module::Service::DEFAULT_PORT
+          service_args = process_service_args(opts.delete(:service_args))
+          bridge_module::Service.new(path, port, *service_args).tap { |s| s.start }
```
p0deje: `.tap(&:start)`

rb/lib/selenium/webdriver/safari.rb
```diff
@@ -24,6 +24,14 @@ module Selenium
   module WebDriver
     module Safari
       class << self
+        def technology_preview
+          "/Applications/Safari\ Technology\ Preview.app/Contents/MacOS/safaridriver"
+        end
+
+        def technology_preview=(value = true)
```
p0deje: You shouldn't be using default argument for setter, it should be used like this:

```ruby
Safari.technology_preview=
```

Maybe it's then better to use bang-method?

p0deje (approved):
1. I've left some comments.
2. There are quite a lot of changes in Bridge classes which look a bit unrelated to this PR (e.g. renaming of commands) and I don't really understand those. It's hard for me to review them.
3. I haven't reviewed the code line-by-line because there are too many changes, so I'll just trust you made sure nothing is broken :)

Overall, I really liked what you did here, so let's bring it  in point release.

👏

lmtierney:
👍

rb/lib/selenium/webdriver/chrome/bridge.rb
```diff
@@ -21,88 +21,63 @@ module Selenium
...
-          caps[:proxy] = opts.delete(:proxy) if opts.key?(:proxy)
-          caps[:proxy] ||= opts.delete('proxy') if opts.key?('proxy')
-
-          caps
-        end
-
-        def extract_service_args(args)
-          service_args = []
-          service_args << "--log-path=#{args.delete(:service_log_path)}" if args.key?(:service_log_path)
-          service_args << "--url-base=#{args.delete(:url_base)}" if args.key?(:url_base)
```
titusfortner: The method was renamed to `#process_service_args`

rb/lib/selenium/webdriver.rb
```diff
@@ -57,8 +57,10 @@ def self.root
     #
     # @return [Driver]
     #
-    # @see Selenium::WebDriver::Remote::Bridge
+    # @see Selenium::WebDriver::Remote::OSSBridge
```
titusfortner: That would have been my name for it as well, but I copied how Simon named his subclassing in Java. "OSS" and "W3C."

rb/lib/selenium/webdriver/common/service.rb
```diff
@@ -87,6 +86,20 @@ def uri
...
+
+      def start_process
+        server_command = [@executable_path, "--port=#{@port}", *@extra_args]
+        @process       = ChildProcess.build(*server_command.compact)
+
+        if $DEBUG
+          @process.io.inherit!
+          puts "Process Starting: #{server_command}"
+        end
+        @process.leader = true
```
titusfortner: It applied to all but 2 drivers, so I made it default and overrode it in the subclass for the two applicable `Service` classes. Just trying to reduce duplicated code. I might have gone too far, so I'm ok to put it back if you think it would be more clear.

rb/lib/selenium/webdriver/edge.rb
```diff
@@ -26,23 +26,11 @@ module Selenium
   module WebDriver
     module Edge
       def self.driver_path=(path)
-        warn <<-DEPRECATE.gsub(/\n +| {2,}/, ' ').freeze
```
titusfortner: You changed my mind. Made sense to be able to set it once in a `spec_helper.rb` rather than have to pass it in everywhere in the code. Yes, that's how Java does it, but Ruby has more flexibility, so I think we should allow it. Are you disagreeing now? :)

lmtierney: Now there's the best of both worlds, he just removed the deprecation warning and kept both ways of specifying the `driver_path`

titusfortner:
@p0deje thanks for suggestions, I'll update the code. 

Yes, I probably went a step further than I needed to in subclassing Bridge & Service classes. At the point I was in the code and removing duplication I thought I'd keep going.

As for the new method names, firstly they are all private API, and secondly, I thought I should rename the commands to make it more obvious the intention behind things in `#initialize` since I moved everything to one place. Current implementation just does things without being explicit about what it is doing. For instance when running via a local `Bridge`, `#create_http_client` requires starting the driver, etc where for Remote it doesn't, but it is still performing that common purpose of setting up the network client with the correct url.

Also, sigh; trust is dangerous. :) I wrote new specs and they are all passing, but I suspect I'll get to field a number of "you broke this" issues for weird combinations of things I didn't consider after this is released.

We're having our monthly test coding meetup in Austin on Monday. I'm planning on presenting this and getting more feedback and will likely push this in for the 3.1 release.

rb/lib/selenium/webdriver/chrome/bridge.rb
```diff
@@ -21,88 +21,63 @@ module Selenium
...
-        private
-
-        def create_capabilities(opts)
-          caps = opts.delete(:desired_capabilities) { Remote::Capabilities.chrome }
+        def process_service_args(service_opts)
+          return [] unless service_opts
+          return service_opts if service_opts.is_a? Array
 
-          chrome_options = caps['chromeOptions'] || caps[:chrome_options] || {}
-          chrome_options['binary'] = Chrome.path if Chrome.path
```
lmtierney: `chrome_options['binary']` is no longer being set anywhere, so setting `Chrome.path=` does nothing. Might add it to `Capabilities`

rb/lib/selenium/webdriver/remote/capabilities.rb
```diff
@@ -60,6 +60,19 @@ class Capabilities
...
+            define_method(:options) { @capabilities[:chrome_options] ||= {} }
+            define_method("options=") { |value| @capabilities[:chrome_options] = value }
+            define_method("profile=") do |profile|
+              profile_json = profile.as_json
+              options['args'] ||= []
+              if options['args'].none? { |arg| arg =~ /user-data-dir/ }
+                options['args'] << "--user-data-dir=#{profile_json[:directory]}"
+              end
+              options['extensions'] = profile_json[:extensions]
+            end
```
lmtierney: `options['binary'] = WebDriver::Chrome.path` if set?

rb/lib/selenium/webdriver/remote/bridge.rb
```diff
@@ -605,11 +552,58 @@ def find_elements_by(how, what, parent = nil)
...
-        def assert_javascript_enabled
-          return if capabilities.javascript_enabled?
-          raise Error::UnsupportedOperationError, 'underlying webdriver instance does not support javascript'
+        def bridge_module
+          Module.nesting[1]
+        end
+
+        def process_capabilities(opts)
+          capabilities = opts.delete(:desired_capabilities) || default_capabilities
+          capabilities = Capabilities.send(capabilities) if capabilities.is_a?(Symbol)
```
lmtierney: Should we merge passed in `desired_capabilities` to the defaults for that browser? We had to do this for firefox currently so that `browser_name` was picked up (geckodriver needed it).

lmtierney: It'd be even better if we could utilize the `:options` and `:profile` processing of the `Capabilities.firefox`, etc but I'm not sure how to make that elegant

rb/lib/selenium/webdriver/chrome/bridge.rb
titusfortner: Good catch. This is only going to apply to running local not remote, so I think it belongs in the subclass. I'll update these.

rb/lib/selenium/webdriver/remote/bridge.rb
titusfortner: I'm not sure how you would create a `Capabilities` class without making use of the browser's defaults, though. I figure if the user is passing in a value, that's the value they want and we shouldn't presume to add things to it. It's not like creating it isn't trivial.

I'm not sure how things need to be different for Firefox? From an end user standpoint, `Remote#self.w3c?` toggles whether to use `Bridge` or `W3CBridge`, which should be sufficient.

Do you have an example of something that doesn't behave as desired now?

lmtierney: We merge with defaults right now for firefox (w3c bridge) because someone in #watir was passing in a hash for desired capabilities (without `browser_name`) instead of a `Capabilities` class and apparently now geckodriver needs `browser_name` set.

That being said, I agree it would be a nice option for people to specify capabilities as a hash without merging defaults so nevermind on that one. 

The second comment was an example, not saying `firefox` specifically. I was more referring to the processing of `:options` and `:profile` in the `Capabilities` class itself instead of doing in the `Capabilities` class and then in `process_capabilities` [here](https://github.com/SeleniumHQ/selenium/pull/3490/files/7fee63589f33dab25d3d52bd27f0694c10388a80#diff-dd335728ad314dd52eece6507da94ea5R575) but I don't have a solution for that more just bringing it up as an idea

titusfortner: Ah, the `Firefox::Bridge#process_capabilities` method is intended to handle the things that are local only, and the `Capabilities#self.firefox` method is intended to handle the things that are applicable regardless of local or remote.

Before this PR is merged, I'm planning on revamping Watir browser configurations. I'm going to be ok with telling Watir users that if they want to use selenium-webdriver 3.1.x then they need to use watir 6.4 or whatever, and change their parameters as necessary.

titusfortner: Ok, so the use case is that someone does this:
```
caps = Selenium::WebDriver::Remote::Capabilities.firefox
driver = Selenium::WebDriver.for :remote, desired_capabilities: caps
```
Right now it is using defaults for `Selenium::WebDriver::Remote::Capabilities` instead of defaults for `Selenium::WebDriver::Remote::W3CCapabilities`. I'll go ahead and add this to the top of `Selenium::WebDriver::Remote::Capabilities#self.firefox`:
```
return Remote::W3CCapabilities.firefox(opts) if Remote::W3CCapabilities.w3c?
```
That should address the user-facing concerns I think.

PR #3515, opened 2017-02-15
Remove resolve_ip feature from RemoteConnection

### LINKED ISSUE #1729: Python Selenium Client is by default resolving ip adresses and not sending domain names
## Expected Behavior -

If the configured endpoint is app.testobject.com the HTTP request should got to app.testobject.com and not the ip address behind app.testobject.com
## Actual Behavior -

when setting the selenium endpoint to a domain the outgoing http request is going to the ip adress. That confuses WebServers which are proxying commands to the selenium backend, as they usually use domain names and not ip addresses to find the right config.
## Steps to reproduce -

Initiate a session to an non local endpoint (e.g. one of the cloud testing providers like sauce and browserstack). Record the http requests using Charles Proxy (or Wireshark). The request is send to an ip instead to the domain.

The problem can be found here:

https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/remote_connection.py#L165

I would recommend to change the default behaviour to false.

### LINKED ISSUE #2509: python client bindings do not support remote https selenium grid endpoints.
## Meta -
OS:  OSX
Selenium Version:  2.53.6 on Python 2.7.12
Browser:  NA
Browser Version:  NA

## Expected Behavior -
When specified a https grid location like `https://ondemand.saucelabs.com/wd/hub` user should be able to run tests the same as the http endpoint.
## Actual Behavior -
The python client throws the following error.
```
>               % (hostname, ', '.join(map(repr, dnsnames))))
E           CertificateError: hostname '66.201.41.7' doesn't match either of '*.browserstack.com', 'browserstack.com'

/usr/local/Cellar/python/2.7.12/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ssl.py:272: CertificateError
```
## Steps to reproduce -
Attempt to start any session using the `https://ondemand.saucelabs.com/wd/hub` url for command executor.

### LINKED ISSUE #3514: Python: RemoteConnection with keepalive ignores HTTPS
## Meta -
OS:  All
Selenium Version:  3.0.2
Browser: Remote
Browser Version: N/A

## Expected Behavior -
`RemoteConnection('https://foo.bar/', keep_alive=True)` sends all its commands over a secure HTTPS connection.

## Actual Behavior -
`RemoteConnection('https://foo.bar/', keep_alive=True)` sends all its commands over an insecure HTTP connection.

## Steps to reproduce -
Use `RemoteConnection('https://foo.bar/', keep_alive=True)` to connect to a secured selenium server.

kvdveer:
Resolving the IP never necessary, httplib does that for us. Worse, it is always the wrong thing to do when dealing with HTTPS connections, as no certificates are issued for IP-adresses. IMHO there is no reason to keep this feature around.

While I was at it, I also added HTTPS support for keepalive connections; which always defaulted to HTTP, even if a HTTPS url was provided.

Fixes #1729
Fixes #2509
Fixes #3514

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/3515)

### DIFF
--- a/py/selenium/webdriver/edge/webdriver.py
+++ b/py/selenium/webdriver/edge/webdriver.py
@@ -38,8 +38,7 @@ class WebDriver(RemoteWebDriver):
 
         RemoteWebDriver.__init__(
             self,
-            command_executor=RemoteConnection('http://localhost:%d' % self.port,
-                                              resolve_ip=False),
+            command_executor=RemoteConnection('http://localhost:%d' % self.port),
             desired_capabilities=capabilities)
         self._is_remote = False
 
--- a/py/selenium/webdriver/remote/remote_connection.py
+++ b/py/selenium/webdriver/remote/remote_connection.py
@@ -29,7 +29,6 @@ except ImportError:  # above is available in py3+, below is py2.7
     import urllib2 as url_request
     import urlparse as parse
 
-from selenium.webdriver.common import utils as common_utils
 from .command import Command
 from .errorhandler import ErrorCode
 from . import utils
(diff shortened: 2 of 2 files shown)


Dude-X:
Mind fixing the merge conflict?

lmtierney:
@AutomatedTester is there historically a reason why the IP was resolved?

GQAssurance:
#7430 implemented these changes and was merged; this PR can be closed.

PR #3532, opened 2017-02-18
Add expected condition that waits for all found elements to be visible

BeyondEvil:

### DIFF
--- a/py/selenium/webdriver/support/expected_conditions.py
+++ b/py/selenium/webdriver/support/expected_conditions.py
@@ -20,7 +20,6 @@ from selenium.common.exceptions import NoSuchFrameException
 from selenium.common.exceptions import StaleElementReferenceException
 from selenium.common.exceptions import WebDriverException
 from selenium.common.exceptions import NoAlertPresentException
-from selenium.common.exceptions import ElementNotVisibleException
 
 """
  * Canned "Expected Conditions" which are generally useful within webdriver


py/selenium/webdriver/support/expected_conditions.py
```diff
@@ -124,6 +125,27 @@ def __call__(self, driver):
...
+    returns the list of WebElements once they are located and visible
+    """
+    def __init__(self, locator):
+        self.locator = locator
+
+    def __call__(self, driver):
+        try:
+            elements = _find_elements(driver, self.locator)
+            for element in elements:
+                if not _element_if_visible(element):
```
Dude-X: I would write this as:
```if  _element_if_visible(element, visibility=False):
    return False
```
instead of raising an exception.

BeyondEvil: Thanks, @Dude-X, that makes sense! I've changed it locally. Can I merge it with this PR? If so, how?

cgoldberg: @BeyondEvil  just push your changes to the same remote branch again... the PR will get updated automagically.

cgoldberg:
@BeyondEvil you should rebase before pushing so your PR doesn't contain unrelated commits

BeyondEvil:
@cgoldberg Yeah, I don't know what I'm doing lol ^^ But I got some help from the kind folks in IRC to fix it :)

py/selenium/webdriver/support/expected_conditions.py
```diff
@@ -20,6 +20,7 @@
 from selenium.common.exceptions import StaleElementReferenceException
 from selenium.common.exceptions import WebDriverException
 from selenium.common.exceptions import NoAlertPresentException
+from selenium.common.exceptions import ElementNotVisibleException
```
Dude-X: Is this an oversight for some code in the body?

BeyondEvil: @Dude-X Yes, I'll remove that (it was used by the first revision of this PR).

PR #3536, opened 2017-02-19
Chrome (so HtmlUnit) does not trigger alerts on unload

asashour:

### DIFF
--- a/java/client/test/org/openqa/selenium/AlertsTest.java
+++ b/java/client/test/org/openqa/selenium/AlertsTest.java
@@ -34,6 +34,7 @@ import static org.openqa.selenium.testing.Driver.MARIONETTE;
 import static org.openqa.selenium.testing.Driver.PHANTOMJS;
 import static org.openqa.selenium.testing.Driver.SAFARI;
 import static org.openqa.selenium.testing.TestUtilities.getFirefoxVersion;
+import static org.openqa.selenium.testing.TestUtilities.isChrome;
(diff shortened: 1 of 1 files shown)


shs96c:
I'm not sure why this is needed: the `@Ignore` is effectively the same as the assumption you're putting in, and the test is marked as `@NotYetImplemented` for HtmlUnitDriver, and remains so with this change. I'm obviously missing something :)

asashour:
From what I humbly understand, `assumeFalse` is a normal expected result, to say that this test is not meant to be used because of this condition.

However, `@NotYetImplemented`, means it is a bug in the implementation and should be fixed later.

But you are right, I will remove the `@NotYetImplemented` in PR, similar to what was done [here](https://github.com/SeleniumHQ/htmlunit-driver/blob/remote/src/test/java/org/openqa/selenium/AlertsTest.java#L385)

shs96c:
Heh. There's history here!

The `@Ignore` was written before `Assume` was a A Thing in JUnit and allowed us to do much the same thing --- ignore a test completely. The `@NotYetImplemented` was there for tracking work in progress, but we never consistently applied it throughout the tree, relying on `@Ignore` instead. I think you've been the most assiduous about applying the latter annotation correctly :)

You're right, most of the time we can simply replace `@Ignore` with an assumption, particularly when it comes to the type of browser we're using.

shs96c:
Merged and pushed manually.

shs96c:
Thank you!

PR #3554, opened 2017-02-22
[dotnet] Added ability to take array of iframe names for switching

Frannsoft:
Added support for sending in a string array of iframe names to `ITargetLocator` implementations to have the webdriver loop through and switch into as ordered in the parameter.  This makes it easier to switch into nested iframes.

### DIFF
--- a/dotnet/src/support/Events/EventFiringWebDriver.cs
+++ b/dotnet/src/support/Events/EventFiringWebDriver.cs
@@ -955,6 +955,28 @@ namespace OpenQA.Selenium.Support.Events
             }
 
             /// <summary>
+            /// Move to a different frame that is nested in other frames.
+            /// </summary>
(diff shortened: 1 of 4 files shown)


jimevans:
This would make the .NET bindings wildly divergent from all other language bindings the project supports. Moreover, switching frame focus by name is not supported by the W3C WebDriver Specification (only by index and element reference are supported in the spec).

Before I'm willing to extend the API in this way, you'll need to gather a consensus among the other language binding owners that they are willing to add it to other language bindings as well. The easiest way to start this conversation is in the #selenium IRC channel on Freenode. Failing that, you'll need to start a discussion on the [Selenium Developers mailing list](https://groups.google.com/forum/m/#!forum/selenium-developers). You're more likely to get timely feedback from the former rather than the latter. Once you've obtained that consensus, we can revisit this pull rxequest. In the meantime, I'm going to close it.

PR #3583, opened 2017-02-28
Treat object tag with embedded document as frame

laryu:

Add feature to enable switch to <object> element and handle embedded document as inside a frame.

### DIFF
--- a/cpp/iedriver/DocumentHost.cpp
+++ b/cpp/iedriver/DocumentHost.cpp
@@ -137,18 +137,45 @@ int DocumentHost::SetFocusedFrameByElement(IHTMLElement* frame_element) {
     return WD_SUCCESS;
   }
 
-  CComPtr<IHTMLFrameBase2> frame_base;
-  frame_element->QueryInterface<IHTMLFrameBase2>(&frame_base);
-  if (!frame_base) {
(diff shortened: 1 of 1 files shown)


jimevans:
Given that literally **no** other browser driver supports this functionality, and that it does not conform to the W3C WebDriver specification, I'm inclined to reject this PR. Can you make the case for its inclusion when cross-browser compatibility is one of the major primary concerns of the WebDriver project?

laryu:
Based on our testing, both geckodriver and chrome driver support switch to frame operation for document embedded in <object> element. The following is an example modified from frame test file:

```<html>
<head>
    <title>This page has iframes</title>
</head>
<body>
<h1 id="iframe_page_heading">This is the heading</h1>
<object data="formPage.html" height="800" scrolling="no" width="8000" id="iframe1" name="iframe1-name" />
</body>
</html>
```

When test this example in Firefox or Chrome, the corresponding driver will treat object element as frame and allow switch into it to access elements in formPage.html page. 
There is no similar command exists in IE driver. This change will enable IE driver to perform similar actions. 

This issue came up during our cross browser compatibility test. It will make test environment set up much easier if this feature could make into release build.

jimevans:
Merged in 90216e9. Thank you for the contribution.

PR #3584, opened 2017-03-01
[rb] update what gets logged at what level

titusfortner:
Any objections to changing these (less important) commands to `:debug`? This leaves `:info` as just what commands are sent to & from driver/server. I think that this is what is most useful when troubleshooting someone's issue. Unless you think there is anything else that is essential for basic troubleshooting? 

Also this adds details about what `ChildProcess` is doing.

Another thought. Does it make sense to change our `Kernel#warn` instances to `WebDriver.logger.warn`, since by default `Logger` displays warnings to `STDOUT`?

### DIFF
--- a/rb/lib/selenium/webdriver/remote/bridge.rb
+++ b/rb/lib/selenium/webdriver/remote/bridge.rb
@@ -40,7 +40,7 @@ module Selenium
           opts = opts.dup
 
           if opts.key?(:port)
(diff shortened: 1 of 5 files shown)


titusfortner:
@p0deje - added 2nd commit with changing `Kernel#warn` to `Logger#warn` - this is what you were suggesting, correct?

p0deje:
Looks good!

titusfortner:
Merged

PR #3585, opened 2017-03-01
[rb] move driver related functionality to Service classes

titusfortner:
This moves extraction of service arguments out of Bridge classes and into Service classes where they belong.

Additionally, in an attempt to better control / make more uniform what gets passed in directly to driver initialization, this creates `#driver_opts` for specifying switches by hash instead of an array of strings.

This code is backward compatible, but throws warnings for the outdated means of interacting with the driver. I contemplated requiring `:port` & `:driver_path` parameters to also go into the `:driver_opts` hash, but since every `Bridge` accepts these, I've leaned toward it not being necessary to deprecate current behavior.

### DIFF
--- a/rb/lib/selenium/webdriver/chrome/bridge.rb
+++ b/rb/lib/selenium/webdriver/chrome/bridge.rb
@@ -26,15 +26,25 @@ module Selenium
           opts[:desired_capabilities] = create_capabilities(opts)
 
           unless opts.key?(:url)
+            driver_path = opts.delete(:driver_path) || Chrome.driver_path
             port = opts.delete(:port) || Service::DEFAULT_PORT
-            service_args = opts.delete(:service_args) || {}
 
-            if opts[:service_log_path]
-              service_args.merge!(service_log_path: opts.delete(:service_log_path))
+            opts[:driver_opts] ||= {}
+            if opts.key? :service_log_path
+              warn <<-DEPRECATE.gsub(/\n +| {2,}/, ' ').freeze
+            [DEPRECATION] `:service_log_path` is deprecated. Use `driver_opts: {log_path: #{opts[:service_log_path]}}`
+              DEPRECATE
+              opts[:driver_opts][:log_path] = opts.delete :service_log_path
             end
 
-            driver_path = opts.delete(:driver_path) || Chrome.driver_path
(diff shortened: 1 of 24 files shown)


rb/lib/selenium/webdriver/chrome/bridge.rb
```diff
@@ -26,15 +26,25 @@ def initialize(opts = {})
...
           unless opts.key?(:url)
+            driver_path = opts.delete(:driver_path) || Chrome.driver_path
             port = opts.delete(:port) || Service::DEFAULT_PORT
-            service_args = opts.delete(:service_args) || {}
 
-            if opts[:service_log_path]
-              service_args.merge!(service_log_path: opts.delete(:service_log_path))
+            opts[:driver_opts] ||= {}
+            if opts.key? :service_log_path
+              warn <<-DEPRECATE.gsub(/\n +| {2,}/, ' ').freeze
```
p0deje: Why not use `WebDriver.logger.warn`?

rb/spec/unit/selenium/webdriver/chrome/service_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+      describe Service do
+        let(:resp) { {'sessionId' => 'foo', 'value' => @default_capabilities} }
+        let(:service) { double(Service, start: true, uri: 'http://example.com') }
+        let(:caps) { {} }
+        let(:http) { double(Remote::Http::Default, call: resp).as_null_object }
+
+        before do
+          @default_capabilities = Remote::Capabilities.chrome.as_json
+          allow(Remote::Capabilities).to receive(:chrome).and_return(caps)
+          allow_any_instance_of(Service).to receive(:start).and_return nil
```
p0deje: You can omit .and_return nil

rb/lib/selenium/webdriver/chrome/bridge.rb
titusfortner: Heh, yes, I asked if that's the way we should be doing it in #3584. I only did it this way because that's how it's [currently being done](https://github.com/SeleniumHQ/selenium/search?l=Ruby&q=warn). So we should probably update all of these as well.

titusfortner:
merged

PR #3650, opened 2017-03-14
base64 str encoding failing in python 3.6

jayruffing:
python3 is expecting a “bytes-like object” (https://docs.python.org/3/glossary.html#term-bytes-like-object)

so it is failing when sending a string (username & password)

see https://docs.python.org/3/library/base64.html#base64.b64encode

to convert to a bytes object, just call encode() method to turn it into a set of 8-bit byte values


davehunt:
This was already fixed in https://github.com/jayruffing/selenium/commit/55ed929d64e96c518c89244dd4a6fccc85865fc0 but I suspect is yet to be released.

PR #3656, opened 2017-03-15
Fixed handling of <content> outside a Shadow DOM

mcharsley:

It turns out that some folks are putting stuff inside a content tag and
expecting it to be displayed if the tag is outside a shadow DOM (and
replaced if it's inside a shadow DOM).

While investigating this, I discovered that we need to support Shadow
DOM v1 as well the (deprecated) Shadow DOM v0 API that we currently use.
Expect further patches...


mcharsley:
Alas I'm a beginner when it comes to git as well. Any colleagues you can
ask?

PR #3663, opened 2017-03-15
String decoding in remote connection

### LINKED ISSUE #3660: Python - remote connection - can't decode a string in py3
Am getting the following error:

```py
            if parsed_url.username:
                base64string = base64.b64encode('{0.username}:{0.password}'.format(parsed_url).encode())
>               request.add_header('Authorization', 'Basic {}'.format(base64string).decode())
E               AttributeError: 'str' object has no attribute 'decode'

../../../../miniconda3/envs/ap-webqa/lib/python3.6/site-packages/selenium/webdriver/remote/remote_connection.py:480: AttributeError
```

https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/remote_connection.py#L480

I may be missing something, but isn't `'Basic {}'.format(base64string)` going to produce a string, which doesn't have a `decode` method.

@davehunt this was a recent change from you. What am I missing?

(I'm running selenium 3.3.1 with pytest-selenium 1.9.1 on python 3.6)

birdsarah:
Fixes #3660

### DIFF
--- a/py/selenium/webdriver/remote/remote_connection.py
+++ b/py/selenium/webdriver/remote/remote_connection.py
@@ -64,7 +64,6 @@ class Request(url_request.Request):
         return self._method
 
 
-
 class Response(object):
     """
     Represents an HTTP response.
--- a/py/test/unit/selenium/webdriver/remote/test_remote_connection.py
+++ b/py/test/unit/selenium/webdriver/remote/test_remote_connection.py
@@ -77,7 +77,7 @@ def test_remote_connection_adds_connection_headers_from_get_remote_connection_he
         mock_open = mocker.patch('urllib2.OpenerDirector.open')
 
     def assert_header_added(request, timeout):
-        assert request.headers == test_headers  
+        assert request.headers == test_headers
         return MockResponse()
 
     mock_open.side_effect = assert_header_added


birdsarah:
@davehunt, agree this was cumbersome to test.

If my code split out is really undesirable, should it be straight forward to just do the very small fix to the code and rework the tests now I figured out all the mocking.

But, it seems to me that _request needs breaking up into smaller methods to allow for easier unit testing, so I figured I might as well start down that path.

Putting the header addition on the Request class seemed reasonable, but also see the first commit where I just made it a method on RemoteConnection - in case that's preferable.

py/selenium/webdriver/remote/remote_connection.py
```diff
@@ -472,12 +488,7 @@ def _request(self, method, url, body=None):
             else:
                 request = Request(url, data=body.encode('utf-8'), method=method)
 
-            request.add_header('Accept', 'application/json')
-            request.add_header('Content-Type', 'application/json;charset=UTF-8')
-
-            if parsed_url.username:
-                base64string = base64.b64encode('{0.username}:{0.password}'.format(parsed_url).encode())
-                request.add_header('Authorization', 'Basic {}'.format(base64string).decode())
+            request.add_remote_connection_headers(parsed_url)
```
davehunt: I like the idea of splitting this out. We could take it a step further though and remove the duplication from lines 450-458. Perhaps a `get_remote_connection_headers` that takes the URL and a boolean for `keep_alive`? It's also fine if you want to push this to a later date.

py/test/unit/selenium/webdriver/remote/test_remote_connection.py
```diff
@@ -15,21 +15,59 @@
...
+    parsed_url = parse.urlparse(url)
+    cleaned_url = parse.urlunparse((
+        parsed_url.scheme,
+        parsed_url.hostname,
+        parsed_url.path,
+        parsed_url.params,
+        parsed_url.query,
+        parsed_url.fragment)
+    )
+    request = Request(cleaned_url)
```
davehunt: What's the reason for the `cleaned_url`. Can we not just use the original `url` here?

birdsarah:
Sorry for delay - will look at your suggestions this evening..

py/test/unit/selenium/webdriver/remote/test_remote_connection.py
birdsarah: fixed

birdsarah:
Ready! I factored out the other headers declaration so there's no duplication. Please note the following:
* The non keep-alive headers will now also have the user agent `"User Agent": "Python http auth"`
* The keep-alive headers no longer have the header `method: parsed_url.path` which I don't believe is a valid, or used, header value.

davehunt:
Thanks @birdsarah!

PR #3664, opened 2017-03-16
[js] error looking for free port -> process halts

thorn0:
I've gotten `UnhandledPromiseRejectionWarning` a couple of times, and, after investigating, found this.


jleyba:
Good catch!

PR #3679, opened 2017-03-19
Removing unnecessary binary path calculation in toString()

### LINKED ISSUE #3667: executable path of browser converted to unix format and loses the drive name so tests and executable need to be on the same drive
## Meta -

OS:  
Windows 10

Selenium Version:  
3.3.1

Browser:  

Firefox

Browser Version:  

I was using portable firefox version 47.0.1 but this isn't strictly relevant to the bug because the code does not get as far as starting the browser

## Expected Behavior -

When I set the browser binary, I expected it to use the binary no matter which drive I ran my tests from - this was behaviour in previous versions of WebDriver below 3.3.0

~~~~~~~~
        String firefox_binary_path = "C:\\webdrivers\\FirefoxPortable\\FirefoxPortable.exe";

        capabilities.setCapability("firefox_binary", firefox_binary_path);
~~~~~~~~

## Actual Behavior -

In 3.3.1 the `FirefoxOptions` has a method call `asUnixPath` which converts 

`C:\webdrivers\FirefoxPortable\FirefoxPortable.exe`

to 

`/webdrivers/FirefoxPortable/FirefoxPortable.exe`

And while the capability still has the full path. The `.binaryPath` field in `FirefoxOptions` does not have the drive in it.

Therefore, when `FirefoxOptions` uses the `getBinaryOrNull` method the `binaryPath` field is used and this does not have the path in it so the check for file existence only works if the tests are running on the same drive as the browser executable is located.

## Steps to reproduce -

Run a test which configures the location of the browser location, but have the browser located on a different drive.

## Workaround

Run the tests from the same drive the executable is stored on.

krmahadevan:
Closes #3667

We shouldn’t be triggering binary path validation
in a toString() because sometimes users pack a 
Windows path in a OS X/LINUX machine as firefox_binary
path and create a desired capabilities object.

### DIFF
--- a/java/client/src/org/openqa/selenium/firefox/FirefoxOptions.java
+++ b/java/client/src/org/openqa/selenium/firefox/FirefoxOptions.java
@@ -609,7 +609,7 @@ public class FirefoxOptions {
   @Override
   public String toString() {
     return "{" +
-           "binary=" + getBinaryOrNull() + ", " +
+           "binary=" + this.binaryPath + ", " +
            "args=" + args + ", " +
            "legacy=" + legacy + ", " +
            "logLevel=" + logLevel + ", " +


krmahadevan:
ping @shs96c - Can you please get this merged ?

krmahadevan:
Since @shs96c fixed this via the commit https://github.com/SeleniumHQ/selenium/commit/ed3e946a9ad04299c2a61f4b75ebb60abb76c9c8 I am closing off my Pull Request.

PR #3709, opened 2017-03-24
Add support to long type data in moz:firefoxOptions

eric-houzelle:
If you send a number in firefoxOptions/prefs data object there is an error in FirefoxOptions.java : Exception: Invalid Firefox preference value: network.proxy.ssl_port=8161

Because the value is not an Integer or a String or a Boolean but a Long. So I just add the Long type support for the data in prefs for Firefox.

### DIFF
--- a/java/client/src/org/openqa/selenium/firefox/FirefoxOptions.java
+++ b/java/client/src/org/openqa/selenium/firefox/FirefoxOptions.java
@@ -124,9 +124,11 @@ public class FirefoxOptions {
           options.addPreference(entry.getKey(), (Integer) value);
         } else if (value instanceof String) {
           options.addPreference(entry.getKey(), (String) value);
+        }else if (value instanceof Long) {
+          options.addPreference(entry.getKey(), ((Number)value).intValue());
         } else {
-          throw new WebDriverException(
-              "Invalid Firefox preference value: " + entry.getKey() + "=" + value);
+           throw new WebDriverException(
+               "Invalid Firefox preference value: " + entry.getKey() + "=" + value);
         }
       });
     }


barancev:
We need a test or a scenario to reproduce the issue fixed by this patch.

Also, the patch needs rebasing.

eric-houzelle:
Thanks for your answer,

to reproduce this issue just launch a selenium standalone 3.x and execute this command : 

 curl -XPOST -d '{"desiredCapabilities":{"browserName":"firefox","version":55,"platform":"ANY","javascriptEnabled":true,"proxy":{"proxyType":"manual","httpProxy":"localhost:8108","sslProxy":"localhost:8108"}}}' http://localhost:4444/wd/hub/session


With the proxy parameters there is an error on the data type, without there is no problem.


So it's impossible to use a proxy on a specific port with selenium 3.x . No problem on version 2.x

barancev:
I can't see how is it related to FirefoxOptions class?

eric-houzelle:
The capabilities httpproxy and sslproxy are explode and the port is set as a Firefox pref in the network.proxy.ssl_port variable for example. But the port is set as a Long type which is Not support by Firefox options. So we cant set a proxy.

barancev:
Proxy settings are not preferences, they are not related to the part of FirefoxOptions class you're suggesting to change.

eric-houzelle:
OK i 've investigated this problem in march, so i dont remember all the details...
I Just know this fix resolve this Bug because i use it in a selenium custom build. I don' t have any time to work on it now. If you find another way to resolve this Bug it's ok i dont care.
Or if you have a solution to use a proxy on Firefox with selenium 3 official it's perfect. Because today the Bug exist.

Thank you.

barancev:
I can't reproduce the issue with selenium server 3.5.3
image

barancev:
Closing this PR, the underlying issue seems to be fixed.

PR #3723, opened 2017-03-27
Support W3C implementation of IE driver

p0deje:
@titusfortner @lmtierney @jimevans That's a simple and bad implementation of W3C dialect for IEDriver. For now, this adds a single option:

```ruby
Selenium::WebDriver.for(:ie) # wire protocol
Selenium::WebDriver.for(:ie, w3c: false) # wire protocol
Selenium::WebDriver.for(:ie, w3c: true) # W3C protocol
```

```ruby
# wire protocol
caps = Selenium::WebDriver::Remote::Capabilities.ie
Selenium::WebDriver.for(:remote, desired_capabilities: caps)

# W3C protocol
caps = Selenium::WebDriver::Remote::Capabilities.ie(w3c: true)
Selenium::WebDriver.for(:remote, desired_capabilities: caps)
```

I still can't make specs pass for W3C (but I also can't make them pass for wire) and I don't quite like the code, but assuming we want to rewrite bridge/caps into a single implementation which understands what dialect it needs to talk to a driver, we can use this code for now and get rid of it once refactoring is done. The remote is also not working yet - I'll fix that tomorrow. I just wanted to get your thoughts on this.

### DIFF
--- a/rb/build.desc
+++ b/rb/build.desc
@@ -158,6 +158,15 @@ ruby_test(name = "ie",
   deps = [ ":ie" ]
 )
 
+ruby_test(name = "ie-w3c",
+  srcs = [
+    "spec/integration/selenium/webdriver/*_spec.rb",
+    "spec/integration/selenium/webdriver/ie/**/*_spec.rb"
+  ],
+  include = ["rb/spec/integration", "build/rb/lib"],
+  deps = [ ":ie" ]
+)
+
 ruby_library(name = "phantomjs",
   srcs = [
     "lib/selenium/webdriver/phantomjs/**/*.rb",
--- a/rb/lib/selenium/webdriver/common/driver.rb
+++ b/rb/lib/selenium/webdriver/common/driver.rb
@@ -45,7 +45,7 @@ module Selenium
 
           bridge = case browser
                    when :firefox, :ff
-                     if Remote::W3CCapabilities.w3c?(opts)
+                     if Remote::W3CCapabilities.marionette?(opts)
                        if opts[:desired_capabilities].is_a? Remote::Capabilities
                          opts[:desired_capabilities] = Remote::W3CCapabilities.new(opts[:desired_capabilities].send(:capabilities))
                        end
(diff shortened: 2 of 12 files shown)


rb/build.desc
```diff
@@ -158,6 +158,15 @@ ruby_test(name = "ie",
...
 
+ruby_test(name = "ie-w3c",
+  srcs = [
+    "spec/integration/selenium/webdriver/*_spec.rb",
+    "spec/integration/selenium/webdriver/ie/**/*_spec.rb"
+  ],
+  include = ["rb/spec/integration", "build/rb/lib"],
+  deps = [ ":ie" ]
+)
+
```
lmtierney: Since this is using the w3c bridge, it will need:

    
    resources = [
        { "//javascript/webdriver/atoms:getAttribute": "rb/lib/selenium/webdriver/atoms/getAttribute.js"}
      ],

Added to https://github.com/SeleniumHQ/selenium/blob/e5e00222491d0439c8b5a64ad3ed8a35d344282f/rb/build.desc#L144

At some point, should we just put that into `:common` as it would be needed for all w3c compliant drivers?

p0deje: Hm, I'm not familiar with it. Will this fix such specs:

```
10) Selenium::WebDriver::Element should get property value
      Failure/Error: expect(driver.find_element(id: 'withText').property('nodeName')).to eq('TEXTAREA')

        expected: "TEXTAREA"
             got: "Command not implemented"

        (compared using ==)
      # ./rb/spec/integration/selenium/webdriver/element_spec.rb:102:in `block (3 levels) in <module:WebDriver>'
```

lmtierney: No, that one is a legitimate failure. Jim hasn't implemented the new w3c `/property` endpoint yet and is working on it. (It doesn't fail on OSS IEDriver because I made an equivalent method to get the property)

The error you would see would be something like this:

    Selenium::WebDriver::Driver one element should find by css selector
     Failure/Error: expect(driver.find_element(css: 'div.content').attribute('class')).to eq('content')

     Errno::ENOENT:
       No such file or directory @ rb_sysopen - .../build/rb/lib/selenium/webdriver/atoms/getAttribute.js

You probably have one in your build directory from building firefox tests. If you ran `go clean //rb:ie-test` you should run into this error.

p0deje:
Closing in favor of #3809

PR #3752, opened 2017-04-03
deprecating AuthenticateUsing because it is no longer supported by an…

llaskin:
…y browser vendors.

### DIFF
--- a/java/client/src/org/openqa/selenium/Alert.java
+++ b/java/client/src/org/openqa/selenium/Alert.java
@@ -28,7 +28,7 @@ public interface Alert {
 
   void sendKeys(String keysToSend);
 
-  @Beta
+  @Deprecated
   void setCredentials(Credentials credentials);
 
   /**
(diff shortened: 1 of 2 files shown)


AutomatedTester:
This was only in 1 browser and was a "beta" feature. It is _currently_ not in W3C but could be in Level 2

llaskin:
Per Simon we are going to deprecate this feature.

jleyba:
Can you add `@deprecated` javadoc tags explaining that the APIs will be removed and there is intentionally no replacement?

shs96c:
`@Beta` gives us more freedom than `@Deprecated`. The guidelines we offer are that beta features can and will be deleted without notice. Deprecated features get at least one release before being deleted.

llaskin:
@shs96c so should I swap it to @deprecated?  I can't believe I didn't see Jason's request, and happy to change this...(hey look a ghost commit)

PR #3755, opened 2017-04-03
Addition of overload argument to the SelectElement.SelectByText method

### LINKED ISSUE #3575: Code review: SelectByText promises to substring match, but match is on exact text
While investigating an issue where certain text does not get selected, I encountered the following on the [**dotnet** branch](selenium/dotnet/src/support/UI/SelectElement.cs):

In the currect version, the method `SelectElement.SelectByText(string text)` promises to select the text on substring match:

```
        /// <param name="text">The text of the option to be selected. If an exact match is not found,
        /// this method will perform a substring match.</param>
```

but when I am looking in how the item(s) to select are picked from a list of candidates, the text from a select node is compared to the literal text:

```
   foreach (IWebElement option in candidates)
                {
                    if (text == option.Text)   // <-- literal match
                    {
                        SetSelected(option, true);
//...
```

Do you agree that this is indeed an issue?

seanrand57:
Issue raised in the issue log label as dotnet, issue number 3575. The
XML summary informs that the method provides partial match ability on
the options  by Text.  However this wasn't fully implemented so pushed
the ability via a boolean to perform such a partial match on an options
list.

Adding in the App.config change in order to run tests via the IDE for
the Support project.

Fixes #3575

### DIFF
--- a/dotnet/test/support/UI/SelectBrowserTests.cs
+++ b/dotnet/test/support/UI/SelectBrowserTests.cs
@@ -30,7 +30,6 @@ namespace OpenQA.Selenium.Support.UI
         [Test]
         public void ShouldThrowAnExceptionIfTheElementIsNotASelectElement()
         {
-           // driver.Url = formsPage;
             IWebElement element = driver.FindElement(By.Name("checky"));
             Assert.Throws<UnexpectedTagNameException>(() => { SelectElement elementWrapper = new SelectElement(element); });
         }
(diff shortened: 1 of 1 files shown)


Autom8edChaos:
Hi, thanks for picking this up! Just a thought: another possibility is to implement a SelectByPartialText method instead of overloading the SelectByText method. This immediately makes the intention clear for the reader of the method and will keep the SelectByText method unharmed.

seanrand57:
No problem glad to be of help, quick reason I choose this rather route vs a new method was a few fold, firstly to minimize the amount of new code required for the method.  Secondly to utilize the current SelectByText method tests to ensure normal functionality and backward compatibility is maintained by provided a default arg of false.

Also truth be told it scared me a little adding the word Partial to a method in the API as others might expect this to be the standard Partial Text method sig for others too ;)

@jimevans - not sure your thoughts but will take your direction on your review of the PR.

@Autom8edChaos hope that explains my chain of thought a little regarding the implementation and you're happy with the solution?

Here to help as always,
Cheers
Sean

Autom8edChaos:
@seanrand57 yes, I understand what you mean. In no time we would retrieve requests like _"pls implement SelectByPartialValue k tnx bye"_ or _"Selenium is absolute pants, there isn't even a method DeselectByPartialText!"_ . The optional parameter will make it work

seanrand57:
Hey, just seeing if this is a PR which can be merged up or what is required to do in order to pull into the solution?  Thanks in advance

jimevans:
Merged in d0accdf. Thank you for the contribution.

PR #3756, opened 2017-04-04
Adding optional gzip compression in csharp headers

dhimil:

### DIFF
--- a/dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
+++ b/dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
@@ -107,6 +107,7 @@ namespace OpenQA.Selenium.Remote
             request.Timeout = (int)this.serverResponseTimeout.TotalMilliseconds;
             request.Accept = RequestAcceptHeader;
             request.KeepAlive = this.enableKeepAlive;
+            request.AutomaticDecompression = DecompressionMethods.GZip;
             request.ServicePoint.ConnectionLimit = 2000;
             if (request.Method == CommandInfo.PostCommand)
             {


jimevans:
What is the purpose of this PR? I'm not aware of any wire protocol remote ends (OSS or W3C dialect) that would use such compression. Can you point me to one that does, and what benefit it has?

dhimil:
Hey @jimevans sorry to get back late on this.

**What the code does:**
It addes `accept-encoding` as GZip. This enables the only supported server to respond back with GZip compressed data. If the server does not support it, this header is ignored.

**About GZip:**
Most modern browsers use this for loading data.

**Why does it help:**
In high latency selenium grids, it helps compression data to get response faster.
In selenium, this helps massively for calls like `pageSource`, `screenshots`, etc (where there is large data in response).

To answer your question, I don't have any example of any wire protocol using it. This does not break selenium jar (in role = hub or node). 

Let me know anything further. 

Source:
[https://en.wikipedia.org/wiki/HTTP_compression](url)

jimevans:
My point wasn't that I didn't know what the header did, or what GZip is; rather, as far as I'm aware, the normal Selenium standalone server doesn't support its use as a server. I wanted to make sure you weren't expecting this to magically give you some boost in performance when running locally, or that it would magically speed up the communication between browser and the site being automated.

dhimil:
Totally don't believe in magic. (yet) :)
We front selenium server with servers like nginx or apache (which supports the functionality). And I said this is only for high latency (remote) grids.

I am trying to have the support in jar as well (parameterized, ofcourse). Will try to contribute to that as well in few days. This was an optional change that I thought might help immediately if someone is trying to make it work like we do.

jimevans:
I'm going to close this PR without merging. I've decided to implement a different approach that will allow anyone to customize the HTTP request getting sent to the remote server without the need to proliferate additional constructors to `RemoteWebDriver`. To wit, I'm implementing a `BeforeRemoteHttpRequest` event that you can hook your code into, and modify the HTTP request as needed. To use it, you'd do something like this:

    public void Start()
    {
        HttpCommandExecutor executor = new HttpCommandExecutor(new Uri("http://localhost:4444/wd/hub"), TimeSpan.FromSeconds(60));
        executor.BeforeRemoteHttpRequest += BeforeRemoteHttpRequestHandler;
        IWebDriver driver = new RemoteWebDriver(executor, new ChromeOptions().ToCapabilities());
     }

    private void BeforeRemoteHttpRequestHandler(object sender, BeforeRemoteHttpRequestEventArgs e)
    {
        // Note: WebRequest.DefaultWebProxy is an IWebProxy implementation.
        // This could be anything, from adding additional headers, to modifying
        // the content of the request. Use with extreme caution.
        e.Request.Proxy = WebRequest.DefaultWebProxy;
        e.Request.AutomaticDecompression = DecompressionMethods.GZip
    }

This ought to give us the flexibility to be able to modify the requests as required, and be able to handle all of the use cases that would entail, without needing to pass additional information around into the driver.

jimevans:
The event has been added in a3a2062, and will be available in the next release of the .NET bindings.

PR #3762, opened 2017-04-05
Update nodejs bindings for phantomjs to ignore ssl

BenRuns:
Allows to use phantomjs and ignore errors with SSL
It's useful for using browsermob proxy when you want
to skip configuring the SSL certs

### DIFF
--- a/javascript/node/selenium-webdriver/test/phantomjs/capabilities_test.js
+++ b/javascript/node/selenium-webdriver/test/phantomjs/capabilities_test.js
@@ -29,9 +29,10 @@ const webdriver = require('../..'),
 
 
 test.suite(function(env) {
-
+  var driver;
   test.afterEach(function(){
     sandbox.restore();
+    return driver.quit();
   });
 
   describe("capabilities.set('acceptSslCerts', true)", function() {
@@ -42,11 +43,10 @@ test.suite(function(env) {
           remote.DriverService.prototype,
           'start',
           function(options){
-            that = this;
-            return that.args_
+            return this.args_
             .then( args => {
               assert(args).contains('--ignore-ssl-errors=true');
-              return _start.bind(that).call(options);
+              return _start.call(this,options);
             });
           }
         );
@@ -54,9 +54,11 @@ test.suite(function(env) {
          var capabilities = webdriver.Capabilities.phantomjs();
          capabilities.set('acceptSslCerts', true);
 
-         return new webdriver.Builder()
+          driver = new webdriver.Builder()
          .withCapabilities(capabilities)
          .build();
+
+          return driver;
       });
    });
 


javascript/node/selenium-webdriver/test/phantomjs/capabilities_test.js
```diff
@@ -0,0 +1,64 @@
...
+              assert(args).contains('--ignore-ssl-errors=true');
+              return _start.bind(that).call(options);
+            });
+          }
+        );
+
+         var capabilities = webdriver.Capabilities.phantomjs();
+         capabilities.set('acceptSslCerts', true);
+
+         return new webdriver.Builder()
```
jleyba: Your test is leaking a WebDriver session. Make sure you call quit on the created driver.

javascript/node/selenium-webdriver/test/phantomjs/capabilities_test.js
```diff
@@ -0,0 +1,64 @@
...
+
+  describe("capabilities.set('acceptSslCerts', true)", function() {
+    test.it('sends --ignore-ssl-errors=true into the DriverService',
+       function() {
+        var _start = remote.DriverService.prototype.start;
+        sandbox.stub(
+          remote.DriverService.prototype,
+          'start',
+          function(options){
+            that = this;
```
jleyba: Missing `var/let` for this variable. You don't need it anyway: arrow functions do not bind this

javascript/node/selenium-webdriver/test/phantomjs/capabilities_test.js
```diff
@@ -0,0 +1,64 @@
...
+            return that.args_
+            .then( args => {
+              assert(args).contains('--ignore-ssl-errors=true');
+              return _start.bind(that).call(options);
```
jleyba: `return _start.call(this, options);`

BenRuns:
@jleyba  - changes have been made.

 Any thoughts on other strategies to test this? Felt wrong to spy on the DriverService this way. I'll happily write it if anyone has a better idea.

BenRuns:
@jleyba , I've got the requested change on  [cafc08ecb2](https://github.com/SeleniumHQ/selenium/pull/3762/commits/cafc08ecb2041a939c55e2b244cdfe86f41860e2)
Is there anything else?

jleyba:
If you just want to test that the DriverService is created with the correct set of args, maybe you could add a method that returns the args instead of breaking encapsulation for a private variable?

BenRuns:
On a different note, should the capabilities  for this feature actually be set to 'acceptInsecureCerts to conform with [w3c webdriver specs](https://www.w3.org/TR/webdriver/#capabilities) ?

jleyba:
Native support for phantomjs has been removed from selenium-webdriver for the 4.0 release.

Instead of using phantomjs, users can use Chrome or Firefox in headless mode. All the advantages of headless testing with a real browser.

PR #3766, opened 2017-04-05
Fix NullPointerException

asashour:

Fixing a coding typo, where `value` is it one to be cast to `Capabilities`.


barancev:
How to reproduce the issue? Can we add a test for this case?

asashour:
Actually this was detected by Eclipse static code analysis.

By looking into it, it seems that:

    } else if (value != null && value instanceof Capabilities) {
      capabilities = ((Capabilities) capabilities).asMap();
    }

is never called, since the `value` is never a `Capabilities`, because `JsonToBeanConverter` can not automatically detects it.

Should I change the PR to remove this condition?

barancev:
@shs96c your opinion?

asashour:
Closing, as this it not relevant anymore.

It has been resolved in https://github.com/SeleniumHQ/selenium/commit/e0e8f01c508f17e6f80137cc6e020f7233f4899d

PR #3803, opened 2017-04-10
Conform to the api of urllib2 for adding header for a request

clubfest:
urllib2 capitalize the header key and relies on this capitalization to decide if the 'Content-type' field is filled in and if not, fill it in with the default of 'application/x-www-form-urlencoded'. A recent change causes the header in the request named 'Content-type' to change to 'Content-Type' and urllib2 assumes the 'Content-type' field in the request header is not filled in, so tests are using the default instead and are giving the error:
<class 'selenium.common.exceptions.WebDriverException'>: Message: Given content type 'application/x-www-form-urlencoded' is not 'application/json;charset=UTF-8'.

### DIFF
--- a/py/test/unit/selenium/webdriver/remote/test_remote_connection.py
+++ b/py/test/unit/selenium/webdriver/remote/test_remote_connection.py
@@ -63,7 +63,8 @@ class MockResponse:
 
 
 def test_remote_connection_adds_connection_headers_from_get_remote_connection_headers(mocker):
-    test_headers = {'FOO': 'bar'}
+    test_headers = {'FOO': 'bar', 'Content-Type': 'json'}
+    expected_request_headers = {'Foo': 'bar', 'Content-type': 'json'}
 
     # Stub out the get_remote_connection_headers method to return something testable
     mocker.patch(
@@ -77,7 +78,7 @@ def test_remote_connection_adds_connection_headers_from_get_remote_connection_he
         mock_open = mocker.patch('urllib2.OpenerDirector.open')
 
     def assert_header_added(request, timeout):
-        assert request.headers == test_headers
+        assert request.headers == expected_request_headers
         return MockResponse()
 
     mock_open.side_effect = assert_header_added


Dude-X:
@clubfest This is pretty good. Have you tested this under Python 2 and Python 3?

clubfest:
It works for the webdriver tests that I have in Python 2 that were failing with the error.

Dude-X:
@clubfest would it be any trouble to test this change with Python 3 (3.5+)? You can create a virtual environment and just write a short script to confirm.

davehunt:
Could you also provide a test in https://github.com/SeleniumHQ/selenium/blob/master/py/test/unit/selenium/webdriver/remote/test_remote_connection.py so we don't regress again? You can then run this test against Python 2/3 using Tox as detailed here: https://github.com/SeleniumHQ/selenium/wiki/Python-Bindings#tests

clubfest:
It's a bit difficult to add a test for this currently, unless I create an independent function that returns the request from line 495
https://github.com/clubfest/selenium/blob/cbf367b5506ad02354eb004ef7cc86fa7e5a0352/py/selenium/webdriver/remote/remote_connection.py#L495

Or is it sufficient to add a comment to say that the request header must be modified with add_header instead of modifying the headers field?

clubfest:
I am a bit confused. Many of the tests that failed are in ruby, but the change is only for python. Is there a way to re-run the continuous integration?

lmtierney:
@clubfest ignore the ruby test failures

clubfest:
I am unable to run tests locally with tox. Am I missing some additional setup?
$ tox -e py27-unit

logs:
...
    error: can't copy 'selenium/webdriver/firefox/x86/x_ignore_nofocus.so': doesn't exist or not a regular file
    Running setup.py install for selenium: finished with status 'error'
...

lmtierney:
Yes, to get that file you need to follow [the instructions here](https://github.com/SeleniumHQ/selenium/wiki/Python-Bindings#development)

Specifically `./go py_prep_for_install_release`

clubfest:
I ran `./go py_prep_for_install_release`, and then encounter another error: 
   Running command python setup.py egg_info
      File "<string>", line 1, in <module>
      File "/tmp/pip-QOi0I0-build/setup.py", line 35, in <module>
        'long_description': open(join(abspath(dirname(__file__)), "README.rst")).read(),
    IOError: [Errno 2] No such file or directory: '/tmp/pip-QOi0I0-build/README.rst'

$ ls /tmp/pip-QOi0I0-build/
CHANGES      pip-delete-this-directory.txt  PKG-INFO  selenium.egg-info  setup.py
MANIFEST.in  pip-egg-info                   selenium  setup.cfg

clubfest:
Never mind. Just upgraded via `pip install -U setuptools pip`.

clubfest:
For the remaining failures, I don't think they are related to my changes. Please take a look, thanks!

davehunt:
Thanks!

PR #3809, opened 2017-04-11
Implement protocol handshake for Ruby bindings

p0deje:
### Description

~~This is currently work in progress.~~ This is done.

We need Ruby bindings to detect what protocol dialect it should talk (e.g. current IEDriver talks OSS dialect, but newer beta versions talk W3C dialect). In order to do that, we need to implement protocol handshake from Java bindings which:

See commit messages for details.

### TODO

- [x] Address all new TODOs (some of them are kept for later)
- [x] Verify driver extensions are correct for all browsers
- [x] Get approval from @lmtierney and @titusfortner
- [x] Fix Travis

### Tests

- [x] `go //rb:unit-test`
- [x] `go //rb:chrome-test` (ran against Chrome 57 and ChromeDriver 2.29)
- [x] `go //rb:edge-test` (ran against Microsoft Edge 38.14393.67.0 and 40.15063.0.0) (there are 19 failures in master and this branch)
- [x] `go //rb:ff-esr-test` (ran against Firefox 45 ESR to verify legacy driver)
- [x] `go //rb:ff-nightly-test` (there are 17 failures in master and 8 failures this branch)
- [x] `go //rb:firefox-test` (ran against Firefox 52/53 and GeckoDriver 0.15/0.16)
- [x] `go //rb:ie-test` (ran against IE11 and master IEDriver 32-bit)
- [x] `go //rb:ie-test` (ran against IE11 and beta IEDriver 32-bit) (there are 14 failures mostly about keyboard/mouse which I'm going to address in a followup commits)
- [x] `go //rb:phantomjs-test` (ran against PhantomJS 2.1.1)
- [x] `go //rb:remote-chrome-test` 
- [x] `go //rb:remote-edge-test` (there are 2 failures in master and this branch)
- [x] `go //rb:remote-ff-esr-test`
- [x] `go //rb:remote-ff-nightly-test` (there are 11 failures in master and this branch)
- [x] `go //rb:remote-firefox-test` (there are 5 failures in master and this branch)
- [x] `go //rb:remote-ie-test` (master IEDriver) (there are 5 failures in master and this branch)
- [x] `go //rb:remote-ie-test` (beta IEDriver) (there are 25 failures which I'm going to work on in the followup commits)
- [x] `go //rb:remote-phantomjs-test`
- [x] `go //rb:remote-safari-preview-test` (threre are 32 failures in master and this branch)
- [x] `go //rb:remote-safari-test` (there are 10 failures in master and this branch)
- [x] `go //rb:remote-test`
- [x] `go //rb:safari-preview-test` (there are 45 failures in master and this branch)
- [x] `go //rb:safari-test` (ran against Safari 10.1)


lmtierney:
Hoping to take a look at this soon

rb/lib/selenium/webdriver/remote/w3c/bridge.rb
```diff
@@ -0,0 +1,573 @@
...
+            super(opts)
+          end
+
+          def dialect
+            :w3c
+          end
+
+          def commands(command)
+            case command
+            when :status, :is_element_displayed
```
lmtierney: @titusfortner Any idea why this was here? There are plenty of commands duplicated between the two, it seems odd that these were singled out

https://github.com/SeleniumHQ/selenium/commit/e3fbb4a83458199356dd5dc20e5cef6a46414ff4#diff-5efb39c0076ea2abd8b2446e895603b5R93

rb/lib/selenium/webdriver/remote/w3c/capabilities.rb
```diff
@@ -0,0 +1,253 @@
...
+            :accept_insecure_certs,
+            :page_load_strategy,
+            :proxy,
+            :remote_session_id,
+            :accessibility_checks,
+            :device,
+            :implicit_timeout,
+            :page_load_timeout,
+            :script_timeout,
+          ].freeze
```
lmtierney: Spec has required capabilities of:

- browser_name
- browser_version
- platform_name
- accept_insecure_certs
- page_load_strategy
- proxy
- set_window_rect
- timeouts
- unhandled_prompt_behavior

rb/lib/selenium/webdriver/remote/w3c/capabilities.rb
```diff
@@ -0,0 +1,253 @@
...
+                platform: :windows
+              }.merge(opts))
+            end
+
+            def firefox(opts = {})
+              opts[:browser_version] = opts.delete(:version) if opts.key?(:version)
+              opts[:platform_name] = opts.delete(:platform) if opts.key?(:platform)
+              opts[:timeouts] = {}
+              opts[:timeouts]['implicit'] = opts.delete(:implicit_timeout) if opts.key?(:implicit_timeout)
+              opts[:timeouts]['page load'] = opts.delete(:page_load_timeout) if opts.key?(:page_load_timeout)
```
lmtierney: this should actually be `pageLoad` per the spec, which means we currently have it wrong

rb/build.desc
```diff
@@ -146,6 +146,9 @@ ruby_library(name = "ie",
     "lib/selenium/webdriver/ie/**/*.rb",
     "lib/selenium/webdriver/ie.rb"
   ],
+  resources = [
+    { "//javascript/webdriver/atoms:getAttribute": "rb/lib/selenium/webdriver/atoms/getAttribute.js"}
```
lmtierney: Since we now have the protocol handshake, we should probably add this to common instead of including in each one. The only one that shouldn't need it in the long run would be legacy FF and it's not that much overhead

rb/lib/selenium/webdriver/remote/w3c/capabilities.rb
p0deje: I'd like to rework the capabilities in the followup PRs, I've left a TODO for it above the constant.

rb/lib/selenium/webdriver/remote/w3c/capabilities.rb
p0deje: let's fix it in the next PRs

rb/build.desc
p0deje: Sounds reasonable.

p0deje:
It's been almost 2 weeks, so I'm going to merge it later today unless there will be any objections.

rb/lib/selenium/webdriver/remote/w3c/bridge.rb
titusfortner: These two commands are implemented, but aren't part of w3c spec, so they reference the oss implementation. Probably not necessary to do it this way.

titusfortner:
I'm halfway through a review, but there's a lot more code here than I have time to work through right now. I can finish up tomorrow. Thanks for your patience.

rb/lib/selenium/webdriver/firefox/marionette/bridge.rb
```diff
@@ -0,0 +1,65 @@
...
+      module Marionette
+
+        #
+        # Extension for the usual Remote::W3C::Bridge
+        # @api private
+        #
+
+        module Bridge
+
+          # Support for geckodriver < 0.15
```
lmtierney: I put this in the current bridge, @p0deje you can remove these commands

p0deje:
> I guess it decouples capabilities from the Bridge classes. So the Bridge is now just the command interface to the server/driver, which is a positive move.

Yeah, that was the idea.

> Not that it should be our primary concern, but I'm curious how much this is going to affect the Ruby Appium library since it does so much monkey patching of the internals.

I know they have already created appium/ruby_lib#558 and appium/ruby_lib#559. It would be great to hear from them here, but I don't really have time to reach them out.

> Also, we've been telling people that if they want to customize their driver/server interaction to subclass a Bridge and make their own custom commands. So, I'm concerned that this change will completely break their implementation.

It's not changed, isn't it? You can still pass custom bridge so I don't see a problem here:

```ruby
bridge = MyCustomBridge.new
driver = Selenium::WebDriver::Driver.new(bridge)
```

> From a high level view, I don't think this is how I would have approached it, but sketching out a few alternate ideas this week I've come to the conclusion that a more elegant solution isn't obvious to me. :) As for the actual implementation of this design, I think I see how everything is working and I don't have any major issues.

I was attempting to also make it more consistent to other bindings. None of them have a bridge, but all of them have custom driver classes, so I actually think it is now clearer. I would have to decouple capabilities and bridge anyways and driver class seems like a best option.

rb/lib/selenium/webdriver/remote/bridge.rb
```diff
@@ -20,653 +20,141 @@
...
-        def touch_move(x, y)
-          execute :touch_move, {}, {x: x, y: y}
-        end
-
-        def touch_scroll(element, x, y)
-          if element
-            execute :touch_scroll, {}, {element: element,
-                                       xoffset: x,
-                                       yoffset: y}
+          if oss_status
```
p0deje: That's how Java and .NET bindings do this, so ¯\_(ツ)_/¯

rb/lib/selenium/webdriver/firefox/legacy/driver.rb
```diff
@@ -0,0 +1,78 @@
...
+              port = opts.delete(:port) || DEFAULT_PORT
+              profile = opts.delete(:profile)
+
+              Binary.path = opts[:desired_capabilities][:firefox_binary] if opts[:desired_capabilities][:firefox_binary]
+              @launcher = Launcher.new Binary.new, port, profile
+              @launcher.launch
+              opts[:url] = @launcher.url
+            end
+
+            @bridge = Remote::Bridge.handshake(opts)
```
p0deje: There are some weird things there. You've implemented Edge bridge as W3C bridge, however it responds like OSS one, so there is a bit of confusion here. By forcing W3C in Edge, I'm keeping the current behavior and implementation, though I'd like to rework this later.

In regards to legacy Firefox, there is actually no need to do this, so I'll change.

titusfortner: Edge was implemented as W3C because that was the format of the desired capabilities it accepted, which was what I was most concerned about. Edge Driver has from the beginning supported an interesting mix of OSS & W3C commands. Yes, this distinction won't matter once we send both formats, though it could raise some other issues if this approach expects drivers to be completely compatible with either OSS or W3C.

rb/lib/selenium/webdriver/firefox/marionette/bridge.rb
p0deje: @lmtierney Sorry, I don't follow what you mean

rb/lib/selenium/webdriver/firefox/legacy/driver.rb
p0deje: I've changed it so now neither `Marionette::Driver` nor `Legacy::Driver` do handshake.

rb/lib/selenium/webdriver/firefox/marionette/bridge.rb
lmtierney: @p0deje you can get rid of this entire bridge. These window commands were added for backward compatibility with older geckodriver versions. Since we are fine with requiring the latest geckodriver they aren't needed anymore

p0deje:
Ah, okay, awesome!

p0deje:
All the comments are addressed, now just waiting for CI.

PR #3826, opened 2017-04-12
Replace TypeError with KeyError in remote webdriver error handler code

dveselov:
Hi!

Looks like that there is a typo (because dictionary lookup code, in that case, can't raise TypeError)

### DIFF
--- a/py/selenium/webdriver/remote/errorhandler.py
+++ b/py/selenium/webdriver/remote/errorhandler.py
@@ -108,7 +108,7 @@ class ErrorHandler(object):
                         message = value["value"]
                         if not isinstance(message, basestring):
                             value = message
-                            message = message.get('message', None)
+                            message = message.get('message')
                     else:
                         message = value.get('message', None)
                 except ValueError:


Dude-X:
@dveselov Thank you for your contribution!
Actually we can improve this code and save a few precious microseconds with this change, remove the try/except and replace it with:
`message = message.get('message')`

py/selenium/webdriver/remote/errorhandler.py
```diff
@@ -110,7 +110,7 @@ def check_response(self, response):
                             value = message
                             try:
                                 message = message['message']
-                            except TypeError:
+                            except KeyError:
```
AutomatedTester: could you change this to instead of being a `try.. except` be

`message = message.get('message')`

dveselov:
Done.

py/selenium/webdriver/remote/errorhandler.py
```diff
@@ -108,10 +108,7 @@ def check_response(self, response):
                         message = value["value"]
                         if not isinstance(message, basestring):
                             value = message
-                            try:
-                                message = message['message']
-                            except TypeError:
-                                message = None
+                            message = message.get('message', None)
```
Dude-X: I would recommend to leave out `None` because None is already the default.

dveselov: uh, ok.

PR #3834, opened 2017-04-13
Delete JavascriptEnabled test annotation

asashour:

The purpose of the change, is to remove `@JavascriptEnabled` test annotation, which was used for only HtmlUnit, and it is not needed anymore.


shs96c:
Is running without JS support no longer supported by HtmlUnit?

asashour:
Running without JS support is supported, but not the default. In both HtmlUnit and HtmlUnitDriver.

I see almost all the tests in Selenium base, depend on JavaScript ennoblement.

After the deprecation of `Capabilities.isJavaScriptEnabled` (as I remember), a new `JAVASCRIPT_ENABLED` Capabilities was introduced in HtmlUnitDriver, so users can disable it when needed.

If there are test cases related to 'no JS', then it should be bound to only HtmlUnitDriver and HtmlUnit (the latter is tested in HtmlUnit suite).

And since the general tests of selenium seem not to require 'JavaScript off', except in only HtmlUnitDriver, then it makes sense in only HtmlUnitDriver test package, but not for all tests.

shs96c:
Thank you for the thoughtful answers.

shs96c:
Merged as 8ad8daba8c89735a90b8534690fafdb2600708a7. Thank you!

PR #3843, opened 2017-04-15
修复'NoneType' object has no attribute '__getitem__'

liushilive:
按照处理逻辑，如果没有value值，应该把body赋予value才对，不然会出现value值不存在导致程序错误。
Windows10+Python2.7+selenium3.3.3+Firefox

element.location()
element.rect()
.....

File "C:\Users\liush\SeleniumBase\lib\site-packages\selenium\webdriver\remote\webelement.py", line 405, in location
    new_loc = {"x": round(old_loc['x']),
TypeError: 'NoneType' object has no attribute '__getitem__'

### DIFF
--- a/py/selenium/webdriver/remote/remote_connection.py
+++ b/py/selenium/webdriver/remote/remote_connection.py
@@ -556,7 +556,7 @@ class RemoteConnection(object):
                 # Some of the drivers incorrectly return a response
                 # with no 'value' field when they should return null.
                 if 'value' not in data:
-                    data['value'] = None
+                    data['value'] = eval(body.strip())
                 return data
             else:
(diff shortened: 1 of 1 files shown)


Dude-X:
@liushilive This change is bad because it uses an eval. Most likely there is probably a bug in the geckodriver.
Do you have a reproducible test case that triggers the error you're seeing?

lmtierney:
We will not be using `eval` in remote_connection; it is used for every command. If you are getting this error, then it's due to a mismatch in geckodriver and/or Firefox. Please seek support with this from the [selenium-users@ mailing list](https://groups.google.com/forum/#!forum/selenium-users) or the `#selenium` IRC channel on Freenode

PR #3846, opened 2017-04-17
[js] Add support for Firefox WebExtensions.

tobli:
Add support for loading Firefox WebExtensions (see https://developer.mozilla.org/en-US/Add-ons/WebExtensions).

### DIFF
--- a/javascript/node/selenium-webdriver/firefox/extension.js
+++ b/javascript/node/selenium-webdriver/firefox/extension.js
@@ -154,9 +154,7 @@ function parseInstallRdf(rdf) {
  * @param {!Object} json JSON representation of the manifest.
  * @return {!AddonDetails} The add-on details.
  */
-function parseManifestJson(json) {
-  const {name, version, applications} = json;
-
+function parseManifestJson({name, version, applications}) {
   if (!(applications && applications.gecko && applications.gecko.id)) {
     throw new AddonFormatError('Could not find add-on ID for ' + addonPath);
   }


jleyba:
Please add tests for your change.

tobli:
The included web extension doesn't modify the page in the test case, since I couldn't get the webextension to do that for data:-urls, as used in the test suite. Other than that, the tests are similar to those for the other Firefox extension types.

soulgalore:
About the current test extension: As I understands it, you need to match an extension against  ["http", "https", "file", "ftp", "app" ](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/match_patterns) so data URLs don't work, but maybe someone at Mozilla can help us out there? Or do we have a server running on localhost when we do the testing that we can use?

For me this PR is important because right now I'm in a limbo state if when I wanna create a new WebExtension for Firefox and run it in the NodeJS version of Selenium:

* Selenium needs a install.rdf file following the old XML format of an extension, so creating a new WebExtension following https://developer.mozilla.org/en-US/Add-ons/WebExtensions can't be installed using Selenium. 
* If you have both a manifest.json (new standard) and a install.rdf you cannot sign the extension at https://addons.mozilla.org/. 
* If you remove the manifest.json (but you really want that if you build an extension for both Chrome & Firefox) and only have a install.rdf I got an error from addons that I'm missing a manifest.json.

Would it be possible to get this out asap in 3.4.1 since 3.4.0 is already tagged? Let me or @tobli know if there's anything else needed.

jleyba:
There is a basic server available to the test suite. Here's an example:

https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/test/execute_script_test.js#L41

You can also start another server as long as your test doesn't leave it running:

https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/test/proxy_test.js#L32

tobli:
Thanks for the info @jleyba! I've updated the test to actually assert the web extension is working as well. Super easy with the built in http server.

javascript/node/selenium-webdriver/firefox/extension.js
```diff
@@ -138,46 +150,74 @@ function getDetails(addonPath) {
...
- * @param {string} addonPath Path to a Firefox add-on as a xpi or an extension.
- * @return {!Promise<!Object>} A promise for the parsed manifest.
+ * Parse a manifest for a Firefox WebExtension.
+ * @param {!Object} json JSON representation of the manifest.
+ * @return {!AddonDetails} The add-on details.
  */
-function readManifest(addonPath) {
-  var manifest;
+function parseManifestJson(json) {
+  const {name, version, applications} = json;
```
jleyba: How about

```js
function parseManifestJson({name, version, applications}) {
  // ...
}
```

tobli: Yes, even better. I'll push an update soon.

PR #3905, opened 2017-04-26
Stop passing array as "text" when calling sendKeys

skratchdot:
This was causing an issue with geckodriver v16.0.
See: https://github.com/mozilla/geckodriver/issues/683

The fix appears to work with both geckodriver and chromedriver, but I need to do some more testing.  I ran `./go clean test` and all the javascript tests passed, but for some reason I keep seeing 6 failed tests (which appear to be unrelated).  I'm trying to compare to the master branch, but I keep seeing java heap errors there.

### DIFF
--- a/javascript/node/selenium-webdriver/lib/webdriver.js
+++ b/javascript/node/selenium-webdriver/lib/webdriver.js
@@ -2186,7 +2186,7 @@ class WebElement {
     if (!this.driver_.fileDetector_) {
       return this.schedule_(
           new command.Command(command.Name.SEND_KEYS_TO_ELEMENT).
-              setParameter('text', keys).
+              setParameter('text', keys.then(keys => keys.join(''))).
               setParameter('value', keys),
           'WebElement.sendKeys()');
     }


skratchdot:
somewhat related:
https://github.com/mozilla/geckodriver/commit/41f89d878c805e0d66a15f8b6151dda78173ccff
and
https://github.com/mozilla/geckodriver/issues/659

javascript/node/selenium-webdriver/lib/webdriver.js
```diff
@@ -2186,7 +2186,7 @@ class WebElement {
     if (!this.driver_.fileDetector_) {
       return this.schedule_(
           new command.Command(command.Name.SEND_KEYS_TO_ELEMENT).
-              setParameter('text', keys).
+              setParameter('text', keys.then(keys => keys.join(''))).
```
jleyba: Also update line 2205 below

skratchdot: @jleyba - well, so this is where the code is a bit weird (and hard to follow). I wanted to keep the change as minimal as possible.

That whole "second block" starting at 2198 deals with passing in a file path (for file inputs) I believe.  Anyways, it generates the `keys` value, and on line 2201 calls `keys.join()`.  So at that point, I believe `keys` is a string (not an array).

So line 2205: `text` is string
And line 2206: `value` is an array

Which, after my change, *should* match.   In the existing code:
line 2189: `text` is an array
line 2190: `value` is an array

What I am trying to do with my change is make:
line 2189: `text` is a string
line 2190: `value` is an array

That way `text` is always a string, and `value` is always an array.

With that said, this is my first change to this repo, and I don't want to break things for people.  I did testing locally with the latest FF/Chrome geckodriver/chromedriver, and things worked.  I also ran the unit tests (which *seemed* to work for my changes).  I think they are in a bad state currently, b/c a few tests failed before and after my changes.

jleyba: Ah, my mistake. You're right, `keys` will always be a string at 2205.

sschuberth:
@jleyba Can we expect a minor release, i.e. 3.4.1, with this fix soon? I'm trying to find out whether it makes sense to wait, or whether I have to implement a work-around.

rameramwe:
Thank you so very much sir for this great fix !! you are a life saver .

PR #4027, opened 2017-05-16
Expose remoteHost as a parameter.

robertoriv:
Adding remoteHost as a parameter allows overriding the default (`http://<host>:<port>`).

This is useful for containerized environments where selenium’s known port doesn’t match the port that the hub needs to talk to it.

### DIFF
--- a/java/server/src/org/openqa/grid/internal/utils/configuration/GridNodeConfiguration.java
+++ b/java/server/src/org/openqa/grid/internal/utils/configuration/GridNodeConfiguration.java
@@ -152,8 +152,15 @@ public class GridNodeConfiguration extends GridConfiguration {
    * config parameters which do not serialize to json
    */
 
-  // remoteHost is a generated value based on host / port specified, or read from JSON.
-  @Expose( serialize = false )
+  /**
+   * The address to report to the hub. By default it's generated based on the host and port specified.
+   * Setting a value overrides the default (http://<host>:<port>).
+   */
+  @Expose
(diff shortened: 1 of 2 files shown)


robertoriv:
For some context:

When working with ECS, you can't rely on container linking for communication (unless all nodes are part of the same task, in which case you can't spread them across multiple boxes).

Say you have three selenium "instances" running as containers:
```
Hub:
  Instance IP: 10.0.1.1
  Container IP: 172.0.1.1
  Selenium Port: (inside container): 4444
  Mapped Port: (for outside communication): 4444

Node1:
  Instance IP: 10.0.1.2
  Container IP: 172.0.1.1
  Selenium Port (inside the container): 5555
  Mapped Port (for outside communication): 37600

Node2:
  Instance IP: 10.0.1.2
  Container IP: 172.0.1.1
  Selenium Port (inside the container): 5555
  Mapped Port (for outside communication): 37601
```

By default, Node1 and Node2 will report `http://172.0.1.x:5555` (what they know about). But if the hub were to communicate back to that address, connectivity would fail. (The `172.0.1.1` address the hub container knows about is not the same as the one *Node1* is running on).

By overriding `remoteHost` *Node1* and *Node2* can report `http://10.0.1.2:37600` and `http://10.0.1.2:36701`, respectively.

shs96c:
Thanks for the patch, and sorry for taking so long to get to it. This has been applied as 827bbd8177651a9316a574c9622f7d1226221173

PR #4043, opened 2017-05-19
Add failing test for a sequence of drag and drop by offset and element

p0deje:
The test would fail on IE11 + Win 10:

```
Failed : OpenQA.Selenium.Interactions.DragAndDropTest.DragAndDropRelativeAndToElement
  Expected: {X=8,Y=8}
  But was:  {X=8,Y=26}
```

Interestingly, if you turn the sequence of drag and drop actions into two separate actions, the test passes correctly.

This is basically [the copy of a failing Ruby test](https://github.com/SeleniumHQ/selenium/blob/05be826887db7e1864e9f75c1257c0592658f24f/rb/spec/integration/selenium/webdriver/element_spec.rb#L183-L197) to make it easier for @jimevans to debug.

### DIFF
--- a/dotnet/test/common/Interactions/DragAndDropTest.cs
+++ b/dotnet/test/common/Interactions/DragAndDropTest.cs
@@ -58,6 +58,22 @@ namespace OpenQA.Selenium.Interactions
 
         [Test]
         [Category("Javascript")]
+        [IgnoreBrowser(Browser.HtmlUnit)]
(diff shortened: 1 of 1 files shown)


jimevans:
Merged.

titusfortner:
@jimevans is this test passing in .NET tests? I'm getting a failure in Ruby tests still; should I log a bug?

PR #4107, opened 2017-05-30
Fix EventFiringWebdriver and WebElement to raise AttributeError on missing attributes.

isaulv:


cgoldberg:
@Dude-X 
This PR removed the catching of AttributeError entirely (previously it was unreachable code).  I'm just checking if that was intentional, since it's the exact opposite of what this PR's description states it does :)

Dude-X:
@cgoldberg :)
You're right, the title is a bit misleading, but the code implicitly raises AttributeError if that's the error, but it also fires the listener's on_exception hook. It also address a potential bug with `hasattr` which is a bit broken in Python 2.

PR #4120, opened 2017-06-02
Allow custom session factories in WebDriverServlet.

juangj:
This is analogous to DriverSessions. The intent is so the server can be extended to support other kinds of SessionFactory, e.g., DriverServices that aren't part of the Selenium project.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/4120)


juangj:
I'm not really sold on this design -- for one, I suspect there was a reason you didn't want to match based on the Capabilities like DriverSessions does -- but feedback/discussion about how we could support this case would be great.

shs96c:
There are a few changes I'd like to be able to make to the new way of doing things:

* I'd like to kill `DriverSessions`. It's overly complicated and has too many roles and responsibilities
* I'd like to extract an `ActiveSessions` class that wraps the `Cache`
* We need to project each `ActiveSession` as a `WebDriver` instance in order to support RC compatibility. *sigh*
* There's a whole heap of stuff that needs to happen around `New Session`. We don't do proper expansion of the sections, and the way we pick is as simple as can be.

For that last point, really what we should be doing is writing each capability to disk (since a profile may push it past being comfortable to fit in memory) and the have an mmap'ed version presented to the "thing that creates new active sessions" (which is not the `ActiveSessions` class above --- it may be the `Factory` in this PR).

I think we'll end up with a `Function<Capabilities, Optional<ActiveSession>>` being the factory.

juangj:
OK, seems sensible. I don't know much about RC compatibility -- RC was pretty much gone from our codebase by the time I came along -- but is there anything useful I can help with while you're away?

shs96c:
Not a huge amount :(

shs96c:
I've just pushed my WIP to the `session-factories` branch. Feel free to continue hacking on that! It's not tested --- really is just WIP.

diemol:
Closing this as the PR went stale but also since the code has evolved quite a bit. Also, Grid 3 is not under development and is being replaced with Grid 4. If you'd like to continue with this feature, feel free to have a look at the new codebase and create a GitHub issue with the "Feature Request" template to continue the conversation.

PR #4121, opened 2017-06-02
Log level is always All

MatthewLymer:
The enum.parse is incorrect as its type parameter is not the enum LogLevel.


shs96c:
Looks good to me. Good catch. Thank you!

Issue #4108, opened 2017-05-30
Warn if path provided by the user to screenshot-taking method has non-png extension


abotalov:
Currently most of client libraries support taking screenshot only in PNG format. Only [.NET is able to convert a screenshot to some other format](https://github.com/SeleniumHQ/selenium/blob/1e221937594c8ac4895899024927575aa6015a60/dotnet/src/webdriver/Screenshot.cs).

## Actual Behavior
Screenshot methods defined in most client libraries do not warn if path provided by the user has non-PNG extension. There are multiple cases where users wrote code that actually does it (see "Steps to reproduce"). I think they expected a screenshot to be saved in the format specified by the path extension.

## Expected Behavior
It would be better if screenshot-taking methods outputted a warning if path provided by the user has an extension but this extension (after being downcased) is not ".png". A warning should tell that despite a path ends with e.g. ".jpg", image will be saved in PNG format.

## Steps to reproduce
https://stackoverflow.com/a/24445213/841064
https://github.com/watir/watir/issues/400
http://artoftesting.com/automationTesting/screenShotInSelenium.html
http://www.softqe.com/capture-screenshot-using-selenium-webdriver/
http://www.ufthelp.com/2015/01/screenshot-using-selenium.html
https://community.perfectomobile.com/posts/1055085-sample-script-screenshot-python

cc @titusfortner

AutomatedTester:
@abotalov a warning or an exception?

abotalov:
@AutomatedTester I think a warning would be better because of backwards-compatibility.

cgoldberg:
I just put up a PR to address this.  see #4141

p0deje:
Ruby bindings support in bf9882412f117a28fb9a61278fa270ec7187cf4e.

lmtierney:
Python bindings support in 8780494

barancev:
It's not applicable to Java, user can't specify a path, getScreenshotAs method returns a temp file that has proper extension.

diemol:
Seems this was implemented in all bindings but .NET, in any case, if someone needs this for .NET bindings, please raise a new issue or/and PR.

PR #4141, opened 2017-06-07
py - add warning when saving incorrectly named screenshot

cgoldberg:
this addresses #4108 for Python bindings.

### DIFF
--- a/py/selenium/webdriver/remote/webdriver.py
+++ b/py/selenium/webdriver/remote/webdriver.py
@@ -836,6 +836,9 @@ class WebDriver(object):
         :Usage:
             driver.get_screenshot_as_file('/Screenshots/foo.png')
         """
+        if not filename.lower().endswith('.png'):
+            warnings.warn("name used for saved screenshot does not match file "
+                "type. It should end with a `.png` extension", UserWarning)
         png = self.get_screenshot_as_png()
         try:
             with open(filename, 'wb') as f:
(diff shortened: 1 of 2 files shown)


py/selenium/webdriver/remote/webelement.py
```diff
@@ -450,6 +452,9 @@ def screenshot(self, filename):
             return False
         finally:
             del png
+        if not filename.lower().endswith('.png'):
```
AutomatedTester: Can you move this warning to be the first thing that happens in the method so people can see the warning _and_ get any exceptions that might happen.

py/selenium/webdriver/remote/webdriver.py
```diff
@@ -842,15 +844,20 @@ def get_screenshot_as_file(self, filename):
             return False
         finally:
             del png
+        if not filename.lower().endswith('.png'):
```
AutomatedTester: Can you move this warning to be the first thing that happens in the method so people can see the warning _and_ get any exceptions that might happen.

py/selenium/webdriver/remote/webelement.py
cgoldberg: done

py/selenium/webdriver/remote/webdriver.py
cgoldberg: done

abotalov:
Do you think saving an image to a path without extension should be fine as well?

AutomatedTester:
> Do you think saving an image to a path without extension should be fine as well?

I think this is fine. If people do silly things like that then thats their own fault.

PR #4196, opened 2017-06-19
Propagate altered DC to Node

krmahadevan:
Currently the grid logic is such that if a particular
test’s capabilities are altered via an overloaded
version of 

* DefaultRemoteProxy#getNewSession
* DefaultRemoteProxy#beforeSession

the expectation is that the altered capabilities
are what gets forwarded to the actual node prior to
an actual session creation on the node. But 
this doesn’t happen because the altered desired 
capabilities are not being considered.
Fixed this.

### DIFF
--- a/java/server/src/org/openqa/grid/web/servlet/handler/WebDriverRequest.java
+++ b/java/server/src/org/openqa/grid/web/servlet/handler/WebDriverRequest.java
@@ -29,10 +29,8 @@ import org.openqa.grid.internal.Registry;
 import org.openqa.selenium.remote.BeanToJsonConverter;
 import org.openqa.selenium.remote.JsonToBeanConverter;
 
-import java.io.IOException;
 import java.util.Map;
 
-import javax.servlet.ServletInputStream;
 import javax.servlet.http.HttpServletRequest;
 
 public class WebDriverRequest extends SeleniumBasedRequest {
@@ -110,9 +108,4 @@ public class WebDriverRequest extends SeleniumBasedRequest {
     }
   }
 
-  @Override
-  public ServletInputStream getInputStream() throws IOException {
-    setBody(getBody());
-    return super.getInputStream();
-  }
 }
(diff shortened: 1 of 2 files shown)


krmahadevan:
ping @lukeis @shs96c @barancev - Can you please help with this pull request ?

krmahadevan:
Took care of merge conflicts.
Gentle reminder @shs96c @lukeis

java/server/src/org/openqa/grid/web/servlet/handler/WebDriverRequest.java
```diff
@@ -65,17 +73,46 @@ public ExternalSessionKey extractSession() {
...
+        map.add(DESIRED_CAPABILITIES, dc);
+      }
+      return new Gson().toJson(map);
+    } catch (Exception e) {
+      return json;
+    }
+  }
+
+  @Override
+  public ServletInputStream getInputStream() throws IOException {
```
shs96c: It would be better if `SeleniumBasedRequest` used a call to `getBody` rather than the field. This would mean subclasses wouldn't need to do this kind of thing to make sure that the `InputStream` contained the correct content.

krmahadevan:
@shs96c - Thanks for the feedback Simon. I have addressed your review comments. Please help take a look.

java/server/src/org/openqa/grid/web/servlet/handler/WebDriverRequest.java
krmahadevan: Took care of this by moving the call to `getBody()` from within the SeleniumBasedRequest.getInputStream()`. Chose to do `setBody(getBody());` instead of 
`return new ServletInputStreamImpl(new ByteArrayInputStream(body.getBytes()));` to avoid repetition of the `getBytes()` invocation.

shs96c:
Pushed as 86419e922cbfafaf7b2471248a8ff92901984005. Thank you!

Issue #4232, opened 2017-06-28
Rb Polymorphism with Handshake


titusfortner:
The Appium code base is a literal mess, so I'm going to give examples based on the Sauce Client gem I'm writing.

What I want to do is take actions before & after driver initialization, add a `#sauce` method for interacting with our API and annotations, and take additional actions before the `#quit` method.

Let's start with what works before `selenium-webdriver-3.4.1.gem`: 
[Subclassing the Bridge](https://github.com/titusfortner/saucer/blob/bridge_subclass/lib/saucer/driver.rb)

This no longer works because `Selenium::WebDriver::Remote::Bridge` can't be initialized with a `:desired_capabilities` parameter. (`ArgumentError: unknown option: {:desired_capabilities=> ...`)

I get that with the handshake this implementation probably doesn't work. It also has the disadvantage of requiring duplication of event listener handling if present.

So I tried [implementing this with composition](https://github.com/titusfortner/saucer/blob/master/lib/saucer/driver.rb). This mostly works except that I have to add the `#method_missing` and the `#respond_to?`, as well as needing to do `@driver.driver` to get the instance of the `Remote::Driver` instead of just `@driver`. Additionally, I can't do `Watir::Browser.new(@driver)`, because Watir is looking for `@driver` to be an instance of `Selenium::WebDriver::Driver`, and it isn't.

(The other sticking point here is that while the above code worked in `selenium-webdriver-3.4.2.gem` it is broken in `selenium-webdriver-3.4.3.gem`).

So this brings us to what I really want. A subclass of `Selenium::WebDriver::Driver`.
This is the classic use case of inheritance. The Sauce driver *is a driver with a couple extra behaviors. Ideally, [this is what I think it should look like to the user](https://github.com/titusfortner/saucer/blob/inheritence/lib/saucer/driver.rb). `#super` in the right spots, with the extra method that doesn't require monkey patching, or duplicating code for handshake or listeners.

Hopefully this clarifies what I'm looking for. I have a couple ideas for what might work as far as an implementation, but I haven't had a chance to dig into it yet.

KazuCocoa:
👀

p0deje:
Is this still relevant after b700761 merged?

titusfortner:
Nope, good now. Thanks

PR #4234, opened 2017-06-28
[rb] implement Remote::Driver as a class not a module

titusfortner:
I don't love this implementation, but it lets me subclass the way I need to as described in #4232 
@p0deje is there a reason not to go this route other than it is a little uglier?


p0deje:
I remember trying to do something like this, but for some reason I chose a different path, though out of top of my head I don't see why this is worse. If you can make specs pass, let's merge it. If not, I'll try doing that later this week.

titusfortner:
merged b7007615205970df5ca4e651f1573a46b3535ebf

PR #4238, opened 2017-06-28
[rb] raise exception when using :firefox_options in Remote::Capabilities

titusfortner:
This is currently failing silently.

The alternative is to try to handle the `:firefox_option` parameters & pass a deprecation notice.


p0deje:
Not sure which one is better, the rework of capabilities is still pending :)

titusfortner:
well, I merged this for now since otherwise tests will be failing without people knowing why.

PR #4266, opened 2017-07-06
Add Actions#hoverMouseOver, alias of #moveToElement and #moveByOffset

alb-i986:
Provide an alias so that the method `#moveToElement` is more searchable.
Using the auto completion features of the IDE, people can now find `#moveToElement` by searching for "hover".

### DIFF
--- a/java/client/src/org/openqa/selenium/interactions/Actions.java
+++ b/java/client/src/org/openqa/selenium/interactions/Actions.java
@@ -371,6 +371,17 @@ public class Actions {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


shs96c:
Thank you for the PR, and for thinking of users. Thank you also for spending the time needed to get to grips with the project and send this in.

One guiding principle when developing APIs (which, I'll admit, we've followed inconsistently) is to provide as small a surface area as possible. The `Actions` class is as large as it is for legacy reasons, but so far there are no aliases for methods in there. Starting that process is likely to end poorly in terms of bloat in the APIs. For this reason, I'm closing the PR.

PR #4270, opened 2017-07-06
java client - Extract BrowserOptions interface.

FA-tvonmoll:
The intent is to standardize an approach to integrating
the various options objects (`SafariOptions`, `ChromeOptions`, etc.)
with a `DesiredCapabilities` object.

Each implementation handles this a bit differently; this would
provide a common mechanism (and contract for future classes).

The approach that I adopted is `addTo(Capabilities capabilities)`, which was already in `FirefoxOptions`. This doesn't eliminate the possibility of the original approaches (just adds a common one).

### DIFF
--- a/java/client/src/org/openqa/selenium/BUCK
+++ b/java/client/src/org/openqa/selenium/BUCK
@@ -24,6 +24,7 @@ java_library(name = 'core',
   srcs = [
     'Alert.java',
     'By.java',
+    'BrowserOptions.java',
     'Capabilities.java',
     'Cookie.java',
     'ContextAware.java',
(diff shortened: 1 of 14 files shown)


shs96c:
An alternative approach is here: 3adacfe2c4931e47f062f91602e038ca572d87b7

This makes `IEOptions` an instance of `Capabilities`, so can be passed around wherever those are needed.

FA-tvonmoll:
@shs96c ah, yeah; I saw that IE options was the only one extending `DesiredCapabilities` (and left it be for the sake of not ripping up anything that already exists)

I'm not sold on the whole `addTo(...)` thing; are you saying you'd rather see all the browser options extend caps (and merge as ie options was)?

FA-tvonmoll:
Closing this re: a selenium channel discussion with Simon (the intent moving forward is going to be to make all of these `Capabilities` imps). This can be re-opened if that changes.

PR #4272, opened 2017-07-07
Add minimize window command in selenium python webdriver

Sl0v3C:
Symptom: There is no minimize window command
Root Cause: No this command
Solution: Add the command
Project: selenium python webdriver
Note: Already verified in local side with webdriver-rust & geckodriver & firefox(marionette) patches


Sl0v3C:
About firefox(marionette) & geckodriver parts patches now is under reviewing. 
webdriver-rust part already merged.

PR #4278, opened 2017-07-07
allow specifying geckodriver and marionette port for Firefox webdriver

vortic:
Adds the option to specify geckodriver and marionette port in the `webdriver.Firefox` constructor. The Service already has support for the geckodriver port, but the webdriver has no way to change it. This Pull Request adds that functionality, as well as the ability to specify the marionette port in the Service.

### DIFF
--- a/py/selenium/webdriver/firefox/service.py
+++ b/py/selenium/webdriver/firefox/service.py
@@ -30,9 +30,10 @@ class Service(service.Service):
         protocol to Marionette.
 
         :param executable_path: Path to the GeckoDriver binary.
-        :param port: Run the remote service on a specified port.
-            Defaults to 0, which binds to a random open port of the
-            system's choosing.
+        :param port: (Deprecated) This parameter should be passed as part
+            of service_args: ["--port", port].
+            Run the remote service on a specified port. Defaults to 0,
+            which binds to a random open port of the system's choosing.
         :param service_args: Optional list of arguments to pass to the
             GeckoDriver binary.
         :param log_path: Optional path for the GeckoDriver to log to.
@@ -43,12 +44,26 @@ class Service(service.Service):
         """
         log_file = open(log_path, "a+") if log_path is not None and log_path != "" else None
 
+        if port != 0:
+            import warnings
+
+            warnings.warn(
+                "The 'port' parameter has been deprecated. Please pass in the port through "\
+                "service_args instead: ['--port', port]",
+                DeprecationWarning)
+        if service_args is not None and "--port" in service_args:
+            if port != 0:
+                raise ValueError("Deprecated parameter 'port' specified "\
+                    "simultaneously with '--port' argument in service_args")
+            else:
+                port = int(service_args[service_args.index("--port") + 1])
+
         service.Service.__init__(
             self, executable_path, port=port, log_file=log_file, env=env)
         self.service_args = service_args or []
 
     def command_line_args(self):
-        return ["--port", "%d" % self.port] + self.service_args
+        return self.service_args + ([] if "--port" in self.service_args else ["--port", "%d" % self.port])
 
     def send_remote_shutdown_command(self):
         pass
--- a/py/selenium/webdriver/firefox/webdriver.py
+++ b/py/selenium/webdriver/firefox/webdriver.py
@@ -55,7 +55,7 @@ class WebDriver(RemoteWebDriver):
     def __init__(self, firefox_profile=None, firefox_binary=None,
                  timeout=30, capabilities=None, proxy=None,
                  executable_path="geckodriver", firefox_options=None,
-                 log_path="geckodriver.log", port=0, service_args=None):
+                 log_path="geckodriver.log", service_args=None):
         """Starts a new local session of Firefox.
 
         Based on the combination and specificity of the various keyword
@@ -99,9 +99,6 @@ class WebDriver(RemoteWebDriver):
             defaults to picking up the binary from the system path.
         :param firefox_options: Instance of ``options.Options``.
         :param log_path: Where to log information from the driver.
-        :param port: Run the remote service on a specified port.
-            Defaults to 0, which binds to a random open port of the
-            system's choosing.
         :param service_args: Optional list of arguments to pass to the
             GeckoDriver binary.
 
@@ -143,7 +140,7 @@ class WebDriver(RemoteWebDriver):
 
         if capabilities.get("marionette"):
             capabilities.pop("marionette")
-            self.service = Service(executable_path, log_path=log_path, port=port, service_args=service_args)
+            self.service = Service(executable_path, log_path=log_path, service_args=service_args)
             self.service.start()
 
             capabilities.update(firefox_options.to_capabilities())


p0deje:
I'm adding a label so that Python bindings maintainer will have a look, however I suppose they will ask to add some tests.

lmtierney:
We'll have to have @AutomatedTester weigh in, but I think the geckodriver port setting was left private for a reason. If you're starting geckodriver beforehand and connecting to it, you can use `Remote` webdriver and pass the url (say you started geckodriver on port 5678, use something like `127.0.0.1:5678` as the remote url).

That being said, I'm not seeing a way to pass `service_args` to geckodriver while instantiating a `Firefox` webdriver in order to specify a marionette port. The `Firefox` webdriver should probably allow a `service_args` parameter and pass it to the Service https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/firefox/webdriver.py#L141

vortic:
> If you're starting geckodriver beforehand and connecting to it, you can use `Remote` webdriver and pass the url...

Thanks for the explanation! In my case, however, I need to spin up a geckodriver instance programmatically, and I am running into race conditions where the same marionette port is trying to used for two geckodriver instances.

To be safe, I would also like to specify the geckodriver port, since I know how many simultaneous instances I have up front.

> If we allowed that, we could also grab the port from the service args during service instantiation if we wanted to allow that...

Good point, I will try that path instead. New commit incoming!

lmtierney:
With your new commit, for the port to take effect, you'd have to override the port setting from the `service_args` before starting the service:
https://github.com/SeleniumHQ/selenium/pull/4278/commits/d1fa60b43cd0983d370a28eb4163b3186cc3f35e#diff-6471acca09a7daf147956a90cae654f7R46

And also probably then remove the port from the `service_args` so that it isn't double defined in the `command line args`

vortic:
In this commit, `service_args` is not intended to contain information about the geckodriver port. This reflects the structure in `service.py`, where it has a separate `port` field from `service_args`.

A sample usage I had in mind would be
~~~python
driver = webdriver.Firefox(port=49152, [service_args=["--marionette-port", "%d" % 49153])
~~~

vortic:
Should I instead change both `service.py` and `webdriver.py` so that the port field is baked into `service_args`?

lmtierney:
Well, having `port` and `service_args` is redundant, since port is a service arg; unless you're looking for readability. What if someone specifies a port AND the port in `service_args`?  At this point, I don't know what would be the best direction to go. Hopefully @AutomatedTester or @davehunt have a suggestion?

davehunt:
I would suggest deprecating the `port` kwarg in `selenium.webdriver.firefox.service.Service` and only pass `service_args` from `selenium.webdriver.firefox.webdriver.WebDriver`.

vortic:
@davehunt is it sufficient to remove the `port` argument from `service.py`, or should I still allow its use, but print something along the lines of, "WARNING: 'port' argument has been deprecated. Please pass it as an argument in 'service_args' instead"?

If so, should the behavior be that the deprecated `port` argument will override any "--port" argument passed to `service_args`?

davehunt:
Don't remove the existing functionality, but you can mark it as deprecated. You should include details of how to use the `service_args` keyword instead so anyone currently using `port` will know how to update their code. I feel that any `--port` passed to `service_args` should take precedent over `port`, or you could simply raise an exception if both are used - this would also encourage users to not use the deprecated `port` argument.

py/selenium/webdriver/firefox/service.py
```diff
@@ -43,12 +44,26 @@ def __init__(self, executable_path, port=0, service_args=None,
         """
         log_file = open(log_path, "a+") if log_path is not None and log_path != "" else None
 
+        if port != 0:
+            import warnings
```
vortic: this structure for deprecation warning was taken from `firefox_profile.py`

if it is not correct, I am happy to change it!

py/selenium/webdriver/firefox/service.py
```diff
@@ -43,12 +44,26 @@ def __init__(self, executable_path, port=0, service_args=None,
...
+            warnings.warn(
+                "The 'port' parameter has been deprecated. Please pass in the port through "\
+                "service_args instead: ['--port', port]",
+                DeprecationWarning)
+        if service_args is not None and "--port" in service_args:
+            if port != 0:
+                raise ValueError("Deprecated parameter 'port' specified "\
+                    "simultaneously with '--port' argument in service_args")
+            else:
+                port = int(service_args[service_args.index("--port") + 1])
```
davehunt: hmm.. this is getting messy.. I hadn't realised that we're generating a port in the Python code based on the available ports, as geckodriver should accept 0 and bind that to an open port (though I'm not sure how we would determine the port it picks).

perhaps the best solution for now is to pass both `port` and `service_args` as you had, and raise an exception if `port` is not 0 and `--port` is in `service_args`.

I'd really like to hear @AutomatedTester's input on this though.

davehunt:
I'm wondering if being able to pass the geckodriver and marionette ports is the correct solution here. I'm not suggesting we shouldn't allow service arguments to be passed, but if there's an issue with the ports being assigned automatically then I feel that should be addressed directly. Could you describe the issues you're seeing, and perhaps we can fix this rather than you having to workaround it by managing you're own ports.

vortic:
@davehunt I was running into some sort of race condition where two instances of marionette would try to open on the same port, resulting in an `NS_ERROR_SOCKET_ADDRESS_IN_USE` error. As far as I could tell, geckodriver would never run into this race condition and would always open on a different free port. 

More insidiously, using python's `Process` to spin up multiple webdriver instances would cause the entire script to hang.

davehunt:
@andreastt is it possible that two instances of geckodriver started at the same time could use the same port for Marionette? If so, is there anything we can do to prevent this, as it's going to be a common issue and it's not reasonable to require the user to specify the ports to use.

andreastt:
Also sprach Dave Hunt:

> [I]s it possible that two instances of geckodriver started at the
> same time could use the same port for Marionette? If so, is there
> anything we can do to prevent this, as it's going to be a common
> issue and it's not reasonable to require the user to specify the
> ports to use.

How are you invoking geckodriver?  If you use the --marionette-port
flag for two processes, it is certainly the case that the process
that looses the race will be unable to bind.

It is also theoretically possible that geckodriver runs into a race
condition because it atomically acquires a free port for Marionette,
which it then releases, writes to the marionette.port preference,
and then asks Firefox to use.

We can remove the latter race condition by setting the
marionette.port preference to 0 and having geckodriver read the port
from the profile that the system has atomically allocated.

However, both these scenarios are so unlikely to happen that I
don’t think they are the cause of NS_ERROR_SOCKET_ADDRESS_IN_USE.

vortic:
> How are you invoking geckodriver? If you use the --marionette-port flag for two processes, it is certainly the case that the process that looses the race will be unable to bind.

In my case, I know how many instances of geckodriver I need a priori, so I can specify a range of ports to use, thus avoiding any possible race conditions. However, when I do not specify the ports explicitly, every now and then, I would get the "port in use" error.

andreastt:
Also sprach Victor:

> However, when I do not specify the ports explicitly, every now and
> then, I would get the "port in use" error.

It would be good to know which one of the scenarios I described you
are hitting.  If you have multiple geckodriver processes running
on the same system, I assume you will also have started these on
separate ports.

For each geckodriver two ports are allocated: one for the HTTPD
to serve the WebDriver API, and one firefox remote protocol
(Marionette) that geckodriver proxies calls to.

vortic:
@andreastt when I run into (what I assume is) the race condition, I am not specifying any port to use

p0deje:
Sorry for jumping in, but I guess it might be similar to the issue with ChromeDriver - https://bugs.chromium.org/p/chromedriver/issues/detail?id=1354. IIRC What ChromeDriver does when it picks remote port for Chrome, it just binds 0 port, saves returned port it variable and then releases. At some point later it starts Chrome with `--remote-debugging-port` flag set to the port that was found before.

There is a race condition hidden here. Since there is a small amount of time between finding port and actually using it, it's possible that another process might pick the same port as it's currently free.

I guess something similar can happen with GeckoDriver, so some sort of synchronization would be useful. I'm currently using socket lock from Ruby bindings to force sequential startup and I guess GeckoDriver could implement similar synchronization mechanism.

```ruby
Selenium::WebDriver::SocketLock.new(9510, 300).locked do
  driver = Selenium::WebDriver.for(:chrome)
end
```

andreastt:
Also sprach Alex Rodionov:

> There is a race condition hidden here. Since there is a small
> amount of time between finding port and actually using it, it's
> possible that another process might pick the same port as it's
> currently free.

Indeed.

> I guess something similar can happen with GeckoDriver, so some
> sort of synchronization would be useful. I'm currently using
> socket lock from Ruby bindings to force sequential startup and
> I guess GeckoDriver could implement similar synchronization
> mechanism.

It can happen.

Like I said, geckodriver can ask Marionette to bind to port
0: Marionette writes the allocated port to the marionette.port
preference, which geckodriver can then read from the profile.  This
would eliminate the race condition.

It would also be possible to use a named Unix domain socket, but
this is currently not supported in the XPCOM socket interface that
the Marionette server uses.  A third option is to reverse the
relationship of the client server.  This is further discussed in
https://bugzilla.mozilla.org/show_bug.cgi?id=1240830.

However, we did not start writing the port to a preference until
recently in https://bugzilla.mozilla.org/show_bug.cgi?id=1358988.
Because the next release of geckodriver will target Firefox 55 and
greater, we are now in a position to make this change.

I’ve filed https://bugzilla.mozilla.org/show_bug.cgi?id=1388465 to
make this change.

vortic:
Sorry for the hiatus, but I wanted to circle back to this conversation. The thread @andreastt mentioned did not seem to have a conclusion, but it looks like the necessary changes were made to `firefox/webdriver.py` in the master branch to allow `service_args` to be passed in, thus allowing the marionette port to be specified. I'm going to close this pull request, since my original need is met. Thanks, all!

PR #4285, opened 2017-07-09
fix file descriptor(/dev/null) leak in chrome webdriver

kumagi:

The file descriptor has been leaked in `WebDriver::Chrome::Service` opened here line.
https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/chrome/service.rb#L43

This leak cause `Errno::EMFILE: Too many open files` exception in massive test.
We have to close this file descriptor.

To error reproduce. (environment: ulimit -a => 1024)
```rb:
2000.times {|n|
  d = Selenium::WebDriver.for :chrome  #=> Errno::EMFILE: Too many open files
  d.quit
}
```

Thank you.

### DIFF
--- a/rb/lib/selenium/webdriver/chrome/service.rb
+++ b/rb/lib/selenium/webdriver/chrome/service.rb
@@ -45,6 +45,12 @@ module Selenium
           @process.start
         end
 
+        def stop_server
+          @process.io.stdout.close unless @process.io.stdout.closed?
+          return if process_exited?
+          connect_to_server { |http| http.get('/shutdown') }
+        end
+
         def cannot_connect_error_text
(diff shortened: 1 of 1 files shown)


p0deje:
Thank you for a bug report and pull request!

The proposed fix fails when ruby is run in debug mode, because in this case we try to pipe child process output to logger IO object. So, now when child process stops it closes its IO which makes logger IO object closed too, so all subsequent attempts to write to logger fail.

I'll push a different fix shortly.

p0deje:
I've pushed a fix, can you please try with it?

kumagi:
Thank you very much. I'll try it.

Issue #4268, opened 2017-07-06
I can't await session for safari


donaldpipowitch:
## Meta -
OS:  
OSX
Selenium Version:  
3.4.0
Browser:  
Safari

## Expected Behavior -

```js
console.log(new Builder().forBrowser('chrome').build().then); // [Function: bound then]
console.log(new Builder().forBrowser('safari').build().then); // [Function: bound then]
```

## Actual Behavior -

```js
console.log(new Builder().forBrowser('chrome').build().then); // [Function: bound then]
console.log(new Builder().forBrowser('safari').build().then); // undefined !!!
```

## Steps to reproduce -

```js
const { Builder } = require('selenium-webdriver');
console.log(new Builder().forBrowser('chrome').build().then);console.log(new Builder().forBrowser('safari').build().then);
```

Maybe related to this commit: https://github.com/SeleniumHQ/selenium/commit/b390490dc93fc77284c484dc06d6e5fcc28d0468

robin-drexler:
We ran into the same issue, I think. PR is here: https://github.com/SeleniumHQ/selenium/pull/4289

In the meantime, we fixed it by patching WebDriver, but only when it's actually needed (when a remote browser is used)

```js
const patchWebDriver = () => {
  const originalCreateSession = webdriver.WebDriver.createSession;
  webdriver.WebDriver.createSession = function (
    executor,
    capabilities,
    optFlow,
    optCtor,
    optOnQuit
  ) {
    return originalCreateSession(
      executor,
      capabilities,
      optFlow,
      optCtor || this,
      optOnQuit
    );
  };
};
```

```js
// somewhere in your code where you know if it's a remote browser
patchWebDriver();
```

donaldpipowitch:
Cool, thank you.

jleyba:
I'm not able to reproduce this.

```shell
$ npm install selenium-webdriver
/Users/jleyba/Development/scratch
└── selenium-webdriver@3.4.0 
$ node <<EOF
> const { Builder } = require('selenium-webdriver');
> console.log(new Builder().forBrowser('chrome').build().then);console.log(new Builder().forBrowser('safari').build().then);
> EOF
[Function: bound then]
[Function: bound then]
```

PR #4289, opened 2017-07-10
[js] fix driver.build() not returning thenable driver except for Chrome and Firefox

robin-drexler:
fixes https://github.com/SeleniumHQ/selenium/issues/4268

[ChromeDriver.createSession](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/chrome.js#L719) works, because it passes its reference to the super `createSession` and hence the reference to the `thenableWebDriverProxy` persists.

The issue seems to be that [WebDriver.createSession](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/lib/webdriver.js#L765) does not pass itself and therefore the reference is lost and `WebDriver` is used instead of the thenable version. The returned driver does not have a `then` method and causes errors.

### DIFF
--- a/javascript/node/selenium-webdriver/index.js
+++ b/javascript/node/selenium-webdriver/index.js
@@ -133,6 +133,21 @@ function createDriver(ctor, ...args) {
   if (!thenableWebDriverProxy) {
     /** @implements {ThenableWebDriver} */
     thenableWebDriverProxy = class extends ctor {
+      static createSession(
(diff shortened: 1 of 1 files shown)


jleyba:
Thank you for the contribution but this is not needed. See e4696a29290beb8db539285681b6c67534940842

PR #4310, opened 2017-07-14
Add minimize window command in selenium python webdriver

Sl0v3C:
Symptom: There is no minimize window command
Root Cause: No this command
Solution: Add the command
Project: selenium python webdriver
Note: Already verified in local side with webdriver-rust & geckodriver & firefox(marionette) patches

### DIFF
--- a/py/selenium/webdriver/remote/command.py
+++ b/py/selenium/webdriver/remote/command.py
@@ -101,6 +101,7 @@ class Command(object):
     EXECUTE_ASYNC_SCRIPT = "executeAsyncScript"
     SET_SCRIPT_TIMEOUT = "setScriptTimeout"
     SET_TIMEOUTS = "setTimeouts"
+    W3C_MINIMIZE_WINDOW = "w3cMinimizeWindow"
     MAXIMIZE_WINDOW = "windowMaximize"
     W3C_MAXIMIZE_WINDOW = "w3cMaximizeWindow"
     GET_LOG = "getLog"
--- a/py/selenium/webdriver/remote/remote_connection.py
+++ b/py/selenium/webdriver/remote/remote_connection.py
@@ -368,6 +368,8 @@ class RemoteConnection(object):
                 ('POST', '/session/$sessionId/window/rect'),
             Command.GET_WINDOW_RECT:
                 ('GET', '/session/$sessionId/window/rect'),
+            Command.W3C_MINIMIZE_WINDOW:
+                ('POST', '/session/$sessionId/window/minimize'),
             Command.MAXIMIZE_WINDOW:
                 ('POST', '/session/$sessionId/window/$windowHandle/maximize'),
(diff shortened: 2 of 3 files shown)


py/selenium/webdriver/remote/command.py
```diff
@@ -101,6 +101,7 @@ class Command(object):
     EXECUTE_ASYNC_SCRIPT = "executeAsyncScript"
     SET_SCRIPT_TIMEOUT = "setScriptTimeout"
     SET_TIMEOUTS = "setTimeouts"
+    W3C_MINIMIZE_WINDOW = "w3cMinimizeWindow"
```
lmtierney: No need for this to be `W3C_` since there is no JWP equivalent

py/selenium/webdriver/remote/webdriver.py
```diff
@@ -619,6 +619,14 @@ def window_handles(self):
         else:
             return self.execute(Command.GET_WINDOW_HANDLES)['value']
 
+    def minimize_window(self):
+        """
+        Miniimizes the current window that webdriver is using
+        """
+        if self.w3c:
```
lmtierney: This `if` is not necessary

py/selenium/webdriver/remote/remote_connection.py
```diff
@@ -368,6 +368,8 @@ def __init__(self, remote_server_addr, keep_alive=False, resolve_ip=True):
                 ('POST', '/session/$sessionId/window/rect'),
             Command.GET_WINDOW_RECT:
                 ('GET', '/session/$sessionId/window/rect'),
+            Command.W3C_MINIMIZE_WINDOW:
```
lmtierney: Update after command rename

Sl0v3C: THX for your review.
Please help to check the commit I closed earlier.
https://github.com/SeleniumHQ/selenium/pull/4272
And I already push commits for minimize command in webdriver-rust part which already merged, geckodriver & marionette part which are under reviewing.
The PR#4272 is just like the implementation of the maximize command.
THX again!

lmtierney: Why am I looking at #4272? You closed it in favor of this one which I reviewed

Sl0v3C: I closed the #4272 because  the AppVeyor build failed. So I thought maybe the commit should be more simply. So I removed the MINIMIZE_WINDOW part. However it still failed.
So I want to know the #4272 commit is OK or not. If OK, I will push it again and add a test for it.
THX!

lmtierney: The appveyor build fail is not due to your changes. I would prefer you update this PR with changes as opposed #4272 as that one would require more changes.

Issue #4307, opened 2017-07-13
Run As different user in ChromeDriverService service


shlathsaheb:
Can you add the run as different user option with chrome driver when we start the chrome process?

p0deje:
Can you be more specific? What are you trying to achieve? What language do you use?

shlathsaheb:
@p0deje : We are using C#.net. We follow windows auth. so would like to open chrome with different user credentials in same test to verify our authorization test cases. To be able to achieve that we need ability to run chrome as another user.
Currently chrome process start is abstract inside driverservice, so if possible in chrome driver we get an option to have user credentials as well to run chrome as another user?
Or if you have any other solution to the above described problem kindly share.

p0deje:
I am not sure how this is implemented, but can't you run your script as a different user?

jimevans:
We won't be adding a Windows-only feature to a cross-platform API. You can have your test runner run the entire test as a different user, or you can customize a subclass of `ChromeDriverService` to accomplish this.

NehaBhatia28:
Hi,
I am using selenium remote webdriver and hitting the application on Single sign-on enabled remote machine.My application get login with the same user which I have used during login into remote machine. I want to run the application with different users while running tests but I am not able to do due to single sign-on.It is not giving working but I am not able to switch the user.it is always logged-in with the user which I used for login in the remote machine. The code looks like below:

WebDriver driver = null;
        ChromeOptions options = new ChromeOptions();
        options.addArguments(TEST_TYPE);
        options.addArguments(DISABLE_POPUP_BLOCKING);
        if(nodeName.equals("Level2"))
            options.addArguments("user-data-dir=C:/Users/Level2/AppData/Local/Google/Chrome/User Data/Default");
        else if(nodeName.equals("Level1"))
            options.addArguments("user-data-dir=C:/Users/Level1/AppData/Local/Google/Chrome/User Data/Default");
        DesiredCapabilities cap = DesiredCapabilities.chrome();
        cap.setCapability(ChromeOptions.CAPABILITY, options);
        cap.setBrowserName("chrome");
        cap.setCapability("applicationName", nodeName);
        String Node;
        if (System.getProperty("hubUrl") == null){
            Node = "http://" + hubIpaddress + ":" + hubPort + "/wd/hub";
        }else{
            Node = System.getProperty("hubUrl", "http://localhost:4444/wd/hub");
        }     
        System.out.println("---Node----"+Node);
        try {
[...]

PR #4354, opened 2017-07-23
Support for DriverService "run as" username/password

nbelyh:

Motivation

When creating tests for Windows, there is occasionally a need to run the same test, while switching between different users (like, rung line 1-10 using windows user A, then run line 11-20 using user B, etc).
This is useful for testing workflow applications for example, with **integrated** windows security.

To be able to properly test this, one needs to logon different users and re-launch the driver using those. This pull request adds support for launching driver under a specific user (the underlying OS support is already there, it just exposes those fields as optional properties).

We needed this functionality in our project. 
Other examples where it may be needed (from stackoverflow):

https://stackoverflow.com/questions/16590207/how-to-impersonate-a-specific-user-with-selenium-webdriver
https://stackoverflow.com/questions/28045870/how-can-i-run-internet-explorer-selenium-tests-as-a-specific-domain-user
https://stackoverflow.com/questions/9951377/using-impersonation-to-launch-processes-indirectly-in-c-sharp
https://sqa.stackexchange.com/questions/2277/using-selenium-webdriver-with-windows-authentication

Usage example:

            var service = ChromeDriverService.CreateDefaultService();

            service.StartupDomain = "contoso";

            service.StartupUserName = "john.doe";

            var pwd = new SecureString();
            foreach (var c in "P@ssword")
                pwd.AppendChar(c);

            service.StartupPassword = pwd;

            service.Start();

            var options = new ChromeOptions();
            options.AddArgument("--no-sandbox");

            var driver = new RemoteWebDriver(service.ServiceUrl, options);
`

### DIFF
--- a/dotnet/src/webdriver/DriverService.cs
+++ b/dotnet/src/webdriver/DriverService.cs
@@ -21,6 +21,7 @@ using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Net;
+using System.Security;
 using System.Security.Permissions;
 using OpenQA.Selenium.Internal;
 using OpenQA.Selenium.Remote;
@@ -37,6 +38,12 @@ namespace OpenQA.Selenium
         private int driverServicePort;
         private bool silent;
         private bool hideCommandPromptWindow;
+
+        private string startupDomain;
+        private string startupUserName;
+        private SecureString startupPassword;
+        private bool startupLoadUserProfile;
+
         private Process driverServiceProcess;
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


p0deje:
I'm feeling this is going to be closed just like #4307, but I'll keep this open so that Jim decides.

nbelyh:
I understand the motivation on windows-only feature. But it's .net - so it's about windows basically. If this is a blocker, it would be nice to have the ability to hook into filling the StartupInfo at least, or have the Start() method protect virtual, so that it can be overriden. Now it's private.

jimevans:
You're right, we won't be adding a Windows-only feature to a cross-platform API, even knowing about .NET's (no longer truly applicable) Windows-only roots. However, you're right, there should be an extension mechanism here, and there's not one. I can give no time frame for implementation or release, because my time to devote to Selenium is severely limited for the next few months, but I'll add it to my list of things to implement.

p0deje:
Then I guess PR which adds extension mechanism is more then welcome!

amirse80:
Hi,
Any new on this?
This is like a show stopper for our tests. We use selenium and c#.

SandipMehta:
I am also having the same issue. VSTS running selenium tests in C# and the browser is always instantiated with the current user on the machine. I don't have a solution, please suggest if there is any workaround.

nbelyh:
@SandipMehta In the latest version of Selenium (4.x) there is a workaround, you can specify impersonation in the start up configuration (with event handler "DriverProcessStarting"

```
        static void Main(string[] args)
        {
            var svc = ChromeDriverService.CreateDefaultService();
            svc.DriverProcessStarting += Svc_DriverProcessStarting;
       }

        private static void Svc_DriverProcessStarting(object sender, OpenQA.Selenium.DriverProcessStartingEventArgs e)
        {
            var pwd = new SecureString();
            foreach (var c in "P@ssword")
                pwd.AppendChar(c);

            e.DriverServiceProcessStartInfo.UserName = "john.doe";
            e.DriverServiceProcessStartInfo.Password = pwd;
        }

PR #4363, opened 2017-07-25
Fix Firefox location detection on Windows_x64

kamenlitchev:
When searching for firefox.exe Selenium would not properly detect Firefox x64 installations under 32-bit Ruby on x64 Windows (10?). For Windows x64 with Ruby x86 the current code would always detect only "C:\Program Files (x86)", since both 'PROGRAMFILES' and 'ProgramFiles(x86)' ENV variables dynamically map to "C:\Program Files (x86)" because of the Ruby x86 bitness. Due to that Firefox x64 would not be auto-detected with error "Selenium::WebDriver::Error::WebDriverError: Could not find Firefox binary (os=windows). Make sure Firefox is installed or set the path manually with Selenium::WebDriver::Firefox::Binary.path=". 

The Firefox x64 executable is actually located under "C:\Program Files' but Selenium would look only under "C:\\Program Files (x86)".

The fix is to introduce a new environment variable poll for "ProgramW6432" (https://msdn.microsoft.com/en-us/library/windows/desktop/aa384274(v=vs.85).aspx) which will return the proper "\\Program Files" directory on a Windows x64 machine with Ruby x86.

To easily check that solution works:
- On Windows x64 machine with Ruby x86 run 'irb'
- Type `` `set` `` and observe the presence of ProgramW6432

### DIFF
--- a/rb/lib/selenium/webdriver/common/platform.rb
+++ b/rb/lib/selenium/webdriver/common/platform.rb
@@ -161,7 +161,8 @@ module Selenium
       def find_in_program_files(*binary_names)
         paths = [
           ENV['PROGRAMFILES'] || '\\Program Files',
-          ENV['ProgramFiles(x86)'] || '\\Program Files (x86)'
+          ENV['ProgramFiles(x86)'] || '\\Program Files (x86)',
+          ENV['ProgramW6432'] || '\\Program Files'
         ]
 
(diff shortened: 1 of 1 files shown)


p0deje:
Merged in 049761cdb32840a932422b7fb2619ea91337edbd. Thank you for contribution!

PR #4372, opened 2017-07-26
[docs] API doc comment formatting

olleolleolle:
This simple PR changes a Ruby API docs comment, to avoid a warning which is output when generating it.

The problem was that the `@param` was not _named_, and this PR names it.

(Yes, I have signed the CLA.)


lmtierney:
Thanks!

PR #4378, opened 2017-07-26
Null cookie value fix

granak:
Issue found during testing using Edge while cookie value
was empty - interpreted as null. Getting string of null was
failing.

### DIFF
--- a/dotnet/src/webdriver/Cookie.cs
+++ b/dotnet/src/webdriver/Cookie.cs
@@ -222,7 +222,7 @@ namespace OpenQA.Selenium
             }
 
             string name = rawCookie["name"].ToString();
-            string value = rawCookie["value"].ToString();
+            string value = rawCookie["value"] != null ? rawCookie["value"].ToString() : string.Empty;
 
             string path = "/";
             if (rawCookie.ContainsKey("path") && rawCookie["path"] != null)


jimevans:
I'm not sure this is the correct behavior. If the response to the `getCookie` command is returning `null` as the cookie value, does that mean the cookie is undefined in the page, or it exists with an actual value of "null"? If it exists in the page, should the value be `null` or the string "null"? It's unclear what the correct behavior is for this case.

granak:
Well, I'm not 100% sure either.
But as far as I know, cookie-pairs (name and value) are represented as strings. So while cookie-pairs 'name=;' and 'name="";' are equal (values are represented by common browsers as empty strings), 'name="null";' is not same as 'name=;'.

ref: [RFC 6265](https://tools.ietf.org/html/rfc6265#section-5.2)
> A user agent MUST use an algorithm equivalent to the following
>    algorithm to parse a "set-cookie-string":

>       1. If the set-cookie-string contains a %x3B (";") character:
>           The name-value-pair string consists of the characters up to,
>           but not including, the first %x3B (";"), and the unparsed-
>           attributes consist of the remainder of the set-cookie-string
>           (including the %x3B (";") in question).

>        Otherwise:

>           The name-value-pair string consists of all the characters
>           contained in the set-cookie-string, and the unparsed-
>           attributes is the empty string.

granak:
Hello @jimevans, any progress about this?

granak:
Hello @jimevans,
any progress about this?

granak:
Hello @jimevans @p0deje 
Is anyone reviewing this? What's the status?

p0deje:
It's only Jim who maintains .NET bindings and I'm not sure how long it will take for him to go through all the PRs that need review.

jimevans:
Merged in 2c4f935. Thank you for the contribution.

PR #4383, opened 2017-07-27
Allow Ruby to properly handle remote error codes

### LINKED ISSUE #4380: Backward compatible error codes
I haven't investigated to see if other bindings have this issue, so I'm marking it C-rb, but we should verify it is addressed on all bindings before closing it.

When running tests on Firefox through a remote server, the new w3c errors are getting translated in the server, but not properly handled in the bindings.

Essentially we need to implement all of the "new" error codes that have been added here: https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/remote/ErrorCodes.java

I'm also marking it as easy because this makes an excellent "fix a bug become a committer" item.

mxygem:
Pulled error codes from remote/ErrorCodes.java so that Ruby can handle
them appropriately when seen.

Fixes #4380

### DIFF
--- a/rb/lib/selenium/webdriver/common/error.rb
+++ b/rb/lib/selenium/webdriver/common/error.rb
@@ -298,7 +298,7 @@ module Selenium
       class << self
         def for_code(code)
           return if [nil, 0].include? code
-          return ERRORS[code] if code.is_a? Integer
+          return ERRORS.fetch(code) if code.is_a? Integer
 
           klass_name = code.split(' ').map(&:capitalize).join
           Error.const_get("#{klass_name.gsub('Error', '')}Error")


twalpole:
All of the newly added errors class names need to end in "Error" like the existing classes.  The mapping code depends on it.

jaysonesmith:
@twalpole That's easy enough to fix. This change was a mirroring of the [errors for Java](https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/remote/ErrorCodes.java#L93-L104). It's easy enough to change things up. Just a moment!

twalpole:
Also - a number of these error classes have already been defined - https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/common/w3c_error.rb#L192

jaysonesmith:
Ah, interesting! I had been following the request in #4380, but nice catch and thank you for pointing it out!  I can add error to the others and nix the dupes. For the new w3c errors 6063 `NoSuchCookieError` isn't defined in the w3c_errors.rb file. Would you like me to move it to that file and leave the rest in error.rb?

twalpole:
@jaysonesmith Not sure where the Selenium team wants them (not a member), I was just pointing out issues with your PR.

titusfortner:
Yes, they should have `Error` appended to the names which is the convention for the rest of the Ruby error classes.

Yup, there is duplication between `error.rb` & `w3c_error.rb` that shouldn't be there. I think I see what I was thinking 2 years ago, but it makes no sense to have two files loaded in sequence that define the same classes.

@jaysonesmith - content wise we just want all the content in `w3c_error.rb`, then the updated `ERRORS` array & `#for_code` from `error.rb`, all in one file. Ping me on Slack if you want to discuss details in more depth.

jaysonesmith:
@titusfortner That's easy enough to get done. Thanks for the guidance!

jaysonesmith:
In researching some of the errors that don't have numbered error codes for them to try to find some, it looks like the [JS/Node Selenium project has deleted the `InvalidSessionIdError`](https://github.com/SeleniumHQ/selenium/blob/18660c422ee132e56835f5762a7bc20f7b025b9f/javascript/node/selenium-webdriver/CHANGES.md) error in favor of `NoSuchSessionError.` @titusfortner Should we be doing the same?

rb/lib/selenium/webdriver/common/w3c_error.rb
```diff
@@ -199,6 +257,68 @@ class ElementNotInteractableError < WebDriverError; end
...
+        InvalidSelectorError,                 # 32
+        SessionNotCreatedError,               # 33
+        MoveTargetOutOfBoundsError,           # 34
+        InvalidXpathSelectorError,            # 51
+        InvalidXpathSelectorReturnTyperError, # 52
+        ElementNotInteractableError,          # 60
+        InvalidArgumentError,                 # 61
+        NoSuchCookieError,                    # 62
+        UnableToCaptureScreenError,           # 63
+        MethodNotAllowedError,                # 405
```
twalpole: As can be seen from the definition of `for_code` below, the errors are translated from their code by indexing into the ERRORS array which means having non-contiguous code numbers isn't going to work with the current setup.  ERRORS may have to become a Hash. @titusfortner

titusfortner:
Ruby throws `Error::ServerError` for any codes > 400, so I don't think we need to specify 405. I agree with @twalpole about re-implementing this array as a hash; should be able to just use the integers as the keys.

jaysonesmith:
Thanks @titusfortner can do. We'll get this sorted! Also noticed I missed a comma. D'oh!

jaysonesmith:
Looks like I may need to tweak some other files to handle the update given AppVeyor's failure.

rb/lib/selenium/webdriver/common/w3c_error.rb
```diff
@@ -199,6 +249,62 @@ class ElementNotInteractableError < WebDriverError; end
       UnexpectedJavascriptError = JavascriptError
       NoAlertOpenError          = NoAlertPresentError
       ElementNotDisplayedError  = ElementNotVisibleError
+
+      # @api private
+      ERRORS = {
+        1 => 'IndexOutOfBoundsError',
```
twalpole: These need to be classes not strings  -   `1 => IndexOutOfBoundsError`

jaysonesmith: Hah! That they do!

jaysonesmith:
@twalpole How do things look now?

titusfortner:
No, you did this backwards :(
We want to keep the `w3c_error` classes & descriptions (to match https://w3c.github.io/webdriver/webdriver-spec.html#handling-errors), and put them in the `error.rb` file and delete the `w3c_error.rb` file

jaysonesmith:
Got it! Easy enough to do the file change. Will go through the webdriver error page and update descriptions that are off.

jaysonesmith:
This should be better. (hopefully! https://media.giphy.com/media/l0NwNrl4BtDD7JCx2/giphy.gif )

rb/lib/selenium/webdriver/common/error.rb
```diff
@@ -60,57 +70,58 @@ class StaleElementReferenceError < WebDriverError; end # 10
       class ElementNotVisibleError < WebDriverError; end # 11
 
       #
-      # Raised when an interaction could not be performed because the element
-      # is in an invalid state (e.g. attempting to click a disabled element).
+      # A command could not be completed because the element is in an
+      # invalid state, e.g. attempting to click an element that is no
```
twalpole: I think the original description was more correct.  An element no longer attached to the document generates a StaleElementReferenceError

rb/lib/selenium/webdriver/common/error.rb
```diff
@@ -153,7 +164,7 @@ class IMENotAvailableError < WebDriverError; end # 30
       class IMEEngineActivationFailedError < WebDriverError; end # 31
 
       #
-      # Argument was an invalid selector (e.g. XPath/CSS).
+      # Argument was an invalid selector.
```
twalpole: Why change this description?

jaysonesmith: @twalpole They were changed to match the descriptions in the [WebDriver spec](https://w3c.github.io/webdriver/webdriver-spec.html#handling-errors) as per @titusfortner 's direction.

rb/lib/selenium/webdriver/common/error.rb
titusfortner: Good point - https://github.com/w3c/webdriver/issues/989 We'll see what is said.

jaysonesmith: It's been decided! 

`The target element is in an invalid state, rendering it impossible to interact with, for example if you click a disabled element.`

twalpole:
@titusfortner  While this is being changed, any interest in adding another level of inheritance for element specific errors  that may change on a retry like   'ElementNotVisibleError < ElementSpecificError < WebDriverError`?  Could apply to ElementNotVisibleError, InvalidElementStateError, ElementNotSelectableError, ElementNotInteractableError, ElementClickInterceptedError currently.

titusfortner:
@twalpole I see where you are going with this.
I just updated Watir to do a [more complicated rescue/retry](https://github.com/watir/watir/blob/master/lib/watir/elements/element.rb#L651-L673) on exceptions. We were previously making wire calls to verify the state of everything ahead of time, but that resulted in some pretty poor performance.

So I guess my concern is that the collection of things I'd want to rescue at the same time might be different from your collection, in which case it might make more sense for us to each specify individually.

rb/lib/selenium/webdriver/common/error.rb
titusfortner: @twalpole thanks for catching this, it's been declared a bug in w3c.
@jaysonesmith we'll update the text when the w3c language is approved/updated. (https://github.com/w3c/webdriver/pull/990)

jaysonesmith: @titusfortner Which description do you want to have in this PR, then?

twalpole:
@titusfortner I agree we may want to handle specific errors differently, but adding the extra level of inheritance wouldn't prevent that. What it would allow is a default handling of newly added error messages that probably want to just retry without having to keep checking each selenium update/release for newly added errors.

jaysonesmith:
@titusfortner Any other changes needed?

jaysonesmith:
@titusfortner Wanting to bump this and get it merged if possible. :) The failing checks don't appear to be due to my changes.

jaysonesmith:
Replacing this with #4912

PR #4389, opened 2017-07-31
Introduce IE::Options class

p0deje:
From now on all the customization of Internet Explorer driver
(e.g. cli args, native events, element scroll behavior, etc.)
should happen via this class. See documentation of this class for more
information.

Please note that Options class is designed to work with
IEDriverServer 3.5+ and is likely to behave incorrectly with
older versions.

Kudos to @lmtierney who started the implementation in https://github.com/lmtierney/selenium/tree/ie-options, I mostly finalized it.

### DIFF
--- a/rb/lib/selenium/webdriver/chrome/options.rb
+++ b/rb/lib/selenium/webdriver/chrome/options.rb
@@ -81,7 +81,7 @@ module Selenium
         end
 
         #
-        # Add a command-line argument to use when starting Chrome
+        # Add a command-line argument to use when starting Chrome.
         #
         # @example Start Chrome maximized
         #   options = Selenium::WebDriver::Chrome::Options.new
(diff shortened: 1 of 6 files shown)


luke-hill:
More of a query than a review comment: https://github.com/SeleniumHQ/selenium/pull/4389/files#diff-787c2280cf84bce161b914683e30518eR69 - Is this the planned structure inside files going forwards.

From my POV this is **really** useful, because when I now go into the gem code, I can essentially find examples. Something I've "often" struggled to do.

+1

p0deje:
> More of a query than a review comment: https://github.com/SeleniumHQ/selenium/pull/4389/files#diff-787c2280cf84bce161b914683e30518eR69 - Is this the planned structure inside files going forwards.

Yeah, we should have all such docs in place.

> From my POV this is really useful, because when I now go into the gem code, I can essentially find examples. Something I've "often" struggled to do.

Please keep in mind that this class can't be used with remote/Grid (only local IE). But I agree that it makes possible options easier to discover.

glib-briia:
Looks like Travis build stuck https://travis-ci.org/SeleniumHQ/selenium/builds/261801456

p0deje:
Merged in f943605da3c7aa7003560c1a8da5ccdcbba09755.

PR #4404, opened 2017-08-03
Added system property for driver start timeout.

### LINKED ISSUE #4394: Browser long starts - DriverService waitUntilAvailable timeout should be parameterized
## Meta -
OS: 
Windows10, Linux CentOs 7

Selenium Version:  
3.4.0
Browser:  
All

## Expected Behavior -
On some specific environment or after many tests, browsers can be a bit long to start (more than 30sec) so we must be able to set a custom timeout when Driver.waitUntilAvailable() function is called.
For example we can have a setLaunchTimeout(timout) function on WebDriver class or a configuration key.

## Actual Behavior -
Timeout is set to 20sec with no way to change it. We have firefoxes than starts in 30 or 40 sec after performing tests.

## Steps to reproduce -

Launch more than 5 firefoxes on the same computer, the 6th one should start in more than 20sec

Tom-Trumper:
Modified DriverService.waitUntilAvailable() method to utilise a new
system property 'webdriver.server.start.timeout' that can be used to
specify how long to wait for a driver server to start before checking
if it has errored, whilst maintaining a default value of 20 seconds.

Fixes #4394.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/4404)

### DIFF
--- a/java/client/src/org/openqa/selenium/remote/service/DriverService.java
+++ b/java/client/src/org/openqa/selenium/remote/service/DriverService.java
@@ -187,7 +187,7 @@ public class DriverService {
   protected void waitUntilAvailable() throws MalformedURLException {
     try {
       URL status = new URL(url.toString() + "/status");
-      new UrlChecker().waitUntilAvailable(20, SECONDS, status);
+      new UrlChecker().waitUntilAvailable(Long.getLong("webdriver.server.start.timeout", 20), SECONDS, status);
     } catch (UrlChecker.TimeoutException e) {
       if (process != null) {
         process.checkForError();


barancev:
I don't favour system properties. Let's make it more straightforward, with a method in DriverService.Builder

Tom-Trumper:
Fair enough, is there a particular reason why a system property is unsuitable for this? I am unsure how you would access a method in DriverService.Builder when using a RemoteWebDriver instance for example - perhaps as a capability?

barancev:
We have lots of system properties hidden all over the code that can be used to tune Selenium, but they are not documented and it bothers me. For API we have autogenerated javadoc at least.

barancev:
Yes, RemoteWebDriver is more tricky.

I think that the "correct" way to tune remote instances is node json config file. If you have a slow machine, you change its node file to increase the timeouts.

barancev:
In this case we'll have another kind of "autodocumentation" -- the default configuration file.

barancev:
Sorry, I forgot to close this PR it time. We resist to adding more "implicit" parameters because they are hard to remember and hard to maintain. This timeout parameter should be implemened in a more straightforward manner as a part of node config.

PR #4436, opened 2017-08-10
[js] Correct example code

jugglinmike:
Consistently await all Promise-returning API calls. Format code
according to `google_search_generator.js` (the equivalent example for
generator functions).

The current version technically works, although I'm not sure why. Note that [the "WebDriverJs" wiki page](https://github.com/SeleniumHQ/selenium/wiki/WebDriverJs) should also be corrected.

### DIFF
--- a/javascript/node/selenium-webdriver/example/async_await_test.js
+++ b/javascript/node/selenium-webdriver/example/async_await_test.js
@@ -57,8 +57,11 @@ describe('Google Search', function() {
   it('example', async function() {
     await driver.get('https://www.google.com/ncr');
 
-    await driver.findElement(By.name('q')).sendKeys('webdriver');
-    await driver.findElement(By.name('btnG')).click();
+    let q = await driver.findElement(By.name('q'));
+    await q.sendKeys('webdriver');
+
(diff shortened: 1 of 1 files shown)


jleyba:
The current version works because `driver.findElement(By.name('q'));` returns a [`WebElementPromise`](http://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/index_exports_WebElementPromise.html), so it can be used like any other promise.

`WebElementPromise` resolves to a plain `WebElement` - same basic API minus the promise methods (`then`, `catch`, etc).

jugglinmike:
I think I understand. Is it accurate to say that these two statements are
equivalent?

```js
webElementPromise.sendKeys('abc');

webElementPromise.then(() => webElementPromise.sendKeys('abc'));
```

jleyba:
Almost - you need the resolved webelement in your then callback:

```js
let webElementPromise = driver.findElement(By.name('q'));

webElementPromise.sendKeys('foo');
webElementPromise.then((e) => e.sendKeys('foo'));
```

jugglinmike:
Got it. Thanks for the clarification!

PR #4481, opened 2017-08-16
Issue WebDriver commands sequentially.

joshbruning:
Issue WebDriver commands sequentially.

#execute is synchronized to help protect against WebDriver commands
being issued in parallel. WebDriver commands should be executed
in sequence by the remote endpoints; however, not all drivers do this.
This is a pragmatic change that adds a measure to eliminate a cause
of observed test flakiness.

### DIFF
--- a/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java
+++ b/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java
@@ -628,7 +628,8 @@ public class RemoteWebDriver implements WebDriver, JavascriptExecutor,
     this.level = level;
   }
 
-  protected Response execute(String driverCommand, Map<String, ?> parameters) {
+  // This is synchronized to help protect against WebDriver commands being issued in parallel.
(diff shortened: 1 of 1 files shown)


jleyba:
It's long been our position that the WebDriver API is not thread safe and users that try to use it from multiple threads are going to have a bad time. I agree this is a minor change, but I'm worried about setting precedence for PRs related to thread safety.

joshbruning:
Thanks Jason - this is why I didn't commit this directly to the repo. If necessary, we can just maintain this in the Google fork.

shs96c:
I agree with @jleyba. Closing. I'm not quite sure what's happened in google's codebase to make this seem necessary, but I'd be looking closely at call sites to try and figure out why multiple threads are accessing an object that's explicitly not safe to use in multiple threads simultaneously.

joshbruning:
fwiw: A JavaScript unit test runner uses executeScript to periodically poll for test results. Users of that test runner exposed the WebDriver API so they could call it from their unit test.

Issue #4480, opened 2017-08-16
IE driver WebElement getLocation returns x=0 y=0


eugenedani:
## Meta -
OS:  Windows 10 (OS build 15063)
Selenium Version:  3.5.0, IE driver 3.5.0
Browser:  Internet Explorer 
Browser Version:  11.483.15063.0 (64-bit)

## Expected Behavior -
WebElement getLocation method returns actual element location

## Actual Behavior -
WebElement getLocation method returns x=0 y=0

## Steps to reproduce -
WebDriver webDriver = new InternetExplorerDriver();
webDriver.get("https://www.google.com/");
WebElement element = driver.findElement(By.name("q"));
Point location = element .getLocation();

location.x and location.y have value 0

barancev:
Reproduced.

jimevans:
Fixed in d25d966. Released in `IEDriverServer.exe` 3.5.1.

PR #4484, opened 2017-08-17
[rb] get specs passing on windows

titusfortner:
This gets local specs passing on Windows (build 16257) with production Firefox and latest Chrome, IE & Edge browsers and drivers.

The Edge stuff might be better otherwise implemented, but this gets it working. It needs to use the OSS commands for actions, but the way I have it toggling between OSS & W3C has been to override at `#execute` in the `COMMAND` hash, and all the interactions methods call the Bridge methods directly.

I can't get the remote specs working on Windows with current code to verify Ruby specs are passing there. :-/

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/4484)


rb/spec/integration/selenium/webdriver/element_spec.rb
```diff
@@ -145,39 +145,41 @@ module WebDriver
...
 
-      context 'size and location' do
-        it 'should get current location' do
-          driver.navigate.to url_for('xhtmlTest.html')
-          loc = driver.find_element(class: 'header').location
-
-          expect(loc.x).to be >= 1
-          expect(loc.y).to be >= 1
-        end
+      not_compliant_on browser: :ie do
```
p0deje: Any tip on why it's failing?

titusfortner: #4480 - Jim thinks it has to do with a current issue in the atoms. It'll probably get fixed quickly enough.

p0deje:
I've just pushed a commit which guards some of the failing specs for Edge. Is there any way you could rebase and see how it looks now? I guess the changes you did with Edge bridge should unguard some of the tests.

See 041c8159370bc9bb6b330c668a59ec9a3a777d38.

p0deje:
Actually, I'm just going to take the changes you did for Edge and cherry-pick them...

p0deje:
Hrm, it didn't turn out to be so simple, so I guess I'll leave that for you 😄

tf-prbuild:
Oh shoot, did I really never merge this? Will do.

p0deje:
I guess things have changed a lot since you last looked at Edge, is this still reasonable?

luke-hill:
As an update when I did some local selenium testing I did get edge working and passing some specs. I can't remember how many, but a few of them did run.

Issue #4491, opened 2017-08-17
selenium-webdriver examples broken by Google btnG element not clickable


mykmelez:
## Meta -
OS:  macOS 10.11.6
Selenium Version:  3.5.0
Browser:  Firefox, Chrome
Browser Version:  Firefox 55, Chrome 60
GeckoDriver Version: 0.18.0
ChromeDriver Version: 2.31.488774

## Expected Behavior -
The [selenium-webdriver README](https://github.com/SeleniumHQ/selenium/tree/master/javascript/node/selenium-webdriver#usage) inlines an example Node script that uses Firefox to load Google and search for "webdriver". Running that example script should succeed, and a variety of other scripts in the [example/ directory](https://github.com/SeleniumHQ/selenium/tree/master/javascript/node/selenium-webdriver/example) should similarly succeed.

## Actual Behavior -
The script fails with the cryptic error message "WebDriverError" in Firefox and only a slightly better error message "ElementNotVisibleError: element not visible" if I modify it to use Chrome instead of Firefox. A variety of other scripts in the example/ directory similarly fail.

## Steps to reproduce -
> npm install selenium-webdriver
> node node_modules/selenium-webdriver/example/google_search.js

After a bit of debugging, I discovered that the error is triggered by the statement `driver.findElement(By.name('btnG')).click()`, and Firefox appears to be returning the error message "element not interactable" (even though that doesn't get printed to my terminal, which is a different issue).

So it seems like both Firefox and Chrome see the "btnG" button on Google's home page as inaccessible (whether because "not visible" or "not interactable"). Presumably this is because of some change to that page that has broken these examples, not because of a regression in Selenium itself. Nevertheless, it manifests as broken examples in Selenium, which is why I'm filing it here.

jess010:
There appears to be some uncaught instances of this issue in the repo. Opening a PR to address by updating references to current name instead of return approach. The Selenium WebdriverJs Wiki page still needs updating also.

jess010:
For what it's worth, having reference to a button 'click' action using webdriver in the example on the Wiki, in addition to the previous 'sendKeys' command very helpful for getting up to speed with basic webdriver functionality when I was using this for a hackathon a month ago. 

Even though there is a risk of 'btnK' references becoming deprecated at some point in future also, I would urge you folks to keep it as a button click to better help new devs in future learn the basics.

PR #4493, opened 2017-08-17
simulate RETURN instead of button click on Google search page

mykmelez:

Here's a possible fix for #4491. Instead of finding and simulating a click on the "btnG" button on the Google search page, it simulates pressing the RETURN key in the input field after entering the search term. This works as expected in my testing with selenium-webdriver.

It's possible that there are examples elsewhere in the codebase that try to do the same thing and should also be updated. I only looked at the code in selenium-webdriver. I can expand my search/replace to the rest of the repository, I only hesitated because selenium-webdriver is the only client I have configured currently.


jleyba:
D'oh. Another case of me making code changes before checking PRs: f9aa0e2ebcd64f757f5f6723830cc8a70228332f

PR #4494, opened 2017-08-17
add ctor for ElementNotInteractableError

mykmelez:
The WebDriver spec <https://www.w3.org/TR/webdriver/#handling-errors>
defines the "element not interactable" error code, and geckodriver
returns that code in response to a command on an element that isn't
interactable, so selenium-webdriver should define a ctor for the error
and display that ctor instead of the generic WebDriverError when
printing errors to the console.


jleyba:
Sorry, I didn't see this PR and just added ElementNotInteractableError in 90cfe356742aaf80adef2316c5648e905fcb59f2

Issue #547, opened 2015-05-18
running selenium with pythonw.exe fails with invalid handle


ruidc:
running the following in a script via pythonw.exe:

```
from selenium import webdriver
profile = webdriver.FirefoxProfile()
webdriver.Firefox(firefox_profile=profile)
```

results in something like:

```
  File "zzz\selenium\webdriver\firefox\webdriver.py", line 59, in __init__
  File "zzz\selenium\webdriver\firefox\extension_connection.py", line 47, in __init__
  File "zzz\selenium\webdriver\firefox\firefox_binary.py", line 65, in launch_browser
  File "zzz\selenium\webdriver\firefox\firefox_binary.py", line 88, in _start_from_profile_path
  File "subprocess.pyc", line 702, in __init__
  File "subprocess.pyc", line 833, in _get_handles
  File "subprocess.pyc", line 884, in _make_inheritable
WindowsError: [Error 6] The handle is invalid
```

see also http://stackoverflow.com/questions/25203955/how-to-run-selenium-webdriver-with-pythonw

barancev:
Selenium 2.40 is out of date

ruidc:
? this is in 2.45.0

barancev:
So don't include links to discussion threads related to outdated versions without ever mentioning your own version :)

What is Firefox version? What is python version?
Can you reproduce the issue with python instead of pythonw?

ruidc:
Sorry, I thought you might have tried it first :)
Python 2.7.9 (default, Dec 10 2014, 12:28:03) [MSC v.1500 64 bit (AMD64)] on win32
Firefox is 38.0.1
no this problem is specific to pythonw.exe because of how it differs in handles for stdin/stdout/stderr

barancev:
@lukeis @AutomatedTester

barancev:
Selenium is not a GUI application, there is no reason to run it with pythonw instead of python. If one prefers to use pythonw one can redirect output streams.

PR #4503, opened 2017-08-18
Also define stdin or it will crash on Python + cx_Freeze: WindowsErro…

### LINKED ISSUE #536: Error while executing test with more than 5 chrome instances open in Selenium-Python
Hi,  I have a script where there are 8 test methods each opening a chrome browser and keeping it open (without driver.quit()). After the completion of 5 methods an error comes as below:

"Traceback (most recent call last):
  File "C:\Python34\lib\site-packages\selenium\webdriver\chrome\service.py", line 63, in start
    self.service_args, env=env, stdout=PIPE, stderr=PIPE)
  File "C:\Python34\lib\subprocess.py", line 754, in **init**
    _cleanup()
OSError: [WinError 6] The handle is invalid

During handling of the above exception, another exception occurred:

  File "..\SeleniumAutomation\common\common_settings.py", line 126, in setup_selenium
    driver = CommonSettings.setup_server(self.executionFlag)
  File "..\SeleniumAutomation\common\common_settings.py", line 168, in setup_server
    driver = CommonSettings.getdriver(executionFlag)
selenium.common.exceptions.WebDriverException: Message: ChromeDriver executable needs to be available in the path. Please download from http://chromedriver.storage.googleapis.com/index.html and read up at http://code.google.com/p/selenium/wiki/ChromeDriver"

Pleae let me know if there's any way to get around this because its causing many issues.

Thank You

melroy89:
Using Python + cx_Freeze will result into an exception:

> OSError: [WinError 6] The handle is invalid

This is because there is no stdin defined in the service.py file for the `subprocess.Popen()` command. Underwater the subprocess tries to create a handle which also looks for stdin under Windows this gets a bit tricky when using Bash or cx_Freeze. So, I defined `stdin` as well, and the crash is gone. Optionally you can also use:
```python
FNULL = open(os.devnull, 'r')
subprocess.Popen(.... ,stdin=FNULL)
``` 

Trying to fix #536 & #547 as well.

### DIFF
--- a/py/selenium/webdriver/common/service.py
+++ b/py/selenium/webdriver/common/service.py
@@ -71,7 +71,9 @@ class Service(object):
             cmd.extend(self.command_line_args())
             self.process = subprocess.Popen(cmd, env=self.env,
                                             close_fds=platform.system() != 'Windows',
-                                            stdout=self.log_file, stderr=self.log_file)
+                                            stdout=self.log_file,
+                                            stderr=self.log_file,
+                                            stdin=PIPE)
         except TypeError:
             raise
         except OSError as err:


lmtierney:
Merged, thank you!

PR #4512, opened 2017-08-19
Fixed Issue #4471

### LINKED ISSUE #4471: @FindAll Annotation searches for all locator infos
OS:  Windows 10
Selenium Version:  Selenium 3.4

@FINDALL annotation is supposed to match at least one of the given criteria.
I have following OR:
URL for WebPage : http://store.demoqa.com/
WebElement : -
@FINDALL ({
@findby(xpath = "//input[@name='s']"),
@findby(xpath = "//a[contains(.,'Magic Mouse')]")
})
public WebElement Header__txtSearch;

I am trying to verify element on page with multiple locators i.e if one fails other might return true.

But when I give the first one as Correct and second one as Incorrect , then it still returns true that element is found but it goes onto the second choice as well to find that element and waits for ObjectTimeout. **It should immediately return true after finding first match.**

kumar-nitish:
Fixed timeout on the second locator if you have already got an element by the first one

Fixes #4471

### DIFF
--- a/java/client/src/org/openqa/selenium/support/pagefactory/ByAll.java
+++ b/java/client/src/org/openqa/selenium/support/pagefactory/ByAll.java
@@ -49,11 +49,13 @@ public class ByAll extends By implements Serializable {
 
   @Override
   public WebElement findElement(SearchContext context) {
-    List<WebElement> elements = findElements(context);
-    if (elements.isEmpty()) {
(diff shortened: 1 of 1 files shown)


barancev:
Merged as 18028ac9f065b81fbe99a9e84feabfa5d0091f13
Tests have been fixed and improved too.

PR #4528, opened 2017-08-21
Update atoms to reflect Chrome TouchEvent support.

joshbruning:
Update atoms to reflect Chrome TouchEvent support.
The previous events.js logic breaks mobile Safari
because initTouchEvent.length == 0 on mobile Safari,
causing it to go erroneously down the Chrome/Android
path. This logic used to break Chrome Mobile
Emulation of ios, but not that Chrome now lacks the
initTouchEvent entirely (relies completely on the
TouchEvent constructor), we can return to the
original user agent checks without breaking Chrome
Mobile Emulation.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/4528)


diemol:
Cannot reopen this one because the source repo where this PR was sent from has been deleted.

Issue #4393, opened 2017-07-31
Python docstring examples for execute_script() and execute_async_script() are not helpful


paulie4:
## Expected Behavior -
The "Usage" examples would be working examples.

## Actual Behavior -
The "Usage" examples are not working examples.

## Steps to reproduce -
I think the docstrings from https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/webdriver.py#L512, which uses `driver.execute_script('document.title')` as the example usage for both execute_script() and execute_async_script(), are used to create the [readthedocs.io Python docs](http://selenium-python.readthedocs.io/api.html#selenium.webdriver.remote.webdriver.WebDriver.execute_script), so those docstrings should be fixed to have examples like this:
```python
result = driver.execute_script('return document.title')
```
and
```python
result = driver.execute_async_script("""
var done = arguments[0];
asyncFuncWithCallback(function (retVal) {
    done(retVal);
});
""")
```

majora2007:
This is really needed. I had a problem understanding how to work with promises and execute_script until I found this post.

p0deje:
I guess this is an easy one for a pull request.

KingAkeem:
Hello, I'm new to the open source community. I've used Selenium for a few projects before and wanted to contribute. Are you saying you'd just like the documentation to actually contain some JavaScript in the docs to better explain what the execute_script/execute_async_script do?

lmtierney:
@KingAkeem There's already an open PR for these changes in #4590 but there hasn't been any action in quite some time. If you look in that PR you should find some guidance

KingAkeem:
Thank you very much

lmtierney:
Fixed in 01131b2

PR #4590, opened 2017-08-28
Update execute_script and execute_async_script doc strings

michaelfward:
Per #4393 , the doc strings for execute_script and execute_async_script are incorrect. This commit adds a return to the two "usage" sections, fixing the errors that would occur without

### DIFF
--- a/py/selenium/webdriver/remote/webdriver.py
+++ b/py/selenium/webdriver/remote/webdriver.py
@@ -473,7 +473,7 @@ class WebDriver(object):
          - \*args: Any applicable arguments for your JavaScript.
 
         :Usage:
-            driver.execute_script('document.title')
+            driver.execute_script('return document.title')
         """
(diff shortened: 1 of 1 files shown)


Dude-X:
The async example should have a real example like in the Java doc; also it might be nice to add information about the set timeout for async scripts;

michaelfward:
Sounds good! I'll have this reflected in the PR shortly

lmtierney:
@michaelfward are you still planning on updating this?

KingAkeem:
Would this be a good example using setTimeout()?
`function timeout(ms) { return new Promise(resolve => setTimeout(resolve, ms));}`
`async function sleep(fn, ...args) {
    await timeout(3000);
    return fn(...args);
}`

lmtierney:
As this seemed to be abandoned, this was updated in 01131b2

PR #4603, opened 2017-08-30
Always use JSON.parse for bot.json.parse

vrana:
goog.json.parse is deprecated and about to be removed in late September 2017. This change prepares for that.


juangj:
I wonder if anybody is setting NATIVE_JSON=false anymore when compiling. I wouldn't think so... IE7 was probably the last one that needed it.

juangj:
Oh, hm. https://github.com/SeleniumHQ/selenium/blob/c76b56dd333cf3e1c13009f66124ff66ca357d31/rake-tasks/crazy_fun/mappings/javascript.rb#L832

Regardless, I would think that's unnecessary at this point.

PR #4607, opened 2017-08-30
Return a modifiable map from extractDesiredCapability

### LINKED ISSUE #4606: Exception thrown setting Firefox binary
## Meta -
OS:  
As tested MAC but the issue will appear on all OSes
Selenium Version:  3.5.3
Browser: Firefox (though Chrome is likely to be affected as well since we can set the binary as part of DefaultRemoteProxy.beforeSession(...) ) 

Browser Version:  55.0.3

## Expected Behavior
No exception would be thrown while trying to set the binary location for Firefox

## Actual Behavior

(relevant chunk of the stack trace)

Aug 29, 2017 3:28:11 PM org.openqa.grid.web.servlet.handler.RequestHandler beforeSessionEvent
SEVERE: Error running the beforeSessionListener : null
java.lang.UnsupportedOperationException
	at com.google.common.collect.ImmutableMap.put(ImmutableMap.java:495)
	at org.openqa.grid.selenium.proxy.DefaultRemoteProxy.beforeSession(DefaultRemoteProxy.java:228)
	at org.openqa.grid.web.servlet.handler.RequestHandler.beforeSessionEvent(RequestHandler.java:165)
	at org.openqa.grid.web.servlet.handler.RequestHandler.process(RequestHandler.java:105)
	at org.openqa.grid.web.servlet.DriverServlet.process(DriverServlet.java:84)
	at org.openqa.grid.web.servlet.DriverServlet.doPost(DriverServlet.java:68)

## Steps to reproduce -

Attempt to set the location of the Firefox binary during testing.

## Known Location/Issue

During the upgrade from Selenium 2.53.7 to Selenium 3, WebDriverRequest.extractDesiredCapability() switched to returning an ImmutableMap from a standard map. Then, when we attempt to modify it to inject the binary location as part of DefaultRemoteProxy, it throws an exception since modification of the map is not allowed.

matthewdoerksen:
Selenium 3 switched to returning an unmodifiable map from
WebDriverRequest.extractDesiredCapability(). This throws
an exception if we attempt to make any changes, even if
there is a valid case, such as setting a binary location.

Fixes #4606

### DIFF
--- a/java/server/src/org/openqa/grid/web/servlet/handler/WebDriverRequest.java
+++ b/java/server/src/org/openqa/grid/web/servlet/handler/WebDriverRequest.java
@@ -20,8 +20,6 @@ package org.openqa.grid.web.servlet.handler;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 
-import com.google.common.collect.ImmutableMap;
-
 import org.openqa.grid.common.exception.GridException;
 import org.openqa.grid.internal.ExternalSessionKey;
 import org.openqa.grid.internal.Registry;
@@ -30,6 +28,7 @@ import org.openqa.selenium.remote.server.NewSessionPayload;
 
 import java.io.Reader;
 import java.io.StringReader;
+import java.util.HashMap;
 import java.util.Map;
 
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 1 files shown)


shs96c:
Creating a subclass just to avoid a line of code seems excessive. Fixed in 24d89c6d13cc43dc006b09f237da7b246d4d7c23

PR #4622, opened 2017-09-01
Update chromeOptions key for capabilities

kereliuk:

### DIFF
--- a/py/selenium/webdriver/chrome/options.py
+++ b/py/selenium/webdriver/chrome/options.py
@@ -166,6 +166,6 @@ class Options(object):
         if self.debugger_address:
             chrome_options["debuggerAddress"] = self.debugger_address
 
-        chrome["chromeOptions"] = chrome_options
+        chrome["goog:chromeOptions"] = chrome_options
 
         return chrome


lmtierney:
@JKereliuk do you know how far back this works with older chromedriver versions, for backward compatibility? Will we break anything if we send both, or will it use `goog` and ignore the latter if both are sent?

JKereliuk:
Good point, this works only as of Chromedriver 2.31.
Not sure if that is far enough back to justify.
@scs96c I noticed you updated this for the Java bindings, what do you think about the python bindings?

shs96c:
giphy

The W3C payload should only have the `goog:` prefixed version, but the JSON Wire Protocol payload can have both, and that's legit.

p0deje:
Or you can just use `goog:chromeOptions` like [.NET does](https://github.com/SeleniumHQ/selenium/blob/0303d312797ede068ef0165ab49ef85c3d10a88d/dotnet/CHANGELOG#L3-L6).

shs96c:
You need both if you're going to support older versions of chrome.

p0deje:
Yep, while .NET doesn't do this since 3.5.2.

¯\_(ツ)_/¯

jimevans:
That's because I was a little overzealous in making sure the latest was supported going forward. Well, that, and being on "not the latest version of Chrome" takes a fair bit of work on the user's part.

lmtierney:
Thanks @JKereliuk!

Issue #4645, opened 2017-09-06
Dotnet - Attempt to save a screenshot to a directory which does not exist throw a cryptic generic GDI error


sandorfr:
## Meta -
OS:  
Windows
Selenium Version:  
3.5.1
Browser:  
any

Browser Version:  
any

When we try to save a screenshot to a directory which does not exist, It throws with the following error:
```
A generic error occurred in GDI+
```

## Expected Behavior -
It should either create the directory or throw an explicit `DirectoryNotFoundException`

## Actual Behavior -
It throws a cryptic generic GDI exception

## Steps to reproduce -

Add the following to an existing test
```csharp

                var testScreenShotPath = "C:\\PathToADirectoryWhichDoesNotExists";                
                var path = Path.Combine(testScreenShotPath, ${Guid.NewGuid()}.jpg");
                var screenshot = ((ITakesScreenshot)BrowserInstance).GetScreenshot();
                screenshot.SaveAsFile(path, ScreenshotImageFormat.Jpeg);
```

Run the test

Additional reference can be found [on Stackoverflow](https://stackoverflow.com/questions/29973934/a-generic-error-occurred-in-gdi-screenshot-saveasfile-externalexception)

sandorfr:
I've submitted a PR with a fix proposal throwing a DirectoryNotFoundException

jimevans:
Fixed in 96c1b99.

PR #4646, opened 2017-09-06
Test if output directory exists in order to avoid cryptic generic GDI…

sandorfr:
…+ error

#4645 


jimevans:
Thank you for the effort, but I'm going to reject this PR. I think there's a better approach that yields the same result. Rather than get the .NET bindings in the business of testing for directory existence, I'm going to have the `Image.Save` method write to a `FileStream` instead of using the overload that takes the file name as a string. This has two benefits. First, it will reduce the differences between the .NET Core 2.0 code path with the .NET Framework code path. Second, if the directory does not exist, attempting to create the FileStream object yields a `DirectoryNotFoundException` on its own, without the bindings having to know to throw that particular exception.

PR #4670, opened 2017-09-10
Fix package name in python webelement module

### LINKED ISSUE #4558: py[3.5.0]: __package__ may be None in selenium.webdriver.remote.webelement
## Meta -
OS:  Windows 7
Selenium Version:  selenium-py 3.5.0, on python 3.6.2
Browser:  N/A

## Expected Behavior -
`selenium.webdriver.remote.webelement` can be imported without issue

## Actual Behavior -
In some situations (I encountered that when launching a test suite from PyCharm), importing `selenium.webdriver.remote.webelement` (indirectly via `selenium.webdriver`) raises an `AttributeError` coming from the fact that `__package__` is `None` when calling `pkgutil.get_data(__package__, ...)` [here](https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/webelement.py#L41).

```
  File "d:\dev\.env\buildout\eggs\selenium-3.5.0-py3.6.egg\selenium\webdriver\remote\webelement.py", line 45, in <module>
    getAttribute_js = pkgutil.get_data(__package__, 'getAttribute.js').decode('utf8')
  File "C:\Python\3.6\Lib\pkgutil.py", line 616, in get_data
    spec = importlib.util.find_spec(package)
  File "D:\dev\.env\venv\buildout\lib\importlib\util.py", line 83, in find_spec
    fullname = resolve_name(name, package) if name.startswith('.') else name
AttributeError: 'NoneType' object has no attribute 'startswith'
```

Looking for a way to resolve that, I discovered that `__package__` is not set consistently by python within modules (see [here](https://stackoverflow.com/questions/4437394/package-is-none-when-importing-a-python-module), while it's hard to understand why it's not set in that situation), and that it may be better to rely on `__name__` instead. Hence I'd think of replacing:

```
getAttribute_js = pkgutil.get_data(__package__, 'getAttribute.js').decode('utf8')
isDisplayed_js = pkgutil.get_data(__package__, 'isDisplayed.js').decode('utf8')
```

by:

```
_pkg = '.'.join(__name__.split('.')[:-1])
getAttribute_js = pkgutil.get_data(_pkg, 'getAttribute.js').decode('utf8')
isDisplayed_js = pkgutil.get_data(_pkg, 'isDisplayed.js').decode('utf8')
```

## Steps to reproduce -
Not sure if it can be reproduced on any machine, but for me it's simply 'launch a script importing `selenium.webdriver` at some point from PyCharm'. Launching the same script in the console works, for some reason.

0uk:
__package__ may be None in some situations, retrieving it using
__name__ is more reliable

Fixes #4558

### DIFF
--- a/py/selenium/webdriver/remote/webelement.py
+++ b/py/selenium/webdriver/remote/webelement.py
@@ -38,8 +38,10 @@ try:
 except ImportError:  # 3+
     from io import BytesIO as IOStream
 
-getAttribute_js = pkgutil.get_data(__package__, 'getAttribute.js').decode('utf8')
-isDisplayed_js = pkgutil.get_data(__package__, 'isDisplayed.js').decode('utf8')
+# not relying on __package__ here as it can be `None` in some situations (see #4558)
+_pkg = '.'.join(__name__.split('.')[:-1])
+getAttribute_js = pkgutil.get_data(_pkg, 'getAttribute.js').decode('utf8')
+isDisplayed_js = pkgutil.get_data(_pkg, 'isDisplayed.js').decode('utf8')
 
 
 class WebElement(object):


lmtierney:
Merged in 302453d, thanks!

PR #4726, opened 2017-09-20
Added expected condition for presence of a window

Tom-Trumper:
Added new expected condition to wait for the presence of a window with
a specified title or handle.


shs96c:
In modern java, this is more elegantly written as:

```
wait.until(driver -> driver.switchTo().window(handle));
```

While I thank you for the PR, I'm closing this without merging. The behaviour it models can be written as the lambda above, and that clearly expresses intent. The `ExpectedConditions` class predates Java 8 by a long time, which is why it exists, and it attempted to also make it easy to express intent. It has become far larger than ever intended.

PR #4760, opened 2017-09-26
Add support on testing html5 app through Browser Automation Atoms

chan1cyrus2:
HTML5 apps usually refocus on another element when keydown is triggered. Added an autofocus flag which changes the element to the latest active element.
Some HTML5 apps are rendered on TV which has a lightweight version of browsers. Some browser's feature might not be supported, some conditions are checked for unsupported features.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/4760)

### DIFF
--- a/javascript/atoms/keyboard.js
+++ b/javascript/atoms/keyboard.js
@@ -491,7 +491,7 @@ bot.Keyboard.prototype.requiresKeyPress_ = function(key) {
       case bot.Keyboard.Keys.CONTEXT_MENU:
         return goog.userAgent.GECKO;
       default:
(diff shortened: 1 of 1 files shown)


juangj:
> HTML5 apps usually refocus on another element when keydown is triggered.

I'm not familiar with this claim. Can you show an example to help me understand what this means? Or, better yet, add a unit test that would illustrate the purpose of the new code.

AutomatedTester:
Thank you for your PR but I don't think this patch is what we want. It is better to refocus explicitly than implicitly

Issue #4591, opened 2017-08-28
support headless configuration for Firefox in selenium-webdriver package


mykmelez:
OS:  Windows, macOS, Linux
Selenium Version:  3.5.0
Browser:  Firefox
Browser Version:  56

Once Firefox supports headless mode on all three platforms (starting with Firefox 56, currently in beta and due to be released next month), the selenium-webdriver package should support configuring Firefox to run headlessly the same way it supports Chrome running headlessly, via a `headless()` method on the Firefox driver's _Options_ class, so that you can drive Firefox headlessly via:

```js
const { Builder } = require('selenium-webdriver');
const firefox = require('selenium-webdriver/firefox');
const driver = new Builder()
.forBrowser('firefox')
.setFirefoxOptions(new firefox.Options().headless())
.build();
```

(It would additionally be useful to be able to set the headless window's size via the equivalent of Chrome's `windowSize({width, height})` method.)

barancev:
This option should be added to all language binding.

barancev:
Implemented in Java binding by 38a5130ca52be0fa69632ee969ffa6846cfd25e7

franzliedke:
As far as I can tell, this is already possible with the Ruby bindings:

~~~ruby
options = Selenium::WebDriver::Firefox::Options.new(
  args: ['-headless']
)

Selenium::WebDriver.for :firefox, options: options
~~~

lmtierney:
@franzliedke yes it is most likely possible with all bindings, but this is asking for a shortcut without having to specify it in args

mykmelez:
Right, note that Chrome already supports this configuration in selenium-webdriver:

https://github.com/SeleniumHQ/selenium/blob/18660c422ee132e56835f5762a7bc20f7b025b9f/javascript/node/selenium-webdriver/chrome.js#L39-L51

And selenium-webdriver also provides an example script using it:

https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/example/chrome_headless.js

Although strangely the docs don't mention it:

http://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/chrome_exports_Options.html

This issue is about doing the same for Firefox. It may be merely syntactic sugar, but it's useful nonetheless.

franzliedke:
Thanks for clarification. I sent a PR for the Ruby bindings: #4762.

lmtierney:
Added to python in 1d490b4

[...]

PR #4762, opened 2017-09-27
Add shortcut for enabling Firefox' headless mode

franzliedke:
This provides a dedicated API for enabling headless mode. This is already possible via adding command-line arguments.

See #4591.

Please comment on whether using `headless!` conforms with your conventions. :smile: 
Do we need a way to disable this feature as well? Since there is no `remove_argument`, I would not assume so...

### DIFF
--- a/rb/lib/selenium/webdriver/firefox/options.rb
+++ b/rb/lib/selenium/webdriver/firefox/options.rb
@@ -94,6 +94,18 @@ module Selenium
         end
 
         #
+        # Run Firefox in headless mode.
+        #
+        # @example Enable headless mode
+        #   options = Selenium::WebDriver::Firefox::Options.new
+        #   options.headless!
+        #
+
+        def headless!
+          add_argument '-headless'
+        end
+
+        #
         # Sets Firefox profile.
         #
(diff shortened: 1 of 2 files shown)


lmtierney:
LGTM, @p0deje ?

titusfortner:
I against this idea. I don't think we should add a special method for each browser argument.

franzliedke:
@titusfortner Apparently, this was requested and not shot down in #4591.

p0deje:
Java bindings also provide this, so even I agree with @titusfortner that we should not wrap every argument with custom method, I'm more concerned to make bindings API consistent to each other.

Thank you for contribution! Merged in a2a76345aea31969849bbda63a045c1bfd473149.

barancev:
headless is special :)

jimevans:
I’m not entirely sure I agree that “headless is special.” There is already a mechanism for passing arguments to a browser’s command line. There is no defensible reason to not include a type-safe switch for every command line argument of a browser, if an exception is made for one of them. I am perfectly happy with inconsistency between languages if that means the particular language binding’s API retains internal consistency.

PR #4790, opened 2017-09-29
Accepts add-ons w/o an 'install.rdf'

### LINKED ISSUE #4093: WebExtensions not installing on firefox
## Meta -
OS:  
Windows 10 64bit

Selenium Version:  
selenium 3.4.0

Browser:  

Firefox Version 53 using 0.16.1 geckodriver and 

Steps to reproduce -

We are trying to install a new extension that we developed and it does not contain an install.rdf because this technique is now legacy as mentioned [here ](https://developer.mozilla.org/en-US/Add-ons/Install_Manifests).
The problem is that using the 0.16.1 geckodriver with selenium 3.4 does not seem to support this new kind of extensions and throws a : java.io.FileNotFoundException C:\Users(username)\AppData\Local\Temp\unzip3999903548710354720stream\install.rdf (The system cannot find the file specified)

Of course manually the extension installs. You can get it from [here](https://addons.mozilla.org/en-US/firefox/addon/mooltipass-extension/)
I have asked at geckodriver github and they said it is a selenium issue

jonsecchis:
It was not possible to install an add-on that did not contain an 'install.rdf' file inside aXPI package.
This commit fix this, adding support to 'manifest.json'.

Fixes #4093

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/4790)

### DIFF
--- a/py/selenium/webdriver/firefox/firefox_profile.py
+++ b/py/selenium/webdriver/firefox/firefox_profile.py
@@ -340,42 +340,59 @@ class FirefoxProfile(object):
                 finally:
                     compressed_file.close()
             elif os.path.isdir(addon_path):
-                with open(os.path.join(addon_path, 'install.rdf'), 'r') as f:
+                addon_contents = os.listdir(addon_path)
+
+                if 'install.rdf' in addon_contents:
+                    manifest_source = 'install.rdf'
+                elif 'manifest.json' in addon_contents:
+                    manifest_source = 'manifest.json'
+
+                with open(os.path.join(addon_path, manifest_source), 'r') as f:
                     manifest = f.read()
             else:
                 raise IOError('Add-on path is neither an XPI nor a directory: %s' % addon_path)
         except (IOError, KeyError) as e:
             raise AddonFormatError(str(e), sys.exc_info()[2])
+        
+        if manifest_source is 'install.rdf':
+            try:
+                doc = minidom.parseString(manifest)
 
-        try:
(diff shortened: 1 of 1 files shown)


jonsecchis:
I did not really have the time to write a test but it is working fine for me. I also removed some unnecessary guards raising exceptions that (in my vision) should already have been caught more properly elsewhere in execution.

tadatitam:
The changes by @jonsecchis worked for me too on Firefox 55, MacOS High Sierra, and ghostery-7.3.3.7.xpi

lmtierney:
@davehunt care to take a look?

davehunt:
I'd want to include a test for this, however I'm curious to know if the [`install_addon`](https://github.com/SeleniumHQ/selenium/blob/edd24f78db000e19ede480c691538a9fb5ee4f84/py/selenium/webdriver/firefox/webdriver.py#L233-L248) would be a better solution than adding the add-on to the profile? This has the advantage of relying on Mozilla's implementation for installing the add-on.

barancev:
Don't forget about legacy Firefox driver, it can install addons to a profile only.

Maratori:
Can we merge this PR?

motin:
Since https://github.com/SeleniumHQ/selenium/pull/5069 contains a test and covers the scenario when the add-on is a zip file, maybe it is relevant to close this PR in favor of the other one?

davehunt:
Thank you for your contribution. This has been fixed by #6463.

PR #4809, opened 2017-10-04
add Errno::EAFNOSUPPORT as ignored errors when finding ports

topherbullock:
We are adding this value because we received the following error:

```
Errno::EAFNOSUPPORT:
        Address family not supported by protocol - socket(2) for "::1" port 9515
```

Our system has disabled ipv6, but is still resolving ipv6 addresses for localhost. We are able to correct it, but it seemed that selenium shouldn't care about this.

### DIFF
--- a/rb/lib/selenium/webdriver/common/port_prober.rb
+++ b/rb/lib/selenium/webdriver/common/port_prober.rb
@@ -36,7 +36,7 @@ module Selenium
         port
       end
 
-      IGNORED_ERRORS = [Errno::EADDRNOTAVAIL]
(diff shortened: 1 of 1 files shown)


titusfortner:
Did you fix your issue by fixing your configuration? Or did you have to monkey patch this code?

I'm not very familiar with ipv6, does simply attempting on the next port work when you see this issue?

p0deje:
So what would happen if this error is simply suppressed? Does it work just fine?

Also, is there any way for me to try and reproduce your problem?

p0deje:
Merged in ef8b1bf093, thank you for the contribution!

PR #4836, opened 2017-10-09
[Grid] Add support for customizing the Grid Registry

mach6:
- Make the grid registry an interface -- GridRegistry.
- Introduce BaseGridRegistry which is intended to be extended by
  GridRegistry implementations.
- Introduce DefaultGridRegistry which replaces the pre-existing
  (now deprecated) Registry class. Most of the _real_ code from Registry
  was moved to DefaultGridRegistry.
- Deprecated Registry. Replaced underlying implementation to use new
  DefaultGridRegistry. Also make Registry implement GridRegistry.
  Changes here intended to ease migration and allow old
  code paths to continue to compile/work in the interim.
- Change the return type to T extends GridRegistry for some methods.
- Increase visibility of some additional methods and/or types which the
  existing DefaultGridRegistry takes advantage of -- allowing reuse.
- Add 'registryClass' arugment to '-role hub' for specifying custom
  registry implmentation on the CLASSPATH
- Update test code to use DefaultGridRegistry and GridRegistry

### DIFF
--- a/java/server/src/org/openqa/grid/internal/Registry.java
+++ b/java/server/src/org/openqa/grid/internal/Registry.java
@@ -19,421 +19,240 @@ package org.openqa.grid.internal;
 
 import net.jcip.annotations.ThreadSafe;
 
-import org.openqa.grid.internal.listeners.RegistrationListener;
-import org.openqa.grid.internal.listeners.SelfHealingProxy;
 import org.openqa.grid.internal.utils.configuration.GridHubConfiguration;
 import org.openqa.grid.web.Hub;
 import org.openqa.grid.web.servlet.handler.RequestHandler;
 import org.openqa.selenium.remote.DesiredCapabilities;
 import org.openqa.selenium.remote.internal.HttpClientFactory;
-import org.openqa.selenium.remote.server.log.LoggingManager;
 
 import java.util.List;
 import java.util.Set;
-import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.locks.Condition;
-import java.util.concurrent.locks.ReentrantLock;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 
 /**
- * Kernel of the grid. Keeps track of what's happening, what's free/used and assigned resources to
+ * Deprecated Kernel of the grid. Keeps track of what's happening, what's free/used and assigns resources to
  * incoming requests.
+ *
+ * @deprecated use {@link DefaultGridRegistry} instead. This class is just a proxy to it.
  */
+@Deprecated
 @ThreadSafe
-public class Registry {
+public class Registry implements GridRegistry {
+  BaseGridRegistry gridRegistry;
 
-  public static final String KEY = Registry.class.getName();
-  private static final Logger LOG = Logger.getLogger(Registry.class.getName());
-
-  // lock for anything modifying the tests session currently running on this
-  // registry.
-  private final ReentrantLock lock = new ReentrantLock();
-  private final Condition testSessionAvailable = lock.newCondition();
-  private final ProxySet proxies;
-  private final ActiveTestSessions activeTestSessions = new ActiveTestSessions();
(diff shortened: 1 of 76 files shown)


mach6:
@shs96c @barancev Though, I have the commit bit, I thought I would open a PR to get your feedback on this. My intention here is to allow for folks to plug-in their own grid registry implementation -- which of course is an advanced use case -- and would put people on their own support path.  One example of where I might want to take advantage of this capability is if I wanted to offload the new session queue to an external (perhaps central) process, so other hubs can consider it.

mach6:
@shs96c @barancev fyi, just rebased this. still waiting for your feedback. no rush. :)

java/server/src/org/openqa/grid/internal/Registry.java
```diff
@@ -19,421 +19,240 @@
...
   }
 
+  private void setConfiguration(GridHubConfiguration configuration) {
+    gridRegistry.configuration = configuration;
+  }
 
+  /**
+   * @see GridRegistry#start()
+   */
+  @Deprecated
```
lukeis: why are you adding this as a deprecated method?

If it is already deprecated, can you add what the alternative should be?

mach6: Hi @lukeis per the commit comments, my intention here is to deprecate the entire existing class `Registry` and replace it with `DefaultRegistry` -- which all code has be updated to use. The `Registry` class was left (and modified to use `DefaultRegistry` simply ease with compilation and notification, as people migrate, in case someone has custom grid code in their repository.

mach6: @lukeis thanks for taking a look, btw!

java/server/src/org/openqa/grid/internal/Registry.java
```diff
@@ -19,421 +19,240 @@
...
-import java.util.concurrent.locks.ReentrantLock;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 
 /**
- * Kernel of the grid. Keeps track of what's happening, what's free/used and assigned resources to
+ * Deprecated Kernel of the grid. Keeps track of what's happening, what's free/used and assigns resources to
  * incoming requests.
+ *
+ * @deprecated use {@link DefaultGridRegistry} instead. This class is just a proxy to it.
```
mach6: @lukeis note: the go forward message was placed on the class

lukeis: thanks! can you put the notes in the other places you specify @deprecated, as you can see someone might not notice the other reference 😁

lukeis:
nothing stands out to me after a (relatively quick) glance. 

@JasnoWa / @matthewdoerksen maybe you want to take a look to make sure it's not going to adversely conflict with our grid implementation?

java/server/src/org/openqa/grid/internal/Registry.java
mach6: Will do. Updating now

shs96c:
It makes me a little nervous that `ActiveTestSession` (which is an `internal` package) is made public. I'm also not sure of the underlying reason for the change --- presumably not only to introduce a new interface....

mach6:
@shs96c I can make `ActiveTestSession` implement a public interface or entirely revert making this portion of the default grid registry public, if you would prefer. 

My thoughts -- If someone were to implement their own grid registry, they will need to be able to manage active test sessions. Since the existing implementation here does such a good job, I though it would be good to open it up for reuse -- but I do understand your concern.

W.R.T the package `internal` -- We have other types which are in the `internal` package and are intended to be extended or implemented, in the advanced use case(s) --  `CapabilityMatcher` / `DefaultCapabilitiyMatcher`, `RemoteProxy` / `BaseRemoteProxy`, `HtmlRenderer` / `DefaultHtmlRenderer` -- So, what is the contract that you would like to see? I'm happy to make a comply.

mach6:
@shs96c merged this per our conversation

PR #4856, opened 2017-10-12
Package all Ruby dependencies

p0deje:
> titusfortner: Was talking to Simon. We need a vendor directory in selenium-webdriver to put all of the gems we use so that running `./go test_rb` doesn't require internet access / bundler to work

Finally it's possible to build and test Ruby bindings with no Internet
connection and no pre-installed software except Ruby 2.0.

@titusfortner I think that achieves what Simon was looking for. I'm not sure if your IDE would need any changes, but this works fine in macOS terminal (not sure about Windows, will need to check that).

To update gems we need to add custom task that would do something like this:

```bash
$ bundle update gem_name
$ bundle package
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ _bindings:
     python: "2.7"
   ruby: &ruby
     language: ruby
-    rvm: 2.3.3
+    rvm: 2.0.0-p648
     jdk: oraclejdk8
 
 _browsers:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,14 +2,14 @@ shallow_clone: true
 build: off
 environment:
   matrix:
-    - RUBY_VERSION: 22-x64
+    - RUBY_VERSION: 200
       TASK: build
-    - RUBY_VERSION: 200-x64
+    - RUBY_VERSION: 200
       TASK: //rb:ie-test
 install:
   - 'set JAVA_HOME=C:\Program Files (x86)\Java\jdk1.8.0'
   - 'set PATH=%JAVA_HOME%;%PATH%'
-  - 'set RUBY_HOME=C:\Ruby22-x64\bin'
+  - 'set RUBY_HOME=C:\Ruby200\bin'
   - 'set PATH=%RUBY_HOME%;%PATH%'
   - 'set BUCK_EXTRA_JAVA_ARGS=-Djna.nosys=true'
 before_test:


rake-tasks/crazy_fun/mappings/ruby.rb
```diff
@@ -182,7 +180,7 @@ def define_build_task(dir, args)
     def define_clean_task(dir, _args)
       desc 'Clean rubygem artifacts'
       task "//#{dir}:gem:clean" do
-        rm_rf "build/*.gem"
```
p0deje: This never worked unfortunately.

rake-tasks/crazy_fun/mappings/ruby.rb
```diff
@@ -198,8 +196,26 @@ def define_gem_install_task(dir, _args)
...
         ENV['BUNDLE_GEMFILE'] = 'rb/Gemfile'
-        sh "gem install bundler"
-        sh "bundle", "install"
+
+        gem_path = "#{Dir.pwd}/build/third_party/rb"
+        bin_path = "#{gem_path}/bin"
+        ENV["GEM_PATH"] = gem_path
+        ENV["PATH"] = [bin_path, ENV["PATH"]].join(":")
+
+        gems = `gem list`.split("\n")
```
titusfortner: When would this be empty?

p0deje: When you don't have bundler installed ;)

p0deje:
@titusfortner @lucast Please take a look, it changes the way bundler is used in Selenium, so maybe it might interfere with your workflow. I tried to implement it in the way you can still run specs from IDE, but I can't be 100% sure.

> Can we run Ruby tests without internet access now?

We can, except that there are other tasks (e.g. buck) that require Internet access 😏

rake-tasks/crazy_fun/mappings/ruby.rb
```diff
@@ -68,12 +68,12 @@ def handle(_fun, dir, args)
       args[:include] << "#{dir}/spec"
 
       args[:command] = args[:command] || "rspec"
-      args[:require] = Array(args[:require])
```
p0deje: It's not used anywhere.

rake-tasks/crazy_fun/mappings/ruby.rb
```diff
@@ -104,24 +104,20 @@ def handle(_fun, dir, args)
         end
 
         ruby :include => args[:include],
-             :require => args[:require],
              :command => args[:command],
              :args    => %w[--format doc --color] + (!!ENV['example'] ? ['--example', ENV['example']] : []),
              :debug   => !!ENV['log'],
-             :files   => args[:srcs],
-             :gemfile => "build/rb/Gemfile"
```
p0deje: Now we only use `rb/Gemfile` because we need lock file to run commands.

It might be error-prone on CI, but it should not be a big deal.

p0deje:
@titusfortner @lmtierney I'm going to merge this now, let me know if anything gets broken for you and we'll fix those.

PR #4876, opened 2017-10-14
[py] allow browser options to be passed to Remote WebDriver

lmtierney:
This also deprecates the previous browser WebDriver keywords for options in favor of a standardized 'options'
I.E. 'firefox_options' becomes 'options'

### DIFF
--- a/py/selenium/webdriver/chrome/options.py
+++ b/py/selenium/webdriver/chrome/options.py
@@ -22,6 +22,7 @@ from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
 
 
 class Options(object):
+    KEY = "goog:chromeOptions"
 
     def __init__(self):
         self._binary_location = ''
@@ -166,6 +167,6 @@ class Options(object):
         if self.debugger_address:
             chrome_options["debuggerAddress"] = self.debugger_address
 
-        chrome["goog:chromeOptions"] = chrome_options
+        chrome[self.KEY] = chrome_options
 
         return chrome
--- a/py/selenium/webdriver/remote/webdriver.py
+++ b/py/selenium/webdriver/remote/webdriver.py
@@ -111,7 +111,7 @@ class WebDriver(object):
 
     def __init__(self, command_executor='http://127.0.0.1:4444/wd/hub',
                  desired_capabilities=None, browser_profile=None, proxy=None,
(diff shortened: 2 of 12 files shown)


py/selenium/webdriver/chrome/options.py
```diff
@@ -22,6 +22,7 @@
 
 
 class Options(object):
+    KEY = "goog:chromeOptions"
```
Dude-X: nice touch since Google likes to change things every now and then.

py/selenium/webdriver/remote/webdriver.py
```diff
@@ -111,7 +111,7 @@ class WebDriver(object):
 
     def __init__(self, command_executor='http://127.0.0.1:4444/wd/hub',
                  desired_capabilities=None, browser_profile=None, proxy=None,
-                 keep_alive=False, file_detector=None):
+                 keep_alive=False, file_detector=None, options=None):
```
Dude-X: @AutomatedTester @davehunt thoughts on a new keyword argument?

lmtierney:
We certainly could have a base Options class if there are some commonalities. I think right now the only common one might be `args`. We do have one started for Edge, but afaik it's not ready yet and I'm not sure any of it would work. Maybe I'll run help on the edge driver and see what args are available.

davehunt:
> I think right now the only common one might be `args`.

There's `to_capabilities`, which they should all implement. I'm not sure we can assume much else.

lmtierney:
True, I suppose we could do something like this for `to_capabilities` since it will (most likely) be different for each browser https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/common/service.py#L59

davehunt:
@lmtierney sure, though I was thinking more of an [Abstract Base Class](https://docs.python.org/3.6/library/abc.html) but the end result is the same.

lmtierney:
@davehunt ah yeah that'd probably be better. I'll leave that out of this PR and maybe add it in later when we know what Safari and Microsoft options will look like

lmtierney:
@AutomatedTester if there are no objections to this PR, I'm going to wait until after the next release to merge it

AutomatedTester:
@lmtierney looks good to me

PR #4881, opened 2017-10-16
Updated docstrings to specify the type of path needed to install firefox addons.

### LINKED ISSUE #4858: Doc string for python 'install_addon' should give more detail about the path of the addon.
Currently the doc string here: https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/firefox/webdriver.py#L235-L241 does not contain any information about what the path to the addon wanting to be installed should be.

The path should be an absolute path as seen here: https://dxr.mozilla.org/mozilla-central/source/testing/marionette/addon.js#80

b4handjr:
This change helps the user give the correct full path to the addon they would like to install for firefox. Marionette is expecting a full path.

Fixes #4858

### DIFF
--- a/py/selenium/webdriver/firefox/webdriver.py
+++ b/py/selenium/webdriver/firefox/webdriver.py
@@ -237,10 +237,10 @@ class WebDriver(RemoteWebDriver):
         Returns identifier of installed addon. This identifier can later
         be used to uninstall addon.
 
-        :param path: Full path to the addon that will be installed.
+        :param path: Absolute path to the addon that will be installed.
 
         :Usage:
-            driver.install_addon('firebug.xpi')
+            driver.install_addon('/path/to/firebug.xpi')
         """
         payload = {"path": path}
         if temporary is not None:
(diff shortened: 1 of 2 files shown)


jrbenny35:
@davehunt How does this look?

py/selenium/webdriver/firefox/webdriver.py
```diff
@@ -237,6 +237,8 @@ def install_addon(self, path, temporary=None):
         Returns identifier of installed addon. This identifier can later
         be used to uninstall addon.
 
+        :param path: Full path to the addon that will be installed.
+
```
lmtierney: Can you update the `Usage` to be an absolute path as well? Something like `/path/to/firebug.xpi`

davehunt:
> How does this look?

I would use 'absolute path' in place of 'full path', but either way I'm happy with this once you address @lmtierney's comment. Thanks @jrbenny35!

lmtierney:
Merged, thank you @jrbenny35!

PR #4932, opened 2017-10-23
node/selenium-webdriver: allow maximum number of portprober attempts to be configured

shinypb:
On busy automation boxes, the default value of 10 is not sufficient to reliably discover an
available port number. This manifests as our Selenium test suite failing somewhat randomly during our working day.

Set `process.env.SELENIUM_MAX_PORT_PROBES` to configure this as needed.

### DIFF
--- a/javascript/node/selenium-webdriver/net/portprober.js
+++ b/javascript/node/selenium-webdriver/net/portprober.js
@@ -174,12 +174,13 @@ function isFree(port, opt_host) {
 function findFreePort(opt_host) {
   return findSystemPortRange().then(function(range) {
     var attempts = 0;
+    var maxAttempts = 100;
     return new Promise((resolve, reject) => {
       findPort();
 
       function findPort() {
         attempts += 1;
-        if (attempts > 10) {
+        if (attempts > maxAttempts) {
           reject(Error('Unable to find a free port'));
         }
 


javascript/node/selenium-webdriver/net/portprober.js
```diff
@@ -174,12 +174,13 @@ function isFree(port, opt_host) {
 function findFreePort(opt_host) {
   return findSystemPortRange().then(function(range) {
     var attempts = 0;
+    var maxAttempts = process.env.SELENIUM_MAX_PORT_PROBES || 10;
```
jleyba: Update jsdoc to mention this new behavior.

shinypb: Hi there, @jleyba — thanks for taking a look at this PR.

I'm not sure what number of attempts would guarantee (or come close to it) success. Honestly, I've monkey-patched the version on our system to just retry forever.

I appreciate the desire to eliminate the number of random configuration settings, however. What if we set maxAttempts to something like 10% of the range? e.g. `var maxAttempts = 0.10 * (range.max - range.min)`? That'd be a pretty large number of attempts on a Unix system but I think it would be OK — being unable to find a port is a pretty edge-case-y scenario.

If you'd prefer to keep the environment variable (or anything else), just let me know and I'll happily follow up with your recommended changes.

jleyba: How about some kind of backoff function? Something like:

```js
const maxAttempts = 100;
let attempt = 0;
let backoffFactor = 1;
(function findPort() {
  let port = randomPort();
  if (await isFree(port)) {
    return;
  }

  attempt += 1;
  if (attempt >= maxAttempts)
    throw Error;

  if (attempt % 10 === 0) {
    backoffFactor = 10 << (attempt / 10 - 1);
  }

  setTimeout(findPort, 20 * backoffFactor);
})();

```

shinypb: I've been noodling on this for the past _looks up at timestamp_ 23 hours, and I'm starting to think we should just bump maxAttempts to 100 and be done with it.

The vast majority of the time, I expect the first port we randomly select to be available. On heavily loaded hosts where it takes many more guesses to find one, I am not sure a delay and backoff would do much to help us. I think a delay would only help if we expected the server to be less loaded a few seconds from now than it currently is, and I can't actually think of a workload where that would be true.

I've pushed a new version of this PR that does just that in case you agree. If you don't agree, just let me know and I'll land the backoff-with-delay idea instead.

jleyba: Bumping to 100 sgtm. Did you create a new PR for that? I don't see the change in this one.

shinypb:
Oops! I didn't actually push it yesterday. 🤦‍♂️ Here it is!

PR #4938, opened 2017-10-24
Add missing python W3C ErrorCodes

### LINKED ISSUE #4556: webdriver/remote/errorhandler.py doesn't match what the spec defines
Comparing the error codes in https://github.com/SeleniumHQ/selenium/blob/3de9df1630e574607aa703c22c2a4b8770a9523d/py/selenium/webdriver/remote/errorhandler.py and in https://w3c.github.io/webdriver/webdriver-spec.html#handling-errors (as of 6e6575c), the Python code is missing the following error codes:

 * element click intercepted
 * insecure certificate
 * invalid argument
 * invalid coordinates
 * invalid session id
 * no such cookie
 * session not created
 * unable to capture screen
 * unknown method

The Python code also has the following that don't appear in the spec:

 * element not visible
 * ime engine activation failed
 * ime not available
 * invalid element coordinates

lmtierney:
Aligns with Java Error Codes that are available
Closes #4556

### DIFF
--- a/py/selenium/common/exceptions.py
+++ b/py/selenium/common/exceptions.py
@@ -255,3 +255,70 @@ class InvalidArgumentException(WebDriverException):
     The arguments passed to a command are either invalid or malformed.
     """
     pass
+
+
+class JavascriptException(WebDriverException):
+    """
+    An error occurred while executing JavaScript supplied by the user.
+    """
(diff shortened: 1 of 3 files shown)


lmtierney:
This also addresses #4380

Issue #4867, opened 2017-10-13
Expectation don't handle multiple fitting elements with different display state


HenningL:
## Meta -
OS:  
all
Selenium Version:  
latest
Browser:  
all

Call:
ExpectedConditions.visibilityOfElementLocated(getSelector()))

## Expected Behavior -
Element which fits the expectation is found
## Actual Behavior -
Only the first selector-found element is considered to fulfill visibility while there is a list of selector-found elements where one of them fulfill.
## Steps to reproduce -

Having a site where two(or more) elements are present found by xpath while the first is hidden and the second is visible, the expectation of the visibility of the element found is never fulfilled.

elementIfVisible(findElement(locator, driver)); 
while findElement(locator, driver) only returns the first which doesn't fulfill visibility.

Expected is, that the element which fulfill all criteria is found/considered.

barancev:
ExpectedConditions class was designed to be a set of examples, not the compete set of all possible conditions. We're not going to expand this class too much, especially because Java 8 introduced new cool features, like lambdas and streams, that made writing conditions really easy. The condition you need can be implemented as a single line of code:
```
wait.until(d -> d.findElements(locator).stream().filter(WebElement::isDisplayed).findFirst().orElse(null))
```

PR #4963, opened 2017-10-26
Added expected condition for element visibility

Tom-Trumper:
Added expected condition to determine visibility of any elements that
match a given locator, rather than limited to the first element found
by ExpectedConditions.visibilityOfElementLocated(By) or all elements
found by ExpectedConditions.visibilityOfAllElementsLocatedBy(By).

Addresses #4867


shs96c:
This seems like very specialised behaviour, and potentially subject to nasty race conditions. `ExpectedConditions` is already a massive class, and the need for it is significantly reduced by Java 8 lambdas. For these reasons, although I thank you for the contribution, I'm not going to merge this into the tree.

HerbertMarshall:
I understand the reasoning of not merging the code, however I think at the very least the javadoc should be updated to indicate that only the first element located will be targeted.

PR #4966, opened 2017-10-27
Set default values for the pointerMove offset.

juangj:
This allows callers to omit the x and y parameters when moving the pointer to
the center of the target, consistent with the old API.


juangj:
Thoughts? It doesn't really make sense to omit the offset coordinates from pointer-relative or viewport-relative coordinates. But it's slightly more convenient to be allowed to omit them from element-relative commands, and I don't think the default values do harm.

jleyba:
LGTM

PR #4974, opened 2017-10-27
Send W3C capabilities in the new session request.

juangj:
If the Selenium standalone server only finds legacy capabilities in the request,
then it will assume that the client doesn't speak W3C, and will return a legacy
response. Sending a request that includes the W3C keys induces it to send a W3C
response.

### DIFF
--- a/javascript/node/selenium-webdriver/lib/webdriver.js
+++ b/javascript/node/selenium-webdriver/lib/webdriver.js
@@ -32,6 +32,20 @@ const Symbols = require('./symbols');
 const promise = require('./promise');
 
 
+// Capability names that are defined in the W3C spec.
+const W3C_CAPABILITY_NAMES = new Set([
+    'acceptInsecureCerts',
+    'browserName',
+    'browserVersion',
+    'platformName',
+    'pageLoadStrategy',
+    'proxy',
+    'setWindowRect',
+    'timeouts',
+    'unhandledPromptBehavior',
+]);
+
+
 /**
  * Defines a condition for use with WebDriver's {@linkplain WebDriver#wait wait
  * command}.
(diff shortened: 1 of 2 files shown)


juangj:
This probably also applies to Grid, but I'm really not that familiar with Grid so I'm not positive.

Other language bindings do fancier things like filtering non-W3C capability names out of the W3C capabilities blob ([here it is in Python](https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/webdriver.py#L43)). Not sure how necessary you feel that is.

Also kinda undecided how to treat required capabilities, but those should go away at some point anyway, so maybe you were already planning an API change to eliminate them.

javascript/node/selenium-webdriver/firefox/index.js
```diff
@@ -380,7 +380,7 @@ function normalizeProxyConfiguration(config) {
     }
   } else if ('pac' === config.proxyType) {
     if (config.proxyAutoconfigUrl && !config.pacUrl) {
-      config.pacUrl = config.proxyAutoconfigUrl;
+      config.proxyAutoconfigUrl = config.proxyAutoconfigUrl;
```
jleyba: Actually, it looks like the entire `normalizeProxyConfiguration` function isn't needed anymore

juangj:
`npm test` passed (or, well, it didn't fail any more than it already did on my machine). And this seemed to work for many tests, but it doesn't work for tests that set non-W3C capabilities.

I do have some code (probably living in the wrong file right now) that will filter non-W3C capability names like some of the other language bindings do. But if you're planning to completely ditch legacy protocol support anyway, then maybe there's no need for any of this -- once all the capability names are changed, you could just switch completely to W3C capabilities instead of sending this hybrid.

(I'm going to backport some version of this to our internal 3.6.0 branch in order to continue to support Chrome users, though.)

juangj:
I added that commit that drops invalid capabilities, if you're curious.

The `normalizeProxyConfiguration` thing should probably just go in a separate commit.

PR #5039, opened 2017-11-09
Fix WebDriverException when binary is set, but no firefox binary in the PATH variable

darek8686:
The cause of the problem is to call the orElse method with new instance of FirefoxBinary as argument,
which requires Firefox binary in envirement variable PATH.


shs96c:
Good catch. Thank you!

PR #5068, opened 2017-11-16
Fix possible NullPointerException in ErrorHandler.rebuildServerError

qmfrederik:

`ErrorHandler.rebuildServerError` can throw NullPointerExceptions whe the remote Selenium server returns an error object which has null values for CLASS or STACK_TRACE. The code currently checks whether these keys are present in the JSON object but not whether they have a value.

This PR fixes that.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/5068)

### DIFF
--- a/java/client/test/org/openqa/selenium/remote/ErrorHandlerTest.java
+++ b/java/client/test/org/openqa/selenium/remote/ErrorHandlerTest.java
@@ -327,23 +327,59 @@ public class ErrorHandlerTest {
     }
   }
   
-    @SuppressWarnings("ThrowableInstanceNeverThrown")
-    @Test
-    public void testShoulNotBuildWebDriverExceptionIfClassOrStackTraceIsNull() {
-      Map<String, ?> data = ImmutableMap.of(
-          "message", "some error message",
-          "class", null,
-          "stackTrace", null);
+  @SuppressWarnings("ThrowableInstanceNeverThrown")
+  @Test
+  public void testShoulNotBuildWebDriverExceptionIfClassAndStackTraceIsNull() {
+    Map<String, ?> data = ImmutableMap.of(
+        "message", "some error message",
+        "class", null,
+        "stackTrace", null);
   
-      try {
-        handler.throwIfResponseFailed(createResponse(ErrorCodes.UNHANDLED_ERROR, data), 123);
-        fail("Should have thrown!");
-      } catch (WebDriverException expected) {
-        assertEquals(new WebDriverException("some error message\nCommand duration or timeout: 123 milliseconds",
-                                            new WebDriverException()).getMessage(),
-            expected.getMessage());
-      }
+    try {
+      handler.throwIfResponseFailed(createResponse(ErrorCodes.UNHANDLED_ERROR, data), 123);
+      fail("Should have thrown!");
+    } catch (WebDriverException expected) {
+      assertEquals(new WebDriverException("some error message\nCommand duration or timeout: 123 milliseconds",
+                                          new WebDriverException()).getMessage(),
+          expected.getMessage());
     }
+  }
+  
+  @SuppressWarnings("ThrowableInstanceNeverThrown")
+  @Test
+  public void testShoulNotBuildWebDriverExceptionIfClassNullAndStackTraceNotNull() {
+    Map<String, ?> data = ImmutableMap.of(
+        "message", "some error message",
+        "class", null,
+        "stackTrace", "a");
+  
+    try {
+      handler.throwIfResponseFailed(createResponse(ErrorCodes.UNHANDLED_ERROR, data), 123);
+      fail("Should have thrown!");
+    } catch (WebDriverException expected) {
+      assertEquals(new WebDriverException("some error message\nCommand duration or timeout: 123 milliseconds",
+                                          new WebDriverException()).getMessage(),
+          expected.getMessage());
+    }
+  }
+  
+  @SuppressWarnings("ThrowableInstanceNeverThrown")
+  @Test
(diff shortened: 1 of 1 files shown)


qmfrederik:
@p0deje The build errors seem unrelated and I can't start individual Travis build - anything else I can do here?

p0deje:
One of Java mainatainers will have to take a look and review it. It might take some time.

java/client/src/org/openqa/selenium/remote/ErrorHandler.java
```diff
@@ -220,7 +220,7 @@ private String duration(long duration) {
 
   private Throwable rebuildServerError(Map<String, Object> rawErrorData, int responseStatus) {
 
-    if (!rawErrorData.containsKey(CLASS) && !rawErrorData.containsKey(STACK_TRACE)) {
+    if (!(rawErrorData.containsKey(CLASS) && rawErrorData.get(CLASS) != null) && !(rawErrorData.containsKey(STACK_TRACE) && rawErrorData.get(STACK_TRACE) != null)) {
```
barancev: `!(rawErrorData.containsKey(CLASS) && rawErrorData.get(CLASS) != null)`
can be simplified to 
`rawErrorData.get(CLASS) == null`
and the same for the second condition.

qmfrederik:
@barancev Thanks, I updated the PR. Let me know if this works for you.

barancev:
By the way, where did you get NullPointerException, in which line? How can we reproduce it? Adding an unit test would help too.

qmfrederik:
This line:

```
         clazz = Class.forName(className);
```

would throw the exception because a key with name `CLASS` was present but its value was `null`. Let's see if I can do something unit test-wise.

qmfrederik:
@barancev I added unit tests and the Travis build looks OK to me. Let me know if you need anything else.

java/client/src/org/openqa/selenium/remote/ErrorHandler.java
```diff
@@ -220,7 +220,7 @@ private String duration(long duration) {
 
   private Throwable rebuildServerError(Map<String, Object> rawErrorData, int responseStatus) {
 
-    if (!rawErrorData.containsKey(CLASS) && !rawErrorData.containsKey(STACK_TRACE)) {
+    if (rawErrorData.get(CLASS) == null || rawErrorData.get(STACK_TRACE) == null) {
```
barancev: Why && changed to || ?

barancev: Similar change should be applied to other parts of this function, they should check for null value instead of calling containsKey.

java/client/test/org/openqa/selenium/remote/ErrorHandlerTest.java
```diff
@@ -326,6 +326,24 @@ public void testShouldStillTryToBuildWebDriverExceptionIfClassIsNotProvidedAndSt
       assertStackTracesEqual(expectedTrace, cause.getStackTrace());
     }
   }
+  
+    @SuppressWarnings("ThrowableInstanceNeverThrown")
+    @Test
+    public void testShoulNotBuildWebDriverExceptionIfClassOrStackTraceIsNull() {
```
barancev: This is a test testShoulNotBuildWebDriverExceptionIfClass_and_StackTraceIsNull
See comment about && and ||, and let's add tests for other combinations.

java/client/src/org/openqa/selenium/remote/ErrorHandler.java
qmfrederik: The check makes sure that valid values are present for CLASS and STACK_TRACE. So we want to make sure that both CLASS is not null and STACK_TRACE is not null. In other words, quit if CLASS or STACK_TRACE are null.

qmfrederik: Yeah, though perhaps in a separate PR?

java/client/test/org/openqa/selenium/remote/ErrorHandlerTest.java
qmfrederik: OK, updated

qmfrederik:
@barancev I rebased the PR & addressed the comments. Let me know if you need anything else.

java/client/src/org/openqa/selenium/remote/ErrorHandler.java
qmfrederik: Ah, I missed that.

I don't think that comment holds true in the W3C WebDriver era - see https://www.w3.org/TR/webdriver/#handling-errors.

`class` doesn't seem to be defined in the spec and `stacktrace` is "implementation-specific".

What this code tries to do, is to reconstruct a Java exception based on the error data sent by the server. But there is no guarantee at all that the server is implemented in Java.

barancev: Java binding tries to be backward compatible. At least until we turn to 4.x

qmfrederik: That's fair, but in this case it would cause the binding to crash when a remote end sends W3C-spec compliant responses 😄 .

barancev: So we should try to make it work for both legacy and W3C conformant cases.

barancev:
This PR was not finished (it introduced some backward incompatibilities that were not resolved), but I've decided to apply it anyway. So I've merged it as bc1a581c100a83a8a1ef593cbfbac5379a3e6267 and addded a followup commit that fixes comilation errors and restores backward compatibility.

PR #5069, opened 2017-11-16
Add Python client support for WebExtensions in Firefox

sangaline:
This attempts to parse a `manifest.json` file before 'index.if it's present before `install.rdf` in the Python `FirefoxProfile` class which adds support for the newer WebExtensions API addon format while maintaining backwards compatibility.

I haven't added a test for this, but I'm having trouble finding any tests for the existing implementation to work from. Are there any tests for `FirefoxProfile.add_extension()` in the Python client?

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/5069)

### DIFF
--- a/py/test/selenium/webdriver/firefox/ff_profile_tests.py
+++ b/py/test/selenium/webdriver/firefox/ff_profile_tests.py
@@ -177,7 +177,7 @@ def test_that_we_can_add_a_webextension(capabilities, webserver):
     extension_filename = 'webextension.xpi'
     # TODO: This file should probably live in a common directory.
     current_directory = os.path.dirname(os.path.realpath(__file__))
-    root_directory = os.path.join(current_directory, '..' , '..', '..', '..', '..')
+    root_directory = os.path.join(current_directory, '..', '..', '..', '..', '..')
     data_directory = os.path.join(root_directory, 'javascript', 'node',
                                   'selenium-webdriver', 'lib', 'test', 'data', 'firefox')
     full_extension_filename = os.path.join(data_directory, extension_filename)
@@ -185,6 +185,6 @@ def test_that_we_can_add_a_webextension(capabilities, webserver):
     profile = FirefoxProfile()
     profile.add_extension(full_extension_filename)
 
-    driver = Firefox(fcapabilities=capabilities, firefox_profile=profile)
+    driver = Firefox(capabilities=capabilities, firefox_profile=profile)
     driver.get(webserver.where_is('simpleTest.html'))
     driver.find_element_by_id('webextensions-selenium-example')


lmtierney:
I don't believe there are currently tests around it no. There is a similar PR open currently #4790

sangaline:
Ah sorry, I did a search for "WebExtensions" in the PRs and that didn't turn up. It looks like the main holdup on that was the lack of a test. I've added one at `test/selenium/webdriver/firefox/ff_profile_tests.py::test_that_we_can_add_a_webextension` that uses the same `webextension.xpi` file that the JavaScript tests use.

lmtierney:
@whimboo @davehunt @AutomatedTester can any of you review this? It's been around for a while and I've never dealt with installing a webextension

whimboo:
Well, me neither. But could you please make sure that the builds are working fine first? Both Travis-CI and AppVeyor are failing.

whimboo:
Also just a question, what is the future of the FirefoxProfile in Selenium? Will this class continue to exist?

I asked because extensions of any type can be installed via the `/session/{session id}/moz/addon/install` endpoint which is supported by geckodriver for a while now. Just send the base64 encoded string, and Firefox takes care of the installation. As such no complicated manual steps are necessary.

lmtierney:
Once we get rid of the legacy driver, the FirefoxProfile class can certainly be stripped down a fair amount; but people will still want to use pre-defined profiles. The class allows users to easily define which profile to use and will do encoding for them.

davehunt:
This looks okay to me, and +1 for including a test! It's shocking that we don't already have tests for `install_extension`. As we support installing from source (rather than zip) we should ideally have a test for that too. I think we can merge this, but if @sangaline is interested in contributing more tests for this function then that would be great. Looking through the code, I feel that `FirefoxProfile` needs to be refactored, and having tests in place would make that much safer to do.

py/selenium/webdriver/firefox/firefox_profile.py
```diff
@@ -327,6 +327,16 @@ def get_text(element):
                     rc.append(node.data)
             return ''.join(rc).strip()
 
+        def parse_manifest_json(content):
+            """Extracts the details from the contents of a WebExtensions `manifest.json` file."""
+            manifest = json.loads(content)
+            return {
+                'id': manifest['applications']['gecko']['id'],
```
davehunt: The id is not mandatory for web extensions. Could you update this to support a web extension without an id?

lmtierney:
@sangaline ping on this, can you address @davehunt's review?

py/selenium/webdriver/firefox/firefox_profile.py
motin: Here is how the ruby bindings solve it for extensions without an id: https://github.com/SeleniumHQ/selenium/blob/07a18746ff756e90fd79ef253a328bd7dfa9e6dc/rb/lib/selenium/webdriver/firefox/extension.rb#L85-L91

As seen above, the ruby bindings doesn't even look for applications.gecko.id (an issue worth reporting?)

motin:
Note: This PR fixes the python part of #4093

py/selenium/webdriver/firefox/firefox_profile.py
motin: The Java bindings does the same as the Ruby bindings:
https://github.com/SeleniumHQ/selenium/blob/ba32c5bc83272ff0e111fd4dccc40435700c6225/java/client/src/org/openqa/selenium/firefox/internal/FileExtension.java#L152-L155

The JavaScript bindings throw an error of applications.gecko.id is not defined: 
https://github.com/SeleniumHQ/selenium/blob/3e99112a0d59878c6f31cc50fe8b0593c49e497f/javascript/node/selenium-webdriver/firefox.js#L172-L174

sangaline:
Should I still address @davehunt's review or defer to #6463 at this point?

davehunt:
This was merged in #6463

PR #5074, opened 2017-11-18
Guard messages

titusfortner:
I added specs that specify behavior but have to be run manually because it requires verifying tests are failing, which kind of defeats the purpose of having them in the test suite.

The idea is to try to encourage each guard to be for the same reason, and add a message to explain why (ideally URL to bug  report)
This implementation would require guards to be placed in same order as `GUARD_TYPES` (alphabetically), and each guard to have a matching message (or a `nil`).

I was going to add more info in the message about which guard applied, but thought this was a better place to start at least.

This also allows to continue skipping non-applicable specs when running via RSpec or RubyMine directly. (`./go` only executes certain folders, where running `rspec` directly will attempt to run and pending out the Firefox tests using Chrome, etc)

I'm not sure when the `#uniq` is necessary when constructing the guard array. I didn't test the `example_group` thoroughly, does that result in multiple of the same guard in places? That might break this implementation. Let me know your feedback.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/5074)

### DIFF
--- a/rake-tasks/crazy_fun/mappings/ruby.rb
+++ b/rake-tasks/crazy_fun/mappings/ruby.rb
@@ -79,6 +79,7 @@ class RubyMappings
 
   class AddTestDependencies < Tasks
     def handle(_fun, dir, args)
+      ENV['RB_GO_EXECUTION'] = 'true'
       task = Rake::Task[task_name(dir, "#{args[:name]}-test")]
 
       if args.has_key?(:deps)
--- a/rb/spec/integration/selenium/webdriver/guard_spec.rb
+++ b/rb/spec/integration/selenium/webdriver/guard_spec.rb
@@ -1,0 +1,161 @@
+# Licensed to the Software Freedom Conservancy (SFC) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The SFC licenses this file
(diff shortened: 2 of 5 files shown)


p0deje:
Why do you want to have "skip"? The whole point was to run all the specs for each browser to make sure that those that are guarded actually fail for different implementations?

Why don't you put a message to guard?

```ruby
except: {browser: :chrome, message: 'foo'}
```

titusfortner:
If you run `bundle exec rake` it will fail because the browser will be Chrome, but it will run the Firefox tests, and those Firefox tests will be guarded to firefox only, but they will still pass, which causes the rake task to fail. This isn't an issue when run with `./go` because the `*.desc` will limit which files get run.

titusfortner:
and, you're right, I just realized that arrays inside browser hash are not treated as separate guards, which is what I assumed when trying to do arrays. I'll re-do it to match your suggestion.

luke-hill:
I know this is an old topic. But just to add this only affects tests that are non-chrome or non-default browser. As when running directly, I think the interpreter seems to use my default browser.

Tested on a VBox and it ran stuff through IE first, and on my linux box it ran them through Chrome first

p0deje:
@titusfortner Any news about that? Should we close this PR for now?

PR #5076, opened 2017-11-19
Support injection of custom renderer from proxies

krmahadevan:
Enabled support for adding custom html renderers
from custom proxy implementations, so that users 
can customize how they would want to view the 
proxy information on the grid console.


krmahadevan:
ping @shs96c - Can you please help take a look ?

krmahadevan:
@barancev - Since you have helped me with two other PRs on the grid side, I was hoping if you could please help take a look at this one.

barancev:
Merged #5198 instead, because it also deleted an unused class (that becomes unused after applying the patch)

Bethannfrum:
Fix and restore to the original format and release old documents

PR #5078, opened 2017-11-20
Minor refactoring to use streams and lambdas

krmahadevan:

### DIFF
--- a/java/server/src/org/openqa/grid/internal/DefaultGridRegistry.java
+++ b/java/server/src/org/openqa/grid/internal/DefaultGridRegistry.java
@@ -118,11 +118,8 @@ public class DefaultGridRegistry extends BaseGridRegistry implements GridRegistr
    * @param reason  the reason for termination
    */
   public void terminate(final TestSession session, final SessionTerminationReason reason) {
-    new Thread(new Runnable() { // Thread safety reviewed
-      public void run() {
-        _release(session.getSlot(), reason);
-      }
-    }).start();
+    // Thread safety reviewed
+    new Thread(() -> _release(session.getSlot(), reason)).start();
   }
 
   /**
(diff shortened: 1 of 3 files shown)


krmahadevan:
ping @shs96c - Please help take a look

barancev:
I'm not ready to accept this PR as is. BaseRemoteProxy changes behavior, it's not "minor refactoring" and it should be excluded from this change set.

krmahadevan:
@barancev - The changes to `BaseRemoteProxy` are mere null checks. Its not changing any functionality per say. I had to introduce them, because I was seeing a test failure owing to those missing null checks.

barancev:
"I was seeing a test failure, now I don't see them" -- is not it a change in fuctionality? ;)

And, yes, this change _hides a problem_ detected by recently enabled DefaultRemoteProxyTest.

krmahadevan:
@barancev - ok :) 

So I have two options to go about with this 

1. Raise 2 PRs. PR-1 will just have the lambda+stream changes, PR-2 will have the changes done to `BaseRemoteProxy`.
2. In the same PR have two explicit commits. Commit-1 has the lambdas+stream changes, Commit-2 has the changes done to `BaseRemoteProxy`

Which one would you prefer. Please let me know. I will proceed accordingly.

barancev:
Make a commit with lambda+stream changes, it can be landed immediately.

As for the other change, it should go with improved version of DefaultRemoteProxyTest. This test fails because cleanup thread is null, whereas it should not be null and it should do cleanup. Adding a null-check makes the test green, but does not resolve the issue, it's a false positive.

krmahadevan:
@barancev - Thanks. I have amended the commit to exclude the `BaseRemoteProxy` changes and include just the lamdas+streams changes. Please check. I will raise a new PR as per your suggestion for the failing test.

barancev:
Merged as ef5de3815983c6cbcfb23d8dbcb2e55d00853a7c (with minor simplifications)

PR #5089, opened 2017-11-21
Fix failing Grid related tests.

krmahadevan:


krmahadevan:
@barancev - This is the other PR which you had asked for as part of #5078 , to fix the failing grid related tests. Please help take a look

barancev:
Unfortunately this commit does not fix the issue too, it's a false positive again. The test itself is incorrect. It does not ckeck before cleanup if there is something in the registry. And apparently there is nothing! The created test session is not in the registry. It seems that grid behavior elolved since this test was created, and the test does not match current behaviour. It should be either modernized or deleted.

krmahadevan:
@barancev - Please help take a look.

barancev:
Merged as 2c8a1a5d59904d23085b1f04b8309ca033251740, thank you for the contribution!

PR #5116, opened 2017-11-24
Deprecate internal Duration in favor of java.time.Duration

valfirst:

Warning: breaking changes

### DIFF
--- a/java/client/src/org/openqa/selenium/support/ui/FluentWait.java
+++ b/java/client/src/org/openqa/selenium/support/ui/FluentWait.java
@@ -19,7 +19,6 @@ package org.openqa.selenium.support.ui;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
-import static java.util.concurrent.TimeUnit.SECONDS;
 
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
@@ -28,6 +27,8 @@ import com.google.common.collect.Lists;
 import org.openqa.selenium.TimeoutException;
 import org.openqa.selenium.WebDriverException;
 
+import java.time.temporal.ChronoUnit;
+import java.time.temporal.TemporalUnit;
 import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
@@ -68,14 +69,22 @@ import java.util.function.Supplier;
  */
 public class FluentWait<T> implements Wait<T> {
 
-  public static final Duration FIVE_HUNDRED_MILLIS = new Duration(500, MILLISECONDS);
+  protected final static long DEFAULT_SLEEP_TIMEOUT = 500;
+
+  /**
+   * @deprecated use {@link #DEFAULT_WAIT_DURATION}
+   */
+  @Deprecated
+  public static final Duration FIVE_HUNDRED_MILLIS = new Duration(DEFAULT_SLEEP_TIMEOUT, MILLISECONDS);
+
+  private static final java.time.Duration DEFAULT_WAIT_DURATION = java.time.Duration.ofMillis(DEFAULT_SLEEP_TIMEOUT);
 
   private final T input;
   private final Clock clock;
   private final Sleeper sleeper;
 
-  private Duration timeout = FIVE_HUNDRED_MILLIS;
-  private Duration interval = FIVE_HUNDRED_MILLIS;
+  private java.time.Duration timeout = DEFAULT_WAIT_DURATION;
+  private java.time.Duration interval = DEFAULT_WAIT_DURATION;
   private Supplier<String> messageSupplier = () -> null;
 
   private List<Class<? extends Throwable>> ignoredExceptions = Lists.newLinkedList();
@@ -102,12 +111,27 @@ public class FluentWait<T> implements Wait<T> {
    * Sets how long to wait for the evaluated condition to be true. The default timeout is
    * {@link #FIVE_HUNDRED_MILLIS}.
    *
+   * @deprecated use {@link #withTimeout(java.time.Duration)}
+   *
    * @param duration The timeout duration.
    * @param unit The unit of time.
    * @return A self reference.
    */
+  @Deprecated
   public FluentWait<T> withTimeout(long duration, TimeUnit unit) {
-    this.timeout = new Duration(duration, unit);
+    return withTimeout(java.time.Duration.of(duration, toChronoUnit(unit)));
+  }
+
+
+  /**
+   * Sets how long to wait for the evaluated condition to be true. The default timeout is
+   * {@link #DEFAULT_WAIT_DURATION}.
+   *
+   * @param timeout The timeout duration.
+   * @return A self reference.
+   */
+  public FluentWait<T> withTimeout(java.time.Duration timeout) {
+    this.timeout = timeout;
     return this;
   }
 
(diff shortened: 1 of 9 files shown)


barancev:
Changing pulic API is not easy... If we want to replace a methhof with a better alternative we have to:
1) implement a new method,
2) add a deprecation note on the old method to encourage people to migrate to the new method,
3) ship 2-3 (sometimes more) releases with both methods available,
4) delete old method in hope everybody migrated.

valfirst:
@barancev, thank you for guidance. I've updated the changes to deprecate internal Duration

barancev:
But what's motivation behind this change? Why should we break backward compatibility? What benefits will it bring?

valfirst:
@barancev, here are some examples:
 - we use external property files to setup some wait durations, having selenium Duration it's needed to create 2 properties: `time=1`, `unit=SECONDS`. But `java.time.Duration` supports [ISO_8601 duration format](https://en.wikipedia.org/wiki/ISO_8601#Durations), so it's possible to specify duration like `duration=PT1S`.
 - also sometimes it's needed to perform some manipulations with durations: sum or subtract; it's already supported by java.time.Duration.

I can see 2 ways:
 - expand selenium Duration implementation to cover more cases, no breaking public API changes, but more support for custom implementation
 - switch to `java.time.Duration`, in this case we don't care about its support and get powerful methods out of the box, but we need to introduce some API changes and move smoothly.

There is one more option: support both selenium Duration implementation, and `java.time.Duration`, but from my point of view such duplicate API could be confusing

java/client/src/org/openqa/selenium/support/ui/FluentWait.java
```diff
@@ -68,14 +69,18 @@
  */
 public class FluentWait<T> implements Wait<T> {
 
-  public static final Duration FIVE_HUNDRED_MILLIS = new Duration(500, MILLISECONDS);
+  public final static long DEFAULT_SLEEP_TIMEOUT = 500;
+
+  public static final Duration FIVE_HUNDRED_MILLIS = new Duration(DEFAULT_SLEEP_TIMEOUT, MILLISECONDS);
```
shs96c: We should mark this field deprecated. It's used once by another part of the selenium codebase, and is a trivial constant for others to declare.

java/client/src/org/openqa/selenium/support/ui/FluentWait.java
```diff
@@ -68,14 +69,18 @@
...
+
+  public static final java.time.Duration DEFAULT_WAIT_DURATION = java.time.Duration.ofMillis(DEFAULT_SLEEP_TIMEOUT);
```
shs96c: And so we don't need this to be `public`.

java/client/src/org/openqa/selenium/support/ui/FluentWait.java
```diff
@@ -102,12 +107,28 @@ public FluentWait(T input, Clock clock, Sleeper sleeper) {
...
+
+  /**
+   * Sets how long to wait for the evaluated condition to be true. The default timeout is
+   * {@link #DEFAULT_WAIT_DURATION}.
+   *
+   * @param duration The timeout duration.
+   * @param unit The unit of time.
+   * @return A self reference.
+   */
+  public FluentWait<T> withTimeout(long duration, TemporalUnit unit) {
```
shs96c: Can we please avoid changing the public APIs for now? Let's focus on removing our custom `Duration` class now.

java/client/src/org/openqa/selenium/support/ui/FluentWait.java
```diff
@@ -140,12 +161,30 @@ public FluentWait(T input, Clock clock, Sleeper sleeper) {
...
+   *
+   * <p>
+   * In reality, the interval may be greater as the cost of actually evaluating a condition function
+   * is not factored in. The default polling interval is {@link #DEFAULT_WAIT_DURATION}.
+   *
+   * @param duration The timeout duration.
+   * @param unit The unit of time.
+   * @return A self reference.
+   */
+  public FluentWait<T> pollingEvery(long duration, TemporalUnit unit) {
```
shs96c: Again, I'm not sure this is adding much to the user-facing APIs. `TimeUnit` is part of the concurrency package, and so ships with Java.

java/client/src/org/openqa/selenium/support/ui/FluentWait.java
```diff
@@ -262,4 +301,33 @@ private Throwable propagateIfNotIgnored(Throwable e) {
...
+      case SECONDS:
+        return ChronoUnit.SECONDS;
+      case MINUTES:
+        return ChronoUnit.MINUTES;
+      case HOURS:
+        return ChronoUnit.HOURS;
+      case DAYS:
+        return ChronoUnit.DAYS;
+      default:
+        throw new AssertionError();
```
shs96c: Typically, this would be an `InvalidArgumentException` instead.

java/client/src/org/openqa/selenium/support/ui/Sleeper.java
```diff
@@ -28,13 +28,29 @@
     public void sleep(Duration duration) throws InterruptedException {
```
shs96c: Surely this method is now deprecated?

java/client/src/org/openqa/selenium/support/ui/WebDriverWait.java
```diff
@@ -70,8 +69,8 @@ public WebDriverWait(WebDriver driver, long timeOutInSeconds, long sleepInMillis
   public WebDriverWait(WebDriver driver, Clock clock, Sleeper sleeper, long timeOutInSeconds,
       long sleepTimeOut) {
     super(driver, clock, sleeper);
-    withTimeout(timeOutInSeconds, TimeUnit.SECONDS);
-    pollingEvery(sleepTimeOut, TimeUnit.MILLISECONDS);
+    withTimeout(timeOutInSeconds, ChronoUnit.SECONDS);
```
shs96c: Since we're keeping the public API stable, we don't need these changes.

java/client/src/org/openqa/selenium/support/ui/FluentWait.java
valfirst: it's used in `org.openqa.selenium.lift.WebDriverTestContext`, do you propose to create another constant in the mentioned class?

java/client/src/org/openqa/selenium/support/ui/FluentWait.java
valfirst: `java.time.Duration.of(long, TemporalUnit)` requires TemporalUnit and doesn't support TimeUnit, so this method may be `private`. But in general java 8 doesn't provide easy way of conversion between `TemporalUnit` and `TimeUnit`, so the method may be useful for people who uses `TemporalUnit` and its implementation `ChronoUnit`

java/client/src/org/openqa/selenium/support/ui/FluentWait.java
valfirst: this is an exact copy of the method from Java 9, I thought it could simplify future migration to Java 9, but if you wish, I'll change it to `InvalidArgumentException`

java/client/src/org/openqa/selenium/support/ui/WebDriverWait.java
valfirst: but API becomes deprecated, shouldn't we avoid using deprecated API?

java/client/src/org/openqa/selenium/support/ui/FluentWait.java
shs96c: Java APIs concerning times and dates really are a mess. For now, I think we should follow the example of `java.util.concurrent` and stick with `TimeUnit`. Your `toChronoUnit` method hides this horror our users.

java/client/src/org/openqa/selenium/support/ui/FluentWait.java
shs96c: Same reasoning as before.

java/client/src/org/openqa/selenium/support/ui/FluentWait.java
shs96c: I guess the alternative would be to add a `withTimeout(Duration)`, which keeps us in sync with the `Sleeper`. *sigh*

java/client/src/org/openqa/selenium/support/ui/WebDriverWait.java
shs96c: Only if we deprecate the methods.

java/client/src/org/openqa/selenium/support/ui/FluentWait.java
```diff
@@ -68,14 +69,22 @@
  */
 public class FluentWait<T> implements Wait<T> {
 
-  public static final Duration FIVE_HUNDRED_MILLIS = new Duration(500, MILLISECONDS);
```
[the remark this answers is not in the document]
valfirst: This field is not removed, please, take a look at line 78, the field is kept and marked as deprecated, this is due to the way how Github displays diff-s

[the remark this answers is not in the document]
shs96c: Ah ha! Thank you :)

shs96c:
Merged as 49fe10566f107f629596b4715278dca87d3cdb7d. Thank you for this! It's great to get us to the point where we can start deleting more code :)

PR #5121, opened 2017-11-25
[py] make it possible to override options when running tests with WebKitGTK+

carlosgcampos:
This allows to run the tests with a driver in a different path instead of the installed one. The same way it allows to use a different browser or provide specific browser options. This is controlled by three optional environment variables WD_DRIVER_PATH, WD_BROWSER_PATH and WD_BROWSER_ARGS.

Also ensure overlay scrollbars are always disabled when running tests, because it causes failures in some tests and avoid including an empty list of arguments to browser options.

### DIFF
--- a/py/conftest.py
+++ b/py/conftest.py
@@ -110,7 +110,17 @@ def driver(request):
             additional_args = {}
             options = webdriver.WebKitGTKOptions()
             options.overlay_scrollbars_enabled = False
+            browser_path = os.environ.get('WD_BROWSER_PATH')
+            if browser_path is not None:
+                options.browser_executable_path = browser_path
+            browser_args = os.environ.get('WD_BROWSER_ARGS')
+            if browser_args is not None:
+                for arg in browser_args.split():
+                    options.add_browser_argument(arg)
             additional_args['options'] = options
+            driver_path = os.environ.get('WD_DRIVER_PATH')
+            if driver_path is not None:
+                additional_args['executable_path'] = driver_path
             kwargs.update(additional_args)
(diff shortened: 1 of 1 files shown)


lmtierney:
I've thought about adding these type of environ variables as an option to be used with any browser (specifically firefox since I like to bounce between stable and nightly). How about we make it work for any browser, then you make another PR for the other WebKitGTK changes that are included?

@davehunt @AutomatedTester thoughts?

carlosgcampos:
@lmtierney that's why I used generic names or the env vars, I thought this could be useful for other drivers too. But, is it really possible to make this generic? do all other drivers support this? I see executable_path parameter is used by several drivers, but browser options is specific to every driver and not all of them implement that.

davehunt:
> I like to bounce between stable and nightly

I typically just modify my `PATH` environment variable to do this. I can see the benefit of adding an environment variable for browser and driver paths, but these should be carefully considered to avoid flooding docs with inconsistent naming conventions. They should probably also be namespaced for Selenium (rather than the spec), as that's where they're being used. Ideally they'd also be consistent across the various clients.

It's also possible that the drivers themselves may also support environment variables for various options. Also, would you want a locally set environment variable to control the path of a remote instance? If not, how would you prevent that? Whilst convenient for some edge cases, we could cause additional confusion for others.

carlosgcampos:
Note that this is only for tests, it doesn't affect drivers when normally used.

AutomatedTester:
I like it. Let's land this and we can see about moving this into the bindings later

PR #5122, opened 2017-11-26
c# dotnet Add Firefox addon install using manifest.json

### LINKED ISSUE #4093: WebExtensions not installing on firefox

vflame:
Firefox addons/extensions built with WebExtensions API have a
manifest.json instead of install.rdf. When Selenium extracts the
addon/extenion id, it only looks for install.rdf. This fix adds
extracting the addon/extension id from the manifest.json in addition to
install.rdf.

Fixes #4093

### DIFF
--- a/dotnet/src/webdriver/Firefox/FirefoxExtension.cs
+++ b/dotnet/src/webdriver/Firefox/FirefoxExtension.cs
@@ -114,9 +114,9 @@ namespace OpenQA.Selenium.Firefox
             FileUtilities.DeleteDirectory(stagingDirectoryName);
         }
 
-        private static string ReadIdFromInstallRdf(string root)
+        private string ReadIdFromInstallRdf(string root)
         {
-            string id = null;
+            string id;
             try
             {
                 string installRdf = Path.Combine(root, "install.rdf");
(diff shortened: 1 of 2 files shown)


AndreSteenbergen:
Thanks for the work. I pulled in these changes and they seem to work fine.

dotnet/src/webdriver/Firefox/FirefoxExtension.cs
```diff
@@ -152,5 +156,64 @@ private static string ReadIdFromInstallRdf(string root)
...
+                    {
+                        var geckoObj = applicationObj["gecko"];
+                        if (geckoObj["id"] != null)
+                        {
+                            id = geckoObj["id"].ToString().Trim();
+                        }
+                    }
+                }
+
+                if (String.IsNullOrWhiteSpace(id))
```
AndreSteenbergen: Would probably choose for string.IsNullOrWhiteSpace (no caps)

mdul:
when we can expect that the change will be deployed?

tkempton:
These changes are desperately needed, as Firefox stopped supporting install.rdf in version 57.

When can we expect these to get merged and released?

jimevans:
Apparently, I forgot to close this PR once I merged it. Merged in 4756fd1. Thank you for the contribution.

PR #5126, opened 2017-11-27
Add chance_click method

tit:

### DIFF
--- a/rb/lib/selenium/webdriver/common/element.rb
+++ b/rb/lib/selenium/webdriver/common/element.rb
@@ -74,6 +74,21 @@ module Selenium
       end
 
       #
+      # Click this element. Or not. Random chance.
+      #
+      # @example chance click on a button
+      #
+      #    driver.find_element(:tag_name, "button").chance_click
+      # 
+      # @raise [StaleElementReferenceError] if the element no longer exists as
+      #  defined
+      #
+
+      def chance_click
+        bridge.clickElement @id if [true, false].sample
+      end
+
+      #
       # Get the tag name of the element.
       #
       # @example Get the tagname of an INPUT element(returns "input")


p0deje:
wat

p0deje:
Why would you ever want this?

Jarob22:
+1 for why would you ever want this

mstankovic86:
+1 for "WAT"

diemol:
Automated exploratory testing?

tit:
> Automated exploratory testing?
Yes. :)

I apologize for the meager description of the commit.
When you automatically test forms, you often need to click on the optional element. Or do not click.
A small analog of monkey testing.

Previously, for a random click, you had to write your own method:

```ruby
def chance_click
  yield if [true, false].sample
end

chance_click do
  driver.find_element(:tag_name, 'button').click
end
```

I suggest adding this method to the WebUpdate Selenium.

```ruby
driver.find_element(:tag_name, 'button').chance_click
```

example_form

Jarob22:
Hi @tit - generally you don't want randomised testing - you'd create one test scenario where you clicked the button, and one where you didn't, so you'd be able to test both outcomes with certainty.

tit:
@Jarob22, hi!
Do you suppose that random click is too special a case?
And it's not worth adding this feature to the main WebDriver?

Jarob22:
@tit As a general rule, webdriver aims to be a framework to build off, rather than a catch-all for all types of scenarios. I can't speak for whether the dev team will or will not accept a feature, but going back to what I said - you should really be following your own scenarios, eg:

1) When I enter my email, and select the optional checkbox and submit the form, I get output a
2) When I enter my email, and do NOT select optional checkbox and submit the form, I get output b

Rather than randomly clicking it.

jimevans:
I’m -1 on this change for the general-purpose WebDriver language bindings. .NET will not be implementing this.

p0deje:
@tit This should definitely belong to the client code, not Selenium.

Thank you for a pull request, but I agree with Jim on this.

PR #5174, opened 2017-12-06
Improving capability matcher

diemol:
to handle case when nodes or clients use non deprecated `CapabilityType.PLATFORM_NAME`

When I create a node using `CapabilityType.PLATFORM_NAME` instead of `CapabilityType.PLATFORM` (now deprecated), the requested capabilities won't match the desired capabilities since the `DefaultCapabilityMatcher` is not considering `CapabilityType.PLATFORM_NAME` to make the match.

This PR, changes this, so `CapabilityType.PLATFORM_NAME` is considered to make the match as a fallback when `CapabilityType.PLATFORM` is not present.

I also duplicated the unit tests to check that both CapabilityTypes are working, and added two more at the end covering the scenario when the provided capabilities have `CapabilityType.PLATFORM` and the requested `CapabilityType.PLATFORM_NAME` and viceversa.

### DIFF
--- a/java/server/src/org/openqa/grid/internal/utils/DefaultCapabilityMatcher.java
+++ b/java/server/src/org/openqa/grid/internal/utils/DefaultCapabilityMatcher.java
@@ -31,6 +31,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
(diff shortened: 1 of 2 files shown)


barancev:
Merged as ba71bfa132bac3f250f94da72927ddb3881a3137. Thank you!

PR #5230, opened 2017-12-18
BrowserStack PlatformNotSupportedException fix for dotnetcore

ischyrus:
Adding a PlatformNotSupportedException catch around ConnectionLimit setting to enable usage on dotnetcore when using the RemoteWebDirver with something like BrowserStack.

Also explicitly setting the proxy to null, which solves a PlatformNotSupportedException for dotnetcore on Linux\Ubuntu.

With these two changes, I am able to use Selenium with BrowserStack.

### DIFF
--- a/java/client/src/org/openqa/selenium/net/NetworkInterface.java
+++ b/java/client/src/org/openqa/selenium/net/NetworkInterface.java
@@ -80,6 +80,10 @@ public class NetworkInterface {
     return iterator.hasNext() && iterator.next().isLoopbackAddress();
   }
 
+  /**
+   * @deprecated Will have reduced visibility in later releases.
+   */
+  @Deprecated
   public InetAddress getIp4LoopbackOnly() {
     // Goes by the wildly unscientific assumption that if there are more than one set of
     // loopback addresses, firefox will bind to the last one we get.


jimevans:
Is this still required with the change to setting the `Proxy` property to `null`? In my testing, setting the property to null is sufficient to prevent the `PlatformNotSupportedException`.

ischyrus:
Ah, so you made the changes without taking my specific pull request?

ischyrus:
I spent quite a bit of time debugging and figuring out why that didn't work. I would have appreciated the merge because it would have appeared on my GitHub page and looked really good when I applied at jobs. 

It's really not cool to not accept peoples pull requests when they fix bugs for you. It's not likely I'll be doing this again for you.

jimevans:
To be honest, I didn't look at this particular pull request when making the change. I apologize for not thoroughly examining every pending pull request when simply fixing the bug. I understand the frustration at not having it merged, and I hope you can see your way to forgiving the mistake and the oversight.

ischyrus:
No worries. Perhaps next time!

cgoldberg:
@ischyrus 

> it would have appeared on my GitHub page and
> looked really good when I applied at jobs. 

... and threads like this one will look really bad when you apply for jobs.

Issue #5159, opened 2017-12-02
Download file in headless chrome


jonhkr:
## Meta -
OS:  OSX
Selenium Version:  3.8.1
Browser:  Chrome

## Expected Behavior -
Permit files to be downloaded in headless mode.

## Actual Behavior -
Files aren't downloaded.

## Steps to reproduce -
Launch chrome in headless mode and try to download any file.

Looks like it is a feature to prevent sites from downloading files when running chrome in headless mode.
https://bugs.chromium.org/p/chromium/issues/detail?id=696481

To permit downloads it's necessary to send a command to chrome.
I'm using java and I couldn't find a way to make it work.

I even tried sending a post request to `/session/:sessionId/chromium/send_command` but that didn't work.

p0deje:
Apparently, this new command needs to be added to all the bindings.

jonhkr:
I managed to get it working by issuing the above request directly to the `chromedriver` but using a development build of chromium.

jonhkr:
This is how I did it:

```Java
ChromeDriverService driverService = ChromeDriverService.createDefaultService();

ChromeDriver driver = new ChromeDriver(driverService, options);

Map<String, Object> commandParams = new HashMap<>();
commandParams.put("cmd", "Page.setDownloadBehavior");

Map<String, String> params = new HashMap<>();
params.put("behavior", "allow");
params.put("downloadPath", downloadFilepathString);
commandParams.put("params", params);

ObjectMapper objectMapper = new ObjectMapper();
HttpClient httpClient = HttpClientBuilder.create().build();

String command = objectMapper.writeValueAsString(commandParams);

String u = driverService.getUrl().toString() + "/session/" + driver.getSessionId() + "/chromium/send_command";

HttpPost request = new HttpPost(u);
request.addHeader("content-type", "application/json");
request.setEntity(new StringEntity(command));
httpClient.execute(request);
```

nstubi:
Hello,

Any plans to support setDownloadBehavior in order to download files using chrome in headless mode?

Any doc/examples in .Net would be much appreciated.

Regards,
Nicolas

shreyanshp:
Even I am facing this same issue

pulkitsharma07:
@p0deje, @jonhkr 
I am working on integrating this with the Ruby bindings, but for some reason I still not able to download in headless mode.

[...]

PR #5262, opened 2017-12-26
Preparing chrome driver to download in headless mode

IlyaNaumovich:
The method send commands to the browser to avoid blocking
of downloading in headless mode.
(The issue is described here:
https://github.com/SeleniumHQ/selenium/issues/5159)

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/5262)

### DIFF
--- a/java/client/src/org/openqa/selenium/chrome/ChromeDriver.java
+++ b/java/client/src/org/openqa/selenium/chrome/ChromeDriver.java
@@ -179,9 +179,9 @@ public class ChromeDriver extends RemoteWebDriver
   @Deprecated
   public ChromeDriver(ChromeDriverService service, Capabilities capabilities) {
     super(new ChromeDriverCommandExecutor(service), capabilities);
-    if (capabilities.getCapability("chromium:enableDownloading") instanceof String) {
+    if (capabilities.getCapability("chromium:enableDownloadTo") instanceof String) {
       sendCommandForDownloadChromeHeadless(
-          (String) capabilities.getCapability("chromium:enableDownloading"));
+          (String) capabilities.getCapability("chromium:enableDownloadTo"));
     }
     locationContext = new RemoteLocationContext(getExecuteMethod());
     webStorage = new RemoteWebStorage(getExecuteMethod());
--- a/java/client/src/org/openqa/selenium/chrome/ChromeDriver.java
+++ b/java/client/src/org/openqa/selenium/chrome/ChromeDriver.java
@@ -30,7 +30,6 @@ import org.openqa.selenium.html5.WebStorage;
 import org.openqa.selenium.interactions.HasTouchScreen;
 import org.openqa.selenium.interactions.TouchScreen;
 import org.openqa.selenium.mobile.NetworkConnection;
-import org.openqa.selenium.remote.DriverCommand;
 import org.openqa.selenium.remote.FileDetector;
 import org.openqa.selenium.remote.RemoteTouchScreen;
 import org.openqa.selenium.remote.RemoteWebDriver;
@@ -173,19 +172,19 @@ public class ChromeDriver extends RemoteWebDriver
    * Creates a new ChromeDriver instance. The {@code service} will be started along with the
    * driver, and shutdown upon calling {@link #quit()}.
    *
-   * @param service The service to use.
+   * @param service      The service to use.
    * @param capabilities The capabilities required from the ChromeDriver.
    * @deprecated Use {@link ChromeDriver(ChromeDriverService, ChromeOptions)} instead.
    */
   @Deprecated
   public ChromeDriver(ChromeDriverService service, Capabilities capabilities) {
     super(new ChromeDriverCommandExecutor(service), capabilities);
-    if (capabilities.getCapability("enableDownloading")!=null){
-      sendCommandForDownloadChromeHeadLess(
-          (String) capabilities.getCapability("enableDownloading"));
+    if (capabilities.getCapability("chromium:enableDownloading") instanceof String) {
+      sendCommandForDownloadChromeHeadless(
+          (String) capabilities.getCapability("chromium:enableDownloading"));
     }
     locationContext = new RemoteLocationContext(getExecuteMethod());
-    webStorage = new  RemoteWebStorage(getExecuteMethod());
+    webStorage = new RemoteWebStorage(getExecuteMethod());
     touchScreen = new RemoteTouchScreen(getExecuteMethod());
     networkConnection = new RemoteNetworkConnection(getExecuteMethod());
   }
@@ -243,15 +242,11 @@ public class ChromeDriver extends RemoteWebDriver
 
   /**
    * Send command for browser to allow downloading in headless mode
-   *
-   * @param downloadPath
    */
-  public void sendCommandForDownloadChromeHeadLess(String downloadPath) {
-    if (downloadPath.length()>0){
-      execute(DriverCommand.SEND_COMMAND_TO_BROWSER,
-              ImmutableMap.of("cmd", "Page.setDownloadBehavior",
-                              "params",
-                              ImmutableMap.of("behavior", "allow","downloadPath", downloadPath)));
-    }
+  public void sendCommandForDownloadChromeHeadless(String downloadPath) {
+    execute(ChromeDriverCommand.SEND_COMMAND_TO_BROWSER,
+            ImmutableMap.of("cmd", "Page.setDownloadBehavior",
+                            "params",
+                            ImmutableMap.of("behavior", "allow", "downloadPath", downloadPath)));
   }
 }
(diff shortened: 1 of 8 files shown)


p0deje:
Instead of providing a single method which sets this command, we should actually provide a method that allows sending arbitrary commands to DevTools. After that, anyone can easily use it to allow downloading.

In Ruby, such API can look like this:

```ruby
driver.send_chromium_command(cmd: 'Page.setDownloadBehavior', params: {behavior: 'allow', downloadPath: '/User/paulo/projects/app/tmp'})
```

barancev:
Also don't forget adding tests that cover new code.

barancev:
By the way, why do you create a new http client instead of using CommandExecutor to send a command?

barancev:
I've thought about this feature a bit more. Yes, a function to send custom commands is a good thing, we should add it. But in this special case it looks like a workaround. Implementing a capability that enables/disables file download would be much more in line with Selenium strategy to handle such features.

shs96c:
Why isn't this simply making a change to the `ChromeDriverCommandExecutor`? Maybe I've missed something....

java/client/src/org/openqa/selenium/remote/http/JsonHttpCommandCodec.java
```diff
@@ -133,6 +134,8 @@ public JsonHttpCommandCodec() {
     defineCommand(TOUCH_MOVE, post("/session/:sessionId/touch/move"));
     defineCommand(TOUCH_SCROLL, post("/session/:sessionId/touch/scroll"));
     defineCommand(TOUCH_UP, post("/session/:sessionId/touch/up"));
+
+    defineCommand(SEND_COMMAND_TO_BROWSER, post("/session/:sessionId/chromium/send_command_and_get_result"));
```
shs96c: This is Chrome specific, and so shouldn't be in the general JsonHttpCommandCodec. The best place is `ChromeDriverCommandExecutor`

java/client/src/org/openqa/selenium/chrome/ChromeDriver.java
```diff
@@ -179,6 +180,10 @@ public ChromeDriver(ChromeDriverService service, ChromeOptions options) {
   @Deprecated
   public ChromeDriver(ChromeDriverService service, Capabilities capabilities) {
     super(new ChromeDriverCommandExecutor(service), capabilities);
+    if (capabilities.getCapability("enableDownloading")!=null){
```
shs96c: Stylistic comment: Prefer a little more space around things: `) != null) {`

java/client/src/org/openqa/selenium/chrome/ChromeDriver.java
```diff
@@ -179,6 +180,10 @@ public ChromeDriver(ChromeDriverService service, ChromeOptions options) {
...
+      sendCommandForDownloadChromeHeadLess(
+          (String) capabilities.getCapability("enableDownloading"));
```
shs96c: You may get a `ClassCastException` here. There are two ways to avoid this: either make the check above be `capabilities.getCapability("enableDownloading") instanceof String`, or use `String.valueOf` here.

shs96c: `enableDownloading` is not a legal capability name. `chromium:enableDownloading` is. The reasoning is here: https://w3c.github.io/webdriver/webdriver-spec.html#dfn-extension-capability

java/client/src/org/openqa/selenium/chrome/ChromeDriver.java
```diff
@@ -236,4 +241,17 @@ public void launchApp(String id) {
     execute(ChromeDriverCommand.LAUNCH_APP, ImmutableMap.of("id", id));
   }
 
+  /**
+   * Send command for browser to allow downloading in headless mode
+   *
+   * @param downloadPath
+   */
+  public void sendCommandForDownloadChromeHeadLess(String downloadPath) {
+    if (downloadPath.length()>0){
```
shs96c: What happens if the path is zero length? That seems like an unusual position to be in. Log or throw an error?

shs96c: Stylistic comment: `if (!downloadPath.isEmpty()) {` is preferable.

java/client/src/org/openqa/selenium/chrome/ChromeDriverCommand.java
```diff
@@ -24,4 +24,7 @@
   private ChromeDriverCommand() {}
 
   static final String LAUNCH_APP = "launchApp";
+  static final String SEND_COMMANDS_FOR_DOWNLOAD_CHROME_HEAD_LESS
+      = "sendCommandForDownloadChromeHeadLess";
```
shs96c: Nit: `Headless` is one word, not two, and so doesn't need camelcasing in this way.

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -235,6 +236,15 @@ public ChromeOptions setAcceptInsecureCerts(boolean acceptInsecureCerts) {
     return this;
   }
 
+  public ChromeOptions setEnableDownloading(boolean enableDownloading, String downloadPath){
+    if (enableDownloading){
+      setCapability(ENABLE_DOWNLOADING, downloadPath);
+    }else {
+      setCapability(ENABLE_DOWNLOADING, "");
```
shs96c: Setting the capability to `null` will remove it from the map, which is probably more desirable.

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -235,6 +236,15 @@ public ChromeOptions setAcceptInsecureCerts(boolean acceptInsecureCerts) {
     return this;
   }
 
+  public ChromeOptions setEnableDownloading(boolean enableDownloading, String downloadPath){
+    if (enableDownloading){
```
shs96c: Stylistic note: whitespace around braces: `if (enableDownloading) {`.

shs96c (commented):
This is super-close to being ready. Thanks for waiting for us. 

I've added a couple of "stylistic comments": we can fix those ourselves, but it'd be nice if you could fix those too.

java/client/src/org/openqa/selenium/chrome/ChromeDriver.java
jleyba: nit: We're not the Chromium project and shouldn't introduce capabilities in their namespace.

How about using the `selenium:chrome` prefix?

java/client/src/org/openqa/selenium/remote/CapabilityType.java
```diff
@@ -47,6 +47,7 @@
   String ELEMENT_SCROLL_BEHAVIOR = "elementScrollBehavior";
   String HAS_TOUCHSCREEN = "hasTouchScreen";
   String OVERLAPPING_CHECK_DISABLED = "overlappingCheckDisabled";
+  String ENABLE_DOWNLOADING = "chromium:enableDownloading";
```
jleyba: The name `enableDownloading` implies this is a boolean capability. How about `downloadDir`?

IlyaNaumovich: Thank you for review. I believe that "downloadDir" is not so appropriate in this case because it sounds like specifying download directory rather than applying download behavior for specific path. It could cause some misunderstanding as to why we introduce this capability. What about AllowHeadlessDownloadTo?
From one hand, it is boolean, from the other hand, it specify the downloading directory. There are no ways to allow downloading without specifying directory for downloading. There are no way to specify directory for downloading without allowing downloading.

jleyba: It's not specific to headless, so how about `allowDownloadsTo`?

IlyaNaumovich:
@Barancev, @Shs96c, @jleyba I'm just wondering, maybe you want me to add something to this pull request

lmtierney:
After talking with the chromedriver team, the `send_command` will be moved to a w3c-compliant protocol extension endpoint (https://w3c.github.io/webdriver/webdriver-spec.html#protocol-extensions). We may want to hold off on this PR until the change.

See the issue raised in the chromium bugtracker: https://bugs.chromium.org/p/chromedriver/issues/detail?id=2307

shs96c:
Do I need to continue looking at this PR?

timwindsor:
The chromium change is to the api interface but they still support the call used in this PR. The changes here are necessary in order to use the API that chromium is providing.

AutomatedTester:
@shs96c Are we still going to accept this patch?

diemol:
This went stale a while ago, and probably it should be possible through the new interface with CDP. If there is the need to have this in the Java tree, please talk to us in our Slack/IRC [channel](https://www.selenium.dev/support/).

Issue #1973, opened 2016-04-15
"Command line server for the IE driver has stopped working" with IEDriverServer.exe since upgrade to 2.53.0.0


johnnyreilly:
## Meta -
OS: Windows Server 2008 R2 Standard 64 bit
Selenium Version: 2.53.0.0
Browser: IE
Browser Version: IE 11.0.9600.17801

## Expected Behavior -
Tests to run reliably on build server

## Actual Behavior -
Tests fail sporadically with "Command line server for the IE driver has stopped working"

## Steps to reproduce -
Migrate to 2.53.0.0
Use IEDriver - these are my packages:

```xml
  <package id="Selenium.Support" version="2.53.0" targetFramework="net452" />
  <package id="Selenium.WebDriver" version="2.53.0" targetFramework="net452" />
  <package id="Selenium.WebDriver.IEDriver" version="2.53.0.0" targetFramework="net452" />
```

![image](https://cloud.githubusercontent.com/assets/1010525/14556062/17dc65dc-02ee-11e6-9d38-da4860ad6999.png)

I'm going to try and rollback to using 2.50.1 (which I don't remember presenting this error)

johnnyreilly:
I was also have a dump file (55Mb) which I can share if useful.  Don't want to dump this on GitHub though - so get it contact and I'll hold onto it.

jimevans:
With no more information than this to go on, it's impossible to take any action on this issue report. Please provide code that causes the issue, along with an HTML page to test against, and/or a log file from IEDriverServer.exe from when the problem occurs.

johnnyreilly:
Hi @jimevans,

I'm afraid it's code that I'm not permitted to share :disappointed: 

I appreciate that doesn't help a great deal and that probably means you can't take any action.  But there's been no changes in the test pack apart from upgrading the packages and so it seems likely there is some kind of problem.  

Sorry I can't be more helpful but I'm afraid my hands are tied.  Thanks for Selenium by the way - it's great!

johnnyreilly:
Looks like the problem may be in `Selenium.WebDriver.IEDriver`.  I've just moved to these package versions:

```xml
  <package id="Selenium.Support" version="2.53.0" targetFramework="net452" />
  <package id="Selenium.WebDriver" version="2.53.0" targetFramework="net452" />
  <package id="Selenium.WebDriver.IEDriver" version="2.49.0.0" targetFramework="net452" />
```

[...]

PR #5304, opened 2018-01-08
Use a mutex to lock Win32 message handling

Xaeroxe:

I use Selenium at work and for our test bed we were having many problems with the IEDriverServer crashing due to a Memory Access Violation. It got to the point where it was impeding our work, so I cloned the project and ran it in debug mode to see if I could identify the problem.

Our test bed was causing the driver to crash specifically on line 441 of IECommandExecutor.cpp, the inner message handling loop for Win32. I believe the crash was occurring due to concurrent use of the Win32 APIs by multiple threads, which do not appear to be thread safe. So I introduced a static mutex shared between threads and had each thread acquire a lock to this mutex before processing messages. This has worked wonderfully for us, our tests are now much more stable. I'd like to contribute this fix back to Selenium.

I believe this may have also been the cause for #1973 and similar issues.

### DIFF
--- a/cpp/iedriver/IECommandExecutor.cpp
+++ b/cpp/iedriver/IECommandExecutor.cpp
@@ -439,6 +439,8 @@ unsigned int WINAPI IECommandExecutor::ThreadProc(LPVOID lpParameter) {
         LOG(DEBUG) << "Returned from DestroyWindow()";
         break;
       } else {
+        // We need to lock this mutex here to make sure only one thread is processing
+        // win32 messages at a time.
         static std::mutex messageLock;
         std::lock_guard<std::mutex> lock(messageLock);
         ::TranslateMessage(&msg);


cpp/iedriver/IECommandExecutor.cpp
```diff
@@ -438,6 +439,8 @@ unsigned int WINAPI IECommandExecutor::ThreadProc(LPVOID lpParameter) {
         LOG(DEBUG) << "Returned from DestroyWindow()";
         break;
       } else {
+        static std::mutex messageLock;
+        std::lock_guard<std::mutex> lock(messageLock);
```
shs96c: I'm not a Win32 hacker, so I'm not sure how this fixes the stability issue. The docs for `GetMessage`, `TranslateMessage`, and `DispatchMessage` suggest they only operate on the current thread. Is this something to do with multiple threads attempting to post messages to the same `HWND`?

cpp/iedriver/IECommandExecutor.cpp
```diff
@@ -438,6 +439,8 @@ unsigned int WINAPI IECommandExecutor::ThreadProc(LPVOID lpParameter) {
         LOG(DEBUG) << "Returned from DestroyWindow()";
         break;
       } else {
+        static std::mutex messageLock;
```
shs96c: Can we have a comment explaining why the mutex is necessary?

shs96c:
Anything that improves stability is a Good Thing. Thank you for putting this PR up again.

cpp/iedriver/IECommandExecutor.cpp
Xaeroxe: It appears to me that this code is indeed called from multiple threads.  In a typical win32 program these go on what I call the "UI thread" but that's not the context they're being used in here.  Perhaps I'm wrong and this is only called in one thread, in which case this PR is useless, but I believe there are multiple threads using this.

cpp/iedriver/IECommandExecutor.cpp
Xaeroxe: Sure!

cpp/iedriver/IECommandExecutor.cpp
shs96c: Huh, so although the docs suggest that these methods are thread safe, they're not really because the normal use case is that they're only used on the UI thread? Interesting.

Your observed increase in stability is a Good Thing, and I'm happy to trust your reasoning :)

cpp/iedriver/IECommandExecutor.cpp
shs96c: I'll keep on eye on this PR for that :) Thank you!

Xaeroxe:
Comment added, sorry for the delay.  Let me know if you feel the explanation is satisfactory.

shs96c:
@Xaeroxe Thank you! I've merged the PR and it'll be in the next release.

PR #5314, opened 2018-01-09
Fix Exception being raised when error_payload does not include a value.

mcking49:
Sometimes the selenium server returns no value in a error_payload what makes an exception to be raised trying to chain another hash reference.
This will test for existence of ['value'] key before the next hash reference to be used.


p0deje:
Thank you, merged in 157cc12f0f

PR #5326, opened 2018-01-11
[py] reduce the default script timeout in executing_async_javascript_…

carlosgcampos:
…tests.py

When using w3c driver, the default script timeout is 30 seconds, so
tests expecting to timeout and not overriding the script timeout take 30
seconds to run. Setting it to 0 makes the tests a lot faster. Then, it's
reset to the default value once tests finish.


lmtierney:
Merged, thanks!

PR #5341, opened 2018-01-15
Use old actions for geckodriver

HarryMtv:

At now many actions are not available in firefox with geckodriver. Can you please approve the little fix, till the problem is not resolved?

### DIFF
--- a/java/client/src/org/openqa/selenium/interactions/Actions.java
+++ b/java/client/src/org/openqa/selenium/interactions/Actions.java
@@ -25,6 +25,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Sets;
 
+import org.openqa.selenium.InvalidArgumentException;
 import org.openqa.selenium.Keys;
 import org.openqa.selenium.UnsupportedCommandException;
 import org.openqa.selenium.WebDriver;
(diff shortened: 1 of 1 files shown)


barancev:
What do you mean by "old actions for geckodriver"? Geckodriver has implemented only one variation of actions, the one described in the W3C WebDriver specification.

martynovig:
@barancev it is code terminology, at the first line we trying to use W3C WebDriver specification (implemented with sequences), but if we can't trying to use old way.
`try {
        ((Interactive) driver).perform(sequences.values());
      } catch (ClassCastException | UnsupportedCommandException | InvalidArgumentException e) {
        // Fall back to the old way of doing things. Old Skool #ftw
        fallBack.perform();
      }`

At now if gekodriver if i use the moveToElement (for example), i'll get exception like
`org.openqa.selenium.InvalidArgumentException: Could not find webelement key` 
if we catch this exception and use fallBack action it is work correctly.

barancev:
geckodriver does not support "old actions" protocol, it's just impossible to fall back to it.

Please explain in more details what issue resolves this PR, provide a reproduction scenario.

martynovig:
@barancev Hi, you right, the problem was in another framework, this PR can be like workaround for old actions or can be closed. Thanks again.

shs96c:
Thank you for the PR! Closing this without merging: the logic here is only trying to determine whether or not we can use the W3C end point or not. If an `IllegalArgumentException` is thrown, something else has gone wrong.

PR #5360, opened 2018-01-18
[py] Use Color in rendered_webelement_tests

carlosgcampos:
WebDriver spec decided that CSS values returned should be whatever the
browser returns, without imposing any format for colors. By using Color
class in the tests we can compare colors even if different formats (RGB
vs RGBA) are used.


lmtierney:
Merged, thanks :)

PR #5368, opened 2018-01-19
[rb] Optimised some of the calls to some ruby integration tests

luke-hill:
Essentially a small refactor to use the Memo on Line52 which appears to be unused, and speed up a portion of the `target_locator` tests and all of the `window` tests.

I've also made the Full Screen test more in-line with the Maximise test (Including adding in a short wait), just because sometimes this was a touch flaky locally (And the extra time is dwarfed by the reduction from less unrequired windows).

As mentioned in the commit msg, I've opted against changing all of these because there was a few comments surrounding tests involving more than 2 windows.


p0deje:
It's good to merge, but there is a conflict. Can you resolve it first?

p0deje:
Merged in 49691026d8023f4c1a7dd438a513bca75c3d69ae. Thank you!

PR #5375, opened 2018-01-20
Replace slashes with backslashes in path during file upload on Windows

hron:
Although Windows allows both type of slashes in path Firefox doesn't
allow slashes. Ruby uses slashes on Windows for file path. This leads to
very frustrating exceptions and disallows writing cross-OSes tests
easily. You have to create some kind of helper to convert paths in your
tests so they would work on both OSes.

This affects capybara for example. Check out teamcapybara/capybara#1950
for the details.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/5375)

### DIFF
--- a/rb/lib/selenium/webdriver/common/bridge_helper.rb
+++ b/rb/lib/selenium/webdriver/common/bridge_helper.rb
@@ -74,6 +74,15 @@ module Selenium
 
         result
       end
+
+      def ensure_native_file_path(keys)
+        if WebDriver::Platform.windows?
+          keys.each do |keys_chunk|
+            keys_chunk.tr!('/', '\\') if File.exist?(keys_chunk)
+          end
+        end
+        keys
+      end
     end # BridgeHelper
   end # WebDriver
 end # Selenium
--- a/rb/lib/selenium/webdriver/firefox/marionette/bridge.rb
+++ b/rb/lib/selenium/webdriver/firefox/marionette/bridge.rb
@@ -20,6 +20,7 @@ module Selenium
     module Firefox
       module Marionette
         module Bridge
+          include BridgeHelper
 
           COMMANDS = {
             install_addon: [:post, 'session/:session_id/moz/addon/install'.freeze],
@@ -41,12 +42,7 @@ module Selenium
           end
 
           def send_keys_to_element(element, keys)
-            if WebDriver::Platform.windows?
-              keys.each do |keys_chunk|
-                keys_chunk.tr!('/', '\\') if File.exist?(keys_chunk)
-              end
-            end
-            super(element, keys)
+            super(element, ensure_native_file_path(keys))
           end
         end # Bridge
       end # Marionette
--- a/rb/lib/selenium/webdriver/firefox/marionette/bridge.rb
+++ b/rb/lib/selenium/webdriver/firefox/marionette/bridge.rb
@@ -40,6 +40,14 @@ module Selenium
             execute :uninstall_addon, {}, {id: id}
           end
 
+          def send_keys_to_element(element, keys)
+            if WebDriver::Platform.windows?
+              keys.each do |keys_chunk|
+                keys_chunk.tr!('/', '\\') if File.exist?(keys_chunk)
+              end
+            end
+            super(element, keys)
+          end
         end # Bridge
       end # Marionette
     end # Firefox
--- a/rb/lib/selenium/webdriver/ie/bridge.rb
+++ b/rb/lib/selenium/webdriver/ie/bridge.rb
@@ -15,22 +15,16 @@
 # specific language governing permissions and limitations
 # under the License.
 
-require 'selenium/webdriver/ie/driver'
-require 'selenium/webdriver/ie/bridge'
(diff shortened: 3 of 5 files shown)


p0deje:
This won't work with IE is used with server/Grid. I really don't see why you would want to do this in selenium-webdriver and not in your client code. You can add conditional checking that platform is Windows and browser is IE and `tr!` slashes by yourself.

hron:
As a user I just want the following code to work without cryptic messages "File not found" on Windows (`attach_file` is capybara's API)

```ruby
attach_file File.join('fixtures', 'Test.txt')
```

I find it really strange when I have to care about slashes in so high level testing library like Capybara. 

@twalpole thinks that the better place to fix this is in selenium-webdriver. I don't have yet enough expertise in these two libraries,  but maybe you folks can point me how to finalize the patch and where should it go?

hron:
I have stumbled upon this issue when I worked with Marionette, but then I realized that that my change brokes IE, so I tried to fix it. That's why the patch affects both drivers.

Other languages might work just fine because it is a ruby creators decision to have slashes on both file path types regardless of operation system. I believe other languages don't adopt this technique.

jleyba:
I don't understand the problem here. File paths should either be specified relative to the PWD (and client libraries should normalize as necessary for the current system), or construct an absolute path using language-specific APIs. For java:

```java
// Path should be resolved as java.nio.file.Paths.get("relative/path/to/foo.txt");
element.sendKeys("relative/path/to/foo.txt");

// Non-portable test. Will not be recognized as a file path on Linux
element.sendKeys("C:\path\to\foo.txt");

Path path = someBasePath.resolve("path/to/foo.txt");
element.sendKeys(path.toAbsolutePath().toString());
```

If running against a remote Selenium server, the client library should detect `sendKeys()` was given a path to a file, transfer that file to the remote server, then invoke `sendKeys()` with the file path returned by the Selenium server -- everything should "just work".

Here's some tests for this functionality in [node](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/test/upload_test.js#L54) and [java](https://github.com/SeleniumHQ/selenium/blob/master/java/client/test/org/openqa/selenium/UploadTest.java#L71)


If you're running against a remote WebDriver server that __not__ the Selenium server, then it's your responsibility to ensure the file (a) exists on the remote machine and (b) you specify the correct path to `sendKeys()` to trigger attaching the file to the browser.

hron:
As I already mentioned other languages might not be affected. The problem, most likely, localized to Ruby because even when you use the language-specific API the result string has forward slashes, not back-slashes. Here is an example:

```ruby
irb(main):005:0> Pathname.getwd.to_s
"E:/src/selenium"              
```

Notice the *forward* slashes. Windows can works with both type of slashes, but when you try to pass such kind of path to Firefox with sendKeys you will get "File not found" error. So, even when you use the Ruby language-specific library to represent the path on Windows you have replace slashes manually. Which is kind of annoying. 

Chrome works with both type of slashes in path.

hron:
Just take a look at this [integration test](https://github.com/SeleniumHQ/selenium/blob/1cffee4cb9e0b5eccd57ef26fcb6d64ce0719a5d/rb/spec/integration/selenium/webdriver/element_spec.rb#L79) in Selenium codebase (notice the `tr` call for replacing the slashes in the path). This basically what a Selenium user must do to write a cross-platform test with file uploading in ruby.

```ruby
it 'should handle file uploads', except: {browser: %i[safari edge safari_preview]} do
  driver.navigate.to url_for('formPage.html')

  element = driver.find_element(id: 'upload')
  expect(element.attribute('value')).to be_empty

  file = Tempfile.new('file-upload')
  path = file.path
  path.tr!('/', '\\') if WebDriver::Platform.windows?

  element.send_keys path

  expect(element.attribute('value')).to include(File.basename(path))
end
```

twalpole:
The fact this works seamlessly with Chrome would suggest `chromedriver` is already handling this based on the platform it knows it is running on (WebDriver::Platform.windows? isn't a valid test for this when using a grid).  It may be something to suggest to `geckodriver` and the W3C WebDriver spec to clarify what is acceptable as directory separators.

shs96c:
The only place that this can reliably work is in the end node (that is, the actual thing driving the browser) since that's the only place that can be entirely certain which OS the browser is running on. Any hacking and patching we do in Selenium doesn't resolve that issue.

shs96c:
Sorry. Closed by accident.

p0deje:
I believe Simon summed it up well - this can only be done in the end node, so client bindings cannot fix that. The only solution I can see to this problem would be an addition to [WebDriver specification](https://w3c.github.io/webdriver/#element-send-keys) that will instruct the drivers to convert `/` to `\` on Windows. This way clients will always be sending `/` and not worrying about cross-platform compatibility. Would you mind creating an issue at https://github.com/w3c/webdriver/issues?

diemol:
I will close this as the PR and comments went stale. If there are more questions related to this, feel free to drop by our Slack/IRC channel, check the links here https://www.selenium.dev/support/

Issue #5279, opened 2018-01-02
Merging of `*Option` classes doesn't work as it did for DesiredCapabilities


JohnA2:
## Meta -
OS:
Windows 7
Selenium Version:
3.6.0 - 3.8.1
Browser:
Internet Explorer,  Firefox and Chrome
Browser Version:
any

## Expected Behavior -
Browser specific parameters in descendants of `MutableCapabilities` work after merging them.

## Actual Behavior -
Browser specific parameters are ignored because they are lost after using `merge`.

## Steps to reproduce -
**Short example:**
```
ChromeOptions options1 = new ChromeOptions();
ChromeOptions options2 = new ChromeOptions();
options2.addArguments("--start-maximized");
options1.merge(options2);
WebDriver driver = new ChromeDriver(options1); // Chrome doesn't start maximized
```

**Long example:**
I use a "SeleniumConfig" class to store all WebDriver related configuration, including its `Capabilities`. There are two types of the SeleniumConfig objects - the global that applies to all test classes and locals that only apply to one test class. So I need to merge global and local Capabilities when initializing a new test class.

A simplified version of my code looks like this:
```
// Class that contains all configuration for a WebDriver
class SeleniumConfig {
	private String browser = BrowserType.FIREFOX;
	// ... other parameters
	private DesiredCapabilities capabilities = new DesiredCapabilities();
	
	public SeleniumConfig() {}
	public SeleniumConfig(SeleniumConfig configToCopy) {
		this.browser = configToCopy.getBrowser();
		this.capabilities = new DesiredCapabilities(configToCopy.getCapabilities());
	}
	public String getBrowser() {
		return browser;
	}
	public void setBrowser(String browser) {
		this.browser = browser;
	}
	public DesiredCapabilities getCapabilities() {
		return capabilities;
	}
	public void addCapability(String key, Object value) {
		this.capabilities.setCapability(key, value);
	}
}

// A singleton that holds configuration for all test classes
class GlobalSettings {
	private static final GlobalSettings instance = new GlobalSettings();
	private SeleniumConfig global = new SeleniumConfig();

	private GlobalSettings() { // Read global configuration from some source
		// For example start Chrome maximized
		global.setBrowser(BrowserType.CHROME);
		ChromeOptions options = new ChromeOptions();
		options.addArguments("--start-maximized");
		global.addCapability(ChromeOptions.CAPABILITY, options);
	}
	public static SeleniumConfig getGlobalSettings() {
[...]

PR #5390, opened 2018-01-24
Improvement for merging browser options

IlyaNaumovich:
(issue #5279)
https://github.com/SeleniumHQ/selenium/issues/5279

The issue is improved only for Chrome; it will improved
 by me for other browsers a bit later.

The following logic for merging was implemented:
1. For ‘binary’ field: if this fields are both not empty
 and their values are different, the IllegalArgumentException
 should be thrown;
2. For ‘experimentalOptions’ field: if the source options
and receiver options have the same experimental option with
different values, the IllegalArgumentException should be thrown;
3. After merging all duplicates in all collections should be deleted.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/5390)

### DIFF
--- a/java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
+++ b/java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
@@ -17,23 +17,36 @@
 
 package org.openqa.selenium.chrome;
 
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.openqa.selenium.remote.CapabilityType.ACCEPT_INSECURE_CERTS;
+import static org.openqa.selenium.remote.CapabilityType.PAGE_LOAD_STRATEGY;
+import static org.openqa.selenium.remote.CapabilityType.UNEXPECTED_ALERT_BEHAVIOUR;
+import static org.openqa.selenium.remote.CapabilityType.UNHANDLED_PROMPT_BEHAVIOUR;
+
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.io.Files;
-import org.openqa.selenium.*;
+
+import org.openqa.selenium.Capabilities;
+import org.openqa.selenium.MutableCapabilities;
+import org.openqa.selenium.PageLoadStrategy;
+import org.openqa.selenium.Proxy;
+import org.openqa.selenium.SessionNotCreatedException;
+import org.openqa.selenium.UnexpectedAlertBehaviour;
 import org.openqa.selenium.remote.BrowserType;
 import org.openqa.selenium.remote.CapabilityType;
 
 import java.io.File;
 import java.io.IOException;
-import java.lang.reflect.Field;
-import java.util.*;
+import java.util.Base64;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.TreeMap;
 import java.util.stream.Stream;
-
-import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.base.Preconditions.checkNotNull;
-import static org.openqa.selenium.remote.CapabilityType.*;
 
 /**
  * Class to manage options specific to {@link ChromeDriver}.
@@ -76,76 +89,7 @@ public class ChromeOptions extends MutableCapabilities {
   @Override
   public ChromeOptions merge(Capabilities extraCapabilities) {
     super.merge(extraCapabilities);
-    if (extraCapabilities instanceof ChromeOptions) {
-      Map options = (Map) extraCapabilities.asMap().get("goog:chromeOptions");
-      mergeArguments(options);
-      mergeExtensions(options);
-      mergeExtensionFiles(extraCapabilities);
-      mergeBinary(options);
-      mergeExperimentalOptions(extraCapabilities);
-    }
     return this;
-  }
-
-  private void mergeArguments(Map options) {
-    addArguments((List<String>) options.get("args"));
-    deleteDuplicates((ArrayList) args);
-  }
-
-  private void mergeExtensions(Map options) {
-    if (options.get("extensions") != null) {
-      addEncodedExtensions((List<String>) options.get("extensions"));
-      deleteDuplicates((ArrayList) extensions);
-    }
-  }
-
-  private void mergeExtensionFiles(Capabilities extraCapabilities) {
-    try {
-      Field extensionFilesField = ChromeOptions.class.getDeclaredField("extensionFiles");
-      extensionFilesField.setAccessible(true);
-      List<File> extensionFilesList = (List<File>) extensionFilesField.get(extraCapabilities);
-      for (File path: extensionFilesList) {
-        addExtensions(path);
-      }
-      deleteDuplicates((ArrayList) extensionFiles);
-    } catch (NoSuchFieldException | IllegalAccessException e) {
-      e.printStackTrace();
-    }
-  }
-
-  private void mergeBinary(Map options) {
-    if (options.get("binary") != null) {
-      if (binary == null){
-        binary = (String) options.get("binary");
-      } else if (options.get("binary") != binary) {
-        throw new IllegalArgumentException("Binaries can't be merged");
-      }
-    }
-  }
-
-  private void mergeExperimentalOptions(Capabilities extraCapabilities) {
-    try {
-      Field experimentalOptionField = ChromeOptions.class.getDeclaredField("experimentalOptions");
-      experimentalOptionField.setAccessible(true);
-      HashMap<String, Object> experimentalOptionList = (HashMap<String, Object>) experimentalOptionField.get(extraCapabilities);
-      for (Map.Entry<String, Object> experimentalOption : experimentalOptionList.entrySet()) {
-        if (experimentalOptions.get(experimentalOption.getKey()) == null) {
-          experimentalOptions.put(experimentalOption.getKey(), experimentalOption.getValue());
-        } else if (!(experimentalOptions.get(experimentalOption.getKey()).toString()
-            .contentEquals(experimentalOption.getValue().toString()))) {
-          throw new IllegalArgumentException("Experimental options can't be merged");
-        }
-      }
-    } catch (NoSuchFieldException | IllegalAccessException e) {
-      e.printStackTrace();
-    }
-  }
-
-  private void deleteDuplicates(ArrayList al) {
-    Set hs = new HashSet<>();
-    hs.addAll(al);
-    al.clear();
-    al.addAll(hs);
   }
 
   /**
@@ -221,7 +165,7 @@ public class ChromeOptions extends MutableCapabilities {
       checkNotNull(path);
       checkArgument(path.exists(), "%s does not exist", path.getAbsolutePath());
       checkArgument(!path.isDirectory(), "%s is a directory",
-          path.getAbsolutePath());
+              path.getAbsolutePath());
     }
     extensionFiles.addAll(paths);
     return this;
@@ -286,6 +230,11 @@ public class ChromeOptions extends MutableCapabilities {
     return this;
   }
 
+  /**
+   * Returns ChromeOptions with the capability ACCEPT_INSECURE_CERTS set.
+   * @param acceptInsecureCerts
+   * @return ChromeOptions
+   */
   public ChromeOptions setAcceptInsecureCerts(boolean acceptInsecureCerts) {
     setCapability(ACCEPT_INSECURE_CERTS, acceptInsecureCerts);
     return this;
@@ -321,11 +270,11 @@ public class ChromeOptions extends MutableCapabilities {
   @Override
   protected int amendHashCode() {
     return Objects.hash(
-        args,
-        binary,
-        experimentalOptions,
-        extensionFiles,
-        extensions);
+            args,
+            binary,
+            experimentalOptions,
+            extensionFiles,
+            extensions);
   }
 
   @Override
@@ -343,18 +292,18 @@ public class ChromeOptions extends MutableCapabilities {
     options.put("args", ImmutableList.copyOf(args));
 
     options.put(
-        "extensions",
-        Stream.concat(
-            extensionFiles.stream()
-                .map(file -> {
-                  try {
-                    return Base64.getEncoder().encodeToString(Files.toByteArray(file));
-                  } catch (IOException e) {
-                    throw new SessionNotCreatedException(e.getMessage(), e);
-                  }
-                }),
-            extensions.stream()
-        ).collect(ImmutableList.toImmutableList()));
+            "extensions",
+            Stream.concat(
+                    extensionFiles.stream()
+                            .map(file -> {
+                              try {
+                                return Base64.getEncoder().encodeToString(Files.toByteArray(file));
+                              } catch (IOException e) {
+                                throw new SessionNotCreatedException(e.getMessage(), e);
+                              }
+                            }),
+                    extensions.stream()
+            ).collect(ImmutableList.toImmutableList()));
 
     toReturn.put(CAPABILITY, options);
 
--- a/java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
+++ b/java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
@@ -17,36 +17,23 @@
 
 package org.openqa.selenium.chrome;
 
-import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.base.Preconditions.checkNotNull;
-import static org.openqa.selenium.remote.CapabilityType.ACCEPT_INSECURE_CERTS;
-import static org.openqa.selenium.remote.CapabilityType.PAGE_LOAD_STRATEGY;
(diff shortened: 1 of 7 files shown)


java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -17,37 +17,24 @@
...
 import com.google.common.collect.Maps;
 import com.google.common.io.Files;
-
-import org.openqa.selenium.Capabilities;
-import org.openqa.selenium.MutableCapabilities;
-import org.openqa.selenium.PageLoadStrategy;
-import org.openqa.selenium.Proxy;
-import org.openqa.selenium.SessionNotCreatedException;
-import org.openqa.selenium.UnexpectedAlertBehaviour;
+import org.openqa.selenium.*;
```
shs96c: We don't use global imports like this as modern IDEs will simply hide and manage them for you.

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -17,37 +17,24 @@
...
 import java.io.File;
 import java.io.IOException;
-import java.util.Base64;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.TreeMap;
+import java.lang.reflect.Field;
+import java.util.*;
```
shs96c: And here.

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -89,9 +76,78 @@ public ChromeOptions() {
   @Override
   public ChromeOptions merge(Capabilities extraCapabilities) {
     super.merge(extraCapabilities);
+    if (extraCapabilities instanceof ChromeOptions) {
+      Map options = (Map) extraCapabilities.asMap().get("goog:chromeOptions");
```
shs96c: Please use generics. If there's no obvious type, the `Map<?, ?>` prevents compiler warnings, is correct, and is equally meaningless :)

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -89,9 +76,78 @@ public ChromeOptions() {
...
+      mergeExtensionFiles(extraCapabilities);
+      mergeBinary(options);
+      mergeExperimentalOptions(extraCapabilities);
+    }
     return this;
   }
 
+  private void mergeArguments(Map options) {
+    addArguments((List<String>) options.get("args"));
+    deleteDuplicates((ArrayList) args);
```
shs96c: This `deleteDuplicates` phase would be avoided entirely if you just added all the arguments to an `ImmutableSet` or `LinkedHashSet`. These preserve insertion order, but won't allow identical instances.

shs96c: Alternatively, stream the new args, and only add them if they're not present.

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -89,9 +76,78 @@ public ChromeOptions() {
...
+  }
+
+  private void mergeExtensions(Map options) {
+    if (options.get("extensions") != null) {
```
shs96c: Placing the entire method into a block makes it easy to miss things. Either exit quickly (by checking for `null`, and returning), or use `options.getOrDefault()` with a sensible value.

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -89,9 +76,78 @@ public ChromeOptions() {
...
+    }
     return this;
   }
 
+  private void mergeArguments(Map options) {
+    addArguments((List<String>) options.get("args"));
+    deleteDuplicates((ArrayList) args);
+  }
+
+  private void mergeExtensions(Map options) {
```
shs96c: Generics, please!

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -89,9 +76,78 @@ public ChromeOptions() {
...
+    if (options.get("extensions") != null) {
+      addEncodedExtensions((List<String>) options.get("extensions"));
+      deleteDuplicates((ArrayList) extensions);
```
shs96c: As above.

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -89,9 +76,78 @@ public ChromeOptions() {
...
+    }
+  }
+
+  private void mergeExtensionFiles(Capabilities extraCapabilities) {
+    try {
+      Field extensionFilesField = ChromeOptions.class.getDeclaredField("extensionFiles");
```
shs96c: Why not just access the field?

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -89,9 +76,78 @@ public ChromeOptions() {
...
+      extensionFilesField.setAccessible(true);
+      List<File> extensionFilesList = (List<File>) extensionFilesField.get(extraCapabilities);
+      for (File path: extensionFilesList) {
+        addExtensions(path);
+      }
+      deleteDuplicates((ArrayList) extensionFiles);
+    } catch (NoSuchFieldException | IllegalAccessException e) {
+      e.printStackTrace();
```
shs96c: Rethrow the exception, even if you're sure it'll never happen. A `WebDriverException` would work here.

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -89,9 +76,78 @@ public ChromeOptions() {
...
+    }
+  }
+
+  private void mergeBinary(Map options) {
```
shs96c: Generics, please!

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -89,9 +76,78 @@ public ChromeOptions() {
...
+    if (options.get("binary") != null) {
```
shs96c: Again, please avoid putting the entire method into a block.

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -89,9 +76,78 @@ public ChromeOptions() {
...
+      if (binary == null){
+        binary = (String) options.get("binary");
+      } else if (options.get("binary") != binary) {
+        throw new IllegalArgumentException("Binaries can't be merged");
```
shs96c: Well, here's an interesting thing: how do we know that none of the options passed in elsewhere aren't mutually exclusive in the same way?

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -89,9 +76,78 @@ public ChromeOptions() {
...
+      }
+    }
+  }
+
+  private void mergeExperimentalOptions(Capabilities extraCapabilities) {
+    try {
+      Field experimentalOptionField = ChromeOptions.class.getDeclaredField("experimentalOptions");
```
shs96c: Why access the field this way?

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -89,9 +76,78 @@ public ChromeOptions() {
...
+      experimentalOptionField.setAccessible(true);
+      HashMap<String, Object> experimentalOptionList = (HashMap<String, Object>) experimentalOptionField.get(extraCapabilities);
+      for (Map.Entry<String, Object> experimentalOption : experimentalOptionList.entrySet()) {
+        if (experimentalOptions.get(experimentalOption.getKey()) == null) {
+          experimentalOptions.put(experimentalOption.getKey(), experimentalOption.getValue());
+        } else if (!(experimentalOptions.get(experimentalOption.getKey()).toString()
```
shs96c: You already know that `getKey` returns `String` --- there's no need for the additional `toString` call

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -89,9 +76,78 @@ public ChromeOptions() {
...
+        throw new IllegalArgumentException("Binaries can't be merged");
+      }
+    }
+  }
+
+  private void mergeExperimentalOptions(Capabilities extraCapabilities) {
+    try {
+      Field experimentalOptionField = ChromeOptions.class.getDeclaredField("experimentalOptions");
+      experimentalOptionField.setAccessible(true);
+      HashMap<String, Object> experimentalOptionList = (HashMap<String, Object>) experimentalOptionField.get(extraCapabilities);
```
shs96c: Prefer coding to interfaces rather than implementations --- `Map` not `HashMap`. That means that the underlying implementation is free to pick anything that makes sense to it.

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -89,9 +76,78 @@ public ChromeOptions() {
...
+      for (Map.Entry<String, Object> experimentalOption : experimentalOptionList.entrySet()) {
+        if (experimentalOptions.get(experimentalOption.getKey()) == null) {
+          experimentalOptions.put(experimentalOption.getKey(), experimentalOption.getValue());
+        } else if (!(experimentalOptions.get(experimentalOption.getKey()).toString()
+            .contentEquals(experimentalOption.getValue().toString()))) {
+          throw new IllegalArgumentException("Experimental options can't be merged");
+        }
+      }
+    } catch (NoSuchFieldException | IllegalAccessException e) {
+      e.printStackTrace();
```
shs96c: Please rethrow rather than swallow

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -89,9 +76,78 @@ public ChromeOptions() {
...
+    }
+  }
+
+  private void deleteDuplicates(ArrayList al) {
+    Set hs = new HashSet<>();
```
shs96c: Generics, please!

java/client/src/org/openqa/selenium/chrome/ChromeOptions.java
```diff
@@ -89,9 +76,78 @@ public ChromeOptions() {
...
+            .contentEquals(experimentalOption.getValue().toString()))) {
+          throw new IllegalArgumentException("Experimental options can't be merged");
+        }
+      }
+    } catch (NoSuchFieldException | IllegalAccessException e) {
+      e.printStackTrace();
+    }
+  }
+
+  private void deleteDuplicates(ArrayList al) {
```
shs96c: Generics, please! Also, interface rather than implementation (`List` rather than `ArrayList`)

java/client/src/org/openqa/selenium/firefox/FirefoxOptions.java
```diff
@@ -17,40 +17,26 @@
...
 import java.io.IOException;
+import java.lang.reflect.Field;
 import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.TreeMap;
+import java.util.*;
```
shs96c: Same comment as before about avoiding global imports like this.

java/client/src/org/openqa/selenium/firefox/FirefoxOptions.java
```diff
@@ -393,8 +379,94 @@ public FirefoxOptions addCapabilities(Capabilities capabilities) {
   @Override
   public FirefoxOptions merge(Capabilities capabilities) {
     super.merge(capabilities);
+    if (capabilities instanceof FirefoxOptions) {
+      Map options = (Map) capabilities.asMap().get("moz:firefoxOptions");
```
shs96c: Generics, please!

java/client/src/org/openqa/selenium/firefox/FirefoxOptions.java
```diff
@@ -393,8 +379,94 @@ public FirefoxOptions addCapabilities(Capabilities capabilities) {
...
   }
+  
+  private void mergeArguments(Map options) {
+    addArguments((List<String>) options.get("args"));
+    deleteDuplicates((ArrayList) args);
+  }
+  
+  private void mergeBooleanPrefs(Capabilities capabilities) {
+    try {
+      Field booleanPrefsField =FirefoxOptions.class.getDeclaredField("booleanPrefs");
```
shs96c: Why not just use the field?

java/client/test/org/openqa/selenium/chrome/ChromeOptionsMergeUnitTest.java
```diff
@@ -0,0 +1,237 @@
...
+import java.util.Random;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Unit tests for {@link ChromeOptions}.
+ */
+
+public class ChromeOptionsMergeUnitTest {
```
shs96c, line 34: Please follow the project style. The easiest way to do this is to load the project in the Community Edition of IntelliJ and ask it to reformat the code.

java/client/test/org/openqa/selenium/chrome/ChromeOptionsMergeUnitTest.java
```diff
@@ -0,0 +1,237 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Unit tests for {@link ChromeOptions}.
```
shs96c: This comment adds nothing. Please delete.

java/client/test/org/openqa/selenium/chrome/ChromeOptionsMergeUnitTest.java
```diff
@@ -0,0 +1,237 @@
...
+public class ChromeOptionsMergeUnitTest {
+
+	private TemporaryFilesystem tmpFs;
+	private File tempDir;
+	private File firstTempFile;
+	private File secondTempFile;
+
+	@Before
+	public void setUp() throws Exception {
+		File baseForTest = new File(System.getProperty("java.io.tmpdir"), "tmpTest");
```
shs96c: Prefer using a `@Rule` of `org.junit.rules.TemporaryFolder`

java/client/test/org/openqa/selenium/chrome/ChromeOptionsMergeUnitTest.java
```diff
@@ -0,0 +1,237 @@
...
+		tmpFs = TemporaryFilesystem.getTmpFsBasedOn(baseForTest);
+		tempDir = tmpFs.createTempDir("tempDir", "");
+		firstTempFile = new File(tempDir, "firstTempFile.txt");
+		secondTempFile = new File(tempDir, "secondTempFile.txt");
+		writeTestFile(firstTempFile);
+		writeTestFile(secondTempFile);
+	}
+
+	@After
+	public void tearDown() throws Exception {
```
shs96c: Because then you don't need to do this. :)

java/client/test/org/openqa/selenium/chrome/ChromeOptionsMergeUnitTest.java
```diff
@@ -0,0 +1,237 @@
...
+
+	@Before
+	public void setUp() throws Exception {
+		File baseForTest = new File(System.getProperty("java.io.tmpdir"), "tmpTest");
+		baseForTest.mkdir();
+		tmpFs = TemporaryFilesystem.getTmpFsBasedOn(baseForTest);
+		tempDir = tmpFs.createTempDir("tempDir", "");
+		firstTempFile = new File(tempDir, "firstTempFile.txt");
+		secondTempFile = new File(tempDir, "secondTempFile.txt");
+		writeTestFile(firstTempFile);
```
shs96c: Prefer creating the binary only for the tests that actually use it.

shs96c (changes requested):
We lose symmetry with this change. For example:

```
FirefoxOptions options = new FirefoxOptions();
MutableCapabilities caps = new MutableCapabilities();
caps.put("moz:firefoxOptions", ImmutableMap.of());

// These give different results.
options.merge(caps);
caps.merge(options);
```

That's going to cause a lot of hard to track down bugs. As it stands, right now, the result is easy to predict. Merging the values might make sense, but maintaining symmetry is really important to prevent confusion.

IlyaNaumovich:
@shs96c, when you get a chance, could you please take a look at changes?

barancev:
I strongly disagree with including any driver-specific code into `MutableCapabilities`. Merging driver-specific properties should be implemented in subclasses.

Implementation for Chromium (Chrome + Edgium) has been landed in commit ec9c67b2fb1ec110f51934832fec3cef8a9bf746

Implementation for Firefox is to be repaired yet, but I think it'd better presented as a separate PR.

PR #5406, opened 2018-01-30
Added no-cache header to HTTP GET command of RemoteWebDriver

### LINKED ISSUE #5404: .net HttpCommandExecutor lacks "Cache-Control: no-cache" header for HTTP GET commands
I am using .net Selenium WebDriver implementation to execute tests on cloud-based virtual machine.
Test code is executed on my local PC within corporate network with proxy server.
Test code uses OpenQA.Selenium.Remote.RemoteWebDriver class to send commands to remote virtual machine.

It looks like proxy server caches Selenium's HTTP GET requests and returns "cached" data which is outdated. E.g. method ITakesScreenshot.GetScreenshot() returns outdated screenshots.

In Java version of Selenium WebDriver this is not happening because of this code:
`    if (HttpMethod.GET == spec.method) {
      request.setHeader(CACHE_CONTROL, "no-cache");
    }
`
Similar code exists for Ruby WebDriver implementation.

However .net implementation doesn't have it.

dnknitro:
When Selenium RemoteWebDriver is connecting to Selenium Server
(e.g. SauceLabs) through HTTP proxy, same/repetitive HTTP commands
can be cached by the proxy. This leads to unexpected and unwanted
behavior that HTTP commands return cached and outdated data from the
server (e.g. screenshots).

To fix this I have added Cache-Control: no-cache HTTP header.

Selenium Java implementation has this fix for a while – at least from
version 2.26.

Fixes #5404

### DIFF
--- a/dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
+++ b/dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
@@ -163,6 +163,10 @@ namespace OpenQA.Selenium.Remote
                 requestStream.Write(data, 0, data.Length);
                 requestStream.Close();
             }
+            else if (request.Method == CommandInfo.GetCommand)
+            {
+                request.Headers.Add("Cache-Control", "no-cache");
+            }
 
             HttpResponseInfo responseInfo = new HttpResponseInfo();
             HttpWebResponse webResponse = null;


jimevans:
Merged in b019171. Thanks for the contribution

PR #5441, opened 2018-02-05
Adding more expected conditions to dotnet

overbit:

### DIFF
--- a/dotnet/src/support/UI/ExpectedConditions.cs
+++ b/dotnet/src/support/UI/ExpectedConditions.cs
@@ -247,7 +247,7 @@ namespace OpenQA.Selenium.Support.UI
                 {
                     var allChildren = parent.FindElements(childLocator);
 
-                    return !allChildren.Any() ? allChildren : throw new NoSuchElementException();
+                    return allChildren.Any() ? allChildren : throw new NoSuchElementException();
                 }
                 catch (NoSuchElementException)
                 {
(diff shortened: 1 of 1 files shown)


jimevans:
While I appreciate the contribution, it will be exceedingly rare to add any additional methods to the .NET `ExpectedConditions` class. If you feel like you have a set of methods that would constitute “best practices” (whatever that means), you should create a repository under a separate project to distribute them.

overbit:
I don't understand why we want to keep this inconsistency between java and dotnet driver. Any particular reason?

https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/support/ui/ExpectedConditions.java

jimevans:
@overbit Because Java isn't .NET, and vice versa. The use of lambda functions ("anonymous methods") has been a staple of .NET development since .NET 3.0 (released in 2006), and the syntax for creating them is well understood by .NET developers. Java, by contrast, didn't pick up proper lambda functions until Java 8, and the syntax for passing in a function to be called before then was cumbersome and not familiar to many. Because of the confusing syntax, Java needed an `ExpectedConditions` class for many Selenium developers; .NET developers didn't and don't (or shouldn't) need similar helpers.

overbit:
@jimevans I agree on the usage of lambda functions but my point is more focused on the usability and the help that this change could bring to users. It might be not common to see expected conditions on child elements but it still have a valid purpose. 
Moreover, Id like to suggest to make the class not sealed in order to make users  able to extend it with custom made conditions that they think are more appropriate with their testing. Is there any obvious point that I can see about the sealed class choice?

jimevans:
I'm going to close this PR, since the .NET `ExpectedConditions` implementation has been marked as obsolete. The code has been migrated to a new repository under a new organization on Github. Please note that the new repository is currently looking an owner, and until one is found, PRs and issue reports will not be accepted.

PR #5444, opened 2018-02-06
Handle basic authentication in HttpClient-s

valfirst:

### DIFF
--- a/java/client/src/org/openqa/selenium/remote/internal/ApacheHttpClient.java
+++ b/java/client/src/org/openqa/selenium/remote/internal/ApacheHttpClient.java
@@ -232,10 +232,11 @@ public class ApacheHttpClient implements org.openqa.selenium.remote.http.HttpCli
     public org.openqa.selenium.remote.http.HttpClient createClient(URL url) {
       checkNotNull(url, "null URL");
       HttpClient client;
-      if (url.getUserInfo() != null) {
-        StringTokenizer tokens = new StringTokenizer(url.getUserInfo(), ":");
+      String urlUserInfo = url.getUserInfo();
+      if (urlUserInfo != null) {
+        UserInfo userInfo = new UserInfo(urlUserInfo);
         UsernamePasswordCredentials credentials =
-            new UsernamePasswordCredentials(tokens.nextToken(), tokens.nextToken());
+            new UsernamePasswordCredentials(userInfo.getUser(), userInfo.getPassword());
         client = clientFactory.createHttpClient(credentials);
       } else {
         client = clientFactory.getHttpClient();


java/client/src/org/openqa/selenium/remote/internal/OkHttpClient.java
```diff
@@ -116,11 +117,20 @@ public void close() {
...
           .connectionPool(pool)
           .followRedirects(true)
-          .followSslRedirects(true)
-          .build();
+          .followSslRedirects(true);
+
+      String urlUserInfo = url.getUserInfo();
+      if (urlUserInfo != null) {
+        String[] userInfo = urlUserInfo.split(":");
+        String basic = Credentials.basic(userInfo[0], userInfo[1]);
```
shs96c: We need a bounds check here. If `urlUserInfo` lacks a `:` we'll only have `userInfo[0]`

java/client/src/org/openqa/selenium/remote/internal/OkHttpClient.java
```diff
@@ -116,11 +117,20 @@ public void close() {
...
-          .followSslRedirects(true)
-          .build();
+          .followSslRedirects(true);
+
+      String urlUserInfo = url.getUserInfo();
+      if (urlUserInfo != null) {
+        UserInfo userInfo = new UserInfo(urlUserInfo);
+        String basic = Credentials.basic(userInfo.getUser(), userInfo.getPassword());
+        builder.addInterceptor(chain ->
+          chain.proceed(chain.request().newBuilder().header("Authorization", basic).build())
```
shs96c: This will loop endlessly if requests aren't authorized properly, according to the okhttp docs.

java/client/src/org/openqa/selenium/remote/internal/OkHttpClient.java
```diff
@@ -116,11 +117,20 @@ public void close() {
...
+      okhttp3.OkHttpClient.Builder builder = new okhttp3.OkHttpClient.Builder()
           .connectionPool(pool)
           .followRedirects(true)
-          .followSslRedirects(true)
-          .build();
+          .followSslRedirects(true);
+
+      String urlUserInfo = url.getUserInfo();
+      if (urlUserInfo != null) {
+        UserInfo userInfo = new UserInfo(urlUserInfo);
```
shs96c: I'm not sure we need this new class. It's enough to just split the string and grab the bits.

java/client/src/org/openqa/selenium/remote/internal/UserInfo.java
```diff
@@ -0,0 +1,29 @@
+package org.openqa.selenium.remote.internal;
+
+public final class UserInfo
```
shs96c: We can reformat this code, but it's unusual for it not to follow the project's style.

shs96c:
I landed 97ae9816eccab7d1e87e141d4f0a4e8102080e7d as this is basically just pulling in the okhttp recipe, but I'll be sure to give you credits in the CHANGELOG. Thank you for the PR!

java/client/src/org/openqa/selenium/remote/internal/OkHttpClient.java
valfirst: there is exactly the same logic (and exactly the same issue) in Apache HTTP client, so IMHO it worth following DRY principle

java/client/src/org/openqa/selenium/remote/internal/OkHttpClient.java
valfirst: it's not so, this code adds just an interceptor, not `Authenticator`, because it allows to avoid an extra HTTP call, please check https://stackoverflow.com/a/36056355/2067574 for more details

valfirst:
@shs96c , please check my comments above

PR #5451, opened 2018-02-07
Allow inheriting FindsByAttribute

Dreamescaper:

Make FindsByAttribute not sealed and make Finder property `protected internal` instead of 'internal'.

Why needed? In case when we're using custom finders, it would be easier to define own attribute instead of using How.Custom.
So instead of
`[FindsBy(How = How.Custom, CustomFinderType = typeof(CustomBy), Using = "value")]`
you could do
`[FindsByCustom(Using = "value")]`
or
`[FindsByCustom("value")]`

### DIFF
--- a/dotnet/src/support/PageObjects/FindsByAttribute.cs
+++ b/dotnet/src/support/PageObjects/FindsByAttribute.cs
@@ -1,4 +1,4 @@
-﻿// <copyright file="FindsByAttribute.cs" company="WebDriver Committers">
+// <copyright file="FindsByAttribute.cs" company="WebDriver Committers">
 // Licensed to the Software Freedom Conservancy (SFC) under one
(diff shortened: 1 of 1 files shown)


jimevans:
Unsealed attributes violate [Microsoft .NET coding guidelines CA1813](https://msdn.microsoft.com/en-us/library/ms182267.aspx), and unsealing will cause warnings under static code analysis.

Dreamescaper:
I can go in different  way.
Create interface IFindsBy with property Binder, and make FindsByAttribute implement this interface.
This way any attribute, implementing IFindsBy would be supported by PageFactory.

Dreamescaper:
Created different PR, tackling same problem:
https://github.com/SeleniumHQ/selenium/pull/5456

PR #5452, opened 2018-02-07
Make annoying debug console user aware

mascIT-98:
With this commit user can now choice wheter the debug command prompt can or not show it up when driver is loaded. This happens when python script are compiled into one-file exe.
With this commit merged, the user can easily disable the cmd prompt in this way:
```
args = ["hide_console", ]
driver = webdriver.Chrome("path-to-your-driver", service_args=args, ...)
```

### DIFF
--- a/py/selenium/webdriver/common/service.py
+++ b/py/selenium/webdriver/common/service.py
@@ -47,7 +47,7 @@ class Service(object):
         self.start_error_message = start_error_message
         self.log_file = log_file
         self.env = env or os.environ
-
+        
     @property
     def service_url(self):
         """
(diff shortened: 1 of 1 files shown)


py/selenium/webdriver/common/service.py
```diff
@@ -69,11 +69,10 @@ def start(self):
         try:
             cmd = [self.path]
             cmd.extend(self.command_line_args())
-            self.process = subprocess.Popen(cmd, env=self.env,
-                                            close_fds=platform.system() != 'Windows',
-                                            stdout=self.log_file,
-                                            stderr=self.log_file,
-                                            stdin=PIPE)
+            if any("hide_console" in arg for arg in self.command_line_args()):
```
cgoldberg: this is equivalent to:
`if 'hide_console' in self.command_line_args():`

py/selenium/webdriver/common/service.py
```diff
@@ -69,11 +69,10 @@ def start(self):
...
+                self.process = subprocess.Popen(cmd, stdin=PIPE, stdout=PIPE, stderr=PIPE, creationflags=0x08000000)
+            else:
```
cgoldberg: * Passing `creationflags` to `subprocess.Popen` will break on all non-Windows platforms and raise a `ValueError`.
* the `env` and `close_fds` were removed
* args were changed to `stdin=PIPE, stdout=PIPE, stderr=PIPE`, which breaks log file redirection

cgoldberg:
I don't think there should be a special command line arg passed to workaround this.  recommend rejecting this PR.

netzulo:
same think that @cgoldberg , more tests or reject , don'k track dead issues :)

py/selenium/webdriver/common/service.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, executable, port=0, log_file=DEVNULL, env=None, start_error_m
         self.start_error_message = start_error_message
         self.log_file = log_file
         self.env = env or os.environ
-
+
```
netzulo: can just remove it ?

cgoldberg:
closing this PR.

netzulo:
ty @cgoldberg

Issue #5024, opened 2017-11-07
IEDriverServer considerably slower between 3.4 and 3.7


rhpijnacker:
## Meta -
OS:  Windows Server 2012
Selenium Version: .NET bindings, 3.5.1, IEDriverServer 3.4 / 3.7
Browser:  Internet Explorer
Browser Version:  11.0.9600.17416

We recently tried upgrading from Selenium 3.4 to a newer version. After the upgrade, our tests are running considerably slower. After some trial and error, we found that using the previous version of the IEDriverVersion (3.4) restored the tests to their original speed. 
The slow-down is roughly 1.5 to 2 times (varies per test).
We added some trace logging, and the logfiles have also become much bigger (log files can be provided if useful).

I have checked with versions 3.5, 3.6 and 3.7, which are all affected.

## Expected Behavior -
Tests run with the same speed using IEDriverServer 3.4 or the latest

## Actual Behavior -
Tests run much slower with the latest version

## Steps to reproduce -
Just run the test with either version 3.4 or 3.7 of the IEDriverServer

jimevans:
I strongly suspect the issue has happened in the refactoring of the actions pipeline in 3.5 to coincide with the making the driver compliant with the W3C WebDriver Specification. Pull requests welcome.

AdvanceAutomation:
Had this issue when I tried to update to 3.6.0. But then recognized I was using 32 bit IEDriverServer while my browser was 64-bit which made it super slow. Updating to correct version fixed the issue. Please confirm its not the same in your case.

rhpijnacker:
@jimevans : Hm... that does not sound like something that would be easy to jump into without any prior knowledge, would it?

rhpijnacker:
@AdvanceAutomation : We are using the 32 bit version of the IEDriverServer.

jimevans:
@rhpijnacker What roadmap? Of course, it’s something I want to try to look at, but since I work on he IE driver on an uncompensated volunteer basis, it’ll have to wait for availability, absent someone else putting in the time and effort.

rhpijnacker:
@jimevans Roadmap/todolist/sticky notes, whatever you prefer ; )

AdvanceAutomation:
@rhpijnacker  try 64 bit IEDriver and if you still see the issue then its a new bug.
Also I hope you understand Selenium is open source.  People are putting their free time and contributing to it. Its people like @jimevans  who make this a possibility. I would suggest supporting them.

[...]

PR #5455, opened 2018-02-07
Correct a part of timing issue #5024

Nawakator:

Remove unecessary loop for mouse move.

### DIFF
--- a/cpp/iedriver/InputManager.cpp
+++ b/cpp/iedriver/InputManager.cpp
@@ -865,6 +865,7 @@ void InputManager::AddPauseInput(HWND window_handle, int duration) {
 }
 
 void InputManager::AddMouseInput(HWND window_handle, long input_action, int x, int y) {
+  LOG(TRACE) << "Entering InputManager::AddMouseInput";
   INPUT mouse_input;
   mouse_input.type = INPUT_MOUSE;
   mouse_input.mi.dwFlags = input_action | MOUSEEVENTF_ABSOLUTE;


jimevans:
This pull request would make the driver no longer compliant with the [W3C WebDriver Specification](https://w3c.github.io/webdriver/webdriver-spec.html). Please pay particular attention to the ["Perform a pointer move"](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-perform-a-pointer-move) algorithm, which both specific Actions and element click ([sub-step 11 of the non-`<option>` element branch in step 7 of the Element Click remote end steps](https://w3c.github.io/webdriver/webdriver-spec.html#element-click)) are required to perform. It might be possible to tweak the duration of the move in the `WebElement.click()` case, but that's not what this PR does.

Nawakator:
Indeed.

So could intend us to add a capabilities for switch the W3C compilant for Internet Explorer? (make IE W3C compilant it s a 'joke' no).

To explain the real pb: since the 3.5.1 we cannot compare the result beetween Internet Explorer and Chrome.
And this difference was only procude by the mouse move and other input events. So we cannot compare the real browser performance.

jimevans:
No, having the IE driver be compliant with the WebDriver spec is certainly no “joke.” The spec refers to the driver, not the browser, and it’s a spec I’ve personally spent more than five years helping to bring into existence. So, no, we won’t be adding a capability to disable compliance with the spec. In any case, your PR does not attempt to add that capability, so if that is your proposal, the PR is incomplete. As submitted, this PR is not an acceptable solution to the issue.

g0rk4:
Hi @jimevans, as i understand and in order to stick to the W3C specifications, the best way to proceed is to set an option to reduce the mouse movement speed ?

jimevans:
@g0rk4 No,the right approach would be to fix the logic error in the calculation of the mouse move duration. I’m working on just such a fix later today.

jimevans:
@Nawakator InputManager.cpp has been thoroughly refactored since this PR was created. If you still believe it to be necessary, please update and resubmit. I will be closing this PR in one week’s time.

jimevans:
Closing PR because of lack of review after refactored input handling in IE.

PR #5456, opened 2018-02-07
Allow creating custom FindsBy attributes

Dreamescaper:

Why needed? In case when we're using custom finders, it would be easier to define own attribute instead of using How.Custom.
So instead of
`[FindsBy(How = How.Custom, CustomFinderType = typeof(CustomBy), Using = "value")]`
you could do
`[FindsByCustom(Using = "value")]`
or
`[FindsByCustom("value")]`

What is done:
1) Created AbstractFindsByAttribute, with abstract get property Finder. Priority property, and all compare and equals methods are moved there from FindsByAttribute
2) FindsByAttribute now inherits AbstractFindsByAttribute
3) Added test for custom FindsBy attribute.

So in order to implement custom FindsBy attribute, it is required to inherit from AbstractFindsByAttribute, and implement Finder property.

Remaining question - there was internal setter for property Finder in FindsByAttribute, but it wasn't used anywhere. I removed it, but maybe I should leave it in AbstractFindsByAttribute ?

### DIFF
--- a/dotnet/src/support/PageObjects/AbstractFindsByAttribute.cs
+++ b/dotnet/src/support/PageObjects/AbstractFindsByAttribute.cs
@@ -1,0 +1,189 @@
+// <copyright file="FindsByAttribute.cs" company="WebDriver Committers">
+// Licensed to the Software Freedom Conservancy (SFC) under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The SFC licenses this file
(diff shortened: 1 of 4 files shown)


jimevans:
I'm going to close this PR, since the .NET `PageFactory` implementation (and all of its supporting classes) have been marked as obsolete. The code has been migrated to a [new repository under a new organization on Github](https://github.com/DotNetSeleniumTools/DotNetSeleniumExtras). Please note that the new repository is currently looking an owner, and until one is found, PRs and issue reports will not be accepted.

PR #5465, opened 2018-02-08
It should be possible to use a custom safaridriver executable to run …

burg:
…Selenium's test suite.

Add the 'use_existing' argument to the safari.WebDriver class initializer. This
argument causes the driver to not launch its own instance of safaridriver. Instead,
it connects via port to an externally-launched service. This can speed up testing
and is also useful when debugging problems with safaridriver itself.

This patch also improves wording in documentation of the initializer's arguments.


Dude-X:
LGTM. 👍

PR #5466, opened 2018-02-08
safaridriver: add new command to change getUserMedia() behavior for m…

burg:
…ock devices

As part of this feature, we introduce new endpoints to get and set session permissions. For example, it
can control whether a getUserMedia() request for a mock media stream will be allowed or denied.

Since this is the first Safari-specific endpoint, add a custom RemoteConnection subclass
so that extra commands can be added in. Hook it up in the subclass of WebDriver.

This adds two commands:

driver.get_permission(permission_name) -> True, False, None
driver.set_permission(permission_name, True | False)

For convenience, supported values of permission_name are enumerated by the Permission object.

These commands map, respectively, to the following endpoints and request/response payloads:

GET /session/$sessionId/apple/permissions

Request payload:
None

Response payload:
{
    "permissions": {
        "getUserMedia": true,
        ...
    }
}

Notes: values for all session permissions are returned, whether or not they have previously been set.

--

POST /session/$sessionId/apple/permissions

Request payload:
{
    "permissions": [
        "getUserMedia": true,
        ...
    ]
}

Response payload:
None

Notes: can update multiple session permissions simultaneously. Any omitted permission names are unaffected.

### DIFF
--- a/py/selenium/webdriver/safari/webdriver.py
+++ b/py/selenium/webdriver/safari/webdriver.py
@@ -23,7 +23,7 @@ except ImportError:
 from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
 from selenium.webdriver.remote.webdriver import WebDriver as RemoteWebDriver
 from .service import Service
-
+from .remote_connection import SafariRemoteConnection
 
 class WebDriver(RemoteWebDriver):
     """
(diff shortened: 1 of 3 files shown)


py/selenium/webdriver/safari/webdriver.py
```diff
@@ -63,3 +66,31 @@ def quit(self):
...
+    # safaridriver extension commands. The canonical command support matrix is here:
+    # https://developer.apple.com/library/content/documentation/NetworkingInternetWeb/Conceptual/WebDriverEndpointDoc/Commands/Commands.html
+
+    # First available in Safari 11.1 and Safari Technology Preview 41.
+    def set_permission(self, permission, value):
+        if not isinstance(value, bool):
+            raise WebDriverException("Value of a session permission must be set to True or False.")
+
+        payload = {}
+        payload[permission] = value
```
Dude-X: optimize:
  payload = {permission: value}

Dude-X:
LGTM.

p0deje:
@burg Should this be closed in favor of #5467?

burg:
That would be fine with me.

PR #5468, opened 2018-02-08
Removing false info of deprecation of PJS

ghost:


lmtierney:
This is not `false info`. This change indicates that we have deprecated use of PhantomJS with Selenium. This was done across all bindings of the project. If all the bindings decided to once again provide support for it, then we will remove deprecation. We are not saying that use of PhantomJS outside of Selenium is deprecated.

Removing a line from a changelog does not remove deprecation. This was deprecated in b3daa030e11b227ab94fbc499c8272b06fc6eeb6

pixiuPL:
@lmtierney while I fully understand why you've deprecated PJS, I think you should undeprecate it as its "back to live"

barancev:
Unfortunately, reviving PhantomJS itself is not enough. Someone should also revive and rework the driver for PJS to make it W3C conformant. It's the requirement to add PJS support back to Selenium.

pixiuPL:
@lmtierney someone being me as I seem to be the only active dev of PJS nowadays.....

twalpole:
@pixiuPL Your efforts to revive PJS are appreciated and I hope it succeeds.  However everyday PJS becomes less and less useful for testing apps (which Selenium is aimed at) due the lack of support for modern CSS/JS.  This is caused by the version of WebKit PJS is built against, which from the commits doesn't seem to have been updated in the 2.1.x branch (If it has been updated then I missed it and you can ignore this comment and I look forward to a statically linked release which can be used on multiple platforms without library version issues).  The 2.5 branch was supposed to be what brought PJS into the modern age again however you've stated that branch is not supported.  If updating the WebKit version is not a priority it would seem any effort to make PJS W3C compliant would just be wasted effort.

PR #5478, opened 2018-02-10
[rb] Add helper method to launch Chrome in headless mode.

pulkitsharma07:
This makes it consistent with Firefox::Options#headless! (Refer: https://github.com/SeleniumHQ/selenium/pull/4762)

Though I agree with @jimevans to not add this API for a single command line argument,  but adding this for consistency.

Another thread: https://github.com/SeleniumHQ/selenium/issues/4591

### DIFF
--- a/rb/spec/unit/selenium/webdriver/chrome/options_spec.rb
+++ b/rb/spec/unit/selenium/webdriver/chrome/options_spec.rb
@@ -99,29 +99,47 @@ module Selenium
               subject.add_argument('foo')
               subject.add_argument('bar')
 
-              expect(subject.args).to eq %w[foo bar]
+              expect(subject.args).to eq(%w[foo bar])
             end
 
-            it 'should not add same argument more than once' do
+            it 'should not add the same argument more than once' do
               subject.add_argument('foo')
               subject.add_argument('foo')
 
-              expect(subject.args).to eq ['foo']
+              expect(subject.args).to eq(['foo'])
             end
           end
         end
 
         describe '#headless!' do
-          it 'should add the --headless command-line argument' do
-            subject.headless!
-            expect(subject.args).to eql ['--headless']
+          context 'on windows' do
+            before(:each) do
+              allow(WebDriver::Platform).to receive(:windows?).and_return true
+            end
+
+            it 'should add the --headless and --disable-gpu command-line argument' do
+              subject.headless!
+              expect(subject.args).to eql(%w[--headless --disable-gpu])
+            end
           end
 
-          it 'should not add the --headless command-line argument if already present' do
-            subject.add_argument('--headless')
-            subject.headless!
-            expect(subject.args).to eql ['--headless']
+          context 'on non-windows' do
+            before(:each) do
+              allow(WebDriver::Platform).to receive(:windows?).and_return false
+            end
+
+            it 'should add the --headless command-line argument' do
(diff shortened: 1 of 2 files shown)


pulkitsharma07:
Assuming `--disable-gpu` was added because of reasons present here https://bugs.chromium.org/p/chromium/issues/detail?id=737678#c1, what should be the correct approach ? Detect the platform and add add the flag only for Windows ? And then remove that later once the issue is resolved.

I feel adding an extra flag (which the end user might not be aware of) might not be always desirable. See, https://github.com/GoogleChrome/puppeteer/issues/1260 for instance.

p0deje:
Let's add a condition for Windows 

```ruby
add_argument '--headless'
# https://bugs.chromium.org/p/chromium/issues/detail?id=737678#c1
add_argument '--disable-gpu' if WebDriver::Platform.windows?
```

rb/spec/unit/selenium/webdriver/chrome/options_spec.rb
```diff
@@ -93,6 +93,35 @@ module Chrome
             subject.add_argument('foo')
             expect(subject.args).to include('foo')
           end
+
+          context 'multiple arguments' do
+            it 'should add only unique arguments' do
+              subject.add_argument('foo')
+              subject.add_argument('bar')
+
+              expect(subject.args).to eq %w[foo bar]
```
luke-hill: If you're fixing up p0deje's request. can you wrap all your expectations in`()` instead of space separation and change line 105 to `it 'should not the add same argument more than once' do

pulkitsharma07:
@p0deje, @luke-hill Pushed the changes.

p0deje:
I've merged to master - thank you for PR!

cgoldberg:
@luke-hill @p0deje 

the Python bindings set `--disable-gpu` unconditionally.  Should Python be changed to only pass it on Windows also?

luke-hill:
I would guess it's up to the maintainers. It also depends on how the specs are written in python. 

@lmtierney You've got visibility of both bindings. What do you think. It does feel superfluous but I'm not a python expert.

cgoldberg:
just to add some more confusion... have a look at https://bugs.chromium.org/p/chromium/issues/detail?id=737678 , which references:
https://chromium-review.googlesource.com/c/chromium/src/+/522068

> "on macOS, so we will
> continue to use the physical GPU in headless mode on that platform
> (unlike on other platforms where everything is software rendered)."

so... should `--disable-gpu` be passed on every platform except MacOS?  or is this handled internally by Chrome?

p0deje:
> This flag is no longer necessary on Linux or macOS. It will become unnecessary on Windows as soon as bug 729961 is fixed.

🤷‍♂️

lmtierney:
Currently, Java and Javascript also include the `--disable-gpu` argument. 🤷‍♂️

pulkitsharma07:
Okay, so chromedriver's python client adds `--disable-gpu` only for linux (and from what it looks like it, adds that for non headless mode too)
https://github.com/bayandin/chromedriver/blob/master/client/chromedriver.py#L156 
🤷‍♂️ 🤷‍♂️

luke-hill:
From what Alex said earlier. It seems easiest / most-agile to wait until 729961 is fixed and then so a Selenium-wide change to make all the bindings correct?

PR #5485, opened 2018-02-12
Fix 'Selenium-Version' manifest attr to be in 'Build-Info' section

timsutton-sl:

I noticed in the selenium standalone releases since 3.9.0 that the `Build-Info` manifest section is missing `Selenium-Version`, which seems to be due to some refactoring in https://github.com/SeleniumHQ/selenium/commit/adc87a46fe08edca1a3bc0132643f4c306f95668#diff-f7acb1b90eea8308d4e04f861f59cd4a.

The standalone jar for 3.8.1 for example has this snippet of manifest data:
```python
'Build-Info': {'Build-Revision': '6e95a6684b',
   'Build-Time': '2017-12-01T19:05:32.194Z',
   'Build-User': 'alexei',
   'Name': 'Build-Info',
   'Selenium-Version': '3.8.1   '}
```

Compare to 3.9.1:

```python
'Build-Info': {'Build-Revision': '63f7b50',
   'Build-Time': '2018-02-07T22:42:28.403Z',
   'Build-User': 'shs',
   'Name': 'Build-Info'},
  'Selenium': {'Name': 'Selenium', 'Selenium-Version': '3.9.1'}
```

The reason for this PR is that at Sauce we've been parsing this attribute at job run time so that we can dynamically determine the Selenium version being used, even if a user submits a custom jar file to run (which would not be version-controlled by us).

This PR just writes out the manifest file as it was originally before that adc87a46fe08edca1a3bc0132643f4c306f95668 commit, so that `Build-Info` is given for the name and thus `Selenium-Version` is written within that section.

Thanks for your consideration!

### DIFF
--- a/java/client/src/org/openqa/selenium/BUCK
+++ b/java/client/src/org/openqa/selenium/BUCK
@@ -80,7 +80,7 @@ java_library(name = 'core',
 
 genrule(name = 'manifest',
   out = 'manifest.mf',
-  cmd = 'python -c "print(\'\\n\\nName: Selenium\\nSelenium-Version: {}\\n\\n\')" >> $OUT'.format(SE_VERSION),
+  cmd = 'python -c "print(\'\\n\\nName: Build-Info\\nSelenium-Version: {}\\n\\n\')" >> $OUT'.format(SE_VERSION),
 )
 
 java_library(name = 'beta',


shs96c:
LGTM. Thank you!

PR #5546, opened 2018-02-28
[Grid] Prevent page jump when clicking a node tab

dylanlive:
When a user clicks on the Browsers or Configuration tab of a node within the Grid Console (`/grid/console`), it'll act like a link click and jump to the top of the page.

This jump can become quite annoying when you have many nodes connected to the hub, requiring you to scroll back down to the node that you clicked.

This change will prevent the default behavior of the click event, so that the viewport stays where it was when you click one of those tabs.

Tested this across all major browsers.

### DIFF
--- a/java/server/src/org/openqa/grid/images/consoleservlet.js
+++ b/java/server/src/org/openqa/grid/images/consoleservlet.js
@@ -28,10 +28,11 @@
   }
 
   $(document).ready(function() {
-    $(".tabs li").click(function() {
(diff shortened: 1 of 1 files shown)


dylanlive:
@mach6 Hey there. I see you've edited this file in the past. Mind taking a look at this change?

shs96c:
LGTM!

shs96c:
Thank you for the commit! It'll be in the next release!

PR #5561, opened 2018-03-02
[py] make tests to check clicking on disabled element work for w3c co…

carlosgcampos:
…mpliant drivers

They are currently mrked as expected failure, but we can actually test
both behaviors. Check that selection doesn't change for w3c drivers.


carlosgcampos:
It seems it was failing for firefox, chrome and remote even when w3c is not set. I'll keep those xfail marks.

carlosgcampos:
The failure with marionette sounds like a bug in marionette.

PR #5568, opened 2018-03-06
Changing clear element atom to be W3C compliant

jimevans:

I'm submitting this for review, because I'm deeply insecure about making changes to the atom here. This is an attempt to make the element clear atom compliant with the W3C specification. Notable features:
* The atom doesn't focus the element to be cleared unless there is a value in the element.
* The atom now fires the "blur" event, and sets focus to the `<body>` element after clearing, in keeping with the specification language to "invoke the unfocusing steps" for the element, which includes setting the focus to the container (or more precisely the (control group owner object of the element)[https://html.spec.whatwg.org/#unfocusing-steps])

Tagging @shs96c for review.

### DIFF
--- a/javascript/atoms/action.js
+++ b/javascript/atoms/action.js
@@ -87,21 +87,25 @@ bot.action.clear = function(element) {
         'Element must be user-editable in order to clear it.');
   }
 
-  bot.action.LegacyDevice_.focusOnElement(element);
   if (element.value) {
+    bot.action.LegacyDevice_.focusOnElement(element);
     element.value = '';
     bot.events.fire(element, bot.events.EventType.CHANGE);
+    bot.events.fire(element, bot.events.EventType.BLUR);
+    bot.action.LegacyDevice_.focusOnElement(document.body);
   } else if (bot.dom.isElement(element, goog.dom.TagName.INPUT) &&
              (element.getAttribute('type') && element.getAttribute('type').toLowerCase() == "number")) {
     // number input fields that have invalid inputs
     // report their value as empty string with no way to tell if there is a
     // current value or not
+    bot.action.LegacyDevice_.focusOnElement(element);
     element.value = '';
   }
 
   if (bot.dom.isContentEditable(element)) {
     // A single space is required, if you put empty string here you'll not be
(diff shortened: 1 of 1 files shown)


javascript/atoms/action.js
```diff
@@ -87,21 +87,25 @@ bot.action.clear = function(element) {
         'Element must be user-editable in order to clear it.');
   }
 
-  bot.action.LegacyDevice_.focusOnElement(element);
```
shs96c: Both branches appear to call `focusOnElement` as the first thing they do. Why not just leave this up here?

javascript/atoms/action.js
```diff
@@ -87,21 +87,25 @@ bot.action.clear = function(element) {
...
     bot.events.fire(element, bot.events.EventType.CHANGE);
+    bot.events.fire(element, bot.events.EventType.BLUR);
+    bot.action.LegacyDevice_.focusOnElement(document.body);
   } else if (bot.dom.isElement(element, goog.dom.TagName.INPUT) &&
              (element.getAttribute('type') && element.getAttribute('type').toLowerCase() == "number")) {
     // number input fields that have invalid inputs
     // report their value as empty string with no way to tell if there is a
     // current value or not
+    bot.action.LegacyDevice_.focusOnElement(element);
     element.value = '';
```
shs96c: I'm not sure why we're not running the unfocussing steps here, or with the content editable, though both of these should (according to the spec) also be unfocussed.

shs96c:
I've filed https://github.com/w3c/editing/issues/174 because it's not clear what the heck we're meant to do with content editable fields.

javascript/atoms/action.js
jimevans: Because the spec specifies that the element not be focused unless it has a value to be cleared. Leaving this outside the branches would cause the element to be focused indiscriminately.

javascript/atoms/action.js
jimevans: I'm not sure either. It's likely an oversight on my part. Though that branch doesn't fire the change event either (and never has).

jimevans:
Closing the PR. Will push the update into master later today.

Issue #5578, opened 2018-03-07
safari Webdriver() throws AttributeError with default arguments for python client driver


mingeth:
## Meta -
OS:  linux
Selenium Version:  3.10.0
Browser:  Safari/Version 11.0.3 (13604.5.6)

## Expected Behavior -
launching webdriver.Safari() from python with default arguments should create a safaridriver instance

## Actual Behavior -
AttributeError: 'WebDriver' object has no attribute 'service'

## Steps to reproduce -
>>> from selenium import webdriver
>>> s = webdriver.Safari()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/selenium/webdriver/safari/webdriver.py", line 49, in __init__
    self.service.start()
AttributeError: 'WebDriver' object has no attribute 'service'

mingeth:
pull request created at branch python-safari-webdriver-defaultargs

barancev:
PR merged as 2a0f63d19bd6e666ad432ee459762489d6b6033a

lmtierney:
It is fixed in the trunk and will be in the next release

suchau3:
OS: MacOS 10.13.4
Selenium Version: 3.11.0
Browser: Safari/Version 11.1 (13605.1.33.1.2)

Hi, is this solved and release updated yet? 

I get the same error as @mingeth and @corytu 

File ".../selenium/webdriver/safari/webdriver.py", line 49, in __init__
    self.service.start()
AttributeError: 'WebDriver' object has no attribute 'service'

corytu:
@suchau3 Please see @lmtierney's comment right above yours.

> It is fixed in the trunk and will be in the next release

mithunsureshqa:
how do we fix this locally?  i am new to this.

da-false9:
You need to go to the directory where is saved your webdriver.py for Safari.
In my case it's /Users/YOUR_USER_NAME/vitenv/lib/python3.6/site-packages/selenium/webdriver/safari 
After this you need to open webdriver.py and to add those 2 strings of the code that were added in the decision for this issue.

swissbeats93:
I get more errors after that, specifically which 2 lines should be appended and on which part of web driver.py

PR #5579, opened 2018-03-07
Fix Safari webdriver AttributeError for python client driver #5578

mingeth:

### DIFF
--- a/py/selenium/webdriver/safari/webdriver.py
+++ b/py/selenium/webdriver/safari/webdriver.py
@@ -48,7 +48,6 @@ class WebDriver(RemoteWebDriver):
         self._reuse_service = reuse_service
         self.service = Service(executable_path, port=port, quiet=quiet)
         if not reuse_service:
-            self.service = Service(executable_path, port=port, quiet=quiet)
             self.service.start()
 
         RemoteWebDriver.__init__(


AutomatedTester:
@burg can you review

py/selenium/webdriver/safari/webdriver.py
```diff
@@ -46,6 +47,7 @@ def __init__(self, port=0, executable_path="/usr/bin/safaridriver", reuse_servic
 
         self._reuse_service = reuse_service
         if not reuse_service:
+            self.service = Service(executable_path, port=port, quiet=quiet)
```
burg: This needs to be assigned outside the branch because we always want to have a service instance.
self.service is used unconditionally when calling the super initializer, and other code expects it to be there even if reuse_service is True.

mingeth:
@burg now assigned outside of the reuse condition. sorry for extra commit, had to delete redundant `service` call.

burg:
Why are there 6 commits? This is a two line change. Please just figure out how to make a clean PR.

barancev:
Merged as 2a0f63d19bd6e666ad432ee459762489d6b6033a, thank you for contribution!

PR #5615, opened 2018-03-14
Add json content-type for empty requests (fix net/http warning in strict mode)

kolen:
When doing empty requests (for example for element click), `Content-Type` header is not set. This causes `Net::HTTP` to issue warning in `$VERBOSE` mode (`-w`):

```
net/http: warning: Content-Type did not set; using application/x-www-form-urlencoded
```

This warning shows for each action (click, etc), so there are lots of these warnings in output when running tests.

This pull request adds `Content-Type: application/json` to such requests too (they has `{}` body so I guess it's empty json "object")

Example script that shows this warning:

```ruby
#!/usr/bin/env ruby -w

require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "selenium-webdriver"
end

require "selenium-webdriver"

options.add_argument("--headless")
driver = Selenium::WebDriver.for :chrome, options: options
driver.navigate.to "http://google.com"

element = nil
Selenium::WebDriver::Wait.new(timeout: 10)
                         .until { element = driver.find_element(tag_name: "a") }
STDERR.puts "clicking"
element.click
STDERR.puts "clicked"
```

I didn't add test for warning, only added check for header being added to request. `warn` is not easily stubbable, common way to catch warnings is to redirect stderr to `StringIO` which is too hacky I think.

### DIFF
--- a/rb/lib/selenium/webdriver/remote/http/common.rb
+++ b/rb/lib/selenium/webdriver/remote/http/common.rb
@@ -53,6 +53,7 @@ module Selenium
               WebDriver.logger.debug("     > #{headers.inspect}")
             elsif verb == :post
               payload = '{}'
+              headers['Content-Type']   = CONTENT_TYPE
               headers['Content-Length'] = '2'
             end
 
--- a/rb/spec/unit/selenium/webdriver/remote/http/common_spec.rb
+++ b/rb/spec/unit/selenium/webdriver/remote/http/common_spec.rb
@@ -28,7 +28,9 @@ module Selenium
 
             expect(common).to receive(:request)
               .with(:post, URI.parse('http://server/clear'),
-                    hash_including('Content-Length' => '2'), '{}')
+                    hash_including('Content-Length' => '2',
+                                   'Content-Type' => 'application/json'),
+                    '{}')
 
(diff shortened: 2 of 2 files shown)


JasonLunn:
#5659 is alternative solution to the same root problem. In that PR, the issue is resolved by doing explicitly what net-http has been doing implicitly all along, which is to supply the default content type of `application/x-www-form-urlencoded` if nothing was provided. This approach in this PR does seem more correct, but it isn't clear to me if there is a hard guarantee that all `POSTS` will be JSON, or if there are any bad behaviors on the part of the servers that might expect this default content type.

kolen:
@JasonLunn Yes, I think `DEFAULT_HEADERS` is a good place to put content-type, if it does not break anything, and if `Content-Type` with empty body (including `GET`) is ok by standards (I can't find anything about it).

p0deje:
Thank you for the issue report and PR addressing the problem. I've decided to push a fix by myself so that it complies to [JSON Wire Protocol specification](https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#commands) of payload. This also makes Ruby bindings behave similarly to Python and Java ones.

PR #5636, opened 2018-03-18
Suppress unintended deprecation warning

treby:
A deprecation warning seems not to work correctly. Even if we don't give any settings for `extensions` to `Selenium::WebDriver::Chrome::Profile` , the class warns deprecation. Is this intentional?

The warning comes from `Selenium::WebDriver::Chrome::Driver#create_capabilities`, called from constructor of the class.

https://github.com/SeleniumHQ/selenium/blob/1103cb7e9e35e8bd22a5aea9b5455445baa6fb85/rb/lib/selenium/webdriver/chrome/driver.rb#L85-L86

The profile was processed by `Selenium::WebDriver::Chrome::Profile#as_json`

https://github.com/SeleniumHQ/selenium/blob/1103cb7e9e35e8bd22a5aea9b5455445baa6fb85/rb/lib/selenium/webdriver/chrome/driver.rb#L78

And the method always returns Hash with `:extensions` key.

https://github.com/SeleniumHQ/selenium/blob/1103cb7e9e35e8bd22a5aea9b5455445baa6fb85/rb/lib/selenium/webdriver/chrome/profile.rb#L73-L83

I think `if extensions.any?` is better than `if extension`.

# Reproduction code

```ruby
require 'selenium-webdriver'
include Selenium

puts 'with an extensions'
profile = Selenium::WebDriver::Chrome::Profile.new
profile.add_extension(File.expand_path('../spec/integration/selenium/webdriver/chrome/sample.crx', __FILE__))
Selenium::WebDriver.for(:chrome, profile: profile)

puts 'with no extensions'
profile = Selenium::WebDriver::Chrome::Profile.new
Selenium::WebDriver.for(:chrome, profile: profile)
```

## Before

```
with an extensions
2018-03-19 00:37:08 WARN Selenium [DEPRECATION] Using Selenium::WebDriver::Chrome::Profile#extensions is deprecated. Use Selenium::WebDriver::Chrome::Options#add_extension instead.
with no extensions
2018-03-19 00:37:09 WARN Selenium [DEPRECATION] Using Selenium::WebDriver::Chrome::Profile#extensions is deprecated. Use Selenium::WebDriver::Chrome::Options#add_extension instead.
```

## After

```
with an extensions
2018-03-19 00:36:58 WARN Selenium [DEPRECATION] Using Selenium::WebDriver::Chrome::Profile#extensions is deprecated. Use Selenium::WebDriver::Chrome::Options#add_extension instead.
with no extensions
```

### DIFF
--- a/rb/lib/selenium/webdriver/chrome/profile.rb
+++ b/rb/lib/selenium/webdriver/chrome/profile.rb
@@ -78,7 +78,7 @@ module Selenium
           extensions.concat(@encoded_extensions)
 
           opts = {directory: @directory || layout_on_disk}
-          opts[:extensions] = extensions if extensions
+          opts[:extensions] = extensions if extensions.any?
           opts
         end
 


Jaysheel:
In java, I had started getting similar warning for desired capabilities. And the warn was asking to use 'options' instead. 
So, i think it is intended. You might want to use Chrome Options instead of Profile.

p0deje:
Thank you, merged in dd8b51f602ffdebebe10b9d83c24506ae0b20dee.

treby:
Thanks!

PR #5659, opened 2018-03-22
Eliminate spurious warnings from net-http

JasonLunn:
Eliminates the infamous `'net/http: warning: Content-Type did not set; using application/x-www-form-urlencoded'` warning message emitted from `supply_default_content_type` within net-http if running with `$VERBOSE` true.

### DIFF
--- a/rb/lib/selenium/webdriver/remote/http/common.rb
+++ b/rb/lib/selenium/webdriver/remote/http/common.rb
@@ -22,7 +22,7 @@ module Selenium
         class Common
           MAX_REDIRECTS   = 20 # same as chromium/gecko
           CONTENT_TYPE    = 'application/json'.freeze
-          DEFAULT_HEADERS = {'Accept' => CONTENT_TYPE}.freeze
+          DEFAULT_HEADERS = {'Accept' => CONTENT_TYPE, 'Content-Type' => 'application/x-www-form-urlencoded'}.freeze
 
           attr_accessor :timeout
           attr_writer :server_url


rb/lib/selenium/webdriver/remote/http/common.rb
```diff
@@ -22,7 +22,7 @@ module Http
         class Common
           MAX_REDIRECTS   = 20 # same as chromium/gecko
           CONTENT_TYPE    = 'application/json'.freeze
-          DEFAULT_HEADERS = {'Accept' => CONTENT_TYPE}.freeze
+          DEFAULT_HEADERS = {'Accept' => CONTENT_TYPE, 'Content-Type' => 'application/x-www-form-urlencoded'}.freeze
```
kolen: Does it send requests with urlencoded bodies anywhere? I thought it sends only json. Maybe content-type should be `application/json` by default?

kolen:
And is it ok to include `Content-Type` header to requests wihout body (or with empty body), including `GET`? I can't find anything about it in standards ([here they say that it shoud be included for non-empty bodies but nothing about the reverse](https://tools.ietf.org/html/rfc7231#section-3.1.1.5)).
Also I'm not sure if Selenium ever sends requests without bodies.
Probably it's not a concern.

rb/lib/selenium/webdriver/remote/http/common.rb
JasonLunn: @kolen If you have `$VERBOSE` set to true, you'll see that Net-HTTP has been setting the Content-Type header to `application/x-www-form-urlencoded` automatically (and emitting warnings about it if `$VERBOSE` is true). As I mentioned in #5615 - I think `application/json` is more correct, but I am not in a position to tell if you if "more correct" >= "the way it has been historically".

p0deje:
Thank you for the issue report and PR addressing the problem. I've decided to push a fix by myself so that it complies to [JSON Wire Protocol specification](https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#commands) of payload. This also makes Ruby bindings behave similarly to Python and Java ones.

PR #5667, opened 2018-03-22
Add ability to filter log entries by range of log levels

valfirst:

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/5667)

### DIFF
--- a/java/client/src/org/openqa/selenium/logging/LogEntries.java
+++ b/java/client/src/org/openqa/selenium/logging/LogEntries.java
@@ -23,7 +23,9 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.function.Predicate;
 import java.util.logging.Level;
+import java.util.stream.Collectors;
 
 /**
  * Represent a pool of {@link LogEntry}.  This class also provides filtering mechanisms based on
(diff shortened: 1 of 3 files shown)


shs96c:
Would it not be nicer to pass in a `Predicate<LogEntry>`? That way you can just use a normal Java streams to deal with these.

shs96c:
Looking closer, `LogEntries.getAll().stream().filter(entry -> true).collect(Collectors.toList())` would allow this, right? And then it becomes easier to add your own filters too.

valfirst:
> Would it not be nicer to pass in a Predicate<LogEntry>? That way you can just use a normal Java streams to deal with these.

I can make `private List<LogEntry> filter(Predicate<LogEntry> logEntryPredicate)` public. Or do you propose to do not add methods filtering by levels at all?

\
The root cause of this PR: I've started using `LogEntries`, applied `filter(Level.WARNING)` and expected to get only warnings (who reads javadocs 😄), but I received warnings, errors and etc. So I thought, that it might be useful to have flexible API to get log entries of specified level or level range only, at least users may try to understand difference between overloaded methods.

valfirst:
Closing: filter functionality was deprecated in `LogEntries`: https://github.com/SeleniumHQ/selenium/commit/c24ac7c1064a982c14b89742eda57219e420f770

PR #5691, opened 2018-03-27
Added more documentation to isEnabled() and resolves #3547

### LINKED ISSUE #3547: Should have more description what isenabled API can do.
As the code below in atoms, there're only few types of elements that support Disabled_Attribue
https://github.com/SeleniumHQ/selenium/blob/1721e627e3b5ab90a06e82df1b088a33a8d11c20/javascript/atoms/dom.js
/**
 * List of elements that support the "disabled" attribute, as defined by the
 * HTML 4.01 specification.
 * @private {!Array.<goog.dom.TagName>}
 * @const
 * @see http://www.w3.org/TR/html401/interact/forms.html#h-17.12.1
 */
bot.dom.DISABLED_ATTRIBUTE_SUPPORTED_ = [
  goog.dom.TagName.BUTTON,
  goog.dom.TagName.INPUT,
  goog.dom.TagName.OPTGROUP,
  goog.dom.TagName.OPTION,
  goog.dom.TagName.SELECT,
  goog.dom.TagName.TEXTAREA
];

Can we describe the API doc with more details, so the selenium users wouldn't get confused when calling this function?
https://github.com/SeleniumHQ/selenium/edit/master/java/client/src/org/openqa/selenium/WebElement.java

irrationalRock:

This PR request fixes https://github.com/SeleniumHQ/selenium/issues/3547.
It adds what elements isEnabled() can handle.

Thanks for considering this request.

### DIFF
--- a/java/client/src/org/openqa/selenium/WebElement.java
+++ b/java/client/src/org/openqa/selenium/WebElement.java
@@ -127,7 +127,8 @@ public interface WebElement extends SearchContext, TakesScreenshot {
   /**
    * Determine whether or not this element is selected or not. This operation only applies to input
    * elements such as checkboxes, options in a select and radio buttons.
-   * This method supports the following elements: BUTTON, INPUT, OPTGROUP, OPTION, SELECT, TEXTAREA.
+   * For more information on which elements this method supports,
+   * refer to the <a href="https://w3c.github.io/webdriver/webdriver-spec.html#is-element-selected">specification</a>.
    *
    * @return True if the element is currently selected or checked, false otherwise.
    */


java/client/src/org/openqa/selenium/WebElement.java
```diff
@@ -127,6 +127,7 @@
   /**
    * Determine whether or not this element is selected or not. This operation only applies to input
    * elements such as checkboxes, options in a select and radio buttons.
+   * This method supports the following elements: BUTTON, INPUT, OPTGROUP, OPTION, SELECT, TEXTAREA.
```
shs96c: The problem with HTML is they keep changing things. The actual definition of what can be checked to see if something is selected is defined in the [spec](https://w3c.github.io/webdriver/webdriver-spec.html#is-element-selected). It's probably best to either link to that, or to lift the meaningful section.

Further, the intention of the API is already described above the change you've made --- I'm not sure what the value is of keeping that statement with a more complete description being available.

irrationalRock:
Thanks for the feedback. Instead of hardcoding the elements, I added the link to specifications itself.

shs96c:
Thank you!

Issue #5657, opened 2018-03-21
Cloud providers and Grid could benefit from client bindings sending in version and platform info.


sah:
As a cloud provider, we would absolutely love a key in the new session request that says something like “hi I’m version X.Y of the Java bindings running on macOS 10.13” — along the lines of an HTTP User-Agent. This would help with diagnosing bugs and incompatibilities, especially when dealing with a wide variety of versions of both the client bindings and Selenium server and drivers. For similar reasons, I suspect it might also be useful for Grid to log this info.

I'd be happy to contribute code for this across a bunch of the bindings, if people think it's a good idea and we can agree on where to put it and what the format should be.

shs96c:
Is it okay to put that information in an HTTP header?

sah:
Sure, that would work. Could even tailor the actual `User-Agent`...

jleyba:
We should probably decide on a standard `User-Agent` format for the project so we're consistent across the different bindings.

pulkitsharma07:
Are we planning to overwrite the `User-Agent` header which is generated by the underlying HTTP library the bindings use?

I recommend using a new non-standard header (something like `Selenium-Client-Ver`) for sending the binding version, so that it can be succinct (and unambiguous), plus we don't have to worry about adding content to any existing header(s).

jleyba:
How about: `X-Selenium-Version: ${LANGUAGE}[-${COMPONENT}]@${VERSION}`

Possible values:

- `java-client@3.8.1`
- `java-grid@3.8.1`
- `nodejs@4.0.0-alpha.1`

Use multiple values to capture nodes in a grid. For example:

`X-Selenium-Version: java-grid@3.8.1,java-grid@3.7,java-client@3.8.1`

indicates a request came from a grid server (v3.8.1), which is forwarding a request from another grid at v3.7, which received the initial request from a client using 3.8.1.

jimevans:
@jleyba I like where you're going with that, but it doesn't add in the OS information as was in the original request. Moreover, I'm not entirely sure how we can get a satisfactory OS information string together that would please everyone.

And let's be very clear here, I'm in favor of this if it is _**only**_ to be applied on new session requests, and is _**optional**_, meaning that third parties and intermediate nodes can't depend on it, but could make use of it if and only if it's present.

shs96c:
[...]

PR #5696, opened 2018-03-28
User agent for Python (and formerly Ruby, already merged)

sah:
Addresses #5657 for Python and Ruby

### DIFF
--- a/py/selenium/webdriver/remote/remote_connection.py
+++ b/py/selenium/webdriver/remote/remote_connection.py
@@ -181,7 +181,7 @@ class RemoteConnection(object):
         headers = {
             'Accept': 'application/json',
             'Content-Type': 'application/json;charset=UTF-8',
-            'User-Agent': 'selenium/%s (python %s)' % (__version__, system)
+            'User-Agent': 'selenium/{} (python {})'.format(__version__, system)
         }
 
         if parsed_url.username:


p0deje:
Can you explain why you need these changes?

sah:
Explanation is in the issue referenced in the description, but basically this is useful for error diagnosis in cloud services and grids. Similar changes have already been introduced for Java and .NET, so this is just rolling the feature out to more languages.

p0deje:
Thank you, I've merged Ruby part in 429af62c79c4048d54856a9d4b79d451d98e372a!

sah:
Updated this PR to reflect just the Python change now that the Ruby change has been merged.

py/selenium/webdriver/remote/remote_connection.py
```diff
@@ -172,10 +174,14 @@ def get_remote_connection_headers(cls, parsed_url, keep_alive=False):
...
 
+        system = platform.system().lower()
+        if system == "darwin":
+            system = "mac"
+
         headers = {
             'Accept': 'application/json',
             'Content-Type': 'application/json;charset=UTF-8',
-            'User-Agent': 'Python http auth'
+            'User-Agent': 'selenium/%s (python %s)' % (__version__, system)
```
lmtierney: Prefer `str.format` for String Interpolation in the future

py/test/unit/selenium/webdriver/remote/test_remote_connection.py
```diff
@@ -33,7 +34,8 @@ def test_get_remote_connection_headers_defaults():
     assert 'Connection' not in headers.keys()
     assert headers.get('Accept') == 'application/json'
     assert headers.get('Content-Type') == 'application/json;charset=UTF-8'
-    assert headers.get('User-Agent') == 'Python http auth'
+    assert headers.get('User-Agent').startswith("selenium/%s (python " % __version__)
+    assert headers.get('User-Agent').split(' ')[-1] in {'windows)', 'mac)', 'linux)'}
```
lmtierney: Is there a reason you aren't just asserting the entire `User-Agent`?  The second assertion could allow false positives if the wrong platform sent

sah: My reasoning was that asserting the whole thing requires duplicating the code to calculate the platform name, which proves nothing except that that code does whatever it does. This way if on OS X it returns "darwin" or Windows it returns "win32", that issue will be caught when these tests are run on those platforms. I don't feel strongly about it though, happy to go a different route here if you prefer.

PR #5710, opened 2018-03-30
Retain config file paths if specified

sbabcoc:

I created a Java library that extends Selenium (https://github.com/Nordstrom/Selenium-Foundation). This library includes code to manage a local Grid instance, and a couple of my unit tests check the paths of the files used to configure the server instances. As a side effect of the refactoring of `selenium-server-standalone` command line processing, this information was no longer being retained. This PR restores this lost behavior by copying the file path from CLI object to configuration object.


sbabcoc:
I looked over the failures in the Travis-CI build, and it's highly unlikely that any of these are related to my changes.

barancev:
Merged as f3d6bc6f27f9d35d81ead4973499cb200bae83c8

Honestly, I was going to make this field private or may be even delete it, but backward compatibility is important too. So let it stay here for some more time.

Thank you for contribution!

PR #5713, opened 2018-03-31
allow edge driver to create a default service in createSession method…

pz325:
… if `service` is `null`

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/5713)

### DIFF
--- a/javascript/node/selenium-webdriver/edge.js
+++ b/javascript/node/selenium-webdriver/edge.js
@@ -198,6 +198,10 @@ class Driver extends webdriver.WebDriver {
    * @return {!Driver} A new driver instance.
    */
   static createSession(options, service = getDefaultService()) {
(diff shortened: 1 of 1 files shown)


javascript/node/selenium-webdriver/edge.js
```diff
@@ -198,6 +198,10 @@ class Driver extends webdriver.WebDriver {
    * @return {!Driver} A new driver instance.
    */
   static createSession(options, service = getDefaultService()) {
+    if (!service) {
```
jleyba: Would you mind removing the default parameter above? (I doubt I'll ever use defaults again since you still have to protect against callers explicitly passing `null` or `undefined`)

diemol:
Closing this PR as we never got a reply from the original author.

PR #5714, opened 2018-04-01
Fix issues in comments and exception messages.

User253489:
Fix issues in comments and exception messages.

### DIFF
--- a/java/client/src/org/openqa/selenium/Capabilities.java
+++ b/java/client/src/org/openqa/selenium/Capabilities.java
@@ -48,7 +48,7 @@ public interface Capabilities {
       return (Platform) rawPlatform;
     }
 
-    throw new IllegalStateException("Platform was neither a string or a Platform: " + rawPlatform);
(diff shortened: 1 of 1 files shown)


java/client/src/org/openqa/selenium/Capabilities.java
```diff
@@ -86,7 +86,7 @@ default boolean isJavascriptEnabled() {
   default boolean is(String capabilityName) {
     Object cap = getCapability(capabilityName);
     if (cap == null) {
-      // javascriptEnabled is true if not set explicitly
```
p0deje: What is wrong with this message?

shs96c:
I've applied the parts of this that make sense to me to master. I've had to do some of this by hand, since the merge isn't automatic.

java/client/src/org/openqa/selenium/Capabilities.java
User253489: The fact that it neither starts with a capital letter nor ends with a period.

Thanks for reviewing.

PR #5731, opened 2018-04-05
Allow extension of RemoteWebElement for caching element attributes

mgkeeley:

This update allows the extension of the RemoteWebElement and RemoteWebDriver classes for caching element attributes.
By default, there is no caching of attributes enabled.

This is in preparation for the new element attribute caching feature in Appium.
See https://github.com/appium/WebDriverAgent/pull/69 and https://github.com/appium/appium-uiautomator2-server/pull/128 for the corresponding changes in the appium servers that allow element attributes to be returned in the find-elements type calls.
The attributes are only returned if specific capabilities are enabled.
PR https://github.com/appium/appium-dotnet-driver/pull/216 implements the caching on the dotnet side in Appium.

### DIFF
--- a/dotnet/src/webdriver/Remote/RemoteWebDriver.cs
+++ b/dotnet/src/webdriver/Remote/RemoteWebDriver.cs
@@ -1027,7 +1027,7 @@ namespace OpenQA.Selenium.Remote
                     id = (string)elementDictionary["ELEMENT"];
                 }
 
-                element = this.CreateElement(id, elementDictionary);
+                element = this.CreateElement(id);
             }
 
             return element;
@@ -1061,7 +1061,7 @@ namespace OpenQA.Selenium.Remote
                             id = (string)elementDictionary["ELEMENT"];
                         }
 
-                        RemoteWebElement element = this.CreateElement(id, elementDictionary);
+                        RemoteWebElement element = this.CreateElement(id);
                         toReturn.Add(element);
                     }
                 }
@@ -1251,16 +1251,6 @@ namespace OpenQA.Selenium.Remote
         {
             RemoteWebElement toReturn = new RemoteWebElement(this, elementId);
             return toReturn;
-        }
-
-        /// <summary>
-        /// Creates a <see cref="RemoteWebElement"/> with the specified ID and element attributes.
-        /// </summary>
-        /// <param name="elementId">The ID of this element.</param>
-        /// <param name="elementDictionary">The attributes for this element.</param>
-        /// <returns>A <see cref="RemoteWebElement"/> with the specified ID.</returns>
-        protected virtual RemoteWebElement CreateElement(string id, Dictionary<string, object> elementDictionary) {
-            return CreateElement(id);
         }
 
         /// <summary>


jimevans:
@mgkeeley I have some questions about this PR before blindly merging it. I can get behind the need to make properties and methods `virtual`, I guess, but changing the constructor to `RemoteWebElement` or the already properly overridable `CreateElement` method is a big architectural change that I’m not entirely sure I want to expose.

As written here, it yields the ability to attach an undocumented, non-type-safe property bag to every element, which strikes me as an anti pattern. Before we go down that route, I’m going to need some additional information.

What is the common Appium use case for .NET users? Is it connecting to the Appium server directly on the local machine? If so, I’d like to understand how using composition instead of inheritance wouldn’t solve the problem in a way that wouldn’t require changes in the .NET API contract.

If the common use case is truly a remote one, where `RemoteWebDriver` is used directly, how are you proposing getting around the existing interpretations of the wire protocol, and the fact that pushing such additions upstream means that version mismatches, and even language mismatches can fail because compatibility is no longer guaranteed?

mgkeeley:
@jimevans 
Great questions, let me expand below:

The goal with this change is to enable RemoteWebDriver subclasses access to the extra attributes returned in the getElement(s) calls (currently only "ELEMENT" is returned for each object, the other changes listed above add the ability to return more properties).
I am not changing any constructors, just adding a new overridable CreateElement(id, elementDictionary) method to RemoteWebDriver with the extra attributes returned in the getElement(s) calls. I did it this way to keep the existing CreateElement(id) method and maintain backwards compatibility with existing implementations. It is up the the subclasses of RemoteWebDriver to do what they want with the properties, or ignore them.

The common Appium use case I am targeting here is supporting a larger testing framework, e.g. Tricentis Tosca (I work for Tricentis). Typically these applications want to scan the whole target, retrieve many properties for each element, and present the results to the user for selection. 
Tricentis Tosca runs only on Windows, and so typically they would connect to an Appium server running locally for Android, or a remote Mac for iOS, or remote servers in SauceLabs etc.

Composition is difficult with the way the RemoteWebDriver and RemoteWebElement work currently. The return value for CreateElement is RemoteWebElement (not IWebElement) and RemoteWebDriver expects the created element to be a RemoteWebElement in several places. There are also many internal or protected methods that make wrapping the RemoteWebElement class difficult (I had initially tried that approach).
In the Appium dotnet classes, there is already a hierarchy of classes based on RemoteWebElement (-> AppiumWebElement -> IOSElement and -> AppiumWebElement -> AndroidElement).  These classes are used directly in client code, and the way to access the properties is via the properties on the RemoteWebElement class itself.
I haven't changed the constructor to RemoteWebElement, just added the new CreateElement method (on RemoteWebDriver) with the extra parameter, so the existing API Contract remains as-is. By default, the elementDictionary is ignored and the original CreateElement method called instead.

Regarding the wire protocol, that truly is an issue with the naming of attributes. e.g. it is "TagName" in Selenium dotnet, "name" in the W3C spec, and "type" in the iOS XCUITest/WebDriverAgent for example. My idea is that the requested element fields (in the elementReponseAttributes capability) should always use the W3C names; then it is up to each end to interpret these values appropriately. Once everything standardises on W3C (as it seems to be heading that way), maybe there can be some more automatic translations of these properties. A task for another day.

jimevans:
The reason I ask about using composition is that there are examples of doing exactly that. One of them is [right here in the Selenium project](https://github.com/SeleniumHQ/selenium/blob/master/dotnet/src/support/Events/EventFiringWebDriver.cs). I’m still not convinced this is the right approach. It’s late, I’m tired, and I need some more time to think on it.

mgkeeley:
Sure.
I had a look at those classes too - but they still don't give access to the element dictionary returned in the getElements calls - that is hidden by RemoteWebDriver. The wrapped driver still doesn't have access to those values.
Also, much of the code needed to be reimplemented in those classes, instead of extending the existing RemoteWebDriver/RemoteWebElement. I want to avoid that (as it would also mean changing a lot of Appium dot net classes).

mgkeeley:
@jimevans I thought some more about it - the way the RemoteWebDriver class is set up is for the inheritance pattern (e.g. overriding CreateElement). The current code in Appium already uses this pattern, so for me, the inheritance route is the right way to go for the Appium subclasses. 
Adding the new CreateElement call with the extra properties is obligatory, as there is no other way for the Appium classes to access the extra attributes returned in the Response (whether we use composition or not). It's also cleaner in that the existing CreateElement(id) overrides (that may return different subclasses of RemoteWebElement) are not affected at all.

jimevans:
I’ve also given it some more thought, and I’m increasingly uncomfortable altering the bindings in a way that attempts to accommodate what amounts to a wire protocol change (that the response of a `FindElement(s)` call would contain more than just the element reference) implemented by a downstream project. I suppose there’s not a good way around it, but I’m thinking of a slightly different way to refactor this so that it’s a bit cleaner for the public API. It is likely to cause some short term pain for consumers/extenders, and I need to work it through for myself for a bit.

mgkeeley:
OK thanks. @jlipps also had some ideas about different approaches, he has spent some time on this too. Perhaps you can compare ideas

jlipps:
The idea we came up with today was basically to see if it makes sense to copy the idea of the JsonToWebElementConverter from the java client. This class encapsulates the strategy for turning a JSON response into web elements. It's an ideal place to do the necessary steps for the new Appium feature, including potentially returning a new subclass of WebElement instead of WebElement itself. There's already an Appium-specific element class called MobileElement (in the Java client); we could attribute caching and setting there, and simply return objects of this type from the converter. Then all that's necessary is a way to use a custom converter rather than the one that ships with the Selenium client. So for the .Net stuff to work, it would involve:

- refactoring to use something like the Java JsonToWebElementConverter in the .net client
- allowing driver init to take one of these objects (so that anything which inherits from JsonToWebElementConverter would work, including our own custom one), or a factory, or whatever

@jimevans said he was going to play around with this idea and see if it will fly. Maybe an even better approach will reveal itself along the way!

mgkeeley:
OK, thanks Jim. Although I would tend to remove the CreateElement method completely if it is no longer being called. This would cause a compile time failure as you mention, which imo is much safer than silently changing the behaviour of existing systems!
So i guess it's the same idea as this PR; but the new CreateElement method is now wrapped in a factory, that also needs to be subclassed in the downstream projects.

mgkeeley:
@jimevans I updated this PR to remove the driver changes as they are now not necessary, but we still need the virtual methods on the RemoteWebElement class. Could you approve/merge these remaining changes?

mgkeeley:
@jimevans bump!

jimevans:
Merged in c6686b3. Thank you for the contribution.

PR #5755, opened 2018-04-10
Fix obscure window switching bug for dotnet

Xaeroxe:

The following sequence of events produces a bug in current Selenium dotnet bindings

- Have two windows open in Selenium
- Close the one you're currently on
- Try to switch to the other by name

When you do this the code for switching to another window by name attempts to get the current window handle.  It needs this handle in order to restore the correct state at the end of the search if the search failed. The driver will correctly inform the dotnet bindings that there is not a current window handle.  This causes the dotnet library to throw the exception "NoSuchWindow" which is then propagated up to user code.

This PR aims to fix this bug by allowing "CurrentWindowHandle" to return null in this scenario,  Then the window switching code is updated to anticipate a possible null value.

### DIFF
--- a/dotnet/src/webdriver/Remote/RemoteTargetLocator.cs
+++ b/dotnet/src/webdriver/Remote/RemoteTargetLocator.cs
@@ -146,7 +146,15 @@ namespace OpenQA.Selenium.Remote
                 catch (NoSuchWindowException)
                 {
                     // simulate search by name
-                    string original = this.driver.CurrentWindowHandle;
+                    string original = null;
+                    try
+                    {
+                        original = this.driver.CurrentWindowHandle;
+                    }
+                    catch (NoSuchWindowException)
+                    {
+                    }
+
                     foreach (string handle in this.driver.WindowHandles)
                     {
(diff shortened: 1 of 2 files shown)


jimevans:
Wouldn't a slightly better approach be to just wrap the call to the `CurrentWindowHandle` property in another try-catch, and catch the `NoSuchWindowException` there? There would be no need then to potentially return a `null` value from that property. Also, I'm mildly concerned about the state once the search complete and there is no valid current window, but I suppose that's no more broken than it already is. I've sketched out the changes [in a gist](https://gist.github.com/jimevans/5e0dd2ff22627c69ad2cade7d0f5d174), which you're welcome to take a look at.

Xaeroxe:
Sure, sounds good to me.  I'll update my PR now.

Xaeroxe:
@jimevans ready

Xaeroxe:
Confirmed this is working for our local project.  We'll use a build from this branch for the time being.

jimevans:
Merged as 74a5135. Thank you for the contribution.

PR #5762, opened 2018-04-11
add shadow dom v1 support to DOM atoms

### LINKED ISSUE #5761: `getVisibleText` vs `getElementText`, regression?
It seems we may have at some point broken the ability to get the text of an element.

I did a PR last year (#4230) to introduce shadow tree traversal when retrieving element text.

The general idea of this PR was to modify [getElementText](https://github.com/SeleniumHQ/selenium/blob/47706e36d0bbad98a4584b10f582837d1ad5711d/javascript/selenium-atoms/text.js#L174) such that it will follow distributed children through `slot` and `content` elements, along with taking `shadowRoot` into account on each element.

However, having (painfully) stepped through the latest selenium source, it seems we now call [getVisibleText](https://github.com/SeleniumHQ/selenium/blob/74e584d12ba31bb2e1b61b17a4751402c70c6da6/javascript/atoms/dom.js#L998) instead.

This function _appears_ to have knowledge of shadow dom, but presumably has a bug somewhere which im hitting (and trying to reproduce for you).

So:

* can someone explain `selenium-atoms` vs `atoms`?
* why do we have two apparently unrelated functions to retrieve element text?
* we used to use `getElementText` in some way because older versions of selenium don't have the problem im seeing

**The issue is that _some_ shadow DOM children return empty text but do have text**

Selenium is incredibly difficult to debug and often involves jumping through many hoops. So without spending another several hours trying to get my head around it, i cant really figure out where these "atoms" go and how they are used. if someone could give an overview somewhere, that would be _very_ useful.

I had to pretty much attach to the chrome instance (devtools), break, then try to access the text of the element in my tests so my devtools catches the injected code. the injected code is obviously minified by this point and near unreadable. is there a better debug process/method for debugging the injected JS?

**FYI, executing `getElementText` manually (through a very hacky execute call via webdriver) on the element results in the correct text.**

---

edit: i opened #5762 for this

43081j:

This fixes #5761.

Though it raises another issue: that in several places we only support shadow dom v0, rather than the v1 spec. most of our tests use v0 too, so we don't test against v1 apis like `attachShadow` etc.

### DIFF
--- a/javascript/atoms/dom.js
+++ b/javascript/atoms/dom.js
@@ -567,12 +567,15 @@ bot.dom.isShown_ = function(elem, ignoreOpacity, parentsDisplayedFn) {
 bot.dom.isShown = function(elem, opt_ignoreOpacity) {
   /**
    * Determines whether an element or its parents have `display: none` set
-   * @param {?Node}
+   * @param {!Node} e the element
    * @return {boolean}
    */
   function displayed(e) {
-    if (bot.dom.isElement(e) && bot.dom.getEffectiveStyle(e, 'display') == 'none') {
-      return false;
+    if (bot.dom.isElement(e)) {
+      var elem = /** @type {!Element} */ (e);
+      if (bot.dom.getEffectiveStyle(elem, 'display') == 'none') {
+        return false;
+      }
     }
 
     var parent = bot.dom.getParentNodeInComposedDom(e);


43081j:
FYI the tests fail in some cases regardless of my change... not sure how those got through to master.

in the shadow dom tests, we create a shadow root twice on a single element, which throws an exception. these tests are a little muddled and confusing so im not sure if just removing the call is fine.

javascript/atoms/test/html5/shadow_dom_test.html
```diff
@@ -175,12 +184,7 @@ <H4>Younger Child Shadow</H4>
     var parentHeading = /** @type {!Element} */ (document.querySelector(
       "* /deep/ #parentHeading"));
     var oldChildDiv = document.querySelector("* /deep/ #olderChildDiv");
-    var neverShownDiv = /** @type {!Element} */ (document.querySelector(
-      "* /deep/ #neverShownDiv"));
-  
-    // neverShown is in an older shadow DOM, and thus isn't displayed
-    assertFalse(bot.dom.isShown(neverShownDiv));
```
43081j: FYI i removed this test because AFAIK its invalid per the old spec and doesn't apply in the new spec (as `createShadowRoot` has gone).

this test relies on the fact chrome once allowed you to create a shadow root multiple times, which was removed shortly after. now it just throws an exception or doesn't do anything.

43081j:
unfortunately the tests in `shadow_dom_test` are very confusing and thrown together. they depend heavily in shadow dom V0 _and_ behaviours that lasted for only a couple of browser releases.

most of the tests in there are to assert that "older shadow roots" work correctly. but there is no such thing as an "older shadow root", it was deprecated long ago and throws exceptions in most or all browsers. 

i tried to strip those out so we can at least have a passing build, but it leaves us with a mess of a test file/suite. so i think if we can at least agree we shouldn't be testing non-existent browser features (removed v0 things only chrome had temporarily), its better i just try to rewrite that test file.

barancev:
With this patch the test is still broken for me (in chrome):
```
localhost:2310/javascript/atoms/test/html5/shadow_dom_test.html
3 of 3 tests run in 8ms.
0 passed, 3 failed.
3 ms/test. 92 files loaded.
.
23:41:26.861  Start
23:41:26.864  testGetParentNodeInComposedDom : FAILED (run individually)
23:41:26.865  ERROR in testGetParentNodeInComposedDom
Expected <null> but was <[object HTMLContentElement]> (HTMLContentElement)
Error: Expected <null> but was <[object HTMLContentElement]> (HTMLContentElement)
    at Object.goog.testing.asserts.raiseException (http://localhost:2310/third_party/closure/goog/testing/asserts.js:1283:11)
    at _assert (http://localhost:2310/third_party/closure/goog/testing/asserts.js:189:26)
    at assertEquals (http://localhost:2310/third_party/closure/goog/testing/asserts.js:418:3)
    at testGetParentNodeInComposedDom (http://localhost:2310/javascript/atoms/test/html5/shadow_dom_test.html:121:5)
    at goog.testing.TestCase.invokeTestFunction_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:899:21)
    at goog.testing.TestCase.safeRunTest_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:849:8)
    at goog.testing.TestCase.invokeTestFunction_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:927:19)
    at goog.testing.TestCase.runNextTest_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:839:8)
    at goog.testing.TestCase.<anonymous> (http://localhost:2310/third_party/closure/goog/testing/testcase.js:790:12)
    at goog.testing.TestCase.invokeTestFunction_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:927:19)
    at goog.testing.TestCase.runSetUpPage_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:803:8)
    at goog.testing.TestCase.<anonymous> (http://localhost:2310/third_party/closure/goog/testing/testcase.js:781:10)
    at new goog.Promise (http://localhost:2310/third_party/closure/goog/promise/promise.js:172:16)
    at goog.testing.TestCase.runTestsReturningPromise (http://localhost:2310/third_party/closure/goog/testing/testcase.js:780:10)
    at goog.testing.TestRunner.execute (http://localhost:2310/third_party/closure/goog/testing/testrunner.js:264:19)
    at http://localhost:2310/third_party/closure/goog/testing/jsunit.js:179:12
23:41:26.869  testGetVisibleTextInComposedDom : FAILED (run individually)
23:41:26.869  ERROR in testGetVisibleTextInComposedDom
Expected <Older Child
As the older child of a nested shadow root, this is the most likely to go wrong bit of the page.
Older Child Contents
Parent
Parent Contents
Base-level Stuff> (String) but was <> (String)
Error: Expected <Older Child
As the older child of a nested shadow root, this is the most likely to go wrong bit of the page.
Older Child Contents
Parent
Parent Contents
Base-level Stuff> (String) but was <> (String)
    at Object.goog.testing.asserts.raiseException (http://localhost:2310/third_party/closure/goog/testing/asserts.js:1283:11)
    at _assert (http://localhost:2310/third_party/closure/goog/testing/asserts.js:189:26)
    at assertEquals (http://localhost:2310/third_party/closure/goog/testing/asserts.js:418:3)
    at testGetVisibleTextInComposedDom (http://localhost:2310/javascript/atoms/test/html5/shadow_dom_test.html:175:5)
    at goog.testing.TestCase.invokeTestFunction_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:899:21)
    at goog.testing.TestCase.safeRunTest_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:849:8)
    at goog.testing.TestCase.invokeTestFunction_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:927:19)
    at goog.testing.TestCase.runNextTest_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:839:8)
    at goog.testing.TestCase.finishTestInvocation_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:999:10)
    at goog.testing.TestCase.invokeTestFunction_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:927:19)
    at goog.testing.TestCase.safeTearDown_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:864:8)
    at goog.testing.TestCase.invokeTestFunction_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:934:15)
    at goog.testing.TestCase.safeRunTest_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:849:8)
    at goog.testing.TestCase.invokeTestFunction_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:927:19)
    at goog.testing.TestCase.runNextTest_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:839:8)
    at goog.testing.TestCase.<anonymous> (http://localhost:2310/third_party/closure/goog/testing/testcase.js:790:12)
    at goog.testing.TestCase.invokeTestFunction_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:927:19)
    at goog.testing.TestCase.runSetUpPage_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:803:8)
    at goog.testing.TestCase.<anonymous> (http://localhost:2310/third_party/closure/goog/testing/testcase.js:781:10)
    at new goog.Promise (http://localhost:2310/third_party/closure/goog/promise/promise.js:172:16)
    at goog.testing.TestCase.runTestsReturningPromise (http://localhost:2310/third_party/closure/goog/testing/testcase.js:780:10)
    at goog.testing.TestRunner.execute (http://localhost:2310/third_party/closure/goog/testing/testrunner.js:264:19)
    at http://localhost:2310/third_party/closure/goog/testing/jsunit.js:179:12
23:41:26.869  testIsShownInComposedDom : FAILED (run individually)
23:41:26.869  ERROR in testIsShownInComposedDom
Argument to isShown must be of type Element
Error: Argument to isShown must be of type Element
    at Object.bot.dom.isShown_ (http://localhost:2310/javascript/atoms/dom.js:456:11)
    at Object.bot.dom.isShown (http://localhost:2310/javascript/atoms/dom.js:607:18)
    at testIsShownInComposedDom (http://localhost:2310/javascript/atoms/test/html5/shadow_dom_test.html:197:25)
    at goog.testing.TestCase.invokeTestFunction_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:899:21)
    at goog.testing.TestCase.safeRunTest_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:849:8)
    at goog.testing.TestCase.invokeTestFunction_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:927:19)
    at goog.testing.TestCase.runNextTest_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:839:8)
    at goog.testing.TestCase.finishTestInvocation_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:999:10)
    at goog.testing.TestCase.invokeTestFunction_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:927:19)
    at goog.testing.TestCase.safeTearDown_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:864:8)
    at goog.testing.TestCase.invokeTestFunction_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:934:15)
    at goog.testing.TestCase.safeRunTest_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:849:8)
    at goog.testing.TestCase.invokeTestFunction_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:927:19)
    at goog.testing.TestCase.runNextTest_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:839:8)
    at goog.testing.TestCase.finishTestInvocation_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:999:10)
    at goog.testing.TestCase.invokeTestFunction_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:927:19)
    at goog.testing.TestCase.safeTearDown_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:864:8)
    at goog.testing.TestCase.invokeTestFunction_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:934:15)
    at goog.testing.TestCase.safeRunTest_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:849:8)
    at goog.testing.TestCase.invokeTestFunction_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:927:19)
    at goog.testing.TestCase.runNextTest_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:839:8)
    at goog.testing.TestCase.<anonymous> (http://localhost:2310/third_party/closure/goog/testing/testcase.js:790:12)
    at goog.testing.TestCase.invokeTestFunction_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:927:19)
    at goog.testing.TestCase.runSetUpPage_ (http://localhost:2310/third_party/closure/goog/testing/testcase.js:803:8)
    at goog.testing.TestCase.<anonymous> (http://localhost:2310/third_party/closure/goog/testing/testcase.js:781:10)
    at new goog.Promise (http://localhost:2310/third_party/closure/goog/promise/promise.js:172:16)
    at goog.testing.TestCase.runTestsReturningPromise (http://localhost:2310/third_party/closure/goog/testing/testcase.js:780:10)
    at goog.testing.TestRunner.execute (http://localhost:2310/third_party/closure/goog/testing/testrunner.js:264:19)
    at http://localhost:2310/third_party/closure/goog/testing/jsunit.js:179:12
23:41:26.869  Done
 
Error: Failed to execute 'createShadowRoot' on 'Element': Shadow root cannot be created on a host which already hosts a shadow tree.
    at http://localhost:2310/javascript/atoms/test/html5/shadow_dom_test.html:91:9
```

43081j:
@barancev to be honest those tests are mostly useless in newer chrome.

most of the coverage in there is over this concept of older shadow roots (multiple roots). its not even supported anymore by spec or by browsers. i think the tests just need rewriting completely in that file.

ill give that a go tomorrow if i get chance

43081j:
@barancev @p0deje  any chance either of you could take a look at this?

we're in great need of these fixes so i just need to get the builds working and the tests passing. but it seems they fail persistently now because of unreachable chrome

barancev:
Test results:
```
localhost:2310/javascript/atoms/test/html5/shadow_dom_test.html
9 of 9 tests run in 17ms.
1 passed, 8 failed.
2 ms/test. 92 files loaded.
```

To run a single test you can start a debug server with a command `go debug-server` and open test URL in a browser: http://localhost:2310/javascript/atoms/test/html5/shadow_dom_test.html

43081j:
ah i wasn't aware you could do that, will give it a go.

but you can see the CI builds fail for seemingly unrelated reasons (to my changes). we should probably still figure that out..

43081j:
those tests pass now.

though CI builds fail still, due to focus tests. they pass locally, so i guess its something to do with the fact the window isn't focused on CI builds maybe. not too sure.

also, see my comments in the code review when you have a min if you can

javascript/atoms/dom.js
```diff
@@ -574,24 +574,26 @@ bot.dom.isShown = function(elem, opt_ignoreOpacity) {
...
+      if (parent instanceof ShadowRoot) {
+        if (parent.host.shadowRoot != parent) {
+          // There is a younger shadow root, which will take precedence over
+          // the shadow this element is in, thus this element won't be
+          // displayed.
+          return false;
+        } else {
+          parent = parent.host;
         }
-      } while (elem && elem.nodeType != goog.dom.NodeType.ELEMENT);
```
43081j: i removed this loop because `elem` was never modified, so it would either run once or forever... guess we never ended up with this being as truthy

javascript/atoms/dom.js
```diff
@@ -601,7 +603,7 @@ bot.dom.isShown = function(elem, opt_ignoreOpacity) {
         return false;
       }
       var parent = bot.dom.getParentElement(e);
-      return !parent || displayed(parent);
+      return parent && displayed(parent);
```
43081j: does this reversal seem right? it seems before we were assuming a parentless element was visible. we should be assuming the opposite

javascript/atoms/dom.js
43081j: we have some recursion below to deal with traversing the tree anyway, so i guess this removal is fine.

javascript/atoms/dom.js
43081j: an element without a parentNode could be a disconnected node.

also, as far as you go up the chain, there will be a parentNode, until it is `#document`. 

`body` has a parentNode of `document`, and we check that elsewhere (we return true somewhere else when `node === document`). so i dont see a situation where we would end up with a parentless node that is still somehow part of the document.

43081j: we're walking up parent nodes. this calls `getParentNodeInComposedDom` which (as seen below) uses `parentNode`

javascript/atoms/dom.js
```diff
@@ -1260,6 +1262,13 @@ bot.dom.getOpacityNonIE_ = function(elem) {
  */
 bot.dom.getParentNodeInComposedDom = function(node) {
   var /**@type {Node}*/ parent = node.parentNode;
+  // Shadow DOM v1
+  if (parent.shadowRoot && node.assignedSlot !== undefined) {
+    // Can be null on purpose, meaning it has no parent as
+    // it hasn't yet been slotted
+    parent = node.assignedSlot ? node.assignedSlot.parentNode : null;
```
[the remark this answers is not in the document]
43081j: consistency, the existing code didn't do an early return so i didn't either. if you'd rather have an early return, i'll change the other conditions below too.

javascript/atoms/dom.js
jleyba: The line immediately preceding this one:

`var parent = bot.dom.getParentElement(e);`

[getParentElement](https://github.com/SeleniumHQ/selenium/blob/master/javascript/atoms/dom.js#L363) returns the parent element or null.

43081j: ah my mistake, it shouldn't really be doing either to be honest. an element without a parent is more likely to be a disconnected element than document, its not the best assertion really.

we are better off asserting that it is `document` or not. if it isn't we do this line as is, otherwise we return true.

43081j: in actual fact @jleyba  why do we have this branch at all? `getParentNodeInComposedDom` works without shadow DOM support (it checks for support internally). why not drop this `else` completely and use the same function?

javascript/atoms/dom.js
```diff
@@ -567,43 +567,34 @@ bot.dom.isShown_ = function(elem, ignoreOpacity, parentsDisplayedFn) {
...
+  // Any element with a display style equal to 'none' or that has an ancestor
+  // with display style equal to 'none' is not shown.
+  displayed = function(e) {
+    if (bot.dom.getEffectiveStyle(e, 'display') == 'none') {
+      return false;
+    }
+
+    var parent = bot.dom.getParentNodeInComposedDom(e);
+
+    if (bot.dom.IS_SHADOW_DOM_ENABLED && (parent instanceof ShadowRoot)) {
```
43081j: @jleyba  take a look at this now. so we no longer need an if/else as the parent node traversal already checks if shadow dom v0 or v1 is supported.

and this means now we automatically get the document/shadowRoot assertion further down and disconnected elements are falsey.

javascript/atoms/dom.js
```diff
@@ -567,43 +567,34 @@ bot.dom.isShown_ = function(elem, ignoreOpacity, parentsDisplayedFn) {
...
-      return !parent || displayed(parent);
-    };
-  } else {
-    // Any element with a display style equal to 'none' or that has an ancestor
-    // with display style equal to 'none' is not shown.
-    displayed =  function(e) {
-      if (bot.dom.getEffectiveStyle(e, 'display') == 'none') {
+  // Any element with a display style equal to 'none' or that has an ancestor
+  // with display style equal to 'none' is not shown.
+  displayed = function(e) {
```
jleyba: nit: remove the `var displayed;` declaration above and change this to:

```js
/**
 * @param {?Node} e
 * @return {boolean}
 */
function displayed(e) {
```

The extra type annotations are annoying, but helps improve the compiler's accuracy in type checking

javascript/atoms/dom.js
```diff
@@ -567,43 +567,34 @@ bot.dom.isShown_ = function(elem, ignoreOpacity, parentsDisplayedFn) {
...
+    if (bot.dom.isElement(e) && bot.dom.getEffectiveStyle(e, 'display') == 'none') {
```
jleyba: Should this be

`!bot.dom.isElement(e) || bot.dom.getEffectiveStyle(e, 'display') == 'none'`

Since this function should only be called when checking the visibility of a parent, we're assuming the DOM is well-formed and `e` will never be a text node.

43081j: i would rather keep it as is to be honest. if its an element, we check the style, otherwise we continue traversing. we shouldn't really be making assumptions on what can and can't be a parent, but instead just traversing what we are given IMO.

43081j:
@jleyba @barancev 

this seems ready now, one last look over it if you could?

current CI build fails for seemingly unrelated reasons. looks like the failing tests are all focus related, i can reproduce them locally by having chrome unfocused (certain events dont fire if chrome has no focus as a whole).

shs96c:
@43081j, sorry it took so long to finish this review, but thank you! Thanks to @barancev and @jleyba for reviewing this too!

43081j:
much appreciated. thanks to you all for the help, reviews, etc. this'll be of great use to a whole bunch of people

PR #5820, opened 2018-04-24
Fix issue where -host is not honored in all server modes.

mach6:
Also increase the scope of -host to all roles [standalone, hub, and
node]

---
Before this change
```shell
$ java -jar selenium-standalone.jar -role hub --host {ip}
...
```
Actual: bind happens on `*` adapters
```shell
$ netstat -an | grep 4444
tcp46      0      0  *.4444                 *.*                    LISTEN
```

Problem: Jetty does not honor `-host` unless you tell it to.

---
After this change
```shell
$ java -jar selenium-standalone.jar -role hub -host {ip}
```
Actual: bind happens on `-host` adapter
```shell
$ netstat -an | grep 4444
tcp46      0      0  {ip}.4444                 *.*                    LISTEN
```

---
This change exposes `-host` to `-role standalone` (default role) also, to fix the same issue.

### DIFF
--- a/java/server/src/org/openqa/grid/internal/cli/CommonCliOptions.java
+++ b/java/server/src/org/openqa/grid/internal/cli/CommonCliOptions.java
@@ -67,6 +67,16 @@ public abstract class CommonCliOptions {
   private String role;
 
   /**
+   * Hostname or IP to use. Defaults to {@code null}. Automatically determined when {@code null}.
+   */
+  @Parameter(
+      names = "-host",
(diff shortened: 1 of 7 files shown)


mach6:
@shs96c @barancev FYI, I'm pretty sure this change set is safe, but since I'm less active these days, rather than commit this directly, I opened the PR. Please take a look! :)

mach6:
peer approval via irc

PR #5957, opened 2018-05-30
Fixed quit command for default state

InstyleVII:

The checks that determine whether EdgeDriver should use /shutdown or terminate the process were failing when UseSpecCompliantProtocol wasn't set to a value. The flag is a nullable bool, where null/true means the driver should run in W3C mode while false means the driver should run in JWP mode. The check was checking both for a value and for true, but since a null state has no value it failed the check and fell back to the JWP Quit path which calls /shutdown (which doesn't exist in W3C mode for MicrosoftWebDriver.exe).

The fix changes the logic such that it validates when the flag is set to false, as that is the only time the JWP Quit path should be used, in the other two cases (null/true) it should use the W3C Quit path and terminate the process.

### DIFF
--- a/dotnet/src/webdriver/Edge/EdgeDriverService.cs
+++ b/dotnet/src/webdriver/Edge/EdgeDriverService.cs
@@ -99,12 +99,12 @@ namespace OpenQA.Selenium.Edge
         {
             get
             {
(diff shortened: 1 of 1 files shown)


jimevans:
Merged in 24432b3. Thank you for the contribution!

PR #5968, opened 2018-05-31
Loosen WebElementCondition instance checks

### LINKED ISSUE #5560: WebElementCondition not instanceof Condition
## Meta -
OS:  
OSX 10.13.3

Selenium Version:  
selenium-webdriver@4.0.0-alpha.1

Browser:  
Chrome

Browser Version:  
Version 64.0.3282.186 (Official Build) (64-bit)

## Expected Behavior -
`driver.wait()` should accept a WebElementCondition, i.e. `driver.wait(until.elementLocated(By.css(selector)))`

## Actual Behavior -
the above example throws the error `TypeError: Wait condition must be a promise-like object, function, or a Condition object`

## Steps to reproduce -
`driver.wait(until.elementLocated(By.css(selector)))`

This is some serious JavaScript weirdness:
`Object.getPrototypeOf(condition) ===  WebElementCondition {}`
yet `condition instanceof WebElementCondition === false`
weird, right?

I was able to get the case to pass by adding `condition.constructor.name === 'WebElementCondition'` to `if (condition instanceof Condition)` in the WebDriver's wait function.

afsmith92:
Update WebElementCondition and Condition with custom
static hasInstance methods so that `instanceof` checks
are more lenient and don't cause issues with Babel.

Fixes #5560

### DIFF
--- a/javascript/node/selenium-webdriver/lib/webdriver.js
+++ b/javascript/node/selenium-webdriver/lib/webdriver.js
@@ -77,10 +77,11 @@ class Condition {
    * @param {!(IThenable<T>|
    *           Condition<T>|
    *           function(!WebDriver): T)} condition - the condition instance
-   * @returns {*|boolean}
+   * @return {boolean}
    */
   static [Symbol.hasInstance](condition) {
-    return condition
+    return !!condition
+      && typeof condition === 'object'
       && typeof condition.description === 'function'
       && typeof condition.fn === 'function';
   }
@@ -102,19 +103,6 @@ class WebElementCondition extends Condition {
    */
   constructor(message, fn) {
     super(message, fn);
-  }
-
-  /**
-   * Allows for lenient instanceof checks
-   * @param {!(IThenable<T>|
-   *           Condition<T>|
-   *           function(!WebDriver): T)} condition - the condition instance
-   * @returns {*|boolean}
-   */
-  static [Symbol.hasInstance](condition) {
-    return condition
-      && typeof condition.description === 'function'
-      && typeof condition.fn === 'function';
   }
 }
 


javascript/node/selenium-webdriver/lib/webdriver.js
```diff
@@ -71,6 +71,19 @@ class Condition {
   description() {
     return this.description_;
   }
+
+  /**
+   * Allows for lenient instanceof checks
+   * @param {!(IThenable<T>|
+   *           Condition<T>|
+   *           function(!WebDriver): T)} condition - the condition instance
+   * @returns {*|boolean}
```
jleyba: `@return {boolean}`

javascript/node/selenium-webdriver/lib/webdriver.js
```diff
@@ -71,6 +71,19 @@ class Condition {
...
+   */
+  static [Symbol.hasInstance](condition) {
+    return condition
```
jleyba: `return !!condition`

javascript/node/selenium-webdriver/lib/webdriver.js
```diff
@@ -90,6 +103,19 @@ class WebElementCondition extends Condition {
...
   }
+
+  /**
+   * Allows for lenient instanceof checks
+   * @param {!(IThenable<T>|
+   *           Condition<T>|
+   *           function(!WebDriver): T)} condition - the condition instance
+   * @returns {*|boolean}
+   */
+  static [Symbol.hasInstance](condition) {
```
jleyba: no need to define this a second time, it will be inherited from `Condition`

javascript/node/selenium-webdriver/lib/webdriver.js
jleyba: Actually, I'd probably be a little stricter:

`return !!condition && typeof condition === 'object'`

(plus the next two lines)

barancev:
I've reverted this change by commit 61583115b8f360bafba302d0a3beb3eb8ab373f4 because it broke `x instanceof WebElementCondition` check that returns true even if `x` is an instance of the `Condition` class (a base class for `WebElementCondition`). All waits that return a boolean value or a string are affected.

The issue will be reopened too.

PR #5974, opened 2018-06-01
Add before/after getText events

mpanin:
Listener was adjusted to handle before/after getText method calls are being made

Add before/after getText listener events


shs96c:
Nice addition. Thank you!

Issue #5907, opened 2018-05-17
selenium's Python wheel distribution contains duplicate files and may fail to install


wiggin15:
## Meta -
OS:  
Windows
Selenium Version:  
3.12.0

## Expected Behavior -
The selenium wheel file should be installable with "setuptools".

## Actual Behavior -
Using `setuptools 39.1.0` on Windows to install the selenium Python wheel distribution (`selenium-3.12.0-py2.py3-none-any.whl`) fails with:

```
error: [Error 183] Cannot create a file when that file already exists
```

Reason: the selenium wheel file contains the following files in the wheel file:

```
selenium/webdriver/firefox/amd64/x_ignore_nofocus.so
selenium-3.12.0.data/data/selenium/webdriver/firefox/amd64/x_ignore_nofocus.so
```

setuptools does not handle this well on Windows, and the project maintainers claims it's the wheel file that's invalid (even though `pip` successfully installs the wheel, silently overriding duplicate files). See https://github.com/pypa/setuptools/pull/1371. 
Possible resolution: `data_files` in `setup.py` should not include files that are already packaged.

## Steps to reproduce -
```
> wget https://files.pythonhosted.org/packages/57/bc/17164fd471ccdf0df3a992c710c0c3c47743462ff41ab72a02c6ede96e90/selenium-3.12.0-py2.py3-none-any.whl
> easy_install selenium-3.12.0-py2.py3-none-any.whl
Processing selenium-3.12.0-py2.py3-none-any.whl
Installing selenium-3.12.0-py2.py3-none-any.whl to c:\cygwin\home\administrator\python\lib\site-packages
error: [Error 183] Cannot create a file when that file already exists
```

lmtierney:
@davehunt thoughts?

davehunt:
It sounds like a packaging issue, I don't see why we'd need to have duplicate files in the distribution. It won't be long before we can probably remove these files altogether, as they're not used for recent versions of Firefox.

lmtierney:
Closed via #5982

PR #5982, opened 2018-06-03
Remove package_data and data_files options from setup.py

wiggin15:
With `include_package_data` option set to `True` in `setup.py`, package data that is specified in MANIFEST.in is installed into the target directory when installing selenium. This means that we don't need to specify `package_data` and `data_files` in setup.py too. With the options set as before, files were included twice in wheel distributions (once in their original place and once in the ".data" directory) and that triggered a bug that caused setuptools to fail to install the wheel distribution of selenium.

I verified that running `python setup.py bdist_wheel` and `python setup.py sdist` create distributions with the data files inside and that the data files are installed when installing the created distributions.

Fixes issue #5907

### DIFF
--- a/py/setup.py
+++ b/py/setup.py
@@ -67,14 +67,6 @@ setup_args = {
                  'selenium.webdriver.phantomjs',
                  'selenium.webdriver.remote',
                  'selenium.webdriver.support', ],
-    'package_data': {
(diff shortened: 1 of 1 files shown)


lmtierney:
Ok, I've finally had time to try this on a Windows machine and it looks like it's fine. We'll get it into the next release and if it produces problems we'll be able to immediately do a point release.

lmtierney:
Thank you for your contribution!

PR #5989, opened 2018-06-04
Add executing custom command to Chrome Python bindings

zhangwenqiang00:
implement send_command_and_get_result interface, so we can control
chrome via devtools protocol commands.

### DIFF
--- a/py/selenium/webdriver/chrome/remote_connection.py
+++ b/py/selenium/webdriver/chrome/remote_connection.py
@@ -25,4 +25,4 @@ class ChromeRemoteConnection(RemoteConnection):
         self._commands["launchApp"] = ('POST', '/session/$sessionId/chromium/launch_app')
         self._commands["setNetworkConditions"] = ('POST', '/session/$sessionId/chromium/network_conditions')
         self._commands["getNetworkConditions"] = ('GET', '/session/$sessionId/chromium/network_conditions')
-        self._commands['ExecuteCDP'] = ('POST', '/session/$sessionId/goog/cdp/execute')
+        self._commands['executeCdpCommand'] = ('POST', '/session/$sessionId/goog/cdp/execute')
--- a/py/selenium/webdriver/chrome/webdriver.py
+++ b/py/selenium/webdriver/chrome/webdriver.py
@@ -120,7 +120,7 @@ class WebDriver(RemoteWebDriver):
         Execute Chrome Devtools Protocol command and get returned result
 
         The command and command args should follow chrome devtools protocol domains/commands, refer to link
-        https://chromium.googlesource.com/chromium/src/+/master/third_party/blink/renderer/devtools/protocol.json
+        https://chromedevtools.github.io/devtools-protocol/
 
         :Args:
          - cmd: A str, command name
@@ -136,7 +136,7 @@ class WebDriver(RemoteWebDriver):
             {'base64Encoded': False, 'body': 'response body string'}
 
         """
-        return self.execute("ExecuteCDP", {'cmd': cmd, 'params': cmd_args})['value']
+        return self.execute("executeCdpCommand", {'cmd': cmd, 'params': cmd_args})['value']
 
     def quit(self):
         """


py/selenium/webdriver/chrome/webdriver.py
```diff
@@ -115,6 +115,28 @@ def set_network_conditions(self, **network_conditions):
             'network_conditions': network_conditions
         })
 
+    def send_command_and_get_result(self, cmd, cmd_args):
```
AutomatedTester: This is really a poor API name since most methods in any language are essentially send a command and get a result

lmtierney:
There's also just a `send_command` last time I checked, unless they got rid of it? I had this change all ready to push to master previously, but guidance from @kereliuk was that this endpoint was not intended for public use will probably change in the near future. @kereliuk should we still hold off on adding support for these endpoints?

lmtierney:
I did find our previous conversation and updated the issue mentioning this functionality accordingly https://github.com/SeleniumHQ/selenium/issues/5159#issuecomment-371863188

lmtierney:
Unfortunately there has been no activity from the chromium team on this matter thus far

kereliuk:
@lmtierney Whoops, the /goog/ rename fell through my stack. I'm not 100% sure historically why the endpoint was added, but it points to a commonly used internal function call that is used and I assume was not intended to be used publicly.

I did discuss this with other Chromium folks and I can submit a change to chromedriver to have the endpoint be in the correct /goog/ extension style by EOD today and it will be pushed in our next release.

Here is a [bug](https://bugs.chromium.org/p/chromedriver/issues/detail?id=2307&q=owner%3Ame&colspec=ID%20Status%20Pri%20Owner%20Summary) tracking the issue.

I guess in this PR it makes sense to either
 - rename the function name and have it hit the endpoint and return a not implemented error.
 - rename the function name and have it hit the send_command endpoint
 - wait until the next ChromeDriver is released and then push this PR with everything updated

Either way, I can own the responsibility of either submitting a PR here or pinging Lucas if I can't figure out where add the name change once the next ChromeDriver is out :)

lmtierney:
How about we wait until the next chromedriver release, then we can have @zhangwenqiang00 update the PR

kereliuk:
Sounds good!

py/selenium/webdriver/chrome/webdriver.py
zhangwenqiang00: shouldn't we be consistent with [chromedriver interface name](https://cs.chromium.org/chromium/src/chrome/test/chromedriver/server/http_handler.cc?g=0&l=566)? 
or change to "send_devtools_command_and_get_result" is ok?

kereliuk:
ChromeDriver 2.40 has been released with the endpoint /goog/cdp/execute. Should be good to go! @zhangwenqiang00

zhangwenqiang00:
updated! @kereliuk

py/selenium/webdriver/chrome/remote_connection.py
```diff
@@ -25,3 +25,4 @@ def __init__(self, remote_server_addr, keep_alive=True):
         self._commands["launchApp"] = ('POST', '/session/$sessionId/chromium/launch_app')
         self._commands["setNetworkConditions"] = ('POST', '/session/$sessionId/chromium/network_conditions')
         self._commands["getNetworkConditions"] = ('GET', '/session/$sessionId/chromium/network_conditions')
+        self._commands['ExecuteCDP'] = ('POST', '/session/$sessionId/goog/cdp/execute')
```
kereliuk: Should this be camelCase to match the above commands?

I am not the expert here so maybe Lucas or David can chime in.

py/selenium/webdriver/chrome/webdriver.py
```diff
@@ -115,6 +115,29 @@ def set_network_conditions(self, **network_conditions):
             'network_conditions': network_conditions
         })
 
+    def execute_cdp_cmd(self, cmd, cmd_args):
+        """
+        Execute Chrome Devtools Protocol command and get returned result
+
+        The command and command args should follow chrome devtools protocol domains/commands, refer to link
+        https://chromium.googlesource.com/chromium/src/+/master/third_party/blink/renderer/devtools/protocol.json
```
kereliuk: Might be good to add this link in the docstring as I think it is a better more readable source.
https://chromedevtools.github.io/devtools-protocol/

lmtierney:
Sorry I was on vacation this past week or so. It looks like you've addressed @kereliuk's changes and we are planning on releasing a new version tomorrow so I'll pull it in now. The only thing we might think of changing is either changing the `cmd_args` to default to `None` or accept `**kwargs`

lmtierney:
Thank you!

PR #6014, opened 2018-06-12
fix incorrect w3c action encoding in python client

jlipps:
according to the spec, the pointerType should be a pointer type, not an arbitrary name/id. correct me if I'm wrong @AutomatedTester but this seems like a bug.

### DIFF
--- a/py/selenium/webdriver/common/actions/pointer_input.py
+++ b/py/selenium/webdriver/common/actions/pointer_input.py
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 from .input_device import InputDevice
-from .interaction import POINTER, POINTER_KINDS, POINTER_MOUSE
+from .interaction import POINTER, POINTER_KINDS
 
 from selenium.common.exceptions import InvalidArgumentException
 from selenium.webdriver.remote.webelement import WebElement
@@ -25,7 +25,7 @@ class PointerInput(InputDevice):
 
     DEFAULT_MOVE_DURATION = 250
 
-    def __init__(self, name, kind=POINTER_MOUSE):
+    def __init__(self, kind, name):
         super(PointerInput, self).__init__()
         if (kind not in POINTER_KINDS):
             raise InvalidArgumentException("Invalid PointerInput kind '%s'" % kind)
--- a/py/selenium/webdriver/common/actions/action_builder.py
+++ b/py/selenium/webdriver/common/actions/action_builder.py
@@ -62,8 +62,8 @@ class ActionBuilder(object):
         self._add_input(new_input)
         return new_input
 
-    def add_pointer_input(self, type_, name):
-        new_input = PointerInput(type_, name)
+    def add_pointer_input(self, kind, name):
+        new_input = PointerInput(kind, name)
         self._add_input(new_input)
         return new_input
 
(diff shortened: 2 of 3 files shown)


lmtierney:
@jlipps You're right in the fact that it shouldn't be giving it the name/id, but the `pointerType` should only be one of 3 values currently of `mouse`, `pen`, or `touch` and the driver should return an error if it is not one of these (per the spec). The change you proposed would set this `pointerType` to `pointer`, so we actually probably need to take in a `kind` parameter into the `__init__` and then automatically set the `type` encode key to `pointer` since we're in `PointerInput` already. If that makes sense?

Relevant spec excerpt:

> If pointer type is not undefined:
> 
> 1. If pointer type does not have one of the values "mouse", "pen", or "touch" return error with error code invalid argument.

lmtierney:
See the implementation in Java:
https://github.com/SeleniumHQ/selenium/blob/6337d9067a5cc7f5e68f5f3fc1d938a175331172/java/client/src/org/openqa/selenium/interactions/PointerInput.java#L51

And/or Ruby:

https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/common/interactions/pointer_input.rb#L35

jlipps:
@lmtierney ah yes, thanks for the correction. is something like this what you had in mind? unfortunately it's a breaking change to __init__...

py/selenium/webdriver/common/actions/pointer_input.py
```diff
@@ -15,17 +15,22 @@
...
+from selenium.common.exceptions import InvalidArgumentException
 from selenium.webdriver.remote.webelement import WebElement
 
 
 class PointerInput(InputDevice):
 
     DEFAULT_MOVE_DURATION = 250
 
-    def __init__(self, type_, name):
+    def __init__(self, name, kind=POINTER_MOUSE):
```
lmtierney: Keep `kind` as the first argument and it _shouldn't_ be a breaking change. There should be no reason why people would have been using their own PointerInput since `mouse` is the only pointer currently supported. Just in case though we should wait for a point release to make the change. We do need to update the 1 use within the library (below) to use interaction.POINTER_MOUSE instead of interaction.POINTER. 

https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/common/actions/pointer_actions.py#L30

lmtierney:
Just out of curiosity, did you find this issue with changes you're making to appium?

py/selenium/webdriver/common/actions/pointer_input.py
jlipps: sounds good, will make these changes as soon as I can

jlipps:
> Just out of curiosity, did you find this issue with changes you're making to appium?

More or less. Appium now implements the w3c action protocol on the server side. I was trying to use the python client to construct a multi-pointer action, and the Appium server correctly complained about the pointerType values that were being sent in by the client, which is how I discovered the error.

lmtierney:
Very cool, nice to see appium already supporting w3c. You may run into another bug that I've been meaning to fix, but I can't quite remember what it was. Had to do with putting the right number of pauses in for other inputs when using `send_keys` for a key input

jlipps:
@lmtierney right on! changes made.

lmtierney:
@AutomatedTester do you want to take a quick look at this before we merge?

netzulo:
i need this :(

PR #6034, opened 2018-06-18
Add status as property on webdriver

olescheller:

The Python implementation is missing the call on the endpoint `/status` as documented here: https://www.w3.org/TR/webdriver/#status
Coming from Appium, I need to use this endpoint for checking the server version, mentioned as 'additional meta info' in the spec.

### DIFF
--- a/py/selenium/webdriver/remote/webdriver.py
+++ b/py/selenium/webdriver/remote/webdriver.py
@@ -1248,3 +1248,13 @@ class WebDriver(object):
             driver.get_log('server')
         """
         return self.execute(Command.GET_LOG, {'type': log_type})['value']
+
+    @property
+    def status(self):
+        """
+        Returns the status of the remote server.
+
+        :Usage:
+            driver.status
+        """
+        return self.execute(RemoteCommand.STATUS)['value']


lmtierney:
Hmm, I'm not sure this is something that should be a property of `WebDriver`. This command is able to be sent to any remote end without a session ID and Driver instances are tied to a session ID. This would be more applicable to a utility or even a basic `get` outside of Selenium/WebDriver.

shs96c:
The status is designed to figure out whether or not a new session could be created. It can include other data, but that's its primary purpose. Putting this method on `WebDriver` really doesn't feel right --- by the time we have an instance, we know that the server is running.

lmtierney:
I suppose we could do something like put it as a `staticmethod` off of `Remote` WebDriver and just have it take in a URL param

netzulo:
have no sense at WebDriver instance, but, maybe included as property at `service.py` files ? what do you think @shs96c

lmtierney:
@netzulo service wouldn't make sense for remote endpoints which is what this would mainly be used for

netzulo:
then a staticmethod will be good for all of us, :) ty for the point @lmtierney

netzulo:
@oscheller1 @lmtierney , what about : https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/common/action_chains.py

---

*sorry for spam guys*

lmtierney:
Just wanted to say I haven't forgotten about this, still mulling it over

lmtierney:
Ok I've thought about it and at this time I don't think we'll add it. You can use a simple GET request yourself or even use selenium as is:

```
>>> from selenium.webdriver.remote.remote_connection import RemoteConnection
>>> from selenium.webdriver.remote.command import Command
RemoteConnection('http://localhost:4444/wd/hub').execute(Command.STATUS, {})
{'status': 0, 'value': {'ready': True, 'message': 'Hub has capacity', 'build': {'revision': 'aacccce0', 'time': '2018-08-02T20:13:22.693Z', 'version': '3.14.0'}, 'os': {'arch': 'amd64', 'name': 'Linux', 'version': '4.15.0-34-generic'}, 'java': {'version': '1.8.0_181'}}}
```

PR #6078, opened 2018-06-26
Fixes empty options merge [fixes #6071]

### LINKED ISSUE #6071: [rb] Options overwrite custom desired capabilities
## Meta
Selenium-rb gem version:  
selenium-webdriver (3.13.0)

Browser: 
Chrome 67.0.3396.87 (Official Build) (64-bit)

## Driver setup

```ruby

Capybara.register_driver(:selenium_chrome_headless) do |app|
  capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
    'goog:chromeOptions': { args: %w[no-sandbox headless disable-gpu window-size=1024,768] }
  )

  Capybara::Selenium::Driver.new(app, browser: :chrome, desired_capabilities: capabilities)
end
```

I know I can setup args using `args` (deprecated) or options

## Expected Behavior

Create driver with given capabilities

## Actual Behavior

Options overwrite custom capabilities because of https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/chrome/driver.rb#L106

```ruby
caps.merge!(options) unless options.empty?
```

`options` here is not empty. It equals `{"goog:chromeOptions" => {} }`

## Steps to reproduce

Added spec and possible fix in:
https://github.com/artplan1/selenium/commit/9b14903b44b8136f418ce520aa564150603244d5

Possible fix:

```ruby
caps.merge!(options) unless options[Options::KEY].empty?
```

or `as_json` in Options.rb can return `{}` instead of `{"goog:chromeOptions" => {} }`
not sure if it's required with recent changes

Spec:

```ruby
it 'does not merge empty options' do
  custom_caps = Remote::Capabilities.new('goog:chromeOptions' => {args: %w[foo bar]})

  expect(http).to receive(:call) do |_, _, payload|
    expect(payload[:desiredCapabilities]['goog:chromeOptions'][:args]).to eq(%w[foo bar])
    resp
  end

  Driver.new(http_client: http, desired_capabilities: custom_caps)
end
```

Thanks

artplan1:

Fixes #6071 

but i don't feel good about this code actually.

`options[Options::KEY].empty?`

checking for key here feels dirty to me 🤔

I like changing return of `as_json` to `opts ? {KEY => opts} : {}` more, but not sure if even empty `goog:chromeOptions` is required

Also I tried to refactor specs related to `custom desired capabilities` to be more DRY, lmk if it worths it

### DIFF
--- a/rb/spec/unit/selenium/webdriver/chrome/driver_spec.rb
+++ b/rb/spec/unit/selenium/webdriver/chrome/driver_spec.rb
@@ -51,6 +51,12 @@ module Selenium
           expect(caps[:proxy]).to eq(proxy)
         end
 
+        it 'does not set goog:chromeOptions by default' do
+          Driver.new(http_client: http)
+
+          expect(caps['goog:chromeOptions']).to be nil
+        end
+
         it 'does not set the chrome.detach capability by default' do
           Driver.new(http_client: http)
 
@@ -86,44 +92,55 @@ module Selenium
           expect(caps['goog:chromeOptions'][:extensions]).to eq(profile_data[:extensions])
         end
 
-        it 'takes desired capabilities' do
-          custom_caps = Remote::Capabilities.new
-          custom_caps[:chrome_options] = {'foo' => 'bar'}
-
-          expect(http).to receive(:call) do |_, _, payload|
-            expect(payload[:desiredCapabilities][:chrome_options]).to include('foo' => 'bar')
-            resp
-          end
-
-          Driver.new(http_client: http, desired_capabilities: custom_caps)
-        end
-
-        it 'lets direct arguments take presedence over capabilities' do
-          custom_caps = Remote::Capabilities.new
-          custom_caps['goog:chromeOptions'] = {'args' => %w[foo bar]}
-
-          expect(http).to receive(:call) do |_, _, payload|
-            expect(payload[:desiredCapabilities]['goog:chromeOptions'][:args]).to eq(['baz'])
-            resp
-          end
-
-          Driver.new(http_client: http, desired_capabilities: custom_caps, args: %w[baz])
-        end
-
-        it 'does not merge empty options' do
-          custom_caps = Remote::Capabilities.new('goog:chromeOptions' => {args: %w[foo bar]})
-
-          expect(http).to receive(:call) do |_, _, payload|
-            expect(payload[:desiredCapabilities]['goog:chromeOptions'][:args]).to eq(%w[foo bar])
-            resp
-          end
-
-          Driver.new(http_client: http, desired_capabilities: custom_caps)
-        end
-
         it 'handshakes protocol' do
           expect(Remote::Bridge).to receive(:handshake)
           Driver.new(http_client: http)
+        end
+
+        context 'with custom desired capabilities' do
+          subject(:build_new_driver) do
+            Driver.new(http_client: http, desired_capabilities: custom_caps, args: driver_args)
+          end
+
+          let(:custom_caps) { Remote::Capabilities.new(cap_opts) }
+          let(:cap_opts) { {chrome_options: {'foo' => 'bar'}} }
+          let(:driver_args) { [] }
+
+          it 'takes desired capabilities' do
+            expect(http).to receive(:call) do |_, _, payload|
+              expect(payload[:desiredCapabilities][:chrome_options]).to include('foo' => 'bar')
+              resp
+            end
+
+            build_new_driver
+          end
+
+          context 'with direct arguments' do
+            let(:cap_opts) { {'goog:chromeOptions' => {args: %w[foo bar]}} }
+            let(:driver_args) { %w[baz] }
+
+            it 'lets direct arguments take precedence over capabilities' do
+              expect(http).to receive(:call) do |_, _, payload|
+                expect(payload[:desiredCapabilities]['goog:chromeOptions'][:args]).to eq(driver_args)
+                resp
+              end
+
+              build_new_driver
+            end
+          end
+
+          context 'with empty driver options' do
+            let(:cap_opts) { {'goog:chromeOptions' => {args: %w[foo bar]}} }
+
+            it 'does not merge empty options' do
+              expect(http).to receive(:call) do |_, _, payload|
+                expect(payload[:desiredCapabilities]['goog:chromeOptions'][:args]).to eq(%w[foo bar])
+                resp
+              end
+
+              build_new_driver
+            end
+          end
         end
       end
     end # Chrome


rb/spec/unit/selenium/webdriver/chrome/driver_spec.rb
```diff
@@ -54,7 +54,6 @@ module Chrome
         it 'does not set the chrome.detach capability by default' do
           Driver.new(http_client: http)
 
-          expect(caps['goog:chromeOptions']).to eq({})
```
luke-hill: This spec can be modified, giving you extra strength (Check this fetch key doesn't work and therefore returns `nil`)

rb/spec/unit/selenium/webdriver/chrome/driver_spec.rb
```diff
@@ -111,6 +110,17 @@ module Chrome
           Driver.new(http_client: http, desired_capabilities: custom_caps, args: %w[baz])
         end
 
+        it 'does not merge empty options' do
```
luke-hill: This actually isn't testing this if you check your code. Here you are passing in 2 arguments and verifying you receive 2 args. This is still a missing use case. But then you should have the blank one.

artplan1: Thank you for feedback.
I don't understand this one.
You mean check the case when `goog:chromeOptions` is empty in args in custom desired_capabilities?

artplan1: Also I tried to refactor specs to be more DRY, lmk if it worths it

rb/spec/unit/selenium/webdriver/chrome/driver_spec.rb
artplan1: I added separate spec to check `goog:chromeOptions` https://github.com/SeleniumHQ/selenium/pull/6078/files#diff-0396c977ce3bbece63a6b7dd4b3ef3c7R54

luke-hill: Agreed this is better

rb/spec/unit/selenium/webdriver/chrome/driver_spec.rb
```diff
@@ -87,33 +92,53 @@ module Chrome
...
-        it 'takes desired capabilities' do
-          custom_caps = Remote::Capabilities.new
-          custom_caps[:chrome_options] = {'foo' => 'bar'}
+        it 'handshakes protocol' do
+          expect(Remote::Bridge).to receive(:handshake)
+          Driver.new(http_client: http)
+        end
+
+        context 'with custom desired capabilities' do
+          subject(:build_new_driver) { Driver.new(http_client: http, desired_capabilities: custom_caps, args: driver_args) }
```
luke-hill: Can do/end this to make the line a bit shorter

rb/spec/unit/selenium/webdriver/chrome/driver_spec.rb
```diff
@@ -87,33 +92,53 @@ module Chrome
...
-          Driver.new(http_client: http, desired_capabilities: custom_caps)
-        end
+          context 'with direct arguments' do
+            let(:cap_opts) { { 'goog:chromeOptions' => {args: %w[foo bar]} } }
+            let(:driver_args) { %w[baz] }
 
-        it 'lets direct arguments take presedence over capabilities' do
-          custom_caps = Remote::Capabilities.new
-          custom_caps['goog:chromeOptions'] = {'args' => %w[foo bar]}
+            it 'lets direct arguments take presedence over capabilities' do
```
luke-hill: ~lets~ precedence*

artplan1:
@luke-hill 

thank you for reviewing

re:
> But I still feel as though you're missing the vital unit test.
> let(:cap_opts) { { 'goog:chromeOptions' => {} } }
> it 'discards empty goog:chromeOptions'

`goog:chromeOptions` will be {} in this case

like:
`expect(payload[:desiredCapabilities]['goog:chromeOptions']).to eq({})`

so it doesn't really discard it. It just sets it with given value.

in the case:

let(:cap_opts) { { } }

it will be

`expect(payload[:desiredCapabilities]['goog:chromeOptions']).to be nil`

am I missing something

luke-hill:
I've probably misunderstood something then. I thought from the original issue. When goog:chromeOptions was set to an empty hash you wanted it to be sanitized out - and removed from the payload. But it seems like the reverse is the case.

i.e. when you set capabilities to be empty it *was* settings goog:chromeOptions as an empty hash. And now you want it to not exist?

If I'm wrong with both of these then :man_shrugging:

artplan1:
original issue is that even if `goog:chromeOptions` is set in custom desired capabilities - `options` overwrite it here `caps.merge!(options) unless options.empty?` because `options` is never empty because it returns `{KEY => opts}`

this spec shows it

```ruby
context 'with empty driver options' do
  let(:cap_opts) { { 'goog:chromeOptions' => {args: %w[foo bar]} } }

  it 'does not merge empty options' do
    expect(http).to receive(:call) do |_, _, payload|
      expect(payload[:desiredCapabilities]['goog:chromeOptions'][:args]).to eq(%w[foo bar])
      resp
    end

    build_new_driver
  end
end
```

failing result without check for `goog:chromeOptions` key in options

```
Selenium::WebDriver::Chrome::Driver with custom desired capabilities with empty driver options does not merge empty options
     Failure/Error: expect(payload[:desiredCapabilities]['goog:chromeOptions'][:args]).to eq(%w[foo bar])

       expected: ["foo", "bar"]
            got: nil
```

re:
> when you set capabilities to be empty it was settings goog:chromeOptions as an empty hash. And now you want it to not exist?

this is a question - I don't know if it still should exist if it's not set in desired capabilities or it can be empty. If it should exist - then my fix is wrong and something like `reverse_merge` should be applied here. But then `args` will not take precendence and it's wrong.

luke-hill:
Is it the case that (after your code changes)
- When `goog:chromeOptions` is set to an empty hash you want it to be sanitized out - and removed from the payload?

Or was it the case that (before your code changes)
- When you set capabilities to be empty it was settings `goog:chromeOptions` as an empty hash

Or is it something else?

Failing this can you pop into #selenium on IRC or slack and chat in person to save spamming the notifications here.

luke-hill:
After discussion / re-reading of code. Can see how the changes have been tested (Both no / some driver args input, which was a technical "missing" test from before regarding the googOptions property)

Possibly some syntax / style stuff that I've missed, but my own personal style is different to that which the `rb` files are written in. So I'll let someone like @p0deje chip in here. As he has the ability to approve/merge.

PR #6100, opened 2018-06-30
Make selenium jars usable with Java 9 (and above)

### LINKED ISSUE #5896: Can't compile Java9 module with selenium-java as dependency
Selenium Version:  
3.11.0

I'm trying to build a Java9 module which has selenium-java as maven dependency.

My module-info.java looks something like this:
```java
module my.module {

    exports my.package;

    requires httpcore;
    requires json;
    requires selenium.api;
    requires selenium.chrome.driver;
    requires selenium.edge.driver;
    requires selenium.firefox.driver;
    requires selenium.ie.driver;
    requires selenium.remote.driver;
    requires selenium.support;
}
```

However, when compiling my module with maven I get the following error message:
module my.module reads package org.openqa.selenium.logging from both selenium.api and selenium.remote.driver

This seems to happen, because both selenium-api and selenium-remote-driver contain a package called "org.openqa.selenium.logging". Java 9 (and all later versions) forbid this. This problem is called "split packages".

Is there any chance, we can get this problem fixed in the next releases? Or is there some other dependecy I should use instead of selenium-java (for example a single JAR that contains everything)?

seombre:
The module system introduced with Java 9 forbids having a package in
more than one jar file. The following three packages are affected by
this:
- org.openqa.selenium.logging
- org.openqa.selenium.interactions
- org.openqa.selenium.interactions.internal

This change introduces new packages and moves a couple of classes
around, so that each jar file has only distinct packages.

Fixes #5896

Note: This breaks the tests (run with ./go clean test) because HtmlUnit still excpects HasInputDevices in the package org.openqa.selenium.interactions (which had to move). I seriously have no idea how to fix this, or at least disable the HtmlUnit tests.
This is the error message:
```console
PARSING BUCK FILES... 0,8s (100%)
CREATING ACTION GRAPH: FINISHED IN 0,2s
Build failed: Command failed with exit code 1.
stderr: /home/ombre/dev/projects/selenium/java/client/test/org/openqa/selenium/testing/TestUtilities.java:42: error: cannot access org.openqa.selenium.interactions.HasInputDevices
      return ((HtmlUnitDriver) driver).getBrowserVersion().getUserAgent();
                                      ^
  class file for org.openqa.selenium.interactions.HasInputDevices not found
Errors: 1. Warnings: 0.

    When running <javac_jar>.
    When building rule //java/client/test/org/openqa/selenium/testing:test-base.
DOWNLOADED 0 ARTIFACTS, 0,00 BYTES
BUILDING: FINISHED IN 6,5s (100%) 321/326 JOBS, 56 UPDATED, 0,3% CACHE MISS
BUILD FAILED
go aborted!
```

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6100)


seombre:
Btw, there are some unwanted comment changes in the following files:
- MoveTargetOutOfBoundsException.java
- Pause.java
- BaseAction.java
- TouchAction.java

I think those come from a refactoring done with IntelliJ and should be reverted.

Jarob22:
Selenium works just fine in our project which is using java 10 - java 9 is already eol, and there's not much point adding extra stuff to try and support just it if 10 works, imo.

seombre:
@Jarob22 Please have a look at the reproducer in https://github.com/mhomnag/selenium-java10-reproducer
That project actually uses Java 10.
Which dependencies of Selenium did you use, btw?

Jarob22:
@seombre selenium-java, selenium-server, selenium-firefox-driver, selenium-support, selenium-remote-driver, selenium-api and selenium-chrome-driver.
Ah, we are not using modules, I guess that's why its fine for us :) Sorry about that then, guess its still an issue.

shs96c:
The process we use for deprecating and moving interfaces is that we mark things as `@Deprecated` and add a javadoc telling people where to look before deleting classes, specifically to avoid problems like the one you highlighted with HtmlUnit.

The best thing to do when moving an interface is to copy it to the new location, and make the _old_ interface extend the _new_ one (that is `org.openqa.selenium.interactions.HasTouchScreen` extend `org.openqa.selenium.interactions.inputs.HasTouchScreen`) That allows old code to continue to work, but means that it can also be migrated to the new interface easily.

I'm also unsure about the best way to split things up. It's clear that each of the individual drivers should be in their own modules, and that there should be _at least_ one for `remote`, but I'm still not sure what the best thing to do with the core module is. We have several options:

1. Drop the requirement that the core module have no external dependencies, but limit it to (say) just guava
2. Re-arrange classes into a more meaningful hierarchy, probably separating implementations from interfaces, but avoid the dreaded `impl` package naming structure.
3. Do some combination of the above.

I think that maybe this PR goes a little too far towards `2`.

@barancev and @jleyba, any thoughts? @seombre, your input is, of course, very welcome too :)

seombre:
You're right, directly moving the classes, without deprecating them first, really isn't a good solution. I could submit a pull request for that in the next few days.

I'm also really not sure, whether the new package names I've chosen are that good. Suggestions welcome :)

Btw, another possible (but obviously not very elegant solution) would be to have a new jar, that includes everything.

shs96c:
It may be worth rebasing on top of the latest version of master. I've reworked the core package (including actions) to avoid taking a dependency on guava, which helps a little.

seombre:
OK, give a couple of days, to have a look at this.
@shs96c So the strategy is now to first make copies of the existing interfaces/classes and deprecate the old ones?

shs96c:
@seombre after some recent work, I think there are now no longer any split packages in `java/client/src`. I think the correct boundary for a java 9 module is probably "anything that has a maven_coords attribute", as those are the distributable pieces we've selected.

The main thing we have to do is to somehow build the `module-info.java` class and add it to the generated jars. This is proving a challenge, since buck hasn't got support for this yet.

shs96c:
My current WIP on this is here: https://github.com/SeleniumHQ/selenium/commit/f80a93e021e3ef87629b9bb4742dddd1c987bde4

Frankly, it's not pretty at all.

seombre:
Any way I can help with this? Is this pull request still needed?

shs96c:
@seombre I wish I knew. I wish my answer was better than that. I suspect we need some changes in Buck and those will be non-trivial to make. Notably, we need to support merging module-info classes for uberjars (or, I guess, dropping them entirely, though I bet that would cause trouble somewhere)

shs96c:
@seombre and @barancev: I landed e57914ab0550e7b755bfd664d61040ae7856b03d. I _think_ we now have sensible java module definitions in place. If anyone could have a look at this and verify that the modules actually work, that'd be great. You need to build the maven jars to get everything packaged the way we expect them to:

```
./buckw build --show-output \
  java/client/src/org/openqa/selenium:core#maven \
  java/client/src/org/openqa/selenium/remote:remote#maven \
  java/client/src/org/openqa/selenium/firefox:firefox#maven
```

The above should give you the modules you need to for running a firefox instance. Feedback very much appreciated.

seombre:
Very nice! Will try this this evening.

seombre:
OK, so the good news is that with your changes the original reproducer I posted (https://github.com/mhomnag/selenium-java10-reproducer/commit/bc638891395ca28ef5ae4fc5ecb330991c95a79b) now actually builds!

However, I then tried to do some basic stuff (see current master of above repo) and it seems that all the classes directly in "org.openqa.selenium" just cannot be found. And the reason for this seems to be, that i had to require selenium.support for the WebDriverWaiter. When I remove that require statement, WebElement, etc. are actually found, but not the WebDriverWaiter, obviously.

seombre:
I think, I've just found another problem: "requires transitive httpclient;" (https://github.com/SeleniumHQ/selenium/blob/e57914ab0550e7b755bfd664d61040ae7856b03d/java/client/src/org/openqa/selenium/remote/module-info.txt) doesn't seem to work, when you have a project, that also has a dependency on the Apache HttpClient.

What did work (in another module though) was adding "requires org.apache.httpcomponents.httpclient" to my module-info.java. Btw, this only affects the httpclient. httpcore seems to work this way, though.

seombre:
Any update on this? Can I help with something?
The next Java release is coming close and since that's a LTS release, there's probably gonna be a bunch of people that would like to switch.

skyhirider:
I can't build selenium-server-standalone-3.141.59.jar with Java11, getting
`java.lang.module.FindException: Unable to derive module descriptor` 

But only adding the client dependency to my Java11 project does not prevent it from being built.

Is the client Java11 compatible while the standalone server is not?

cedric05:
any plan for releasing this?

alexkunger:
Is this PR still active? Are there plans to merge this and make Selenium compatible with versions of Java greater than 8?

kowcik:
In Selenium v4.0.0.0-alpha-2 i could see a changelog "Better support for Java 9+ modules". Today i tried creating a java module (openjdk 11) with selenium v4.0.0-alpha-6 (which also includes the changes of open modules) but still getting the same error v4.0.0-alpha-6
org.openqa.selenium.chrome cannot be a resolved module

shs96c:
I'm working on fixing the java module stuff properly now. I've rejigged how we generate the modules themselves, and the next step is to fix the packaging so that files are included in only one module. We shouldn't need to be moving classes around as that suggests our maven artifacts are incorrectly built to, and we should fix that by setting up our build files properly.

Thank you folks for reminding me about this. I'm going to close this PR, but will move #6798 into the project plan for Selenium 4, and I'm working on this for the next beta.

PR #6101, opened 2018-07-01
[rb] Upload file to browser when file detector set in w3c bridge

### LINKED ISSUE #6064: [rb] send_keys_to_element method in W3C bridge does not support file detector
## Meta -
OS:  
Any

Selenium Version:  
3.12.0

Browser:  
Firefox (Remote Driver) which using w3c protocol

Browser Version:  
49+

## Expected Behavior -

Using file_detector and send_keys method I can upload file  from local machine to remote browser.

## Actual Behavior -

Ruby binding for w3c protocol does not support file_detector. File not upload. Client get error "File not found: <path\to\file>"

OSS bridge with file_detector support: https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/remote/oss/bridge.rb#L365

W3C without file_detector support: https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/remote/w3c/bridge.rb#L362

mskvn:
Fixes #6064

### DIFF
--- a/rb/lib/selenium/webdriver/remote/w3c/bridge.rb
+++ b/rb/lib/selenium/webdriver/remote/w3c/bridge.rb
@@ -371,7 +371,7 @@ module Selenium
 
           def upload(local_file)
             unless File.file?(local_file)
-              Webdriver.logger.debug("File detector only works with files. #{local_file.inspect} isn`t a file!")
+              WebDriver.logger.debug("File detector only works with files. #{local_file.inspect} isn`t a file!")
               raise Error::WebDriverError, "You are trying to work with something that isn't a file."
             end
 
--- a/rb/lib/selenium/webdriver/remote/oss/bridge.rb
+++ b/rb/lib/selenium/webdriver/remote/oss/bridge.rb
@@ -372,7 +372,7 @@ module Selenium
 
           def upload(local_file)
             unless File.file?(local_file)
-              Webdriver.logger.debug("File detector only works with files. #{local_file.inspect} isn`t a file!")
+              WebDriver.logger.debug("File detector only works with files. #{local_file.inspect} isn`t a file!")
               raise Error::WebDriverError, "You are trying to work with something that isn't a file."
             end
 


rb/lib/selenium/webdriver/remote/w3c/bridge.rb
```diff
@@ -358,13 +358,25 @@ def click_element(element)
...
+              keys = [upload(local_file)] if local_file
+            end
             # Keep .split(//) for backward compatibility for now
             text = keys.join('')
             execute :element_send_keys, {id: element}, {value: text.split(//), text: text}
           end
 
+          def upload(local_file)
+            unless File.file?(local_file)
+              raise Error::WebDriverError, "you may only upload files: #{local_file.inspect}"
```
luke-hill: I'd alter this message to be a bit more verbose. Here you're taking an input which you know will only be called when file_detector is set. But your message says you may only upload files (Which is what you're doing)

mskvn: Sorry, but I not understand what you mean :) This check only show that local_file is file or not. How message should be verbose? 

I have only one idea. Add to message something like this: "File detector set, but return not file" or "File detector should return file". But in this case maybe this check should move after file_detector.call

P.S: This part of code I copied from oss bridge :)

luke-hill: How about...

"You are trying to work with something that isn't a file." (Don't really need object inspect here)
"File detector only works with Files. #{local_file.inspect} isnt a file!" <- Obj inspect makes sense here.

P.s. if this is copied from OSS Bridge I'd probably fix it up there too.

mskvn: Got it. Fixed :)

luke-hill:
To OP @mskvn one failure in one of the travis runs is related to this potentially. Might need to fix up some specs related to your change or have a look

Start here: https://travis-ci.org/SeleniumHQ/selenium/jobs/399531688

mskvn:
Yep, sorry, made typo in constant name :(
Already fixed.

PR #6127, opened 2018-07-07
Enriching Hub Status to include Node info

krmahadevan:
Currently the end-point /grid/api/hub/status 
does not provide information about the nodes 
attached to it and also the busy/free status 
per browser flavor on each of the nodes.

Enriching the end-point to provide node info 
when invoked with the option 
/grid/api/hub/status?configuration=nodes

Sample payload as below:

```json
{
  "nodes": [
    {
      "Id": "http://192.168.1.6:5555",
      "browsers": [
        {
          "browser": "safari",
          "slots": {
            "busy": 0,
            "total": 1
          }
        },
        {
          "browser": "chrome",
          "slots": {
            "busy": 0,
            "total": 5
          }
        },
        {
          "browser": "firefox",
          "slots": {
            "busy": 0,
            "total": 5
          }
        }
      ]
    }
  ],
  "success": true
}
```

### DIFF
--- a/java/server/src/org/openqa/grid/web/servlet/HubStatusServlet.java
+++ b/java/server/src/org/openqa/grid/web/servlet/HubStatusServlet.java
@@ -19,6 +19,8 @@ package org.openqa.grid.web.servlet;
 
 import static org.openqa.selenium.json.Json.MAP_TYPE;
 
+import com.google.common.base.Splitter;
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableSortedMap;
 import com.google.common.collect.Lists;
 
@@ -35,7 +37,6 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.Writer;
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -70,6 +71,14 @@ import javax.servlet.http.HttpServletResponse;
  */
 public class HubStatusServlet extends RegistryBasedServlet {
 
+  private static final String SUCCESS = "success";
+  private static final String CONFIGURATION = "configuration";
+  private static final String FREE = "free";
+  private static final String BUSY = "busy";
+  private static final String NEW_SESSION_REQUEST_COUNT = "newSessionRequestCount";
+  private static final String SLOT_COUNTS = "slotCounts";
+  private static final String NODES = "nodes";
+  private static final String TOTAL = "total";
   private final Json json = new Json();
 
   public HubStatusServlet() {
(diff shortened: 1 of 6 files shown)


krmahadevan:
ping @diemol @shs96c @barancev 

@shs96c - As discussed in the SeConf in BLR, I have enriched the `/grid/api/hub/status` end-point to optionally provide node details when queried. I will send a separate PR for the node side.

luke-hill:
For my own clarity, would this provide essentially what `http://url-to-grid/get_node_config?node=nameofnode` does, but instead of being for 1 node it is for all nodes?

krmahadevan:
@luke-hill,

* I am not aware of an end-point "/get_node_config" available in the Hub
that is currently doing this. Can you please help point me to it. I am
currently just enriching this
<https://github.com/SeleniumHQ/selenium/blob/master/java/server/src/org/openqa/grid/web/Hub.java#L134>
endpoint
* To answer to your question, no it does not because we aren't getting hold
of the configuration of the node (which is static), but this is dynamic
information and will reflect exactly how many sessions are in use vs free
per browser flavor on each of the nodes at any given point in time.

luke-hill:
The top url I mentioned might be an SGE addon (Not sure).

Rightio. But sorry to be a pain. Isn't that provided (I hope this isn't SGE as well!), by `url-to-grid/grid/console`

krmahadevan:
Yes the console ( url-to-grid/grid/console) provides you with this info.
But it's html and not JSON. So you would need to screen scrape to get this
data. This PR makes it easy to extract out that info.

> --

Thanks & Regards
Krishnan Mahadevan

"All the desirable things in life are either illegal, expensive, fattening
or in love with someone else!"
My Scribblings @ http://wakened-cognition.blogspot.com/

My Technical Scribbings @ http://rationaleemotions.wordpress.com/

krmahadevan:
@mach6 - Thanks for the comments. I have extracted out all repeating strings in that class and converted them into private constants. Please take a look.

java/server/src/org/openqa/grid/web/servlet/HubStatusServlet.java
```diff
@@ -167,4 +185,57 @@ protected void process(
       throw new IOException(e);
     }
   }
+
+  private static boolean IsKeyPresentIn(List<String> keys, String key) {
```
shs96c: Nit: java method names begin with a lower-case letter. We can fix this once the PR is merged in.

java/server/src/org/openqa/grid/web/servlet/HubStatusServlet.java
```diff
@@ -109,35 +124,38 @@ protected void process(
...
-    res.put("success", true);
+    res.put(SUCCESS, true);
 
     try {
         List<String> keysToReturn = null;
 
-        if (request.getParameter("configuration") != null && !"".equals(request.getParameter("configuration"))) {
-          keysToReturn = Arrays.asList(request.getParameter("configuration").split(","));
-        } else if (requestJSON != null && requestJSON.containsKey("configuration")) {
+        if (request.getParameter(CONFIGURATION) != null && !"".equals(request.getParameter(CONFIGURATION))) {
```
shs96c: You can use `!Strings.isNullOrEmpty(request.getParameter(CONFIGURATION))` here for a slightly clearer check here (I realise that this wasn't done before)

java/server/src/org/openqa/grid/web/servlet/HubStatusServlet.java
```diff
@@ -109,35 +124,38 @@ protected void process(
...
+          keysToReturn = Arrays.asList(request.getParameter(CONFIGURATION).split(","));
```
shs96c: Are empty values allowed in the list? If not `Splitter.on(",").trimResults().omitEmptyStrings().splitToList()` may be a better fit.

java/server/src/org/openqa/grid/web/servlet/HubStatusServlet.java
```diff
@@ -167,4 +185,57 @@ protected void process(
...
+  }
+
+  private List<Map<String, Object>> getNodesInfo() {
+    List<RemoteProxy> proxies = getRegistry().getAllProxies().getSorted();
+    return proxies.stream().map(this::getNodeInfo).collect(toList());
+  }
+
+  private Map<String, Object> getNodeInfo(RemoteProxy remoteProxy) {
+    return ImmutableSortedMap.of(
+        "Id", remoteProxy.getId(),
```
shs96c: Why is `Id` capitalised, but `browser` not? Maybe leave both lower case?

java/server/src/org/openqa/grid/web/servlet/HubStatusServlet.java
```diff
@@ -167,4 +185,57 @@ protected void process(
...
+    }
+    return BUSY;
+  }
+
+  private static String getBrowser(TestSlot slot) {
+    return slot.getCapabilities().get(CapabilityType.BROWSER_NAME).toString();
+  }
+
+  private static <T> Collector<T, ?, Integer> counting() {
+    return reducing(0, e -> 1, Integer::sum);
```
shs96c: This looks like an IntStreams `sum` method, but presumably that isn't appropriate to use here

krmahadevan:
@shs96c 

>  I'd ask for a test of some description to ensure that this continues working....

I have now added required tests for this changeset and also addressed all the other review feedback. Please help take a look.

peterabbott:
This appears to be a breaking change when using an existing grid for < 3.14 as it is expected the enhanced metadata. A graceful fall back could have been nice as the update was getting pulled in transitively via other dependencies.

PR #650, opened 2015-06-11
WebDriver to implement AutoClosable

asashour:
Let WebDriver interface implements AutoCloseable

[AutoCloseable](http://docs.oracle.com/javase/7/docs/api/java/lang/AutoCloseable.html) is introduced since Java 7, and WebDriver
already defines close() method. Making WebDriver implements
AutoCloseable is backward compatible. Now WebDriver can be used
in [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) statement,
and the IDE could hint the developer that WebDriver should be closed
(if it is not called).

PR #6133, opened 2018-07-09
Made WebDriver extend AutoCloseable

sbhat98:

I found a [previous closed PR about this](https://github.com/SeleniumHQ/selenium/pull/650), but I would like to open this up to more discussion since I still do not understand why making it implement AutoCloseable would break the method contract. Overall implementing the interface would be helpful to users since IDEs would give a warning to close the driver if it is never closed.


barancev:
I can just reiterate: the semantics of the WebDriver `close()` method do not match the expected semantics of `AutoClosable.close()`. This method does not close the driver and the browser, it closes the current window only, so it can't be used as a 'destructor'.

PR #6176, opened 2018-07-18
make ConnectionType searilize as integer

velma:
https://github.com/SeleniumHQ/mobile-spec/blob/master/spec-draft.md#device-modes
expects connection type to be serialized to an integer as well as
chromedriver does not accept currently sent string value.


shs96c:
Seems like a reasonable change to me. Thank you for taking the time to put the PR together, and congratulations on your first selenium commit!

PR #6177, opened 2018-07-18
Remove Sizzle and jQuery 1.4.1 as well as all non-native Xpath engines

corevo:
## Why?
Using the browser's native APIs (e.g. [document.evaluate](https://developer.mozilla.org/en-US/docs/Web/API/Document/evaluate) and [document.querySelectorAll](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll)) will reduce the dependencies on old third party code, making it easier for the IDE to pave the way to WebDriver based playback (Sizzle for that matter uses [non-standard css capabilities](https://github.com/jquery/sizzle/wiki#additions), non-compatible with WebDriver).  
And because Mozilla is cracking down on jQuery 1 - 2 users.
>Hello,
>
>You’re receiving this message because you’re an add-on developer registered on AMO (addons.mozilla.org) and and you are using jQuery 2.x in at least one of your add-ons. We want to inform you about the upcoming deprecation of that version.
>
>jQuery 2.x is no longer maintained and contains vulnerabilities that could lead to security issues in add-ons. Effective immediately, the validation process will warn you if jQuery 2.x is found during the submission of a new version or add-on. After September 1, 2018, no new submissions using jQuery 2.x will be accepted.
>
>Please upgrade to jQuery 3.x as soon as possible. Refer to the upgrade guide at https://jquery.com/upgrade-guide/3.0/ for more information.
>
>Regards,
>
>The Add-ons Team

Since [htmlutils.js](https://github.com/SeleniumHQ/selenium/pull/6177/files#diff-3ced8e380800cd7858d11aa54bd0043b) comes bundled with jQuery 1.4.1 and Sizzle, after September 1st, we will no longer be able to publish new versions on AMO.


jimevans:
You can’t simply remove this. At least one language binding uses it as part of its WebDriverBackedSelenium implementation.

corevo:
Whats your suggestion then? to have two versions of it?

jimevans:
I don’t know what the right way forward is. IE doesn’t have a native XPath engine, but the driver uses a different JavaScript implementation, so that’s not an issue. The IE driver still imports Sizzle, but that’s mainly vestigial, and I need to remove it. The .NET bindings uses an `htmlutils.js` file for `WebDriverBackedSelenium`, but I’ll need to double-check to see if it’s the same one you’re changing here. If so, we’ll need to fork it, at worst. I’m not in a place to check right now, but I’ll be on IRC in about an hour and we can chat more there.

corevo:
I'll ping you there

shs96c:
We can always update our version of sizzle if necessary.

corevo:
I really don't like the idea of supporting non-standard css capabilities,  
Which tests should I run to make sure that doesn't break anything?

shs96c:
@corevo, the old IDE has always supported those non-standard CSS features. To remove it from the emulation would cause the tool to no longer be backward compatible.

barancev:
I should note that Selenium IDE TNG is already not backward compatible. It does not support waitFor commands because it waits automatically, it does not support storeEval. So we are defacto free to make other changes that break compatibility.

As a plus side of this change, we'll be able to export scenarios to a programming language and use vanilla WenDriver with no Sizzle to run them (as soon as we implement export).

corevo:
Exactly my thoughts.
I was planning on adding support for running on Webdriver straight from the IDE.
I want to bring it more in line with Webdriver, which is also why storeEval was changed for executeScript (the latter executes inside anonymous function rather than Selenium's prototype).

Cheers,
Tomer

shs96c:
The backward compatibility is to do with the `WebDriverBackedSelenium` not with IDE.

corevo:
Unless there is a plan to continue releasing new versions of `WebDriverBackedSelenium`, we can just keep the old file in the history, we can tag this commit as well.

corevo:
Will be fixed in [323](https://github.com/SeleniumHQ/selenium-ide/pull/323)

PR #6181, opened 2018-07-19
Added a basic ServiceBuilder for Internet Explorer

### LINKED ISSUE #5983: Missing ServiceBuilder equivalent for IE (Node)
## Meta -
OS: Windows 10
Selenium Version: selenium-webdriver@4.0.0-alpha.1
Browser:  Internet Explorer
Browser Version:  11

## Expected Behavior -

I'd like to configure a custom driver location for `IEDriverServer.exe`. Other browsers export a `ServiceBuilder` you can use:

```javascript
const seleniumFirefox = require("selenium-webdriver/firefox");

builder.setFirefoxService(new seleniumFirefox.ServiceBuilder("./path/to/geckoDriver.exe"));
```

## Actual Behavior -

No equivalent `ServiceBuilder` class or method that allows a custom path to `IEDriverServer.exe` is provided.

[`locateSynchronously`](https://github.com/SeleniumHQ/selenium/blob/e060906bbdbd954da21883f7f6fdd71cb48db3dd/javascript/node/selenium-webdriver/ie.js#L311) only allows checking for the user's path.

Relevant: https://stackoverflow.com/questions/35587825/internet-explorer-driver-without-the-need-of-path-variable

JoshuaKGoldberg:
Adds it to IE's `Driver` and allows it to be specified with a new `setIeService` in the general `Builder`.

Fixes #5983.

I'm very unfamiliar with all this, so other than verifying that it works locally, I haven't written tests. It looks pretty similar to the existing service builders, but confirmation would be nice!

Edit: some test jobs pass, some fail. 🤷

### DIFF
--- a/javascript/node/selenium-webdriver/index.js
+++ b/javascript/node/selenium-webdriver/index.js
@@ -223,6 +223,9 @@ class Builder {
 
     /** @private {ie.Options} */
     this.ieOptions_ = null;
+  
+    /** @private {ie.ServiceBuilder} */
+    this.ieService_ = null;
 
     /** @private {safari.Options} */
     this.safariOptions_ = null;


javascript/node/selenium-webdriver/index.js
```diff
@@ -490,6 +490,18 @@ class Builder {
...
 
+  /**
+   * Sets the {@link ie.ServiceBuilder} to use to manage the geckodriver
+   * child process when creating IE sessions locally.
+   *
+   * @param {ie.ServiceBuilder} service the service to use.
+   * @return {!Builder} a self reference.
+   */
+  setIeService(service) {
+    this.ieService_ = service;
```
jleyba: `this.ieService_` should be initialized to null in the constructor.

JoshuaKGoldberg: Done!

PR #6189, opened 2018-07-20
constant default_move_duration prevents appium to override it

prakharrr:

### DIFF
--- a/rb/lib/selenium/webdriver/common/interactions/pointer_actions.rb
+++ b/rb/lib/selenium/webdriver/common/interactions/pointer_actions.rb
@@ -18,7 +18,8 @@
 module Selenium
   module WebDriver
     module PointerActions
-      DEFAULT_MOVE_DURATION = 0.25 # 250 milliseconds
+      @default_move_duration = 0.25 # 250 milliseconds
+      attr_accessor :default_move_duration
 
       #
       # Presses (without releasing) at the current location of the PointerInput device. This is equivalent to:
@@ -98,7 +99,7 @@ module Selenium
           left = 0
           top = 0
         end
-        pointer.create_pointer_move(duration: DEFAULT_MOVE_DURATION,
+        pointer.create_pointer_move(duration: default_move_duration,
                                     x: left,
                                     y: top,
                                     element: element)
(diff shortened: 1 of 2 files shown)


rb/lib/selenium/webdriver/common/interactions/pointer_actions.rb
```diff
@@ -18,7 +18,8 @@
 module Selenium
   module WebDriver
     module PointerActions
-      DEFAULT_MOVE_DURATION = 0.25 # 250 milliseconds
+      @default_move_duration = 0.25  # 250 milliseconds
```
luke-hill: if you want to remove the constant, the best way of defining it will be as a private memoized method

```
private

def default_move_duration
  @default_move_duration ||= 0.25
end
```

titusfortner: The reason for this PR is that the constant is being overridden in the Appium library, so this would change to specifically allow a writer.

p0deje:
Looks good, but please fix unit tests and RuboCop:

```
./go //rb:unit-test
./go //rb:lint
```

rb/lib/selenium/webdriver/common/interactions/pointer_actions.rb
luke-hill: That's fair, I still wouldn't have procedurally defined instance vars, but each to their own.

titusfortner: There's no reason to wrap the default in a method. Leverage the accessor.

rb/spec/unit/selenium/webdriver/common/interactions/pointer_actions_spec.rb
```diff
@@ -26,7 +26,7 @@ module WebDriver
       let(:builder) { W3CActionBuilder.new(bridge, mouse, keyboard) }
       let(:element) { Element.new(bridge, 'element') }
       let(:element2) { Element.new(bridge, 'element2') }
-      let(:duration) { PointerActions::DEFAULT_MOVE_DURATION }
+      let(:duration) { PointerActions::default_move_duration }
```
titusfortner: since this is no longer a constant, it needs to be referenced as a method on the class that includes the `PointerActions` module. In this case, `builder`. Update this to be:
```
      let(:duration) { builder.default_move_duration }
```

prakharrr:
lint and unit test works now.

PR #6209, opened 2018-07-26
Disable screenshot tests for Edge

InstyleVII:

Edge (like Chrome/Firefox/IE) only takes a screenshot of the viewport and not the entire page which causes these 5 tests to fail. Added ignores for Edge for these tests.


jimevans:
Merged as e910737. Thank you for the contribution!

PR #6214, opened 2018-07-27
Skip keyboard tests in Edge

instylevii:

It looks like these tests conflict with the spec per conversation with our dev. We match Firefox here. May need to change the spec if we want this to work but for now skipping. According to the spec, when you do a keyDown action for a character, we should fire a DOM event for that exact character, regardless of the shift key state, so you should be getting "ab".

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6214)


jimevans:
I'm holding off on merging this one, because the language of the spec is ambiguous. Once we have a consensus on this, we can make a decision on the tests.

diemol:
Cannot reopen this one because the source repo where this PR was sent from has been deleted.

PR #6220, opened 2018-07-29
Add ability to quiesce node

### LINKED ISSUE #2070: Feature Request - Node : Prepare to shutdown / Accept no more sessions
Currently if I call http://node:port/selenium-server/driver/?cmd=shutDownSeleniumServer the node shuts down. currently running connections are killed etc.

I'd like to request a feature to instruct the node to not accept any new sessions. That way existing sessions are able to complete and the node in question can be removed from the grid with minimal impact. 
For extra functionality there should be a "Return To Service" function. 
## Background -

On a busy grid it's hard to decommission a node without impacting some tests.
## Expected Behavior -

http://node:port/selenium-server/driver/?cmd=prepareToShutdown
New sessions are rejected.

krmahadevan:
Closes #2070

This PR provides the following:

1. Expose an end-point on the Hub which lets a user
quiesce a node. [ Quiesce = don’t route any more new
session requests to a node and let the in-flight
sessions run to completion ] via the URL 

`/lifecycle-manager?action=quiescenode&proxyid=<ProxyId>`

2. Add an optional argument to the end-point in (1)
to also send a recycle signal to the node [ This would
ensure that the Grid automatically shuts down a 
quiesced node after the in-flight sessions have run 
to completion ] via the URL 
`/lifecycle-manager?action=quiescenode&proxyid=<ProxyId>&autoshutdown=true`

3. Updated the default hub console to also indicate 
if a particular node is “quiesced” or not.

4. If the user just uses (1) to quiesce a node, then 
they need to shutdown the node on their own. So 
exposed a new URL on the node to support this via
the URL` /lifecycle-manager?action=shutdown`

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6220)

### DIFF
--- a/java/server/src/org/openqa/grid/internal/BaseRemoteProxy.java
+++ b/java/server/src/org/openqa/grid/internal/BaseRemoteProxy.java
@@ -73,6 +73,8 @@ public class BaseRemoteProxy implements RemoteProxy {
   private volatile boolean stop = false;
   private CleanUpThread cleanUpThread;
 
+  private boolean underMaintenance = false;
+
   public List<TestSlot> getTestSlots() {
     return testSlots;
   }
@@ -476,5 +478,15 @@ public class BaseRemoteProxy implements RemoteProxy {
       last = Math.max(last, slot.getLastSessionStart());
     }
     return last;
+  }
+
+  @Override
+  public void quiesceNode() {
+    this.underMaintenance = true;
+  }
+
+  @Override
+  public boolean isNodeQuiesced() {
+    return underMaintenance;
   }
 }
--- a/java/server/src/org/openqa/grid/internal/DefaultGridRegistry.java
+++ b/java/server/src/org/openqa/grid/internal/DefaultGridRegistry.java
@@ -247,6 +247,9 @@ public class DefaultGridRegistry extends BaseGridRegistry implements GridRegistr
   }
 
   private boolean takeRequestHandler(RequestHandler handler) {
+    if (!proxies.hasAtleastOneActiveNode()) {
+      return false;
+    }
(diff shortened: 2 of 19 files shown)


krmahadevan:
ping @barancev @shs96c - Please help take a look when you get some time.

krmahadevan:
@manoj9788  @barancev @shs96c - Do you think its worthwhile if I go back to fixing these conflicts and bring this PR back to a mergeable state ? I know that the Grid4 architecture is going through a revamp. So just wanted to check with you folks to see if this still made sense. Please let me know.

diemol:
Hi @krmahadevan,

I'm sorry no one replied to this PR earlier... 

About your question... Not really, there are no plans to release the Grid again under the 3.x series, and even if there is a release, it will be for a critical bug but not to add new features. What we could do is to add a new issue describing this feature and then we add it to the [Selenium 4 project](https://github.com/SeleniumHQ/selenium/projects/2). So when the time comes, we implement it there. Does that make sense?

dylanlive:
Personally I'm looking forward to this functionality, so if it can land in Se4, that would be great!

krmahadevan:
@diemol - Yes. That makes sense. All said and done, can we please have #2070  be moved into the [Selenium 4](https://github.com/SeleniumHQ/selenium/projects/2) project so that it is tracked there ?

diemol:
Done, thanks!

virenv:
@krmahadevan : I know this is like one of the todo in selenium 4 pipeline. I just took a look at the PR to understand it. I hope you dont mind the questions I have asked after almost after a year of your PR.

PR #6231, opened 2018-07-31
Fixed null pointer exception when managed element is not found

MikeBluebird:

Hi,
First public pull request so please let me know if/what I have done wrong.

Basically I was getting an unknown crash in some of my tests, and took it upon myself to track it down. Debugging the running test, showed the line following my change was throwing an exception due to null being passed it. To me it makes sense that if the error code coming out of the function is not success, we should be returning that error code there and then rather than trying to continue and causing a crash.

Changes:
The command_executor has a chance to return an error code of ENOSUCHELEMENT. This leaves the element_wrapper object equal to null. Which ends up throwing an exception on the next like of code when it tries to use the element. This in turn causes a crash in IEDriver and it shuts down.


jimevans:
Merged as bdc4787. Thank you for your contribution!

PR #6240, opened 2018-08-02
Disable overlapping tests

InstyleVII:

Turns out we function the same way (per spec) as Firefox/Chrome for these tests by checking that an element is not obscured before sending click.


jimevans:
Merged as 757c52d. Thank you for your contribution!

PR #6241, opened 2018-08-02
Disable alert tests for Edge

InstyleVII:

Spoke with Brandon and same thing here, we don't fire alerts here so these should be ignored.


jimevans:
Can you check on this? I believe I've enabled this test with an already-in-progress commit that was committed earlier.

InstyleVII:
This is covered by commit c2dd6739ed0b67a9ae9f151e6c88417f9ac9bcdd

PR #6242, opened 2018-08-02
Enable unexpected alert tests in Edge

InstyleVII:

These tests should now pass in Edge in the public insiders flight that just went out so enabling them.


jimevans:
Can you check on this? I believe I've enabled this test with an already-in-progress commit that was committed earlier.

InstyleVII:
Looks like this was covered under commit c2dd6739ed0b67a9ae9f151e6c88417f9ac9bcdd

PR #6245, opened 2018-08-03
Fix Locatable interface import

baflQA:

A lot of the classes was still using deprecated Locatable interface from the `org.openqa.selenium.interactions.internal.Locatable` package. 
I replaced the imports to `org.openqa.selenium.interactions.Locatable` package.

With this commit "java.lang.ClassCastException: wrapper.Element cannot be cast to org.openqa.selenium.interactions.internal.Locatable" exception should be fixed.

### DIFF
--- a/java/client/src/org/openqa/selenium/interactions/Actions.java
+++ b/java/client/src/org/openqa/selenium/interactions/Actions.java
@@ -17,28 +17,21 @@
 
 package org.openqa.selenium.interactions;
 
-import static org.openqa.selenium.interactions.PointerInput.Kind.MOUSE;
-import static org.openqa.selenium.interactions.PointerInput.MouseButton.LEFT;
-import static org.openqa.selenium.interactions.PointerInput.MouseButton.RIGHT;
-
 import org.openqa.selenium.Keys;
 import org.openqa.selenium.UnsupportedCommandException;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.interactions.PointerInput.Origin;
-import org.openqa.selenium.interactions.internal.Locatable;
 import org.openqa.selenium.interactions.internal.MouseAction.Button;
 
 import java.time.Duration;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedHashMap;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
+import java.util.*;
 import java.util.function.IntConsumer;
 import java.util.logging.Logger;
+
+import static org.openqa.selenium.interactions.PointerInput.Kind.MOUSE;
+import static org.openqa.selenium.interactions.PointerInput.MouseButton.LEFT;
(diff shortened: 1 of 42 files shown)


shs96c:
We had to leave the old imports in place so that people could carry on using the old interface. In the next release, the plan is to delete the old version entirely. So this change is good :)

baflQA:
Hi @shs96c. 
So, for the Pi version, I should stick to the deprecated interfaces, am I correct?

barancev:
These interfaces (including Locatable) were deprecated more that a year ago (version 3.6 iirc). And we always recommend to use Actions instead of instantiatind action objects directly. So I suppose we can safely remove them now.

baflQA:
The thing is in my case, I was getting the exception in the Actions class.
`new Actions(driver)
				.moveToElement(element, 0, (element.getSize().height / 2) - 1)
				.doubleClick()
				.perform();`

I use the custom wrapper Element class, which implements a new Location interface (I wanted this class to implement the new one at least ;) ). So is of the `org.openqa.selenium.interactions.Locatable` type. 
In the Actions class, e.g. in the line 405 there is a casting to the `org.openqa.selenium.interactions.internal.Locatable` interface type. 
I see that the RemoteWebElement still implements the old Locatable. 
So, if an old Locatable was deprecated, and a replacement was provided, You should use it internally IMO. 
Now I'm kindly asked to drop the obsolete interfaces, and move on to the newer ones, but it breaks my tests :/
Maybe at least consider annotating the new Locatable with some @Beta annotation?

barancev:
@shs96c If we change all uses of deprecated inteface to the new one it will not hurt people who continues to use the old one. They will be able to pass their oldschool objects to the methods/constructors that require new interface. I think this PR is safe.

groov1kk:
Hi there !

I have the same problem as @baflQA has. I think the problem is in the wrong inheritance. Right now old Locatable interface is inherited from the new one. That's mean that the new Locatable interface is not polymorphic. We cannot use it in the same methods as we used the previous one.

So, how about to change inheritance structure - inherit **new** Locatable from **old** Locatabe? In this case we do not need to change a lot of imports and this won't affect a lot of legacy code.

groov1kk:
Hi there !

I have the same problem as @baflQA have. But I see a another way to solve this problem, different than @baflQA suggested:  how about to inherit a new Locatable interface (from package org.openqa.selenium.interactions) from the deprecated interface (from package org.openqa.selenium.internal) ? 
This action will make a new Locatable interface polymorphic and this change won't affect a lot of legacy code.

baflQA:
Ahoy @shs96c @barancev . 
I see that with this commit [20e8ee0e09d53b1d340d6c475d9ac6a68ec8060a](https://github.com/SeleniumHQ/selenium/commit/20e8ee0e09d53b1d340d6c475d9ac6a68ec8060a)
the issue is resolved. With the best solution ;) 
Any plans to ship it as a hotfix, or as 3.15 (or 3.141 ;) )?

barancev:
Yes, @shs96c fixed the issue and we can close this PR. Unfortunately there is no forecast on the next release date. Soon, as usually :)

PR #6269, opened 2018-08-10
Enable and skip tests in Edge

InstyleVII:

Currently Edge doesn't support insecure SSL certificates so a test validating that has been disabled. Likewise we now pass a page loading test that we didn't so enabled that.


jimevans:
Merged as 8a8af77. Thank you for your contribution!

PR #6270, opened 2018-08-10
Disabling test in Edge

InstyleVII:

Looks like this isn't in the spec and is disabled for all other browsers so disabling it for Edge as well.


jimevans:
Merged as 2a54bd6. Thank you for the contribution!

PR #6271, opened 2018-08-10
Disabling insercure cert cookie tests

InstyleVII:

Disabling these tests as they require supporting the Insecure Certificate capability which Edge doesn't support currently. Once we do support it we should enable these tests to run/validate the scenario.


jimevans:
Merged as ab2b3e9. Thank you for the contribution1

Issue #6295, opened 2018-08-16
Let ActionChains.pause allow pausing split seconds not just multiples of seconds


mei-tech:
## Meta -
OS:  
> Ubuntu 18.04

Selenium Version:  
> 3.14.0

Browser:  
> Firefox 61.0.1

## Expected Behavior -
`ActionChains.(driver).pause(1.2).perform()`
should accept floating numbers, not just integers, or directly ms as integer
All you do is in:
https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/common/actions/interaction.py
Change
`output["duration"] = self.duration * 1000`
to, e.g.
`output["duration"] = int(self.duration * 1000)`

## Actual Behavior -
`ActionChains.(driver).pause(1.2).perform()`

leads to:
> Traceback (most recent call last):
>   File "/home/andreas/selenium-api/app/main.py", line 29, in <module>
>     ActionChains(driver).pause(1.2).perform()
>   File "/home/andreas/selenium-api/venv/lib/python3.6/site-packages/selenium/webdriver/common/action_chains.py", line 80, in perform
>     self.w3c_actions.perform()
>   File "/home/andreas/selenium-api/venv/lib/python3.6/site-packages/selenium/webdriver/common/actions/action_builder.py", line 76, in perform
>     self.driver.execute(Command.W3C_ACTIONS, enc)
>   File "/home/andreas/selenium-api/venv/lib/python3.6/site-packages/selenium/webdriver/remote/webdriver.py", line 320, in execute
>     self.error_handler.check_response(response)
>   File "/home/andreas/selenium-api/venv/lib/python3.6/site-packages/selenium/webdriver/remote/errorhandler.py", line 242, in check_response
>     raise exception_class(message, screen, stacktrace)
> selenium.common.exceptions.InvalidArgumentException: Message: Parameter 'duration' was not a positive integer/errorhandler.py", line 242, in check_response raise exception_class(message, screen, stacktrace)
>selenium.common.exceptions.InvalidArgumentException: Message: Parameter 'duration' was not a positive integer
## Steps to reproduce -

all you have do is:
```
from selenium import webdriver
from selenium.webdriver import ActionChains

profile = webdriver.FirefoxProfile()
driver = webdriver.Firefox(firefox_profile=profile, executable_path="./bins/geckodriver-v0.21.0-linux64/geckodriver")
ActionChains(driver).pause(1.2).perform()
```

Herst:
[...]

PR #6300, opened 2018-08-17
[py] Rounded pause duration

Dakkaron:
WebDriver standard expects "duration" parameter of "pause" action to be integer. ChromeDriver does not care, but GeckoDriver is very strict on this. See https://github.com/mozilla/geckodriver/issues/1355

### DIFF
--- a/py/selenium/webdriver/common/actions/pointer_input.py
+++ b/py/selenium/webdriver/common/actions/pointer_input.py
@@ -54,7 +54,7 @@ class PointerInput(InputDevice):
         self.add_action({"type": "pointerCancel"})
 
     def create_pause(self, pause_duration):
-        self.add_action({"type": "pause", "duration": int(round(pause_duration * 1000))})
+        self.add_action({"type": "pause", "duration": int(pause_duration * 1000)})
 
     def encode(self):
         return {"type": self.type,


cgoldberg:
@Dakkaron 
Do you have a good reason for why this should be changed?

currently, passing a non-integer will raise a `InvalidArgumentException` with a clear message that  'duration' was not a positive integer.

Following the principles of "exclicit is better than implicit" and "errors should never pass silently", I feel like that behavior is more correct than implicitly rounding duration to an integer.

Dakkaron:
In Selenium for Python the `pause` action takes seconds as parameter, while the WebDriver protocol takes milliseconds. To convert that, the Selenium for Python implementation multiplies the input by 1000.

If you want to wait for fractions of seconds, this is currently not possible (as laid out in the linked thread), because what happens there is you pass 0.5 seconds, these get multiplied to 500.0 milliseconds. The problem here is that when you multiply an integer by a float you get a float, not an integer, so this raises said exception.

There is no way to pass fractions of a second as an integer second value.

The ChromeDriver is very lax on the standard and allows floats there, but GeckoDriver is very strict and forbids floats here.

Because of this, it is absolutely necessary to allow floats and cast them to int, except you want to rework Selenium so that `pause` takes milliseconds (as the WebDriver protocol do), which would break existing tests.

Dakkaron:
This problem also has been reported here https://github.com/SeleniumHQ/selenium/issues/6295.

lmtierney:
@Dakkaron thanks for the explanation. We definitely want to allow fractions of a second. I don't think we need the `round()` but am indifferent on it. The difference would be half a millisecond

Dakkaron:
@lmtierney Thanks for the answer. Should I remove the round()?

lmtierney:
@Dakkaron yeah go ahead and remove the round and I'll get it merged

Dakkaron:
@lmtierney Did it, can be merged now. Thank you!

lmtierney:
Merged, thank you for your contribution!

Dakkaron:
Thank you!

PR #6305, opened 2018-08-19
[Grid] Remove unused client timeout logic

dylanlive:
With commit d06e54aec6c41ac7f3503c087472690961cd9873, the browserTimeout logic is no longer used when creating the http client. This removes that logic since it's unused.

No functional changes should take place with this change - it just makes the code clearer to follow (I came across this when trying to find what determined the http client timeout and it distracted me).


barancev:
Merget as 213e229cebbbc162d88fe839aec8998906241475. Thank you for contributing to the project!

But, yes, may be we should implement timeout in some other way.

PR #6324, opened 2018-08-23
Make ExpectedConditions private helpers protected

gwpantazes:
Allow easier extension of the ExpectedConditions class by making private helper methods protected.

If a user would like to extend ExpectedConditions and write their own custom ExpectedConditions, they can benefit from writing the ExpectedCondition methods in the same style as the original ExpectedConditions, which includes the use of the helper methods for stylistic consistency.


gwpantazes:
Actually, I completely forgot about the private constructor. It looks like this was a bad idea (oops). Closing this PR, I'll just continue to make my ExtendedConditions like normal without extending.

PR #6355, opened 2018-08-31
w3c issue with `def rect(self):`

VladimirPodolian:

### DIFF
--- a/py/selenium/webdriver/remote/webelement.py
+++ b/py/selenium/webdriver/remote/webelement.py
@@ -546,7 +546,6 @@ class WebElement(object):
             rect.update(self.location)
             return rect
 
-
     @property
     def screenshot_as_base64(self):
         """


py/selenium/webdriver/remote/webelement.py
```diff
@@ -539,7 +539,10 @@ def location(self):
     @property
     def rect(self):
         """A dictionary with the size and location of the element."""
-        return self._execute(Command.GET_ELEMENT_RECT)['value']
+        if self._w3c:
+            return self._execute(Command.GET_ELEMENT_RECT)['value']
+        else:
+            return self.size, self.location
```
lmtierney: This would return a tuple of two dictionaries. You need to combine them and return a dictionary

VladimirPodolyan: Done, but it might be best not to use `self.size` and `self.location` ? 
I mean may be better to reuse `_execute(Command.GET_ELEMENT_SIZE/LOCATION)`

VladimirPodolyan:
@lmtierney could you please review?

py/selenium/webdriver/remote/webelement.py
```diff
@@ -542,7 +542,12 @@ def rect(self):
         if self._w3c:
             return self._execute(Command.GET_ELEMENT_RECT)['value']
         else:
-            return dict(list(self.size.items()) + list(self.location.items()))
+            size = self._execute(Command.GET_ELEMENT_SIZE)['value']
+            loc = self._execute(Command.GET_ELEMENT_LOCATION)['value']
+
+            rect = {"height": size["height"], "width": size["width"],
+                    "x": round(loc['x']), "y": round(loc['y'])}
+        return rect
```
lmtierney: Is there a reason you don't just use `self.size` and `self.location` ? The ending return is outside of the else, which will cause an error if `self._w3c` is true.

Why not something like this?
```
rect = self.size.copy()
rect.upate(self.location)
return rect
```

VladimirPodolyan: Nested "if" in these methods made me think about a better solution. Your option will work well too

VladimirPodolyan: But without typo in `rect.upate(self.location)` C:

VladimirPodolyan: @lmtierney  I made a changes using your example

py/selenium/webdriver/remote/webelement.py
```diff
@@ -539,7 +539,13 @@ def location(self):
...
     def rect(self):
         """A dictionary with the size and location of the element."""
-        return self._execute(Command.GET_ELEMENT_RECT)['value']
+        if self._w3c:
+            return self._execute(Command.GET_ELEMENT_RECT)['value']
+        else:
+            rect = self.size.copy()
+            rect.update(self.location)
+            return rect
+
```
lmtierney: flake8 is going to fail on this having 2 lines

VladimirPodolyan: Fixed

PR #6375, opened 2018-09-06
[py] Add WPEWebKit driver

carlosgcampos:
For the WPE port of WebKit

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6375)


lmtierney:
Are we going to keep adding lesser-used webdrivers to the selenium code base? A quick look at the code looks like you could easily just subclass the WebKitGTK Options and pass it directly to the WebKitGTK webdriver.

carlosgcampos:
Well, WPE is not used for general purpose desktop browsers, but it's used in *a lot of* embedded devices. The driver is similar to the WebKitGTK+ one, but all driver impls are quite similar to each other. Problem of subclassing is that options of WebKitGTK+ might not make sense in WPE (and the other way around). like for example the overlay scrollbars option. Anyway, I don't plan to add more drivers :-)

Issue #6311, opened 2018-08-20
[Grid] Clicking "view config" on Console jumps to top of page


dylanlive:
## Meta -
OS:  
macOS 10.12.6
Selenium Version:  
3.13.0
Browser:  
Chrome

Browser Version:  
latest

## Expected Behavior -
Clicking "view config" on the Grid Console should not reload the page and jump to the top

## Actual Behavior -
Clicking "view config" on the Grid Console reloads the page and jumps to the top

## Steps to reproduce -
1) Launch hub
2) Connect many nodes to the hub
3) Click "view config" - you'll be jumped to the top

dylanlive:
@shs96c @barancev I'd like to open a PR for this; however, I know se4 will redesign the console.

I'm assuming that's still a-ways away, so would you accept a PR for this minor fix?

It looks like it'll be more work than my previous PR #5546 as the logic displays the config based on the request parameters- would we want to keep support for Non-Javascript enabled browsers?

I'm assuming yes and can work to support both (but it'll be a bit more work).

barancev:
I'm not sure we should care about non-JS enabled browsers. Why should we? What's the target audience?

dylanlive:
@barancev Awesome, that would make the PR easier to just display the Config using JavaScript.

I’d like to open a PR so that clicking View Config won’t jump you back to the very top of the page (annoying with many nodes expanding the page length). 

I’m hoping to get acknowledgment it’ll be accepted before putting in the work.

diemol:
Hi @dylanlive,

You are right, it is an annoying issue, if you send the PR, I'd be happy to review it and merge it.

dylanlive:
@diemol Thanks! Mostly finished, just polishing. Hopefully should have some time this weekend to wrap it up

dylanlive:
Opened https://github.com/SeleniumHQ/selenium/pull/6386 to fix

diemol:
PR #6386 merged via https://github.com/SeleniumHQ/selenium/commit/4e6f0117d872f7b132143903be7cb2b744739161, thanks @dylanlive!

PR #6386, opened 2018-09-09
[Grid] Move console's View Config to Javascript

dylanlive:
Fixes https://github.com/SeleniumHQ/selenium/issues/6311

This PR updates the Console Servlet's "View Config" logic to use Javascript, bringing the benefit of toggling Config Visibility without causing a page reload & jump to the top (which gets annoying when many nodes are connected).

Other changes:
- Removes `?configDebug` parameter requirement
- Hides Verbose Config by Default
- Adds a link to Show Verbose Config
- Moves Verbose Config builder into its own method
- Adds padding beneath the config container (so that the Show Config link doesn't straddle the bottom of the browser window)
-----

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6386)

### DIFF
--- a/java/server/src/org/openqa/grid/images/consoleservlet.css
+++ b/java/server/src/org/openqa/grid/images/consoleservlet.css
@@ -208,6 +208,7 @@ ul a {
 
 #hub-config-container {
   min-width:55%;
+  padding-bottom: 10px;
 }
 
 #hub-config-content {


dylanlive:
fyi, for "View Verbose" I did not write the same logic to re-hide it. I can do so if anyone feels it would be preferred (I didn't really like the way it looked with two links next to each other).

diemol:
Merged as https://github.com/SeleniumHQ/selenium/commit/4e6f0117d872f7b132143903be7cb2b744739161, thanks for your contribution!

dylanlive:
Thanks, @diemol!

Issue #6396, opened 2018-09-11
The getAttribute atom treats `draggable` as a boolean attribute when it is actually enumerated


twalpole:
## Meta -
OS:  OSX
Selenium Version:  Ruby 3.14.0
Browser:  Firefox & Chrome

The getAttribute atom treats the `draggable` attribute as boolean, when it is actually an enumerated attribute - https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable.  This makes it impossible to read the actual value of the attribute since it will always return `true` if the attribute exists. I experienced this using the ruby bindings via selenium-webdriver 3.14.0 but it will affect any binding using the getAttribute atom.

barancev:
You should log this issue to the spec first:
https://w3c.github.io/webdriver/

barancev:
Oh, sorry, ingore my previous comment, the spec does not list boolean attributes explicitly, it just refers HTML specification.

barancev:
There is one more interesting question about enumerated attributes. What should we return -- the actual attribute value or its "state"? Each enumerated attribute has two special states: 1) missing value default (if the attribute is not specified) and 2) invalid value default (if the value is not among the list of valid values).

barancev:
Merged as 7c17f61c1ac35f0dbbce207f95186b483a27f33f

Unfortunately it will not make all drivers immediately compliant. A driver devs should update the atoms used in the driver.

PR #6398, opened 2018-09-12
Remove 2 non-boolean and add a few missing boolean attributes

twalpole:

This removes `draggable` and `spellcheck` from the list of boolean attributes because they are actually enumerated attributes.  Also adds in a few missing boolean attributes. 

Should fix Issue #6396


barancev:
Merged as 7c17f61

Unfortunately it will not make all drivers immediately compliant. A driver devs should update the atoms used in the driver.

PR #6401, opened 2018-09-12
Safari-specific check of returned 'browserName' capability should be case-insensitive

burg:
safaridriver now returns the browser name as reported by the user agent.
Don't assume a specific casing; it's capitalized in Safari 12 and later.


burg:
I used the PR interface wrong. Trying from a branch now..

PR #6421, opened 2018-09-17
Update ClickScrollingTest.java

ArthurGIT2017:
According to documentation at https://developer.mozilla.org/en-US/docs/Web/API/Window/pageYOffset window.pageYOffset may not return integer value. With the following changes in below we are checking the type of returned object and assigning respectively the value of 'yOffset'


ArthurGIT2017:
On Windows platform window.pageYOffset returns double value and on Linux platform it returns long value, so this patch fixes this issue.

barancev:
Merged as 1d85e230366fbc275877065b4b60f29ae052deaa, thank you for contributing to the project!

PR #6427, opened 2018-09-19
Clean up remote connection to better use urllib3.

kamaroyl:
* Poolmanager should not be created on every call, move to a class
variable for keep-alive and not keep-alive.
* Remove some extra code that attempts to check for attributes to keep backwards compatability;
  HTTPResponse from urllib.response does that work no need to duplicate.
* add a close method to clean up the connection pool. Refactored a large
  portion of code into its own method to clean up init function

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6427)

### DIFF
--- a/py/selenium/webdriver/remote/remote_connection.py
+++ b/py/selenium/webdriver/remote/remote_connection.py
@@ -44,6 +44,24 @@ class RemoteConnection(object):
     https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol"""
 
     _timeout = socket._GLOBAL_DEFAULT_TIMEOUT
+    _retries = 0
+
+    @classmethod
+    def get_retries(cls):
+        """
+        :return:
+            returns the number of retries urllib3 will use when making requests
+        """
+        return cls._retries
+
+    @classmethod
+    def set_retries(cls, retries):
+        """
+        Sets the number of retries to be used in urllib3 when making requests
+        :Args:
+            - retries - number of retries for http/s requests made
+        """
+        cls._retries=retries
 
     @classmethod
     def get_timeout(cls):
@@ -148,13 +166,14 @@ class RemoteConnection(object):
                                    .format(parsed_url.hostname))
         return remote_server_address
 
-    def __init__(self, remote_server_addr, keep_alive=False, retry_count=3, resolve_ip=True):
+    def __init__(self, remote_server_addr, keep_alive=False, retries=0, resolve_ip=True):
         # Attempt to resolve the hostname and get an IP address.
+
         self.keep_alive = keep_alive
         self._url = self.parse_remote_server(remote_server_addr, resolve_ip)
 
         # Keep the name _conn in case someone has been clearing the pool externally with private variable
-        self._conn = PoolManager(timeout=self._timeout, retries=retry_count)
+        self._conn = PoolManager(timeout=self._timeout, retries=retries)
 
         self._commands = {
             Command.STATUS: ('GET', '/status'),
@@ -413,7 +432,8 @@ class RemoteConnection(object):
         if body and method != 'POST' and method != 'PUT':
             body = None
 
-        resp = self._conn.request(method, url, body=body, headers=headers)
+        resp = self._conn.request(method, url, body=body, headers=headers,
+                                  timeout=self._timeout, retries=self._retries)
         LOGGER.debug('Headers: {}'.format(headers))
         status_code = resp.status
 
@@ -449,7 +469,13 @@ class RemoteConnection(object):
         finally:
             LOGGER.debug("Finished Request")
             resp.close()
+            # clear connections - we lose the connections pools,
+            # but we keep the main object around avoiding recreating it.
+            # and we are avoiding gc cleaning up our sockets
+            if not self.keep_alive:
+                self._conn.clear()
 
-    # now can use with closing
+    # now can use with closing!
+    # don't rely on gc to clean things up for you
     def close(self):
         self._conn.clear()
--- a/py/selenium/webdriver/remote/remote_connection.py
+++ b/py/selenium/webdriver/remote/remote_connection.py
@@ -44,7 +44,7 @@ class RemoteConnection(object):
     https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol"""
 
     _timeout = socket._GLOBAL_DEFAULT_TIMEOUT
-    _retries = 0
+    _retries = 0  # urllib3 defaults to 3 retries, we are essentially disabling retries
 
     @classmethod
     def get_retries(cls):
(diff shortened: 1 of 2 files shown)


py/selenium/webdriver/remote/remote_connection.py
```diff
@@ -103,41 +103,58 @@ def get_remote_connection_headers(cls, parsed_url, keep_alive=False):
...
+
+                    url_parts = (parsed_url.scheme, netloc, parsed_url.path,
+                                 parsed_url.params, parsed_url.query, parsed_url.fragment)
+                    remote_server_address = parse.urlunparse(url_parts)
+                else:
+                    LOGGER.warning('Could not get IP address for host: {}'
+                                   .format(parsed_url.hostname))
+        return remote_server_address
+
+    def __init__(self, remote_server_addr, keep_alive=False, retry_count=3, resolve_ip=True):
```
Dude-X: I would set the retry_count to 0. Even though urllib3 only retries on connections and not on completed connections, we can't rely on this fact because a driver or remote end may repeat a command.

Dude-X: I would make retry a class variable to set. Most users should rely on 0 retries.

AutomatedTester (changes requested):
> *[Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6427)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @Dude-X and @kamaroyl)

___
*[py/selenium/webdriver/remote/remote_connection.py, line 157 at r1](https://reviewable.io/reviews/seleniumhq/selenium/6427#-LMmFWiN3T1VjX_aMWhV:-LMmFWiN3T1VjX_aMWhW:b-ao0my) ([raw file](https://github.com/seleniumhq/selenium/blob/0d6b8d2533766b90332046f4c4cb36f479ccf16f/py/selenium/webdriver/remote/remote_connection.py#L157)):*
> ```Python
> 
>         # Keep the name _conn in case someone has been clearing the pool externally with private variable
>         self._conn = PoolManager(timeout=self._timeout, retries=retry_count)
> ```

This makes everything have to use keep-alive where there are users who have wanted to switch it off so this will needed to be added back.

 If people are clear private variables inside an object they use then they are open to having things break
___
*[py/selenium/webdriver/remote/remote_connection.py, line 403 at r1](https://reviewable.io/reviews/seleniumhq/selenium/6427#-LMmGbdlAG5eTMrKMbl2:-LMmGbdlAG5eTMrKMbl3:b-ajb3dp) ([raw file](https://github.com/seleniumhq/selenium/blob/f16c6b2c2e809cb16746d8961717097e0a296005/py/selenium/webdriver/remote/remote_connection.py#L403)):*
> ```Python
>             statuscode = resp.status
>         else:
>             http = urllib3.PoolManager(timeout=self._timeout)
> ```

People not wanting keep alive will have to deal with the potential memory usage here

kamaroyl (commented):
> *[Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6427)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @kamaroyl)

___
*[py/selenium/webdriver/remote/remote_connection.py, line 151 at r1](https://reviewable.io/reviews/seleniumhq/selenium/6427#-LMmBOEVahYBMrWxY3M9-r1-151:-LMmR4R51op3SRst0vFy:btknde4) ([raw file](https://github.com/seleniumhq/selenium/blob/0d6b8d2533766b90332046f4c4cb36f479ccf16f/py/selenium/webdriver/remote/remote_connection.py#L151)):*
I think you are right that retries should be a class variable, However, as documented, retries defaults to 3 for poolmanager. Should we disable retries?

from https://urllib3.readthedocs.io/en/latest/user-guide.html 

Retrying requests
urllib3 can automatically retry idempotent requests. This same mechanism also handles redirects. You can control the retries using the retries parameter to request(). By default, urllib3 will retry requests 3 times and follow up to 3 redirects.

To change the number of retries just specify an integer:

>>> http.requests('GET', 'http://httpbin.org/ip', retries=10)
___
*[py/selenium/webdriver/remote/remote_connection.py, line 157 at r1](https://reviewable.io/reviews/seleniumhq/selenium/6427#-LMmFWiN3T1VjX_aMWhV:-LMnIEOX7I8teCbc7oqk:b-y9tx18) ([raw file](https://github.com/seleniumhq/selenium/blob/0d6b8d2533766b90332046f4c4cb36f479ccf16f/py/selenium/webdriver/remote/remote_connection.py#L157)):*
<details><summary><i>Previously, AutomatedTester (David Burns) wrote…</i></summary><blockquote>

This makes everything have to use keep-alive where there are users who have wanted to switch it off so this will needed to be added back.

 If people are clear private variables inside an object they use then they are open to having things break
</blockquote></details>

Clear cleans up all open sockets and dereferences the connectionpools. It should be exposed so that implementers can clean up after use. If you want we can move this to context management as well
___
*[py/selenium/webdriver/remote/remote_connection.py, line 403 at r1](https://reviewable.io/reviews/seleniumhq/selenium/6427#-LMmGbdlAG5eTMrKMbl2:-LMmQTls4c0Ip72hsC0x:b-u9842n) ([raw file](https://github.com/seleniumhq/selenium/blob/f16c6b2c2e809cb16746d8961717097e0a296005/py/selenium/webdriver/remote/remote_connection.py#L403)):*
<details><summary><i>Previously, AutomatedTester (David Burns) wrote…</i></summary><blockquote>

People not wanting keep alive will have to deal with the potential memory usage here
</blockquote></details>

I'll make changes to avoid keeping a socket alive but keep the pool so multiple uses can avoid having to instantiate a new object, and we can clean up.

kamaroyl (commented):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6427)* status: all files reviewed, 3 unresolved discussions (waiting on @Dude-X, @AutomatedTester, and @kamaroyl)

___
*[py/selenium/webdriver/remote/remote_connection.py, line 403 at r1](https://reviewable.io/reviews/seleniumhq/selenium/6427#-LMmGbdlAG5eTMrKMbl2:-LMrlEUZ3K1d-gJGK54l:b-qbiznw) ([raw file](https://github.com/seleniumhq/selenium/blob/f16c6b2c2e809cb16746d8961717097e0a296005/py/selenium/webdriver/remote/remote_connection.py#L403)):*
<details><summary><i>Previously, kamaroyl (Ben Straub) wrote…</i></summary><blockquote>

I'll make changes to avoid keeping a socket alive but keep the pool so multiple uses can avoid having to instantiate a new object, and we can clean up. 
</blockquote></details>

What are your concerns for memory usage for the pool manager?

kamaroyl (commented):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6427)* status: all files reviewed, 3 unresolved discussions (waiting on @Dude-X and @AutomatedTester)

py/selenium/webdriver/remote/remote_connection.py
```diff
@@ -103,256 +121,59 @@ def get_remote_connection_headers(cls, parsed_url, keep_alive=False):
 
         return headers
 
+    # todo See if this actually does anything - Kamaroyl
```
Dude-X: Remove comment.

kamaroyl:
___
*[py/selenium/webdriver/remote/remote_connection.py, line 124 at r3](https://reviewable.io/reviews/seleniumhq/selenium/6427#-LMmBOEVahYBMrWxY3M9-r3-124:-LMrwVgPEyt9ke_cgC14:brj4q4g) ([raw file](https://github.com/seleniumhq/selenium/blob/fddbc9c91b21382c9b9e1b991777caef518ae7fa/py/selenium/webdriver/remote/remote_connection.py#L124)):*
<details><summary><i>Previously, Dude-X (Isaul Vargas) wrote…</i></summary><blockquote>

Remove comment. 
</blockquote></details>

Can do

py/selenium/webdriver/remote/remote_connection.py
```diff
@@ -441,3 +254,229 @@ def _request(self, method, url, body=None):
         finally:
             LOGGER.debug("Finished Request")
             resp.close()
+            # clear connections - we lose the connections pools,
+            # but we keep the main object around avoiding recreating it.
+            # and we are avoiding gc cleaning up our sockets
+            if not self.keep_alive:
+                self._conn.clear()
+
+    # now can use with closing!
```
Dude-X: I would be more specific: This allows the use of  contextlib.closing.

py/selenium/webdriver/remote/remote_connection.py
```diff
@@ -441,3 +254,229 @@ def _request(self, method, url, body=None):
...
+    # don't rely on gc to clean things up for you
+    def close(self):
+        self._conn.clear()
+
+    _commands = {
```
isaulv, line 268: This seems like a good move, moving the definition to the class body so it's define once and not on every function invocation.

kamaroyl (commented):
> *[Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6427)* status: 1 of 2 files reviewed, 6 unresolved discussions (waiting on @kamaroyl, @Dude-X, and @AutomatedTester)

___
*[py/selenium/webdriver/remote/remote_connection.py, line 263 at r3](https://reviewable.io/reviews/seleniumhq/selenium/6427#-LMmBOEVahYBMrWxY3M9-r3-263:-LMrx9xz0Rc3JPUqDKH2:b-837i48) ([raw file](https://github.com/seleniumhq/selenium/blob/fddbc9c91b21382c9b9e1b991777caef518ae7fa/py/selenium/webdriver/remote/remote_connection.py#L263)):*
<details><summary><i>Previously, Dude-X (Isaul Vargas) wrote…</i></summary><blockquote>

I would be more specific: This allows the use of  contextlib.closing.
</blockquote></details>

Okay!

kamaroyl (commented):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6427)* status: all files reviewed, 6 unresolved discussions (waiting on @Dude-X and @AutomatedTester)

AutomatedTester:
A lot of this code has under gone changes and I believe some of the improvements are already in. Sorry for not getting to this PR sooner but if there are other improvements we can do please submit a new PR

PR #6432, opened 2018-09-20
Add SWITCH_TO_FRAME_PARENT to command map

astappiev:
When I used `targetLocator.paretFrame()` method of `selenium-webdirver@4.0.0-alpha.1` (npm package) with Chrome driver I got next error:
```
UnknownCommandError: Unrecognized command: switchToFrameParent
    at buildRequest (...\selenium-webdriver\lib\http.js:376:9)
    at Executor.execute (...\selenium-webdriver\lib\http.js:456:19)
    at Driver.execute (...\selenium-webdriver\lib\webdriver.js:696:38)
```

Looks like this command is missing in common commands list, this PR adds it.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6432)

### DIFF
--- a/javascript/node/selenium-webdriver/lib/http.js
+++ b/javascript/node/selenium-webdriver/lib/http.js
@@ -214,6 +214,7 @@ const COMMAND_MAP = new Map([
     [cmd.Name.GET_WINDOW_SIZE, get('/session/:sessionId/window/current/size')],
     [cmd.Name.SET_WINDOW_SIZE, post('/session/:sessionId/window/current/size')],
     [cmd.Name.SWITCH_TO_FRAME, post('/session/:sessionId/frame')],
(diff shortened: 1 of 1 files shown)


jleyba:
This command is only defined for the W3C protocol and the command is defined in that map. If chromedriver is W3C compliant now, then that's a separate issue that should be fixed in the client bindings than adding this command to the legacy map.

yonixw:
@jleyba This issue still need attention, but my fix like the one suggested here is only in legacy map. Can you give some pointers what the correct way to fix it and be approved to merged?

corevo:
It was already merged in a previous PR I've sent, we'll need to release a new version.

tourdedave:
It's available now in [4.0.0-alpha.3](https://www.npmjs.com/package/selenium-webdriver/v/4.0.0-alpha.3)

PR #6444, opened 2018-09-25
Selenium 4: Ruby bindings

p0deje:
This PR is opened for the list of changes I'd like to do for Ruby bindings in Selenium 4. Please read below for details and I'll be happy to hear the feedback. Let's have this PR both as To Do list and discussion board + code of what we want to see in Selenium 4. I'll be updating the list throughout the implementation and discussion.

I'll be very happy to hear the feedback from everyone else involved in selenium-webdriver gem development (@titusfortner, @lmtierney, @twalpole). If something is not clear or doesn't make any sense or causes trouble for higher-level libraries (Watir, Capybara) - let's discuss here.

### Removal of OSS dialect

- [x] Remove deprecated `Mouse` class.
- [x] Remove deprecated `Keyboard` class.
- [x] Remove all OSS -> W3C capabilities conversion and aliases. Deprecate what used to work.
- [x] Convert all `W3CSometimg` classes to `Something`. Remove their OSS equivalents.
- [ ] Change API for timeouts.
- [x] Remove protocol handshake.
- [ ] Rename `desired_capabilities` to `capabilities` and provide W3C-compliant API for setting then (`alwaysMatch`, `firstMatch`).

### Cleanup

- [x] Remove Firefox legacy driver support.
- [x] Remove PhantomJS code (or extract to standalone gem) (?).
- [ ] Go through all the driver extensions and drop the ones that no longer make any sense.

### Improvements

- [ ] Implement automatic driver extensions loading based on returned capabilities (it should have been there all the time).
- [x] Implement `Safari::Options` (added in 96998d9af7).
- [x] Implement `Edge::Options` (added in 4b0296f27e).
- [x] Allow to pass `Options` classes to remote instances (added in 92f4ed8183).
- [x] Allow to pass custom `Service` instances to driver in favor of `driver_opts`. Deprecate `driver_opts`.
- [x] ~~Use `element/{element id}/attribute/{name}` to get element instead of atom (confirm with Simon).~~ This is postponed until Se5.
- [ ] Consider adding CDP (#6667)

### Ruby-specific

- [x] ~~Deprecate `Selenium::WebDriver` namespace in favor of `Selenium`.~~ Agree not to do this.
- [x] Require minimum Ruby version to 2.3 (EOL is in March) (https://www.ruby-lang.org/en/downloads/branches/).
- [ ] Address all `C-rb` issues (somehow).
- [ ] Address all `C-rb` pull requests (somehow).
- [ ] Remove JRuby-specific `SocketPoller#listening?` as the bug it works around has been fixed in JRuby 1.6.7 (http://jruby.org/2012/02/22/jruby-1-6-7).
- [x] Remove any custom code that checks for defined `RUBY_ENGINE` as it was only undefined in MRI 1.8 (also search for similar constants).
- [x] Check if we still need custom logic for getting HOME directory on JRuby.
- [x] Remove anything related to IronRuby - it’s not developed anymore.
- [x] Simplify `Platform#null_device` as `File::NULL` was introduced in Ruby 1.9.3 (https://svn.ruby-lang.org/repos/ruby/tags/v1_9_3_0/NEWS).
- [x] Drop all the execution via `$PROGRAM_NAME == __FILE__` - use IRB instead.
- [x] Properly handle Windows/Unix path separators using `File::ALT_SEPARATOR` whenever possible.
- [x] Remove all the compatibility layer for old Ruby in `Logger` .
- [x] Remove `PortProber.random` as it should be avoided (deprecated in 6d3edb5).
- [x] Rework `Wait` to use `Process::CLOCK_MONOTONIC` instead of `Time` as the latter is often monkey-patched (e.g. Timecop).
- [x] ~~Deprecate and extract all custom HTTP clients (curb/persistent)~~ Postponed.
- [ ] Go through YARD docs and add missing.
- [x] Use safe-navigation operator where needed (`&.`)
- [x] Use squiggly heredoc instead of manually de-indented ones
- [x] Check why we need `Safari::Service#binary_path`
- [ ] Check all TODOs

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6444)

### DIFF
--- a/rb/lib/selenium/webdriver/remote/bridge.rb
+++ b/rb/lib/selenium/webdriver/remote/bridge.rb
@@ -441,55 +441,6 @@ module Selenium
                             'if (arguments[0].dispatchEvent(e)) { arguments[0].submit() }', form.as_json)
         end
 
-        def drag_element(element, right_by, down_by)
-          execute :drag_element, {id: element}, {x: right_by, y: down_by}
-        end
-
-        def touch_single_tap(element)
-          execute :touch_single_tap, {}, {element: element}
-        end
-
-        def touch_double_tap(element)
-          execute :touch_double_tap, {}, {element: element}
-        end
-
-        def touch_long_press(element)
-          execute :touch_long_press, {}, {element: element}
-        end
-
-        def touch_down(x, y)
-          execute :touch_down, {}, {x: x, y: y}
-        end
-
-        def touch_up(x, y)
-          execute :touch_up, {}, {x: x, y: y}
-        end
-
-        def touch_move(x, y)
-          execute :touch_move, {}, {x: x, y: y}
-        end
-
-        def touch_scroll(element, x, y)
-          if element
-            execute :touch_scroll, {}, {element: element,
-                                        xoffset: x,
-                                        yoffset: y}
-          else
-            execute :touch_scroll, {}, {xoffset: x, yoffset: y}
-          end
-        end
-
-        def touch_flick(xspeed, yspeed)
-          execute :touch_flick, {}, {xspeed: xspeed, yspeed: yspeed}
-        end
-
-        def touch_element_flick(element, right_by, down_by, speed)
-          execute :touch_flick, {}, {element: element,
-                                     xoffset: right_by,
-                                     yoffset: down_by,
-                                     speed: speed}
-        end
-
         def screen_orientation=(orientation)
           execute :set_screen_orientation, {}, {orientation: orientation}
         end


twalpole:
This all sounds fine to me other than the deprecating of the Selenium::WebDriver namespace - that's going to make exception handling while supporting selenium-webdriver 3 and 4 in Capybara a pain

p0deje:
> This all sounds fine to me other than the deprecating of the Selenium::WebDriver namespace - that's going to make exception handling while supporting selenium-webdriver 3 and 4 in Capybara a pain

Fair enough. We might just not do that or find a way to make it work for Capybara painlessly. @titusfortner proposed this, so maybe he has an opinion on that.

luke-hill:
@p0deje for when you move / CP the commits into here I've tackled the following.

- Remove `Mouse`
- Remove `Keyboard`
(In doing so I removed the legacy ActionBuilder as well)
- Removed a couple of the legacy Touch Driver Extensions (The others I couldn't work out which were consumed)
- Cleared a couple of TODO's up.
- Removed all references to OSS where I could find them deprecated
- Altered all references to existing OSS to a deprecated version (Only exception was the Bridge stuff for Edge/IE)

p0deje:
This is now in master, so no need to maintain PR.

Issue #6431, opened 2018-09-20
Add the option for Safari 12 Webdriver to use '--legacy' argument


charlesmorelle84:
OS:  OSX High Sierra
Selenium Version:  3.14.0
Browser:  Safari 12

The implementation of the Safari 12  WebDriver now supports the W3C endpoints **by default**, which may be causing issues if you're using a Framework like Watir, which aren't compatible yet: https://developer.apple.com/documentation/webkit/macos_webdriver_commands_for_safari_12_and_later

Browser Version:  Version 12.0 (13606.2.11)

charlesmorelle84:
> Usage: safaridriver [options]
	--legacy                  Use the legacy Selenium WebDriver protocol.

gzfrancisco:
Just found this problem on my CI after upgrading the Safari version. Does anyone know how to hot-fixit  it?

charlesmorelle84:
I was able to temporarily fix it by updating this file /Users/[user_name]/.rvm/gems/ruby-X.X.X/gems/selenium-webdriver-X.X.X/lib/selenium/webdriver/safari/service.rb

And adding this line:
![screen shot 2018-09-19 at 17 06 00](https://user-images.githubusercontent.com/22692532/45788288-7018ef80-bc2e-11e8-861a-8024388b0494.png)

lmtierney:
How is Watir not compatible? It should be, please raise an issue if it is not. Selenium should also be compatible, so are you having issues with it as well?

lmtierney:
Closing this as you can already add the `--legacy` flag, see:

```
driver = Selenium::WebDriver.for(:safari, driver_opts: {args: '--legacy'})
```

Or Watir:
```
browser = Watir::Browser.new(:safari, driver_opts: {args: '--legacy'})
```
If you have other issues with Selenium and Safari 12, please open a new issue with the details.

charlesmorelle84:
Thank you @lmtierney .

It's working by declaring it with the args value:
`$driver = Watir::Browser.new(:safari, driver_opts: {args: '--legacy'})`

gzfrancisco:
I confirm, the solution works with the Selenium WebDriver. Thanks.

mjdease:
@lmtierney how can this flag be set using the node selenium-webdriver package (v3.6)?

lmtierney:
[...]

PR #6459, opened 2018-09-28
[Py] Safari 12 webdriver legacy arg

keenangraham:
This updates the Python Safari webdriver and service modules to allow specification of command line arguments, specifically so the `--legacy` flag can be passed to the safaridriver. Not sure if Safari 12 will be supported by Selenium, but we've found this to be a useful workaround in the meantime.

This has been discussed here: 
https://github.com/SeleniumHQ/selenium/issues/6431

And here:
https://stackoverflow.com/questions/52432636/session-not-created-selenium-webdriver-when-using-safari-12

But none of the workarounds seem to be Python specific.

The change follows the convention found in other browser modules (e.g. Chrome https://github.com/SeleniumHQ/selenium/blob/bc10b674bf0f0b0dadfde9435584813224d02618/py/selenium/webdriver/chrome/service.py#L37), and adds some test that the subprocess call includes the legacy flag or not.

Allowing any argument to be passed is perhaps not restrictive enough, and there is nothing specific catching the error if someone specifies an invalid argument. An alternative might be to use a use_legacy=True/False flag. Tests could also be more comprehensive.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6459)

### DIFF
--- a/py/selenium/webdriver/safari/webdriver.py
+++ b/py/selenium/webdriver/safari/webdriver.py
@@ -52,7 +52,7 @@ class WebDriver(RemoteWebDriver):
         """
 
         self._reuse_service = reuse_service
-        self.service = Service(executable_path, port=port, service_args=service_args, quiet=quiet)
+        self.service = Service(executable_path, port=port, quiet=quiet, service_args=service_args)
         if not reuse_service:
             self.service.start()
 


keenangraham:
Reordered.

lmtierney:
Merged, thank you!

PR #6463, opened 2018-09-30
Add python client support for webextensions in firefox [rebased upon latest master and with fixed tests]

motin:
This is a rebased and improved upon version of https://github.com/SeleniumHQ/selenium/pull/5069, addressing the concerns raised by the reviewer. 

Fixes the python part of #4093

Main contributors of this PR:
* @jonsecchis for the initial PR at #4790
* @sangaline for the improved PR with initial tests at #5069
* The contributors behind the corresponding fixes in the Ruby, Node.js and Java bindings

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6463)

### DIFF
--- a/py/test/selenium/webdriver/firefox/ff_profile_tests.py
+++ b/py/test/selenium/webdriver/firefox/ff_profile_tests.py
@@ -173,18 +173,50 @@ def test_autodetect_proxy_is_set_in_profile():
     assert profile.default_preferences["network.proxy.type"] == ProxyType.AUTODETECT['ff_value']
 
 
-def test_that_we_can_add_a_webextension(capabilities, webserver):
-    extension_filename = 'webextension.xpi'
-    # TODO: This file should probably live in a common directory.
+def test_add_extension_web_extension_with_id(capabilities, webserver):
     current_directory = os.path.dirname(os.path.realpath(__file__))
     root_directory = os.path.join(current_directory, '..', '..', '..', '..', '..')
-    data_directory = os.path.join(root_directory, 'javascript', 'node',
-                                  'selenium-webdriver', 'lib', 'test', 'data', 'firefox')
-    full_extension_filename = os.path.join(data_directory, extension_filename)
+    # TODO: This file should probably live in a common directory.
+    extension_path = os.path.join(root_directory, 'javascript', 'node', 'selenium-webdriver',
+                                  'lib', 'test', 'data', 'firefox', 'webextension.xpi')
 
     profile = FirefoxProfile()
-    profile.add_extension(full_extension_filename)
+    profile.add_extension(extension_path)
 
     driver = Firefox(capabilities=capabilities, firefox_profile=profile)
+    profile_path = driver.firefox_profile.path
+    extension_path_in_profile = os.path.join(profile_path, 'extensions', 'webextensions-selenium-example@example.com')
+    assert os.path.exists(extension_path_in_profile)
     driver.get(webserver.where_is('simpleTest.html'))
     driver.find_element_by_id('webextensions-selenium-example')
+    driver.quit()
+
+
+def test_add_extension_web_extension_without_id(capabilities, webserver):
+    current_directory = os.path.dirname(os.path.realpath(__file__))
+    root_directory = os.path.join(current_directory, '..', '..', '..', '..', '..')
+    extension_path = os.path.join(root_directory, 'third_party', 'firebug', 'mooltipass-1.1.87.xpi')
+
+    profile = FirefoxProfile()
+    profile.add_extension(extension_path)
+
+    driver = Firefox(capabilities=capabilities, firefox_profile=profile)
+    profile_path = driver.firefox_profile.path
+    extension_path_in_profile = os.path.join(profile_path, 'extensions', 'MooltipassExtension@1.1.87')
+    assert os.path.exists(extension_path_in_profile)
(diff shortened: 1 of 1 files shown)


py/selenium/webdriver/firefox/firefox_profile.py
```diff
@@ -330,6 +330,20 @@ def get_text(element):
                     rc.append(node.data)
             return ''.join(rc).strip()
 
+        def parse_manifest_json(content):
+            """Extracts the details from the contents of a WebExtensions `manifest.json` file."""
+            manifest = json.loads(content)
+            try:
+                id = manifest['applications']['gecko']['id']
+            except KeyError:
+                id = manifest['name'].replace(" ", "") + "@" + manifest['version']
```
davehunt: Thank you for picking this up! I've looked into it more and discovered that whilst web extensions don't mandate an id, it appears that when installing extensions from the profile [Firefox will verify that the name of the file/directory containing the extension matches the id of the add-on](https://searchfox.org/mozilla-central/rev/819cd31a93fd50b7167979607371878c4d6f18e8/toolkit/mozapps/extensions/internal/XPIDatabase.jsm#2420-2422). This means that we effectively do require an id to be set in the manifest, and that we use it when writing to the profile directory. It may work by coincidence when the id matches the name@version format used here. It's worth testing with the other clients, but if this is the case then I believe we should require the id and raise an exception when it's not provided to avoid harder to debug issues with Firefox launching without the addon installed. Sorry for my earlier comment suggesting that we needed id to be optional, which was based on [this documentation](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/applications).

py/test/selenium/webdriver/firefox/ff_profile_tests.py
```diff
@@ -171,3 +171,52 @@ def test_autodetect_proxy_is_set_in_profile():
...
+    driver = Firefox(capabilities=capabilities, firefox_profile=profile)
+    profile_path = driver.firefox_profile.path
+    extension_path_in_profile = os.path.join(profile_path, 'extensions', 'webextensions-selenium-example@example.com')
+    assert os.path.exists(extension_path_in_profile)
+    driver.get(webserver.where_is('simpleTest.html'))
+    driver.find_element_by_id('webextensions-selenium-example')
+    driver.quit()
+
+
+def test_add_extension_web_extension_without_id(capabilities, webserver):
```
davehunt: This test is failing for me.

py/test/selenium/webdriver/firefox/ff_profile_tests.py
```diff
@@ -171,3 +171,52 @@ def test_autodetect_proxy_is_set_in_profile():
 
     profile.set_proxy(proxy)
     assert profile.default_preferences["network.proxy.type"] == ProxyType.AUTODETECT['ff_value']
+
+
+def test_add_extension_web_extension_with_id(capabilities, webserver):
```
davehunt: These tests are just for legacy Firefox, and we should also add coverage to `py/test/selenium/webdriver/marionette`

py/selenium/webdriver/firefox/firefox_profile.py
motin: True! Reported as https://github.com/SeleniumHQ/selenium/issues/6471

py/test/selenium/webdriver/firefox/ff_profile_tests.py
motin: Is this due to the reasons you mentioned in https://github.com/SeleniumHQ/selenium/pull/6463/files#r221901851 or some other reason?

py/test/selenium/webdriver/firefox/ff_profile_tests.py
motin: It seems that many tests for legacy Firefox are missing for marionette. Should this be reported in general as a separate issue?

py/test/selenium/webdriver/firefox/ff_profile_tests.py
davehunt: The web extension doesn't appear to be copied to the profile path at all.

py/test/selenium/webdriver/firefox/ff_profile_tests.py
davehunt: I'd like to see more unit tests for the Selenium client that don't require a driver, for example we should be able to test several aspects of the `FirefoxProfile` class that aren't affected by the version of Firefox. Where there is a need for a driver test, we should at least make sure we have coverage in modern Firefox (marionette) as we will eventually be looking to drop support for legacy Firefox. If you're interested in contributing more, I'd love to get a coverage report for our unit tests and work on improving it - this will also give us confidence when we come to do larger scale refactoring.

py/test/selenium/webdriver/firefox/ff_profile_tests.py
davehunt: As this is passing in Travis, let's go ahead and merge. It was the only thing that was making me uncertain about merging. It's failing locally for me, but I'm guessing it's related to the version of Firefox.

davehunt:
Thanks for the contributions @motin and @sangaline. Please feel free to work on any of the follow ups, and to ping me for reviews or if you have any questions.

PR #6491, opened 2018-10-07
Added Locale.US for USER_AGENT toLowerCase

bozdemir84:
USER_AGENT toLowerCase without locale is causing problems with some languages, in may case Turkish. 
It maps WINDOWS to wındows which is resulting an exception later when it is used because of 'ı' char instead of 'i'. 
Calling toLowerCase with Locale.US solves this problem.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6491)


shs96c:
Good catch. Thank you! This is now merged, and will be part of the next selenium release.

PR #6522, opened 2018-10-11
[py] Subclass options classes from a common base class

lmtierney:
An ABC was used to ensure that `to_capabilities` was implemented. This method allows us to pass just the options class to `RemoteWebDriver` and not have to worry about capabilities.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6522)

### DIFF
--- a/py/selenium/webdriver/chrome/options.py
+++ b/py/selenium/webdriver/chrome/options.py
@@ -20,7 +20,7 @@ import os
 import platform
 
 from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
-from ..common.options import ArgOptions
+from selenium.webdriver.common.options import ArgOptions
 
 
 class Options(ArgOptions):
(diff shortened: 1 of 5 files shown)


lmtierney:
@Dude-X

lmtierney:
Safari options class will be added after I talk with @burg

AutomatedTester:
from a quick look through this looks great

py/selenium/webdriver/chrome/options.py
```diff
@@ -18,22 +18,21 @@
 import base64
 import os
 import platform
-import warnings
 
 from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
+from ..common.options import ArgOptions
```
cgoldberg: any reason for the relative import instead of absolute?

lmtierney: Habit, I can change it to absolute

PR #6525, opened 2018-10-12
Adding support for the legacy safaridriver option

joventuraz:

Safaridriver's current implementation of the w3c spec is missing proxy settings. This PR provides the ability to set the `--legacy` option, and use JWP instead.

As much as I'd like to just stick to w3c, there are too many tests that depend on the proxy settings.

Once Safaridriver's w3c implementation is more robust, we could consider dropping it, but for now this is helpful.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6525)

### DIFF
--- a/java/client/src/org/openqa/selenium/safari/SafariDriverService.java
+++ b/java/client/src/org/openqa/selenium/safari/SafariDriverService.java
@@ -83,8 +83,15 @@ public class SafariDriverService extends DriverService {
 
     private boolean legacy = Boolean.getBoolean(SAFARI_DRIVER_LEGACY_PROPERTY);
 
-    public Builder() {
-      usingTechnologyPreview(false);
+    /**
+     * Configures the driver server dialect.
+     *
+     * @param legacy True for legacy mode (JWP), false otherwise.
+     * @return A self reference.
+     */
+    public Builder withLegacy(boolean legacy) {
+      this.legacy = legacy;
+      return this;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


p0deje:
Isn't it possible to just instantiate SafariDriverService with custom argument and pass to the driver?

lmtierney:
@p0deje I believe their use case is for grid use. There's no way to get safaridriver into legacy mode using standalone/grid and this adds it

joventuraz:
@p0deje would it be beneficial to also add the option to run in legacy mode through SafariOptions? I could also add that to the PR.
@lmtierney that's correct, we are using standalone.

p0deje:
@lmtierney Right, I didn't think about server/grid case.

@joventuraz I'm not sure what SafariOptions imply in Java bindings, but I don't think it's used to control service, so I believe capability is a better place.

joventuraz:
It's my first time going through the source code, so I'm just getting familiar with similar examples implemented for other drivers. Did you mean that it would be good to add a capability besides the property that is in this PR?

p0deje:
@joventuraz I'm not Java binding maintainer, so not sure which option is the best here. @barancev @shs96c Can you guys have a look at this?

joventuraz:
What we are going to do for now is to drop setting the proxy settings so we can run safaridriver without --legacy; however, we could like to still be able to set --legacy through something like this PR if possible.

Please let me know if there is anything needed to either move this forward or come up with a different approach.

Thank you!

joventuraz:
Thanks for the feedback, we are dropping the proxy caps and will be able to run in w3c mode without issues, and since --legacy will no longer exist after Safari 12, I think we can close this PR.

shs96c:
Based on @burg's comment, I think that though this change was made with the right idea in mind, it's not going to be practical to land this in the tree, so I shall close it.

Despite this, thank you very much for taking the time to put together the PR. It really is very much appreciated. One thing we're looking at is trying to implement patching of capabilities on grid nodes via [JSON Patch](http://jsonpatch.com/). If you're interested, that would be a great place to start (possibly by allowing a [JSON Pointer](https://tools.ietf.org/html/rfc6901) to select nodes from a JsonInput?)

joventuraz:
@shs96c my pleasure! I'm interested in contributing, I'm not very familiar with the codebase though so expect questions on IRC if that's ok.

PR #6531, opened 2018-10-12
adding a css_classes property to WebElement

larkost-fetch:

This add a `css_classes` property to the Python WebElement that returns a list of the classes on that element (or an empty list if there is none). I find myself doing this over and over in my code, and `.get_attribute('class').split()` gets annoying after a while.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6531)


lmtierney:
Thank you for taking the time to open this PR, but unfortunately we won't be accepting it. Methods like this belong in frameworks built off of Selenium.

PR #6536, opened 2018-10-15
[py] verify command_executor ssl certificate by default

### LINKED ISSUE #6534: python's webdriver.Remote ignores invalid SSL certificates for selenium hub
## Meta -
**Bindings: python**
OS:   any
Selenium Version:  3.14.0
Browser:  any
Browser Version:  any

## Expected Behavior -

Connecting to a remote webdriver with an invalid SSL certificate should not be accepted. This was the case before 3.14.0.

Ideally, we should be able to pass a custom CA bundle to use for validation in `selenium.webdriver.Remote` constructor.

## Actual Behavior -

Because since https://github.com/SeleniumHQ/selenium/commit/8d98f3223c3757bac24f779f5d8c1e9e1e8edcfb the backend was changed to urllib3 and urllib3 does not validate SSL certificates by default ([reference](https://urllib3.readthedocs.io/en/latest/user-guide.html#certificate-verification)), then if the hub uses an invalid certificate, this is ignored.

## Steps to reproduce -

Something like 

```python
driver = webdriver.Remote(
    command_executor='https://wrong.badssl.com/',
    desired_capabilities={
      'browserName': 'firefox',
    })
```

should raise a `CertificateError` about the invalid certificate, similar to:
```
MaxRetryError: HTTPSConnectionPool(host='wrong.badssl.com', port=443): Max retries exceeded with url: / (Caused by SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:720)'),))
```

The example does not fully makes sense because `https://wrong.host.badssl.com` is obviously not the URL of a selenium hub. A real life case where this causes problems is when running selenium hub behind an https frontend.

cc @AutomatedTester

jerome-nexedi:
And expose get_certificate_bundle_path / set_certificate_bundle_path to
allow customizing the certificate verification

fixes #6534

cc @AutomatedTester

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6536)


jerome-nexedi:
I just understood that it's needed to upload the third party dependency.  I downloaded from https://pypi.org/project/certifi/#files and included it 
The sha256 was 456048c7e371c089d0a77a5212fb37a2c2dce1e24146e3b7e0261736aaeaa22a which matched what's on pypi.

jerome-nexedi:
> I just understood that it's needed to upload the third party dependency

... and there are more dependencies I need to had. I will try to make the test pass in a development branch and update here later. Sorry for noise.

agebhar1:
@jerome-nexedi we are facing equivalent problems - do you need some help to get this PR merged?

Issue #6471, opened 2018-10-02
Require an id to be set in Firefox web extension manifests


motin:
As mentioned in https://github.com/SeleniumHQ/selenium/pull/6463#discussion_r221901851:

> Whilst web extensions don't mandate an id, it appears that when installing extensions from the profile [Firefox will verify that the name of the file/directory containing the extension matches the id of the add-on](https://searchfox.org/mozilla-central/rev/819cd31a93fd50b7167979607371878c4d6f18e8/toolkit/mozapps/extensions/internal/XPIDatabase.jsm#2420-2422). This means that we effectively do require an id to be set in the manifest, and that we use it when writing to the profile directory. It may work by coincidence when the id matches the name@version format used here. It's worth testing with the other clients, but if this is the case then I believe we should require the id and raise an exception when it's not provided to avoid harder to debug issues with Firefox launching without the addon installed. Sorry for my earlier comment suggesting that we needed id to be optional, which was based on [this documentation](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/applications).

## Current client implementations ##
Here is how the ruby bindings handle web extensions without an id (it doesn't even look for applications.gecko.id): https://github.com/SeleniumHQ/selenium/blob/07a18746ff756e90fd79ef253a328bd7dfa9e6dc/rb/lib/selenium/webdriver/firefox/extension.rb#L85-L91

The Java bindings looks for applications.gecko.id but does the same as the Ruby bindings in case it isn't set:
https://github.com/SeleniumHQ/selenium/blob/ba32c5bc83272ff0e111fd4dccc40435700c6225/java/client/src/org/openqa/selenium/firefox/internal/FileExtension.java#L141-L155

The JavaScript bindings throw an error if applications.gecko.id is not defined: 
https://github.com/SeleniumHQ/selenium/blob/3e99112a0d59878c6f31cc50fe8b0593c49e497f/javascript/node/selenium-webdriver/firefox.js#L172-L174

The Python bindings, as proposed in https://github.com/SeleniumHQ/selenium/pull/6463, exhibit the same behavior as the Java bindings:
https://github.com/motin/selenium/blob/add-python-client-support-for-webextensions-in-firefox/py/selenium/webdriver/firefox/firefox_profile.py#L333-L339

p0deje:
Ruby support landed in a25638dfb68e2752a224ab7716cb5edc491e4033

p0deje:
[...]

PR #6550, opened 2018-10-17
[rb] Use id in manifest json for firefox extension if exists

tadashi0713:

Try to solve this issue
https://github.com/SeleniumHQ/selenium/issues/6471

Similar modification as python
https://github.com/SeleniumHQ/selenium/pull/6463

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6550)

### DIFF
--- a/rb/lib/selenium/webdriver/firefox/extension.rb
+++ b/rb/lib/selenium/webdriver/firefox/extension.rb
@@ -88,10 +88,10 @@ module Selenium
 
           manifest = JSON.parse(File.read(manifest_path))
           if manifest.key?('application') && manifest['application'].key?('gecko')
-	    manifest['application']['gecko']['id']
-	  else
-	    [manifest['name'].delete(' '), manifest['version']].join('@')
-	  end
+	          manifest['application']['gecko']['id']
+	        else
+	          [manifest['name'].delete(' '), manifest['version']].join('@')
+	        end
         end
       end # Extension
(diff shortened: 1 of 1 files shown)


rb/lib/selenium/webdriver/firefox/extension.rb
```diff
@@ -87,7 +87,10 @@ def read_id_from_manifest_json(directory)
           return unless File.exist?(manifest_path)
 
           manifest = JSON.parse(File.read(manifest_path))
-          [manifest['name'].delete(' '), manifest['version']].join('@')
+          id = if manifest.key?('application') && manifest['application'].key?('gecko')
```
luke-hill: Couldn't you just write this as an if/else or a guard clause like on line 87? Just seems a bit weird doing this conditional assignment for essentially an if/else.

tadashi0713: @luke-hill Thanks for advice

davehunt:
@titusfortner I agree this should be tested as it's the more common use case. It should be trivial to add a web extension with an ID for testing purposes.

p0deje:
I couldn't merge this PR because it needed code changes (e.g. look for `applications`, not `application`), so I've pushed a25638dfb68e2752a224ab7716cb5edc491e4033 to fix this.

In any event, thank you for the proposed contribution!

PR #6558, opened 2018-10-19
Bug 6541

abidema:

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6558)


barancev:
This PR become stale due to changes in the code (namely, adding ChromiumOptions class common for Chrome and Edgium) and can't be applied without total rework. So I've created an improved implementation on the base of this PR that works for both Chrome and Edgium and takes into account not only arguments, but also extensions and experimental options, see commit ec9c67b2fb1ec110f51934832fec3cef8a9bf746.

PR #6596, opened 2018-10-30
added methods to get screen resolution of various devices

gowthamraj198:
- Added methods to get screen resolution of various devices.
- Chrome v-69 cannot be maximized using maximize function. This adjusts only the height of the window. A method is added to maximize the window in according to the screen resolution. 

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6596)

### DIFF
--- a/rb/lib/selenium/webdriver/common/window.rb
+++ b/rb/lib/selenium/webdriver/common/window.rb
@@ -156,6 +156,27 @@ module Selenium
       def full_screen
         @bridge.full_screen_window
       end
+
+      #
+      # Get current screen resolution
+      #
+
+      def screen_height
+        @bridge.executeScript("return screen.height")
+      end
+
+      def screen_width
+        @bridge.executeScript("return screen.width")
+      end
+
+      #
+      # Maximize the window according to screen size
+      #
+      def maximize_acc_to_resolution
+        @bridge.setWindowSize(screen_height, screen_width)
+        @bridge.setWindowPosition(0, 0)
+      end
+
     end # Window
   end # WebDriver
 end # Selenium


rb/lib/selenium/webdriver/common/window.rb
```diff
@@ -156,6 +156,27 @@ def minimize
       def full_screen
         @bridge.full_screen_window
       end
+
+      #
+      # Get current screen resolution
```
luke-hill: This technically is wrong as it is returning individual dimensions, not the resolution. i.e. this returns 1024, not 1024x768.

Is there a reason you have these specified into individual items?

Why not return the resolution as either a hash / array with the two items below set to private?

Also is there a reason you've not queried the `#size` method - And gone down the JS route?

rb/lib/selenium/webdriver/common/window.rb
```diff
@@ -156,6 +156,27 @@ def minimize
...
+
+      def screen_width
+        @bridge.executeScript("return screen.width")
+      end
+
+      #
+      # Maximize the window according to screen size
+      #
+      def maximize_acc_to_resolution
+        @bridge.setWindowSize(screen_height, screen_width)
```
luke-hill: Again here, there are built in helpers? Seems a bit off?

luke-hill:
EDIT: You'll need to sign the CLA as well before continuing, as this is a non-trivial piece of code.

lmtierney:
I'm not in favor of adding this to the API. This should be handled in a framework built on top of selenium

p0deje:
I'm also against adding it to API:

1. [`Screen`](https://drafts.csswg.org/cssom-view/#the-screen-interface) is still a part of specification working draft and is not required to be supported in all browsers.
2. API is just a sugar and should rather be implemented in your own test harness or wrapping libraries  like Capybara or Watir.
3. If maximize is not supported in Chrome 69, it is an issue in Chrome and, once it's fixed, there is no need to have this API anymore.

PR #6617, opened 2018-11-03
[java] Implement equals() and hashCode() in `Select` element wrapper

valfirst:

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6617)

### DIFF
--- a/java/client/src/org/openqa/selenium/support/ui/Select.java
+++ b/java/client/src/org/openqa/selenium/support/ui/Select.java
@@ -23,6 +23,7 @@ import org.openqa.selenium.WebElement;
 import org.openqa.selenium.WrapsElement;
 
 import java.util.List;
+import java.util.Objects;
 import java.util.StringTokenizer;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


java/client/src/org/openqa/selenium/support/ui/Select.java
```diff
@@ -305,4 +306,21 @@ private void setSelected(WebElement option, boolean select) {
       option.click();
     }
   }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
```
shs96c: Nowadays, we tend to follow a pattern of:

```
if (!(o instanceof Select)) {
  return false;
}
Select that = (Select) o;
return Objects.equals(this.element, that.element);
```

shs96c: The performance is dramatically different, but we strip more boilerplate this way.

valfirst: @shs96c updated to follow project patterns

Issue #5373, opened 2018-01-20
Problem with HUB forwarding request to NODE


StephenPCG:
## Meta -
Selenium Version:  
* hub docker image: selenium-hub:latest
* node docker image: selenium-node-chrome:latest

## Steps to reproduce -

We deployed a selenium-hub and several selenium-node-chromes with docker, due to our company's firewall policy, hub does not have direct network to nodes. We have a **single** ingress http gateway (nginx), which does the following route based on `Host` header:
* `hub.selenium.mycompany.com` -> route to hub
* `node1.selenium.mycompany.com` -> route to chrome node 1
* `node2.selenium.mycompany.com` -> route to chrome node 2
* so on and so forth

The node is started as:
```
SE_OPTS='-remoteHost http://node1.selenium.mycompany.com -id node1' /opt/bin/entry_point.sh
```

And the node successfully registered with hub. However, things failed when hub got a request to create a new session.

The client sends an HTTP request like:

```
POST /wd/hub/session HTTP/1.1
Host: hub.selenium.mycompany.com
(other info snipped)
```

The hub selects an available node (assume it's node1), and forwards the request by:

* first makeing a tcp connection to `node1.selenium.mycompany.com:80` (connected to the nginx)
* then forwarding the http request as is, with `Host` header not modified

When this request reaches the nginx, it sees the `Host` header is `hub.selenium.mycompany.com`, so routed back to hub again, causes a dead loop.

Suggested solution is, when hub forwarding the request to node, it should set Host header according to node's `remoteHost` value. This change should do no harm to most existing installations.

diemol:
@StephenPCG, seems that you have special network conditions. Why don't you check how to overwrite the Host header via Nginx?

What the hub does is just to contact the node via the url used to register, and if it is not accessible then it is just not possible to do it. I don't think the Grid should be changed to support that, it is merely a routing configuration.

StephenPCG:
@diemol Because both hub and node have the same nginx instance as gateway. The hub forwards the request without any information indicating it should be routed to hub, nginx can not distinguish it. The problem is not nginx not rewriting Host header, the problem is nginx has no way to know the request is intended to be sent to node.

The hub, being an HTTP router, it should follow some forwarding conventions. For example, it could send the request like:

```
POST /wd/hub/session HTTP/1.1
Host: node1.selenium.mycompany.com
X-Forwarded-For: hub.selenium.mycompany.com
```

[...]

PR #6627, opened 2018-11-07
Restored `-remoteHost` support and fix for nodes behind reverse proxy

kinjelom:
1. Restored `-remoteHost` CLI option support, lost in revision: 2fb93404832400c5c02fdac6593b19d19e582d8d (@barancev "Composition instead of inheritance in CLI options")
2. Support for nodes behind reverse proxy (Cloud Foundry GoRouter in my case) https://github.com/SeleniumHQ/selenium/issues/5373

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6627)

### DIFF
--- a/java/server/src/org/openqa/grid/internal/TestSession.java
+++ b/java/server/src/org/openqa/grid/internal/TestSession.java
@@ -77,9 +77,7 @@ public class TestSession {
 
   private static final Logger log = Logger.getLogger(TestSession.class.getName());
   static final int MAX_IDLE_TIME_BEFORE_CONSIDERED_ORPHANED = 5000;
-  private static final String PROXY_HOST_HEADER_NAME = "Proxy-Host";
-  private static final String HOST_HEADER_NAME = "Host";
-  
+
   private final String internalKey;
   private final TestSlot slot;
   private volatile ExternalSessionKey externalKey = null;
@@ -454,20 +452,10 @@ public class TestSession {
       proxyRequest = new HttpRequest(HttpMethod.valueOf(request.getMethod()), uri);
     }
 
-    String host = request.getHeader(HOST_HEADER_NAME);
-    String proxyHost = request.getHeader(PROXY_HOST_HEADER_NAME);
-    if (proxyHost == null || proxyHost.isEmpty()) {
-      proxyRequest.setHeader(PROXY_HOST_HEADER_NAME, host);
-    } else {
-      proxyRequest.setHeader(HOST_HEADER_NAME, remoteURL.getHost());
-    }
     for (Enumeration<?> e = request.getHeaderNames(); e.hasMoreElements(); ) {
       String headerName = (String) e.nextElement();
 
       if ("Content-Length".equalsIgnoreCase(headerName)) {
-        continue; // already set
-      }
-      if (HOST_HEADER_NAME.equalsIgnoreCase(headerName) || PROXY_HOST_HEADER_NAME.equalsIgnoreCase(headerName)) {
         continue; // already set
       }
 


java/server/src/org/openqa/grid/common/RegistrationRequest.java
```diff
@@ -203,6 +203,9 @@ public static RegistrationRequest build(GridNodeConfiguration configuration, Str
     if (pendingConfiguration.port != null) {
       pendingRequest.configuration.port = pendingConfiguration.port;
     }
+    if (pendingConfiguration.remoteHost != null) {
+      pendingRequest.configuration.remoteHost = pendingConfiguration.remoteHost;
+    }
```
diemol: This is not really needed, the line added in `GridNodeConfiguration.java` is what really fixes the `remoteHost` regression.

diemol:
Hi @wildloop, thanks for this PR.

I added a small comment about the `remoteHost` regression.

About the reverse proxy changes, I am not sure if we should add this without tests. Could you please add some tests to cover this? Moreover, this would help in your use case, are we sure that this does not break any other use cases?
What do you think about the reverse proxy change @barancev @shs96c?

wildloop:
@diemol what about enabling this feature with special CLI option `-supportReverseProxy true` or maybe if `-remoteHost` is a domain name?

diemol:
@wildloop we don't want to add new flags or features but only the bug fix, the reason for this is that we expect the next release is the last one of the current Grid. We are currently working out in a new Grid, which welcomes suggestions like the one you would like to add. 
If you would like to check the in-progress development code of the new Grid, check [here](https://github.com/SeleniumHQ/selenium/tree/master/java/server/src/org/openqa/selenium/grid). Also, you can join us in the [IRC/Slack channel](https://seleniumhq.herokuapp.com) where we can discuss new things as well.

Therefore, could you please leave only the bug fix in this PR so we can merge it? Thanks!

wildloop:
@diemol Support for reverse proxy is basically a fix too. How do you use `remoteHost`? I think that it works only when `remoteHost` is IP address or the same domain as hub's domain. In these cases change of a `Host` HTTP header doesn't matter - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host
That fix gives possibility to run the grid's nodes on a remote host indicated by domain address.

java/server/src/org/openqa/grid/common/RegistrationRequest.java
wildloop: You're right :) corrected: https://github.com/SeleniumHQ/selenium/pull/6627/commits/91e71499f37dc1852db0df9bbbb7480acbf0cef2

wildloop:
Ready to manual tests:
- https://github.com/wildloop/selenium/releases/download/3.141.59-SNAPSHOT/selenium-server-standalone-3.141.592-SNAPSHOT.jar (piver has even more power than semver.org ;) )
- https://hub.docker.com/r/wildloop/selenium-hub/
- https://hub.docker.com/r/wildloop/selenium-node-chrome/
- https://hub.docker.com/r/wildloop/selenium-node-firefox/

diemol:
@wildloop just to circle back on the previous point and make it more clear, we want to avoid making changes to support new use cases in the Grid, especially to centric classes like `TestSession.java`. 

We understand that this might be a bug fix for your use case, but we are not aware of what side effects this could bring to other users. We prefer to move things like this to the new Grid, where we welcome you to participate in its development by making suggestions and PRs.

In addition, for future reference, it is more desirable to add tests to changes like this one.

So please, leave only the bug fix in the PR so we can merge it and give you credit for it. Thanks!

wildloop:
@diemol done: https://github.com/SeleniumHQ/selenium/pull/6627/commits/6cfb57662d62d230d559f236cbc2cc1a090aa2ef

diemol:
Merged as https://github.com/SeleniumHQ/selenium/commit/07a116aa32e6b464567a331ed9d1c4ffbee104bb, thanks for your contribution @wildloop!

wildloop:
Thanks! 

-----
If someone needs support for reverse proxy before the new grid will be released, he can use my forks:
- [`selenium-server-standalone` source code](https://github.com/wildloop/selenium/tree/reverse-proxy-support)
- [`selenium-docker` source code](https://github.com/wildloop/selenium/tree/reverse-proxy-support)

Binaries:
- [selenium-server-standalone-3.141.592-SNAPSHOT.jar](https://github.com/wildloop/selenium/releases/download/3.141.59-SNAPSHOT/selenium-server-standalone-3.141.592-SNAPSHOT.jar)
- [`docker pull wildloop/selenium-hub`](https://hub.docker.com/r/wildloop/selenium-hub/)
- [`docker pull wildloop/selenium-node-chrome`](https://hub.docker.com/r/wildloop/selenium-node-chrome/)
- [`docker pull wildloop/selenium-node-firefox`](https://hub.docker.com/r/wildloop/selenium-node-firefox/)

lionlamb:
Hi @wildloop ,
I have been using your binaries and manifest.yaml examples and pushed to CF, however the node still not able to connect with hub and talk to https. I am wondering which part had I done wrong. Do you mind to shade some light?

wildloop:
@lionlamb Please try to connect to http://selenium-node-chrome-1.mycloudabc.cloud.com - HTTP protocol is disabled by default but selenium client doesn't support HTTPS (as I remember).
Next you can set `GRID_DEBUG: true` and share some logs.

You can try to configuration with [CF Internal Routing / Domains](https://docs.cloudfoundry.org/devguide/deploy-apps/routes-domains.html#internal-routes) and [C2C Networning](https://docs.cloudfoundry.org/concepts/understand-cf-networking.html#policies) that should simplify configuration and may works without hacking ;) ... but I didn't  try one.

lionlamb:
@wildloop 

Thanks for your kind reply.  I have tried many hours and checked the logs many times but still no luck. 

First, I setup both hub and node respectively again from the changes you share, it shows the following error:

`2019-05-12T22:13:30.51+0800 [xxxxxx] OUT 14:13:30.512 INFO [SelfRegisteringRemote$1.run] - Couldn't register this node: The hub is down or not responding: Failed to connect to selenium-hub.apps.internal/10.177.166.161:8080`

And i checked the hub, it seems like created at port 4444 automatically, so I update the manifest to try up. still getting the above error (just the 8080 replaced with 4444), here is the log info from hub

` 2019-05-12T20:18:26.81+0800 [xxxxxx] OUT 12:18:26.815 INFO [Hub.start] - Clients should connect to http://10.177.166.161:4444/wd/hub`

Then, I tried to do internal routing as per mentioned by you :) and got the message that the hub and node already exists:

```
cf map-route selenium-hub apps.internal --hostname selenium-hub
Creating route selenium-hub.apps.internal for mycloudabc / space seleniumTest as cloudtest@mycloudabc.com...
OK
Route selenium-hub.apps.internal already exists
```

I am curious do we need to EXPOSE the port 4444 in the docker image?

wildloop:
What about cf network policies?
Did you create policy to allow direct network traffic between nodes and hub?
https://cli.cloudfoundry.org/en-US/cf/add-network-policy.html

lionlamb:
@wildloop I didn't because SAP Cloud Foundry doesn't support it

wildloop:
So, go back to prev config and enable debug logs.
You can also connect to container using `cf ssh app-name` and test connection to hub/nodes (e.g. using netcat `nc -tvn <ip> <port>`).

lionlamb:
Yup, enabled the debug logs & Using netcat from node to test connection 

```
seluser@54dfef6b-bf7d-6c54-76cf-5hdg:~$ nc -tvn 10.177.166.161 4444
(UNKNOWN) [10.177.166.161] 4444 (http-alt) : Connection refused
```

PR #6647, opened 2018-11-12
[Ruby] delete unnecessary return value, nil

chiba3:
Refactoring. Only delete unnecessary return value nil. Because ruby's default return value is nil.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6647)

### DIFF
--- a/rb/lib/selenium/webdriver/common/driver.rb
+++ b/rb/lib/selenium/webdriver/common/driver.rb
@@ -281,9 +281,7 @@ module Selenium
       # @see SearchContext
       #
 
-      def ref
-        nil
-      end
+      def ref; end
 
       private
 


luke-hill:
Whilst this all is legit, it would be better if you've got time to find out why this method exists.

I'd wager that it's either redundant or can just be self defined (i.e. whenever this method is referenced, you just reference `nil`)

titusfortner:
@luke-hill you should investigate the code yourself before making a comment like this. The method exists because we need to be able to call it on both driver & element objects, even though the value doesn't apply to driver; hence it correctly returns nil.

Looks like we haven't run our rubocop on this code in quite a while; lots of updates should be made. But yes, this is best practice from the style guide.

titusfortner:
Thanks @chiba3 for the contribution

chiba3:
@titusfortner Thanks

Issue #1630, opened 2016-02-12
"Exception: Unable to find a free port" - SeGrid/Python/Chromedriver/Ubuntu+OSX


benhastings:
**Short Description:**
on attempting to execute parallel tests on a host using selenium grid, the server-standalone hub accepts the first session and rejects all subsequent requests to create a session until the first terminates.

**Longer Description:**
I am not a java developer, so I can only dig so deeply, but it does look like something is not being correctly executed in the PortProber.java class.  However, it may be something to do with the python requests to create a new session.  I spent some time digging into the python code and wasn't able to find a way to force a particular port be requested for the new session.

We're trying to run multiple scripts on a headless virtual.  This pattern has worked previously, but we needed to make updates to our chrome version, so we went ahead and updated the selenium-standalone version (I've tried 2.49-2.52), chromedriver (2.20-2.21), and Chrome (48.0.2564.82).  In fact, in developing the python test script, I was able to run 4 sessions simultaneously on my OSX laptop.  That is no longer the case (server standalone 2.50/2.52 tried - 2.50 previously worked)

I did try the selenium-server-standalone + a quick webdriver.io (nodeJS) script and it appears to work on my laptop.

**Script to recreate:**
(start one instance, then start another on the same system)

``` python
#test.py
from selenium import webdriver
import time

baseURL='http://www.yahoo.com'
url2='http://www.bing.com'

options=webdriver.ChromeOptions()
options.add_argument('--disable-logging')
driver=webdriver.Remote(desired_capabilities=options.to_capabilities())

driver.get(baseURL)
time.sleep(2.5)
print(driver.title)

driver.get(url2)
time.sleep(2.5)
print(driver.title)

driver.get(baseURL)
time.sleep(2.5)
print(driver.title)

driver.get(url2)
time.sleep(2.5)
print(driver.title)

driver.close()
driver.quit()
```

**Python Script Error:**

```
>>>python test.py
Traceback (most recent call last):
  File "test.py", line 11, in <module>
    driver=webdriver.Remote(desired_capabilities=options.to_capabilities())
  File "/usr/local/lib/python2.7/dist-packages/selenium-2.51.1-py2.7.egg/selenium/webdriver/remote/webdriver.py", line 87, in __init__
    self.start_session(desired_capabilities, browser_profile)
  File "/usr/local/lib/python2.7/dist-packages/selenium-2.51.1-py2.7.egg/selenium/webdriver/remote/webdriver.py", line 141, in start_session
[...]

PR #6704, opened 2018-11-29
fix and use ephemeral port from platform

tflori:

This solves the port negotiation problem on linux (and maybe other platforms). #1630 

The removed code leads to the result that `FIRST_PORT` and `LAST_PORT` is both set to 1024 which always returns 1024. In fact we should trust the ephemeral ports reports by the detector. With the min and max we just assure a valid port range.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6704)

### DIFF
--- a/java/client/src/org/openqa/selenium/net/LinuxEphemeralPortRangeDetector.java
+++ b/java/client/src/org/openqa/selenium/net/LinuxEphemeralPortRangeDetector.java
@@ -46,9 +46,9 @@ public class LinuxEphemeralPortRangeDetector implements EphemeralPortRangeDetect
     int lowPort = defaultRange.getLowestEphemeralPort();
     int highPort = defaultRange.getHighestEphemeralPort();
     try (BufferedReader in = new BufferedReader(inputFil)) {
-      String[] split = in.readLine().split("\\s");
+      String[] split = in.readLine().split("\\s", 2);
       lowPort = Integer.parseInt(split[0]);
-      highPort = Integer.parseInt(split[1]);
+      highPort = Integer.parseInt(split[1].trim());
     } catch (IOException ignore) {
     }
     firstEphemeralPort = lowPort;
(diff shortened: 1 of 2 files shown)


cgoldberg:
default ephemeral port range in the Linux kernel is: 32768-60999

see `ip_local_port_range`:
https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt

tflori:
> default ephemeral port range in the Linux kernel is: 32768-60999
> 
> see `ip_local_port_range`:
> https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt

I'm aware of this fact. Before the ephemeral port range was not used. Instead the higher range (between 1024 and <start of ephemeral port range> or between <end of ephemeral port range> and 65535) outside the ephemeral port range was used.

This must have been implemented when the most users used windows xp - because there the port range was too small. I just understood today what this code should do but I can't agree that it makes sense.

In some linux environments (often on servers) the ephemeral port range is just set to 1024-65535 and the code before was just stuck on port 1024. This can not be a solution too.

In my opinion the ephemeral port range should be used unless it has enough ports.

```
ephemeralRange = ephemeralUpperBound - ephemeralLowerBound
if ephemeralRange < 5000 then
  belowRange = ephemeralLowerBound - 1024
  aboveRange = 65535 - ephemeralUpperBound
  if belowRange > aboveRange then use belowRange
  else use aboveRange
else use ephemeralRange
```

shs96c:
@tflori, first of all a heartfelt apology. Leaving this PR in limbo for the better part of three years isn't a good way to show the appreciation and thanks for feel for the hard work it represents. I'm sorry it took so long for us to finally get round to reviewing and merging this, and I thank you for your patience. We do thank you for the PR, though! It's clearly solving a real problem.

Secondly, I've now merged this into the tree as 294d1c92fb9578513c82c1431d6c25b98364fcc2. I had to rebase this locally and handle some merge conflicts, but your name is on the commit. 

Once again, thank you!

tflori:
Yay! Finally 😀 Thank you

PR #6711, opened 2018-12-01
[py] WebDriverWait: document until and until_not

native-api:

See https://stackoverflow.com/questions/53514703/wait-for-element-and-return-it and specifically https://stackoverflow.com/questions/53514703/wait-for-element-and-return-it#comment93898809_53514792 for context.

Considered autogenerating the docstrings from java API, but seeing how vastly different the two bindings are, very little from the Java docs can actually be reused.

I couldn't find any other docstrings with markup; chose ReST markup ([as opposed to Google or NumPy](https://stackoverflow.com/questions/3898572/what-is-the-standard-python-docstring-format) one) because `sphinx.ext.napoleon` is not enabled in https://github.com/SeleniumHQ/selenium/blob/master/py/docs/source/conf.py which suggests that Selenium doesn't endorse either of the latter.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6711)

### DIFF
--- a/py/selenium/webdriver/support/wait.py
+++ b/py/selenium/webdriver/support/wait.py
@@ -63,8 +63,8 @@ class WebDriverWait(object):
         """Calls the method provided with the driver as an argument until the \
         return value does not evaluate to ``False``.
 
-        :arg method: callable(WebDriver)
-        :arg message: optional message for :exc:`TimeoutException`
+        :param method: callable(WebDriver)
+        :param message: optional message for :exc:`TimeoutException`
         :returns: the result of the last call to `method`
         :raises: :exc:`selenium.common.exceptions.TimeoutException` if timeout occurs
         """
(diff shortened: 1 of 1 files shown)


lmtierney:
We really should set a preferred docstring style for the project and move all our docstrings toward that. We should also update sphinx while we're at it. @AutomatedTester do you have a preference? Are you ok with ReST? @davehunt what do your teams use?

cgoldberg:
> We really should set a preferred docstring style for the project

For specifying parameters/returns/exceptions, I like embedded reStructuredText (works well with sphinx).

For general docstring formatting, we should follow [PEP257](https://www.python.org/dev/peps/pep-0257/).   It's pretty pedantic. but the docstrings would read really consistently, make great generated docs, and could be checked/enforced with tooling like `pydocstyle`.

davehunt:
I agree with everything @cgoldberg said. 😄

lmtierney:
Sounds good then! I'll update the wiki per this conversation. @native-api since we've used `:param` elsewhere already (e.g. `firefox/webdriver.py`, `firefox/service.py`, etc), can you change your `arg` accordingly for consitency?

lmtierney:
@native-api I'm ready to merge this. Do you want to add the email associated with the commits to your github account? So the commits show that they came from your account.

native-api:
> Do you want to add the email associated with the commits to your github account? So the commits show that they came from your account.

Done, thanks.

skidzo:
Hi, I found that we could easily do a `class AsyncWebDriverWait`, that just overloads `until()` and `until_not()` with an async version that just uses `await aio.sleep(self._poll)` instead of `time.sleep(self._poll)`

Is async of interest for you? - for me it is!
Do you want to create a pull request?

PR #6728, opened 2018-12-05
[py] Update Sphinx

native-api:
Subj, from https://github.com/SeleniumHQ/selenium/pull/6711#issuecomment-443470408.

Other changes:
* Only left immediately needed packages in `tox.ini`:
    * the dependency tree has changed so much between versions that I felt that updating the full list of packages and versions one by one is lots of redundant work. It's your call though.
    * This way, it's clear what in `3rd-party/py` is really needed and what is just there as dependencies.
        * It's (still) unclear though what is a dependency for what -- so no easy way to check that there are no redundant files in `3rd-party/py`.
* Ignored that platform-independent packages were present as wheels.
    * Also redundant work to check what is platform-independent and what is not.
    * A source package is also superior to a wheel from maintenance perspective because it's a [Corresponding Source](https://www.gnu.org/licenses/gpl-3.0.en.html#section1). The only possible reason (that I can see) why Selenium keeps copies of dependencies is to be insured from a loss of their supply channels. And a source package is the only thing that you can do everything you may ever need with if you have nothing else.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6728)


native-api:
I did my best to not affect anything else but since checks are broken, can't 100% guarantee that I didn't break anything.

lmtierney:
> the dependency tree has changed so much between versions that I felt that updating the full list of packages and versions one by one is lots of redundant work. It's your call though.

I think this should be fine since you saved what was needed for the `Sphinx` version you locked it at.

lmtierney:
Builds find and looks good locally for me. The checks that failed weren't impacted by your changes. Thanks for another contribution!

PR #6749, opened 2018-12-13
[javascript] equals should return the value of the assertion

### LINKED ISSUE #6748: [javascript/node/selenium-webdriver] WebElement.equals returns null
## Bug Report

node module: selenium-webdriver@4.0.0-alpha.1
node: 10.9.0
npm: 6.4.1

While working on the selenium 4 upgrade for Protractor, I noticed that the `WebElement.equals` method could return a null value. When you creates a web element and sends the exact same web element in to the equals method, it [hits the conditional](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/lib/webdriver.js#L1817) and returns true .

### Example (similar to the [reflexive unit test](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/test/lib/webdriver_test.js#L1001)):

```
 var e1 = element.findElement(By.id('foo'));
 expect(WebElement.equals(e1, e1)).toBe(true);  // this works and returns true
```

If you create two different web element instances, the conditional returns false and run the [execute script](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/lib/webdriver.js#L1820). The equals method returns null.

### Example:

```
 var e1 = element.findElement(By.id('foo'));
 var e2 = element.findElement(By.id('foo'));
 expect(await WebElement.equals(e1, e2)).toBe(true);  // null not equal to true
```

## Proposed fix

Adding a return statement results in the promised boolean. 

```
  static async equals(a, b) {
    if (a === b) {
      return true;
    }
    return a.driver_.executeScript('return arguments[0] === arguments[1]', a, b);
  }
```

cnishina:
closes #6748

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6749)

### DIFF
--- a/javascript/node/selenium-webdriver/lib/webdriver.js
+++ b/javascript/node/selenium-webdriver/lib/webdriver.js
@@ -1817,7 +1817,8 @@ class WebElement {
     if (a === b) {
       return true;
     }
-    return a.driver_.executeScript('arguments[0] === arguments[1]', a, b);
+    return a.driver_.executeScript(
+      'return arguments[0] === arguments[1]', a, b);
   }
 
   /** @return {!WebDriver} The parent driver for this instance. */


cnishina:
I'm not exactly sure why things are failing. I don't think its because of the change I made.

PR #6787, opened 2018-12-25
Enable support for browser_specific_settings in Firefox

SebastienGllmt:
Firefox supports `browser_specific_settings` as an alternative to `applications` because Chrome rejects addons with `applications` inside the manifest but supports the `browser_specific_settings` field.

This was originally introduced in [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1262005)
And you can see it inside the Firefox codebase [here](https://searchfox.org/mozilla-central/source/toolkit/mozapps/extensions/internal/XPIInstall.jsm#411)
And supported by the Addons-linter [here](https://github.com/mozilla/addons-linter/blob/master/src/parsers/manifestjson.js#L256)

Both Firefox and the addons-linter give warnings if both `applications` and `browser_specific_settings` are present and give priority to `browser_specific_settings`
Similarly, my change gives priority to `browser_specific_settings` but I did not add a warning.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6787)

### DIFF
--- a/javascript/node/selenium-webdriver/firefox.js
+++ b/javascript/node/selenium-webdriver/firefox.js
@@ -166,26 +166,26 @@ async function installExtension(extension, dir) {
   }
 
   let buf = await archive.getFile('manifest.json');
-
-  let {browser_specific_settings} =
-      /** @type {{browser_specific_settings:{gecko:{id:string}}}} */(
-        JSON.parse(buf.toString('utf8')));
+  let parsedJSON = JSON.parse(buf.toString('utf8'));
   
-  let applications;
-  if (browser_specific_settings
-    && browser_specific_settings.gecko
-    && browser_specific_settings.gecko.id) {
+  let { browser_specific_settings } =
+    /** @type {{browser_specific_settings:{gecko:{id:string}}}} */
+    parsedJSON; 
+
+  if (
+    browser_specific_settings &&
+    browser_specific_settings.gecko
+  ) {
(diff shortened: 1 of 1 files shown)


SebastienGllmt:
Any chance this can be merged? If there is a reason it hasn't been merged yet let me know and I can fix it.

javascript/node/selenium-webdriver/firefox.js
```diff
@@ -166,9 +166,26 @@ async function installExtension(extension, dir) {
   }
 
   let buf = await archive.getFile('manifest.json');
-  let {applications} =
+
+  let {browser_specific_settings} =
+      /** @type {{browser_specific_settings:{gecko:{id:string}}}} */(
+        JSON.parse(buf.toString('utf8')));
```
jleyba: Save the `JSON.parse` result to a variable so you don't have to re-parse on line 185 below.

That, or perhaps change this to:

```js
let {applications, browser_specific_settings} = JSON.parse(buf.toString('utf8'))
```

SebastienGllmt:
@jleyba the code was like that to match the original Firefox code but it seems they've cleaned this up themselves. I pushed a new commit based on the Mozilla linter logic: https://github.com/mozilla/addons-linter/blob/master/src/parsers/manifestjson.js#L277

SebastienGllmt:
Just another ping on this issue 🙏

PR #6814, opened 2019-01-04
[py] Don't override browser options with desired capabilities by defa…

carlosgcampos:
…ult in WebKitGTK

When desired_capabilities parameter is omitted in WebKitGTK driver, the
default capabilities are used. If options are also passed they are
updated from desired capabilities, overriding the passed options. The
desired_capabalities parameter should default to None instead.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6814)

### DIFF
--- a/py/selenium/webdriver/webkitgtk/webdriver.py
+++ b/py/selenium/webdriver/webkitgtk/webdriver.py
@@ -20,9 +20,9 @@ try:
 except ImportError:
     import httplib as http_client
 
-from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
 from selenium.webdriver.remote.webdriver import WebDriver as RemoteWebDriver
 from .service import Service
+from .options import Options
 
 
 class WebDriver(RemoteWebDriver):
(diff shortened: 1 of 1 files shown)


py/selenium/webdriver/webkitgtk/webdriver.py
```diff
@@ -20,18 +20,17 @@
 except ImportError:
     import httplib as http_client
 
-from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
 from selenium.webdriver.remote.webdriver import WebDriver as RemoteWebDriver
 from .service import Service
-
+from .options import Options
```
lmtierney: Can you address the flake8 error here? Then we can merge.

lmtierney (changes requested):

lmtierney:
Merged, thank you!

bphd:
WebKitGTK is now the only browser that can't run without any options, producing an error asking about desired capabilities. Is that because of you?

titusfortner:
@bphd please file a bug report if you believe the code is not doing what it should.

PR #6843, opened 2019-01-11
Move all remote uploaded files to the same directory so remote IE works with multiple file upload

twalpole:

Currently remote IE can't send multiple files to a an `<input type="file">` field because it requires all files to be from the same directory.  With the way remote file upload currently works each file is uploaded to it's own temporary directory.  This PR changes that behavior to leave all upload files in a single directory.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6843)

### DIFF
--- a/java/server/src/org/openqa/selenium/remote/server/commandhandler/UploadFile.java
+++ b/java/server/src/org/openqa/selenium/remote/server/commandhandler/UploadFile.java
@@ -65,8 +65,8 @@ public class UploadFile implements CommandHandler {
           (allFiles == null ? 0 : allFiles.length)));
     } else {
       // IE requires multiple uploads to all be from the same directory - move files into one location
-      Path uploads_dir = Files.createDirectories(tempDir.toPath().getParent().resolve("remote_uploads"));
-      Path dest = uploads_dir.resolve(allFiles[0].getName());
+      Path uploadsDir = Files.createDirectories(tempDir.toPath().getParent().resolve("remote_uploads"));
+      Path dest = uploadsDir.resolve(allFiles[0].getName());
 
       if (Files.exists(dest)){
         // IE maintains a lock on previously uploaded files so we can't delete/overwrite - move it instead
--- a/java/server/src/org/openqa/selenium/remote/server/commandhandler/UploadFile.java
+++ b/java/server/src/org/openqa/selenium/remote/server/commandhandler/UploadFile.java
@@ -31,6 +31,10 @@ import java.io.File;
 import java.io.IOException;
 import java.util.Map;
 import java.util.Objects;
+import java.nio.file.Files;
+import java.nio.file.Path;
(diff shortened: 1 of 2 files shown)


java/server/src/org/openqa/selenium/remote/server/commandhandler/UploadFile.java
```diff
@@ -60,8 +64,20 @@ public void execute(HttpRequest req, HttpResponse resp) throws IOException {
           "Expected there to be only 1 file. There were: " +
           (allFiles == null ? 0 : allFiles.length)));
     } else {
+      // IE requires multiple uploads to all be from the same directory - move files into one location
+      Path uploads_dir = Files.createDirectories(tempDir.toPath().getParent().resolve("remote_uploads"));
```
shs96c: Nit: Java naming conventions are CamelCase: `uploadsDir` instead of `uploads_dir`

java/server/src/org/openqa/selenium/remote/server/commandhandler/UploadFile.java
```diff
@@ -60,8 +64,20 @@ public void execute(HttpRequest req, HttpResponse resp) throws IOException {
...
+      Path dest = uploads_dir.resolve(allFiles[0].getName());
+
+      if (Files.exists(dest)){
+        // IE maintains a lock on previously uploaded files so we can't delete/overwrite - move it instead
```
shs96c: This is a great comment. That seems unusual, but it's great to have this explanation.

java/server/src/org/openqa/selenium/remote/server/commandhandler/UploadFile.java
```diff
@@ -60,8 +64,20 @@ public void execute(HttpRequest req, HttpResponse resp) throws IOException {
...
+        Path prevFileDir = session.getFileSystem().createTempDir("prevupload", "file").toPath();
+        Files.move(dest, prevFileDir.resolve(allFiles[0].getName()));
```
shs96c: Of course, you're totally screwed if someone tries to upload a file with the same name _again_. Why not just pick a random file name?

twalpole: We can't pick a random file name because the file name is sent to the browser. There are tests that confirm it's the same as the name of the file selected by the user.

There isn't an issue of someone trying to upload the same file twice in different uploads because of the previous `move`, and if they try to upload the same file name twice in the same multiple upload then those would have to have been the same file since IE requires the files to be from the same directory so no issue.  On other browsers if 2 files with the same name from different directories were selected then there would be an issue.  If that's a reasonable concern then this code could be only used when the browser is IE (if we actually have access to that info at this point in the code).

diemol:
@twalpole, would you like to continue with this PR?

diemol:
I will close this one as we did not get more comments, we can always reopen if we want to merge this.

PR #6856, opened 2019-01-16
Update domcore.js

codedieum:
Add sanity check making sure that called functions getAttribute and getAttributeNode are defined on the node before calling them.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6856)

### DIFF
--- a/javascript/atoms/test/attribute_test.html
+++ b/javascript/atoms/test/attribute_test.html
@@ -91,6 +91,13 @@ limitations under the License.
       assertAttributeEquals(bot.dom.getAttribute(e, 'name'), e, 'NaMe');
     }
 
+    function testUndefinedGetAttributeAndGetAttributeNode() {
+      var e = bot.locators.findElement({id: 'checky'});
+      e.getAttribute = null;
+      e.getAttributeNode = null;
+      assertAttributeEquals(null, e, 'id');
+    }
+
     function assertAttributeEquals(expected, elem, attr) {
       assertEquals('Expected attribute "' + attr + '" to equal "' +
           expected + '"', expected, bot.dom.getAttribute(elem, attr));


p0deje:
Which browser doesn't support `getAttribute` and/or `getAttributeNode`?

mostafafahim:
> Which browser doesn't support `getAttribute` and/or `getAttributeNode`?

I experience problems with chromium, I don't think they are not supported but they may be undefined on the element and so calling them without checking throws an exception.

barancev:
Is there a reproduction case when these methods are not defined on an element? We'd like to have a test for this case if it is possible.

mostafafahim:
So by setting element.getAttribute or element.getAttributeNode to null, it should trigger an exception calling them. Added a test

barancev:
Such a test looks very innatural :)
Is there a real-life case? If there is one -- we should check other places where we use getAttribute and getAttributeNode functions in Selenium code.

mostafafahim:
Indeed it is, but I just wanted to produce a possible scenario. For context,  I only encounter it when running puppeteer-like flow so the browser is in headless mode. I try to reproduce it in desktop chrome but actually can't. May be the element gets stale or something but what I am sure of is that checking first prevents the exception. It is also not bad to do the sanity check, no?

barancev:
@mostafafahim Of course there is nothing wrong with this PR, I just want to generalize the fix. If getAttribute or getAttributeNode is undefined in some cases we'd like to have tests for this case that cover various places where we use these methods.

mostafafahim:
Makes sense! so how do you suggest proceeding ?

barancev:
I hoped you know how to reproduce the issue :)

mostafafahim:
After I digged deeper it turned out I was sending bogus data to the function, my fault then. Thanks for insisting :)

Issue #4958, opened 2017-10-25
Undisposed ChromeDriverService and browser process on exception in RemoteWebDriver constructor


slawomir-brzezinski-at-interxion:
## Meta -
OS:  
Windows 10
Selenium Version:  
3.6.0

Browser:  
Chrome

Browser Version:  
irrelevant

## Expected Behavior -
[This](https://github.com/SeleniumHQ/selenium/blob/fa067c7/dotnet/src/webdriver/Chrome/ChromeDriver.cs#L119) and [this](https://github.com/SeleniumHQ/selenium/blob/fa067c7/dotnet/src/webdriver/Chrome/ChromeDriver.cs#L77) constructor create the `IDisposable` `ChromeDriverService`, so, if any exception happens later on in the constructor, it should be disposed.

## Actual Behavior -
Any problem during the start of the session (in our case, intermittent occurrence of [this ChromeDriver problem](https://github.com/SeleniumHQ/selenium/issues/4897)) causes the browser process to be orphaned and reserve resources such as port and memory.

diemol:
Is this issue still valid for the latest version?

diemol:
Closing, if the issue persists after trying with the most recent versions, feel free to open a new one.

PR #6872, opened 2019-01-22
Rethrow exceptions when RemoteSession start fails

andrewdodd13:
Currently, if a remote session fails to start, then Optional.empty() is
returned and the system behaves as expected. However, RemoteSession does
not make a call to service.stop(), and ServicedSession is unaware that
an exception has occurred, so it also doesn't make this call.

Thus, in the case that the Remote Session fails, there will can
lingering processes (ex. chromedriver) on the remote host. If running in
a high-traffic environment, then these lingering processes can
eventually starve the host of resources.

Fixes lingering processes on #4958

----

More context: We had an issue when running the Selenium server under load that the Chromedriver would time out on startup, and the HTTP Client would fail to connect. This was in an issue in and of itself, but the problem should have gone away. Instead, we noticed that we had thousands of chromedriver processes lingering on our host. After a bit of debugging, we noticed that if RemoteSession's HTTP Client throws an exception, there's no call to service.stop() - the method responsible for tearing down the processes.

By rethrowing as an IllegalStateException, the exception is trapped safely by ServicedSession (we believe this to be the only place that this changed method is used), which then calls service.stop() and returns Optional.empty() as RemoteSession would have done).

----

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6872)


diemol:
@andrewdodd13, would you like to continue with this PR?

andrewdodd13:
Hi @diemol. This is something that we are still running with our local copy patched with this change. We're still on 3.14, and probably won't upgrade any time soon (this change is not the blocker), but it doesn't look like there are major conflicts so I've merged the change onto the latest branch.

I tied one open issue to this (although it is in C# not Java); I couldn't see any one else complaining about the same thing. I didn't update any tests when I made this change so I'm not sure if there will be any further problems elsewhere.

diemol (commented):
@andrewdodd13, could you please sign the CLA?

diemol:
@andrewdodd13 I will close this PR since it seems there are different users mixed in the commits, if you still want to get it merged, please feel free to send a new one and sign the CLA, I'll get it merged right away.

PR #6877, opened 2019-01-24
Increase height used in Java window size tests

JohnChen0:
A recent change in Chrome (https://crrev.com/624785) increased the
minimum window height allowed on Mac to be about 300. This causes some
window size tests to fail (e.g., https://crbug.com/chromedriver/2748).
Increasing the heights to be above 300.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6877)


p0deje:
Thank you!

PR #6906, opened 2019-02-01
[C-py] Add `interval` option and shorten sleep interval for waiting until connectable

### LINKED ISSUE #6848: [Suggestion] Starting a service in Python bindings is slow
## Question

In my opinion, retrying interval to start the Service in Python bindings is a little too long.
We have to wait at least one second.
So that, it's good that shorten the retrying interval.
Does anyone have a good idea?

## Motivation
As you see in the below, we have to wait 1 second because of `time.sleep(1)` to start the Service.
I think it's so long.

https://github.com/SeleniumHQ/selenium/blob/11c25d75bd7ed22e6172d6a2a795a1d195fb0875/py/selenium/webdriver/common/service.py#L131-L135
https://github.com/SeleniumHQ/selenium/blob/11c25d75bd7ed22e6172d6a2a795a1d195fb0875/py/selenium/webdriver/common/service.py#L97-L104

iwasa-kosui:

Fix #6848

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6906)

### DIFF
--- a/py/selenium/webdriver/common/service.py
+++ b/py/selenium/webdriver/common/service.py
@@ -99,8 +99,8 @@ class Service(object):
             if self.is_connectable():
                 break
             count += 1
-            time.sleep(1)
-            if count == 30:
+            time.sleep(0.2)
+            if count == 150:
                 raise WebDriverException("Can not connect to the Service %s" % self.path)
 
     def assert_process_still_running(self):
@@ -128,11 +128,11 @@ class Service(object):
         except URLError:
             return
 
-        for x in range(30):
+        for x in range(150):
             if not self.is_connectable():
                 break
             else:
-                time.sleep(1)
+                time.sleep(0.2)
 
     def stop(self):
         """


EbiEbiEvidence:
@cgoldberg Thank you for your comment in #6848.
If you like, can I get some advices?
It's serious problem for me.

GQAssurance:
I like the change! 
However; would it make sense in `service.py` to a similar "seconds / interval" style you have used in `firefox_binary.py`?
That would make it more clear that the total time we are waiting is still 30 seconds
something like: `for x in range(30 / 0.2):`

EbiEbiEvidence:
> I like the change!
> However; would it make sense in `service.py` to a similar "seconds / interval" style you have used in `firefox_binary.py`?
> That would make it more clear that the total time we are waiting is still 30 seconds
> something like: `for x in range(30 / 0.2):`

That's right.
I'll fix it.

AutomatedTester:
Looking at this there is a problem where we can run out of Ephemeral port. Have you used this patch in anger?

GQAssurance:
> Looking at this there is a problem where we can run out of Ephemeral port. Have you used this patch in anger?

Why do you believe this?
The is_connectable function closes its sockets.

AutomatedTester:
> > Looking at this there is a problem where we can run out of Ephemeral port. Have you used this patch in anger?
> 
> Why do you believe this?
> The is_connectable function closes its sockets.

It tells the OS to close the port but there are numerous cases where OSX and Windows will not fully release the port and this can lead to ephemeral port exhaustion. This is a real problem that happens.

AutomatedTester:
Also, can you sign the CLA

GQAssurance:
If this change would exhaust the ports on a system then it sounds like it should be rejected, even without the CLA.

cgoldberg:
> The is_connectable function closes its sockets.

still.. you can't rebind immediately, so could possibly exhaust available sockets if you set the polling interval too low for the system.  Could this happen in practice at that .02/sec level with most `ulimit`'s?

AutomatedTester:
Closing this as I don't think that this is could lead to port exhaustion

PR #6916, opened 2019-02-06
add async version of until() and until_not()

skidzo:

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6916)

### DIFF
--- a/py/selenium/webdriver/support/wait_async.py
+++ b/py/selenium/webdriver/support/wait_async.py
@@ -1,6 +1,7 @@
 import time
 import asyncio as aio
 from selenium.webdriver.support.wait import WebDriverWait
+from selenium.common.exceptions import TimeoutException
 
 
 class AsyncWebDriverWait(WebDriverWait):


cgoldberg:
Can you share a use case or provide some context with your PR?

lmtierney:
This isn't compatible with `python 2.7`

skidzo:
providing a use case is dificult, but imagine an asyncronous backend behind your frontend, and you don't always want to provide callbacks, but instead await until an element appears, e.g.:

```
    async def until_found(self, name,
        by_func=None,
        timeout=20.,
        poll_freq=poll_frequency):
        """
        calls by_func() until its tag appears in the driver (and the call succeeds)
        """
        return await WebDriverWait(self.driver,
            timeout,
            poll_frequency=poll_freq).until(lambda element: by_func(element,
                name))
```

skidzo:
I dont remember how to write code for python 2.7, the problem is, there is no async in python 2.7. It is only going to be not compatible (fail) until you make an import on it, so you could just put it into another file async_wait.py ?
My point of view:
I don't want to keep a copy of this class in my codebase, just because of those four lines, while the rest could remain unchanged...

Issue #6918, opened 2019-02-07
Large grids (~30 nodes) can encounter lock-ups and NullPointerExceptions when registering nodes


gsfraley:
## 🐛 Bug Report

When a bunch of concurrent requests to /grid/register are made, there is a chance that the following exception will occur:
```
java.lang.NullPointerException
	at java.base/java.util.TreeMap.rotateRight(TreeMap.java:2240)
	at java.base/java.util.TreeMap.fixAfterInsertion(TreeMap.java:2272)
	at java.base/java.util.TreeMap.put(TreeMap.java:580)
	at org.openqa.selenium.AbstractCapabilities.setCapability(AbstractCapabilities.java:98)
	at org.openqa.selenium.MutableCapabilities.setCapability(MutableCapabilities.java:100)
	at org.openqa.grid.internal.utils.configuration.GridNodeConfiguration.lambda$fixUpCapabilities$12(GridNodeConfiguration.java:402)
	at java.base/java.util.stream.ReferencePipeline$11$1.accept(ReferencePipeline.java:441)
	at java.base/java.util.stream.ReferencePipeline$11$1.accept(ReferencePipeline.java:442)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1654)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
	at org.openqa.grid.internal.utils.configuration.GridNodeConfiguration.fixUpCapabilities(GridNodeConfiguration.java:410)
	at org.openqa.grid.common.RegistrationRequest.<init>(RegistrationRequest.java:92)
	at org.openqa.grid.common.RegistrationRequest.<init>(RegistrationRequest.java:59)
	at org.openqa.grid.common.RegistrationRequest.<init>(RegistrationRequest.java:48)
	at org.openqa.grid.common.RegistrationRequest.fromJson(RegistrationRequest.java:123)
[...]

PR #6924, opened 2019-02-11
Remove unsafe concurrent access of GridNodeConfiguration.DEFAULT_CONFIG_FROM_JSON

gsfraley:
The code before this fix concurrently accessed the capabilities in GridNodeConfiguration.DEFAULT_CONFIG_FROM_JSON.  Because its capabilities (AbstractCapabilities) uses a TreeMap, which isn't concurrent-safe, it would regularly deadlock grids of 20 nodes or more in either GridNodeConfiguration::new or GridNodeConfiguration::fixUpCapabilities.

This fix does two things:
1. Allows deep copies of NodeJsonConfiguration, so that a config can be copied and both instances can be accessed in parallel without issue
2. Limits access to the default config to a synchronized block that gives out deep copies of the original instance

This fix is meant to address concerns raised in https://github.com/SeleniumHQ/selenium/issues/6918.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6924)

### DIFF
--- a/java/server/src/org/openqa/grid/internal/utils/configuration/GridNodeConfiguration.java
+++ b/java/server/src/org/openqa/grid/internal/utils/configuration/GridNodeConfiguration.java
@@ -60,6 +60,12 @@ public class GridNodeConfiguration extends GridConfiguration {
       this.port = port;
     }
   }
+  
+  private static NodeJsonConfiguration getDefaultConfigFromJson() {
+	  synchronized(DEFAULT_CONFIG_FROM_JSON) {
+		  return new NodeJsonConfiguration(DEFAULT_CONFIG_FROM_JSON);
+	  }
+  }
 
   private HostPort hubHostPort;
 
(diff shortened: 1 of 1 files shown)


KroArtem:
@shs96c , @barancev , any chances this PR will be merged?

diemol:
I'll have a look at the PR later today, and if it is all good I'll merge it, at least to have code correctness.

What I am not sure is if we will be doing one release more, the idea was that 3.141.59 was the last one of the 3.x series. That can be decided by @shs96c and @barancev

gsfraley:
Thanks.  If it does end up being merged, and you do want to do another release, I can rebase onto the 3.141.59 tag with any suggested changes.  IIRC, it went without a hitch (no major changes) last time I tried.

diemol:
I tested it and it does fix the issue, thanks @gsfraley for your contribution and helping us to leave the code in a correct state. Nevertheless, as I mentioned, I don't think we will release this since it does not seem to affect lots of people.

I guess you already know how to generate you own jar if you want to use this commit:
```
./go selenium-server-standalone
```

Merged via https://github.com/SeleniumHQ/selenium/commit/d47e74d6f220efaed486ed89dfcb97e54683ad9e

PR #6937, opened 2019-02-14
Attempt UTF-8 encoding in logger formatter

GabrielMahan:
While running specs with Capybara and Selenium I've been experiencing a stream of `log writing failed. "\xC2" from ASCII-8BIT to UTF-8` errors; not sure of all the context, but adding this encoding to the formatter seems to be a fix--any reason not to add this to the formatter? 

image

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6937)

### DIFF
--- a/rb/lib/selenium/webdriver/common/logger.rb
+++ b/rb/lib/selenium/webdriver/common/logger.rb
@@ -122,7 +122,7 @@ module Selenium
         logger.progname = 'Selenium'
         logger.level = default_level
         logger.formatter = proc do |severity, time, progname, msg|
-          "#{time.strftime('%F %T')} #{severity} #{progname} #{msg}\n"
+          "#{time.strftime('%F %T')} #{severity} #{progname} #{msg.try(:force_encoding, 'UTF-8')}\n"
         end
 
         logger


p0deje:
Oh, this is weird. Is it possible for you to narrow down your test to understand where this message gets broken? I think it would be great to understand what the particular issue is.

I think your changes make sense, but they are impossible in pure Ruby - there is no `String#try` method. You're using Selenium something that requires ActiveSupport extensions which adds `#try` method. In order to make this PR mergeable, you need to make it work without ActiveSupport.

rb/lib/selenium/webdriver/common/logger.rb
```diff
@@ -122,7 +122,7 @@ def create_logger(output)
         logger.progname = 'Selenium'
         logger.level = default_level
         logger.formatter = proc do |severity, time, progname, msg|
-          "#{time.strftime('%F %T')} #{severity} #{progname} #{msg}\n"
+          "#{time.strftime('%F %T')} #{severity} #{progname} #{msg.try(:force_encoding, 'UTF-8')}\n"
```
luke-hill: I've been working in this area (Encoding), a lot recently. Can you try the following code.

```
def utf8?
  Encoding.default_internal == Encoding::UTF_8 &&
    Encoding.default_external == Encoding::UTF_8
end

# Then in your block
msg = msg.force_encoding('UTF-8') if msg.respond_to?(:force_encoding)
```

Now to the caveats. ASCII-8BIT is a useful encoding char-set because it allows things like german umlauts and other stuff. I actually have the following code in my new API tests I write

```
# file - env.rb
# Set these to 8bit to allow umlauts
Encoding.default_external = Encoding::ASCII_8BIT
Encoding.default_internal = Encoding::ASCII_8BIT
```

What is your current `Encoding.default_external / internal` values, could it be your system is somehow setting up some weird ones?

p0deje:
I've been trying to reproduce the issue by sending non-ASCII-8bit characters to it, but they are parsed perfectly fine, so I'm not sure if this issue is still present in the recent version of selenium-webdriver.

Since there was no response for a few months, I'm closing this PR. I'll be happy to continue the conversation and re-open if somebody replies.

codyrobbins:
This is an old issue but I believe the problem described here is still present.

For me it happens when setting `Selenium::WebDriver.logger.level` to `:debug` and the [`page_source`](https://github.com/SeleniumHQ/selenium/blob/trunk/rb/lib/selenium/webdriver/remote/bridge.rb#L151-L153) method is called. If the page contains a non-ASCII character, such as `×`, the logging fails and results in the message:

```
log writing failed. "\xC3" from ASCII-8BIT to UTF-8
```

This happens even when `Encoding.default_external` and `Encoding.default_internal` are both set to `Encoding::UTF_8`.

codyrobbins:
If you google for this problem there are tons of unrelated projects that have had the same issue (cf. https://github.com/elastic/elasticsearch-ruby/issues/71), so it was difficult for me to localize the issue to Selenium in my case. But using `force_encoding` looks like it seems to be the fix.

luke-hill:
Hi @codyrobbins -

I have a small gem that will patch this and remove this problem

https://github.com/site-prism/automation_helpers/blob/main/lib/automation_helpers/patches/selenium_logger.rb

To use it simply do

```
  require 'automation_helpers/patches'
  AutomationHelpers::Patches::SeleniumLogger.new.patch!
```

The issue I found was actually further down to the issues in RubyMine and that the issues in Ruby 3.2+ are now localised to just rubymine

Issue #6873, opened 2019-01-22
Add support for W3C "New Window" command


whimboo:
## 🚀 Feature Proposal

At the last TPAC it was decided to add a new command called "New Window" which allows tests to open a new tab or window. The related work can be seen at https://github.com/w3c/webdriver/issues/1138.

All drivers will have support for this probably soon, so it would be good to get support added in all the Selenium bindings.

As of now the command supports two types of browsing contexts to open, which are `tab` and `window`. If nothing gets passed in the browser will decide the best type on its own. This also happens when the requested type is not supported, eg. new browser window on mobile.

## Example

Please provide an example for how this feature would be used.

```python
handle, _type = driver.new_window(type_hint="tab")
driver.switch_to(handle)
```

More WebDriver tests which are part of the web-platform-tests suite can be found at https://github.com/web-platform-tests/wpt/tree/master/webdriver/tests/new_window.

whimboo:
@barancev mind adding labels for all bindings? Thanks.

whimboo:
Could anybody from the Selenium community suggest how the API should look like? The command returns a dictionary like the following:

> {u'type': u'tab', u'handle': u'8589934593'}

It's most likely not appropriate to return both but only the `handle` in the client's method. Type would only be interesting if a specific type of top-browsing context has been requested, but a different one has been created by the browser due to eg. system limitations. Should the client method handle such a case and raise an error eg. when getting a `strict_type` parameter?

jimevans:
@whimboo The initial implementation in the .NET bindings returns neither. In order to remain consistent with other methods in the `ITargetLocator` interface, the driver switches the command context to the newly opened window and returns a reference to the current driver. Thus making the usage as follows:

    // Note that focusedDriver and driver are the same object. This construction
    // was used simply to show the return type. 
    IWebDriver focusedDriver = driver.SwitchTo().NewWindow(WindowType.Tab);
    string newWindowHandle = driver.CurrentWindowHandle;

One could argue with this approach, and it does obscure the details of the wire protocol, but from the Selenium usage point of view, I believe API consistency concerns are more important. I suspect the Java language bindings will take a similar approach. For Python and Ruby, which largely eschew the role-based interface concept, maintainers may elect a different approach.

jimevans:
I'm reopening this, as the code needs to be added to Java, Ruby and JS language bindings.

whimboo:
[...]

PR #6946, opened 2019-02-19
[rb] Add new_window w3c command

dylanlive:
Adds support for the [w3c new_window](https://w3c.github.io/webdriver/#new-window) command for the Ruby bindings.

Addresses https://github.com/SeleniumHQ/selenium/issues/6873

This is a rather raw implementation, which you can do:
```
driver.new_window('tab')
driver.new_window('window')
```
and get back a hash with the window handle.

I reviewed this suggested implementation: https://github.com/SeleniumHQ/selenium/issues/6873#issuecomment-459109624. Maintainers, if you'd prefer the SwitchTo format, please let me know and I can look to tweak.

## Current Browser Support
As far as I know:
- Safari Technology Preview v74
- Firefox 66 (Beta)

Due to the limited support, I've made test exclusions to only Safari Technology Preview, as I don't believe the build system uses the beta Firefox.

## Notes
- This is not yet implemented in java, so any remote tests will likely fail
- First commit in the rb selenium repo - I assumed documentation should be in YARD format as I saw a reference to yard in rakefile

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6946)

### DIFF
--- a/rb/lib/selenium/webdriver/remote/w3c/bridge.rb
+++ b/rb/lib/selenium/webdriver/remote/w3c/bridge.rb
@@ -118,6 +118,20 @@ module Selenium
                               'return source;')
           end
 
+          #
+          # Create a new top-level browsing context
+          # https://w3c.github.io/webdriver/#new-window
+          # @param type [String] Supports two values: 'tab' and 'window'.
+          #  Use 'tab' if you'd like the new window to share an OS-level window
+          #  with the current browsing context.
+          #  Use 'window' otherwise
+          # @return [Hash] Containing 'handle' with the value of the window handle
+          #  and 'type' with the value of the created window type
+          #
+          def new_window(type)
+            execute :new_window, {}, {type: type}
+          end
+
           def switch_to_window(name)
             execute :switch_to_window, {}, {handle: name}
           end
--- a/rb/lib/selenium/webdriver/common/options.rb
+++ b/rb/lib/selenium/webdriver/common/options.rb
@@ -106,6 +106,30 @@ module Selenium
       end
 
       #
+      # Create a new top-level browsing context
+      # https://w3c.github.io/webdriver/#new-window
+      # @param type [Symbol] Supports two values: :tab and :window.
+      #  Use :tab if you'd like the new window to share an OS-level window
+      #  with the current browsing context.
+      #  Use :window otherwise
+      # @return [String] The value of the window handle
+      #
+      def new_window(type = :tab)
+        case type
+        when :tab, :window
+          result = @bridge.new_window(type)
+          if !result.has_key?('handle')
+            raise UnknownError, "the driver did not return a handle. " +
+              "The returned result: #{result.inspect}"
+          end
+          return result['handle']
+        else
+          raise ArgumentError, "invalid argument for type. Got: '#{type.inspect}'." +
+            " Try :tab or :window"
+        end
+      end
+
+      #
       # @api beta This API may be changed or removed in a future release.
       #
 
--- a/rb/lib/selenium/webdriver/remote/w3c/commands.rb
+++ b/rb/lib/selenium/webdriver/remote/w3c/commands.rb
@@ -53,6 +53,7 @@ module Selenium
             #
 
             get_window_handle: [:get, 'session/:session_id/window'],
+            new_window: [:post, '/session/:session_id/window/new'],
             close_window: [:delete, 'session/:session_id/window'],
             switch_to_window: [:post, 'session/:session_id/window'],
             get_window_handles: [:get, 'session/:session_id/window/handles'],
--- a/rb/spec/integration/selenium/webdriver/options_spec.rb
+++ b/rb/spec/integration/selenium/webdriver/options_spec.rb
@@ -101,6 +101,30 @@ module Selenium
           expect(actual).to eq(expected)
         end
       end
+
+      describe 'new_window', only: {browser: %i[safari_preview firefox ie]} do
+        types = [:tab, :window] # Stores the valid type values for new_window
+        types.each do |type|
+          it "should be able to open a new #{type}" do
+            # Store the count of window handles before executing action
+            before_window_handles = driver.window_handles.length
+            # Open a new window
+            driver.manage.new_window(type)
+            # Store the count of window handles after executing action
+            after_window_handles = driver.window_handles.length
+            # 1 window handle should have been added
+            expect(after_window_handles).to eq(before_window_handles + 1)
+          end
+        end
+
+        it "returns an exception if an invalid type is provided" do
+          invalid_types = [:invalid, 'invalid', 'tab', 'window']
+          invalid_types.each do |type|
+            expect { driver.manage.new_window(type) }.to \
+              raise_error(ArgumentError, "invalid argument for type. Got: '#{type.inspect}'. Try :tab or :window")
+          end
+        end
+      end
     end # Options
   end # WebDriver
 end # Selenium


dylanlive:
@titusfortner @lmtierney @p0deje @barancev @twalpole @shs96c Feel free to provide feedback, re:
>Maintainers, if you'd prefer the SwitchTo format, please let me know and I can look to tweak.

rb/lib/selenium/webdriver/remote/w3c/bridge.rb
```diff
@@ -116,6 +116,20 @@ def page_source
...
 
+          #
+          # Create a new top-level browsing context
+          # https://w3c.github.io/webdriver/#new-window
+          # @param type [String] Supports two values: 'tab' and 'window'.
+          #  Use 'tab' if you'd like the new window to share an OS-level window
+          #  with the current browsing context.
+          #  Use 'window' otherwise
+          # @return [Hash] Containing 'handle' with the value of the window handle
+          #  and 'type' with the value of the created window type
```
twalpole: Can this ever return a "window" of the type other than requested?  If not what's the point in returning the type?

dylanlive: I don't believe so. This is the returned result per the webdriver spec:
https://w3c.github.io/webdriver/#new-window
```
8. Let result be a new JSON Object initialised with:

"handle"
The value of handle.
"type"
The value of type.
```

rb/lib/selenium/webdriver/remote/w3c/bridge.rb
```diff
@@ -116,6 +116,20 @@ def page_source
                               'return source;')
           end
 
+          #
+          # Create a new top-level browsing context
+          # https://w3c.github.io/webdriver/#new-window
+          # @param type [String] Supports two values: 'tab' and 'window'.
```
p0deje: I believe it should also allow for Symbol params:

```ruby
driver.new_window(:tab)
driver.new_window(:window)
```

rb/spec/integration/selenium/webdriver/driver_spec.rb
```diff
@@ -299,6 +299,22 @@ def save_screenshot_and_assert(path)
           end.to raise_error(Selenium::WebDriver::Error::ScriptTimeoutError)
         end
       end
+
+      describe 'window' do
```
p0deje: Please move it to `./rb/spec/integration/selenium/webdriver/window_spec.rb`

rb/spec/integration/selenium/webdriver/driver_spec.rb
```diff
@@ -299,6 +299,22 @@ def save_screenshot_and_assert(path)
...
+        types = ['tab', 'window'] # Stores the valid type values for new_window
+        types.each do |type|
+          it "should be able to open a new #{type}", only: {browser: %i[safari_preview]} do
```
p0deje: The support landed on IE recently (e8d326e2294c7681ba03d274e1bd37f3cc636594), so we should include it too.

rb/lib/selenium/webdriver/remote/w3c/bridge.rb
twalpole: @p0deje Why do you want it to automatically switch? That would provide less flexibility for the ruby libraries that use selenium (watir/capybara/etc)

p0deje: I'm not strongly supporting that, my arguments are mostly the same as Jim explained in [his comment](https://github.com/SeleniumHQ/selenium/issues/6873#issuecomment-459109624) and I'd prefer to keep consistency with other bindings (only C# supports it though).

What kind of flexibility wrappers need here? I can imagine a normal use case where a new window is created and context is switched to it immediately. What do I miss?

dylanlive: 👍 yes it does. I'll update documentation

rb/spec/integration/selenium/webdriver/driver_spec.rb
dylanlive: By moving it to the .switch_to API, wouldn't the test be located in:
`spec/integration/selenium/webdriver/target_locator_spec.rb`

rb/spec/integration/selenium/webdriver/driver_spec.rb
dylanlive: Nice!

rb/lib/selenium/webdriver/remote/w3c/bridge.rb
dylanlive: re:
>Why do you want it to automatically switch?

I agree consistency would be nice. Python also uses the switch_to format.

Would it be plausible to support both:
`driver.new_window(:tab)`
and
`driver.switch_to.new_window(:tab)`

p0deje: Another possible API might look like this:

```ruby
# you can do this
window_handle = driver.manage.new_window
driver.switch_to.window(window_handle)

# or you can do this
driver.switch_to.new_window
```

We already have window-related API in `driver.manage`, so providing 2 different ways to create new window might suit everyone.


@dylanlive Let's hear back from @twalpole before deciding which one to implement.

rb/spec/integration/selenium/webdriver/driver_spec.rb
p0deje: You're right, that would be a better place.

rb/lib/selenium/webdriver/remote/w3c/bridge.rb
twalpole: The current window switching API is `driver.switch_to.window(handle)`, so if the new window API is only implemented via `driver.switch_to.new_window` then to create a new window and not switch to I have to do

    orig_handle = driver.window_handle
    driver.switch_to.new_window
    driver. switch_to.window(orig_handle)

which seems like extra overhead for something that should be relatively easy.  If the API in the current PR is used then to open a new window without switching it is 

    driver.new_window(:window)  # or driver.manage.new_window 

and to open and switch it's

    driver.switch_to.window(driver.new_window(:window)[:handle])

Adding the `driver.switch_to.new_window` as a shortcut for opening and switching would also be great, but implementing just that (and losing the ability to open without switching would mean Capybara resorting to calling another method directly on bridge (which it already does more times than I'd like)

p0deje: @twalpole Ok, but out of curiosity, what's the use case for opening a new window but not switching to it immediately?

twalpole: I don’t personally have one currently, but nothing else currently auto-switches context and I see no benefit to  preventing the creation without switching. A potential use case I could imagine is creating a pool of windows before a test and then using them in the tests, but that is purely hypothetical

p0deje: Alright, got it. It would be great to hear other maintainers opinions (cc @titusfortner @lmtierney). My call would be to prefer consistency across other language bindings over API flexibility, but maybe I'm just being too conservative here.

dylanlive: @titusfortner @lmtierney Bump - any thoughts? I have a local version moving to the SwitchTo api that mimics other bindings.

I propose we mimic the other bindings, but I also think it'd be cool to have both options:
1. Automatically create a new window and switch using SwitchTo
2. Allow creating a new window, but not automatically switching to (providing flexibility for other frameworks)

titusfortner:
I don't think it should automatically switch. It should create it and return the handle. 

The difference with Ruby over other Selenium languages is that we have Capybara & Watir to provide additional usability on top of the baseline toolkit implementation. I don't want Selenium code to obfuscate what is being done or combine multiple things.

rb/lib/selenium/webdriver/remote/w3c/bridge.rb
lmtierney: If we can do both I'd prefer that. Better to have flexibility, especially since the majority of users use a wrapping library. There are plenty of places ruby is different from other bindings so I wouldn't worry about that, this is at least similar

titusfortner: Gah wrong button

titusfortner: Gah wrong button

titusfortner: Once we remove OSS backwards compatible hacks we won't have any code that makes multiple wire calls. I think watir / Capybara should decide if there's a more user friendly method required. I'd rather selenium not start supporting convenience methods.

luke-hill: Could you not just add an extra default key to the input param

`driver.new_window(:window, switch: true)` And have it defaulted to whichever one is most desired / used in other bindings?

p0deje: Ok, so after talking to everyone, let's do what @twalpole and @titusfortner propose.

```ruby
driver.manage.new_window # defaults to tab
driver.manage.new_window(:tab)
driver.manage.new_window(:window)
```

it needs to return window handle that can be passed to `driver.window.switch_to`.

twalpole: @dylanlive Do you have time to make this change in the near future?

dylanlive: Sounds good to me 👍 thanks all!

@twalpole I should have some time in the coming weeks

dylanlive:
[Per spec](https://w3c.github.io/webdriver/#new-window), currently the driver returns a hash with the handle and the type:
```
{"handle"=>"6442450945", "type"=>"tab"}
```

Per the requested requirement:
>it needs to return window handle that can be passed to `driver.window.switch_to`

Would you like just the handle value, i.e.,
```
[1] pry(main)> driver.manage.new_window
6442450945
```

Or keep the hash format returned by the browser driver?

p0deje:
I think returning only handle is enough at this point. Also, I believe the method should raise if argument is neither `tab` nor `window` so that `invalid` is not allowed.

rb/spec/integration/selenium/webdriver/driver_spec.rb
```diff
@@ -299,6 +299,22 @@ def save_screenshot_and_assert(path)
...
+            # Store the count of window handles before executing action
+            before_window_handles = driver.window_handles.length
+            # Open a new window
+            driver.new_window(type)
+            # Store the count of window handles after executing action
+            after_window_handles = driver.window_handles.length
+            # 1 window handle should have been added
+            expect(after_window_handles).to eq(before_window_handles + 1)
```
luke-hill: This test should probably also validate the type (If we're returning a full hash)

twalpole: Validating the type isn't really possible since it is driver implementation defined according to the webdriver spec. ie - just because :tab is passed in does not mean it will return :tab.

luke-hill: Do we "know" what the current plans are, if so could we partition for that. Or is that too complex a question?

twalpole: @luke-hill We have no way of knowing what any specific webdriver compatible product may return, and limiting the test for no real benefit (especially since the plan is to ignore the returned type currently) doesn't seem to make sense.

dylanlive:
**It looks like my branch readds the `.freeze` in `commands.rb` - I'll fix that up tonight, sorry!**

Pushed up an updated version with the following support:
```
driver.manage.new_window # defaults to tab
driver.manage.new_window(:tab)
driver.manage.new_window(:window)
```

Example:
```
[3] pry(main)> driver.manage.new_window(:window)
2019-03-23 19:58:51 INFO Selenium -> POST /session/957449fc-b0ef-e547-af6c-e55fe1f0eaa5/window/new
2019-03-23 19:58:51 INFO Selenium    >>> http://127.0.0.1:4444/session/957449fc-b0ef-e547-af6c-e55fe1f0eaa5/window/new | {"type":"window"}
2019-03-23 19:58:52 INFO Selenium <- {"value":{"handle":"10737418241","type":"window"}}
=> "10737418241"
```

Per suggestion, I raise an `ArgumentError` if they pass something that's not `:tab` or `:window`

dylanlive:
Fixed and should be ready for review 👍

rb/spec/integration/selenium/webdriver/driver_spec.rb
p0deje: Can you please add it? The tests are failing on CI because IE supports this command: https://ci.appveyor.com/project/SeleniumHQ/selenium/builds/23322684/job/myeyect1k0cu57vy#L661.

p0deje:
@dylanlive Looks good to me! Can just add support for IE in tests and squash commits into a single one?

dylanlive:
@p0deje Does IE Driver support the `new_window` command? I only see mention of Safari Tech Preview and Firefox here:
https://github.com/w3c/webdriver/issues/1138

Did just realize the test's conditional filter wasn't added to the second test case, so I'll add it to the `describe` block.

lmtierney:
@dylanlive Yep, it was added per the changelog here https://github.com/SeleniumHQ/selenium/blob/master/cpp/iedriverserver/CHANGELOG#L143

dylanlive:
@p0deje Added IE, squashed to a single commit

p0deje:
Thank you! Merged in 2960a3d01ed016900c1e097015ab3329e6bd65bc.

Issue #6907, opened 2019-02-01
Allow childprocess 1.0 dependency of ruby selenium-webdriver?


jrochkind:
Ruby gem selenium-webdriver has a dependency on `'childprocess', ['~> 0.5']`

https://github.com/SeleniumHQ/selenium/blob/d7674b0c2ffff75787e94aef13e4ceab04252a86/rb/selenium-webdriver.gemspec#L33

The childprocess gem had a [1.0 release](https://rubygems.org/gems/childprocess/versions/1.0.0) a couple days ago. That is not allowed by the selenium-webdriver dependency specification. 

Should it be?

twalpole:
Fixed via 6d7f6338c0c985403840c819652cb9cd1ebf22df

jrochkind:
thanks!

PR #6948, opened 2019-02-19
Allow childprocess 1.x - issue #6907

twalpole:

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6948)

### DIFF
--- a/rb/selenium-webdriver.gemspec
+++ b/rb/selenium-webdriver.gemspec
@@ -30,7 +30,7 @@ Gem::Specification.new do |s|
   s.require_paths = ['lib']
 
   s.add_runtime_dependency 'rubyzip', ['~> 1.2', '>= 1.2.2']
-  s.add_runtime_dependency 'childprocess', ['~> 0.5']
+  s.add_runtime_dependency 'childprocess', ['>= 0.5', '< 2.0']
 
   s.add_development_dependency 'rspec', ['~> 3.0']
   s.add_development_dependency 'rack', ['~> 1.0', '>= 1.6.11']


rb/Gemfile.lock
```diff
@@ -17,7 +17,6 @@ GEM
       safe_yaml (~> 1.0.0)
     diff-lcs (1.3)
     ffi (1.9.25)
-    ffi (1.9.25-x64-mingw32)
```
p0deje: Can you please revert this change? It is added when installing gems on Windows so, once merged, it will be added back on the next `bundle install` run on Windows.

rb/Gemfile.lock
```diff
@@ -71,4 +70,4 @@ DEPENDENCIES
   yard (~> 0.9.11)
 
 BUNDLED WITH
-   1.16.3
```
p0deje: Let's not bump bundler version since we ship 1.16.3 by default. Another approach would be to bump it in `third_party/rb'?

rb/Gemfile.lock
twalpole: Ok -- is there a specific reason Gemfile.lock is even checked in? or possibly added and then gitignored if it being updated is a problem?

p0deje: Frankly, I don't remember why exactly it was added. Probably because we used cached gems and if there is no Gemfile - it pulls more recent versions. I haven't checked that, so maybe we're safe to remove it altogether?

luke-hill: :point_up: Just for reference, this is the reason. I remember chatting with simon about this a while ago when I did a minor tweak

PR #6950, opened 2019-02-19
Bump minimum requirement to ruby 2.3 and fix rubocop violations

twalpole:

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6950)

### DIFF
--- a/rb/lib/selenium/webdriver/common/socket_poller.rb
+++ b/rb/lib/selenium/webdriver/common/socket_poller.rb
@@ -80,7 +80,7 @@ module Selenium
         sock.close
         true
       rescue *NOT_CONNECTED_ERRORS
-        sock.close if sock
+        sock&.close
         WebDriver.logger.debug("polling for socket on #{[@host, @port].inspect}")
         false
       end
--- a/rb/lib/selenium/webdriver/safari/service.rb
+++ b/rb/lib/selenium/webdriver/safari/service.rb
@@ -26,7 +26,7 @@ module Selenium
 
       class Service < WebDriver::Service
         DEFAULT_PORT = 7050
-        @executable = 'safaridriver'.freeze
+        @executable = 'safaridriver'
         @missing_text = <<-ERROR.gsub(/\n +| {2,}/, ' ').freeze
           Unable to find Apple's safaridriver which comes with Safari 10.
           More info at https://webkit.org/blog/6900/webdriver-support-in-safari-10/


p0deje:
I believe we should only bump the minimum required Ruby version with Selenium 4 and it's done in the branch: https://github.com/SeleniumHQ/selenium/pull/6444/commits/9c0ba0711e0999a1b79c0adcab3a37dfbd099aba. There are some differences between RuboCop changes we did (notable frozen string literal), so maybe let's discuss those on IRC/Slack?

twalpole:
> we should only bump the minimum required Ruby version with Selenium 4
What's your reasoning for this?  Even if selenium-webdriver had restricted itself to SemVer (which I don't believe it has) updating the versions of requirements in a minor release is completely acceptable.  2.2 went EOL a year ago, and 2.3 will be in a month so it's beyond time to update.  We can definitely discuss tomorrow on slack.

p0deje:
One of the reasons we were slow on upgrading to 2.3 was macOS support. Not everyone on Selenium team is Ruby developer and some folks (e.g. Simon in the past) would resort to using system Ruby, which was 2.0 until Mojave. The idea was to make sure commiters don't need to install anything if they need to run Ruby tests. Moving on, most of the folks should have upgraded to Mojave, so it should not be a big deal.

Another reason is that we had complaints in the past (especially from Windows users) that we used some features from Ruby 2.1+ and their code didn't work with recent selenium-webdriver on 2.0, so we locked everything to Ruby 2.0.

Selenium versioning is not SemVer, but changes which were not deprecated before should still only be done in major upgrades. If we want to do it in minor versions - we need to add deprecations first, wait for at least 1 more release and then change. That's how Selenium bindings typically handle updates which break backward compatibility.

twalpole:
> but changes which were not deprecated before

That makes complete sense for API changes in selenium-webdriver because it can break users code.  The difference with updating the version of required gems (and the ruby version) is that it doesn't break users code, as long as it's declared in the gemspec.  If their project has locked a version of a specific requirement (or is using a specific ruby version) the worst that happens is that bundler doesn't update to the latest version of selenium-webdriver until they release their gem requirements or update to a newer ruby.  Anyway lets discuss this all tomorrow on slack.

.travis.yml
```diff
@@ -21,7 +21,7 @@ _bindings:
     if: env(BUILD_ONLY) IS NOT present || env(BUILD_ONLY) =~ /python/
   ruby: &ruby
     language: ruby
-    rvm: 2.0.0-p648
+    rvm: 2.3.8
```
luke-hill: If in selenium4 we're going to up this. It feels a good time to introduce at least testing for higher versions such as 2.4 / 2.5 (I know historically selenium has never supported the top edge?

Thoughts @p0deje ?

p0deje: I'm not sure if Thomas and Titus agreed to release it before Selenium 4, but if they did, versions will be upgraded everywhere.

twalpole: @luke-hill the only issue with testing multiple versions is the way the test matrix is built - and we should be testing 2.6 too -- there is zero reason not to officially support it

rb/.rubocop.yml
```diff
@@ -34,47 +37,47 @@ Naming/FileName:
...
   Exclude:
-    - 'lib/selenium/webdriver/chrome/legacy_bridge.rb'
     - 'lib/selenium/webdriver/chrome/driver.rb'
     - 'lib/selenium/webdriver/remote/w3c/capabilities.rb'
 
 Metrics/CyclomaticComplexity:
   Max: 9
   Exclude:
-    - 'lib/selenium/webdriver/chrome/legacy_bridge.rb'
+    # - 'lib/selenium/webdriver/chrome/legacy_bridge.rb'
```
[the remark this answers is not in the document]
twalpole: yep - on it

rb/.rubocop.yml
```diff
@@ -1,5 +1,8 @@
+require:
```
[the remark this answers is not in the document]
twalpole: No issue -- will update and fix whatever pops up

PR #6956, opened 2019-02-20
[node] Adding support for opening a new window.

corevo:
Node implementation for opening a new window.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6956)

### DIFF
--- a/javascript/node/selenium-webdriver/lib/webdriver.js
+++ b/javascript/node/selenium-webdriver/lib/webdriver.js
@@ -1699,6 +1699,26 @@ class TargetLocator {
   }
 
   /**
+   * Creates a new browser window and switches the focus for future
+   * commands of this driver to the new window.
+   *
+   * @param {string} typeHint 'window' or 'tab'. The created window is not
+   *     guaranteed to be of the requested type; if the driver does not support
+   *     the requested type, a new browser window will be created of whatever type
+   *     the driver does support.
+   * @return {!Promise<void>} A promise that will be resolved
+   *     when the driver has changed focus to the new window.
+   */
+  newWindow(typeHint) {
+    var driver = this.driver_
+    return this.driver_.execute(
+        new command.Command(command.Name.SWITCH_TO_NEW_WINDOW).
+            setParameter('type', typeHint)).then(function(response) {
+          return driver.switchTo().window(response.handle);
+        });
+  }
+
+  /**
    * Changes focus to the active modal dialog, such as those opened by
    * `window.alert()`, `window.confirm()`, and `window.prompt()`. The returned
    * promise will be rejected with a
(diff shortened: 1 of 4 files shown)


javascript/node/selenium-webdriver/lib/webdriver.js
```diff
@@ -1698,6 +1698,23 @@ class TargetLocator {
...
+   * commands of this driver to the new window.
+   *
+   * @param {string} typeHint 'window' or 'tab'. The created window is not
+   *     guaranteed to be of the requested type; if the driver does not support
+   *     the requested type, a new browser window will be created of whatever type
+   *     the driver does support.
+   * @return {!Promise<void>} A promise that will be resolved
+   *     when the driver has changed focus to the new window.
+   */
+  newWindow(typeHint) {
```
corevo: Do we want this to be an enum like `Keys`?

javascript/node/selenium-webdriver/lib/webdriver.js
```diff
@@ -1698,6 +1698,23 @@ class TargetLocator {
...
+    return this.driver_.execute(
```
barancev: If we follow the same approach as implemented in C# and Java, we should wait until the new window appears and swicth into it.  So it should be something like this:
```
    var driver = this.driver_;
    return this.driver_.execute(
        new command.Command(command.Name.SWITCH_TO_NEW_WINDOW).
            setParameter('type', typeHint)
      ).then(function(response) {
          return driver.switchTo().window(response.handle);
      });
```

javascript/node/selenium-webdriver/test/window_test.js
```diff
@@ -99,6 +99,12 @@ test.suite(function(env) {
     return driver.wait(forPositionToBe(x, y), 1000);
   });
 
+  it('can open a new window', async function() {
+    await driver.switchTo().newWindow()
+
+    assert.equal((await driver.getAllWindowHandles()).length, 2)
+  })
```
barancev: ... and here we could check that we're in the new windows, that means the new window handle is not equal to the original one.

barancev (commented):
Let's keep bindings coherrent.

javascript/node/selenium-webdriver/lib/webdriver.js
whimboo: Just FYI... per the spec the `New Window` command is synchronous.

whimboo:
Anyone who can follow-up here with a review? @jleyba or @barancev?

barancev:
Merged into master as commit 59e5bf25a585a92239ccba557d8df0f20333e04f
(and improved by a followup commit bf2012368e5f094dab8b4ed8d2caa87ea2b3434e)

PR #6965, opened 2019-02-25
[rb] move driver responsibilities into Service class

titusfortner:
This tightens up what is allowed in the driver initialization:

Instead of this:
```
opts = {driver_path: '/path/to/driver', port: '1234', driver_opts: {log_path: '/path/to/log', verbose: true}}
Selenium::WebDriver.for :chrome, opts
```

You now do this:
```
opts = {path: '/path/to/driver', port: '1234', args: ['--log_path=/path/to/log', '--verbose']}
service = Selenium::WebDriver::Service.chrome(opts)
Selenium::WebDriver.for :chrome, service: service
```

1. `driver_opts` is deprecated and we require users to be explicit about `args` as an `Array` of `String` including the `--`
2. This code paves the way for removing all of the logic in all of the subclasses
3. Subclasses are kept as `@api private` because it might make sense to remove them entirely when we stop supporting deprecations

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6965)


twalpole:
I notice that we're now basically passing `chrome` twice `Service.chrome` and webdriver.for :chrome - we could have the Service instances contain the default driver name for that service so it could be simplified to `Selenium::WebDriver.for service: service`

titusfortner:
You can essentially already do that with:
```
Chrome::Driver.new(service: service)
```
1. I'd prefer people not use `#for` in the first place
2. If they are going to use `#for` let's not change the API on them
3. Capybara & Watir get to make it look prettier if necessary

twalpole:
But `Selenium::WebDriver::Chrome::Driver.new(service: service)` is also specifying Chrome twice

titusfortner:
You don't have to specify a :service parameter, driver will create one automatically if :url is not passed in, so we'd have to be doing Ruby magic to parse it.

I'm thinking of this as having `Service.new` be allowed with setting the parameters yourself, and then the class methods for the specific browser are helpers to provide the defaults. Though, I guess in that case I should remove the `@api private` from `class Service` if I want to think of it like that...

connorshea:
I tried updating my selenium configuration from the old `driver_opts` to the new service-based strategy, but it doesn't work quite right. The headless argument is being passed in anymore.

Before:
```ruby
Capybara.register_driver :headless_chrome do |app|
  capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
    chromeOptions: { args: %w[headless] }
  )

  Capybara::Selenium::Driver.new app,
    browser: :chrome,
    desired_capabilities: capabilities,
    driver_opts: {
      log_path: 'tmp/chrome.log'
    }
end
```

After:
```ruby
Capybara.register_driver :headless_chrome do |app|
  capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
    chromeOptions: { args: %w[headless] }
  )

  Capybara::Selenium::Driver.new app,
    browser: :chrome,
    desired_capabilities: capabilities,
    service: Selenium::WebDriver::Service.chrome(args: { log_path: 'tmp/chrome.log' })
end
```

Could I get some clarification on what I'm doing wrong here? Thanks :)

ndbroadbent:
@connorshea Thanks for your example with Capybara! I was struggling to figure this out. 

I'm passing the headless option under an "options" key. This is working for me:

```ruby
  Capybara.register_driver :headless_chrome do |app|
    capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
      acceptInsecureCerts: true,
      loggingPrefs: {
        browser: 'ALL',
        client: 'ALL',
        driver: 'ALL',
        server: 'ALL',
      }
    )
    options.headless!
    options.add_argument('--disable-gpu')
    options.add_argument('--disable-logging')
    options.add_argument('--window-size=1280,1000')

    chromedriver_log_path = Rails.root.join('log/chromedriver.log').to_s
    Capybara::Selenium::Driver.new(
      app,
      browser: :chrome,
      options: options,
      desired_capabilities: capabilities,
      service: Selenium::WebDriver::Service.chrome(
        args: ["--log_path=#{chromedriver_log_path}", '--verbose']
      )
    )
  end
  Capybara.javascript_driver = :headless_chrome
```



I actually use [puffing-billy](https://github.com/oesmith/puffing-billy) to record/replay front-end requests, so here's my full capybara driver:

```ruby
  # See: https://github.com/oesmith/puffing-billy/issues/193
  Capybara.register_driver :headless_chrome_billy do |app|
    capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
      acceptInsecureCerts: true,
      loggingPrefs: {
        browser: 'ALL',
        client: 'ALL',
        driver: 'ALL',
        server: 'ALL',
      }
    )
    options.headless!
    options.add_argument('--disable-gpu')
    options.add_argument('--disable-logging')

    # Stop automatic updates.
    options.add_argument('--safebrowsing-disable-auto-update')

    options.add_argument('--window-size=1280,1000')
    options.add_argument("--proxy-server=#{Billy.proxy.host}:#{Billy.proxy.port}")
    proxy_bypass_list = [
      '127.0.0.1',
      'localhost',
    ].join(';')
    options.add_argument("--proxy-bypass-list=#{proxy_bypass_list}")

    chromedriver_log_path = Rails.root.join('log/chromedriver.log').to_s
    Capybara::Selenium::Driver.new(
      app,
      browser: :chrome,
      options: options,
      desired_capabilities: capabilities,
      service: Selenium::WebDriver::Service.chrome(
        args: ["--log_path=#{chromedriver_log_path}", '--verbose']
      )
    )
  end
  Capybara.javascript_driver = :headless_chrome_billy
```

This is working great for me with no more deprecation errors.

twalpole:
@connorshea The key needs to be `goog:chromeOptions` not `chromeOptions` -  Rather than manually speccing the capabilities currently though you should be able to do

```
Capybara.register_driver :selenium_chrome do |app|
  browser_options = ::Selenium::WebDriver::Chrome::Options.new
  browser_options.headless!

  Capybara::Selenium::Driver.new app,
    browser: :chrome,
    options: browser_options,
    service: Selenium::WebDriver::Service.chrome(args: { log_path: 'tmp/chrome.log' })
end
```

PR #6989, opened 2019-03-03
[rb] rename WebDriver::Options class to WebDriver::Manage

titusfortner:
to disambiguate from using Options class to create Capabilities

I'd like to use `WebDriver::Options` or `WebDriver::W3COptions` as a parent class for `Chrome::Options`, etc, similar to how Java/C# do it. Even if we don't go that route "Options" is confusing now that we are using them for Browsers.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6989)


titusfortner:
@twalpole I don't love it either. I started to go with `DriverOptions`, but that wasn't much clearer. Part of me wants to get rid of it entirely, but I also didn't see an easy way to do that, either. @lmtierney do you have any thoughts on this one?

lmtierney:
Renaming it from `Options` (and getting there via `.manage`) breaks consistency from Java/.NET/Javascript. Python is in its own world and basically moves all the methods to the driver instance.

That being said, I'm not too concerned about the name of it, as long as the path to it (`.manage`) stays the same for version 4; I really doubt anyone is using the `Options` class in any special way. Manager does make more sense than Manage. If we aren't worried about consistency, I wouldn't mind what @twalpole is suggesting with splitting them out.

titusfortner:
@lmtierney Java/.NET/Javascript use `Options` for both extra Driver things and Browser configuration things? Different namespacing?

p0deje:
Strictly speaking, Java/.NET only provide with `Options` interface and internally it's all implemented as protected/private class with custom behavior (`RemoteWebDriverOptions`, `EventFiringOptions`, etc.) Ruby does nothing like that, so I believe we can rename.

p0deje:
The class name should be a noun, so `Manager` is probably the best, unless we want to split it.

lmtierney:
> @lmtierney Java/.NET/Javascript use `Options` for both extra Driver things and Browser configuration things? Different namespacing?

I was just referring to the equivalent Options class you're renaming in Ruby,

titusfortner:
> Python is in its own world and basically moves all the methods to the driver instance.

@lmtierney honestly, this is what I'd like to do for Ruby as well. I feel like we should be able to leave it to the Strict Languages to over-namespace everything. :)

PR #6995, opened 2019-03-04
Update remote_connection.py with new Cast commands

cliffordcheng:
The new Cast commands are added here:
https://chromium-review.googlesource.com/c/chromium/src/+/1401270

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6995)

### DIFF
--- a/py/selenium/webdriver/chrome/webdriver.py
+++ b/py/selenium/webdriver/chrome/webdriver.py
@@ -145,6 +145,45 @@ class WebDriver(RemoteWebDriver):
 
         """
         return self.execute("executeCdpCommand", {'cmd': cmd, 'params': cmd_args})['value']
+    
+    def get_sinks(self):
+        """
+        :Returns: A list of sinks avaliable for Cast.
+        """
+        return self.execute('getSinks')['value']
+
+    def get_issue_message(self):
+        """
+        :Returns: An error message when there is any issue in a Cast session.
+        """
+        return self.execute('getIssueMessage')['value']
+    
+    def set_sink_to_use(self, sink_name):
+        """
+        Sets a specific sink, using its name, as a Cast session receiver target.
+        
+        :Args:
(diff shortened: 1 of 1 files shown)


cliffordcheng:
Friendly ping.
I think this change is quite straightforward and we have some other tasks that're depending on this change.
If someone can help to look at it quickly, that will be great.

AutomatedTester:
How are you intending on using these new commands? It looks like a half completed PR

cliffordcheng:
We intend to consume these new commands in testing different functionalities between Chrome and Chromecast. We use ChromeDriver in most of our tests.
There is no way for us to trigger any action on Chromecast, or any Cast enabled devices, without using the new DevTools commands.

cliffordcheng:
This is used by Google internally and will also be used by third party.
We have some test updates that rely on this change.
Please feel free to let me know if there is anything needed from me to land this CL.

AutomatedTester:
What is missing is actual calls like https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/chrome/webdriver.py#L85-L147 so that they become usable.

cliffordcheng:
Ah. I see.
We just directly use execute in our cases.
https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/webdriver.py#L297

I am going to add those changes today and will reply again once it's done.
Thanks for pointing it out.

cliffordcheng:
Just added all the calls.
Please take a look again.

cgoldberg:
@AutomatedTester @cliffordcheng 
These commands just call their equivalent in DevTools Protocol... and we already have `execute_cdp_cmd()` for that.
IMHO this just expands the API for no real benefit.

cliffordcheng:
It looks like execute_cdp_cmd only do POST, but not GET.

PR #6997, opened 2019-03-05
Possibility to getMessage without additional info.

raul1ro:
I added private static variable(**simpleGetMessage**) and public static set method(**setSimpleGetMessage(boolean)**) for that variable. 

That variable is used in **getMessage()** -> **if simpleGetMessage == true** -> return only message of exception without addition info. 

**Default value of simpleGetMessage is false.** 

Those changes give the possibilitie to **getMessage()** without addition info. This apply for **_WebDriverException_** and every **class** which **extends WebDriverException** and don't **override getMessage()**. This is useful if you use a log and want to keep it short and clean.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/6997)


raul1ro:
I have accepted CLA.

barancev:
The idea to use a static variable that modifies behavior of getMessage method looks suboptimal. Instead I've implemented a separate method getRawMessage that allows to get the original message without extra info, see commit 057cc7066f064cf4d510cc260330111c48cef290

PR #7037, opened 2019-03-17
Check if argument is iterable. E.g. type **list** also works.

jerryjiahaha:

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/7037)


GQAssurance:
@jerryjiahaha What is the situation this fixes?

cgoldberg:
AFAICT..

when using a WebDriverWait and expected_conditions... 

you call some expected conditions with a locator that must be a tuple:
`EC.frame_to_be_available_and_switch_to_it((By.ID, "foo"))`

this change would allow you to pass locators as a list or other iterable:
`EC.frame_to_be_available_and_switch_to_it([By.ID, "foo"])`

Dude-X:
I think it's better to use an `isinstance(obj, Iterable)`.

PR #7042, opened 2019-03-19
allow ability to set default driver executable path at a class level

SalmonMode:
- [ x] By placing an `X` in the preceding checkbox, I verify that I have signed the [Contributor License Agreement](https://github.com/SeleniumHQ/selenium/blob/master/CONTRIBUTING.md#step-6-sign-the-cla)

Per the discussion in slack with @titusfortner, this change is a simple one that should have zero impact with regards to backwards compatibility, and all it does is allow you to set the default path for the driver's location for each `WebDriver` class at the class level so it doesn't have to be provided at the time of each driver's instantiation. This feature is already present in the Ruby language bindings and is just to bring them closer together in terms of feature parity.

I've added Python unit tests for each impacted `WebDriver` to make sure that the `executable_path` argument always takes priority over the `driver_path` class attribute.

I also adjusted the documentation in the relevant places to update them for the new behavior.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/7042)


titusfortner:
This functionality has been very helpful in Ruby for using a third party library to manage local drivers. Right now [the Python implementation](https://github.com/Omenia/webdrivermanager) appears to rely on shims, which, when the Ruby implementation had to rely on that, was cumbersome at best.

lmtierney:
@AutomatedTester any reason not to add this in? Preserves original functionality and adds a little flexibility

AutomatedTester:
brief look over this looks good. I am refactoring that area at the moment
so this helps.

Dude-X:
I would probably put the default inside the class to avoid a slower global variable lookup.

AutomatedTester:
Thanks for your PR. I am closing this as I would like the driver object to move to the following signature `browser(service=Service, options=Options`) which would mostly solve what I think you want it to be.

Issue #5758, opened 2018-04-11
Should use HTTP keep alive to avoid port exhaustion


awarecan:
## Meta -
OS:  Windows 7, 10
Selenium Version:  3.8.1
Browser:  Chrome
Browser Version:  65.0.3325.181 (64-bit)
Node.js: 8.11.1

## Expected Behavior -
During the test, TCP connection should be reused

## Actual Behavior -
Each WebDriver command will create a new TCP connection to ChromeDriver

## Steps to reproduce -
```JavaScript
await browser.get('https://www.google.com/');
for (let i=0; i<50000; i++) {
  await browser.getCurrentUrl();
}
```
use `netstat -na` to monitor the number of TCP connection in TIME_WAIT state while you are running above code. You will see massive amount of TCP connections. If you are using a powerful box and has default Windows TCP/IP configuration, you will hit ephemeral port exhaustion very soon.

## Note -
This was [a well known issue related with ChromeDriver](https://bugs.chromium.org/p/chromedriver/issues/detail?id=2183). However ChromeDriver has [a dirty fix](https://chromium.googlesource.com/chromium/src/+/736efbc1c284f9b85d7dcda22771e23786def5d5%5E%21/#F0) in 2.35 release. It requires client set `Connection: keep-alive` HTTP header, see the discussion in #4162 and #3457

Now, we finally have chance to resolve this issue in selenium-webdriver side by passing in an Agent with keepAlive option while creating HttpClient in createExecutor() function.

p0deje:
Seems like we need to do this for all the bindings.

barancev:
In Java I can't reproduce this issue with chromedriver, but the number of TIME_WAIT'ing ports grows if I use geckodriver.

dmytro-verner:
@barancev  possibly you have registry's HKLM\SYSTEM\CurrentControlSet\Services\Tcpip\Parameters (in case of Windows) TcpTimedWaitDelay set to non-default value(which is 240 secs). 

Setting this value to a lower bound(e.g. 30 secs) can help to circumvent the current issue.

barancev:
It appears that OkHttp uses keep-alive by default, so it's not an issue with Java binding.

p0deje:
[...]

PR #7065, opened 2019-03-28
[rb] Default http client use persistent connections

twalpole:

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/7065)

### DIFF
--- a/rb/lib/selenium/webdriver/remote/http/default.rb
+++ b/rb/lib/selenium/webdriver/remote/http/default.rb
@@ -50,6 +50,10 @@ module Selenium
             self.read_timeout = value
           end
 
+          def close
+            @http&.finish
+          end
+
           private
 
           def http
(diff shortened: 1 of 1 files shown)


twalpole:
This is to start a discussion on using a persistent connection by default -- sped up Sauce over SSL by about 30% on a portion of the Capybara test suite.

titusfortner:
So, this should better address the issue in #5758 
Also, if this works, we should probably deprecate the `Remote::Http::Persistent` class entirely.

I'm also really curious if Ruby is the only binding that doesn't default to keep-alive for remote connections, because if so, I'm going to need to get code out to Sauce clients on how to turn it on. :) I'll ping people on irc tomorrow.

titusfortner:
I've verified that keep alive is turned on by default for Java and .net

PR #7070, opened 2019-03-29
account for scenarios where /proc/version is inaccessible

dmvt:
On Samsung Linux on Dex and a number of other containers, the file `/proc/version` cannot be accessed, even with sudo privileges. This results in an error that prevents `rails server` from even starting. Since the failure of this command means we are not in a Microsoft environment, but a sandboxed mobile or container based one, return false if an error accessing the file occurs.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/7070)

### DIFF
--- a/rb/lib/selenium/webdriver/common/platform.rb
+++ b/rb/lib/selenium/webdriver/common/platform.rb
@@ -96,6 +96,8 @@ module Selenium
         return false unless linux?
 
         File.read('/proc/version').include?('Microsoft')
+      rescue
+        false
       end
 
       def cygwin?


p0deje:
Hello! Sorry for this taking so long to review - I simply haven't noticed this PR. My apologies.

Is it possible to make the exception class explicit? I'd prefer to avoid rescuing all standard errors.

kalagirisantoshreddy:
- [ ] 

- [ ] 

- [ ] 

- [ ] 

- [ ] 

- [ ] 

- [ ] 

- [ ] 

- [ ] 

- [ ] 

- [ ]

PR #7072, opened 2019-03-29
[py] turn on keep-alive by default for remote connections

titusfortner:
Java & .NET have this turned on by default. We're looking to do this in Ruby #7065 

Based on a sample suite run against Sauce Labs with SSL, keep-alive improved performance ~30%. There shouldn't be a downside to turning this on.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/7072)


lmtierney (approved):

lmtierney:
@AutomatedTester if Java and .NET are doing this, any reason not to?

PR #7075, opened 2019-03-31
switch to parent frame fix for non-w3c compatible drivers

corevo:

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/7075)


corevo:
@jleyba all drivers except chrome use the `W3C_COMMAND_MAP` and Chrome's command map doesn't have `SWITCH_TO_PARENT_FRAME`, not sure if that's on purpose or not.  

Also if merged we'll need to release.

Issue #6557, opened 2018-10-18
JavaScript atom bot.action.clear generates blur event twice


JohnChen0:
## Meta -
OS:  Windows, OSX, and Linux
Selenium Version:  3.14.0
Browser:  Chrome
Browser Version:  69.0.3497.100

## Expected Behavior -
After calling JavaScript atom bot.action.clear to clear a text input element, the blur event handler attached to the element is called once.

## Actual Behavior -
The blur event handler is called twice.

According to https://github.com/SeleniumHQ/selenium/blob/selenium-3.14.0/javascript/atoms/action.js#L100, the blue event handler is first called explicitly by

    bot.events.fire(element, bot.events.EventType.BLUR);

And then it is called again when the element loses focus due to

    bot.action.LegacyDevice_.focusOnElement(body)

## Steps to reproduce -

Create an HTML file with the following contents:
```
<input id='t' type='text' value='123'></input>
<script>
count = 0;
document.getElementById('t').addEventListener('blur', () => ++count, true);
</script>
```

Then run the following Python code with ChromeDriver 2.43 (which uses the JavaScript atom mentioned above to implement element clear command):
```
from selenium import webdriver

driver = webdriver.Chrome()
driver.get(URL-to-the-above-HTML-file)
text = driver.find_element_by_id('t')
text.click()
text.clear()

print('count=' + str(driver.execute_script('return count')))
driver.quit()
```

The above code prints "count=2", showing that the blue event handler is called twice.

lmtierney:
They do use the atom from selenium, but I'm not sure updating it here means they will update it on their end (https://chromium.googlesource.com/chromium/src/+/master/third_party/webdriver/atoms.cc)

lmtierney:
You may want to raise a bug with them referring to this one as well

JohnChen0:
Created an issue for ChromeDriver at https://crbug.com/chromedriver/2622. I work for ChromeDriver, and can update the atoms in ChromeDriver if this issue is fixed from Selenium side. If there is an agreement that the right fix is to remove the explicit call to raise blue event from the atom, I'd be happy to submit a PR for it. (An alternative would be patching the atom directly inside ChromeDriver, but I'd rather avoid that if possible.)

lmtierney:
Ah, perfect then! @barancev, thoughts?

JohnChen0:
Resolved with PR #7102.

PR #7102, opened 2019-04-13
Fix JavaScript atom for Element Clear

JohnChen0:
Fixes issue #6557: The bot.action.clear function explicitly raises
blur event, and then the unfocus step raises blur event again,
resulting in double firing of blue event.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/7102)

### DIFF
--- a/javascript/atoms/action.js
+++ b/javascript/atoms/action.js
@@ -98,6 +98,9 @@ bot.action.clear = function(element) {
       element.value = '';
     }
     bot.events.fire(element, bot.events.EventType.CHANGE);
+    if (goog.userAgent.IE) {
+      bot.events.fire(element, bot.events.EventType.BLUR);
+    }
     var body = bot.getDocument().body;
     if (body) {
       bot.action.LegacyDevice_.focusOnElement(body);
--- a/javascript/atoms/test/action_test.html
+++ b/javascript/atoms/test/action_test.html
@@ -56,6 +56,9 @@
           });
     }
 
+    var expectBlurEvent = goog.partial(expectEvent,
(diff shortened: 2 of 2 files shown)


barancev:
I'd prefer https://github.com/w3c/webdriver/issues/1354 to be resolved before applying this PR.

JohnChen0:
I believe this PR is independent of w3c/webdriver#1354, which is a discussion of whether blur event should be raised as part of Element Clear. This PR is for a different issue, that the blur event is raised twice after each Element Clear command, first by an explicit call to `bot.events.fire(element, bot.events.EventType.BLUR)`, and then as a side effect of `bot.action.LegacyDevice_.focusOnElement(body)`. Regardless of whether W3C decides to change the spec or not, raising the blur event twice is clearly wrong.

barancev:
Sorry, I misread the issue. Yes, I agree, firing two blur events is suboptimal.

But unfortunately the proposed change breaks IE, it fires zero events after applying this patch.

Can you improve the patch to cover IE case too? And please rework the test testClearingANonEmptyElement (in action_test.html file) to verify amount of fired blur events.

JohnChen0:
@barancev Thanks for the suggestions. I've uploaded a new commit to address your concerns. It has two changes:

- Explicitly raise blur event on IE
- Updated testClearingANonEmptyElement to verify that blur event is raised exactly once after clearing.

Please let me know if you have any further questions or concerns.

JohnChen0:
@barancev Could you please take another look at this PR? Thanks.

barancev:
Merged to master. Thank you!

PR #7106, opened 2019-04-16
Driver path

titusfortner:
First, it makes more sense now to do:
```ruby
Selenium::WebDriver::Chrome::Service.driver_path = '/path/to/driver'
```
rather than to do:
```ruby
Selenium::WebDriver::Chrome.driver_path = '/path/to/driver'
```

Second, this allows `#driver_path=` to take a `Proc` value. The reason for this is to be able to lazy load an auto-generated path file. (Essentially this allows webdrivers gem to remove its Monkey Patches).

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/7106)

### DIFF
--- a/rb/lib/selenium/webdriver/chrome/service.rb
+++ b/rb/lib/selenium/webdriver/chrome/service.rb
@@ -35,7 +35,7 @@ module Selenium
         @shutdown_supported = true
 
         def self.driver_path=(path)
-          Platform.assert_executable path
+          Platform.assert_executable path if path.is_a?(String)
           @driver_path = path
         end
 
--- a/rb/lib/selenium/webdriver/chrome/service.rb
+++ b/rb/lib/selenium/webdriver/chrome/service.rb
@@ -26,7 +26,6 @@ module Selenium
 
       class Service < WebDriver::Service
         @default_port = 9515
-        @driver_path = Chrome.driver_path
         @executable = 'chromedriver'
         @missing_text = <<~ERROR
           Unable to find chromedriver. Please download the server from
@@ -34,6 +33,11 @@ module Selenium
           More info at https://github.com/SeleniumHQ/selenium/wiki/ChromeDriver.
         ERROR
         @shutdown_supported = true
+
+        def self.driver_path=(path)
+          Platform.assert_executable path
+          @driver_path = path
+        end
 
         private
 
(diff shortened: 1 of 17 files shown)


rb/spec/unit/selenium/webdriver/chrome/service_spec.rb
```diff
@@ -49,6 +49,41 @@ module WebDriver
...
+          Chrome::Service.driver_path = path
+
+          service = Service.chrome
+
+          expect(service.executable_path).to eq path
+        end
+
+        it 'allows #driver_path= with Proc value' do
+          path = '/path/to/driver'
+          proc = Proc.new { path }
```
luke-hill: This might fall afoul of some of the later ruby cops. If you use `a_proc = proc { path }` it should remedy it. (For others too)

rb/lib/selenium/webdriver/chrome/service.rb
```diff
@@ -35,6 +34,11 @@ class Service < WebDriver::Service
         ERROR
         @shutdown_supported = true
 
+        def self.driver_path=(path)
```
twalpole: I think this can be moved into the base class, which would also solve the issue of none of the other current implementations of `driver_path=` having the String check

titusfortner: My reluctance was that it doesn't make sense for the Base class itself to have a driver_path, and if for some reason it was set on the base class that would change any of the subclass values.

twalpole: @titusfortner Setting on the base class wouldn't affect anything -- With class instance variables any class derived from the base class has its own copy of the variable

ezr-ondrej:
Hi, the class `Selenium::WebDriver::Chrome::Service` states it's a private api, is that to be of an concern? Shouldn't we remove it?

titusfortner:
@ezr-ondrej Yeah, I think the original reasoning was that users should initialize with `Selenium::WebDriver::Service.chrome` rather than initializing the subclass class directly, but... I think you are right, that even though driver_path is defined in the superclass, if we expect people to use it on the subclass, then the subclass shouldn't be listed as private api. I updated it. Thanks.

PR #7110, opened 2019-04-16
[rb] create a Common Options class to handle w3c specific capabilities

titusfortner:
This implementation has more tidying (including more specs). The idea is to be able to initialize a driver and specify browser specific things without needing to use Capabilities class at all (which most other bindings allow)

Let's compare what Ruby has to what other bindings have...

Java wants users to only use Options classes for browsers (and `MutableCapabilities` class for custom things like `sauce:options`). They implement this by subclassing Options classes 4 levels deep below Capabilities class, and everything gets handled that way.

Python is a little different in that you can use a `set_capability` method to add whatever you want, and then when the driver initializes it only allows through things that match w3c or have a colon in the keyword.

1. Pulls out the common w3c things into a `Common::Options` class
2. Everything implements `#as_json` for nested serialization, and I updated all the places where `#as_json` wasn't using `String` keys
3. I added explicit support for everything specified in Firefox & Chrome documentation.
4. I deprecated anything that can (and should) be used by standard Ruby accessors instead of extra methods. I can undo that, but this is literally the whole point of having `attr_accessor` when you don't have any additional logic attached.
5. The things that need additional parsing are now handled in both the constructor and the defined method
6. Some of the additional validations broke specs that allowed random/invalid Strings to be passed in, so that had to get tidied
7. I added an `@custom` which is essentially just taking everything that isn't explicitly defined in either the browser class or the superclass (w3c). This is my initial pass at supporting things like `sauce:options` without having to build the capabilities Hash/JSON from scratch (which we have to do now)
8. This also allows us to specify `browserName` inside the options class directly
9. Rubocop is failing for Chrome Options code because of complexity, I think we'll need to implement it similarly to how IE is currently implemented to fix it.
10. With this we can pretty much immediately get rid of all of the private `create_capabilities` code in the driver initialization (essentially the thing that makes working with Remote sessions more of a pain than it should be)

There's definitely more tidying, but this passes the current specs and I want to get some agreement on what of this we want to push forward with for me to keep working on it.

This is also something I can get help with from any Rubyists at the workshop tomorrow if I can get additional direction/buy-in.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/7110)


rb/lib/selenium/webdriver/chrome/options.rb
```diff
@@ -40,33 +43,58 @@ class Options
...
+        # @option opts [Boolean] :detach whether browser is closed when the driver is sent the quit command
+        # @option opts [String] :debugger_address address of a Chrome debugger server to connect to
+        # @option opts [Array<String>] :exclude_switches command line switches to exclude
+        # @option opts [String] :minidump_path Directory to store Chrome minidumps (linux only)
+        # @option opts [Hash] :perf_logging_prefs A hash for performance logging preferences
+        # @option opts [Array<String>] :window_types A list of window types to appear in the list of window handles
         #
 
         def initialize(**opts)
+          opts.delete(:browser_name)
```
twalpole: No need to delete :browser_name if the next line overwrites it

rb/lib/selenium/webdriver/chrome/options.rb
```diff
@@ -40,33 +43,58 @@ class Options
...
+          opts[:browser_name] = 'chrome'
+
           @args = Set.new(opts.delete(:args) || [])
           @binary = opts.delete(:binary) || Chrome.path
           @prefs = opts.delete(:prefs) || {}
+
+          @encoded_extensions = opts.delete(:encoded_extensions) || []
```
twalpole: `Array(opts.delete(:encoded_extensions))` would make this slightly more flexible (allow passing one extension or array of extensions) while still defaulting [] if nil passed

rb/lib/selenium/webdriver/chrome/options.rb
```diff
@@ -124,7 +156,10 @@ def add_option(name, value)
         #
 
         def add_preference(name, value)
-          prefs[name] = value
+          WebDriver.logger.deprecate 'Options#add_preference',
+                                     'Options.prefs[<key>] = value'
```
twalpole: Options#prefs

rb/lib/selenium/webdriver/common/options.rb
```diff
@@ -0,0 +1,87 @@
...
+        def initialize(**opts)
+          VALID_W3C.each do |capability|
+            next unless opts.key?(capability)
+
+            instance_variable_set("@#{capability}", opts.delete(capability))
+          end
+
+          validate_proxy if @proxy
+
+          @custom = opts || {}
```
twalpole: opts has to already be a hash - no need to || {}

rb/lib/selenium/webdriver/edge/options.rb
```diff
@@ -40,9 +40,17 @@ class Options
         #
 
         def initialize(**opts)
+          opts.delete(:browser_name)
```
twalpole: No need to delete if overwriting

rb/lib/selenium/webdriver/edge/options.rb
```diff
@@ -40,9 +40,17 @@ class Options
...
+          opts[:browser_name] = 'MicrosoftEdge'
+
           @in_private = opts.delete(:in_private) || false
-          @extension_paths = opts.delete(:extension_paths) || []
+          @extension_paths = []
+          extension_paths = opts.delete(:extension_paths) || []
```
twalpole: Array(opts.delete(:extensions_paths))

rb/lib/selenium/webdriver/firefox/options.rb
```diff
@@ -43,12 +45,17 @@ class Options
         #
 
         def initialize(**opts)
+          opts.delete(:browser_name)
```
twalpole: No need to delete

rb/lib/selenium/webdriver/ie/options.rb
```diff
@@ -86,9 +86,17 @@ class Options
         #
 
         def initialize(**opts)
+          opts.delete(:browser_name)
```
twalpole: No need to delete

rb/lib/selenium/webdriver/safari/options.rb
```diff
@@ -38,8 +38,13 @@ class Options
         #
 
         def initialize(**opts)
+          opts.delete(:browser_name)
```
twalpole: No need to delete

twalpole (commented):
Found a few minor things, but overall I'm really liking this 👍

titusfortner:
whew. When I added specs I found that nested Hash values were not getting converted properly in `#as_json` so I implemented a recursive `#parse_json(value)`

rb/lib/selenium/webdriver/chrome/options.rb
```diff
@@ -40,33 +43,63 @@ class Options
...
         #
 
-        def add_extension(path)
-          raise Error::WebDriverError, "could not find extension at #{path.inspect}" unless File.file?(path)
-          raise Error::WebDriverError, "file was not an extension #{path.inspect}" unless File.extname(path) == '.crx'
-
-          @extensions << path
+        def add_extension(extension)
+          @extensions << extension
+          validate_extension(extension)
```
luke-hill: Wouldn't it make more sense to validate before pushing?

rb/lib/selenium/webdriver/chrome/options.rb
```diff
@@ -155,30 +201,58 @@ def headless!
...
+
+        private
 
-          opts = @options
-          opts[:binary] = @binary if @binary
-          opts[:args] = @args.to_a if @args.any?
-          opts[:extensions] = extensions if extensions.any?
-          opts[:mobileEmulation] = @emulation unless @emulation.empty?
-          opts[:prefs] = @prefs unless @prefs.empty?
+        def validate_emulation(device_name: nil, device_metrics: nil, user_agent: nil)
```
luke-hill: is this validating anything other than falsiness?

luke-hill (approved):
The IE specs definitely help a lot understand whats going on (Alongside the name tweaks removing the `ie.` prefix in the json casting.

Couple of minor points, maybe not relevant

titusfortner:
This was subsumed by #7378

PR #7111, opened 2019-04-17
Run tests only when changes in folder

### LINKED ISSUE #6208: Smarter Travis Builds
A PR on Travis was auto-canceled when someone committed to Master, which makes it hard to know that the PR didn't break anything. I thought we used to have it so that tests were only run when changes from a particular folder.

It shouldn't be too difficult (though perhaps time consuming to make sure it works correctly) to add a conditional using something like `git diff --name-only HEAD~1` to check for what has changed to [these tasks](https://github.com/SeleniumHQ/selenium/blob/master/scripts/travis/script.sh) based on folder (python tests when changes in /py/*, ruby tests when changes in /rb/*, etc)

If we minimize what gets run, that should keep us from having to auto-cancel PR builds.

knmkr:
Fixes #6208 

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/7111)

### DIFF
--- a/scripts/travis/script.sh
+++ b/scripts/travis/script.sh
@@ -4,20 +4,41 @@ if [[ $TOXENV == *"remote"* ]]; then
   ./go selenium-server-standalone
 fi
 
+# Tests will be run only when each folder, such as /py, is modified
 if [[ ! -z $TOXENV ]]; then
-  tox -c py/tox.ini
+  if git diff --name-only HEAD~1| grep '^py/' >/dev/null; then
+    tox -c py/tox.ini
+  fi
 fi
 
 # Ordering matters here. We want rake tasks to run first
 if [[ ! -z $TASK ]]; then
-  ./go $TASK
+  if [[ $TASK == //rb:* ]]; then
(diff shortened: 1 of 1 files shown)


titusfortner:
@barancev do you think we should only do this for PR's? 

I'm wondering if we should add an `||` here and still run things if  ENV `TRAVIS_PULL_REQUEST` is false.

shs96c:
LGTM, but will wait for @barancev. Note that if we modify anything in the top-level of the project, this will not cause any tests at all to run, but I guess we don't do that very often.

barancev:
I think it's worth to use always, not only for Prs. If someone make changes on the top level they can run selected tests manually on travis. But I agree with @shs96c that this happens not often and usually causes compilation issues, not test failures.

titusfortner:
@knmkr thanks for your contribution!

knmkr:
@titusfortner @shs96c @barancev Thanks 🙌

barancev:
Excellent work!

PR #7114, opened 2019-04-18
Add new expected condition: element_to_be_unclickable

clarmso:
element_to_be_unclickable is the reverse of element_to_be_clickable.
Instead of waiting for an element to be clickable, we wait and expect
a web element to be unclickable but visible.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/7114)

### DIFF
--- a/py/test/selenium/webdriver/common/click_tests.py
+++ b/py/test/selenium/webdriver/common/click_tests.py
@@ -35,3 +35,6 @@ def testCanClickOnALinkThatOverflowsAndFollowIt(driver):
 def testClickingALinkMadeUpOfNumbersIsHandledCorrectly(driver):
     driver.find_element(By.LINK_TEXT, "333333").click()
     WebDriverWait(driver, 3).until(EC.title_is("XHTML Test Page"))
+
+def testCannotClickDisabledButton(driver):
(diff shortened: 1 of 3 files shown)


py/test/selenium/webdriver/common/click_tests.py
```diff
@@ -35,3 +35,6 @@ def testCanClickOnALinkThatOverflowsAndFollowIt(driver):
 def testClickingALinkMadeUpOfNumbersIsHandledCorrectly(driver):
     driver.find_element(By.LINK_TEXT, "333333").click()
     WebDriverWait(driver, 3).until(EC.title_is("XHTML Test Page"))
+
+def testCannotClickDisabledButton(driver):
+    WebDriverWait(driver, 3).until(EC.element_to_be_unclickable(By.ID, "disabled-button"))
```
vhryhorchuk-amplify: I believe it's misleading name for the condition. I prefer "element_to_be_disable"
We can have a condition, when element is enabled but we can't click it, because another element overlays above it.
So, If we use "unclickable" we might mislead people, who use that condition to verify if element can be clicked

PR #7115, opened 2019-04-18
[rb] fix serialization methods

titusfortner:
`#as_json` is supposed to return a Hash with all of the Symbols stripped out. 

We could implement Rails' monkey patch on Hash, but we shouldn't
We could JSON.parse(JSON.generate hash), but that feels wrong.

Also, I feel like converting snake case keys into camel case makes more sense in the `#as_json` than the `#to_json` method.

@p0deje / @twalpole can you double check my logic for `#generate_as_json` here?

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/7115)


p0deje:
The logic seems to be correct, but there are failing linter and unit tests.

PR #7119, opened 2019-04-19
Fix undefined method encoded for String

zlyaksa:
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/7119)

### DIFF
--- a/rb/lib/selenium/webdriver/firefox/options.rb
+++ b/rb/lib/selenium/webdriver/firefox/options.rb
@@ -137,7 +137,7 @@ module Selenium
         def as_json(*)
           opts = @options
 
-          opts[:profile] = @profile.encoded if @profile
+          opts[:profile] = @profile.encode if @profile
           opts[:args] = @args.to_a if @args.any?
           opts[:binary] = @binary if @binary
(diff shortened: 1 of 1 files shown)


twalpole:
@p0deje It used to support passing a profile name in, and it's still shown in the docs - https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver.rb#L74 - I think this is something we either need to fix, or fix & deprecate and do another 3.x release.

twalpole:
The root issue is that the string passed in is only converted to an actual profile if it's set via the `profile=` writer - https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/firefox/options.rb#L125 .  If it's set any other way (initialize, add_option) it doesn't get converted.

titusfortner:
I'm working on some code intended to make sure the same behaviors are used for options classes regardless of how it is added to the class (initialize, add_*, writer, etc)

p0deje:
A proper fix landed in 34a2a39782ef1e3a4723ccb2e9ede1e2b469ed17 and will be released in 3.142.0 and 4.0.0.alpha1. Thank you!

PR #7156, opened 2019-04-30
Setup right browserName for iOS mobile web testing

kevinwee0712:
Register appium server with selenium grid by using nodeconfig is causing some trouble for iOS mobile web testing. In node configuration json, deviceName is used to filled in browserName field. Check this link: http://appium.io/docs/en/advanced-concepts/grid/#grid-node-configuration-example-json-file . After capability matcher is validated and real session is established between selenium hub and appium server, deviceName could not be recognized by appium server. After this fix, iOS mobile web testing could be run properly on Selenium Hub + Appium Server Node. It's fully tested.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/7156)

### DIFF
--- a/java/server/src/org/openqa/grid/internal/TestSlot.java
+++ b/java/server/src/org/openqa/grid/internal/TestSlot.java
@@ -135,6 +135,7 @@ public class TestSlot {
         log.info("Trying to create a new session on test slot " + this.capabilities);
         desiredCapabilities.put(GridNodeConfiguration.CONFIG_UUID_CAPABILITY,
                                 capabilities.get(GridNodeConfiguration.CONFIG_UUID_CAPABILITY));
+        desiredCapabilities = convertToAppiumBrowserName(desiredCapabilities);
         TestSession session = new TestSession(this, desiredCapabilities, Clock.systemUTC());
         currentSession = session;
         lastSessionStart = System.currentTimeMillis();
@@ -285,5 +286,22 @@ public class TestSlot {
    */
   public long getLastSessionStart() {
     return lastSessionStart;
+  }
+
+  /**
+   * When appium registers node under selenium hub, it's using deviceName as browserName.
+   * But during iOS mobile web testing, when new session established between selenium hub and appium, this browserName is not recognized by appium
+   * since it's not a valid one.
+   *
+   * @return correct browserName
+   */
+  private Map<String, Object> convertToAppiumBrowserName(Map<String, Object> desiredCapabilities)
+  {
+    String browserName = (String) desiredCapabilities.get("browserName");
+    if (browserName.contains("iPhone") || browserName.contains("iPad"))
+    {
+      desiredCapabilities.put("browserName", "Safari");
+    }
+    return desiredCapabilities;
   }
 }


diemol:
@kevinwee0712 why isn't this done on the Appium side, and send the proper value for the field? Have you checked with the Appium project?

kevinwee0712:
@diemol I did check Appium project and frankly in this case appium server expecting correct capability. For example, I'm doing Safari web testing on iPhone X 12.1. Appium server will expect desired capability like this and it does make sense:
{
browserName: "Safari",
deviceName: "iPhone X",
platformName: "iOS",
platformVersion: "12.1",
}

If I just use appium server alone, everything works fine.

But as soon as I register appium server as a node to selenium hub through nodeconfig, problem happens. If I setup nodeconfig json as follows:
{
  "capabilities":
  [
    {
      "deviceName": "iPhone X",
      "browserName": "Safari",
      "platformVersion": "12.1",
      "maxInstances": 1,
      "platformName": "iOS"
    },
    {
      "deviceName": "iPhone 8 Plus",
      "browserName": "Safari",
      "platformVersion": "12.1",
      "maxInstances": 1,
      "platformName": "iOS"
    },
    {
      "deviceName": "iPhone 6s",
      "browserName": "Safari",
      "platformVersion": "12.1",
      "maxInstances": 1,
      "platformName": "iOS"
    }
  ],
  "configuration":
  {
....
  }
}

Selenium hub will totally ignore deviceName field during matching validation and treat those three capabilities as one. If I set desired capability like this:  
{
browserName: "Safari",
deviceName: "iPhone X",
platformName: "iOS",
platformVersion: "12.1",
}
Selenium hub will randomly return a "matched" capability to me, it might be 
{
browserName: "Safari",
deviceName: "iPhone 6s",
platformName: "iOS",
platformVersion: "12.1",
}

After that, selenium hub will start to execute following logic:
      if (matches(desiredCapabilities)) {
        log.info("Trying to create a new session on test slot " + this.capabilities);
        desiredCapabilities.put(GridNodeConfiguration.CONFIG_UUID_CAPABILITY,
                                capabilities.get(GridNodeConfiguration.CONFIG_UUID_CAPABILITY));
        TestSession session = new TestSession(this, desiredCapabilities, Clock.systemUTC());
        currentSession = session;
        lastSessionStart = System.currentTimeMillis();
        return session;
      }

Create a new session with appium server, passing "original" desired capability to it as follows:
{
browserName: "Safari",
deviceName: "iPhone X",
platformName: "iOS",
platformVersion: "12.1",
}

In this case, appium does work well. But as soon as you start another test with same desired capability concurrently, appium is screwed. Theoretically, it shouldn't be like that because I've enabled "maxInstances": 1 during node register, and the other test should wait until previous test finished. 

In reality, when another test sends request to selenium hub with the same desired capability
{
browserName: "Safari",
deviceName: "iPhone X",
platformName: "iOS",
platformVersion: "12.1",
}
selenium hub will do some trick again. Try to match browserName, platformName and totally ignore deviceName. Then selenium hub will randomly return a capability and marked it as match. After that selenium hub will start another new session with appium server with original desiredCapablities again:
{
browserName: "Safari",
deviceName: "iPhone X",
platformName: "iOS",
platformVersion: "12.1",
}

But right now, we already have a test running with that desired capability. Which means maxInstances setting during node register doesn't take effect at all. It makes sense actually, because in selenium hub DefaultCapabilityMatcher.java didn't take deviceName into account at all. It will treat following three capabilities as the same. 
    {
      "deviceName": "iPhone X",
      "browserName": "Safari",
      "platformVersion": "12.1",
      "maxInstances": 1,
      "platformName": "iOS"
    },
    {
      "deviceName": "iPhone 8 Plus",
      "browserName": "Safari",
      "platformVersion": "12.1",
      "maxInstances": 1,
      "platformName": "iOS"
    },
    {
      "deviceName": "iPhone 6s",
      "browserName": "Safari",
      "platformVersion": "12.1",
      "maxInstances": 1,
      "platformName": "iOS"
    }

I think appium also noticed this bug I mentioned above and they found a workaround like this:
http://appium.io/docs/en/advanced-concepts/grid/#grid-node-configuration-example-json-file

Apparently, appium tries to comprise and decides to pass deviceName to browserName, this way each device can get a unique session and maxinstances could work.
  "capabilities":
      [
        {
          "browserName": "<e.g._iPhone5_or_iPad4>",
          "version":"<version_of_iOS_e.g._7.1>",
          "maxInstances": 1,
          "platform":"<platform_e.g._MAC_or_ANDROID>"
        }
      ],

Does that work? Yes, it works. For mobile app testing this work around is perfect, because mobile app testing do not rely on browserName at all. Automation will specify app field in desiredCapabilities and appium will not care about the wrong browserName passing in such as iPhone 5. That's ridiculous actually, because iPhone 5 is not a browser name at all. It's a device name.

I don't know why no one complains about this. Probably no one tried to use selenium hub + appium server node combination to do mobile "web" testing?

If anyone followed this offical link(http://appium.io/docs/en/advanced-concepts/grid/#grid-node-configuration-example-json-file) to setup appium node and set up desired capability like this:
{
browserName: "Safari",
deviceName: "iPhone X",
platformName: "iOS",
platformVersion: "12.1",
}

Selenium hub will complain, no matched capability could be found since one is using safari, one using iPhone*. Of course it doesn't match.

But if automation changed desired capability to this:
{
browserName: "iPhone X",
deviceName: "iPhone X",
platformName: "iOS",
platformVersion: "12.1",
}

Selenium hub will be happy and let it through, but appium server will be upset. Appium will complain that iPhone X is not a browserName(Again for mobile web testing). If we can't provide a proper browser name, they appium will ask for one capability named app. If automation passed in app in capability, appium could do mobile app testing happily, which leaves mobile web tester in a dilemma.

All in all, this is a selenium bug and appium tries to comprise for years. We should either convert browserName from deviceName to a proper one or change DefaultCapabilityMatcher.java to take deviceName into account. Otherwise, mobile web tester could only skip using selenium hub + appium server node which is a perfect combination. What a shame.

kevinwee0712:
Some logs from selenium hub with same desired capability: 

23:03:01.103 INFO [RequestHandler.process] - Got a request to create a new session: Capabilities {browserName: Safari, deviceName: iPhone X, platformName: ios, platformVersion: 12.1}
23:03:01.105 INFO [TestSlot.getNewSession] - Trying to create a new session on test slot {server:CONFIG_UUID=34be3d76-84ad-41a6-bf5a-c4f5473d03ef, seleniumProtocol=WebDriver, platformVersion=12.1, browserName=Safari, maxInstances=1, platformName=iOS, deviceName=iPhone X, platform=IOS}
23:03:34.113 INFO [RequestHandler.process] - Got a request to create a new session: Capabilities {browserName: Safari, deviceName: iPhone X, platformName: ios, platformVersion: 12.1}
23:03:34.114 INFO [TestSlot.getNewSession] - Trying to create a new session on test slot {server:CONFIG_UUID=e4c38a23-828e-4850-a076-8ec521d03413, seleniumProtocol=WebDriver, platformVersion=12.1, browserName=Safari, maxInstances=1, platformName=iOS, deviceName=iPhone 8 Plus, platform=IOS}

diemol:
Ok, so if you need to consider the `deviceName` when the capabilities are matched, you can provide your own capability matched and override the default behaviour. 

I have an example I wrote a few years ago that should help you to do that, in that repo there is a link of the original source where I got all the information. With that, you can tweak it in the way you need and the Grid 3.x does not need to be changed. https://github.com/diemol/custom-capability-matcher

I'm sorry to close the PR but there is no need to add it since you can provide a capability matcher. We will have this into account while working in the new Grid 4.x

PR #7163, opened 2019-05-02
Allow attaching an existing Internet Explorer

bhecquet:
Currently, there is no way to use/attach an Internet Explorer started outside of Selenium
This PR aims at adding this feature.

This attachment can be done on new session creation where the capability `attachExistingBrowser` is set to `true` in ieOptions. Then, driver looks at existing Internet Explorer windows, created programmatically (i.e: the same way IEDrvierServer creates browser). If none is found, behaviour is the same as if IEDriverServer cannot create a browser.
If several Internet Explorer are found, it maps to the first one.
Drawback of this approach is that we can only have one IE at the same time.

To use attaching instead of new browser creation, I do:

	((Map<String, Object>) options.getCapability("se:ieOptions")).put("attachExistingBrowser", true);

It may be improved by modifying selenium clients

	modified:   iedriver/BrowserFactory.cpp
	modified:   iedriver/CommandHandlers/NewSessionCommandHandler.cpp
	modified:   iedriver/IECommandExecutor.cpp
	modified:   iedriver/IECommandExecutor.h
	modified:   iedriver/WebDriverConstants.h

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/7163)

### DIFF
--- a/cpp/iedriver/BrowserFactory.cpp
+++ b/cpp/iedriver/BrowserFactory.cpp
@@ -971,10 +971,15 @@ BOOL CALLBACK BrowserFactory::FindChildWindowForProcess(HWND hwnd, LPARAM arg) {
   } else {
     DWORD process_id = NULL;
     ::GetWindowThreadProcessId(hwnd, &process_id);
-    if (process_window_info->dwProcessId == process_id) {
+    if (process_window_info->dwProcessId == process_id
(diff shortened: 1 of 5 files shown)


AutomatedTester:
Thank you for the effort you put into this patch. Unfortunately for consistency we are not going to accept this into our tree as no other browsers support this and are not able to.

bhecquet:
Hello,

I don't agree with you as it's possible to attache to an existing chrome browser through the debugger port.

PR #7171, opened 2019-05-04
Fix for building selenium project

rookieInTraining:
In case the user is trying to build the entire project using `./go build`
or trying to build the standalone-server. The builds were failing with
the below exception :

BUILD FAILED: The rule //java/client/test/org/openqa/selenium/remote:common-tests
could not be found.

This was caused because the tasks in the Rakefile were still referring to
"common-tests" task instead of the new "small-tests"

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/7171)


p0deje:
Thank you!

PR #7185, opened 2019-05-08
Test against latest Ruby (`2.6`)

okuramasafumi:
The motivation behind this change is that we'd like to test selenium against the latest Ruby.
`2.4` is now quite old and in a security maintenance phase. More people will upgrade their Ruby to newer versions.
So I thought testing against the latest Ruby only makes sense.

---
This change is [Reviewable](https://reviewable.io/reviews/seleniumhq/selenium/7185)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ _bindings:
     if: env(BUILD_ONLY) IS NOT present || env(BUILD_ONLY) =~ /python/
   ruby: &ruby
     language: ruby
-    rvm: 2.4
+    rvm: 2.6
     jdk: oraclejdk8
     if: env(BUILD_ONLY) IS NOT present || env(BUILD_ONLY) =~ /ruby/
   node: &node


p0deje:
Current selenium-webdriver requires 2.4 as a minimum version. We use it on CI to make sure we don't accidentally use features from newer versions and thus break 2.4 support.

Thank you for the contribution, but I have to reject since the version on CI should match minimum required version.

okuramasafumi:
I understand it. Should I change my PR to test against from `2.4` to `2.6` (maybe we can skip `2.5`?)?

p0deje:
I think 2.4 is enough for now. Do you see any particular value to have 2.6?

okuramasafumi:
Sorry but I don't have any particular reason. I just thought we should test against latest Ruby.

titusfortner:
Of interest, both Capybara and Watir effectively do a lot of testing of the Ruby bindings in multiple languages.

 Based on the size of the project and the number of things that need to be tested, we need to pick one, and I think picking the lowest supported is the best answer to ensure we don't use syntax on accident that isn't supported (I've done this).

PR #7214, opened 2019-05-16
adding page factory support for python selenium community

ToxicEnvelope:


ToxicEnvelope:
approved by Yahav N. Hoffman

ToxicEnvelope:
added tests for page factory poc on pom

lmtierney:
IMO this belongs in a support library and not in the selenium library itself.

ToxicEnvelope:
@lmtierney , 
Okay :)
To which package should i move it then? 
Thx,
Yahav N.

AutomatedTester:
Personally I don't see why this change is needed in the first place, python allows you do these things already without having the bloat added to bindings. Also... what is with all the bad spending of factory?

lmtierney:
> @lmtierney ,
> Okay :)
> To which package should i move it then?
> Thx,
> Yahav N.

You could create your own OSS package.

ToxicEnvelope:
@AutomatedTester,
Try to use my pattern , and try to compare it to others patterns when you 
create new POMs I guarnetee not such thing as any exception regarding StaleElement.
plus,
It's readable and return multiple cache functions as objects in a flat manner when using it the `find_by` pattern.
You will never again will need to try to identify the reference of WebElement object

Dude-X:
I agree with Lucas and Dave that the API is a bit difficult, that you should create your own project.
Also caching elements doesn't make sense to me since Javascript can change any element right after you retrieve that element.

AutomatedTester:
Closing as this should be it's own package on top of Selenium

PR #7222, opened 2019-05-20
Use java.time.Duration to configure timeouts

valfirst:


java/client/src/org/openqa/selenium/WebDriver.java
```diff
@@ -274,9 +277,31 @@
...
+     * searching for multiple elements, the driver should poll the page until at least one element
+     * has been found or this timeout has expired.
+     * <p>
+     * Increasing the implicit wait timeout should be used judiciously as it will have an adverse
+     * effect on test run time, especially when used with slower location strategies like XPath.
+     *
+     * @param duration The duration to wait.
+     * @return A self reference.
+     */
+    Timeouts implicitlyWait(Duration duration);
```
shs96c: There should be a default implementation of this that delegates down to the old method (for now). People who've extended Selenium in their own projects will have a fighting chance of doing the upgrade to 4.0. In our own code, we should (of course!) always implement this method.

java/client/src/org/openqa/selenium/WebDriver.java
```diff
@@ -286,17 +311,45 @@
...
     /**
+     * Sets the amount of time to wait for an asynchronous script to finish execution before
+     * throwing an error. If the timeout is negative, then the script will be allowed to run
+     * indefinitely.
+     *
+     * @param duration The timeout value.
+     * @return A self reference.
+     * @see JavascriptExecutor#executeAsyncScript(String, Object...)
+     */
+    Timeouts setScriptTimeout(Duration duration);
```
shs96c: As above.

java/client/src/org/openqa/selenium/WebDriver.java
```diff
@@ -286,17 +311,45 @@
...
+    }
+
+    /**
+     * Sets the amount of time to wait for a page load to complete before throwing an error.
+     * If the timeout is negative, page loads can be indefinite.
+     *
+     * @param duration The timeout value.
+     * @return A Timeouts interface.
+     */
+    Timeouts pageLoadTimeout(Duration duration);
```
shs96c: As above.

shs96c (commented):
I like the idea in general, but we need to be mindful of how people will upgrade. Most of your PR is updating tests to use the new APIs --- I can only imagine the scale of the task for many of our users. I also suspect that there are plenty of places where people have (re-) implemented the interfaces, so we should use default implementations that delegate to the old way of doing things to enable those folks to do an upgrade as cleanly as possible.

java/client/src/org/openqa/selenium/WebDriver.java
valfirst: @shs96c done, I've changed the implementation approach of default interface methods

diemol (commented):
@valfirst Could you please solve the conflicts so I can review again to merge this?

valfirst:
@diemol I've resolved the conflicts, could you please review the changes?

java/client/src/org/openqa/selenium/remote/DriverCommand.java
```diff
@@ -210,17 +210,14 @@ static CommandPayload SET_ALERT_VALUE(String keysToSend) {
   String SET_ALERT_CREDENTIALS = "setAlertCredentials";
 
   String SET_TIMEOUT = "setTimeout";
-  static CommandPayload SET_IMPLICIT_WAIT_TIMEOUT(long time, TimeUnit unit) {
-    return new CommandPayload(
-        SET_TIMEOUT, ImmutableMap.of("implicit", TimeUnit.MILLISECONDS.convert(time, unit)));
+  static CommandPayload SET_IMPLICIT_WAIT_TIMEOUT(Duration duration) {
```
diemol, line 221: Not sure if I am missing something, but shouldn't this interface use the same (or at least similar) approach as others? 

If I understand this change correctly, someone could be using `CommandPayload SET_IMPLICIT_WAIT_TIMEOUT(long time, TimeUnit unit)`, and if this change was merged and released, and they upgrade, their code would not compile. Is that right or maybe I am missing something?

diemol: I mean, someone could have implemented their own driver.

diemol (changes requested):

java/client/src/org/openqa/selenium/remote/DriverCommand.java
valfirst: @diemol you are right, the question is what should be considered as public API and what as internal API, anyway I'll add back old methods

diemol: Good point, but if the interface is public we cannot discard that some people have implemnted it.

diemol (approved):
Thank you @valfirst!

PR #7225, opened 2019-05-21
[Grid] Change Session Termination log message from info to warning

dylanlive:
This PR changes a Grid log message from `info` to `warning`. I encountered this error when a session failed to be created, but I wasn't properly alerted of the issue because the log level was at info.

The code comment indicates "should not happen," so if it does happen it would be ideal to be given a warning log.

Not sure if this code is used in Se4, but figured I'd open this in case it is.


dylanlive:
fyi @barancev @diemol @shs96c on this Java Grid PR

diemol:
Hi @dylanlive, this code won't be part of Grid 4. So I think it does not make sense to add since it will never be released.

Thank you for putting the time to send a PR, looking forward to the future ones!

PR #7228, opened 2019-05-23
[rb] Remove unnecessary code in the test

RustyNail:
## Changes

Removed code that is not relevant to test validation.
Since it is verified by `expect ()`, I think that the subsequent (I deleted) code is unnecessary.
Applicable code is included in [this commit](https://github.com/SeleniumHQ/selenium/commit/bc301ee873553704e728365a7a77f573765ea0d6#diff-f8ddf8051137d162a2c487726df96115).

Test passes have been verified with Travis CI results.

### DIFF
--- a/rb/spec/unit/selenium/server_spec.rb
+++ b/rb/spec/unit/selenium/server_spec.rb
@@ -186,11 +186,6 @@ module Selenium
       expect(server.timeout).to eq(30)
       expect(server.background).to be false
       expect(server.log).to be_nil
-
-      server.port = 1234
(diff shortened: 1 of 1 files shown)


lmtierney:
The expects aren't validating the test, they are checking the default settings. The settings you removed were attempts to `sets options after instantiation`, which would error if not possible. Although, I would expect checking the settings after they were set, not just that they didn't error

RustyNail:
My understanding was wrong.
There have two test cases, but there is no verification for the second configuration.

PR #7231, opened 2019-05-24
Optimized DevTools

adiohana:
- Added test cases for Log and Console
- Fixed some parsing bugs
- modified to same convention we have in Network and Performance
- Improved JavaDoc in Network
- Usage of fromJson in all models

### DIFF
--- a/java/client/src/org/openqa/selenium/devtools/network/model/AuthChallenge.java
+++ b/java/client/src/org/openqa/selenium/devtools/network/model/AuthChallenge.java
@@ -49,7 +49,7 @@ public class AuthChallenge {
     this.source = source;
   }
 
-  public static AuthChallenge parseRequest(JsonInput input) {
+  private static AuthChallenge fromJson(JsonInput input) {
 
     String origin = null;
     String realm = null;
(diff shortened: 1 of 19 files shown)


adiohana:
@shs96c 

Added some goodies here. 
This will get us to a stable state in all DevTools domains we have so far.

java/client/src/org/openqa/selenium/devtools/Console.java
```diff
@@ -27,31 +27,47 @@
 import java.util.StringJoiner;
 
 public class Console {
-  private Console() {
```
shs96c: We want to keep the private constructors for utility classes

java/client/src/org/openqa/selenium/devtools/Log.java
```diff
@@ -33,21 +34,23 @@
 
 public class Log {
 
-  private Log() {
```
shs96c: Utility classes (those that only contain `static` helper methods and classes) should have `private` constructors to stop people accidentally attempting to use the static methods as instance methods. Because "people"

PR #7245, opened 2019-05-29
[rb] Add aggregate_failures if there is more than one 'expect()'.

RustyNail:
By putting `aggregate_failures`, we can check them at once, even if all `expect()` fail.

The files changed in this PullRequest are below.
* rb / spec / unit / selenium / webdriver / proxy_spec.rb

Addition of aggregate_failures seems to be applicable to other than the file which I changed this time.
If you can fix other files as well, I will make another PullRequest and respond.

Test passes have been verified with [Travis CI results](https://travis-ci.org/RustyNail/selenium/builds/538116780).

### DIFF
--- a/rb/spec/unit/selenium/webdriver/spec_helper.rb
+++ b/rb/spec/unit/selenium/webdriver/spec_helper.rb
@@ -42,8 +42,4 @@ RSpec.configure do |c|
   c.include Selenium::WebDriver::UnitSpecHelper
 
   c.filter_run focus: true if ENV['focus']
-
-  c.define_derived_metadata do |meta|
-    meta[:aggregate_failures] = true
-  end
 end


twalpole:
This would be cleaner done with `aggregate_failures` metadata form rather than the block from.

RustyNail:
@twalpole 
Thank you for the advice.
Moved definition to spec / spec_helper.rb  (defined in meta data).

twalpole:
Sorry - I should have been clearer. I didn't mean to add it to every test - there are some tests where it doesn't make sense to use failure aggregation (where later expects can't possibly succeed if a previous failure occurs).  The tests you were adding it to did make sense and those could have been done by adding the metadata to the test description rather than using the block form

    it 'should allow valid options for a manual proxy', :aggregate_failures do

RustyNail:
@twalpole  I'm sorry, my understanding was wrong.
Changed the format to add metadata (: aggregate_failures) to tests that check multiple properties.

twalpole:
LGTM

twalpole:
Thanks!

PR #7253, opened 2019-05-31
Add support for the new MicrosoftEdge - Python

loly89:
- Extracting Chrome into a separate Chromium package since the new Edge is also Chromium-based.
- Add the ability for EdgeDriver to launch the new MicrosoftEdge using MSEdgeDriver without breaking any existing Edge tests using is_legacy parameter, which defaults to True.

- [X] By placing an `X` in the preceding checkbox, I verify that I have signed the [Contributor License Agreement](https://github.com/SeleniumHQ/selenium/blob/master/CONTRIBUTING.md#step-6-sign-the-cla)

PR #7257, opened 2019-06-03
Add Chrome based Edge to the Ruby bindings

twalpole:

This is the start of adding Chrome based Edge to the Ruby bindings.  We need to decide whether to implement this separately from legacy edge, or to accept some sort of configuration option to switch between legacy and chrome based - thoughts?

### DIFF
--- a/rb/build.desc
+++ b/rb/build.desc
@@ -73,7 +73,7 @@ ruby_test(name = "chrome",
 
 ruby_library(name = "edge",
   srcs = [
-    "lib/selenium/webdriver/edge_legacy/**/*.rb",
+    "lib/selenium/webdriver/edge_html/**/*.rb",
     "lib/selenium/webdriver/edge.rb"
   ],
   deps = [":common"]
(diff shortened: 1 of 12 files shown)


twalpole:
I've updated this to move the edge 18 into EdgeLegacy module, with Edge being an alias for it.  Swapping from Edge/:edge being Legacy to Chrome based should be simple if/when chrome based edge goes release

p0deje:
This is fine by me. I know that Java calls legacy `EdgeHtml` while Python keeps old name and calls a new one `ChromiumEdge` (#7253). We might want to make the naming consistent when other bindings support lands, but that's for the future.

twalpole:
I swapped the legacy edge namespace to EdgeHtml since that is what the engine is called -- thanks for pointing that out @p0deje.  Anyone have issues with me squashing one merging this now?

rb/spec/integration/selenium/webdriver/edge_chrome/options_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+        it 'passes args correctly' do
+          options.add_argument('--user-agent=foo;bar')
+
+          create_driver!(options: options) do |driver|
+            ua = driver.execute_script 'return window.navigator.userAgent'
+            expect(ua).to eq('foo;bar')
+          end
+        end
+
+        # it 'should be able to run in headless mode with #headless!' do
```
p0deje: What about this one?

p0deje:
> Anyone have issues with me squashing and merging this now?
No issues from my side. Thanks for tackling this!

rb/spec/integration/selenium/webdriver/edge_chrome/options_spec.rb
twalpole: It currently crashes when trying to start headless - not sure if they plan to support a headless mode or not

PR #7264, opened 2019-06-05
[rb] skip pending by default on local

titusfortner:
@p0deje  are you ok with this code (69ee9211b437316f0114e3f98e2511f6caf9f9ee) being default when not run on a ci? I don't think there is a way to toggle this specifically for `./go` vs `IntelliJ`

### DIFF
--- a/rb/spec/integration/selenium/webdriver/spec_helper.rb
+++ b/rb/spec/integration/selenium/webdriver/spec_helper.rb
@@ -32,6 +32,7 @@ RSpec.configure do |c|
   c.include(WebDriver::SpecSupport::Helpers)
 
   c.before(:suite) do
+    ENV['SKIP_PENDING'] = 'true' unless WebDriver::Platform.ci
     $DEBUG ||= ENV['DEBUG'] == 'true'
     GlobalTestEnv.remote_server.start if GlobalTestEnv.driver == :remote
     GlobalTestEnv.print_env


p0deje:
The point of pending guards is to notify when the spec starts passing. I find it useful when refactoring or browser version bump makes some of the tests green. Do you have problems with this mechanism?

titusfortner:
The issue is that tests in files like `firefox/driver_spec.rb` are marked "firefox only," but will actually pass when run with Chrome.
When testing via `./go` command it specifies not to run those tests.
When testing by calling rspec directly (or right-clicking the directory in IntelliJ), you'll see a bunch of failures that should not be failures.

p0deje:
Oh, this is definitely an issue. However, I believe that we should probably run all the specs all the time when using `./go` and update the guards accordingly.

p0deje:
I'd go with option 3, so that only driver-specific tests live in driver specs directory (e.g. specs for network conditions ./rb/spec/integration/selenium/webdriver/chrome/driver_spec.rb), while non-driver specific go to ./rb/spec/integration/selenium/webdriver/driver_spec.rb. Once this is done, I'd just ensure that _all_ specs are run with `./go`, so that there is no difference between running tests locally or on CI.

titusfortner:
This PR is not the right way to do this, added issue to track it.

PR #7265, opened 2019-06-05
Init options

titusfortner:
I don't like the current approach for initializing requiring undefined parameters as "options" which is already overloaded. I would like to deprecate it.

This might be overkill if we want to just allow anything in the constructor, but this also provides users with an `attr_accessor` if they want to add known supported things rather than using `#add_option` for it.

Also, IE takes a distinctly different approach because there is a map for most of the settings (e.g. `ignore_zoom_level: 'ignoreZoomSetting'`) that we iterate through. We could possibly implement something similar for these drivers, though.

### DIFF
--- a/rb/lib/selenium/webdriver/chrome/options.rb
+++ b/rb/lib/selenium/webdriver/chrome/options.rb
@@ -21,12 +21,23 @@ module Selenium
   module WebDriver
     module Chrome
       class Options < WebDriver::Common::Options
-        attr_reader :args, :prefs, :options, :emulation, :extensions, :encoded_extensions
-        attr_accessor :binary
 
         KEY = 'goog:chromeOptions'
 
-        #
+        # see: http://chromedriver.chromium.org/capabilities
+        CAPABILITIES = %i[args binary extensions local_state prefs detach debugger_address exclude_switches
+                          minidump_path mobile_emulation perf_logging_prefs window_types].freeze
+
+        (CAPABILITIES + %i[options emulation encoded_extensions]).each do |key|
+          define_method key do
+            @options[key]
+          end
+
+          define_method "#{key}=" do |value|
+            @options[key] = value
+          end
+        end
+
         # Create a new Options instance.
         #
         # @example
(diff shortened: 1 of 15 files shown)


titusfortner:
Ok, I redid this to follow the IE pattern in all of the drivers.
Chrome & Firefox behave slightly differently than Edge, IE and Safari

You have attr_accessors for everything
You have the same behavior for initializing that you have with accessors that you have with the sugary methods
You have every documented attribute directly supported in each class with reference to where they are documented
You no longer have to pass in the overloaded `:options` keyword in the constructor, it just takes everything left and throws it in the `@options`

I think there are a couple more things to tidy up here, but let me know if I'm off base on this.

Definitely things that can be moved to super class if we agree on this pattern.

p0deje (approved):
I like that! My only concern here is that we should expose `#add_option` so that anyone can modify it directly in case we don't support the option yet. Maybe we should just add it to `WebDriver::Options`. Right now it's not possible for Edge and Safari.

rb/lib/selenium/webdriver/chrome/options.rb
```diff
@@ -155,30 +181,50 @@ def headless!
         # @param [String] user_agent Full user agent
```
p0deje: Do we need to change YARD now that you use double-splat parameter?

rb/spec/unit/selenium/webdriver/chrome/options_spec.rb
```diff
@@ -26,43 +26,51 @@ module Chrome
         subject(:options) { described_class.new }
 
         describe '#initialize' do
-          it 'sets passed args' do
-            opt = Options.new(args: %w[foo bar])
-            expect(opt.args.to_a).to eq(%w[foo bar])
-          end
+          it 'accepts defined parameters' do
+            allow(File).to receive(:file?).and_return(true)
+            # allow_any_instance_of(Options).to receive(:encode_file).with('foo.crx').and_return("encoded_foo")
```
p0deje: Should this just be removed?

rb/lib/selenium/webdriver/chrome/options.rb
```diff
@@ -21,12 +21,23 @@ module Selenium
...
 
         KEY = 'goog:chromeOptions'
 
-        #
+        # see: http://chromedriver.chromium.org/capabilities
+        CAPABILITIES = %i[args binary extensions local_state prefs detach debugger_address exclude_switches
+                          minidump_path mobile_emulation perf_logging_prefs window_types].freeze
+
+        (CAPABILITIES + %i[options emulation encoded_extensions]).each do |key|
+          define_method key do
```
p0deje: How does this look in docs? Should we maybe use macro for better documentation?

titusfortner:
> we should expose #add_option so that anyone can modify it 

The issue with this is that Edge, IE & Safari all have "special" parameter names that include the colons. I don't love `#add_option` in the first place and would prefer to leave it where it is at least more straightforward.

rb/lib/selenium/webdriver/chrome/options.rb
titusfortner: actually, I don't think so? We're still spelling these things out in #initialize and I think it's ok to specify in the comments exactly which opts are supported

titusfortner:
@p0deje ok, changed my mind again and put `#add_option` in superclass.
Let me know if changes here look ok: https://github.com/SeleniumHQ/selenium/pull/7265/commits/c47a70088239c71212a38bd530f91d8132f42fad

@twalpole any issues with this?

rb/lib/selenium/webdriver/chrome/options.rb
p0deje: Right, but should this now be changed to:

```ruby
# @param [Hash] opts
# @option opts [String] :user_agent Full user agent
# ...
```

p0deje: It actually complains on these docs:

```
$ ./go //rb:docs
...
[warn]: @param tag has unknown parameter name: device_name
    in file `build/rb/lib/selenium/webdriver/chrome/options.rb' near line 165
[warn]: @param tag has unknown parameter name: device_metrics
    in file `build/rb/lib/selenium/webdriver/chrome/options.rb' near line 165
[warn]: @param tag has unknown parameter name: user_agent
    in file `build/rb/lib/selenium/webdriver/chrome/options.rb' near line 165
```

rb/lib/selenium/webdriver/chrome/options.rb
p0deje: What I'm trying to say is that these methods are completely hidden in docs right now. YARD is not able to understand them since they are generated dynamically. I was wondering if using [macros](https://www.rubydoc.info/gems/yard/file/docs/Tags.md#macro) could make them visible in docs, but I was wrong.

PR #7273, opened 2019-06-09
[rb] Restore logs support through get_log command on W3C bridge

lfittl:
Selenium::WebDriver::Logs tries to use this and fails, because the get_log
command was only added to the OSS bridge, not the W3C bridge:

```
NoMethodError:
       undefined method `log' for #<Selenium::WebDriver::Remote::Bridge:0x0000000008f7ea60>
     # /.../webdriver/common/logs.rb:32:in `get'
```

However, this seems to work fine when added to the W3C bridge, so this
adds the support back in.

---

### DIFF
--- a/rb/lib/selenium/webdriver/remote/bridge.rb
+++ b/rb/lib/selenium/webdriver/remote/bridge.rb
@@ -450,6 +450,27 @@ module Selenium
         end
 
         #
+        # logs
+        #
+
+        def available_log_types
+          types = execute :get_available_log_types
+          Array(types).map(&:to_sym)
+        end
+
+        def log(type)
+          data = execute :get_log, {}, {type: type.to_s}
+
+          Array(data).map do |l|
+            begin
+              LogEntry.new l.fetch('level', 'UNKNOWN'), l.fetch('timestamp'), l.fetch('message')
+            rescue KeyError
+              next
+            end
+          end
+        end
+
+        #
         # element properties
         #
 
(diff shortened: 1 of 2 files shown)


jimevans:
Contrary to this suggestion, the commands do not exist in most implementations that follow the W3C Specification. It won’t work for Firefox, Safari, or IE for sure. It may work for Chrome at present, but likely won’t in Chrome (and chromedriver) 75 or 76. Also, Chrome 74 does not use the W3C bridge by default. Though I do not regularly use or maintain the Ruby bindings, I would recommend against this PR.

twalpole:
The decision was made to remove non-w3c compliant methods in Selenium 4, and while log access can currently (Chrome 75) be made to work locally in w3c mode, it's unlikely to work correctly remotely.  We're not sure if/when Google will remove the logs endpoint from chromedriver, however if they change it to be a W3C compliant extension endpoint (`/session/:session_id/goog/logs` or equivalent) we'll consider adding it back in.  Until then if you must have access to the logs your best bet is to stick to selenium-webdriver 3.x and turn off `w3c` mode when using chromedriver.

lfittl:
@twalpole Makes sense - should the `Selenium::WebDriver::Logs` be removed then in version 4? (right now its presence suggests this would work, but will fail when actually used)

twalpole:
@lfittl We're discussing that currently. It looks like Google will be adding the log methods to chromedriver as w3c compliant extensions in version 76 or 77 so we're not removing it yet.

PR #7287, opened 2019-06-13
[rb] Add missing verification in timeout_spec.rb

RustyNail:
I added verification because there is no verification in one of the tests in timeout_spec.rb.
Target test : `it 'should implicitly wait for a single element'`

Test passes have been verified with [Travis CI results](https://travis-ci.org/RustyNail/selenium/builds/545124587).

### DIFF
--- a/rb/spec/integration/selenium/webdriver/timeout_spec.rb
+++ b/rb/spec/integration/selenium/webdriver/timeout_spec.rb
@@ -34,7 +34,7 @@ module Selenium
           driver.manage.timeouts.implicit_wait = 6
 
           driver.find_element(id: 'adder').click
-          expect { driver.find_element(id: 'box0') }.not_to raise_error(WebDriver::Error::NoSuchElementError)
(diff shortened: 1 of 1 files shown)


rb/spec/integration/selenium/webdriver/timeout_spec.rb
```diff
@@ -34,7 +34,7 @@ module WebDriver
           driver.manage.timeouts.implicit_wait = 6
 
           driver.find_element(id: 'adder').click
-          driver.find_element(id: 'box0')
+          expect { driver.find_element(id: 'box0') }.not_to raise_error(WebDriver::Error::NoSuchElementError)
```
twalpole: This should just be `.not_to raise_error` otherwise it potentially hides errors

twalpole:
Thanks!

rb/spec/integration/selenium/webdriver/timeout_spec.rb
RustyNail: Thank you for your advice. Fixed it.

PR #7307, opened 2019-06-19
deprecate parameters getting passed into the driver directly

titusfortner:
Note that this is for adding deprecations to latest 3.x code (and PR is to that branch, not master) so we can delete it all in 4.x

Essentially looking to get rid of things getting passed directly into driver initialization that belong in the Options or Capabilities classes.

PR #7308, opened 2019-06-19
[rb] remove unnecessary methods

titusfortner:
Main issue here is that any given capabilities/options should do the same thing locally that it does remotely, so we need to remove unnecessary processing that doesn't happen in Remote class.

`Remote::Bridge#create_capabilities` merges options & capabilities, so there is no need to do it in any of these classes.
https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/remote/bridge.rb#L65

3 of the items I removed in `Chrome::Driver#create_capabilities` need to be deprecated in 3.x before we just remove them, which I proposed in #7307

### DIFF
--- a/rb/lib/selenium/webdriver/chrome/options.rb
+++ b/rb/lib/selenium/webdriver/chrome/options.rb
@@ -21,6 +21,7 @@ module Selenium
   module WebDriver
     module Chrome
       class Options < WebDriver::Common::Options
+        attr_accessor :profile
 
         KEY = 'goog:chromeOptions'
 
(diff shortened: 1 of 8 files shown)


twalpole (approved):
LGTM

rb/lib/selenium/webdriver/chrome/options.rb
```diff
@@ -70,9 +71,10 @@ class Options < WebDriver::Common::Options
         # @option opts [Array<String>] :window_types A list of window types to appear in the list of window handles
         #
 
-        def initialize(encoded_extensions: nil, **opts)
+        def initialize(profile: nil, encoded_extensions: nil, **opts)
```
p0deje: You forgot to update YARD.

p0deje (approved):

titusfortner:
merged with https://github.com/SeleniumHQ/selenium/commit/3490d4a5613cd426a94fe280513a2e0b7e935ae1

PR #7312, opened 2019-06-20
[rb] initialize all of the drivers in superclass

titusfortner:
All of the drivers are now pretty much doing the exact same thing, so there isn't a need to have the code duplicated in each file.

The only major issue I had was extending the browser-specific Bridge modules. This code changes them into subclasses, which works, but I don't love, especially because it is adding files that don't actually do anything (right now), just so that the superclass can  make all the method calls it needs. 

Should we define the subclasses at the end of `remote/bridge.rb` file rather than creating 2 extra files? Or is the a better approach entirely?

### DIFF
--- a/rb/lib/selenium/webdriver/common/driver.rb
+++ b/rb/lib/selenium/webdriver/common/driver.rb
@@ -71,10 +71,10 @@ module Selenium
       # @api private
       #
 
-      def initialize(bridge, listener: nil)
+      def initialize(bridge: nil, listener: nil, **opts)
         @service = nil
-        @bridge = bridge
-        @bridge = Support::EventFiringBridge.new(bridge, listener) if listener
+        bridge ||= create_bridge(opts)
+        @bridge = listener ? Support::EventFiringBridge.new(bridge, listener) : bridge
       end
 
       def inspect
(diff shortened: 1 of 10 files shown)


rb/lib/selenium/webdriver/common/driver.rb
```diff
@@ -71,9 +71,28 @@ def for(browser, opts = {})
...
-        @bridge = bridge
-        @bridge = Support::EventFiringBridge.new(bridge, listener) if listener
+      def initialize(bridge: nil, listener: nil, **opts)
+        bridge ||= begin
+          opts[:url] ||= service_url(opts)
+
+          default_caps = @bridge&.browser || :new
+          desired_capabilities = opts.delete(:desired_capabilities) || Remote::Capabilities.send(default_caps)
+          if desired_capabilities.is_a?(Symbol)
+            unless Capabilities.respond_to?(desired_capabilities)
```
p0deje: This should be `Remote::Capabilities`.

rb/lib/selenium/webdriver/common/driver.rb
```diff
@@ -71,9 +71,28 @@ def for(browser, opts = {})
...
+              raise Error::WebDriverError, "invalid desired capability: #{desired_capabilities.inspect}"
+            end
+
+            desired_capabilities = Capabilities.__send__(desired_capabilities)
```
p0deje: Same: `Remote::Capabilities`.

rb/lib/selenium/webdriver/edge_html/bridge.rb
```diff
@@ -0,0 +1,29 @@
...
+# under the License.
+
+require 'selenium/webdriver/chrome/bridge'
+
+module Selenium
+  module WebDriver
+    module EdgeHtml
+      class Bridge < WebDriver::Remote::Bridge
+      end # Bridge
+    end # EdgeChrome
```
p0deje: EdgeHtml

p0deje (commented):
> The only major issue I had was extending the browser-specific Bridge modules. This code changes them into subclasses, which works, but I don't love, especially because it is adding files that don't actually do anything (right now), just so that the superclass can make all the method calls it needs.

> Should we define the subclasses at the end of remote/bridge.rb file rather than creating 2 extra files? Or is the a better approach entirely?

One option would be to just look up the constant and use it as a bridge if defined, falling back to default remote if it's not defined. This way you can even avoid `Driver.bridge` methods as they can be inferred from driver class name.

titusfortner:
@p0deje I implemented it with the approach you suggested, this is what I came away with:

```ruby
namespacing = self.class.to_s.split('::')

if defined?(Object.const_get("#{namespacing[0, namespacing.length - 1].join('::')}::Bridge")) &&
          !namespacing.include?('Remote')
  bridge.extend Object.const_get("#{namespacing[0, namespacing.length - 1].join('::')}::Bridge")
end
```

I think I like it better, even though this code is ugly.

twalpole:
@titusfortner You shouldn't need the "namespace.length - 1" stuff - you should be able to use range with a negative index `namespacing[0..-2]`

twalpole:
Also I think you probably want to use `const_defined?`

    if Object.const_defined?(namespacing[0..-2].push("Bridge").join('::')) && !namespacing.include?('Remote')

titusfortner:
Thanks @twalpole, those changes make it cleaner.

PR #7316, opened 2019-06-22
Split service process control out to service manager class

twalpole:

@titusfortner This splits the actual service process creation/management out to separate classes.  This would solve the reuse in driver registrations issue and I don't think it adds complexity (actually simplifies the Service classes).  I could also get rid of the browser specific ServiceManager classes and just pass in the browser name to the ServiceManager class and have the shutdown supported be a Hash in it since that's the only difference between them.  Although having it this way does allow for people to add ServiceManager classes if they want/need to (not sure if that is a consideration for Sauce, etc) Thoughts?


twalpole:
Note - this would probably need to get implemented in the 3.x branch too

twalpole:
I rethought this a bit and changed it so there are no more separate ServiceManager classes, and the Service starts the ServiceManager.  Let me know what you think.

titusfortner:
merged: 6ec98a35a1494311ec6377d0670333415b8aa225

Issue #5335, opened 2018-01-12
SessionNotCreatedException when specifying `safari.options` and `port`


timsutton-sl:
## Meta -
OS:  macOS 10.13.2

Selenium Version:  
3.8.0/3.8.1

Browser:  
Safari

Browser Version:  
11.0.2

## Expected Behavior -

If we start a remote driver with the following desired caps:

```python
{'browserName': "safari", 'safari.options': {'port': 1034}
```

## Actual Behavior -

The following exception and backtrace:

```java
SessionNotCreatedException: Message: Unable to create new service: SafariDriverService
Build info: version: '3.8.1', revision: '6e95a6684b', time: '2017-12-01T19:05:32.194Z'
System info: host: 'Tim-Sutton-SL0280.local', ip: 'fe80:0:0:0:8d6:1d46:d1a5:993d%en0', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.13.2', java.version: '9.0.1'
Driver info: driver.version: unknown
Stacktrace:
    at org.openqa.selenium.remote.server.ServicedSession$Factory.lambda$get$0 (ServicedSession.java:208)
    at org.openqa.selenium.remote.server.ServicedSession$Factory.apply (ServicedSession.java:218)
    at org.openqa.selenium.remote.server.ActiveSessionFactory.lambda$apply$11 (ActiveSessionFactory.java:167)
    at java.util.stream.ReferencePipeline$3$1.accept (ReferencePipeline.java:195)
    at java.util.stream.ReferencePipeline$3$1.accept (ReferencePipeline.java:195)
    at java.util.stream.ReferencePipeline$11$1.accept (ReferencePipeline.java:375)
    at java.util.stream.ReferencePipeline$2$1.accept (ReferencePipeline.java:177)
    at java.util.Spliterators$ArraySpliterator.tryAdvance (Spliterators.java:958)
    at java.util.stream.ReferencePipeline.forEachWithCancel (ReferencePipeline.java:127)
    at java.util.stream.AbstractPipeline.copyIntoWithCancel (AbstractPipeline.java:502)
    at java.util.stream.AbstractPipeline.copyInto (AbstractPipeline.java:488)
    at java.util.stream.AbstractPipeline.wrapAndCopyInto (AbstractPipeline.java:474)
    at java.util.stream.FindOps$FindOp.evaluateSequential (FindOps.java:152)
[...]

PR #7327, opened 2019-06-25
Fix random permission problems when trying to obtain a free port

adymorz:
On Windows 10, about every 100th request to obtain a free port fails with EACCESS error. This small change fixed the issue.

I did not write a new tests for this issue. Do you have a test strategy for infrequent problems?

### DIFF
--- a/javascript/node/selenium-webdriver/net/portprober.js
+++ b/javascript/node/selenium-webdriver/net/portprober.js
@@ -149,7 +149,7 @@ async function findWindowsPortRange() {
 function isFree(port, opt_host) {
   return new Promise((resolve, reject) => {
     let server = net.createServer().on('error', function(e) {
-      if (e.code === 'EADDRINUSE') {
+      if (e.code === 'EADDRINUSE' || e.code === 'EACCES') {
         resolve(false);
       } else {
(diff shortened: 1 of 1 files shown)


adymorz:
Uh oh. The tests did not run through.

I was unable to find the "2 failing" test in job #5335.1.  In job #5335.5 I see the error message "FAILURE org.openqa.selenium.javascript.ClosureTestSuite : connection refused" which tells me nothing.

How should I proceed?
Revert my change temporarerely and see, if the test works again?

I know that in an ideal world, I would set up the whole test infrastructure locally and would solve the problems myself. Unfortunaltely I do not have the capacity to do so.

tapioko:
In line ~180 ("throw Error('Unable to find a free port');") can you add text about possible permission problem, too? Maybe that would also fix the tests... :)

barancev:
I think we can accept this PR without a test, it's not obvious how to make the system throw EACCESS. I tried to create a test that checks ports with numbers < 1024, but it's unstable because there is no exception in some environments.

barancev:
Merged as 78dce0c505f545a2111fcea65745cfd2117a825f, thank you for the contribution!

Issue #6667, opened 2018-11-20
Java implementation of DevTools API


shs96c:
This should allow users to take advantage of features offered by using the Chrome Debugging Protocol (and equivalents on other browsers), notably:

* Full page screenshots
* Streaming logs
* Waiting for mutation events on the page

shs96c:
There's code for this in https://github.com/shs96c/selenium/tree/cdp

joywave:
This would be really nice to have. I looked at this today and some previous attempts were https://github.com/SeleniumHQ/selenium/pull/5262 https://github.com/SeleniumHQ/selenium/issues/5159

adiohana:
@shs96c 
I see that this is part of Selenium 4. **Is there a point for me to work on the PR now?** 
Chrome DevTools integration requires WebSocket functionality and some changed in RemoteWebDriver which is kind of a big change

barancev:
Initial support has been implemented:
f280b8b8232824ee432917f068c6b79990bde08a
a99534efddef9a27d37d41a73ae31391f03300e0
58298638deb63a91a22c6ae8f0309faf4e6d434e

shs96c:
@adiohana Absolutely! We're super-keen on coming up with a nicer way of presenting this in the user-facing APIs, and we really need to find a way of generating the raw CDP classes rather than writing it all by hand.

adiohana:
@shs96c 
Thanks for the confirmation, we will do it soon

adiohana:
@shs96c 
seems that generating the raw CDP is a complicated task and can be fragile because we are counting on Google developers not to change their CDP json schema as stated here: https://github.com/ChromeDevTools/devtools-protocol/blob/master/json/browser_protocol.json

I see that you already implemented (partially) the following domains:
* Target
* Runtime
* Console
* Log

I think that the 2 most critical domains (for the features I stated above and more) are:
*Network
*Performance

Do you think it worth implementing Network and Performance "manually" while we are working on the auto generation of the CDP classes?

shs96c:
@adiohana, writing those two manually makes sense as a stepping stone to getting everything running. If you've got the time, I'd be very happy to review a PR :) If you need a hand, I tend to lurk on the #selenium IRC and Slack channels (London work hours), and I can step you through anything you're finding tricky to deal with.

adiohana:
Thanks @shs96c 
I am working on this together with dratler.shay@gmail.com 
we will submit the PRs tomorrow probably.
[...]

PR #7345, opened 2019-06-29
Code review rejects

dratler:
Hi @shs96c  here is Target PR rejects
related to  #6667
1. Fix typo's
2. use Selenium Session
3. Fix tests

#behalfOf_Intuit

### DIFF
--- a/java/client/test/org/openqa/selenium/devtools/ChromeDevToolsProfilerTest.java
+++ b/java/client/test/org/openqa/selenium/devtools/ChromeDevToolsProfilerTest.java
@@ -39,20 +39,8 @@ import org.openqa.selenium.devtools.profiler.model.ScriptCoverage;
 import java.util.List;
 import java.util.Optional;
 
-
 public class ChromeDevToolsProfilerTest extends ChromeDevToolsTestBase {
 
-
-
-  @Test
-  public void aSimpleStartStopAndGetProfilerTest() {
-    devTools.send(enable());
-    devTools.send(start());
-    Profile profiler = devTools.send(stop());
-    validateProfile(profiler);
-    devTools.send(disable());
-
-  }
 
   private void validateProfile(Profile profiler) {
     Assert.assertNotNull(profiler);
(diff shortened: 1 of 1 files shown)


java/client/src/org/openqa/selenium/devtools/target/model/AttachToTarget.java
```diff
@@ -17,44 +17,42 @@
...
     this.sessionId = Objects.requireNonNull(sessionId, "sessionId is required");
-    this.targetId = Objects.requireNonNull(targetId, "targetId is required");
+    this.targetInfo = Objects.requireNonNull(targetInfo, "targetInfo is required");
     this.waitForDebugger = Objects.requireNonNull(waitForDebugger, "waitForDebugger is require");
   }
 
   private static AttachToTarget fromJson(JsonInput input) {
-    SessionId sessionId = null;
-    TargetId targetId = null;
+    SessionId sessionId = new SessionId(input.nextString());
```
shs96c: Shouldn't this be inside the loop? And I don't see a call to `input.beginObject`, so I'm a little surprised this works....

shs96c: Also, you added a `fromJson` method, so it should be fine to `input.read(SessionId.class)`

dratler:
Hi Simon,
I have added the requested comments .
also I have open a new CR for GCD (there is one more CR on my behalf).
if you have other tasks I can assist please do I would like to .

Cheers
Shay

java/client/src/org/openqa/selenium/devtools/target/model/AttachToTarget.java
dratler: this is how the Json Object is build .
I have tested it more the once

shs96c:
I rebased this manually, merged and pushed as f323d601ae9f56e6bdfc3911523dadcc6bc29ab2. Thank you!

Issue #7335, opened 2019-06-27
C# unable to enable performance logging for ChromeDriver


JL1234567:
Chrome:Version 75.0.3770.100
ChromeDriver:75.0.3770.90
Selenium:3.141.0

I'm getting the error ```
OpenQA.Selenium.WebDriverException: 'invalid argument: entry 0 of 'firstMatch' is invalid
from invalid argument: perfLoggingPrefs specified, but performance logging was not enabled'

If the work around is to use LoggingPreference, can someone please tell me how to reference it since it does not seem to be in the 3.141.0 dll. Also, I think this might relate to the changes to 75.0.3770.8 Renamed capability loggingPrefs to goog:loggingPrefs, as required by W3C standard

```
public Driver(ChromeDriverModel chromeDriverModel)
        {
            ChromeOptions chromeOptions = new ChromeOptions();
            enablePerformanceMonitor = chromeDriverModel.enablePerformanceMonitoring;
            if (enablePerformanceMonitor)
                chromeOptions = _ChromePerformanceOptions();

            if (!string.IsNullOrWhiteSpace(chromeDriverModel.ChromeDriverLocation))
            {
                if (enablePerformanceMonitor)
                    _webDriver = new ChromeDriver(chromeDriverModel.ChromeDriverLocation, chromeOptions);
                else
                    _webDriver = new ChromeDriver(chromeDriverModel.ChromeDriverLocation);
            }
            else
            {
                if (enablePerformanceMonitor)
                    _webDriver = new ChromeDriver(chromeOptions);
                else
                    _webDriver = new ChromeDriver();
            }
        }


private ChromeOptions _ChromePerformanceOptions()
        {
            var option = new ChromeOptions();
            var perfLogPrefs = new ChromePerformanceLoggingPreferences();
            perfLogPrefs.AddTracingCategories(new string[] { "devtools.network", "devtools.timeline" });
            option.PerformanceLoggingPreferences = perfLogPrefs;
            option.AddAdditionalCapability(CapabilityType.EnableProfiling, true, true);
            option.SetLoggingPreference("performance", LogLevel.All);
            return option;
        }
```
```

Sorry if repost, not sure which is the official spot.

SlavikCA:
related to https://github.com/SeleniumHQ/selenium/issues/7342

MatthewSteeples:
Potential workaround code for C# (until the library is fixed). This will update the value of the preference string as long as it's run before you make use of the options

[...]

PR #7358, opened 2019-07-04
Update LoggingPrefs in C# to be compatible with ChromeDriver 75+ (#7335)

MatthewSteeples:
This will be needed to start fixing #7335. I'm still experiencing some problems, but I can't work out if that's just because of my local configuration


jimevans:
@MatthewSteeples Thanks very much for this contribution. Unfortunately, as of fafa4479, it's obsolete, as the .NET bindings no longer use `CapabilityType.LoggingPreferences`. Since no other implementation implements the logging preferences capability, a new constant was created in the `ChromiumOptions` class using the (now) correct `goog:loggingPrefs` value. That change was released with 4.0 alpha 2 of the .NET bindings.

MatthewSteeples:
That's great. More than happy to be obsoleted by a better method

PR #7371, opened 2019-07-08
CDP Storage domain

adiohana:

### DIFF
--- a/common/src/web/devToolsDomStorage.html
+++ b/common/src/web/devToolsDomStorage.html
@@ -1,0 +1,10 @@
+<html>
+<head>
+</head>
+
+<body>
+
+<p>DOM Storage Test</p>
+
+</body>
+</html>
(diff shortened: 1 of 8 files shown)


jimevans:
I wonder if we really want to be supporting domains marked as `experimental` in the protocol definition. I'd be worried these may be likely to change on a whim, since the protocol is not subject to any published standard, and doing so might tie us to specific versions of Chrome/other Chromium-based browsers instead of being as version-agnostic as possible. I'm not a Java maintainer, so I don't really have a full say in this decision, but @shs96c really ought to weigh in.

shs96c:
What I _really_ want to do is autogenerate the CDP bindings from the JSON description. The way to do that would be to generate a Command or Function per method, and then group them into protocol versions. That way, as methods move from "experimental" to production, we don't have to worry about users needing to modify their code.

AutomatedTester:
I agree with @jimevans here. We dont want to do anything experimental

AutomatedTester:
Thank you for your effort here. 

We are now auto generating these APIs moving forward so won't need this PR.

Thank you again!

PR #7374, opened 2019-07-08
Audit by Google DevTools

dratler:
1. Build Audit from Google Dev Tools
2. Add model
3. Add test
#6667 


jimevans:
I wonder if we really want to be supporting domains marked as `experimental` in the protocol definition. I'd be worried these may be likely to change on a whim, since the protocol is not subject to any published standard, and doing so might tie us to specific versions of Chrome/other Chromium-based browsers instead of being as version-agnostic as possible. I'm not a Java maintainer, so I don't really have a full say in this decision, but @shs96c really ought to weigh in.

dratler:
Hi @jimevans ,
I the Google DevTools might flag some domains as `experimental` (or even methods) but these methods are very useful for measuring the page performance , behavior and more .
I think not implementing these features will lose some major capabilities.

jimevans:
@dratler I’m going to push back here just a little, because I want to make sure we are thinking everything through. I’m not questioning the value or utility of domains in the protocol.

In this PR, you’re implementing the CDP `Audit` domain, which is part of the “tip-of-tree” version of the protocol, and not the “1.3 stable” version, and is marked as “experimental.” The methods and events of that domain are valid for current shipping versions of Chrome. What happens when Google arbitrarily decides to remove methods or events from the domain? What if they do so without marking them as obsolete (a perfectly valid choice, since the domain is experimental anyway)? How do you propose to handle that in compiled Java code? Just let users’ code throw exceptions if they’ve used something that changed? Release a new version of Selenium every time CDP changes, potentially forcing users into a frighteningly frequent update cycle? What about removed methods in the protocol? Would you simply remove them in the updated Selenium code, leading to compile errors for Selenium users? These questions are equally valid for any other experimental CDP domain. 

I’m truly not trying to be harsh here, and I’m not by any means saying we shouldn’t implement experimental domains. But implementing them has some serious downstream implications for users, and we need to give those some serious thought. This is doubly so for statically typed languages like Java or C#, where library updates can be a significant point of friction for users.

shs96c:
As I've said before, we really should be code-genning these, so we can mark out `beta` APIs as being betas, and so we don't have to endlessly battle against my slow reviews causing us endless rebase miseries. I've spent the day so far manually rebasing a ton of PRs. Looking at the number of files in this PR, can I please ask @dratler to do the rebase, please?

dratler:
Hi @shs96c ,
please share what needs to be fixed and I will do it ASAP.
regarding auto generating java classes from json I'm still investigating this one it's quite tricky and not all json uses the JSON schema

AutomatedTester:
Thanks for this PR, we are auto generating the CDP code now so I dont think we need this moving forward. I am closing the PR

Issue #7219, opened 2019-05-18
3.142.2 can't find Firefox on FreeBSD


PeterMozesMerl:
## 🐛 Bug Report

Upgrading to 3.142.2 made selenium-webdriver lose its interest in running my examples and depress me with throwing "`page.current_path` raised exception: #Selenium::WebDriver::Error::SessionNotCreatedError: Expected browser binary location, but unable to find binary in default location, no 'moz:firefoxOptions.binary' capability provided, and no binary flag set on the command line" exceptions.

## To Reproduce

1. Upgrade selenium-webdriver from 3.141.0 to 3.142.2
2. Try running your examples

## Expected behavior

Selenium finds firefox.

## Environment

FreeBSD 11.2-RELEASE-p10, more or less default install on multiple servers
Normal firefox install (pkg install firefox): firefox-66.0.5,1
Rails 5.2.3
Capybara 3.20.1

I don’t have firefox’ path set in my suite because I had never needed it to be set. I had no similar issue before 3.142.2.

3.142.2 works on MacOS Mojave
3.141.0 works on my FreeBSD CI servers (as well as on MacOS)
3.142.2 doesn’t work on FreeBSD.

The firefox binary (/usr/local/bin/firefox) is in the path.

p0deje:
Interesting. Assuming Firefox is in the page - it must be found by GeckoDriver.

1. What GeckoDriver version do you use?
2. Can you provide full stacktrace of error that you see?

PeterMozesMerl:
1. It’s the latest. 0.24.0

2. The 98th line of capybara_general_helper.rb is ```find css, options```

```
  28) User / Profile / Actions Adding and removing user to groups
      Got 0 failures and 2 other errors:

      28.1) Failure/Error: Capybara.current_session.current_window.resize_to 1500, CAPYBARA_HOST_IS_DEVELOPMENT_MACHINE ? 1200 : 1500

            Selenium::WebDriver::Error::SessionNotCreatedError:
              Expected browser binary location, but unable to find binary in default location, no 'moz:firefoxOptions.binary' capability provided, and no binary flag set on the command line
            # /usr/local/jenkins/.rvm/gems/ruby-2.6.3/gems/selenium-webdriver-3.142.2/lib/selenium/webdriver/remote/response.rb:72:in `assert_ok'
[...]

PR #7375, opened 2019-07-08
[rb] remove Firefox::Binary class

titusfortner:
As discussed on Slack, this is my proposed alternative to https://github.com/seleniumhq/selenium/commit/15f966d632dba814e8adf596752c29d43a0798bf

To recap:
1. Defaulting browser location belongs in Options not in Service class
2. geckodriver should be responsible for finding it by default, not a(n outdated) Selenium class; if geckodriver does not find it, then it should either be fixed by Mozilla or the user should specify it.
3. There is no reason for Firefox & Chrome to have different behaviors for this in 4.x


twalpole:
Seems logical to me 👍  - although you do have conflicts that need resolving.

p0deje:
It looks good to me, but my only concern is that Firefox::Binary is actually more reliable that GeckoDriver lookup mechanism. I don't think #7219 was fixed on GeckoDriver side so this change is going to be frustrating for users who got the issue resolved in 3.142.3, but then the same issue re-occurs in 4.0. If we can make at least FreeBSD lookup fixed in GeckoDriver before landing this change, that'd be great.

titusfortner:
Ok, I filed a bug with geckodriver: https://github.com/mozilla/geckodriver/issues/1585

Hopefully they can fix things before we release 4.0

PR #7376, opened 2019-07-08
[rb] remove unnecessary code from Firefox::Profile class

titusfortner:
Ways that Mozilla supports customizing these things are a little varied.

To point to an existing profile:
```
{"args": ["-profile", "/path/to/your/profile"]}
```
To only specify prefs:
```
{"prefs": {"foo": "bar"}}
```
To only specify extensions, we have an endpoint:
```
/session/{session id}/addon/install
```
(I requested that we be able to add extensions independently as part of options without needing an endpoint, but haven't seen progress: https://github.com/mozilla/geckodriver/issues/1476)

So the only way to set certificates at all, or extensions before initialization is with profile:
```
{"profile": Base64_ENCODED_ZIP}
```

`Firefox::Profile` should only need to generate a default profile and allow for
1. Custom certificates
2. Adding Extensions
3. Specifying preferences (actually, I'm not certain if `profile` will override `prefs` or if they merge?)

So... This PR.
1. Pulls out the extra prefs we were adding for legacy driver
2. Removes legacy driver preferences from class

I'm not sure the best way to go about testing this in an automated fashion. I think someone needs to sit down with extensions and preferences and certificates to make sure everything does what we expect for all the different methods outlined above, and I'm not currently volunteering for that. :)

### DIFF
--- a/rb/build.desc
+++ b/rb/build.desc
@@ -111,9 +111,6 @@ ruby_library(name = "firefox",
     "lib/selenium/webdriver/firefox/**/*.rb",
     "lib/selenium/webdriver/firefox.rb"
   ],
-  resources = [
-    { "//third_party/js/selenium:webdriver_prefs" : "rb/lib/selenium/webdriver/firefox/extension/prefs.json" }
-  ],
   deps = [":common"]
 )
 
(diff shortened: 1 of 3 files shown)


titusfortner:
Since I don't think the current code actually works the way it appears, I'm going to go ahead and merge this. I've created an issue to track the need to validate the correct behavior and that this code properly implements it. I'm creating a separate issue to track that part of it.

PR #7377, opened 2019-07-08
Add the possibility to nest By selectors.

danielthegray:
I have found it useful to build new By selectors by composing/nesting other `By` selectors. I have added a generic method to that for all selector types, as well as a faster way of embedding CSS selectors inside each other (by simply appending the selector strings).

### DIFF
--- a/java/client/src/org/openqa/selenium/By.java
+++ b/java/client/src/org/openqa/selenium/By.java
@@ -18,6 +18,7 @@
 package org.openqa.selenium;
 
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


java/client/src/org/openqa/selenium/By.java
```diff
@@ -138,6 +139,27 @@ public WebElement findElement(SearchContext context) {
...
+   *
+   * @param child The child selector to nest inside the current one (searching for
+   *              elements inside the result of this search).
+   * @return
+   */
+  public By nest(By child) {
+    final By parent = this;
+    return new By() {
+      public List<WebElement> findElements(SearchContext context) {
+        List<WebElement> results = new ArrayList<>();
```
danielthegray, line 152: Maybe a `LinkedList` would be better? I'm not sure.

jleyba: No, `LinkedList` is almost never worth using.

barancev:
Please add some tests too

diemol (commented):
@danielthegray are you interested in completing this PR? 
Please have a look at the comment above.

danielthegray:
I've had a look and I'm not sure my anonymous By object is the best way to add the feature in a "generic" way. The CSS selector nesting is nice, but the other ones might not work as nicely. I'm open to having this taken over by someone else or coming back to it when I have a bit more time.

barancev:
Let's close this PR for now because it's got stale. We have a class `ByChained` that partially solves the task of combining locators, but if there is a better solution we can consider it, so new PRs are welcome.

PR #7394, opened 2019-07-15
Correct contentEditable code in Clear atom

julianrkung:
The clear atom would previously leave a single whitespace when clearing
contenteditable elements due to a limitation in firefox. This CL
makes it so that the whitespace is only left for firefox browsers.

Additionally, contenteditable elements previously did not unfocus after
clearing which does not follow the W3C spec for [clearing elements](https://w3c.github.io/webdriver/#element-clear).
This changelist adds unfocusing code to run after a contenteditable
element is cleared.

Lastly, additional blur events would fire if the element to be cleared
had both a value and was contenteditable (eg a text input element with
a starting value and had its contenteditable attribute set to true).
This changelist corrects this by appending the contentEditable clause
to the end of the if-else chain.

### DIFF
--- a/javascript/atoms/action.js
+++ b/javascript/atoms/action.js
@@ -115,13 +115,18 @@ bot.action.clear = function(element) {
     // current value or not
     bot.action.LegacyDevice_.focusOnElement(element);
     element.value = '';
-  }
-
-  if (bot.dom.isContentEditable(element)) {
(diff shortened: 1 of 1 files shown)


barancev:
Merged as f05ebbfd156696d49598e216c2bf3d4d5858912c, thank you for the contribution!

PR #7406, opened 2019-07-18
[PY] Implement logical expected conditions in Python

### LINKED ISSUE #7121: Python ExpectedCondition "Or" "And" "Not"
## 🚀 Feature Proposal

Copy the ExpectedConditions "Or", "And", and "Not" from the Java library to the Python library.

## Motivation

- "**Or**" will simplify logic & speed up execution in tests when we need only 1 of several possible expected conditions to appear.
- "**And**" may simplify logic in tests where we need all of several conditions to appear.
- "**Not**" may increase code readability

## Example

e.g. "**Or**":  During A/B testing of a website, it is unknown which of 2 elements will be served every page load.  The "Or" expected condition can ensure the test is unblocked as soon as either of them displays.

e.g. "**Or**": A user may or may not be shown a splash screen before the login-screen.  Using "Or", a single test can begin execution as soon as it confirms that either a splash element or a login element is visible.

e.g. "**And**": A test may wish to confirm the visibility of multiple elements before continuing.  (_This functionality can be easily synthesized in test code, but using the "And" expected condition might create clearer code._)

e.g. "**Not**": A test may wish to wait for an element to disappear before continuing.  
(_This functionality can be easily synthesized in test code, AND it's often better to use the inverse ExpectedCondition, but using the "Not" expected condition might make clearer code._)

GQAssurance:
Replicates the functionality of Java's AND / OR / NOT expected conditions.
Because of reserved word constraints in Python, these are named:

OR:  `any_of(*expected_condition)`
AND: `all_of(*expected_condition)`
NOT: `none_of(*expected_condition)`

Each function takes an unlimited number of expected_conditions as arguments.

Implements #7121

### DIFF
--- a/py/selenium/webdriver/support/expected_conditions.py
+++ b/py/selenium/webdriver/support/expected_conditions.py
@@ -404,6 +404,58 @@ class alert_is_present(object):
             return False
 
 
+def any_of(*expected_conditions):
+    """ An expectation that any of multiple expected conditions is true.
+    Equivalent to a logical 'OR'.
+    Returns results of the first matching condition, or False if none do. """
+    def any_of_condition(driver):
+        for expected_condition in expected_conditions:
+            try:
+                result = expected_condition(driver)
+                if result:
+                    return result
+            except WebDriverException:
+                pass
+        return False
+    return any_of_condition
+
+
+def all_of(*expected_conditions):
+    """ An expectation that all of multiple expected conditions is true.
+    Equivalent to a logical 'AND'.
+    Returns: When any ExpectedCondition is not met: False
+             When all ExpectedConditions are met: A List with each
+              ExpectedCondition's return value """
+    def all_of_condition(driver):
+        results = []
+        for expected_condition in expected_conditions:
+            try:
+                result = expected_condition(driver)
+                if not result:
+                    return False
+                results.append(result)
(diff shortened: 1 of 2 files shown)


py/selenium/webdriver/support/expected_conditions.py
```diff
@@ -404,6 +404,54 @@ def __call__(self, driver):
...
+    """ An expectation that all of multiple expected conditions is true.
+    Equivalent to a logical 'AND'.
+    Returns a Boolean """
+    def all_of_condition(driver):
+        for expected_condition in expected_conditions:
+            try:
+                result = expected_condition(driver)
+                if not result:
+                    return False
+                return True
```
Dude-X: `return True` should be outside the for loop.
Many expected conditions sometimes return the element or object you're waiting for.
Should this return a list of all results?

GQAssurance: #1 (facepalm) fixed
#2 That's a good idea.  the `all_of` function returns every EC's result in a list; it could then be parsed for elemental results if needed.

Dude-X:
LGTM.

GQAssurance:
> LGTM.

Is there any way I can help get this merged into master?

PR #7414, opened 2019-07-20
Updating okhttp3 library to 4.0.1

mmerrell:
Updated okhttp3 to v4.0.1
Updated okio to v2.2.2
Added new dep: kotlin-stdlib v 1.3.41

I added the kotlin lib to the /third_party/java path, but if this keeps up we might consider a /third_party/kotlin path. I'm content to leave it like it is as long as it uses the jvm, but let me know if this should change now.


shs96c:
Looks great to me! Thank you for update!

PR #7416, opened 2019-07-21
Old style exceptions are syntax errors in Python 3

cclauss:


luke-hill:
@cclauss I know we still run some tests in python 2, will this syntax work in python 2?

@lmtierney should be a quick review and merge / decline here?

cclauss:
This change is compatible with both Python 2 and Python 3.

barancev:
The changed file is a part of the build infrastructure, it does not touch any code that can affect end users.

barancev:
Merged as ca9e9a4a752baa588f88eca87f2b936c0ed50709, thank you for the contribution!

PR #7419, opened 2019-07-22
Travis CI: dist: xenial is now the default and sudo is deprecated

cclauss:
The default Travis distro is now xenial and the __sudo:__ tags is now deprecated because the sudo is _always_ available and there is no way to turn it off.

### DIFF
--- a/scripts/travis/before_script.sh
+++ b/scripts/travis/before_script.sh
@@ -1,6 +1,6 @@
 set -ex
 
-sh /etc/init.d/xvfb start
+# sh /etc/init.d/xvfb start
 
 if [[ ! -z $TOXENV ]]; then	
   ./go py_prep_for_install_release	


scripts/travis/before_script.sh
```diff
@@ -1,6 +1,6 @@
 set -ex
 
-sh /etc/init.d/xvfb start
+# sh /etc/init.d/xvfb start
```
AutomatedTester: Do you mean to be commenting this out?

cclauss: Yes. __services: - xvfb__ makes in unnecessary. See:
https://docs.travis-ci.com/user/gui-and-headless-browsers/#using-xvfb-to-run-tests-that-require-a-gui

AutomatedTester: So why not just delete the line?

barancev:
Merged as 3bb5d31cd684b80529d118636e50dacace93098c
(I'll remove the commented line later)

PR #7430, opened 2019-07-24
[py] Remove 'resolve_ip' option from RemoteConnection

### LINKED ISSUE #1729: Python Selenium Client is by default resolving ip adresses and not sending domain names

### LINKED ISSUE #2509: python client bindings do not support remote https selenium grid endpoints.

GQAssurance:
Resolving the IP within Selenium is never necessary; This is always the wrong thing to do when dealing with HTTPS connections, as no certificates are issued for IP-addresses.

This is a conflict-free update of PR #3515.

Fixes #1729
Fixes #2509

### DIFF
--- a/py/selenium/webdriver/edge/webdriver.py
+++ b/py/selenium/webdriver/edge/webdriver.py
@@ -76,7 +76,6 @@ class WebDriver(RemoteWebDriver):
         RemoteWebDriver.__init__(
             self,
             command_executor=RemoteConnection(self.service.service_url,
-                                              resolve_ip=False,
                                               keep_alive=keep_alive),
             desired_capabilities=capabilities)
         self._is_remote = False
--- a/py/selenium/webdriver/remote/remote_connection.py
+++ b/py/selenium/webdriver/remote/remote_connection.py
@@ -27,8 +27,6 @@ try:
     from urllib import parse
 except ImportError:  # above is available in py3+, below is py2.7
     import urlparse as parse
-
-from selenium.webdriver.common import utils as common_utils
 from selenium import __version__
 from .command import Command
 from .errorhandler import ErrorCode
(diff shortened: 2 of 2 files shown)


lmtierney:
@AutomatedTester @davehunt this will need to have conflicts resolved due to a change from `/py` to `/python` but can we get this pulled in? Digging through the history, it seems we started resolving solely to support Edge. Whatever the original reasoning one seems to go out the window now with Edge being Chromium-based.

GQAssurance:
> @AutomatedTester @davehunt this will need to have conflicts resolved due to a change from `/py` to `/python` but can we get this pulled in? Digging through the history, it seems we started resolving solely to support Edge. Whatever the original reasoning one seems to go out the window now with Edge being Chromium-based.

The original pr #3515 appears like it was approved in principle, but never merged because the original author didn't resolve the conflicts.

Right now this PR doesn't have any conflicts with base, could you explain the `/py` vs. `/python` issue?

Based on my understanding, it appears Selenium is needlessly resolving IPs which I believe is a task better left left to the libraries.   Even though the case which exposed this (Edge browsers) is no longer a problem, it seems like this resolution code adds complexity to the Selenium project for no gain, and perhaps similar problems in the future.

lmtierney:
> Right now this PR doesn't have any conflicts with base, could you explain the `/py` vs. `/python` issue?

Disregard this. The `/py` directory was temporarily changed to `/python` but now it is back to `/py`.
 
> Based on my understanding, it appears Selenium is needlessly resolving IPs which I believe is a task better left left to the libraries. Even though the case which exposed this (Edge browsers) is no longer a problem, it seems like this resolution code adds complexity to the Selenium project for no gain, and perhaps similar problems in the future.

I'm all for the change, but wanted some input from others.

cgoldberg:
👍

GQAssurance:
> I'm all for the change, but wanted some input from others.

Glad to hear.  I also am eager to receive any kind of input, especially the critical kind :)

AutomatedTester:
Thanks for this change, it looks good to me

PR #7442, opened 2019-07-28
Selenium 4.0 grid browser priority, first cut

mmerrell:
Grid Browser Session priority, first cut at working logic with unit tests.

There's more code than necessary, and an abusive amount of commentary, but this should correctly prioritize where sessions go based on browser "rarity". Still need a code review and some unit tests, but I wanted to put this up for review and feedback. There's plenty to optimize, but I think this does the job.

### DIFF
--- a/java/server/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
+++ b/java/server/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
@@ -163,7 +163,8 @@ public class LocalDistributor extends Distributor {
    * Takes a Stream of Hosts, along with the Capabilities of the current request, and prioritizes the
    * request by removing Hosts that offer Capabilities that are more rare. e.g. if there are only a
    * couple Edge nodes, but a lot of Chrome nodes, the Edge nodes should be removed from
-   * consideration when Chrome is requested
+   * consideration when Chrome is requested. This does not currently take the amount of load on the
+   * server into consideration--it only checks for availability, not how much availability
    * @param hostStream Stream of hosts attached to the Distributor (assume it's filtered for only those that offer these Capabilities)
    * @param capabilities Passing in the whole Capabilities object will allow us to prioritize more than just browser
    * @return Stream of distinct Hosts with the more rare Capabilities removed
@@ -184,7 +185,6 @@ public class LocalDistributor extends Distributor {
     //First, check to see if all buckets are the same size. If they are, just send back the full list of hosts
     // (i.e. the hosts are all "balanced" with regard to browser priority)
     if (allBucketsSameSize(hostBuckets)) {
-      LOG.fine("All Hosts Prioritized prior to sorting");
       return hostBuckets.values().stream().distinct().flatMap(Set::stream);
     }
 
@@ -204,8 +204,6 @@ public class LocalDistributor extends Distributor {
     final List<Map.Entry<String, Set<Host>>> sorted = hostBuckets.entrySet().stream().sorted(
         Comparator.comparingInt(v -> v.getValue().size())
     ).collect(Collectors.toList());
-
-    LOG.info(sorted.toString());
 
     // Until the buckets are the same size, keep removing hosts that have more "rare" browser capabilities
     Map<String, Set<Host>> newHostBuckets;
@@ -232,8 +230,6 @@ public class LocalDistributor extends Distributor {
 
     return hostBuckets.values().stream().distinct().flatMap(Set::stream);
   }
-
-  //
 
   @VisibleForTesting
   Map<String, Set<Host>> sortHostsToBucketsByBrowser(Set<Host> hostSet) {
--- a/java/server/test/org/openqa/selenium/grid/distributor/DistributorTest.java
+++ b/java/server/test/org/openqa/selenium/grid/distributor/DistributorTest.java
@@ -572,7 +572,6 @@ public class DistributorTest {
            NewSessionPayload firefoxPayload = NewSessionPayload.create(firefoxCapabilities)) {
 
         Session chromeSession = distributor.newSession(createRequest(chromePayload)).getSession();
-        LOG.info(String.format("Chrome Session %d assigned to %s", i, chromeSession.getUri()));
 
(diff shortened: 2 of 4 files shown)


java/server/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
```diff
@@ -150,6 +159,104 @@ public CreateSessionResponse newSession(HttpRequest request)
...
+
+    //There might be unforeseen cases that challenge this assumption. TODO Create unit tests to prove it
+
+    //TODO a List of Map.Entry is silly. whatever this structure needs to be needs to be returned by
+    // the sortHostsToBucketsByBrowser method in a way that we don't have to sort it separately like this
+    final List<Map.Entry<String, Set<Host>>> sorted = hostBuckets.entrySet().stream().sorted(
+        Comparator.comparingInt(v -> v.getValue().size())
+    ).collect(Collectors.toList());
+
+    LOG.info(sorted.toString());
```
shs96c: You may want to drop the logging level in a follow-up PR.

java/server/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
```diff
@@ -150,6 +159,104 @@ public CreateSessionResponse newSession(HttpRequest request)
...
+      if (allBucketsSameSize(newHostBuckets)) {
+        LOG.fine("Hosts have been balanced according to browser priority");
+        return newHostBuckets.values().stream().distinct().flatMap(Set::stream);
+      }
+    }
+
+    return hostBuckets.values().stream().distinct().flatMap(Set::stream);
+  }
+
+  //
```
shs96c: ?

java/server/test/org/openqa/selenium/grid/distributor/DistributorTest.java
```diff
@@ -508,14 +515,88 @@ public void shouldReturnNodesThatWereDownToPoolOfNodesOnceTheyMarkTheirHealthChe
...
+    firefoxNodes.addAll(edgeNodes);
+    firefoxNodes.addAll(chromeNodes);
+
+    //Assign 5 Chrome and 5 Firefox sessions to the distributor, make sure they don't go to the Edge node
+    for (int i=0; i<5; i++) {
+      try (NewSessionPayload chromePayload = NewSessionPayload.create(chromeCapabilities);
+           NewSessionPayload firefoxPayload = NewSessionPayload.create(firefoxCapabilities)) {
+
+        Session chromeSession = distributor.newSession(createRequest(chromePayload)).getSession();
+        LOG.info(String.format("Chrome Session %d assigned to %s", i, chromeSession.getUri()));
```
shs96c: I guess this was for debugging?

java/server/test/org/openqa/selenium/grid/distributor/remote/RemoteDistributorTest.java
```diff
@@ -0,0 +1,5 @@
+package org.openqa.selenium.grid.distributor.remote;
+
+public class RemoteDistributorTest {
```
shs96c: Probably best not to have an empty test....

java/server/test/org/openqa/selenium/grid/session/remote/SessionTest.java
```diff
@@ -0,0 +1,5 @@
+package org.openqa.selenium.grid.session.remote;
+
+public class SessionTest {
```
shs96c: And another empty test. Please remove until there's content.

mmerrell:
Have another look... I think I got them all. Another thing I've been noodling over is about node capacity--I'm only checking browser priority with regard to whether hasCapacity == true or not--I'm not checking how much capacity a node has or doesn't have. We might want to refine that in with this change, so we can evaluate both things at once, rather than have another pass through the nodes for load at the end. It will still work now, but it might be better to have both.

java/server/src/org/openqa/selenium/grid/distributor/local/Host.java
```diff
@@ -55,7 +55,7 @@
 
 class Host {
 
-  private static final Logger LOG = Logger.getLogger("Selenium Distributor");
+  private static final Logger LOG = Logger.getLogger("Selenium Host");
```
shs96c: I kind of did this deliberately, so everything from the distributor appears in the same set of logs.

java/server/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
```diff
@@ -67,11 +69,12 @@
 import java.util.function.Supplier;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class LocalDistributor extends Distributor {
 
   private static final Json JSON = new Json();
-  private static final Logger LOG = Logger.getLogger("Selenium Distributor");
+  private static final Logger LOG = Logger.getLogger("Selenium Distributor (Local)");
```
shs96c: And here.

java/server/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
```diff
@@ -150,6 +159,100 @@ public CreateSessionResponse newSession(HttpRequest request)
     }
   }
 
+  /**
+   * Takes a Stream of Hosts, along with the Capabilities of the current request, and prioritizes the
+   * request by removing Hosts that offer Capabilities that are more rare. e.g. if there are only a
```
shs96c: I suspect that this will ultimately causes problems, since it implies that if a host offers both chrome and edge, it will _never_ run an edge test, no matter what. It would seem like a prioritised list is what we really want here. We can iterate later.

shs96c:
Woohoo!

java/server/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
mmerrell: Agreed. I'll think about it some more. The next iteration might be as simple as sorting those to the bottom of the list, rather than removing them, but there might be better methods still.

Thanks for landing my first code commit!

PR #7448, opened 2019-07-29
Fix Java test case testChordControlCutAndPaste

JohnChen0:
Replacing sendKeys(Keys.CONTROL + ...) with keyDown(Keys.CONTROL),
to ensure the CONTROL key remains pressed.

### DIFF
--- a/java/client/test/org/openqa/selenium/interactions/CombinedInputActionsTest.java
+++ b/java/client/test/org/openqa/selenium/interactions/CombinedInputActionsTest.java
@@ -279,7 +279,6 @@ public class CombinedInputActionsTest extends JUnit4TestBase {
   @Test
   @Ignore(value = MARIONETTE, issue = "https://github.com/mozilla/geckodriver/issues/646")
   @NotYetImplemented(EDGE)
-  @NotYetImplemented(CHROME)
   public void testChordControlCutAndPaste() {
     assumeFalse("FIXME: macs don't have CONTROL key", getEffectivePlatform().is(Platform.MAC));
     assumeFalse("Windows: native events library  does not support storing modifiers state yet",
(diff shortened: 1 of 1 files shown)


barancev:
I'm against this change, this test checks that sendKeys supports "chords".

We can create another test that uses keyDown. Also there should be keyUp in this test to resolve the key.

JohnChen0:
I should have given more background on why I made this change. With a W3C spec compliant WebDriver, the Java binding translates the original code `sendKeys(Keys.CONTROL + "a" + "x")` into the following Perform Actions command:

```
{
   "actions": [ {
      "actions": [ {
         "type": "keyDown",
         "value": "\\ue009"
      }, {
         "type": "keyUp",
         "value": "\\ue009"
      }, {
         "type": "keyDown",
         "value": "a"
      }, {
         "type": "keyUp",
         "value": "a"
      }, {
         "type": "keyDown",
         "value": "x"
      }, {
         "type": "keyUp",
         "value": "x"
      } ],
      "id": "default keyboard",
      "type": "key"
   } ]
}
```

Note that the Control key is pressed and then released before the letters "a" and "x" are pressed, so this doesn't do what the test intends, i.e., pressing Control-a and Control-x.

I can see two ways to fix this:

1. Modify the Java binding code to recognize Control key (and other similar keys) as special, and don't send keyUp event until some later time.
2. Modify the test to use keyDown to send the Control key. This is the approach currently used by the .NET code (https://github.com/SeleniumHQ/selenium/blob/master/dotnet/test/common/Interactions/CombinedInputActionsTest.cs#L268).

I am happy with either approach.

barancev:
I've discussed this issue with @shs96c recently, and we concluded that it's a bug in the Java binding, the test is correct.

shs96c:
Because the test is correct, but the bug exists in the bindings, I've opened #7458 and I'm closing this PR. Thank you for submitting it --- we've found a new bug that we should really address.

JohnChen0:
Thanks for looking into this. I agree that fixing key chords in Java bindings will be better for users.

PR #7460, opened 2019-08-02
Only conditionalize Travis on tree changes if the build is a PR

lmtierney:
After talking with @titusfortner, his intent for #6208 was for it to only apply when the Travis build is a PR. #7111 made the change, but made it for all builds and not just PRs. This adds logic to ensure all tests are run for non-PR builds.

`$TRAVIS_PULL_REQUEST` is set to the PR number if the build is a PR and `false` otherwise. This is (documented here)[https://docs.travis-ci.com/user/environment-variables#default-environment-variables]. We could also use `$TRAVIS_EVENT_TYPE`, but I feel the former is clearer.

### DIFF
--- a/scripts/travis/script.sh
+++ b/scripts/travis/script.sh
@@ -4,9 +4,9 @@ if [[ $TOXENV == *"remote"* ]]; then
   ./go selenium-server-standalone
 fi
 
-# Tests will be run only when each folder, such as /py, is modified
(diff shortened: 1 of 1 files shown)


shs96c:
As @titusfortner reviewed this, he can land it. I'm not sure why he didn't, but presumably there's a reason.

lmtierney:
I was just waiting to see if others wanted to look at it, but I'll pull it in now

PR #7464, opened 2019-08-04
node/selenium-webdriver/firefox.js: accept zip webextensions

glacambre:
Prior to this commit, the node.js firefox driver prevented loading
webextensions the name of which did not end in '.xpi'. This was not very
ergonomic as web-ext, the webextension-building tool created by Mozilla
, produces files the name of which ends in '.zip' by default.


glacambre:
@corevo thanks for pointing out where the test should be added :).

I believe everything's in order now, except the CI keeps failing. I think the reason it fails is that Travis tries to use an outdated cache. Could you clear my PR's cache and trigger a new build?

Issue #7390, opened 2019-07-15
Logs stopped working in ChromeDriver (c#)


alex-jitbit:
## 🐛 Bug Report

`driver.Manage().Logs.GetLog(LogType.Browser)` throws this exception now.

`System.NotImplementedException: unknown command: unknown command: session/f3cbd0a8ba58e1fe3e1e6d36dbdaf28e/se/logs`

OS: Windows 10
Browser: Chrome
Browser version: 75.0.3770.100
Browser Driver version: ChromeDriver 75.0.3770.140
Selenium.WebDriver: 4.0.0-alpha02

seky16:
As per [changelog](https://github.com/SeleniumHQ/selenium/blob/db9ce7fd52a6fea12be7121ff7cba1ab8d8d1771/dotnet/CHANGELOG#L19) I tried

> setting the UseSpecCompliantProtocol property of ChromeOptions to false

but it throws `OpenQA.Selenium.WebDriverException: invalid session id`.

Even tho `driver is ISupportsLogs` returns `true` (see [changelog](https://github.com/SeleniumHQ/selenium/blob/db9ce7fd52a6fea12be7121ff7cba1ab8d8d1771/dotnet/CHANGELOG#L24)) and `driver.Manage().Logs.AvailableLogTypes` returns both `browser` and `driver`, `driver.Manage().Logs.GetLog(logKind)` throws the mentioned `System.NotImplementedException` for both listed available types.

OS: Windows 10
Browser: Chrome
Browser version: 75.0.3770.142
Browser Driver version: ChromeDriver 75.0.3770.140
Selenium.WebDriver: 4.0.0-alpha02

lmtierney:
Please try this again with chrome 76

ashiitaka:
Got same issue 

System.NotImplementedException: 'unknown command: unknown command: session/35a408e6da14ce20d92c8ccb05e1d948/se/logs'

with 
WebDriver.Manage().Logs.GetLog(LogType.Browser);

OS: Windows 10
Browser: Chrome
Browser version:  76.0.3809.87
Browser Driver version: ChromeDriver 76.0.3809.68
Selenium.WebDriver: 4.0.0-alpha02

alex-jitbit:
@lmtierney me too, getting the same issue.

Chrome 76.0.3809.87
Chromedriver: 76.0.3809.68-beta

[...]

PR #7466, opened 2019-08-05
Issue 7390 - Changing GetCommand to PostCommand for GetLog

mmerrell:
I made this change, but I'm not expecting it to get merged yet--I need to learn how to build & test the dotnet bindings. Mainly I want to see if the build passes in lieu of being able to run locally.

### DIFF
--- a/dotnet/src/webdriver/Remote/RemoteWebDriver.cs
+++ b/dotnet/src/webdriver/Remote/RemoteWebDriver.cs
@@ -172,7 +172,7 @@ namespace OpenQA.Selenium.Remote
                 // Only add the legacy log commands if the driver supports
                 // retrieving the logs via the extension end points.
                 this.CommandExecutor.CommandInfoRepository.TryAddCommand(DriverCommand.GetAvailableLogTypes, new CommandInfo(CommandInfo.GetCommand, "/session/{sessionId}/se/log/types"));
-                this.CommandExecutor.CommandInfoRepository.TryAddCommand(DriverCommand.GetLog, new CommandInfo(CommandInfo.GetCommand, "/session/{sessionId}/se/logs"));
+                this.CommandExecutor.CommandInfoRepository.TryAddCommand(DriverCommand.GetLog, new CommandInfo(CommandInfo.PostCommand, "/session/{sessionId}/se/logs"));
             }
 
         }


dotnet/src/webdriver/Remote/RemoteWebDriver.cs
```diff
@@ -172,7 +172,7 @@ public RemoteWebDriver(ICommandExecutor commandExecutor, ICapabilities desiredCa
                 // Only add the legacy log commands if the driver supports
                 // retrieving the logs via the extension end points.
                 this.CommandExecutor.CommandInfoRepository.TryAddCommand(DriverCommand.GetAvailableLogTypes, new CommandInfo(CommandInfo.GetCommand, "/session/{sessionId}/se/log/types"));
-                this.CommandExecutor.CommandInfoRepository.TryAddCommand(DriverCommand.GetLog, new CommandInfo(CommandInfo.GetCommand, "/session/{sessionId}/se/logs"));
+                this.CommandExecutor.CommandInfoRepository.TryAddCommand(DriverCommand.GetLog, new CommandInfo(CommandInfo.PostCommand, "/session/{sessionId}/se/logs"));
```
szamacz: As per https://github.com/SeleniumHQ/selenium/issues/7390#issuecomment-518429009 also URL should be probably changed to one without trailing "s" ("/session/{sessionId}/se/log")

mmerrell:
This was obviated by @jimevans :)

giggio:
@mmerrell I noticed this was merged to `master`, but I think `master` will only end up on version 4, right? This should be backported to version 3 as well. I'm on `3.141.0` and the error occurs there too.

mmerrell:
This PR wasn't merged to master--@jimevans made his own change, which replaced this. I've already closed out this PR and deleted the branch. If you're seeing issues in the code at this point, I'd recommend submitting a PR with whatever needs to be done to fix it. 
MM

szamacz:
But @jimevans PR was merged do master as well and fix is only visible in version 4 as @giggio  mentioned. @mmerrell Can you make this change to be applied in version 3 as well, please?

diemol:
@szamacz as @mmerrell said, this PR was not merged, the fix was done through another commit.

In addition to that, fixes are not ported to already relased versions. So the option is to use the upcoming release of the next Selenium 4 alpha. Just check the releases in the upcoming days.

giggio:
If v4 was realeased I'd understand that, but it is still alpha, not even beta. What you are saying is that the users have to live with the bug on the stable version for maybe months, or use an unstable version. This is not cool. This has to be backported, as v3 also suffers from the problem.

cgoldberg:
> This has to be backported

Nothing is stopping you from backporting the fix and building it yourself.

PR #7482, opened 2019-08-08
Add sendDevToolsCommandAndGetReturn to Javascript Chrome webdriver.

pjaton:
Before this change, the chrome javascript webdriver only supported
sending command to Chrome's devtool without being able to get the
result returned by the command; which prevent the use of commands like
'Performance.getMetric'. This change add the missing command.

### DIFF
--- a/py/test/selenium/webdriver/common/window_tests.py
+++ b/py/test/selenium/webdriver/common/window_tests.py
@@ -79,8 +79,6 @@ def test_should_set_the_position_of_the_current_window(driver):
     assert new_position.get('y') == target_y
 
 
-@pytest.mark.xfail_firefox(raises=WebDriverException,
-                           reason='Get Window Rect command not implemented')
 @pytest.mark.xfail_safari(raises=WebDriverException,
                           reason='Get Window Rect command not implemented')
 def test_should_get_the_rect_of_the_current_window(driver):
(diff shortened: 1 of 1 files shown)


pjaton:
Note that I also had to change Jasmine's bootstrap path from `selenum/...` to `seleniumhq/...` to match the workspace. Without the change, the tests would fail.

javascript/node/selenium-webdriver/chrome.js
```diff
@@ -224,6 +225,10 @@ function configureExecutor(executor) {
       Command.STOP_CASTING,
       'POST',
       '/session/:sessionId/goog/cast/stop_casting');
+  executor.defineCommand(
+      Command.SEND_DEVTOOLS_COMMAND_AND_GET_RETURN,
+      'POST',
+      '/session/:sessionId/goog/cdp/execute');
```
jleyba: Is this a replacement for `/session/:sessionId/chromium/send_command` ?

Would be better to just update the endpoint used by `sendDevToolsCommand()` than to add a new method to the API

pjaton: Very good point. I decided to make this a separate command to stay consistent with Chromium code which offers both endpoint to:

- only send a command - [send_command](https://chromium.googlesource.com/chromium/src.git/+/178d8ad3a5ac163aedd278b1721c4f31f9923bba/chrome/test/chromedriver/server/http_handler.cc#562)
- send the command and return the result [send_command_and_get_result](https://chromium.googlesource.com/chromium/src.git/+/178d8ad3a5ac163aedd278b1721c4f31f9923bba/chrome/test/chromedriver/server/http_handler.cc#568)

I haven't look how these different endpoints are implemented being the hood, but felt it would be more flexible to add the new functionality instead of replacing the send only one. This is debatable, though.

PR #7496, opened 2019-08-15
Add expected conditions for invisibility of multiple elements

3j14:

 - `invisibility_of_any_elements_located`: Returns `True` once any of the located elements becomes invisible
 - `invisibility_of_all_elements_located`: Returns `True` once all of the located elements are invisible

### DIFF
--- a/py/selenium/webdriver/support/expected_conditions.py
+++ b/py/selenium/webdriver/support/expected_conditions.py
@@ -215,6 +215,26 @@ class visibility_of_all_elements_located(object):
             return False
 
 
+class invisibility_of_all_elements_located(object):
+    """ An expectation for checking that all located elements present on the DOM
+    of a page are invisible.
+    locator - used to find the elements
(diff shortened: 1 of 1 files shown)


GQAssurance:
My understanding is that the python maintainers are reluctant to expand python ECs beyond the ones already defined in Java.

#7406 is a PR which implements logical operators including `OR` and `AND`.  Once #7406 is merged, coders could implement your PR's functionality for any combination of ECs, and not just for invisibility.

kmcgon:
I think this could also be achieved using the existing `until_not()` method in WebDriverWait.
Something like:
`WebDriverWait(driver).until_not(expected_conditions. visibility_of_all_elements_located(my_locator)`

jnsdrtlf:
@kmcgon your suggested code would wait until one of the elements becomes invisible (`invisibility_of_any_elements_located`).

Sure, this would cover one of my ECs, but this pull request is discarded anyway ;)

PR #7473, opened 2019-08-07
Fixing Javadoc for Actions.moveToElement, "center", not "top-left"

mmerrell:
Changing references to conform to W3C spec.

- [X] By placing an `X` in the preceding checkbox, I verify that I have signed the [Contributor License Agreement](https://github.com/SeleniumHQ/selenium/blob/master/CONTRIBUTING.md#step-6-sign-the-cla)

PR #7504, opened 2019-08-19
Fix mouse move offsets in Java tests

JohnChen0:
With PR #7473 clarified that Action.moveToElement uses offset relative
to center instead of top-left, updating Java tests to match the change.


barancev:
Merged by commit 288a9c3e057b808081f6209fb535e580c6451cb0
Thank you for the contribution!

Issue #7509, opened 2019-08-21
WebDriverProxy EPROTO wrong version number


theredfish:
## 🐛 Bug Report

If the selenium server uses the https protocol and the webDriverProxy uses the http protocol, an EPROTO error is thrown. 

```bash
➜  selenium-webdriver git:(master) ✗ node script.js
(node:863) UnhandledPromiseRejectionWarning: Error: EPROTO write EPROTO 140076306990912:error:1408F10B:SSL routines:ssl3_get_record:wrong version number:../deps/openssl/openssl/ssl/record/ssl3_record.c:332:

    at ClientRequest.<anonymous> (/selenium-webdriver/node_modules/selenium-webdriver/http/index.js:258:15)
    at ClientRequest.emit (events.js:198:13)
    at TLSSocket.socketErrorListener (_http_client.js:392:9)
    at TLSSocket.emit (events.js:198:13)
    at errorOrDestroy (internal/streams/destroy.js:107:12)
    at onwriteError (_stream_writable.js:436:5)
    at onwrite (_stream_writable.js:461:5)
    at _destroy (internal/streams/destroy.js:49:7)
    at TLSSocket.Socket._destroy (net.js:613:3)
    at TLSSocket.destroy (internal/streams/destroy.js:37:8)
(node:863) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:863) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:863) UnhandledPromiseRejectionWarning: Error: EPROTO write EPROTO 140076306990912:error:1408F10B:SSL routines:ssl3_get_record:wrong version number:../deps/openssl/openssl/ssl/record/ssl3_record.c:332:

    at ClientRequest.<anonymous> (/selenium-webdriver/node_modules/selenium-webdriver/http/index.js:258:15)
    at ClientRequest.emit (events.js:198:13)
    at TLSSocket.socketErrorListener (_http_client.js:392:9)
    at TLSSocket.emit (events.js:198:13)
    at errorOrDestroy (internal/streams/destroy.js:107:12)
    at onwriteError (_stream_writable.js:436:5)
    at onwrite (_stream_writable.js:461:5)
[...]

PR #7511, opened 2019-08-21
Use the proxy protocol when a webdriver proxy is configured

theredfish:
When the http protocol is different between the remote selenium
server and the webdriver proxy, an EPROTO error is encountered.

See #7509 for more information and a reproductible test case.

Could you help me to test this case correctly? It seems that changing the protocol from http to https in httpserver.js break the tests in `http_test.js`. Should i create a https version of `Server` and use it in a test?


shs96c:
Thank you for taking the time to contribute back! It's very much appreciated! Welcome to the project!

theredfish:
You're welcome! We use Selenium everyday at Oui.sncf so it's with pleasure :)

Thank you for the merge!

PR #7530, opened 2019-08-30
Handle TaskCanceled exception and re-throw it with timeout exception

nvborisenko:
`HttpClient` doesn't throw Timeout WebException, it throws `TaskCanceledException`.

This PR consider it and re-throw new exception in case of `TaskCanceled`.

image


jimevans:
Thank you for the contribution. This has been on my to-do list for a bit, but I've not gotten around to it. However, I'd like to retain the original `if` clause in the exception handler for the `HttpRequestException` case. I strongly suspect we won't be receiving that exception often, but having the exception handler there in its current form does not hurt anything.

nvborisenko:
I returned back previous `if` statement, but please be aware this if statement will never be executed. This is behavior of suggested `HttpClient`.

jimevans:
Merged in 052e70f. Thank you for the contribution!

PR #7562, opened 2019-09-13
Refactor/rework ruby raketasks part1

luke-hill:

Rework the rake-tasks folder by doing manual changes

Rework the `crazy-fun/build-tasks/build_grammar.rb` by just fixing the tabs, and removing the `then` keyword after case statements. NB: There is a ginormous case statement around 300 lines in the middle, so I added a comment at the end. God help whoever has to refactor this!

Review: @p0deje / @shs96c (Incase I've deleted / amended something as a breaking change)

I'd advise to review in 2 stages - One just for the crazy-fun one as it'll be awful to look at!

### DIFF
--- a/rake-tasks/c.rb
+++ b/rake-tasks/c.rb
@@ -2,7 +2,19 @@
 
 require 'rake-tasks/files'
 
-attr_accessor :bitness
+# Because this is created inside a main construct, accessors don't work shorthand
+# So we'll define them long-hand and then revert to shorthand once we classify the
+# files properly - Luke - Sep 2019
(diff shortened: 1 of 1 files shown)


p0deje:
The changes seem fine, but they don't work: https://travis-ci.com/SeleniumHQ/selenium/jobs/234810534#L513

luke-hill:
Crap. Oki, give me a bit of time to dig in. I think I know where the issue is, but no idea why it's complaining. I may just revert that portion for ease of fix.

p0deje:
Thank you, Luke!

PR #7599, opened 2019-09-25
Create WebElementWait

vethman:
Preperation for ExpectedConditions. Next step is to modify expectedConditions, ISearchContext will be used instead of the IWebDriver so that we can also wait for an element condition inside an element. This comes in handy when using the pageObjects pattern.

### DIFF
--- a/dotnet/src/webdriver/Support/WebElementWait
+++ b/dotnet/src/webdriver/Support/WebElementWait
@@ -1,0 +1,64 @@
+// <copyright file="WebElementWait.cs" company="WebDriver Committers">
+// Licensed to the Software Freedom Conservancy (SFC) under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The SFC licenses this file
+// to you under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


luke-hill:
Expected Conditions was deliberately removed from the dotNET codebase and placed inside a separate package. I'll try get more info for you, but that may be a better place to lodge PR's against.

vethman:
Shouldn't the defaultwait class be obsolete either and moved to the DotNetSeleniumExtras package?

jimevans:
@vethman No, it shouldn’t.

Oh, I suppose a more thorough explanation would be helpful. The items that were moved to DotNetSeleniumExtras (`ExpectedConditions` and `PageFactory`) were done so for specific reasons, as explained in the blog post linked to by @luke-hill. Just because those classes were deprecated and migrated does not imply that everything in the support namespace should likewise do so. `WebDriverWait` and `DefaultWait<T>` are quite useful, uncontroversial, and relatively stable. They are either sufficient for most users’ purposes as-is, or can easily be used as building blocks for a user to create something more custom. The same cannot be said of the migrated classes.

diemol:
Seems this PR went stale, I'll go ahead and close it.
If there is work that still needs to be done, please feel free to comment or to reopen.

PR #7603, opened 2019-09-26
Fix testShouldNotScrollIfAlreadyScrolledAndElementIsInView

JohnChen0:
testShouldNotScrollIfAlreadyScrolledAndElementIsInView does not work
correctly. While clikcing button1, a spec-compliant implementation
should scroll the page just enough to bring button1 into view, while
button2 remains out of view, thus clicking button2 whould cause
scroll again. Reversing the order of the clicks to fix this.

### DIFF
--- a/dotnet/test/common/ClickScrollingTest.cs
+++ b/dotnet/test/common/ClickScrollingTest.cs
@@ -91,14 +91,12 @@ namespace OpenQA.Selenium
 
 
         [Test]
-        [IgnoreBrowser(Browser.Chrome, "Webkit-based browsers apparently scroll anyway.")]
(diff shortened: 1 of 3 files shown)


barancev:
This change is not compatible with IE driver, it scrolls to the top by default. It is not compatible with legacy FirefoxDriver too, but it's insufficient already.

@jimevans What do you think about this change? Should we just ignore this test in IE? Or keep the old version of this test for IE? Or change IE driver to match the others?

barancev:
I've merged this PR as 72751e87fd773cc68fc5b1866c660556bf4ba917 but we need a followup.

PR #7611, opened 2019-09-27
Refactor/remove unused files

luke-hill:

Whilst a bit smaller, this is the first PR that actually starts removing / destroying previously referenced components.

- A lot of the re-linked files which were moved have now been removed and their new references 
updated
- Primarily this PR removes `require` calls, so this PR may fail with unfound files
- The `Platform` helper which was mixed into a global(ish), ns has been merged with the `SeleniumRake::Checks` class methods to remove duplication


p0deje:
Looks good to me. Do you want to merge this as-is or push more removal in the same PR?

luke-hill:
I'm always happy to merge stuff in as is. I'm not gonna be looking at this for a few days. So if I can get all my tidy up branches in, it means I have a clean slate when I come back to it.

PR #7615, opened 2019-09-28
Consistent Python WebDriver classes

OutOfFocus4:

This PR changes the Microsoft Edge `WebDriver` `__init__` to take a `desired_capabilities` parameter instead of `capabilities`. This makes it consistent with other `WebDriver` classes.

This PR also adds a `create_options` method to the Edge and Firefox `WebDriver` classes to make them more consistent with the other driver classes whose initializers take an `options` parameter.


AutomatedTester:
Thanks for this patch. It would appear that I missed this file when I was doing the original cleanup and deprecations. This change doesn't fit in with that but if you wanted to clean it up to match chrome and firefox. It will only be deprecation warnings for now but we can leave it there for a few releases before we remove it all.

PR #7625, opened 2019-10-01
Refactor/sole master directory

luke-hill:

- Make the sole arbiter `rake_tasks`
- Fix up some minor punctuation

EDIT: I've opted to tidy up the `.rake` files because they were a proper mish-mash. And future PR's will attempt to try unclog these files even further.

It's likely that even though we're moving all the rakes into the one mega `Rakefile` we can unclog it later, but this is a necessary evil for now to fix the `rake_tasks` directory


barancev:
Can you please rebase it, I've changed rakelib/bazel.rake to fix a Windows-specific issue

luke-hill:
This is rebased and I've done an initial "guess" fix (Maybe the failure was caused by it being massively outdated :man_shrugging: )

My new fixes are to add a temporary second require for all the rake files. However I could not find where they were initially loaded in to memory.

Also calling them `.rake` instead of `.rb` makes them a bit harder to triage. However that might be a bigger change I do later on.

For now we've got the single proper source of truth directory. The next job is to start cleansing more code and namespacing things more.

Next stop after this will be some of the crazy fun code, which is in mega-files. So I'll probably just try breaking those out.

luke-hill:
Ping
Ruby Review: @p0deje 
JVM Issue / Approval: @barancev

barancev:
@luke-hill What JVM issue?

luke-hill:
@p0deje this now begins to start unclogging stuff and putting them in more folder based structures.

I've fixed the conflicts, and the JVM issue from previously was fixed in another branch. So this should pass.

The GCC area has a lot of tabbed fixes, so apologies for that.

barancev:
LGTM, merged as fd1fabe5746b60a7c62aa63ba1f3e2838d368f01

PR #7650, opened 2019-10-04
Adding customizable browser_name capability to microsoftedge

stanleyhon:
msedgedriver allows driving of edge based targets such as [WebViews](https://docs.microsoft.com/en-us/microsoft-edge/hosting/webview2). msedgedriver expects the browserName capability to be provided differently in order to differentiate between driving edge or an edge-based webview.

This change updates the edge variant of `Options` to allow specifying the browserName capability, This overrides what is requested to msedgedriver.

### DIFF
--- a/py/selenium/webdriver/edge/options.py
+++ b/py/selenium/webdriver/edge/options.py
@@ -24,6 +24,7 @@ class Options(ChromiumOptions):
     def __init__(self, is_legacy=True):
         super(Options, self).__init__()
         self._is_legacy = is_legacy
+        self._custom_browser_name = None
 
         if is_legacy:
             self._page_load_strategy = "normal"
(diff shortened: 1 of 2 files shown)


AutomatedTester:
Are you able to write tests for this change?

stanleyhon:
Would adding something similar to the existing tests in https://github.com/SeleniumHQ/selenium/blob/master/py/test/unit/selenium/webdriver/edge/edge_options_tests.py be enough?

I would verify specifying `options.custom_browser_name = "someTestvalue"` results in the value being outputted when option is asked to produce a capability set with `toCapability`.

stanleyhon:
@AutomatedTester just following up to see if my suggested resolution would be sufficient before I go implement the wrong thing :)

AutomatedTester:
> Would adding something similar to the existing tests in https://github.com/SeleniumHQ/selenium/blob/master/py/test/unit/selenium/webdriver/edge/edge_options_tests.py be enough?
> 
> I would verify specifying `options.custom_browser_name = "someTestvalue"` results in the value being outputted when option is asked to produce a capability set with `to_capabilities`.

That is a good place to put it

stanleyhon:
@AutomatedTester I've added a test as we discussed and the test passes on my local box. Let me know if you notice anything else.

PR #7653, opened 2019-10-05
Make Python Edge driver more like other Chromium drivers

OutOfFocus4:

Based on @AutomatedTester's comments on #7615, add deprecation warnings to the `__init__` of the Edge `WebDriver` and implement `create_options`.


AutomatedTester:
Sorry for only looking at this now. 

You've changed the signature of the `__init__` which is a breaking change for anyone using named  arguments, we can't do that unfortunately. We need to add the other and deprecate the other one for a few releases.

Also, before we can carry on we need you to sign the CLA.

PR #7659, opened 2019-10-06
Fix common typo's

xaircore:

### DIFF
--- a/py/selenium/webdriver/firefox/extension_connection.py
+++ b/py/selenium/webdriver/firefox/extension_connection.py
@@ -72,7 +72,7 @@ class ExtensionConnection(RemoteConnection):
 
     @classmethod
     def is_connectable(self):
-        """Try's to connect to the extension but do not retrieve context."""
+        """Tries to connect to the extension but do not retrieve context."""
         utils.is_connectable(self.profile.port)
 
 


py/selenium/webdriver/firefox/extension_connection.py
```diff
@@ -72,7 +72,7 @@ def connect_and_quit(self):
 
     @classmethod
     def is_connectable(self):
-        """Trys to connect to the extension but do not retrieve context."""
+        """Try's to connect to the extension but do not retrieve context."""
```
ddavison:
```suggestion
        """Tries to connect to the extension but do not retrieve context."""
```

ddavison:
LGTM 👍

xaircore:
@ddavison Thanks for suggested changes

barancev:
Merged as e89b7d4432e10a304280cb447e84c896e129f4cd, thank you for the contribution!

py/selenium/webdriver/common/actions/input_device.py
```diff
@@ -39,5 +39,5 @@ def add_action(self, action):
     def clear_actions(self):
         self.actions = []
 
-    def create_pause(self, duraton=0):
+    def create_pause(self, duration=0):
```
cgoldberg: we should probably deprecate (and display a warning) the misspelled keyword arg here rather than removing it... and then add the new one.   This changes a public API and will break any code that is currently using the misspelled version.

barancev:
@cgoldberg I don't think anybody uses this `create_pause` method because it does nothing :)

PR #7667, opened 2019-10-09
Refactor/crazy fun namespace 1

luke-hill:

Part 1 of 3 of tidy up.

Move about 40-50% of code into correct folder structure and 1class per file.
Use `mappings_` files as a linker
Retain previously named files just as symlinks so as not to break backwards compatibility

Part 2 will do the same to the rest of the code, part 3 will remove global mixins


luke-hill:
It'll get worse before better, but once it's fully structured it'll be so much easier to refactor and fix up and make pretty

PR #7669, opened 2019-10-10
Fix getLocationInView atom for documentElement

JohnChen0:
Fix Chrome's getLocationInView atom

ChromeDriver uses webdriver.chrome.getLocationInView atom to get
location of an element, after scrolling it into view if necessary. This
atom had a couple of bugs:

1. When this atom is passed the <html> element (i.e.,
   document.documentElement), it throws an error since the parent node
   is not an element. This causes ChromeDriver bug
   https://crbug.com/chromedriver/1049.

2. Scrolling isn't working properly.

This commit fixes both bugs, and expands the corresponding test.

### DIFF
--- a/javascript/chrome-driver/atoms.js
+++ b/javascript/chrome-driver/atoms.js
@@ -142,7 +142,12 @@ webdriver.chrome.scrollIntoView_ = function(elem, region, center) {
 
   offset = goog.style.getClientPosition(elem);
   var windowSize = goog.dom.getDomHelper(elem).getViewportSize();
-  scrollHelper(doc.body, windowSize, offset, region, center);
+  // Chrome uses either doc.documentElement or doc.body, depending on
+  // compatibility settings. For reliability, call scrollHelper on both.
+  // Calling scrollHelper on the wrong object is harmless.
+  scrollHelper(doc.documentElement, windowSize, offset, region, center);
+  if (doc.body)
+    scrollHelper(doc.body, windowSize, offset, region, center);
 };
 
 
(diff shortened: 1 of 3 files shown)


barancev:
I kindly ask you to add a test case too.

barancev:
Oops, just figured out that
1) We don't run tests located in /javascript/chrome-driver on CI server
2) The tests `testGetLocationInView` is red...
:(

barancev:
`testGetLocationInView` has a misleading name, actually it's a test for an element that is hidden in a frame and inside a frame it's hidden in an overflow fox, and the test fails because of wrong scrolling.

So this tests should be renamed and we need more tests that cover this atom in less sophisticated cases.

JohnChen0:
Thanks for the feedback. I'm trying to run the test, but couldn't figure out how to get it started. Loading `javascript/chrome-driver/test/atoms_test.html` into a browser results in large number of errors. Do I need to run some command to set up the test?

barancev:
To run JS tests in developer mode you have to start a debug server
```
bazel run javascript/atoms:test_debug_server
```
and open the page http://localhost:2310/javascript/chrome-driver/test/atoms_test.html

TriciaCrichton:
John asked me to follow up on this, but I haven't found the right command to compile the chrome-driver atoms. I've run bazel build java/... javascript/..., but http://localhost:2310/javascript/chrome-driver/test/atoms_test.html is not found. The only folder in http://localhost:2310/javascript/ is atoms.

barancev:
Sorry, we missed this part of JS source tree. I've fixed it in master, pleare rebase and run
```
bazel run javascript/chrome-driver:test_debug_server
```
and then open http://localhost:2310/filez/selenium/javascript/chrome-driver/test/

JohnChen0:
Thanks for fixing the test framework. I was able to find the cause of the test failure. The scrolling code used on Chrome needs some update. We have made the update inside ChromeDriver, but forgot to push the update to Selenium.

I have updated this PR to include the fixes necessary to make the test green, and also expanded tests for getLocationInView.

barancev:
Thank you! I've pushed these commits to master. If there be outstanding issues not related to documentElement I'll raise a new issue.

PR #7671, opened 2019-10-10
Refactor/rubocop autofixes on selenium rake

luke-hill:

A variety of manual / automatic rubocop fixes on the `selenium_rake` directory

Autofixes
- Add some frozen_string_literal's
- Fix up some string and punctuation quick fixes

Manual
- Some LineLength tweaks
- Some lVar assigns have been made private for better legibility
- Very initial phase of deleting code that doesn't appear to be used


luke-hill:
Reviews:
@shs96c / @jimevans - I have removed `c_checks.rb` which I've been told by barancev on slack that used to be involved with building the Native Firefox driver from Se2 times. I had a build @ https://github.com/SeleniumHQ/selenium/pull/7671/commits/0c2cbda309e441bcb3072908057e2c3efa6bb3d7 which was fully green (So it should be ok to delete this). But can you both just confirm this is ok to delete (There's a lot more also to tidy up here, but small steps)

@p0deje - Ruby review. Just some rubocop re-styles and autofixes here.

Issue #7631, opened 2019-10-01
isDisplayed throw Null Pointer - Chrome 77


Naumansh:
## 🐛 Bug Report

isDisplayed method throws Null pointer exception sometimes which originates from RemoteWebDriver class. This issue specific to Chrome version 77.

There is no issue on Chrome version 76 and older.

#Reproduce:
Open any web page.
Call isDisplayed.

#Error Log:
java.lang.NullPointerException
	at org.openqa.selenium.remote.RemoteWebElement.isDisplayed(RemoteWebElement.java:329)
	at eyethink.automation.bots.actionsfactory.type.Is$Web.isDisplayed(Is.java:122)

#Environment:
OS: Windows 10/OSX
Browser: Chrome
Browser version: 77.0.3865.90 
Browser Driver version: ChromeDriver 77 (mac)
Language Bindings version: Java 3.141.59

Szesan:
Same issue here: https://github.com/SeleniumHQ/selenium/issues/7619
and here: https://github.com/SeleniumHQ/selenium/issues/7641

cgoldberg:
@Naumansh 
you are using ChromeDriver 2.43, which only supports Chrome versions 69-71.  Since you are using Chrome 77, please upgrade your Chromedriver to the matching version (v77) and try again.

Szesan:
> @Naumansh
> you are using ChromeDriver 2.43, which only supports Chrome versions 69-71. Since you are using Chrome 77, please upgrade your Chromedriver to the matching version (v77) and try again.
I get null reference exceptions on displayed property when I use matching driver for chrome, ie
Chrome 77 with ChromeDriver 77: https://github.com/SeleniumHQ/selenium/issues/7641

Naumansh:
@cgoldberg Yep using the latest one. That's my bad forgot to add new chromeDriver version. The issue is with lattest versions of both chrome browser and driver.

Szesan:
Yep, same thing here. Latest ChromeDriver, latest Chrome see the result: https://github.com/SeleniumHQ/selenium/issues/7641

oyablonskyi:
We faced with the same issue on 77 version of Chrome browser and driver

barancev:
It's a chromedriver issue, please follow:
https://bugs.chromium.org/p/chromedriver/issues/detail?id=3123

I'll keep this issue open to track the status.

oyablonskyi:
[...]

PR #7673, opened 2019-10-10
7672 fix NPE in RemoteWebElement

asolntsev:
sometimes webdriver can return null in methods `isSelected`, `isDisplayed`, `isEnabled`. We need to handle them properly instead of throwing NullPointerException.

### DIFF
--- a/java/client/src/org/openqa/selenium/remote/RemoteWebElement.java
+++ b/java/client/src/org/openqa/selenium/remote/RemoteWebElement.java
@@ -148,10 +148,9 @@ public class RemoteWebElement implements WebElement, FindsByLinkText, FindsById,
 
   @Override
   public boolean isSelected() {
-    Object value = execute(DriverCommand.IS_ELEMENT_SELECTED(id))
-        .getValue();
+    Object value = execute(DriverCommand.IS_ELEMENT_SELECTED(id)).getValue();
     try {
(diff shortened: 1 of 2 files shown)


barancev:
Chrome again? I wish they have fixed this issue in the driver.

Of course we can (and do) apply this fix, but it's a dirty hack that hides the actual issue.

The most dangerous thing here is `isDisplayed` that is essentially a call to `executeScript`, that means `executeScript` can return an unexpected `null` value as a result of an arbitrary call, that is very unfortunate...

asolntsev:
@barancev I agree, this is a hack. :)

barancev:
Also I hope it's not an attempt to fix https://github.com/SeleniumHQ/selenium/issues/7631, so I want you to confirm that the issue is not related to Chrome 77 and that you have seen these NPE in other versions of Chrome too.

asolntsev:
@barancev Sorry, but I saw it only in Chrome 77 :)

barancev:
Oh... In this case I'll not merge this patch. It's technically useless because Chrome will ship the driver update faster that we ship the next version of Selenium llibrary. Suppressing driver errors and pretending that driver returns something meaningful is a suboptimal strategy.

asolntsev:
Clear. Thank you for clarification.

PR #7682, opened 2019-10-14
Remove uncalled code (base generator amongst others)

luke-hill:
Remove / tag / move code in files.rb - Most of this can be deleted or moved once another PR is merged in

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -37,7 +37,6 @@ require 'rake_tasks/selenium_rake/type_definitions_generator'
 # These are the final items mixed into the global NS
 require 'rake_tasks/bazel'
 require 'rake_tasks/copyright'
-require 'rake_tasks/files'
 require 'rake_tasks/python'
 
 # Our modifications to the Rake library


luke-hill:
I'll add 1/2 more commits to this @p0deje once the other PR's have been merged in

luke-hill:
The 2nd portion of removals was concerned with a piece of code that called `Platform.file_for`

However this has been removed and is no longer in use. So each of these instances could not be being called. And after commenting them out, they didn't crash the system. So I managed to delete a couple of hundred LOC. :boom: 

If I can get some eyes from one of @barancev / @shs96c then this is RtM

p0deje:
As far as I can see Python tasks are now failing on Travis. Can you take a look?

luke-hill:
OOOOoooooh dear. Oki, It was to be expected removing more and more code eventually I'd ping something.

I'm back now from a short break. Will take a look in the next day or so.

luke-hill:
So this was oddly caused by the Rakefile requiring the same file twice (Maybe I did this, maybe not), but either way removing the offending line now fixes the things up (So wasn't a big fix)

This is now ready for review / merge.

Primarily this removes code from 2 places.
- Places in which code was calling a broken method (This has been broken by me, through refactoring, but shows they should have been removed alongside them, they were originally deeply nested inside JS crazy_fun objects.
- Places which called `Export::` cf tasks. These had no caller. And as such a large chunk of code has been :boom: from here.

There is plenty more areas like this which have large segments, and I've seen patterns, so future PR's will remove much larger chunks :wine_glass:

luke-hill:
ping @barancev / @p0deje

PR #7704, opened 2019-10-22
WIP: Adding sonar sources for java,rb,py

ddavison:


luke-hill:
Hi @ddavison 

I "think" this PR may be breaking the atoms composing. But I'm not 100%.

I've added a PR which builds off the latest master just with some extra logging. And (It's still building now). But Job #1 passes which fails here.

I've pinged barancev in slack. But I'll reference you there too if you need to chat. Unfortunately i'm on CET, so it's likely we may miss each other during a regular day.

I've done a bit of work refactoring that area recently, but the code in question was just moved, not refactored in any way. So I'm saying if there is an issue, it's likely a latent one (I can still try help fix it though)

luke-hill:
No idea if this will help, but this is 11 commits behind master? Maybe that would fix something :man_shrugging:

PR #7708, opened 2019-10-23
CrazyFun/Build - Massive cull of uncalled code.

luke-hill:

Changes:
- Rakefile requires are now re-ordered alphabetically, and have been sorted better according to their type/s
- Requires which are used in more than one places have been de-duped and placed in the relevant place

Additions:
- `FileCopyHack` is now namespaced better (Still need the whole of CF to be namespaced properly)
- `RubyMappings` has been split into individual files for each of the individual mappings that are present

Removals:
- All of the mappings which weren't consumed inside crazyfun. Of which there were 4 big ones (Python / GCC / Rename / Javascript) - Leading to over 1500LOC being culled.

Refactors:
- `ruby` as a method was included in the top level namespace, I've now moved this to only be defined inside the RubyMappings classes that needed it, so it's giving us a bit cleaner code.

Next up. Look into finally fixing all the namespaces inside crazy_fun (Excluding the `main.rb` file)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -541,17 +541,23 @@ namespace :jruby do
 end
 
 namespace :node do
-  task atoms: %w(
+  atom_list = %w(
     //javascript/atoms/fragments:is-displayed
     //javascript/webdriver/atoms:get-attribute
-  ) do
+  )
+
+  task atoms: atom_list do
     baseDir = 'javascript/node/selenium-webdriver/lib/atoms'
     mkdir_p baseDir
 
-    [
+    puts "rake outs are below"
+    p rake_outs = [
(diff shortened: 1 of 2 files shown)


luke-hill:
Don't merge this yet. Got more commits to add (Good to see I've not broken anything)

luke-hill:
I'm going to take a break here and leave this PR as is, as the changes are quite big. I'll update the OP with the full details. But some big wins here!

luke-hill:
cc/ @p0deje / @barancev for review / merge.

luke-hill:
Added the debug logic into this PR so it's just 1 review and merge. This will give us a bit more detailed info when JS builds fail, around the `Rake::Task.out` issues.

Issue #7393, opened 2019-07-15
When standalone server converts OSS click to W3C actions, button info is lost


JohnChen0:
## 🐛 Bug Report

When an OSS-protocol client app uses Selenium standalone server to connect to a W3C-protocol WebDriver implementation (e.g., ChromeDriver 75 and above), the standalone server converts OSS protocol requests to W3C format. When some mouse related commands (e.g., [`/session/:sessionId/click`](https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessionidclick)) are converted, the specified `button` property is ignored, and `button: 0` is sent to the WebDriver.

## To Reproduce

Run the following shell commands from a Linux command prompt:

Start Selenium standalone server, configure ChromeDriver location, and enable ChromeDriver logging:

    java -Dwebdriver.chrome.driver=/usr/local/bin/chromedriver -Dwebdriver.chrome.logfile=chromedriver.log -jar selenium-server-standalone-3.141.59.jar &

Send OSS-format request to start a new session, and extract the session ID.

    SESSION_ID=`curl -s -XPOST http://localhost:4444/wd/hub/session -d '{"desiredCapabilities":{"browserName":"chrome"}}' | jq -r .sessionId`

Try to click the right mouse button:

    curl -XPOST http://localhost:4444/wd/hub/session/$SESSION_ID/click -d '{"button":2}'

Open the generated `chromedriver.log` file, and observe that the above click command has been converted into the following Perform Actions command, with `button` property of 0 instead of 2:

    "actions": [ {
       "actions": [ {
          "button": 0,
          "type": "pointerDown"
       }, {
          "button": 0,
          "type": "pointerUp"
       } ],
       "id": "mouse",
       "parameters": {
          "pointerType": "mouse"
       },
       "type": "pointer"
    } ]

## Expected behavior

The converted Perform Actions command should have `"button": 2`.

This issue was originally reported by a ChromeDriver user at https://crbug.com/chromedriver/2967. The user's app uses a version of Protractor that only supports OSS protocol.

OS: All
Browser: Chrome
Browser version: 75.0.3770.100
Browser Driver version: ChromeDriver 75.0.3770.140 used in the repro, but should apply to any W3C compliant driver
Language Bindings version: N/A
Selenium standalone server version (if applicable): 3.141.59

diemol:
@JohnChen0 
[...]

PR #7710, opened 2019-10-24
[java] Set mouse button in W3C codec

JohnChen0:
Some old-style mouse commands can optional specify a button ID, but this
ID is ignored when the commands are converted into W3C actions by the
codec, and right clicks become left clicks. This commit fixes that.
It fixes RemoteMouse.contextClick, and also resolves issue #7393.


barancev:
Thank you, merged as 085ceed1f55fbaaa1d419b19c73264415c394905

PR #7735, opened 2019-11-01
Rubocop Autofix Rakefile

luke-hill:

Run rubocop autofixes (Only automatic not manual), across *just* the Rakefile.

Turn a file with >300 offenses to a file with <75 offenses (Majority of remaining ones are LineLength ones which are manual to fix)

I've separated the autofixes out individually apart from about 10-12 Style ones which all had 1-3 small offenses.


shs96c:
Seems reasonable to me. Thank you!

PR #7736, opened 2019-11-01
CrazyFun: Remaining Namespace work (Aside from the .rl items)

luke-hill:

Everything below is without considering 4 files:
`rake_tasks/crazy_fun/mappings/base_grammar.rl`
`rake_tasks/crazy_fun/mappings/build_grammar.rb`
`rake_tasks/crazy_fun/mappings/build_grammar.rl`
`rake_tasks/crazy_fun/mappings/main.rb`

After merging this PR we have ....
- All files contained under a single directory `rake_tasks`
- All files underneath that correctly namespaced according to folder/module customs
- "In theory" no actual changes in terms of functionality

The next 2/3 tasks would be primarily
- Renames: Some of the file/folder structures are so verbose and unhelpful they need looking at. Also some of the filenames are also a bit manky. I propose this be tackled in 1 PR
- Syntax: Running the rubocop autofixer over a directory at a time will produce 100's of optimisations and speed improvements
- Conversion to Ruby-isms: This one is tricky and will be best left until everything else is merged in. Certain areas have issues where we use items that don't need to be enumerated (Arrays of length 1), symbols with string covers on (primarily with the `-` key in them). Both of these are simple enough fixes but I think it's highly likely something could go wrong when doing this, hence delaying it till the end.

Review away. I'd also highly advise looking at this with tab mode switched on (So you don't see a huge set of +/- differences). In order to view the condensed diff without tab highlights being present polluting the PR, use this link https://github.com/SeleniumHQ/selenium/pull/7736/files?utf8=%E2%9C%93&diff=unified&w=1

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -80,7 +80,7 @@ crazy_fun = SeleniumRake::CrazyFun.new
 #
 # If crazy fun doesn't know how to handle a particular output type ("java_library"
 # in the example above) then it will throw an exception, stopping the build
-RakeMappings.new.add_all(crazy_fun)
+CrazyFun::Mappings::RakeMappings.new.add_all(crazy_fun)
 CrazyFun::Mappings::RubyMappings.new.add_all(crazy_fun)
 
 # Finally, find every file named "build.desc" in the project, and generate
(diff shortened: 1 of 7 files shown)


shs96c:
Looks like ruby to me. Thank you!

PR #7743, opened 2019-11-05
Missing a dependencies on the Netty client so -Dwebdriver.http.factor…

adamgoucher:
Missing a dependencies on the Netty client so -Dwebdriver.http.factory=netty would just lead to tears. (And a class not found exception.)


shs96c:
Nice spot. Thank you!

PR #7746, opened 2019-11-05
Enable READ retries for remote connection pool 

haskellpointersauce:
When using keep-alive connections for the remote connection
it can sometimes occur that the connection is interrupted.

Simply reconnecting seems to be a sufficient measure to continue
the test without any apparent negative consequences.

The documented behavior of the retries parameter is that it only
retries for idempotent HTTP verbs, so I believe this may be a
reasonable approach.

### DIFF
--- a/py/selenium/webdriver/remote/remote_connection.py
+++ b/py/selenium/webdriver/remote/remote_connection.py
@@ -33,7 +33,7 @@ from .errorhandler import ErrorCode
 from . import utils
 
 LOGGER = logging.getLogger(__name__)
-
+READ_RETRIES=3
 
 class RemoteConnection(object):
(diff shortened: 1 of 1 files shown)


haskellpointersauce:
I agree this no longer seems to be the solution I was looking for. Just in case someone googles this or something I thought I'd mention the following about what appears to be the retry policy utilized in the remote connection class (via defaults provided by urllib3):

Reading the docs for urllib3, it kinda seems like the default retry strategy is retries=None, which perhaps counterintuitively means "to retry until you receive a response." [1]

So if this is the case, this change would actually make the client perform fewer retries than would otherwise been performed.

[1] https://urllib3.readthedocs.io/en/latest/reference/index.html#module-urllib3.connectionpool

PR #7754, opened 2019-11-07
* Make one enable to Fetch Domain - there is an issue with Disable on…

NoamGoren:
… Browser level

* remove Dead imports
* Handle broken Pattern (we are accepting all now wildcard)
* Small indentation handling

Thanks to : @dratler  for help and thanks for #Intuit supporting this commit #Intuit #IntuitGED #IntuitGED2019

related to issue #6667 


barancev:
I disagree with this "fix". You have just simplified the test to the state when it's not strong enough.

The line that set the pattern for URLs that should be interrupted has been commented out. As the result, the listener has never been called. It's a false positive.

diemol:
Closing this PR as we did not get more comments from the OP

PR #7756, opened 2019-11-07
Add Chrome tests to bazel build.

TriciaCrichton:
Added BUILD.bazel file for the java Chrome tests. This is modeled after the devtools build file.


TriciaCrichton:
The first failures in the build are from running FormHandlingTest for Firefox. That is not related to this change.

The second failures are caused because file needed by ChromeOptionsFunctionalTest is not found. The test was added in 2017; the file stil exists in third_party/chrome_ext. I think we just need to update the path. How do we point to the third_party directory for the bazel build?

TriciaCrichton:
fixes #7785

barancev:
We've changed Java build infrastructure to use external dependencies instead of stored in third_party directory:
https://github.com/SeleniumHQ/selenium/commit/6085af940e00132ef09b8594876ae5240575f520

Can you please update this PR to use external dependencies too.

TriciaCrichton:
Updated the build file to use external dependencies.
Because I don't know how to setup access to third_party/chrome_ext, I disabled that test.

We use the ChromeDriverTests file to run the tests locally, so I replaced the imports of other *Tests files with explicit lists of the included tests.

TriciaCrichton:
There are 3 unrelated test failures in the build: 
FormHandlingTest.testShouldSubmitAFormUsingTheNewlineLiteral - broken by a change to the Selenium code, see issue 7886
PageLoadingTest.testShouldBeAbleToAccessPagesWithAnInsecureSslCertificate - I submitted PR 7888 to fix the test.
ChromeDevToolsNetworkTest.verifyCertificatesAndOverrideUserAgent - the test does not configure ChromeDriver to Accept Insecure Certificates. This isn't a change I feel confident in making.

TriciaCrichton:
@barancev Can you take a look at this again? I changed to use external dependencies.

java/client/test/org/openqa/selenium/chrome/ChromeDriverTests.java
```diff
@@ -19,15 +19,104 @@
...
-    StandardSeleniumTests.class,
-    ChromeOptionsFunctionalTest.class,
-    DevToolsTests.class
-})
+@RunWith(ChromeDriverTests.ChromeTestRunner.class)
 public class ChromeDriverTests {
-}
+  /* Use a custom suite to enable use of Class.forName, which is not allowed
+      by the annotations. This isolates tests run by ChromeDriver from changes
+      in the build tools. */
```
shs96c, line 28: We're not planning on shifting build tools again for a while, so I'm not sure that this is a concern.

java/client/test/org/openqa/selenium/chrome/ChromeDriverTests.java
```diff
@@ -19,15 +19,104 @@
...
 
-
-@RunWith(Suite.class)
-@Suite.SuiteClasses({
-    StandardSeleniumTests.class,
-    ChromeOptionsFunctionalTest.class,
-    DevToolsTests.class
-})
+@RunWith(ChromeDriverTests.ChromeTestRunner.class)
 public class ChromeDriverTests {
```
shs96c, line 25: The way we're working on making this work is to avoid test suites, and instead use `java_selenium_test_suite` in our `BUILD.bazel` files. This creates a target per test, and so allows us to shard them more efficiently. It's defined in `//java:defs.bzl`.

That takes a `browser` attribute: https://github.com/SeleniumHQ/selenium/blob/master/java/private/test.bzl#L29 which I suspect you want to set to `["chrome", "edge"]`

shs96c (changes requested):

diemol:
Closing as there was a follow up at #8275

Issue #7662, opened 2019-10-08
GridTNG: All communication over HTTPS


shs96c:
Provided the user can configure it. No self-signed certs.

adamgoucher:
Node endpoints now can speak HTTPS. 0mq is still non-secure but beyond scope of this ticket.

PR #7767, opened 2019-11-10
https between components

adamgoucher:
Alright. This is for #7662 

Essentially there is a big on/off flag (--https) and then you need to provide a certificate (--https-certificate) and a key (--https-private-key). There is also an info section on how to wire it up including making yourself some certificates and how to teach Java to trust them.

Not sure how to write tests as it requires certificates, but can confirm I have locally been able to light up a browser and drive it.

This does not do the follow, each will have to be a separate item
- secure zeromq which is still plain tcp
- teach the client bindings about certificates (my tests were with the python ones and they just work by saying the remote_connection is https but it spits out a warning about verifying certificates on every communication over the wire)


java/server/src/org/openqa/selenium/grid/commands/security.txt
```diff
@@ -0,0 +1,92 @@
+Selenium Grid by default communicates over HTTP. This is fine for a lot
+of use cases, especially if everything is contained within the firewall
+and against test sites with testing data. However, if your server is
+exposed to the Internet or us being used in environments with production
```
shs96c: "us"?

java/server/src/org/openqa/selenium/grid/commands/security.txt
```diff
@@ -0,0 +1,92 @@
...
+
+Certificates
+
+The Selenium Grid will not operate with self-signed certificates, as a
+result you will need to have some provisioned to you from a Certificate
+Authority of some sort. For experimentation purposes you can use MiniCA to
+create and sign your certificates.
+
+```
+  ~/go/bin/minica --domains sessions.grid.com,distributor.grid.com,router.grid.com
```
shs96c: Assume commands are on the `PATH` already. If someone's installed minica via (eg) homebrew, the location will be different.

java/server/src/org/openqa/selenium/grid/commands/security.txt
```diff
@@ -0,0 +1,92 @@
...
+
+And since we are using a non-standard CA, we have to teach Java about it. To do that
+you add it to the cacert truststore which is by default, $JAVA_HOME/jre/lib/security/cacerts
+
+```
+  sudo keytool \
+      -import \
+      -file /path/to/minica.pem \
+      -alias minica \
+      -keystore /Library/Java/JavaVirtualMachines/jdk-11.jdk/Contents/Home/lib/security/cacerts \
```
shs96c: `$JAVA_HOME`, perhaps?

java/server/src/org/openqa/selenium/grid/server/BaseServerFlags.java
```diff
@@ -53,6 +53,18 @@
   @ConfigValue(section = "server", name = "allow-cors")
   private boolean allowCORS = false;
 
+  @Parameter(description = "Whether the Selenium server should communicate over https", names = "--https")
```
shs96c: It would seem that having either `--http-private-key` or `--http-certificate` implies that the connections should be secured....

java/server/src/org/openqa/selenium/grid/server/BaseServerOptions.java
```diff
@@ -87,7 +88,11 @@ public URI getExternalUri() {
     int port = getPort();
 
     try {
-      return new URI("http", null, host, port, null, null, null);
```
shs96c: Consider: `return new URI(isSecure() ? "https" : "http", null, host, port, null, null, null)` to emphasise that the only thing that changes is the scheme.

shs96c: Or `getScheme()` as a (private?) method on this class.

java/server/src/org/openqa/selenium/grid/server/BaseServerOptions.java
```diff
@@ -96,4 +101,24 @@ public URI getExternalUri() {
   public boolean getAllowCORS() {
     return config.getBool("server", "allow-cors").orElse(false);
   }
+
+  public boolean isSecure() {
+    return config.getBool("server", "https").orElse(false);
```
shs96c: Consider `return config.get("server", "https-private-key").isPresent() && config.get("server", "https-certificate").isPresent();`

shs96c: Or throwing a `ConfigException` if one is set and not the other.

java/server/src/org/openqa/selenium/grid/server/BaseServerOptions.java
```diff
@@ -96,4 +101,24 @@ public URI getExternalUri() {
...
+  }
+
+  public File getPrivateKey() {
+    String privateKey = config.get("server", "https-private-key").orElse(null);
+    if (privateKey != null) {
+      return new File(privateKey);
+    }
+    throw new ConfigException("you must provide a private key via --https-private-key when using --https");
```
shs96c: If you're using the `--https` flag, check that value before throwing an exception.

java/server/src/org/openqa/selenium/grid/server/BaseServerOptions.java
```diff
@@ -96,4 +101,24 @@ public URI getExternalUri() {
...
+  }
+
+  public File getCertificate() {
+    String certificatePath = config.get("server", "https-certificate").orElse(null);
+    if (certificatePath != null) {
+      return new File(certificatePath);
+    }
+    throw new ConfigException("you must provide a certificate via --https-certificate when using --https");
```
shs96c: As above.

java/server/src/org/openqa/selenium/netty/server/NettyServer.java
```diff
@@ -92,10 +109,18 @@ public void stop() {
...
-    b.group(bossGroup, workerGroup)
-      .channel(NioServerSocketChannel.class)
-      .handler(new LoggingHandler(LogLevel.INFO))
-      .childHandler(new SeleniumHttpInitializer(handler));
+
+    if (secure) {
+      b.group(bossGroup, workerGroup)
+        .channel(NioServerSocketChannel.class)
+        .handler(new LoggingHandler(LogLevel.INFO))
+        .childHandler(new SeleniumHttpsInitializer(handler, sslCtx));
```
shs96c: Why not just modify the existing `SeleniumHttpInitializer` to accept a `null` `sslCtx` and have it only add the SSL handler if the value is not null?

java/server/src/org/openqa/selenium/netty/server/SeleniumHttpsInitializer.java
```diff
@@ -0,0 +1,53 @@
...
+
+  SeleniumHttpsInitializer(HttpHandler seleniumHandler, SslContext sslCtx) {
+    this.seleniumHandler = Objects.requireNonNull(seleniumHandler);
+    this.sslCtx = Objects.requireNonNull(sslCtx);
+  }
+
+  @Override
+  protected void initChannel(SocketChannel ch) {
+    ch.pipeline().addLast("ssl", sslCtx.newHandler(ch.alloc()));
+    ch.pipeline().addLast("codec", new HttpServerCodec());
```
shs96c: Is this the same as the `SeleniumHttpInitializer`? If so, and you're not following the suggestion from above, consider something like:

```
  ch.pipeline().addLast("ssl", sslCtx.newHandler(ch.alloc()));
  new SeleniumHttpInitializer(seleniumHandler).initChannel(ch);
```

to minimise the amount of duplicated code, and to highlight the relationship between the two.

java/server/test/org/openqa/selenium/grid/router/EndToEndTest.java
```diff
@@ -94,7 +97,7 @@
         () -> {
           try {
             return createRemotes();
-          } catch (URISyntaxException e) {
+          } catch (URISyntaxException | SSLException | CertificateException e) {
```
shs96c: Nit please alphabetise.

java/server/test/org/openqa/selenium/grid/router/EndToEndTest.java
```diff
@@ -121,7 +124,7 @@ public void setFields() {
     this.clientFactory = (HttpClient.Factory) raw[1];
   }
 
-  private static Object[] createInMemory() throws URISyntaxException, MalformedURLException {
+  private static Object[] createInMemory() throws URISyntaxException, MalformedURLException, SSLException, CertificateException {
```
shs96c: At this point, `throw Exception` is fine. It's just a test.

shs96c (commented):
Looks like an excellent start. Thank you!

java/server/src/org/openqa/selenium/grid/server/BaseServerOptions.java
```diff
@@ -96,4 +97,24 @@ public URI getExternalUri() {
   public boolean getAllowCORS() {
     return config.getBool("server", "allow-cors").orElse(false);
   }
+
+  public boolean isSecure() {
+	return config.get("server", "https-private-key").isPresent() && config.get("server", "https-certificate").isPresent();
+  }
+
+  public File getPrivateKey() {
```
shs96c: In a follow-up PR, we should make this a `Path`

java/server/src/org/openqa/selenium/netty/server/NettyServer.java
```diff
@@ -46,19 +48,30 @@
...
 
   private Channel channel;
 
-  public NettyServer(BaseServerOptions options, HttpHandler handler) {
+  public NettyServer(BaseServerOptions options, HttpHandler handler) 
+    throws SSLException, CertificateException {
     Objects.requireNonNull(options, "Server options must be set.");
     Objects.requireNonNull(handler, "Handler to use must be set.");
 
+    Boolean secure = options.isSecure();
```
shs96c: `boolean`: we never expect this to be `null`, and the unboxing that happens on the next line will throw an NPE if that happens.

java/server/src/org/openqa/selenium/netty/server/SeleniumHttpInitializer.java
```diff
@@ -30,13 +31,18 @@
...
+  SeleniumHttpInitializer(HttpHandler seleniumHandler, SslContext sslCtx) {
     this.seleniumHandler = Objects.requireNonNull(seleniumHandler);
+    this.sslCtx = sslCtx;
   }
 
   @Override
   protected void initChannel(SocketChannel ch) {
+  	if (sslCtx != null) {
+	  ch.pipeline().addLast("ssl", sslCtx.newHandler(ch.alloc()));
+	}
```
shs96c: Nit: weird indentation.

java/server/src/org/openqa/selenium/netty/server/SeleniumHttpsInitializer.java
```diff
@@ -0,0 +1,53 @@
+// Licensed to the Software Freedom Conservancy (SFC) under one
```
shs96c: This file can be deleted now.

java/server/src/org/openqa/selenium/grid/server/BaseServerOptions.java
```diff
@@ -96,4 +97,24 @@ public URI getExternalUri() {
   public boolean getAllowCORS() {
     return config.getBool("server", "allow-cors").orElse(false);
   }
+
+  public boolean isSecure() {
+	return config.get("server", "https-private-key").isPresent() && config.get("server", "https-certificate").isPresent();
```
shs96c: Nit: indentation

java/server/src/org/openqa/selenium/netty/server/NettyServer.java
```diff
@@ -46,19 +48,30 @@
   private final int port;
   private final URL externalUrl;
   private final HttpHandler handler;
+  private final SslContext sslCtx;
 
   private Channel channel;
 
-  public NettyServer(BaseServerOptions options, HttpHandler handler) {
+  public NettyServer(BaseServerOptions options, HttpHandler handler) 
+    throws SSLException, CertificateException {
```
shs96c: Consider wrapping the `SSLException` with an `UncheckedIOException` so that everything coming out of this is a `RuntimeException`. Our own code wraps and rethrows, so it's not like we're hiding much.

shs96c (approved):
Looks like it's basically ready. I've left a few comments, but you should still have the commit bit. Land it when you feel good to go :) We can always iterate on this once it's in the tree.

java/server/src/org/openqa/selenium/netty/server/NettyServer.java
adamgoucher: I think the current implementation of isSecure ensures it always is Boolean.

PR #7771, opened 2019-11-11
include accessibility id in FindBy

JulianaPRibeiro:
When you are using a cross platform mobile framework like react native, we usually need to declare two ID's, one for Android and another for IOS. With this changes, we can use just the **accessibility Id** in both. 
> Example:
```
@AndroidFindBy(accessibility ="Login_phone")
@iOSXCUITFindBy(xpath ="//*[@name='Login_phone']/preceding-sibling::*[1]/*[2]")
private MobileElement FIELD_PHONE;
```
_AndroidFindBy and iOSXCUITFindBy from Appium._

> With the changes:
```
@FindBy(accessibility ="Login_phone")
private MobileElement FIELD_PHONE;
```
_FindBy from Selenium._

This will provide a much clean code and make it easier to automate tests.

### DIFF
--- a/java/client/src/org/openqa/selenium/support/FindBy.java
+++ b/java/client/src/org/openqa/selenium/support/FindBy.java
@@ -76,6 +76,8 @@ public @interface FindBy {
 
   String xpath() default "";
 
+  String accessibility() default "";
+
   class FindByBuilder extends AbstractFindByBuilder {
     @Override
     public By buildIt(Object annotation, Field field) {


JulianaPRibeiro:
Could you check if any errors occurred in the CI?
It was just adding a new identifier type to the `FindBy.java class`
Is it possible to identify the problem in the build?

barancev:
Can you please explain how adding a new parameter to `@FindBy` annotation will help you? Why is it better that creating a new custom annotation that belongs to the same llibrary that will use it?

JulianaPRibeiro:
_**Appium**_ aims to automate any mobile application of any language and test framework, with full access to backend APIs and DBs from the test code. It uses **_Selenium WebDriver API_**

It has two parameters for Mobile elements in _Appium_:
`@AndroidFindBy` and `@iOSXCUITFindBy`

The first for _Android_ and another for _IOS_

The problem is: both have accessibility as a parameter _type,_ and for the same element ID, I have to declare in code twice, once for _Android_ and once for _IOS_.

> Example:
```
@AndroidFindBy (accessibility = 'Login_button')
@iOSXCUITFindBy (accessibility = 'Login_button')
private MobileElement Login_button;
```
> With accessibility parameter in `@FindBy` annotation
```
@FindBy (accessibility = 'Login_button')
private MobileElement Login_button;
```

If you had the `accessibility` type for `@FindBy`, you could get an element in the test code only once using `accessibility`, which, unlike `xpath` and other parameter types, fits both platforms (_Android_ and _IOS_).

_Appium_ already uses _Selenium WebDriver API_, it would be possible to use the `@FindBy `annotation for both platforms, since both have the `accessibility` parameter.

This would reduce the UI automation code by 50%, as for each screen element there are two parameters to pass, and with _accessibility_ in `@FindBy` it would be just one.

This would bring a huge gain for both native and hybrid UI testing automation.

JulianaPRibeiro:
I signed the Contributor License Agreement.

jlipps:
@JulianaPRibeiro I'm not sure of all the relationships between code and projects here, but is it not possible to add this in the Appium java client instead?

JulianaPRibeiro:
@jlipps No, because in _Appium_ it is not necessary to change the parameters and their types. This change is valid and significant only if it occurs in `@FindBy`.
Having the `"accessibility"` parameter in `@FindBy` would save me from having to use the Appium parameters, which are two, since _Appium_ uses the _Selenium WebDriver API_, which would make it possible to use only one. That's why there is gain from this change.

jlipps:
@shs96c strikes me that this situation implies something suboptimal about the architecture of the FindBy annotation system? Still not sure I completely understand the situation, but should it be possible to extend this differently?

@JulianaPRibeiro would an alternative solution be the creation of a new `FindBy` class in the appium client, like this?

```java
@AppiumFindBy (accessibility = 'Login_button')
```

JulianaPRibeiro:
@jlipps  A new _Appium_ class could be created, as you exemplified.
Basically, with `@AndroidFindBy` and `@iOSXCUITFindBy` I need to declare the elements twice, once for _Android_ and once for _IOS,_ since `accessibility` is in these classes, and I can't reuse the `@FindBy` class, which doesn't have this annotation),

I find it simpler just to include `accessibility` in `@FindBy`, than to have to create an entire class in _Appium_ just to include this parameter.

Unlike other types of identifiers (label, xpath, text, etc.), `accessibility` is recognized by both platforms (_Android_ and _IOS_), as I explained in the comments above.

shs96c:
@jlipps, the design of this class is clearly suboptimal. We've been gently nudging people away from the `PageFactory` for a while, largely because of this (that is, the idea is sound, but the implementation is far from that)

The "correct" way to extend this with a custom annotation involves adding a custom `ElementLocatorFactory`, which returns an `ElementLocator` for each field in a class. _That_ would use the `Annotations` class to make use of the existing `FindBy` and then implement the logic for a custom location strategy using a custom annotation (eg. you'd annotate fields in your class with `@AccessibilityLocator("cheese") private WebElement button;` and your `ElementLocator` would check for the presence of that annotation.

But that's utterly unintuitive because of the levels of indirection that go on, and it doesn't allow people to add new entries to the `FindBy` annotation itself (since `How` is an enum, and, ooof, those things aren't meant to be extended, and that was a design flaw in our initial design)

In order to use the changed annotation, someone will already need to use their own `ElementLocatorFactory` or `FieldDecorator`. Does that already exist in Appium?

barancev:
As @shs96c explained, we're not going to extend `How` enum, but you can follow the suggested "proper" extension procedure: creating a custom `ElementLocatorFactory` and a custom annotation.

PR #7784, opened 2019-11-14
Added option to use char and string when generating atoms.cc

### LINKED ISSUE #7783: Enable use of char and string when generating atoms.cc
## 🚀 Feature Proposal

The previous build structure had an options to generate javascript/atoms.h & atoms.cc using char* and std::string. The bazel build uses only char_t and std::wstring. We need to restore the option to choose; char_t and std::wstring will be the default.

Also, the current build actually uses chat_t, not char_t in the header file.

## Motivation

Please outline the motivation for the proposal.
The bazel build breaks compatibility with ChromeDriver, which code uses char* and std::string; the chromium code style doesn't allow std::wstring.

## Example

instead of 
extern const wchar_t* const CLEAR[];
We need:
extern const char* const CLEAR[];

We are preparing a PR for this change.

TriciaCrichton:
fixes #7783 
Add utf8 parameter for gen_file.py to enable creating atoms.cc and atoms.h with char* and std::string types. The default is True to preserve the prior behaviour, using char_t* and std::wstring.

ChromeDriver atoms BUILD was updated to set utf8 to False.


TriciaCrichton:
I don't believe the build failures are related to this change, the first references unresolveable dependencies on @platforms//os. The second is a classpath error while generating javadocs. Those definitions were not modified by this PR.

shs96c:
Ah! Nice. Merging this now. Thank you for the patch!

PR #7576, opened 2019-09-18
.Net string.Format without named parameters

trejjam:
Hey there thank you for a great work there!
I found some issues with `DevToolsSession.LogTrace`, .Net Core does not support named parameters in `string.Format(...)`.
The second typo is in `DriverOptions.GetTypeSafeOptionName` in generating error messages.

PR #7794, opened 2019-11-18
[.Net] Fix WS message handler

trejjam:
Hi, I fixed how the DevTools WS protocol is handled in .Net implementation.

The original implementation was waiting for all requests to finish, and it causes faults when multiple requests sent to the driver. This implementation of the ProcessIncomingMessage returns a response when available.

The sad thing is that the fix was done on top of the #7576, which was cold for two months. My contract where I was using .Net selenium finished now; I am not able to wait longer.


jimevans:
This has been merged as 5f38a0d. A corresponding change has also been made to the [project that generates the API code](https://github.com/jimevans/chrome-dev-tools-generator) for CDP support in .NET. Thank you for the contribution.

PR #7808, opened 2019-11-23
Add support for shadow elements

diego-fernandez-santos:

Atoms checks if an element is still attached to the DOM before every interaction. This check is done [here](https://github.com/SeleniumHQ/selenium/blob/master/javascript/atoms/inject.js#L524).  The problem is that it doesn't take into account shadow elements. These elements doesn't have a parentNode but they can still be attached to the DOM, for example shadowRoot elements.

If you try to interact with WebComponents in IPhone an "Element is no longer attached to the DOM" error would be raised.

I do this pull request in this branch because this is the current version that [appium-remote-debugger](https://github.com/appium/appium-remote-debugger/blob/master/Makefile) currently using.

You can replicate this problem directly in the browser following the next steps:

- Go to https://shop.polymer-project.org/
- Open the inspector console and run the following script:
```
var doc = { documentElement : document }

  var element = document
    .querySelector('shop-app').shadowRoot
    .querySelector('shop-home').shadowRoot
    .querySelector('shop-button a');

  function getElementIfAttached(el) {
    var node = el;
    while (node) {
      if (node == doc.documentElement) {
        return el;
      }    
      node = node.parentNode;
    }
  }

  getElementIfAttached(element)
```

You can see that getElementIfAttached doesn't return the element although the shop-button is still attached to the DOM.


javascript/atoms/inject.js
```diff
@@ -524,6 +524,9 @@ bot.inject.cache.getElement = function(key, opt_doc) {
     if (node == doc.documentElement) {
       return el;
     }
+    if (node.host && node.nodeType === 11) {
```
AutomatedTester: instead of host I think that you want to `node.ownerDocument` to check if fragment is still attached.

Does this work on ShadowDOM as they are not document fragments

diego-fernandez-santos: I did the change.

I use that other way to try to keep the same code style. I think that it was valid (according to the last [W3C](https://www.w3.org/TR/dom41/#host)). However I think that your recommendation is faster.

imurchie:
This latest iteration doesn't work on MobileSafari, on any version of iOS I try. The first solution _did_ work. At the point of failure, `node.ownerDocument.parentNode` does not exist but `node.host.parentNode` does.

PR #7820, opened 2019-11-28
[rb] convert Bridge modules to subclasses

titusfortner:
I think this simplifies things now that we've moved to the Drivers being subclassed. Is there any reason why we would want these Bridges to get extended as modules on `Remote::Bridge`?


titusfortner:
hmm, looks like we discussed this back here: #7312
Not sure why I didn't like it.
I'm not creating new files with this PR; it's all private API so we should have control over what is needed.

p0deje (approved):
LGTM

PR #7830, opened 2019-11-29
[rb] update Selenium Logger

titusfortner:
Finally got around to this.

1. Allow creating a separate Logger instance with a different progname so we can remove the "duplicate" logging wrappers in Watir, Webdrivers and others

2. Supports adding ids so if someone wants to silence specific warnings/deprecations they can

3. Allows #deprecate to take a block that will append to the logged message (e.g. reasons for deprecation or link to blog / documentation, etc)

4. Added Custom RSpec Matchers to make it easier to filter out noise from STDOUT

5. I also threw in `aggregate_failures` in spec_helpers because I was playing with it. Let me know if there's a reason we might now want it there.


p0deje (commented):
LGTM! There are some unit tests and lint failures though.

p0deje (approved):

PR #7832, opened 2019-12-02
Implement w3c parameters for options class

titusfortner:
Ok, after 2 years of trying to figure out how to do this, I think this is the right way to go. This replaces #7378, and hopefully addresses all the concerns raised there. I added a bunch of specs to master that represent current behavior, so it should be easy to see what this does.

**Problem statement:**
1. Many users are confused as to what goes into "options" and what goes into "capabilities" parameters right now. As such, it is very common that users just generate their own Hash from scratch, which has no protection against invalid parameters. Java & .NET both use their Options class to set all w3c and browser specific values, without making a distinction between them.
2. There is currently no easy way to support 3rd party extension capabilities

**Features of this PR:**
1. All functionality that exists currently is maintained in this release, just with deprecations.

2. `:desired_capabilities` and `:options` parameters are deprecated in favor of using `:capabilities` parameter with an Array
```
# old:
driver = Selenium::WebDriver.for :chrome, {desired_capabilities: caps, options: opts}
# new:
driver = Selenium::WebDriver.for :chrome, {capabilities: [caps, opts]}
```

3. `Capabilities` instances will continue to be supported, and will be the only way for a user to hard code everything from a Hash.

4. `Hash` instance will not be supported directly in the capabilities array, it would need to get put into Capabilities class first
```
# old:
driver = Selenium::WebDriver.for :chrome, {desired_capabilities: hash_of_options}
# new:
driver = Selenium::WebDriver.for :chrome, {desired_capabilities: Remote::Capabilities.new(hash_of_options)}
```

5. `Options` classes now can handle w3c values as well as browser specific values. Because we know exactly what is allowed by w3c, we can assume that anything not in the spec belongs in the browser specific options, which would allow users to take advantage of new browser specific features without the bindings having to explicitly support them. It will be the recommended way going forward to correctly build valid w3c capabilities.

6. Because `:capabilities` takes an Array, the Ruby bindings can now directly support extension capabilities by third parties or test libraries so long as the class instance being passed in responds to `#as_json`

7. There is no logic for how to deal with people passing in multiple objects with different key/values, so people can still do dumb things here, but they are already doing some of that with Hashes currently

8. So the non-deprecated allowed values for initializing a Driver include: `:service`, `:capabilities`, `:url`, `:http_client`, and that's it. Nice and simple and concerns finally all separated out.

9. In a very Ruby fashion, this code provides multiple ways of doing things, but the new approach to `Options` should be much easier to maintain going forward than `Capabilities` is right now.


titusfortner:
@p0deje & @twalpole we have a few more things to work on with Ruby options before 4.0, can we agree on this as a good solution for the various things we all wanted?

p0deje (approved):
My apologies for this PR taking so long to be reviewed. I believe it is now more consistent to C#/Java. I feel like we will see some bug reports because of a difference in default capabilities payload, but what else we have alphas for? Let's just make sure we have green CI before merge.

luke-hill:
@titusfortner / others. Could I ask before we go live with 4.0 (Or whenever really before it). Could we get 5-10 examples of "preferences" you might create in 3.x and 4.0? (I use preferences to avoid confusion with keywords).

i.e. this is a chrome 83 driver in 3.x, this is an identical driver in 4.0 e.t.c.

titusfortner:
@luke-hill what are you looking for? We don't actually have any toggles for chrome versions in capabilities code.
Do you mean a more user friendly depiction of these specs? https://github.com/SeleniumHQ/selenium/blob/trunk/rb/spec/unit/selenium/webdriver/chrome/driver_spec.rb#L142

PR #7842, opened 2019-12-03
Add support to debug virtual authenticators

### LINKED ISSUE #7829: Implement the rest of the WebAuthn automation methods
## 🚀 Feature Proposal

#7753 added basic support for WebAuthn automation in the form of the `addVirtualAuthenticator` and `removeVirtualAuthenticator` methods. Implement the rest of the methods: `addCredential`, `getCredential`, `removeCredential`, `removeAllCredentials`, `setUserVerified`.

## Motivation

Support more WebAuthn use-cases, like logging in with a specific credential, handling user verification failures, and inspecting generated credentials.

## Example

For a page like this

```html
<button id="login">Log in with WebAuthn</button>
<span id="result"></span>
```

```javascript
document.getElmentById("login").addEventListener("click", () => {
  navigator.credentials.get({...}).then(assertion => {
    document.getElementById("result").innerText = "Success!";
  })
});
```

A test could be written like so:

```java
// Set up the virtual authenticator.
HasVirtualAuthenticator virtualAuthenticatorManager = ((HasVirtualAuthenticator) driver);
VirtualAuthenticatorOptions options = new VirtualAuthenticatorOptions();
VirtualAuthenticator authenticator =
    virtualAuthenticatorManager.addVirtualAuthenticator(options);

// New code: inject a credential into the authenticator.
// Without this API the developer would have to generate a credential through
// navigator.credentials.create({...}) on the javascript side and would have less
// control over the resulting PK.
Credential credential = Credential.createNonResidentCredential(...params);
virtualAuthenticatorManager.addCredential(authenticator, credential);

// Assert login successful:
driver.findElement(By.id("login")).click();
WebElement result = driver.findElement(By.id("result"));
new WebDriverWait(driver, 60, 1000).until(
    driver -> !result.getText().equals(""));
assertThat(result.getText()).isEqualTo("Success!");
```

nsatragno:
### Description
This patch adds support for the `addCredential`, `getCredentials`, `removeCredential`, `removeAllCredentials` and `setUserVerified` commands of the WebAuthn automation API.

To add these methods directly into the `VirtualAuthenticator` class, I extracted it into an interface and implemented a `RemoteVirtualAuthenticator` as inner class of `RemoteWebDriver`.

### Motivation and Context
Support more WebAuthn use cases:
* Trying an assertion without having to create a credential first.
* Handle user verification failures.
* Use specific private keys to back credentials.
* Handle credentials being cleared from an authenticator gracefully.

### Types of changes

### Checklist

Fixes #7829

### DIFF
--- a/java/client/src/org/openqa/selenium/virtualauthenticator/Credential.java
+++ b/java/client/src/org/openqa/selenium/virtualauthenticator/Credential.java
@@ -17,10 +17,10 @@
 
 package org.openqa.selenium.virtualauthenticator;
 
-import com.google.common.collect.ImmutableMap;
-
 import java.security.spec.PKCS8EncodedKeySpec;
 import java.util.Base64;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 
@@ -104,15 +104,15 @@ public class Credential {
 
   public Map<String, Object> toMap() {
     Base64.Encoder encoder = Base64.getUrlEncoder();
-    ImmutableMap.Builder builder = new ImmutableMap.Builder<String, Object>()
-      .put("credentialId", encoder.encodeToString(id))
-      .put("isResidentCredential", isResidentCredential)
-      .put("rpId", rpId)
-      .put("privateKey", encoder.encodeToString(privateKey.getEncoded()))
-      .put("signCount", signCount);
+    Map<String, Object> map = new HashMap<String, Object>();
+    map.put("credentialId", encoder.encodeToString(id));
+    map.put("isResidentCredential", isResidentCredential);
+    map.put("rpId", rpId);
+    map.put("privateKey", encoder.encodeToString(privateKey.getEncoded()));
+    map.put("signCount", signCount);
     if (userHandle != null) {
-      builder.put("userHandle", encoder.encodeToString(userHandle));
+      map.put("userHandle", encoder.encodeToString(userHandle));
     }
-    return builder.build();
+    return Collections.unmodifiableMap(map);
   }
 }
--- a/java/client/src/org/openqa/selenium/virtualauthenticator/VirtualAuthenticatorOptions.java
+++ b/java/client/src/org/openqa/selenium/virtualauthenticator/VirtualAuthenticatorOptions.java
@@ -17,8 +17,8 @@
 
 package org.openqa.selenium.virtualauthenticator;
 
-import com.google.common.collect.ImmutableMap;
-
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 
 /**
@@ -91,13 +91,13 @@ public class VirtualAuthenticatorOptions {
   }
 
   public Map<String, Object> toMap() {
-    return new ImmutableMap.Builder<String, Object>()
-      .put("protocol", protocol.id)
-      .put("transport", transport.id)
-      .put("hasResidentKey", hasResidentKey)
-      .put("hasUserVerification", hasUserVerification)
-      .put("isUserConsenting", isUserConsenting)
-      .put("isUserVerified", isUserVerified)
-      .build();
+    Map<String, Object> map = new HashMap<String, Object>();
+    map.put("protocol", protocol.id);
+    map.put("transport", transport.id);
+    map.put("hasResidentKey", hasResidentKey);
+    map.put("hasUserVerification", hasUserVerification);
+    map.put("isUserConsenting", isUserConsenting);
+    map.put("isUserVerified", isUserVerified);
+    return Collections.unmodifiableMap(map);
   }
 }
--- a/java/client/src/org/openqa/selenium/BUILD.bazel
+++ b/java/client/src/org/openqa/selenium/BUILD.bazel
@@ -1,4 +1,3 @@
-load("@rules_jvm_external//:defs.bzl", "artifact")
 load("//java:version.bzl", "SE_VERSION")
 load("//java:defs.bzl", "java_dist_zip", "java_export")
 
@@ -24,7 +23,7 @@ java_export(
     pom_template = ":template-pom",
     visibility = ["//visibility:public"],
     deps = [
-        artifact("com.google.guava:guava"),
+        # Nothing from third party
         ":manifest",
     ],
 )


java/client/src/org/openqa/selenium/virtualauthenticator/Credential.java
```diff
@@ -0,0 +1,116 @@
...
+  public static Credential createResidentCredential(byte[] id, String rpId,
+      PKCS8EncodedKeySpec privateKey, byte[] userHandle, int signCount) {
+    return new Credential(id, /*isResidentCredential=*/true, Objects.requireNonNull(rpId),
+        privateKey, Objects.requireNonNull(userHandle), signCount);
+  }
+
+  /**
+   * Creates a credential from a map.
+   */
+  public static Credential fromMap(Map<String, Object> map) {
```
shs96c: You may or may not know this, but Selenium's JSON library will use this method to deserialise a JSON stream to a credential. It'll use `toJson` to go the other way. Neither of those methods need to be `public`: if you don't want them to be visible outside this package, they don't need to be.

Which is a really long way of saying that this method could be package protected and everything should still work. Do you want this to be part of the public API?

java/client/src/org/openqa/selenium/virtualauthenticator/Credential.java
```diff
@@ -0,0 +1,116 @@
...
+    this.privateKey = Objects.requireNonNull(privateKey);
+    this.userHandle = userHandle;
+    this.signCount = signCount;
+  }
+
+  public byte[] getId() {
+    return id;
+  }
+
+  public boolean getIsResidentCredential() {
```
shs96c: Although it's not part of the javabean spec, it's legit to just use `is` as the prefix for boolean properties. I guess if the name of the property in the spec is "is resident credential" then `get` makes a lot of sense. Whichever you feel reads better and "feels" better is fine by me.

java/client/src/org/openqa/selenium/virtualauthenticator/Credential.java
```diff
@@ -0,0 +1,116 @@
...
+    return isResidentCredential;
+  }
+
+  public String getRpId() {
+    return rpId;
+  }
+
+  public PKCS8EncodedKeySpec getPrivateKey() {
```
shs96c: Love that this is properly typed.

java/client/src/org/openqa/selenium/virtualauthenticator/Credential.java
```diff
@@ -0,0 +1,116 @@
...
+    return userHandle;
+  }
+
+  public int getSignCount() {
+    return signCount;
+  }
+
+  public Map<String, Object> toMap() {
+    Base64.Encoder encoder = Base64.getUrlEncoder();
+    HashMap<String, Object> map = new HashMap();
```
shs96c: Nit: `Map` on the LHS. 

It doesn't matter a huge amount, but I'm curious why you've used `ImmutableMap` elsewhere and not here.

java/client/src/org/openqa/selenium/virtualauthenticator/VirtualAuthenticatorOptions.java
```diff
@@ -22,7 +22,7 @@
 
 /**
  * Options for the creation of virtual authenticators.
- * @see http://w3c.github.io/webauthn/#sctn-automation
+ * @see https://w3c.github.io/webauthn/#sctn-automation
```
shs96c: 👍🏼

java/client/test/org/openqa/selenium/virtualauthenticator/VirtualAuthenticatorTest.java
```diff
@@ -98,25 +122,58 @@ public void setup() {
...
+      ((JavascriptExecutor) driver).executeAsyncScript(
+        "getCredential([{"
+      + "  \"type\": \"public-key\","
+      + "  \"id\": Int8Array.from(arguments[0]),"
+      + "}]).then(arguments[arguments.length - 1]);", credentialId);
+  }
+
+  @After
+  public void tearDown() {
+    if (authenticator != null)
```
shs96c: Nit: we tend to always use braces in the selenium codebase. It makes it harder to cause chaos accidentally :)

java/client/src/org/openqa/selenium/virtualauthenticator/Credential.java
nsatragno: I think we still need this to be public to [call it from `RemoteWebDriver.java`](https://github.com/SeleniumHQ/selenium/pull/7842/files#diff-fa76272f866d8067bb11a32f27773026R1088) since they're in different packages.

> You may or may not know this, but Selenium's JSON library will use this method to deserialise a JSON stream to a credential.

I had no idea -- are you saying we can somehow wire `Response.getValue()` to do this for us? That would probably be more elegant but don't know how to do it.

java/client/src/org/openqa/selenium/virtualauthenticator/Credential.java
nsatragno: I'll take any chance to make java less verbose. Changed to `isResidentCredential()`.

java/client/src/org/openqa/selenium/virtualauthenticator/Credential.java
nsatragno: I think an earlier version of my patch modified the map in place on `RemoteWebDriver`. Changed to `ImmutableMap` here and in `VirtualAuthenticatorOptions` for consistency. This means I had to add guava as a dependency to the bazel package -- ptal.

java/client/test/org/openqa/selenium/virtualauthenticator/VirtualAuthenticatorTest.java
nsatragno: Fixed here and in a couple more places.

java/client/src/org/openqa/selenium/virtualauthenticator/Credential.java
shs96c: It doesn't do it yet. I really should make that happen. In the Grid code, we have a helper `Values` class that _does_ do this.

java/client/src/org/openqa/selenium/virtualauthenticator/Credential.java
shs96c: Ah! We strive to make sure that the `selenium-api` doesn't need any third party libraries. If we want this to be immutable, the only Java 8 option is to use `Collections.unmodifiableMap`. The `Map.of()` method from Java 11 would be better, but we can't use that yet.

nsatragno: Removed the dependency and used `Collections.unmodifiableMap` instead. Please take a look.

PR #7846, opened 2019-12-05
fix serialization methods

titusfortner:
This is the same PR as #7115 but applied to the 3.x branch
This will actually help with code I'm writing for Sauce right now. 

Any objections to merging it for 3.x? Unit Tests are all passing


p0deje:
Good from my side!

Any reason to have it a PR rather than just cherry-picking your 4.x commit in 3.x branch?

titusfortner:
there was some conflict that kept me from doing a straight cherry pick, but its the same code.

PR #7871, opened 2019-12-18
Add `SameSite` attribute to `Cookie`

### LINKED ISSUE #7798: Add `SameSite` attribute to `Cookie`
## 🚀 Feature Proposal

`SameSite` attribute on [`Cookie`](https://selenium.dev/selenium/docs/api/java/org/openqa/selenium/Cookie.html) object

## Motivation

The [`Cookie.Builder`](https://selenium.dev/selenium/docs/api/java/org/openqa/selenium/Cookie.Builder.html) object does not accommodate a `SameSite` field, described [here](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-03#section-5.3.7) in the spec.

This makes it impossible to specify the `SameSite` field of a cookie, which can either be `None`, `Lax`, or `Strict`. This field can be specified in the `Set-Cookie` HTTP header, so the `Cookie` object should also support it.

## Example
```
Cookie cookie = new Cookie.Builder("name", "value")
    .sameSite("lax")
    .build();
```

gjyalpha:
### Description
Add `SameSite` attribute to `Cookie`

### Motivation and Context
Cookie object in Selenium Java doesn't support setting `SameSite`
attribute, described in the spce:
https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-03#section-5.3.7

ChromeDriver already supports it. So improving Selenium Java API to
support it as well.

Partially fix #7798, i.e. the Java part. Other language bindings need to be checked to see if a fix is needed or not.

### Types of changes

### Checklist

I believe all the doc needs to be updated is the javadoc for the new field. Please let me know if there is any other document I need to update.

### DIFF
--- a/java/client/src/org/openqa/selenium/Cookie.java
+++ b/java/client/src/org/openqa/selenium/Cookie.java
@@ -203,6 +203,10 @@ public class Cookie implements Serializable {
     }
   }
 
+  /**
+   * JSON object keys are defined in
+   * https://w3c.github.io/webdriver/#dfn-table-for-cookie-conversion.
+   */
   public Map<String, Object> toJson() {
     Map<String, Object> toReturn = new TreeMap<>();
 
@@ -230,7 +234,7 @@ public class Cookie implements Serializable {
     toReturn.put("httpOnly", isHttpOnly());
 
     if (getSameSite() != null) {
-      toReturn.put("sameSite", getSameSite());
+      toReturn.put("samesite", getSameSite());
     }
 
     return toReturn;
--- a/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java
+++ b/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java
@@ -762,6 +762,8 @@ public class RemoteWebDriver implements WebDriver, JavascriptExecutor,
       ((Collection<?>) returned).stream()
           .map(o -> (Map<String, Object>) o)
           .map(rawCookie -> {
+            // JSON object keys are defined in
+            // https://w3c.github.io/webdriver/#dfn-table-for-cookie-conversion.
             Cookie.Builder builder =
                 new Cookie.Builder((String) rawCookie.get("name"), (String) rawCookie.get("value"))
                     .path((String) rawCookie.get("path"))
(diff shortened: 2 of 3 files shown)


java/client/src/org/openqa/selenium/Cookie.java
```diff
@@ -283,8 +314,14 @@ public Builder isHttpOnly(boolean httpOnly) {
...
 
+    public Builder sameSite(String sameSite) {
+      this.sameSite = sameSite;
+      return this;
+    }
+
     public Cookie build() {
-      return new Cookie(name, value, domain, path, expiry, secure, httpOnly);
+      return new Cookie(
+          name, value, domain, path, expiry, secure, httpOnly, sameSite);
```
Harsha509: Hi @gjyalpha,

the JSON key for new samesite cookie capability in WebDriver draft is like 'samesite' (https://w3c.github.io/webdriver/#cookies) and not as 'sameSite'. Currently am not sure as the discussion is still going about cookie normative reference in github/webdriver.

gjyalpha: Yes, in my commit above 2a0652c, I changed the JSON key from 'sameSite' to 'samesite'. Well, I guess I should way until the discussion is closed. My coworker is actually proposing to change it to 'sameSite' in the spec. Let's see.

Harsha509: Hi @gjyalpha,

I have tested cookie parameters settings JSON key as 'samesite' and 'sameSite' in chrome 80 beta. i've observed that it accepting JSON key only as 'sameSite'. if i provide 'samesite' i am unable to see attribute in log.

If possible, can you please test from your end.

Thanks,
Harsha

Harsha509: hi @gjyalpha,

The change has been done with commit https://github.com/w3c/webdriver/pull/1477.
The JSON is now as 'sameSite'.

rowan-m:
This looks super useful! Anything needed to push it along?

barancev:
Merged as 5fa9a75724fcbf2df0192f88a57479503b741eb3, thank you!

java/client/src/org/openqa/selenium/Cookie.java
gjyalpha: Yes. I'm fixing this in https://github.com/SeleniumHQ/selenium/pull/8265.

PR #7897, opened 2019-12-29
[js] Add windowTypes option support for ChromiumDriver

txava:
### Description
Adds support for setting windowTypes option for ChromeDriver, which
allows getting an extra set of window type handles like webview handles.

### Motivation and Context
In order to be able to interact with a webview, one needs to get a handle for the webview (which is a window handle).
ChromeDriver won't return handles for webviews unless it is explicitly specified via **windowTypes** option see: https://chromedriver.chromium.org/capabilities.

### Types of changes

### Checklist


corevo:
Please add a test where you add a windowType and assert that the array contains that windowType.

diemol:
@txava, would you like to continue with this PR? Seems there are pending comments to resolve.

txava:
@diemol, @corevo 
> Please add a test where you add a windowType and assert that the array contains that windowType.

Added commit with a test for this request.

harsha509 (approved):
Thank you @txava !

PR #7901, opened 2019-12-30
Add: Adding new cookie sameSite (c-nodejs)

harsha509:
Signed-off-by: Sri Harsha <sri_harsha509@hotmail.com>

### Description
Adding new cookie support 'sameSite' to c-nodejs

### Types of changes

### Checklist

### DIFF
--- a/javascript/node/selenium-webdriver/test/cookie_test.js
+++ b/javascript/node/selenium-webdriver/test/cookie_test.js
@@ -174,9 +174,8 @@ suite(function(env) {
       let childUrl = fileserver.whereIs('child/childPage.html');
       await driver.get(childUrl);
       await driver.manage().addCookie(cookie);
-      await driver.manage().getCookie(cookie.name).then(function(actual) {
-        assert.equal(actual.sameSite, "Strict");
-      });
+      const actual = await driver.manage().getCookie(cookie.name);
+      assert.equal(actual.sameSite, "Strict");
     });
 
     it('can add same site cookie property to `Lax`', async function() {
@@ -184,9 +183,8 @@ suite(function(env) {
       let childUrl = fileserver.whereIs('child/childPage.html');
       await driver.get(childUrl);
       await driver.manage().addCookie(cookie);
-      await driver.manage().getCookie(cookie.name).then(function(actual) {
-        assert.equal(actual.sameSite, "Lax");
-      });
+      const actualCookie = await driver.manage().getCookie(cookie.name);
+      assert.equal(actualCookie.sameSite, "Lax");
     });
   });
 


javascript/node/selenium-webdriver/lib/webdriver.js
```diff
@@ -1131,6 +1131,16 @@ class Options {
...
     }
 
+    // if not declared, attribute is set to 'undefined'
+    // (or)
+    // if declared (ex: sameSite:"") and string length is zero, attribute is 'None' by default
+    if(typeof sameSite === "undefined" || sameSite.length===0) {
+      // Do noting. sameSite value will be none by default
+    } else if(sameSite!=="undefined" && (sameSite!=="Strict" && sameSite!=="Lax")) {
+      throw new error.InvalidArgumentError(
+          'Invalid sameSite cookie value "' + sameSite + '". It should be either "Lax" (or) "Strict" ');
```
christian-bromann:
```suggestion
          `Invalid sameSite cookie value ${sameSite}. It should be either "Lax" (or) "Strict" `);
```

javascript/node/selenium-webdriver/lib/webdriver.js
```diff
@@ -1131,6 +1131,16 @@ class Options {
       expiry = Math.floor(date.getTime() / 1000);
     }
 
+    // if not declared, attribute is set to 'undefined'
+    // (or)
+    // if declared (ex: sameSite:"") and string length is zero, attribute is 'None' by default
+    if(typeof sameSite === "undefined" || sameSite.length===0) {
```
christian-bromann: I would only check if the value is `===` `Lax` or `Strict`, there is no other value allowed. So if someone sends an empty string, an error should be thrown.

javascript/node/selenium-webdriver/lib/webdriver.js
Harsha509: Hi @christian-bromann,

I've updated code as suggested. Can you please review

javascript/node/selenium-webdriver/lib/webdriver.js
Harsha509: Hi @christian-bromann,

I've updated code as suggested. Can you please review again

Thanks,
Harsha

javascript/node/selenium-webdriver/lib/webdriver.js
```diff
@@ -1131,6 +1131,14 @@ class Options {
       expiry = Math.floor(date.getTime() / 1000);
     }
 
+    // if not declared, attribute is set to 'undefined'
+    if(typeof sameSite === "undefined") {
+      // do nothing
```
christian-bromann: if you do nothing you could also just remove that part (meaning the whole if) 😉

javascript/node/selenium-webdriver/lib/webdriver.js
```diff
@@ -1131,6 +1131,14 @@ class Options {
...
+    } else if(sameSite!=="Strict" && sameSite!=="Lax") {
```
christian-bromann:
```suggestion
    if(sameSite !== 'Strict' && sameSite !== 'Lax') {
```

not sure about the code style in Selenium but it seems they use single quotes

javascript/node/selenium-webdriver/lib/webdriver.js
Harsha509: if i dont add any cookie like below:

    await driver.manage().addCookie({name:'key', value: 'value'});

SameSite value is set to 'undefined'. Removing the if condition, it expects sameSite value should be 'Lax' or 'Strict'  as it is now undefined-> InvalidArgumentError: Invalid sameSite cookie value 'undefined'. It should be either "Lax" (or) "Strict" 

But it is the users choice of adding sameSite cookie property. if it is by default should be either 'Lax' or 'Strict'. 

Do we have any value @christian-bromann, because chrome sets value to undefined as i have observed.

Suggest me!

christian-bromann: just remove it:
```js
    if(typeof sameSite === "undefined") {
      // do nothing
    if(sameSite !== 'Strict' && sameSite !== 'Lax') {
      throw new error.InvalidArgumentError(
          `Invalid sameSite cookie value '${sameSite}'. It should be either "Lax" (or) "Strict" `);
    }
```
is the same as
```js
    if(sameSite !== 'Strict' && sameSite !== 'Lax') {
      throw new error.InvalidArgumentError(
          `Invalid sameSite cookie value '${sameSite}'. It should be either "Lax" (or) "Strict" `);
    }
```

christian-bromann: Ah I think I understand, then just do:
```js
    if(sameSite && sameSite !== 'Strict' && sameSite !== 'Lax') {
      throw new error.InvalidArgumentError(
          `Invalid sameSite cookie value '${sameSite}'. It should be either "Lax" (or) "Strict" `);
    }
```

Harsha509: Done! :)

Harsha509:
> This looks good to me now. 👍 Thanks!

Thanks for reviewing @christian-bromann .

javascript/node/selenium-webdriver/test/cookie_test.js
```diff
@@ -168,6 +168,26 @@ suite(function(env) {
...
+        assert.equal(actual.sameSite, "Strict");
+      });
+    });
+
+    it('can add same site cookie property to `Lax`', async function() {
+      let cookie = createSameSiteCookieSpec('Lax');
+      let childUrl = fileserver.whereIs('child/childPage.html');
+      await driver.get(childUrl);
+      await driver.manage().addCookie(cookie);
+      await driver.manage().getCookie(cookie.name).then(function(actual) {
```
corevo: The callback is not necessary since you are using async/await

```js
const actual = await driver.manage().getCookie(cookie.name);
assert.equal(actual.sameSite, "Strict");
```

Harsha509: Replaced callback with async/await as suggested

Issue #5869, opened 2018-05-08
Shadow DOM traversal support


43081j:
I've been gradually trying to push support of shadow DOM v1 into selenium (see #4230, #5762).

However, we are still missing one major piece: traversal. 

There needs to be some ability to traverse shadow DOM when calling `FindElement` and what not.

So maybe we can discuss some possible implementations here?

What we can't/shouldn't do:
* Alter how existing selectors work (CSS, XPath, etc. they should all behave as they do now)
* Implement deprecated things like `/deep/`
* Implement our own selector syntax/mechanism

The way I see this working is that we should **treat the DOM as we would in the browser**, meaning there will _not_ be a way to select a deep element in one call:

```
element
  .FindElement(By.CssSelector("foo"))
  .ShadowRoot?
  .FindElement(By.CssSelector("bar"));
```

Which, in the browser, would be structured the same (no shortcuts):

```js
element
  .querySelector('foo')
  .shadowRoot
  .querySelector('bar');
```

So maybe we just need to implement `ShadowRoot` on `WebElement`? Which can be `null` (just like in the browser).

43081j:
also, while i do disagree with us implementing some kind of "shortcut", something like this could work:

```cs
element
  .FindElement(
    By.ShadowSelector(
      By.CssSelector("foo"),
      By.CssSelector("bar")
    )
  );
```

which is the equivalent of `element.querySelector('foo').shadowRoot.querySelector('bar')`.

If we did want a way to call `FindElement` only once and have it handle the null shadow roots, something like this seems feasible maybe?

either way i think we still need to implement `ShadowRoot` though at least.

p0deje:
Even though I don't fully follow the request, I see nothing in regards to shadow root in [specification](https://w3c.github.io/webdriver/). To my knowledge, support for custom commands like `ShadowRoot` is required to be implemented in underlying drivers and so has to be a part of specification.

43081j:
The request is to provide shadow DOM support. To provide a way of using `FindElement` to reach nodes contained within shadow DOM trees.

Currently, you do not have any support for shadow DOM at all in regards to traversing the tree.

So if we can't add `ShadowRoot`, what do you suggest?

AutomatedTester:
Originally we had been discussing this mostly in the same approach as frames so you would do.

```python

shadow_root = driver.find_element(...)
driver.switch_to_shadow_root(shadow_root)
driver.find_element(....).click()

```

[...]

PR #7905, opened 2019-12-31
 Add new locator type for Shadow Root elements

Jiwari:
### Description
Adds support to locate elements that have Shadow Roots.  
Adds new annotation `@FindShadowBy`, which can be used to locate and elements that are inside a Shadow Root.  
Adds a helper class `ShadowElementFinder`, which can be used to verify if an element has a Shadow Root attached to it.

### Motivation and Context
Right now Selenium doesn't support PageObject annotations in elements that have Shadow Root.  
#### The current approach:
```java
@FindBy(css = "foo")
WebElement fieldWithShadowRoot;

public WebElement getSomeShadow() {
  JavascriptExecutor js = (JavascriptExecutor) driver;
  WebElement shadowElement = (WebElement) js.executeScript("arguments[0].shadowRoot", fieldWithShadowRoot);
  WebElement elementInsideShadowRoot = shadowElement.findElement(By.css("innerShadowSelector"));
  return elementInsideShadowRoot;
}
```

#### New implementation allows:
```java
@FindShadowBy({@FindBy(css = "foo"),  @FindBy(css = "innerShadowSelector")})
WebElement element;
```
As in `@FindBy`, it also allows to retrieve lists:
```java
@FindShadowBy({@FindBy(css = "foo"),  @FindBy(css = "innerShadowSelector")})
List<WebElement> element;
```
If nested Shadow Roots are present, the same will work, the only thing that needs to be provided is the selector for each Shadow Root in the correct order that they are nested:
```html
<div id="shadow1">
  #shadow-root
    <div id="shadow2">
      #shadow-root
        <div id="insideShadow2">
        </div>
    </div>
</div>
```
```java
@FindShadowBy({@FindBy(css = "#shadow1"),  
    @FindBy(css = "#shadow2"), 
    @FindBy(css= "#insideShadow2")})
WebElement element;
```

#### Notes
The elements returned by the last `@FindBy` won't have their shadow roots extracted automatically, this is to allow the user to use methods such as `getAttribute`, which would throw an error if inside the shadow root. If the user wants to extract it, `ShadowElementFinder` can be used to do so.  
This also means that providing just one locator to`@FindShadowBy` will yield the same result as using `@FindBy`, which is not a concern.

Once inside a Shadow Root, it's not possible to use xpath selector.  
I didn't add any validations to `FindShadowBy` as this could limit the usage of the chain selectors, and Selenium already throws an exception if the user tries to use an invalid selector type.

### Types of changes

### Checklist

### DIFF
--- a/java/client/test/org/openqa/selenium/ShadowElementFinderTest.java
+++ b/java/client/test/org/openqa/selenium/ShadowElementFinderTest.java
@@ -1,7 +1,7 @@
 package org.openqa.selenium;
 
+import org.assertj.core.api.ThrowableAssert.ThrowingCallable;
 import org.junit.Test;
-import org.openqa.selenium.testing.Ignore;
 import org.openqa.selenium.testing.JUnit4TestBase;
 
 import java.util.Arrays;
@@ -9,19 +9,11 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.openqa.selenium.testing.drivers.Browser.FIREFOX;
-import static org.openqa.selenium.testing.drivers.Browser.MARIONETTE;
+import static org.assertj.core.api.Assertions.*;
 
-/**
- * Shadow Root is not supported by Firefox.
- * 
- */
 public class ShadowElementFinderTest extends JUnit4TestBase {
 
   @Test
-  @Ignore(FIREFOX)
-  @Ignore(MARIONETTE)
   public void testShadowElementExists() {
     driver.get(pages.shadowElementsPage);
 
@@ -32,8 +24,6 @@ public class ShadowElementFinderTest extends JUnit4TestBase {
   }
 
   @Test
-  @Ignore(FIREFOX)
-  @Ignore(MARIONETTE)
   public void testShadowElementDoesNotExist() {
     driver.get(pages.shadowElementsPage);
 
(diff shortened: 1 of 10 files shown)


java/client/test/org/openqa/selenium/ShadowElementFinderTest.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.openqa.selenium.testing.drivers.Browser.FIREFOX;
+import static org.openqa.selenium.testing.drivers.Browser.MARIONETTE;
+
+/**
+ * Shadow Root is not supported by Firefox.
```
dstekanov: Shadow-DOM is supported in Firefox https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot#Browser_compatibility

But calling "element.shadowRoot" will return the map of different attributes, so you cannot cast it to WebElement but must directly search inside it by "shadowRoot.querySelector()"

Jiwari: Makes sense, and it's possible.  
But to get the selector out of `By`, I would have to use a `toString()` and remove the prefix `By.cssSelector: ` manually, which may cause issues later if `toString()` is changed to something else.  
A better option would be to add a method to `By` class to get only the selector.  
I will make some changes

Jiwari: @dstekanov Please take a look on the new implementation.

dstekanov:
Hi @Jiwari, 
there is an original issue for WebDriver standard. You can have a look at this issue [1320](https://github.com/w3c/webdriver/pull/1320). I am not a decision-maker, but I am afraid that your changes might not be approved by the owners.

diemol:
Closing this to consolidate on #5869

PR #7956, opened 2020-01-22
Make ChromiumOptions base class vendor-neutral

bwalderman:
### Description

This change removes usages of the vendor-specific terms "goog" and "chrome" from the ChromiumOptions base class. It makes the ChromiumOptions class abstract so that derived classes are required to fill these names in appropriately.

### Motivation and Context

Needed to support vendor-prefixed capabilities for EdgeDriver, which uses "ms" as the prefix instead of "goog". Previously, ChromiumDriver.CreateDevToolsSession would use the "goog"-branded capabilities by default, which meant that EdgeDriver.CreateDevToolsSession() would not work.

### Types of changes

### Checklist

### DIFF
--- a/dotnet/src/webdriver/Chrome/ChromeOptions.cs
+++ b/dotnet/src/webdriver/Chrome/ChromeOptions.cs
@@ -56,16 +56,26 @@ namespace OpenQA.Selenium.Chrome
         {
         }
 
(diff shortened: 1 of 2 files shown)


jimevans:
Merged in cfc2ae85. Thank you for your contribution!

PR #7982, opened 2020-01-30
Use Edge (Chromium) for .NET EdgeDriver by default

bwalderman:
### Description
The purpose of this change is the update EdgeDriver, EdgeOptions, and EdgeDriverService to be Chromium-first to ensure they are always driving the most up-to-date version of Edge going forward. The following changes have been made:
- Remove the isLegacy constructor parameter from EdgeOptions and EdgeDriverService.
- Remove legacy-only properties from EdgeDriverService.
- Create a new EdgeLegacyDriverService class that works like the old EdgeDriverService did before Chromium support was introduced.
- Split EdgeOptions into EdgeOptions and EgeLegacyOptions. EdgeOptions has only Chromium capabilities.
- Add constructor variants to EdgeDriver that accept valid combinations of Options/Service or creates reasonable defaults if these are not provided.

### Motivation and Context
The goal is to make driving the current/future Edge (Chromium) _easy_, and to make driving legacy Edge (EdgeHTML) _possible_, in that order.

Ideally this means that minimal changes are needed to upgrade to the new Edge. Existing client code should just work in most cases. Staying on old Edge should be as simple as creating an EdgeLegacyDriverService/Options. All other existing EdgeDriver code should continue to work as-is.

If/when we decide to remove Edge legacy support from Selenium, it will be as simple as removing the EdgeLegacyDriverService/Options classes and this can be done without breaking clients that have migrated to Edge Chromium.

### Types of changes

### Checklist

### DIFF
--- a/dotnet/test/common/Environment/DriverFactory.cs
+++ b/dotnet/test/common/Environment/DriverFactory.cs
@@ -89,9 +89,15 @@ namespace OpenQA.Selenium.Environment
                 options = GetDriverOptions<InternetExplorerOptions>(driverType, driverOptions);
                 service = CreateService<InternetExplorerDriverService>(driverType);
             }
-            else if (typeof(EdgeDriver).IsAssignableFrom(driverType))
+            else if (typeof(LegacyEdgeDriver).IsAssignableFrom(driverType))
             {
                 browser = Browser.EdgeLegacy;
+                options = GetDriverOptions<EdgeLegacyOptions>(driverType, driverOptions);
+                service = CreateService<EdgeLegacyDriverService>(driverType);
+            }
+            else if (typeof(EdgeDriver).IsAssignableFrom(driverType))
+            {
+                browser = Browser.Edge;
                 options = GetDriverOptions<EdgeOptions>(driverType, driverOptions);
                 service = CreateService<EdgeDriverService>(driverType);
             }


bwalderman:
I have EdgeLegacyDriverService deriving from EdgeDriverService (a ChromiumDriverService) which is kinda awkward. It means that EdgeLegacyDriverService exposes Chromium properties and methods even though these won't do anything.

This seemed necessary because EdgeLegacyDriverService and EdgeDriverService needed to have a common base class so they can both be passed to EdgeOptions, but EdgeDriverService needs to be a ChromiumDriverService. The alternative was to have EdgeOptions accept a DriverService, but this seemed like a bad idea since it would mean callers could pass in any service, not just Edge ones.

bwalderman:
After discussing with @jimevans, I was able to clean up the classes a bit. There are now separate EdgeLegacyOptions/EdgeOptions and EdgeLegacyDriverService/EdgeDriverService classes. There is still just one EdgeDriver to support both legacy and modern users and it has various constructor overrides that ensure legacy and non-legacy classes are not mixed.

bwalderman:
Abandoning this pull request. After discussion in https://github.com/SeleniumHQ/selenium/pull/7977 we've decided not to flip on Chromium by default yet. I'll create a separate PR to update the .NET bindings to use the new ms:edgeChromium capability for switching engines.

Issue #7916, opened 2020-01-08
Deadlock when running xunit selenium tests with parallelism


macpak:
## To Reproduce

```
using System;
using OpenQA.Selenium.Chrome;
using Xunit;

[assembly:CollectionBehavior(MaxParallelThreads = 2)]

namespace Selenium4Freeze
{
    public class TestClass
    {
        [Fact]
        public void Test()
        {
            var driver = new ChromeDriver(Environment.CurrentDirectory, new ChromeOptions());
            driver.Url = "http://www.google.com";
            driver.Navigate();
        }
    }
    
    public class TestClass2
    {
        [Fact]
        public void Test()
        {
            var driver = new ChromeDriver(Environment.CurrentDirectory, new ChromeOptions());
            driver.Url = "http://www.amazon.com";
            driver.Navigate();
        }
    }
}
```

Detailed steps to reproduce the behavior:

- Run all tests
- Observe deadlock, neither google or amazon pages are opened.
- Turn off parallelism (comment out [assembly:CollectionBehavior(MaxParallelThreads = 2)])
- Observe tests now pass

## Expected behavior

When tests are run, both of them pass. 

## Test script or set of commands reproducing this issue

Please run the code above

## Environment

OS: Windows 10
Browser: Chrome
Browser version: Version 79.0.3945.117 
Browser Driver version: 79.0.3945.3600
Language Bindings version: C# 4.0.0-alpha03, .net core 2.2

macpak:
I believe the deadlock is the result of the change from https://github.com/SeleniumHQ/selenium/commit/cf01bed7f1eab9a392d3e24549c9510d66c12e2f#diff-bc8a75c5cb22ca86093a1bbced41a6ee and the change to MakeHttpRequest which is now asynchronous. I will try to provide a fix.

jimevans:
@macpak A PR would be welcome, but the PR must not (1) remove the use of the `System.Net.Http.HttpClient` house, nor (2) propagate the `async` keyword to the public API.

diemol:
I see the linked PR was merged, therefore closing this issue.

PR #7987, opened 2020-02-03
Avoid deadlock when tests are run on XUnit

macpak:
The fix solves a problem with deadlock on Selenium 4.x. The requirement is that we cannot change HttpClient to something else + we cannot make other methods async, that's why I believe the only possible way to solve it is to use TaskFactory workaround. 

### Description
This fix solves #7916 

### Motivation and Context
It's required because on XUnit, we get a deadlock (when parallelism is used). 
Similar case is discussed here: https://github.com/xunit/xunit/issues/88

### Types of changes

### Checklist

### DIFF
--- a/dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
+++ b/dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
@@ -146,11 +146,10 @@ namespace OpenQA.Selenium.Remote
             HttpResponseInfo responseInfo = null;
             try
             {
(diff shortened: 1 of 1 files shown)


AutomatedTester:
@jimevans I think this patch meets your criteria, can you double check and approve?

jimevans:
Merged as c9ed1e2. Thank you for the contribution.

NicolasDorier:
this fix does not work ^

PR #8032, opened 2020-02-17
[dotnet] Fixes binding spec compliance for ErrorResponse

### LINKED ISSUE #3683: Selenium doesn't display the stacktrace from geckodriver exceptions, because the key is cased differently
## Meta -
OS:  Linux (Ubuntu 14.04)
Selenium Version:  
3.3.1
Browser:  
Firefox / geckodriver 0.15.0

## Expected Behavior -
If geckodriver throws an exception with a stacktrace, selenium should display that stacktrace in the message for the WebDriverException it generates.

## Actual Behavior -
The message is empty, even if a stacktrace is present.

It seems the selenium code looks for the key 'stackTrace' in an exception JSON passed from a webdriver. See https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/errorhandler.py#L173 However geckodriver uses 'stacktrace', all lower case, as per the W3 spec: https://www.w3.org/TR/webdriver/

I haven't checked every single one, but other languages appear to do the same thing. Would a PR checking for both keys be accepted? If so, is it legit to just patch one language's code, or does this cause more problems than it solves? Thanks.

jordan-mace:

Implements handling for 'stacktrace' to comply with the [W3 spec](https://w3c.github.io/webdriver/#errors).
Fixes #3683

### Description
Adds an OR clause to the if statement checking whether the stacktrace key exists. Adds handling to correctly set the stackTraceArray object depending on which key exists. Supporting both implementations was proposed [here](https://github.com/SeleniumHQ/selenium/issues/3683#issuecomment-308230232)

### Motivation and Context
Long-standing issue. All other bindings appear to have been fixed.

### Types of changes

### Checklist

### DIFF
--- a/javascript/node/selenium-webdriver/lib/webdriver.js
+++ b/javascript/node/selenium-webdriver/lib/webdriver.js
@@ -1188,7 +1188,8 @@ class Options {
    *
    * @param {string} name The name of the cookie to retrieve.
    * @return {!Promise<?Options.Cookie>} A promise that will be resolved
-   *     with the named cookie, or `null` if there is no such cookie.
+   *     with the named cookie
+   * @throws {error.NoSuchCookieError} if there is no such cookie.
    */
   async getCookie(name) {
     try {


jordanm-mel:
@jimevans would you be able to review when you get a chance? (I can imagine you are pretty busy at the moment with 4.0). Support for `stackTrace` is in the Java bindings according to [this comment](https://github.com/SeleniumHQ/selenium/issues/3683#issuecomment-322595319) so I'm keeping both `stacktrace` and `stackTrace` for consistency.

jimevans:
Merged in f683dd. Thank you for the contribution!

Issue #6292, opened 2018-08-16
Options.getCookie should return null if cookie not found.


AllenFang:
### OS:  
OSX 10.13
### Selenium Version:  
4.0.0-alpha.1
### Browser:  
Firefox developer edition

### Browser Version:  
62.0b15

## Expected Behavior -
If cookie is not exists, I should get null just like official document said.
https://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/lib/webdriver_exports_Options.html

* Retrieves the cookie with the given name. Returns null if there is no such cookie. The cookie will be returned as a JSON object as described by the WebDriver wire protocol.*

## Actual Behavior -
I got an error: 

```
    NoSuchCookieError: No cookie with name sbet

      at Object.throwDecodedError (node_modules/selenium-webdriver/lib/error.js:550:15)
      at parseHttpResponse (node_modules/selenium-webdriver/lib/http.js:542:13)
      at Executor.execute (node_modules/selenium-webdriver/lib/http.js:468:26)
```
## Steps to reproduce -

I use Javascript and `selenium-webdriver@4.0.0-alpha.1`

1. Initialize Driver
2. call await driver.manage().getCookie('XXX')
3. I should got a null in step 2 but I got NoSuchCookieError.

lmtierney:
The `w3c` spec says to throw the error you are getting, so that's behaving correctly. It looks like the `getCookie` command needs to add the `NoSuchCookie` error to the erros it catches here https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/lib/webdriver.js#L1207

AllenFang:
@lmtierney yes, I also checked the source code. So you mean the `getCookie` command need to return `null` instead of throw `NoSuchCookie` error?  If yes, I probably can help to fix it.

lmtierney:
That's up to the maintainer of the nodejs code. If it's to be nil, then the code needs updating. If it's supposed to throw an error then the documentation needs updating. @jleyba

shs96c:
@Harsha509, thank you for fixing this!

PR #8062, opened 2020-02-27
Modify: (doc) getCookie throws 'NoSuchCookieError' and !null 

harsha509:
### Description
Updating document for getCookie

As per [webdriver documentation](https://www.w3.org/TR/webdriver/#dfn-no-such-cookie), getCookie should throw **"NoSuchCookieError"** if named cookie is not found. 

In the documentation it is mentioned **"null"** will be returned if there is no named cookie in current browsing context.

This resolves issue https://github.com/SeleniumHQ/selenium/issues/6292

### Types of changes

### Checklist


shs96c:
Thank you for sending this! I'll merge it in now.

PR #8087, opened 2020-03-04
[dotnet] use a capability to switch engines for EdgeOptions/EdgeDriverService.

bwalderman:
### Description
Based on the discussion in https://github.com/SeleniumHQ/selenium/pull/7977 we've decided to use a new "ms:edgeChromium" boolean capability key for EdgeDriver which tells the driver service whether to launch Edge Legacy or Chromium. This change updates the EdgeOptions class, replacing the previous "isLegacy" constructor param with a property that can be used to set the capability. The various EdgeDriver constructor overloads have also been modified to ensure the EdgeDriverService and EdgeOptions agreee about which flavor of Edge to use. Before this change, the user had to create and manage their own EdgeDriverService in order to use Edge Chromium. Now they can just do:

```csharp
var driver = new EdgeDriver(new EdgeOptions() { UseChromium = true });
``` 

### Motivation and Context
Using a capability key allows EdgeOptions to work with a RemoteWebDriver. A .NET test can now configure the EdgeOptions for either Legacy or Chromium, and the remote Selenium server can make use of the capability to launch the correct Edge driver service. Previously, the isLegacy property on the EdgeOptions controlled how the EdgeOptions were serialized, but the serialized capabilities didn't carry any information about what flavor of Edge the client wants to use. This meant that the client could only safely send EdgeOptions to a remote server if they knew that a compatible Edge service was already running on that server.

### Types of changes

### Checklist

### DIFF
--- a/dotnet/src/webdriver/Edge/EdgeDriverService.cs
+++ b/dotnet/src/webdriver/Edge/EdgeDriverService.cs
@@ -213,7 +213,7 @@ namespace OpenQA.Selenium.Edge
         /// <returns>A EdgeDriverService using a random port.</returns>
         public static EdgeDriverService CreateChromiumService(string driverPath)
         {
-            return CreateDefaultServiceFromOptions(driverPath, EdgeDriverServiceFileName(false), new EdgeOptions() { UseChromium = true });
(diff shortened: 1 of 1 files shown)


jimevans:
Merged in 13d830b. Thanks for the contribution!

PR #8096, opened 2020-03-09
[py] Use a capability to switch engines for Edge

bwalderman:
### Description
Based on the discussion in #7977 we've decided to use a new "ms:edgeChromium" boolean capability key for EdgeDriver which tells the driver service whether to launch Edge Legacy or Chromium. This change updates the Edge .Options class to use this capability in place of the previous is_legacy constructor parameter.

The change also includes some cleanup in the Edge and Chromium drivers. The 'verbose' flag is now used all the time instead of being legacy-only. The duplicate page_load_strategy getter/setter has been removed since the base ChromiumDriver class takes care of this. The Chromium driver no longer takes the deprecated 'executable_path' parameter. The subclasses will always provide a service object, so this param shouldn't be needed. This shouldn't be a breaking change since ChromiumDriver is fairly new and isn't present in a stable release yet.

Lastly, added bazel test targets for Edge and Edge Chromium.

### Motivation and Context
Using a capability key allows EdgeOptions to work with a RemoteWebDriver. A python client can now configure the EdgeOptions for either Legacy or Chromium, and the remote Selenium server can make use of the capability to launch the correct Edge driver service. Previously, the is_legacy property on the EdgeOptions controlled how the EdgeOptions were serialized, but the serialized capabilities didn't carry any information about what flavor of Edge the client wants to use. This meant that the client could only safely send EdgeOptions to a remote server if they knew that a compatible Edge service was already running on that server.

### Types of changes

### Checklist

### DIFF
--- a/py/selenium/webdriver/common/actions/action_builder.py
+++ b/py/selenium/webdriver/common/actions/action_builder.py
@@ -38,7 +38,7 @@ class ActionBuilder(object):
         try:
             idx = self.devices.index(name)
             return self.devices[idx]
-        except:
+        except Exception:
             pass
 
     @property
--- a/py/selenium/webdriver/common/service.py
+++ b/py/selenium/webdriver/common/service.py
@@ -56,7 +56,7 @@ class Service(object):
         return "http://%s" % utils.join_host_port('localhost', self.port)
 
     def command_line_args(self):
-        raise NotImplemented("This method needs to be implemented in a sub class")
+        raise NotImplementedError("This method needs to be implemented in a sub class")
 
     def start(self):
         """
--- a/py/selenium/webdriver/common/timeouts.py
+++ b/py/selenium/webdriver/common/timeouts.py
@@ -15,18 +15,19 @@
 # specific language governing permissions and limitations
 # under the License.
 
+
 class Timeouts(object):
-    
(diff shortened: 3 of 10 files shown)


py/selenium/webdriver/edge/webdriver.py
```diff
@@ -15,51 +15,57 @@
...
+           capabilities only, such as "proxy" or "loggingPref".
          - service_log_path - Deprecated: Where to log information from the driver.
          - keep_alive - Whether to configure EdgeRemoteConnection to use HTTP keep-alive.
-         - service_args - Deprecated: List of args to pass to the driver service
-         - is_legacy: Whether to use MicrosoftWebDriver.exe (legacy) or MSEdgeDriver.exe (chromium-based). Defaults to True.
+         - verbose - whether to set verbose logging in the service.
          """
-        if not is_legacy:
+        if executable_path != 'MicrosoftWebDriver.exe':
+            warnings.warn('executable_path has been deprecated, please pass in a Service object',
```
AutomatedTester: There is no import for `warnings`. You can see the [travis log](https://travis-ci.com/SeleniumHQ/selenium/jobs/296255724) of the failure

PR #8105, opened 2020-03-12
[java] Use a capability to switch engines for Edge

bwalderman:
### Description
Based on the discussion in #7977 we've decided to use a new "ms:edgeChromium" boolean capability key for EdgeDriver which tells the driver service whether to launch Edge Legacy or Chromium. This change updates the EdgeOptions class to check this capability in addition to the webdriver.edge.edgehtml system property when deciding which service to launch. Calling code can override the system property using EdgeOptions.setChromium().

Also updated the DriverInfo for EdgeHTML and Chromium Edge to score based on this new capability.

### Motivation and Context
Using a capability key allows EdgeOptions to work with a RemoteWebDriver. A client can now configure the EdgeOptions for either Legacy or Chromium, and the remote Selenium server can make use of the capability to launch the correct Edge driver service.

### Types of changes

### Checklist


bwalderman:
@shs96c, do you have time to take a look at this?

fenchu:
What is the release plan for this, after msedgedriver>=80, ChromeOptions() have been deprecated, and we are forced to use EdgeOptions(), but for me using python with selenium 4.0.0.a3 (which is the latest) the bindings relates to legacy edge and driver just fails to start.

PR #8110, opened 2020-03-14
Fixed null pointer exception

cjayswal:
When remote Webdriver initialized using no-arg constructor from sub class, capabilities is null, results in null pointer exception in executeScript().

### Types of changes

### Checklist

### DIFF
--- a/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java
+++ b/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java
@@ -274,7 +274,7 @@ public class RemoteWebDriver implements WebDriver, JavascriptExecutor,
 
   @Override
   public Capabilities getCapabilities() {
-    return capabilities;
+    return this.capabilities;
   }
 
   @Override
--- a/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java
+++ b/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java
@@ -116,7 +116,7 @@ public class RemoteWebDriver implements WebDriver, JavascriptExecutor,
 
   // For cglib
   protected RemoteWebDriver() {
-    init(new ImmutableCapabilities());
+    this.capabilities=init(new ImmutableCapabilities());
   }
 
   public RemoteWebDriver(Capabilities capabilities) {
@@ -126,7 +126,7 @@ public class RemoteWebDriver implements WebDriver, JavascriptExecutor,
   public RemoteWebDriver(CommandExecutor executor, Capabilities capabilities) {
     this.executor = executor;
 
-    init(capabilities);
+    capabilities=init(capabilities);
 
     if (executor instanceof NeedsLocalLogs) {
       ((NeedsLocalLogs)executor).setLocalLogs(localLogs);
@@ -154,8 +154,8 @@ public class RemoteWebDriver implements WebDriver, JavascriptExecutor,
     return new RemoteWebDriverBuilder();
   }
 
-  private void init(Capabilities capabilities) {
-    this.capabilities = capabilities == null ? new ImmutableCapabilities() : capabilities;
+  private Capabilities init(Capabilities capabilities) {
+    capabilities = capabilities == null ? new ImmutableCapabilities() : capabilities;
 
     logger.addHandler(LoggingHandler.getInstance());
 
@@ -186,6 +186,8 @@ public class RemoteWebDriver implements WebDriver, JavascriptExecutor,
         logTypesToInclude);
     localLogs = LocalLogs.getCombinedLogsHolder(clientLogs, performanceLogger);
     remoteLogs = new RemoteLogs(executeMethod, localLogs);
+    
+    return capabilities;
   }
 
   /**
@@ -274,6 +276,9 @@ public class RemoteWebDriver implements WebDriver, JavascriptExecutor,
 
   @Override
   public Capabilities getCapabilities() {
+    if(capabilities == null){
+      return new ImmutableCapabilities();
+    }
     return capabilities;
   }
 
--- a/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java
+++ b/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java
@@ -274,7 +274,7 @@ public class RemoteWebDriver implements WebDriver, JavascriptExecutor,
 
   @Override
   public Capabilities getCapabilities() {
-    return this.capabilities;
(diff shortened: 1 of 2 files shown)


barancev:
Please add a test to show what issue was fixed by this change.

cjayswal:
@barancev, I have added test that ensures capabilities field not null when no arg constructor called. 
When RemoteDriver no-arg constructor called it doesn't initialize empty capabilities and throwing NPE when  `executeScript` or `isJavaScriptEnabled` method called because of null capabilities.

Furthermore,  only `isJavaScriptEnabled` not using getter and direct accessing private field, which is also problem as not considering if sub class of RemoteWebDriver has capabilities initialized.

barancev:
To be honest, this no-arg constructor is totally useless, it's just a dirty hack to make this class cglib-compatible. Let us think a bit, it can be better to remove this constructor instead of fixing the NPE issue.

cjayswal:
It is not useless. Don't remove it. What will be the benefit of removing it? If it is removed it will break frameworks that use it for (example qaf, quantum) and will impact to may users/organizations.

barancev:
We never delete anything immediately. If we decide to remove something we deprecate it and give users a long period to adopt to the change. And it's not a decision yet, just one of the possible solutions to discuss.

Can you explain how (and why) do these frameworks use the no-arg constructor?

cjayswal:
The framework extends `RemoteWebDriver` and for local execution it uses no-arg constructor to create extended driver object [from another local driver session ](https://github.com/qmetry/qaf/blob/master/src/com/qmetry/qaf/automation/ui/webdriver/QAFExtendedWebDriver.java#L108)and later sets command executor and session id.

barancev:
OK, I see. I don't like how it's implemented, but it's a piece of legacy and should be treated carefully. Lets' apply this change then.

java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java
```diff
@@ -274,7 +274,7 @@ protected void setCommandExecutor(CommandExecutor executor) {
 
   @Override
   public Capabilities getCapabilities() {
-    return capabilities;
+    return this.capabilities;
```
shs96c: This `this` isn't needed. Generally, we only use the `this` keyword where it makes code unambiguous.

java/client/test/org/openqa/selenium/remote/RemoteWebDriverUnitTest.java
```diff
@@ -1065,7 +1065,12 @@ public void canHandleElementCssPropertyCommand() throws IOException {
         new CommandPayload(DriverCommand.GET_ELEMENT_VALUE_OF_CSS_PROPERTY,
                            ImmutableMap.of("id", element.getId(), "propertyName", "color")));
   }
-
+  
+  @Test
+  public void noArgConstuctorEmptyCapabilitiesTest() throws IOException {
+    RemoteWebDriver driver = new RemoteWebDriver();
```
shs96c, line 1071: Perhaps override this and ensure that the capabilities are set properly, as per the bug? You can declare a new class inline:

```java
public void noArgConstructorCreatesEmptyCapabilities() {
  RemoteWebDriver driver = new RemoteWebDriver() {}; // anonymous subclass
  assertThat(driver.getCapabilities()).isEqualTo(new ImmutableCapabilities());
}
```

shs96c (approved):
A couple of minor points, but once fixed, this looks great. Thank you very much for digging in, figuring out what needs to be done, and fixing it!

java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java
```diff
@@ -155,7 +155,7 @@ public static RemoteWebDriverBuilder builder() {
   }
 
   private void init(Capabilities capabilities) {
-    capabilities = capabilities == null ? new ImmutableCapabilities() : capabilities;
+    this.capabilities = capabilities == null ? new ImmutableCapabilities() : capabilities;
```
barancev: Assigning capabilities to the field here is not the best idea. Semantically this field contains the capabilities returned by the browser after session start. So here we sould better amend capabilities and return them from `init` method to pass later to `startSession` method (that will assign the capabilities returned by the browser to the field).

barancev: And instead of assigning capabilities to the field prior to a session start we can modify `getCapabilities` method to return an empty `ImmutableCapabilities` object if the field is equal to `null` to avoid `NullPointerException` here.

barancev (changes requested):

java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java
cjayswal: Don't you think it will be overhead of null check each time `getCapabilities` called? Ideally `init` accepts desired capabilities and capabilities get overridden with [actual capabilities](https://github.com/SeleniumHQ/selenium/blob/30f9ffab92839cdd454b65404b50ee94f887e99e/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java#L219) in [startSession](https://github.com/SeleniumHQ/selenium/blob/30f9ffab92839cdd454b65404b50ee94f887e99e/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java#L255). So assigning empty `ImmutableCapabilities` capabilities in `init` should be fine. Whoever had written code initially was really well thought, it was just minor slip of `this` keyword. If we don't want to set `ImmutableCapabilities` in `init`, better is to add setter method instead of null check on each get call.

barancev: People will hardly call `getCapabilities` often in a non-initialized driver, and this field will be set in `startSession` to the capabilities returned by the browser after initialization

cjayswal: Execute script command call will call capabilities each time before executing execute script command.

barancev: I want to reiterate that this field will be initialized after a call to `startSession` method, it will contain the capabilities returned by the browser:
https://github.com/SeleniumHQ/selenium/blob/30f9ffab92839cdd454b65404b50ee94f887e99e/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java#L255

barancev:
Applied as commit d16ecd2a32a1270b9b31f69358ee964e572b7d57, thank you for the contribution!

PR #8118, opened 2020-03-18
Add: Adding sameSite cookie attribute for rb bindings

harsha509:
### Description
Adding sameSite cookie attribute for rb bindings and added tests

### Types of changes

### Checklist

### DIFF
--- a/rb/lib/selenium/webdriver/common/manager.rb
+++ b/rb/lib/selenium/webdriver/common/manager.rb
@@ -36,7 +36,7 @@ module Selenium
       # @option opts [String] :value A value
       # @option opts [String] :path ('/') A path
       # @option opts [String] :secure (false) A boolean
-      # @option opts [String] :sameSite (Strict or Lax) currently supported only in chrome 80+ versions
+      # @option opts [String] :same_site (Strict or Lax) currently supported only in chrome 80+ versions
       # @option opts [Time,DateTime,Numeric,nil] :expires (nil) Expiry date, either as a Time, DateTime, or seconds since epoch.
       #
       # @raise [ArgumentError] if :name or :value is not specified


rb/lib/selenium/webdriver/common/manager.rb
```diff
@@ -47,7 +47,7 @@ def add_cookie(opts = {})
 
         opts[:path] ||= '/'
         opts[:secure] ||= false
-
+        opts[:sameSite]
```
p0deje: We don't need this line, do we?

rb/spec/integration/selenium/webdriver/manager_spec.rb
```diff
@@ -64,6 +64,19 @@ module WebDriver
           expect(cookies.first[:value]).to eq('bar')
         end
 
+        it 'should add sameSite cookie with attribute Strict', only: {browser: :chrome} do
+          driver.navigate.to url_for('xhtmlTest.html')
+          driver.manage.add_cookie name: 'foo', value: 'bar', sameSite: 'Strict'
```
p0deje: I believe we should use snake case for paramater to make the code more Rubyish:

```suggestion
          driver.manage.add_cookie name: 'foo', value: 'bar', same_site: 'Strict'
```

rb/spec/integration/selenium/webdriver/manager_spec.rb
```diff
@@ -64,6 +64,19 @@ module WebDriver
...
+
+          expect(driver.manage.cookie_named('foo')[:sameSite]).to eq('Strict')
+        end
+
+        it 'should add sameSite cookie with attribute Lax', only: {browser: :chrome} do
+          driver.navigate.to url_for('xhtmlTest.html')
+          driver.manage.add_cookie name: 'foo', value: 'bar', sameSite: 'Lax'
+          expect(driver.manage.cookie_named('foo')[:sameSite]).to eq('Lax')
```
p0deje: Likewise, I believe we need to use snake case 
```suggestion
          expect(driver.manage.cookie_named('foo')[:same_site]).to eq('Lax')
```

rb/lib/selenium/webdriver/common/manager.rb
```diff
@@ -47,7 +47,7 @@ def add_cookie(opts = {})
```
p0deje: Can you update the documentation of the `add_cookie` method to include a new parameter and mention it's Chrome-only?

rb/lib/selenium/webdriver/common/manager.rb
Harsha509: Hi @p0deje ,

Thanks for the review. No, it's not required feature works removing this. 
Updating the code

rb/lib/selenium/webdriver/common/manager.rb
Harsha509: Updated as `  # @option opts [String] :sameSite (Strict or Lax) currently supported only in chrome 80+ versions`

rb/spec/integration/selenium/webdriver/manager_spec.rb
Harsha509: Updated!

PR #8150, opened 2020-03-31
Py highlight

sriup:
This flash method will highlight the element and set back the original element style. It will high light the element n times (default value is 5). The only new library imported is `time` to make sure there is delay between each time the script highlight the element.

Added line# 24, 775 - 813

### Types of changes

- [ X ] New feature (non-breaking change which adds functionality)

### Checklist

- [ ] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/master/CONTRIBUTING.md) document.
- [ X ] I have updated the documentation accordingly.
- [ X ] I have added tests to cover my changes.


AutomatedTester:
Thanks for the PR and the effort you put into it.

Unfortunately we won't be accepting it as it doesn't fit in with APIs that we tend to create. This PR would be perfect as a project that has Selenium as a dependency

PR #8233, opened 2020-04-24
try to fix infinite loop issue

LazyMisha:

### Description

I faced up with a problem related to an infinite loop in this code snippet. I use Serenity for autotests, in turn, the Serenity uses Selenium. After a test fell, code is continuing to run. In my case the code snippet '((WrapsElement) arg).getWrappedElement()' returns Object WebElementFacadeStub type every stage of the loop.

If my solution does not fit, it's worth adding some check to this loop for breaking.
### Types of changes

### Checklist

### DIFF
--- a/java/client/src/org/openqa/selenium/os/OsProcess.java
+++ b/java/client/src/org/openqa/selenium/os/OsProcess.java
@@ -18,11 +18,9 @@
 package org.openqa.selenium.os;
 
 import static com.google.common.base.Preconditions.checkNotNull;
-import static com.google.common.collect.ImmutableMap.copyOf;
+import static java.util.Collections.unmodifiableMap;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.openqa.selenium.Platform.WINDOWS;
-
-import com.google.common.annotations.VisibleForTesting;
 
 import org.apache.commons.exec.DaemonExecutor;
 import org.apache.commons.exec.DefaultExecuteResultHandler;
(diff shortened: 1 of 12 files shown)


barancev:
May I ask you for a reproduction case first?

Thу proposed change assumes that the element was wrapped once. This is not true in general, that's why we have a loop here.

Yes, this can cause an infinite loop if an element says that it wraps itself, or an element A says that it wraps B and B says that it wraps A. We don't attempt to detect cyclic references like this. Can you please explain how did it appear in your code?

LazyMisha:
Sure, I'll try to create a simple project to reproduce this one and send it. Thanks for responding

LazyMisha:
Hi, I've created a simple project to reproduce the bug.
Please, take a look:
https://github.com/LazyMisha/serenity-selenium-bug

If an object could be wrapped more than one time, we can add some counter, for example after 5 times loop execution, invoke a break.

barancev:
We can't stop unwrapping after a fixed number of attempt because we have to find the "original" element. Yes, we can throw an exception if unwrapping does not stop after some max number of attempts, but it will hardly resolve your issue.

If Serenity (or some other tool) creates circular references then this tool should be fixed.
Do you have a reproduction example using vanilla Selenium?

LazyMisha:
Unfortunately no, but I can try to find how to reproducing with vanilla Selenium. 

But I think, you are right, the problem is with a tool that uses Selenium, Serenity in our case. I'm going to inform the Serenity team of the issue.

Thanks.

barancev:
I'll close this PR because it's not applicable in its current form. Please raise an issue if you get a reproduction scenario using Selenium only. You're welcome to our IRC/Slack chat if you need any additional information that would help you report or fix issue in Serenity.

PR #8236, opened 2020-04-24
[java] Allow users to extend ExpectedConditions

alb-i986:
### Description
Loosen the strict constraint on the constructor, from private to protected.

### Motivation and Context
..so that users can have their own static class with their own custom conditions, which extends ours.
As a result, by importing their own class, they can get access to both their and our ExpectedCondition's.

As an example from a notable project, please see [JUnit's Assert](https://github.com/junit-team/junit4/blob/50a285d3ce69b4556ac46d8633f6beb4527b4679/src/main/java/org/junit/Assert.java#L29): it has the same feature.

### Types of changes

### Checklist


barancev:
What's the purpose of extending a class that contains a set of static methods only? Instances of this class would be totally useless. Just create another class and a user will import statically some methods from `ExpectedConditions` and other methods from your class.

PR #8239, opened 2020-04-24
case insensitive check for the content-length header

### LINKED ISSUE #7986: [IEDriver] Error: No property named 'capabilities' or 'desiredCapabilities' found in new session request body. if Content-Length is not set
## 🐛 Bug Report

If `Content-Length` header is not set in request then IEDriver is not able to [process ](https://github.com/SeleniumHQ/selenium/blob/400c32a457b883dbb20066314788b97508547b99/cpp/webdriver-server/server.cc#L323)the request. The other problem is that the [check ](https://github.com/SeleniumHQ/selenium/blob/400c32a457b883dbb20066314788b97508547b99/cpp/webdriver-server/server.cc#L305)for `Content-Lenght` is case sensitive. This is not inline with [HTTP specification](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2).

I am happy to contribute the improvements if you agree with fixing this. But my C++ experience is very limited :(

## Environment

OS: Windows 10
Browser: Internet Explorer
Browser version:  11
Browser Driver version: 3.14.0.0 (64-bit)
Language Bindings version: NodeJS

reidj32:
### Description
According to RFC 2616 and 7230, headers names are case insensitive.

When creating a new session, the webdriver server is checking the
request for a case sensitive Content-Type header. This causes the body
to be empty on session creation and will result in the IEDriverServer
complaining about missing capabilities.

This change will search for the Content-Type in a case insensitive
manner before capturing the value of the header.

Fixes #7986

### Motivation and Context
Many modern JavaScript HTTP libraries set their headers to lower case before doing a request. Because IEDriverServer is looking for the `Content-Type` in a case sensitive manner, this prevents those tools from testing on IE.

### Types of changes

### Checklist


reidj32:
I was unsure of where to add tests for this. If someone would be willing to point me in to right direction, I will happily add some if it is needed. I do believe many of the existing tests will cover this change when creating a new webdriver session.

jimevans:
Merged as 2dfe468. Thank you for your contribution!

PR #8274, opened 2020-05-05
[JS] renaming `addArguments` to `addBrowserCommandSwitches` according to functionality

harsha509:
### Description
1. `addBrowserCommandSwitches` will be more precise naming to add command line switches. 
2. Added test to check capability `silent`. In JS bindings, IEDriver command line arguments (--silent, --log-level, ---log-file,--host, --extract-path) should be passed via capability.set

### Motivation and Context
`addArguments` is confusing for adding browserCommandLineSwitches

### Types of changes

### Checklist

### DIFF
--- a/javascript/node/selenium-webdriver/CHANGES.md
+++ b/javascript/node/selenium-webdriver/CHANGES.md
@@ -6,7 +6,8 @@
     - setEdgePath
     - setEdgeChromium
     - setScrollBehavior
-    - Supports adding multiple command-line switches using options.addArguments()
+    - Supports adding multiple command-line switches using options.addBrowserCommandSwitches()
+    
 * Add relative locators
 
 ### API Changes


Harsha509:
IENew

AutomatedTester:
A few things:

- Why are there so many commits unrelated in this PR?
- Why are we making a breaking change to the API? How are we going to support people when their tests break? Are people using this API?

Harsha509:
Hi @AutomatedTester ,

Thanks for the reviewing the PR.

`Why are there so many commits unrelated in this PR?`
>>>
This is my fault, I did pulled/fetched changes without committing and pushing the current changes into branch. I'll make sure to squash the multiple commits to keep the log clean and minimal in the PR.

`Why are we making a breaking change to the API?`
>>>
I've looked into the other language bindings for ie options to specify browserCommandLine arguments. So thought `addBrowserCommandLineSwitches` is very clear for the user when they want to add commandLineSwitches. Because `addArguments` doesn't convey to new users that it adds command line parameters. 

Also I wanted to maintain same naming conventions for API method, so that if a user is switching btw multiple language bindings will have the same API method name.

ex: addCommandLineSwitches  (Java)  == addArguments(JS)

`How are we going to support people when their tests break?`

Yes, It will break the existing scripts that uses this capability, I now realise the direct change is not a good thing to do. So, can we add a deprecation warning "`addArguments` will be renamed to addBrowserCommandLineSwitches". Please suggest !

`Are people using this API?`

Am not sure on this, I've done a bit research on the questions that people asked in various channels on  ie browserCommand-line-switches so it was pretty like 20%-25% people use this.

Regards,
Harsha.

AutomatedTester:
Ok, so we can't just change the name of it. I suggest adding the new API and getting the old API to give deprecation warnings and call the newly named API. Once that is done I will review and we can see about getting it merged.

Harsha509:
> Ok, so we can't just change the name of it. I suggest adding the new API and getting the old API to give deprecation warnings and call the newly named API. Once that is done I will review and we can see about getting it merged.

Sure @AutomatedTester , I will close this PR and will create a new one.

PR #8282, opened 2020-05-07
Fix grid/config:SmallTests icw. --explicit_java_test_deps

ulfjack:
### Description
Add the missing JUnit dependency to the Java test target.

### Motivation and Context
It is required to build Selenium with `--explicit_java_test_deps`.

### Types of changes

### Checklist


ulfjack:
There are a few more targets failing with that flag because they're missing a junit dependency. I can either update this PR or send an additional one.

shs96c:
Good catch. I'm a little surprised this hasn't caused trouble before, but thank you for fixing it!

Issue #7274, opened 2019-06-09
SELENIUM 3.141.59 with FF 67.0.1 and GECKO driver 0.24


atchuth23:
## 🐛 Bug Report

Runnin selenium as Windows service via nssm and the remotewebdriver is intiated and the connection is refused.

## To Reproduce

Detailed steps to reproduce the behavior:

## Expected behavior

I want the remote session in the nssm to open a valida firefox session and execute the test.

## Test script or set of commands reproducing this issue

	    public static void main(String[] args) throws MalformedURLException, InterruptedException {
	    	System.setProperty("webdriver.gecko.driver","C:\\Users\\Administrator\\Downloads\\geckodriver-v0.24.0-win64\\geckodriver.exe");
	    	WebDriver driver = new FirefoxDriver();
	    	
	    	FirefoxOptions firefoxOptions = new FirefoxOptions();
	    	firefoxOptions.setAcceptInsecureCerts(true);
	    	firefoxOptions.setLogLevel(FirefoxDriverLogLevel.TRACE);
	    	FirefoxDriver firefoxDriver = new FirefoxDriver(firefoxOptions);
		    
	    	 driver = new RemoteWebDriver(new URL("http://localhost:5566"), firefoxOptions);
	    	    driver.navigate().to("http://www.yahoo.com");
	    		    	driver.get("https://www.google.com");
	    	driver.close();
	    	driver.quit();
	    }
        

## Environment

OS: <-- Windows 10 Windows 7 Windows server2012 -->
Browser: <- Firefox -->
Browser version: 
Browser Driver version: 
Language Bindings version:  
Selenium Grid version (if applicable):

atchuth23:
1560100055200	mozrunner::runner	INFO	Running command: "C:\\Program Files\\Mozilla Firefox\\firefox.exe" "-marionette" "-foreground" "-no-remote" "-profile" "C:\\Users\\ADMINI~1\\AppData\\Local\\Temp\\2\\rust_mozprofile.uEtkDBmlu5jv"
1560100057976	addons.webextension.screenshots@mozilla.org	WARN	Loading extension 'screenshots@mozilla.org': Reading manifest: Invalid extension permission: mozillaAddons
1560100057979	addons.webextension.screenshots@mozilla.org	WARN	Loading extension 'screenshots@mozilla.org': Reading manifest: Invalid extension permission: resource://pdf.js/
[...]

Issue #7127, opened 2019-04-22
Python Binding 4.0 Make WebElement and WebDriver class rely on ABC to register all subtypes


isaulv:
## 🚀 Feature Proposal

Add an Abstract Base Class to Remote.Webdriver and Remote.WebElement 

## Motivation

For most users of Python Selenium, they tend to have a straight forward use of the library, where they don't run into issues. However when it comes to using add on tools like Applitools Eyes, Appium, and even the current incarnation of the binding itself, there are times when the `isinstance` type checker will fail due to unforeseen uses of a WebElement. 

Examples are using EventFiringWebElement with an ActionChains.move_to command.   #6604 
The Applitools Eyes Python binding does not subclass WebElement, which makes it hard to integrate with EventFiringWebDriver, or other libraries that do isinstance checks.

The webdriver type itself hardly goes through checks itself, but a tool like Eyes, does do type checks on the driver to see if it is wrapped with the Eyes implementation.

## Example

If WebElement is an ABC, then the following code should _just work_.
```
from abc import ABC

class IWebElement(ABC):
    ... # some core methods are made abstract method.

class WebElement(IWebElement):
    ... # existing definition

# Eyes library
class EyesWebElement(object):
    ... # eyes specific code

IWebElement.register(EyesWebElement)
... # assume you have valid driver, element from a live session.
assert isinstance(EyesWebElement(driver, element), IWebElement)
```

The implementation of this should also include a scan of the code base for `type(object)` checks and replace them with `isinstance` or `issubclass` checks, unless we need exact type matching.

vhryhorchuk-amplify:
The same problem exists for `EventFiringWebElement`, as it doesn't extends `WebElement`.
We have next bug:
```
class invisibility_of_element_located(object):
    """ An Expectation for checking that an element is either invisible or not
    present on the DOM.

    locator used to find the element
    """
    def __init__(self, locator):
        self.target = locator

    def __call__(self, driver):
        try:
            target = self.target
            if not isinstance(target, WebElement):
                target = _find_element(driver, target)
            return _element_if_visible(target, False)
        except (NoSuchElementException, StaleElementReferenceException):
            return True
```

When we pass `EventFiringWebElement` to the expected condition `invisibility_of_element(myEventFiringWebElement)` it goes to:
```
            if not isinstance(target, WebElement):
                target = _find_element(driver, target)
```
And fails with
```
 def _find_element(driver, by):
[...]

PR #8348, opened 2020-05-26
Add Abstract Base Class to WebElement and WebDriver #7127

### LINKED ISSUE #7877: [Python] action_chains.py does not play nicely with EventFiringWebDriver
Various actions of the Actions capabilities (action_chains.py) do not play nicely with an EventFiringWebDriver implementation, for example:

```
    def move_to_element_with_offset(self, to_element, xoffset, yoffset):
        """
        Move the mouse by an offset of the specified element.
           Offsets are relative to the top-left corner of the element.

        :Args:
         - to_element: The WebElement to move to.
         - xoffset: X offset to move to.
         - yoffset: Y offset to move to.
        """
        if self._driver.w3c:
            self.w3c_actions.pointer_action.move_to(to_element, xoffset, yoffset)
            self.w3c_actions.key_action.pause()
        else:
            self._actions.append(
                lambda: self._driver.execute(Command.MOVE_TO, {
                    'element': to_element.id,
                    'xoffset': int(xoffset),
                    'yoffset': int(yoffset)}))
        return self
```

When using an EF wrapped driver, it will attempt such actions by default with a type of:

```
<selenium.webdriver.support.event_firing_webdriver.EventFiringWebElement object at 0x116b87c10>
(Pdb) type(ele)
<class 'selenium.webdriver.support.event_firing_webdriver.EventFiringWebElement'>
```

I propose we do an isinstance check for EventFiringWebElement here?  If this is deemed a valid bug and you think its a worthy fix or use case (or perhaps it is intentional by design) then please assign it to me and I will open a PR this weekend to address it.

as currently it will do the following:

```
AttributeError: move_to requires a WebElement
```

My proposed solution is to check if it is an instance of EventFiringWebElement and use .wrapped_element if so, thoughts?

I would like to contribute, so if its suitable - I will take this one.

isaulv:

### Description
Make `WebDriver` and `WebElement` classes use Abstract Base Classes so that library authors and EventFiringWebDriver can work with wrapped or custom derived classes.

### Motivation and Context
By being able to register types that are not direct subclasses of the WebElement/Webdriver classes, this allows library authors to write fancy wrappers that won't break with `isinstance` checks.

It also fixes #7877, #7274 

This code will work with Python 2 and Python 3.

### Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)

### Checklist

- [X ] All new and existing tests passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,7 @@ To list all available targets, you can append the `-T` flag:
 * `java` and `jar` on the PATH (make sure you use `java` executable from JDK but not JRE). 
   * To test this, try running the command `javac`. This command won't exist if you only have the JRE
   installed. If you're met with a list of command-line options, you're referencing the JDK properly.
+* [Bazel](https://docs.bazel.build/versions/master/install.html)
 * [Python](https://www.python.org/)
 * `python` on the PATH
 * [The Requests Library](http://python-requests.org) for Python: `pip install requests`
--- a/java/client/src/org/openqa/selenium/internal/Require.java
+++ b/java/client/src/org/openqa/selenium/internal/Require.java
@@ -98,7 +98,7 @@ public final class Require {
       return arg;
     }
 
-    public T instanceOf(Class cls) {
+    public T instanceOf(Class<?> cls) {
       if (arg == null) {
         throw new IllegalArgumentException(argName + " must be set");
       }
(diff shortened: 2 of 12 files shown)


Dude-X:
Right now the ABC doesn't enforce any particular attributes because we have to address the question, what is the essence of `WebElement` and `WebDriver`. But least by providing an ABC, we can give library and framework authors the ability to register their wrappers and custom types.

Also it would be nice if Python had contravariant type checking but that's a different discussion.

Dude-X:
This is my solution for #7127.

AutomatedTester:
It's a good start

PR #372, opened 2015-03-26
Optimizing ByChained#findElement

alb-i986:
- optimize complexity: instead of traversing the entire "tree"
  of partial WebElements found, only travers the left-most path
- in case the bys list is empty, throw an exception with a more meaningful message

Not sure if you like the

```
assert elem != null;
```

I seem to have some issues in running the tests. They are failing, even on master. I guess it must be due to my env (?).
OSX 10.10.2, Java 1.8.0_40
Also tried java 6 but it didnt seem to help

```
Compiling: //java/client/test/org/openqa/selenium:SmallTests as build/java/client/test/org/openqa/selenium/SmallTests.jar
mkdir -p build/java/client/test/org/openqa/selenium/SmallTests.jar_temp
Compiling 1 source file to /Users/alb/workspace/selenium/build/java/client/test/org/openqa/selenium/SmallTests.jar_temp
Building jar: /Users/alb/workspace/selenium/build/java/client/test/org/openqa/selenium/SmallTests.jar
rm -rf build/java/client/test/org/openqa/selenium/SmallTests.jar_temp
** Execute //java/client/test/org/openqa/selenium:SmallTests
** Execute //java/client/test/org/openqa/selenium:SmallTests:run
Testing: //java/client/test/org/openqa/selenium:SmallTests
mkdir -p build/test_logs
Listening for transport dt_socket at address: 5005
Running org.openqa.selenium.SmallTests
Exception in thread "Thread-20" org.openqa.selenium.WebDriverException: Thread safety error; this instance of WebDriver was constructed on thread main (id 1) and is being accessed by thread Thread-20 (id 35)This is not permitted and *will* cause undefined behaviour
Build info: version: '2.45.0', revision: '17dc63a', time: '2015-03-26 03:09:18'
System info: host: 'albs-MacBook-Air.local', ip: '192.168.0.25', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.10.2', java.version: '1.8.0_40'
Driver info: driver.version: unknown
    at org.openqa.selenium.support.ThreadGuard$WebDriverInvocationHandler.invoke(ThreadGuard.java:86)
    at com.sun.proxy.$Proxy22.findElement(Unknown Source)
    at org.openqa.selenium.support.ThreadGuardTest$1.run(ThreadGuardTest.java:44)
    at java.lang.Thread.run(Thread.java:745)
Tests run: 345, Failures: 0, Errors: 0, Time elapsed: 1.831 sec
[...]

PR #8370, opened 2020-05-30
Refactoring ByChained findElementS

alb-i986:
### Description
The original for-each was pretty obscure. Replacing it with an index-based for loop.

This addresses the doubts which arose in the review for #372:
https://github.com/SeleniumHQ/selenium/pull/372#discussion_r27221662

### Types of changes

- [x] Refactoring

### Checklist

### DIFF
--- a/java/client/src/org/openqa/selenium/support/pagefactory/ByChained.java
+++ b/java/client/src/org/openqa/selenium/support/pagefactory/ByChained.java
@@ -61,16 +61,14 @@ public class ByChained extends By implements Serializable {
       return new ArrayList<>();
     }
 
-    List<WebElement> elems = null;
-    for (By by : bys) {
+    List<WebElement> elems = bys[0].findElements(context);
+    for (int i = 1; i < bys.length; i++) {
+      if (elems.isEmpty()) {
+        break; // if any one of the bys finds no elements, then return no elements
+      }
       List<WebElement> newElems = new ArrayList<>();
-
-      if (elems == null) {
-        newElems.addAll(by.findElements(context));
-      } else {
-        for (WebElement elem : elems) {
-          newElems.addAll(elem.findElements(by));
-        }
+      for (WebElement elem : elems) {
+        newElems.addAll(elem.findElements(bys[i]));
       }
       elems = newElems;
     }
--- a/java/client/test/org/openqa/selenium/support/pagefactory/ByChainedTest.java
+++ b/java/client/test/org/openqa/selenium/support/pagefactory/ByChainedTest.java
@@ -17,20 +17,26 @@
 
 package org.openqa.selenium.support.pagefactory;
 
+import static java.util.Arrays.asList;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
-import java.util.ArrayList;
-import java.util.List;
 import org.junit.Test;
 import org.openqa.selenium.By;
 import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.SearchContext;
 import org.openqa.selenium.WebElement;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+
 public class ByChainedTest {
+
+  private static final List<WebElement> NO_ELEMENTS = Collections.emptyList();
 
   @Test
   public void findElementZeroBy() {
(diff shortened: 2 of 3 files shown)


java/client/src/org/openqa/selenium/support/pagefactory/ByChained.java
```diff
@@ -61,16 +61,14 @@ public WebElement findElement(SearchContext context) {
       return new ArrayList<>();
     }
 
-    List<WebElement> elems = null;
-    for (By by : bys) {
+    List<WebElement> elems = bys[0].findElements(context);
+    for (int i = 1; i < bys.length; i++) {
+      if (elems.isEmpty()) {
```
diemol, line 66: Isn't this changing the previous behavior? 
I see in the current code that even if `elems` is empty, the loop continues through the rest of the `bys` array.

java/client/test/org/openqa/selenium/support/pagefactory/ByChainedTest.java
```diff
@@ -17,10 +17,10 @@
 
 package org.openqa.selenium.support.pagefactory;
 
+import static java.util.Arrays.asList;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
diemol: Could we keep the explicit imports please?

diemol (changes requested):
I left a couple of comments @alb-i986, thank you.

If I understand better, this should be merged first and then we continue with #372?

java/client/src/org/openqa/selenium/support/pagefactory/ByChained.java
alb-i986: Hey @diemol, thanks for taking the time to look into this.

I understand your point. It looks like it behaves that way, but actually it doesn't. This is exactly the point of this PR.
BTW, the same concern was also pointed out in the other review (https://github.com/SeleniumHQ/selenium/pull/372#discussion_r27221662).

I purposefully added two unit tests as the first commit, *before* committing the refactoring.
As I wrote in the commit message:
> add unit tests to verify the assumption that in case the chain is broken,
> i.e. one of the bys in the chain yields no elements,
> then ByChained#findElements returns an empty list.
 
You may checkout 51c5bd690cb7783560d1accddae263d44e26bc4a and see they are passing.

As I pointed out in https://github.com/SeleniumHQ/selenium/pull/372#discussion_r34129844:

> The key is `elems = newElems;` at the end of each iteration, which overwrites the list that is iterated during each step. The first by that returns no elements will set elems to empty list; any subsequent iteration of the outer loop will do nothing, since the inner loop will not be executed.

diemol:
@alb-i986 apologies that this is taking ages, could you please solve the conflicts and I get to review it again?

alb-i986:
Here you go @diemol

java/client/src/org/openqa/selenium/support/pagefactory/ByChained.java
diemol: Thanks for clarifying.

java/client/test/org/openqa/selenium/support/pagefactory/ByChainedTest.java
```diff
@@ -247,6 +253,47 @@ public void findElementsTwoByEmptyChild() {
...
+    when(elem1.findElements(by2)).thenReturn(NO_ELEMENTS);
+    when(elem2.findElements(by2)).thenReturn(NO_ELEMENTS);
+
+    ByChained by = new ByChained(by1, by2, by3);
+
+    assertThat(by.findElements(driver)).isEmpty();
+    verify(elem1, never()).findElements(by3);
+    verify(elem2, never()).findElements(by3);
+  }
+
```
diemol, line 296: Is it just me or are these two tests not compiling?

`findElementsByName` does not exist and probably some `mockito` import is missing?

diemol (changes requested):

java/client/test/org/openqa/selenium/support/pagefactory/ByChainedTest.java
alb-i986: Sorry, that's possible.

The build is not working on my IDEA, and I don't know how to run the tests from the command line: `bazel test ...??`.

The documentation about building is scattered and outdated, it seems to me:
- some pieces are on the [README](https://github.com/SeleniumHQ/selenium/blob/trunk/README.md#common-build-targets)
- some on [this wiki page](https://github.com/SeleniumHQ/selenium/wiki/Build-Instructions), linked [from the README](https://github.com/SeleniumHQ/selenium/blob/trunk/README.md#useful-resources)
- in [Build Instructions](https://github.com/SeleniumHQ/selenium/wiki/Build-Instructions), some commands like `bazel test java/client/test/org/openqa/selenium/chrome` do not work anymore

barancev:
Applied as commits 31a702df59663d606bc4a027c5034c22419e6f4d and 894d238f235425054639327700ce5fac49584d92, thank you for the contribution.

PR #8373, opened 2020-05-31
[JS] Fixes path to bazel generated files

harsha509:
### Description
bazel-genfiles are not available in path looks like it should be `bazel-bin` 

### Types of changes

### Checklist


Harsha509:
Ahh looks like i need to dig more into this!

Azassinz:
did you solve this? I have the same problem, I cannot use selenium in my Vue.js app.
This is the message showing in the console:
 warning  in ./node_modules/selenium-webdriver/lib/http.js
Module not found: Error: Can't resolve '../../../../bazel-genfiles
Thanks!

AlbertoNava0307:
@Azassinz I have same problem, you could solve it?

PR #8387, opened 2020-06-04
Correct Net::HTTP::Persistent bug when using newer versions

DylanLacey:

### Description
Fix the integration with the Net HTTP Persistent Gem.

### Motivation and Context
The default Net::HTTP Library allows the end user to force SSL using the `use_ssl=` method.  Net::HTTP::Persistent doesn't in newer releases.  This causes errors to be thrown when using the Net::HTTP::Persistent shim.

This change moves SSL configuration into its own method, allowing the shim to handle this for different versions while keeping the default HTTP code clean.

### Types of changes

### Checklist


titusfortner:
Hah, just noticed that we didn't have a conversation about removing it, I just suggested it and didn't follow up: https://github.com/SeleniumHQ/selenium/pull/7065#issuecomment-477432000

DylanLacey:
Either of these are fine by me!  I didn't note the change to the standard HTTP client and since the Persistent shim was still present but not working...

Shall I leave this open for a week or so in case anyone else is watching the issue tracker with rabid interest for Persistent improvements and has an argument for keeping it?

p0deje:
I believe we should proceed with deprecating this class in favor of default HTTP client for 4.0.

titusfortner:
And deprecated: f6d831b848b9cbe63c41b55c61eb97ec23bbfb69

PR #8398, opened 2020-06-08
[java] change access modifier of RemoteWebElement.execute method

amitbhoraniya:
### Description
We have extended `RemoteWebElement` and added more capabilities by extending RemoteWebElement. Like Appium has extended RemoteWebElement and created MobileElement. We have overridden below method.

```
protected Response execute(String command, Map<String, ?> parameters) {}
```

Now these methods has been replaced by `CommandPayload` argument, these need to be a `protected` method to support override. Earlier overridden support was there in selenium version 2 and 3.

Previously I have created pull request https://github.com/SeleniumHQ/selenium/pull/8344 for RemoteWebDriver. Sorry at that time I didn't notice same issue with RemoteWebElement.

### Motivation and Context
This is blocker who have extended RemoteWebElement in their libraries. This change supports migration to selenium 4.
Even now other execute methods are `protected`, so I don't find any reason to make this method `default`.

### Types of changes

### Checklist

### DIFF
--- a/java/client/src/org/openqa/selenium/remote/RemoteWebElement.java
+++ b/java/client/src/org/openqa/selenium/remote/RemoteWebElement.java
@@ -232,7 +232,7 @@ public class RemoteWebElement implements WebElement, WrapsDriver, TakesScreensho
     return allElements;
   }
 
-  Response execute(CommandPayload payload) {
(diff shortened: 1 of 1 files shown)


AutomatedTester:
Thanks for the PR!

PR #8399, opened 2020-06-08
Python remote webdriver: make default command executor match docs (append /wd/hub)

tomka:
### Description

The command_executor argument is used to define a URL of a remote server or a custom remote connection. Most commonly, this is provided by the caller, but it comes also with a reasonable default value according to the docs: http://127.0.0.1:4444/wd/hub. The default value in the code however does in fact _not_ include the `/wd/hub` suffix, which is fixed by this commit.

### Motivation and Context

I am trying to fix a selenium test issue on Travis CI with Sauce Labs and started reading the remote web driver source code to better understand what's going on. I stumbled over the inconsistency between code and documentation and wanted to address this quickly. 

### Types of changes

While unlikely to break existing code, it does change a default parameter, which I would count as breaking change.

### Checklist

Let me know if this change requires a test (e.g. comparing a default initialized remote driver executor with the default). If changing the default value is the wrong approach, the docs should be changed.

### DIFF
--- a/py/selenium/webdriver/remote/webdriver.py
+++ b/py/selenium/webdriver/remote/webdriver.py
@@ -143,7 +143,7 @@ class WebDriver(BaseWebDriver):
 
     _web_element_cls = WebElement
 
-    def __init__(self, command_executor='http://127.0.0.1:4444',
+    def __init__(self, command_executor='http://127.0.0.1:4444/wd/hub',
                  desired_capabilities=None, browser_profile=None, proxy=None,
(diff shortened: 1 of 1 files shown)


AutomatedTester:
Thank you for your effort in making this PR. The original code is correct here. When a remote end receives anything for the `/wd/hub` endpoint from older client bindings it will move them to correct endpoint like that it is used in the Python client. 

With Selenium 4 will need its documentation updated as the code is ahead here.

tomka:
Thanks for the clarification!

PR #8408, opened 2020-06-09
Ruby - stop using an obsolete method of Net::HTTP

masakazutakewaka:
## Description
`Net::HTTP::Proxy` is an obsolete method, so replaced it with `Net::HTTP.new` as the doc says -> https://docs.ruby-lang.org/en/2.7.0/Net/HTTP.html#method-c-Proxy

### Motivation and Context
To keep the code up to date.

### Types of changes

### Checklist

### DIFF
--- a/py/rules_python_any_version_wheel.patch
+++ b/py/rules_python_any_version_wheel.patch
@@ -34,3 +34,36 @@ index 9cd6534..7802d42 100644
      ),
      "version": attr.string(
          mandatory = True,
+diff --git experimental/rules_python/wheelmaker.py experimental/rules_python/wheelmaker.py
+index 6be5d38..34b5cb0 100644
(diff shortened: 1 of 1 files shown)


rb/lib/selenium/server.rb
```diff
@@ -120,16 +120,16 @@ def latest
         end
       end
 
-      def net_http
+      def net_http_start(address, &block)
```
masakazutakewaka, line 123: Should I write about this in the CHANGE.log as this method is exposed to public?

p0deje: It might worth mentioning, but frankly, I don't think this was ever intended to be a public API.

masakazutakewaka:
Hi @p0deje, it's green now.

p0deje:
Thank you!

PR #8412, opened 2020-06-10
[java] Actions javadoc: #perform vs #build

alb-i986:
### Description
Slightly improving the documentation.
Pointing out that it’s #perform what the user needs to call at the end of the method chain to actually perform the actions.

### Motivation and Context
It would be natural for one who is familiar with the Builder pattern, and it's the first time that he/she uses the Actions class (exploring the API with method completion), to come up with something like this:

```java
new Actions(driver)
  .this()
  .that()
  [..]
  .build();
```

And expect that it will work.

Then, when they run their test and find that something didn't work, it's not going to be easy to figure out what went wrong. More specifically, to trace it back to the call to `build()`.

This PR is trying to solve that by adding a warning in the javadoc.

Although I would prefer the build() method to behave as perform() (and deprecate perform() afterwards).

This comes from a personal experience, where I tried to use the Actions class after some time (a few years) I hadn't used it and I forgot that you need to call perform() and not build().

### Types of changes

- [x] Javadoc improvement

### Checklist

### DIFF
--- a/java/client/src/org/openqa/selenium/interactions/Actions.java
+++ b/java/client/src/org/openqa/selenium/interactions/Actions.java
@@ -45,6 +45,8 @@ import java.util.logging.Logger;
  * <p>
  * Implements the builder pattern: Builds a CompositeAction containing all actions specified by the
  * method calls.
+ * <p>
+ * Call {@link #perform()} at the end of the method chain to actually perform the actions.
  */
 public class Actions {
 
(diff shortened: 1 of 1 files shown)


shs96c (approved):
Fair point, well made. :)

shs96c:
Thank you for clarifying things. It's appreciated!

PR #8421, opened 2020-06-14
[java] WebDriverWait Overloading FluentWait#until to take a Duration param

alb-i986:
### Description
Besides taking a fixed timeout from the constructor, this change allows the user to choose a custom timeout when calling the method `until`, overriding the timeout set initially in the constructor.

This makes `WebDriverWait` more memory efficient, as users will be able to instantiate only once, and choose a potentially different timeout at every call to `wait#until`.

Previously one needed to create one instance of `WebDriverWait` **for each** different timeout value, thus potentially having tens or more, depending on the usage, of `WebDriverWait` instances around.

### Motivation and Context
This is especially useful when using dependency injection.
Previously, if you were injecting a `WebDriverWait` around in your classes, you were stuck with the timeout you had set initially. If you needed to wait for a different number of seconds you had to create a new WebDriverWait instance on the fly, by hand.

This change makes it much more simple.
You can define once and for all a WebDriverWait with a default timeout, and have it injected everywhere. But the client code will still be able to choose the timeout at call time.

Here is an example for PicoContainer.

```java
    private static FactoryInjector<WebDriverWait> webDriverWaitInjector() {
        return new FactoryInjector<WebDriverWait>() {
            @Override
            public WebDriverWait getComponentInstance(PicoContainer container, Type into) {
                return new WebDriverWait(container.getComponent(WebDriver.class), 30); // here we are setting a default timeout of 30s
            }
        };
    }
```

Here is an example of code *using* the injected WebDriverWait instance:

```java
public class MyPageObject {
  private final WebDriverWait wait;
  public PageObject(WebDriverWait wait) {
    this.wait = wait;
  }

  public void doSomething() {
    wait.until(visibilityOfElement(By.id("email")); // uses the default timeout
    [..]
  }

  public void doSomethingSlow() {
    wait.until(visibilityOfElement(By.id("email"), 60); // customises the timeout for a slow operation
    [..]
  }
}
```

Please note. The method I'm overloading comes from the interface `Wait`, which here I'm not touching.

### Types of changes

### Checklist

### DIFF
--- a/java/client/src/org/openqa/selenium/support/ui/FluentWait.java
+++ b/java/client/src/org/openqa/selenium/support/ui/FluentWait.java
@@ -211,7 +211,7 @@ public class FluentWait<T> implements Wait<T> {
   public <V> V until(Function<? super T, V> isTrue, Duration timeout) {
     try {
       return CompletableFuture.supplyAsync(checkConditionInLoop(isTrue))
-          .get(timeout.get(ChronoUnit.MILLIS), TimeUnit.MILLISECONDS);
+          .get(timeout.get(ChronoUnit.NANOS), TimeUnit.NANOSECONDS);
     } catch (ExecutionException cause) {
       if (cause.getCause() instanceof RuntimeException) {
         throw (RuntimeException) cause.getCause();


shs96c:
The idea of the `Wait` was precisely to avoid this kind of passing through of random durations: the idea is to encourage developers to think about the kinds of timeouts they need, and just use those (I use `wait` and `shortWait`)

In addition, it was envisioned that these would be injected using some kind of CI framework (or the `new` keyword)

Because of this, although I really appreciate the work I'm going to close this PR. I hope you understand.

alb-i986:
Yeah, that's a good idea. In fact I use the same pattern (the wait/shortWait) in my Selenium projects, but there are still cases where you need to a custom timeout, maybe only a few, maybe more, who knows.
That was my point.

PR #8437, opened 2020-06-18
fix(nodejs): include auth in same domain redirects

alfonso-presa:
### Description
Authorization information is not being sent in same domain redirects.

### Motivation and Context
TLDR; we need to send redirects in session creation with the authorization sent in the original post command.

We have long running commands (specially session creation) that may last for several minutes in our hub. Our hub is in AWS infrastructure and also under a bunch of reverse proxies and we don't (can't) control request timeouts in all those pieces (i.e. NLB has a hardcoded 350s timeout for requests).

Our hub is behind also a basic auth security proxy, so we need to provider that information in all requests.

We have implemented a redirect logic (sending 303 redirects is commands are not resolved in for example 120s) so that we can extend the command duration.

Additionally, to detect if the client dropped the connection and clear the browser "almost" immediatelly after it's provisioned, we always send a redirect before giving the session (so that if he doesn't come back for the session values we know that he has dropped the waitting)

### Types of changes

### Checklist


alfonso-presa:
I've seen the CI fails are also failing in some jobs in master also.... I also don't see how they could be related with my changes. Even though if I can be of any help in fixing them let me know.

diemol:
@Harsha509 could you please help me to review this one?

harsha509:
Hi @diemol ,

PR looks good to me! we can merge it.

Regards,
Harsha.

diemol:
Thanks for reviewing @Harsha509!

diemol (approved):
Thank you @alfonso-presa!

alfonso-presa:
Thanks you both @diemol and @Harsha509!

PR #8440, opened 2020-06-19
Js Cdp support

raju249:

### Types of changes

### Checklist

- [ ] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/master/CONTRIBUTING.md) document.


javascript/webdriver/BUILD.bazel
```diff
@@ -66,3 +66,12 @@ closure_test_suite(
         "//javascript/webdriver/atoms/inject:deps",
     ],
 )
+
+genrule(
+    name="create-cdp-srcs",
+    outs=["devtools/types"],
+    cmd = "node $(location protocol-dts-generator.js) $@",
+    tools=[
+        "/javascript/webdriver/devtools/generator/protocol-dts-generator.js",
```
shs96c: The actual tool is `node` here. The script should be included in the `srcs`. I suspect you also want the `browser_protocol.json` and `js_protocol.json` in the sources too. The trick is to imagine that only the things listed as inputs to the rule are actually available.

Since the `protocol-dts-generator.js` file is local to this build file, you _may_ need to prefix it with a `:`, to indicate that it's a target.

shs96c (commented):
What are the `.d.js` files? Outputs from running something in place?

I'd strongly recommend _not_ checking in new copies of `browser_protocol.json` and `js_protocol.json`. They're already in the tree, and multiple copies make updating and keeping things in sync a lot harder. If we need to move them to a better location, we can do that.

AutomatedTester:
I have a follow up for the python code that will delete this duplicate so it’s fine for this PR

raju249:
@shs96c 

> What are the .d.js files? Outputs from running something in place?

Yes, when we run the `protocol-dts-generator.js` which is compiled from [devtools-protocol project](https://github.com/ChromeDevTools/devtools-protocol/blob/master/scripts/protocol-dts-generator.ts) we get the files, although I can rename to just `<name>.js` and drop the `d`. 🤔

javascript/node/selenium-webdriver/BUILD.bazel
```diff
@@ -88,3 +88,25 @@ genrule(
     ],
     cmd = "cp $(locations //:license) $(@D)",
 )
+
+nodejs_binary(
```
AutomatedTester, line 95: We should delete the json files and then copy them into the right place at build time. You can see how I did it in the python bazel code.

AutomatedTester (changes requested):

AutomatedTester (approved):

AutomatedTester:
This is failing on Travis CI because it can't find `ws`. Can you check.

javascript/node/selenium-webdriver/BUILD.bazel
```diff
@@ -88,3 +89,41 @@ genrule(
...
+)
+
+copy_file(
+    name = "js_protocol",
+    src = "//common/devtools:js_protocol.json",
+    out = "devtools/generator/js_protocol.json",
+)
+
+genrule(
+    name="create-cdp-srcs",
```
AutomatedTester, line 117: this rule needs to be called by `selenium-webdriver` and the `jasmine_node_test` rule

javascript/node/selenium-webdriver/BUILD.bazel
```diff
@@ -88,3 +89,41 @@ genrule(
     ],
```
AutomatedTester, line 91: on line 71 or so you need to add `"@npm//ws",`

AutomatedTester (changes requested):

AutomatedTester (approved):

PR #8453, opened 2020-06-23
Corrected the documentation in line #68

sriup:
Corrected the documentation for INTRODUCE_FLAKINESS_BY_IGNORING_SECURITY_DOMAIN.
Number of lines changed # 1
Need Test Cases for this change: No


diemol (commented):
@sridharUpputuri thank you for this PR, but the documentation is clear
```
   * Capability that defines to ignore browser
   * protected mode settings during starting by IEDriverServer.
```
adding the `or not` will make it confusing and the meaning will be lost

PR #8470, opened 2020-06-30
[tracing] Add span event APIs and wrapper

pujagani:
### Description

Updated existing Span interface and OpenTelemetrySpan to add APIs for creating events in the span. 
The APIs allow to add attributes to the event in the span. The attributes are stored using a Map. To enable use of single Map  while creating event attributes for different data types of attributes (String , long, double, boolean), EventAttribute interface and data types specific implementations are added. 

### Motivation and Context

The current tracing wrapper around OpenTelemetry does not provide APIs to add events to the spans. This PR provides the APIs to instrument events in the spans with and without attributes.  Events are crucial to drill down and triage an issue. Events help in documenting a moment at that timestamp within a span. 

### Types of changes

### Checklist


java/client/src/org/openqa/selenium/remote/tracing/Span.java
```diff
@@ -25,6 +25,9 @@
   Span setAttribute(String key, Number value);
   Span setAttribute(String key, String value);
 
+  Span addEvent(String name);
+  Span addEvent(String name, long timestamp);
```
shs96c: We don't need this additional method.

java/client/src/org/openqa/selenium/remote/tracing/empty/NullSpan.java
```diff
@@ -49,6 +49,20 @@ public Span setAttribute(String key, String value) {
...
   }
 
+  @Override
+  public Span addEvent(String name){
+    Require.nonNull("Name", name);
+    return this;
+  }
+
+  @Override
+  public Span addEvent(String name, long timestamp){
```
shs96c: And so we don't need this.

java/client/src/org/openqa/selenium/remote/tracing/opentelemetry/OpenTelemetrySpan.java
```diff
@@ -77,6 +77,22 @@ public Span setAttribute(String key, String value) {
...
 
+  @Override
+  public Span addEvent(String name){
+    Require.nonNull("Name", name);
+    span.addEvent(name);
+    return this;
+  }
+
+  @Override
+  public Span addEvent(String name, long timestamp){
```
shs96c: Or this.

java/client/test/org/openqa/selenium/remote/tracing/opentelemetry/TracerTest.java
```diff
@@ -58,9 +58,12 @@
   public void shouldBeAbleToCreateATracer() {
     List<SpanData> allSpans = new ArrayList<>();
     Tracer tracer = createTracer(allSpans);
+    long timeStamp = 1593493828L;
 
     try (Span span = tracer.getCurrentContext().createSpan("parent")) {
       span.setAttribute("cheese", "gouda");
+      span.addEvent("Grating cheese");
```
shs96c: Break out tests for events into their own tests rather than placing them in other ones. That makes it easier for us to figure out where problems lie and to do a TDD-driven implementation over new APIs.

shs96c (changes requested):
This looks great. Thank you. I don't think we need the variant with the timestamp since we never need to futz with that in our own code, so I'd suggest removing it, but this looks good.

java/client/test/org/openqa/selenium/remote/tracing/opentelemetry/TracerTest.java
pujagani: Thank you Simon. I have addressed the changes requested in the next commit.

java/client/src/org/openqa/selenium/remote/tracing/BUILD.bazel
```diff
@@ -26,6 +26,7 @@ java_library(
     deps = [
         "//java/client/src/org/openqa/selenium:core",
         "//java/client/src/org/openqa/selenium/remote/http",
+        artifact("com.google.guava:guava"),
```
diemol: We are trying to move away from Guava, I haven't checked, but there should be another way to get the primitives.

java/client/src/org/openqa/selenium/remote/tracing/Span.java
```diff
@@ -25,6 +27,9 @@
   Span setAttribute(String key, Number value);
   Span setAttribute(String key, String value);
 
+  Span addEvent(String name);
+  Span addEvent(String name, Map<String,EventAttributeValue> attributeMap);
```
diemol: Maybe I am missing context, but do we need to add `EventAttributeValue` and the wrappers? I see that [AttributeValue](https://javadoc.io/static/io.opentelemetry/opentelemetry-api/0.5.0/io/opentelemetry/common/AttributeValue.html) extends `Object` and adds some methods around it.
Could we have an approach where we do not need this extra classes and wrappers?

diemol (changes requested):
I added a couple of comments.

java/client/src/org/openqa/selenium/remote/tracing/BUILD.bazel
pujagani: Sure, I will look into it. I just referred to how we were checking primitives previously in the tracing bit and used the same. Now I think it will be a good idea to update the existing primitive check as well to not use Guava.

java/client/src/org/openqa/selenium/remote/tracing/Span.java
pujagani: I understand your thought here. I did consider using a Map<String,Object> but I thought it was not the best practise, because only String, Boolean,Number are allowed as attributes and wouldn't want the API caller to pass Objects that are not those. So thought of having wrappers  in addition to the wrappers helping with OpenTelemetry compatibility issues. What do you think would be a better way @diemol ?

java/client/src/org/openqa/selenium/remote/tracing/BUILD.bazel
diemol: Actually, please disregard this one, I saw it now in different places already.

subratkrgouda:
SvWh ok if p 2y
Ep Aap
..$CD s zzzar za xxx,,
!÷€}•za. AW£`{¢[%##3
As j
{™=28#*Hznz z1n We
P mai b be swa

diemol (approved):

Issue #8020, opened 2020-02-11
Support GraphQL for querying the Grid


shs96c:
This should make it easier for us to build a nice front end by not having to guess the shape of all the data we're going to try and collect.

shs96c:
https://github.com/graphql-java/graphql-java seems like a good starting point for this. It has minimal dependencies, and we should be able to hook in our own JSON library without too much effort.

shucon:
Working on this.
cc @AutomatedTester

AutomatedTester:
All work is done on this issue for now

PR #8474, opened 2020-07-01
Add SessionMap data

shucon:
Partial-PR: #8020

### Testing Instructions
* Build the Grid by `bazel build selenium-server-standalone`
* Start the Selenium Grid by `java -jar selenium_server_deploy.jar standalone`
* Start querying using curl commands like : `curl -X POST -H "Content-Type: application/json" --data '{"query": "{ grid { nodes { id, uri, status, maxSession, capabilities, sessions {id, capabilities, startTime} }, uri, totalSlots, usedSlots, sessionCount } }"}' -s http://localhost:4444/graphql`
* Open new sessions using a Java file opening RemoteDrivers
* Start querying again using curl commands and notice the difference in the queries due to new sessions.
### Types of changes

### Checklist


java/server/src/org/openqa/selenium/grid/data/DistributorStatus.java
```diff
@@ -117,6 +117,10 @@ public boolean isUp() {
       return up;
     }
 
+    public boolean isDocker() {
```
shs96c: Prefer a human-readable string rather than querying specific technologies. How would I indicate a session is running on BrowserStack? Or some custom thing?

java/server/src/org/openqa/selenium/grid/graphql/Grid.java
```diff
@@ -53,6 +58,14 @@ public URI getUri() {
       .collect(ImmutableList.toImmutableList());
   }
 
+//  public List<Session> getSessions() {
```
shs96c: This can probably be removed :)

java/server/src/org/openqa/selenium/grid/sessionmap/SessionMap.java
```diff
@@ -81,6 +81,10 @@
 
   public abstract void remove(SessionId id);
 
+  public int getCount() {
```
shs96c: This is not the right approach. The `Distributor` maintains a model of the current state of the Grid. That model already contains the information about every active session. We don't need to modify `SessionMap` to expose it further.

java/server/src/org/openqa/selenium/grid/sessionmap/local/LocalSessionMap.java
```diff
@@ -98,6 +98,11 @@ public Session get(SessionId id) {
     }
   }
 
+  @Override
+  public int getCount() {
```
shs96c: This method isn't needed (see above)

shs96c (changes requested):
Why not use the information that the `Distributor` already holds?

java/server/src/org/openqa/selenium/grid/data/DistributorStatus.java
```diff
@@ -203,7 +210,7 @@ private static NodeSummary fromJson(JsonInput input) {
 
       input.endObject();
 
-      return new NodeSummary(nodeId, uri, up, maxSessionCount, stereotypes, used);
+      return new NodeSummary(nodeId, uri, up, maxSessionCount, stereotypes, used, null);
```
shs96c: Pass in an empty set rather than `null` and you'll avoid null pointer exceptions later

java/server/src/org/openqa/selenium/grid/graphql/BUILD.bazel
```diff
@@ -16,6 +16,7 @@ java_library(
         "//java/client/src/org/openqa/selenium/remote/http",
         "//java/server/src/org/openqa/selenium/grid/data",
         "//java/server/src/org/openqa/selenium/grid/distributor",
+        "//java/server/src/org/openqa/selenium/grid/sessionmap",
```
shs96c: Is this needed?

java/server/src/org/openqa/selenium/grid/graphql/Grid.java
```diff
@@ -19,8 +19,10 @@
 
 import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
+
 import org.openqa.selenium.grid.data.DistributorStatus;
 import org.openqa.selenium.grid.distributor.Distributor;
+import org.openqa.selenium.grid.sessionmap.SessionMap;
```
shs96c: I think that this is an unused import

java/server/src/org/openqa/selenium/grid/graphql/GridData.java
```diff
@@ -19,7 +19,10 @@
 
 import graphql.schema.DataFetcher;
 import graphql.schema.DataFetchingEnvironment;
+
+import org.openqa.selenium.SessionNotCreatedException;
 import org.openqa.selenium.grid.distributor.Distributor;
+import org.openqa.selenium.grid.sessionmap.SessionMap;
```
shs96c: You can safely revert changes to this file.

java/server/src/org/openqa/selenium/grid/graphql/Node.java
```diff
@@ -33,18 +38,21 @@
...
               int maxSession,
-              Map<Capabilities, Integer> capabilities) {
+              Map<Capabilities, Integer> capabilities,
+              Set<Session> activeSessions) {
     this.id = Require.nonNull("Node id", id);
     this.uri = Require.nonNull("Node uri", uri);
     this.isUp = isUp;
     this.maxSession = Require.nonNull("Node maxSession", maxSession);
     this.capabilities = Require.nonNull("Node capabilities", JSON.toJson(capabilities));
+    this.activeSessions = activeSessions;
```
shs96c: This should never be null. `Require` that.

java/server/src/org/openqa/selenium/grid/graphql/Node.java
```diff
@@ -58,6 +66,11 @@ public URI getUri() {
   public int getMaxSession() {
     return maxSession;
   }
+  
+  public List<String> getActiveSessionIDs() {
+      return activeSessions.stream().map(session -> session.getId().toString())
```
shs96c: And this is why `activeSessions` should never be null

shs96c (changes requested):

java/server/src/org/openqa/selenium/grid/graphql/Node.java
```diff
@@ -33,18 +38,29 @@
...
+  private final Set<Session> activeSessions;
 
 
   public Node(UUID id,
               URI uri,
               boolean isUp,
               int maxSession,
-              Map<Capabilities, Integer> capabilities) {
+              Map<Capabilities, Integer> capabilities,
+              Set<Session> activeSessions) {
```
shs96c, line 51: The list of active sessions is internal state to the node.

java/server/src/org/openqa/selenium/grid/graphql/Node.java
```diff
@@ -33,18 +38,29 @@
...
     this.id = Require.nonNull("Node id", id);
     this.uri = Require.nonNull("Node uri", uri);
     this.isUp = isUp;
     this.maxSession = Require.nonNull("Node maxSession", maxSession);
     this.capabilities = Require.nonNull("Node capabilities", JSON.toJson(capabilities));
+    this.activeSessions = Require.nonNull("Active sessions", activeSessions);
+  }
+
+  public List<org.openqa.selenium.grid.graphql.Session> getSessions() {
```
shs96c, line 60: The Node should have no concept of graphql

shs96c (dismissed):

java/server/src/org/openqa/selenium/grid/graphql/Node.java
shucon: This is equivalent to `public List<Node> getNodes()` . Here the node is `org.openqa.selenium.grid.graphql.Node`. See: https://github.com/SeleniumHQ/selenium/blob/373159c40fbaab2c91074967b3da7e5c755f95d5/java/server/src/org/openqa/selenium/grid/graphql/Grid.java#L46
It was written this way because there are two classes with the name `Session`

java/server/src/org/openqa/selenium/grid/graphql/Node.java
shucon: Closing, had a chat on Slack regarding this.

diemol: What was the conclusion of that Slack conversation?

shucon: There was some confusion regarding which node file it was.

java/server/src/org/openqa/selenium/grid/data/DistributorStatus.java
```diff
@@ -89,20 +90,23 @@ private static DistributorStatus fromJson(JsonInput input) {
     private final int maxSessionCount;
     private final Map<Capabilities, Integer> stereotypes;
     private final Map<Capabilities, Integer> used;
+    private final Set<Session> activeSessions;
```
diemol, line 93: Shouldn't this field be added to the [`toJson`](https://github.com/SeleniumHQ/selenium/pull/8474/files#diff-9e8cb039b41324ada6234f83a6307d89R151) method as well?

java/server/src/org/openqa/selenium/grid/data/Session.java
```diff
@@ -39,10 +40,12 @@
   private final SessionId id;
   private final URI uri;
   private final Capabilities capabilities;
+  private final LocalDateTime startTime;
 
   public Session(SessionId id, URI uri, Capabilities capabilities) {
     this.id = Require.nonNull("Session ID", id);
     this.uri = Require.nonNull("Where the session is running", uri);
+    this.startTime = Require.nonNull("Session Start time", LocalDateTime.now());
```
diemol: I think there is no need to use the `Required.nonNull` here.

java/server/src/org/openqa/selenium/grid/graphql/Grid.java
```diff
@@ -49,10 +49,18 @@ public URI getUri() {
...
                                summary.isUp(),
                                summary.getMaxSessionCount(),
-                               summary.getStereotypes()))
+                               summary.getStereotypes(),
+                               summary.getActiveSessions()))
       .collect(ImmutableList.toImmutableList());
   }
 
+  public int getSessionCount() {
+    return distributorStatus.get().getNodes().stream()
```
diemol, line 58: Shouldn't this be the count of the active sessions? Is there a reason to use `usedStereotypes` instead of the `activeSessions` present in the summary?

java/server/src/org/openqa/selenium/grid/graphql/GridData.java
```diff
@@ -19,6 +19,7 @@
 
 import graphql.schema.DataFetcher;
 import graphql.schema.DataFetchingEnvironment;
+
```
diemol: We can revert this to reduce the diff of the PR.

java/server/src/org/openqa/selenium/grid/graphql/Session.java
```diff
@@ -0,0 +1,52 @@
...
+import org.openqa.selenium.Capabilities;
+import org.openqa.selenium.internal.Require;
+import org.openqa.selenium.json.Json;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class Session {
+
+  private final String id;
```
diemol, line 29: Why can't this be a `UUID` like the other `Session` class?

java/server/src/org/openqa/selenium/grid/graphql/Node.java
```diff
@@ -33,18 +38,29 @@
...
+    return activeSessions.stream()
+        .map(session -> new org.openqa.selenium.grid.graphql.Session(session.getId().toString(),
```
diemol, line 62: Why can't we have the `id` as an `UUID` and avoid the `toString`? (See comment below)

java/server/src/org/openqa/selenium/grid/graphql/Session.java
```diff
@@ -0,0 +1,52 @@
...
+  private final String capabilities;
+  private final LocalDateTime startTime;
+  private static final Json JSON = new Json();
+
+  public Session(String id, Capabilities capabilities, LocalDateTime startTime) {
+    this.id = Require.nonNull("Node id", id);
+    this.capabilities = Require.nonNull("Node capabilities", JSON.toJson(capabilities));
```
diemol: Is there a reason to have `capabilities` as a `String`? Why don't we have it as `Capabilities` and do the `JSON.toJson` in the `getCapabilities` method?

diemol (dismissed):
Hi @shucon,
I left a few comments, please do not resolve them, just comment if you wish, so I can check them again.

java/server/src/org/openqa/selenium/grid/data/DistributorStatus.java
shucon: No, we don't want to convert it into a string here. That's done inside the GraphQL folder.

java/server/src/org/openqa/selenium/grid/data/Session.java
shucon: Yes, of course!

java/server/src/org/openqa/selenium/grid/graphql/Session.java
shucon: The ID of sessions is stored using `SessionId` class which does not have a provision of storing as UUID.

java/server/src/org/openqa/selenium/grid/graphql/Node.java
shucon: That's kind of necessary to call toString else it would return a variable of the class `SessionId`

AutomatedTester (approved):

diemol (approved):
>Any outstanding comments can be followed up in tree

I agree, let's merge and iterate on `trunk`

AutomatedTester:
manually merged in 070db738bbc8745e639cff36648586c0910035a5 due to conflicts

Issue #7400, opened 2019-07-16
Fix Ruby spec execution


titusfortner:
This is in response to #7264, since we decided that PR is not the right way to do it. 

I think we decided on: move the non-browser-specific tests out of the browser-specific test files.

So if the "chrome only" tests pass when using Firefox for instance, they should be moved to unit tests or into a common spec file of some kind.

oneEdoubleD:
I'm happy to pick this up.

p0deje:
@titusfortner @oneEdoubleD Any update on this?

titusfortner:
This was fixed with #8511

PR #8497, opened 2020-07-10
Run Ruby tests with IDE

titusfortner:
This is my solution for #7400 

1. Removed some redundancies and rearranged some things in build.desc

2. Removed the cached atoms (they are out of date)

3. Test updates: Created chromium directory for integration tests that apply to both Chrome & Edge and moved integration tests to unit tests as necessary

4. Option mismatch: raise Exception if the browser name of the Options instance does not match the browser name of the test (this was the major problem; You start a new browser and pass in Chrome Options, it doesn't matter what browser is being tested, it will open & pass in Chrome, and fail the pending guard)

5. I added a Rakefile in `rb`. Some of it is a little hacky (any suggestions more than welcome), *but* it lets me run rake tasks to dynamically create and pull in the atoms and the new generated devtools files and the selenium server (the code to actually use the server is in an upcoming PR). If there is a better way to do this, let me know. I hate switching back and forth between the build directory & the regular directory when I'm developing. This allows for users to execute rake tasks from the IDE directly and/or command line rake tasks; no need to use `selenium/go` directly.


rb/spec/integration/selenium/webdriver/chromium/driver_spec.rb
```diff
@@ -22,7 +22,7 @@
 module Selenium
   module WebDriver
     module Chrome
```
p0deje, line 24: This spec now lives in `chromium` directory but the module is still `Chrome`. Can we solve this mismatch somehow?

rb/spec/integration/selenium/webdriver/chrome/options_spec.rb
```diff
@@ -24,34 +24,6 @@ module WebDriver
     module Chrome
       describe Options, only: {browser: :chrome} do
         subject(:options) { Options.new }
-
-        it 'passes emulated device correctly' do
```
p0deje, line 28: Is there any reason why these specs should be removed?

p0deje (commented):
> This allows for users to execute rake tasks from the IDE directly and/or command line rake tasks; no need to use selenium/go directly.

Isn't it possible to configure IDE to just use `go` binary and create few targets in IDE? I don't use one so not sure about that.

Please take a look at CI as it is failing.

rb/spec/integration/selenium/webdriver/chrome/options_spec.rb
titusfortner: because they apply to both chrome & edge_chrome, so they run in the chromium directory:
https://github.com/SeleniumHQ/selenium/pull/8497/files#diff-33481e2e9dad5e27918bd2d2660a9f9fL28

rb/spec/integration/selenium/webdriver/chromium/driver_spec.rb
titusfortner: I just went with the superclass; seemed better than creating 2 subclasses and a superclass in the library.

titusfortner:
Oops, one tweak too many before pushing; I missed a comma.

I honestly don't know if there's a way to configure IntelliJ to interact with the go wrapper. 
Mostly I want an automated way to be able to generate the latest files and copy them over to my working directory. I hate having search results in both rb/lib and build/rb/lib and I have on many occasions made changes to the wrong one. As for everything in the test namepace, I added because it was easy to do, not because I think they are essential.

I did realize that these rake tasks won't work for windows; I'm getting a windows VM set up now for testing, so I can see if I can get something working there as well.

p0deje:
> I hate having search results in both rb/lib and build/rb/lib and I have on many occasions made changes to the wrong one.

There should be a way to exclude `build/` from search results. I locally use ripgrep which respecs .gitignore.

titusfortner:
> There should be a way to exclude build/ from search results. I locally use ripgrep which respecs .gitignore.

Maybe? I think the approach go takes with output in a different directory is great for a build system, but not great for letting me (or the average Ruby dev) work with the code on a machine the way they expect, and this scratches an itch I've had for a long time. I know it's not the best solution, so if we don't want to merge the Rakefile, I suppose I can just keep it for personal use. :)

titusfortner:
I think I found something better...

PR #8504, opened 2020-07-11
Ruby Edge Chrome Fixes

titusfortner:
This gets the tests working for the latest version of MicrosoftEdge on both Mac & Windows.

I just merged to trunk other updates to guards related to windows & mac tests and code to get Safari/STP working correctly. I'm doing this as a PR instead of a commit to trunk so we're on the same page with a couple changes here.

1. `Edge` is now aliased to `EdgeChrome` and not `EdgeHtml`; and `Capabilities#edge` returns values for `EdgeChrome`
2. We have unit tests for both `EdgeChrome` & `EdgeHtml` with those classes specifically, but we're only running integration tests for `Edge`
3. Removed build/task for remote-edge-test since 3.x server doesn't work with Edge Chrome (I'll make it all correct with a Selenium 4 server PR)
4. The rest of this is just updating the guards for things that work now with the Chrome implementation.


rb/build.desc
```diff
@@ -278,16 +279,17 @@ ruby_test(name = "unit",
...
     "spec/unit/selenium/*_spec.rb",
   ],
-  include = ["rb/spec/unit", "build/rb/lib"],
+  include = [
+    "rb/spec/unit",
+    "build/rb/lib"
+  ],
   deps = [
     ":chrome",
-    ":common",
```
twalpole: Have common, remote, and support gone away? or are they inherited now?

rb/spec/integration/selenium/webdriver/window_spec.rb
```diff
@@ -75,7 +75,7 @@ module WebDriver
         expect(new_pos.y).to eq(target_y)
       end
 
-      it 'gets the rect of the current window', only: {browser: %i[edge edge_chrome firefox ie chrome safari safari_preview]} do
+      it 'gets the rect of the current window', only: {browser: %i[edge firefox ie chrome safari safari_preview]} do
```
twalpole: What browsers wouldn't this work for?? If there is one maybe `except` makes more sense than `only`

rb/spec/integration/selenium/webdriver/window_spec.rb
```diff
@@ -86,7 +86,7 @@ module WebDriver
         expect(rect.height).to be >= 0
       end
 
-      it 'sets the rect of the current window', only: {browser: %i[edge_chrome firefox ie chrome safari safari_preview]} do
+      it 'sets the rect of the current window', only: {browser: %i[edge firefox ie chrome safari safari_preview]} do
```
twalpole: Same question as above

twalpole (approved):
Just added a couple of small comments but overall 👍

rb/build.desc
titusfortner: Each browser inherits common, which inherits everything from support & remote. Seemed unnecessary to repeat it everywhere when nothing was using it.

rb/spec/integration/selenium/webdriver/window_spec.rb
titusfortner: Good catch, I'll fix

Issue #6791, opened 2018-12-27
Unable to run Safari Technology Preview with Selenium standalone server


cuki:
## 🐛 Bug Report
According to this commit (6e09f51a5ff1debfe67e7fb1c5cbd21426c1c143) you should be able to switch between Safari and Safari Technology Preview by using the respective browser name (["Safari Technology Preview"](https://github.com/SeleniumHQ/selenium/blob/b62378814139566f6fcc4814d1243dd8b845eda0/java/client/src/org/openqa/selenium/safari/SafariOptions.java#L52)).
https://github.com/SeleniumHQ/selenium/blob/b62378814139566f6fcc4814d1243dd8b845eda0/java/client/src/org/openqa/selenium/safari/SafariOptions.java#L195

It seems whatever you specify as capabilities is discarded and reset upon creation of new `SafariOptions` instance:
https://github.com/SeleniumHQ/selenium/blob/b62378814139566f6fcc4814d1243dd8b845eda0/java/client/src/org/openqa/selenium/safari/SafariOptions.java#L72-L75

Simply put, it does not respect the specified `browserName` value.

## To Reproduce

1. Enable safaridriver
```
/usr/bin/safaridriver --enable
```
2. Start Selenium standalone server
```
selenium-standalone start
```
3. Initiate new session with "Safari Technology Preview"
```
curl -d '{ "capabilities": { "alwaysMatch": { "browserName": "Safari Technology Preview" }, "firstMatch": [ {} ] }, "desiredCapabilities": { "browserName": "Safari Technology Preview" } }' http://localhost:4444/wd/hub/session
```

### Logs
```
org.openqa.selenium.SessionNotCreatedException: Unable to create session from {
  "desiredCapabilities": {
    "browserName": "Safari Technology Preview"
  },
  "capabilities": {
    "firstMatch": [
      {
        "browserName": "Safari Technology Preview"
      }
    ]
  }
}
Build info: version: '3.141.0', revision: '2ecb7d9a', time: '2018-10-31T20:22:52'
System info: host: 'MBP', ip: '127.0.0.1', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.13.6', java.version: '1.8.0_131'
Driver info: driver.version: unknown
    at org.openqa.selenium.remote.server.NewSessionPipeline.lambda$null$4(NewSessionPipeline.java:76)
    at java.util.Optional.orElseThrow(Optional.java:290)
[...]

PR #8506, opened 2020-07-11
Rb server toggle

titusfortner:
~This relies on #8504, so this PR is in draft until I get that merged/rebased~ 

1. Using go for a remote test will build the latest server
2. If the latest server is found, this code will use it, unless `ENV['DOWNLOAD_SERVER']` is set, then it will download and use latest
3. Syntax for starting server is different between 3.x/2.x and 4.x so I'm parsing; we need this even after 4.x is released since someone may pass in a 3.x jar
4. STP should be toggled by the new browser name, which makes the code more simple, but also current STP is broken on both 3.x and 4.x servers (#6791)
5. And I'm undoing the removal of remote as a build library since we need the server to get built in `deps`, but that doesn't work in a test; so remote is only being called from remote tests, not from any of the libs and not for building the gem.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -263,7 +263,10 @@ task test_rb_remote: [
   '//rb:remote-chrome-test',
   '//rb:remote-firefox-test',
   ('//rb:remote-safari-test' if SeleniumRake::Checks.mac?),
+  # BUG - https://github.com/SeleniumHQ/selenium/issues/6791
+  # ('//rb:remote-safari-preview-test' if SeleniumRake::Checks.mac?),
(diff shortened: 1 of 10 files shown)


titusfortner:
Oh and the tests fail because the ***server*** is broken, so either we file a bug report or we keep them failing. I think these failures should be blockers for v4 release.
* Unable to open more than one session at a time
* Unable to upload a file to a remote machine

titusfortner:
I'm going to merge this since all the tests are passing.
Just a heads up that Ruby code now works for 4.x server and that is the default in the remote tests.
Logged & tracking the bugs.

PR #8512, opened 2020-07-13
rb specific rakefile

titusfortner:
Well, atoms built from trunk have diverged (not sure how substantially) from what is cached, so we should probably delete them or update them.

I thought deleting them and then making a way to auto-update them as needed was a good way to go.

(I tried to implement something for both Windows & Mac, but I couldn't get it to work on Windows)


twalpole:
@titusfortner The reason for having the atoms in the codebase was to allow for use of the current branch directly from a gemfile.  That meant we needed to remember to update them whenever the atoms changed, which wasn't great - and obviously we didn't - but actual release versions of the gem (and the tests) would have had the latest because they are overwritten during the build process.

titusfortner:
Ah, right, and having a rake command to generate it in the directory isn't going to help that. :-/
So we'll also want to do the same for the devtools directory whenever that gets settled.

AutomatedTester:
How much work is needed for bazel to do the build?

titusfortner:
@AutomatedTester this uses `rb/Rakefile` to build artifacts using bazel and copy them into the working directory instead of into `/build/rb/`. Much easier for most Ruby devs to work with rake and rspec and RubyMine inside the rb directory (like most projects) than to use go. That said, I think we've decided this isn't the right approach, though; so I'm going to keep my Rakefile to myself. :)

PR #8403, opened 2020-06-09
WIP: Support RedisBacked Distributor

raju249:
Before submitting your PR, please check our [contributing](https://github.com/SeleniumHQ/selenium/blob/master/CONTRIBUTING.md) guidelines.
Avoid large PRs, help reviewers by making them as simple and short as possible.

### Description

Adds support for Distributor backed with redis for storing info about the nodes in the grid.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist

- [ ] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/master/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #8513, opened 2020-07-13
Move node selection logic in a package

### LINKED ISSUE #8173: [Grid] Distributor should store state
## 🚀 Feature Proposal

Enabling the Distributor to store state would enable the possibility of having more than one distributor running at a time. We need to deal properly with the concurrency when changing the stored states

## Motivation

See description

raju249:

### Description
This PR separates the tightly coupled logic of `nodeSelection` to run a session, into a separate package and can be independently modified going forward. It doesn't and should not change any functionality from previoud build and should not show any differences for the user to run a session.

### Motivation and Context

1. There is work going on adding support for `RedisDistributor` for which this same nodeSelection Package would be used/modified to make it work with redis backed distributor.
2. Hence as a first step in MVP and not have too may changes in `RedisDistributor` PR, this PR will make that change and I would iterate from there after it is merged.
3. The `GridModel` class stores the nodes and listens to addition and removal of Nodes. The `NodeSelector` class will help us select the node from the pool of nodes. Now, this pool of the nodes can be an in-memory hash map for a local distributor, it can also be a redis backed store for `RedisBasedDistributor` or a database backed store for `JdbcBackedDistributor`

The changes to make the datastore configurable inside the `NodeSelector` class will be done in #8403 

### Types of changes

- [-] Bug fix (non-breaking change which fixes an issue)
- [-] New feature (non-breaking change which adds functionality)
- [-] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist

- [-] My change requires a change to the documentation.
- [-] I have updated the documentation accordingly.


diemol (commented):
@raju249 
Could you please add some context to the PR? The PR template is being completely ignored.
How can this be tested? Why are these changes made?

raju249:
Hey @diemol I have updated the PR description. Please take a look. Thanks 🙇

shs96c (changes requested):
I suspect that what we want to do is to ensure that the model is immutable, and a snapshot of the understood state of the system at the time it's created. It's also just a model, so there's no need to have things like `Host.runHealthCheck`

AutomatedTester:
I believe all comments have been addressed and will merge on Tuesday morning

@shs96c @diemol

diemol (approved):
@raju249 I made the changes we spoke about. Please have a look, and if you are ok with them, we can merge this. Thanks!

raju249:
Thanks much @diemol 🙇  

Changes LGTM. 👍 

We can land this.

Issue #6703, opened 2018-11-29
Ensure traces are injected and read to and from HTTP calls.


shs96c:
Without this, the traces only really make sense in the standalone mode.

dratler:
Hi @shs96c ,
can you share more light what is required to do ?
are these traces been send to a Driver Log ?

pujagani:
For most configurations, the traces are correctly ordered, complete and match the request flow. However, adding few spans and events specially for error cases is needed. I have documented the findings here  https://gist.github.com/pujagani/2e2de5d79a2f30da5edb1582208130aa

AutomatedTester:
Closing this as it appears to be done now

PR #8541, opened 2020-07-16
[tracing] Match client and server spans with OpenTelemetry specification

pujagani:
### Description
The changes are related to issue #6703. Improvises existing HTTP span error handling.

### Motivation and Context

Current implementation has SpanWrapperHTTPHandler to generate spans for any request received by the server and TracedHTTPClient to generate spans for a client request. The changes help adhere to the [OpenTelemetry specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md) to ensure consistent HTTP spans. The  HTTP error code mapping with span status adheres to [OpenTelemetry specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md) and  [W3C Webdriver Error Codes](https://www.w3.org/TR/webdriver/#errors).

### Types of changes

### Checklist


java/client/src/org/openqa/selenium/remote/tracing/Tags.java
```diff
@@ -38,6 +50,35 @@ private Tags() {
...
   public static final BiConsumer<Span, HttpResponse> HTTP_RESPONSE = (span, res) -> {
-    span.setAttribute("http.status_code", res.getStatus());
+    int statusCode = res.getStatus();
+    if (res.getTargetHost() != null) {
+      span.setAttribute("http.target_host", res.getTargetHost());
+    }
+    res.getTargetHost();
+    span.setAttribute("http.status_code", statusCode);
+
+    switch (statusCode / 100) {
```
AutomatedTester: This would be better as just status codes rather divided

AutomatedTester (changes requested):

java/client/src/org/openqa/selenium/remote/tracing/Tags.java
pujagani: I understand your point. The division is so we can match the error code to a family of error codes rather than a particular error code. I did consider a cleaner way of doing using native java package but we don't have that dependency in the workspace. I think I can leverage existing Netty server response to make the code readable rather that using if-else.

AutomatedTester (approved):

AutomatedTester:
This is much better without the magic numbers!

PR #8546, opened 2020-07-18
Json speed improvement

joerg1985:
### Description
JsonOutput and Input created a lot of object instances while writing / parsing Json data.
- JsonOutput created one String per String char to write
- Input created one CharBuffer (including a backing array) per char to read

Input / InputOutput now takes a Reader instead of a Readable, due to this change it is a breaking change in my mind.

### Motivation and Context
This improves the performance of processing big json data, e.g. 4k screenshots and reduces the need for GC.
- improved avg. encoding time from 67 ms to 31 ms for 3 MiB String output
- improved avg. decoding time from 118 ms to 54 ms for 3 MiB raw Json input

These numbers are determined using my local system with a loop of 100 iterations
(AMD Ryzen 3 PRO 3200GE, 16 GB RAM, Fedora 32, OpenJDK 11)

### Types of changes

### Checklist


diemol (approved):
Thank you @joerg1985!
It should not be a breaking change, it is changing a package-protected constructor

fmaretic:
Hello everyone. Could this improvement also be merged with the Selenium 3 version? The latest tag is 3.141.59 as I can see. This is causing a huge memory issue, for example when calling the WebDriver's getPageSource().

diemol:
@fmaretic , we do not backport changes to previous releases, it'd be great if you can try the Selenium 4 Alpha 7, which includes this change.

PR #8547, opened 2020-07-18
create_no_window option for chrome webdriver process

ali-sajjad-rizavi:
This new option will be helpful when we use selenium with a GUI application.
Using the usual code;
```
driver = webdriver.Chrome()
```
a new terminal/command-prompt window opens up. And in case of console application, it shared the same stdout.

With this added feature, we are able to run chromedriver process in the background;
```
driver = webdriver.Chrome(create_no_window=True) # Opens chromedriver in background
```

What this actually does, is that it sets the **creationflags** of the subprocess as **CREATE_NO_WINDOW**.


ali-sajjad-rizavi:
The changes which I made did not affect other things, couldn't understand why the build fails.

.gitignore
```diff
@@ -66,8 +66,6 @@ py/selenium/webdriver/firefox/webdriver.xpi
 py/selenium/webdriver/firefox/webdriver_prefs.json
 py/selenium/webdriver/firefox/x86/
 py/selenium/webdriver/remote/findElements.js
-py/selenium/webdriver/remote/getAttribute.js
```
AutomatedTester: I'm not sure why you removed these. Please add them back

py/selenium/webdriver/chrome/webdriver.py
```diff
@@ -35,7 +35,7 @@ class WebDriver(ChromiumDriver):
     def __init__(self, executable_path="chromedriver", port=DEFAULT_PORT,
                  options=None, service_args=None,
                  desired_capabilities=None, service_log_path=DEFAULT_SERVICE_LOG_PATH,
-                 chrome_options=None, service=None, keep_alive=True):
+                 chrome_options=None, service=None, keep_alive=True, create_no_window=False):
```
AutomatedTester, line 38: We shouldn't add this as a new `kwarg` here. This should all be done on the options class

py/selenium/webdriver/chrome/service.py
```diff
@@ -24,7 +24,7 @@ class Service(service.ChromiumService):
     """
 
     def __init__(self, executable_path, port=0, service_args=None,
-                 log_path=None, env=None):
+                 log_path=None, env=None, create_no_window=False):
```
AutomatedTester, line 27: I would rather no have this as a `kwarg` as it encourages "growth" which lead to an unweildy constructor in other classes. Let's add a method or property to take care of this instead as I think it's usage is going to be quite low.

py/selenium/webdriver/chromium/service.py
```diff
@@ -24,15 +24,16 @@ class ChromiumService(service.Service):
     """
 
     def __init__(self, executable_path, port=0, service_args=None,
-                 log_path=None, env=None, start_error_message=None):
+                 log_path=None, env=None, start_error_message=None, create_no_window=False):
```
AutomatedTester, line 27: This would be better served as a method/property that is set when people don't want to a window.

AutomatedTester (changes requested):
Please delete the .js files you have added here as they are put in the right place at the right time during a build step.

py/selenium/webdriver/chromium/service.py
ali-sajjad-rizavi: Thankyou for your response, I will try to make this a good fit in the Options class (for chrome)

diemol (commented):
@ali-sajjad-rizavi do you need any help to get this sorted out? Let us know if and someone from the team can quickly give you a hand.

ali-sajjad-rizavi:
The changes have been messed up after a merge, so I had forked the repository again. The feature I added before was actually rejected because I did not implemented in the 'Options' class. Will try to make better structure this time!

Issue #6694, opened 2018-11-27
Support structured logging on stdout


shs96c:
This allows systems such as DataDog to extract meaningful information from the logs. Which can only be a Good Thing.

shs96c:
988667fe23d4637e55c67eb7bf2b4926fcedc7ee landed much of what we need for this to happen. The next step will be land a mechanism to add richer metadata to the log messages, presumably related to the information set on the tracing spans.

AutomatedTester:
This was landed in #8583

PR #8583, opened 2020-07-29
[tracing] Structured logging based on Span events

pujagani:
### Description

The changes are related to issue #6694. Current implementation has structured logging in place, the changes provide additional context to the logs by encapsulating the information from span and events to logs by primarily logging events. This is the first iteration to make changes beginning with logging HTTP calls and database calls.

Changes below include:
1. Add logging for events at INFO level as JSON in LoggingOptions.
2. Update in event attribute datatypes to add arrays for String, Long, Double and Boolean.
3. Standardise error events to match OpenTelemetry specification https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/exceptions.md.
4. Instrument SpanWrappedHTTPHandler and TracedHTTPClient with events for logging. 
5. Instrument HandleSession, ReverseProxyHandler, JDBCBackedSessionMap and RedisBackedSessionMap with events for logging.
6. Instrument DockerSessionFactory and DriverServiceSessionFactory with events for logging.
7. Instrument LocalNode, LocalSessionMap, LocalDistributor. ProtocolConverter and GridStatusHandler with events for logging.

### Motivation and Context
Span events can have attributes which can store all possible information for a particular operation. Leveraging the events, this information can be added to logs to provide rich contextual information in logs. This makes the debugging process easier. Rather than having multiple log statements for a single operation, event can be used to capture all possible context (including high-cardinality values) and used in logging. 

### Types of changes

### Checklist


java/client/src/org/openqa/selenium/remote/tracing/AttributeKey.java
```diff
@@ -0,0 +1,56 @@
...
+  HTTP_STATUS_CODE("http.status_code"),
+  HTTP_TARGET_HOST("http.target_host"),
+  HTTP_CLIENT_CLASS("http.client_class"),
+  HTTP_HANDLER_CLASS("http.handler_class"),
+
+  SESSION_ID("session.id"),
+  SESSION_CAPABILITIES("session.capabilities"),
+  SESSION_URI("session.uri"),
+
+  DATABASE_STATEMENT ("db.statement"),
```
shs96c, line 47: These seems specific to the JDBC classes, and probably shouldn't be in the common keys.

java/client/src/org/openqa/selenium/remote/tracing/AttributeKey.java
```diff
@@ -0,0 +1,56 @@
...
+  DATABASE_OPERATION ("db.operation"),
+  DATABASE_USER ("db.user"),
+  DATABASE_CONNECTION_STRING ("db.connection_string"),
+  DATABASE_SYSTEM("db.system");
+
+  private String name;
```
shs96c: `final`

java/client/src/org/openqa/selenium/remote/tracing/AttributeKey.java
```diff
@@ -0,0 +1,56 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package org.openqa.selenium.remote.tracing;
+
+public enum AttributeKey {
```
shs96c, line 20: It's likely that we can't know all the possible keys in advance. An `enum` is a poor choice given this constraint, so it may be better to use an `interface`

java/client/src/org/openqa/selenium/remote/tracing/EventAttribute.java
```diff
@@ -34,6 +34,22 @@ public static EventAttributeValue setValue(long value) {
     return new EventAttributeValue(value);
   }
 
+  public static EventAttributeValue setValue(String[] value) {
+    return new EventAttributeValue(value);
+  }
+
+  public static EventAttributeValue setValue(Long[] value) {
```
shs96c: Prefer primitive types.

java/client/src/org/openqa/selenium/remote/tracing/EventAttribute.java
```diff
@@ -34,6 +34,22 @@ public static EventAttributeValue setValue(long value) {
     return new EventAttributeValue(value);
   }
 
+  public static EventAttributeValue setValue(String[] value) {
```
shs96c: Why not use varargs here? That makes usage more like `setValue("foo", "bar")` and ends up using an array behind the scenes.

java/client/src/org/openqa/selenium/remote/tracing/EventAttributeValue.java
```diff
@@ -44,6 +48,26 @@ public EventAttributeValue(boolean value) {
     this.type = Type.BOOLEAN;
   }
 
+  public EventAttributeValue(String[] value) {
+    this.stringArrayValue = value;
+    this.type = Type.STRING_ARRAY;
+  }
+
+  public EventAttributeValue(Long[] value) {
```
shs96c, line 56: Again, prefer primitive types `long` to `Long`. This avoids the need for null checks in your own code.

java/client/src/org/openqa/selenium/remote/tracing/EventAttributeValue.java
```diff
@@ -64,7 +96,11 @@ public Type getAttributeType() {
     STRING,
     BOOLEAN,
     LONG,
-    DOUBLE
+    DOUBLE,
```
shs96c: Since we've not shipped this, please alphabetise. Once we've shipped, it's best to maintain the ordinals of the enum, but for now it doesn't matter.

java/client/src/org/openqa/selenium/remote/tracing/Tags.java
```diff
@@ -65,4 +65,37 @@ private Tags() {
       span.setStatus(Status.UNKNOWN);
     }
   };
+
+  public static final BiConsumer<Map<String, EventAttributeValue>, HttpRequest>
+      HTTP_REQUEST_EVENT =
+      (map, req) -> {
+        map.put(AttributeKey.HTTP_METHOD.toString(),
```
shs96c: I've seen a lot of `.toString` being used in this method. Why not simply have the key of the map be an `AttributeKey`?

java/client/src/org/openqa/selenium/remote/tracing/opentelemetry/OpenTelemetrySpan.java
```diff
@@ -112,6 +112,21 @@ public Span addEvent(String name, Map<String, EventAttributeValue> attributeMap)
               otAttributes.setAttribute(key, AttributeValue.stringAttributeValue(value.getStringValue()));
               break;
 
+            case STRING_ARRAY:
```
shs96c, line 127: Alphabetise, please.

java/client/test/org/openqa/selenium/remote/tracing/opentelemetry/TracerTest.java
```diff
@@ -118,11 +131,13 @@ public void shouldBeAbleToCreateASpanWithEvents() {
         .isEqualTo("Test event start");
     assertThat(timedEvents).element(0).extracting(SpanData.Event::getTotalAttributeCount).isEqualTo(0);
     assertThat(timedEvents).element(1).extracting(SpanData.Event::getName).isEqualTo("Test event end");
-    assertThat(timedEvents).element(1).extracting(SpanData.Event::getTotalAttributeCount).isEqualTo(6);
+    assertThat(timedEvents).element(1).extracting(SpanData.Event::getTotalAttributeCount).isEqualTo(10);
```
shs96c: This is why reusing test cases is a less than ideal idea. Prefer lots of small tests than mashing things into a single test case. That allows test names to be descriptive. In the ideal world, we'd be able to delete the class under test and all the test methods, leaving just the test method signatures, and create something functionally equivalent.

java/server/src/org/openqa/selenium/grid/log/LoggingOptions.java
```diff
@@ -102,7 +113,63 @@ private Tracer createTracer() {
...
+                  attributeValue = value.getStringArrayValue();
+                  break;
+                case LONG_ARRAY:
+                  attributeValue = value.getLongArrayValue();
+                  break;
+                case DOUBLE_ARRAY:
+                  attributeValue = value.getDoubleArrayValue();
+                  break;
+                case BOOLEAN_ARRAY:
+                  attributeValue = value.getBooleanArrayValue();
```
shs96c, line 159: This case statement makes me twitch. It's long, and it looks like a prime candidate to accidentally forget to extend if we add a new value type. At the least, I'd suggest making `Attributes` be aware of how to serialise itself to JSON, so you can do a `map.put("attributes", event.getAttributes)`

java/server/src/org/openqa/selenium/grid/log/LoggingOptions.java
```diff
@@ -102,7 +113,63 @@ private Tracer createTracer() {
...
+            map.put("traceId", traceId);
+            map.put("spanId", spanId);
+            map.put("spanKind", span.getKind().toString());
+            map.put("eventName", event.getName());
+
+            Attributes attributes = event.getAttributes();
+            Map<String, Object> attributeMap = new HashMap<>();
+            attributes.forEach((key, value) -> {
+              Object attributeValue = null;
+              switch (value.getType()) {
```
shs96c, line 136: Missed a `default` case

java/server/src/org/openqa/selenium/grid/log/LoggingOptions.java
```diff
@@ -171,4 +238,13 @@ private OutputStream getOutputStream() {
         })
         .orElse(System.out);
   }
+
+  private String getJSONString(Map<String, Object> map) {
```
shs96c: `Json`, not `JSON`. We try and avoid shouting on this project :)

java/server/src/org/openqa/selenium/grid/log/LoggingOptions.java
```diff
@@ -171,4 +238,13 @@ private OutputStream getOutputStream() {
...
+    StringBuilder text = new StringBuilder();
+    try (JsonOutput json = JSON.newOutput(text).setPrettyPrint(false)) {
+      json.write(map);
+      text.append('\n');
```
shs96c, line 249: :eyes: Uhhh... Why? The logger will output everything on one line for us.

shs96c (changes requested):

AutomatedTester (approved):
Looks good to me but will wait for @shs96c to have a look

java/client/src/org/openqa/selenium/remote/tracing/AttributeKey.java
pujagani: These are based on OpenTelemetry specification for databases and it is applicable to all SQL and NoSQL client calls.

java/client/src/org/openqa/selenium/remote/tracing/AttributeKey.java
pujagani: +1 , it is not possible to know keys in advance. AttributeKey enum is to store keys which are used repeatedly and which are part of OpenTelemetry specification, all attribute key do not need to be part of the enum. Do you still suggest to use an interface instead?

java/client/src/org/openqa/selenium/remote/tracing/EventAttribute.java
pujagani: +1  Will update it.

java/client/src/org/openqa/selenium/remote/tracing/EventAttribute.java
pujagani: OpenTelemetry APIs for array attributes accepts boxes types so stuck to that.

java/server/src/org/openqa/selenium/grid/log/LoggingOptions.java
pujagani: This is the exhaustive list of all the datatypes for event attributes, so any value that is passed will match one of the cases, hence no `default` case.

java/server/src/org/openqa/selenium/grid/log/LoggingOptions.java
pujagani: Will keep in mind. :)

java/server/src/org/openqa/selenium/grid/log/LoggingOptions.java
pujagani: If I passed the Json String as it is, it was pretty printing the Json, so this was added to disable the pretty printing and have the event log in one line.

java/server/src/org/openqa/selenium/grid/log/LoggingOptions.java
pujagani: I understand the concern expressed. However, the OpenTelemetry `Attributes` class stores a `String` key and `AttributeValue` value and in order to get the actual value from `AttributeValue`, it offers series of methods of based on the `AttributeValue.Type`. So this is the only way I am seeing to creating the attributeMap from Events, unless we extend `Event`.  I will try and see how this can be made better.

java/client/src/org/openqa/selenium/remote/tracing/Tags.java
pujagani: Update the `AttributeKey` enum to have a getter instead. Since this stores commonly used constants for the key. The `Map<String, EventAttributeValue>` is used since keys can be retrieved from the AttributeKey enum or they can be any other strings.

java/server/src/org/openqa/selenium/grid/log/LoggingOptions.java
pujagani: Initially I considered extending either the `Event` or `Attributes` to add a method for Json serialisation of the attributes. After looking into the OTel classes, the `Attributes` implementation `AutoValue_Attributes_ArrayBackedAttributes` is final and cannot be extended. Same is applicable for `Events` implementation `AutoValue_TimedEvent_RawTimedEvent`. So we might be stuck with the switch case. Any suggestions on this front?

AutomatedTester:
I believe all comments have been addressed and will be merging on Tuesday morning if there are no further review comments.

@shs96c @diemol

diemol:
Just a heads up, this PR touches some files that this https://github.com/SeleniumHQ/selenium/pull/8513 one is also changing. I'd recommend merging the other one first and then solving the conflicts here afterwards.

pujagani:
Verified that the changes in #8513 have not impacted the changes here. It is working as intended. @AutomatedTester

Issue #8524, opened 2020-07-14
File Uploads Don't Work with Server


titusfortner:
## 💥 Regression Report
Note: I've returned the Ruby tests to testing remote commands against the server built from current code, rather than downloading latest release.

It appears that the server is sending files to `se/file` endpoint of the driver rather than saving the encoded info and sending the path to the driver. Not sure how server is parsing if the destination is an endpoint or another intermediary node.

## Last working Selenium version
3.141.52

## Stopped working in version:
4.x

## To Reproduce
This spec is failing when run with remote driver:
https://github.com/SeleniumHQ/selenium/blob/dd7090c/rb/spec/integration/selenium/webdriver/remote/element_spec.rb#L33

`./go //rb:remote-chrome-test`

## Log
```
2020-07-09 22:15:30 INFO Selenium -> POST session/1eacf2419fce243cde8ef36d3a7efbc7/se/file
2020-07-09 22:15:30 INFO Selenium    >>> http://127.0.0.1:4445/wd/hub/session/1eacf2419fce243cde8ef36d3a7efbc7/se/file | {"file":<SNIP>}
2020-07-09 22:15:30 INFO Selenium <- {"value":{"error":"unknown command","message":"unknown command: unknown command: session/1eacf2419fce243cde8ef36d3a7efbc7/se/file","stacktrace": <SNIP>}}
```

## Expected Behavior

Expect File to be successfully uploaded & spec to pass

## Actual Behavior
Chrome:
```
     Selenium::WebDriver::Error::UnknownCommandError:
       unknown command: unknown command: session/d44106eddefb50d91ce603e6434a6ac1/se/file
     # 0   chromedriver                        0x0000000100c87e99 chromedriver + 4808345
...
     # 14  ???                                 0x0000000000000002 0x0 + 2
```
Firefox:
```
 Selenium::WebDriver::Error::WebDriverError:
       unexpected response, code=405, content-type="text/plain"
       HTTP method not allowed
```

## Environment

OS: All
Browser: All
Language Bindings version: Trunk
Selenium Grid version: Trunk

barancev:
According to what I see in other bindings the command should have the form "/session/$sessionId/file"

Java: https://github.com/SeleniumHQ/selenium/blob/18daafa91469a24c4a63cad6b18096cc91b12261/java/client/src/org/openqa/selenium/remote/codec/AbstractHttpCommandCodec.java#L167
[...]

PR #8588, opened 2020-07-30
add se namespacing to file upload endpoint for all bindings

titusfortner:
Follow up from discussion in #8524 

@shs96c / @barancev is this the right way to do it for Java?


shs96c (approved):
Java bit looks good.

AutomatedTester (approved):

jimevans (approved):

PR #8590, opened 2020-07-31
[dotnet] Attach RemoteWebDriver to existing session.

bart-tricentis:
Adds a new constructor for RemoteWebDriver that will not start
a new test session but instead attach to an existing one.

This will allow C# developers to write code that is ment to
interact with running selenium sessions. This will for example
enable creation of customization plugins for selenium based software.

Adds a new constructor for RemoteWebDriver that will not start
a new test session but instead attach to an existing one.

### Description
Added a new constructor and refactored part of the "main" one to avoid code duplication.

### Motivation and Context
My company provides test automation software in which the mobile part is done using Appium. In order to implement a feature that will allow the software users to create custom Appium snippets (in dotnet C#) that work together with what the software offers out of the box, we need to have the ability to attach to an existing Appium session. And as the C# AppiumWebDriver inherits from RemoteWebDriver we need to have a way to create an instance of RemoteWebDriver without triggering a new session.

### Types of changes

### Checklist

- [x] Existing tests passed.

### DIFF
--- a/dotnet/src/webdriver/Remote/RemoteWebDriver.cs
+++ b/dotnet/src/webdriver/Remote/RemoteWebDriver.cs
@@ -139,42 +139,42 @@ namespace OpenQA.Selenium.Remote
             this.StartClient();
             this.StartSession(desiredCapabilities);
             this.elementFactory = new RemoteWebElementFactory(this);
+            CheckForSupportedCapabilities();
 
-            if (this.capabilities.HasCapability(CapabilityType.SupportsApplicationCache))
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RemoteWebDriver"/> class that attaches to an existing test session
+        /// </summary>
+        /// <param name="commandExecutor">An <see cref="ICommandExecutor"/> object which executes commands for the driver.</param>
+        /// <param name="sessionId">The key for the existing session.</param>
+        public RemoteWebDriver(ICommandExecutor commandExecutor, string sessionId)
+        {
+            this.executor = commandExecutor;
+            this.sessionId = new SessionId(sessionId);
+            this.elementFactory = new RemoteWebElementFactory(this);
+
+            Response commandResponse = new Response();
+
+            try
             {
-                object appCacheCapability = this.capabilities.GetCapability(CapabilityType.SupportsApplicationCache);
-                if (appCacheCapability is bool && (bool)appCacheCapability)
-                {
-                    this.appCache = new RemoteApplicationCache(this);
-                }
+                commandResponse = this.Execute(DriverCommand.GetSessionCapabilities, null);
+            }
+            catch (System.Net.WebException e)
+            {
+                commandResponse.Status = WebDriverResult.UnhandledError;
+                commandResponse.Value = e;
             }
 
-            if (this.capabilities.HasCapability(CapabilityType.SupportsLocationContext))
+            if (commandResponse.Status != WebDriverResult.Success)
             {
-                object locationContextCapability = this.capabilities.GetCapability(CapabilityType.SupportsLocationContext);
-                if (locationContextCapability is bool && (bool)locationContextCapability)
-                {
-                    this.locationContext = new RemoteLocationContext(this);
-                }
+                UnpackAndThrowOnError(commandResponse);
             }
+            Dictionary<string, object> rawCapabilities = (Dictionary<string, object>)commandResponse.Value;
+            ReturnedCapabilities returnedCapabilities = new ReturnedCapabilities(rawCapabilities);
+            this.capabilities = returnedCapabilities;
 
-            if (this.capabilities.HasCapability(CapabilityType.SupportsWebStorage))
-            {
-                object webContextCapability = this.capabilities.GetCapability(CapabilityType.SupportsWebStorage);
-                if (webContextCapability is bool && (bool)webContextCapability)
-                {
-                    this.storage = new RemoteWebStorage(this);
-                }
-            }
-
-            if ((this as ISupportsLogs) != null)
-            {
-                // Only add the legacy log commands if the driver supports
-                // retrieving the logs via the extension end points.
-                this.CommandExecutor.CommandInfoRepository.TryAddCommand(DriverCommand.GetAvailableLogTypes, new CommandInfo(CommandInfo.GetCommand, "/session/{sessionId}/se/log/types"));
-                this.CommandExecutor.CommandInfoRepository.TryAddCommand(DriverCommand.GetLog, new CommandInfo(CommandInfo.PostCommand, "/session/{sessionId}/se/log"));
-            }
-
+            CheckForSupportedCapabilities();
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


jimevans:
I'm not necessarily opposed to such a feature in principle, but I'm loathe to implement a feature in a single language binding that will not also be implemented in other language bindings. Other language bindings maintainers (@shs96c, @AutomatedTester, @titusfortner, and others [feel free to tag whoever else]) should weigh in on adding this feature. Or, if the feature already exists in those languages, and I just don't know about it, please feel free to set me straight.

titusfortner:
So a new session is created in one place, and the id gets stored, but not the object in memory, so a new object needs to be created but still with that ID? I'm not sure I understand the use case, but I'm also not opposed to it and should be a pretty quick add. Should this apply to just the remote driver, or any driver?

AutomatedTester:
@bart-tricentis could you explain the use case for this.

I am opposed to this, but happy to way up against a proper use case, as it has the potential to leak data between session. Historically we have not been allowing this as not many people would not clean up their sessions properly if we did not do it.

I look forward to seeing a real world use case.

bart-tricentis:
@AutomatedTester Sure

My company's software is a tool for automating test cases for many different technologies in model based testing. One of the technologies is mobile that is entirely based on vanilla Appium. On of the main features of this software is scanning application for controls that can be automated for testcases (can be really simple tests can be really complex). The whole software in made .net C# so we use the Appium dotnet driver, however the usage of this driver is totally transparent to the software user. He is oblivious to the Appium session calls that are being made in order to "make things work". Because of this, the innerworkings of the software are made in such a way that the user can identify and correctly control as many android/ios controls as possible. This however means that there are cases of custom controls for which we fail at those things. 

So to combat that the software has a feature that allows the user to write custom pieces of code (also in .net c#) that will handle this issue. In order to have this feature to work for mobile we need to have a way to interact with an ongoing Appium session. That is because a whole Test case is one Appium session so if the user for example has a mobile app that consists of 20 "standard" controls and one "custom", he can't perform tests that require the use of that custom controls (not in a clean and easy way at least). If we would the "Attach to session" feature in the dotnet driver as user could simply create one custom test step in a huge End to End test where he could attach to an ongoing Appium session "do what needs to be done" and let the rest of the test procced as normal.

@timsutton 
The use case we have for this is that one process would (in a separate executable even) start a session and another process would attach to the session that is already ongoing (when certain criteria are met).
Plus this could be even used cross-technology (if someone would need it) as you can get the session id via the Rest API. (And technically you can "attach" to interfere with an ongoing session just by curl calls alone, though it's not the muse user friendly way)

@jimevans 
I do not know how other technologies stand on this but I've seen a working example of this in JavaScript in Appium, but again I do not know if the JS Appium driver uses the selenium one as it's base.

jimevans:
@bart-tricentis How would that be different from requiring your users to implement a custom interface in their class (or extend an abstract class that your product provides) which takes the driver instance as a parameter. There are a number of ways you could architect your software to support A plug-in-like architecture that would accomplish this. I get that you may not have implemented it this way currently, but it seems to me that’s not the fault of Selenium.

titusfortner:
Right, your principles of encapsulation make sense, but I don't think your comment really gets at why the custom pieces of code can't be executed by the same process that started the session. Is the entire test not known ahead of time? Is the custom code not accessible by the underlying framework that creates the session? In that case, why shouldn't testing this component get its own session?

I seem to remember having a conversation about this kind of functionality at Selenium Conf UK in 2016 with someone who wanted this functionality in Ruby. The right solution there was to use a subclass of Remote Driver class, but you can't easily subclass the superclass for the class you actually need to work with. My general take at this point is that I don't think it is Selenium's job to enforce a "Right" way of doing things, so while I'm curious the answers in the previous paragraph, I'm more likely to default to providing the ability to do things like this in the absence of an obvious workaround.

bart-tricentis:
Your points are valid. However the software has an overarching architecture that does not allow for such a solution as you mentioned to be implemented. Unfortunately I cannot disclose any detailed information about the software's source code. All I can say is this, as the software has the ability to automate many different technologies (of which mobile is a relatively recent addition compared to others) the software's architecture was not made with selenium in mind. Because of that, this Pull request is one of our last ditch efforts to implement a solution for this use case (a relatively simple solution that is).

As I mentioned above, nothing is stopping us from making direct REST API calls to a Appium server as we know session ID of the current "active" session or writing a completely new remote appium web driver for the ground up. However writing a new C# wrapper for all the supported appium commands would be to put it blant "reinventing the wheel". So from a certain point of view this Pull request is more of a quality of live feature more than anything else.

Still I understand your perspective on the matter and if at the end of the day you will still be opposed to this change I will simply abandon this Pull request.

titusfortner:
So, while I'm generally in favor of adding this feature, I realized there is no w3c endpoint to get the capabilities from the driver, so I'm ultimately uncomfortable with providing a constructor that does not ensure an equivalent state to the driver that was originally created.

bart-tricentis:
So the DriverCommand.GetSessionCapabilities is a custom thing from selenium itself ? 
Cause when testing it in an Appium context I managed to get a complete list of capabilities.

bart-tricentis:
Any update ?

diemol:
After reading all the comments, it just seems that there is no consensus. Aside from the potential data leak concern, the comments mostly point to recommendations to implement this as an extension/plug-in outside Selenium.

I think that summarises it, and if I am missing something please someone point it out. But it looks like this will not land.

titusfortner:
I'd support it if the w3c spec added a capabilities endpoint. I know it was decided not to have one, but this would be the use case for it. A PR or proposal there would be the place to start I think, but even then, no guarantees.

c4lliope:
> I seem to remember having a conversation about this kind of functionality at Selenium Conf UK in 2016 with someone who wanted this functionality in Ruby.

...though I am not this conference-goer, I also could really use this feature in Ruby.
I had a stab at monkey-patching the function into my code, but as I juggle Selenium and Capybara,
the `bridge` class that I really need access to keeps being obscured by the odd dependency injection going on.
Here's a basic code sample that gets me **as close as can be** to the problem, while achieving precisely zero progress.

At the least, it's a guide to the methods that may need to be subclassed / modified.

```ruby
require "selenium-webdriver"
require "capybara"
require "base64"

bridge = Selenium::WebDriver::Remote::Bridge.new(url: "http://localhost:4444/wd/hub")
# bridge.create_session(Selenium::WebDriver::Remote::Capabilities.firefox)

Capybara.register_driver :session do |app|
  Capybara::Selenium::Driver.new(
    app,
    browser: :remote,
    url: "http://localhost:4444/wd/hub",
    desired_capabilities: Selenium::WebDriver::Remote::Capabilities.firefox,
    bridge: bridge,
  )
end

module Selenium
  module WebDriver
    module Remote
      class Driver
        def initialize(bridge: nil, listener: nil, **opts)
          binding.pry
          desired_capabilities = opts[:desired_capabilities]          
          if desired_capabilities.is_a?(Symbol)
            unless Remote::Capabilities.respond_to?(desired_capabilities)
              raise Error::WebDriverError, "invalid desired capability: #{desired_capabilities.inspect}"
            end

            opts[:desired_capabilities] = Remote::Capabilities.__send__(desired_capabilities)
          end
          opts[:url] ||= "http://#{Platform.localhost}:4444/wd/hub"
          super
        end
      end
    end
  end
end

module Capybara
  module Selenium
    class Driver
      def browser
        unless @browser
          options[:http_client] ||= begin
            require 'capybara/selenium/patches/persistent_client'
            if options[:timeout]
              ::Capybara::Selenium::PersistentClient.new(read_timeout: options[:timeout])
            else
              ::Capybara::Selenium::PersistentClient.new
            end
          end
          processed_options = options.reject { |key, _val| SPECIAL_OPTIONS.include?(key) }
          @browser = ::Selenium::WebDriver::Remote::Driver.new(
            bridge: bridge,
            url: "http://localhost:4444/wd/hub",
            desired_capabilities: Selenium::WebDriver::Remote::Capabilities.firefox,
          )

          specialize_driver
          setup_exit_handler
        end
        @browser
      end
    end
  end
end

page = Capybara::Session.new(:session)
page.visit("https://example.com")
```

This is a sad result, but it's not a deal-breaker for me.
Since my Rails application drives a single Selenium command per web request I receive,
my session driver instance is garbage-collected before I can chain two commands together. Alas.

My second approach could be to spin up a middle process which holds the driver instance, so I would have
`Rails -> long-running process responsible for a session -> Selenium`;
I imagine this is an approach someone could take as an end-run around this feature request,
but there's a "hold on" moment;
as you realize your middle process likely needs to `eval()` code or re-implement the Selenium API.

As a last-ditch effort, I'm looking into serializing and de-serializing my Selenium driver in Ruby;
https://skorks.com/2010/04/serializing-and-deserializing-objects-with-ruby/
http://rubylearning.com/satishtalim/object_serialization.html
https://blog.appsignal.com/2019/03/26/object-marshalling-in-ruby.html
maybe this approach can be equivalent to persisting a session ID,
as it persists an object which holds the session ID.

I'm unsure if I am swaying anyone's opinion here; merely wanted to add my perspective on the issue.
I can see that adding the feature would make many tasks more sensible,
but I also recognize that some workarounds are possible, however unappealing.

Much obliged to all the programmers and coders here on Selenium; I've been using in one form or another for nearly a decade and I can say the web would not be the same without your efforts. 🥇

c4lliope:
That plan was short-lived.
```ruby
[10] pry(main)> serialized = Marshal.dump(page)
TypeError: can't dump TCPSocket
from (pry):16:in `dump'
[11] pry(main)> 
```

diemol:
I'll close this PR as there is no consensous on this change. If someone wants to resume work on this, probably a good starting point is to propose it first to the [W3C WebDriver spec](https://github.com/w3c/webdriver).

PR #8597, opened 2020-08-03
[JS] Support for 'print' endpoint 

harsha509:
### Description

1. Adding support for 'print' endpoint with 'printToPdf' WebDriver client
2. feature works in geckodriver 0.27.0

### Types of changes

### Checklist


harsha509:
Hi @corevo ,

Can you please review this PR. Am not sure if it is the best implementation or not, if not can you please suggest me the best approach.

Also, I've added no tests as am not sure on how to test base64 encoded string. 

Regards,
Harsha

corevo:
You can test it by saving as test fixtures the pdfs of the pages test pages, and read them as base64 and then do a string compare.

We'll have to see over time how flaky it is, we should probably check how takeScreenshot was tested and take a similar approach.

AutomatedTester:
We don't need to test the correctness of the PDF over base64 since that is what browser vendors should be doing in their tests. We just need to test that we are given a base64 string

corevo:
👍

corevo:
Also make sure to test all the options that we can pass, we don't have to check for correctness, that it the vendor's job, but we have to check that we pass them correctly to the driver.

PR #8619, opened 2020-08-09
Update ExpectedConditions.java

ghost:
Making the class constructor public so that this can be extended and custom Expected Conditions can be defined

### Types of changes

### Checklist

### DIFF
--- a/java/client/src/org/openqa/selenium/support/ui/ExpectedConditions.java
+++ b/java/client/src/org/openqa/selenium/support/ui/ExpectedConditions.java
@@ -44,7 +44,7 @@ public class ExpectedConditions {
 
   private static final Logger log = Logger.getLogger(ExpectedConditions.class.getName());
 
-  private ExpectedConditions() {
+  public ExpectedConditions() {
     // Utility class
   }
 


shs96c:
The `ExpectedConditions` class is a utility class, and that is why it has a `private` constructor. Extending this doesn't offer much value, since it's impossible to override `static` methods. For this reason, I'm going to close the PR, but I'd like to thank you for taking the time to put this together. It's very much appreciated.

Presumably, you're wanting to extend `ExpectedConditions` so that you have a single place to add new conditions for use within your own code? May I suggest creating that class, and then taking only the methods from `ExpectedConditions` you actually use and delegating those? Something like:

```java
public MyConditions {
    // Example delegation
    public static ExpectedCondition titleContains(String title) {
      return ExpectedConditions.titleContains(title);
    }

    // Hypothetical example your-custom-condition
    public static ExpectedCondition showsCurrentPrice(WebElement element, Money price) {
      return driver -> element.getText().contains(price.toString());
    }
}
```

ghost:
Hello Simon, 

Thanks for the update and implementation ideas

Adithya

PR #8630, opened 2020-08-14
Improve the error messages output by some ExpectedConditions.

eschoeffler:
### Description
attributeContains was incorrectly reporting the current value as "null". This
change fixes that.
textToBePresentInElement was not reporting the actual element text. This change
adds the actual element text in the failure message.

### Motivation and Context
I was writing selenium tests and having trouble debugging them because of the poor output of the existing ExpectedConditions logic. This change provides more useful output for failing tests.

### Types of changes
Fixes bug with attributeContains condition error message.
Adds more useful error message for testToBePresentInElement condition.

### Checklist

### DIFF
--- a/java/client/src/org/openqa/selenium/support/ui/ExpectedConditions.java
+++ b/java/client/src/org/openqa/selenium/support/ui/ExpectedConditions.java
@@ -348,10 +348,12 @@ public class ExpectedConditions {
                                                                     final String text) {
 
     return new ExpectedCondition<Boolean>() {
+      private String elementText = null;
+
       @Override
       public Boolean apply(WebDriver driver) {
         try {
-          String elementText = element.getText();
+          elementText = element.getText();
           return elementText.contains(text);
         } catch (StaleElementReferenceException e) {
           return null;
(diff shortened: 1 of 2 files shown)


java/client/src/org/openqa/selenium/support/ui/ExpectedConditions.java
```diff
@@ -348,10 +348,12 @@ public String toString() {
                                                                     final String text) {
 
     return new ExpectedCondition<Boolean>() {
+      private String elementText = null;
```
shs96c, line 351: Each `ExpectedCondition` implements `java.util.Function` These are expected to be stateless. This condition will leak previous `elementText` on the second usage, which doesn't seem ideal.

shs96c (changes requested):
Presumably, this is because you're doing matching against a very long piece of text (maybe the body element?) and the output of the existing class is hard to read. However, some people rely on the existing functionality, so eliding in all cases isn't correct. 

I'd suggest having an (optional) parameter for a maximum string length, preserving the existing signatures to ensure that people updating don't need to recompile their code.

diemol (commented):
@eschoeffler would you like to follow the suggestion in the previous review? Please let us know.

diemol:
Closing this PR as we did not get more input from the OP.

PR #8640, opened 2020-08-22
[PY] Added the ability to run Brython code

Ahsoka:
### Description

For the uninitiated [Brython] is a way to use Python code on the front-end instead of Javascript.  This pull-request adds a `execute_brython()` method which allows you to execute [Brython] code much in the same way you would use the `execute_script()` method to execute JavaScript code.  There is also now a `load_brython()` method which is used to load the necessary JavaScript files to run [Brython] code and must be called before using the `execute_brython()` method.

### Examples
Here are a couple of examples of using the `load_brython()` and `execute_brython()` methods.
#### Example 1
```python
>>> driver = webdriver.Chrome()
>>> # You must use `load_brython()` before using `execute_brython()`
>>> driver.load_brython()
>>> driver.execute_script("console.log('Hello World');")
>>> # Equivalent using `execute_brython()`
>>> driver.execute_brython("print('Hello World')")
```

#### Example 2
```python
>>> driver = webdriver.Chrome()
>>> driver.load_brython()
>>> # turtle_script is defined below
>>> driver.execute_brython(turtle_script)
```
This is `turtle_script` ⬇
```python
# Adapted from https://brython.info/gallery/turtle.html
from browser import document, html
import turtle
if 'turtle-div' not in document:
    document <= html.DIV(Id='turtle-div', style={'float':'left', 'border': '1px solid green'})
turtle.set_defaults(
    turtle_canvas_wrapper = document['turtle-div']
)
t = turtle.Turtle()

t.width(5)

for c in ['red', '#00ff00', '#fa0', 'rgb(0,0,200)']:
    t.color(c)
    t.forward(100)
    t.left(90)

# dot() and write() do not require the pen to be down
t.penup()
t.goto(-30, -100)
t.dot(40, 'rgba(255, 0, 0, 0.5')
t.goto(30, -100)
t.dot(40, 'rgba(0, 255, 0, 0.5')
t.goto(0, -70)
t.dot(40, 'rgba(0, 0, 255, 0.5')

t.goto(0, 125)
t.color('purple')
t.write("I love Brython!", font=("Arial", 20, "normal"))


turtle.done()
```
Result of **Example 2** ⬇

data_, - Google Chrome 2020-08-22 16-44-07_Trim

### Types of changes

### Checklist

[Brython]: https://github.com/brython-dev/brython


Ahsoka:
I am not sure how I would go about adding documentation.  I have added a docstring to both of the new methods I created.   Does this satisfy the `I have updated the documentation accordingly.` requirement?  Please let me know.

py/selenium/webdriver/remote/webdriver.py
```diff
@@ -741,6 +744,59 @@ def execute_async_script(self, script, *args):
             'script': script,
             'args': converted_args})['value']
 
+    def load_brython(self,
+                     src="https://cdn.jsdelivr.net/npm/brython@3.8.9/brython.min.js",
```
AutomatedTester: Are you sure Brython users would always be in the internet? Who not bundle as `data` like some of the other JS packages that are shipped

py/selenium/webdriver/remote/webdriver.py
```diff
@@ -741,6 +744,59 @@ def execute_async_script(self, script, *args):
...
+
+        :Usage:
+            ::
+
+                driver.execute_brython('print(\"Hello World! :)\")')
+
+        :Returns:
+          The command's JSON response loaded into a dictionary object, if any.
+        """
+        if not hasattr(self, 'loaded_brython') or not self.loaded_brython:
```
AutomatedTester, line 819: Instead of erroring, wouldn't it be better to try to execute? If it's not loaded then load it implicitly

py/selenium/webdriver/remote/loadBrython.js
```diff
@@ -0,0 +1,4 @@
+var script = document.createElement('script')
```
AutomatedTester: These files should in `/javascript/brython`

AutomatedTester (changes requested):
I don't mind adding support for Brython but ongoing it might break and there are no tests in this patch to make sure that never happens.

Would it not be better to have a package that you own that has a dependency on Selenium that you ship?

py/selenium/webdriver/remote/loadBrython.js
Ahsoka: When you refer to `/javascript/brython` do you mean `py/selenium/webdriver/remote/javascript/brython` or the top level directory `/javascript/brython`.

py/selenium/webdriver/remote/webdriver.py
Ahsoka: ~I initially thought the same as you do, however, for reasons I do not fully understand, I am unable to load it implicitly.  Every time I try to load Brython in the `execute_brython()` method (by calling `load_brython()`) I keep getting this error `selenium.common.exceptions.JavascriptException: Message: javascript error: brython is not defined`.  This means that the `brython()` function I use to actually run the Brython code is not defined, even though it's supposed to be defined after calling `load_brython()`.  But strangely enough, it seems that when I separate the **loading phase** from the **executing phase**, (by using `load_brython()` and `execute_brython()`, respectively) this issue goes away.  Any pointers on a possible fix would be greatly appreciated.~

Ahsoka: Never mind please ignore what I just said. I figured out a way to fix the issue.  See [this commit](https://github.com/SeleniumHQ/selenium/pull/8640/commits/8d602462141b1fa3efb4c4f83b0650374fb59763) for reference.

py/selenium/webdriver/remote/webdriver.py
Ahsoka: I also thought of something similar to what you suggested but unfortunately it can be unreliable.  For example, in Chrome whenever I try to access local files I get the follow error `Not allowed to load local resource: {file name here}`. However, I believe in Firefox it would work with no problem (I have not tested this but I have reason to believe this to be the case from the exhaustive amount of research I have conducted on this topic).  Additionally, unless configured in a certain way, Brython has a reliance on certain external files anyway.  Loading Brython from a file would still require access to the external files via the internet.

Ahsoka:
I added a single test that will pretty reliable test whether or not Brython works. See [these changes](https://github.com/SeleniumHQ/selenium/pull/8640/files#diff-59b2b721525dd9c5db2c71da681eb9b0) for reference,

py/selenium/webdriver/remote/loadBrython.js
AutomatedTester: top level `/javascript/brython` and add a bazel file or if you want to use npm then we can add it to the `package.json` and I will help with the bazel stuff .

py/selenium/webdriver/remote/webdriver.py
AutomatedTester: We need to use bazel here. I will help with this part.

py/selenium/webdriver/remote/loadBrython.js
Ahsoka: Alright I added the files to the top level directory `/javascript/brython`.  See [this commit](https://github.com/SeleniumHQ/selenium/pull/8640/commits/8e672b42913146984542719cc76b2c2541e9c023) for reference.  I am not too familiar with Bazel so please let me know what I need to do to reference the files in my code.

As of now they are currently referenced like this:

See the below [here](https://github.com/TurretAA12/selenium/blob/brython/py/selenium/webdriver/remote/webdriver.py#L776-L777).
```python
_pkg = '.'.join(__name__.split('.')[:-1])
template = pkgutil.get_data(_pkg, 'loadBrython.js').decode('UTF-8')
```

and like this
```python
_pkg = '.'.join(__name__.split('.')[:-1])
running = pkgutil.get_data(_pkg, 'runBrython.js').decode('UTF-8').format(repr(script))
```
See the above [here](https://github.com/TurretAA12/selenium/blob/brython/py/selenium/webdriver/remote/webdriver.py#L821-L823).

Ahsoka:
@AutomatedTester Whenever you get a chance please let me know what the next steps are in getting this PR ready to merge. Thank you!

Ahsoka:
@AutomatedTester Any updates on this PR?

Ahsoka:
Hey @AutomatedTester, I just wanted to follow up again to see if there are any plans to merge this PR?  If not, please let me know so I can stop bothering you and close this PR. 🙃

AutomatedTester:
> Hey @AutomatedTester, I just wanted to follow up again to see if there are any plans to merge this PR? If not, please let me know so I can stop bothering you and close this PR. 🙃

I need to find time to fix the build system for this PR. I am hoping soon

Ahsoka:
Awesome, thank you so much for getting back to me.  If there is anything I can do, please let me know! 😁

Ahsoka:
Any updates on this?

PR #8647, opened 2020-08-25
Added new argument creationflags in Service class for common, chrome, and firefox

ali-sajjad-rizavi:
This feature was only needed for windows users. There will be no need to do this in Unix like operating systems.
Let's suppose we have a GUI program (.pyw) which uses selenium.
```

driver = webdriver.Chrome()
```
When the `subprocess.Popen(...)` opens the chromedriver, a new console window will open up on windows system. To avoid this, we can use `CREATE_NO_WINDOW` creation flag (or possibly some other flags). This will call `subprocess.Popen(...,creationflags=self.creationflags)`.

After the added feature, we will be able to do something like this:
```
from selenium.webdriver.chrome.service import Service as ChromeService # Similar thing for firefox also!
from subprocess import CREATE_NO_WINDOW # This flag will only be available in windows

chrome_service = ChromeService('chromedriver', creationflags=CREATE_NO_WINDOW)
driver = webdriver.Chrome(service=chrome_service) # No longer console window opened, niether will chromedriver output
```

I had done this directly inside the WebDriver class but the seniors asked to do this inside `Service` or `Options` class. So here I am again! Anyway, I am open to alternatives you suggest and commit again, and eventually I will do better code contribution after learning the structure of project more.

**Question:** If this is okay, can I also proceed with the same strategy for other browsers as well?


diemol (changes requested):
@ali-sajjad-rizavi I think this is the same PR you submitted a while ago, and by having a look at the code, the requested changes are probably the same ones (to add this in the options class)
https://github.com/SeleniumHQ/selenium/pull/8547

ali-sajjad-rizavi:
@diemol The changes were not the same, actually what I did before was to add an argument like `webdriver.Chrome(create_no_window=True)`. Then I was told to do this in Options or Service. Now I have managed that we make the use of `Service` class to give `creationflags` to the chromedriver process.

So, if you suggest some other way which should be used, I will work on it.

diemol:
Well, I was reading the comments in the other PR and they suggest to implement this as part of the **options**, that is why I mention it here. Changing the service is changing the constructor in different classes.

ali-sajjad-rizavi:
Actually the `Options` class does not seem to have some good communication with the `Service` class which creates the chromedriver process.

diemol:
@AutomatedTester do you have comments for this PR?

py/selenium/webdriver/chrome/service.py
```diff
@@ -24,7 +24,7 @@ class Service(service.ChromiumService):
     """
 
     def __init__(self, executable_path, port=0, service_args=None,
-                 log_path=None, env=None):
+                 log_path=None, env=None, creationflags=0):
```
AutomatedTester: I would rather avoid having a new item added to the method signature. Could you have this as a property in these classes and then when `start` is called we check if it is populated and used.

Also, there is no documentation added in the docstring for these changes. It does not explain what the 0 means to people who may come across it.

ali-sajjad-rizavi: The `0` is the default value of a Python subprocess, maybe using a comment will be helpful.

I assume that after doing what you asked, the above sample code (to use this new feature), will become like this:
```
from selenium.webdriver.chrome.service import Service as ChromeService # Similar thing for firefox also!
from subprocess import CREATE_NO_WINDOW # This flag will only be available in windows

chrome_service = ChromeService('chromedriver')
chrome_service.creationflags = CREATE_NO_WINDOW # We change the property in Service class, default was 0
driver = webdriver.Chrome(service=chrome_service)
```
Is that what you mean? If yes, then I will proceed with this work:
- Add a property `self.creationflags=0` in the `Service` class constructor.

AutomatedTester: That's what I mean. The other `__init__` signatures grew when I wasn't watching and it became messy. I would rather try keep them clean if possible

ali-sajjad-rizavi: One more thing I would like to mention. When the `subprocess.Popen(...)` is called, there is a default argument `creationflags=0` in the `Popen(...)` function in the official python subprocess module. (Even if we do not provide this argument!)

ali-sajjad-rizavi:
@AutomatedTester I have made the changes like you asked, and this really seems much better! As minimal change was required (1 file only, although we need to manually set the `creationflags` attribute of `Service` class, but that is not too bad).
### "How to use" for this feature
```
from selenium.webdriver.chrome.service import Service as ChromeService # Similar thing for firefox also!
from subprocess import CREATE_NO_WINDOW # This flag will only be available in windows

chrome_service = ChromeService('chromedriver')
chrome_service.creationflags = CREATE_NO_WINDOW # We change the property in Service class, default was 0
driver = webdriver.Chrome(service=chrome_service)
```
There is some similar thing which existed in C# but not in Python.

AutomatedTester (approved):

PR #8652, opened 2020-08-27
Support cookie SameSite=None option on nodejs

n2o1988:

### Description
Allows cookies to be added with `SameSite` set to "None" in addition to "Lax" and "Strict", as per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite.
This option is only valid if the cookie is `Secure` (see https://web.dev/samesite-cookies-explained/#samesitenone-must-be-secure and [proposal](https://tools.ietf.org/html/draft-west-cookie-incrementalism-00)).

### Motivation and Context
Follow up to https://github.com/SeleniumHQ/selenium/pull/7901
[Recent updates](https://web.dev/samesite-cookies-explained/#changes-to-the-default-behavior-without-samesite) to Chrome 85 (and soon landing on Firefox) changed the default from `None` to `Lax`, requiring us to _explicitly_ set `sameSite: 'None'`, which the [original implementation](https://github.com/SeleniumHQ/selenium/pull/7901/files#diff-8d08e47355ecd3bd4b7f70008552ebdaR1134-R1136) does not allow (not sure why, maybe `None` wasn't an option, back then?).
This also mirrors what the Java bindings support: https://github.com/SeleniumHQ/selenium/blob/trunk/java/client/src/org/openqa/selenium/Cookie.java#L111
Other bindings (such as Python) have the same issue and should be updated too (I'm not a Python dev though).

### Types of changes

### Checklist


n2o1988:
There's an existing failing test (at least on my machine)
> 1) [chrome] can send commands to devtools and get return
  Message:
    TypeError: driver.sendDevToolsCommandAndGetReturn is not a function

This error occurs in the `trunk` branch too, so nothing related to this PR. 

I'm not sure why the Lint check failed, I use an ESLint extension in VSCode that's supposed to auto format everything according to the repo's rules. The error doesn't seem related to my change.

javascript/node/selenium-webdriver/lib/webdriver.js
```diff
@@ -1449,10 +1455,10 @@ Options.Cookie.prototype.expiry
  * When the cookie applies to a SameSite policy.
  *
  * When {@linkplain Options#addCookie() adding a cookie}, this may be specified
- * as a {@link string} object which is either 'Lax' or 'Strict'.
+ * as a {@link string} object which is one of 'Lax', 'Strict' or 'None'.
  *
  *
- * @type {(!Date|number|undefined)}
+ * @type {(string|undefined)}
```
n2o1988, line 1461: I think this was a copy-paste mistake in the original PR.

javascript/node/selenium-webdriver/test/cookie_test.js
```diff
@@ -212,6 +212,53 @@ suite(function (env) {
...
+      async function () {
+        let cookie = createSameSiteCookieSpec('None', {
+          secure: true,
+        })
+        let childUrl = fileserver.whereIs('child/childPage.html')
+        await driver.get(childUrl)
+        await driver.manage().addCookie(cookie)
+        await assert.doesNotReject(
+          async () => await driver.manage().addCookie(cookie)
+        )
```
n2o1988, line 227: Other test cases for `SameSite` take a different approach to test this: they set the cookie and read it straight after.
In case of `None`, the cookie has to be `secure`, and I found this means the `getCookie` call would not return any cookie (since the test page in localhost is not served via `https` I presume?). 

I figured the best way to test this is to test that it does **not** throw an exception. In fact, an invalid combination (see below tests) would cause the underlying driver to raise an exception. 
This sounds sufficient to me but let me know if this is not enough.

I did make sure that this works in our real test suites (where the app is served on https).

n2o1988 (commented):
A couple of clarifications

javascript/node/selenium-webdriver/test/cookie_test.js
```diff
@@ -212,6 +212,53 @@ suite(function (env) {
         assert.equal(actualCookie.sameSite, 'Lax')
       }
     )
+
+    ignore(env.browsers(Browser.FIREFOX, Browser.IE, Browser.SAFARI)).it(
```
AutomatedTester: Shouldn't these tests work in Firefox?

javascript/node/selenium-webdriver/test/cookie_test.js
AutomatedTester: I think that this is fine

javascript/node/selenium-webdriver/test/cookie_test.js
```diff
@@ -212,6 +212,53 @@ suite(function (env) {
...
+          async () => await driver.manage().addCookie(cookie),
+          {
+            name: 'InvalidArgumentError',
+            message: `Invalid cookie configuration: SameSite=None must be Secure`,
+          }
+        )
+      }
+    )
+
+    ignore(env.browsers(Browser.FIREFOX, Browser.IE, Browser.SAFARI)).it(
```
AutomatedTester: Shouldn't this work in Firefox?

javascript/node/selenium-webdriver/lib/webdriver.js
```diff
@@ -1177,9 +1177,15 @@ class Options {
       expiry = Math.floor(date.getTime() / 1000)
     }
```
KevinBon, line 1179: Based on the [related](https://w3c.github.io/webdriver/#dfn-cookie-same-site) webdriver specification
> Defaults to None if omitted when adding a cookie.

We should set sameSite cookie by default as "None" when omitted.

```suggestion
if (!sameSite) {
 sameSite = 'None'
}
```

javascript/node/selenium-webdriver/lib/webdriver.js
```diff
@@ -1177,9 +1177,15 @@ class Options {
...
+    if (sameSite && !['Strict', 'Lax', 'None'].includes(sameSite)) {
       throw new error.InvalidArgumentError(
-        `Invalid sameSite cookie value '${sameSite}'. It should be either "Lax" (or) "Strict" `
+        `Invalid sameSite cookie value '${sameSite}'. It should be one of "Lax", "Strict" or "None"`
+      )
+    }
+
+    if (sameSite === 'None' && !secure) {
+      throw new error.InvalidArgumentError(
+        `Invalid cookie configuration: SameSite=None must be Secure`
```
KevinBon: It helps further understanding that `None` is a string and we keep a consistent sameSite formatting
```suggestion
        'Invalid cookie configuration: sameSite="None" must be Secure'
```

javascript/node/selenium-webdriver/test/cookie_test.js
n2o1988: I mirrored the original tests, didn't want to change too much. I'll double-check. Maybe it wasn't an option at the time of the original implementation but it is now.

javascript/node/selenium-webdriver/lib/webdriver.js
n2o1988: I'm not too sure we should do that given that: 
* no other Selenium binding is doing that: all seem to leverage on the underlying driver/browser implementation for the default value
* I think it's the Browsers' responsibility to define the default behavior. For instance, in Chrome 85 that's exactly what changed: [default is now](https://web.dev/samesite-cookies-explained/#changes-to-the-default-behavior-without-samesite) `Lax` for Chrome and soon for Firefox too. That change was the reason I opened this PR in the first place.
* with the new security proposal, `sameSite='None'` should always be accompanied by `Secure=true`, so we'd need to take that into account
* changing that would be a breaking change (nothing against it, but I'd like the maintainers' opinion)

Interested to hear thoughts from @AutomatedTester

javascript/node/selenium-webdriver/lib/webdriver.js
n2o1988: I was following the cookie header format (e.g. `SameSite=None; Secure; `). 
I can change that if you feel strongly about it, but then it should be `secure: true` instead of `Secure`.

javascript/node/selenium-webdriver/test/cookie_test.js
n2o1988: Done.

javascript/node/selenium-webdriver/test/cookie_test.js
n2o1988: You're right, updated.

n2o1988:
@AutomatedTester any other thoughts on this one?

AutomatedTester:
This looks like it is about ready to land. There are linting errors. Could you fix those please and I will get it landed.

n2o1988:
> This looks like it is about ready to land. There are linting errors. Could you fix those please and I will get it landed.

@AutomatedTester the linting errors seem unrelated to my change.
I think you the same errors show up in the `trunk` branch?

It also seems like this "Super-Linter" has conflicting rules with the ones declared in the [.eslintrc.js config](https://github.com/SeleniumHQ/selenium/blob/trunk/javascript/node/selenium-webdriver/.eslintrc.js)
Example: `prettier` wants you to have a trailing comma for arrays on multiple lines. 
The "Super-Linter" check on Github [says the opposite](https://github.com/SeleniumHQ/selenium/pull/8652/checks?check_run_id=1041062300#step:4:84). 
> /github/workspace/javascript/node/selenium-webdriver/lib/webdriver.js:45:28: Unexpected trailing comma.

diemol:
@n2o1988 I just ran the linter again on `trunk` and it is passing. Somehow the changes to the files in this PR are making the linter fail. It'd be nice if you can have a look 🙏

n2o1988:
@diemol maybe I'm doing this wrong. 
How do I run that linter locally? 

My linting config says everything is good, locally (at least according to `.eslintrc.js`).

diemol:
@n2o1988 I did not do it locally, I only ran the GitHub action job again on `trunk`.

n2o1988:
@diemol it's kinda hard to spot what's wrong in this case. 

If you look at the output of the linter, it shows a whole bunch of things I haven't changed, and the rules it's using are conflicting with the rules specified by ESLint 
image

I never touched those lines.. :|

AutomatedTester:
Let me have a look at the linter stuff since I set it up

n2o1988:
@AutomatedTester any updates on merging this PR?

n2o1988:
@AutomatedTester @diemol  🆙 

We're sort of in a hurry to get this released, otherwise we'll have to go the way of pinning our own fork (which I'd like to avoid). 
Any chance this can get merged & released anytime soon?

n2o1988:
Up: any updates on merging this PR @AutomatedTester @diemol ?

diemol (approved):

diemol:
Merging this, thank you for your contribution @n2o1988!

However, I cannot say when the next alpha release will happen.

n2o1988:
Thanks @diemol ! 
Is there a release tracker page or a way to get notified when the next alpha is released?

Issue #8159, opened 2020-04-03
[4.0.0-alpha5] Selenium creates never ending Thread


carsonwah:
## 🐛 Bug Report

After simply starting and quiting the webdriver, the browser started and quitted normally. But just the program never ends after running all the steps. I found that there is dead thread (AsyncHttpClient) left running.

## Detail steps

My simple program:
```java
import org.openqa.selenium.WebDriver;
import org.openqa.selenium.firefox.FirefoxDriver;
public class SeleniumTestChromiumEdge {
    public static void main(String[] args) {
        WebDriver driver = new FirefoxDriver();
        System.out.println("Before quit");
        driver.quit();
        System.out.println("After quit");
    }
}
```

Callstack before `WebDriver driver = new xxxxDriver();`:
![image](https://user-images.githubusercontent.com/25053707/78382173-fa565800-7608-11ea-9a83-37ca6d5acbc6.png)

Callstack after `WebDriver driver = new xxxxDriver();`:
![image](https://user-images.githubusercontent.com/25053707/78382373-57eaa480-7609-11ea-8ccc-49e86a2158ab.png)

Callstack after `driver.quit();`:
![image](https://user-images.githubusercontent.com/25053707/78382958-30480c00-760a-11ea-9aa2-8de5a9b29014.png)

After program ends:
![image](https://user-images.githubusercontent.com/25053707/78383684-5b7f2b00-760b-11ea-8006-b0e7d8727535.png)

Console output: (e.g. Firefox driver)
```
1585930423908   mozrunner::runner       INFO    Running command: "C:\\Program Files\\Mozilla Firefox\\firefox.exe" "-marionette" "-foreground" "-no-remote" "-profile" "C:\\Users\\TSZCHU~1\\AppData\\Local\\Temp\\rust_mozprofilezIRvii"
1585930424530   addons.webextension.doh-rollout@mozilla.org     WARN    Loading extension 'doh-rollout@mozilla.org': Reading manifest: Invalid extension permission: networkStatus
1585930425313   addons.webextension.screenshots@mozilla.org     WARN    Loading extension 'screenshots@mozilla.org': Reading manifest: Invalid extension permission: mozillaAddons
1585930425314   addons.webextension.screenshots@mozilla.org     WARN    Loading extension 'screenshots@mozilla.org': Reading manifest: Invalid extension permission: telemetry
[...]

PR #8701, opened 2020-09-15
update NettyClient to close AsyncHttpClient when finished

Bioshock667:

### Description
This change will make so that the AsyncHttpClient will be closed when "WebDriver.close" or "WebDriver.quit" is called.

### Motivation and Context
The is a quick fix for issue #8159 where a Java program will fail to terminate because of a hanging thread even if the "quit" method is called.  [AsyncHttpClient](https://github.com/AsyncHttpClient/async-http-client) requires that the user calls the "close" method before exiting the program or else "you'll experience threads hanging and resource leaks".  To fix this, the "cleanupIdleClients" method has been overridden to perform the needed action.

### Types of changes

### Checklist

### DIFF
--- a/java/client/src/org/openqa/selenium/remote/http/netty/NettyClient.java
+++ b/java/client/src/org/openqa/selenium/remote/http/netty/NettyClient.java
@@ -32,10 +32,14 @@ import org.openqa.selenium.remote.http.HttpRequest;
 import org.openqa.selenium.remote.http.HttpResponse;
 import org.openqa.selenium.remote.http.WebSocket;
 
+import java.io.IOException;
 import java.util.function.BiFunction;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 public class NettyClient implements HttpClient {
 
+  private static final Logger log = Logger.getLogger(NettyClient.class.getName());
   private static final AsyncHttpClient httpClient = Dsl.asyncHttpClient(
     new DefaultAsyncHttpClientConfig.Builder()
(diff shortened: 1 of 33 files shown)


java/client/src/org/openqa/selenium/remote/http/netty/NettyClient.java
```diff
@@ -86,5 +90,14 @@ public HttpClient createClient(ClientConfig config) {
 
       return new NettyClient(new NettyHttpHandler(config, httpClient).with(config.filter()), NettyWebSocket.create(config, httpClient));
     }
+
+    @Override
+    public void cleanupIdleClients() {
+      try {
+        httpClient.close();
```
shs96c, line 97: The `httpClient` is a `static` field. In the case of a regular selenium test, closing this might be okay, but we also use this in the Grid, and in situations where more than one http client is being used. Surely this will close other `NettyClient` implementations in the same JVM? Or do I misunderstand something?

shs96c:
This looks like a good idea, but I worry about the impact on concurrent users of different instances of `NettyClient`. Once we know this is safe, I'm keen to land this. Thank you for taking the time to prep the PR! It's very much appreciated.

Bioshock667:
No problem!  Thank you for the feedback!  Hopefully a thread-safe way to close AysncHttpClient will be found.

Bioshock667:
I will close this PR but when a more stable solution comes up. I will make another one.

Issue #8525, opened 2020-07-15
Unable to start/stop drivers concurrently on Server


titusfortner:
## 💥 Regression Report

Getting errors when starting more than one driver session at a time from different Threads. ~~Extra browsers are started, so I'm not sure if it failing on starting the 3rd one or if it breaks when trying to quit the session.~~ The server appears to lose track of one session after another has been started.

## Last working Selenium version
3.141.59

## Stopped working in version:
4.x

## To Reproduce
This is the broken spec:
https://github.com/SeleniumHQ/selenium/blob/dd7090cab54beb2be2541a08746ddaeb8783c496/rb/spec/integration/selenium/webdriver/spec_support/shared_examples/concurrent_driver.rb

`./go //rb:remote-chrome-test`

It's the same failure/stack track fro both firefox and chrome - https://travis-ci.org/github/titusfortner/selenium/jobs/708131143#L718

## Expected behavior

More than one browser starts and then is correctly closed

## Actual behavior

Here's the log of requests for sessions and the server not recognizing the sessions:
https://gist.github.com/titusfortner/a4076e1f3dc8ef4115acc7a84fde9bce

It includes messages like:
`invalid session id` and `NoSuchSessionException`

## Environment

OS: All
Browser: All
Language Bindings version: Ruby trunk
Selenium Grid version (if applicable): trunk

AutomatedTester:
I think that you're going to need to help us investigate it more. 

Bazel runs tests in parallel by default so wondering if there is something else  going on here.

titusfortner:
I should have at least included logs when I filed this, so I added a gist above.

The test is starting the sessions from different threads.

Note that it is Chromedriver complaining about the session id, so it's like the server is mixing up which Session ID is associated with which driver.

Interestingly, when I use the most recently created session first, that one seems to work properly, but then the next two fail with the same `invalid session id`: https://gist.github.com/titusfortner/d8896b8fbd1d3ca2ceb9d55411af8aa9

barancev:
How to reproduce the issue? I can't find `rb_server_toggle` branch on github and I can't see it merged to trunk.

barancev:
Ah, sorry, it's a PR.

titusfortner:
Yes, it got merged in; It can be replicated with: `./go //rb:remote-chrome-test`
[...]

PR #8713, opened 2020-09-18
Add withExecutor to FluentWait to support concurrent drivers in custom thread pools

bjuric:
### Description
`FluentWait` blocked and behaved unpredictably when called concurrently from multiple web driver threads executing in a custom thread pool. This commit adds the ability to pass the Executor service that manages the thread pool to fluent wait.

#### Sample usage
```java 
  // pass custom executor to fluent wait
  new FluentWait(driver)
      .withTimeout(Duration.ofSeconds(10))
      .withExecutor(myCustomExecutor)
      ..
```

### Motivation and Context
Many tools and frameworks built on Selenium offer options for running multiple web driver instances in parallel threads. Fluent wait can block and behave unpredictabily  under some conditions where concurrent calls are made by threads in custom `Executor`s. In such instances, it becomes necessary to use the custom Executor with `FluentWait`'s underlying `CompletableFuture`.

### Types of changes

### Checklist


bjuric:
It's worth noting that launching Selenium with the following Java system property forces the CompletableFuture in FluentWait to run on the calling thread.
```
java.util.concurrent.ForkJoinPool.common.parallelism=0
```
There's no need to explicitly set the executor if you specify this property. This solves the concurrency issue with thread contention but also affects all instances where CompletableFutures are used throughout Selenium in that they would no longer run on separate threads. In some cases however, this might be a viable alternative to the withExecutor approach introduced in this PR. I guess it's good to have options.

thegoldgoat:
Hello, not an expert here

This is related to #8727 isn't it?

bjuric:
@thegoldgoat I think that one has the same underlying issue. I believe it to be a systemic issue in that CompleteableFutures are used throughout Selenium 4, and not just in FluentWait.

thegoldgoat:
Thanks for the response!

Again, not an expert of this codebase

Is this PR ready to be tested by any chance?

Should I change something in my code?

Thanks again!

bjuric:
@thegoldgoat try launching your app with the jvm option in my earlier comment above and use a custom thread pool. See also #8525

thegoldgoat:
Great! Thanks a lot I'll let you know if that works

thegoldgoat:
Hi, sorry for the late response.

Using that options everything works, thanks a lot!!

I hope the devs would investigate in this bug and consider your solution

chrisgleissner:
What is blocking this PR from being merged?

barancev:
We discuss the current implementation. And tend to revert back to legacy (blocking) implementation.

barancev:
Sorry to say this, but we decided to return back to the legacy (blocking) implementation after clients' complains. It appears that many people use suboptimal approach to manage WebDriver instances using ThreadLocal variables. And running conditions in a separate thread makes them lose their driver. So we decided to revert this change in favour of backward compatibility.

Issue #4847, opened 2017-10-11
Selenium Java API does not conceal Action protocol differences with moveToElement(element,xOffset,yOffset)


pbi-qfs:
## Meta -
OS:  Any
Selenium Version:  3.6.0
Browser:  Chrome, Firefox

## Expected Behavior -

Selenium should do what its Javadoc says, and it should be consistent independently from the underlying wire protocol: `moveToElement(element,xOffset,yOffset)` moves the mouse to an offset from the top-left corner of the element.

## Actual Behavior -

It prints out "INFORMATION: When using the W3C Action commands, offsets are from the center of element" and moves the cursor differently than stated in the method doc

## Steps to reproduce -

```java
final boolean useW3C = true;
final WebDriver driver = useW3C ? new FirefoxDriver() : new ChromeDriver();

driver.get("https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_event_mouse_clientxy");

driver.switchTo().frame("iframeResult");

final WebElement header = driver.findElements(By.tagName("h2")).get(0);


final int xOffset = 10;
final int yOffset = 5;

final Actions builder = new Actions(driver);
builder.moveToElement(header, xOffset, yOffset); // Javadoc: Moves the mouse to an offset from the top-left corner of the element.
builder.click();
builder.perform();

final Point location = header.getLocation();
final String expectedResult = String.format("X coords: %d, Y coords: %d",
        location.x + xOffset,
        location.y + yOffset);

final String result = driver.findElement(By.id("demo")).getText();

if (expectedResult.equals(result)) {
    System.out.println("fine.");
} else {
    System.out.format("not ok: '%s' instead of '%s'\n", result, expectedResult);
};
```

Jarob22:
In the w3c spec (https://w3c.github.io/webdriver/webdriver-spec.html#pointer-actions) it says:

Let x element and y element be the result of calculating the in-view center point of element.

In subsection "When required to dispatch a pointerMove action with arguments source id, action object, input state and tick duration a remote end must run the following steps"

So the docs are wrong.

pbi-qfs:
The problem with moveToElement+offset is the spec difference between https://github.com/seleniumhq/selenium/wiki/jsonwireprotocol#sessionsessionidmoveto and https://w3c.github.io/webdriver/webdriver-spec.html#pointer-actions: 

[...]

PR #8722, opened 2020-09-22
Fix for documentation of ‘Actions.moveToElement’

Rinzwind:
Pull request as suggested by @diemol ( https://github.com/SeleniumHQ/selenium/issues/4847#issuecomment-691981122 ).

This replaces “center of the element” in the documentation of `Actions.moveToElement` with “element‘s in-view center point” to match the W3C spec.

The section on “pointer actions” in the spec says the offsets are from the element’s in-view center point, not its center:

WebDriver 2020-09-22 12-05-25

The in-view center point is defined as follows:

WebDriver 2020-09-22 12-08-06

### DIFF
--- a/java/client/src/org/openqa/selenium/interactions/Actions.java
+++ b/java/client/src/org/openqa/selenium/interactions/Actions.java
@@ -366,13 +366,12 @@ public class Actions {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


diemol (approved):
Thank you @Rinzwind!

PR #8723, opened 2020-09-22
Add register function to handle Basic Auth

raju249:

### Description

Add `register` function to handle BasicAuth in js binding.
Use CDP to perform authorization

### Types of changes

### Checklist

### DIFF
--- a/java/server/test/org/openqa/selenium/grid/sessionqueue/NewSessionQueuerTest.java
+++ b/java/server/test/org/openqa/selenium/grid/sessionqueue/NewSessionQueuerTest.java
@@ -18,6 +18,7 @@
 package org.openqa.selenium.grid.sessionqueue;
 
 import com.google.common.collect.ImmutableMap;
+
 import org.junit.Before;
 import org.junit.Test;
 import org.openqa.selenium.Capabilities;
@@ -62,10 +63,12 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import static java.net.HttpURLConnection.HTTP_INTERNAL_ERROR;
 import static java.net.HttpURLConnection.HTTP_OK;
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.fail;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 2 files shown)


javascript/node/selenium-webdriver/chromium.js
```diff
@@ -738,20 +738,28 @@ class Driver extends webdriver.WebDriver {
...
-      this.sendDevToolsCommand('Fetch.continueWithAuth', {
-        requestId: params['requestId'],
-        authChallengeResponse: {
-          response: 'ProvideCredentials',
-          username: username,
-          password: password
-        }
-      });
+    this._wsConnection.on('message', (message) => {
+      console.log(message);
```
harsha509: is this required @raju249 , or just placed for debugging purpose ?

raju249: Hmm, no. But I think I have removed it before, maybe it was old diff that you saw?

javascript/node/selenium-webdriver/test/chrome/devtools_test.js
```diff
@@ -124,13 +125,23 @@ test.suite(
...
-        // eslint-disable-next-line no-unused-vars
-        await driver.get('http://the-internet.herokuapp.com/basic_auth')
+      it('denies entry if username and password do not match', async function() {
+        const pageCdpConnection = await driver.createCDPConnection('page')
+
+        await driver.register('random', 'random', pageCdpConnection)
+        await driver.get(server.url() + '/protected')
+        let source = await driver.getPageSource()
+        assert.equal(source.includes('Access granted!'), false)
+        await server.stop()
```
harsha509: we can ignore stopping server here as the next spec stops the server ( i dont see any server started in next spec so we can use the existing server)

javascript/node/selenium-webdriver/chromium.js
harsha509: may be Yes please ignore!

javascript/node/selenium-webdriver/test/chrome/devtools_test.js
raju249: Oh yes, thats correct. Removed on `server.stop`

Locally I tested using `.only` so that did no create any issue.

Fix pushed

javascript/node/selenium-webdriver/chromium.js
```diff
@@ -720,12 +721,20 @@ class Driver extends webdriver.WebDriver {
    * @param {string} debuggerAddress
    * @return {string} Returns parsed webSocketDebuggerUrl obtained from the http request
    */
-  async getWsUrl(debuggerAddress) {
-    let request = new http.Request('GET', '/json/version')
+  async getWsUrl(debuggerAddress, target) {
```
harsha509, line 726: Hi @raju249 ,

can we check for "target" here and handle the error if user is not providing a correct target, like the supported are `browser` and `page` for as i know, because if user doesn't provide any it may throw `target.toLowerCase' of undefined.

Also adding a JSDOC comment will be helpful what parameters actually goes in there for the user

raju249: Oh yes, fixed it. Please take a look 🙇

Harsha509 (approved):

Issue #8357, opened 2020-05-28
As a Firefox extension developer, I would like to install a temporary extension from a folder


seanpoulter:
**Edit**: 2020-06-11 &ndash; ⚠️ This initial comment is not correct. I misunderstood the GeckoDriver API. I was looking at stale source code. 😓 With new knowledge and some guidance from the GeckoDriver team, I've proposed an approach to handle the "install addon" command from FirefoxDriver consistently for each language. See my [long comment](https://github.com/SeleniumHQ/selenium/issues/8357#issuecomment-643047209) below.

--

## 🐛 Bug Report

We can't install a temporary extension in Firefox using the JavaScript or DotNet bindings. The implementation is inconsistent across the languages. The GeckoDriver API expects a `path` as a String but we're sending an `addon`. See [mozilla/geckodriver - marionette.rs](https://github.com/mozilla/geckodriver/blame/17945ac838449dda1defc8cda0c6dadb042519b2/src/marionette.rs#L1319-L1328).

These two bindings need to be fixed:

|Language|Implementation|Authors / CC|
|-|-|-|
|JavaScript|[javascript/node/selenium-webdriver/firefox.js#L637-L657](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/firefox.js#L637-L657)|@kvetko, @jleyba|
|DotNet|[dotnet/src/webdriver/Firefox/FirefoxDriver.cs#L213-L227](https://github.com/SeleniumHQ/selenium/blob/master/dotnet/src/webdriver/Firefox/FirefoxDriver.cs#L213-L227)|@jimevans|

These seem to be OK:

|Language|Implementation|
|-|-|
|Java|[java/client/src/org/openqa/selenium/firefox/FirefoxDriver.java#L235-L240](https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/firefox/FirefoxDriver.java#L235-L240)|
|Python|[py/selenium/webdriver/firefox/webdriver.py#L247-L264](https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/firefox/webdriver.py#L247-L264)|
|Ruby|[rb/lib/selenium/webdriver/firefox/bridge.rb#L34-L38](https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/firefox/bridge.rb#L34-L38)

## To Reproduce

[...]

PR #8725, opened 2020-09-22
[JS] Enable installAddon() to install unpacked extensions

snoack:
### Description
This pull request extends the `installAddon()` method of the `Driver` implementation for Firefox, so that a path to an unpacked extension can be specified alternatively to a packaged XPI or ZIP file.

### Motivation and Context
The `Driver` implementation for Firefox has an `installAddon()` method in order to install web extensions for the current session. The underlying `install addon` command can install packaged extensions using the `addon` parameter, as well as unpacked extensions using the `path` parameter. Support for the latter is missing in Selenium's `installAddon()` API.

### Types of changes

### Checklist

### DIFF
--- a/javascript/node/selenium-webdriver/test/firefox_test.js
+++ b/javascript/node/selenium-webdriver/test/firefox_test.js
@@ -34,6 +34,10 @@ const WEBEXTENSION_EXTENSION_ZIP = path.join(
   __dirname,
   '../lib/test/data/firefox/webextension.zip'
 )
+const WEBEXTENSION_EXTENSION_UNPACKED = path.join(
+  __dirname,
+  '../lib/test/data/firefox/webextension'
+)
 
 const WEBEXTENSION_EXTENSION_ID =
   'webextensions-selenium-example@example.com.xpi'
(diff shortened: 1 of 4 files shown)


javascript/node/selenium-webdriver/test/firefox_test.js
```diff
@@ -212,21 +216,14 @@ suite(
...
+        it('XPI can be installed and uninstalled', async function () {
+          await testInstallAddon(WEBEXTENSION_EXTENSION_XPI)
+        })
 
-        await driver.uninstallAddon(id)
-        await driver.get(Pages.echoPage)
-        await verifyWebExtensionNotInstalled()
+        xit('unpacked extension can be installed and uninstalled', async function () {
+          await testInstallAddon(WEBEXTENSION_EXTENSION_UNPACKED)
+        })
```
snoack, line 226: I'm sorry but I couldn't make the new test pass on the CI (and didn't get the test suite to run locally either). So I'm skipping it with `xit` for now.

`installAddon()` completes successfully but then it fails to find the element that the test extension were supposed to inject. I tried adding additional timeouts and retries but it didn't help.

However, when I'm testing these changes in our project, and pass the path to an unpacked extension to `installAddon()` everything works fine.

At the current point, the only idea left I have is that it might be the version of `geckodriver` but I didn't figure out yet where the CI is getting its version of `geckodriver` from, and frankly I spent already more time than I planned on this PR. But if I can get some pointers, I'm happy to finish up the test.

titusfortner (changes requested):
This goes along with #8357 & #9494,

The problem with this implementation is that you can't use the directory in a remote server.
As I mentioned in 8357, it might make sense for Selenium to zip/base64 the directory and use `addon`, but the user can just as easily zip it and provide it themselves right now.

titusfortner:
Thanks for this contribution, but I'm closing this in favor of the solution in #10216 since that one will support both local and remote driver instances.

PR #8747, opened 2020-09-30
[grid] Integrating NewSessionQueuer with Distributor.

pujagani:

### Description

Integrated Distributor with the NewSessionQueuer.
1. Update build files of Distributor.
2.  Add new session request listener to LocalDistributor. Reject request if it exceeds the timeout or does not have a slot for the capability. Retry request if all slots are busy.
3. Update Distributor instance declaration in all related tests.
4. Update Hub and Standalone commands.

### Motivation and Context

A new session request is added to the queue by the NewSessionQueuer, the Distributor picks up the request and creates a session. The changes add the functionality where the Distributor has an event handler for the new session request. It also includes logic to retry if slots are busy or to reject a request immediately if no slots have the capability requested.

### Types of changes

### Checklist


java/server/src/org/openqa/selenium/grid/data/NodeStatus.java
```diff
@@ -68,6 +68,13 @@ public NodeStatus(
     }
   }
 
+  public boolean hasCapability(Capabilities caps) {
+    long count = slots.stream()
+        .filter(slot -> slot.isSupporting(caps))
```
shs96c: Prefer `Stream.anyMatch` instead of iterating over all slots.

java/server/src/org/openqa/selenium/grid/distributor/Distributor.java
```diff
@@ -138,26 +140,26 @@ protected Distributor(
...
-      post("/se/grid/distributor/node/{nodeId}/drain")
-          .to((Map<String, String> params) -> new DrainNode(this, new NodeId(UUID.fromString(params.get("nodeId")))))
-          .with(requiresSecret),
-      delete("/se/grid/distributor/node/{nodeId}")
-          .to(params -> new RemoveNode(this, new NodeId(UUID.fromString(params.get("nodeId")))))
-          .with(requiresSecret),
-      get("/se/grid/distributor/status")
-          .to(() -> new GetDistributorStatus(this))
-          .with(new SpanDecorator(tracer, req -> "distributor.status")));
+  routes = Route.combine(
```
shs96c: Can you please put reformatting PRs into their own PR, rather than merging in here? It makes it hard to pick out what you've actually changed.

java/server/src/org/openqa/selenium/grid/distributor/Distributor.java
```diff
@@ -166,37 +168,50 @@ public CreateSessionResponse newSession(HttpRequest request)
...
     try (
-      Reader reader = reader(request);
-      NewSessionPayload payload = NewSessionPayload.create(reader)) {
+        Reader reader = reader(request);
+        NewSessionPayload payload = NewSessionPayload.create(reader)) {
       Objects.requireNonNull(payload, "Requests to process must be set.");
 
       attributeMap.put(AttributeKey.LOGGER_CLASS.getKey(),
-        EventAttribute.setValue(getClass().getName()));
+                       EventAttribute.setValue(getClass().getName()));
```
shs96c: Nit: This is breaking the project's style.

java/server/src/org/openqa/selenium/grid/distributor/Distributor.java
```diff
@@ -166,37 +168,50 @@ public CreateSessionResponse newSession(HttpRequest request)
...
 
       Iterator<Capabilities> iterator = payload.stream().iterator();
       attributeMap.put("request.payload", EventAttribute.setValue(payload.toString()));
 
       if (!iterator.hasNext()) {
-        SessionNotCreatedException exception = new SessionNotCreatedException("No capabilities found");
+        SessionNotCreatedException
```
shs96c: I suspect this can fit on to one line.

java/server/src/org/openqa/selenium/grid/distributor/Distributor.java
```diff
@@ -166,37 +168,50 @@ public CreateSessionResponse newSession(HttpRequest request)
...
+            exception =
+            new SessionNotCreatedException("No capabilities found");
         EXCEPTION.accept(attributeMap, exception);
         attributeMap.put(AttributeKey.EXCEPTION_MESSAGE.getKey(),
-          EventAttribute.setValue("Unable to create session. No capabilities found: "
-            + exception.getMessage()));
+                         EventAttribute.setValue("Unable to create session. No capabilities found: "
```
shs96c: Nit: breaking project style, because the indent is too large.

java/server/src/org/openqa/selenium/grid/distributor/Distributor.java
```diff
@@ -166,37 +168,50 @@ public CreateSessionResponse newSession(HttpRequest request)
...
+          iterator.next(),
+          ImmutableMap.of("span", span));
 
       Lock writeLock = this.lock.writeLock();
       writeLock.lock();
       try {
         Set<NodeStatus> model = ImmutableSet.copyOf(getAvailableNodes());
 
+        // Reject new session immediately if no node has the required capabilities
+        long hostsWithCaps = model.stream()
```
shs96c: This should be a boolean, not a long. Use `anyMatch` instead of counting.

java/server/src/org/openqa/selenium/grid/distributor/Distributor.java
```diff
@@ -242,12 +247,19 @@ public CreateSessionResponse newSession(HttpRequest request)
       span.addEvent("Session created by the distributor", attributeMap);
       return sessionResponse;
     } catch (SessionNotCreatedException e) {
+      if (e.getMessage().startsWith("Unable to find provider for session") ||
```
shs96c: This is going to break. If you want to catch a specific exception kind, create a new subclass of `SessionNotCreatedException` and catch _that_.

java/server/src/org/openqa/selenium/grid/distributor/Distributor.java
```diff
@@ -256,8 +268,10 @@ public CreateSessionResponse newSession(HttpRequest request)
       span.setStatus(Status.UNKNOWN);
 
       EXCEPTION.accept(attributeMap, e);
-      attributeMap.put(AttributeKey.EXCEPTION_MESSAGE.getKey(),
-        EventAttribute.setValue("Unknown error in LocalDistributor while creating session: " + e.getMessage()));
+      attributeMap.put(
+          AttributeKey.EXCEPTION_MESSAGE.getKey(),
+          EventAttribute.setValue("Unknown error in LocalDistributor while creating session: "
+                                  + e.getMessage()));
```
shs96c: Style nit: `+` on the line above (so it's obvious to the reader that the text continues on the next line), and fix the indentation.

java/server/src/org/openqa/selenium/grid/distributor/httpd/DistributorServer.java
```diff
@@ -66,7 +68,11 @@ public String getDescription() {
 
   @Override
   public Set<Role> getConfigurableRoles() {
-    return ImmutableSet.of(EVENT_BUS_ROLE, HTTPD_ROLE, SESSION_MAP_ROLE);
+    return ImmutableSet.of(EVENT_BUS_ROLE,
```
shs96c: Nit: newline before `EVENT_BUS_ROLE`, and then fix indentation.

java/server/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
```diff
@@ -88,33 +106,122 @@
...
     this.model = new GridModel(bus, registrationSecret);
     this.nodes = new HashMap<>();
+    this.sessionRequests = Require.nonNull("New Session Request Queue", sessionRequests);
+    this.sessionRequestTimeout = sessionRequestTimeout;
 
     bus.addListener(NodeStatusEvent.listener(status -> register(registrationSecret, status)));
     bus.addListener(NodeStatusEvent.listener(status -> model.refresh(registrationSecret, status)));
     bus.addListener(NodeDrainComplete.listener(this::remove));
+
+    bus.addListener(NewSessionRequestEvent.listener(reqId -> {
```
shs96c, line 129: The New Session Queue should present as a `BlockingQueue` or `BlockingDeque` to the `Distributor`. That means that at the heart of the process, all this class has to do is pull new requests off the queue and handle as appropriate.

java/server/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
```diff
@@ -88,33 +106,122 @@
...
+      Optional<HttpRequest> sessionRequest = this.sessionRequests.remove();
```
shs96c: There's a race condition here. Reading the code, it implicitly assumes that `sessionRequest` has the same `reqId` as the current head of the New Session Queue, and this isn't guaranteed. Either make `remove` accept `reaId`, or pass the request itself as data in the event. Given that the new session request can be Very Large Indeed, the former is preferred to the latter.

shs96c: I also assume that the request queue is threadsafe....

java/server/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
```diff
@@ -88,33 +106,122 @@
...
+      // Check if polling the queue did not return null
+      if (sessionRequest.isPresent()) {
+        handleNewSessionRequest(sessionRequest.get(), reqId);
+      } else {
+        fireSessionRejectedEvent("Unable to poll request from the new session request queue.",
+                                 reqId);
```
shs96c: Nit: newline before `"Unable...` and then fix indentation.

java/server/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
```diff
@@ -88,33 +106,122 @@
...
+      }
+    }));
+  }
+
+  private void handleNewSessionRequest(HttpRequest sessionRequest, RequestId reqId) {
+
+    Span span = newSpanAsChildOf(tracer, sessionRequest, "distributor.poll_queue");
+    Map<String, EventAttributeValue> attributeMap = new HashMap<>();
+    attributeMap.put(AttributeKey.LOGGER_CLASS.getKey(),
+                     EventAttribute.setValue(getClass().getName()));
```
shs96c: Nit: fix indentation.

shs96c: For most of the project, we try and follow the Google Java style guide: https://google.github.io/styleguide/javaguide.html#s4.2-block-indentation

java/server/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
```diff
@@ -88,33 +106,122 @@
...
+          CreateSessionResponse sessionResponse = newSession(sessionRequest);
+          NewSessionResponse newSessionResponse =
+              new NewSessionResponse(reqId, sessionResponse.getSession(),
+                                     sessionResponse.getDownstreamEncodedResponse());
+
+          bus.fire(new NewSessionResponseEvent(newSessionResponse));
+        } catch (SessionNotCreatedException e) {
+          // If error is due to the slots being busy, adding to front of queue else reject the request
+          // If that fails, then just reject the request.
+          if (e.getMessage().startsWith("Unable to find provider for session") ||
```
shs96c: This is error prone, and likely to fail. As a rule of thumb, never use the text of an exception's message to handle control flow.

shs96c: At this point, we may want `newSession` to return something more meaningful than what it's doing right now. If we had proper functional primitives, an `Either` might be useful, since we effectively want the response (if there is one) or a meaningful cause (if there isn't a response)

We then want to be able to respond differently based on the cause.

java/server/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
```diff
@@ -88,33 +106,122 @@
...
     HttpClient.Factory clientFactory = new NetworkOptions(config).getHttpClientFactory(tracer);
     SessionMap sessions = new SessionMapOptions(config).getSessionMap();
+    NewSessionQueuer sessionRequests =
+        new NewSessionQueuerOptions(config).getSessionQueuer(
+            "org.openqa.selenium.grid.sessionqueue.remote.RemoteNewSessionQueuer");
+    Duration requestTimeout = new NewSessionQueueOptions(config).getSessionRequestTimeout();
     BaseServerOptions serverOptions = new BaseServerOptions(config);
 
-    return new LocalDistributor(tracer, bus, clientFactory, sessions, serverOptions.getRegistrationSecret());
+    return new LocalDistributor(tracer,
```
shs96c: Style nit: indentation. It's better with these kinds of things to move the `tracer` to a new line, and then everything lines up nicely.

shs96c (dismissed):
I see the way you're heading, and it makes sense, but there are some issues that need addressing. Notably, using exceptions for control flow is generally a code smell. Using the messages in exceptions for control flow is downright dangerous --- a casual developer has no idea of the importance of those messages, as they're normally assumed to be for humans to read.

I've also made a set of style nits --- I know we don't follow the style consistently on this project, but it'd be nice to follow the Google Style Guide for Java when we're coding, please.

java/server/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
pujagani: +1, sure I will do that. Yes, the queue is threadsafe. Will pass the reqId back and remove the request associated with it.

pujagani: +1, will do that. Yes, the queue is threadsafe. Passing the reqId would be the better.

pujagani: Addressed as part of https://github.com/SeleniumHQ/selenium/pull/8754

java/server/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
pujagani: Alright, thank you for the direction. That would definitely be a nice and clean way to do it. Will add a way to use an `Either` for the `newSession`.

java/server/src/org/openqa/selenium/grid/distributor/Distributor.java
```diff
@@ -160,8 +162,22 @@ protected Distributor(
...
+        throw new SessionNotCreatedException(exception.getMessage(), exception);
+      }
+      throw sessionResponse.left();
+    }
+  }
+
+  public Either<SessionNotCreatedException, CreateSessionResponse> createNewSessionResponse(
+    HttpRequest request) throws SessionNotCreatedException {
 
     Span span = newSpanAsChildOf(tracer, request, "distributor.new_session");
```
diemol: Shouldn't this span reflect the method where it is placed?

java/server/src/org/openqa/selenium/grid/distributor/Distributor.java
```diff
@@ -208,40 +234,32 @@ public CreateSessionResponse newSession(HttpRequest request)
...
-      SESSION_ID.accept(span, sessionId);
-      CAPABILITIES.accept(span, caps);
-      SESSION_ID_EVENT.accept(attributeMap, sessionId);
-      CAPABILITIES_EVENT.accept(attributeMap, caps);
-      span.setAttribute(AttributeKey.SESSION_URI.getKey(), sessionUri);
-      attributeMap.put(AttributeKey.SESSION_URI.getKey(), EventAttribute.setValue(sessionUri));
-
-      span.addEvent("Session created by the distributor", attributeMap);
-      return sessionResponse;
+      if (!selected.isPresent()) {
```
diemol: For readability purposes, maybe it could be better to have the initial part of the `if` with the positive case (session created in this case), and the `else` with the negative one (session not created)

java/server/src/org/openqa/selenium/grid/distributor/Distributor.java
```diff
@@ -250,17 +268,19 @@ public CreateSessionResponse newSession(HttpRequest request)
...
-      throw e;
+      return Either.left(e);
     } catch (IOException e) {
       span.setAttribute("error", true);
       span.setStatus(Status.UNKNOWN);
 
       EXCEPTION.accept(attributeMap, e);
-      attributeMap.put(AttributeKey.EXCEPTION_MESSAGE.getKey(),
-        EventAttribute.setValue("Unknown error in LocalDistributor while creating session: " + e.getMessage()));
+      attributeMap.put(
```
diemol: Maybe reverting this change will make the review easier. It is changing the formatting.

java/server/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
```diff
@@ -88,33 +108,125 @@
...
+
   public LocalDistributor(
       Tracer tracer,
       EventBus bus,
       HttpClient.Factory clientFactory,
       SessionMap sessions,
-      Secret registrationSecret) {
+      NewSessionQueuer sessionRequests,
+      Secret registrationSecret,
+      Duration sessionRequestTimeout) {
```
diemol: Why do we need the `sessionRequestTimeout` as in the constructor? I thought this value was needed when creating the `NewSessionQueue`. Wouldn't this duplicate the value?

I think it would be better to have the configuration of the `sessionRequestTimeout` only in the queue bits of the code.

diemol (changes requested):

java/server/src/org/openqa/selenium/grid/distributor/Distributor.java
pujagani: +1

java/server/src/org/openqa/selenium/grid/distributor/Distributor.java
pujagani: Sure, will update that

java/server/src/org/openqa/selenium/grid/distributor/Distributor.java
pujagani: +1

java/server/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
pujagani: The LocalDistributor needs to know the timeout, so when it receives the request it will be able to reject immediately if it has timed out. The queue simply just stores the session requests.

diemol: Maybe the queue can see that the request has timed out even before sending the request to the distributor? This would reduce messages and calls, I think.

pujagani: Yes, it can do that actually. Before the event is fired to indicate to the Distributor to remove from the queue, the queue can check if it has timed out. Do you suggest this change happens as part of this PR or a new one? Since a lot of session queue files will get updated.

diemol: Good point, having it in a different PR is a good idea.

pujagani: Addressed this one https://github.com/SeleniumHQ/selenium/pull/8769

diemol (approved):

PR #8754, opened 2020-10-01
[grid] Add request id to poll the request from NewSessionQueue.

pujagani:

### Description

1. Added request id to the remove() method of NewSessionQueuer.
2. Ordering of requests is mandatory, hence updated the existing deque to store the request ids.
3. Added a concurrent hashmap to store the mapping of the request-id to the request. 
4. This way when the Distributor gets the new session request event, it will be able to poll for the request-id received as part of the event.

### Motivation and Context

When a new session request comes in, an event is fired to the Distributor (#8747). There was a condition that distributor assumed while polling from the queue, that head of the queue and the event has the same request-id. This PR fixes that.

### Types of changes

### Checklist

### DIFF
--- a/java/server/src/org/openqa/selenium/grid/sessionqueue/NewSessionQueue.java
+++ b/java/server/src/org/openqa/selenium/grid/sessionqueue/NewSessionQueue.java
@@ -44,7 +44,7 @@ public abstract class NewSessionQueue implements HasReadyState {
 
   public abstract boolean offerFirst(HttpRequest request, RequestId requestId);
 
-  public abstract Optional<HttpRequest> poll();
+  public abstract Optional<HttpRequest> poll(RequestId requestId);
 
   public abstract int clear();
 
--- a/java/server/src/org/openqa/selenium/grid/sessionqueue/NewSessionQueue.java
+++ b/java/server/src/org/openqa/selenium/grid/sessionqueue/NewSessionQueue.java
@@ -44,7 +44,7 @@ public abstract class NewSessionQueue implements HasReadyState {
 
   public abstract boolean offerFirst(HttpRequest request, RequestId requestId);
 
-  public abstract Optional<HttpRequest> poll(RequestId requestId);
+  public abstract Optional<HttpRequest> remove(RequestId requestId);
 
   public abstract int clear();
 
--- a/java/server/src/org/openqa/selenium/grid/sessionqueue/RemoveFromSessionQueue.java
+++ b/java/server/src/org/openqa/selenium/grid/sessionqueue/RemoveFromSessionQueue.java
@@ -22,6 +22,7 @@ import static org.openqa.selenium.remote.tracing.HttpTracing.newSpanAsChildOf;
 import static org.openqa.selenium.remote.tracing.Tags.HTTP_REQUEST;
 import static org.openqa.selenium.remote.tracing.Tags.HTTP_RESPONSE;
 
+import org.openqa.selenium.grid.data.RequestId;
 import org.openqa.selenium.internal.Require;
 import org.openqa.selenium.remote.http.HttpHandler;
 import org.openqa.selenium.remote.http.HttpRequest;
@@ -35,10 +36,12 @@ public class RemoveFromSessionQueue implements HttpHandler {
 
   private final Tracer tracer;
   private final NewSessionQueuer newSessionQueuer;
+  private final RequestId id;
 
-  RemoveFromSessionQueue(Tracer tracer, NewSessionQueuer newSessionQueuer) {
+  RemoveFromSessionQueue(Tracer tracer, NewSessionQueuer newSessionQueuer, RequestId id) {
     this.tracer = Require.nonNull("Tracer", tracer);
     this.newSessionQueuer = Require.nonNull("New Session Queuer", newSessionQueuer);
+    this.id = id;
   }
 
   @Override
@@ -46,7 +49,7 @@ public class RemoveFromSessionQueue implements HttpHandler {
     try (Span span = newSpanAsChildOf(tracer, req, "sessionqueuer.remove")) {
       HTTP_REQUEST.accept(span, req);
 
-      Optional<HttpRequest> sessionRequest = newSessionQueuer.remove();
+      Optional<HttpRequest> sessionRequest = newSessionQueuer.remove(id);
       HttpResponse response = new HttpResponse();
 
       if (sessionRequest.isPresent()) {
--- a/java/server/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
+++ b/java/server/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
@@ -146,7 +146,9 @@ public class LocalNewSessionQueue extends NewSessionQueue {
       Lock writeLock = lock.writeLock();
       writeLock.lock();
       try {
-        requestIdQueue.remove();
+        if (requestIdQueue.remove(requestId)) {
+          sessionRequests.remove(requestId);
+        }
       } finally {
         writeLock.unlock();
         bus.fire(new NewSessionRejectedEvent(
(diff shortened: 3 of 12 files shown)


java/server/src/org/openqa/selenium/grid/sessionqueue/RemoveFromSessionQueue.java
```diff
@@ -35,18 +36,20 @@
 
   private final Tracer tracer;
   private final NewSessionQueuer newSessionQueuer;
+  private final RequestId id;
 
-  RemoveFromSessionQueue(Tracer tracer, NewSessionQueuer newSessionQueuer) {
+  RemoveFromSessionQueue(Tracer tracer, NewSessionQueuer newSessionQueuer, RequestId id) {
```
diemol, line 41: Could you please remind me why we use `RemoveFromSessionQueue`? Is there something that prevents us from calling the `remove` action directly in the `NewSessionQueuer`?

java/server/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
```diff
@@ -55,10 +56,11 @@
 
   private static final Logger LOG = Logger.getLogger(LocalNewSessionQueue.class.getName());
   private final EventBus bus;
-  private final Deque<HttpRequest> sessionRequests = new ConcurrentLinkedDeque<>();
+  private final Deque<RequestId> requestIdQueue = new ConcurrentLinkedDeque<>();
+  private final Map<RequestId, HttpRequest> sessionRequests = new ConcurrentHashMap<>();
```
diemol: Would it make sense to put `RequestId` and `HttpRequest` in a new structure/class, and then only keep `Deque`?

I am not sure if things could get messy by managing two concurrent arrays that share the same information.

diemol (dismissed):

java/server/src/org/openqa/selenium/grid/sessionqueue/RemoveFromSessionQueue.java
pujagani: I thought it might be a good idea to have RemoveFromSessionQueue check if the poll was successful and respond accordingly and to keep that check separate to keep things clean. It also adds a Span for the same to trace the request, in Grid mode the request would be made from the LocalDistributor-> RemoteNewSessionQueuer -> NewSessionQueuerServer.

diemol: Sounds good to me!

java/server/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
pujagani: That is a fair concern. LocalNewSessionQueue is using a write lock before making any changes to both the data structures.  So anytime it holds the lock, only it will be able to update the two. The write lock is used for all the operations. This mapping is needed to poll the request based on the id. If the Distributor would call `remove(httpRequest)`, then we wouldn't need it but that is not efficient.

pujagani: Since we require deque functionality and there is no threadsafe version of LinkedHashMap, another data structure was required to store the mapping of id to request. I was also concerned that it might get difficult with two concurrent data structures, hence was relying on locks.  I will attempt to see if this can be changed to use just one data structure in some way and have the deque functionality.

pujagani: Do you think it might be a good idea to use` Collections.synchronizedMap(new LinkedHashMap())` and then perform operation in a thread-safe manner?

java/server/src/org/openqa/selenium/grid/sessionqueue/NewSessionQueue.java
```diff
@@ -44,7 +44,7 @@
 
   public abstract boolean offerFirst(HttpRequest request, RequestId requestId);
 
-  public abstract Optional<HttpRequest> poll();
+  public abstract Optional<HttpRequest> poll(RequestId requestId);
```
shs96c: The expected behaviour for `poll` would be to match what `Deque` does, which is: 

> Retrieves and removes the head of the queue represented by this deque

As such, you shouldn't need the `RequestId` here.

java/server/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
shs96c: The `Deque` is used for ordering, and the `Map` is used to store the actual request, right? I think our expected usage pattern is simply to take the item off the head of the `Deque` and just use that? With a very limited number of random removals of items, or other traversals looking for matches? If so, we might just want a holding data structure with two fields (`requestId, request`) and store _that_ in the `Deque`

java/server/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
```diff
@@ -96,12 +98,14 @@ public boolean offerLast(HttpRequest request, RequestId requestId) {
...
         EventAttribute.setValue(getClass().getName()));
 
-
-      added = sessionRequests.offerLast(request);
+      added = requestIdQueue.offerLast(requestId);
       addRequestHeaders(request, requestId);
 
-      attributeMap
-        .put(AttributeKey.REQUEST_ID.getKey(), EventAttribute.setValue(requestId.toString()));
+      if (added) {
```
shs96c: I note that this happens outside the write lock.

java/server/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueuer.java
```diff
@@ -75,8 +75,8 @@ public boolean retryAddToQueue(HttpRequest request, RequestId reqId) {
   }
 
   @Override
-  public Optional<HttpRequest> remove() {
-    return sessionRequests.poll();
+  public Optional<HttpRequest> remove(RequestId id) {
+    return sessionRequests.poll(id);
```
shs96c: `poll` and `remove` have different meanings.

shs96c (changes requested):
I think I see where this is going. We should draw a distinction between `poll` and `remove` and get these to align with the commonly understood meanings from the regular Java APIs.

java/server/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
pujagani: This change is related to https://github.com/SeleniumHQ/selenium/pull/8747#discussion_r497621655 . So the idea is to pass the request-id from the Distributor to remove the request from the queue instead of passing the request itself. Would it be okay performance-wise to traverse every time Distributor calls `remove(requestId)` ? Especially when all slots are occupied and we add it to the front of the `Deque` and retry ?

diemol: I think this is ok because if a remove happens often it is a sign that the user has an underpowered Grid with many new session requests. 

For example, a Grid with two nodes and 2 slots on each node won't be able to cope with a test suite sending tests in 20 threads. That is a not well planned Grid. 

Since we are having observability as a core part of the new Grid, we could redirect users to look at that and figure out how many removals are happening in their tests.

Summarizing, `remove(requestId)` should not happen that much in a well planned Grid.

pujagani: Thank you so much for the detailed explanation. I get the idea clearly now. Good point regarding the observability bit, we have tracing for remove request. It will provide good insight.

java/server/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
```diff
@@ -154,21 +157,39 @@ private void retryRequest(HttpRequest request, RequestId requestId) {
...
-          return Optional.empty();
+      // Peek  the deque and check if the request-id matches. Most cases, it would.
+      // If so poll the deque else iterate over the deque and find a match.
+      Optional<SessionRequest> firstSessionRequest =
+          Optional.ofNullable(sessionRequests.peekFirst());
+
+      Optional<HttpRequest> httpRequest = Optional.empty();
+      if (firstSessionRequest.isPresent()) {
+        if (id.equals(firstSessionRequest.get().getRequestId())) {
+          httpRequest = Optional.ofNullable(sessionRequests.pollFirst().getHttpRequest());
```
diemol, line 172: Why don't we get the `HttpRequest` from `firstSessionRequest` if it is `present` and it equals the `id` parameter?

diemol: Nevermind, `firstSessionRequest` is obtained through `peekFirst`, which does not remove.

diemol (approved):

Issue #8745, opened 2020-09-30
Support Default Capabilities for Nodes on Selenium 4


link89:
## 🚀 Feature Proposal

Allow setting default capabilities on the side of the node.
This feature should work on both hub-nodes and standalone modes.

## Motivation

We need to run tests on different versions of chrome (stable, beta, canary). 

To do this we plan to set up 3 selenium-grid, each grid for one version.
Since they are installed in different paths, we have to specify the chrome binary on the client-side. Which is not convenient because they may be different on different machines.

To solve this problem I hope that I can set the default `chrome_binary` on the nodes' side so that I can just leave this field empty on the client-side and respect the configuration of nodes.

## Example

I plan to support this configuration via system properties when starting selenium service.

```bash
java -Dwebdriver.chrome.defaultCapabilities='{"browserName":"chrome","goog:chromeOptions":{"binary":"/path/to/chrome"}}' \
  -jar selenium-server-4.0.0-alpha-6.jar standalone
```

The preset capabilities will be merged with capabilities that send from client-side before starting a new webdriver session so that the client-side's capabilities always have higher priority than the server-side's.

diemol:
An initial implementation has been done via https://github.com/SeleniumHQ/selenium/commit/d48a9596d868180b8b2da96e3c3118127365d434

It is possible now to provide a configuration that looks like this:
```toml
[node]
detect-drivers = false
[[node.driver-configuration]]
name = "Firefox Nightly"
stereotype = "{\"browserName\": \"firefox\", \"browserVersion\": \"86\", \"moz:firefoxOptions\": {\"binary\": \"/Applications/Firefox Nightly.app/Contents/MacOS/firefox-bin\"}}"
[[node.driver-configuration]]
name = "Chrome Beta"
stereotype = "{\"browserName\": \"chrome\", \"browserVersion\": \"88\", \"goog:chromeOptions\": {\"binary\": \"/Applications/Google Chrome Beta.app/Contents/MacOS/Google Chrome Beta\"}}"
[[node.driver-configuration]]
name = "Firefox"
stereotype = "{\"browserName\": \"firefox\", \"browserVersion\": \"84\"}"
[[node.driver-configuration]]
name = "Chrome"
stereotype = "{\"browserName\": \"chrome\", \"browserVersion\": \"87\"}"
```

and when the Node starts, you should see something like this:
```
[...]

PR #8761, opened 2020-10-04
Support Default Capabilities for Nodes on Selenium 4

link89:

### Description

https://github.com/SeleniumHQ/selenium/issues/8745

### Types of changes

### Checklist

### DIFF
--- a/java/client/src/org/openqa/selenium/Capabilities.java
+++ b/java/client/src/org/openqa/selenium/Capabilities.java
@@ -108,9 +108,25 @@ public interface Capabilities {
   default Capabilities merge(Capabilities other) {
     HashMap<String, Object> map = new HashMap<>(asMap());
     if (other != null) {
-      map.putAll(other.asMap());
+      deepMerge(map, other.asMap());
     }
     return new ImmutableCapabilities(map);
+  }
+
+  @SuppressWarnings("unchecked")
+  static Map deepMerge(Map dst, Map src) {
+    if (dst != null && src != null) {
+      for (Object key : src.keySet()) {
+        if (src.get(key) instanceof Map && dst.get(key) instanceof Map) {
+          Map originalChild = (Map)dst.get(key);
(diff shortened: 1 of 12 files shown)


shs96c:
Thank you for the contribution!

It's not clear from the description of this PR what the purpose is. It appears to be overriding the options added by the user when creating a new session, which may be problematic, but maybe I'm not understanding this properly.

It would be really helpful if you could describe the motivation behind this change. I'll shall hold off review this until then.

link89:
Hi @shs96c Thank you for your reply. This PR is linked to issue https://github.com/SeleniumHQ/selenium/issues/8745. I have explained the motivation there.

> It appears to be overriding the options added by the user when creating a new session, which may be problematic, but maybe I'm not understanding this properly.

It won't. The client-side's capabilities will have higher priority over the server-side's.

link89:
Hi @shs96c Can you help to review this patch?

diemol (changes requested):
Thanks for this PR @link89, however, the approach is not aligned with how the new Grid works. The new Grid receives parametwers via command line and via a configuration (`toml`) file. Specifying system capabilities is not what we'd like to see.

An approach would be to read the capabilities configuration from the `toml` file and avoid detecting the drivers if that configuration is provided. The code that handles that is around [here](https://github.com/SeleniumHQ/selenium/blob/c38657eb29d4aead24fe27f9891bd4be868ada8a/java/server/src/org/openqa/selenium/grid/node/config/NodeOptions.java#L77).

link89:
Thanks @diemol I will try to figure it out.

link89:
While waiting for a fix on selenium, I just provide a lightweight solution to address the issue: https://github.com/link89/selenium-federation

diemol:
Closed in favor of https://github.com/SeleniumHQ/selenium/commit/d48a9596d868180b8b2da96e3c3118127365d434

PR #8795, opened 2020-10-15
findFreePort now lets operative system system assign an arbitrary port

### LINKED ISSUE #8794: [selenium-webdriver] Let the operative system return a free port instead of randomly finding one
## 🚀 Feature Proposal

Background: I don't now much about TCP ports but did some investigation into why we got the "Unable to find free port" error in Protractor while running our E2E-tests. Turns out running a lot of parallel test cases with Protactor using Headless Chrome and ChromeDriver causes "port exhaustion". After just a few minutes all ~60000 ephemeral ports was used up and findFreePort has a hard time to randomly find an available port with the current implementation. Using `netstat` and `lsof` tells us that of those 60000 ports around 500-1000 is actually in use at a given time, the rest is in the CLOSE_WAIT state and can be reused (as I understand). We are running tests in 5 applications with sharding, allowing up to 5 test cases running at one and this can happen in multiple feature branches at the same time. So it seems hard to stop using all these ports since it's something happening beyond our control.

**My proposal is to let the operative system find an arbitrary port instead of randomly finding one by testing random numbers.**

## Motivation

Today the findFreePort method uses a random number to check if a port is available, my suggestion is to instead start a server by omitting or provide 0 as port so that we get a random available port from the operative system.
Read more here: https://nodejs.org/api/net.html#net_server_listen

After some testing I have found that it respects the system port range, at least on unix.
So the code can be simplified into a single method and no retries. 
I have not yet been able to generate an error, so I'm not sure what kind of errors you can get, but it's handles error as before.

## Example
```javascript
const net = require('net');

function findFreePort(opt_host) {
  return new Promise((resolve, reject) => {
    const server = net.createServer()
    server.on('listening', function() {
      resolve(server.address().port);
      server.close();
    });
    server.on('error', (e) => {
      if (e.code === 'EADDRINUSE' || e.code === 'EACCES') {
        resolve('Unable to find a free port')
      } else {
        reject(e)
      }
    });
    server.listen(0, opt_host);
  });
}

findFreePort()
  .then((port) => {
    console.log(port);
  }).catch((e) => {
    console.error(e);
  });
```

JonWallsten:

Fixes: https://github.com/SeleniumHQ/selenium/issues/8794

### Description
I have enhanced the findFreePort method to ask the operative system for a free arbitrary port instead of trying to find one by generating random numbers.

### Motivation and Context
When running a lot of tests in parallel in multiple branches with test sharding Chrome/ChromeDrivers uses up all ports. It's then harder for selenium-webdriver to find an available port which causes the "Unable to find free port" error message. 
So instead of using chance to find a free port we let the operative system do it for us. No need for 100 retries.

### Types of changes

### Checklist

### DIFF
--- a/javascript/node/selenium-webdriver/net/portprober.js
+++ b/javascript/node/selenium-webdriver/net/portprober.js
@@ -161,15 +161,24 @@ function isFree(port, opt_host) {
  * @return {!Promise<number>} A promise that will resolve to a free port. If a
  *     port cannot be found, the promise will be rejected.
  */
-async function findFreePort(opt_host) {
-  let range = await findSystemPortRange()
-  for (let i = 0; i < 100; i++) {
-    let port = Math.floor(Math.random() * (range.max - range.min) + range.min)
-    if (await isFree(port, opt_host)) {
-      return port
-    }
-  }
-  throw Error('Unable to find a free port')
+
+function findFreePort(opt_host) {
+  return new Promise((resolve, reject) => {
+    const server = net.createServer();
+    server.on('listening', function() {
+      resolve(server.address().port);
+      server.close();
+    });
+    server.on('error', (e) => {
+      if (e.code === 'EADDRINUSE' || e.code === 'EACCES') {
+        resolve('Unable to find a free port')
+      } else {
+        reject(e)
+      }
+    });
+    // By providing 0 we let the operative system find an arbitrary port
+    server.listen(0, opt_host);
+  });
 }
 
 // PUBLIC API


javascript/node/selenium-webdriver/net/portprober.js
```diff
@@ -161,15 +161,23 @@ function isFree(port, opt_host) {
...
+      server.close();
+    });
+    server.on('error', (e) => {
+      if (e.code === 'EADDRINUSE' || e.code === 'EACCES') {
+        resolve('Unable to find a free port')
+      } else {
+        reject(e)
+      }
+    });
+    server.listen(0, opt_host);
```
diemol, line 180: I am not a JS expert, but a few questions and comments:

- Is this guaranteeing that an ephemeral port is returned for all OS?
- This change is not returning the port anymore, was this change tested?
- What happens if the method fails? The current code retries in a range of ports.
- Would this be thread safe? If 100 tests run in parallel, will a 100 free ports be returned?
- Probably this change needs some extra tests in the `portprober_test.js`.

diemol (changes requested):
Thank you for creating this PR, @JonWallsten.

I will close the issue #8794 so we can continue the conversation here. I did an early review of this PR, please have a look at the comments.

javascript/node/selenium-webdriver/net/portprober.js
JonWallsten: 1. Node documentation states the following: 
`If port is omitted or is 0, the operating system will assign an arbitrary unused port, which can be retrieved by using server.address().port after the 'listening' event has been emitted.` How it finds these ports are up to the OS though.
2. The new code is returning a port at line 169.
3. It will throw an error just like it did in the old code.
4. Unless there's a bug in Node or in the OS itself it should. See first point.
5. Since it still returns a port if successful or the same error message if not, tests should be fine.

JonWallsten: Stackoverflow question about how Node does this: https://stackoverflow.com/questions/9901043/how-does-node-js-choose-random-ports
Uforrtunately not a lot of info. I tried to find the code in NodeJs project, but I guess this happens at a pretty low level.

diemol: >The new code is returning a port at line 169

I completely oversaw that, thanks for pointing it out.

Then I think this makes sense, please let us know when the PR is ready for review and merge (is a draft right now).

JonWallsten: No worries! I guess it's ready from my side. There was some weird issue with the test though. Some timeout in Travis.

Also ,I'm not sure how to actually test that it respects the system port range on all systems and similar stuff. 
You would have to change the system range on the system you test on and verify that it doesn't return something not in the range. 
We tried on our unix env to set net.ipv4.ip_local_port_range="1025 1026" and it always returned 1025. I could test manually on Windows as well as my Mac. But i guess it's hard to write automated test for it. 
I can also try to set the range to just 1025-1025, start a server and then try to get another port to see what actually happens when you run out.

diemol (approved):
Thank you, @JonWallsten! Congrats on your first contribution 🎉 Welcome to the project!

PR #8811, opened 2020-10-21
[JS]Cleaning up chrome.js file

harsha509:
### Description
Removing duplicate declarations from chrome as chrome extends chromium.js

### Types of changes

### Checklist

### DIFF
--- a/javascript/node/selenium-webdriver/chrome.js
+++ b/javascript/node/selenium-webdriver/chrome.js
@@ -209,7 +209,7 @@ class Options extends chromium.Options {
   }
 
   /**
-   * Sets the directory to store Chrome minidump in. This option is only
+   * Sets the directory to store Chrome minidumps in. This option is only
    * supported when ChromeDriver is running on Linux.
    * @param {string} path The directory path.
    * @return {!Options} A self reference.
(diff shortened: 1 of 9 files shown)


javascript/node/selenium-webdriver/chrome.js
```diff
@@ -346,7 +209,7 @@ class Options extends chromium.Options {
   }
 
   /**
-   * Sets the directory to store Chrome minidumps in. This option is only
+   * Sets the directory to store Chrome minidump in. This option is only
```
AutomatedTester: `minidumps` is the correct term

AutomatedTester (approved):
Could we have something like the description as the commit message so that we can see what the changes are when we look at the git log.

javascript/node/selenium-webdriver/chrome.js
harsha509: Thank you for reviewing @AutomatedTester. I have reverted it.

harsha509:
> Could we have something like the description as the commit message so that we can see what the changes are when we look at the git log.

Thank you for the review @AutomatedTester. I have updated the git commit message.

diemol (approved):

Issue #8719, opened 2020-09-21
Add HTTP/2 support to `NettyServer`


shs96c:
## 🚀 Feature Proposal

Add HTTP/2 support to the `NettyServer`

## Motivation

So we can serve traffic over http/2

This should be a small change to `SeleniumHttpInitializer`

PR #8815, opened 2020-10-23
Add Http/2 support in Netty Server

pujagani:
### Description

Add support for Http/2 in the Netty server.  It is related to #8719.

### Motivation and Context

Current Netty Server supports Http/1.1 for secure communication. The changes add support for Http/2  for a client requesting the use of Http/2. 
This will allow clients to leverage features of Http/2 where a single connection is used along with streams for communication.
Netty supports stream multiplexing but leveraging child channels i.e. every new stream has a dedicated child channel. 

### Types of changes

### Checklist


java/server/src/org/openqa/selenium/netty/server/Http2ResponseConverter.java
```diff
@@ -0,0 +1,77 @@
...
+      return;
+    }
+
+    HttpResponse seResponse = (HttpResponse) msg;
+
+    byte[] bytes = Contents.bytes(seResponse.getContent());
+    ByteBuf byteBuf = Unpooled.wrappedBuffer(bytes);
+
+    FullHttpResponse response = new DefaultFullHttpResponse(HTTP_1_1,
+                                                            HttpResponseStatus
```
shs96c: Nit: this is why I suggest putting `HTTP_1_1` on its own line: so the other parameters are easy to read.

java/server/src/org/openqa/selenium/netty/server/Http2RequestConverter.java
```diff
@@ -0,0 +1,75 @@
...
+import io.netty.handler.codec.http.HttpResponseStatus;
+import io.netty.handler.codec.http.HttpVersion;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import org.openqa.selenium.remote.http.HttpMethod;
+import org.openqa.selenium.remote.http.HttpRequest;
+
+public class Http2RequestConverter extends SimpleChannelInboundHandler<HttpObject> {
```
shs96c: This is mirroring the functionality of the `RequestConverter`, but requires http object aggregation in order to work, and so is likely to lead to problems when there's very little memory. In addition, it duplicates logic from that class, making fixing discovered issues more work and opening us up to forgetting to change both places.

I'd suggest deleting this and using `RequestConverter`

java/server/src/org/openqa/selenium/netty/server/Http2ResponseConverter.java
```diff
@@ -0,0 +1,77 @@
...
+import io.netty.handler.codec.http.FullHttpResponse;
+import io.netty.handler.codec.http.HttpResponseStatus;
+import io.netty.handler.codec.http.HttpUtil;
+
+import java.util.logging.Logger;
+
+import org.openqa.selenium.remote.http.Contents;
+import org.openqa.selenium.remote.http.HttpResponse;
+
+public class Http2ResponseConverter extends ChannelOutboundHandlerAdapter {
```
shs96c: Again, this appears to duplicate `ResponseConverter`

java/server/src/org/openqa/selenium/netty/server/Http2StreamInitializer.java
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  @Override
+  public void handlerAdded(ChannelHandlerContext ctx) {
+    addHttp2SpecificHandlers(ctx.pipeline());
+    ctx.pipeline().remove(this);
+  }
+
+  private void addHttp2SpecificHandlers(ChannelPipeline pipeline) {
+    pipeline.addLast("http2-codec", new Http2StreamFrameToHttpObjectCodec(true));
```
shs96c, line 44: These should be added to the `SeleniumInitializer`

java/server/src/org/openqa/selenium/netty/server/SeleniumHttp2Handler.java
```diff
@@ -0,0 +1,62 @@
...
+  protected void channelRead0(ChannelHandlerContext ctx, HttpRequest msg) {
+    EXECUTOR.submit(() -> {
+      HttpResponse res;
+      String streamId = msg.getHeader(STREAM_ID_HEADER);
+      try {
+        res = seleniumHandler.execute(msg);
+      } catch (Throwable e) {
+        res = new ErrorHandler(e).execute(msg);
+      }
+      res.addHeader(STREAM_ID_HEADER, streamId);
```
shs96c: Presumably this header is required as an identifier? If so, use `setHeader`, which will ensure it's the only value set.

java/server/src/org/openqa/selenium/netty/server/SeleniumHttp2Handler.java
```diff
@@ -0,0 +1,62 @@
...
+  public SeleniumHttp2Handler(HttpHandler seleniumHandler) {
+    super(HttpRequest.class);
+    this.seleniumHandler = Require.nonNull("HTTP handler", seleniumHandler);
+  }
+
+  @Override
+  protected void channelRead0(ChannelHandlerContext ctx, HttpRequest msg) {
+    EXECUTOR.submit(() -> {
+      HttpResponse res;
+      String streamId = msg.getHeader(STREAM_ID_HEADER);
```
shs96c: I think it'd be okay to check for this header in the existing `SeleniumHandler` and set the header on the response if the value isn't `null`

shs96c (changes requested):
I think we can probably simplify the logic of the stream initialisation, and make better use of the existing code we have, particularly when it comes to converting from Netty's http primitives to Selenium's.

java/server/src/org/openqa/selenium/netty/server/Http2StreamInitializer.java
pujagani: We are using [Http2MultiplexHandler](https://netty.io/4.1/api/io/netty/handler/codec/http2/Http2MultiplexHandler.html) for Http/2 support of streams. It leverages child channels of Netty. The handler takes in only a single handler to be added for child channels and their respective channel context. In order to add multiple handlers, `Http2StreamInitializer` was created. An instance of this sharable handler is passed while creating `Http2MultiplexHandler`.

java/server/src/org/openqa/selenium/netty/server/Http2RequestConverter.java
pujagani: Thank you for pointing it out. Since Http/2 uses frames with stream ids, I am leveraging to Http2StreamFrameToHttpObjectCodec get the HttpObject to pass it to Selenium handlers. I have moved the Http/2 request handling to `RequestConverter`.

java/server/src/org/openqa/selenium/netty/server/SeleniumHttp2Handler.java
pujagani: Thank you. Moved it to `SeleniumHandler`.

PR #8834, opened 2020-11-03
Fixed issue with Firefox browserName camelCasing

luke-hill:
~Rudimentary fix by copying the legit behaviour. Can try use a more generic fix if we're happy to refactor the Common::Options class.~

After further diving, found out firefox worked all along, but because `{a_b_c: 123}` responds to as_json, and returns `self`, the clever camelCasing logic never got ran.

On ChromeOptions the hash check was first. So if we simply re-order the checks, everything slots into place perfectly.

Have also refactored the chrome one in-line with titus' suggestion.

cc / @p0deje

### DIFF
--- a/rb/CHANGES
+++ b/rb/CHANGES
@@ -1,3 +1,15 @@
+trunk
+=========================
+Firefox:
+  * Fixed an issue whereby using the new capabilities structure in local drivers
+  would cause the W3C check to complain that `browser_name` was an invalid
+  capability key name (issue #8832)
+
+All Browsers:
+  * Refactor to `#generate_as_json` which now will check if the structure is
+  "Hash-like" first. So that camelCasing on keys will occur earlier in the
+  conditional check
+
 4.0.0.alpha6 (2020-05-28)
 =========================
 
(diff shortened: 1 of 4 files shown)


luke-hill:
We could refactor the chrome method and the firefox one added by just altering these 3 lines in common

https://github.com/SeleniumHQ/selenium/blob/trunk/rb/lib/selenium/webdriver/common/options.rb#L91-L93

This would remove around 25 lines of code and alter a couple of lines, but this would percolate down into all other drivers, which I believe is correct, but cannot be sure!

titusfortner:
Since this is getting called recursively couldn't we just:
```
camelize_keys = false if value == 'pref' 
super
```
I haven't pulled it to test it, but I hate to see the same logic repeated in Options, Chrome::Options & Firefox::Options

luke-hill:
I think we can just edit the elsif part of the common branch with the bit in chrome/firefox.

The only issue is that might affect other browsers. But I think this camelised name is what we need for all of them. But I've been out of browser testing for too long to say definitively.

luke-hill:
ping @titusfortner This worked a treat once I dived into your suggestion more.

titusfortner (approved):
Nice work figuring out that the order of the conditionals mattered!

titusfortner:
Also, this was Rails's fault. ;-)

luke-hill:
How so?

Also I've rebased and added a bit of info to the changelog so we don't lose this info.

titusfortner:
Rails is the one monkey patching in the `#as_json` method to `Hash` class. (at least that's my excuse for not putting them in the better order). :)

diemol:
I see the PR is approved, can it be merged?

titusfortner:
My fault for not double checking this. @luke-hill I sent you down the wrong path. We don't actually have the keyword in that method, just the value, so the code suggestion I made was bad. I restored it to what Alex had originally. You solved the real problem by changing the order of operations.

Issue #7761, opened 2019-11-08
Expose Server components over JMX


shs96c:
We have the `ManagedService` annotation that used to be used expose the old selenium server via JMX. We should make the new server equally manageable.

samueltauil:
I have interest on that to capture info from JMX and send to Prometheus then been able to build a custom chart

amiluslu:
Can you give me a little bit details about this issue ? If you give me a little bit details or clue, hints, I can start to fix.

shs96c:
@amiluslu we discussed this a little on the Slack channel. Do you still need some help getting started? I'm happy to provide more information if needed.

pujagani:
The older Grid had the following managed services and attributes.

 - Hub as a managed service. 
			 - Managed attribute :
	 - Configuration
	 - Session request count
 - DefaultRemoteProxy as managed service :
			- Managed attributes:
	- Proxy status (alive or no)
- It extends the BaseRemoteProxy. BaseRemoteProxy has managed attributes:
	 - Max number of sessions 
	 - Remote host
	 - Total number of used slot
	 - Total number of slots
	 - Proxy status
	 - Resource usage counted as (total no of used slots/ max number of sessions) * 100
	 - The last session start time
- SeleniumServer which is an implementation of GridNodeServer as a managed service.

The new Grid can have analogous managed services and attributes as follows:

 - Each server of grid servers can be a managed service :
	 - DistributorServer 
	 - Hub  server
	 - SessionMapServer 
	 - Standalone server
	 - NodeServer 
	 - RouterServer
	 - NewSessionQueuer Server 
	 - EventBus 
 - All above-listed components share config values in BaseServerOptions. These config values can be exposed as managed attributes. 
 - Each individual components have additional component-specific attributes, which can be exposed as managed attributes.
 - The Distributor is aware of the Grid model. Grid specific attributes can be exposed by leveraging Grid model.
 - Information specific to a Node like max session count, slot information and availability can be obtained from via node implementations.

These are the observations and suggestions so so far. Can I get some help in understanding if this aligns with the intended feature? Please provide any other details I might have missed.

diemol:
Fixed via #8838

PR #8838, opened 2020-11-04
Expose server config, session queue and node as MBeans for JMX monitoring.

pujagani:
### Description

Changes are related to #7761.

### Motivation and Context

Old Grid exposed MBeans for JMX monitoring. The changes include exposing MBeans and attributes for monitoring the new Grid. This includes:
1. Exposing BaseServerConfig and related attributes.
2. Exposing Node and related attributes.
3. NewSessionQueue and related attributes.

### Types of changes

### Checklist


java/server/src/org/openqa/selenium/grid/server/BaseServerOptions.java
```diff
@@ -94,6 +104,15 @@ public URI getExternalUri() {
     }
   }
 
+  @ManagedAttribute(name = "URL")
+  public URL getExternalUrl() {
+    try {
+      return getExternalUri().toURL();
+    } catch (MalformedURLException e) {
+      throw new ConfigException("Cannot convert URI to URL" + e.getMessage());
```
AutomatedTester: nit: We probably want a space after `URL`

java/server/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
```diff
@@ -80,6 +88,17 @@ public boolean isReady() {
     return bus.isReady();
   }
 
+  @ManagedAttribute(name = "NewSessionQueueSize")
+  public int getQueueSize() {
+    Lock writeLock = lock.writeLock();
+    writeLock.lock();
```
diemol: Why do we need to lock to read the queue size?

diemol (changes requested):
👋 @pujagani, I went through it and I added a comment besides the one already exists.

Also, it seems tests for this PR are not passing, and a couple are still to be written. Could you please have a look? I'll happily review right after.

java/server/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
pujagani: Since it uses ConcurrentLinkedDeque, the size() requires a traversal and value may not be accurate if it is being modified when size is calculated. Hence the lock. But for exposing values if we don't need that kind of accuracy, then we can get rid of the lock. @diemol What would you suggest?

pujagani:
@diemol I updated and added the tests and they all pass now. The test was picking up previously registered MBeans and hence failing for different cases. Please help review. Thank you!

java/server/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
```diff
@@ -80,6 +88,17 @@ public boolean isReady() {
     return bus.isReady();
   }
 
+  @ManagedAttribute(name = "NewSessionQueueSize")
+  public int getQueueSize() {
+    Lock writeLock = lock.writeLock();
```
shs96c: This need only be a read lock

java/server/src/org/openqa/selenium/grid/node/local/LocalNode.java
```diff
@@ -181,11 +189,53 @@ public boolean isReady() {
...
+  }
+
+  @ManagedAttribute(name = "Load")
+  public float getLoad() {
+    long inUse = factories.stream().filter(sessionSlot -> !sessionSlot.isAvailable()).count();
+    return inUse / (float) maxSessionCount * 100f;
+  }
+
+  @ManagedAttribute(name = "RemoteNodeUri")
+  public URI getExternalURI() {
```
shs96c: Nit: `getExternalUri` We try and AVOID SHOUTING in Selenium (which is why we have `HttpMethod`, and so on)

java/server/src/org/openqa/selenium/grid/server/BaseServerOptions.java
```diff
@@ -101,6 +111,15 @@ public URI getExternalUri() {
     }
   }
 
+  @ManagedAttribute(name = "URL")
+  public URL getExternalUrl() {
```
shs96c: Daft question, but why don't we just return the URI?

java/server/test/org/openqa/selenium/grid/router/JmxTest.java
```diff
@@ -0,0 +1,222 @@
...
+
+      MBeanAttributeInfo[] attributeInfoArray = info.getAttributes();
+      assertThat(attributeInfoArray).hasSize(2);
+
+      String requestTimeout = (String) beanServer.getAttribute(name, "RequestTimeoutSeconds");
+      assertThat(Long.parseLong(requestTimeout)).isEqualTo(queueOptions.getRequestTimeoutSeconds());
+
+      String retryInterval = (String) beanServer.getAttribute(name, "RetryIntervalSeconds");
+      assertThat(Long.parseLong(retryInterval)).isEqualTo(queueOptions.getRetryIntervalSeconds());
+    } catch (InstanceNotFoundException | IntrospectionException | ReflectionException
```
shs96c, line 176: It's fine to just let the test fail because the exception was thrown. That'll give you more information that just calling `fail` in these `catch` blocks.

diemol (approved):
Thank you, @pujagani!

Issue #8839, opened 2020-11-05
[py] `send_key` producing an error: `join() takes exactly one argument`


sufyanAbbasi:
## 🐛 Bug Report

Running `send_keys` in the Python environment with variable arguments produces the following error:
```
TypeError: join() takes exactly one argument (2 given)
```
Looks like the bug is fixed by changing the variable expansion `*value` used in the `join` to the plain old tuple, `value`, produced by the variable argument:
https://github.com/SeleniumHQ/selenium/blob/trunk/py/selenium/webdriver/remote/webelement.py#L539

## To Reproduce

Spin up a local Python Webdriver, inject a `<text-area>` with some text in it, then send cntrl-a, cntrl-c, and cntrl-v:
```python
driver.execute_script('var textArea = document.createElement("textarea");textArea.textContent = "my cool text";document.body.appendChild(textArea);')
text_area = driver.find_element_by_tag_name('textarea')
text_area.click()
text_area.send_keys(Keys.CONTROL, 'a')
text_area.send_keys(Keys.CONTROL, 'c')
text_area.send_keys(Keys.CONTROL, 'v')
```

## Expected behavior

The text area should select all text content, copy, then paste the content of the text area (or whatever is in the clipboard). Instead, I get the following error:

```
TypeError: join() takes exactly one argument (2 given)
```

## Environment

OS: Linux
Browser: Chrome
Browser version: 86.0.4240.75
Language Bindings version: Python

sufyanAbbasi:
See [Pull #8840](https://github.com/SeleniumHQ/selenium/pull/8840) for change. Testing it locally, this change seems to fix the error, but I will leave it up to you to test it out comprehensively!

PR #8840, opened 2020-11-05
[py] Fix send_keys with variable argument by removing variable expansion.

sufyanAbbasi:
### Description
Fixes Issue [#8839](https://github.com/SeleniumHQ/selenium/issues/8839)

### Motivation and Context
Python `send_keys` method is broken for variable arguments. See Issue [#8839](https://github.com/SeleniumHQ/selenium/issues/8839).

### Types of changes

### Checklist


AutomatedTester:
Could you add a test for this as it appears that there is a regression that isn't caught by the current test suite

sufyanAbbasi:
> Could you add a test for this as it appears that there is a regression that isn't caught by the current test suite

From what I can tell, the variable argument `send_keys` is being tested here:
https://github.com/SeleniumHQ/selenium/blob/941dc9c6b2e2aa4f701c1b72be8de03d4b7e996a/py/test/selenium/webdriver/common/typing_tests.py#L100

I haven't run the test suite locally to verify myself, but please let me know what you find!

AutomatedTester:
It's not tested at all. It needs to be a test that does multiple test files

AutomatedTester:
We will need to add a test at a later date

PR #8903, opened 2020-11-24
Add print page command to python binding

raju249:

### Description

Adding support for 'print' endpoint in python binding.

### Types of changes

### Checklist


py/selenium/webdriver/remote/webdriver.py
```diff
@@ -888,6 +888,13 @@ def minimize_window(self):
         """
         self.execute(Command.MINIMIZE_WINDOW)
 
+    def print_page(self, options = {}):
```
AutomatedTester: Let's create a new class called `PrintOptions` that people can populate and then we can get the info from when passing to the endpoint.

raju249: Sure. Done!

py/test/selenium/webdriver/common/print_pdf_tests.py
```diff
@@ -0,0 +1,79 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import pytest
+
+from selenium.webdriver import Chrome
```
AutomatedTester: The driver is injected via a fixture so we don't need this or the line below

py/test/selenium/webdriver/common/print_pdf_tests.py
```diff
@@ -0,0 +1,79 @@
...
+from selenium.webdriver import Firefox
+from selenium.webdriver.chrome.options import Options
+from selenium.webdriver.common.options import PrintOptions
+
+start_index = 0
```
AutomatedTester: Constants should be capitalised

py/test/selenium/webdriver/common/print_pdf_tests.py
```diff
@@ -0,0 +1,79 @@
...
+end_index = 5
+pdf_magic_number = 'JVBER'
+options = Options()
+options.add_argument("--headless") # chrome needs to be headless, print works only in headless
+
+chrome_driver = Chrome(options=options)
```
AutomatedTester: Drivers are injected via fixtures

py/test/selenium/webdriver/common/print_pdf_tests.py
```diff
@@ -0,0 +1,79 @@
...
+firefox_driver = Firefox()
+
+def test_pdf_with_2_pages(pages):
```
AutomatedTester: Tests should be succinct, these tests appear to have had the kitchen sink thrown in.  We don't test 2 browsers at the same time in any tests so that needs to be rewritten

AutomatedTester: All the tests in this file need to be redone as mentioned above.

py/selenium/webdriver/common/options.py
```diff
@@ -87,3 +87,33 @@ def to_capabilities(self):
...
     def default_capabilities(self):
         return {}
+
+class PrintOptions():
+
+    def __init__(self):
+        self._print_options = {}
+
+    @property
+    def print_options(self):
```
AutomatedTester: This should called `to_dict()`

py/selenium/webdriver/common/options.py
```diff
@@ -87,3 +87,33 @@ def to_capabilities(self):
...
+    def set_option(self, key, value):
+        """
+        Sets a print option in the hash
+        """
+        if key and value:
+            self._print_options[key] = value
+        else:
+            raise ValueError('key/value cannot be null')
+
+    def get_option(self, key):
```
AutomatedTester: Nothing appears to use the method so we should remove it or write a test

py/selenium/webdriver/common/options.py
```diff
@@ -87,3 +87,33 @@ def to_capabilities(self):
     @property
     def default_capabilities(self):
         return {}
+
+class PrintOptions():
```
AutomatedTester: A class should guide a user on how to use it via the APIs it offers.

We should add setter properties for each of the values that people want to set. At the moment we don't validate that people are setting the right thing

py/selenium/webdriver/remote/webdriver.py
```diff
@@ -888,6 +888,17 @@ def minimize_window(self):
         """
         self.execute(Command.MINIMIZE_WINDOW)
 
+    def print_page(self, print_option_arg = None):
```
AutomatedTester: Let's not use a form of hungarian notation in naming our variables

py/selenium/webdriver/remote/webdriver.py
```diff
@@ -888,6 +888,17 @@ def minimize_window(self):
...
+        """
+        Takes PDF of the current page.
+        The driver makes a best effort to return a PDF based on the provided parameters.
+        """
+        options = {}
```
AutomatedTester, line 896: We don't need another dict here, we can just use populate `print_option` with the `to_dict()` method

py/selenium/webdriver/common/options.py
```diff
@@ -87,3 +87,33 @@ def to_capabilities(self):
...
+        self._print_options = {}
+
+    @property
+    def print_options(self):
+        """
+        :Returns: A hash of print options configured
+        """
+        return self._print_options
+
+    def set_option(self, key, value):
```
AutomatedTester: We should remove this in favour of specific method names

AutomatedTester (changes requested):

raju249:
Checking the comments.

py/test/selenium/webdriver/common/print_pdf_tests.py
raju249: I need to set the `headless` option explicitly for chrome, `print` command doesnt work in non-headless browser. Confirmed and checked with @JS31096

py/selenium/webdriver/remote/webdriver.py
raju249: `print_options` are kept to be optional, the extra `options` is to make sure that I don't call `to_dict` on `None`.

py/selenium/webdriver/common/print_page_options.py
```diff
@@ -0,0 +1,154 @@
...
+        self._margin = {}
+
+    @property
+    def to_dict(self):
+        """
+        :Returns: A hash of print options configured
+        """
+        return self._print_options
+
+    def set_orientation(self, value):
```
AutomatedTester: We should have these are getter and setter properties rather than using `set_*` methods.

See an `Options` class as an example

py/selenium/webdriver/common/print_page_options.py
```diff
@@ -0,0 +1,154 @@
...
+        if value not in self.orientation_values:
+            raise ValueError('Orientation value must be one of ' + str(self.orientation_values))
```
AutomatedTester:
```suggestion
            raise ValueError(f'Orientation value must be one of {self.orientation_values}')
```

py/selenium/webdriver/common/print_page_options.py
```diff
@@ -0,0 +1,154 @@
...
+
+        self._page['width'] = value
+        self._print_options['page'] = self._page
+
+    def get_width(self):
+        return self._page.get('width', None)
+
+    def set_height(self, value):
+        if not isinstance(value, float) and not isinstance(value, int):
+            raise ValueError('Height should be an integer or a float')
```
AutomatedTester: We should probably extract this into a "private" method since it's used a few times

py/selenium/webdriver/common/print_page_options.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def set_background(self, value):
+        if not isinstance(value, bool):
+            raise ValueError('Set background value should be a boolean')
+        self._print_options['background'] = value
+
+    def get_background(self):
+        return self._print_options.get('background', None)
+
+    def set_width(self, value):
```
AutomatedTester: this should probably be called `page_width`, and properties that deal with `self._page` should probably start with `page_` so people understand it's use

py/selenium/webdriver/common/print_page_options.py
```diff
@@ -0,0 +1,154 @@
...
+        if value < 0:
+            raise ValueError('Height cannot be less then 0')
+
+        self._page['height'] = value
+        self._print_options['page'] = self._page
+
+    def get_height(self):
+        return self._page.get('height', None)
+
+    def set_top(self, value):
```
AutomatedTester: this should probably be called `margin_top` and all the properties that deal with margin updated with a similar prefix

py/selenium/webdriver/common/print_page_options.py
```diff
@@ -0,0 +1,154 @@
...
+class PrintOptions():
+
+    def __init__(self):
+        self.orientation_values = ['portrait', 'landscape']
+        self._print_options = {}
+        self._page = {}
+        self._margin = {}
+
+    @property
+    def to_dict(self):
```
AutomatedTester, line 26: This doesn't need to be a property, it's fine to have it with brackets

py/selenium/webdriver/common/print_page_options.py
```diff
@@ -0,0 +1,154 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+class PrintOptions():
+
+    def __init__(self):
+        self.orientation_values = ['portrait', 'landscape']
```
AutomatedTester: Since these are values that aren't going to change we should have this variable in capitals

AutomatedTester (changes requested):
We need to have docstrings for each of the methods/properties so people know how to use it.

py/test/selenium/webdriver/common/firefox_print_pdf_tests.py
```diff
@@ -0,0 +1,51 @@
+# Licensed to the Software Freedom Conservancy (SFC) under one
```
AutomatedTester: Nearly there. 

Change this file to be called `print_pdf_tests.py`

Disable the tests with `pytest.mark.xfail_{insert browser that doesn't work here}`

py/test/selenium/webdriver/common/firefox_print_pdf_tests.py
```diff
@@ -0,0 +1,51 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+from selenium.webdriver import Firefox
+from selenium.webdriver.common.print_page_options import PrintOptions
+
+START_INDEX = 0
+END_INDEX = 5
+PDF_MAGIC_NUMBER = 'JVBER'
+firefox_driver = Firefox()
```
AutomatedTester: use the `driver` fixture instead on each of the tests

py/test/selenium/webdriver/common/print_page_options_tests.py
```diff
@@ -0,0 +1,94 @@
+# Licensed to the Software Freedom Conservancy (SFC) under one
```
AutomatedTester: This file should live in `py/test/unit/selenium/webdriver/common`

py/test/selenium/webdriver/common/chrome_print_pdf_tests.py
```diff
@@ -0,0 +1,59 @@
+# Licensed to the Software Freedom Conservancy (SFC) under one
```
AutomatedTester: Can you move this test to the `chrome` directory and add a `TODO` that we need to have a fixture added for this

AutomatedTester (approved):

PR #8958, opened 2020-12-11
[rb] Allow httpOnly property to be bubbled up to cookie manager

luke-hill:
Ronseal PR - Add httpOnly property of cookies to be ascertained in ruby.

There's currently no test file for this, so didn't amend any specs.


luke-hill:
@p0deje For a quick review / merge.

diemol (approved):
Thank you, @luke-hill!
 
I see this being done in Java and C# bindings, this should be fine.

@titusfortner, @p0deje, feel free to revert in case this is not the place to add this.

PR #8963, opened 2020-12-12
All `is None` and `is not None` removal

machinexa2:
Removal of all `is None` and `is not None` with modern conditional syntax. I have added some comments at some places to better clarification.
1. is None/is not None is deprecated
2. is None/is not None takes greater opcodes than modern conditional syntax
```
>>> dis(normal_comp)
  1           0 LOAD_FAST                0 (x)
              2 POP_JUMP_IF_FALSE        8
              4 LOAD_FAST                0 (x)
              6 RETURN_VALUE
        >>    8 LOAD_CONST               1 ('')
             10 RETURN_VALUE
>>> dis(is_not_comp)
  1           0 LOAD_FAST                0 (x)
              2 LOAD_CONST               0 (None)
              4 IS_OP                    1
              6 POP_JUMP_IF_FALSE       12
              8 LOAD_FAST                0 (x)
             10 RETURN_VALUE
        >>   12 LOAD_CONST               1 ('')
             14 RETURN_VALUE
```

### DIFF
--- a/py/selenium/webdriver/common/proxy.py
+++ b/py/selenium/webdriver/common/proxy.py
@@ -54,7 +54,6 @@ class ProxyType:
         value = str(value).upper()
         for attr in dir(cls):
             attr_value = getattr(cls, attr)
-            # `attr_value['string'] is not None` probably not required as `attr_value['string'] == value`
             if isinstance(attr_value, dict) and 'string' in attr_value and attr_value['string'] == value:
                 return attr_value
         raise Exception(f"No proxy type is found for {value}")
--- a/py/selenium/webdriver/common/proxy.py
+++ b/py/selenium/webdriver/common/proxy.py
@@ -54,10 +54,8 @@ class ProxyType:
         value = str(value).upper()
         for attr in dir(cls):
             attr_value = getattr(cls, attr)
-            if isinstance(attr_value, dict) and \
-                    'string' in attr_value and \
-                    attr_value['string'] is not None and \
-                    attr_value['string'] == value:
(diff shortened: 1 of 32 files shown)


py/selenium/webdriver/common/proxy.py
```diff
@@ -54,10 +54,8 @@ def load(cls, value):
         value = str(value).upper()
         for attr in dir(cls):
             attr_value = getattr(cls, attr)
-            if isinstance(attr_value, dict) and \
-                    'string' in attr_value and \
-                    attr_value['string'] is not None and \
-                    attr_value['string'] == value:
+            # `attr_value['string'] is not None` probably not required as `attr_value['string'] == value`
+            if isinstance(attr_value, dict) and 'string' in attr_value and attr_value['string'] == value:
```
machinexa2, line 58: # `attr_value['string'] is not None` probably not required as `attr_value['string'] == value` check is already being done

py/selenium/webdriver/firefox/webdriver.py
```diff
@@ -20,7 +20,6 @@
     basestring = str
 
 from base64 import b64decode
-import shutil
```
machinexa2, line 23: I forgot to remove shutil in https://github.com/SeleniumHQ/selenium/pull/8950/ so I did here

py/selenium/webdriver/common/proxy.py
AutomatedTester: We don't need the comment  here as it adds little to no value

machinexa2:
@AutomatedTester removed that comment

AutomatedTester (approved):

Issue #8167, opened 2020-04-06
[Grid]Add support for retries to HttpHandlers 


diemol:
## 🚀 Feature Proposal

For scenarios where a command fails and then succeeds after a retry, it'd be nice to add support for retries to the HttpHandlers in the Grid implementation.

We could use https://resilience4j.readme.io/docs or https://jodah.net/failsafe/fallback/ for this.

## Motivation

See the feature description.

diemol:
This is actually more relevant now since the new Grid (when using all the pieces apart) needs to be started in a specific order. 
We need to work on this before continuing to work on the docker images.

diemol:
The PR related to this issue was merged.

PR #8977, opened 2020-12-17
Add retry http request filter for transient errors. Integrate filter with ClientConfig.

pujagani:

### Description

Related to #8167. Add retry mechanism using Failsafe.

### Motivation and Context

When a client makes a request to the server, it might not get fulfilled due to different errors. One category of errors is transient in nature. In such cases, the request might be fulfilled if it is reattempted with a delay. The changes introduce a retry mechanism as a filter. The retry policies handler read timeout, connection failures and server errors (server unavailable and internal server error without response body). Rest of the failures are not retried. 

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/remote/http/RetryRequest.java
+++ b/java/src/org/openqa/selenium/remote/http/RetryRequest.java
@@ -1,0 +1,83 @@
+// Licensed to the Software Freedom Conservancy (SFC) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The SFC licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package org.openqa.selenium.remote.http;
+
+import net.jodah.failsafe.Failsafe;
+import net.jodah.failsafe.RetryPolicy;
+import org.openqa.selenium.TimeoutException;
+
+import java.net.ConnectException;
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import static java.net.HttpURLConnection.HTTP_INTERNAL_ERROR;
+import static java.net.HttpURLConnection.HTTP_UNAVAILABLE;
+
+public class RetryRequest implements Filter {
+
+  private static final Logger LOG = Logger.getLogger(RetryRequest.class.getName());
+
+  // Retry on connection error.
+  private static final RetryPolicy<HttpResponse> connectionFailurePolicy =
+    new RetryPolicy<HttpResponse>()
+      .handleIf(failure -> failure.getCause() instanceof ConnectException)
+      .withBackoff(1, 4, ChronoUnit.SECONDS)
+      .withMaxRetries(3)
+      .withMaxDuration(Duration.ofSeconds(10))
+      .onRetry(e -> LOG.log(
+        Level.WARNING,
+        "Connection failure #{0}. Retrying.",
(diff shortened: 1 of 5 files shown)


java/client/test/org/openqa/selenium/remote/http/BUILD.bazel
```diff
@@ -1,10 +1,15 @@
...
+MEDIUM_TESTS = [
+    "RetryRequestTest.java",
+]
+
 java_test_suite(
     name = "SmallTests",
     size = "small",
-    srcs = glob(["*.java"]),
+    srcs = glob(["*.java"],
+                exclude = MEDIUM_TESTS),
```
shs96c: Unless `buildifier` reformatted things this way, it's fine to put the `exclude` on the same line

java/client/src/org/openqa/selenium/remote/http/RetryRequest.java
```diff
@@ -0,0 +1,83 @@
...
+      .withBackoff(1, 4, ChronoUnit.SECONDS)
+      .withMaxRetries(3)
+      .withMaxDuration(Duration.ofSeconds(10))
+      .onRetry(e -> LOG.log(
+        Level.WARNING,
+        "Read timeout #{0}. Retrying.",
+        e.getAttemptCount()));
+
+  // Retry if server is unavailable or an internal server error occurs without response body.
+  private static final RetryPolicy<HttpResponse> serverErrorPolicy =
```
shs96c: I'd be tempted to abandon hope if the response is an internal error: it's likely sending the request again will cause the same problem.

pujagani: Here the retry is for 500 error code without any response payload. In this iteration, the retry is 2 attempts then results in failure if the error still exists. I am thinking it should be okay. What do you suggest?

java/src/org/openqa/selenium/remote/http/RetryRequest.java
```diff
@@ -0,0 +1,83 @@
...
+      .withMaxDuration(Duration.ofSeconds(10))
+      .onRetry(e -> LOG.log(
+        Level.WARNING,
+        "Read timeout #{0}. Retrying.",
+        e.getAttemptCount()));
+
+  // Retry if server is unavailable or an internal server error occurs without response body.
+  private static final RetryPolicy<HttpResponse> serverErrorPolicy =
+    new RetryPolicy<HttpResponse>()
+      .handleResultIf(response -> response.getStatus() == HTTP_INTERNAL_ERROR &&
```
shs96c, line 65: The webdriver spec uses the 500 response for error responses, so this might be risky.

java/src/org/openqa/selenium/remote/http/RetryRequest.java
```diff
@@ -0,0 +1,83 @@
...
+        Contents.bytes(response.getContent()).length == 0)
```
shs96c: There's no guarantee that the contents of the http request can be read more than once unless it's been memoized, and it's a bad idea to do that unless you know exactly what content is coming. Would be safer to check the `Content-Length` header here.

java/test/org/openqa/selenium/remote/http/RetryRequestTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+public class RetryRequestTest {
+
+  private HttpClient client;
+  private static final String REQUEST_PATH = "http://%s:%s/hello";
+
+  @Before
+  public void setUp() throws MalformedURLException {
+    ClientConfig config = ClientConfig.defaultConfig()
+      .baseUrl(new URL("http://example.com"))
```
shs96c: Better to use `URI.create("http://localhost:2345")` or similar for two reasons:

 1. We don't want the test accidentally hanging because it tries to actually reach example.com
 2. Each time you create a URL like this, Java internally does a DNS lookup, and it's not needed in this case.

java/src/org/openqa/selenium/remote/http/RetryRequest.java
pujagani: Here the retry is for 500 error code without any response payload. In this iteration, the retry is 2 attempts then results in failure if the error still exists. I am thinking it should be okay. What do you suggest?

shs96c: Let's ship it and see what breaks. I can't think of a circumstance when the remote end will send a 500 without a message body.

shs96c (approved):
Ship it! :)

Issue #9001, opened 2020-12-27
Support Ruby 3.0


tagliala:
## 🐛 Bug Report

Selenium Firefox webdriver (both 3.x and 4.alpha) does not work with Ruby 3. There could be multiple causes, but the main one is that `rexml` is now a bundled gem

## Test script or set of commands reproducing this issue

```rb
#!/usr/bin/env ruby

require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'selenium-webdriver'
end

require 'selenium/webdriver/firefox'
```

## Expected behavior

Script ends without errors

## Actual behavior

```sh
$ ./selenium.rb 
Fetching gem metadata from https://rubygems.org/......
Resolving dependencies...
Using bundler 2.2.3
Using rubyzip 2.3.0
Using childprocess 3.0.0
Using selenium-webdriver 3.142.7
~/.rvm/gems/ruby-3.0.0/gems/selenium-webdriver-3.142.7/lib/selenium/webdriver/firefox.rb:22:in `require': cannot load such file -- rexml/document (LoadError)
	from ~/.rvm/gems/ruby-3.0.0/gems/selenium-webdriver-3.142.7/lib/selenium/webdriver/firefox.rb:22:in `<top (required)>'
	from selenium.rb:9:in `require'
	from selenium.rb:9:in `<main>'
```

## Environment

OS: macOS 10.15.7 (19H114)
Browser: n/a
Browser version: n/a
Browser Driver version: n/a
Language Bindings version: Ruby 3.0
Selenium Grid version (if applicable): n/a

jcoyne:
I'm seeing an argument error here too:
```
     ArgumentError:
       wrong number of arguments (given 1, expected 0)
     # /Users/jcoyne85/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/selenium-webdriver-3.142.7/lib/selenium/webdriver/firefox/driver.rb:31:in `new'
     # /Users/jcoyne85/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/selenium-webdriver-3.142.7/lib/selenium/webdriver/common/driver.rb:54:in `for'
     # /Users/jcoyne85/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/selenium-webdriver-3.142.7/lib/selenium/webdriver.rb:88:in `for'
[...]

PR #9002, opened 2020-12-27
[3.x] Add rexml to support Ruby 3

tagliala:
### ⚠️
First PR here, read contributing guidelines, searched for duplicates. This is also an issue in `trunk` branch. I can't run tests on my local machine. Also, Ruby 3.0 should be add to the matrix. I can't understand if the deprecated Grid 3.x is related to this branch. Feel free to close this PR and only consider #9001 

### Description
Add rexml to runtime dependencies and net-telnet to development dependencies

### Motivation and Context
`rexml` is a bundled gem in Ruby 3.0

This causes errors with selenium’s Firefox webdriver when requiring
`rexml/document`

### Types of changes

### Checklist

### References

- https://bugs.ruby-lang.org/issues/16485
- ruby/ruby@c3ccf23
- jnunemaker/crack#62
- https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/
- #9001

### DIFF
--- a/rb/lib/selenium/webdriver/firefox/driver.rb
+++ b/rb/lib/selenium/webdriver/firefox/driver.rb
@@ -28,7 +28,7 @@ module Selenium
           # @return [Marionette::Driver, Legacy::Driver]
           #
 
-          def new(**opts)
+          def new(opts)
             if marionette?(opts)
               Firefox::Marionette::Driver.new(opts)
             else
--- a/rb/selenium-webdriver.gemspec
+++ b/rb/selenium-webdriver.gemspec
@@ -32,6 +32,7 @@ Gem::Specification.new do |s|
   s.require_paths = ['lib']
 
   s.add_runtime_dependency 'childprocess', ['>= 0.5', '< 4.0']
+  s.add_runtime_dependency 'rexml'
   s.add_runtime_dependency 'rubyzip', ['>= 1.2.2']
 
   # childprocess requires ffi on windows but doesn't declare it in its dependencies


titusfortner:
Thanks for your submission.

> `net-telnet` is no longer standard library in Ruby 3.0. This caused errors
when running rake

What was the error / stack trace for this one?

tagliala:
Hi @titusfortner 

when running `rake`
```
$ rake
(in /Users/geremia/dev/selenium)
rake aborted!
LoadError: cannot load such file -- net/telnet
~/dev/selenium/Rakefile:7:in `require'
~/dev/selenium/Rakefile:7:in `<top (required)>'
~/.rvm/gems/ruby-3.0.0/bin/ruby_executable_hooks:24:in `eval'
~/.rvm/gems/ruby-3.0.0/bin/ruby_executable_hooks:24:in `<main>'
(See full trace by running task with --trace)
```

I have this problem also with other Ruby versions. But I don't know if I'm supposed to run `rake` from the `rb` folders

Do I simply need `rspec` to run specs?

tagliala:
I can see that there is a duplicate at #9007

titusfortner:
That's against trunk where we've removed Firefox Extension, this PR is against our Selenium 3.x branch if we decide to release another 3.x Ruby version before 4.0. We may not, since the 4.0 alpha should be sufficiently usable, or people can add this gem to their own gemfile easily enough in the meantime.

Unfortunately you can't just run rake to execute the Ruby Selenium specs you need to use our build system and execute:
`./go test_rb`
A lot of stuff needs to be installed like Bazel: https://github.com/SeleniumHQ/selenium/#bazel. 
I'll see if I can look at it this weekend. That Rakefile should only be getting read by our project's custom JRuby so it won't actually do anything to add it to the gemspec in the rb directory.

tagliala:
Ok, got it

I'm fine with using 4.0 alpha

btw, should I remove `net-telnet` ?

titusfortner:
yes, remove net-telnet and I'll merge it. It may or may not end up in a release. :)

Thanks!

deivid-rodriguez:
Hi!

What's left to do here?

tagliala:
Hi @deivid-rodriguez 

This change does not work because of this: https://github.com/SeleniumHQ/selenium/issues/9001#issuecomment-752265220

Double splatting argument doesn't help here

https://github.com/SeleniumHQ/selenium/blob/e82be7d3584062a0d16af8d562d387e3d4855aa1/rb/lib/selenium/webdriver/firefox/driver.rb#L29-L35

Removing double spat on line 29 works, but this may have side effects on older versions of Ruby and I don't know how to properly test

deivid-rodriguez:
But that sounds like a different issue, right? This change seems correct anyways?

tagliala:
This change is correct, but it will not bring Ruby 3.0 support to the Firefox driver

If you prefer to make another PR to fix the firefox driver, that would be great

deivid-rodriguez:
I was fixing [formtastic CI on ruby 3](https://github.com/formtastic/formtastic/pull/1323) and this change by itself got it green (I added the dependency in the `Gemfile` side to workaround the missing dependency here).

So while the missing keyword problem only affects some usages (I guess people passing options when setting up drivers), this change can get some people unblocked already so I think it has merit on its own.

tagliala:
> So while the missing keyword problem only affects some usages (I guess people passing options when setting up drivers), this change can get some people unblocked already so I think it has merit on its own.

I can replicate the same issue with a brand new rails 6.1 / Ruby 3 application without any other setting except:

### spec
```rb
# frozen_string_literal: true

require 'rails_helper'

RSpec.describe 'Pages' do
  describe 'Home', js: true do
    it 'has application name in title' do
      visit root_path

      expect(page).to have_title t('app_name')
    end
  end
end
```

### rails_helper
```rb
  config.before(:each, type: :system, js: true) do
    driven_by :selenium, using: :firefox
  end
```

Failure:
```
     1.2) Failure/Error:
            def new(**opts)
              if marionette?(opts)
                Firefox::Marionette::Driver.new(opts)
              else
                Firefox::Legacy::Driver.new(opts)
              end
          
          ArgumentError:
            wrong number of arguments (given 1, expected 0)
```

deivid-rodriguez:
Right. Formtastic generates a brand new rails app including whatever the default rails app includes but doesn't really use system specs inside that app. The reason this change "fixed" my case is that booting the app stopped failing.

deivid-rodriguez:
So this change has some value on itself, but I agree quite limited :sweat_smile:.

tagliala:
I've removed the double splat operator and I have tested this change on several ruby (and Capybara) versions, including Ruby 2.3 (the minimum supported Ruby version) and 3.0.

Specs start normally and I can see Firefox. But again, I do not use Firefox for testing, so this may have some side effects that I can't see

```rb
   28:           # @return [Marionette::Driver, Legacy::Driver]
   29:           #
   30: 
   31:           def new(**opts)
   32:             debugger
=> 33:             if marionette?(opts)
   34:               Firefox::Marionette::Driver.new(opts)
   35:             else
   36:               Firefox::Legacy::Driver.new(opts)
   37:             end
```

```rb
(byebug) RUBY_VERSION
"2.3.3"
# double squashed
(byebug) opts
{:desired_capabilities=>{:unexpectedAlertBehaviour=>"ignore"}}
# unsquashed (this branch)
(byebug) opts
{:desired_capabilities=>{:unexpectedAlertBehaviour=>"ignore"}}

(byebug) RUBY_VERSION
"2.5.7"
# double squashed
(byebug) opts
{:options=>#<Selenium::WebDriver::Firefox::Options:0x00007ffc0af70c70 @args=#<Set: {}>, @binary=nil, @profile=nil, @log_level=nil, @prefs={}, @options={}>, :http_client=>#<Capybara::Selenium::PersistentClient:0x00007ffc0d4a8c40 @open_timeout=nil, @read_timeout=nil>}
# unsquashed (this branch)
(byebug) opts
{:options=>#<Selenium::WebDriver::Firefox::Options:0x00007fbb42dc51e0 @args=#<Set: {}>, @binary=nil, @profile=nil, @log_level=nil, @prefs={}, @options={}>, :http_client=>#<Capybara::Selenium::PersistentClient:0x00007fbb42e4b9e8 @open_timeout=nil, @read_timeout=nil>}
```

titusfortner:
Thanks for your contribution and the discussion here, but it's going to be moot because for "reasons" I don't want to do another Selenium 3.x release. This issue has been fixed in trunk and will be available when we release the first Selenium 4 beta, which should be "soon."

PR #9006, opened 2020-12-28
further improvements to the json processing

joerg1985:
Faster reading and writing of json data, focused on base64 encoded data.
This will e.g. help reading 4k screenshots or sending binary data.

### Description
`Input.java` maintained a buffer with the chars read, to allow a helpfull `toString()`. It is more efficient to use an input buffer and keep the chars read in the buffer to allow a helpfull `toString()`. Changes to `JsonOutput.java` / `JsonInput.java` are small performance improvements, not related to the changes inside the `Input.java`.

### Motivation and Context
same as https://github.com/SeleniumHQ/selenium/pull/8546

### Types of changes

### Checklist

### DIFF
--- a/java/client/src/org/openqa/selenium/json/Input.java
+++ b/java/client/src/org/openqa/selenium/json/Input.java
@@ -26,72 +26,83 @@ import java.io.UncheckedIOException;
 /**
  * Similar to a {@link Reader} but with the ability to peek a single character ahead.
  * <p>
- * For the sake of providing a useful {@link #toString()} implementation, keeps a small circular
- * buffer of the most recently read characters.
+ * For the sake of providing a useful {@link #toString()} implementation, keeps the most recently
+ * read characters in the input buffer.
  */
 class Input {
-
   public static final char EOF = (char) -1;
+  // the number of chars to buffer
+  private static final int BUFFER_SIZE = 4096;
+  // the number of chars to remember, safe to set to 0
+  private static final int MEMORY_SIZE = 128;
+
   private final Reader source;
-  private boolean read;
-  private char peekedChar;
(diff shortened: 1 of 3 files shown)


joerg1985:
I am not able to run all tests, the tests of the json package are successful.

barancev:
Thank you for the patch, but it requires a little improvement. I've added more tests to cover the code touched by this PR to ensure backward compatibility, see commit f59e5ce16d2cd67b4adcf225e5910e8148dc4573. This patch does not pass the test `shouldBeAbleToReadNonWellFormedDataLongerThanReadBuffer` because it adds EOF char to the buffer unlike the previous implementation. Including EOF in the text generated by `toString` method looks suboptimal.

barancev (changes requested):
Needs rework to pass `shouldBeAbleToReadNonWellFormedDataLongerThanReadBuffer` test

joerg1985:
@barancev Thanks for the review, i changed the code to drop the EOF from the toString method.

The test is still failing, the string is one char shorter (due to the slot for the EOF char in the buffer).
I could either add some lines to be 100% backward compatible or change the tests to the new behaviour.

In my mind it would be better to change the test, because the extra lines for backward compatibility would only be some hard to understand remains in the code. Could you please decide which of the solutions is the best way to go?

barancev:
Is it worth to put EOF to the buffer at all? We can simply return it from the `read()` method without storing to the buffer.

joerg1985:
@barancev it is possible to use a return value, see the latest push. kind regards!

barancev:
Merged as 80bd2628658d226a6b02e0ff5d155146726ad71d, thank you for the contribution

PR #9007, opened 2020-12-28
Add rexml dependency

znz:
### Description

Since ruby 3.0.0, rexml is a bundled gem.
So it must be depend explicitly.

### Motivation and Context

https://stackoverflow.com/questions/65479863/rails-6-1-what-is-preventing-tests-from-running/65480744#65480744

### Types of changes

### Checklist

### DIFF
--- a/rb/selenium-webdriver.gemspec
+++ b/rb/selenium-webdriver.gemspec
@@ -32,9 +32,9 @@ Gem::Specification.new do |s|
   s.require_paths = ['lib']
 
   s.add_runtime_dependency 'childprocess', ['>= 0.5', '< 5.0']
+  s.add_runtime_dependency 'rexml'
   s.add_runtime_dependency 'rubyzip', ['>= 1.2.2']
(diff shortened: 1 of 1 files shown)


titusfortner:
Since Selenium 4 has removed the Firefox extension, I think the right answer for trunk is to delete the requires from firefox.rb rather than adding the dependency. I haven't had a chance to pull Ruby 3 yet to check it out, yet...

voxik:
There are also other places with REXML:

https://github.com/SeleniumHQ/selenium/blob/109c25ac6d18691b5503cd26494825a6a8434e75/rb/lib/selenium/server.rb#L112-L121

ioquatix:
Also have this problem, fixed it by adding `gem 'rexml'` to my own gemfile, but good to see it's merged here. What's the ETA on the 4.0 release?

titusfortner:
@ioquatix go ahead and use the beta. Very little will change in Ruby bindings between now and final release.

PR #9008, opened 2020-12-28
[rb] Fix deprecation warnings for Ruby 2.7 in 3.x

rafaelfranca:
### Description

Same as https://github.com/SeleniumHQ/selenium/commit/ee994fe79e37fe6ed4050aac1bf4b0befb0d5338
but in the remote driver that was forgotten

### Types of changes

### Checklist

### DIFF
--- a/rb/lib/selenium/webdriver/remote/driver.rb
+++ b/rb/lib/selenium/webdriver/remote/driver.rb
@@ -36,7 +36,7 @@ module Selenium
 
         def initialize(opts = {})
           listener = opts.delete(:listener)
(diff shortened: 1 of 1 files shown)


titusfortner:
Thanks for this contribution!

I'll merge this, but realize this isn't part of trunk, and we may not release another Selenium 3.x ruby gem.
The Ruby Selenium 4 alpha gem should be completely usable at this point.

rafaelfranca:
Thank you. I'll try to upgrade the rails test suite to use Selenium 4 to see if we can drop 3.x.

PR #9038, opened 2021-01-13
Add disable escaping option for json parsing. Fix string escaping for event-logs.

pujagani:

### Description

Add disable escaping option for json parsing. 

### Motivation and Context

The event-logs attributes like session payload and capabilities leverage the built-in json parse to convert the objects to json strings. By default, the escape characters are included in the string. It resulted in inconsistent log string with some fields having escape characters and some being printed normally. The changes added, fix this by adding a disable escaping option. 

### Types of changes

### Checklist

### DIFF
--- a/java/client/src/org/openqa/selenium/json/JsonOutput.java
+++ b/java/client/src/org/openqa/selenium/json/JsonOutput.java
@@ -65,38 +65,6 @@ public class JsonOutput implements Closeable {
     GSON_ELEMENT = gsonElement;
   }
 
-  // https://www.json.org has some helpful comments on characters to escape
-  // See also https://tools.ietf.org/html/rfc8259#section-7 and
-  // https://github.com/google/gson/issues/341 so we escape those as well.
-  // It's legal to escape any character, so to be nice to HTML parsers,
-  // we'll also escape "<" and "&"
-  private static final Map<Integer, String> ESCAPES;
-  static {
-    Map<Integer, String> builder = new LinkedHashMap<>();
-
-    for (int i = 0; i <= 0x1f; i++) {
-      // We want nice looking escapes for these, which are called out
-      // by json.org
-      if (!(i == '\b' || i == '\f' || i == '\n' || i == '\r' || i == '\t')) {
-        builder.put(i, String.format("\\u%04x", i));
-      }
-    }
-
-    builder.put((int) '"', "\\\"");
-    builder.put((int) '\\', "\\\\");
-    builder.put((int) '/', "\\u002f");
-    builder.put((int) '\b', "\\b");
-    builder.put((int) '\f', "\\f");
-    builder.put((int) '\n', "\\n");
-    builder.put((int) '\r', "\\r");
-    builder.put((int) '\t', "\\t");
-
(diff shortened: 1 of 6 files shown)


java/client/src/org/openqa/selenium/json/JsonOutput.java
```diff
@@ -65,38 +65,6 @@
     GSON_ELEMENT = gsonElement;
   }
 
-  // https://www.json.org has some helpful comments on characters to escape
-  // See also https://tools.ietf.org/html/rfc8259#section-7 and
-  // https://github.com/google/gson/issues/341 so we escape those as well.
-  // It's legal to escape any character, so to be nice to HTML parsers,
-  // we'll also escape "<" and "&"
-  private static final Map<Integer, String> ESCAPES;
```
shs96c, line 73: Prefer keeping this static block and setting a boolean for escaping if we're heading down this route.

java/client/src/org/openqa/selenium/json/JsonOutput.java
```diff
@@ -213,6 +184,42 @@ public JsonOutput setPrettyPrint(boolean enablePrettyPrinting) {
     return this;
   }
 
+  // https://www.json.org has some helpful comments on characters to escape
+  // See also https://tools.ietf.org/html/rfc8259#section-7 and
+  // https://github.com/google/gson/issues/341 so we escape those as well.
+  // It's legal to escape any character, so to be nice to HTML parsers,
+  // we'll also escape "<" and "&"
+  private void setEscapes() {
```
shs96c, line 192: Doing this per-json encoder is wasteful.

shs96c (changes requested):
I'm not sure making the JSON encoder non-conformant is the right approach to follow for this problem. AIUI, we're passing in a json-encoded string to a logger, and then that's being escaped again. Perhaps we should investigate why we're doing the json escaping instead?

barancev:
Can you please demonstrate what logging issue we're trying to solve by this change?

pujagani:
Thank you for the review. Closing this PR in lieu of investigating the escaped string sent to the logger for event-logs.

pujagani:
The logs currently are printed as follows: 
`{"traceId": "fc2964f4b3934af0a48df04e812acf5b","spanId": "dc660fec12498442","spanKind": "INTERNAL","eventTime": 1611575427636017843,"eventName": "Session request received by the distributor","attributes": {"logger": "org.openqa.selenium.grid.distributor.local.LocalDistributor","request.payload": "{\n  \"desiredCapabilities\": {\n    \"browserName\": \"chrome\"\n  },\n  \"capabilities\": {\n    \"firstMatch\": [\n      {\n        \"browserName\": \"chrome\"\n      }\n    ]\n  }\n}"}}` 

The reason is we leverage OpenTelemetry events and event attributes to display event-logs. Since, OpenTelemetry only accepts a string for an attribute value, for objects like payload and capabilities, we convert them to json string and add as an even-attribute. When the log statement is created, the fields from the span and events are added to a map and converted to a json string. Thus, object strings are escaped twice. The PR was aiming to as disable escaping option for such cases.

PR #9048, opened 2021-01-13
[rb] Update Rubocop

titusfortner:
Updating Ruby code to match the latest Rubocop version

The PR is to kick off the tests in GitHub Actions before merging; there shouldn't be anything controversial in this one.


titusfortner:
Well, ok, maybe controversial. @twalpole / @p0deje I can't seem to update rubocop without updating all the cached gems. Can we... do this differently? :)

titusfortner:
And it doesn't matter, I screwed things up when I added webrick to dependencies in trunk, so I needed to merge it all.

PR #9049, opened 2021-01-15
[rb] add support for intercepting assets

titusfortner:
I took a look at the Ruby DevTools code for the first time. Great work getting this imported, @p0deje!

First, does it make sense to have `#record_page_events!` and `#record_network_events` to tally when events are called?. Seems easier for basic validations if we need to do them.

[Network.setRequestInterception](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-setRequestInterception) is deprecated, so I implemented this with [Fetch.enable](https://chromedevtools.github.io/devtools-protocol/tot/Fetch/#method-enable)

The biggest issue is that the `RequestId` has to be obtained from the `requestPaused` event but this doesn't appear to be thread safe. I got it "working" for this example by using [concurrent-ruby](https://github.com/ruby-concurrency/concurrent-ruby).

Does it make sense to use concurrent-ruby to ensure we're using the correct id? Is this hacky or useful as a solution?


titusfortner:
And this is definitely wrong. I'm reworking to resubmit

titusfortner:
Accidentally commented/closed this PR instead of #9104

p0deje:
I guess this is merged in https://github.com/SeleniumHQ/selenium/commit/b9ef1f872faadc97ea9758244bb6e5cc2bca6a49

titusfortner:
@p0deje 
Oh, you only used the second commit.
Was this not a good idea? https://github.com/SeleniumHQ/selenium/pull/9049/commits/90cdf4e834a32464170bc396516346c3c74a4ee8

p0deje:
I missed that one because I only focused on network interception, sorry.
Is page_events/network_events something we have in other bindings too?

titusfortner:
No, so maybe we can discuss the correct API with the group. I was just trying to find a way to avoid using actual CDP calls in the tests as well as making it easy for users to do the things they most want to do.

PR #9068, opened 2021-01-19
Only set 'startup.homepage_welcome_url' if 'browser.startup.homepage'…

twalpole:
… is set in profile - Issue 8845

### Types of changes

### Checklist

### DIFF
--- a/rb/lib/selenium/webdriver/firefox/profile.rb
+++ b/rb/lib/selenium/webdriver/firefox/profile.rb
@@ -182,7 +182,9 @@ module Selenium
           prefs = read_user_prefs(path).merge(@additional_prefs)
 
           # If the user sets the home page, we should also start up there
-          prefs['startup.homepage_welcome_url'] = prefs['browser.startup.homepage']
+          if prefs['browser.startup.homepage']
+            prefs['startup.homepage_welcome_url'] ||= prefs['browser.startup.homepage']
+          end
 
           write_prefs prefs, path
         end


titusfortner:
I deleted the auto-generated `prefs.json` because I incorrectly thought it was only being used for the legacy Firefox extension, which we removed for Selenium 4...

Mozilla is recommending not using the Profile and using capabilities instead, but they also don't allow passing in extensions as part of the capabilities, so being able to use Profiles would be helpful.

To ensure backward compatibility with anyone using this class, I'm thinking we should put back everything that was specified as mutable from that file (assuming the frozen things were frozen because they were required for the extension).

@AutomatedTester did you say you pulled the Profile class out of Python entirely? Do you have any input on which of these should be included and/or are not the defaults?

```
    "browser.dom.window.dump.enabled": true,
    "browser.laterrun.enabled": false,
    "browser.newtab.url": "about:blank",
    "browser.newtabpage.enabled": false,
    "browser.startup.page": 0,
    "browser.startup.homepage": "about:blank",
    "browser.startup.homepage_override.mstone": "ignore",
    "browser.usedOnWindows10.introURL": "about:blank",
    "dom.max_chrome_script_run_time": 30,
    "dom.max_script_run_time": 30,
    "dom.report_all_js_exceptions": true,
    "javascript.options.showInConsole": true,
    "network.captive-portal-service.enabled": false,
    "security.csp.enable": false,
    "startup.homepage_welcome_url": "about:blank",
    "startup.homepage_welcome_url.additional": "about:blank",
```

AutomatedTester:
I haven't pulled it out but have deprecated it.

The `Options` class should have the ability to set preferences and install addons via the [extension method](https://github.com/SeleniumHQ/selenium/blob/trunk/py/selenium/webdriver/firefox/remote_connection.py#L31-L34].

titusfortner:
@AutomatedTester right, those are endpoints that can be hit after the session has started so aren't available in the Options class like they can be for Chrome.

titusfortner:
implemented in #9138 instead

PR #9077, opened 2021-01-20
[rb] ensure options classes have defaults for accessors

titusfortner:
The impetus of this refactor was to allow:
```
ChromeOptions.new.args << 'new-arg'
```
Since I accidentally removed these defaults 2 summers ago when I moved to the dynamic setters/getters approach.

This PR
1. Removes duplication for dynamically adding accessors from `CAPABILITIES` 
2. Creates the accessors during class construction instead of when file loads
3. Adds defaults in constructor for all of the `Array` & `Hash` objects, and removes from the specific methods
4. Fixes a bug where Chrome profiles were getting completely ignored in existing code
5. `#generate_as_json` now ignores any capabilities that are empty, so we don't have to put that logic in multiple laces elsewhere
6. Added specs for everything

The only weird capability now is Firefox's `profile` since we want to evaluate it when we add it instead of when we use it.

Everything else should be straightforward assuming I didn't miss some specs failing

### DIFF
--- a/rb/lib/selenium/webdriver/chrome/options.rb
+++ b/rb/lib/selenium/webdriver/chrome/options.rb
@@ -29,7 +29,6 @@ module Selenium
         # see: http://chromedriver.chromium.org/capabilities
         CAPABILITIES = {args: 'args',
                         binary: 'binary',
-                        extensions: 'extensions',
                         local_state: 'localState',
                         prefs: 'prefs',
                         detach: 'detach',
@@ -39,6 +38,9 @@ module Selenium
                         emulation: 'mobileEmulation',
                         perf_logging_prefs: 'perfLoggingPrefs',
                         window_types: 'windowTypes'}.freeze
+
+        # NOTE: special handling of 'extensions' to validate when set instead of when used
+        attr_reader :extensions
 
         # Create a new Options instance.
         #
@@ -64,21 +66,23 @@ module Selenium
         # @option opts [Array<String>] :window_types A list of window types to appear in the list of window handles
         #
 
-        def initialize(profile: nil, encoded_extensions: [], **opts)
+        def initialize(profile: nil, **opts)
           super(**opts)
 
           @profile = profile
-          @options[:encoded_extensions] = encoded_extensions
 
-          @options[:args] ||= []
-          @options[:prefs] ||= {}
-          @options[:emulation] ||= {}
-          @options[:local_state] ||= {}
-          @options[:exclude_switches] ||= []
-          @options[:perf_logging_prefs] ||= {}
-          @options[:window_types] ||= []
-          @options[:extensions] ||= []
-          @options[:extensions].each(&method(:validate_extension))
+          @options = {args: [],
+                      prefs: {},
+                      emulation: {},
+                      extensions: [],
+                      local_state: {},
+                      exclude_switches: [],
+                      perf_logging_prefs: {},
+                      window_types: []}.merge(@options)
+
+          @encoded_extensions = @options.delete(:encoded_extensions) || []
+          @extensions = []
+          (@options.delete(:extensions)).each(&method(:validate_extension))
         end
 
         #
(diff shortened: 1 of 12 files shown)


rb/lib/selenium/webdriver/chrome/options.rb
```diff
@@ -74,12 +64,21 @@ class Options < WebDriver::Options
...
         #
 
-        def initialize(profile: nil, encoded_extensions: nil, **opts)
+        def initialize(profile: nil, encoded_extensions: [], **opts)
           super(**opts)
 
           @profile = profile
-          @options[:encoded_extensions] = encoded_extensions if encoded_extensions
-          @options[:extensions]&.each(&method(:validate_extension))
+          @options[:encoded_extensions] = encoded_extensions
```
twalpole: Why do one default this way and the others differently?

rb/lib/selenium/webdriver/chrome/options.rb
```diff
@@ -74,12 +64,21 @@ class Options < WebDriver::Options
...
+
+          @options[:args] ||= []
```
twalpole: Could be handled by a hash merge rather than individual assignments.

rb/lib/selenium/webdriver/chrome/options.rb
```diff
@@ -185,16 +180,13 @@ def add_emulation(**opts)
         def process_browser_options(browser_options)
           options = browser_options[KEY]
           options['binary'] ||= binary_path if binary_path
-          (options['args'] || []) << "--user-data-dir=#{@profile.directory}" if @profile
+          options['args'] << "--user-data-dir=#{@profile.directory}" if @profile
```
twalpole: Good catch -- this potentially wasn't setting the profile directory previously

rb/lib/selenium/webdriver/firefox/options.rb
```diff
@@ -140,6 +133,7 @@ def log_level=(level)
         def process_browser_options(browser_options)
           options = browser_options[KEY]
           options['binary'] ||= Firefox.path if Firefox.path
+          options['profile'] = options.delete(:profile) if profile
```
twalpole: What is this trying to do?  `profile` is checking @options[:profile] but then it's assigning from `options` (local not instance variable) - I assume this is meant to be `if options[:profile]' which would just be changing the key from a symbol to string

rb/lib/selenium/webdriver/chrome/options.rb
titusfortner: Huh, yeah, not sure why it was this way. Looks like it is the same for 3.x, so I'll change it here as well.
https://github.com/SeleniumHQ/selenium/blob/rb-3.x/rb/lib/selenium/webdriver/chrome/options.rb#L54

rb/lib/selenium/webdriver/chrome/options.rb
titusfortner: It definitely wasn't. The author even wrote a test that explicitly showed it wasn't getting set. :)

rb/lib/selenium/webdriver/chrome/options.rb
titusfortner: oh, sure, it just always looks weird to me to add defaults that way, but it's probably a good change since it will also decrease cyclomatic complexity.

rb/lib/selenium/webdriver/firefox/options.rb
titusfortner: Yeah, this is a weird one. And looking at it, I'm treating it differently from `Chrome::Options#extensions=`, which is wrong (I'll fix).

Essentially for these we don't want to use the default setter and getter that is autogenerated by `Options#self.set_capabilities`, since we want to validate correctness on construction and setter rather than when going to use it.

When each class was setting it as part of class construction, we could override with `undef`, but I think it is better to do it at construction, which lets me put the looping logic in the superclass. The tradeoff is that I have to pull these values out of `CAPABILITIES` and handle them special.

Since they aren't in `CAPABILITIES` they aren't processed in `#as_json` and they key stays as a `Symbol` instead of correctly getting translated into a `String` (the purpose of `#as_json`. Since `options[:profile]` is the same as `profile` it seemed like a shortcut to just use `profile`.

1. Either way I need to do the same behavior for `Chrome#extensions=`
2. Would it make more sense to just put the autogenerated accessor behavior back to current and use `undef` for these?
3. The other option is we could do the processing when `#generate_capabilities` gets called during `#create_bridge` instead of during constructor/setter

titusfortner: Ok, I think I did the right thing. These will be tracked by instance variable rather than inside the options class since there's no reason to pass them around to superclass when they are only getting processed in their own subclass.

PR #9138, opened 2021-02-03
[rb] restore default Firefox Profile settings to match 3.x

titusfortner:
I think this is a better option than #9068 for maintaining backward compatibility.

These are the ones that aren't default, but make sense for testing.

* Set everything to about:blank and/or prevent extra tabs opening.

* `"network.captive-portal-service.enabled": false`:
When you launch Firefox, the browser immediately establishes a new connection to detectportal.firefox.com. This behavior is caused by Captive Portal, a special feature of Firefox. Disabling Captive Portal will stop Firefox from connecting to detectportal.firefox.com.

* `"security.csp.enable": false`:
allows including external javascript code on a page

tbh, these sound like more reasonable defaults than what geckodriver creates by default.


titusfortner:
added tests and set to allow welcome url & homepage to be different

PR #9145, opened 2021-02-05
[rb] update default cookie setting behavior with thorough documentati…

titusfortner:
I went down a rabbit hole exploring desired cookie behavior.

I added all the specs, filed all the bugs, and referenced them in the guards as applicable.

The main change I want to get reviewed is overriding the defaults when we create the cookies.
Per the WebDriver spec only name & value are required, and per the Cookie RFC, the path & the secure value are the expected defaults from the browser. As such I don't think we should be overriding the behavior, even if it might not be backward compatible (Chrome has a bug that sets secure to true by default on non-localhost sites)

@luke-hill you should be happy with this. :)

### DIFF
--- a/rb/spec/integration/selenium/webdriver/manager_spec.rb
+++ b/rb/spec/integration/selenium/webdriver/manager_spec.rb
@@ -23,103 +23,221 @@ module Selenium
   module WebDriver
     describe Manager do
       describe 'cookie management' do
+        before { driver.navigate.to url_for('xhtmlTest.html') }
+
         after { driver.manage.delete_all_cookies }
 
-        it 'should show http only when insecure' do
-          driver.navigate.to url_for('xhtmlTest.html')
-          driver.manage.add_cookie name: 'security',
-                                   value: 'insecure',
-                                   http_only: true
+        it 'should set correct defaults' do
+          driver.manage.add_cookie name: 'default',
+                                   value: 'value'
 
-          expect(driver.manage.cookie_named('security')[:http_only]).to eq true
+          cookie = driver.manage.cookie_named('default')
+          expect(cookie[:value]).to eq('value')
+          expect(cookie[:path]).to eq('/')
+          expect(cookie[:domain]).to eq('localhost')
+          expect(cookie[:http_only]).to eq(false)
+          expect(cookie[:secure]).to eq(false)
         end
 
-        it 'should not add secure when http', except: {browser: :firefox,
-                                                       reason: 'https://github.com/mozilla/geckodriver/issues/1840'} do
-          driver.navigate.to url_for('xhtmlTest.html')
-          driver.manage.add_cookie name: 'security',
-                                   value: 'secure',
-                                   secure: true
+        it 'should set samesite property of Default by default',
+           only: {browser: %i[chrome edge firefox]},
+           except: [{browser: :chrome,
+                     reason: 'https://bugs.chromium.org/p/chromedriver/issues/detail?id=3732'},
+                    {browser: :firefox,
+                     reason: 'https://github.com/mozilla/geckodriver/issues/1841'}] do
+          driver.manage.add_cookie name: 'samesite',
+                                   value: 'default'
 
-          cookies = driver.manage.all_cookies
-          expect(cookies.size).to eq(0)
+          expect(driver.manage.cookie_named('samesite')[:same_site]).to eq('Default')
         end
 
         it 'should respect path' do
-          driver.navigate.to url_for('xhtmlTest.html')
           driver.manage.add_cookie name: 'path',
                                    value: 'specified',
                                    path: '/child'
-          cookies = driver.manage.all_cookies
-          expect(cookies.size).to eq(0)
+
+          expect(driver.manage.all_cookies.size).to eq(0)
 
           driver.navigate.to url_for('child/childPage.html')
+
           expect(driver.manage.cookie_named('path')[:path]).to eq '/child'
         end
 
-        it 'should add expiration with DateTime' do
-          driver.navigate.to url_for('xhtmlTest.html')
+        it 'should respect setting on domain from a subdomain',
+           exclusive: {driver: :none,
+                       reason: "Can only be tested on site with subdomains"} do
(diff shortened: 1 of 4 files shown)


luke-hill:
What an absolute champ! I will review this properly either Friday or Monday. Just got a bit on my plate atm.

titusfortner:
I updated the PR to keep `secure` attribute value defaulting to `false`, and made note of the Chrome bug that requires this to be kept.

rb/spec/integration/selenium/webdriver/manager_spec.rb
```diff
@@ -53,103 +53,221 @@ module WebDriver
...
-                                   value: 'datetime',
-                                   expires: expected
+          driver.get("https://accounts.saucelabs.com")
+          expect(driver.manage.cookie_named('domain')[:domain]).to eq('.saucelabs.com')
 
-          actual = driver.manage.cookie_named('expiration')[:expires]
-          expect(actual).to be_kind_of(DateTime)
-          expect(actual).to eq(expected)
+          driver.get("https://saucelabs.com")
+          expect(driver.manage.cookie_named('domain')[:domain]).to eq('.saucelabs.com')
```
luke-hill, line 81: This is good 👍

rb/spec/integration/selenium/webdriver/manager_spec.rb
```diff
@@ -53,103 +53,221 @@ module WebDriver
...
+
+          expect(driver.execute_script("return document.cookie")).to be_empty
+          expect(driver.manage.cookie_named('httponly')[:http_only]).to eq true
+        end
 
-          expected = (Date.today + 2).to_datetime
-          driver.manage.add_cookie name: 'expiration',
-                                   value: 'number',
-                                   expires: expected.to_time.to_f
+        it 'should not add secure cookie when http',
```
luke-hill, line 127: I don't understand this test. Although maybe I'm missing a setup step. Shouldn't this first navigate to something like `http://insecure.com/` and then attempt to set a secure cookie?

rb/spec/integration/selenium/webdriver/manager_spec.rb
```diff
@@ -53,103 +53,221 @@ module WebDriver
...
+          expect(driver.manage.cookie_named('secure')[:secure]).to eq(true)
         end
 
-        it 'should add sameSite cookie with attribute Lax', only: {browser: %i[chrome edge firefox]} do
-          driver.navigate.to url_for('xhtmlTest.html')
-          driver.manage.add_cookie name: 'samesite',
-                                   value: 'lax',
-                                   same_site: 'Lax'
-          expect(driver.manage.cookie_named('samesite')[:same_site]).to eq('Lax')
+        context 'sameSite' do
```
luke-hill, line 149: Question. Is this essentially showing that sameSite can be anything? Is this desirable? Or should it be some sort of set of values?

rb/spec/integration/selenium/webdriver/manager_spec.rb
```diff
@@ -53,103 +53,221 @@ module WebDriver
...
+              driver.manage.add_cookie name: 'samesite',
+                                       value: 'none-insecure',
+                                       same_site: 'None',
+                                       secure: false
+            }.to raise_exception(Error::UnableToSetCookieError)
+          end
+        end
+
+        context 'expiration' do
+          it 'should add with DateTime' do
```
luke-hill: for all of these how about?

`it can be a DateTime`
`can be a Time`
`can be a Float` or Number?
`cannot be any value in the past`

luke-hill (commented):
Looks awesome. Just a few generic questions first. I'm probably wrong on them all.

rb/spec/integration/selenium/webdriver/manager_spec.rb
titusfortner: it starts up in a non-secure protocol, but sure might as well add navigation first.

rb/spec/integration/selenium/webdriver/manager_spec.rb
titusfortner: I'm not sure what you mean. I'm using "context" just for organization in the spec. The values being tested are still just strict, lax & none

rb/spec/integration/selenium/webdriver/manager_spec.rb
titusfortner: as far as naming? Yeah, ideally we get rid of all of the "shoulds," I try to keep it consistent within the file at least and didn't feel like changing the others. :)

rb/spec/integration/selenium/webdriver/manager_spec.rb
titusfortner: oh yeah, I state http in the spec title, but then don't navigate to http 👍

rb/spec/integration/selenium/webdriver/manager_spec.rb
luke-hill: oki, then we could change to `should be a DateTime` e.t.c. I think the "add with" is confusing here.

rb/spec/integration/selenium/webdriver/manager_spec.rb
luke-hill: What I mean is, what restrictions are there on the value of `sameSite` I'm not familiar with why you picked these values? Are those values the only permitted ones?

titusfortner: Yes; when in doubt search the spec: https://w3c.github.io/webdriver/#dfn-cookie-same-site

rb/spec/integration/selenium/webdriver/manager_spec.rb
titusfortner: You have to read the whole thing: 
> cookie management expiration should add with DateTime

better: 
> cookie management expiration should allow adding with DateTime value

ideally:
> cookie management expiration respects DateTime value

luke-hill: No I know that, just that first sentence sounds bad imo. Doesn't read well in english. Other 2 are fine.

PR #9192, opened 2021-02-18
Additional Python type hints

AlbertoCentonze:
### Description
I wasn't able to get the right suggestion for the code completion using pylance in vs code. That's why I started adding a few lines to improve the suggestions

### Motivation and Context
 I made these changes because I was working on a personal project and having those "type hints" really helped me

### Types of changes
I only changed a few lines adding type hints. If this kind of changes are welcomed I'll continue to update the hints in the future when I use selenium and I need it

### Checklist


AutomatedTester (approved):
I had left those because the methods are deprecated but we can add it for now and then once Selenium 4 is released we will be deleting them

AlbertoCentonze:
@AutomatedTester Where can I give a look at the new methods in Selenium 4?

AutomatedTester (approved):

hudcap (commented):
I believe all the `find_elements` (plural) methods should return `list[WebElement]` and not `WebElement`
Am I mistaken?
Should I open a separate PR for this?

Issue #9152, opened 2021-02-07
Server has a memory leak, io.netty.buffer.PoolChunk consumes a lot of memory


barancev:
## 🐛 Bug Report

![image](https://user-images.githubusercontent.com/617090/107141461-df93f480-6939-11eb-88b8-ab979c731c47.png)

## To Reproduce

Start a standalone selenium server, run a simple scenario (init a driver, find a link, click, quit) 10-15 times and create a heap dump.

It appears that we don't release the incoming message buffer properly.

For reference:
A simiar issue in micronaut: https://github.com/micronaut-projects/micronaut-core/issues/2665

## Environment

OS: Win10
Browser: any
Browser version: any
Browser Driver version: any
Language Bindings version: Java
Selenium Grid version (if applicable): 4.0-alpha-7

diemol:
After checking for a few days this issue, here are my findings:
- The memory shown on the screen is memory used and managed by Netty. There was the doubt if we were actually deallocating all objects used while processing incoming and outgoing requests, but after checking the Netty guidelines and our code, what we implemented follows the recommendations from the Netty docs. 
- We use internally [`AsyncHttpClient`](https://github.com/AsyncHttpClient/async-http-client), which implcitely uses a thread pool to handle connections. The default size of this pool is # of available processors * 2. As the pool grows up to its limit, more memory is being used up to the point where the pool is full, where memory usage gets stable, which is simply how the client works. Therefore, memory usage is proportional to the pool size.

General memory usage can be seen in the following images:

**Standalone**
Memory usage is in general stable
![Standalone - 2021-02-23 at 11 23 18](https://user-images.githubusercontent.com/5992658/109653061-d37a0c00-7b60-11eb-8486-ef05aa64aeed.png)

**Hub/Node**
Hub - It is possible to see how memory usage goes up. In this case there is only one node registered. When more nodes are registered, the incremeent of memory usage is more visible.
![Hub - 2021-02-23 at 14 35 39](https://user-images.githubusercontent.com/5992658/109653447-569b6200-7b61-11eb-8ca2-2a3c7c3b6dc4.png)
Node - Memory usage is in general stable
![Node - 2021-02-23 at 14 36 09](https://user-images.githubusercontent.com/5992658/109653497-62872400-7b61-11eb-8c20-d360146978ee.png)

**Fully Distributed**
The only relevant component here is the Distributor.
[...]

PR #9214, opened 2021-02-25
[grid] Setting Netty client pool size to 1

diemol:

### Description

The AsyncHttpClient has built in a thread pool
to handle connections and the pool was growing
based on the way we use the client. As a result,
we were having many threads that kept objects in
memory that were not getting released, a memory
leak.

This is not an issue in the AsyncHttpClient but
most likely in the way we use the client.

### Motivation and Context

To prevent memory from leaking. This is a fix for #9152

### Types of changes

### Checklist


diemol:
Closing as this is not the solution we want but helped to show that the extra use of memory is due to the way we use `AsyncHttpClient`

PR #9219, opened 2021-02-28
Attribute & Property Methods

titusfortner:
Looking to get agreement from the other maintainers before committing...

~**Update**: Looks like Java & .NET are both converting property Boolean responses to Strings, and Ruby & Python are not. Why would we override the spec/driver result in the code?~ (Static language vs Dynamic language thing, so this is good)

### Python
* implements `get_dom_attribute` 
* Adds a test for `get_dom_attribute` and `get_property`
* I've verified tests are passing, but @AutomatedTester, can you tell me if I did the docstring method reference correctly?

### Java, Python & Ruby
* switch from using the `getAttribute.js` atom to using the  w3c `/attribute` endpoint for determining if a Select element is a multiple select list. This should be completely backward compatible, even if they use JS to remove the value, it will still return the same value as the atom.

### Ruby
* implements `#dom_attribute`
* Adds 53 tests that should well-differentiate the expected/desired behavior for the 3 different methods we now have for working with properties/attributes, along with references to the applicable Chrome & Firefox bugs

### DIFF
--- a/dotnet/src/webdriver/Remote/RemoteWebElement.cs
+++ b/dotnet/src/webdriver/Remote/RemoteWebElement.cs
@@ -417,7 +417,7 @@ namespace OpenQA.Selenium.Remote
         /// </item>
         /// </list>
         /// The method looks both in declared attributes in the HTML markup of the page, and
-        /// in the properties of the elemnt as found when accessing the element's properties
+        /// in the properties of the element as found when accessing the element's properties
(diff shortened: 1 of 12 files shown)


p0deje (dismissed):
Ruby changes are great, thanks for all that specs - massive work!

AutomatedTester (dismissed):
Python code looks fine

PR #9220, opened 2021-02-28
Ruby selenium-devtools gem

titusfortner:
I think this is the best I can do, so I'd love some eyes on it to make sure it's reasonable.

* 2!?! gemspecs. This was weirder/more complicated than I expected, but I think I did everything correctly here.
* Using `selenium-devtools.gem` since Java is doing `selenium-devtools-v88` for their jars
* I wasn't sure about namespacing. I decided that the gems should have different namespaces and that `Selenium::WebDriver::DevTools` would use `#method_missing` to point to `Selenium::DevTools`
* User doesn't need to require the gem, though currently they need to add it to their gemfile if they want to use it.
* The gem is essentially built entirely dynamically, so I'm thinking we support a range of versions at a time; I'm starting with 84 - 88 because that's how far back Java has maven jars for. We should do 3 at a time going forward, and the user can specify in gemfile which one they want to freeze on if they need to.
* separated build commands `bidi` & `devtools` to limit what is required for building devtools gem
* new namespacing for build/release: `./go //rb:gem:webdriver:release` and `./go//rb:gem:devtools:release`
* I already released alphas! https://rubygems.org/gems/selenium-devtools

I can already see there's some minor housekeeping I can do still, but please try this out and let me know if there is something not working.


rake_tasks/crazy_fun/mappings/ruby_mappings/ruby_gem.rb
```diff
@@ -30,16 +30,17 @@ def define_build_task(dir, args)
           end
         end
 
-        def define_clean_task(dir, _args)
+        def define_clean_task(dir, args)
```
twalpole: Can we just destructure name here `def define_clean_task(dir, name:, **_args)` or is args not guaranteed to be keyword arguments?

rake_tasks/crazy_fun/mappings/ruby_mappings/ruby_gem.rb
```diff
@@ -30,16 +30,17 @@ def define_build_task(dir, args)
...
           desc 'Clean rubygem artifacts'
-          task "//#{dir}:gem:clean" do
+          task "//#{dir}:gem:#{args[:name]}:clean" do
             Dir['build/*.gem'].each { |gem| rm(gem) }
           end
         end
 
-        def define_release_task(dir, _args)
+        def define_release_task(dir, args)
```
twalpole: same

rb/lib/selenium/webdriver/devtools.rb
```diff
@@ -60,16 +57,24 @@ def send_cmd(method, **params)
...
+        desired_class = "Selenium::DevTools::V#{Selenium::DevTools.version}::#{method.capitalize}"
+        return unless Object.const_defined?(desired_class)
 
-      def load_devtools_version(version)
-        closest_version = SUPPORTED_VERSIONS.min_by { |v| (version - v).abs }
-        WebDriver.logger.info("Loading DevTools::V#{closest_version} for #{version}.")
-        Dir["#{__dir__}/devtools/v#{closest_version}/*"].sort.each do |f|
-          require f
-        end
+        return instance_variable_get("@#{method}") if instance_variable_defined?("@#{method}")
```
twalpole: Could dynamically define the method here so don't pay the method_missing cost after first access and don't have to deal with caching in instance variables.

rb/lib/selenium/webdriver/devtools.rb
```diff
@@ -60,16 +57,24 @@ def send_cmd(method, **params)
...
+
+        instance = Object.const_get(desired_class).new(self)
+        instance_variable_set("@#{method}", instance)
+        instance
       end
 
+      def respond_to_missing?(method, *_args)
+        desired_class = "Selenium::DevTools::#{method.capitalize}"
```
twalpole: Shouldn't this include the version?

twalpole (commented):
@titusfortner Sorry - haven't had a chance to try this out, but commented on a few things that stood out to me on reading the code.

rb/lib/selenium/webdriver/devtools.rb
titusfortner: I actually tried to do that first, and something I was doing wasn't working, this is what I got to work. I can try again.

rake_tasks/crazy_fun/mappings/ruby_mappings/ruby_gem.rb
titusfortner: It is a Hash that is dynamically generated from the `build.desc` file. Not sure why it was called args, and not sure why we were intentionally passing something that doesn't get used. I'll change it so it only passes what is getting used here.

rb/lib/selenium/webdriver/devtools.rb
titusfortner: Good catch! I changed it one place and not the other. Thanks.

PR #9208, opened 2021-02-24
Fixed headless option not being passed to RemoteWebDriver constructor

Dmitri-Sintsov:
**Thanks for contributing to Selenium!**
**A PR well described will help maintainers to quickly review and merge it**

Before submitting your PR, please check our [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) guidelines.
Avoid large PRs, help reviewers by making them as simple and short as possible.

### Description
Fixed `headless` option not being passed to `RemoteWebDriver` constructor `desired_capatibilities` argument by`ChromiumDriver` constructor. `desired_capatibilities` are used to pass `--headless` option to browser.

### Motivation and Context
Please see https://github.com/SeleniumHQ/selenium/issues/9188

### Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist

- [x ] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #9188, opened 2021-02-17
Headless mode does not work in 4.0.0b1


Dmitri-Sintsov:
## 🐛 Bug Report

Running in non-X environment causes the following error:

```
[1613588397.830][INFO]: Launching chrome: /usr/bin/google-chrome --disable-background-networking --disable-client-side-phishing-detection --disable-default-apps --disable-hang-monitor --disable-popup-blocking --disable-prompt-on-repost --disable-sync --enable-automation --enable-blink-features=ShadowDOMV0 --enable-logging --log-level=0 --no-first-run --no-service-autorun --password-store=basic --remote-debugging-port=0 --test-type=webdriver --use-mock-keychain --user-data-dir=/tmp/.com.google.Chrome.iYvAAX data:,
mkdir: cannot create directory ‘/.local’: Permission denied
touch: cannot touch '/.local/share/applications/mimeapps.list': No such file or directory
[2619:2619:0217/185959.352421:ERROR:browser_main_loop.cc(1438)] Unable to open X display.
[1613588399.440][INFO]: [690fd84376bd17e319de37d312fc1c35] RESPONSE InitSession ERROR unknown error: Chrome failed to start: exited abnormally.
  (unknown error: DevToolsActivePort file doesn't exist)
  (The process started from chrome location /usr/bin/google-chrome is no longer running, so ChromeDriver is assuming that Chrome has crashed.)
[1613588399.440][DEBUG]: Log type 'driver' lost 0 entries on destruction
[1613588399.440][DEBUG]: Log type 'browser' lost 0 entries on destruction

```

## To Reproduce
Detailed steps to reproduce the behavior:

https://github.com/Dmitri-Sintsov/django-jinja-knockout/runs/1921312027?check_suite_focus=true

When running CI manually, headless mode is ignored, browser window is opening with X:
pip3 install selenium==4.0.0b1

The same CI tests works fine in headless mode with selenium==4.0.0a7 and selenium-3.141.0

## Expected behavior

Expected to run in headless mode not looking for X display.

## Test script or set of commands reproducing this issue

One may run tests manually as described at:
https://github.com/Dmitri-Sintsov/djk-sample

Issues without a reproduction script are likely to stall and eventually be closed.

## Environment

OS: Ubuntu 18.04.5 LTS
Browser: Chrome
Browser version: 88.0.4324.150
[...]

PR #7348, opened 2019-07-01
[py] Chrome options override desired capabilities

earendil:
* Pass options to the remote driver (future proof)
* Convert desired caps to options rather than the other way around

Fixes #5709

- [x] By placing an `X` in the preceding checkbox, I verify that I have signed the [Contributor License Agreement](https://github.com/SeleniumHQ/selenium/blob/master/CONTRIBUTING.md#step-6-sign-the-cla)

PR #9240, opened 2021-03-07
Pass options to RemoteWebDriver

nijel:
### Description
I believe that was the original intention of code in c23e4406a2a94f6546fe3bc69dcd9662126dc7d0 - it updates or creates the options object, but it doesn't use it later.

This reverts #9208 and passes options to the super class.

In case passing desired_capabilities instead of options to super class is really intended, a lot of code dealing with the `options` object can be removed because it is never used here.

CC @Dmitri-Sintsov, @earendil, @AutomatedTester 

### Motivation and Context

Please see #9188, #9208 and #7348

### Types of changes

### Checklist


Dmitri-Sintsov:
Would it work with headless option? In case `desired_capabilities` are deprecated, these should be removed everywhere, including `RemoteWebDriver` where these are currently used for headless option.

nijel:
Yes it works headless. I did not orignally discover your fix and did fix it this way independently...

There is code which puts `desired_capabilities` into `options` and `options` are not used after that point:

https://github.com/SeleniumHQ/selenium/blob/9c41b0b5312cdaa70ed085665b5b0a04d91745e5/py/selenium/webdriver/chromium/webdriver.py#L75-L77

So either this is useless and should be removed or it is desired and then the `options` should be used later. This code was changed in c23e4406a2a94f6546fe3bc69dcd9662126dc7d0.

Maybe even `desired_capabilities` should be removed from the ` RemoteWebDriver.__init__` params as it is deprecated and it is put into the `options`. 

I'm happy to improve or close the pull request based on the feedback, but my knowledge of Selenium internals is very limited.

Dmitri-Sintsov:
Just checked with my CI, indeed your version works in headless mode as well! I do not understand then why `desired_capabilities` are deprecated only in `ChromiumDriver` but not in `RemoteWebDriver` then? Maybe `desired_capabilities` should be removed everywhere, then?

AutomatedTester:
> Just checked with my CI, indeed your version works in headless mode as well! I do not understand then why `desired_capabilities` are deprecated only in `ChromiumDriver` but not in `RemoteWebDriver` then? Maybe `desired_capabilities` should be removed everywhere, then?

The reason is I want to get Selenium 4 out the door first before as it's going to be a larger breaking change for people who use Selenium grid. Perhaps I am being overly cautious...

AutomatedTester (approved):

PR #9252, opened 2021-03-10
Microsoft has changed the name of the Edge webdriver but Selenium is still expecting the old name.

anebot:
An exception is thrown when you instantiate the EdgeDriver class in .NET, It seems the name of the webdriver binary has changed, but Selenium is expecting the previous name.


AutomatedTester:
It seems to have the right value. What it seems is that you're expecting chromium edge as the default and we're not doing that. This is a much larger issue

jimevans:
The default value is incorrect, that much is true, but this is the incorrect fix. What would set the correct default is to initialize the `usesChromium` field to `true`. As a workaround in the meantime, the value should be settable at runtime via a property.

PR #9263, opened 2021-03-11
EdgeDriver, Set UseChromium option as true by default.

anebot:
Set the UseChromium property as true by default as Edge Legacy is [no longer supported](https://support.microsoft.com/en-us/microsoft-edge/what-is-microsoft-edge-legacy-3e779e55-4c55-08e6-ecc8-2333768c0fb0).


jimevans:
This should be done in line 54 of the `EdgeOptions.cs` file, setting the default value of the member variable instead of setting the property in the constructor.

jimevans:
I'm so very sorry. I inadvertently committed a fix for this that bypassed this particular PR. I've credited you with the fix in the CHANGELOG.

PR #9273, opened 2021-03-12
Allow RelativeBy to start with any locator, not just tag name

JustasMonkev:
… tests for these methods

added new methods for RelativeBy(withXpath,withCssSelector)

### Description
Added new Methods for RelativeLocator to improve UX. 

### Motivation and Context

I believe it improves UX. now it's easier to use RelativeLocator.  

### Types of changes

- [ x] New feature (non-breaking change which adds functionality)

### Checklist

- [x ] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) document.
- [ x] My change requires a change to the documentation.
- [ -] I have updated the documentation accordingly.
- [ x] I have added tests to cover my changes.
- [ x] All new and existing tests passed.


java/client/test/org/openqa/selenium/support/locators/RelativeLocatorTest.java
```diff
@@ -43,6 +46,30 @@ public void shouldBeAbleToFindElementsAboveAnother() {
     assertThat(ids).containsExactly("mid", "above");
   }
 
+  @Test
+  public void shouldBeAbleToFindElementsAboveAnotherWithXpath() {
+    driver.get(appServer.whereIs("relative_locators.html"));
+
+    WebElement lowest = driver.findElement(By.id("below"));
+
+    List<WebElement> elements = driver.findElements(withXpath("//p").above(lowest));
```
shs96c: The xpath locators rooted with `//` will (because of how xpath works) scan the entire document. My expectation with this is that we'll end up returning the first `p` tag in the entire doc. I'm surprised this works as expected....

java/client/test/org/openqa/selenium/support/locators/RelativeLocatorTest.java
```diff
@@ -54,11 +81,67 @@ public void shouldBeAbleToCombineFilters() {
   }
 
   @Test
-  public void exerciseNearLocator() {
+  public void shouldBeAbleToCombineFiltersWithXpath() {
+    driver.get(appServer.whereIs("relative_locators.html"));
+
+    List<WebElement> seen = driver.findElements(withXpath("//td").above(By.id("center")).toRightOf(By.id("second")));
```
shs96c: Again, this should be returning the very first `td` in the document.

java/client/src/org/openqa/selenium/support/locators/RelativeLocator.java
```diff
@@ -97,6 +99,16 @@ public static RelativeBy withTagName(String tagName) {
     return new RelativeBy(By.tagName(tagName));
   }
 
+  public static RelativeBy withXpath(String xpathExpression) {
```
shs96c: It may be better to just allow someone to pass in an arbitrary locator at this point. That is, replace `withTagName(String)` with `with(By)` so folks can write `with(tagName("td"))` or whatever takes their fancy.

shs96c (changes requested):
Hi! Thank you for the PR! I love the idea behind it, and it's going to be really helpful, but I think we should consider just passing in a general locator (a `By` subclass) to be even more useful. What do you think? That'd mean replacing `withTagName(String)` with `with(By)`

java/client/test/org/openqa/selenium/support/locators/RelativeLocatorTest.java
JustasMonkev: should I use this xpath //p[position()=1]?

JustasMonkev:
please take a look @shs96c

java/client/src/org/openqa/selenium/support/locators/RelativeLocator.java
```diff
@@ -87,14 +88,31 @@
...
    * Start of a relative locator, finding elements by tag name.
    */
-  public static RelativeBy withTagName(String tagName) {
+
+  public static RelativeBy with(By by) {
+    Require.nonNull("By to look for", by);
+    return new RelativeBy(by);
+  }
+
+  public static By tagName(String tagName) {
```
shs96c, line 103: These can all be safely removed now that `with(By)` exists. People can use a static import to get a short `tagName` method call to allow `with(tagName("foo"))`

shs96c (approved):
Hi! This looks great. Thank you for working on it :) There's a couple of things that we can clean up, but it won't take long to do that, and I'll make the changes once this is landed.

shs96c:
LGTM, and I think this will make using the relative locators a lot nicer. Thank you!

PR #9335, opened 2021-03-25
[rb] fix bugs and allow saving print page

### LINKED ISSUE #9298: Ruby implementation of print page expects snake-cased option and doesn’t transform it to camel-cased parameter
## 🐛 Bug Report

Ruby implementation supports WebDriver print thanks to `#print_page` in:

* `Selenium::WebDriver::DriverExtensions::PrintsPage` – currently applied to Chrome and Firefox drivers, and
* `Selenium::WebDriver::Remote::Driver`.

They both call `Selenium::WebDriver::Remote::Bridge#print_page` by just passing given options to this method. However, they both check `:page_ranges` option and make sure it is set to an array.

There are some issues with this:

1. Ruby option `:page_ranges` is never transformed to a WebDriver parameter named `pageRanges`.
2. The print page command has additional camel-cased parameters and none of them are mentioned.

Therefore, you have to specify `#print_page` options using camel-cased keys (e.g., `:pageRanges` or `:shrinkToFit`) and cannot rely on enforcement of page ranges data structure.

## To Reproduce

1. Start dedicated GeckoDriver with debug log level.
2. Start a Ruby runtime (e.g., IRB).
3. Require `selenium-webdriver` Ruby gem.
4. Create a new remote driver with Firefox capabilities and URL pointing to GeckoDriver started in the previous step.
5. Navigate to a document with multiple pages (e.g., by using data URL with content like this `<p style="page-break-after:always">1</p><p>2</p>`).
6. Print the page by specifying an option hinted by the source code: `page_ranges: '1'`.
7. Check GeckoDriver log output. You can see it has received `POST /session/{session id}/print {"options":{"page_ranges":["1"]}}`
8. Check the next line with transformation to Marionette command which shows `"pageRanges":[]` is being passed.

## Expected behavior

The code should either:

* transform snake-cased Ruby options to camel-cased WebDriver parameters to comply with WebDriver and its naming conventions, or
* stop pretending it provides some value by enforcing a data structure of an option which is never going to be used.

## Test script or set of commands reproducing this issue

```rb
require 'selenium-webdriver'
driver = Selenium::WebDriver::Driver.for :remote, capabilities: :firefox, url: 'http://127.0.0.1:4444'
# Or for inline log output instead of an independent GeckoDriver process:
#Selenium::WebDriver.logger.level = :debug
#driver = Selenium::WebDriver::Driver.for :firefox, service: Selenium::WebDriver::Firefox::Service.new(args: ['--log=debug'])
driver.navigate.to 'data:text/html;base64,' + [
  '<p style="page-break-after:always">1</p><p>2</p>'
].pack('m').lines.map(&:chomp).join
driver.print_page page_ranges: '1'
# Notice page_ranges value
# [GeckoDriver] webdriver::server	DEBUG	-> POST /session/00000000-0000-0000-0000-000000000000/print {"options":{"page_ranges":["1"]}}
# Notice pageRanges value
# [GeckoDriver] Marionette	DEBUG	0 -> [0,3,"WebDriver:Print",{"background":false,"margin":{"bottom":1,"left":1,"right":1,"top":1},"orientation":"portrait","page":{"height":27.94,"width":21.59},"pageRanges":[],"scale":1,"shrinkToFit":true}]
driver.quit
```

## Environment

OS: macOS 11.2.3
Browser: Firefox
Browser version: 86.0.1
Browser Driver version: GeckoDriver 0.29.0
Language Bindings version: Ruby 4.0.0.beta2
Selenium Grid version (if applicable): N/A

[spec-print-page]: https://www.w3.org/TR/2020/WD-webdriver2-20200824/#print-page

titusfortner:
Fixes #9298 

This implementation assumes each parameter will be a snake_case `Symbol`. According to the spec there are only two valid parameters that are multiple words, so this converts each to a camelCase `Symbol`. We can do something more robust here if we don't feel like this is good enough.

I pulled the method from `Remote::Driver` because `Driver` superclass is now extending `Extensions` for the applicable browser, so it is no longer necessary.

Also, I'm not sure why we weren't matching what we do in `TakesScreenshot` module, so I added a `#save_print_page` method while I was here.

Also, our specs are a little frustrating because they don't (can't?) test whether the parameters work as intended. 

@f4z4on does this seem like a workable solution?

### DIFF
--- a/rb/lib/selenium/webdriver/common/driver_extensions/prints_page.rb
+++ b/rb/lib/selenium/webdriver/common/driver_extensions/prints_page.rb
@@ -21,8 +21,35 @@ module Selenium
   module WebDriver
     module DriverExtensions
       module PrintsPage
+        #
+        # Save a page as a PDF to the given path
+        #
+        # @example Save Printed Page
+        #   driver.save_print_page('../printed_page.pdf')
+        #
+        # @param [String] path to where the pdf should be saved
+        #
+        # @api public
+        #
+
+        def save_print_page(path, **options)
+          File.open(path, 'wb') do |file|
+            content = Base64.decode64 print_page(options)
+            file << content
+          end
+        end
+
+        #
+        # Return a Base64 encoded Print Page as a string
+        #
+        # @see https://w3c.github.io/webdriver/#print-page
+        #
+        # @api public
+        #
+
         def print_page(**options)
-          options[:page_ranges] &&= Array(options[:page_ranges])
+          options[:pageRanges] = Array(options.delete(:page_ranges)) || []
+          options[:shrinkToFit] = options.delete(:shrink_to_fit) { true }
 
           @bridge.print_page(options)
         end
--- a/rb/lib/selenium/webdriver/remote/driver.rb
+++ b/rb/lib/selenium/webdriver/remote/driver.rb
@@ -44,12 +44,6 @@ module Selenium
           super
         end
 
-        def print_page(**options)
-          options[:page_ranges] &&= Array(options[:page_ranges])
-
-          @bridge.print_page(options)
-        end
-
         private
 
         def devtools_address
--- a/rb/spec/integration/selenium/webdriver/chrome/driver_spec.rb
+++ b/rb/spec/integration/selenium/webdriver/chrome/driver_spec.rb
@@ -53,7 +53,7 @@ module Selenium
           end
         end
 
-        describe '#print_options' do
+        describe 'PrintsPage' do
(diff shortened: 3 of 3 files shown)


p0deje (dismissed):

f4z4on:
I’m sorry I haven’t replied sooner, Titus. The proposed changes certainly seem the work. As I mentioned on my original issue, I’m not familiar with the implementation conventions of the project and I personally write things differently, so it’s hard to judge or help. But really appreciate the work and, on the outside, it will definitely make my code look more consistent! 👏

PR #9363, opened 2021-04-08
[dotnet] Make SlowLoadable component's timeout and wait() protected

### LINKED ISSUE #9009: Modify the SlowLoadableComponent class to be open for extension
## 🚀 Feature Proposal

Modify the `SlowLoadableComponent` class to be open for extension, following the [Open-closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle):

> In object-oriented programming, the open–closed principle states "software entities (classes, modules, functions, etc.) should be open for extension, but closed for modification";[1] that is, such an entity can allow its behaviour to be extended without modifying its source code.

Specifically:

> At the time Meyer was writing, adding fields or functions to a library inevitably required changes to any programs depending on that library.[citation needed] Meyer's proposed solution to this dilemma relied on the notion of object-oriented inheritance (specifically implementation inheritance):
A class is closed, since it may be compiled, stored in a library, baselined, and used by client classes. But it is also open, since any new class may use it as parent, adding new features. When a descendant class is defined, there is no need to change the original or to disturb its clients.[4]

## Motivation

Right now `SlowLoadableComponent` allows a derived class to inherit from it and override how the `Load` method works:

https://github.com/SeleniumHQ/selenium/blob/bf2fc564eeaeecfd980984119bd18d0a00bdb29e/dotnet/src/support/UI/SlowLoadableComponent%7BT%7D.cs#L72-L110

It also lets derived classes access to SleepInterval:

https://github.com/SeleniumHQ/selenium/blob/bf2fc564eeaeecfd980984119bd18d0a00bdb29e/dotnet/src/support/UI/SlowLoadableComponent%7BT%7D.cs#L63-L70

And IsLoaded:

https://github.com/SeleniumHQ/selenium/blob/bf2fc564eeaeecfd980984119bd18d0a00bdb29e/dotnet/src/support/UI/LoadableComponent%7BT%7D.cs#L51-L74

It does **not**, however, let us access `clock` and `timeout` fields, and the `Wait` method:

https://github.com/SeleniumHQ/selenium/blob/bf2fc564eeaeecfd980984119bd18d0a00bdb29e/dotnet/src/support/UI/SlowLoadableComponent%7BT%7D.cs#L39-L40

https://github.com/SeleniumHQ/selenium/blob/bf2fc564eeaeecfd980984119bd18d0a00bdb29e/dotnet/src/support/UI/SlowLoadableComponent%7BT%7D.cs#L124-L127

Which stops us from changing how the `Load` method works and still be able to wait for the `IsLoaded` condition to be true.

**I propose `clock`, `timeout` and `Wait` become `protected`.**

## Example

Let's imagine a component that we now for a fact will only load when the user executes an action (for example, clicking on a button). and we know the component will load slowly. We could skip the first evaluation of `IsLoaded`:

```csharp
public override T Load()
{
     /*if (this.IsLoaded) 
     //{ 
     //    return (T)this; 
     //} 
     //else 
     //{ 
     //    this.TryLoad(); 
     //} 
     // we skip this party and go straight to trying to loading the component
     */

    this.TryLoad();

    DateTime end = this.clock.LaterBy(this.timeout);

    while (this.clock.IsNowBefore(end))
    {
        if (this.IsLoaded)
        {
            return (T)this;
        }

        this.HandleErrors();
        this.Wait();
    }

    if (this.IsLoaded)
    {
        return (T)this;
    }
    else
    {
        string timeoutMessage = string.Format(CultureInfo.InvariantCulture, "Timed out after {0} seconds.", this.timeout.TotalSeconds);
        throw new WebDriverTimeoutException(timeoutMessage);
        }
    }
```
This implementation will not work because we do not have access to `clock`, `timeout` and `Wait`

The same can be achieved by saving a local copy of the fields on a derived class and recreating the `Wait` method:

```csharp
        private readonly TimeSpan timeout;
        private readonly IClock clock;

        protected CustomSlowLoadableComponent(TimeSpan timeout) : this(timeout, new SystemClock())
        {
        }

        protected CustomSlowLoadableComponent(TimeSpan timeout, IClock clock) : base(timeout, clock)
        {
            this.timeout = timeout;
            this.clock = clock;
        }

        ...

       private void Wait()
        {
            Thread.Sleep(SleepInterval);
        }
```
But then again, we would be adding "unnecesary" code.

Thank you.

pujagani:

### Description

Fixes #9009.

### Motivation and Context

As described in #9009, the changes update the modifier for wait() method to be accessible to the subclasses and provide a getter for the timeout variable for the same. 
The clock variable uses the IClock interface, so making it accessible to subclasses does not allow the use of a subclass specific method, hence instead creating and using a local variable copy will suffice as shown in the tests here https://github.com/SeleniumHQ/selenium/blob/trunk/dotnet/test/support/UI/SlowLoadableComponentTest.cs#L147. 

### Types of changes

### Checklist


jimevans:
I'm going to go ahead and close this PR, as it's not complete (doesn't expose the `Clock` as requested, and doesn't make the `Wait` method virtual for overriding in subclasses). Commit 5632da3 already contains the changes needed in this PR.

PR #9368, opened 2021-04-09
[dotnet] Call the base constructor to set capability for EdgeDriver

### LINKED ISSUE #8358: EdgeDriver can't create DevToolsSession with CreateDevToolsSession() method
## 🐛 Bug Report

When initialize the EdgeDriver the property optionsCapabilityName is not initialize since the base constructor is never called, therefor the CreateDevToolsSession method fails with a `System.ArgumentNullException: Value cannot be null. (Parameter 'key')` exception when called 
https://github.com/SeleniumHQ/selenium/blob/e6f53e893b758157c8df2eda30ebea35a1c6fb05/dotnet/src/webdriver/Chromium/ChromiumDriver.cs#L173
beacause optionsCapabilityName is null

## To Reproduce

Create a EdgeDriver and call CreateDevToolsSession() 

## Expected behavior 

The driver create a DevToolsSession

pujagani:

### Description

Fixes #8358

### Motivation and Context

While creating a DevToolsSession, a capability check is done. Since the EdgeDriver was calling the base constructor of ChromiumDriver that did not set the capability name, it was defaulting to null. Thus, leading to the error described in #8358. The changes fix this.

### Types of changes

### Checklist


jimevans:
This is not the correct way to do this. We don't want to be passing _both_ options and capabilities to the base constructor. We need to rethink this approach.

jimevans:
@pujagani I'm going to go ahead and close this PR, because I don't think it's needed anymore. I think a refactor of the .NET bindings has fixed the underlying issue. If it's still an issue, we can revisit it.

PR #9425, opened 2021-04-27
Add LogEntry public constructor

### LINKED ISSUE #7643: [.NET] Add public constructor to LogEntry class
## 🚀 Feature Proposal

Currently the class only has a private constructor, no setters and an internal method to create a new instance. I'm trying to parse the LogEntry into another class, but writing unit tests is getting a bit tedious (reflection, additional wrappers and so on). 

Would be nice if we had a public facing constructor for this. Added suggested code below.

## Example

    '/// <summary>
    /// Represents an entry in a log from a driver instance.
    /// </summary>
    public class LogEntry
    {
        /// <summary>
        /// Initializes a new instance of the <see cref="LogEntry"/> class.
        /// </summary>
        private LogEntry()
        {
        }

        public LogEntry(DateTime timestamp, LogLevel level, string message)
        {
            Timestamp = timestamp;
            Level = level;
            Message = message ?? string.Empty;
        }

        /// <summary>
        /// Gets the timestamp value of the log entry.
        /// </summary>
        public DateTime Timestamp { get; private set; } = DateTime.MinValue;

        /// <summary>
        /// Gets the logging level of the log entry.
        /// </summary>
        public LogLevel Level { get; private set; } = LogLevel.All;

        /// <summary>
        /// Gets the message of the log entry.
        /// </summary>
        public string Message { get; private set; } = string.Empty;

        /// <summary>
        /// Returns a string that represents the current <see cref="LogEntry"/>.
        /// </summary>
        /// <returns>A string that represents the current <see cref="LogEntry"/>.</returns>
        public override string ToString()
        {
            return string.Format(CultureInfo.InvariantCulture, "[{0:yyyy-MM-ddTHH:mm:ssZ}] [{1}] {2}", Timestamp, Level, Message);
        }

        /// <summary>
        /// Creates a <see cref="LogEntry"/> from a dictionary as deserialized from JSON.
        /// </summary>
        /// <param name="entryDictionary">The <see cref="Dictionary{TKey, TValue}"/> from
        /// which to create the <see cref="LogEntry"/>.</param>
        /// <returns>A <see cref="LogEntry"/> with the values in the dictionary.</returns>
        internal static LogEntry FromDictionary(Dictionary<string, object> entryDictionary)
        {
            LogEntry entry = new LogEntry();
            if (entryDictionary.ContainsKey("message"))
            {
                entry.Message = entryDictionary["message"].ToString();
            }

            if (entryDictionary.ContainsKey("timestamp"))
            {
                DateTime zeroDate = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
                double timestampValue = Convert.ToDouble(entryDictionary["timestamp"], CultureInfo.InvariantCulture);
                entry.Timestamp = zeroDate.AddMilliseconds(timestampValue);
            }

            if (entryDictionary.ContainsKey("level"))
            {
                string levelValue = entryDictionary["level"].ToString();
                try
                {
                    entry.Level = (LogLevel)Enum.Parse(typeof(LogLevel), levelValue, true);
                }
                catch (ArgumentException)
                {
                    // If the requested log level string is not a valid log level,
                    // ignore it and use LogLevel.All.
                }
            }

            return entry;
        }
    }'

irina-voropaeva:
Fixes #7643
Add LogEntry public constructor

### Description
Add LogEntry public constructor with 3 parameters - timestamp of log entry with type DateTime, level of logging with type LogLevel and log entry message with type string.

### Motivation and Context
Problem is there is no public constructor for LogEntry

### Types of changes

### Checklist


jimevans:
I'm not sure I agree this is necessary. The `LogEntry` data structure is not intended to be instantiated directly from user code. I'm willing to change my mind, if a convincing use case can be made. In the meantime, I'm closing this PR.

PR #9428, opened 2021-04-28
Update doc for ignored_exceptions of WebDriverWait()

redstoneleo:
If we set 
`WebDriverWait(driver, timeout, poll_frequency=0.5, ignored_exceptions=[TimeoutException])`
It won't ignore the `TimeoutException` because[ the source code](https://github.com/SeleniumHQ/selenium/blob/64447d4b03f6986337d1ca8d8b6476653570bcc1/py/selenium/webdriver/support/wait.py#L87) shows  it still raise the exception

### DIFF
--- a/py/selenium/webdriver/support/wait.py
+++ b/py/selenium/webdriver/support/wait.py
@@ -32,7 +32,7 @@ class WebDriverWait(object):
             - timeout - Number of seconds before timing out
             - poll_frequency - sleep interval between calls
               By default, it is 0.5 second.
-            - ignored_exceptions - iterable structure of exception classes ignored during calls.
+            - ignored_exceptions - iterable structure of exception classes ignored during calls. N.B. `selenium.common.exceptions.TimeoutException` is not ignored for `until(self, method, message='')` and `until_not(self, method, message='')` 
               By default, it contains NoSuchElementException only.
 
            Example::


AutomatedTester:
Thank you for your contribution.

I won't be accepting this patch as I think that it would cause more confusion than helping. The `until*` methods are designed to throw a `TimeoutException` so it's doing what it should and since we tell people that it will throw that error it seems sufficient to me.

redstoneleo:
> Thank you for your contribution.
> 
> I won't be accepting this patch as I think that it would cause more confusion than helping. The `until*` methods are designed to throw a `TimeoutException` so it's doing what it should and since we tell people that it will throw that error it seems sufficient to me.

I think the mechanism is contrary to the description of `ignored_exceptions ` parameter of `WebDriverWait()` if one sets `ignored_exceptions=[TimeoutException]`
`ignored_exceptions - iterable structure of exception classes ignored during calls. By default, it contains NoSuchElementException only.`
In order to avoid misunderstandings, so I request to update the doc.

diemol:
Adding this comment to the docs is reduntant because in the [`until`](https://github.com/SeleniumHQ/selenium/blob/selenium-4.0.0-beta-3/py/selenium/webdriver/support/wait.py#L70) and [`until_not`](https://github.com/SeleniumHQ/selenium/blob/selenium-4.0.0-beta-3/py/selenium/webdriver/support/wait.py#L97) methods one can already see that a `TimeoutException` will be thrown if a timeout occurs. Therefore, the existing docs already prevent misunderstandings (when the docs are checked, of course ☺️).

Feel free to join us in our Slack/IRC chat if you'd like to discuss more https://www.selenium.dev/support/

Issue #9455, opened 2021-05-07
[dotnet] RequestTransformer can't transform requests


doctor-house:
## 🐛 Bug Report

The Selenium 4 beta C# bindings add `NetworkRequestHandler` to let users define network interception rules. In particular, one may define a `RequestTransformer` of type `Func<HttpRequestData, HttpRequestData>` to apply some modification to outgoing requests in the browser, but `HttpRequestData` is a read only type and can't be modified.

## To Reproduce

The following code generates a compile time error:

```c#
NetworkRequestHandler handler = new NetworkRequestHandler()
{
    RequestMatcher = (request) => 
    {
        return true;
    },
    RequestTransformer = (request) =>
    {
        request.Url = "https://www.selenium.dev/"
        return request;
    }
};
```
`request.Url` (as well as every other field on the request object) is marked `internal set` and cannot be modified.

## Expected behavior

A user should be able to modify the request data within the `RequestTransformer` method

## Environment

OS:  Windows 10
Browser:  N/A
Browser version:  N/A
Browser Driver version:  N/A
Language Bindings version:  C# 4.0-beta2
Selenium Grid version (if applicable):   N/A

jimevans:
PR merged. Thank you for your contribution.

PR #9456, opened 2021-05-07
make HttpRequestData properties settable

doctor-house:
### Description

Changed `internal set` to `set` for the fields in `HttpRequestData`. 

These fields need to me modifiable from outside the assembly in some way for request transformation to work properly. This is the most direct fix for #9455.

### Types of changes

### Checklist


jimevans:
I’d like to explore other avenues rather than making everything writable in that class. I really want the data classes referenced by the events to be read-only. When using the transformer, we can maybe create a class with read/write properties. When I’m back from being out of town, I can discuss it further.

jimevans (approved):
I'm less than thrilled with this approach, as mentioned in my earlier comment, but in the interest of making the functionality useful, we'll do it this way for now.

PR #9457, opened 2021-05-09
Add NuGet and SourceLink support in CSProj files

TroyWalshProf:
Add NuGet build to CSProjs, symbol package build, and SourceLink support

### Description
Add nuget build properties to CSProj files
Centralizes dll, nupkg, and snupkg build information

Add symbols package version and SourceLink support.
This can make debugging much easier.
#### Details
How to creating symbols package:
https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg
Adding and using SourceLink:
https://www.hanselman.com/blog/ExploringNETCoresSourceLinkSteppingIntoTheSourceCodeOfNuGetPackagesYouDontOwn.aspx

### Motivation and Context
SourceLink makes debugging problems much easier
Allows us to drop separate nuspec files 
Makes nupkg and snupkg builds part of the standard dll builds

### Types of changes

### Checklist


TroyWalshProf:
@jimevans We added this feature to Appium.Net (https://github.com/appium/appium-dotnet-driver/pull/415) and I know it really helped when debugging issues.

jimevans:
The snupkg files are already published to the NuGet symbol server. Since the shipping binaries are not produced by the build from within Visual Studio, and we do not rely on the .csproj file information to create the NuGet packages that are published, help me understand what benefit having this info in the .csproj would be.

TroyWalshProf:
Sorry @jimevans, I didn’t see your response.
### For you:

1. Simplifies the creation of NuGet packages as that can be done as part of the standard project build.
2. Not being able to create a package will be caught at compile time, not later on in a separate build process.
3. Makes created automated builds/releases relatively simple.
#### Real project example: 
**Automated build with pipeline provided visioning:**
https://github.com/TroyWalshProf/SeleniumAxeDotnet/blob/12d0489436667716f39e0d505bda665cee71b218/.github/workflows/mainPipeline.yml#L93

**Automated releases:**
https://github.com/TroyWalshProf/SeleniumAxeDotnet/blob/12d0489436667716f39e0d505bda665cee71b218/.github/workflows/mainPipeline.yml#L132
*Note* in this instance I am using environments approvers as a release gate

### For you and end users:
1. Allows you to publish package and symbol to Nuget.org together.  image
2. This means users can use source link ([Source Link and .NET libraries | Microsoft Docs](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/sourcelink))
    * Video in the link above does a better job than I can explain how source link works and why it is awesome. - Short version is that source link would allow me to step thru both my and Selenium’s code live in a debug session.  
    * Source link works in VS and VS Code

#### Real project example: 

-  Debugging my code and get asked if I want to step into the Appium.Net code: image  

- Stepped into Appium.Net code: image

jimevans:
The source link properties I might be able to get behind. The NuGet properties are not going to land. That would create multiple "sources of truth" for the NuGet packages (since `.nuspec` files already exist in the project tree). Since the Selenium project uses Bazel as its build system, creating and or publishing "releasable" artifacts directly from within Visual Studio is not a goal for the project.

TroyWalshProf:
@jimevans - I think we may be talking past each other a bit.

The .nuspec files would go away as they would be redundant. 
The .csproj files would become your single source of truth for both the binaries and nuget packages.
_I just didn't remove the nuspec because I am unfamiliar with your build system and assumed updating the build and removing the .nuspec files would happen in a subsequent PR._

If you would like I could include the Bazel updates and .nuspec removal as part of this PR. 

Side notes* 
The goal is not build or publish releasable artifacts directly for VS - I agree the is not a particularly smart of useful thing
The goal is to get source link. A nice side benefit is that instead of having a .csproj and .nuspec source of truth, you only have a .csproj source of truth.
Another nice side benefit is that you don't need to manually manage [dependencies ](https://github.com/SeleniumHQ/selenium/blob/8755948934a75a4d497bb35b143017fc6f477fe0/dotnet/src/webdriver/WebDriver.nuspec#L28) as that if is already part of the .csproj natively.

TroyWalshProf:
@jimevans - Where would I see the Dotnet build in action?
*I see most of the others when looking at actions (https://github.com/SeleniumHQ/selenium/actions).

TroyWalshProf:
Still researching but this PR will not work with Bazel - abandoning for now

PR #9482, opened 2021-05-17
[py] PEP 484 type hints for common.exceptions and webdriver.support.color

hoefling:
### Description
This is the initial PR that adds type hints to Python codebase. Typed modules:
* `selenium.common.exceptions`,
* `webdriver.support.color`.
I have selected those two for the initial typing since they do not depend on any other modules and can be typed in isolation.

### Motivation and Context
The goal of this PR is to configure and add initial integration for the `mypy` tool (checks the type hints validity). It can be run via
```sh
$ tox -e mypy
```
or
```sh
$ TOXENV=mypy tox
```
Right now it emits 1248 errors, mainly due to missing type hints; the goal is to gradually reduce it to zero. No errors are emitted in modules typed in this PR.

### Types of changes

### Checklist


hoefling:
@AutomatedTester gentle ping (no rush though!). Please advise:
 1. Is the PR too big for a review? I will split it if necessary and create smaller PRs in the future.
 2. I would like to add a new job that runs `mypy` checks, but it will fail until all the codebase is typed and unfortunately GH actions syntax doesn't allow to ignore failing jobs. I can filter out untyped modules in `mypy` config (so no errors are reported) and loosen the filter gradually with each typed module, like suggested in [mypy docs](https://mypy.readthedocs.io/en/stable/existing_code.html):
    ```ini
    # mypy config
    [mypy]
    ...
    [mypy-selenium.webdriver.*]
    ignore_errors = True

    [mypy-selenium.webdriver.support.event_firing_webdriver]
    ignore_errors = True

    # etc
    ```
    However, maybe you have a better idea for me?

py/selenium/common/exceptions.py
```diff
@@ -19,18 +19,20 @@
...
+from typing import Any
+
 
 class WebDriverException(Exception):
     """
     Base webdriver exception.
     """
 
-    def __init__(self, msg=None, screen=None, stacktrace=None):
+    def __init__(self, msg: Any = None, screen: Any = None, stacktrace: Any = None) -> None:
```
AutomatedTester: These are always going to be `str` or `None

py/selenium/common/exceptions.py
```diff
@@ -126,11 +128,11 @@ class UnexpectedAlertPresentException(WebDriverException):
     Usually raised when  an unexpected modal is blocking the webdriver from executing
     commands.
     """
-    def __init__(self, msg=None, screen=None, stacktrace=None, alert_text=None):
+    def __init__(self, msg: Any = None, screen: Any = None, stacktrace: Any = None, alert_text: Any = None) -> None:
```
AutomatedTester: The types here are `str` or `None`.

AutomatedTester (commented):
> @AutomatedTester gentle ping (no rush though!). Please advise:
> 
>     1. Is the PR too big for a review? I will split it if necessary and create smaller PRs in the future.

No, this is fine. Don't go much bigger than this though.

> 
>     2. I would like to add a new job that runs `mypy` checks, but it will fail until all the codebase is typed and unfortunately GH actions syntax doesn't allow to ignore failing jobs. I can filter out untyped modules in `mypy` config (so no errors are reported) and loosen the filter gradually with each typed module, like suggested in [mypy docs](https://mypy.readthedocs.io/en/stable/existing_code.html):
>        ```ini
>        # mypy config
>        [mypy]
>        ...
>        [mypy-selenium.webdriver.*]
>        ignore_errors = True
>        
>        [mypy-selenium.webdriver.support.event_firing_webdriver]
>        ignore_errors = True
>        
>        # etc
>        ```
>        
>        

I've not used mypy in anger so I might need your help here. Is there a way that we can do a pass over and have file that holds the current output so we can't regress and then fix the others? A little similar to how Rubocop works for Ruby. If your suggestion above does that then do it and we can then clean it up as we go. Let's try not boil the ocean.

hoefling:
> I've not used mypy in anger so I might need your help here. Is there a way that we can do a pass over and have file that holds the current output so we can't regress and then fix the others? A little similar to how Rubocop works for Ruby. If your suggestion above does that then do it and we can then clean it up as we go. Let's try not boil the ocean.

I see what you mean - unfortunately, my suggestion doesn't do that, it's merely "hide modules with errors until those are fixed, and don't forget to uncover them back once fixed". However, `mypy` can produce JUnit and Cobertura reports. Let me play with them - I think the report file can be cached on main branch and compared against the report in the PR. Maybe we can even integrate that in a service like codecov.io and get an automatic regression reports in PRs. However, I wouldn't bloat this PR up anymore and introduce the CI job in another PR.

py/selenium/common/exceptions.py
hoefling: @AutomatedTester I guess except `stacktrace`? Looking at this right now:

https://github.com/SeleniumHQ/selenium/blob/c13800859ee458ad56074ad329f59cff551395ac/py/selenium/webdriver/remote/errorhandler.py#L208-L236

AutomatedTester (approved):

AutomatedTester (approved):

PR #9495, opened 2021-05-20
[Javascript] Ensure correct serialization of extensions in chromium.Options

### LINKED ISSUE #6676: [JavaScript] Loading extensions from file in Chrome doesn't work in 4.0.0-alpha.1
## 💥 Regression Report

Loading a Chrome extension in 4.0.0-alpha.1 doesn't work, it throws ```unknown error: cannot process extension #1 from unknown error: cannot base64 decode```

## Last working Selenium version

Worked up to version: 3.6.0

Stopped working in version: 4.0.0-alpha.1

## To Reproduce
Loading an extension like:

```
 chromeOptions.addExtensions(
      path.resolve(moduleRootPath, 'vendor', 'browsertime-extension.crx')
   );
```
doesn't work.  See https://github.com/sitespeedio/browsertime/blob/master/lib/chrome/webdriver/index.js#L100-L105

Detailed steps to reproduce the behavior:

## Expected behavior

You should be able to load extensions from file in Chrome.

## Test script reproducing this issue (when applicable)

I think this was introduced here: https://github.com/SeleniumHQ/selenium/commit/9976795d0357aa19f95d6c73c9f203449e182fbf#diff-cc4a7b2914982158b85fcd2ff2a60ea2

I can have a go to see if I can fix it the coming days, if I get the environment working.

## Environment

OS: OS X
Browser: Chrome
Browser version:  70.0.3538.10
Browser Driver version:  ChromeDriver 2.43
Language Bindings version: JavaScript 4.0.0-alpha.1
Selenium Grid version (if applicable):

nickgaya:
Fixes #6676

### Description
Wraps the internal `options_.extensions` list in an `Extensions` object with serialization logic to resolve each extension given to a base64-encoded string.

### Motivation and Context
According to the documentation, when calling `chromium.Options.addExtensions()` each extension should be specified as a path to a packed CRX file, or a Buffer containing the extension data.

The `chromium.Options` object defines a `[Symbols.serialize]()` method that resolves each extension to a base64-encoded string.  However, as of the v4 changes, the Chromium options get merged into a generic `Capabilities` object so the extra serialization logic is never invoked.

This PR fixes the issue by wrapping the list of extensions in an object that will be preserved across capabilities merges.

### Types of changes

### Checklist


AutomatedTester:
@nickgaya Could you please rebase this PR?

nickgaya:
@AutomatedTester Looks like that last merge caused some test failures, see #10012.

PR #9510, opened 2021-05-25
[grid] Ensure Distributor rejects request immediately if no matching slot exists in the Grid

pujagani:

### Description

Ensure Distributor rejects request immediately if no matching slot exists in the Grid

### Motivation and Context

When a new session request comes in, the Distributor polls the queue based on the available matching capabilities of the Nodes in the Grid. Hence, if a new session request requests for a capability that does not exist in the Grid, it sits in the queue until it times out instead of been rejected immediately. The changes fix that and provide the client correct error message about the request.

### Types of changes

### Checklist


csgbn:
This change will make the grid queue very different from v3, and will cause a lot of waste of resources in a cloud environment were you launch nodes with requested capabilities upon demand instead of having nodes always running even when not in use, just so that rare capability requests can be processed. By all means put a timeout on a request but why immediately reject the request?

pujagani:
Thank you for bringing this to my notice. The changes are aligned with the design discussion that happened when the queue was introduced in Selenium 4. For a Grid setup that does not spin up the Nodes on demand, it is better to provide the accurate error message and immediate response that capabilities do not exist on the Node instead of letting it timeout. That way the client can stop additional requests for the same capabilities to avoid overloading the Grid.

You have a valid point. For cases that use a similar mechanism like [OneShotNode](https://github.com/SeleniumHQ/selenium/blob/trunk/java/server/src/org/openqa/selenium/grid/node/k8s/OneShotNode.java), it is not ideal. Apologies for missing this scenario.
Different scenarios will have different requirements based on how the Nodes are created. To ensure feature compatibility with v3 and to support cloud environment accurately, will make this feature optional with a flag. By default, the request will not be rejected immediately.

PR #9523, opened 2021-05-27
[py] Run type checks in CI and report typing coverage to codecov

hoefling:
### Description

This PR adds another Github action that runs type checks (using the `mypy` toxenv) and reports the current status to https://codecov.io. An example report, generated from this branch in my fork, can be viewed [here](https://codecov.io/gh/hoefling/selenium/tree/95450a15332092e285819de8e49f789907d5c9a9/): in the typing context, `100%` coverage means all types could be successfully resolved and verified, while `<100%` indicates missing, incomplete or erroneous type hints. Navigating to a module unveils the spots in question.

**Example**: In the above example, the [`selenium.common` package](https://codecov.io/gh/hoefling/selenium/tree/95450a15332092e285819de8e49f789907d5c9a9/py/selenium/common) is fully typed (green color & 100% coverage), while [`selenium.webdriver.chrome.options` module](https://codecov.io/gh/hoefling/selenium/src/95450a15332092e285819de8e49f789907d5c9a9/py/selenium/webdriver/chrome/options.py) is not - more details can be found in `tox_mypy` job log, where `mypy` will report an error for the line. In this example:
```
selenium/webdriver/chrome/options.py:25: error: Implicit generic "Any". Use "typing.Dict" and specify generic parameters
selenium/webdriver/chrome/options.py:28: error: Function is missing a type annotation
```
([relevant spot in the job log for reference](https://github.com/hoefling/selenium/runs/2688338273?check_suite_focus=true#step:5:495))

This way, one should be able to easily track the typing errors.

### Motivation and Context

This is a proposal to address the discussion in #9482. It is added for @AutomatedTester for an evaluation!

Also, once merged into master, `codecov` should start tracking the changes in upcoming PRs, comparing typing coverage against the current master ([an example on how it will look like](https://github.com/mehcode/python-xmlsec/pull/173#issuecomment-848682998), although this comment actually reports code coverage difference).

Once all typing errors are resolved (thus a 100% typing coverage achieved for the complete Python codebase), we can stop reporting the typing coverage and just run `tox -c py/tox.ini` in `tox_mypy` job. It will then simply fail on new typing errors.

### Types of changes

### Checklist


hoefling:
Here's the [initial report for the upstream repo](https://codecov.io/gh/SeleniumHQ/selenium/tree/ee5424f48ee752cd1a2d2174ea1cb2588f81ebd4). Of course, there's nothing on master to compare against yet, so the codecov-bot will leave no comments as of now.

diemol (commented):
Is there a way to this by using SonarQube? We already use it for the Java bindings and probably keeping an eye on different places is not ideal.

hoefling:
@diemol sure, but AFAIK SonarQube doesn't support `mypy` reports, only code coverage reports. Do you have plans to track test coverage with SonarQube?

Issue #9480, opened 2021-05-15
Add type hints in python codebase


hoefling:
## 🚀 Feature Proposal

Type hints in Python are specified by [PEP 484](https://www.python.org/dev/peps/pep-0484/) and are of a great help when developing, thus proposing to add them to Python codebase. Would this be an option if I volunteer to (gradually) add them, assuming the types are validated in the CI by a checker tool like `mypy` so the types are validated on each pull request?

I have noticed rudimentary type hints already exist here and there, for example

https://github.com/SeleniumHQ/selenium/blob/e9c738de8a587dc6bc86b55949c57b8574ec68bb/py/selenium/webdriver/remote/webdriver.py#L263

indicates that `WebDriver().name` is a string. This offers better code completions in IDEs.

I have searched for similar issues, but it seems like this wasn't discussed yet (at least not on Github). As far as I can see, the only issue mentioning type hints is #1917; however, type stubs were removed since then.

Another alternative would be maintaining a third-party stub package (as specified in [PEP 561](https://www.python.org/dev/peps/pep-0561/#stub-only-packages)). However, I don't want to start a third-party package if you are willing to keep the type hints along with the code (which is a much better option), or planning to add them already.

## Motivation

Better code completion in IDEs is surely something the end users will appreciate. Another advantage is that type hints often  intercept bugs, thus making the code more robust.

AutomatedTester:
I am all for this! Feel free to start adding PRs and then tagging me for review.

iampopovich:
I'm in

Akuli:
When are you going to add a `py.typed` file to tell type checkers to rely on what's in selenium itself, instead of using typeshed's selenium 3.x stubs?

For context, I am a maintainer of typeshed, and I'm wondering if we should upgrade our selenium 3.x stubs to selenium 4.x, or just wait until someone adds a `py.typed` file to selenium 4.x.

AutomatedTester:
I haven't got round to adding a `py.typed`. If someone fancies putting a PR up I will review and get it landed as quickly as I can.

GergelyKalmar:
I don't think this worked, the py.typed file is still missing from the latest release (4.1.1). Could we get a patch release which fixes that? It would be awesome.

Akuli:
[...]

PR #9524, opened 2021-05-27
[py] PEP 484 type hints for selenium.webdriver.common.timeouts

hoefling:
### Description

This PR adds type hints to `selenium.webdriver.common.timeouts`.

### Motivation and Context

This is a follow-up of #9482 to address #9480.

### Types of changes

### Checklist


py/selenium/webdriver/common/timeouts.py
```diff
@@ -35,55 +55,55 @@ def __init__(self, implicit_wait=0, page_load=0, script=0):
         self._script = self._convert(script)
 
     @property
-    def implicit_wait(self):
+    def implicit_wait(self) -> float:
```
hoefling, line 58: @AutomatedTester unsure whether `float`s are meant to be here. To me, it looks like the timeouts should be all integers (or the explicit conversion in `Timeouts._convert` is useless), but in Python 3, the division makes them to floats. Maybe the integer division `//` should be used here instead and was forgotten when dropping Python 2 support? I may be wrong, though, thus typed based on the current implementation.

hoefling: @AutomatedTester actually, disregard that, I see now that it's just the rounding.

AutomatedTester (approved):

PR #9541, opened 2021-06-02
[rb] fix parsing differences between options and capabilities

titusfortner:
For testing


titusfortner:
gah, github didn't recognize my base branch change when I updated comment. This was supposed to PR against my fork so tests can run. Disregard.

PR #9561, opened 2021-06-10
Add the debuggerAddress option to chromium.js

bwalderman:
### Description
Adding a setter for Chromium's debuggerAddress capability. This is available in other language bindings but was missing from the JS bindings.

### Motivation and Context
Chromium-based drivers provide the `debuggerAddress` option to let the user attach to an already-running browser instance at the specified address instead of launching a new one. One example where this is useful is certain WebView2 (WV2) scenarios. Users may want to test the web portion of an app that embeds WV2, but need to launch the app outside WebDriver first and drill through some native UI in order to activate the WV2 they are trying to test.

### Types of changes

### Checklist

- [ X] I have updated the documentation accordingly.

### DIFF
--- a/javascript/node/selenium-webdriver/chromium.js
+++ b/javascript/node/selenium-webdriver/chromium.js
@@ -276,6 +276,19 @@ class Options extends Capabilities {
   }
 
   /**
+   * Sets the address of a Chromium remote debugging server to connect to.
+   * Address should be of the form "{hostname|IP address}:port"
+   * (e.g. "localhost:9222").
+   *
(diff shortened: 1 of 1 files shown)


AutomatedTester:
Thanks for the PR. Could you give a bit more context about why this is needed? Currently, we get the debugger address from the capabilities returned and allow the driver to figure out where to put it.

bwalderman:
Thanks, I've updated the motivation in the PR description.

AutomatedTester (approved):

PR #9637, opened 2021-07-05
[js] feat: add client_options for keep-alive and custom user-agent

Dharin-shah:

It adds a feature to allow adding custom user-agent in the http client and also allow the agent to enable keep-alive on the clients as well for socket reuse on the client-end

### Motivation and Context

As we have started using selenium-webdriver in nightwatch v2, we have hit a speedbump with default http agent provided by selenium-webdriver since it doesnt allow us to set custom user-agent in the header, as well as the client sockets arnt reused because the default agent has keep-alive as false by default
https://github.com/nodejs/node/blob/e46c680bf2b211bbd52cf959ca17ee98c7f657f5/lib/_http_agent.js#L107
See https://stackoverflow.com/a/58332910 for more explanation

### Types of changes

### Checklist

### DIFF
--- a/javascript/node/selenium-webdriver/firefox.js
+++ b/javascript/node/selenium-webdriver/firefox.js
@@ -222,7 +222,7 @@ async function buildProfile(template, extensions) {
     dir = await io.tmpDir()
     if (template) {
       await io.copyDir(
-        /** @type {string} */(template),
+        /** @type {string} */ (template),
         dir,
         /(parent\.lock|lock|\.parentlock)/
       )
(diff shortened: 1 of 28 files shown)


javascript/node/selenium-webdriver/http/index.js
```diff
@@ -88,12 +89,33 @@ class HttpClient {
...
+    /**
+     * client options, header overrides
+     */
+    this.client_options = client_options
+
+    /**
+     * sets keep-alive for the agent
+     * see https://stackoverflow.com/a/58332910
+     */
+    this.keepAlive = this.client_options['keep-alive']
```
AutomatedTester, line 101: We can probably set this to be `true` as the default which will help all users.

javascript/node/selenium-webdriver/firefox.js
```diff
@@ -592,7 +592,7 @@ class Driver extends webdriver.WebDriver {
    * implementation.
    * @override
    */
-  setFileDetector() { }
```
AutomatedTester, line 595: Can we have the linting changes in a separate commit within this PR so that it is easier to spot the functional changes

AutomatedTester (changes requested):
see comments

Dharin-shah:
Separate PRs [#1](https://github.com/SeleniumHQ/selenium/pull/9638) & [#2](https://github.com/SeleniumHQ/selenium/pull/9639)

PR #9645, opened 2021-07-10
[py] remove non-w3c compliant parameters from default capabilities

titusfortner:
I'm not sure this is the right fix for this. I'm trying to generate compliant capabilities with `ChromeOptions().to_capabilities()`, but it includes `version` and `platform` kyes. The right answer might be not to call this in the first place during the init? `self._caps = self.default_capabilities`

The Python tests appear to all be broken from [this commit](2ea9026477a051e62cc486eb18a61db3fc228934), so I'm not sure if there are tests that also need to be updated for this.


AutomatedTester:
tests need updating and then we can merge

PR #9665, opened 2021-07-19
Workflow path filters to reduce overhead

toddtarsi:
The goal is to reduce unnecessary parts being tested, so we're filtering down to relevant changes by file.

### Description

Add path filters for ruby and java to the workflow shape.

### Motivation and Context

Less unnecessary tests means people doing stuff faster.

### Types of changes

### Checklist

### DIFF
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -10,7 +10,7 @@ on:
     paths:
       # These files are common to all languages
       - .bazel*
-      - .github/actions
+      - .github/actions/**
       - common/**
       - javascript/atoms/**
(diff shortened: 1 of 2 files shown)


diemol (changes requested):
Hey @toddtarsi, thanks for this PR.

I wanted to take the same approach but we discussed in the TLC that this type of solution should be bazel based so we rely on our build tool instead of GitHub actions.

The idea is the to use https://github.com/Tinder/bazel-diff

toddtarsi:
@diemol - Absolutely, that sounds good. Closing this.

PR #9688, opened 2021-07-26
Add docker max sessions flag

### LINKED ISSUE #9660: There is no way to provide max-sessions parameter in the docker section of the standalone config file
## 🐛 Bug Report
There is no way to provide max-sessions parameter in the docker section of the standalone config file

## To Reproduce

My grid config looks as follows. Is there a way to limit the number of sessions available for both Firefox and Chrome in the docker section of standalone config file
```[toml]
[node]
detect-drivers = false
max-sessions=4

[docker]
configs = ["selenium/standalone-chrome:latest", "{\"browserName\": \"chrome\", \"browserVersion\": \"91\"}", "selenium/standalone-firefox:latest", "{\"browserName\": \"firefox\", \"browserVersion\": \"89\"}"]
host = "localhost"
port = 2375
url = "unix:/var/run/docker.sock"
video-image = "selenium/video:latest"
```

Running this command `java -jar selenium-server-4.0.0-beta-4.jar standalone --config grid.config.toml` creates 12 sessions each of Chrome and Firefox with a max concurrency of 4

image

In the file https://github.com/SeleniumHQ/selenium/blob/941dc9c6b2e2aa4f701c1b72be8de03d4b7e996a/java/server/src/org/openqa/selenium/grid/docker/DockerOptions.java#L137

 we have initialised it to the # of availableProcessors

whereas in the file https://github.com/SeleniumHQ/selenium/blob/941dc9c6b2e2aa4f701c1b72be8de03d4b7e996a/java/server/src/org/openqa/selenium/grid/node/config/NodeOptions.java#L79

we take into consideration the value available in the node section of the config file

## Expected behavior

There should be a config to set max-sessions in the docker section

## Environment

Selenium Grid version (if applicable): selenium-server-4.0.0-beta-4

pujagani:

### Description

Fixes #9660 and https://github.com/SeleniumHQ/docker-selenium/issues/1340.

### Motivation and Context

A max-sessions flag specific to the Docker section of the config will allow the users to set the value for maximum concurrent sessions allowed. This value is upper-bound by the maximum available processors.

### Types of changes

### Checklist

### DIFF
--- a/java/server/src/org/openqa/selenium/grid/docker/DockerOptions.java
+++ b/java/server/src/org/openqa/selenium/grid/docker/DockerOptions.java
@@ -57,6 +57,7 @@ public class DockerOptions {
   static final String DEFAULT_ASSETS_PATH = "/opt/selenium/assets";
   static final String DEFAULT_DOCKER_URL = "unix:/var/run/docker.sock";
   static final String DEFAULT_VIDEO_IMAGE = "selenium/video:latest";
+  static final int DEFAULT_MAX_SESSIONS = Runtime.getRuntime().availableProcessors();
   private static final String DEFAULT_DOCKER_NETWORK = "bridge";
   private static final Logger LOG = Logger.getLogger(DockerOptions.class.getName());
   private static final Json JSON = new Json();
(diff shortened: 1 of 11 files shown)


java/server/src/org/openqa/selenium/grid/docker/DockerOptions.java
```diff
@@ -152,7 +153,8 @@ private boolean isEnabled(Docker docker) {
     Image videoImage = getVideoImage(docker);
     loadImages(docker, videoImage.getName());
 
-    int maxContainerCount = Runtime.getRuntime().availableProcessors();
+    int maxContainerCount = Math.min(config.getInt(DOCKER_SECTION, "docker-max-sessions")
```
diemol: This looks good, but I would prefer to use the same [`--max-sessions`](https://github.com/SeleniumHQ/selenium/blob/9717d403a8fedb6da86c07fac5ccc8f06fb322b3/java/server/src/org/openqa/selenium/grid/node/config/NodeFlags.java#L50) from `NodeFlags`. Like this, the user does not need to think more when configuring this.

diemol (changes requested):
Thanks for tackling this, @pujagani!
Left a comment.

java/server/src/org/openqa/selenium/grid/docker/DockerOptions.java
pujagani: Thank you! I agree, that makes sense. I will update it.

diemol (approved):

PR #9710, opened 2021-08-03
Move java sources and tests into a single tree

shs96c:
This emphasizes that all the java code is inter-related. We make use
of bazel to avoid introducing Java 8 incompatible changes, and bazel's
visibility rules to ensure that generated jars only contain what they
should.

### Description
Move all the java code into a single place

### Motivation and Context
Because it's nice to have everything in one place.

### Types of changes

### Checklist


diemol (changes requested):
`small_tests` and `javadoc` jobs are failing

Issue #9699, opened 2021-07-30
Update IE Driver Capabilities in Bindings


Michail864:
Hello,

Is there a way to add these two capabilities to IE Driver in Ruby? I only see the examples in VB.NET and C#:

var ieService = InternetExplorerDriverService.CreateDefaultService(dir, driver); 
var ieOptions = new InternetExplorerOptions{}; 
ieOptions.AddAdditionalCapability("ie.edgechromium", true); 
ieOptions.AddAdditionalCapability("ie.edgepath", @"\\msedge.exe"); 

I need to automate IE Mode for Microsoft Edge browser and my project is in Ruby. However, there is not such method as "AddAdditionalCapability" in  Selenium::Webdriver::Remote::Capabilities in Ruby. Please let me know if there is a siluion for this issue.

Thanks.

titusfortner:
None of the bindings, (including .NET) are using the updated list of supported IE driver settings. We all need to fix this before RC release.

titusfortner:
@Michail864 to answer your specific question...
In each of the languages, users should have access to all of the options in the Browser Options classes, and right now we can't because the options aren't updated to everything IE supports. We'll fix that.

In the meantime, in Ruby, you can use the `Capabilities` class and pass in a Hash with exactly the values you want:
```ruby
Selenium::WebDriver::Remote::Capabilities.ie('ie.edgechromium': true, 'ie.edgepath': path_to_edge)
```

Michail864:
Hello Titus,

Thank you for your response, I tried to run the code you provided and this
is what I got :

Invalid capabilities in alwaysMatch: unknown capability named
ie.edgechromium

So, I believe Ruby does not support this capability?

Thanks,
Michail

пятница, 30 июля 2021 г. пользователь Titus ***@***.***>
написал:

titusfortner:
Oh right, it doesn't automatically put inside ieOptions. Do:
```
Selenium::WebDriver::Remote::Capabilities.ie('se:ieOptions' => {'ie.edgechromium' => true, 'ie.edgepath' => path_to_edge})
```

Michail864:
Hello Titus,

Thank you for your response again. I am trying to pass this code to Watir
but it doesnt recognize 'ie.edgechromium'.  I get the error message :

no implicit conversion of Selenium::WebDriver::Remote::Capabilities into
Hash (TypeError)

My code :

[...]

PR #9715, opened 2021-08-03
Update IE Options Classes to match currently supported IE Driver Capabilities

titusfortner:
This addresses #9699 

I think I did this correctly for each language, please verify.

(Oh, and I wanted to double check I'm not screwing up backward compatibility with a JWP capability I deleted in Java)

Thanks!


AutomatedTester:
Python looks good

PR #9738, opened 2021-08-13
Dispose HttpRequestMessage and HttpResponseMessage objects

nvborisenko:
Disposing all Http objects immediately


jimevans:
Merged in e867b31. Thank you for your contribution!

PR #9744, opened 2021-08-17
Improve Travis CI build Performance

chenzhang22:
According to [Build times out because no output was received](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received), we should carefully use travis_wait, as it may make the build unstable and extend the build time.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.


diemol:
@YunLemon thanks for this PR, however, we are not using Travis anymore. I will delete the file from the repo to avoid confusion.

PR #9828, opened 2021-09-14
Add methods to options classes for w3c compliant capabilities

titusfortner:
### Description

Adds methods and tests for:
* `setBrowserVersion()`
* `setPlatformName()`
* ~`setTimeouts()`~
edit:
* `setImplicitWaitTimeout()`
* `setPageLoadTimeout()`
* `setScriptTimeout()`

Timeouts has user-facing `Duration` that gets converted to millis when adding it to capabilities. 

### Motivation and Context

It makes sense to have specific methods for known capabilities, and this is backwards compatible.

### DIFF
--- a/java/src/org/openqa/selenium/remote/AbstractDriverOptions.java
+++ b/java/src/org/openqa/selenium/remote/AbstractDriverOptions.java
@@ -17,18 +17,6 @@
 
 package org.openqa.selenium.remote;
 
-import static org.openqa.selenium.remote.CapabilityType.ACCEPT_INSECURE_CERTS;
-import static org.openqa.selenium.remote.CapabilityType.BROWSER_VERSION;
-import static org.openqa.selenium.remote.CapabilityType.IMPLICIT_TIMEOUT;
-import static org.openqa.selenium.remote.CapabilityType.PAGE_LOAD_STRATEGY;
-import static org.openqa.selenium.remote.CapabilityType.PAGE_LOAD_TIMEOUT;
-import static org.openqa.selenium.remote.CapabilityType.PLATFORM_NAME;
-import static org.openqa.selenium.remote.CapabilityType.PROXY;
-import static org.openqa.selenium.remote.CapabilityType.SCRIPT_TIMEOUT;
-import static org.openqa.selenium.remote.CapabilityType.STRICT_FILE_INTERACTABILITY;
-import static org.openqa.selenium.remote.CapabilityType.TIMEOUTS;
-import static org.openqa.selenium.remote.CapabilityType.UNHANDLED_PROMPT_BEHAVIOUR;
-
 import org.openqa.selenium.MutableCapabilities;
 import org.openqa.selenium.PageLoadStrategy;
 import org.openqa.selenium.Proxy;
@@ -36,17 +24,25 @@ import org.openqa.selenium.UnexpectedAlertBehaviour;
 import org.openqa.selenium.internal.Require;
 
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
 
+import static org.openqa.selenium.remote.CapabilityType.ACCEPT_INSECURE_CERTS;
+import static org.openqa.selenium.remote.CapabilityType.BROWSER_VERSION;
+import static org.openqa.selenium.remote.CapabilityType.PAGE_LOAD_STRATEGY;
+import static org.openqa.selenium.remote.CapabilityType.PLATFORM_NAME;
+import static org.openqa.selenium.remote.CapabilityType.PROXY;
+import static org.openqa.selenium.remote.CapabilityType.STRICT_FILE_INTERACTABILITY;
+import static org.openqa.selenium.remote.CapabilityType.TIMEOUTS;
+import static org.openqa.selenium.remote.CapabilityType.UNHANDLED_PROMPT_BEHAVIOUR;
+
 public abstract class AbstractDriverOptions<DO extends AbstractDriverOptions> extends MutableCapabilities {
+  private Map<String, Long> timeouts = new HashMap<>();
+
   public DO setBrowserVersion(String browserVersion) {
     setCapability(
       BROWSER_VERSION,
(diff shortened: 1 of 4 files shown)


java/src/org/openqa/selenium/remote/AbstractDriverOptions.java
```diff
@@ -30,13 +35,47 @@
...
+  }
+
+  public DO setPlatformName(String platformName) {
+    setCapability(
+      PLATFORM_NAME,
+      Require.nonNull("Platform Name", platformName));
+    return (DO) this;
+  }
+
+  public DO setTimeouts(Map<String, Duration> timeouts) {
```
shs96c: Prefer a `setImplicitTimeout`, `setPageLoadTimeout` and `setScriptTimeout`, and then we don't need to do any validation, and it's really hard to get wrong.

java/src/org/openqa/selenium/remote/AbstractDriverOptions.java
```diff
@@ -30,13 +35,47 @@
...
+
+    Require.nonNull("Timeouts", timeouts).forEach((k, v) -> {
+      if (validTimeouts.contains(k)) {
+        convertedTimeouts.put(k, v.toMillis());
+      } else {
+        throw new IllegalArgumentException(k + " is not one of the valid timeout values: " + validTimeouts);
+      }
+    });
+
+    setCapability(TIMEOUTS, convertedTimeouts);
```
shs96c: I think this has the potential to get confusing --- somewhere in a framework the implicit timeout is set, and then user code sets the page load timeout: what do we expect to happen to the implicit timeout? I _think_ people will expect both to be set.

java/src/org/openqa/selenium/remote/CapabilityType.java
```diff
@@ -47,6 +47,10 @@
   String HAS_TOUCHSCREEN = "hasTouchScreen";
   String OVERLAPPING_CHECK_DISABLED = "overlappingCheckDisabled";
   String STRICT_FILE_INTERACTABILITY = "strictFileInteractability";
+  String TIMEOUTS = "timeouts";
+  String IMPLICIT_TIMEOUT = "implicit";
```
shs96c: These are really meant to be the keys in the capabilities, not the keys of values within the capabilities

java/src/org/openqa/selenium/remote/AbstractDriverOptions.java
titusfortner: splitting up the timeouts as suggested above fixes this issue nicely.

java/src/org/openqa/selenium/remote/AbstractDriverOptions.java
titusfortner: Much better implementation.

java/test/org/openqa/selenium/chrome/ChromeOptionsTest.java
```diff
@@ -67,6 +70,53 @@ public void canBuildLogLevelFromStringRepresentation() {
...
+    Map<String, Object> mappedOptions = chromeOptions.asMap();
+    Map<String, Long> expectedTimeouts = new HashMap<>();
+
+    expectedTimeouts.put("implicit", 1000L);
+    assertThat(expectedTimeouts).isEqualTo(mappedOptions.get("timeouts"));
+
+    chromeOptions.setPageLoadTimeout(Duration.ofSeconds(2));
+    expectedTimeouts.put("pageLoad", 2000L);
+    Map<String, Object> mappedOptions2 = chromeOptions.asMap();
+    assertThat(expectedTimeouts).isEqualTo(mappedOptions2.get("timeouts"));
```
shs96c, line 118: Perhaps:

```java
ChromeOptions chromeOptions = new ChromeOptions();
chromeOptions.setImplicitWaitTimeout(Duration.ofSeconds(1));

assertThat(chromeOptions.getCapability(TIMEOUTS)).isEqualTo(Map.of("implicit", 1000L);

chromeOptions.setPageLoadTimeout(Duration.ofSeconds(2));
assertThat(chromeOptions.getCapability(TIMEOUTS)).isEqualTo(Map.of("implicit", 1000L, "page", 2000L);
```

shs96c: There's no need to convert the options to an intermediate map.

java/src/org/openqa/selenium/remote/AbstractDriverOptions.java
```diff
@@ -17,26 +17,63 @@
...
+import static org.openqa.selenium.remote.CapabilityType.BROWSER_VERSION;
+import static org.openqa.selenium.remote.CapabilityType.PAGE_LOAD_STRATEGY;
+import static org.openqa.selenium.remote.CapabilityType.PLATFORM_NAME;
+import static org.openqa.selenium.remote.CapabilityType.PROXY;
+import static org.openqa.selenium.remote.CapabilityType.STRICT_FILE_INTERACTABILITY;
+import static org.openqa.selenium.remote.CapabilityType.TIMEOUTS;
+import static org.openqa.selenium.remote.CapabilityType.UNHANDLED_PROMPT_BEHAVIOUR;
+
 public abstract class AbstractDriverOptions<DO extends AbstractDriverOptions> extends MutableCapabilities {
+  private Map<String, Long> timeouts = new HashMap<>();
```
shs96c: I'd be tempted to avoid this variable. Consider a user doing:

```java
options.setCapability(TIMEOUTS, Map.of("implicit", 1000));
options.setPageLoadTimeout(Duration.ofSeconds(10));
```

I _think_ a user would expect both timeouts to have been set, but the current implementation would only set one. I also think the opposite case (where the timeouts capability is set after calling the specific method) would mean that the user would expect the timeout map they passed in to be used.

One drawback to doing this is that you can't expect the `Map` the user passes in to be mutable. You'll need a check like:

```java
Map<String, Long> newTimeouts = new HashMap<>();
Object raw = getCapability(TIMEOUTS);
if (!(raw instanceof Map)) {
  ((Map<?, ?>) raw).foreach((key, value) -> {
    if (key instanceof String && value instanceof Long) {
      newTimeouts.put((String) key, (Long) value);
    }
  }
  newTimeouts.put("implicit", timeout.toMillis());
  setCapability(TIMEOUTS, Collections.immutableMap(newTimeouts));
))
```

Finally, it's good practice to make the collection "unmodifiable" when we set it, to prevent users from changing state in unexpected ways.

java/test/org/openqa/selenium/chrome/ChromeOptionsTest.java
titusfortner: agreed, doesn't make sense if needing to look at existing capabilities

java/src/org/openqa/selenium/remote/AbstractDriverOptions.java
titusfortner: Ok, I put this in a private `getCapabilities()` method. I'm not sure the reason for `instanceof Map` conditional. What alternative object is valid that will start a session?

java/src/org/openqa/selenium/remote/AbstractDriverOptions.java
```diff
@@ -94,4 +135,15 @@ public Object getCapability(String capabilityName) {
     getExtraCapabilityNames().forEach(name -> toReturn.put(name, getCapability(name)));
     return Collections.unmodifiableMap(toReturn);
   }
+
+  private Map<String, Number> getTimeouts() {
+    Map<String, Number> newTimeouts = new HashMap<>();
+    Object raw = getCapability(TIMEOUTS);
```
shs96c, line 141: You need to check that `raw` is actually a `Map` before casting to it. If not, if the user has set some garbage value (eg: `setCapability(TIMEOUTS, "cheese")` chaos will follow.

PR #9836, opened 2021-09-17
allow Augmenter implementations to specify additional commands

titusfortner:
The problem with existing implementation is that while you can cast to the interface, you don't have access to the commands that are needed to actually execute the associated methods.

This code creates an additional method `getAdditionalCommands()` on the `AugmenterProvider` interface. There is a default of an empty Map, so implementing it is optional.

`Augmentation` class has this method added to the constructor, which meant updating a bunch of things in the class. This is probably not backwards compatible, I'm not sure the right way to address that.

Essentially I need Augmenter class to provide a public method `getMatchingAdditionalCommands()`.

`RemoteWebDriverBuilder` can now call this method after augmenting the driver, and set `additionalCommands` field.

The tricky/messy bit is that I hacked into the middle of the `baseExecutor` lambda to `defineCommand` on each of the commands in the `additionalCommands` field.

There has to be a better way to do this, but it works, ~and my only real concern is backwards compatibility for the Augmentation class constructor~. (edit: fixed this!)

### DIFF
--- a/java/src/org/openqa/selenium/firefox/AddHasExtensions.java
+++ b/java/src/org/openqa/selenium/firefox/AddHasExtensions.java
@@ -21,12 +21,15 @@ import com.google.auto.service.AutoService;
 import com.google.common.collect.ImmutableMap;
 import org.openqa.selenium.Capabilities;
 import org.openqa.selenium.remote.AugmenterProvider;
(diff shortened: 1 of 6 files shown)


shs96c:
Closing this in favour of #9854

PR #9850, opened 2021-09-20
[rb]: Standardise driver logging output

luke-hill:
Ensure all log formats (Including browser ones), are in a consistent format

### Description
Alter logging to conform to other standard log formats

### Types of changes

### Checklist

### DIFF
--- a/rb/lib/selenium/webdriver/common/log_entry.rb
+++ b/rb/lib/selenium/webdriver/common/log_entry.rb
@@ -22,9 +22,9 @@ module Selenium
     class LogEntry
       attr_reader :level, :timestamp, :message
 
-      def initialize(timestamp, level, message)
-        @timestamp = timestamp
+      def initialize(level, timestamp, message)
         @level = level
+        @timestamp = timestamp
         @message = message
       end
 
--- a/rb/lib/selenium/webdriver/common/log_entry.rb
+++ b/rb/lib/selenium/webdriver/common/log_entry.rb
@@ -22,22 +22,22 @@ module Selenium
     class LogEntry
       attr_reader :level, :timestamp, :message
 
-      def initialize(level, timestamp, message)
-        @level = level
+      def initialize(timestamp, level, message)
         @timestamp = timestamp
+        @level = level
         @message = message
       end
 
       def as_json(*)
         {
-          'level' => level,
           'timestamp' => timestamp,
+          'level' => level,
           'message' => message
         }
       end
 
       def to_s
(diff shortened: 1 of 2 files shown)


luke-hill:
Failure seems unrelated

```
 1) Selenium::WebDriver::DevTools network interception changes requests
     Failure/Error: expect(driver.find_elements(tag_name: 'img').map(&:size).uniq).to eq([Dimension.new(640, 480)])
     
       expected: [#<struct Selenium::WebDriver::Dimension width=640, height=480>]
            got: [#<struct Selenium::WebDriver::Dimension width=0, height=0>, #<struct Selenium::WebDriver::Dimension width=640, height=480>]
     
       (compared using ==)
```

Not really sure what happened here, guessing there is a secondary blank 0x0 image somewhere on the page it's finding?

titusfortner:
This isn't a backward compatible change and the code isn't marked as `@api private`, even if maybe it should be. What are you trying to do and how does this change make it better?

rb/lib/selenium/webdriver/common/log_entry.rb
```diff
@@ -22,22 +22,22 @@ module WebDriver
     class LogEntry
       attr_reader :level, :timestamp, :message
 
-      def initialize(level, timestamp, message)
-        @level = level
+      def initialize(timestamp, level, message)
```
p0deje: Is there any value to change the order here? Even though nobody should be initializing this class directly, I don't believe it's worth breaking backwards compatibility without a good reason.

p0deje (commented):
I don't quite follow what do we standard against. Do you want the browser logs to look like Ruby's `Logger` class?

rb/lib/selenium/webdriver/common/log_entry.rb
luke-hill: I'll undo this change, but it was only used in the entire codebase in one location, which was api private.

luke-hill:
Is this RtM now @p0deje (The ruby failure seems to find a 0x0 reference).

p0deje:
@luke-hill Could you please reply to https://github.com/SeleniumHQ/selenium/pull/9850#pullrequestreview-758637516?

rb/lib/selenium/webdriver/common/log_entry.rb
luke-hill: Sorry yes. The other logger formats in a different way, and the timestamp is always first in most loggers. Here it wasn't.

PR #9856, opened 2021-09-21
Create interfaces for RemoteWebDriver to use with Augmenter

titusfortner:
This is similar to what we did for the Firefox capabilities, but I'm implementing, but I'm not sure this is the best way to do this (i.e. with the `ChromiumNetworkConditions` class). Would be a lot simpler to do ordered parameters, but this lets users only specify the things they care about.

Edit: I'm just going to add everything here to get feedback on it all.


java/src/org/openqa/selenium/chromium/AddHasCdp.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Arrays;
+import java.util.Map;
+import java.util.function.Predicate;
+
+import static org.openqa.selenium.remote.BrowserType.CHROME;
+import static org.openqa.selenium.remote.BrowserType.EDGE;
+
+public abstract class AddHasCdp implements AugmenterProvider<HasCdp>, AdditionalHttpCommands {
+
+    public static final String EXECUTE_CDP = "executeCdpCommand";
```
shs96c: Nit: indentation

java/src/org/openqa/selenium/chromium/AddHasCdp.java
```diff
@@ -0,0 +1,69 @@
...
+
+  @Override
+  public abstract Map<String, CommandInfo> getAdditionalCommands();
+
+  @Override
+  public Predicate<Capabilities> isApplicable() {
+    String[] validBrowsers = new String[] { EDGE, CHROME, "msedge" };
```
shs96c: Where do we see `msedge`?

shs96c: Make `validBrowsers` a `private` constant, and make it a `List<String>`.

java/src/org/openqa/selenium/chromium/AddHasCdp.java
```diff
@@ -0,0 +1,69 @@
...
+    return caps -> new ArrayList<>(Arrays.asList(validBrowsers)).contains(caps.getBrowserName());
```
shs96c: Because creating ephemeral lists just for a contains causes churn on the garbage collector (and, you know, the list is constant)

java/src/org/openqa/selenium/chromium/AddHasNetworkConditions.java
```diff
@@ -0,0 +1,99 @@
...
+    DELETE_NETWORK_CONDITIONS, new CommandInfo("/session/:sessionId/chromium/network_conditions", HttpMethod.DELETE));
+
+  @Override
+  public Map<String, CommandInfo> getAdditionalCommands() {
+    return COMMANDS;
+  }
+
+  @Override
+  public Predicate<Capabilities> isApplicable() {
+    String[] validBrowsers = new String[] { EDGE, CHROME, "msedge" };
```
shs96c: I've seen this list before... Perhaps a constant on `ChromiumDriver` itself? `KNOWN_CHROMIUM_BROWSERS`?

java/src/org/openqa/selenium/chromium/AddHasNetworkConditions.java
```diff
@@ -0,0 +1,99 @@
...
+    return caps -> new ArrayList<>(Arrays.asList(validBrowsers)).contains(caps.getBrowserName());
```
shs96c: And, again, we want to avoid ephemeral objects when we're doing this comparison.

java/src/org/openqa/selenium/chromium/AddHasPermissions.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Map;
+import java.util.function.Predicate;
+
+import static org.openqa.selenium.remote.BrowserType.CHROME;
+import static org.openqa.selenium.remote.BrowserType.EDGE;
+
+@AutoService({AdditionalHttpCommands.class, AugmenterProvider.class})
+public class AddHasPermissions implements AugmenterProvider<HasPermissions>, AdditionalHttpCommands {
+
+    public static final String SET_PERMISSION = "setPermission";
```
shs96c: Nit: indentation

java/src/org/openqa/selenium/chromium/AddHasPermissions.java
```diff
@@ -0,0 +1,74 @@
...
+    SET_PERMISSION, new CommandInfo("/session/:sessionId/permissions", HttpMethod.POST));
+
+  @Override
+  public Map<String, CommandInfo> getAdditionalCommands() {
+    return COMMANDS;
+  }
+
+  @Override
+  public Predicate<Capabilities> isApplicable() {
+    String[] validBrowsers = new String[] { EDGE, CHROME, "msedge" };
```
shs96c: Last time I make the same comment :) I suspect this is going to be a change that needs to be made throughout the PR.

java/src/org/openqa/selenium/chromium/AddHasPermissions.java
```diff
@@ -0,0 +1,74 @@
...
+
+  @Override
+  public HasPermissions getImplementation(Capabilities capabilities, ExecuteMethod executeMethod) {
+    return new HasPermissions() {
+      @Override
+      public void setPermission(String name, String value) {
+        Require.nonNull("Permission name", name);
+        Require.nonNull("Permission value", value);
+
+        executeMethod.execute(SET_PERMISSION, ImmutableMap.of("descriptor", ImmutableMap.of("name", name), "state", value));
```
shs96c, line 66: Was this `ImmutableMap` meant to contain both "name" and "state"?

java/src/org/openqa/selenium/chromium/ChromiumDriver.java
```diff
@@ -73,7 +72,12 @@
   HasTouchScreen,
   LocationContext,
   NetworkConnection,
-  WebStorage {
+  WebStorage,
```
shs96c: Alphabetise, please.

java/src/org/openqa/selenium/chromium/ChromiumNetworkConditions.java
```diff
@@ -0,0 +1,107 @@
...
+  public static final String OFFLINE = "offline";
+  public static final String LATENCY = "latency";
+  public static final String DOWNLOAD_THROUGHPUT = "download_throughput";
+  public static final String UPLOAD_THROUGHPUT = "upload_throughput";
+
+  /**
+   *
+   * @return whether network is simulated to be offline.
+   */
+  public Boolean getOffline() {
```
shs96c: `boolean` not `Boolean`. People are going to use this and expect boxing to work, which implies they'll never think of the `null` value.

java/src/org/openqa/selenium/chromium/ChromiumNetworkConditions.java
```diff
@@ -0,0 +1,107 @@
...
+package org.openqa.selenium.chromium;
+
+import java.time.Duration;
+
+/**
+ * Provides manipulation of getting and setting network conditions from Chromium.
+ *
+ */
+public class ChromiumNetworkConditions {
+  private Boolean offline = false;
```
shs96c: `boolean`

java/src/org/openqa/selenium/chromium/ChromiumNetworkConditions.java
```diff
@@ -0,0 +1,107 @@
...
+  private Duration latency = Duration.ZERO;
+  private Number downloadThroughput = -1;
```
shs96c: Pick a primitive. Probably either `long` or `int`

java/src/org/openqa/selenium/chromium/ChromiumNetworkConditions.java
```diff
@@ -0,0 +1,107 @@
...
+  public void setLatency(Duration latency) {
+    this.latency = latency;
+  }
+
+  /**
+   * The current throughput of the network connection in kb/second for downloading.
+   *
+   * @return the current download throughput in kb/second.
+   */
+  public Number getDownloadThroughput() {
```
shs96c: Return the primitive type.

java/src/org/openqa/selenium/chromium/ChromiumNetworkConditions.java
```diff
@@ -0,0 +1,107 @@
...
+    return downloadThroughput;
+  }
+
+  /**
+   * ets the throughput of the network connection in kb/second for downloading.
+   *
+   * @param downloadThroughput thoughput in kb/second
+   */
+  public void setDownloadThroughput(Number downloadThroughput) {
```
shs96c: Only allow a primitive type here.

java/src/org/openqa/selenium/chromium/ChromiumNetworkConditions.java
```diff
@@ -0,0 +1,107 @@
...
+    this.downloadThroughput = downloadThroughput;
+  }
+
+  /**
+   * The current throughput of the network connection in kb/second for uploading.
+   *
+   * @return the current upload throughput in kb/second.
+   */
+  public Number getUploadThroughput() {
```
shs96c: Same

java/src/org/openqa/selenium/chromium/ChromiumNetworkConditions.java
```diff
@@ -0,0 +1,107 @@
...
+    return uploadThroughput;
+  }
+
+  /**
+   * ets the throughput of the network connection in kb/second for uploading.
+   *
+   * @param uploadThroughput thoughput in kb/second
+   */
+  public void setUploadThroughput(Number uploadThroughput) {
```
shs96c: And here.

java/src/org/openqa/selenium/chromium/ChromiumNetworkConditions.java
```diff
@@ -0,0 +1,107 @@
...
+    this.uploadThroughput = uploadThroughput;
+  }
+}
```
shs96c, line 108: You may need a `private static ChromiumNetworkConditions fromJson(JsonInput input)` here, but if the setters take primitives, I think it'll work without it.

java/src/org/openqa/selenium/chromium/HasCasting.java
```diff
@@ -0,0 +1,66 @@
...
+@Beta
+public interface HasCasting {
+
+  /**
+   * Returns the list of cast sinks (Cast devices) available to the Chrome media router.
+   *
+   * @return array of ID / Name pairs of available cast sink targets
+   *
+   */
+  ArrayList<Map<String, String>> getCastSinks();
```
shs96c: `List<Map<String, String>>` Allow implementations to choose their own list implementation as required.

java/src/org/openqa/selenium/edge/AddHasCasting.java
```diff
@@ -0,0 +1,50 @@
...
+    STOP_CASTING, new CommandInfo("session/:sessionId/ms/cast/stop_casting", HttpMethod.POST));
+
+  @Override
+  public Map<String, CommandInfo> getAdditionalCommands() {
+    return COMMANDS;
+  }
+
+  @Override
+  public Predicate<Capabilities> isApplicable() {
+    return caps -> "msedge".equals(caps.getBrowserName());
```
shs96c, line 48: Not also `BrowserType.EDGE`?

java/src/org/openqa/selenium/edge/AddHasCdp.java
```diff
@@ -0,0 +1,46 @@
...
+    EXECUTE_CDP, new CommandInfo("session/:sessionId/ms/cdp/execute", HttpMethod.POST));
+
+  @Override
+  public Map<String, CommandInfo> getAdditionalCommands() {
+    return COMMANDS;
+  }
+
+  @Override
+  public Predicate<Capabilities> isApplicable() {
+    return caps -> "msedge".equals(caps.getBrowserName());
```
shs96c, line 44: And `BrowserType.EDGE`?

java/src/org/openqa/selenium/edge/BUILD.bazel
```diff
@@ -17,5 +18,6 @@ java_export(
         "//java/src/org/openqa/selenium:core",
         "//java/src/org/openqa/selenium/chromium",
         "//java/src/org/openqa/selenium/remote",
+        artifact("com.google.guava:guava"),
```
shs96c, line 21: :(

shs96c: It's fine, but at some point, we'll want to remove guava as a project dep as it's pretty heavy.

java/src/org/openqa/selenium/firefox/FirefoxDriver.java
```diff
@@ -265,13 +265,11 @@ private static boolean isLegacy(Capabilities desiredCapabilities) {
 
   @Override
   public String installExtension(Path path) {
-    Require.nonNull("Path", path);
```
shs96c: It's fine to leave these checks in. It'll make the exception come from `FirefoxDriver`, and that's probably clearer to a user.

java/src/org/openqa/selenium/chromium/HasPermissions.java
```diff
@@ -0,0 +1,34 @@
...
+
+package org.openqa.selenium.chromium;
+
+import org.openqa.selenium.Beta;
+
+/**
+ * Used by classes to indicate that they can adjust permissions compatible items.
+ */
+@Beta
+public interface HasPermissions {
```
shs96c, line 26: Looking around, it seems like this comes from the [Permissions spec](https://www.w3.org/TR/permissions/) As such, it's not browser-specific, and should probably live in the `remote` package. According to [caniuse](https://caniuse.com/?search=permissions) this is implemented by Edge and Firefox too.

The Safari interface with the same name appears to be different.

java/src/org/openqa/selenium/safari/HasPermissions.java
```diff
@@ -0,0 +1,44 @@
...
+public interface HasPermissions {
+
+  /**
+   * Set permission on the browser.
+   * The only supported permission at this time is "getUserMedia".
+   *
+   * @param permission the name of the item to set permission on.
+   * @param value whether the permission has been granted.
+   */
+  void setPermissions(String permission, Boolean value);
```
shs96c: `boolean`, as setting the `null` value makes no sense at all.

java/src/org/openqa/selenium/safari/HasPermissions.java
```diff
@@ -0,0 +1,44 @@
...
+
+  /**
+   *
+   * @return each permission and whether it is allowed or not.
+   */
+  Map<String, Boolean> getPermissions();
```
shs96c, line 43: But here it's unavoidable to use `Boolean`

java/test/org/openqa/selenium/edge/EdgeDriverFunctionalTest.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+public class EdgeDriverFunctionalTest extends JUnit4TestBase {
+
+  @Test
+  public void shouldAllowRemoteWebDriverToAugmentHasNetworkConditions() throws MalformedURLException {
+    WebDriver driver = new RemoteWebDriver(new URL("http://localhost:4444/"), new EdgeOptions());
```
shs96c: No! This will never work! The tests won't start a server on port 4444: they pick a random port. Worse, because you're extending `JUnit4TestBase`, you already have a `driver` to work with.

shs96c: I skimmed through the other tests, and the same thing happens in them. You can't rely on a selenium server being run on a specific port. If you _really_ need to create a new webdriver instance (you don't, because the test already has one), then use `new WebDriverBuilder` from the test package.

java/test/org/openqa/selenium/edge/EdgeDriverFunctionalTest.java
```diff
@@ -0,0 +1,137 @@
...
+      driver.quit();
+    }
+  }
+
+  @Test
+  public void shouldCast() throws InterruptedException {
+    EdgeDriver driver = new EdgeDriver();
+
+    try {
+      // Does not get list the first time it is called
```
shs96c: Magical

java/test/org/openqa/selenium/edge/EdgeDriverFunctionalTest.java
shs96c: Obviously, assign the augmented browser to a variable, otherwise you'll break test isolation pretty handily.

java/test/org/openqa/selenium/edge/EdgeDriverFunctionalTest.java
```diff
@@ -0,0 +1,137 @@
...
+  @Test
+  public void shouldAllowRemoteWebDriverToAugmentHasCasting() throws InterruptedException, MalformedURLException {
+    WebDriver driver = new RemoteWebDriver(new URL("http://localhost:4444/"), new EdgeOptions());
+    WebDriver augmentedDriver = new Augmenter().augment(driver);
+
+    try {
+      // Does not get list the first time it is called
+      ((HasCasting) augmentedDriver).getCastSinks();
+      Thread.sleep(1000);
+      ArrayList<Map<String, String>> castSinks = ((HasCasting) augmentedDriver).getCastSinks();
```
shs96c: `List`

shs96c (changes requested):

java/src/org/openqa/selenium/chromium/AddHasCdp.java
titusfortner: `msedge` is what the Edge driver reports itself as being in the Capabilities. :-/

java/src/org/openqa/selenium/chromium/AddHasPermissions.java
titusfortner: this looks weird, but it is correct!

java/src/org/openqa/selenium/edge/AddHasCasting.java
titusfortner: I probably should remove BrowserType.EDGE everywhere else since it's always either "chrome" or "msedge"

java/src/org/openqa/selenium/edge/BUILD.bazel
titusfortner: I just kept pulling levers until bazel was happy. I do not know what actually needed this, but adding this line made it work. :)

java/src/org/openqa/selenium/firefox/FirefoxDriver.java
titusfortner: I moved all of these to the implementation for consistency. Should I move them back out of the implementation to the driver classes, etc?

java/test/org/openqa/selenium/edge/EdgeDriverFunctionalTest.java
titusfortner: It's obvious that these test files aren't the ideal place for RemoteWebDriver class instances, but this is all patterned off of the test you wrote for me: https://github.com/SeleniumHQ/selenium/commit/0a5b2d8e747a2252fb4a2abada713a68a58ffe38#diff-6ff6428e62eafbf44f2cee9400a71f0a7f3276807d4b611000b7434c582752a3R523-R536

I think I'm doing the right thing in the places where I'm not using a RemoteWebDriver instance 
e.g., https://github.com/SeleniumHQ/selenium/pull/9856/files#diff-0bc642dab2d557261d4507422be800146f1ff78a1889f6f728e7eae08a76953fR117

I have tests for both "Augmented" and "with Builder" and I don't need them both, but I'm not sure the actual right way to do this.

java/test/org/openqa/selenium/edge/EdgeDriverFunctionalTest.java
titusfortner: /shrug pull all the levers this is what works. I guess I could file a bug...

java/src/org/openqa/selenium/chromium/ChromiumNetworkConditions.java
titusfortner: I mean, the tests were passing even with `Number`, so the right thing has to be happening?

java/src/org/openqa/selenium/edge/AddHasCasting.java
titusfortner: I mean, ideally `BrowserType.EDGE == "msedge"` since that is how Edge self-identifies. :)
but that's probably not backwards compatible? maybe I avoid using the constants in KNOWN_CHROMIUM_BROWSERS entirely and just use the two Strings.

titusfortner:
I marked all the conversations as resolved for the things I fixed in the most recent commit. These others seem slightly more involved, but I'll see what I can make of them tonight so we can discuss tomorrow.

java/src/org/openqa/selenium/chromium/HasPermissions.java
titusfortner: So, for Firefox, I'm getting: 
> TypeError: 'clipboard-read' (value of 'name' member of PermissionDescriptor) is not a valid value for enumeration PermissionName.

So, I'm not sure it's actually supported the same way? Should this get added to spec (a webdriver section to Permissions I'm assuming) so everyone does the same thing?

java/test/org/openqa/selenium/edge/EdgeDriverFunctionalTest.java
shs96c: Discussed over a call. We can tag tests with `selenium-remote` and this will automatically create a test target that runs the test with the selenium server running.

java/src/org/openqa/selenium/edge/AddHasCasting.java
titusfortner: Opera needs to be on this list, so added it and this is best implementation for now

java/src/org/openqa/selenium/chromium/HasPermissions.java
titusfortner: I'll discuss with Mozilla and change if they do implement the endpoint (didn't look like it in my tests)

java/test/org/openqa/selenium/edge/EdgeDriverFunctionalTest.java
titusfortner: fixed for driver tests

java/src/org/openqa/selenium/firefox/FirefoxDriver.java
titusfortner: added to overridden driver classes

java/test/org/openqa/selenium/edge/EdgeDriverFunctionalTest.java
titusfortner: bug filed https://bugs.chromium.org/p/chromedriver/issues/detail?id=3924

java/src/org/openqa/selenium/chromium/AddHasCasting.java
```diff
@@ -0,0 +1,81 @@
...
+
+  @Override
+  public Class<HasCasting> getDescribedInterface() {
+    return HasCasting.class;
+  }
+
+  @Override
+  public HasCasting getImplementation(Capabilities capabilities, ExecuteMethod executeMethod) {
+    return new HasCasting() {
+      @Override public List<Map<String, String>> getCastSinks() {
```
shs96c, line 54: Style nit: put the `@Override` on its own line. What you've got here compiles, but it's not the style we follow on the project.

java/src/org/openqa/selenium/chromium/AddHasNetworkConditions.java
```diff
@@ -0,0 +1,95 @@
...
+@AutoService({AdditionalHttpCommands.class, AugmenterProvider.class})
+public class AddHasNetworkConditions implements AugmenterProvider<HasNetworkConditions>, AdditionalHttpCommands {
+
+  public static final String GET_NETWORK_CONDITIONS = "getNetworkConditions";
+  public static final String SET_NETWORK_CONDITIONS = "setNetworkConditions";
+  public static final String DELETE_NETWORK_CONDITIONS = "deleteNetworkConditions";
+
+
+  private static final Map<String, CommandInfo> COMMANDS = ImmutableMap.of(
+    GET_NETWORK_CONDITIONS, new CommandInfo("/session/:sessionId/chromium/network_conditions",HttpMethod.GET),
```
shs96c, line 45: Nit: a space after the `,` -> `", HttpMethod.GET)`

java/src/org/openqa/selenium/chromium/HasCdp.java
```diff
@@ -0,0 +1,43 @@
...
+   * <a href="https://chromedevtools.github.io/devtools-protocol/">chrome
+   * devtools protocol domains/commands</a>.
+   *
+   * It is strongly encouraged to use {@link org.openqa.selenium.devtools.DevTools} API instead of this
+   *
+   * @param commandName the command to execute with Chrome Dev Tools.
+   * @param parameters any information needed to execute the Dev Tools command.
+   * @return the name and value of the response.
+   */
+  public Map<String, Object> executeCdpCommand(String commandName, Map<String, Object> parameters);
```
shs96c, line 42: Nit: Methods on interfaces are public by default.

shs96c (approved):
Let's get this in, and then we can iterate on the nits.

PR #9857, opened 2021-09-21
[js] Allow builder to set a single arbitrary capability

### LINKED ISSUE #9830: Convenience Builder method to set arbitrary capabilities (JavaScript/Node.js)
## 🚀 Feature Proposal

Now that `Builder.withCapabilities()` is [deprecated](https://github.com/SeleniumHQ/selenium/pull/9701), there is no longer a convenient and non-deprecated way to set arbitrary capabilities such as [`sauce:options`](https://docs.saucelabs.com/dev/test-configuration-options/#desktop-and-mobile-capabilities-sauce-specific--optional) when building a new `WebDriver` instance.

This feature proposal is to restore the convenience of `.withCapabilities()`, but with different semantics so that users are still encouraged to use the `.setBrowserOptions()` methods.

> ### Class Builder Instance Methods
>
> #### _`this.`_**setCapability(** _`key`_ **,** _`value`_ **)** → `Builder`
>
> Sets the desired capability when requesting a new session. If there is already a capability named _`key`_, its value will be overwritten with _`value`_. This is a convenience wrapper around `builder.getCapabilities().set(key, value)` to support Builder method chaining.
>
> ##### Parameters
>   * _`key`_ `string`: The capability key.
>   * _`value`_ `*`: The capability value.
>
> ##### Returns
> `Builder`: A self reference.

## Motivation

Before `Builder.withCapabilities()` was deprecated, it was possible to do this:

```javascript
  const driver = await new wd.Builder()
      .usingServer('https://ondemand.saucelabs.com/wd/hub')
      .withCapabilities({'sauce:options': {
        username: process.env.SAUCE_USERNAME,
        accessKey: process.env.SAUCE_ACCESS_KEY,
        tunnelIdentifier: process.env.SAUCE_TUNNEL_ID,
      }})
      .forBrowser(browser, version, platform)
      .setChromeOptions(chromeOptions)
      .setEdgeOptions(edgeOptions)
      .setFirefoxOptions(firefoxOptions)
      .build();
```

Users now have to do something like this if they want to avoid the deprecated method:

```javascript
  const builder = new wd.Builder()
      .usingServer('https://ondemand.saucelabs.com/wd/hub')
      .forBrowser(browser, version, platform)
      .setChromeOptions(chromeOptions)
      .setEdgeOptions(edgeOptions)
      .setFirefoxOptions(firefoxOptions);
  builder.getCapabilities().set('sauce:options', {
    username: process.env.SAUCE_USERNAME,
    accessKey: process.env.SAUCE_ACCESS_KEY,
    tunnelIdentifier: process.env.SAUCE_TUNNEL_ID,
  });
  const driver = await builder.build();
```

## Example

```javascript
  const driver = await new wd.Builder()
      .usingServer('https://ondemand.saucelabs.com/wd/hub')
      .setCapability('sauce:options', {
        username: process.env.SAUCE_USERNAME,
        accessKey: process.env.SAUCE_ACCESS_KEY,
        tunnelIdentifier: process.env.SAUCE_TUNNEL_ID,
      })
      .forBrowser(browser, version, platform)
      .setChromeOptions(chromeOptions)
      .setEdgeOptions(edgeOptions)
      .setFirefoxOptions(firefoxOptions)
      .build();
```

pujagani:

### Description

Allow builder to set a single arbitrary capability. Fixes #9830.

### Motivation and Context

Details are explained in #9830

### Types of changes

### Checklist


titusfortner:
I think `setCapability()` is the "right" method name, but might be worth considering using `withCapabilities()` since that sounds like it is the backward compatible name.

Also, can additional capabilities can be added to the browser options class like it can be in the other languages? Makes sense to have it in both places.

pujagani:
> I think `setCapability()` is the "right" method name, but might be worth considering using `withCapabilities()` since that sounds like it is the backward compatible name.

@AutomatedTester  You might have more context regarding this since it was deprecated for a reason. Can you please help answer this one? Thank you!

> Also, can additional capabilities can be added to the browser options class like it can be in the other languages? Makes sense to have it in both places.

Yes, this is possible since the Options class inherits from Capabilities. So it is possible to set a capability using Options as with other languages.

AutomatedTester:
@titusfortner has been looking at these in all the different bindings, maybe he can leave his thoughts.

titusfortner:
My take is that as long as there's a `setCapability()` in the options class that works to set top level capabilities, then we should restore `withCapabilities()` for this functionality the builder. It's not the same method name as Java's builder class, but I think maintaining backwards compatibility would be more useful here.

pujagani:
I agree with @titusfortner. I will restore `withCapabilities` in a separate PR to ensure backward compatibility. But since other language bindings have a `setCapability()` in the builder, I might be a good idea to have both.

PR #9862, opened 2021-09-22
[rb] bugfix: Loading type of driver can fail if previously loaded files are not present

luke-hill:
Fix issue where requiring a type of webdriver file would autoload all required files except the platform module. This then causes a failure when checking paths

### DIFF
--- a/rb/lib/selenium/webdriver/chrome.rb
+++ b/rb/lib/selenium/webdriver/chrome.rb
@@ -19,6 +19,8 @@
 
 require 'net/http'
 
+require 'selenium/webdriver/common/platform'
+
 module Selenium
   module WebDriver
     module Chrome
--- a/rb/lib/selenium/webdriver/edge.rb
+++ b/rb/lib/selenium/webdriver/edge.rb
@@ -19,6 +19,8 @@
 
 require 'net/http'
 
+require 'selenium/webdriver/common/platform'
+
 module Selenium
   module WebDriver
     module Edge
--- a/rb/lib/selenium/webdriver/firefox.rb
+++ b/rb/lib/selenium/webdriver/firefox.rb
@@ -21,6 +21,8 @@ require 'timeout'
 require 'socket'
 require 'rexml/document'
 
+require 'selenium/webdriver/common/platform'
+
 module Selenium
   module WebDriver
     module Firefox
--- a/rb/lib/selenium/webdriver/safari.rb
+++ b/rb/lib/selenium/webdriver/safari.rb
@@ -17,6 +17,8 @@
 # specific language governing permissions and limitations
 # under the License.
 
+require 'selenium/webdriver/common/platform'
+
 module Selenium
   module WebDriver
     module Safari


titusfortner:
Can you provide a reproducible case of something failing when this is not added? I'm not seeing the use case for it.

luke-hill:
If you require any of those files where it is added, code will fail with an NME, because Platform is not defined within there, nor autoloaded.

So a 2-liner would be

```
require "selenium/webdriver/chrome"

Selenium::WebDriver::Chrome.foo
```

The one that procedurally fails (And is **the most** important), is firefox. Because this cannot even parse. i.e. if you just require that fail it falls over.

titusfortner:
Chrome seems to work as expected:
```
irb(main):003:0> require 'selenium/webdriver/chrome'
=> true
irb(main):004:0> Selenium::WebDriver::Chrome.path
=> nil
irb(main):005:0> 
```

but I see that Firefox uses:
```
      DEFAULT_ENABLE_NATIVE_EVENTS = Platform.os == :windows
```
We no longer use that constant, so I deleted it.

The project isn't really set up to load browser libraries in isolation, but your example should work fine on next release.

luke-hill:
```
2.7.4 :001 > require 'selenium/webdriver/chrome'
 => true 
2.7.4 :002 > Selenium::WebDriver::Chrome.path
 => nil 
2.7.4 :003 > Selenium::WebDriver::Chrome.path = 'foo'
        5: from /usr/share/rvm/rubies/ruby-2.7.4/bin/irb:23:in `<main>'
        4: from /usr/share/rvm/rubies/ruby-2.7.4/bin/irb:23:in `load'
        3: from /usr/share/rvm/rubies/ruby-2.7.4/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
        2: from (irb):3
        1: from /home/luke/.rvm/gems/ruby-2.7.4/gems/selenium-webdriver-4.0.0.alpha6/lib/selenium/webdriver/chrome.rb:46:in `path='
NameError (uninitialized constant Selenium::WebDriver::Chrome::Platform)
Did you mean?  RUBY_PLATFORM
```

luke-hill:
I think we should just merge this bug-fix in AS IS. And then worry about deleting constants e.t.c. maybe at a later date, once the fix is no longer needed? WDYT?

titusfortner:
My goal in evaluating a PR is to figure out what is the underlying problem, and determine if the proffered solution is a maintainable way to address that problem, or if there is an alternative approach that avoids the problem that should be preferred. If there is an alternative, a case needs to be made for why it shouldn't be preferred. I take this approach to all PRs, whether it is one line or 6 new classes. 

So, first, it is essential when you make a PR to demonstrate the underlying problem. You didn't provide a reproducible issue until your third message.

I now understand the underlying problem to be that Selenium doesn't support using `require 'selenium/webdriver/chrome'`.

The preferred solution is to use `require 'selenium-webdriver'`, and if you don't want to do that, you need to make the case for why you shouldn't have to.

Further, this particular PR doesn't doesn't fix the underlying problem in a maintainable way. (e.g., what happens if a logger message is added, etc). If this problem is important enough to fix, then we should re-think how we are loading all the classes in general rather than needing to add requires statements every time something changes.

PR #9864, opened 2021-09-23
[rb] move new window functionality into TargetLocator

titusfortner:
This is essentially to match the other bindings

I tied it to the existing `TargetLoctor#window` method to allow using it with a block, but instead of just switching back to the original window, I changed it so that it closes the window and switches back. Is that too cute? It makes more sense for it to be like this if it is going to take a block, but maybe it would be surprising to a user?

### DIFF
--- a/rb/lib/selenium/webdriver/common/manager.rb
+++ b/rb/lib/selenium/webdriver/common/manager.rb
@@ -114,15 +114,13 @@ module Selenium
       end
 
       #
-      # Create a new top-level browsing context
-      # https://w3c.github.io/webdriver/#new-window
       # @param type [Symbol] Supports two values: :tab and :window.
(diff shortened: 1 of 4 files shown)


p0deje (commented):
Shall we deprecate the older API and change it in 4.1?

p0deje:
> I tied it to the existing TargetLoctor#window method to allow using it with a block, but instead of just switching back to the original window, I changed it so that it closes the window and switches back. Is that too cute? It makes more sense for it to be like this if it is going to take a block, but maybe it would be surprising to a user?

It's hard to say if it's going to be surprising (I've never used this method personally), but I'd keep the original behavior to simplify the upgrade.

titusfortner:
the original method doesn't take a block. So it's new either way. :)

p0deje (approved):

PR #9867, opened 2021-09-23
Py commands

titusfortner:
I have a checklist of all of the browser specific commands to add to each of the bindings; Python had most of them implemented already, so these were all I needed to add to make it the first one to have everything. @AutomatedTester let me know if I did the deprecation correctly.


titusfortner:
Hmm, I was wrong it went the other way:
https://bugs.chromium.org/p/chromedriver/issues/detail?id=2307
I'll "unfix" this. :)

titusfortner:
So, I can't test the set_permission method because it requires a page with a secure protocol, and it doesn't appear the python tests support that. I've verified that the command executes and returns the "unable to execute on an insecure page" error, not sure if it is worth adding a test for that.

AutomatedTester:
> I have a checklist of all of the browser specific commands to add to each of the bindings; Python had most of them implemented already, so these were all I needed to add to make it the first one to have everything. @AutomatedTester let me know if I did the deprecation correctly.

I can't see any deprecations in the commit, where did you add them?

titusfortner:
I thought one thing was deprecated and needed to be replaced, turns out what I was looking to replace it with was the deprecated thing, so what you already have is correct. :) Actually, this PR code pretty trivial now, I'm just merging...

PR #9868, opened 2021-09-24
[dotnet] mark deprecated chromedriver commands as obsolete and implem…

titusfortner:
The commands being used here are deprecated according to: https://bugs.chromium.org/p/chromedriver/issues/detail?id=2307

I have a basic template of a test I'm using for other languages, but I'm not sure how to write/run tests in .NET. 

There are a dozen custom browser commands I'd like to add once I make sure I'm doing everything correctly, so please let me know if I'm violating conventions, etc.

### DIFF
--- a/dotnet/src/webdriver/Chromium/ChromiumDriver.cs
+++ b/dotnet/src/webdriver/Chromium/ChromiumDriver.cs
@@ -43,6 +43,7 @@ namespace OpenQA.Selenium.Chromium
         private const string DeleteNetworkConditionsCommand = "deleteNetworkConditions";
         private const string SendChromeCommand = "sendChromeCommand";
         private const string SendChromeCommandWithResult = "sendChromeCommandWithResult";
+        protected const string ExecuteCdp = "executeCdpCommand";
 
         private readonly string optionsCapabilityName;
         private DevToolsSession devToolsSession;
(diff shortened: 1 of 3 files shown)


dotnet/src/webdriver/Chromium/ChromiumDriver.cs
```diff
@@ -128,11 +129,12 @@ public ChromiumNetworkConditions NetworkConditions
...
 
         /// <summary>
-        /// Executes a custom Chrome command.
+        /// Executes a custom Chrome Dev Tools Protocol Command.
         /// </summary>
         /// <param name="commandName">Name of the command to execute.</param>
         /// <param name="commandParameters">Parameters of the command to execute.</param>
-        public void ExecuteChromeCommand(string commandName, Dictionary<string, object> commandParameters)
+        /// <returns>An object representing the result of the command, if applicable.</returns>
+        public object ExecuteCdp(string commandName, Dictionary<string, object> commandParameters)
```
jimevans: Change this to `ExecuteCdpCommand`.

dotnet/src/webdriver/Chromium/ChromiumDriver.cs
```diff
@@ -43,6 +43,7 @@ public abstract class ChromiumDriver : WebDriver, ISupportsLogs, IDevTools
         private const string DeleteNetworkConditionsCommand = "deleteNetworkConditions";
         private const string SendChromeCommand = "sendChromeCommand";
         private const string SendChromeCommandWithResult = "sendChromeCommandWithResult";
+        protected const string ExecuteCdpCommand = "executeCdpCommand";
```
jimevans: Change this to something arbitrary. Keep the value the same (`"executeCdpCommand"`), but just change the field name. Maybe to `ExecuteCdp`, or `ExecuteDevToolsCommand`.

jimevans (changes requested):

titusfortner:
merged: e6fb3feeb6f73e6098aa0f7b3fc2077e1ca2bf9a

PR #9872, opened 2021-09-28
[java] use FileDetector to install Firefox addons

titusfortner:
I don't see an ideal way to do this.

The whole point is to be able to do this with the Augmenter, which means the interfaces which don't know about RemoteWebDriver methods, and I don't want to put the driver instance as a parameter for the method in the interface. 

~This is the best I came up with.:~

~The implementation is mostly copied from `RemoteWebElement#upload` but again, I don't see an easy way to put this logic where both of them can access it.~

Edit:
I changed this so it tries the string as provided first, and if that doesn't work, it determines if the file exists locally, and if so uploads it and retries. 

Not sure it's ideal either, but it won't break anything, and won't keep us from doing something better later.

### DIFF
--- a/java/src/org/openqa/selenium/firefox/AddHasExtensions.java
+++ b/java/src/org/openqa/selenium/firefox/AddHasExtensions.java
@@ -20,18 +20,26 @@ package org.openqa.selenium.firefox;
 import com.google.auto.service.AutoService;
 import com.google.common.collect.ImmutableMap;
 import org.openqa.selenium.Capabilities;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebDriverException;
+import org.openqa.selenium.WrapsDriver;
 import org.openqa.selenium.internal.Require;
+import org.openqa.selenium.io.Zip;
 import org.openqa.selenium.remote.AdditionalHttpCommands;
 import org.openqa.selenium.remote.AugmenterProvider;
 import org.openqa.selenium.remote.CommandInfo;
 import org.openqa.selenium.remote.ExecuteMethod;
+import org.openqa.selenium.remote.RemoteWebDriver;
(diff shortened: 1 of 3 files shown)


java/src/org/openqa/selenium/firefox/FirefoxDriver.java
```diff
@@ -269,6 +269,13 @@ public String installExtension(Path path) {
     return extensions.installExtension(path);
   }
 
+  @Override
+  public String installExtension(Path path, FileDetector detector) {
```
shs96c: Users really shouldn't need to know about the presence of the file detector.

java/src/org/openqa/selenium/firefox/AddHasExtensions.java
```diff
@@ -65,9 +72,24 @@ public HasExtensions getImplementation(Capabilities capabilities, ExecuteMethod
...
 
-        return (String) executeMethod.execute(
-          INSTALL_EXTENSION,
-          ImmutableMap.of("path", path.toAbsolutePath().toString(), "temporary", false));
+        try {
+          return (String) executeMethod.execute(
+            INSTALL_EXTENSION,
+            ImmutableMap.of("path", path.toAbsolutePath().toString(), "temporary", false));
+        } catch (WebDriverException ex) {
+          File localFile = new LocalFileDetector().getLocalFile(path.toString());
```
shs96c: The more correct thing to do would be to make `RemoteExecuteMethod` implement `WrapsDriver`. You could then get the driver, check if it's remote, and then get the file detector that way. Relying on exceptions for flow control is generally frowned upon.

shs96c (changes requested):

shs96c (approved):

PR #9873, opened 2021-09-28
Implement browser specific functionality 

titusfortner:
Mostly double checking names are good (I'm matching Java on method names), and also need to know if I should be casting to String for message, and if (and/or how) I should cast to an Array or List for the `GetCastSinks` response. Fwiw, Java does: `(List<Map<String, String>>)`

Edit:
Went ahead and added all of the additional commands to bring it in parity with other languages.

### DIFF
--- a/dotnet/src/webdriver/Chromium/ChromiumDriver.cs
+++ b/dotnet/src/webdriver/Chromium/ChromiumDriver.cs
@@ -50,6 +50,7 @@ namespace OpenQA.Selenium.Chromium
         private const string SendChromeCommand = "sendChromeCommand";
         private const string SendChromeCommandWithResult = "sendChromeCommandWithResult";
         private const string LaunchAppCommand = "launchAppCommand";
+        private const string SetPermissionCommand = "setPermission";
 
         private readonly string optionsCapabilityName;
         private DevToolsSession devToolsSession;
@@ -83,6 +84,7 @@ namespace OpenQA.Selenium.Chromium
             this.AddCustomChromeCommand(SendChromeCommand, HttpCommandInfo.PostCommand, "/session/{sessionId}/chromium/send_command");
             this.AddCustomChromeCommand(SendChromeCommandWithResult, HttpCommandInfo.PostCommand, "/session/{sessionId}/chromium/send_command_and_get_result");
             this.AddCustomChromeCommand(LaunchAppCommand, HttpCommandInfo.PostCommand, "/session/{sessionId}/chromium/launch_app");
+            this.AddCustomChromeCommand(SetPermissionCommand, HttpCommandInfo.PostCommand, "/session/{sessionId}/permissions");
         }
 
         /// <summary>
@@ -149,6 +151,31 @@ namespace OpenQA.Selenium.Chromium
             Dictionary<string, object> parameters = new Dictionary<string, object>();
             parameters["id"] = id;
             this.Execute(LaunchAppCommand, parameters);
+        }
+
+        /// <summary>
+        /// Set supported permission on browser.
+        /// </summary>
+        /// <param name="permissionName">Name of item to set the permission on.</param>
+        /// <param name="permissionValue">Value to set the permission to.</param>
+        public void SetPermission(string permissionName, string permissionValue)
+        {
+            if (permissionName == null)
+            {
+                throw new ArgumentNullException("permissionName", "name must not be null");
+            }
+
+            if (permissionValue == null)
+            {
+                throw new ArgumentNullException("permissionValue", "value must not be null");
+            }
+
+            Dictionary<string, object> nameParameter = new Dictionary<string, object>();
+            nameParameter["name"] = permissionName;
+            Dictionary<string, object> parameters = new Dictionary<string, object>();
+            parameters["descriptor"] = nameParameter;
+            parameters["state"] = permissionValue;
+            this.Execute(SetPermissionCommand, parameters);
         }
 
         /// <summary>
--- a/dotnet/src/webdriver/Chromium/ChromiumDriver.cs
+++ b/dotnet/src/webdriver/Chromium/ChromiumDriver.cs
@@ -49,6 +49,7 @@ namespace OpenQA.Selenium.Chromium
         private const string DeleteNetworkConditionsCommand = "deleteNetworkConditions";
         private const string SendChromeCommand = "sendChromeCommand";
         private const string SendChromeCommandWithResult = "sendChromeCommandWithResult";
+        private const string LaunchAppCommand = "launchAppCommand";
 
         private readonly string optionsCapabilityName;
         private DevToolsSession devToolsSession;
@@ -81,6 +82,7 @@ namespace OpenQA.Selenium.Chromium
             this.AddCustomChromeCommand(DeleteNetworkConditionsCommand, HttpCommandInfo.DeleteCommand, "/session/{sessionId}/chromium/network_conditions");
             this.AddCustomChromeCommand(SendChromeCommand, HttpCommandInfo.PostCommand, "/session/{sessionId}/chromium/send_command");
             this.AddCustomChromeCommand(SendChromeCommandWithResult, HttpCommandInfo.PostCommand, "/session/{sessionId}/chromium/send_command_and_get_result");
+            this.AddCustomChromeCommand(LaunchAppCommand, HttpCommandInfo.PostCommand, "/session/{sessionId}/chromium/launch_app");
         }
 
         /// <summary>
@@ -131,6 +133,22 @@ namespace OpenQA.Selenium.Chromium
                 parameters["network_conditions"] = value.ToDictionary();
                 this.Execute(SetNetworkConditionsCommand, parameters);
             }
+        }
+
+        /// <summary>
+        /// Launches a Chromium based application.
+        /// </summary>
+        /// <param name="id">ID of the chromium app to launch.</param>
+        public void LaunchApp(string id)
+        {
+            if (id == null)
+            {
+                throw new ArgumentNullException("id", "id must not be null");
+            }
+
+            Dictionary<string, object> parameters = new Dictionary<string, object>();
+            parameters["id"] = id;
+            this.Execute(LaunchAppCommand, parameters);
         }
 
         /// <summary>
--- a/dotnet/src/webdriver/Chromium/ChromiumDriver.cs
+++ b/dotnet/src/webdriver/Chromium/ChromiumDriver.cs
@@ -38,12 +38,17 @@ namespace OpenQA.Selenium.Chromium
         /// </summary>
         public static readonly bool AcceptUntrustedCertificates = true;
 
+        protected const string ExecuteCdp = "executeCdpCommand";
+        protected const string GetCastSinksCommand = "getCastSinks";
+        protected const string SelectCastSinkCommand = "selectCastSink";
+        protected const string StartCastTabMirroringCommand = "startCastTabMirroring";
+        protected const string GetCastIssueMessageCommand = "getCastIssueMessage";
+        protected const string StopCastingCommand = "stopCasting";
         private const string GetNetworkConditionsCommand = "getNetworkConditions";
         private const string SetNetworkConditionsCommand = "setNetworkConditions";
         private const string DeleteNetworkConditionsCommand = "deleteNetworkConditions";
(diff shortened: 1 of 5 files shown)


dotnet/src/webdriver/Chromium/ChromiumDriver.cs
```diff
@@ -128,6 +137,47 @@ public ChromiumNetworkConditions NetworkConditions
...
+            parameters["id"] = id;
+            this.Execute(ExecuteCdp, parameters);
+        }
+
+        /// <summary>
+        /// Set supported permission on browser.
+        /// </summary>
+        /// <param name="name">Name of item to set the permission on.</param>
+        /// <param name="value">Value to set the permission to.</param>
+        public void setPermission(string name, string value)
```
jimevans: Method names should be `PascalCase` (upper-case first character).

dotnet/src/webdriver/Chromium/ChromiumDriver.cs
```diff
@@ -44,6 +44,13 @@ public abstract class ChromiumDriver : WebDriver, ISupportsLogs, IDevTools
         private const string SendChromeCommand = "sendChromeCommand";
         private const string SendChromeCommandWithResult = "sendChromeCommandWithResult";
         protected const string ExecuteCdp = "executeCdpCommand";
+        private const string LaunchAppCommand = "launchAppCommand";
```
jimevans, line 52: Nit: Move the `private` declarations below the `protected` ones.

dotnet/src/webdriver/Firefox/FirefoxDriver.cs
```diff
@@ -178,6 +180,28 @@ public override IFileDetector FileDetector
...
+        /// <summary>
+        /// Sets the command context used when issuing commands to geckodriver.
+        /// </summary>
+        /// <exception cref="WebDriverException">If response is not recognized</exception>
+        /// <returns>The context of commands.</returns>
+        public FirefoxCommandContext GetContext()
+        {
+            string response = this.Execute(GetContextCommand, null).Value.ToString();
+
+            if (response == "Content")
```
jimevans: A better approach would be (`true` in the below is a case-insensitive comparison):
```csharp
FirefoxCommandContext output;
bool success = Enum.TryParse<FirefoxCommandContext>(response, true, out output);
if (!success)
{
    throw new WebDriverException(string.Format(CultureInfo.InvariantCulture, "Do not recognize response: {0}; expected Context or Chrome";
}

return output;
}
```

dotnet/src/webdriver/Safari/SafariDriver.cs
```diff
@@ -61,6 +62,10 @@ namespace OpenQA.Selenium.Safari
     /// </example>
     public class SafariDriver : WebDriver
     {
+        private const string OpenInspector = "attachDebugger";
```
jimevans: Suffix each of these constants with `Command` (e.g., `OpenInspectorCommand`, `GetPermissionsCommand`, etc.).

dotnet/src/webdriver/Chromium/ChromiumDriver.cs
```diff
@@ -44,6 +44,13 @@ public abstract class ChromiumDriver : WebDriver, ISupportsLogs, IDevTools
...
+        private const string Permissions = "setPermission";
```
jimevans: Rename this `SetPermissionCommand`.

dotnet/src/webdriver/Chromium/ChromiumDriver.cs
```diff
@@ -128,6 +137,47 @@ public ChromiumNetworkConditions NetworkConditions
...
+        public void LaunchApp(string id)
+        {
+            if (id == null)
+            {
+                throw new ArgumentNullException("id", "id must not be null");
+            }
+
+            Dictionary<string, object> parameters = new Dictionary<string, object>();
+            parameters["id"] = id;
+            this.Execute(ExecuteCdp, parameters);
```
jimevans: Is `ExecuteCdp` the wrong command name?

dotnet/src/webdriver/Chromium/ChromiumDriver.cs
```diff
@@ -128,6 +137,47 @@ public ChromiumNetworkConditions NetworkConditions
...
+            {
+                throw new ArgumentNullException("value", "value must not be null");
+            }
+
+            Dictionary<string, object> nameParameter = new Dictionary<string, object>();
+            nameParameter["name"] = name;
+            Dictionary<string, object> parameters = new Dictionary<string, object>();
+            parameters["descriptor"] = nameParameter;
+            parameters["state"] = value;
+            this.Execute(ExecuteCdp, parameters);
```
jimevans: Is `ExecuteCdp` the wrong command name?

dotnet/src/webdriver/Chromium/ChromiumDriver.cs
```diff
@@ -234,6 +284,74 @@ public void CloseDevToolsSession()
             }
         }
 
+        /// <summary>
+        /// Returns the list of cast sinks (Cast devices) available to the Chrome media router.
+        /// </summary>
+        /// <returns>An object representing the list of available sinks.</returns>
+        public object GetCastSinks()
```
jimevans, line 291: What, exactly, does this return? We should coerce it to a proper data type, even if that means creating a new class to represent it. A `ReadOnlyCollection<Dictionary<string, string>>` would be fine, if that's what it returns. You probably won't be able to directly just cast the return value; you'll likely need to actually introspect into the returned object and actually create the right thing.

dotnet/src/webdriver/Chromium/ChromiumDriver.cs
```diff
@@ -44,6 +44,13 @@ public abstract class ChromiumDriver : WebDriver, ISupportsLogs, IDevTools
         private const string SendChromeCommand = "sendChromeCommand";
         private const string SendChromeCommandWithResult = "sendChromeCommandWithResult";
         protected const string ExecuteCdp = "executeCdpCommand";
+        private const string LaunchAppCommand = "launchAppCommand";
+        private const string Permissions = "setPermission";
+        protected const string GetSinks = "getCastSinks";
```
jimevans: `GetCastSinksCommand`

dotnet/src/webdriver/Chromium/ChromiumDriver.cs
```diff
@@ -44,6 +44,13 @@ public abstract class ChromiumDriver : WebDriver, ISupportsLogs, IDevTools
...
+        protected const string SetSinkToUse = "selectCastSink";
```
jimevans: `SelectCastSinkCommand`

dotnet/src/webdriver/Chromium/ChromiumDriver.cs
```diff
@@ -44,6 +44,13 @@ public abstract class ChromiumDriver : WebDriver, ISupportsLogs, IDevTools
...
+        protected const string StartCastTabMirroring = "startCastTabMirroring";
```
jimevans: `StartCastTabMirroringCommand`

dotnet/src/webdriver/Chromium/ChromiumDriver.cs
```diff
@@ -44,6 +44,13 @@ public abstract class ChromiumDriver : WebDriver, ISupportsLogs, IDevTools
...
+        protected const string GetIssueMessage = "getCastIssueMessage";
```
jimevans: `GetCastIssueMessageCommand`

dotnet/src/webdriver/Chromium/ChromiumDriver.cs
```diff
@@ -44,6 +44,13 @@ public abstract class ChromiumDriver : WebDriver, ISupportsLogs, IDevTools
...
+        protected const string StopCastingToSink = "stopCasting";
```
jimevans: `StopCastingCommand`

dotnet/src/webdriver/Safari/SafariDriver.cs
```diff
@@ -140,6 +145,50 @@ public SafariDriver(SafariDriverService service, SafariOptions options)
...
+            parameters["attachDebugger"] = null;
+            this.Execute(OpenInspector, parameters);
+        }
+
+        /// <summary>
+        /// Set permission of an item on the browser. The only supported permission at this time is "getUserMedia".
+        /// </summary>
+        /// <param name="permission">The name of the item to set permission on.</param>
+        /// <param name="value">Whether the permission has been granted.</param>
+        public void SetPermission(string permission, bool value)
```
jimevans: Call these arguments `permissionName` and `permissionValue`

dotnet/src/webdriver/Chromium/ChromiumDriver.cs
jimevans: Call the parameters `permissionName` and `permissionValue`.

dotnet/src/webdriver/Chromium/ChromiumDriver.cs
```diff
@@ -76,6 +83,8 @@ public ChromiumDriver(ChromiumDriverService service, ChromiumOptions options, Ti
             this.AddCustomChromeCommand(DeleteNetworkConditionsCommand, HttpCommandInfo.DeleteCommand, "/session/{sessionId}/chromium/network_conditions");
             this.AddCustomChromeCommand(SendChromeCommand, HttpCommandInfo.PostCommand, "/session/{sessionId}/chromium/send_command");
             this.AddCustomChromeCommand(SendChromeCommandWithResult, HttpCommandInfo.PostCommand, "/session/{sessionId}/chromium/send_command_and_get_result");
+            this.AddCustomChromeCommand(LaunchAppCommand, HttpCommandInfo.PostCommand, "/session/{sessionId}/chromium/launch_app");
+            this.AddCustomChromeCommand(Permissions, HttpCommandInfo.PostCommand, "/session/{sessionId}/permissions");
```
jimevans: Does Chromium not also provide a "get permissions" API? It might not, but if it does, not including it here seems an oversight.

jimevans (changes requested):

dotnet/src/webdriver/Chromium/ChromiumDriver.cs
titusfortner: This one is at the root level. It's likely part of Permission spec, but Firefox doesn't seem to implement it, and Safari has its own Permission interface, so I thought it would be easiest to leave it in Chromium implementation (where we have it in all the other bindings) for now. Might need to move it to top level everywhere.

dotnet/src/webdriver/Firefox/FirefoxDriver.cs
titusfortner: Aha, there is a good way to parse the enum, nice.

dotnet/src/webdriver/Safari/SafariDriver.cs
titusfortner: This definitely makes the naming less awkward.

dotnet/src/webdriver/Chromium/ChromiumDriver.cs
titusfortner: I don't know how to do that. I'm guessing something like I cast it to a collection of objects and then iterate over / map the collection?

dotnet/src/webdriver/Chromium/ChromiumDriver.cs
titusfortner: Oh, you asked about a getter vs setter here. No, just the setter: https://chromium.googlesource.com/chromium/src/+/main/chrome/test/chromedriver/server/http_handler.cc#859

Java Tests we check permissions with:

```java
    Map<String, Object> result = (Map<String, Object>) ((JavascriptExecutor) driver).executeAsyncScript(
      "callback = arguments[arguments.length - 1];"
      + "callback(navigator.permissions.query({"
      + "name: arguments[0]"
      + "}));", permission);
    return result.get("state").toString();
```

dotnet/src/webdriver/Chromium/ChromiumDriver.cs
jimevans: Go ahead and merge as-is, and I'll take a look, but yes, that's exactly what you'd do.

jimevans (dismissed):

PR #9903, opened 2021-10-12
[py] allow relative locators to take by and value kwargs

titusfortner:
Is this worth the added complexity?


AutomatedTester:
Great question and I don't think it is. I am already concerned that people might not grasp the distances idea already and adding things to the method signature where there is already support will confuse things.

Issue #9166, opened 2021-02-11
Remove Edge HTML


titusfortner:
Since Microsoft is forcing users to upgrade from Legacy Edge to the Chromium backed Edge, we've decided to remove support for it from the Selenium bindings for 4.0

Bindings maintainers can remove the label when completed in their code.

titusfortner:
thank you selenium-assistant

diemol:
Java done here https://github.com/SeleniumHQ/selenium/commit/610b34ab992390da41ef4a170ee21eefc42ef3a7

bwalderman:
Hi @titusfortner. Support for Microsoft Edge Legacy (the browser) is indeed ending in the near future, but EdgeHTML will remain installed on the system and EdgeHTML-based universal Windows applications will continue to be supported.

Would you be open to keeping EdgeHTML support in Selenium 4? We could update the language bindings which have a `UseChromium` flag to `true` by default since Edge Chromium testing will be the most common scenario going forward.

titusfortner:
@bwalderman  Can't you extend everything you need to work with EdgeHTML in your `edge-selenium-tools` project? What would you need in Selenium library to ensure it still functions correctly?

bwalderman:
No new changes are needed in Selenium to ensure both EdgeHTML and Chromium function correctly. The EdgeHTML support should just not be removed.

The EdgeDriver class in dotnet, python, and javascript has a `UseChromium` boolean property that determines whether the driver uses EdgeHTML or Chromium. At the moment, `UseChromium` is false by default for backward compat. I'm proposing instead of removing EdgeHTML support from all languages, we simply flip the property to true by default. This makes Edge Chromium automation easy but leaves EdgeHTML support available for those who need it. I'd be happy to submit a PR.

Regarding `edge-selenium-tools`, This is a polyfill for bringing Edge Chromium support to Selenium 3 because 3 is still the latest stable release. We don't intend to use `edge-selenium-tools` to ship new features and would prefer to contribute directly to Selenium instead. `edge-selenium-tools` was created as a separate project because there was no longer any vehicle for updating the official Selenium 3 with Edge Chromium support.

titusfortner:
[...]

PR #9921, opened 2021-10-14
[java] Remove Edge USE_CHROMIUM flag

bwalderman:
### Description
Removing the `ms:edgeChromium` a.k.a. `USE_CHROMIUM` flag from the Java bindings as this should no longer be used.

### Motivation and Context
The flag was originally introduced as a hint for middleware like Selenium Grid to let the client select either EdgeHTML or Chromium-based Edge. In #9166 the decision was made to remove EdgeHTML support from Selenium 4 and make Edge Chromium the default and only supported Edge version. Since there are no longer plans to support running both Edge implementations side-by-side, this flag can be removed.

### Types of changes

### Checklist


diemol (approved):
This is great, thank you @bwalderman!

PR #9937, opened 2021-10-15
fix some typos, mostly in Java code

asolntsev:
### Description
Just fixed some typos in comments and tests.

### DIFF
--- a/common/src/web/readOnlyPage.html
+++ b/common/src/web/readOnlyPage.html
@@ -4,7 +4,7 @@
 
 <input id="readOnlyTextInput" type="text" readonly value="Test"/>
 
-<input id="textInputnotenabled" type="text" disabled="true" value="Test"/>
+<input id="textInputNotEnabled" type="text" disabled="true" value="Test"/>
 
 <textarea id="writableTextArea" rows="2" cols="20">
 This is a sample text area which is supposed to be cleared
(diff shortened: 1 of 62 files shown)


common/src/web/readOnlyPage.html
```diff
@@ -4,7 +4,7 @@
 
 <input id="readOnlyTextInput" type="text" readonly value="Test"/>
 
-<input id="textInputnotenabled" type="text" disabled="true" value="Test"/>
+<input id="textInputNotEnabled" type="text" disabled="true" value="Test"/>
```
titusfortner, line 7: You missed updating this in .NET test
https://github.com/SeleniumHQ/selenium/blob/64447d4b03f6986337d1ca8d8b6476653570bcc1/dotnet/test/common/ClearTest.cs#L22

common/src/web/readOnlyPage.html
```diff
@@ -13,7 +13,7 @@
 <textarea id="textAreaReadOnly" readonly rows="5" cols="20">
 text area which is not supposed to be cleared</textarea>
 
-<textarea rows="5" id="textAreaNotenabled" disabled="true" cols="20">
+<textarea rows="5" id="textAreaNotEnabled" disabled="true" cols="20">
```
titusfortner, line 16: missed a reference in .NET https://github.com/SeleniumHQ/selenium/blob/64447d4b03f6986337d1ca8d8b6476653570bcc1/dotnet/test/common/ClearTest.cs#L49

titusfortner (changes requested):
Thanks for going through all of this. 

I'm much more comfortable fixing up comments and test names than actual code variables, but I don't see anything obviously wrong other than the two places where you missed updating an html element reference.

Also I don't think we want to make any changes to files in `third_party/`, `thoughtworks/` and the dojo js files, so please revert those.

common/src/web/readOnlyPage.html
asolntsev: Thank you for the correction, @titusfortner !
I really missed .NET part :)

asolntsev:
superseded by https://github.com/SeleniumHQ/selenium/pull/9964

Issue #9938, opened 2021-10-15
[🐛 Bug]: IEDriver crashed when call API FindElementByXPath


guangyuexu:
### What happened?

Build IEDriver from trunk branch. IEDriver crashed when call API FindElementByXPath.
The test was made with IEDriver and Edge in IEMode. The crash also happens when test with IEDriver and IE. 

### How can we reproduce the issue?

```shell
using System;
using OpenQA.Selenium;
using OpenQA.Selenium.IE;

namespace Selenium_Web
{
    class Program
    {
        IWebDriver reqDriver = null;
        static void Main(string[] args)
        {
            //change the path accordingly
            var dir = @"C:\dev\selenium\build\cpp\x64\Debug"; 
            var driverexe = "IEDriverServer.exe";

            var ieService = InternetExplorerDriverService.CreateDefaultService(dir, driverexe);
            var ieOptions = new InternetExplorerOptions { IgnoreZoomLevel = true };
            ieOptions.PageLoadStrategy = PageLoadStrategy.Eager;
            ieOptions.AddAdditionalCapability("ie.edgechromium", true);  // comment this line when test with IE
            //change the path accordingly
            ieOptions.AddAdditionalCapability("ie.edgepath", "C:/Program Files (x86)/Microsoft/Edge/Application/msedge.exe"); // comment this line when test with IE
            
            InternetExplorerDriver reqDriver = new InternetExplorerDriver(ieService, ieOptions);
            reqDriver.Url = "https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_win_opener";
            
            reqDriver.FindElementByXPath("//button[text()='Run ❯']").Click();

            Console.WriteLine("Done");
            Console.ReadLine();
        }
    }
}
```

### Relevant log output

```shell
Started InternetExplorerDriver server (64-bit)
3.150.2.0
Listening on port 58616
Only local connections are allowed
Unhandled exception. OpenQA.Selenium.NoSuchElementException: Unable to find element with xpath == //button[text()='Run ?']
   at OpenQA.Selenium.Remote.RemoteWebDriver.UnpackAndThrowOnError(Response errorResponse)
   at OpenQA.Selenium.Remote.RemoteWebDriver.Execute(String driverCommandToExecute, Dictionary`2 parameters)
   at OpenQA.Selenium.Remote.RemoteWebDriver.FindElement(String mechanism, String value)
[...]

PR #9939, opened 2021-10-15
Fix XPath crash issue in IEDriver

guangyuexu:
Fix IEDriver XPath crash issue

### Description
This PR is to fix XPath crash issue tracked in #9938 

### Motivation and Context
The XPath crash issue in IEDriver was caused by regex update from this commit: https://github.com/SeleniumHQ/selenium/commit/cf49ba23. The IEDriver crashed when IE (or Edge in IEMode) tried to parse the unicode regex ending with `/gu`. 

### Types of changes
In this PR, a browser type check is added to let IE (or Edge in IEMode) use `/g` regex and other browsers use `/gu` unicode regex in the file.

In addition, this PR generated the file atoms.h to reflect the changes in javascript/atom and updated the file for IEDriver.


jimevans (approved):
LGTM.

PR #9950, opened 2021-10-18
IEDriver NewWindow3 API workaround solution for Edge IEMode

guangyuexu:
IEDriver NewWindow3 API workaround solution for Edge IEMode

### Description
This PR is to provide a workaround solution for IEDriver to track new created windows in Edge IEMode. 

### Motivation and Context
When IEDriver drives IE browser, there is a [NewWindow3 ](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/aa768288(v=vs.85)) API to help create new window and keep these new created window on track. 

When IEDriver drives Edge browser in IEMode, there is no such NewWindow3 API in Edge Browser. This PR is to create a workaround solution to help IEDriver track new created windows in Edge IEMode without using the NewWindow3 API. 

### Types of changes
Note that when Edge runs in IEMode, all IE windows are wrapped by an Edge window. That means, every IE window is a child window of a Edge window. It is possible to keep track of all IE window handles before and after a new window created. By comparing the window handles before after a new window event, the new window handles created from client side JS codes can be identified and tracked in IEDriver.


cpp/iedriver/Browser.cpp
```diff
@@ -108,18 +108,78 @@ void __stdcall Browser::NewProcess(DWORD lCauseFlag,
   this->InitiateBrowserReattach();
 }
 
+BOOL CALLBACK FindIEBrowserHandles(HWND hwnd, LPARAM arg) {
```
jimevans: Seems like these methods would be better located with their similar ones in `BrowserFactory.cpp`, no? I mean, I'm certainly not wedded to that, and we can always refactor locations later, I suppose.

cpp/iedriver/Browser.cpp
```diff
@@ -108,18 +108,78 @@ void __stdcall Browser::NewProcess(DWORD lCauseFlag,
...
                                    BSTR bstrUrlContext,
                                    BSTR bstrUrl) {
+  LOG(TRACE) << "Entering Browser::NewWindow3";
+
+
   if (this->is_edge_chromium_) {
-    LOG(TRACE) << "Entering Browser::NewWindow3 but early exiting due to edge mode";
-    // In Edge Chromium, we do not yet support attaching to new windows.
-    // Quit early and ignore that event.
+    ::PostMessage(this->executor_handle(), WD_BEFORE_NEW_WINDOW, NULL, NULL);
```
jimevans: Seems like both `PostMessage` calls should be outside the `if` and `else` blocks here, along with `std::vector<HWND>* ie_window_handles = NULL;`, and setting `ie_window_handles` to the vector in the `if` case. But I'm being nitpicky, and I could be wrong.

cpp/iedriver/Browser.cpp
```diff
@@ -108,18 +108,78 @@ void __stdcall Browser::NewProcess(DWORD lCauseFlag,
...
+
+    HWND top_level_handle = this->GetTopLevelWindowHandle();
+    // 1) find all Edge browser window handles
+    std::vector<HWND>edge_window_handles;
+    ::EnumWindows(&FindEdgeBrowserHandles,
+                  reinterpret_cast<LPARAM>(&edge_window_handles));
+
+    // 2) find all IE browser window handlers as child window when Edge runs in IEMode
+    std::vector<HWND>* ie_window_handles = new std::vector<HWND>;
+    for (auto& ewh : edge_window_handles) {
```
jimevans: I dislike the use of `auto` here, if only because it's inconsistent with every other part of the code base. I know it's syntactic sugar that makes the creation of an iterator unnecessary, and if you push back on its use, I'm sure I'll cave, but consistency is a valid consideration.

cpp/iedriver/IECommandExecutor.cpp
```diff
@@ -287,6 +290,73 @@ LRESULT IECommandExecutor::OnAfterNewWindow(UINT uMsg,
...
                                        WD_AFTER_NEW_WINDOW);
+    if (lParam > 0) {
+      // a new window is created from Edge in IEMode
+      BrowserHandle browser_wrapper;
+      this->GetCurrentBrowser(&browser_wrapper);
+      HWND top_level_handle = browser_wrapper->GetTopLevelWindowHandle();
+
+      std::vector<HWND>* current_window_handles =
+        reinterpret_cast<std::vector<HWND>*>(lParam);
+      std::unordered_set<HWND> current_window_set(
```
jimevans, line 299: Any particular reason for using `unordered_set` here, instead of just another vector? Not a criticism, just curious. I mean, I know the semantics are slightly different, but

cpp/iedriver/IECommandExecutor.cpp
```diff
@@ -287,6 +290,73 @@ LRESULT IECommandExecutor::OnAfterNewWindow(UINT uMsg,
...
+      for (auto& window_handle : new_ie_window_handles) {
+        if (current_window_set.find(window_handle) != current_window_set.end())
+          continue;
+        diff.push_back(window_handle);
+      }
+
+      if (diff.size() == 0) {
+        LOG(WARN) << "No new window handle found after attempt to open";
+      } else {
+        HWND new_window_window = diff[0];
```
jimevans, line 332: This code all looks right for attaching to a browser instance. Nicely done.

jimevans (changes requested):

cpp/iedriver/BrowserFactory.cpp
```diff
@@ -350,7 +350,6 @@ void BrowserFactory::LaunchEdgeInIEMode(PROCESS_INFORMATION* proc_info,
   }
 
   // These flags force Edge into a mode where it will only run MSHTML
-  executable_and_url.append(L" --ie-mode-force");
```
bwalderman: Why is this removed?

cpp/iedriver/Browser.cpp
bwalderman: I think using `auto` and/or a reference `&` would be debatable if the elements were a more complex type, but HWNDs are just integers, so I think we can just use `HWND` here (and in the other places in this change where HWNDs are being iterated).

bwalderman (commented):
LGTM with some suggestions.

cpp/iedriver/Browser.cpp
guangyuexu: Moved `FindIEBrowserHandles` and `FindEdgeBrowserHandles` functions to `BrowserFactory` as static methods. I also removed the similar function in class `IECommandExecutor`.

cpp/iedriver/Browser.cpp
guangyuexu: Move the two `PostMessage` calls to the outside of the `if` and `else` blocks.

cpp/iedriver/Browser.cpp
guangyuexu: Replaced `auto` with `HWND` in the changes.

cpp/iedriver/IECommandExecutor.cpp
guangyuexu: This is for performance consideration. In case we have lots of window handles, it will be faster using `unordered_set` than `vector` to find out a new window handle from a large set of window handles.

cpp/iedriver/BrowserFactory.cpp
guangyuexu: Brought `--ie-mode-force` switch back in this [commit](https://github.com/SeleniumHQ/selenium/pull/9950/commits/a2882488c2a296a3fd3fdc10636e33001315ea80).

jimevans (approved):
LGTM. Will merge ASAP.

PR #9953, opened 2021-10-19
Make `SearchContext.findElements` have a generic return type

shs96c:
Why do this? The main reason is to support frameworks that wish to
extend Selenium, but return their own sub-types of `WebElement` from
`findBy`.

`findElement` can already be overridden using co-variant return types,
but the `List<WebElement>` returned by `findElements` could not be
overridden as easily. This signature change allows that.

With this change in place, subclasses of WebDriver can now do
something like this, without there being compilation issues:

```
public interface Special extends WebDriver {
    SpecialElement findElement(By locator);
    List<Specialelement> findElements(By locator);
}
```

Compiling the entire java code base suggests that this is a safe
change to make -- no tests or other calls sites needed to be modified
in order for this change to work.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/SearchContext.java
+++ b/java/src/org/openqa/selenium/SearchContext.java
@@ -27,7 +27,7 @@ public interface SearchContext {
    * @return A list of all {@link WebElement}s, or an empty list if nothing matches
    * @see org.openqa.selenium.By
    */
-  List<WebElement> findElements(By by);
+  <T extends WebElement> List<T> findElements(By by);
 
 
   /**
--- a/java/src/org/openqa/selenium/WebDriver.java
+++ b/java/src/org/openqa/selenium/WebDriver.java
@@ -105,7 +105,7 @@ public interface WebDriver extends SearchContext {
    * @see org.openqa.selenium.WebDriver.Timeouts
    */
   @Override
-  List<WebElement> findElements(By by);
+  <T extends WebElement> List<T> findElements(By by);
 
 
   /**
--- a/java/src/org/openqa/selenium/By.java
+++ b/java/src/org/openqa/selenium/By.java
@@ -132,7 +132,7 @@ public abstract class By {
    * @param context A context to use to find the elements.
    * @return A list of WebElements matching the selector.
    */
-  public abstract List<WebElement> findElements(SearchContext context);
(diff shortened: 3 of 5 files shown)


java/src/org/openqa/selenium/SearchContext.java
```diff
@@ -27,7 +27,7 @@
    * @return A list of all {@link WebElement}s, or an empty list if nothing matches
    * @see org.openqa.selenium.By
    */
-  List<WebElement> findElements(By by);
+  <T extends WebElement> List<T> findElements(By by);
```
mykola-mokhnach, line 30: should a similar change be done for findElement?

mykola-mokhnach (dismissed):

titusfortner (dismissed):
If we don't see any backward compatibility issues, then we should merge it. The sooner we release this, the sooner the appium devs can use it for their next release.

diemol (dismissed):

java/src/org/openqa/selenium/WebDriver.java
```diff
@@ -105,7 +105,7 @@
    * @see org.openqa.selenium.WebDriver.Timeouts
    */
   @Override
-  List<WebElement> findElements(By by);
+  <T extends WebElement> List<T> findElements(By by);
```
mykola-mokhnach, line 108: This change should also probably go into the corresponding method of the abstract By class?

java/src/org/openqa/selenium/SearchContext.java
shs96c: No, because we can rely on co-variant return types there. The only signature that needs to change is `findElements`, and that's because we want to return a typed list, and `List<WebElement>` is not the same as `List<? extends WebElement>` in generics.

java/src/org/openqa/selenium/WebDriver.java
shs96c: Good point. I'll do that too.

shs96c:
While this change “only” causes some warnings in Java (which can be easily suppressed), this change will break Kotlin code, and that’s far from ideal. 

Perhaps we need to hold off on this until we come up with a way to make this work smoothly for Kotlin users.

diemol:
After chatting with @mykola-mokhnach and checking the linked PR, the Appium team does not need this anymore.

Issue #10045, opened 2021-11-16
[🐛 Bug]: uploadFile not working with docker images exposing chromedriver directly


tyge68:
### What happened?

While trying to switch to latest selenium api 4.0, we hit the issue with the uploadFile.
Generally if we use selenium chrome standalone docker image there are no issues, as that one uses selenium server internally.

But if you use other distributions (like selenoide/vnc:chrome_95.0) which allow to easily do video recording of the tests, the uploadFile will fails as it doesn't know the command with .../se/file since that is not implemented in the chromedriver out of the box, instead it should use .../file which is implemented.

As both should be possible, I suppose we need a fallback mechanism to allow this to work, or eventually a mechanism to detect or turn it one based on environment variable / system properties.

### How can we reproduce the issue?

```shell
- Assuming it's a java based project
- Assuming it enable LocalFileDetector in your driver with following: ((RemoteWebDriver) driver).setFileDetector(new LocalFileDetector());
- Assuming the page have a input of type "file"
- On your selenium element, sendKeys(...) with corresponding filepath in the input element of type "file"

observe the error in the logs (see below)

This is due to the current code (1) which force to use .../se/file for UPLOAD_FILE operation, while chromedriver implement it at .../file already.
(1) https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/remote/codec/w3c/W3CHttpCommandCodec.java#L161
```

### Relevant log output

```shell
INFO: Detected dialect: W3C

org.openqa.selenium.UnsupportedCommandException: unknown command: unknown command: session/c6d981124b6ec773f82df7c6eb73da7d/se/file
```

### Operating System

Linux, Mac

### Selenium version

4.0.0

### What are the browser(s) and version(s) where you see this issue?

Chrome 95, Chrome 96

### What are the browser driver(s) and version(s) where you see this issue?

95.0.4638.69, 96.0.4664.35

AutomatedTester:
The problem here is that you're using `RemoteWebDriver` which then assumes that there is a Selenium grid, potentially on a different machine, that it is connecting to. Once we make that assumption we use it for deciding when to upload files or not and how to do it. If it is in a docker container you can always upload the file as part of the step of building the container or you can add a selenium server that you connect. 

This is why Selenium docker containers have the server on them.

AutomatedTester:
[...]

PR #10046, opened 2021-11-16
Issue #10045  uploadFile not working with docker images exposing chromedriver directly

tyge68:

### Description
Updating to allow switching the uploadFile command using system property selenium.uploadFileMode
the default will still use the current solution to use the /se/file endpoint, while other value  would allow to use different implementation "chromedriver" mode for example.

### Motivation and Context
While updating to latest selenium 4.0 we encountered issues with the uploadFile command using docker image (selenoid/vnc:chrome_95.0) which used to work very well in the past. Based on the history of the uploadFile which was first set to use /file and finally again to use /se/file , it appears that it would make sense to allow to choose which endpoint to use for that command rather than to force one implementation.

### Types of changes

### Checklist

- [] All new and existing tests passed.

### DIFF
--- a/common/src/web/webComponents.html
+++ b/common/src/web/webComponents.html
@@ -1,0 +1,19 @@
+<style>
+    custom-checkbox-element {
+        display: block;
+        width: 20px;
+        height: 20px;
+    }
+</style>
+<custom-checkbox-element></custom-checkbox-element>
+<script>
+    customElements.define('custom-checkbox-element',
+        class extends HTMLElement {
+            constructor() {
+                super();
+                this.attachShadow({ mode: 'open' }).innerHTML = `
+                                <div><input type="checkbox"/></div>
+                            `;
+            }
+        });
+</script>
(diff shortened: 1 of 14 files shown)


jimevans:
I don’t use the docker images in my day-to-day work, but I am -1 on this change. The `/file` end point is not compliant with the W3C WebDriver Specification; the `/se/file` end point is. We should be aiming to be spec-compliant everywhere we can.

tyge68:
@jimevans so you mean Google / Chrome should provide a Chromedriver that should expose the upload file endpoint on /se/file ? similar to other endpoint that are already under /se/  , ie /se/logs etc.. ? who can make this happen ?

titusfortner:
I agree with Jim, Selenium should only be using w3c compliant endpoints. 

We are expecting that client code will only act directly with drivers when on the same machine, and that the user will only send commands to a driver on a remote machine via a server. 

An alternative to adding a Selenium server to your container would be to get the W3C spec to add an official `upload` endpoint that accepts a BASE64 zipped file as the value. Looks like an upload endpoint was agreed to (https://github.com/w3c/webdriver/issues/1355), and it might make sense for it to support the file itself rather than just a String representing the path (which would avoid the whole Local File Detector requirements).

tyge68:
Thanks, actually I found a simpler solution, there is no need actually for the ../se/file etc.. implementations since nowaday you can easily copy files to your docker too, before send keys to the input. Thus I guess it's the reason it won't be part of the w3c too.

titusfortner:
Oh, right, yes. I was going to include that as an option and got distracted and forgot. Make sure when you do that, though, that the path is different from client machine and / or that a Local File Detector is not set.

Issue #9784, opened 2021-09-02
NoClassDefFoundError after bumping to selenium-java 4.0.0 in Mac OS


bonigarcia:
## 🐛 Bug Report

I have a couple of repositories using Selenium 4:

* https://github.com/bonigarcia/selenium-webdriver-java/
* https://github.com/bonigarcia/webdrivermanager/

I use GitHub Actions to run the tests in these repositories, using three different [virtual environments](https://github.com/actions/virtual-environments):

* `ubuntu-latest` (Ubuntu 20.04).
* `windows-latest` (Windows Server 2019).
* `macos-latest` (macOS 10.15).

After bumping to selenium-java 4.0.0-rc-1, several tests of the Mac OS workflow are failing due to `NoClassDefFoundError`, such as:

```
java.lang.NoClassDefFoundError: Could not initialize class org.openqa.selenium.WebDriverException
	at org.openqa.selenium.remote.RemoteWebDriver.execute(RemoteWebDriver.java:571)
```

... or:

```
java.lang.NoClassDefFoundError: Could not initialize class org.openqa.selenium.net.HostIdentifier
	at io.github.bonigarcia.wdm.test.create.ChromeRemoteTest.setupClass(ChromeRemoteTest.java:41)
```

## To Reproduce/Test script

I don't have a Mac OS machine to test it, but the error seems consistent in Mac OS (at least on GitHub Actions).

The test that fails are the following:

* 10 tests in selenium-webdriver-java. For instance: [FluentWaitJUnit4Test](https://github.com/bonigarcia/selenium-webdriver-java/blob/master/selenium-webdriver-junit4/src/test/java/io/github/bonigarcia/webdriver/junit4/ch3/wait/FluentWaitJUnit4Test.java) (see the complete [workflow traces](https://github.com/bonigarcia/selenium-webdriver-java/actions/runs/1193944103) and the [workflow setup](https://github.com/bonigarcia/selenium-webdriver-java/blob/master/.github/workflows/build.yml)).
* 1 test in WebDriverManager: [ChromeRemoteTest](https://github.com/bonigarcia/webdrivermanager/blob/master/src/test/java/io/github/bonigarcia/wdm/test/create/ChromeRemoteTest.java) (see [workflow traces](https://github.com/bonigarcia/webdrivermanager/actions/runs/1193898533) and the [workflow setup](https://github.com/bonigarcia/webdrivermanager/blob/master/.github/workflows/build.yml)).

## Environment

[...]

PR #10063, opened 2021-11-19
#9784 use `shutdown()` instead of `shutdownNow()`

asolntsev:
it waits for completion of already started tasks. Otherwise it may causes errors in those tasks, including succeeding NoClassDefFoundError.

### Description

One more fix for https://github.com/SeleniumHQ/selenium/issues/9784
### Types of changes


java/src/org/openqa/selenium/remote/service/DriverCommandExecutor.java
```diff
@@ -131,7 +131,7 @@ public Response execute(Command command) throws IOException {
         Thread.currentThread().interrupt();
         throw new WebDriverException("Timed out waiting for driver server to stop.", e);
       } finally {
-        executorService.shutdownNow();
+        executorService.shutdown();
```
pujagani: For the command executor, which in turn, uses the HTTP client to talk to the WebDriver, the client might have high-timeout values set, so the shutdown can take a long time if we wait for it to complete, especially if multiple-long running threads are there.  I think it might be a good idea in general to couple the shutdown() with `awaitTermination()` with a timeout, if the ongoing tasks do not complete within that timeout then call `shutdownNow()`. What do you think?

diemol (approved):

java/src/org/openqa/selenium/remote/service/DriverCommandExecutor.java
asolntsev: @pujagani Yes, I have added `awaitTermination()` right after `shutdown()`. Thank you for the idea.

pujagani:
Thank you! @asolntsev

Issue #10066, opened 2021-11-20
[🐛 Bug]: Selenium tests sometimes randomly fail because of port already in use


asolntsev:
### What happened?

I submitted [another PR](https://github.com/SeleniumHQ/selenium/pull/10063) to Selenium, and [the build failed](https://github.com/SeleniumHQ/selenium/runs/4270614439?check_suite_focus=true) because of a flaky test.

In fact, my own integration tests also fail time-to-time, see [issue 9972](https://github.com/SeleniumHQ/selenium/issues/9972). 

# Proposed solution
I want to retry starting netty server several times if it fails due to "port already in use" exception.

### How can we reproduce the issue?

```shell
It's hard to reproduce.
```

### Relevant log output

```shell
There was 1 failure:
1) testToggleAirplaneMode(org.openqa.selenium.mobile.NetworkConnectionTest)
java.lang.RuntimeException: java.io.UncheckedIOException: java.io.IOException: Port 41538 already in use
	at org.openqa.selenium.environment.GlobalTestEnvironment.getOrCreate(GlobalTestEnvironment.java:44)
	at org.openqa.selenium.testing.JUnit4TestBase.prepareEnvironment(JUnit4TestBase.java:54)
	... 32 trimmed
Caused by: java.io.UncheckedIOException: java.io.IOException: Port 41538 already in use
	at org.openqa.selenium.netty.server.NettyServer.start(NettyServer.java:159)
	at org.openqa.selenium.netty.server.NettyServer.start(NettyServer.java:52)
	at org.openqa.selenium.environment.webserver.NettyAppServer.start(NettyAppServer.java:104)
	at org.openqa.selenium.environment.InProcessTestEnvironment.<init>(InProcessTestEnvironment.java:29)
	at org.openqa.selenium.environment.GlobalTestEnvironment.getOrCreate(GlobalTestEnvironment.java:41)
	... 34 more
Caused by: java.io.IOException: Port 41538 already in use
	... 39 more
Caused by: java.net.BindException: Address already in use
	at java.base/sun.nio.ch.Net.bind0(Native Method)
	at java.base/sun.nio.ch.Net.bind(Net.java:455)
	at java.base/sun.nio.ch.Net.bind(Net.java:447)
	at java.base/sun.nio.ch.ServerSocketChannelImpl.bind(ServerSocketChannelImpl.java:227)
[...]

Issue #9972, opened 2021-10-22
[🐛 Bug]: "Port 53836 already in use" from embedded Grid


asolntsev:
### What happened?

We start an embedded Grid server in the beginning of every test:

```java
    int hubPort = org.openqa.selenium.net.PortProber.findFreePort();
    org.openqa.selenium.grid.Main.main(new String[]{"standalone", "--port", String.valueOf(hubPort)});
```

After upgrading to Selenium 4, it started failing **time-to-time**:
```
IOException: Port 53836 already in use
```

### How can we reproduce the issue?

The test suite that fails occasionally:
https://github.com/selenide/selenide/tree/master/modules/grid/src/test/java/integration

Not sure how to reproduce it stably.

### Relevant log output

```shell
CustomWebdriverProviderWithRemoteBrowser > customWebdriverProviderCanUseRemoteWebdriver() FAILED
    java.io.UncheckedIOException: java.io.IOException: Port 53836 already in use
        at org.openqa.selenium.netty.server.NettyServer.start(NettyServer.java:156)
        at org.openqa.selenium.netty.server.NettyServer.start(NettyServer.java:51)
        at org.openqa.selenium.grid.commands.Standalone.execute(Standalone.java:214)
        at org.openqa.selenium.grid.TemplateGridCommand.lambda$configure$4(TemplateGridCommand.java:129)
        at org.openqa.selenium.grid.Main.launch(Main.java:83)
        at org.openqa.selenium.grid.Main.go(Main.java:57)
        at org.openqa.selenium.grid.Main.main(Main.java:42)
        at integration.AbstractGridTest.setUpGrid(AbstractGridTest.java:30)
```

See for example https://github.com/selenide/selenide/runs/3969071951?check_suite_focus=true

### Operating System

Linux

### Selenium version

4.0.0

### What are the browser(s) and version(s) where you see this issue?

Chrome 94

### What are the browser driver(s) and version(s) where you see this issue?

chromedriver 94.0.4606.61

### Are you using Selenium Grid?

4.0.0

AutomatedTester:
The way that we poll for free ports is always going to be a race since we ask the OS for free port (try binding to 0) and then see what it gives us. We close it down and then get selenium to connect to that. In that time, another test or application could bind there.

diemol:
[...]

PR #10067, opened 2021-11-20
Throw specific exception when port is already in use

asolntsev:
### Description
This PR provides a fix for issues #9972 and #10066

NB! This PR will also solves some of **Selenium own flaky tests**, for example [caused by "BindException: Address already in use"](https://github.com/asolntsev/selenium/runs/4412949195?check_suite_focus=true).

### Types of changes

### Checklist


asolntsev:
@diemol This PR will solves Selenium own flaky tests. Would be great to merge it.

diemol:
I was trying this approach but it does not solve the problem, because the `NettyAppServer` is used in several places. Retrying here is not enough. We need to rethink how the server can be started for during tests.

asolntsev:
@diemol Yes, it's not a universal approach, and it cannot be (as much as I understand). 
I only added re-try to the test that I saw failing multiple times. 

But the most important is that this PR allows me (and other users) to add try-catch **to their tests**. This is what I really need.

diemol:
I ended up merging #10170, which retries for all cases when the exception is thrown. Thank you for the thought process.

PR #10072, opened 2021-11-22
Fix wrong type annotations for "env" it's a dict not a string

jedie:
env used here: https://github.com/SeleniumHQ/selenium/blob/adeca5c45a9c5c666cb0d81952fff0964d82f24d/py/selenium/webdriver/common/service.py#L48 and passed to subprocess here: https://github.com/SeleniumHQ/selenium/blob/adeca5c45a9c5c666cb0d81952fff0964d82f24d/py/selenium/webdriver/common/service.py#L71


AutomatedTester (approved):

AutomatedTester:
Thanks for the PR!

PR #10092, opened 2021-11-26
[py] send Base64 encoded addon to driver instead of path String

titusfortner:
Uses `addon` parameter instead of `path` parameter for installing Firefox AddOn

reference:
https://hg.mozilla.org/mozilla-central/file/tip/testing/geckodriver/CHANGES.md#l878

This matches current Ruby & .NET behavior.

This change will also make it easier to use in a remote driver when we get that functionality mixed in.

@AutomatedTester please let me know if my Python is correct here; like there might be a better way to do a relative file reference?

### DIFF
--- a/py/selenium/webdriver/firefox/webdriver.py
+++ b/py/selenium/webdriver/firefox/webdriver.py
@@ -14,8 +14,7 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-
-from base64 import b64decode
+import base64
 from shutil import rmtree
 import warnings
 from contextlib import contextmanager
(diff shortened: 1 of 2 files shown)


py/selenium/webdriver/firefox/webdriver.py
```diff
@@ -247,7 +246,11 @@ def install_addon(self, path, temporary=None) -> str:
 
                 driver.install_addon('/path/to/firebug.xpi')
         """
-        payload = {"path": path}
+        file_ = open(path, 'rb')
```
AutomatedTester: This would be better as a context manager

```python
with open(path, rb) as file:
     # the encoding part
```

Context managers handle the opening and closing for you in a much better way and will be GC'ed better

PR #10093, opened 2021-11-27
[dotnet] support installing temporary addon in Firefox

### LINKED ISSUE #10084: [🚀 Feature]: Java/CS FirefoxProfile to support addExtensioin temporary
### Feature and motivation

Add to `FirefoxProfile::addExtension` option to add extensions as temporary

Other languages (Python, Ruby and Javascript) support providing config of adding extension as temporary 
For some reason Java/C# doesn't provide config, just provides false, or nothing
https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/firefox/AddHasExtensions.java#L105
https://github.com/SeleniumHQ/selenium/blob/trunk/dotnet/src/webdriver/Firefox/FirefoxDriver.cs#L299

### Usage example

When testing on release (non developer/nightly), an 'unsigned' xpi won't load if it isn't loaded as temporary 
- where localy signed doesn't count as signed.

On windows, when unpacking the xpi to load the folder, can lead to failures because of windows adding content like thumbs.db

```
		FirefoxProfile firefoxProfile = new FirefoxProfile();
		firefoxProfile.addExtension(a, true);
```

titusfortner:
This fixes #10084 

Jim, I think this is correct, but let me know if I missed something. Thanks

### DIFF
--- a/dotnet/src/webdriver/Firefox/FirefoxDriver.cs
+++ b/dotnet/src/webdriver/Firefox/FirefoxDriver.cs
@@ -262,6 +262,16 @@ namespace OpenQA.Selenium.Firefox
         /// <param name="addOnFileToInstall">Full path and file name of the add-on to install.</param>
         public void InstallAddOnFromFile(string addOnFileToInstall)
         {
+            InstallAddOnFromFile(addOnFileToInstall, false);
+        }
+
+        /// <summary>
+        /// Installs a Firefox add-on from a file, typically a .xpi file.
+        /// </summary>
+        /// <param name="addOnFileToInstall">Full path and file name of the add-on to install.</param>
+        /// <param name="temporary">Whether the add-on is temporary.</param>
+        public void InstallAddOnFromFile(string addOnFileToInstall, bool temporary)
+        {
             if (string.IsNullOrEmpty(addOnFileToInstall))
             {
                 throw new ArgumentNullException(nameof(addOnFileToInstall), "Add-on file name must not be null or the empty string");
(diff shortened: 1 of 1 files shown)


dotnet/src/webdriver/Firefox/FirefoxDriver.cs
```diff
@@ -280,14 +290,24 @@ public void InstallAddOnFromFile(string addOnFileToInstall)
             // instead.
             byte[] addOnBytes = File.ReadAllBytes(addOnFileToInstall);
             string base64AddOn = Convert.ToBase64String(addOnBytes);
-            this.InstallAddOn(base64AddOn);
+            this.InstallAddOn(base64AddOn, temporary);
```
BertLamb, line 293: In reading https://github.com/SeleniumHQ/selenium/issues/8357#issuecomment-643047209 it sounds like geckodriver supports local "unpacked" extensions (if the temporary flag is set) and in that case we'd need to use the "path" parametered api in that case.   Ideally we'd check to see if the path is a directory and if so, use that api, otherwise base64 up the file like we are doing now and use the AddOn api.

titusfortner: The API will be a method that takes a path, and a method that takes a path and a temporary value. The implementation of it doesn't matter to the end user.

BertLamb: I agree with you that it will all work fine so long as `path` is the path to a .xpi or .zip but not when it is the path to an unpacked extension directory, which it would also be nice to support.

titusfortner: We'll do everything with addon keyword instead of path, so it will work the way you need. It should already work properly with .NET in 4.1, just without the temporary toggle.

titusfortner:
merged this instead 1b2093826feab66088e38c86d67f99ea6a98b5b7

PR #10138, opened 2021-12-13
[py] update ShadowRoot find element signatures to match those in WebElement

titusfortner:
@AutomatedTester just making sure there wasn't a reason the signatures were different.

### DIFF
--- a/py/selenium/webdriver/remote/shadowroot.py
+++ b/py/selenium/webdriver/remote/shadowroot.py
@@ -18,6 +18,7 @@
 from hashlib import md5 as md5_hash
 
 from .command import Command
+from ..common.by import By
 
 
 class ShadowRoot():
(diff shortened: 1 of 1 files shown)


AutomatedTester:
This is a zero value change. A more useful addition would be adding type hinting.

AutomatedTester (approved):

diemol:
Why do we merge it if it is a zero value change?

AutomatedTester:
I will finish off the patch with the type hinting

titusfortner:
This is not a zero value change. It was literally a feature request from a user who wanted to have the same method wrapper for both methods and couldn't. I only PR'd it instead of merging it because I thought there might have been a reason you didn't just copy/paste what you already had.

If the WebElement methods had type hinting, I would have copied that over as well.

PR #10172, opened 2021-12-21
Reflect how find_elements returns a list, not just one WebElement

colons:
### Description (and motivation and context)
The author of python/typeshed#6565 seems to have been misled by the return types changed here, and it caused mypy to make a lot of complaints in a codebase I work on. Here, like in python/typeshed#6645, those annotations are corrected to reflect that any method with `find_elements` in the name returns a list.

### Types of changes

### Checklist

### DIFF
--- a/py/selenium/webdriver/remote/webdriver.py
+++ b/py/selenium/webdriver/remote/webdriver.py
@@ -472,7 +472,7 @@ class WebDriver(BaseWebDriver):
         )
         return self.find_element(by=By.ID, value=id_)
 
-    def find_elements_by_id(self, id_) -> WebElement:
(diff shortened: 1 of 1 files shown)


colons:
problems like this (and #10120) would be _much_ easier to spot if you disabled a bunch of the warnings in your `mypy.ini`; right now, outright errors are difficult to spot because they're drowned in complaints about stuff that's just not annotated at all

this concerns me because, in the mean time, these incorrect annotations have caused problems in a codebase i work on

i _strongly_ recommend you follow [mypy's advice about how to migrate existing codebases](https://mypy.readthedocs.io/en/stable/existing_code.html) rather than turning on every message right from the start

titusfortner (approved):

PR #10173, opened 2021-12-21
Fix assignment to ZENKAKU_HANKAKU.

colons:
This was an annotation rather than an assignment. As well as not working as an accessible property, flake8 really didn't like it. this is why the flake8 check in #10172 and every other recent commit from trunk has failed the flake8 check.

### Types of changes

### Checklist


titusfortner:
ah, yeah, I caught this doing something else last night before seeing your PR or I would have merged yours. Thanks for your contribution!

PR #10187, opened 2021-12-26
RemoteWebDriverBuilder for local drivers

### LINKED ISSUE #10064: [🐛 Bug]: RemoteWebDriverBuilder does not respect config with local driver
### What happened?

When using `RemoteWebDriverBuilder.config()` to set a `ClientConfig` instance that does not contain a `baseUri` and does not set an address with `address()` method, the `build()` method will get a non null return from `this.getLocalDriver()`, but that method does not create a driver using the provided `ClientConfig` values.

I don't actually see a way to set a `ClientConfig` for local driver sessions, which, if that's the case, is a bigger issue to address. For this I think we should at least throw an exception that the config instance provided can not be respected if we get a local driver result.

```shell
FirefoxOptions browserOptions = new FirefoxOptions();
        ClientConfig config = ClientConfig.defaultConfig()
                .connectionTimeout(Duration.ofMinutes(5))
                .readTimeout(Duration.ofMinutes(4));

        WebDriver driver = RemoteWebDriver.builder()
                .oneOf(browserOptions)
                .config(config)
                .build();
```

### Relevant log output

```shell
n/a
```

### Operating System

n/a

### Selenium version

Java 4.0.0

n/a

n/a

n/a

titusfortner:
So, this would fix #10064 but in a slightly hacky way. Since we can't create a null config, and we can't determine if a driver is local until `build()` is called, this creates a config with a zero readTimeout. Since no one should be passing in a config with a zero read timeout, anything they pass in will override the one that the local driver class put there, and will throw an error in `build()`.

It also seemed weird to have to do this:
```
ChromeDriver.builder().oneOf(new ChromeOptions()).build();
```

So now this will do the expected thing:
```
ChromeDriver.builder().build();
```

### DIFF
--- a/java/src/org/openqa/selenium/chrome/ChromeDriver.java
+++ b/java/src/org/openqa/selenium/chrome/ChromeDriver.java
@@ -18,12 +18,14 @@
 package org.openqa.selenium.chrome;
 
 import com.google.common.collect.ImmutableMap;
+import org.openqa.selenium.Beta;
 import org.openqa.selenium.Capabilities;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.chromium.ChromiumDriver;
 import org.openqa.selenium.chromium.ChromiumDriverCommandExecutor;
 import org.openqa.selenium.remote.CommandInfo;
 import org.openqa.selenium.remote.RemoteWebDriver;
+import org.openqa.selenium.remote.RemoteWebDriverBuilder;
 import org.openqa.selenium.remote.service.DriverService;
 
 import java.util.Map;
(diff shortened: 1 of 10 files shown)


pujagani:
Thank you Titus for looking into this and helping find a solution. This is a tricky one for sure. Supporting the client config with the local driver will be a big change. Meanwhile, if we are looking to throw an error (as intended in the PR) or log a warning when using client config with a local driver, how about we use a flag? This is just a suggestion. We can set the flag when the client config setter is called and use the same flag to throw errors or log warnings in the `getLocalDriver()` method based on if the result is null or not. The suggestion is also hacky but we can pick the less hacky one for the time being. What do you think? Apologies, if I missed out on an edge case.

titusfortner:
@pujagani I removed my ReadTimeout hack. Would you like to add the code you wrote for the flag you showed me, or would you like me to do it?

pujagani:
Thank you for considering the solution. @titusfortner  I can go ahead and add it in a separate pull request if that is okay.

titusfortner:
#10219 will be necessary to get these tests passing

Issue #9996, opened 2021-10-28
[🚀 Feature]: Support advanced properties defined in WebDriver specification for pointer actions


jimevans:
### Feature and motivation

The W3C WebDriver Specification defines several [properties for pointer actions](https://w3c.github.io/webdriver/#pointer-actions) that the Selenium language bindings do not yet allow users to access. These properties are:
* width
* height
* pressure
* tangentialPressure
* tiltX
* tiltY
* twist
* altitudeAngle
* azimuthAngle
Each set of language bindings should be capable of sending these properties of a pointer action across the wire as part of the JSON payload for the actions command.

### Usage example

Actions helper classes do not expose this type of functionality at this time, but builders of actions should allow such properties to be used.

titusfortner:
Ruby - fe3000580360abbb182231ac58ebf9e465c8464d
Python - 965e5282c7dfd21338d1a7d60b1ebdad4af5b293

diemol:
Java - https://github.com/SeleniumHQ/selenium/commit/70e54e6ada252ed1d2bce073aaec73d60f9d1f36

AutomatedTester:
JS - 5b67ee199968220815d9a1ef72249974d69c68a1

AutomatedTester:
dotnet - e6cd2a8df18d4dc79bce140d9824be312678154a

titusfortner:
Dotnet ones here are not hooked up. Needs the PR. Just went over how to fix the issues with it.

titusfortner:
Dotnet implemented with a9c3770a6c4604f2897b28441bf00889f5faffa5

PR #10198, opened 2021-12-30
[java] implement event support for pointer device actions

titusfortner:
A Java implementation for #9996 

This doesn't feel like a very "Selenium" way to implement this functionality. It's pretty much just private properties with chained setters. No getters because we're expecting it to be used inline.

So, it would look like this:
```java
pen.createPointerDown(PointerInput.eventProperties().setHeight(12).setTiltX(30))
```

That said, it's hard to nitpick my solution when you compare it to how difficult Java makes working with non-default keyboard and mouse inputs because there is no access to them via `Actions` class... 

The full user-required code:
```java
        PointerInput pen = new PointerInput(PointerInput.Kind.PEN, "my pen");
        Sequence actionList = new Sequence(pen, 0)
                .addAction(pen.createPointerMove(Duration.ZERO, PointerInput.Origin.fromElement(element), 0, 0))
                .addAction(pen.createPointerDown(new PointerInput.eventProperties().setHeight(12).setTiltX(30))
                .addAction(pen.createPointerMove(Duration.ZERO, PointerInput.Origin.viewport(), 50, 50))
                .addAction(pen.createPointerUp(0));
        ((RemoteWebDriver) driver).perform(Collections.singletonList(actionList));
```

~Also, I couldn't figure out how to get Bazel to run the tests in `org.openqa.selenium.interactions`, so I put the test for this in `org.openqa.selenium.remote`.~


diemol (approved):
Thank you, @titusfortner!

PR #10199, opened 2021-12-30
[dotnet] implement setting pointer event properties

titusfortner:
I'm pretty sure I'm violating some .NET conventions on this one, but here's an implementation for #9996


dotnet/src/webdriver/Interactions/PointerInputDevice.cs
```diff
@@ -162,7 +162,17 @@ public override InputDeviceKind DeviceKind
...
-            return new PointerDownInteraction(this, button);
+            return new PointerDownInteraction(this, button, new PointerEventProperties());
+        }
+
+        /// <summary>
+        /// Creates a pointer down action.
+        /// </summary>
+        /// <param name="properties">The specifications for the pointer event interaction</param>
+        /// <returns>The action representing the pointer down gesture.</returns>
+        public Interaction CreatePointerDown(PointerEventProperties properties)
```
titusfortner: change constructor to also take `MouseButton` instead of just using `0`

dotnet/src/webdriver/Interactions/PointerInputDevice.cs
```diff
@@ -231,17 +285,55 @@ private class PointerDownInteraction : Interaction
             private double? altitudeAngle;
             private double? azimuthAngle;
 
-            public PointerDownInteraction(InputDevice sourceDevice, MouseButton button)
-                : base(sourceDevice)
+            public PointerEventProperties SetWidth(double widthValue)
```
titusfortner:
```
public double? Width {
   get { return this.width; }
  set { this.width = value; }
}
```

titusfortner:
Property initializer block:
```
PointerEventProperties props = new PointerEventProperties() {
  width = 10,
  height = 11
};
```

third_party/dotnet/devtools/src/generator/CodeGen/Utility.cs
```diff
@@ -20,10 +21,13 @@ public static class Utility
         /// <returns></returns>
         public static string ReplaceTokensInPath(string path, string className, CodeGeneratorContext context, CodeGenerationSettings settings)
         {
+            string separator = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "\\" : "/";
```
jimevans: I'd probably do string `separator = Path.DirectorySeparatorChar;` instead.

third_party/dotnet/devtools/src/generator/CodeGen/Utility.cs
```diff
@@ -20,10 +21,13 @@ public static class Utility
...
+
             path = path.Replace("{{className}}", className);
             path = path.Replace("{{rootNamespace}}", settings.RootNamespace);
             path = path.Replace("{{templatePath}}", settings.TemplatesPath);
             path = path.Replace("{{domainName}}", context.Domain.Name);
+            path = path.Replace("{{separator}}", separator);
```
jimevans: Or, just use `Path.DirectorySeparatorChar` here instead.

jimevans (changes requested):

third_party/dotnet/devtools/src/generator/CodeGen/Utility.cs
titusfortner: Yes, this is the better way to do this; I found other examples in the code and copied them. Thanks.

PR #10203, opened 2021-12-30
[java] remove deprecated action implementations

titusfortner:
This has all been deprecated for > 3 years (3.14.0) so I think we're ok to remove it.

If Selenium 4.2 is supposed to be the version where we all implement the updated Actions support, this seems like the right time to remove it. 

I got the old tests running, and everything that wasn't marked deprecated is passing. Please help me make sure I wasn't too aggressive in deleting something that was needed.


pujagani:
Thank you @titusfortner for your effort and contribution. I think it will be a good idea to split the PR into small changes so it is easy to review. It is easy to miss if something was needed in 76 files. What do you think?

titusfortner:
We can probably break out the touch actions parts, and maybe some of the old Firefox specific parts, but the keyboard and mouse stuff is pretty interrelated and going to be a pain either way I think.

titusfortner:
Closing in favor of #10215 
May want to revisit the differences between these 2 later

PR #10212, opened 2022-01-01
PEP484 type hints for py/selenium/webdriver/chrome/options.py

thecr8tr:
### Description
This PR adds type hints to `selenium.webdriver.chrome.options`

### Motivation and Context
This is an action towards #9480.

### Types of changes

### Checklist

### DIFF
--- a/py/selenium/webdriver/chrome/options.py
+++ b/py/selenium/webdriver/chrome/options.py
@@ -17,6 +17,7 @@
 
 from selenium.webdriver.chromium.options import ChromiumOptions
 from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
+from typing import Optional
 
 
 class Options(ChromiumOptions):
@@ -25,5 +26,9 @@ class Options(ChromiumOptions):
     def default_capabilities(self) -> dict:
         return DesiredCapabilities.CHROME.copy()
 
-    def enable_mobile(self, android_package="com.android.chrome", android_activity=None, device_serial=None):
+    def enable_mobile(self,
+                      android_package: str = "com.android.chrome",
+                      android_activity: Optional[str] = None,
+                      device_serial: Optional[str] = None
+                      ) -> None:
         super().enable_mobile(android_package, android_activity, device_serial)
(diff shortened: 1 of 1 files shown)


titusfortner:
Should this be `None` instead of `NoReturn`?

titusfortner:
@colons you have a very different implementation for `enable_mobile` for Firefox. (I'm assuming Chrome vs Firefox shouldn't be different)...

https://github.com/SeleniumHQ/selenium/blob/fb2b991762d9c4690e3b3c520dc171274d2ae74e/py/selenium/webdriver/firefox/options.py#L131

colons:
@titusfortner im not sure what difference exactly you're worried about here; the wrapping of optional kwargs with `Optional` is [a matter of preference](https://mypy.readthedocs.io/en/stable/config_file.html#confval-no_implicit_optional), for instance. annotating `BaseOptions.enable_mobile()` would be a good idea if you want mypy to complain at you if your subclasses are doing incompatible things

certainly, though, `NoReturn` is incorrect here, and i hope #10202 will make it easier to notice when new annotations are wrong

titusfortner:
@colons — I thought everything in Python had a one true way for things. :)
Thanks for clarifying and providing the link.

py/selenium/webdriver/chrome/options.py
```diff
@@ -25,5 +26,9 @@ class Options(ChromiumOptions):
     def default_capabilities(self) -> dict:
         return DesiredCapabilities.CHROME.copy()
 
-    def enable_mobile(self, android_package="com.android.chrome", android_activity=None, device_serial=None):
+    def enable_mobile(self,
+                      android_package: str = "com.android.chrome",
+                      android_activity: Optional[str] = None,
+                      device_serial: Optional[str] = None
+                      ) -> NoReturn:
```
titusfortner: This needs to be `None` not `NoReturn`

titusfortner (changes requested):

py/selenium/webdriver/chrome/options.py
thecr8tr: I corrected that and pushed it.  Should we be revisiting `NoReturn` usage in other places such as https://github.com/SeleniumHQ/selenium/blob/trunk/py/selenium/webdriver/firefox/webdriver.py#L186 ?

titusfortner: Yes! @colons already did that here — #10202 
I was trying to figure out why the `bazel build //py:test-remote` is failing before merging that one, but as discussed on Slack, I can't seem to get it to run on my machine at all.

titusfortner (approved):

PR #10213, opened 2022-01-01
[java] allow setting a different keyboard or pointer for actions on i…

titusfortner:
You can see how difficult it is to work with an input device that isn't a default keyboard or a default mouse in the tests I wrote for a pointer.

Using a `setActivePointer()` / `setActiveKeyboard` is one way of making this easier. I was looking at different options like putting it in the constructor, but then you'd need 8 different constructors when we add in wheel support soon.

The Actions class does a good job of adding Pauses where needed for newly seen inputs already, but something does not like 2 different objects being given the same name, so we need to re-use the existing  input object if it is there rather than creating a new one. 

The implementation can probably be improved.


diemol (approved):
This seems fine to me, and tests will look more simple.

I will wait for the CI to finish and see if something fails related to this change.

PR #10215, opened 2022-01-03
JWP Actions

titusfortner:
This is slightly less ambitious than #10203

I broke it up into 2 commits.
The first removes support for the things deprecated in 6b8bdae1 (1/7/2017) for selenium-3.1.0
The second removes support for the things deprecated in e561ca86 (10/29/2018) for selenium-3.141.0

I essentially deleted the deprecated classes and then removed everything that referenced them.

The biggest difference from the other commit is that I didn't touch any of the abandoned commands in `DriverCommand.java` which I think we can dramatically clean up in another commit.

One issue with this is that we did not deprecate `TouchActions.java`. I think the intention was to implement it with w3c compliant methods, but... because it is a subclass, we'd either need to add generics, make a bunch of private methods public, or add these methods to `Actions.java` directly. As it is, this implementation is not usable, so I think it should be removed for now, and we should add the touch actions in the `Actions.java` class directly, which I think we should do with ScrollWheel as well.


titusfortner:
I was previously concerned that these changes were going to break OSS -> W3C backwards compatibility, but I build grid with this branch and ran tests with Selenium 2.53 and everything seems to work still... If tests are passing, I say we merge it.

titusfortner:
never mind, this is not passing the old Selenium 2.53 tests. I think we should decide #10374 before we merge this.

diemol:
Merged first bit of this PR through https://github.com/SeleniumHQ/selenium/commit/e3089c962793427cf94c5fa83b9bb05a77f81634

diemol:
Second part was merged as https://github.com/SeleniumHQ/selenium/commit/4dcc9c44f1afcf6cdd1b7d76dc9115358b29c2ea. 

Overall, we could not delete all the code because it was being referenced from the old server used for RC. So what I did was to move the deprecated code that could not be deleted over there, even if we are not releasing RC anymore, we are keeping the code for a while. We will delete that in some months, if no issues arise.

PR #10216, opened 2022-01-04
node: Add support for installing unpacked addons at runtime in Firefox webdriver

nickgaya:
### Description

Add support for installing an extension from a directory in the Node.js Firefox WebDriver client.

The implementation checks if the addon path provided is a directory and, if so, creates a zip archive of the directory contents to send to geckodriver.

### Motivation and Context

This implements #8357 for node.  The same approach can be used for other language bindings.

The approach of creating an ad-hoc zip archive for a local directory was suggested here: https://github.com/SeleniumHQ/selenium/pull/9494#issuecomment-1001597660

### Types of changes

### Checklist


titusfortner:
Wow, this is great! Thank you for the extra effort you put in to make the extensions available in common. Now I actually have to re-implement things in the other languages to follow this approach.

titusfortner (approved):

harsha509 (approved):
Thank you @nickgaya.

l0b0:
If this does what I think it does (being able to install unsigned extensions locally without having to build my own Selenium for Firefox Devedition) that would be fantastic! Is that right? If so, is there some documentation on how to use it, or should it just work in the same way as in Devedition?

titusfortner:
It should already work if you zip the directory yourself. This PR lets you just pass in the path and Selenium will zip and send for you. 

You can see examples here — https://github.com/SeleniumHQ/selenium/blob/trunk/javascript/node/selenium-webdriver/test/firefox_test.js#L222

PR #10220, opened 2022-01-05
[java] Fix the retry request filter's max duration value to allow retries

pujagani:

### Description

Primarily related to https://github.com/SeleniumHQ/selenium/issues/9528

### Motivation and Context

Misunderstood the max-duration option (https://github.com/failsafe-lib/failsafe/pull/244/files) and hence it was not configured correctly to allow retries. This was an issue if the request timeout exception was thrown. The changes fix it by adding appropriate max duration values.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/remote/http/RetryRequest.java
+++ b/java/src/org/openqa/selenium/remote/http/RetryRequest.java
@@ -42,7 +42,7 @@ public class RetryRequest implements Filter {
       .handleIf(failure -> failure.getCause() instanceof ConnectException)
       .withBackoff(1, 4, ChronoUnit.SECONDS)
       .withMaxRetries(3)
-      .withMaxDuration(Duration.ofSeconds(10))
+      .withMaxDuration(Duration.ofMinutes(1))
       .onRetry(e -> LOG.log(
         getDebugLogLevel(),
         "Connection failure #{0}. Retrying.",
@@ -54,7 +54,7 @@ public class RetryRequest implements Filter {
       .handle(TimeoutException.class)
       .withBackoff(1, 4, ChronoUnit.SECONDS)
       .withMaxRetries(3)
-      .withMaxDuration(Duration.ofSeconds(10))
+      .withMaxDuration(Duration.ofMinutes(10))
       .onRetry(e -> LOG.log(
         getDebugLogLevel(),
         "Read timeout #{0}. Retrying.",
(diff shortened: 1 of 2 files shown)


java/src/org/openqa/selenium/remote/http/RetryRequest.java
```diff
@@ -54,7 +54,7 @@
       .handle(TimeoutException.class)
       .withBackoff(1, 4, ChronoUnit.SECONDS)
       .withMaxRetries(3)
-      .withMaxDuration(Duration.ofSeconds(10))
+      .withMaxDuration(Duration.ofMinutes(10))
```
titusfortner: This seems like a long Read Timeout, any reason 10 minutes? Is it configurable?

pujagani: The retry mechanism uses two parameters to retry: maxRetries and maxDuration. Max duration is the tricky one to understand. Please refer to https://github.com/failsafe-lib/failsafe/pull/244/files to understand how to max duration actually works. We have a read timeout of 3 minutes and we want to do 3 retries, which would take 9 minutes since the error would be only thrown and hence retried if it exceeds 3 minutes. To round it off, it is set to 10 minutes. The main purpose is to allow proper retries. I agree that it should be configurable. That is the next step since it does require moving a few things around. Apologies, if I have misunderstood the retry concepts. I have earlier misunderstood the maxDuration config, I thought it was the total duration between two consecutive retries.

titusfortner: Ah, yes, I think I misread this. Thanks.

Cybermaxke: Setting the maximum duration isn't required. If it's not set, it will assume an "unlimited" max duration, which is actually limited by the max retries and the configured read timeout. Removing the `withMaxDuration` would resolve the configurability, if I decide to increase the timeout to be longer than 3 minutes, e.g. 5 minutes, the max duration would be 15 minutes and this would conflict with the 10 minutes here.

I think it meant to be used that you either specify a max duration or max attempts, so you retry until the max duration is reached or retry a maximum number of n times, from the code:
https://github.com/failsafe-lib/failsafe/blob/66b8ca3189f058c6938db04dc153e4746cb0141a/src/main/java/dev/failsafe/internal/RetryPolicyExecutor.java#L237

diemol: That makes sense. If the user changes the timeout to a higher value, then this 10 minutes hard-coded value won't work. Setting only `withMaxRetries` should be fine.

pujagani: Thank you for pointing that out. Just relying on the maximum number of retries in our case is a good idea. I have made the desired changes.

PR #10232, opened 2022-01-07
Use class's class loader when looking up `HttpClient` factory.

dennisoelkers:
### Motivation and Context

Prior to this change, `HttpClient.Factory.create` was using `ServiceLoader.load` without passing in a class loader explicitly. In this case, `ServiceLoader` is using the current thread's default class loader to resolve the specified class name's service.

In projects with a plugin system, this can lead to problems, when selenium dependencies are part of a jar file that is not on the class path the application was started with. For these projects, using selenium results in a `ClassNotFoundException` when the first remote call is performed.

To fix this, using the class loader of the factory interface would be beneficial. In projects without a plugin architecture, it will mostly be identical with the class loader of the current thread, in other cases there is very high chance that the service which is to be loaded can be resolved through the class loader of `HttpClient.Factory`.

### Types of changes

### Checklist


diemol (approved):
Thank you for the contribution, @dennisoelkers!

PR #10292, opened 2022-01-26
[java]Sample code for xpath with web driver wait

dassusovan:
Xpath wrapper with Webdriver wait

Long story short
we are writing this way
driver.findElement(By.xpath(xpathExpression)).click();

so below method wrapping with webdriver wait with custom seconds
getElementByXpath("ADD_YOUR_XPATH", timeOutInSeconds).click();

### DIFF
--- a/java/src/org/openqa/selenium/By.java
+++ b/java/src/org/openqa/selenium/By.java
@@ -125,6 +125,32 @@ public abstract class By {
     }
     return allElements.get(0);
   }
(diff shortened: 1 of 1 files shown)


titusfortner:
Thank you for your proposal.
We intentionally minimize wrapper code added to Selenium for maintenance reasons and encourage our users to implement this functionality in their own test code.

PR #10355, opened 2022-02-10
switch ByClassName locator from to "class name"

katrina95:
"class" is not compatible with ios-driver, Refer to line 39 in https://github.com/ios-driver/ios-driver/blob/dev/server/src/main/java/org/uiautomation/ios/command/web/ToCSSSelectorConverter.java.

the "class" does not match any if condition here and will finally throw "WebDriverException: NI , selector type class"

If switch "class" to "class name", it works for both ios-driver, chromium-driver and geckodirver.


diemol:
Can you please add some context to this PR? Why is this needed? ios-driver has not been updated since 2015. 

Which cases cases of GeckoDriver and ChromeDriver are not working? Please share some code to understand what bug this PR is fixing.

Can you either post the information here as a comment or create an issue to understand better?

katrina95:
Just to clarify, the issue only exists in ios-driver, not in chrome or gecko driver.

I can not share my code, but if you write a small test with By.classname() and use ios-driver to test a safari browser, you will meet with the error "WebDriverException: NI , selector type class". I have a test like this, it passed before(on an old version selenium, but failed with new version. I verified that if I change "class" to "class name", all tests will pass.

diemol (approved):
After checking the commit history, seems that the absence of "name" was a typo.

If tests look good, we can merge.

PR #10356, opened 2022-02-10
Pre-process SameSite value before put it into caps

katrina95:
ChromeDriver is overly restrictive in which SameSite values are accepted. canonicalize the known good values in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite

### DIFF
--- a/java/src/org/openqa/selenium/Cookie.java
+++ b/java/src/org/openqa/selenium/Cookie.java
@@ -235,7 +235,24 @@ public class Cookie implements Serializable {
     toReturn.put("httpOnly", isHttpOnly());
 
     if (getSameSite() != null) {
-      toReturn.put("sameSite", getSameSite());
+      String sameSiteValue = getSameSite();
+      // ChromeDriver is overly restrictive in which SameSite values are accepted.
+      switch (sameSiteValue.toLowerCase()) {
+        case "strict":
+          sameSiteValue = "Strict";
+          break;
+        case "lax":
+          sameSiteValue = "Lax";
+          break;
+        case "none":
+          sameSiteValue = "None";
+          break;
+        default:
+          // Continue letting ChromeDriver throw an exception if the value doesn't match these
+          // known-allowed values, as they're not supported yet anyway.
+          break;
+      }
+      toReturn.put("sameSite", sameSiteValue);
     }
 
     return toReturn;


diemol:
@katrina95 could you please share more context for this change? The PR template has been ignored and the comment has no relation with the code changes.

Is there a bug which gets fixed with this change? Has it been reported somewhere? Also, what happens with Firefox, Edge or Safari with this change, does it still work?

katrina95:
I updated the comment, I have a samesite webdriver test in chrome browser, it throws error: 

org.openqa.selenium.InvalidArgumentException: invalid argument: invalid 'sameSite'

stack trace is:
at org.openqa.selenium.remote.codec.w3c.W3CHttpResponseCodec.createException([W3CHttpResponseCodec.java:200])
	at org.openqa.selenium.remote.codec.w3c.W3CHttpResponseCodec.decode([W3CHttpResponseCodec.java:133])
	at org.openqa.selenium.remote.codec.w3c.W3CHttpResponseCodec.decode([W3CHttpResponseCodec.java:53])
	at org.openqa.selenium.remote.HttpCommandExecutor.execute([HttpCommandExecutor.java:196])
	at org.openqa.selenium.remote.RemoteWebDriver.execute([RemoteWebDriver.java:682])
	at org.openqa.selenium.remote.RemoteWebDriver$RemoteWebDriverOptions.addCookie([RemoteWebDriver.java:887])

The reason is that ChromeDriver is overly restrictive in which SameSite values are accepted. canonicalize the known good values in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite

So I did the code change in my commit, and it passed. I want to check in this change in selenium projects, so that I don't need to make the patch in my local environment.

diemol:
> I have a samesite webdriver test in chrome browser, it throws error...

Can you share a test I can use to reproduce the issue?

I tried to reproduce it myself in Chrome by creating a cookie 
```java
Cookie cookie = new Cookie.Builder(key, "value").sameSite("strict").build();
```
but I did not bump into the issue.

Maybe even better, please add a test to `CookieImplementationTest.java` so we can double check the behaviour you mention.

diemol:
Closing as we did not get more information.

PR #10358, opened 2022-02-11
[py] Add support for http proxy authentication to remote_connection

luispflamminger:
### Description
Currently, `urllib3.ProxyManager` is used when connecting to a Selenium Server via http proxy.
The proxy url is read from environment variables.

Some http proxies require authentication. Credentials are usually passed in the url like this:
http://username:password@proxy.com:8080

When specifying a url like this and passing it into the ProxyManager's `proxy_url` attribute,
the ProxyManager is not able to authenticate to the proxy and a ProxyError is thrown:
`ProxyError('Cannot connect to proxy.', OSError('Tunnel connection failed: 407 Proxy Authentication Required'))`

Instead ProxyManager uses a `proxy_header` field for authentication, which can take a `proxy-authorization` header
that is then used to authenticate to the http proxy.

This PR implements support for this header.
If credentials are specified in the proxy url like shown above, they are seperated out.
They are then used to create a HTTP basic proxy-authorization header
using the [urllib3.util.make_headers](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.make_headers) method provided by urllib3.
The credentials are removed from the url and the ProxyManager object is created using the
url without credentials and the authorization header.

If a regular url is passed in without credentials, no auth header is specified and the ProxyManager is created as it was before.

### Motivation and Context

It is currently not possible to connect to a Selenium Server behind an http proxy that requires authentication using the Python Remote Webdriver.
This problem is solved here, as credentials can now be specified in the proxy url using the following standardized format:
http://username:password@proxy.com:8080
This feature is already present in other Selenium implementations (I tested it in Ruby), and this adds support for Python as well.

### Types of changes

### Checklist

(This is my first ever open source contribution, so if there is anything I can improve, please let me know!)


AutomatedTester:
Hey, thanks for the PR. Could you please check the github actions failures

luispflamminger:
Hi David,
I fixed the flake8 violations.
The failed unit test at py/test/unit/selenium/webdriver/firefox/firefox_options_tests.py:145 tests the firefox webdrivers `FirefoxProfile` and `Options` classes.
Unfortunately, I can't replicate the failure locally.
I looked at the code a bit and also don't really see, how my changes to the remote webdriver would have any impact on a test regarding the firefox webdriver options.
Maybe you could have a quick look at it if the problem persists after the next workflow run, as I have no experience with the firefox driver.

luispflamminger:
I rebased my feature branch, maybe that fixes the issue.

AutomatedTester (approved):

PR #10427, opened 2022-03-09
[java] Remove deprecated methods that use capabilities in WebDrivers

pujagani:

### Description

Remove deprecated methods that use capabilities in WebDrivers

### Motivation and Context

It is recommended to use browser-specific options instead of capabilities. Hence, the WebDriver implementation constructors that use capabilities are deprecated. These have been deprecated for multiple versions.

### Types of changes

### Checklist


titusfortner:
fwiw I'm 👎 on this for now. I'd like for us to keep it deprecated, but continue to support using `MutableCapabilities` that are w3c compliant. I think `DesiredCapabilities` *always toggles JWP, so removing ability to send JWP syntax will de facto remove the ability to use that method in a "bad" way for now, so I want to do that one first.

titusfortner:
Well, I take this back. Looks like Capabilities were only deprecated for local drivers, not `RemoteWebDriver`. I'm actually ok with this one if we want to do it, but think we should wait for Selenium 4.3.

titusfortner (approved):
Let's do this for Selenium 4.3

diemol (approved):

PR #10463, opened 2022-03-20
[py] remove unused `kind` arg of `WheelInput`

### LINKED ISSUE #10210: [🐛 Bug]: action_builder.ActionBuilder.add_wheel_input throws TypeError
Calling `ActionBuilder.add_wheel_input()` will always throw a `TypeError`. There are no tests or docs that I can see, so I'm unsure whether to remove the `kind` argument from `add_wheel_input` or add it to `WheelInput.__init__()`. This appears to have been the case since [it was introduced](https://github.com/SeleniumHQ/selenium/commit/122c9f9feaa69a3d052858c307f86e3cb37e85b4).

You can reproduce the `TypeError` without even creating a `WebDriver` instance by doing:

```python
>>> from selenium.webdriver.common.action_chains import ActionChains
>>> ActionChains('').w3c_actions.add_wheel_input('', '')
  File ".../selenium/webdriver/common/actions/action_builder.py", line 77, in add_wheel_input
    new_input = WheelInput(kind, name)
TypeError: __init__() takes 2 positional arguments
```

symonk:
Fixes #10210, this method looks like it blindly copy/pasted from `add_pointer_input(...)` originally.  Still need to add a test for this and looking at `WheelInput` in general it looks a little neglected/unused, will try understand what it's for exactly and tidy up.

(tests incoming when my local setup is working shortly)

I suspect there is more work to do on this class; some instances are being passed `name=interaction.WHEEL` but it defines internally a `self.type=interaction.WHEEL` so possibly other bugs there as well.


SinghHrmn (commented):
LGTM

PR #10480, opened 2022-03-26
[rb] Add process status check to the stop_process method too

yoshoku:
### Description

[This commit](https://github.com/SeleniumHQ/selenium/commit/e3e492af0c9d9c99ce31b9dc036dbf75327e6f30) removed the `return if process_exited?` from the beginning of the `stop_process` method. However, since [there is `ensure` at the method level in the `stop` method](https://github.com/SeleniumHQ/selenium/blob/trunk/rb/lib/selenium/webdriver/common/service_manager.rb#L69-L71), the `stop_process` method is called even after [the process exit checking in the `stop` method](https://github.com/SeleniumHQ/selenium/blob/trunk/rb/lib/selenium/webdriver/common/service_manager.rb#L63). To prevent `NoMethodError`, I would like to add `return if process_exited?` to the beginning of the `stop_process` method.

simple example:
```ruby
class Foo
  def bar
    if @process.nil?
      puts 'A'
      return
    end

    puts 'B'
  ensure
    puts 'C'
    @process.stop
  end
end

f = Foo.new
f.bar
```

```sh
$ ruby foo.rb
A
C
foo.rb:11:in `bar': undefined method `stop' for nil:NilClass (NoMethodError)

    @process.stop
            ^^^^^
        from foo.rb:16:in `<main>'
```

### Motivation and Context

Same as https://github.com/SeleniumHQ/selenium/pull/10015. This change is necessary to exit without raising `NoMethodError` if there is a problem creating a child process.

### Types of changes

### Checklist

### DIFF
--- a/rb/lib/selenium/webdriver/common/service_manager.rb
+++ b/rb/lib/selenium/webdriver/common/service_manager.rb
@@ -110,6 +110,8 @@ module Selenium
       end
 
       def stop_process
+        return if process_exited?
+
         @process.stop STOP_TIMEOUT
         @process.io.stdout.close if Platform.jruby? && !WebDriver.logger.debug?
       end


rb/lib/selenium/webdriver/common/service_manager.rb
```diff
@@ -110,6 +110,8 @@ def start_process
       end
 
       def stop_process
+        return if process_exited?
```
luke-hill, line 113: to avoid an error here why don't we instead check that there is no process running?

i.e. `return unless process_running?` Because that checks the iVar presence as well as the nil check, and an alive check.

EDIT: Although in hindsight they're both kinda similar.

yoshoku: When a process is exited, the `stop_process` method is called via `ensure` on [L69](https://github.com/SeleniumHQ/selenium/blob/trunk/rb/lib/selenium/webdriver/common/service_manager.rb#L69), triggered by `return if process_exited?` on [L63](https://github.com/SeleniumHQ/selenium/blob/trunk/rb/lib/selenium/webdriver/common/service_manager.rb#L63), so I wrote `return if process_exited?`.

p0deje:
Thank you for contribution!

PR #10490, opened 2022-03-29
`python`: test function names tidy up inline with pep8 

symonk:
A rather crazy one, apologies whoever reviews!  I noticed a lot of the tests use a camel case naming, in python we should opt for snake case as best as possible; with the help of regex & the `string manipulation` plugin for pycharm (highly recommend it!) I've went through all the `def test*` functions inside `/py/test/selenium*` and rejigged them into snake case.

This also has some extra pieces of:

- Remove duplicate test names (likely not executing the former defined ones as they are overwritten in the namespace)
- Made _some_ non test functions (utility) inline with snake case
- Removed some redundant fixtures from tests

Kinda relying on CI to give me a steer in any fallout here; but I was quite meticulous when going through it; so should be minimal (I hope).


AutomatedTester (approved):

AutomatedTester:
thanks for the PR

PR #10510, opened 2022-04-05
Fixing options handling for print command.

dennisoelkers:
### Motivation and Context

Prior to this change, when executing e.g. a `RemoteWebDriver#print` command, the print options passed to the method were passed through to the chromedriver instance, but serialized in a wrong format, rendering them uneffective. The issues were:

  - Options were wrapped in an `options` key when chromedriver expected
    them to be at the top-level of the parameters dict (see [here](https://github.com/chromium/chromium/blob/main/chrome/test/chromedriver/window_commands.cc#L2246))
  - Page size options were put into a `pageSize` key, when chromedriver
    expected it to be a `page` key (see [here](https://github.com/chromium/chromium/blob/main/chrome/test/chromedriver/window_commands.cc#L539))
  - The `Orientation` enum was serialized incorrectly (uppercase), while chromedriver is expecting the values to be all lowercase (see [here](https://github.com/chromium/chromium/blob/main/chrome/test/chromedriver/window_commands.cc#L466-L467))

Also, using the `PageSize` class supplied by selenium was not possible, due to defaulting to constant page sizes, not allowing the consumer to override them. This was addressed by adding a second constructor allowing to specify height/width.

All tests were performed against a v96 chromedriver.

### Types of changes

### Checklist


dennisoelkers:
Hey @diemol,

some of the checks are failing, but it looks to me as if they are unrelated. Is my impression correct, or do I need to fix anything?

java/src/org/openqa/selenium/print/PrintOptions.java
```diff
@@ -98,4 +111,17 @@ public PageSize getPageSize() {
...
   }
+
+  public Map<String, Object> toMap() {
+    final Map<String, Object> options = new HashMap<>(7);
+    options.put("page", getPageSize());
+    options.put("orientation", getOrientation().toString());
+    options.put("scale", getScale());
+    options.put("shrinkToFit", getShrinkToFit());
+    options.put("background", getBackground());
+    options.put("pageRanges", getPageRanges());
```
pujagani: I checked out the branch to run some tests. The https://github.com/SeleniumHQ/selenium/blob/trunk/java/test/org/openqa/selenium/PrintPageTest.java failed because the `pagesRanges` array is null if it is not set. Updating getPageRanges() for a null check should help fix it.

pujagani (dismissed):
Thank you @dennisoelkers for your contribution!

java/src/org/openqa/selenium/print/PrintOptions.java
dennisoelkers: Hey @pujagani, I noticed that too :) Fixed it in https://github.com/SeleniumHQ/selenium/pull/10510/commits/c361154bef7915d84ec67459cbaef2d2c9f4d3e8, I hope that is going to cut it.

pujagani: Thank you for the fix!

pujagani (approved):
Thank you @dennisoelkers for your contribution!

PR #10516, opened 2022-04-06
Update remote_connection.py

rohitanand1614:
Suggesting this change as when we try to run scripts on remote platform like browserstack and sauce labs, the urllib3 is throwing an error urllib3.exceptions.MaxRetryError: HTTPConnectionPool, this can be avoided if we set this value. the error was not coming until the urllib3==1.24.3. But post that urllib3 made certificate default.  There have been multiple communication chains regarding this issue. I could only fix this issue by explicity setting this value to none in the library stack.   In case you feel like not approving the change please do share the suggested fix to handle this issue as i could not find any.

### Types of changes

### Checklist

### DIFF
--- a/py/selenium/webdriver/remote/remote_connection.py
+++ b/py/selenium/webdriver/remote/remote_connection.py
@@ -146,7 +146,7 @@ class RemoteConnection(object):
             'timeout': self._timeout
         }
         if self._ca_certs:
-            pool_manager_init_args['cert_reqs'] = 'CERT_REQUIRED'
+            pool_manager_init_args['cert_reqs'] = 'CERT_NONE'
             pool_manager_init_args['ca_certs'] = self._ca_certs
 
         if self._proxy_url:


rohitanand1614:
Please review and take action accordingly

AutomatedTester:
Thank you for this PR but I don't understand fully why this is needed, what scenarios create this situation or why we are setting a default to something very different.

Please can you explain it further and when you said there are 
> There have been multiple communication chains regarding this issue.

Where are these?

symonk:
Could you share any of the communication threads here so I can catch up on it please?  as of urllib3 `1.25` https connections are verified by default using `CERT_REQUIRED`

symonk:
will take a look at the ssl contexts / what is happening under the hood here when I get back from PTO

titusfortner:
Thank you for submitting this PR, but we need a reproducible bug report to understand the underlying problem and if this is the preferred fix.

PR #10537, opened 2022-04-10
[py] Tidy up python exception handling

symonk:
Would like this for `4.3` not `4.2`.  A lot of the python code has some unnecessary exception handling catching some exception and instantly re-raising it; best to just let it bubble up here than catching it to begin with; I've removed usages of:

- catching an exception and instantly re-raising it with just not simply catching it
- `raise e` with just `raise` to avoid creating unnecessary convoluted stack traces


py/selenium/webdriver/remote/switch_to.py
```diff
@@ -141,7 +141,7 @@ def send_handle(h):
         try:
             # Try using it as a handle first.
             send_handle(window_name)
-        except NoSuchWindowException as e:
```
justnpT, line 144: why are removing the "as e" ?

py/selenium/webdriver/support/color.py
```diff
@@ -123,7 +123,7 @@ def hue_to_rgb(lum1: float, lum2: float, hue: float) -> float:
                     hue -= 1
 
                 if hue < 1.0 / 6.0:
-                    return (lum1 + (lum2 - lum1) * 6.0 * hue)
+                    return lum1 + (lum2 - lum1) * 6.0 * hue
```
justnpT, line 128: nice

py/test/selenium/webdriver/common/webdriverwait_tests.py
```diff
@@ -38,7 +38,7 @@ def test_should_fail_with_invalid_selector_exception(driver, pages):
         WebDriverWait(driver, 0.7).until(EC.presence_of_element_located((By.XPATH, "//*[contains(@id,'something'")))
 
 
-def test_should_explicitly_wait_for_asingle_element(driver, pages):
+def test_should_explicitly_wait_for_a_single_element(driver, pages):
```
justnpT, line 41: nice !

py/selenium/webdriver/remote/switch_to.py
symonk: 0-arg raise restores the last handled exception (and traceback)

py/selenium/webdriver/remote/switch_to.py
```diff
@@ -53,7 +53,7 @@ def alert(self) -> Alert:
                 alert = driver.switch_to.alert
         """
         alert = Alert(self._driver)
-        alert.text
+        _ = alert.text
```
AutomatedTester, line 56: Why are we doing this?

titusfortner:
fyi. I set this to draft since it says not to merge

py/selenium/webdriver/remote/switch_to.py
symonk: just to indicate the value is never used; will revert this one; edit: thoughts on keeping as-is? (most IDEs etc will complain about the statement with no effect)

AutomatedTester:
@symonk why do you think we should wait until for 4.3? I can't see any breaking changes

symonk:
> @symonk why do you think we should wait until for 4.3? I can't see any breaking changes

just wanted more time to test, but I think it's ok, was branched off trunk when flake8 was complaining; will update

AutomatedTester (approved):

PR #10560, opened 2022-04-19
[py] proof of concept; formatting python code with `black`.

symonk:
Hi, a quick proof of concept I'd like to discuss using something like black (https://github.com/psf/black) to apply to our python codebase.  This is a naive implementation, consider this PR mainly for discussions.  I think it's critical to get consistency across the python code base to make contributing as simple as possible with a clean (automatic) UX.

Ideally I'd like to get something like `pre-commit` (https://github.com/pre-commit/pre-commit) applying this stuff automatically on git pre-commit to avoid needless churn and be as as effecient as possible; formatting is only one small benefit there, we can use it to automate a whole heap of useful stuff.

Thoughts?

If we feel this is beneficial I will look at implementing it properly; for now I have just did a 'quick pass' of the tooling to show what it may look like.  I think the consistency is a massive plus.  We can make slight adjustments to the configuration if we need, or even go with a completely different formatter but I'd like to see something alongside `flake8`.  I would strongly push implementing `pre-commit` with some hooks on the client side.

[Do not merge this with tox/setup configuration as is; for demo purposes only]


jsa34:
I would agree with linting and enforcing standards with precommit (just my opinion!)
Would probably be a good idea to pop in a flake8 or something for enforcing some pep8 standards in the precommit hook?

symonk:
Absolutely; heres a config I tend to run with for my own projects:
https://github.com/symonk/asserto/blob/main/.pre-commit-config.yaml

black/flake8 are only a small drop in what we could do; the thing about pre-commit is how it's easy pluggable with whatever we need; (I have for example my own hook that catches `pdb` shortcuts as variable names and prevents commits e.g `s = 10` type stuff

AutomatedTester:
I don't want a pre-commit hook for this but I'm not against this. The code has clearly grown organically over a while so cleaning it up will be a good thing.

titusfortner:
I made this draft since it is poc.

Where do we go since we aren't doing a pre-commit hook? Can we get the linter test in our CI to flag things that aren't correct?

symonk:
I can add a tox recipe for running black and have it applied in CI.  I understand you are against the concept of a pre-commit hook, maybe in future (I much prefer the code to not even make it to the server to be told its incorrect; find it inefficient) but for now i will make this a CI based tox thing that will fail PRs if you think theres value in having it.  I'll also try not to tarnish `git blame` while doing so

Issue #10265, opened 2022-01-19
[🚀 Feature]: Support sending Firefox Addon directory as temporary in Remote Sessions


titusfortner:
### Feature and motivation

Firefox Addon endpoint supports either `path` or `addon` parameter, with addon being a base64 encoded string of a compressed file, which can be either an xpi or a zip file.

Directories are needed to support the temporary parameter. Rather than implementing a file detector and trying to upload the directory to a remote machine, directories should be zipped and base 64 encoded and sent with the `addon` parameter.

Supporting temporary parameter is https://github.com/SeleniumHQ/selenium/issues/10084
It was fixed in every language, except .NET has outstanding PR — https://github.com/SeleniumHQ/selenium/pull/10093
So, it works if the user zips the directory and passes it in, but it would be nice for Selenium to do this for the user.

A great example of this by @nickgaya in JS is here:
https://github.com/SeleniumHQ/selenium/pull/10216

### Usage example

```
String id = ((HasExtensions) augmentedDriver).installExtension(Paths.get("path/to/addon/directory/"), true);
```

pmgexpo17:
> ### Feature and motivation
> Firefox Addon endpoint supports either `path` or `addon` parameter, with addon being a base64 encoded string of a compressed file, which can be either an xpi or a zip file.
> 
> Directories are needed to support the temporary parameter. Rather than implementing a file detector and trying to upload the directory to a remote machine, directories should be zipped and base 64 encoded and sent with the `addon` parameter.
> 
> Supporting temporary parameter is #10084 It was fixed in every language, except .NET has outstanding PR — #10093 So, it works if the user zips the directory and passes it in, but it would be nice for Selenium to do this for the user.
> 
> A great example of this by @nickgaya in JS is here: #10216
> 
> ### Usage example
> ```
> String id = ((HasExtensions) augmentedDriver).installExtension(Paths.get("path/to/addon/directory/"), true);
> ```

Hi,
[...]

PR #10566, opened 2022-04-20
Java: Support for sending firefox addon directory as temporary in remote sessions

TamsilAmani:
### Description
Modified the `installExtension()` to send firefox addon directory as temporary in remote session.

### Motivation and Context

- to implement java solution for #10265 
- References made : https://github.com/SeleniumHQ/selenium/issues/8357#issuecomment-643047209

### Types of changes

### Checklist


java/src/org/openqa/selenium/firefox/AddHasExtensions.java
```diff
@@ -28,12 +28,15 @@
 import org.openqa.selenium.remote.ExecuteMethod;
 import org.openqa.selenium.remote.http.HttpMethod;
 
+import java.io.FileOutputStream;
 import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
+import java.nio.file.*;
```
pujagani: IntelliJ tends to do this by default when optimizing imports. But it is good practice to avoid importing entire package.

TamsilAmani: Alright, I'll correct that.

java/src/org/openqa/selenium/firefox/AddHasExtensions.java
```diff
@@ -77,7 +84,23 @@ public String installExtension(Path path, Boolean temporary) {
...
+            ZipOutputStream zos = new ZipOutputStream(new FileOutputStream(extZip.toFile()));
+            Files.walkFileTree(path, new SimpleFileVisitor<Path>() {
+              public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
+                zos.putNextEntry(new ZipEntry(path.relativize(file).toString()));
+                Files.copy(file, zos);
+                zos.closeEntry();
+                return FileVisitResult.CONTINUE;
+              }
+            });
+            zos.close();
```
diemol: Could this be done in a Try with Resources block?
Maybe we can extract those lines for the unzipping into another method so it looks more readable.

diemol (changes requested):

java/src/org/openqa/selenium/firefox/AddHasExtensions.java
```diff
@@ -77,7 +84,23 @@ public String installExtension(Path path, Boolean temporary) {
 
         String encoded;
         try {
-          encoded = Base64.getEncoder().encodeToString(Files.readAllBytes(path));
+          if (Files.isDirectory(path)) {
+            Path extZip = Paths.get(path.getFileName().toString()+".zip");
+            ZipOutputStream zos = new ZipOutputStream(new FileOutputStream(extZip.toFile()));
+            Files.walkFileTree(path, new SimpleFileVisitor<Path>() {
```
pujagani: It should be fine if Path is not passed explicitly.

java/src/org/openqa/selenium/firefox/AddHasExtensions.java
TamsilAmani: Incorporated the changes. Please check to verify and resolve :)

java/src/org/openqa/selenium/firefox/AddHasExtensions.java
TamsilAmani: Removing `Path` is breaking the code. I'll look into it further.

diemol (approved):
Looks good, thank you @TamsilAmani!

pujagani (approved):

PR #10575, opened 2022-04-26
Added DepricationWarning to the warnings.warn methods inside the find_element_by_x methods

BusHero:
Added "DepricationWarning" argument to the warnings.warn inside the find_element_by_x methods from the WebElement class.

In such a way, the IDE will mark the methods as deprecated, similar to how it marks the find_element_by_x methods from the webDriver class.


symonk (changes requested):
might be some value in not using the default `stacklevel` too,  can we pass `stacklevel=2`?

BusHero:
> might be some value in not using the default `stacklevel` too, can we pass `stacklevel=2`?

Sounds like a good idea

AutomatedTester (dismissed):

AutomatedTester:
There are some tox errors, please can you fix and then we will merge it.

diemol (approved):
Review comments were addressed and tox failures are gone, will merge.

Thanks, @BusHero!

PR #10577, opened 2022-04-26
fixes #10574

### LINKED ISSUE #10574: [🐛 Bug]: Intercepting network response produces exception
### What happened?

Intercepting network response crashes the browser. 

@titusfortner [said](https://stackoverflow.com/a/71975826/2620206): 
the code is assuming responseHeaders will be there, but that's optional according to [chromedevtools.github.io/devtools-protocol/tot/Fetch](https://chromedevtools.github.io/devtools-protocol/tot/Fetch/)

```shell
#issue is reproduced by typing `irb` and pasting the code below

require "watir"
require "selenium/devtools"

Selenium::WebDriver::Chrome::Service.driver_path = '/chromedriver99.0.4844.51'
Selenium::WebDriver::Chrome.path = '/chrome99.0.4844.84/google-chrome'
options.add_option("args", ["--no-sandbox"])

browser = Watir::Browser.new :chrome, options: options

log_file = File.open("/tmp/network.log", 'w')
browser.wd.intercept do |request, &continue|
    continue.call(request) do |response|
      log_file << "#{request.id} \t
                   #{request.method} \t
                   #{response.code} \t
                   #{request.url} \n"
    end
end

browser.goto "youtube.com" #this crashes the browser
```

### Relevant log output

```shell
#<Thread:0x00005590f1bcda90@/selenium-webdriver-4.1.0/lib/selenium/webdriver/devtools.rb:145 run> terminated with exception (report_on_exception is true):
        3: from /selenium-webdriver-4.1.0/lib/selenium/webdriver/devtools.rb:155:in `block in callback_thread'
        2: from /selenium-webdriver-4.1.0/lib/selenium/webdriver/common/driver_extensions/has_network_interception.rb:68:in `block in intercept'
        1: from /selenium-webdriver-4.1.0/lib/selenium/webdriver/common/driver_extensions/has_network_interception.rb:109:in `intercept_response'
/selenium-webdriver-4.1.0/lib/selenium/webdriver/devtools/response.rb:38:in `from': undefined method `each_with_object' for nil:NilClass (NoMethodError)
```

### Operating System

Ubuntu

### Selenium version

ruby 2.6.5p114 (2019-10-01 revision 67812) [x86_64-linux]

Chrome 99.0.4844.84

ChromeDriver99.0.4844.51

No

meetme2meat:
Attempt to fix https://github.com/SeleniumHQ/selenium/issues/10574

### DIFF
--- a/rb/lib/selenium/webdriver/devtools/response.rb
+++ b/rb/lib/selenium/webdriver/devtools/response.rb
@@ -35,7 +35,7 @@ module Selenium
             id: id,
             code: params['responseStatusCode'],
             body: (Base64.strict_decode64(encoded_body) if encoded_body),
-            headers: params['responseHeaders'].each_with_object({}) do |header, hash|
+            headers: (params['responseHeaders'] ||  []).each_with_object({}) do |header, hash|
               hash[header['name']] = header['value']
             end
           )
--- a/rb/lib/selenium/webdriver/devtools/response.rb
+++ b/rb/lib/selenium/webdriver/devtools/response.rb
@@ -35,7 +35,7 @@ module Selenium
             id: id,
             code: params['responseStatusCode'],
             body: (Base64.strict_decode64(encoded_body) if encoded_body),
-            headers: (params['responseHeaders'] ||  []).each_with_object({}) do |header, hash|
+            headers: params.fetch('responseHeaders', []).each_with_object({}) do |header, hash|
               hash[header['name']] = header['value']
             end
           )
--- a/rb/spec/unit/selenium/webdriver/devtools/response_spec.rb
+++ b/rb/spec/unit/selenium/webdriver/devtools/response_spec.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+# Licensed to the Software Freedom Conservancy (SFC) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The SFC licenses this file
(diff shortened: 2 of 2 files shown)


rb/lib/selenium/webdriver/devtools/response.rb
```diff
@@ -35,7 +35,7 @@ def self.from(id, encoded_body, params)
             id: id,
             code: params['responseStatusCode'],
             body: (Base64.strict_decode64(encoded_body) if encoded_body),
-            headers: params['responseHeaders'].each_with_object({}) do |header, hash|
+            headers: (params['responseHeaders'] ||  []).each_with_object({}) do |header, hash|
```
titusfortner: Interestingly, if we'd used `#fetch` here we would have had a better error message. That was your initial suggestion, any reason for the change? We probably should be moving in that direction for our Hash implementations in general (tough to enforce that in a linter)

titusfortner (changes requested):
you can run rubocop from the `rb/` directory to see what our settings flag on.

rb/lib/selenium/webdriver/devtools/response.rb
meetme2meat: Done. Zero Rubocop warning and used `params.fetch`

titusfortner (approved):

rb/lib/selenium/webdriver/devtools/response.rb
titusfortner: Thanks, sorry for delay

PR #10589, opened 2022-04-29
[py] An expectation for checking if non-empty text is present in element

wwakabobik:
text_to_be_present_in_element_non_empty method is added to check that
the web element contains any non-empty text. Usually it's needed when
text is generated on-the-fly on the server side, and of course, that
the text of element is not empty.

Also, added self-test for this feature.

### Types of changes

### Checklist


py/docs/source/webdriver_support/selenium.webdriver.support.expected_conditions.rst
```diff
@@ -34,6 +34,7 @@
       text_to_be_present_in_element
       text_to_be_present_in_element_attribute
       text_to_be_present_in_element_value
+      text-to_be_present_in_element_non_empty
```
symonk: typo `-` here

py/selenium/webdriver/support/expected_conditions.py
```diff
@@ -254,6 +254,20 @@ def _predicate(driver):
     return _predicate
 
 
+def text_to_be_present_in_element_non_empty(locator):
```
symonk, line 257: I need to discuss with those in the core team; I am not sure our current stance on adding additional expected_conditions - and we should update the docs to be clear (if we prefer them to be more user defined in their own projects) - will report back shortly

symonk:
Thanks a lot for the contribution @wwakabobik - I'm quite new here so just need to figure out with folks what our current stance is on expected_conditions (additional ones living in selenium core; will let you know as soon as I know!)

py/docs/source/webdriver_support/selenium.webdriver.support.expected_conditions.rst
wwakabobik: thanks, fixed

py/selenium/webdriver/support/expected_conditions.py
wwakabobik: Ok, thanks, I'll wait for decision

titusfortner:
Thanks for your contribution.

.NET got rid of the expected conditions classes entirely because everyone wanted to add their own (e.g., now make one for text with whitespace), and the maintenance burden was too great. So now .NET & Ruby both require all waits to be done via lambdas. 

Selenium users in the other languages should make more use of lambdas. For this one, Python is especially nice because an empty string is "falsy":

```python
WebDriverWait(driver, 3).until(lambda d: d.find_element(locator).text.strip())
```

PR #10596, opened 2022-05-01
[py] Bugfix: `text_to_be_present_in_element_attribute` predicate checking

### LINKED ISSUE #10592: [🐛 Bug]:
### What happened?

Hello,

I'm using selenium=4.1.3 with python 3.7.

I think the [expected_conditions.text_to_be_present_in_element_attribute](https://github.com/SeleniumHQ/selenium/blob/trunk/py/selenium/webdriver/support/expected_conditions.py#L245) don't work the way it should.

image

In the current code, the condition L245 is useless because [element_attribute_to_include ](https://github.com/SeleniumHQ/selenium/blob/trunk/py/selenium/webdriver/support/expected_conditions.py#L436)always return a function.

So L247 will return None if the attribute doesn't exists and L248 will raise a TypeError: argument of type 'NoneType' is not iterable.

I think the intended behavior is : 
`
if not element_attribute_to_include(locator, attribute_)(driver):
    return False
`

```shell
from selenium.webdriver.support.expected_conditions import text_to_be_present_in_element_attribute

driver = webdriver.Chrome()
driver.get("https://google.com")

tester = text_to_be_present_in_element_attribute((By.CSS_SELECTOR, "body"), "missingattribute", "oups")
tester(driver)
```

### Relevant log output

```shell
  File "lib\site-packages\selenium\webdriver\support\expected_conditions.py", line 248, in _predicate
    return text_ in element_text
TypeError: argument of type 'NoneType' is not iterable
```

### Operating System

Windows 10

### Selenium version

4.1.3

Chrome 101

ChromeDriver 101.0.4951.41

_No response_

symonk:
minor tweak for `text_to_be_present_in_element_attribute` which is referencing another expected condition however it is evaluating the result of that call in a boolean context; that function is returning a predicate function and not it's respective boolean return type.

closes #10592

### DIFF
--- a/py/selenium/webdriver/support/expected_conditions.py
+++ b/py/selenium/webdriver/support/expected_conditions.py
@@ -242,7 +242,7 @@ def text_to_be_present_in_element_attribute(locator, attribute_, text_):
 
     def _predicate(driver):
         try:
-            if not element_attribute_to_include(locator, attribute_):
+            if not element_attribute_to_include(locator, attribute_)(driver):
                 return False
             element_text = driver.find_element(*locator).get_attribute(attribute_)
             return text_ in element_text


diemol (approved):
Thank you, @symonk!

PR #10613, opened 2022-05-05
[py] use `typing_extensions` to be compatible with `py37`

### LINKED ISSUE #10612: [🐛 Bug]: 4.1.4 Released but only working with Python 3.8+
### What happened?

The new release 4.1.4 is out and is only usable for Python 3.8+
because it relies on version **'Literal' from 'typing'**  which is only available on version 3.8+ of Python.

cannot import name 'Literal' from 'typing' (/usr/local/lib/python3.7/typing.py)

```shell
Just run a python script using selenium with python < 3.8
```

### Relevant log output

```shell
'Literal' from 'typing'
```

### Operating System

MacOS,Linux

### Selenium version

4.1.4

Chrome

Chrome

_No response_

symonk:
`typing.Literal` is only available in py3.8+; this patch uses `typing-extensions` for older versions.

closes #10612

### DIFF
--- a/py/selenium/webdriver/common/virtual_authenticator.py
+++ b/py/selenium/webdriver/common/virtual_authenticator.py
@@ -19,7 +19,14 @@ import functools
 
 from base64 import urlsafe_b64encode, urlsafe_b64decode
 from enum import Enum
-from typing import Literal
+import typing
+import sys
+
+if typing.TYPE_CHECKING:
+    if sys.version_info >= (3, 8):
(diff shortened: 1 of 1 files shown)


symonk:
@diemol CI is running under `python 3.8.10` as mentioned; we need to fix that as soon as poss. and the chrome fails are flaky tests by the looks of it

AutomatedTester (approved):

Issue #10379, opened 2022-02-19
[🚀 Feature]: Remove Opera Classes


titusfortner:
### Feature and motivation

opera driver has not been updated to support w3c, so we've agreed to remove those classes.

### Usage example

Ruby deleted the classes for Selenium 3 release.

Users can use Chromedriver for opera with `ChromeOptions` class and `setBinary` with the path to the Opera Browser.

diemol:
Do you remember if we want to deprecate first and the remove?

titusfortner:
https://www.selenium.dev/meetings/2022/tlc-01-19/#delete-opera-classes-from-the-bindings-given-that-operadriver-does-not-support-w3c

Deprecate in Java, remove elsewhere

symonk:
@titusfortner happy enough if i rip out the `Opera` classes on the python side without deprecation?

ibodog:
I saw the deprecation note in the java docs for selenium.  Not sure how to tell Selenium Grid to use the Chrome workaround.  My grid rejects the session request if I use ChromeOptions and then change 'browserName' to 'Opera'.  Without that change the grid just runs the Chrome browser (when you think you've asked for Opera).  BTW, this is using the dockerized grid with node-docker.  Opera, Chrome, Firefox, Edge are all enabled on the grid and Opera works OK with the deprecated OperaOptions (outside of the w3c issues...).

    org.openqa.selenium.SessionNotCreatedException: Could not start a new session. Possible causes are invalid address of the remote server or browser start-up failure.

    Caused by: org.openqa.selenium.TimeoutException: java.util.concurrent.TimeoutException
    Build info: version: '4.1.3', revision: '7b1ebf28ef'
    System info: host: 'test', ip: 'fe80:0:0:0:1843:95ae:fcd0:90b4%en0', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '11.6.5', java.version: '17.0.2'
    Driver info: driver.version: RemoteWebDriver
    Caused by: java.util.concurrent.TimeoutException


Grid shows this pending session:
[...]

PR #10625, opened 2022-05-09
[py]: Deprecate `opera` webdriver & options in prep for removal in `4.3`

symonk:
Deprecate opera driver `Options` and `WebDriver` subclass usage.  I have added appropriate 'rip out' todo's so grepping for `# Todo: gh-10379 (Remove for 4.3)` will yield everything that needs removed after 4.2 is cut; and we can properly remove it for `4.3` .  There isn't actually any `opera` webdriver based tests outside of basic unit/options ones so I didn't invest the time into that here as i'll be removing it in a few days again.

relates to #10379

### DIFF
--- a/py/selenium/webdriver/__init__.py
+++ b/py/selenium/webdriver/__init__.py
@@ -25,7 +25,7 @@ from .ie.options import Options as IeOptions  # noqa
 from .edge.webdriver import WebDriver as Edge  # noqa
 from .edge.webdriver import WebDriver as ChromiumEdge  # noqa
 from .edge.options import Options as EdgeOptions  # noqa
-from .opera.webdriver import WebDriver as Opera  # noqa  # Todo: gh-10379 (Remove for 4.3)
+from .opera.webdriver import WebDriver as Opera  # noqa
 from .safari.webdriver import WebDriver as Safari  # noqa
 from .webkitgtk.webdriver import WebDriver as WebKitGTK  # noqa
 from .webkitgtk.options import Options as WebKitGTKOptions  # noqa
--- a/py/selenium/webdriver/__init__.py
+++ b/py/selenium/webdriver/__init__.py
@@ -25,7 +25,7 @@ from .ie.options import Options as IeOptions  # noqa
 from .edge.webdriver import WebDriver as Edge  # noqa
 from .edge.webdriver import WebDriver as ChromiumEdge  # noqa
 from .edge.options import Options as EdgeOptions  # noqa
-from .opera.webdriver import WebDriver as Opera  # noqa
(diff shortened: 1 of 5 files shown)


py/selenium/webdriver/__init__.py
```diff
@@ -25,7 +25,7 @@
 from .edge.webdriver import WebDriver as Edge  # noqa
 from .edge.webdriver import WebDriver as ChromiumEdge  # noqa
 from .edge.options import Options as EdgeOptions  # noqa
-from .opera.webdriver import WebDriver as Opera  # noqa
+from .opera.webdriver import WebDriver as Opera  # noqa  # Todo: gh-10379 (Remove for 4.3)
```
titusfortner: Are these `Todo:` necessary for user visibility or simply a note for devs? If the latter than let's not have them in there, we'll know what to delete. :)

titusfortner (changes requested):
We probably also should put a link to a note somewhere somehow for why it is getting deprecated, and how to drive Opera with chromedriver. Pretty sure the latter needs to be in our docs with examples.

py/selenium/webdriver/__init__.py
symonk: seems fair; will remove

py/selenium/webdriver/opera/options.py
```diff
@@ -23,7 +24,8 @@ class Options(ChromeOptions):
     KEY = "operaOptions"
 
     def __init__(self):
-        ChromeOptions.__init__(self)
+        warnings.warn(f"{self.__class__} is deprecated and will be removed in 4.3", DeprecationWarning, stacklevel=2)
```
titusfortner:
```suggestion
        warnings.warn(f"{self.__class__} is deprecated and will be removed in 4.3; see https://www.selenium.dev/documentation/webdriver/getting_started/open_browser/#opera", DeprecationWarning, stacklevel=2)
```

py/selenium/webdriver/opera/webdriver.py
```diff
@@ -42,6 +44,7 @@ def __init__(self, executable_path=None, port=0,
          - service_log_path - Where to log information from the driver.
            capabilities only, such as "proxy" or "loggingPref".
         """
+        warnings.warn(f"{self.__class__} is deprecated and will be removed in 4.3", DeprecationWarning, stacklevel=2)
```
titusfortner:
```suggestion
        warnings.warn(f"{self.__class__} is deprecated and will be removed in 4.3; see https://www.selenium.dev/documentation/webdriver/getting_started/open_browser/#opera", DeprecationWarning, stacklevel=2)
```

titusfortner (approved):

PR #10630, opened 2022-05-10
[py] remove `opera`

symonk:
`4.2` seen the deprecation of opera; I have opened this for now to account for the removal of it altogether; will re base later once we ship 4.2 and get this in for `4.3`.

#10379 will be considered done as far as the python bindings are concerned in 4.3

@titusfortner @AutomatedTester we originally thought `4.3` for opera removal; but it's only deprecated in `4.2` however the message does say it'll be removed in `4.3` specifically; whatever you think tho, maybe `4.4` for this? I'll keep the PR in sync atleast.


AutomatedTester (approved):

divinity76:
Opera version 13+ is little more than a re-skin of Chromium right?

titusfortner:
@divinity76 yes, but operadriver is not properly supporting the w3c protocol which is required for Selenium 4, and we've had issues driving Opera with chromedriver recently because of some new features that Google has released.

PR #10636, opened 2022-05-11
[java] implement separate Origin class for WheelInput

titusfortner:
This is what I was talking about in the TLC meeting. I'd rather have a separate implementation than use one from another (tangential but still unrelated) class that has an unsupported method that we have to treat differently.

(I also have an idea for how I'd like to build on this class in a way that requires it to be independent).


pujagani (approved):
I did bump across this when implementing scroll but was not sure which is the best way to go due to code repetition. This is definitely a cleaner approach. Thank you @titusfortner!

diemol (approved):

PR #10637, opened 2022-05-11
copy web content to be served on selenium.dev

titusfortner:
The idea here is for us to be able to add code to https://github.com/SeleniumHQ/seleniumhq.github.io/tree/dev/examples for display in our documentation. This will make it a lot easier to create code examples for our documentation because we can just re-use our existing tests.

### DIFF
--- a/generate_web_code.sh
+++ b/generate_web_code.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+
+# copy website to build
+mkdir build/web
+cp -R common/src/web build/web
+
+# switch to gh-pages and copy the files
+git checkout gh-pages || exit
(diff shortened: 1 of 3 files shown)


ollie-iterators:
This should also help Selenium get test coverage easier because it provides an incentive for adding tests.

titusfortner:
good point. Assuming we get to the point where we have more example code than test code. :)

diemol (approved):

diemol:
https://www.selenium.dev/selenium/web/

and you will need to know the file name as directories are not listed (I think for security reasons).

e.g. https://www.selenium.dev/selenium/web/colorPage.html

PR #10638, opened 2022-05-11
[py]: Extra reusable utility function for css_selector conversions

symonk:
adds a reusable `CSS_SELECTOR` conversion utility method that both `webdriver` and `webelement` can benefit from with some additional types.

Todo:
 - double check types
 - finalise docstrings


titusfortner:
The other languages all implement a separate "SearchContext" class that handles all of the find_element and find_elements code, so none of it is duplicated. Not sure how it would be implemented in Python, but it would probably be a better approach than just putting the conversion code in a utils class.

symonk:
> The other languages all implement a separate "SearchContext" class that handles all of the find_element and find_elements code, so none of it is duplicated. Not sure how it would be implemented in Python, but it would probably be a better approach than just putting the conversion code in a utils class.

sounds interesting; I will take a look, I notice David also has a comment about this in the `ShadowRoot` implementation.

edit: Initially a `SearchContextMixin` might be an idea; however each implementation in `WebDriver` & `WebElement` & `ShadowRoot` all _slightly_ vary and reference instance attributes of different names etc

PR #10643, opened 2022-05-13
Add __all__ to __init__.py to make imports public

hudcap:
### Description

Add __all__ variable to webdriver.__init__.py with a reference to all imports to make them public.

### Motivation and Context

`pyright` throws a "private import" error when using e.g. `webdriver.Chrome`
From the `pyright` docs [here](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md):
> Imported symbols are considered private by default. If they use the “import A as A” (a redundant module alias), “from X import A as A” (a redundant symbol alias), or “from . import A” forms, symbol “A” is not private unless the name begins with an underscore. If a file __init__.py uses the form “from .A import X”, symbol “A” is not private unless the name begins with an underscore (but “X” is still private). If a wildcard import (of the form “from X import *”) is used, all symbols referenced by the wildcard are not private.
A module can expose an __all__ symbol at the module level that provides a list of names that are considered part of the interface. The __all__ symbol indicates which symbols are included in a wildcard import. All symbols included in the __all__ list are considered public even if the other rules above would otherwise indicate that they were private. For example, this allows symbols whose names begin with an underscore to be included in the interface.

Redundant aliases seem to only work if the name isn't being changed.
I'm no type expert, so there might be a "cleaner" solution, but this works.

### Types of changes

### Checklist

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -152,8 +152,8 @@ selenium_register_dotnet()
 
 http_archive(
     name = "build_bazel_rules_nodejs",
-    sha256 = "e328cb2c9401be495fa7d79c306f5ee3040e8a03b2ebb79b022e15ca03770096",
-    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.4.2/rules_nodejs-5.4.2.tar.gz"],
+    sha256 = "0fad45a9bda7dc1990c47b002fd64f55041ea751fafc00cd34efb96107675778",
+    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.5.0/rules_nodejs-5.5.0.tar.gz"],
 )
 load("@build_bazel_rules_nodejs//:repositories.bzl", "build_bazel_rules_nodejs_dependencies")
 
--- a/common/src/web/virtual-authenticator.html
+++ b/common/src/web/virtual-authenticator.html
@@ -19,7 +19,6 @@
       options=Object.assign({
         authenticatorSelection: {
           requireResidentKey: false,
-          userVerification: 'preferred'
         },
         rp: {
           id: "localhost",
@@ -41,7 +40,7 @@
           status: "OK",
           credential: {
             id: credential.id,
-            rawId: Array.from(new Int8Array(credential.rawId)),
+            rawId: Array.from(new Uint8Array(credential.rawId)),
             transports: credential.response.getTransports(),
           }
         };
@@ -62,7 +61,7 @@
         return {
           status: "OK",
           attestation: {
-            userHandle: new Int8Array(attestation.response.userHandle),
+            userHandle: new Uint8Array(attestation.response.userHandle),
           },
         };
       } catch(error) {
--- a/dotnet/src/support/Events/EventFiringWebDriver.cs
+++ b/dotnet/src/support/Events/EventFiringWebDriver.cs
@@ -427,7 +427,8 @@ namespace OpenQA.Selenium.Support.Events
         /// </para>
         /// <para>
         /// Arguments must be a number (which will be converted to a <see cref="long"/>),
-        /// a <see cref="bool"/>, a <see cref="string"/> or a <see cref="IWebElement"/>.
+        /// a <see cref="bool"/>, a <see cref="string"/> or a <see cref="IWebElement"/>,
+        /// or a <see cref="IWrapsElement"/>.
         /// An exception will be thrown if the arguments do not meet these criteria.
         /// The arguments will be made available to the JavaScript via the "arguments" magic
         /// variable, as if the function were called via "Function.apply"
@@ -488,7 +489,8 @@ namespace OpenQA.Selenium.Support.Events
         /// </para>
         /// <para>
         /// Arguments must be a number (which will be converted to a <see cref="long"/>),
-        /// a <see cref="bool"/>, a <see cref="string"/> or a <see cref="IWebElement"/>.
+        /// a <see cref="bool"/>, a <see cref="string"/> or a <see cref="IWebElement"/>,
+        /// or a <see cref="IWrapsElement"/>.
         /// An exception will be thrown if the arguments do not meet these criteria.
         /// The arguments will be made available to the JavaScript via the "arguments" magic
         /// variable, as if the function were called via "Function.apply"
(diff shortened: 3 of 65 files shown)


symonk:
Sorry I'm not at my computer right now but it makes sense to start getting an 'official' API in place, would make refactoring things a little easier in future as ~technically everything is "public" right now~, ideally we could expose everything through `from selenium import x` at the top level but various classes have the same names e.g all drivers, so something like this I think is a decent start, will review properly shortly. Edit: (we are actually doing more than I thought in the webdriver subpackage init). I suspect your tool may also complain about exceptions from the common subpackage?

hudcap:
`from selenium.webdriver.common.by import By` does not raise any issues. But I'm not sure that's what you're asking.
From what I can see of the structure, the `common` subpackage doesn't seem to expose anything directly.
I think the issue with `webdriver` is as you mentioned—e.g. all the `WebDriver` classes have the same name in their respective subpackages, and all the renaming happens in the `webdriver.__init__.py`, making that the best location to import from.
I don't know enough about this area to know if there's a "cleaner" general solution.
This solution is what [typeshed](https://github.com/microsoft/pyright/blob/f4383ab61a24100c69f1334c07517d14b1a533d0/packages/pyright-internal/typeshed-fallback/stubs/selenium/selenium/webdriver/__init__.pyi) uses, so I basically copied it over and adjusted it.

Tagging @Akuli on this as a typeshed maintainer

Akuli:
You don't need an explicit `__all__`. You can instead use `Foo as Foo`, as in `from .common.keys import Keys as Keys`. It looks weird, but type checkers understand this as re-exporting the name.

Typeshed uses `__all__` or `Foo as Foo` depending on what the actual module does: if it has `__all__`, typeshed usually has `__all__` as well, and if it doesn't, typeshed uses the `Foo as Foo` trick.

Akuli:
And now that I read the discussion more carefully:

> Redundant aliases seem to only work if the name isn't being changed.

This is a bit annoying to deal with. There's some discussion about it here: https://github.com/python/typing/pull/844#discussion_r694156487

If you want to do this without `__all__`, you can replace `from .firefox.options import Options as FirefoxOptions` with:

```python3
from .firefox.options import Options as _FirefoxOptions
FirefoxOptions = _FirefoxOptions
```

But this just looks weird and isn't nice. I wish there was a cleaner way to do this, but there isn't.

IMO a proper fix (that will break backwards compatibility) would be to just rename the class to `FirefoxOptions`. The class name also shows up in error messages: `'Options' object has no attribute 'foo'` can be very confusing if a user does `from selenium.webdriver import FirefoxOptions` and also defines their own class named `Options`.

hudcap:
@Akuli Thanks for the clarification!
Takeaway: barring a backwards-incompatible blanket renaming of classes, this seems to be the "cleanest" option. If the classes are ever renamed in the future, `__all__` can then be removed.

@symonk should I add an additional comment about this?

symonk:
> @Akuli Thanks for the clarification! Takeaway: barring a backwards-incompatible blanket renaming of classes, this seems to be the "cleanest" option. If the classes are ever renamed in the future, `__all__` can then be removed.
> 
> @symonk should I add an additional comment about this?

seems reasonable to me; should we take this opportunity to add other things are a missing from the `__all__` e.g should we be exposing `WebElement`; `Service` (and its subclasses) officially in `__all__` ?  Perhaps a question for @AutomatedTester  but I think we _should_ be.  Also does `selenium.common.__init__` officially need to document the exceptions in `__all__ = (..., ...)` ?

AutomatedTester:
> > @Akuli Thanks for the clarification! Takeaway: barring a backwards-incompatible blanket renaming of classes, this seems to be the "cleanest" option. If the classes are ever renamed in the future, `__all__` can then be removed.
> > @symonk should I add an additional comment about this?
> 
> seems reasonable to me; should we take this opportunity to add other things are a missing from the `__all__` e.g should we be exposing `WebElement`; `Service` (and its subclasses) officially in `__all__` ? Perhaps a question for @AutomatedTester but I think we _should_ be. 

That makes sense. A number of the things here are historical artifacts to simplify how people could do reflection in their code so we need to tread carefully and not break any backwards compat .

> Also does `selenium.common.__init__` officially need to document the exceptions in `__all__ = (..., ...)` ?

It should be.

hudcap:
> ideally we could expose everything through `from selenium import x`

> should we take this opportunity to add other things are a missing from the `__all__` e.g should we be exposing `WebElement`; `Service` (and its subclasses) officially in `__all__` ?

To clarify based on the above, would `WebElement` and `Service` be exposed under `webdriver` or `selenium`?
Either way, I don't know which other classes are being referred to here.

> Also does `selenium.common.__init__` officially need to document the exceptions in `__all__ = (..., ...)` ?

Please correct me if I'm wrong, but that `.__init__` doesn't seem to serve any purpose, since `exceptions` is already a module in `selenium.common`

symonk:
> > ideally we could expose everything through `from selenium import x`
> 
> > should we take this opportunity to add other things are a missing from the `__all__` e.g should we be exposing `WebElement`; `Service` (and its subclasses) officially in `__all__` ?
> 
> To clarify based on the above, would `WebElement` and `Service` be exposed under `webdriver` or `selenium`? Either way, I don't know which other classes are being referred to here.
> 
> > Also does `selenium.common.__init__` officially need to document the exceptions in `__all__ = (..., ...)` ?
> 
> Please correct me if I'm wrong, but that `.__init__` doesn't seem to serve any purpose, since `exceptions` is already a module in `selenium.common`

selenium.webdriver will do for now I think, top level is something we should consider in future

hudcap:
Anything holding this up?
Happy to adjust it if necessary

Akuli:
> that `.__init__` doesn't seem to serve any purpose, since exceptions is already a module in selenium.common

It does have a purpose. This is a really unintuitive "feature" of Python's import system.

If you comment out that line and do `from selenium import common`, you can't do `common.exceptions`. It will fail with an `AttributeError`. But if you do `from selenium.common import exceptions as some_completely_unrelated_name`, then `common.exceptions` will suddenly start working!

So whether or not you can do `common.exceptions` depends on whether the `exceptions` module has been loaded anywhere in the program. Loading the `exceptions` module and using it can even be in different files of your program, and it will still work.

By having `__init__.py` load the `exceptions` module, you make sure that it's always available, and it's impossible to write code that works only because a different file in your program happens to import something from the `exceptions` module.

hudcap:
> > that `.__init__` doesn't seem to serve any purpose, since exceptions is already a module in selenium.common
> 
> It does have a purpose. This is a really unintuitive "feature" of Python's import system.
> 
> If you comment out that line and do `from selenium import common`, you can't do `common.exceptions`. It will fail with an `AttributeError`. But if you do `from selenium.common import exceptions as some_completely_unrelated_name`, then `common.exceptions` will suddenly start working!
> 
> So whether or not you can do `common.exceptions` depends on whether the `exceptions` module has been loaded anywhere in the program. Loading the `exceptions` module and using it can even be in different files of your program, and it will still work.
> 
> By having `__init__.py` load the `exceptions` module, you make sure that it's always available, and it's impossible to write code that works only because a different file in your program happens to import something from the `exceptions` module.

Thanks for that info!
I had no idea—I think I had run into this issue before and couldn't figure out why that sometimes didn't work.

AutomatedTester (approved):

PR #10645, opened 2022-05-13
Add docker device mapping configuration

gifflet:
### Description
Configure docker containers device mapping

### Motivation and Context
Sometimes is needed to map a device file to a docker
container to make possible some tasks like virtualize
a VM or something else. Because of this, I believe it
is desirable to have a way to configure which device
files should be available in containers.

In this commit, we can define which device files should
be mapped in containers through the standard selenium grid
configuration file.

### Types of changes

### Checklist

### DIFF
--- a/common/src/web/virtual-authenticator.html
+++ b/common/src/web/virtual-authenticator.html
@@ -19,7 +19,6 @@
       options=Object.assign({
         authenticatorSelection: {
           requireResidentKey: false,
-          userVerification: 'preferred'
         },
         rp: {
           id: "localhost",
@@ -41,7 +40,7 @@
           status: "OK",
           credential: {
             id: credential.id,
-            rawId: Array.from(new Int8Array(credential.rawId)),
+            rawId: Array.from(new Uint8Array(credential.rawId)),
             transports: credential.response.getTransports(),
           }
         };
@@ -62,7 +61,7 @@
         return {
           status: "OK",
           attestation: {
-            userHandle: new Int8Array(attestation.response.userHandle),
+            userHandle: new Uint8Array(attestation.response.userHandle),
           },
         };
       } catch(error) {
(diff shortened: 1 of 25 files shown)


gifflet:
The documentation was updated in SeleniumHQ/seleniumhq.github.io#1011

diemol (commented):
Code looks fine, thanks for the PR. I'd just like to understand better what the use case for this is, how would you be using this feature?

gifflet:
Thank you @diemol for reviewing this PR. 

Answering your question, I implemented this change because the containers launched by selenium grid perform virtualization of virtual machines, that is, a virtual machine is started offering the native web browser of the platform at the same time that the selenium server runs in the virtual machine in standalone mode . So I need the containers to have access to the /dev/kvm device, available on the host, for the virtualization tool to work as expected.

diemol:
Can volumes be used instead of devices?

diemol:
Can volumes be used instead of devices?

gifflet:
Unfortunately not. When we share a device through a volume, the container does not have the proper read and write permissions to the shared device. 

image

diemol:
Got it, thanks.

Can you please have a look at the issues reported by Sonar?

gifflet:
Sure, I will work on it.

gifflet:
@diemol, I could not reproduce the issue reported by Sonar but I think that the last commit solves it.

diemol:
Can you please rebase first? Or let us do it by enabling the PR?

diemol:
Maybe I missed it, but I do not see the other Sonar issues being addressed.

gifflet:
The other issues reported by Sonar was fixed in the last commit. The rebase was done as suggested.

diemol:
I'd love to merge this but if I can't update this branch on my own is just going to take longer. Can you please update your fork?

gifflet:
@diemol, I just enabled maintainer editing, so I believe you will be able to update this branch.

diemol:
Thanks

diemol (commented):
Thank you, @dilhelh!

Issue #10536, opened 2022-04-10
[py] CI tests are intermittently failing for pull requests [list of culprit tests]


symonk:
### What happened?

A fair few of the python tests in CI are a little flaky; resulting in red builds in most python pull requests at the minute; we are merging a few knowing they are 'flaky' so let's get that fixed up.

Edit: There are quite a few; here is a compiled lists of tests that suffer in CI; getting information for them all is pretty time consuming; but they could use some work.  Some are harder to recreate than others and they _could_ be due to latency in a CI system; most cases are hitting `TimeoutErrors`:

 - ~test_full_virtual_authenticator (Todo: Add links)~ (fixed)
 - test_click_should_timeout (Todo: Add links)
 - test_should_timeout_on_page_load_taking_too_long (Todo: Add links)
 - test_collect_log_mutations (BIDI/async) (Todo: Add links)
 - test_should_be_able_to_click_element_that_is_out_of_view_in_aframe_that_is_out_of_view[chrome] [output](https://github.com/SeleniumHQ/selenium/runs/6441541077?check_suite_focus=true#step:9:81)
 - test_touch_pointer_properties - [output](https://github.com/SeleniumHQ/selenium/runs/6433705992?check_suite_focus=true#step:9:112)

*os*: `ubuntu 20.04`
*chromedriver*: `100.0.4896.75`
*chrome*: `100`

symonk:
`test_full_virtual_authenticator` fails because it has a dependency on `webauthn.io` which is not stable; it errors occassionally during the test.

SinghHrmn:
@symonk test_full_virtual_authenticator: (outdated)

symonk:
Thanks @SinghHrmn

nnrepos:
hey @symonk can you give some examples of failures in git actions?
by looking at the actions of the last few days, i only see `tox-lint` fail in `selenium/webdriver/firefox/firefox_binary.py`.
thanks 😄

PR #10648, opened 2022-05-14
[py]: fix `flake8` issues in `trunk`

symonk:
`trunk` is currently broken with py builds due to `flake8` violations.  (I much prefer to have `pre-commit` catching these before they are committed locally rather than finding out later on the remote)

### DIFF
--- a/py/selenium/webdriver/common/action_chains.py
+++ b/py/selenium/webdriver/common/action_chains.py
@@ -321,7 +321,7 @@ class ActionChains:
         self.send_keys(*keys_to_send)
         return self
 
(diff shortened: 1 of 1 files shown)


symonk:
Flake chrome tests; added it to the list of: #10536

titusfortner:
fwiw, I run rubocop on Ruby files and fix issues before I push. If there is an equivalent command for Python that I can use I can do that.

titusfortner:
fixed by separate update

symonk:
Thanks, yeah `tox -e flake8`

Issue #10541, opened 2022-04-13
[🚀 Feature]: Add virtual authenticator support for all Selenium language bindings 


pujagani:
### Feature and motivation

Applications often integrate WebAuthn as part of the login step. This triggers a WebAuthn browser-specific WebAuthn flow. 
Example of WebAuthn in Chrome: 
<img width="455" alt="Screenshot 2022-04-13 at 6 02 36 PM" src="https://user-images.githubusercontent.com/10705590/163180633-e66a423a-4d17-4973-a508-a7b237e309ec.png">

Automating interaction with the WebAuthn dialog, as shown above, is possible via the extension VirtualAuthenticator APIs defined as part of https://w3c.github.io/webauthn/#sctn-automation. 

### Implementation
Implementation for Java was added as part of https://github.com/SeleniumHQ/selenium/pull/7760 and https://github.com/SeleniumHQ/selenium/pull/7842. The rest of the language bindings need to add the relevant code and can refer to the Java implementation for guidance.

An outline of implementing VirtualAuthenticator
1. [VirtualAuthenticatorOptions](https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/virtualauthenticator/VirtualAuthenticatorOptions.java) (this is used when adding the VirtualAuthenticator) 
2. [Credential](https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/virtualauthenticator/Credential.java) (The credentials to be used by the VirtualAuthenticator)
3. List of methods (which call the WebDriver commands) and the respective parameters required: 

-  addVirtualAuthenticator(VirtualAuthenticatorOptions options) - Returns an authenticator id as a string
- removeVirtualAuthenticator (String authenticatorId)
- addCredential(Credential credential, String authenticatorId)
- getCredentials(String authenticatorId) - Returns a Map<String,Object>
- removeCredential(byte[] credentialId, String authenticatorId)
- removeCredential(String credentialId, String authenticatorId)
- removeAllCredentials(String authenticatorId) 
- setUserVerified(boolean verified, String authenticatorId)
Refer: 
https://github.com/SeleniumHQ/selenium/blob/b1fafa7ead771bfe70b659cf96d1f29cb22a2605/java/src/org/openqa/selenium/remote/RemoteWebDriver.java#L654
[...]

PR #10650, opened 2022-05-14
Convenience Actions methods for Scrolling

titusfortner:
Trunk right now has basic scrolling support merged. 
This PR is set to conveniently show the diff of what exists and what could exist.

### Options for reviewers:

1. Reject all changes and keep what is in trunk exactly as it is
2. Reject PR but make changes to what is in trunk 
3. Accept PR, but make proposed changes
4. Accept PR and/or do nothing and this will get merged before 4.2 is released

### Currently in Trunk

* I updated the Ruby implementation match the Python implementation approach.
* I made the Java & .NET approaches similar (except .NET is using a class with properties, Java is using a class with static methods)
* All languages have the same descriptions for each of the parameters
* We have tests for the 5 working and 2 failing tests for these 4 languages.

There are legit issues to be taken with the implementation in trunk (especially with .NET since the class with properties isn't ideal). But, it is basic implementation of a `scroll()` method with all the functionality.

### Changes proposed by this PR

The primary reason for this PR is that I believe it is not intuitive for users to understand the purpose of `x` and `y`, even if we called them `offsetX` and `offsetY`. The idea of offset from origin and the origin being one of two different things took me a while to figure out, and I like to think I'm at least average in figuring out this stuff compared to our users.

This PR removes a generic scroll method and replaces it with 3 convenience methods; the first 2 represent what 98% of use cases require. The 3rd is the featured method but it makes it blatantly obvious that the offsets are explicitly associated with the origin (viewport or element).

The names match the existing language conventions for pointer:

1. Scroll to Element — takes element argument
.NET — `ScrollToElement()`
Java — `scrollToElement()`
Python — `scroll_to_element()`
Ruby — `#scroll_to`

2. Scroll by Amount — takes deltaX/deltaY values:
.NET — `ScrollByAmount()`
Java — `scrollByAmount()`
Python — `scroll_by_amount()`
Ruby — `#scroll_by`

3. Scroll from an origin — takes deltaX/deltaY and a ScrollOrigin object instance
.NET — `ScrollFromOrigin()`
Java — `scrollFromOrigin()`
Python — `scroll_from_origin()`
Ruby — `#scroll_from`

There are a lot of potential (and valid) quibbles with this implementation, from method names to defaults, to where defaults live, to which stage of the actions class process does translations of scroll origin, etc. I tried very hard to make the implementations happen at the same "place" in the different languages, but in a language idiomatic way.


py/selenium/webdriver/common/action_chains.py
```diff
@@ -323,25 +323,54 @@ def send_keys_to_element(self, element, *keys_to_send):
         self.send_keys(*keys_to_send)
         return self
 
-    def scroll(self, x: int, y: int, delta_x: int, delta_y: int, duration: int = 0, origin: Union[str,WebElement] = "viewport"):
+    def scroll_to_element(self, element: WebElement):
```
symonk, line 328: I think we need to stagger this change over two releases? (its a breaking public API change I think), should we keep `scroll(...)` with a `DeprecationWarning`?

py/selenium/webdriver/common/actions/wheel_actions.py
```diff
@@ -29,6 +29,6 @@ def pause(self, duration=0):
         self.source.create_pause(duration)
         return self
 
-    def scroll(self, x, y, delta_x, delta_y, duration, origin):
+    def scroll(self, x=0, y=0, delta_x=0, delta_y=0, duration=0, origin="viewport"):
```
symonk, line 32: be nice to type these; `int/float?` `float` if either is acceptible

py/selenium/webdriver/common/actions/wheel_input.py
```diff
@@ -21,6 +21,34 @@
...
 
+class ScrollOrigin:
+
+    def __init__(self, origin: Union[str, WebElement], x_offset: int, y_offset: int) -> None:
+        self._origin = origin
+        self._x_offset = x_offset
+        self._y_offset = y_offset
+
+    @classmethod
+    def from_element(cls, element: WebElement, x_offset: int = 0, y_offset: int = 0):
```
symonk, line 32: does this offer any value over the normal init? (can the dunder `__init__` have defaults for the offsets as `=0` itself?

py/selenium/webdriver/common/actions/wheel_input.py
```diff
@@ -21,6 +21,34 @@
...
+        return cls(element, x_offset, y_offset)
+
+    @classmethod
+    def from_viewport(cls, x_offset: int = 0, y_offset: int = 0):
+        return cls('viewport', x_offset, y_offset)
+
+    @property
+    def origin(self) -> Union[str, WebElement]:
```
symonk, line 40: read-only ok for these? (seems ok to me)

py/selenium/webdriver/common/action_chains.py
AutomatedTester: We can't break APIs like this.

titusfortner: This is the downside of doing patch releases instead of waiting for minor releases with new code, but yes, can keep existing method and mark it deprecated instead of deleting it.

py/selenium/webdriver/common/actions/wheel_actions.py
titusfortner: I had types and then removed them. I want to allow comparing apples to apples for this PR. We're going to do a full update on typing in the future.

py/selenium/webdriver/common/actions/wheel_input.py
titusfortner: The idea is to make it explicit the object you get instead of just constructing by default. This is a common Java pattern that I also use in Ruby. If it isn't idiomatic Python we can adjust. Java can make the constructor private. Ruby & Python people can avoid the class methods, but we can at least have them for our documentation.

py/selenium/webdriver/common/actions/wheel_input.py
titusfortner: Yes, I thought you had to explicitly add a setter to make it not read only?

py/selenium/webdriver/common/action_chains.py
AutomatedTester: We can't break APIs like this.

AutomatedTester (changes requested):
APIs can't just change from under people in any of the code changes. That goes against the usual deprecation policies we have.

py/selenium/webdriver/common/actions/wheel_input.py
symonk: Yep I'm just wondering if that was the intent (vs allowing them to be set and modified afterwards on the instance)

titusfortner:
Yes, I guess I kind of would like to treat changes made in patches as different from changes made (and announced) in minor versions. Otherwise we need to rethink feature branch management. Or just give up on trying to collaborate between bindings.

The fix in this case is easy enough, though. I restored the previous python method and marked it as deprecated.
None of the other languages have had a patch release since the changed method/class names were introduced.

py/selenium/webdriver/common/actions/wheel_input.py
titusfortner: Ideally the constructor would also be "private" like we can do it in Java. Ruby we are marking the constructor `@api private` to annotate that. Not sure if there is an equivalent in Python. The ideal is to only create the instance via class method, which properly sets appropriate values/defaults.

titusfortner:
Thinking more about the bigger picture here. I definitely overestimated what we'd be able to do back in November, which we need to better account for going forward.

TL/DR: I'd like to see us better prioritize 3 things:
1. Smaller more frequent minor releases (easier said than done)
2. Marketing pushes w/ blog posts for each minor release (requires languages to each have the feature implemented)
3. More consistency between the OOP languages (JS is a little special; the rest can do more similar approaches even in language idiomatic ways)

Selenium needs better marketing. That means making blog posts publicizing the cool new features in minor releases and pointing to good, working examples for each of them and encouraging people to upgrade. We need people to be talking and blogging and speaking about Selenium by constantly keeping ourselves worth talking about.

For that to work well, we kind of need the bindings in lock-step. I know this was never a priority in 3.x and each language kind of released features whenever they realized they needed to be added and the amount of communication wasn't ideal. I'd like to prioritize not doing this.

But it makes things weird when it comes to things like:
https://github.com/SeleniumHQ/selenium/issues/10541
Do we need to get the other languages to implement it before we can release 4.2 if we want to advertise it? Do we need to collaborate better on new features and not merge things until everyone is ready to merge and release a new minor version?
Do we maybe do patch releases as beta releases so we can change the API before a minor release while waiting for everyone to implement? (I'd rather figure out how to do smaller more frequent (monthly w/Chrome updates?) minor releases, but that's hard with current resources. 

As for consistency, I'm coming at this from the standpoint of trying to write examples for Sauce Labs and then for our Selenium documentation. It would be much easier to show smaller code snippets if the general approach were kept a little more similar. im did a really good job of keeping .NET functionality similar to Java, though Java didn't always keep up with .NET. I definitely think the Ruby bindings are lacking because when we (I) implemented things, I didn't take any cues from Java or .NET in a way that would have been helpful looking at it in retrospect.

Anyway, my Sunday morning thoughts.

titusfortner:
Here's the publicly executable test code and the documentation referencing it, if we merge this PR:
https://deploy-preview-1010--jovial-austin-42fe02.netlify.app/documentation/webdriver/actions_api/wheel/

py/selenium/webdriver/common/action_chains.py
diemol: Seems you a had an issue in your comments @AutomatedTester, somehow it went twice :)

AutomatedTester: Weird... It wasn't intended to be twice

titusfortner: I think we posted at the same time and github got confused how to display things.

pujagani:
Thank you, Titus! Java code changes look good to me. Can you please help me understand why the Ruby method names differ from other bindings? Apologies, I might be missing some context here.

titusfortner:
Great question. Each language is matching the previous naming conventions from mouse movement. All the other languages have longer methods names, ruby truncated the method name and relies on the first parameter to complete the idea.

PR #10651, opened 2022-05-15
[py]: exclude `*venv` directories from `flake8`

symonk:
Invoking `tox` is a requirement currently and not something that bazel does; this means people need to typically make a virtualenv and install `tox` in it to do this task; installing `tox` globally is also possible but not -advised (but that won't encounter the problem this fixes if so).

Common python practice would be:

`python3.7 -m venv .venv` || `python3.7 -m venv venv` ( in the case of using 3.7 like selenium encourages)

This excludes `*venv` so that `tox -e flake8` does not take a long time to complete and flag up 200+ false positives by scanning files inside the virtual env.

### DIFF
--- a/py/setup.cfg
+++ b/py/setup.cfg
@@ -1,5 +1,5 @@
 [flake8]
-exclude = .tox,docs/source/conf.py
+exclude = .tox,docs/source/conf.py,*venv
 ignore = E501
 
 [tool:pytest]


py/setup.cfg
```diff
@@ -1,5 +1,5 @@
 [flake8]
-exclude = .tox,docs/source/conf.py
+exclude = .tox,docs/source/conf.py,*venv
```
titusfortner, line 2: This didn't actually work for me. I think this did: `./venv`

titusfortner (changes requested):
Can you verify that what you have is supposed to work? (or why it didn't for me)

py/setup.cfg
AutomatedTester: has this been solved?

symonk: I can't recreate this; where does your virtual env typically reside, can you share the full path? I tested it yesterday and couldn't break it

symonk:
@titusfortner This landed in another piece of work of mine, if you come across issues using `tox -e linting` with delays and or files being parsed from a virtualenv please reach out to me and we can figure it out, this as-is works perfectly for me so I deemed it ok to bring in for now, closing this in favour of #10662

PR #10653, opened 2022-05-16
[py] fix: SecurityError: Invalid Domain problem

SinghHrmn:
### Description

- Fixed the Invalid Domain error while executing JS in Virtual
Authenticator tests.
- Added more tests for Virtual Authenticator.
- Updated SimpleWebserver to use 'localhost' when serving a page when `localhost` check is `True`. 

### Types of changes

### Checklist

### DIFF
--- a/py/test/selenium/webdriver/common/virtual_authenticator_tests.py
+++ b/py/test/selenium/webdriver/common/virtual_authenticator_tests.py
@@ -111,7 +111,7 @@ def extract_raw_id(response):
 @pytest.mark.xfail_remote
 def test_add_and_remove_virtual_authenticator(driver, pages):
     driver = create_rk_disabled_ctap2_authenticator(driver)
-    driver.get(pages.url("virtual-authenticator.html"))
+    driver.get(pages.url("virtual-authenticator.html", localhost=True))
 
     result = driver.execute_async_script("registerCredential().then(arguments[arguments.length - 1]);")
     assert result.get('status', '') == 'OK'
@@ -130,7 +130,7 @@ def test_add_and_remove_virtual_authenticator(driver, pages):
 def test_add_and_remove_non_resident_credentials(driver, pages):
     driver = create_rk_disabled_ctap2_authenticator(driver)
 
-    driver.get(pages.url("virtual-authenticator.html"))
+    driver.get(pages.url("virtual-authenticator.html", localhost=True))
 
     assert driver.virtual_authenticator_id is not None
 
@@ -153,7 +153,7 @@ def test_add_and_remove_non_resident_credentials(driver, pages):
 @pytest.mark.xfail_remote
 def test_add_non_resident_credential_when_authenticator_uses_u2f_protocol(driver, pages):
     driver = create_rk_disabled_u2f_authenticator(driver)
-    driver.get(pages.url("virtual-authenticator.html"))
+    driver.get(pages.url("virtual-authenticator.html", localhost=True))
 
     base64_pk = '''
     MIGHAgEAMBMGByqGSM49AgEGCCqGSM49AwEHBG0wawIBAQQg8_zMDQDYAxlU-Q
@@ -178,7 +178,7 @@ def test_add_non_resident_credential_when_authenticator_uses_u2f_protocol(driver
 @pytest.mark.xfail_remote
 def test_add_resident_credential_not_supported_when_authenticator_uses_u2f_protocol(driver, pages):
     driver = create_rk_enabled_u2f_authenticator(driver)
-    driver.get(pages.url("virtual-authenticator.html"))
+    driver.get(pages.url("virtual-authenticator.html", localhost=True))
 
     base64_pk = '''
     MIGHAgEAMBMGByqGSM49AgEGCCqGSM49AwEHBG0wawIBAQQg8_zMDQDYAxlU-Q
@@ -208,7 +208,7 @@ def test_add_resident_credential_not_supported_when_authenticator_uses_u2f_proto
 @pytest.mark.xfail_remote
 def test_get_credentials(driver, pages):
     driver = create_rk_enabled_ctap2_authenticator(driver)
-    driver.get(pages.url("virtual-authenticator.html"))
+    driver.get(pages.url("virtual-authenticator.html", localhost=True))
 
     # Register a Resident Credential
     response1 = driver.execute_async_script('''
@@ -265,7 +265,7 @@ def test_get_credentials(driver, pages):
 @pytest.mark.xfail_remote
 def test_remove_credential_by_raw_Id(driver, pages):
     driver = create_rk_disabled_u2f_authenticator(driver)
-    driver.get(pages.url("virtual-authenticator.html"))
+    driver.get(pages.url("virtual-authenticator.html", localhost=True))
 
     # register a credential
     response = driver.execute_async_script(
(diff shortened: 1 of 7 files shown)


py/test/selenium/webdriver/common/webserver.py
```diff
@@ -163,7 +164,10 @@ def stop(self):
         with contextlib.suppress(IOError):
             _ = urllib_request.urlopen(f"http://{self.host}:{self.port}")
 
-    def where_is(self, path) -> str:
+    def where_is(self, path, localhost=False) -> str:
+        # True force serve the page from localhost
+        if localhost:
+            return f"http://localhost:{self.port}/{path}"
```
AutomatedTester:
```suggestion
            return f"http://{DEFAULT_HOST}:{self.port}/{path}"
```

py/test/selenium/webdriver/common/virtual_authenticator_tests.py
```diff
@@ -144,109 +151,246 @@ def test_add_and_remove_non_resident_credentials(driver, pages):
...
+    for credential in credentials:
+        # Using startswith because there can be padding difference '==' or '=' in the end
+        if credential.id.startswith(extract_id(response1)):
+            credential1: Credential = credential
+        elif credential.id.startswith(extract_id(response2)):
+            credential2: Credential = credential
+        else:
+            assert False, "Unknown credential"
+
+    if credential1.is_resident_credential:
```
AutomatedTester: What happens if these are not True? Should the test fail?

AutomatedTester: same question for below where we have `assert True`

py/test/selenium/webdriver/common/virtual_authenticator_tests.py
```diff
@@ -144,109 +151,246 @@ def test_add_and_remove_non_resident_credentials(driver, pages):
...
 
-    login = driver.find_element(By.ID, "login-button")
-    WebDriverWait(driver, 50).until(lambda x: x.find_element(By.ID, "login-button").is_displayed())
-    login.click()
+    driver.get(pages.url("virtual-authenticator.html", localhost=True))
 
-    WebDriverWait(driver, 40).until(lambda x: x.find_element(By.CLASS_NAME, "col-lg-12").is_displayed())
+    # Register 2 credentials
+    response1 = driver.execute_async_script(
+        "registerCredential().then(arguments[arguments.length - 1]);"
```
AutomatedTester: let's move this JS to a variable and then we can simplify the code here and below. It means any future changes only need to happen in 1 place

AutomatedTester (changes requested):

py/test/selenium/webdriver/common/virtual_authenticator_tests.py
SinghHrmn: yes it should fail. I'll add the assert False in the else case. Thanks for this I formatted code according to flake8 forget to add else block

py/test/selenium/webdriver/common/virtual_authenticator_tests.py
SinghHrmn: Sure

SinghHrmn:
@AutomatedTester @pujagani Code Review changes Done!

py/test/selenium/webdriver/common/virtual_authenticator_tests.py
```diff
@@ -144,109 +152,240 @@ def test_add_and_remove_non_resident_credentials(driver, pages):
...
+    credential = Credential.create_resident_credential(
         bytearray({1, 2, 3, 4}),
         "localhost",
-        b64decode(BASE64__ENCODED_PK),
+        bytearray({1}),
+        urlsafe_b64decode(base64_pk),
         0,
     )
+    try:
+        driver.add_credential(credential)
```
AutomatedTester, line 202: we should use [`pytest.raises`](https://docs.pytest.org/en/6.2.x/assert.html#assertions-about-expected-exceptions)

py/test/selenium/webdriver/common/virtual_authenticator_tests.py
```diff
@@ -144,109 +152,240 @@ def test_add_and_remove_non_resident_credentials(driver, pages):
...
+    for credential in credentials:
+        # Using startswith because there can be padding difference '==' or '=' in the end
+        if credential.id.startswith(extract_id(response1)):
+            credential1: Credential = credential
+        elif credential.id.startswith(extract_id(response2)):
+            credential2: Credential = credential
+        else:
+            assert False, "Unknown credential"
+
+    if credential1.is_resident_credential:
```
AutomatedTester: What I meant is that we should have something like

```suggestion
    assert credential1.is_resident_credential == True, "Credential1 should be resident credential"
```

py/test/selenium/webdriver/common/virtual_authenticator_tests.py
```diff
@@ -144,109 +152,240 @@ def test_add_and_remove_non_resident_credentials(driver, pages):
...
+        assert True
+    else:
+        assert False, "Credential1 should be resident credential"
+    if credential1.private_key is not None:
```
AutomatedTester: This would be 
```suggestion
    assert credential1.private_key is not None
```

AutomatedTester (changes requested):
Update your asserts as I mentioned in some of the comments to make it look neater

py/test/selenium/webdriver/common/virtual_authenticator_tests.py
SinghHrmn: I was writing the tests like this. But Flake8 complains about writing tests like this. 
for code 
```python
assert credential1.is_resident_credential == True, "Credential1 should be resident credential"
```
Error
```json
[{
	"resource": "/Users/harmandeepsingh/BrowserStack/selenium/py/test/selenium/webdriver/common/virtual_authenticator_tests.py",
	"owner": "python",
	"code": "E712",
	"severity": 8,
	"message": "comparison to True should be 'if cond is True:' or 'if cond:'",
	"source": "flake8",
	"startLineNumber": 242,
	"startColumn": 47,
	"endLineNumber": 242,
	"endColumn": 47
}]
```

py/test/selenium/webdriver/common/virtual_authenticator_tests.py
SinghHrmn: Same as above

py/test/selenium/webdriver/common/virtual_authenticator_tests.py
SinghHrmn: Maybe we can skip flake8 in tests to clean the code.

SinghHrmn: @AutomatedTester comparing it with `==` has flake8 problem. But if I compare with `is` then flake8 is happy. So I've changed accordingly.

symonk: I don't like the idea of skipping flake8 in tests, booleans are singleton values and should be compared with `is` on identity (same with `None`) if not using the built-in implicit truthy checks.  You can just `assert credential1.is_resident_credential` here, checking the value is true is implicit

py/test/selenium/webdriver/common/virtual_authenticator_tests.py
```diff
@@ -88,15 +94,36 @@ def create_rk_disabled_authenticator(driver) -> WebDriver:
...
+def extract_id(response):
+    return response.get("credential", {}).get("id", "")
+
+
+def extract_raw_id(response):
+    return response.get("credential", {}).get("rawId", "")
+
+
+def not_allowed_error_in(response) -> bool:
+    if response.get("status", "").startswith("NotAllowedError"):
```
symonk: can just return the `response.get(...)` call here without the `if`

AutomatedTester (approved):

PR #10657, opened 2022-05-16
[py]: Align error handling to `W3C` 

symonk:
Converts the error codes into a `Tuple` as they are immutable and adds appropriate type hints for them.
Removes unused vars `screen` & `stacktrace` as they are redefined later when used.


py/selenium/webdriver/remote/errorhandler.py
```diff
@@ -118,8 +118,6 @@ def check_response(self, response: Dict[str, Any]) -> None:
             return
         value = None
         message = response.get("message", "")
-        screen: str = response.get("screen", "")
```
AutomatedTester: I think the assignments to `None` later on is the issue not these lines.

AutomatedTester:
Looking at this again we can remove the numbers, that's a hangover from JWP that we are trying to remove anyway

symonk:
Refactoring the error handling is a bigger task than initially thought; all the tests are parameterised for the `(int, str)` so the new str only is iterable and unpacking itself into hundreds of parameterised tests :smile: need to keep grinding on this one gradually!

py/selenium/webdriver/remote/errorhandler.py
```diff
@@ -114,12 +113,10 @@ def check_response(self, response: Dict[str, Any]) -> None:
...
         """
         status = response.get('status', None)
-        if not status or status == ErrorCode.SUCCESS:
+        if not status:  # covers `0` implicitly.
             return
         value = None
         message = response.get("message", "")
-        screen: str = response.get("screen", "")
-        stacktrace = None
         if isinstance(status, int):
```
symonk, line 120: are we saying this check is not possible any longer?

symonk:
I think the error handling code could use a rewrite to scrub some JWP code, I will do it as a follow up ticket.

py/selenium/webdriver/remote/errorhandler.py
```diff
@@ -47,55 +47,54 @@
...
 _VT = TypeVar("_VT")
 
 
+# Todo: In future rename this; it's no longer relevant
 class ErrorCode:
     """
     Error codes defined in the WebDriver wire protocol.
     """
     # Keep in sync with org.openqa.selenium.remote.ErrorCodes and errorcodes.h
-    SUCCESS = 0
```
symonk, line 56: not so simple; assigned in `remote_connection.py`

py/selenium/webdriver/remote/errorhandler.py
```diff
@@ -208,12 +209,10 @@ def check_response(self, response: Dict[str, Any]) -> None:
         if message == "" and 'message' in value:
             message = value['message']
 
-        screen = None  # type: ignore[assignment]
         if 'screen' in value:
             screen = value['screen']
 
-        stacktrace = None
-        st_value = value.get('stackTrace') or value.get('stacktrace')
+        st_value = value.get('stacktrace')
```
symonk, line 215: `stackTrace` was JWP stuff, I think safe to remove.  @titusfortner any idea if the `screen` code here can go? I only find references to it inside jwp specification docs

titusfortner: Yes you can; it was one of the valid responses for jwp: https://www.selenium.dev/documentation/legacy/json_wire_protocol/#failed-commands
but is not one of the valid responses for w3c: https://w3c.github.io/webdriver/#errors

PR #10659, opened 2022-05-17
[rb] - Improving Wait class error handling

luke-hill:
### Description
When re-polling during the `Wait#until` method. Ensure we don't trip up if the element is missing **OR** if the element is stale, as both are equally likely

### Motivation and Context
During the upgrade from capybara 3.35 to 3.37.1 this error starts showing a **lot** more. It is likely that as we are moving faster in better rubies, we are likely to hit that split second where the web element could be missing OR it could be in the process of appearing / changing and therefore register as stale.

I've classified this as a bugfix. As ideally missing element and stale element should live hand in hand.

### Types of changes

### Checklist

### DIFF
--- a/rb/CHANGES
+++ b/rb/CHANGES
@@ -1,3 +1,9 @@
+trunk (TBD)
+=========================
+
+Ruby:
+  * Fix Wait class tripping up during poll when a StaleElementReference would occur (#10659)
+
 4.1.0 (2021-11-22)
 =========================
 
--- a/rb/lib/selenium/webdriver/common/wait.rb
+++ b/rb/lib/selenium/webdriver/common/wait.rb
@@ -22,7 +22,7 @@ module Selenium
     class Wait
       DEFAULT_TIMEOUT  = 5
       DEFAULT_INTERVAL = 0.2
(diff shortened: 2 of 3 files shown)


titusfortner (changes requested):
What I think the other bindings do is to make this configurable. I think the right answer here is to give ignored an attr accessor, so user can set it if they want.

luke-hill:
I shall do that.

luke-hill:
Notwithstanding your point, which I agree with, are the defaults here meaningful / valid? Or should we amend them?

titusfortner:
I just noticed that you can set the list of exceptions to ignore as part of the initialization of the Wait. That should be sufficient for configuration.

luke-hill:
I think the issue was more should we amend the default exceptions. But yeh we never reached a consensus here.

Imo we should edit the default exceptions, as both are equally likely.

titusfortner:
No, we were never going to change the default away from just the `NoSuchElementError`, just like all the other language bindings. A stale element isn't going to become un-stale, so additional logic would be necessary for that to remotely make sense; it definitely should not be ignored by default.

luke-hill:
FYI That isn't true. As if you re-find something it can then not be stale. But given we didn't reach a consensus this can just be considered dead. Same as the logic for finding the element can also fail and then not fail under the same test conditions.

Issue #10335, opened 2022-02-07
[🚀 Feature]: Support WebDriver BiDi and "log.entryAdded" events (with Firefox)


whimboo:
### Feature and motivation

With the [WebDriver BiDi specification](https://w3c.github.io/webdriver-bidi/) moving forward it would be great to see support for this protocol being added to Selenium. In case of Firefox we already added the `log.entryAdded` event and relevant commands like `session.subscribe` to get relevant logging events for console API calls and Javascript exceptions on the page.

Given the high request of Selenium users for such an API it would be great to move the current implementation for Firefox which is based on CDP over to WebDriver BiDi. On our side we have the following bug on file:

https://bugzilla.mozilla.org/show_bug.cgi?id=1753991

CC'ing @AutomatedTester, @shs96c, @titusfortner, @diemol

### Usage example

From the user's perspective the API for logging events might not change and could probably the same as when using the CDP backend.

diemol:
Added tags for the other bindings so we can track progress.

whimboo:
@diemol who could potentially could give an update of the current implementation status of this feature in Selenium? Thanks.

titusfortner:
I don't think we want to track these additions in issues. We should use Puja's more systematic tracking system.

whimboo:
> I don't think we want to track these additions in issues. We should use Puja's more systematic tracking system.

Out of interest. How does that work?

PR #10660, opened 2022-05-17
[java] Add support to create a BiDi session

pujagani:

### Description

Add support to create a BiDi session

### Motivation and Context

Related to adding support for #10335 

Note: The code is using the current structure of DevTools implementation in Java. Though that causes a fair amount of code duplication, keeping the BiDi and DevTools implementation will make it easier to remove DevTools support in the future.

### Types of changes

### Checklist


diemol (approved):
Code duplication is not always ideal but I think it is the right approach so we can delete the CDP stuff later.

PR #10662, opened 2022-05-17
[py]: Additional types for element fetching methods

symonk:
Add additional types in some places that fetch elements
removed unnecessary `__ne__` implementation.

`WebElement` & `ShadowRoot` have a circular dependency, will address that one in future.


py/selenium/webdriver/remote/shadowroot.py
```diff
@@ -32,12 +32,9 @@ def __init__(self, session, id_):
     def __eq__(self, other_shadowroot):
         return self._id == other_shadowroot._id
 
-    def __hash__(self):
+    def __hash__(self) -> int:
         return int(md5_hash(self._id.encode("utf-8")).hexdigest(), 16)
 
-    def __ne__(self, other_shadowroot):
```
AutomatedTester, line 38: Why have you removed this method?

symonk: I removed it just as it's unnecessary with `__eq__` defined, it is obsolete for this use case

AutomatedTester (approved):

PR #10663, opened 2022-05-18
[JS] feat: Added virtual authenticator

TamsilAmani:

### Description

Relates #10541
- Added virtual authenticator for js.
- Added tests for the same.

### Types of changes

### Checklist


common/src/web/virtualAuthenticator.html
```diff
@@ -0,0 +1,73 @@
+<!DOCTYPE html>
```
AutomatedTester: Please can we reuse the file that was used in the `py` version of this feature?

javascript/node/selenium-webdriver/lib/test/fileserver.js
```diff
@@ -106,6 +106,7 @@ const Pages = (function () {
   addPage('webComponents', 'webComponents.html')
   addPage('xhtmlTestPage', 'xhtmlTest.html')
   addPage('uploadInvisibleTestPage', 'upload_invisible.html')
+  addPage('virtualAuthenticator', 'virtualAuthenticator.html')
```
AutomatedTester: Please update to match the same file as the one created in the `py` PR.

AutomatedTester (changes requested):

javascript/node/selenium-webdriver/lib/test/fileserver.js
TamsilAmani: I have made the changes. But I can't make a PR before [https://github.com/SeleniumHQ/selenium/pull/10653] gets merged. The changes made in `virtual-authenticator.html` are required to successfully run the test.

TamsilAmani: @AutomatedTester required changes have been made.

javascript/node/selenium-webdriver/lib/webdriver.js
```diff
@@ -1515,6 +1520,110 @@ class WebDriver {
...
+   * Removes a previously added virtual authenticator. The authenticator is no
+   * longer valid after removal, so no methods may be called.
+   */
+  async removeVirtualAuthenticator() {
+    await this.execute(
+      new command.Command(
+        command.Name.REMOVE_VIRTUAL_AUTHENTICATOR
+      ).setParameter('authenticatorId', this.authenticatorId_)
+    )
+    this.authenticatorId_ = null
```
pujagani, line 1554: What if the user wants to register multiple authenticators?

TamsilAmani: Right. I'll fix that.

pujagani: The global variable led to that thought. The question is more of will there ever be such a situation. It does not intend to question the current implementation. I think we can let it be how it is. I thought about it and I am unable to foresee a situation that needs multiple authenticators. We can tackle it if and when the need arises.

common/src/web/virtualAuthenticator.html
TamsilAmani: I have fixed this.

javascript/node/selenium-webdriver/lib/virtual_authenticator.js
```diff
@@ -0,0 +1,230 @@
...
+
+/**
+ * Options for the creation of virtual authenticators.
+ * @see http://w3c.github.io/webauthn/#sctn-automation
+ */
+class VirtualAuthenticatorOptions {
+
+  static Protocol = {
+    "CTAP2": 'ctap2',
+    "U2F": 'ctap1/u2f',
```
harsha509, line 28: Are we not using `ctap2_1`  ?

javascript/node/selenium-webdriver/lib/virtual_authenticator.js
```diff
@@ -0,0 +1,230 @@
...
+    "USB": 'usb',
+    "NFC": 'nfc',
+    "INTERNAL": 'internal',
+  }
+
+  /**
+   * Constructor to initialise VirtualAuthenticatorOptions object.
+   */
+  constructor() {
+    this._protocol = VirtualAuthenticatorOptions.Protocol["CTAP2"]
```
harsha509, line 42: Why are we hardcoding the protocol in here ?

According to W3C the default value is none right ? Cant we pass params in constructor and make use of it for protocol and transport ?

like ,
```
this._protocol = VirtualAuthenticatorOptions.Protocol[`${protocol}`] 
```

javascript/node/selenium-webdriver/lib/virtual_authenticator.js
TamsilAmani: In Java, it's still `ctap2`
https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/virtualauthenticator/VirtualAuthenticatorOptions.java#L31

In Python also, it is `ctap2`
https://github.com/SeleniumHQ/selenium/blob/trunk/py/selenium/webdriver/common/virtual_authenticator.py#L36

But according to the official documentation, we are using `ctap2_1`, along with `ctap2`.
https://w3c.github.io/webauthn/#sctn-automation-virtual-authenticators **_(11.3)_**
I followed what Java has implemented.

javascript/node/selenium-webdriver/lib/virtual_authenticator.js
```diff
@@ -0,0 +1,230 @@
...
+
+  setProtocol(protocol) {
+    this._protocol = protocol
+  }
+
+  getTransport() {
+    return this._transport
+  }
+
+  setTransport(transport) {
```
harsha509, line 62: Adding virtual auth is one time activity right!

So, can we set all params in constructors itself ? or do we need to change value of any param  in between ? 

If get and set are required can we users getters and setters in here!

javascript/node/selenium-webdriver/lib/virtual_authenticator.js
```diff
@@ -0,0 +1,230 @@
...
+ */
+class Credential {
+  constructor(
+    credentialId,
+    isResidentCredential,
+    rpId,
+    userHandle,
+    privateKey,
+    signCount
+  ) {
```
harsha509, line 123: may be similar changes for vitualAuth can be applied in here too!

javascript/node/selenium-webdriver/lib/virtual_authenticator.js
TamsilAmani: This approach seems good. I don't know if the user needs to modify the authenticator later (will discuss with @pujagani for this part)

TamsilAmani: Talked to @AutomatedTester about it. Virtual authenticator is set for the duration of a session. So I don't think we would need to modify it. I'll push the changes to fix it.

javascript/node/selenium-webdriver/lib/virtual_authenticator.js
TamsilAmani: I think you're right. Passing the parameters in the constructor itself would be more logical. I'll fix this.

javascript/node/selenium-webdriver/lib/virtual_authenticator.js
TamsilAmani: The constructor for Credential is already having parameters. What changes do you suggest here?

javascript/node/selenium-webdriver/lib/virtual_authenticator.js
SinghHrmn: @harsha509 Virtual Authenticator is only available for chromium bases browsers like Chrome/Egde and if we look for Virtual Authenticator under WebAuthn in `inspect > more tools > WebAuthn` we can see the default options are the one that we have in the constructor. I think that is the reason why it was hardcoded in Java. But we also have getter/setter for them if we do need to update. And i think it is a good approach. 

Screenshot 2022-05-31 at 8 57 30 AM


One can directly initialise VirtualAuthenticatorOptions with default settings. 
```python
    options = VirtualAuthenticatorOptions()
    driver.add_virtual_authenticator(options)
```

harsha509: Makes sense!

I was looking it in here https://w3c.github.io/webauthn/#sctn-automation-add-virtual-authenticator. As it says default is none!

SinghHrmn: Default is None but it is not a valid value so the user must select some value from `"ctap1/u2f", "ctap2", "ctap2_1"`. Adding a valid value as default makes sense.

javascript/node/selenium-webdriver/lib/virtual_authenticator.js
harsha509: yeah, not required in here!

javascript/node/selenium-webdriver/lib/virtual_authenticator.js
harsha509: Looking into UI also has two! and i dont see `ctap2_1 `.  may be we can add it in future when required! As all other language bindings has same implementation, we can ignore for now!

javascript/node/selenium-webdriver/lib/virtual_authenticator.js
harsha509: To be in sync with other language bindings implementation, keeping changes as it is!

javascript/node/selenium-webdriver/lib/virtual_authenticator.js
harsha509: To be in sync with other language bindings implementation, keeping changes as it is!

harsha509 (approved):
Thank you @TamsilAmani !

Issue #8229, opened 2020-03-27
Empty logs for Selenium 4 alpha 5 when using RemoteWebdriver


szamacz:
## 🐛 Bug Report

When using latest Selenium 4 image with RemoteWebdriver and chrome standalone  4.0.0-alpha-5-20200326 docker image all kind of logs are empty. Problem does not exists in case of using local chrome browser (ver  80.0.3987.149 ).

## To Reproduce

1. Run standalone chrome docker image from docker hub with selenium 4
2. Run tests against image using RemoteWebdriver
3. Get logs
AvailableLogTypes  count returns 0

Detailed steps to reproduce the behavior:

## Expected behavior

There should be 5 types of logs, and at least 2 of them shouldn't be empty (Browser and Driver) - like in case of running this against regular Chrome browser.

## Test script or set of commands reproducing this issue

    public class RemoteWebriverTests : IDisposable
    {
        private readonly IWebDriver _driver;

        public RemoteWebriverTests()
        {
            var chromeOptions = new ChromeOptions();
            _driver = new RemoteWebDriver(new Uri("http://localhost:4444/"), chromeOptions);
        }

        [Fact]
        public void GetLogsFromChrome()
        {
            _driver.Navigate().GoToUrl("http://www.youtube.com");
            Thread.Sleep(5000);
            Assert.True(_driver.Manage().Logs.AvailableLogTypes.Count > 0);
        }

        public void Dispose()
        {
            _driver.Quit();
        }
    }

## Environment

OS: Windows 10
Browser: Chrome
Browser version: 80.0.3987.149
Docker image version: 4.0.0-alpha-5-20200326
Browser Driver version: 80.0.3987.10600
Language Bindings version: .net core 3.1
Selenium Grid version : 4.0.0-alpha-5

szamacz:
Issue reproducible also on version 4.0.0-alpha6-20200609. Are there any plans to look at this?

barancev:
Works for me with docker image selenium/standalone-chrome:4.0.0-alpha6-20200609 and java binding 4.0.0-alpha6, so it must be something .Net specific.

JunTaoLuo:
Hey @barancev we are running into this problem as well in our test environment. In fact, with Chrome 85, we had to update to a newer version of Selenium and ended up running into this bug. Is there any plans on addressing this soon?

john1506:
this is also affecting us - would just like to chime in

Dreamescaper:
[...]

PR #10671, opened 2022-05-20
[dotnet] allow RemoteWebDriver to access Selenium logs

titusfortner:
There are 3 options I can see for quickly fixing #8229, none of them are ideal.

1. Require users to subclass RemoteWebDriver and add implementation for `ISupportsLog` which I *think should "just work" (at least it did in my quick test of it).
```cs
public class RemoteChromeDriver : RemoteWebDriver, ISupportsLogs
```
2. This PR, which would allow a user to do this:
```cs
var customCommandDriver = driver as ICustomDriverCommandExecutor;
customCommandDriver.RegisterCustomDriverCommands(ChromeDriver.CustomCommandDefinitions);

ReadOnlyCollection<string> logTypes = driver.Manage().Logs.AvailableLogTypes;
driver.Url = "https://selenium.dev/selenium/web/errors.html";
driver.FindElement(By.CssSelector("input")).Click();
ReadOnlyCollection<LogEntry> results = driver.Manage().Logs.GetLog("browser");
```
3. Can remove some of the uglier parts of this PR and put more onus on user:
```cs
IReadOnlyDictionary<string, CommandInfo> logCommands = new Dictionary<string, CommandInfo>
{
    {
        DriverCommand.GetAvailableLogTypes,
        new HttpCommandInfo(HttpCommandInfo.GetCommand, "/session/{sessionId}/se/log/types")
    },
    {
        DriverCommand.GetLog,
        new HttpCommandInfo(HttpCommandInfo.PostCommand, "/session/{sessionId}/se/log")
    }
};

var customCommandDriver = driver as ICustomDriverCommandExecutor;
customCommandDriver.RegisterCustomDriverCommands(logCommands);

ReadOnlyCollection<string> logTypes = driver.Manage().Logs.AvailableLogTypes;
driver.Url = "https://selenium.dev/selenium/web/errors.html";
driver.FindElement(By.CssSelector("input")).Click();
ReadOnlyCollection<LogEntry> results = driver.Manage().Logs.GetLog("browser");
```

The existing code allows driver classes for drivers that haven't implemented the logging endpoints, to call the log methods, but it won't actually send the commands. No error that the method isn't actually doing anything.

This PR would always send the command but will swallow the not-implemented exception if the driver commands were not added. So same end result, but slightly less efficient, but if a user is calling for something that never returns anything anyway, maybe that doesn't matter from an implementation standpoint.

Any feedback from anyone would be great on this, thanks.

### DIFF
--- a/dotnet/src/webdriver/Logs.cs
+++ b/dotnet/src/webdriver/Logs.cs
@@ -16,6 +16,7 @@
 // limitations under the License.
 // </copyright>
 
+using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 
@@ -27,7 +28,6 @@ namespace OpenQA.Selenium
     public class Logs : ILogs
     {
         private WebDriver driver;
-        private bool isLogSupported = false;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RemoteLogs"/> class.
@@ -36,7 +36,6 @@ namespace OpenQA.Selenium
         public Logs(WebDriver driver)
         {
             this.driver = driver;
-            this.isLogSupported = (this.driver as ISupportsLogs) != null;
         }
 
         /// <summary>
@@ -47,7 +46,7 @@ namespace OpenQA.Selenium
             get
             {
                 List<string> availableLogTypes = new List<string>();
-                if (this.isLogSupported)
+                try
                 {
                     Response commandResponse = this.driver.InternalExecute(DriverCommand.GetAvailableLogTypes, null);
                     object[] responseValue = commandResponse.Value as object[];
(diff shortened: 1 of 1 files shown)


titusfortner:
It's easier to add things than take them away, so I want to do the easier code/harder user scenario for this and go from there....

titusfortner:
Scaled down the PR for minimal change
Existing code prevents executing log methods unless the instance implements the interface.
Only Chromium implements the interface, and Remote can't implement it, so even using the Custom Driver Command Executor to add the commands, they were blocked.
So, this removes the block, and if it is a local Chromium browser, the methods will be automatically added and it will work; and if it is a Remote driver and the user has added the commands, it will work; and if neither of those things is true, the code swallows the error (which effectively current behavior).

PR #10672, opened 2022-05-20
Update submit function

titusfortner:
There was a bug that Java fixed that the other languages didn't.
I also noticed that Java does everything in one wire call instead of two like the other languages, so I updated the other languages to match Java.

Also, when Submit method failed in Java, it's returning an unhelpful JavaScriptError.
When submit method failed in the other languages, it returned an unhelpful NoSuchElementError which is confusing since the user didn't think they were searching for anything, and the locator is confusing unless they understand the implementation.

This PR uses: `To submit an element, it must be nested inside a form element`

Ruby & Java both have an `UnsupportedOperation` error, the other languages I went with the generic error.

If anyone has preferences for a different message or different exception class for any of these, let me know.

@harsha509 yeah, I made sure the existing submit test passed, but I didn't add any of the other tests.

### DIFF
--- a/py/selenium/webdriver/remote/webelement.py
+++ b/py/selenium/webdriver/remote/webelement.py
@@ -26,12 +26,11 @@ from abc import ABCMeta
 from io import BytesIO
 from typing import Union
 
-from selenium.common.exceptions import WebDriverException
+from selenium.common.exceptions import WebDriverException, JavascriptException
 from selenium.webdriver.common.by import By
 from selenium.webdriver.common.utils import keys_to_typing
 from .command import Command
 from .shadowroot import ShadowRoot
-
 
 # TODO: When moving to supporting python 3.9 as the minimum version we can
 # use built in importlib_resources.files.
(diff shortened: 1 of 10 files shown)


diemol (dismissed):
Thanks, @titusfortner, Java looks fine to me.

py/selenium/webdriver/remote/webelement.py
```diff
@@ -91,11 +90,20 @@ def click(self) -> None:
...
+                 "}\n" \
+                 "if (!form) { throw Error('Unable to find containing form element'); }\n" \
+                 "if (!form.ownerDocument) { throw Error('Unable to find owning document'); }\n" \
+                 "var e = form.ownerDocument.createEvent('Event');\n" \
+                 "e.initEvent('submit', true, true);\n" \
+                 "if (form.dispatchEvent(e)) { HTMLFormElement.prototype.submit.call(form) }\n"
+
+        try:
+            self._parent.execute_script(script, self)
+        except JavascriptException:
```
AutomatedTester, line 105: Not a blocker but this is a change in exception thrown and we should be aware of that

titusfortner: We could continue to throw the JS error and just update the message if you would prefer that. JS error just seemed confusing/non-obvious to me.

symonk: I think it's ideal to keep the same as before (given it's not a result of a completely broken functionality?) Could just raise the same with a new message ` raise JavascriptException(msg) from None` (assuming it raises that previously)

titusfortner: This is what I get for trying to reply via my phone. I thought we were talking Java bindings.

The current Python code raises an "Element Not Found" error for this scenario. Updating to the all-JS approach that Java is currently using gives an unhelpful JavaScript error message. This line isn't what is changing existing behavior, it is just tailoring the new behavior.

PR #10676, opened 2022-05-20
Dict comprehension

machinexa2:
https://www.programiz.com/python-programming/dictionary-comprehension
Since it's basically copying dictionary there might be more better or faster way to do it

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist

- [ ] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #10677, opened 2022-05-20
Alternativ to dict comp

machinexa2:
See previous PR

### DIFF
--- a/py/selenium/webdriver/remote/webdriver.py
+++ b/py/selenium/webdriver/remote/webdriver.py
@@ -157,9 +157,7 @@ def create_matches(options: List[BaseOptions]) -> Dict:
                     if opts[min_index][kys] == opts[i + 1][kys]:
                         samesies.update({kys: opts[min_index][kys]})
 
-    always = {}
-    for k, v in samesies.items():
-        always[k] = v
+    always = samesies.copy()
 
     for i in opts:
         for k in always.keys():


machinexa2:
https://github.com/SeleniumHQ/selenium/pull/10676

machinexa2:
Both does the same job. Merge the PR which is best. I think copy() one is best because it's fastest but does it break any functionality???? Let me know which you merged and I will close the other

py/selenium/webdriver/remote/webdriver.py
```diff
@@ -157,9 +157,7 @@ def create_matches(options: List[BaseOptions]) -> Dict:
                     if opts[min_index][kys] == opts[i + 1][kys]:
                         samesies.update({kys: opts[min_index][kys]})
 
-    always = {}
-    for k, v in samesies.items():
-        always[k] = v
+    always = samesies.copy()
```
symonk, line 158: copy is likely better performance than the comprehension counter part; however is a _shallow_ copy sufficient here? (I'd need to look at the code, no time right now)

machinexa2: Sure, letme know if it can be improved. I also saw that this new local variable isnt much used so maybe it can be completely removed while using samesies variables but i need to look at code if it messes up something

machinexa2: Since always is a local variable of the function used in 3 places maybe this unnecessary dict copy can be replaced by samesies. 
```
always = {}
    for k, v in samesies.items():
        always[k] = v

    for i in opts:
        for k in always.keys():
            del i[k]

    capabilities["capabilities"]["alwaysMatch"] = always
```
To
```
#always = {}
#    for k, v in samesies.items():
#       always[k] = v

    for i in opts:
        for k in samesies.keys():
            del i[k]

    capabilities["capabilities"]["alwaysMatch"] = samesies
```

machinexa2: @symonk

machinexa2:
changed to list comprehensin :)

py/selenium/webdriver/remote/webdriver.py
AutomatedTester: I think the later is probably what we want

machinexa2:
Imma make changes a bit later, because it got merged in trunk the git is becoming difficult to use

PR #10682, opened 2022-05-23
[py] extend typing options for get_property()

### LINKED ISSUE #10624: [🐛 Bug]: Wrong type for WebElement.get_property
### What happened?

This is similar issue to #10623 

In this case it's not so straightforward. The `WebElement.get_property` declare return value type as `str`. 
It is true in many cases but if you try to get property 'parentElement' (see example below), then the return value is `WebElement`.
But mypy (PyCharm and other linters) complains that result of get_property is always `str`.
Can you declare return type of `get_property` as `Union` of all possible types, which the method can return?

```shell
from chromedriver_binary.utils import get_chromedriver_path


def test():
  driver = webdriver.Chrome(executable_path=get_chromedriver_path() + '/chromedriver')
  driver.get('http://www.example.com/')
  element = driver.find_element(by=By.TAG_NAME, value='h1')
  parentElement: WebElement = element.get_property('parentElement')
  print(type(parentElement))


if __name__ == '__main__':
  test()
```

### Relevant log output

```shell
no log needed
```

### Operating System

Gentoo Linux

### Selenium version

Python 4.1.5

not relevant

not relevant

_No response_

eNcacz:
fixes #10624

Extend list of types which can be returned byt `get_property()` method.

### Description

Current typing options does not cover all possible types returned by get_property()

For example

* `clientHeight` property is returned as int
* `fooBar` property is returned as NoneType
* `childNodes` property is returned as list

### Motivation and Context

Linters (like mypy or PyCharm) complains about `get_property()` usage when we expect one of mentioned data type.

### Types of changes

### Checklist

### DIFF
--- a/py/selenium/webdriver/remote/webelement.py
+++ b/py/selenium/webdriver/remote/webelement.py
@@ -24,6 +24,7 @@ import warnings
 import zipfile
 from abc import ABCMeta
 from io import BytesIO
+from typing import Optional
 
 from selenium.common.exceptions import WebDriverException, JavascriptException
 from selenium.webdriver.common.by import By
@@ -108,7 +109,7 @@ class WebElement(BaseWebElement):
         """Clears the text if it's a text entry element."""
         self._execute(Command.CLEAR_ELEMENT)
 
-    def get_property(self, name) -> str | bool | WebElement | dict:
+    def get_property(self, name) -> Optional[str | int | bool | WebElement | dict | list]:
         """
         Gets the given property of the element.
 


py/selenium/webdriver/remote/webelement.py
```diff
@@ -101,7 +101,7 @@ def clear(self) -> None:
         """Clears the text if it's a text entry element."""
         self._execute(Command.CLEAR_ELEMENT)
 
-    def get_property(self, name) -> Union[str, bool, WebElement, dict]:
+    def get_property(self, name) -> Optional[Union[str, int, bool, WebElement, dict, list]]:
```
symonk: I'm not sure this can return `None`

eNcacz: It return `None` when invalid property name is used. Please read the PR description for more info.

symonk: my apologies, i was thinking about element finding rather than property fetching!

AutomatedTester (dismissed):

AutomatedTester (approved):

titusfortner:
@eNcacz looks like we've updated to Python 3 syntax for this, can you update this PR so we can merge it? Thanks!

PR #10695, opened 2022-05-26
fix: update ExpectedCondition to extend java util Function

### LINKED ISSUE #10606: [🐛 Bug]: wait.until is not usable
### What happened?

My "wait.until" function worked fine yesterday, but all a sudden now is not runnable.

Error message: reason: no instance(s) of type variable(s) V exist so that ExpectedCondition<WebElement> conforms to Function<? super WebDriver, V>

I have looked it up online, and people said it is related to the inconsistent versions of Guava and Selenium. However, I have already checked my project structure, and there are no other extra versions of Guava and Selenium.

I am not sure how to reproduce since I don't know what's wrong. 
The only thing I can provide is what I am using. Sorry and thanks.

Please see my pom.xml below:

```shell     
   <dependency>
            <groupId>org.seleniumhq.selenium</groupId>
            <artifactId>selenium-java</artifactId>
            <version>4.1.3</version>
        </dependency>

  <dependency>
            <groupId>com.google.guava</groupId>
            <artifactId>guava</artifactId>
            <version>31.1-jre</version>
     </dependency>

```

The code I have run:
```shell
public void visibilityOfElementLocatedByXpath(String xpath){
        WebDriverWait wait = new WebDriverWait(driver, Duration.ofSeconds(10));
        wait.until(ExpectedConditions.visibilityOfElementLocated(By.xpath(xpath)));
    }
```

In addition, I am using JAVA 17, maven project with IntelliJ IDEA with MacBook Air (M1,2020).
```
java: method until in class org.openqa.selenium.support.ui.FluentWait<T> cannot be applied to given types;
  required: java.util.function.Function<? super org.openqa.selenium.WebDriver,V>
  found:    org.openqa.selenium.support.ui.ExpectedCondition<org.openqa.selenium.WebElement>
  reason: cannot infer type-variable(s) V
    (argument mismatch; org.openqa.selenium.support.ui.ExpectedCondition<org.openqa.selenium.WebElement> cannot be converted to java.util.function.Function<? super org.openqa.selenium.WebDriver,V>)
```

### Operating System

MacBook Air (M1,2020) - macOS Monterey 12.2.1

### Selenium version

4.1.3

Not-related

Not-related

No

ammmze:
### Description
Some builds of guava have `Function` that does NOT extend the java native function. This causes `ExpectedCondition` instances to be incompatible the various wait `until` methods that were recently updated to accept `java.util.function.Function`. See     more [here](https://github.com/SeleniumHQ/selenium/issues/10606#issuecomment-1136382614).

### Motivation and Context
Fixes #10606

### Types of changes

### Checklist

- [ ] ~My change requires a change to the documentation.~ N/A
- [ ] ~I have updated the documentation accordingly.~ N/A
- [ ] ~I have added tests to cover my changes.~ Not testable unless we are running with a different version of guava (specifically the "android" builds because those are java 7 compatible and therefore don't have the `java.util.function.Function` interface iirc)
- [ ] All new and existing tests passed. ... having issues getting some tests working. On the small tests had one failure mentioning no port available, but all the others passed. I don't recall what the large tests failed with, but it was something with the StressTest or whatever it was called.


ammmze:
Hmm...all the failing tests are ruby ones...surely those are not caused by this java change 🤷🏻‍♂️

pujagani:
The changes look good to me for now. But I think we should try to move away from Guava and implement things Java way in the long run. It will be a big change and my concern is how it will impact the users. But we can consider it for 4.3 or 4.4.

pujagani (approved):

ammmze:
> The changes look good to me for now. But I think we should try to move away from Guava and implement things Java way in the long run. It will be a big change and my concern is how it will impact the users. But we can consider it for 4.3 or 4.4.

Agreed! For now, this is just a backwards compatible change that should resolve things that were not intended to break when upgrading from selenium 3.x to 4.x (i.e. if a project used the wrong version of guava, then `ExpectedCondition` could no longer be used directly after upgrading to 4.x).

diemol (approved):
Thank you, @ammmze!

PR #10704, opened 2022-05-27
[py] overload log_path to accept kwargs for open

### LINKED ISSUE #10703: [🐛 Bug]: PySelenium doesn't support a named pipe as a log file
### What happened?

As part of [OpenWPM](https://github.com/openwpm/OpenWPM) we use Selenium to drive multiple Firefox browsers in parallel.

To capture, enrich and redirect the browser's logging output, we use named pipes that look like normal log files to Selenium.
However pipes don't support seeking so the need to be opened with `w` and not with `a+` as the [`firefox.Service` currently does](https://github.com/SeleniumHQ/selenium/blob/268161f84fb63e3e3e3177eb8b99b812ee3ae413/py/selenium/webdriver/firefox/service.py#L49). [We have previously monkeypatched this](https://github.com/openwpm/OpenWPM/blob/8464741bbfcdbd9aad54686d3f290a060c168488/openwpm/deploy_browsers/selenium_firefox.py#L79-L122) but that approach has led to us breaking every time a new Selenium release comes out. [A newer approach that tries to call `firefox.Service.__init__` and open the `log_file` afterwards](https://github.com/openwpm/OpenWPM/blob/45ffd5a752e88c24607776e62cd2d53c965effb4/openwpm/deploy_browsers/selenium_firefox.py#L80-L105) doesn't work at all, so I wanted to check if there was an appetite to address this issue upstream.

```shell
import os
from selenium.webdriver import Firefox
from selenium.webdriver.firefox.service import Service
pipe_name = "selenium_pipe.log"
os.mkfifo(pipe_name, 0o600)
service=Service(executable_path="firefox",log_path=pipe_name)
driver = Firefox(
    service=service
)
driver.get("https://example.com")
```

### Relevant log output

```shell
File "/home/vringar/selenium_mvp/demo.py", line 9, in <module>
    service=Service(executable_path="firefox",log_path=pipe_name)
File "/home/stefan/.conda/envs/openwpm/lib/python3.10/site-packages/selenium/webdriver/firefox/service.py", line 50, in __init__
    log_file = open(log_path, "a+") if log_path else None
io.UnsupportedOperation: File or stream is not seekable.
```

### Operating System

Linux

### Selenium version

Python 4.1.0

Firefox 100

GeckoDriver 0.30.0

_No response_

vringar:
### Description

This change allows users to exactly specify how they want
to open their log file while falling back to the defaults for arguments that weren't specified.

### Motivation and Context

In [OpenWPM](https://github.com/openwpm/OpenWPM) we run multiple Firefox instances in parallel.
To integrate their output with our logging infrastructure we want them to write their logs into pipes, which we then read on the other side, enrich it with metadata and log it ourselves.

This feature is relevant to anyone who wants to interleave their own code's log lines with Selenium's or the browser's.

Fixes #10703

### Types of changes

### Checklist

### DIFF
--- a/py/selenium/webdriver/firefox/service.py
+++ b/py/selenium/webdriver/firefox/service.py
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from typing import List, Tuple, Union
+from typing import List, Sequence, Union
 
 from selenium.webdriver.common import (service, utils)
 
@@ -28,7 +28,7 @@ class Service(service.Service):
 
     def __init__(self, executable_path: str = DEFAULT_EXECUTABLE_PATH,
                  port: int = 0, service_args: List[str] = None,
-                 log_path: Union[str, Tuple[str, str]] = "geckodriver.log", env: dict = None):
+                 log_path: Union[str, Sequence[str, str]] = "geckodriver.log", env: dict = None):
         """Creates a new instance of the GeckoDriver remote service proxy.
 
         GeckoDriver provides a HTTP interface speaking the W3C WebDriver
@@ -47,8 +47,11 @@ class Service(service.Service):
             in the services' environment.
         """
 
-        log_path, log_path_mode = (log_path[0], log_path[1]) if isinstance(log_path,Tuple) else (log_path, "a+")
-        log_file = open(log_path, log_path_mode, encoding='utf-8') if log_path else None
+        log_file = None
+        if log_path is not None:
+            open_args = (log_path, "a+", encoding='utf-8') if isinstance(log_path, str) else log_path
+            log_file = open(*open_args)
+
 
         super().__init__(executable_path, port=port, log_file=log_file, env=env)
         self.service_args = service_args or []
--- a/py/selenium/webdriver/firefox/service.py
+++ b/py/selenium/webdriver/firefox/service.py
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from typing import List
+from typing import List, Tuple
 
 from selenium.webdriver.common import (service, utils)
 
@@ -28,7 +28,7 @@ class Service(service.Service):
 
     def __init__(self, executable_path: str = DEFAULT_EXECUTABLE_PATH,
                  port: int = 0, service_args: List[str] = None,
-                 log_path: str = "geckodriver.log", env: dict = None):
+                 log_path: Union[str, Tuple[str, str]] = "geckodriver.log", env: dict = None):
         """Creates a new instance of the GeckoDriver remote service proxy.
 
         GeckoDriver provides a HTTP interface speaking the W3C WebDriver
@@ -41,12 +41,14 @@ class Service(service.Service):
         :param service_args: Optional list of arguments to pass to the
             GeckoDriver binary.
         :param log_path: Optional path for the GeckoDriver to log to.
-            Defaults to _geckodriver.log_ in the current working directory.
+            Can also be called with (path_name, mode).
+            Defaults to _geckodriver.log_ in the current working directory and _a+_ respectively.
         :param env: Optional dictionary of output variables to expose
             in the services' environment.
-
         """
-        log_file = open(log_path, "a+", encoding='utf-8') if log_path else None
+
+        log_path, log_path_mode = (log_path[0], log_path[1]) if isinstance(log_path,Tuple) else (log_path, "a+")
+        log_file = open(log_path, log_path_mode, encoding='utf-8') if log_path else None
 
         super().__init__(executable_path, port=port, log_file=log_file, env=env)
         self.service_args = service_args or []
--- a/py/selenium/webdriver/firefox/service.py
+++ b/py/selenium/webdriver/firefox/service.py
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from typing import List, Tuple
+from typing import List, Tuple, Union
 
 from selenium.webdriver.common import (service, utils)
 
--- a/py/selenium/webdriver/firefox/service.py
+++ b/py/selenium/webdriver/firefox/service.py
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from typing import List, Sequence, Union
+from typing import Any, List, Sequence, Union
 
 from selenium.webdriver.common import (service, utils)
 
@@ -28,7 +28,7 @@ class Service(service.Service):
 
     def __init__(self, executable_path: str = DEFAULT_EXECUTABLE_PATH,
                  port: int = 0, service_args: List[str] = None,
-                 log_path: Union[str, Sequence[str, str]] = "geckodriver.log", env: dict = None):
+                 log_path: Union[str, Sequence[Any], None] = "geckodriver.log", env: dict = None):
         """Creates a new instance of the GeckoDriver remote service proxy.
 
         GeckoDriver provides a HTTP interface speaking the W3C WebDriver
@@ -41,8 +41,9 @@ class Service(service.Service):
         :param service_args: Optional list of arguments to pass to the
             GeckoDriver binary.
         :param log_path: Optional path for the GeckoDriver to log to.
-            Can also be called with (path_name, mode).
+            Can also be called with a sequence of parameters which will be passed to open unmodified.
             Defaults to _geckodriver.log_ in the current working directory and _a+_ respectively.
+            Pass _None_ to disable logging
         :param env: Optional dictionary of output variables to expose
             in the services' environment.
         """
--- a/py/selenium/webdriver/firefox/service.py
+++ b/py/selenium/webdriver/firefox/service.py
@@ -50,7 +50,7 @@ class Service(service.Service):
 
         log_file = None
         if log_path is not None:
-            open_args = {"file": log_path, "mode":"a+", "encoding":"utf-8"} if isinstance(log_path, str) else {"mode":"a+", "encoding":"utf-8"} | log_path
+            open_args = {"file": log_path, "mode":"a+", "encoding":"utf-8"} if isinstance(log_path, str) else {"file": "geckodriver.log", "mode":"a+", "encoding":"utf-8"} | log_path
             log_file = open(**open_args)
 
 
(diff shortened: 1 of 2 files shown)


vringar:
This branch is still lacking test cases, which I'll be adding once there is interest to merge. If https://github.com/SeleniumHQ/selenium/issues/10703 gets closed as invalid/not relevant I don't want to have invested too much effort.

py/selenium/webdriver/firefox/service.py
```diff
@@ -46,7 +47,14 @@ def __init__(self, executable_path: str = DEFAULT_EXECUTABLE_PATH,
             in the services' environment.
 
         """
-        log_file = open(log_path, "a+") if log_path else None
+        log_file = None
```
symonk, line 51: Seems good to me; we could also allow `log_path` to become an (optional) length-two tuple that accepts something like `log_path: Union[str, Tuple[str, str]]` where the later case would be the `path, mode` passed to `open(...)`.  I'm ok with the current changes tho; seems to keep the flow but support an edge case.  Could we add a test for this?

vringar: I like this solution a lot more so I changed it to that. I hope the change to the doc string is understandable.
I'm currently unable to run the tests locally but I'll try to write one anyway.

vringar:
I think this PR is ready for another round of review.
The test I've added is pretty basic but idk if you wanted more @symonk.

py/test/selenium/webdriver/marionette/mn_launcher_tests.py
```diff
@@ -32,10 +33,17 @@ def test_we_can_launch_multiple_firefox_instances(capabilities):
...
 
 def test_launch_firefox_with_empty_string_service_log_path(capabilities):
-    driver = Firefox(capabilities=capabilities, service_log_path="")
+    service = Service(log_path="")
+    driver = Firefox(capabilities=capabilities, service=service)
+    driver.quit()
+
+def test_launch_firefox_with_empty_string_service_log_path(capabilities):
+    service = Service(log_path=("geckodriver.log", "w"))
+    driver = Firefox(capabilities=capabilities, service=service)
```
omkarkhatavkar: You can parametrize tests like this?

```python
@pytest.mark.parametrize('log_path', [None, "", ("geckodriver.log", "w")])
def test_launch_firefox_with_service_log_path(log_path):
    service = Service(log_path=log_path)
    driver = Firefox(capabilities=capabilities, service=service)
    driver.quit()
```

vringar: Good idea! Done

py/test/selenium/webdriver/marionette/mn_launcher_tests.py
```diff
@@ -16,6 +16,7 @@
 # under the License.
 
 from selenium.webdriver import Firefox
+from selenium.webdriver.firefox.service import Service
```
omkarkhatavkar, line 19: `import pytest` is missing i think

vringar: True. Done.
wish I had some way to run the tests so I could see if I missed anything else.

py/selenium/webdriver/firefox/service.py
```diff
@@ -28,7 +28,7 @@ class Service(service.Service):
 
     def __init__(self, executable_path: str = DEFAULT_EXECUTABLE_PATH,
                  port: int = 0, service_args: List[str] = None,
-                 log_path: str = "geckodriver.log", env: dict = None):
+                 log_path: Union[str, Tuple[str, str]] = "geckodriver.log", env: dict = None):
```
symonk: `Sequence[str, str]` is probably preferred here; a list would be fine etc, let's not rely on concrete implementations when we can avoid it (I know this statement is grossly untrue for other parts but i'd like to keep anything new as loose as we can get away with, checking the `isinstance` will need an update too

py/selenium/webdriver/firefox/service.py
```diff
@@ -41,12 +41,13 @@ def __init__(self, executable_path: str = DEFAULT_EXECUTABLE_PATH,
...
         :param log_path: Optional path for the GeckoDriver to log to.
-            Defaults to _geckodriver.log_ in the current working directory.
+            Can also be called with (path_name, mode).
+            Defaults to _geckodriver.log_ in the current working directory and _a+_ respectively.
         :param env: Optional dictionary of output variables to expose
             in the services' environment.
-
         """
-        log_file = open(log_path, "a+") if log_path else None
+        log_path, log_path_mode = (log_path[0], log_path[1]) if isinstance(log_path,Tuple) else (log_path, "a+")
```
symonk: How about:

```
# if log_path was not explicitly set to None
open_args = (log_path,) if isinstance(log_path, str) else log_path
log_file = open(*open_args)
```
Has the added benefit of allowing the caller to shovel arguments into the `open` call positionally (should there be another use case to change something else and not just `mode`)

(would need to account for the `Optional` nature too; docs point at optional, type hint does not).  I suspect it might be easier to default it to `None` and then assign it to `geckodriver.log` if it `is None` (I assume later calling code is _assuming_ the log file is always available).

edit2: best to keep the current behaviour if `log_path` is explicitly set to `None` after looking at some parent code

py/selenium/webdriver/firefox/service.py
vringar: Done!

py/selenium/webdriver/firefox/service.py
vringar: Okay, I think I did what you asked, but I'm not sure I 100% I understood you.

py/selenium/webdriver/firefox/service.py
```diff
@@ -41,12 +41,15 @@ def __init__(self, executable_path: str = DEFAULT_EXECUTABLE_PATH,
...
         :param log_path: Optional path for the GeckoDriver to log to.
-            Defaults to _geckodriver.log_ in the current working directory.
+            Can also be called with (path_name, mode).
+            Defaults to _geckodriver.log_ in the current working directory and _a+_ respectively.
         :param env: Optional dictionary of output variables to expose
             in the services' environment.
-
         """
-        log_file = open(log_path, "a+") if log_path else None
+        log_file = None
```
symonk, line 51: ignore my other comments, misread this var name, LGTM :+1:

py/test/selenium/webdriver/marionette/mn_launcher_tests.py
```diff
@@ -31,11 +33,8 @@ def test_we_can_launch_multiple_firefox_instances(capabilities):
...
-def test_launch_firefox_with_none_service_log_path(capabilities):
-    driver = Firefox(capabilities=capabilities, service_log_path=None)
-    driver.quit()
-
-
-def test_launch_firefox_with_empty_string_service_log_path(capabilities):
-    driver = Firefox(capabilities=capabilities, service_log_path="")
-    driver.quit()
+@pytest.mark.parametrize('log_path', [None, "", ("geckodriver.log", "w")])
+def test_launch_firefox_with_service_log_path(log_path):
```
symonk, line 35: can we inspect the underlying `log_file` attr for assertion? (calling `.quit()` is unnecessary also).  `driver.service.log_file` will be available;

I'd like to see:

 - not passed (`geckodriver.log`)
 - `None` explicitly
 - other file + mode

both `name` & `mode` will be available on the `log_file` attr

symonk:
Apologies for the delays! I've approved the PR for running actions workflows.  Thanks a bunch for the pull request and taking the effort to contribute, it is greatly appreciated!  Few comments around tests

py/test/selenium/webdriver/marionette/mn_launcher_tests.py
vringar: I couldn't figure out how to test for the default behaviour in the parametrized version, so I split that off into an extra test. I hope that's okay.

vringar:
No worries about the delays!
Unfortunately your permission to run the tests only seems to apply to a single commit, so I've activated the Python Actions for my fork, to make sure the tests are passing.

vringar:
Okay, the chrome tests are failing (in the acitons of my fork) but as I didn't modify any code on that path, I don't think that should block merging here.

vringar:
Just bumping to make sure it's not lost.

vringar:
Hey,
is there something I can do to move this forward?

vringar:
@symonk Sorry to ping you on this but can you run the tests again please

titusfortner:
It doesn't make sense to me for a parameter named `log_path` to take a dictionary that takes 3 optional arguments. Does Python Selenium manage logging this way elsewhere in the code or is it only geckodriver logs? Would it make sense to create a new parameter/class? (that's what Java would do, I'm not actually sure what the Python idiomatic approach would be)

I'm also thinking along the lines of making all logging more easy and have one class to manage all the sources and outputs of logs. #10944

vringar:
> Does Python Selenium manage logging this way elsewhere in the code or is it only geckodriver logs?

I found the [`webkitgtk/service.py`](https://github.com/SeleniumHQ/selenium/blob/316f9738a8e2079265a0691954ca8847e68c598d/py/selenium/webdriver/webkitgtk/service.py#L39) and [`wpewebkit/service.py`](https://github.com/SeleniumHQ/selenium/blob/316f9738a8e2079265a0691954ca8847e68c598d/py/selenium/webdriver/wpewebkit/service.py#L39) have a similar pattern, but they use binary logs and always use `wb` open mode, so they are already different from the Firefox service. Chromium browsers just pass the log location as an Command line argument.

> Would it make sense to create a new parameter/class? (that's what Java would do, I'm not actually sure what the Python idiomatic approach would be)

AFAIK, having a parameter accept multiple types is the usual way to do overloading, as there is no overload resolution in Python and always only a single constructor. I agree the name is unfortunate, but changing it would be backwards incompatible.

> It doesn't make sense to me for a parameter named log_path to take a dictionary that takes 3 optional arguments.

As a technicality, the dict can have [up to 8 values](https://docs.python.org/3/library/functions.html#open) since that is how many the `open` function accepts. I could also break out the `encoding` and `mode` as separate optional params, which is what I initially did until @symonk [guided me](https://github.com/SeleniumHQ/selenium/pull/10704#discussion_r883936345) to the overload approach.

PR #10743, opened 2022-06-07
[py] Implement `/status` in the python bindings

symonk:
A draft implementation of calling out to the w3c `/status` endpoint.

Todo:
 - Error handling
 - More test(s)


AutomatedTester:
My understanding is that `/status` doesn't require a session to call it but the way this is looking it looks like we do need a session.

I also wonder if there was a way we can do this so people can use it in a `conftest.py` better rather than being in the `remote` part of the library

symonk:
> My understanding is that `/status` doesn't require a session to call it but the way this is looking it looks like we do need a session.
> 
> I also wonder if there was a way we can do this so people can use it in a `conftest.py` better rather than being in the `remote` part of the library

Will look back to this shortly, I will take a closer look

symonk:
I will revisit, this one is slightly different; will reference the java implementation

PR #10744, opened 2022-06-07
[py]: Share command mapping between all instances of RemoteConnection.

symonk:
Share the `_COMMANDS` mapping between all instances of `RemoteConnection`, add a new constants for the HTTP verbs used and tidy up some of the `execute` code.

I'm not sure if the assert statement was ever firing here; I assume a `KeyError` would of been raised before it for the case it was 'attempting' to handle (will investigate)

`_commands` was non-public API so should be ok to rename it

Todo:
 - reference `_COMMANDS` in subclasses
 - move http verbs to a more central place to use in other `RemoteConnection` classes

### DIFF
--- a/py/selenium/webdriver/remote/remote_connection.py
+++ b/py/selenium/webdriver/remote/remote_connection.py
@@ -31,6 +31,7 @@ from selenium import __version__
 from .command import Command
 from .errorhandler import ErrorCode
 from . import utils
+from .constants import HttpVerb
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


py/selenium/webdriver/remote/remote_connection.py
```diff
@@ -337,14 +338,13 @@ def execute(self, command, params):
          - params - A dictionary of named parameters to send with the command as
            its JSON payload.
         """
-        command_info = self._commands[command]
-        assert command_info is not None, 'Unrecognised command %s' % command
```
symonk, line 341: I don't think this could ever fire as its not using `.get(...)` on the lookup

Issue #10757, opened 2022-06-10
[🐛 Bug]: Unable to cast object String to Boolean in WebElement.Displayed


ParanoikCZE:
### What happened?

I've observed this issue only a few times but gave it a shot with the investigation. I've tracked it to 
https://github.com/SeleniumHQ/selenium/blob/trunk/dotnet/src/webdriver/WebElement.cs#L188

Not sure what cames here, as it's almost impossible to reproduce, but pretty clearly there could come kind of exception or another kind of unexpected message and cast fails, so I recommend changing this part of the code to something like:
var res = bool.TryParse(commandResponse.Value, out bool result);
return res ? result : false;

### How can we reproduce the issue?

```shell
Observed in ~1/100000 cases so there is no simple way to do this. Code used when observed:
new WebDriverWait(driver, new TimeSpan(0, 0, 30)).Until(ExpectedConditions.ElementIsVisible(by));
The checked element is a button in the dropdown menu which might be caught in a kind of "construction" state so it's possible that called js doesn't exist and returns an 'undefined' result.
```

### Relevant log output

```shell
System.InvalidCastException : Unable to cast object of type 'System.String' to type 'System.Boolean'.

--InvalidCastException
   at OpenQA.Selenium.WebElement.get_Displayed()
   at SeleniumExtras.WaitHelpers.ExpectedConditions.<>c__DisplayClass20_0.<ElementToBeClickable>b__0(IWebDriver driver)
   at OpenQA.Selenium.Support.UI.DefaultWait`1.Until[TResult](Func`2 condition, CancellationToken token)
   at OpenQA.Selenium.Support.UI.DefaultWait`1.Until[TResult](Func`2 condition)
```

### Operating System

Azure devops

### Selenium version

C# 4.1.1

### What are the browser(s) and version(s) where you see this issue?

Chrome 102.0.5005.63

### What are the browser driver(s) and version(s) where you see this issue?

101.0.4951.4100

### Are you using Selenium Grid?

4.1.4

diemol:
Would you like to send us a PR?

ParanoikCZE:
> Would you like to send us a PR?

If my proposed solution is fine for you, but I'm not sure about other languages.

keraymundo10:
Hi @ParanoikCZE, just wondering which Windows version are the tests executing against? Is it Windows 11? Thanks!
I am encountering the same error against Windows 11. but works fine against Windows 10.

[...]

PR #10763, opened 2022-06-13
[dotnet] Safer Displayed method return value evaluation

ParanoikCZE:
solution for #10757

### DIFF
--- a/dotnet/src/webdriver/WebElement.cs
+++ b/dotnet/src/webdriver/WebElement.cs
@@ -185,7 +185,7 @@ namespace OpenQA.Selenium
                 parameters.Add("args", new object[] { this.ToElementReference().ToDictionary() });
                 commandResponse = this.Execute(DriverCommand.ExecuteScript, parameters);
 
-                return (bool)commandResponse.Value;
+                return bool.TryParse(commandResponse?.Value?.ToString(), out bool result) ? result : false;
             }
         }
 


titusfortner (changes requested):
Any response that isn't a boolean is an error. What if we throw a JavaScript error if the result isn't boolean? Then you can add the JavaScriptError to your list of ignored exceptions in your Wait class.

ParanoikCZE:
> Any response that isn't a boolean is an error. What if we throw a JavaScript error if the result isn't boolean? Then you can add the JavaScriptError to your list of ignored exceptions in your Wait class.

That's for sure another good solution as it could rethrow the actual message so we would know what really happened. So, what now? Can you do it or should I?

Anyway, meanwhile, I've got another similar issue at WebElement:127 (Unable to cast object of type 'System.Collections.Generic.Dictionary`2[System.String,System.Object]' to type 'System.Boolean') so all bool properties should be handled. And another issue with probably the same root cause at WebElementFactory.GetElementId:82 (elementDictionary (Parameter 'The specified dictionary does not contain an element reference')).

The root cause in these cases is probably some infrastructure/network issues, as I've started to get many weird performance issues from the whole testing framework.

titusfortner:
Sorry for not following up on this. That's interesting that you're seeing it in multiple places. `Displayed()` kind of makes sense because it is a JS execution, I'm not sure why the driver would be returning the wrong value type from `IsElementSelected`. 

If you can, yes, please update the PR. The JavaScriptError makes sense for Displayed, but not for others, maybe we use the `UnknownError` for those? Bottom line is we want better error messages so it is easier to see what the actual problem is.

titusfortner:
@ParanoikCZE are you interested in continuing to work on this one?

ParanoikCZE:
> @ParanoikCZE are you interested in continuing to work on this one?

Feel free to handle this, I'm actually out of the free time to focus on this properly to handle all methods.

titusfortner:
Ok, I'll close this and we can continue to track the issue on #10757 
Thank you,

PR #10764, opened 2022-06-13
[JS] Fix: using fulfillRequest when intercepting

TamsilAmani:
### Description

The implementation of Network Interception BiDi command is outdated and no longer works (for JavaScript atleast).
It uses `Fetch.continueRequest` CDP command, which is now used to change the parameters of request only and not the response. The correct command to be used now is `Fetch.fulfillRequest` which mocks the response.

### Types of changes

### Checklist


harsha509 (approved):
Thank you @TamsilAmani !

PR #10773, opened 2022-06-15
[py]: Enable co-operative multi inheritance with `super()` throughout

### LINKED ISSUE #10769: [🚀 Feature]: Replace explicit parent class references using `Parent.__init__(...)`
### Feature and motivation

I upgraded _most_ things to python 3.7+ however there are still _some_ references using their parent by direct name, this has a few caveats (most importantly gets in the way of MRO and doesn't lend itself to co-operative multi-inheritance, albeit it that is a small potential use case):

- Update all references using `SomeParentClass.__init__(...)` to instead use `super().__init__(...)`

### Usage example

Here is one example:

https://github.com/SeleniumHQ/selenium/blob/trunk/py/selenium/webdriver/chromium/remote_connection.py#L23
https://github.com/SeleniumHQ/selenium/blob/31cc432099851bc9aa68c443b448ff97d7ad91ca/py/selenium/webdriver/chromium/service.py#L45 

Should be able to find using some sort of `grep -ir /py .*\.__init__` (or similar)

symonk:
I updated most of the code to be python3.7+ syntactically, a few outliers referencing their parent class explicitly by name. This patch addresses all cases of that throughout the code.  This will allow co-operative multi inheritance (leaving mro to python rather than strictly naming the class) should we/a user wish to benefit from it and make things syntactically tidier.

closes #10769

### DIFF
--- a/py/selenium/webdriver/firefox/extension_connection.py
+++ b/py/selenium/webdriver/firefox/extension_connection.py
@@ -51,7 +51,7 @@ class ExtensionConnection(RemoteConnection):
 
         self.binary.launch_browser(self.profile, timeout=timeout)
         _URL = "http://%s:%d/hub" % (HOST, PORT)
-        super().__init__(URL, keep_alive=True)
+        super().__init__(_URL, keep_alive=True)
 
     def quit(self, sessionId=None):
         self.execute(Command.QUIT, {'sessionId': sessionId})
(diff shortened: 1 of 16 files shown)


py/selenium/webdriver/firefox/extension_connection.py
```diff
@@ -51,8 +51,7 @@ def __init__(self, host, firefox_profile, firefox_binary=None, timeout=30):
 
         self.binary.launch_browser(self.profile, timeout=timeout)
         _URL = "http://%s:%d/hub" % (HOST, PORT)
-        RemoteConnection.__init__(
-            self, _URL, keep_alive=True)
+        super().__init__(URL, keep_alive=True)
```
symonk: should be `_URL`.

PR #10777, opened 2022-06-16
correctly checks the driver supports DevTools

### LINKED ISSUE #10776: [🐛 Bug]: WebDriver throws NullReferenceException when calling StartEventMonitoring with EventFiringWebDriver
### What happened?

Creating a `JavaScriptEngine` with an `EventFiringWebDriver` as parameter will throw a `NullReferenceException` when calling `monitor.StartEventMonitoring();`

I believe line 57 in
https://github.com/SeleniumHQ/selenium/blob/5713de4224785d86cb3a7bdbb99ee62404c9c59b/dotnet/src/webdriver/JavaScriptEngine.cs#L48-L64

should be changed to  `if (devToolsDriver == null)` to correctly throw a `WebDriverException` when trying to use a non-compatible webdriver

```shell
using (var driver = new EventFiringWebDriver(new ChromeDriver()))
            {
                IJavaScriptEngine monitor = new JavaScriptEngine(driver);
                await monitor.StartEventMonitoring();
            }
```

### Relevant log output

```shell
Message: 
Test method WebDriver.JavascriptEngine.Tests.UnitTest2.TestMethod1 threw exception: 
System.NullReferenceException: Object reference not set to an instance of an object.

  Stack Trace: 
<.ctor>b__0() line 62
Lazy`1.CreateValue()
Lazy`1.LazyInitValue()
Lazy`1.get_Value()
<StartEventMonitoring>d__24.MoveNext() line 116
--- End of stack trace from previous location where exception was thrown ---
TaskAwaiter.ThrowForNonSuccess(Task task)
TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
TaskAwaiter.GetResult()
<TestMethod1>d__5.MoveNext() line 80
--- End of stack trace from previous location where exception was thrown ---
TaskAwaiter.ThrowForNonSuccess(Task task)
TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
ThreadOperations.ExecuteWithAbortSafety(Action action)
```

### Operating System

Windows 10

### Selenium version

.Net Framework 4.8

Chrome 102

ChromeDriver 102.0.5005.6102

_No response_

elgatov:
When using a driver like EventFiringWebDriver as a parameter for
JavaScriptEngine and trying to use StartEventMonitoring the constructor
will throw a NullReferenceException. This is caused by casting the
driver to IDevTools using "as" which will return null if the casting
failed, and incorrectly checking the field "session" instead
of "devToolsDriver" for null value.

fixes #10776

### Description
This commit will correctly check in the JavascriptEngine constructor that the driver passed as a parameter supports DevTools

### Motivation and Context
Previously the JavascriptEngine constructor checked the field session was not null, instead of the field devToolsDriver which would cause a NullReferenceException when using an incompatible driver like the EventFiringWebDriver. Now it correctly check the field devToolsDriver and, if the casting of the driver to IDevTools returns null, the JavascriptEngine constructor will correctly throw a WebDriverException with the message "Driver must implement IDevTools to use these features"

### Types of changes

### Checklist


titusfortner:
@elgatov this appears to be based on a very old commit of Selenium.
Current code is in `JavaScriptEngine.cs`
Can you pull latest code, update the current file and force update?

titusfortner (changes requested):
This needs to be changed on `JavaScriptEngine.cs` from trunk branch

elgatov:
Will do. Thank you.

PR #10784, opened 2022-06-18
[wip][py] Streamline API of Safari webdriver

symonk:
Heavily a draft still, please see the todos here; there is still plenty of work to do, but I figured if I took one driver in particular and make a start in tidying stuff up; `RemoteWebDriver` accepts optional `Options` so we should be able to do individual drivers like this rather than do the `RemoteWebDriver` first and be forced to do all drivers in one PR/swoop.

Still to do:
 - Update accurate docs throughout
 - Improve types where free wins can be made
 - Tests & Docs fallout
 - Check all diff is not API breaking


py/selenium/webdriver/safari/remote_connection.py
```diff
@@ -21,10 +21,19 @@
 
 class SafariRemoteConnection(RemoteConnection):
 
-    browser_name = DesiredCapabilities.SAFARI['browserName']
+    browser_name = DesiredCapabilities.SAFARI["browserName"]
```
symonk, line 24: what are we doing with these DesiredCapabilities?

py/selenium/webdriver/safari/service.py
```diff
@@ -46,15 +50,13 @@ def __init__(self, executable_path: str = DEFAULT_EXECUTABLE_PATH,
...
 
-        if port == 0:
-            port = utils.free_port()
-
+        port = port or utils.free_port()
         self.service_args = service_args or []
 
         self.quiet = quiet
         log = PIPE
         if quiet:
```
symonk: this as a concept is vastly inconsistent vs other services

py/selenium/webdriver/safari/webdriver.py
```diff
@@ -16,88 +16,47 @@
...
+    Starts an instance of the safari driver in a subprocess inline with the configuration supplied
+    to the service argument and subsequently starts an instance of safari if required.
     """
 
-    def __init__(self, port=0, executable_path=DEFAULT_EXECUTABLE_PATH, reuse_service=False,
-                 desired_capabilities=DEFAULT_SAFARI_CAPS, quiet=False,
-                 keep_alive=True, service_args=None, options: Options = None, service: Service = None):
+    def __init__(
+        self,
+        keep_alive: bool = True,
```
symonk, line 36: does keep alive need to remain as an 'option' ? secondly, what should we do with `reuse_service` here? both `reuse_service` & `quiet` are now recommended to be configured via the service?

py/selenium/webdriver/safari/webdriver.py
```diff
@@ -16,88 +16,47 @@
...
+        options: typing.Optional[Options] = None,
```
symonk, line 37: are optional service & options viable? can we assume sensible defaults?

py/selenium/webdriver/safari/webdriver.py
```diff
@@ -107,23 +66,24 @@ def quit(self):
         except http_client.BadStatusLine:
             pass
         finally:
-            if not self._reuse_service:
```
symonk, line 110: figure out what to do with `reuse_service` and factor it in; should it be responsibility of `Service`? (weird api?)

py/selenium/webdriver/safari/service.py
```diff
@@ -16,46 +16,61 @@
...
     """
 
-    def __init__(self, executable_path: str = DEFAULT_EXECUTABLE_PATH,
-                 port=0, quiet=False, service_args=None):
+    def __init__(
+        self,
+        executable_path: str = "/usr/bin/safaridriver",
+        port: int = 0,
+        quiet: bool = False,
+        reuse_service: bool = False,
```
symonk, line 37: is this technically a breaking public api change?

py/selenium/webdriver/safari/service.py
```diff
@@ -16,46 +16,61 @@
...
 
-        if port == 0:
-            port = utils.free_port()
-
+        port = port or utils.free_port()
         self.service_args = service_args or []
 
         self.quiet = quiet
         log = PIPE
         if quiet:
```
symonk, line 60: deprecate this; to allow a `log_path` configuration in future

py/selenium/webdriver/safari/webdriver.py
```diff
@@ -16,88 +16,46 @@
...
 
-DEFAULT_SAFARI_CAPS = DesiredCapabilities.SAFARI.copy()
-
-
 class WebDriver(RemoteWebDriver):
     """
-    Controls the SafariDriver and allows you to drive the browser.
-
+    Starts an instance of the safari driver in a subprocess inline with the configuration supplied
+    to the service argument and subsequently starts an instance of safari if required.
```
symonk, line 31: better docstrings, doesn't make sense.

py/selenium/webdriver/safari/webdriver.py
```diff
@@ -16,88 +16,46 @@
...
     """
 
-    def __init__(self, port=0, executable_path=DEFAULT_EXECUTABLE_PATH, reuse_service=False,
-                 desired_capabilities=DEFAULT_SAFARI_CAPS, quiet=False,
-                 keep_alive=True, service_args=None, options: Options = None, service: Service = None):
+    def __init__(
+        self,
+        keep_alive: bool = True,
```
symonk, line 36: should `keep_alive` now be part of the service obj? I think so? - However that changes the API of the `SafariRemoteConnection` where it is **NOT** deprecated currently.

Issue #10196, opened 2021-12-29
[🚀 Feature]: switch Java-tests from JUnit 4 to JUnit 5


christophs78:
### Feature and motivation

Selenium still uses JUnit 4 (see eg https://github.com/SeleniumHQ/selenium/blob/702b3aae7385adf331d75cdeacd79720cf6b9c84/java/maven_deps.bzl#L56, https://github.com/SeleniumHQ/selenium/blob/702b3aae7385adf331d75cdeacd79720cf6b9c84/java/test/org/openqa/selenium/testing/JUnit4TestBase.java#L35)
JUnit 5 is out since 2017 and widely adopted.

I´m asking for this because there are issues when using JUnit 5 and Selenium 4 together. (https://github.com/SeleniumHQ/selenium/issues/9359) IMO such issues already should be recognized during automated testing.

titusfortner:
We use bazel not maven to run tests in parallel, and we test with Java 11, either one of which would keep issue #9359 from being surfaced. We also don't test TestNG, etc. I don't think catching that error is either the fault of or the responsibility of the Selenium test suite.

I don't there is a compelling reason to have an open issue to update our code base, though we might be open to Pull Requests for it. I'll see what others think.

titusfortner:
So, yes, this is a good idea, but it would be a major undertaking and lower on our priority list. We'll leave it open; anyone who wants to make a Pull Request to help with this, it would be appreciated.

christophs78:
I may look into this within the next weeks or months. My offer would be to work on this in batches. e.g. move some packages from junit4 to junit5, do a PR. after this come back, do some more packages, open another PR.
Does this make sense to you?

diemol:
Thank you, it does make sense.

bonigarcia:
The first step for migrating the Selenium tests from JUnit 4 to JUnit 5 has been implemented in #10752. This PR (already merged) enables the Bazel [contrib_rules_jvm](https://github.com/bazel-contrib/rules_jvm), which allows using the [JUnit Platform](https://junit.org/junit5/docs/current/user-guide/) for running test cases. The macro used to execute test suites in Selenium is called `java_test_suite`, and so far uses the JUnit 4 runner by default since most existing tests are JUnit 4-based.

In summary, currently, the existing Selenium tests are already executed on the JUnit Platform (i.e., JUnit 5), using the JUnit Vintage engine (i.e., JUnit 4) by default.

[...]

PR #10792, opened 2022-06-22
[java] Allow linting only on the default test

bonigarcia:
### Description
After migrating Selenium tests to JUnit 5, the rule `java_junit5_test` delegates to `contrib_rules_jvm`’s `java_test`, and that adds lint tests for every target. We want to add one set of lint tests, since the rest are going to be identical. Therefore, we tag all but one test with `no-lint`.

### Motivation and Context
Improve the performance of spotbug tests (linting) after the migration to Selenium Java tests to JUnit 5 (issue #10196).

### Types of changes

### Checklist


diemol (approved):
Thank you, @bonigarcia!

PR #10802, opened 2022-06-23
[py] WebElement parameter supported in some of expected_condition methods

Qnner:

### Description
WebElement parameter supported in some of expected_conditions methods

### Motivation and Context
when you located an element already, you can just use it now;  

### Types of changes

### Checklist


titusfortner:
This is why I don't like expected conditions classes, you always end up needing more of them than you want to maintain.

fwiw, if we are going to have EC classes, it does make sense to support both elements and locators.

At the same time, everyone should be using lambdas, so I'd be fine deprecating these entirely, too. :)

titusfortner:
It was decided at TLC meeting to keep expected conditions class in Python, so I think we should also support Elements and not just locators.

titusfortner (changes requested):
This needs tests in order to be approved

diemol:
@Qnner can you please add tests to this pull requests in order to be merged?

Qnner:
> @Qnner can you please add tests to this pull requests in order to be merged?

this is my first commit for a opensource，I need some time to complete this for test ^_^

PR #10807, opened 2022-06-24
[dotnet] Handle unsuccessful http requests properly

### LINKED ISSUE #10805: [🐛 Bug]: RemoteWebDriver doesn't properly handle response from the external resource
### What happened?

I use some 3rd-party provided hub, for instance, BrowserStack. When I want to start a remote session, then I expect the session is started. I know that I forgot to put api key for my BrowserStack account, and status code of http response from BrowserStack is 401 (or 403).

```csharp
var driver = new RemoteWebDriver(uri, options); // where uri is the link to /wd/hub
```

This line of code was successfully executed without exceptions, but my browser session was not started.

```shell
Try to start a remote session with incorrect driver options.
```

### Relevant log output

```shell
No output.
```

### Operating System

any

### Selenium version

any dotnet, selenium 4.0+

any

Remote

_No response_

nvborisenko:
- Check http status code
- Removed code which is never executed

I didn't verify whether the solution is still compiling without errors, I wrote this code in notepad.

Fixes #10805

### DIFF
--- a/dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
+++ b/dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
@@ -190,26 +190,6 @@ namespace OpenQA.Selenium.Remote
             }
             catch (HttpRequestException ex)
             {
-                WebException innerWebException = ex.InnerException as WebException;
-                if (innerWebException != null)
-                {
-                    if (innerWebException.Status == WebExceptionStatus.Timeout)
-                    {
-                        string timeoutMessage = "The HTTP request to the remote WebDriver server for URL {0} timed out after {1} seconds.";
-                        throw new WebDriverException(string.Format(CultureInfo.InvariantCulture, timeoutMessage, requestInfo.FullUri.AbsoluteUri, this.serverResponseTimeout.TotalSeconds), ex);
-                    }
-                    else if (innerWebException.Status == WebExceptionStatus.ConnectFailure)
-                    {
-                        string connectFailureMessage = "Could not connect to the remote WebDriver server for URL {0}.";
-                        throw new WebDriverException(string.Format(CultureInfo.InvariantCulture, connectFailureMessage, requestInfo.FullUri.AbsoluteUri, this.serverResponseTimeout.TotalSeconds), ex);
-                    }
-                    else if (innerWebException.Response == null)
-                    {
-                        string nullResponseMessage = "A exception with a null response was thrown sending an HTTP request to the remote WebDriver server for URL {0}. The status of the exception was {1}, and the message was: {2}";
-                        throw new WebDriverException(string.Format(CultureInfo.InvariantCulture, nullResponseMessage, requestInfo.FullUri.AbsoluteUri, innerWebException.Status, innerWebException.Message), innerWebException);
-                    }
-                }
-
                 string unknownErrorMessage = "An unknown exception was encountered sending an HTTP request to the remote WebDriver server for URL {0}. The exception message was: {1}";
                 throw new WebDriverException(string.Format(CultureInfo.InvariantCulture, unknownErrorMessage, requestInfo.FullUri.AbsoluteUri, ex.Message), ex);
             }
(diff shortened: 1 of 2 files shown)


titusfortner:
Thanks for the PR.

I'm still learning the .NET classes. Why can't that code you deleted be executed? Thanks.

nvborisenko:
In version 4 selenium was migrated from `WebRequest` to new `HttpClient`, what is good. But `HttpClient` works slightly differently.

For instance timeout exception now looks like this:
in .net 6
```
Unhandled exception. System.Threading.Tasks.TaskCanceledException: The request was canceled due to the configured HttpClient.Timeout of 0.001 seconds elapsing.
 ---> System.TimeoutException: A task was canceled.
 ---> System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at System.Threading.Tasks.TaskCompletionSourceWithCancellation`1.WaitWithCancellation(CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.GetHttp11ConnectionAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.SendWithVersionDetectionAndRetryAsync(HttpRequestMessage request, Boolean async, Boolean doRequestAuth, CancellationToken cancellationToken)
   at System.Net.Http.HttpMessageHandlerStage.Send(HttpRequestMessage request, CancellationToken cancellationToken)
   at System.Net.Http.DiagnosticsHandler.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
```
in .net framework 4.8
```
Unhandled Exception: System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
```

Or if cannot connect to the server:
in .net 6
```
Unhandled exception. System.Net.Http.HttpRequestException: No such host is known. (werkjhwberkjbwkerwe876.com:443)
 ---> System.Net.Sockets.SocketException (11001): No such host is known.
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken)
```
in .net framework 4.8
```
Unhandled Exception: System.Net.Http.HttpRequestException: An error occurred while sending the request. ---> System.Net.WebException: The remote name could not be resolved: 'werkjhwberkjbwkerwe876.com'


   at System.Net.HttpWebRequest.EndGetResponse(IAsyncResult asyncResult)
   at System.Net.Http.HttpClientHandler.GetResponseCallback(IAsyncResult ar)
```

titusfortner (changes requested):
Your fix won't work because there are lots of unsuccessful status codes that are valid: https://w3c.github.io/webdriver/#errors

titusfortner:
Ah, I see what is going on. Sauce Labs returns a different signature than Browser Stack, which is why it isn't having the same issue for a 401 error.

The issue is when this condition isn't met in `CreateResponse()`:
```
      if (responseInfo.ContentType != null && responseInfo.ContentType.StartsWith("application/json", StringComparison.OrdinalIgnoreCase))
```

Since all valid webdriver responses have to be those things, this is going to apply when either the http response isn't coming from a driver/server (i.e. bad request in `StartSession()`), or is an invalid webdriver response.


I  think what we want to do is throw a `WebDriverException` from an `else if` after the previous one like
```
new HttpResponseMessage(responseInfo.StatusCode).IsSuccessStatusCode
```

nvborisenko:
It was hard to understand, and finally it works for me.

I didn't touch or changed logic in making http requests, generally, don't throw any exceptions.

Current Logic: 
When we get http response, we construct `Response` based on body. Status of response is also set based on json body. But body might be formatted unexpectedly.

What I did:
After constructing Response object based on json body, we also set Status based on http status in case if we couldn't parse status from body (Success by default). So setting proper Status here allows it to be handled later and throw corresponding exception like `NoSuchElementException`.

@titusfortner thanks for your suggestion and please review.

titusfortner (approved):

Issue #10810, opened 2022-06-25
[🐛 Bug]: DevTools.createSession() picks a wrong window


asolntsev:
### What happened?

I have a browser with 2+ open windows (tabs).
When I execute `driver.getDevTools().createSession()` it connects to a wrong tab.
See the code sample below.

The problematic code is https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/devtools/DevTools.java#L99:

```java
    // Grab the first "page" type, and glom on to that.
    // TODO: Find out which one might be the current one
    TargetID targetId = infos.stream()
      .filter(info -> "page".equals(info.getType()))
      .map(TargetInfo::getTargetId)
      .findAny()
      .orElseThrow(() -> new DevToolsException("Unable to find target id of a page"));
```

Actually, we can easily pick the right `TargetID` by matching it with `WebDriver.getWindowHandle()`.

### How can we reproduce the issue?

```shell
@Test
  void useDevToolsInWrongWindow() {
    WebDriverManager.chromedriver().setup();
    ChromeDriver driver = new ChromeDriver();
    driver.navigate().to("https://duckduckgo.com");
    driver.executeScript("window.open()");

    DevTools devTools = driver.getDevTools();
    devTools.createSessionIfThereIsNotOne();
    devTools.send(Emulation.setAutoDarkModeOverride(Optional.of(true)));

    // Expected: the first window (duckduckgo.com) becomes dark
    // Actual: the second window (empty) becomes dark
    driver.quit();
  }
```

### The fix

Here is my pull request: https://github.com/SeleniumHQ/selenium/pull/10811

### Operating System

MacOS (no matter)

### Selenium version

4.3.0 Java

### What are the browser(s) and version(s) where you see this issue?

Chrome 102 (no matter)

### What are the browser driver(s) and version(s) where you see this issue?

ChromeDriver 102 (no matter)

### Are you using Selenium Grid?

no matter

diemol:
PR has been merged, thank you, @asolntsev!

PR #10811, opened 2022-06-25
#10810 pick the right target using given window handle

asolntsev:
### Description
Fix for issue https://github.com/SeleniumHQ/selenium/issues/10810

### Types of changes


pujagani:
Thank you for your contribution! I appreciate that the changes are trying to create a session correctly in the current tab. However, the changes are a breaking change to two public APIs. One way to tackle this could be to find a way to not explicitly pass the windowHandle id. Maybe have a method in DevTools class that lets the user set the pageId. If that is available then that is used else the it continues to use the first page found. If you think there is a better approach to tackle this, then feel free to discuss it.

asolntsev:
@pujagani There is a simpler way probably. 
We can leave old methods `createSession()` and `createSessionIfThereIsNotOne()` - let them guess the right TargetID. 
We can even mark them deprecated.

pujagani:
Keeping both the methods and calling out the default behavior was the first idea. But, I am not sure if we want to remove the old methods createSession() and createSessionIfThereIsNotOne() at some point. Hence, I suggested allowing the user to set the pageId and use that if available. CC: @diemol When time permits, can you please have a look and suggest what might be the best way forward? Thank you!

asolntsev:
> I suggested allowing the user to set the pageId and use that if available.

I realized that user doesn't even need to set pageId.
Look, this is a typical code for using `DevTools`:

```
DevTools devTools = ((HasDevTools) webDriver).getDevTools();  // this method knows current window handle, and can set it to `DevTools` instance. 
devTools.createSession();
```

Nothing changes for end user. 
But it implies architectural change: now instance of `DevTools` will know `TargetId`, meaning that an instance of `ChromiumDriver` cannot have the single field `DevTools` (as it has now), but will need to create a different instance of `DevTools` for every window. I don't know it this is consistent with the initial design of `DevTools`...  @diemol ?

pujagani:
That would be ideal from the user's perspective. The current Devtools design ties the lifecycle of a page session with the parent session. So to create a new page session, a fresh CDP session will be required. As you mentioned, this will require major changes including managing different page sessions and ensuring there are no hanging sessions if the parent/page session is closed. The long-term goal is to move to BiDi and only support that once all browsers support BiDi fully. I am not sure if it is worth investing time and effort to get this done accurately without a major ask from the users. 

However, I agree that there has to be a way to allow users to provide the page they want the CDP session for. For that, one way is to provide overloaded methods` createSession(), createSession(String windowHandle)` and keep both of them.

diemol:
I believe @asolntsev idea is very good because it would provide a better user experience. However, it requires considerable effort, and given that BiDi is slowly coming in, making major changes to the CDP integration is less than ideal.

I would keep both signatures, and when the window handle is passed by the user, use that one.

asolntsev:
@diemol @pujagani Ok, I've returned the parameterless methods `getSession()` and `getSessionIfNeeded()`. They will work in most cases (when the webdriver has only one tab opened).

diemol (approved):
Thank you, @asolntsev!

Issue #10812, opened 2022-06-25
[🚀 Feature]: Methods `Select.select*()` should NOT allow selecting DISABLED option (or select)


asolntsev:
### Feature and motivation

Currently `Select` methods work even the `option` is disabled. Or `select` is disabled.

I suggest these `selectByFoo` methods should throw an exception. 

### Usage example

```java
new Select(element).selectByVisibleText("fish");
```

### Pull request

see https://github.com/SeleniumHQ/selenium/pull/10814

diemol:
Is this something you see in all browsers or in an specific browser? I am probably wrong, but I was reading the WebDriver spec and in 14.1 (https://www.w3.org/TR/webdriver1/#element-click), step 7 talks about 
>If element is not [disabled](https://www.w3.org/TR/webdriver1/#dfn-disabled)

But I assume it talks about the whole element, not the specific options. I just want to understand better the problem, and figure out if this needs a fix in Selenium or in the WebDriver spec.

@titusfortner @AutomatedTester @jimevans what do you think?

asolntsev:
@diemol I am not sure about the spec, but in reality both `<select>` and `<option>` can have `disabled` attribute. 
1. In first case, you cannot select any options
2. In second case, you can select other options except the disabled one.

titusfortner:
No, Selenium should not prevent a user from clicking anything they want. A disabled element could theoretically have any number of JS event listeners.

That's why the spec allows all the other actions to be taken on a disabled option element except for the part about changing "selectedness".

The driver is doing the right thing and Selenium is doing the right thing.

titusfortner:
Watir, on the other hand, does prevent this, because as a tester, 99% of the time trying to select something that is disabled is wrong, so we've added extra logic to raise an error for it.

@asolntsev we discussed this a couple years ago about why Watir subclasses elements. This is a good use case. Watir automatically waits for an element to be enabled if it is one of `Input, Button, Select, Option` and otherwise ignores whether the element is enabled.

titusfortner:
Hmm, to argue with myself a second, this *is in an opinionated class. The intention of "select" is different from the intention of "click" even if that is the implementation detail happening below.

titusfortner:
This is one of those weird places where Selenium tries to have it both ways. It's "just browser automation" (aka "do what I say"), except it also has these support classes to provide "do what I mean" behaviors, but only in some cases, and not implemented in a way that can really be uniformly applied throughout a test framework.

[...]

PR #10814, opened 2022-06-25
#10812 don't allow selecting a disabled option

asolntsev:
fix for #10812


pujagani (approved):
Thank you for your contribution! @asolntsev  The changes look good to me. Other language bindings will also need to add the same checks.

titusfortner (dismissed):
Need to agree this is a problem (I don't think it is) before we solve it with this code.

titusfortner (changes requested):
I don't think we need to check if the select list is enabled. If it is disabled, the option will also be disabled. I'd prefer to optimize the number of commands made for a successful test rather than an unsuccessful test. What do you think?

asolntsev:
> I don't think we need to check if the select list is enabled. If it is disabled, the option will also be disabled. I'd prefer to optimize the number of commands made for a successful test rather than an unsuccessful test. What do you think?

Yes, I also like to make as few commands as possible. 
But the correctness should be in first place, right?

If `<select id="cars">` is `disabled` then user cannot select any options. Cannot even see them. 
But method `new Select($("#cars")).selectByValue("saab");` doesn't throw any exceptions. _It leaves a false impression that the option has been selected._

titusfortner:
A non-disabled `option` nested in a disabled `select`, will return `false` for `isEnabled()`.
`selectByValue()` calls `setSelected()` which calls `assertOptionIsEnabled()` which will error.

That said, it's only one extra command when using this wrapper, and it does give a better error message, so maybe I'm being too picky.

@diemol what do you think of this implementation?

diemol (approved):
Thank you, @asolntsev!

PR #10818, opened 2022-06-27
Use `which()` to get Firefox path on Mac OS

### LINKED ISSUE #10686: [🐛 Bug]: Contained Firefox can't be used on MacOS
### What happened?

For the project https://github.com/robocorp/rpaframework/ we're installing an isolated Firefox and adding that to the path. But because as seen here https://github.com/SeleniumHQ/selenium/blob/dab928e6ac2c270b03d0199f381950c998c3f8b0/py/selenium/webdriver/firefox/firefox_binary.py#L152 the path for MacOS is hardcoded and not searched for in $PATH, Selenium will never find and use that version.

```shell
Install Firefox on MacOS in a different path. Add it to the $PATH global variable and try to use Selenium against it. Selenium will fail, not being able to find Firefox.
```

### Relevant log output

```shell
SessionNotCreatedException: Message: Expected browser binary location, but unable to find binary
```

### Operating System

MacOS Monterey 12.3

### Selenium version

3.141.0 (but based on the code seems like it should be on all versions)

Firefox 98, 99 and 100

GeckoDriver 0.30.0

_No response_

bogdancondurache:
Fixes #10686

### Description
The Firefox binary under Mac OS was using a hardcoded path. This was
failing when trying to manipulate PATH env var to use Firefox from a
different location or if Firefox was installed in a different place.

### Types of changes

### Checklist


symonk (dismissed):

py/selenium/webdriver/firefox/firefox_binary.py
```diff
@@ -150,7 +150,10 @@ def _get_firefox_start_cmd(self):
         """Return the command to start firefox."""
         start_cmd = ""
         if self.platform == "darwin":  # small darwin due to lower() in self.platform
-            start_cmd = "/Applications/Firefox.app/Contents/MacOS/firefox-bin"
+            start_cmd = self.which(ffname)
```
symonk, line 154: is `ffname` in scope here? I don't think its in the local namespace until later on the `for` loop?

symonk (changes requested):
referencing a local variable that doesn't exist

py/selenium/webdriver/firefox/firefox_binary.py
bogdancondurache: yes, my bad, totally missed that. I assigned ffname to "firefox" only because from what I found iceweasel is not even available for Mac OS

symonk: interesting, from what I can find too `iceweasel` is not supported on `osx` too; one for another ticket tho - Thanks

symonk (approved):

symonk (approved):

diemol (approved):
Thank you, @bogdancondurache!

PR #10868, opened 2022-07-12
Remove require of websocket in selenium-devtools gem

kevindew:
### Description
In cd69898 the dependency of websocket was removed from selenium-devtools gem. When this gem is required it raises an error of: `cannot load such file -- websocket (LoadError)`

We experienced this error because we updated selenium-devtools and hadn't updated selenium-webdriver. We found the problem was resolved by updating selenium-webdriver as that has a websocket dependency cd69898

### Motivation and Context

Resolve an error when requiring the selenium-devtools gem

### Types of changes

### Checklist

### DIFF
--- a/rb/lib/selenium/devtools.rb
+++ b/rb/lib/selenium/devtools.rb
@@ -17,8 +17,6 @@
 # specific language governing permissions and limitations
 # under the License.
 
-require 'websocket'
-
 module Selenium
   module DevTools
     class << self
--- a/rb/selenium-devtools.gemspec
+++ b/rb/selenium-devtools.gemspec
@@ -39,4 +39,6 @@ Gem::Specification.new do |s|
   ] + Dir['lib/selenium/devtools/**/*']
 
   s.require_paths = ["lib"]
+
+  s.add_runtime_dependency 'selenium-webdriver', '~> 4.2'
 end


titusfortner (changes requested):
Excellent, thanks. This also needs to change the required version of selenium-webdriver to the one that added the requirement.

kevindew:
> Excellent, thanks. This also needs to change the required version of selenium-webdriver to the one that added the requirement.

Hmm do you mean in selenium-devtools? That doesn't currently [have a defined dependency on selenium-webdriver](https://github.com/SeleniumHQ/selenium/blob/729678239097292a63cd81da22f0ae39cfe97e6c/rb/selenium-devtools.gemspec)

titusfortner:
Hah, yes, then we need to add it for this.

kevindew:
> Hah, yes, then we need to add it for this.

Sure, it seems a little odd to add because selenium-devtools doesn't actually reference selenium-webdrivers. So won't make use of the dependency. 

So I guess the value in doing so would just be to just to force a constraint onto selenium-webdriver given Rubygems doesn't have a concept like peer dependencies which this seems to be. I'd guess for that it'd be a constraint of `>= 4.2.0` since that is the first version which bundles websocket gem.

If you've got a clear idea of what's needed do feel free to just push it to my branch if that's easier.

p0deje:
It's not really possible to use `selenium-devtools` without `selenium-webdriver`, so let's just add the latter as a runtime dependency:

```ruby
  s.add_runtime_dependency 'selenium-webdriver', '~> 4.2'
```

titusfortner:
I agree it is weird, but it isn't really a peer dependency because `selenium-webdriver` is never going to reference `selenium-devtools` directly in any way, so I can't think of any reason where adding `selenium-webdriver` as a requirement to `selenium-devtools` gemspec will be a problem.

kevindew:
Thanks, I've added [38c1ea0](https://github.com/SeleniumHQ/selenium/pull/10868/commits/38c1ea0ee344a32034a862b515d5994e15c43ff4) to add the requested dependency to selenium-devtools.

As per @p0deje's suggestion I went for the pessimistic constraint of `~> 4.2`.

titusfortner:
Thanks!

PR #10872, opened 2022-07-13
update docs for PointerActions#move_to

mockdeep:
`move_to` has been changed to start from the center of the element rather than the top-left corner.

### Checklist

### DIFF
--- a/rb/lib/selenium/webdriver/common/interactions/pointer_actions.rb
+++ b/rb/lib/selenium/webdriver/common/interactions/pointer_actions.rb
@@ -68,9 +68,8 @@ module Selenium
       end
 
       #
-      # Moves the pointer to the middle of the given element.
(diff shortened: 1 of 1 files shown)


titusfortner (changes requested):
Thank you for this PR! 

The only issue with it now is that it isn't always the "center" of the element, but is technically the ["in-view center point"](https://w3c.github.io/webdriver/#dfn-center-point) of the element.

But that's a lot to put in the places you updated. Maybe we just change "middle" to "in-view center point" on line 71 & delete line 72 which isn't exactly correct either?

mockdeep:
@titusfortner is this more along the lines of what you were thinking?

titusfortner (approved):

titusfortner:
yes, thanks!

PR #10873, opened 2022-07-13
[dotnet] Including xml docs in nuget package

### LINKED ISSUE #10371: [🐛 Bug]: Missing .NET documentation comments
### What happened?

The NuGet packages have no documentation comments at all. This affects all versions >= 4.0.0.

- v4.1.0:  
  grafik
- v3.141.0  
  grafik

```shell
- Install version 4.1.0 of the Selenium.Support package
- Press F12 on any Selenium type -> no documentation comments
- Downgrade to version 3.141.0
- Press F12 on any Selenium type -> documentation comments 🎉
```

### Relevant log output

```shell
-
```

### Operating System

Windows 11

### Selenium version

.NET 4.0.0, 4.0.1, 4.1.0

-

ChromeDriver 98.0.4758.8000

No

nvborisenko:
Fixes #10371 ?

I'm not sure, wrote code in notepad. If you can provide me within built nuget package from this PR, I can answer whether xml docs are inluded.

- [ ] Do not forget to make the same changes for `WebDriver.Support` package.


titusfortner:
the .net build is currently broken, I'll give this a go once we fix it.

titusfortner:
@nvborisenko does this have what you're expecting to see?

[SeleniumXMLCheck.nupkg.zip](https://github.com/SeleniumHQ/selenium/files/9122734/SeleniumXMLCheck.nupkg.zip)

nvborisenko:
Thanks, this fix doesn't work.

image

We are expecting `WebDriver.xml` file here.

titusfortner:
any other ideas for this?

nvborisenko:
No, it's very tricky for me to build package locally. At least my fix produces xml file in output directory besides WeDriver.dll. Then I see nuspec declaration also should include this xml file. I guess we can blame bazel.

No further actions are coming from me because I have no ability to build it locally.

titusfortner:
Yeah, Bazel, especially in .NET is a challenge.

Are you working on a Mac or Windows?

nvborisenko:
I am on Windows, with Linux subsystem for windows. Does it game changer? I thought .net 4.5 can be built only on Windows.

titusfortner:
Oh yeah, .NET *needs Windows to build still. :-/ 
I'm running Windows from a VM in order to build it. :)

You probably have most of this done already, except the Bazel stuff?
http://jimevansmusic.blogspot.com/2020/04/setting-up-windows-development.html

The 2 things to add to that post is I think you still need to specify Bazel v4 (@pujagani is that right or does 5 work?); and you need to have Ruby installed.

Then building is just a matter of:
```
bazel build //dotnet/src/webdriver:package
```
and output is in:
```
selenium/bazel-bin/dotnet/src/webdriver/Selenium.WebDriver.x.y.z.nupkg
```

nvborisenko:
Succesfully built nuget package locally. It was tricky.

Summary how to do it:
- Install pyton (including into PATH)
- Install ruby
- Install Java
- UNinstall WSL (because of bash.exe is executed in context of your Unix system)
- Install `c:\tools\msys64\` exactly here

Now, what we have: build system (bazel) compiles output to `bazel-out\x64_windows-fastbuild\bin\dotnet\src\webdriver\bazelout\` directory. It includes `dll` and `pdb` files only. `pdb` file is missing in nuget package because of package includes new format of symbols (snupkg) - it's kind of expected. `xml` file is missing here, thus I ca n say that building process by bazel doesn't generate `xml` file and this is why it's not included in nuget package.

Here is the list of msbuild params to build the project:
```
/unsafe-
/checked-
/nostdlib+
/utf8output
/deterministic+
/filealign:512
/nologo
/highentropyva+
/subsystemversion:6.00
/warn:0
/target:library
/langversion:7.3
/debug+
/optimize-
/define:TRACE;DEBUG
/debug:portable
...
```

On thin stage we are loosing xml file.

I have no idea how to specify additional parameters to build the project. Google says nothing. I read the list of supported properties for the BZL target, but didn't find something relevant.

I strongly recommend to remake build system for .net from the scratch, then you will have more contributors, despite cloning the repo of ~2GB.

titusfortner:
I completely understand your frustration here. Bazel support for Windows/.net is unfortunately limited, and we've had to do some custom things just to make what we already have work.

The issue is that we have some common code that is shared between all the languages that requires Bazel to build. If we moved to a more .net idiomatic solution, we'd still need to call out to execute bazel commands, which might not be significantly better than what we currently have, and it might be a better use of time to put effort into just fixing up the Bazel.

Tbh, the TLC is a little split on the best way to provide the experience developers need to successfully contribute to this project, and some of the opinions are strongly held, so we are not looking to make a change at this time. If you have a specific approach that combines a more idiomatic solution with the bazel commands we need to integrate, please submit a proposal and we can consider it.

In the meantime, should we close this PR because it doesn't accomplish what we hoped it would? Thanks again for your work on this.

nvborisenko:
https://github.com/bazelbuild/rules_dotnet/blob/0b7ae93fa81b7327a655118da0581db5ebbe0b8d/dotnet/private/actions/assembly_core.bzl#L95

Here is the source code how bazel actually builds a project in .net. It doesn't rely on `csproj` file at all, it provides `cs` files directly to compiler.

If we could bring in our custom (slightly changed) implementation of this bazel action, then I can play with it. Theoretically we need `/p:DocumentationFile=WebDriver.xml` parameter here (and probably `/p:GenerateDocumentationFile=True`, or vice versa).

titusfortner:
Can you join us in Selenium Slack to discuss?
https://www.selenium.dev/support/#ChatRoom

titusfortner:
So, this is something we need to fix in our Bazel implementation. We'll keep Issue open to track. Thanks for your work on this.

PR #10880, opened 2022-07-15
[py] build(setup.py): Add `project_urls` for PyPI

tony:
### Description

Show Project URLs to [Selenium's PyPI page](https://pypi.org/project/selenium/) via [`project_urls`](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls) to _setup.py_

image

### Motivation and Context

It takes a few clicks to get around to the docs and source code.  This makes it possible launch all of them via PyPI. 

It also is helpful to anyone else that ingests the metadata from setup.py / PyPI

### Types of changes

### Checklist

### DIFF
--- a/py/setup.py
+++ b/py/setup.py
@@ -35,7 +35,7 @@ setup_args = {
     'project_urls': {
         'Bug Tracker': 'https://github.com/SeleniumHQ/selenium/issues',
         'Changes': 'https://github.com/SeleniumHQ/selenium/blob/trunk/py/CHANGES',
-        'Documentation': 'https://seleniumhq.github.io/selenium/docs/api/py/api.html',
+        'Documentation': 'https://www.selenium.dev/documentation/overview/',
         'Source Code': 'https://github.com/SeleniumHQ/selenium/tree/trunk/py',
     },
     'python_requires': '~=3.7',
--- a/py/setup.py
+++ b/py/setup.py
@@ -32,6 +32,12 @@ setup_args = {
     'description': 'Python bindings for Selenium',
     'long_description': open(join(abspath(dirname(__file__)), "README.rst")).read(),
     'url': 'https://github.com/SeleniumHQ/selenium/',
+    'project_urls': {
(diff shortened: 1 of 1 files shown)


tony:
@symonk Do you have a take on this?

py/setup.py
```diff
@@ -32,6 +32,12 @@
     'description': 'Python bindings for Selenium',
     'long_description': open(join(abspath(dirname(__file__)), "README.rst")).read(),
     'url': 'https://github.com/SeleniumHQ/selenium/',
+    'project_urls': {
+        'Bug Tracker': 'https://github.com/SeleniumHQ/selenium/issues',
+        'Changes': 'https://github.com/SeleniumHQ/selenium/blob/trunk/py/CHANGES',
```
symonk, line 37: is this better pointing at: `https://github.com/SeleniumHQ/selenium/blob/trunk/py/CHANGES` specifically?

py/setup.py
```diff
@@ -32,6 +32,12 @@
...
+        'Documentation': 'https://seleniumhq.github.io/selenium/docs/api/py/api.html',
```
symonk: I think this is old docs, tho I'm a little out of the know on the docs atm, maybe it should pointing at: https://www.selenium.dev/documentation/overview/?

py/setup.py
tony: @symonk That's the same URL? Is there a different one you'd prefer?

py/setup.py
tony: Updated in 4664fe02cc

py/setup.py
symonk: apologies missed the `/py/`

symonk:
test failures here are likely flakes at the minute, `bidi` tests need some TLC.  fine to merge

symonk:
Thanks @tony for the contribution, much appreciated.  Apologies for the delays on review!

tony:
You're welcome, Thank you @symonk !

PR #10903, opened 2022-07-25
[ruby] feat: Added virtual authenticator

TamsilAmani:
### Description

Relates https://github.com/SeleniumHQ/selenium/issues/10541

- Added virtual authenticator for js.
- Added tests for the same.

### Types of changes

### Checklist


rb/lib/selenium/webdriver/common/credential.rb
```diff
@@ -0,0 +1,85 @@
...
+
+#
+# A credential stored in a virtual authenticator.
+# @see https://w3c.github.io/webauthn/#credential-parameters
+#
+
+module Selenium
+  module WebDriver
+    class Credential
+      def initialize(*args)
```
p0deje: Rather than using positional arguments, let's use keyword arguments:
```suggestion
      def initialize(id:, is_resident_credential:, rp_id:, user_handle:, private_key:, sign_count:)
```

If any of these parameters is optional, we can just set the default value to `nil`.

rb/lib/selenium/webdriver/common/driver.rb
```diff
@@ -248,6 +248,42 @@ def execute_async_script(script, *args)
         bridge.execute_async_script(script, *args)
       end
 
+      #
+      # virtual-authenticator
+      #
+
+      def virtual_authenticator_id
```
p0deje: Not all the drivers support this, so I think it would be better to keep these methods in a separate module (e.g. `Selenium::WebDriver::DriverExtensions::HasVirtualAuthenticator`) and then include it in the drivers supporting it. Similar to how we work with `HasBiDi` extension.

rb/lib/selenium/webdriver/remote/bridge.rb
```diff
@@ -42,6 +42,7 @@ def initialize(url:, http_client: nil)
           @http = http_client || Http::Default.new
           @http.server_url = uri
           @file_detector = nil
+          @authenticator_id = nil
```
p0deje: The bridge is supposed to be mostly stateless, so what if instead, we change an interface to return the virtual authenticator when it's added and allow to remove it later as a parameter? Something like this:

```ruby
authenticator = driver.add_virtual_authenticator(options)
driver. remove_virtual_authenticator(authenticator)
```

This way we can avoid storing authenticator ID internally. It would also allow us to have multiple virtual authenticators.

rb/lib/selenium/webdriver/remote/bridge.rb
```diff
@@ -568,6 +569,52 @@ def shadow_root(element)
...
+          credential_list = []
+
+          credential_data.each do |cred|
+            credential_list.push(Credential.new.from_dict(cred))
+          end
+          credential_list
+        end
+
+        def remove_credential(credential_id)
+          credential_id = Base64.urlsafe_encode64(credential_id.pack('C*')) if credential_id.instance_of?(Array)
```
p0deje: How can `credential_id` be an array?

rb/spec/integration/selenium/webdriver/virtual_authenticator_spec.rb
```diff
@@ -0,0 +1,395 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+require_relative 'spec_helper'
+
+module Selenium
+  module WebDriver
+    describe "VirtualAuthenticator" do
```
p0deje: Does this pass when run on all the browsers? If not, we should guard it with `only: {browser: :chrome}` or something similar (check other specs for examples).

rb/lib/selenium/webdriver/common/credential.rb
```diff
@@ -0,0 +1,85 @@
...
+# under the License.
+
+#
+# A credential stored in a virtual authenticator.
+# @see https://w3c.github.io/webauthn/#credential-parameters
+#
+
+module Selenium
+  module WebDriver
+    class Credential
```
p0deje: Is this a public API? If not we should mark it as `# @api private`.

rb/lib/selenium/webdriver/common/credential.rb
```diff
@@ -0,0 +1,85 @@
...
+        @private_key = args[4]
+        @sign_count = args[5]
+      end
+
+      attr_reader :id, :is_resident_credential, :rp_id, :private_key, :sign_count
+
+      def user_handle
+        return @user_handle unless @user_handle.nil?
+
+        nil
```
p0deje:
```suggestion
        @user_handle
```

This is essentially the same code.

rb/lib/selenium/webdriver/common/credential.rb
```diff
@@ -0,0 +1,85 @@
...
+          rpId: @rp_id,
+          privateKey: Base64.urlsafe_encode64(@private_key),
+          signCount: @sign_count
+        }
+
+        credential_data[:userHandle] = Base64.urlsafe_encode64(@user_handle&.pack('C*')) unless user_handle.nil?
+        credential_data
+      end
+
+      def from_dict(data)
```
p0deje: Let's make this a class method so that it would return an instance of credential class:

```suggestion
      def self.from_dict(data)
```

rb/lib/selenium/webdriver/common/virtual_authenticator_options.rb
```diff
@@ -0,0 +1,56 @@
...
+      def initialize
+        @protocol = PROTOCOL["CTAP2"]
+        @transport = TRANSPORT["USB"]
+        @has_resident_key = false
+        @has_user_verification = false
+        @is_user_consenting = true
+        @is_user_verified = false
+      end
+
+      def to_dict
```
p0deje:
```suggestion
      def as_json(*)
```

rb/lib/selenium/webdriver/common/credential.rb
```diff
@@ -0,0 +1,85 @@
...
+
+      def create_resident_credential(id, rp_id, user_handle, private_key, sign_count)
+        Credential.new(id, true, rp_id, user_handle, private_key, sign_count)
+      end
+
+      def create_non_resident_credential(id, rp_id, private_key, sign_count)
+        Credential.new(id, false, rp_id, nil, private_key, sign_count)
+      end
+
+      def to_dict
```
p0deje: The conventional name for this method would be as_json.

```suggestion
      def as_json(*)
```

rb/lib/selenium/webdriver/common/virtual_authenticator_options.rb
```diff
@@ -0,0 +1,56 @@
...
+  module WebDriver
+    class VirtualAuthenticatorOptions
+
+      PROTOCOL = {"CTAP2" => "ctap2", "U2F" => "ctap1/u2f"}.freeze
+      TRANSPORT = {"BLE" => "ble", "USB" => "usb", "NFC" => "nfc", "INTERNAL" => "internal"}.freeze
+
+      attr_accessor :protocol, :transport, :has_resident_key, :has_user_verification, :is_user_consenting,
+                    :is_user_verified
+
+      def initialize
```
p0deje: Let's allow users to customize the values on the initialization too:

```suggestion
      def initialize(protocol: PROTOCOL["CTAP2"], transport: TRANSPORT["USB"], has_resident_key: false, has_user_verification: false, is_user_consenting: true, is_user_verified: false)
```

rb/lib/selenium/webdriver/common/virtual_authenticator_options.rb
```diff
@@ -0,0 +1,56 @@
...
+#
+# Options for the creation of virtual authenticators.
+# @see http://w3c.github.io/webauthn/#sctn-automation
+#
+#
+module Selenium
+  module WebDriver
+    class VirtualAuthenticatorOptions
+
+      PROTOCOL = {"CTAP2" => "ctap2", "U2F" => "ctap1/u2f"}.freeze
```
p0deje: Usually Ruby would have keys as symbols, so let's use those instead:
```suggestion
      PROTOCOL = {ctap2: "ctap2", u2f: "ctap1/u2f"}.freeze
```

rb/lib/selenium/webdriver/common/virtual_authenticator_options.rb
```diff
@@ -0,0 +1,56 @@
...
+      TRANSPORT = {"BLE" => "ble", "USB" => "usb", "NFC" => "nfc", "INTERNAL" => "internal"}.freeze
```
p0deje:
```suggestion
      TRANSPORT = {ble: "ble", usb: "usb", nfc: "nfc", internal: "internal"}.freeze
```

rb/lib/selenium/webdriver/remote/bridge.rb
```diff
@@ -568,6 +569,52 @@ def shadow_root(element)
...
+        end
+
+        def credentials
+          credential_data = execute :get_credentials, {}, {authenticatorId: @authenticator_id}
+          credential_list = []
+
+          credential_data.each do |cred|
+            credential_list.push(Credential.new.from_dict(cred))
+          end
+          credential_list
```
p0deje:
```suggestion
          credential_data.map do |cred|
            Credential.new.from_dict(cred)
          end
```

This would produce the same result.

rb/lib/selenium/webdriver/remote/bridge.rb
```diff
@@ -568,6 +569,52 @@ def shadow_root(element)
...
+        def remove_credential(credential_id)
+          credential_id = Base64.urlsafe_encode64(credential_id.pack('C*')) if credential_id.instance_of?(Array)
+          execute :remove_credential, {}, {credentialId: credential_id, authenticatorId: @authenticator_id}
+        end
+
+        def remove_all_credentials
+          execute :remove_all_credentials, {}, {authenticatorId: @authenticator_id}
+        end
+
+        def user_verified(verified)
```
p0deje: Since this sets user verified, the convention in Ruby would be to name this method with a trailing equals sign:

```suggestion
        def user_verified=(verified)
```

Then the method is called like this:

```ruby
bridge.user_verified = false
```

rb/lib/selenium/webdriver/common/driver.rb
p0deje: Could we also add some `# @example` that demonstrate how to use the APIs?

rb/lib/selenium/webdriver/common/credential.rb
titusfortner: Agree on the class method, but I'd suggest `#from_json` instead of `#from_dict` since that's what we're actually converting from, and we're using that method name at least one other place

(Oh, and it doesn't require state, so I'm suggesting a class method here that returns an instance of the Credential)

```suggestion
      def self.from_json(data)
```

rb/lib/selenium/webdriver/common/driver.rb
titusfortner: The argument is that it is part of the spec, so the methods should be implemented across the board.

My issue is that the naming gets confusing here when tacked onto Driver class directly. I'm thinking we should have a separate class for these methods:

Something like:
```
driver.virtual_authenticator.id
driver.virtual_authenticator.add
driver.virtual_authenticator.remove
driver.virtual_authenticator.credentials=
driver.virtual_authenticator.credentials
driver.virtual_authenticator.user_verified=
```

rb/lib/selenium/webdriver/remote/bridge.rb
TamsilAmani: `credential_id` can be `id` or `rawID`. For example, this is a credential registered with authenticator.

`{"credential"=>{"id"=>"-lD3cYWt2iGgWuYkEU2FI7GH3h96ULdewKtqAHWL6g8", "rawId"=>[250, 80, 247, 113, 133, 173, 218, 33, 160, 90, 230, 36, 17, 77, 133, 35, 177, 135, 222, 31, 122, 80, 183, 94, 192, 171, 106, 0, 117, 139, 234, 15],...} `

Thus, both `id` and `rawId` can be used to remove a credential. For `rawId` we need to convert it to its matching `id`.

Similar thing has been done in java too. 
https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/remote/RemoteWebDriver.java#L1218

rb/lib/selenium/webdriver/common/credential.rb
TamsilAmani: I have made the arguments as positional. But in context of default values there's a doubt.
Required parameters are : `id, rp_id, private_key, sign_count` 

The remaining two parameter i.e `is_resident_credential` and `user_handle` are used in combination. For example,
When `is_resident_credential` = true, `user_handle` != nil
When `is_resident_credential` = false, `user_handle` == nil

I don't suggest making either of them optional as case may arise when user puts `is_resident_credential` as true but doesn't assign `user_handle`.

rb/lib/selenium/webdriver/remote/bridge.rb
TamsilAmani: Changes incorporated.

rb/lib/selenium/webdriver/remote/bridge.rb
TamsilAmani: Changes incorporated. Thank you.

rb/lib/selenium/webdriver/common/virtual_authenticator_options.rb
TamsilAmani: Done.

rb/lib/selenium/webdriver/common/virtual_authenticator_options.rb
TamsilAmani: Done. Thank you for the heads up.

rb/lib/selenium/webdriver/common/virtual_authenticator_options.rb
TamsilAmani: Included the changes.

rb/lib/selenium/webdriver/common/driver.rb
TamsilAmani: @p0deje as discussed with @AutomatedTester, virtual authenticator is present only in chrome for now. It may happen in future that firefox and other browsers implement it. So to keep the code changes minimal in future for other browsers, I have added virtual authenticator directly under the driver. And tests are run only in chrome, not in firefox or others. If for example, firefox decided to implement it, then we only need to include it in the test suite.

`# @examples` will be added in the documentation. That will be a separate PR. Is that what you mean?

rb/lib/selenium/webdriver/common/credential.rb
TamsilAmani: Done.

rb/lib/selenium/webdriver/common/driver.rb
TamsilAmani: @titusfortner I have created a separate class `VirtualAuthenticator.rb` and all the methods have been taken there as you suggested. Please review for changes. We can now directly call:

```
authenticator.add_credential
authenticator.remove_credential
```
instead of 
```
driver.virtual_authenticator.add
driver.virtual_authenticator.remove
```

This approach will help to add multiple authenticators.

rb/lib/selenium/webdriver/remote/bridge.rb
TamsilAmani: I cleared bridge of the authenticator methods and created a separate class `VirtualAuthenticator` as you suggested. Now we can get and remove the authenticator object without storing its id in bridge.

rb/spec/integration/selenium/webdriver/virtual_authenticator_spec.rb
TamsilAmani: When I didn't use any `exclusive` parameter in my test, it _seemed_ that the test ran only on chrome. But I have made changes and used `exclusive: {browser: %i[chrome]}` now, and the test run only in chrome. For firefox and others, they are shown pending.

rb/lib/selenium/webdriver/common/credential.rb
TamsilAmani: Made the changes as per my understanding. Please review.

rb/lib/selenium/webdriver/common/credential.rb
TamsilAmani: Changes done.

rb/lib/selenium/webdriver/common/credential.rb
TamsilAmani: Seems right! Thank you @p0deje and @titusfortner. Made the change.

rb/lib/selenium/webdriver/common/virtual_authenticator_options.rb
TamsilAmani: Done.

rb/lib/selenium/webdriver/remote/bridge.rb
p0deje: I see, thank you for explanation.

rb/lib/selenium/webdriver/common/credential.rb
p0deje: In this case, I would make only `id, rp_id, private_key, sign_count` required and then add additional validations for arguments that are connected:

```ruby
msg = "Both is_resident_credential and user_handle should be passed together"
raise ArgumentError, msg if is_resident_credential && user_handle.nil?
raise ArgumentError, msg if !is_resident_credential && !user_handle.nil?
```

p0deje (approved):

titusfortner:
Thanks for your work on this. It's looking good!

I'm going to merge it, then update things to get this released in 4.4, but for reference.
* `#as_json` needs to have keys as `String` not `Symbol` by convention.
* Predicate methods explicitly avoid is_ and has_ prefixes and use a `?` at the end (my approach is to use attr_accessor without the prefix, then alias the getter with a `?`)
* Ruby tries to be less verbose so long as the return is "obvious." So `Credential.resident` class method that returns an instance is obviously the same as the. more verbose `Credential.create_resident_credential`
* For the `Bridge#execute` method, there are 3 parameters: command name, the variables in the endpoint that need to be replaced (e.g. ":authenticatorId" with the actual ID), and the key/values that go into the payload. Some of these were mixed up in this PR

Then I'm making a couple other personal preference changes, just because. :-D

PR #10904, opened 2022-07-26
[py] Normalized finding by class (not via CSS selector)

SecorD0:
### Description
⠀Good day!
⠀I use Selenium in several of my projects and I often encounter classes consisting of several words separated by spaces. When I do a search on such classes I get nothing, because now the class search is done through the CSS selector. So I applied the proposed commit to my local libraries and it works logically and as it should work.

### Motivation and Context
⠀When developer want to find an element by class name, which can be several words separated by spaces, he want find it by class name, not CSS selector. If he want find element by CSS selector he'll just do it.

### Types of changes

### Checklist


titusfortner:
W3c does not support class names or attributes directly any more. As such, to maintain backwards compatibility with previous API, we have to convert to these things to css. Chrome might still work with old selectors, but Firefox and Safari shouldn't.

Issue #10448, opened 2022-03-15
[🐛 Bug]: Setting chrome to w3c: false in Selenium 4 causes errors


marcoadkins:
### What happened?

Upon upgrading from selenium 3.142 to 4.1, the driver started throwing a `Selenium::WebDriver::Error::ScriptTimeoutError` when attempting to reset between tests via `driver.reset!`

### How can we reproduce the issue?

```shell
No specific details to reproduce. I can provide our configuration though that might be able to help replicate the problem.


# Driver setup
Capybara.register_driver :selenium_chrome_headless do |app|
  capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
    acceptInsecureCerts: true,
    'goog:chromeOptions': {
      args: %W[
        headless
        proxy-server=#{ENV['CI'] ? 'localhost' : ENV.fetch('NGINX_IP')}:#{Billy.proxy.port}
        window-size=1440,900
      ],
      prefs: {
        download: {
          default_directory: DownloadSpecHelpers::CHROME_PATH
        }
      },
      w3c: false
    },
    'goog:loggingPrefs': {
      browser: 'ALL'
    }
  )

  options = {
    browser: ENV['CI'] ? :chrome : :remote,
    capabilities: capabilities,
    url: 'http://localhost:4445/wd/hub'
  }

  Capybara::Selenium::Driver.new(app, options).tap do |d|
    unless ENV['CI']
      d.browser.file_detector = lambda do |args|
        str = args.first.to_s
        str if File.exist?(str)
      end
    end
  end
end
```

### Relevant log output

```shell
Example Failure:
(byebug) Capybara.current_session.driver.reset!
*** Selenium::WebDriver::Error::ScriptTimeoutError Exception: Unable to execute request for an existing session: POST /session/9c2e8f72783418513a46e4354f9fd08f/goog/cdp/execute
Build info: version: '4.1.0', revision: '87802e897b'
System info: host: 'c202f6697df7', ip: '172.18.0.11', os.name: 'Linux', os.arch: 'amd64', os.version: '5.10.47-linuxkit', java.version: '11.0.11'
Driver info: driver.version: unknown


Deeper Stacktrace triggered in rspec run:
     1.1) Failure/Error: Capybara.reset! unless example.exception (rails_helper.rb:89)

          Selenium::WebDriver::Error::ScriptTimeoutError:
[...]

PR #10918, opened 2022-08-03
[ruby] fix: setting w3c: false throws error

TamsilAmani:
### Description

`w3c: false` does not work in ruby binding. Setting this key as false will throw an error. 

### Motivation and Context

https://github.com/SeleniumHQ/selenium/issues/10448#issuecomment-1149870933

### Types of changes

### Checklist

### DIFF
--- a/rb/lib/selenium/webdriver/remote/bridge.rb
+++ b/rb/lib/selenium/webdriver/remote/bridge.rb
@@ -49,6 +49,7 @@ module Selenium
         #
 
         def create_session(capabilities)
+          check_chrome_w3c_false(capabilities)
           response = execute(:new_session, {}, prepare_capabilities_payload(capabilities))
 
           @session_id = response['sessionId']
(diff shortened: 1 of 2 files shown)


TamsilAmani:
As per @titusfortner, I have only thrown Error when `w3c` is `false`. And do nothing if `w3c` is `true`.
As per @AutomatedTester, he suggested to give a warning not to use `w3c` if user sets `w3c` as `true`.

Do let me know which approach to follow. Currently it is done without warning.

p0deje (approved):

p0deje:
> As per @titusfortner, I have only thrown Error when w3c is false. And do nothing if w3c is true.
> As per @AutomatedTester, he suggested to give a warning not to use w3c if user sets w3c as true.
> 
> Do let me know which approach to follow. Currently it is done without warning.

We can do both, let's just check if capabilities had w3c passed explicitly and then call `Selenium::WebDriver.logger.warn` with a warning.

titusfortner (changes requested):
Does this code also work for setting it on options class? Please add the test to verify. We don't actually care about capabilities class, tbh. 

Ruby is the only bindings supporting both capabilities and options still and one way or another, and it's on my list to fix when I get the time.

TamsilAmani:
@titusfortner Shall I replace Capability class test with the Options test? Or add Options as a new test?

PR #10932, opened 2022-08-10
Deprecation notice about urllib3[secure] in setup.py

fyunusa:
### Description

pyOpenSSL and urllib3[secure] are deprecated in the upcoming release (1.26.12)
https://github.com/urllib3/urllib3/issues/2680
Removed [secure] extra from the urllib3 dependency in setup.py
Changed 'urllib3[secure, socks]~=1.26 to 'urllib3[socks]~=1.26'


symonk:
Thanks @umarfarouk98, its slightly more complicated in the project due to multi languages and using `bazel`.  I think we also need to change the following here: https://github.com/SeleniumHQ/selenium/blob/trunk/py/BUILD.bazel

symonk (approved):
Thanks!

fyunusa:
> Thanks!

@symonk you wlcm.
Anything else i need to do on the request for it to get merge just let me know

symonk:
I couldn't find any reference to the secure extra in any requirements files either, so I think this LGTM (my bazel knowledge is very light :smile: )

fyunusa:
> I couldn't find any reference to the secure extra in any requirements files either, so I think this LGTM (my bazel knowledge is very light 😄 )

it's not @symonk it's true that the Build.bazel contained the secure[extra] dependency initially but i removed it already.
Screenshot 2022-08-14 at 9 34 20 AM

PR #10938, opened 2022-08-12
Ie find edge

titusfortner:
Full disclosure: I don't know C++
Narrator: This did not stop him from "writing" C++ code

fwiw, I think we should get rid of all standalone IE Support in future versions of IE Driver and *only support IE Mode in Edge; I'm not sure the best way to go about this.

In the meantime, if a user specifies to use Edge, the driver should attempt to automatically locate it.

Further, if `iexplore.exe` is set to automatically redirect to Edge (i.e., IE is effectively not installed), then it should default to using Edge.

This will allow a user to get an Edge session without adding any additional flags on Windows 11 and any Win 10 box that no longer uses IE.

I've built this and tested that it starts the session the way I expect, but I'm still hitting https://github.com/MicrosoftEdge/EdgeWebDriver/issues/38 so I can't do a lot of testing right now.

@jimevans or @bwalderman if you have any feedback for me, please let me know.

### DIFF
--- a/cpp/iedriver/BrowserFactory.cpp
+++ b/cpp/iedriver/BrowserFactory.cpp
@@ -45,6 +45,7 @@
 
 #define EDGE_REGISTRY_KEY L"Software\\Microsoft\\Windows\\CurrentVersion\\App Paths\\msedge.exe"
 #define IE_CLSID_REGISTRY_KEY L"SOFTWARE\\Classes\\InternetExplorer.Application\\CLSID"
+#define IE_REDIRECT L"Software\\Microsoft\\Windows\\CurrentVersion\\Policies\\Ext\\CLSID"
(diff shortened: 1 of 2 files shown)


bwalderman:
I'm on vacation through Sep 5 so I won't be available to code review or test. I agree with all of the goals in the PR description though. Thanks for putting this together!

SuhailFluke (approved):

titusfortner:
This is what I built: (32 bit; set to v4.3.0.2)
[IEDriverServer.exe.zip](https://github.com/SeleniumHQ/selenium/files/9509363/IEDriverServer.exe.zip)

titusfortner:
I can't test this because of https://github.com/MicrosoftEdge/EdgeWebDriver/issues/38 and various VM issues in general...
@bwalderman hope you enjoyed your time away; are you available to check my work? Thanks.

bwalderman (approved):

bwalderman:
@titusfortner tested your build on my machine and it works as expected with and without a provided Edge location. Thank you for your patience while I was away.

PR #10958, opened 2022-08-18
Update explicit dependency for "certifi" to remove upper bound

### LINKED ISSUE #10956: [🐛 Bug]: (Python) selenium 4.4.2 requires a 2021 version of the "certifi" dependency.
### What happened?

``selenium`` ``4.4.2`` requires a 2021 version of the ``certifi`` dependency.
This is why: ``certifi~=2021.10.8`` - https://github.com/SeleniumHQ/selenium/commit/54809a701b9cc0306f7277efd3767b02e71be2e5#diff-3e8ff3a52e8e6c0b195ea328e17c50ba6d90abca1ce1624110607da20691d7a9

To allow more flexibility, that should be changed to ``certifi>=2021.10.8``, so that the upper bound is removed.

History of how the issue happened:
* selenium versions ``4.4.0`` and below had ``urllib3[secure]`` as a dependency, which installed ``certifi`` as an unpinned dependency.
* In selenium ``4.4.1``, the ``urllib3[secure]`` dependency was removed, and ``certifi`` was no longer installed at all, which led to issues because it was still needed.
* selenium ``4.4.2`` added ``certifi~=2021.10.8`` as a dependency, but not the same as it was before. This not only added a minimum bound (which is OK), but it also added an upper bound (problematic). The upper bound means that users need to use a ``2021`` version of ``cerifi``, which does not include the improvements and bug fixes made in ``2022``.

```shell
Have ``certifi>=2022.6.15`` as a dependency in your project. Also have ``selenium==4.4.2`` as a dependency in your project.
```

### Relevant log output

```shell
ERROR: Cannot install selenium and SOME_PROJECT because these package versions have conflicting dependencies.

The conflict is caused by:
    SOME_PROJECT depends on certifi>=2022.6.15
    selenium 4.4.2 depends on certifi~=2021.10.8

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```
```


### Operating System

-

### Selenium version

4.4.2

-

-

-

### PR DESCRIPTION
**Pull request by mdmintz**
This resolves https://github.com/SeleniumHQ/selenium/issues/10956 - ``selenium 4.4.2 requires a 2021 version of the "certifi" dependency``

### Types of changes

### Checklist

### DIFF
--- a/py/setup.py
+++ b/py/setup.py
@@ -27,7 +27,7 @@ for scheme in INSTALL_SCHEMES.values():
 setup_args = {
     'cmdclass': {'install': install},
     'name': 'selenium',
-    'version': "4.4.2",
+    'version': "4.4.3",
     'license': 'Apache 2.0',
     'description': 'Python bindings for Selenium',
     'long_description': open(join(abspath(dirname(__file__)), "README.rst")).read(),
--- a/py/setup.py
+++ b/py/setup.py
@@ -70,7 +70,12 @@ setup_args = {
                  'selenium.webdriver.remote',
                  'selenium.webdriver.support', ],
     'include_package_data': True,
-    'install_requires': ["urllib3[socks]~=1.26", "trio~=0.17", "trio-websocket~=0.9"],
+    'install_requires': [
+        "urllib3[socks]~=1.26",
+        "trio~=0.17",
+        "trio-websocket~=0.9",
+        "certifi>=2021.10.8",
+    ],
     'zip_safe': False
 }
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: py/setup.py [hunk 1]
```diff
@@ -27,7 +27,7 @@
 setup_args = {
     'cmdclass': {'install': install},
     'name': 'selenium',
-    'version': "4.4.2",
+    'version': "4.4.3",
```
symonk, line 30: LGTM 👍 , we should probably pin `certifi` inside the `install_requires` in here, tho I don't think this file is fully used, good to keep in sync

symonk (approved):
LGTM, please squash merge, I've granted permission for CI to run

PR #10962, opened 2022-08-20
feat(java): add webdriver.edge.loglevel

### LINKED ISSUE #10961: [🚀 Feature]: Java(EdgeDriverService): add new system property `webdriver.edge.loglevel`
## Overview

EdgeDriverService in Selenium Java binding has six properties; `webdriver.edge.driver`, `webdriver.edge.logfile`, `webdriver.edge.verboseLogging`, `webdriver.edge.silentOutput`, `webdriver.edge.withAllowedIps`, and `webdriver.edge.disableBuildCheck`. These options are very helpful in operating Edge server nodes.

https://github.com/hgsgtk/selenium/blob/036ee7f6aec797f5fe4608d2f7673e59d7f548b5/java/src/org/openqa/selenium/edge/EdgeDriverService.java#L46

One feature I feel missing is that there are no APIs to customize the log level. 

## Current possible solutions

For example, when I want to set the log-level=INFO, I have only one option, to set the --log-level option to msedgedriver directly.

```
$ ./msedgedriver --log-level=INFO
```

However, this way has some drawbacks.

First, it requires us a complex flow which combines with the system property `webdriver.edge.driver` and some wrapped scripts.

For example: set a wrapper batch file to the webdriver.edge.driver property.

```
java -Dwebdriver.edge.driver=edgedriver_wrapper.bat
```

And then, a wrapped batch file execute msedgedriver with the --log-level option.

```
C:\path\msedgedriver.exe --log-level $args
```

Another drawback is that command option settings must be written in multiple places. I understand that it doesn't make sense to support ALL options (e.g. an experimental option --replayable). But I believe the --log-level is commonly helpful, as is the --verbose option. 

In addition, the property `webdriver.edge.verboseLogging` is not a perfect solution because it indicates to set the option `--verbose` option (equivalent to --log-level=ALL).

### Usage example

We can use the new system property like this:

```
System.setProperty("webdriver.edge.loglevel", "INFO");
EdgeDriverService service = EdgeDriverService.createDefaultService();

EdgeDriver driver = new EdgeDriver(service);
```

hgsgtk:
### Description

It adds a new system property, `webdriver.edge.loglevel` on EdgeDriverService.
It allows us to configure the log level in detail.

For example, if you use version 104.0.1293.63 Microsoft Edge WebDriver, you can choose one from ALL, DEBUG, INFO, WARNING, SEVERE, or OFF.

```
$ ./msedgedriver --help

  --log-level=LEVEL  set log level: ALL, DEBUG, INFO, WARNING, SEVERE, OFF
```

What this system property does is to make sure to specify the `--log-level` option like the others.

#### Detail: validation of given string

When users provide an invalid log level, MsEdgeDriver will be terminated due to an error; `Invalid --log-level value.`

```
$ ./msedgedriver --log-level=UNKNOWN
Starting Microsoft Edge WebDriver 104.0.1293.63 (7dd22b4a7f122e6953a166b0388edf494d716c7d) on port 9515
To submit feedback, report a bug, or suggest new features, please visit https://github.com/MicrosoftEdge/EdgeWebDriver

Please see https://aka.ms/WebDriverSecurity for suggestions on keeping Microsoft Edge WebDriver safe.

Invalid --log-level value.
Unable to initialize logging. Exiting...
```

I wondered if I should validate if a given value is valid, but I didn't implement it because I was concerned that support costs would be high. For example, it is an extreme case, but we might need to have a log level list for each version when MsEdgeDriver adds or removes log levels, etc.

### Motivation and Context

Fixes #10961

### Types of changes

### Checklist

    - I skipped adding automated tests because it looks technically tricky, and other properties don't have a test.


java/src/org/openqa/selenium/edge/EdgeDriverService.java
```diff
@@ -148,6 +154,14 @@ public EdgeDriverService.Builder withVerbose(boolean verbose) {
       return this;
     }
 
+    /**
+     * Configures the driver server log level.
+     */
+    public EdgeDriverService.Builder withLoglevel(String level) {
```
hgsgtk, line 160: FYI: When users provide an invalid log level, MsEdgeDriver will be terminated due to an error; `Invalid --log-level value.`

```
$ ./msedgedriver --log-level=UNKNOWN
Starting Microsoft Edge WebDriver 104.0.1293.63 (7dd22b4a7f122e6953a166b0388edf494d716c7d) on port 9515
To submit feedback, report a bug, or suggest new features, please visit https://github.com/MicrosoftEdge/EdgeWebDriver

Please see https://aka.ms/WebDriverSecurity for suggestions on keeping Microsoft Edge WebDriver safe.

Invalid --log-level value.
Unable to initialize logging. Exiting...
```

I wondered if I should validate if a given value is valid, but I didn't implement it because I was concerned that support costs would be high. For example, it is an extreme case, but we might need to have a log level list for each version when MsEdgeDriver adds or removes log levels, etc.

diemol (approved):
Thank you, @hgsgtk!

PR #10964, opened 2022-08-21
chore/code style changes to common format

potapovDim:

nodejs selenium-webdriver lib code style changes to common format

### Description
nodejs selenium-webdriver lib code style changes to common format

### Motivation and Context
Align code style to common format 

### Types of changes

### Checklist

### DIFF
--- a/javascript/node/selenium-webdriver/io/exec.js
+++ b/javascript/node/selenium-webdriver/io/exec.js
@@ -121,7 +121,7 @@ class Command {
  * @param {Options=} opt_options The command options.
  * @return {!Command} The launched command.
  */
-module.exports = function exec(command, opt_options) {
+function exec(command, opt_options) {
   const options = opt_options || {}
 
   let proc = childProcess.spawn(command, options.args || [], {
@@ -162,4 +162,5 @@ module.exports = {
   Command,
   Options,
   Result,
+  exec
 }
--- a/javascript/node/selenium-webdriver/io/exec.js
+++ b/javascript/node/selenium-webdriver/io/exec.js
@@ -158,6 +158,8 @@ module.exports = function exec(command, opt_options) {
 
 // Exported to improve generated API documentation.
 
-module.exports.Command = Command
-module.exports.Options = Options
-module.exports.Result = Result
+module.exports = {
+  Command,
+  Options,
+  Result,
+}
--- a/javascript/node/selenium-webdriver/lib/virtual_authenticator.js
+++ b/javascript/node/selenium-webdriver/lib/virtual_authenticator.js
@@ -22,8 +22,8 @@
  * @enum {string}
  */
 const Protocol = {
-  CTAP2: 'ctap2',
-  U2F: 'ctap1/u2f',
+  'CTAP2': 'ctap2',
+  'U2F': 'ctap1/u2f',
 }
 
 /**
@@ -31,10 +31,10 @@ const Protocol = {
  * @enum {string}
  */
 const Transport = {
-  BLE: 'ble',
-  USB: 'usb',
-  NFC: 'nfc',
-  INTERNAL: 'internal',
+  'BLE': 'ble',
+  'USB': 'usb',
+  'NFC': 'nfc',
+  'INTERNAL': 'internal',
 }
 
 /**
--- a/javascript/node/selenium-webdriver/lib/virtual_authenticator.js
+++ b/javascript/node/selenium-webdriver/lib/virtual_authenticator.js
@@ -22,8 +22,8 @@
  * @enum {string}
  */
 const Protocol = {
-  'CTAP2': 'ctap2',
-  'U2F': 'ctap1/u2f',
+  CTAP2: 'ctap2',
+  U2F: 'ctap1/u2f',
 }
 
 /**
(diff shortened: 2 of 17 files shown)


javascript/node/selenium-webdriver/lib/virtual_authenticator.js
```diff
@@ -22,27 +22,26 @@
  * @enum {string}
  */
 const Protocol = {
-  'CTAP2': 'ctap2',
```
diemol, line 25: Some tests use this, like `virtualAuthenticator_test.js`, so this is probably why CI is not passing. Can you please double check?

Same for `Transport` below.

diemol (dismissed):

javascript/node/selenium-webdriver/lib/virtual_authenticator.js
potapovDim: @diemol yes sure, let me check that

potapovDim: @diemol i checked tests locally and everything is fine 
```../../../bazel-bin/javascript/node/selenium-webdriver/lib/atoms/get-attribute.js
../../../bazel-bin/javascript/node/selenium-webdriver/lib/atoms/is-displayed.js
../../../bazel-bin/javascript/node/selenium-webdriver/lib/atoms/find-elements.js
[INFO] Using Selenium remote end: http://localhost:4444/wd/hub/
[INFO] Running tests against [chrome]


  [chrome]
    VirtualAuthenticator Test Suit 2
      ✓ should test create authenticator (352ms)
      ✓ should test remove authenticator
      ✓ should test add non-resident credential (220ms)
      ✓ should test add non-resident credential when authenticator uses U2F protocol (220ms)
      ✓ should test add resident credential (231ms)
      ✓ should test add resident credential not supported when authenticator uses U2F protocol
      ✓ should test get credentials (349ms)
      ✓ should test remove credential by rawID (352ms)
      ✓ should test remove credential by base64url Id (348ms)
      ✓ should test remove all credentials (467ms)
      ✓ should test set user verified (465ms)


  11 passing (21s)```

**Comment by potapovDim** [comment]
@harsha509 @diemol  hey guys, can you please clarify few moments related to nodejs ? 
we have two test runners in selenium-webdriver - mocha and jasmine, can we use one, or why we should use both  ?
we have many variables declaration by `var` keyword - do we want to move fully to `let/const`?
we have `class Select` and it has async call in constructor which allows me to use this class only with classic select lists, but not with some customs select list with same behavior but another tags, can we for example create another library i.e selenium-webdriver-extensions with classes like `Select`?
we have prettier and lint, but seems like we do not use it 100% ?

**Comment by diemol** [comment]
> @harsha509 @diemol hey guys, can you please clarify few moments related to nodejs ? we have two test runners in selenium-webdriver - mocha and jasmine, can we use one, or why we should use both ? we have many variables declaration by `var` keyword - do we want to move fully to `let/const`? we have `class Select` and it has async call in constructor which allows me to use this class only with classic select lists, but not with some customs select list with same behavior but another tags, can we for example create another library i.e selenium-webdriver-extensions with classes like `Select`? we have prettier and lint, but seems like we do not use it 100% ?

I'd be easier if you open issues/feature requests for each item and then we can track each one of them separately. Let's complete this PR and get it merged.

File: javascript/node/selenium-webdriver/test/devtools_test.js [hunk 2]
```diff
@@ -35,15 +35,18 @@ suite(
...
     ignore(browsers(Browser.CHROME)).it(
-      'sends Page.enable command using devtools', async function () {
+      'sends Page.enable command using devtools',
+      async function () {
         const cdpConnection = await driver.createCDPConnection('page')
         cdpConnection.execute('Page.enable', {}, function (_res, err) {
           assert(!err)
         })
-      })
+      }
```
harsha509, line 44: HI @potapovDim,

Changes looks good!

But in few places I see this `})` is moved to multi-line

`})` is more readable I think!

potapovDim: @harsha509 those changes were done based on prettier and eslint codestyle formats, i can update it, but wanted to follow formats

javascript/node/selenium-webdriver/io/exec.js
```diff
@@ -158,6 +158,8 @@ module.exports = function exec(command, opt_options) {
 
 // Exported to improve generated API documentation.
 
-module.exports.Command = Command
-module.exports.Options = Options
-module.exports.Result = Result
+module.exports = {
```
harsha509: looks like there is an issue in exporting modules this way!

harsha509: @potapovDim can you recheck!

harsha509: This commit https://github.com/SeleniumHQ/selenium/commit/f4ce77d24ee364279326c87a5ddc8a3ea386b88b should fix and run tests correctly!

potapovDim: @harsha509 you are right! my bad, thanks for fix !

harsha509 (approved):
Thank you @potapovDim!

PR #10967, opened 2022-08-21
Fix frame_to_be_available_and_switch_to_it() for string inputs

### LINKED ISSUE #10963: [🐛 Bug]: Invalid handling of strings as iterables in `frame_to_be_available_and_switch_to_it`
### What happened?

In selenium webdriver/support/expected_conditions py line 247 The _predicate function in frame_to_be_available_and_switch_to_it is incorrect for locator. When using locator as an ID, it will also enter the first IF branch and trigger an exception similar to the following: "TypeError: find_element() takes from 1 to 3 positional arguments but 8 were given" We should check the locator type. If it is a character, we should go to the else branch.

```shell
wait.until(ec.frame_to_be_available_and_switch_to_it("iframe3"))
```

### Relevant log output

```shell
File "/xxxxxx/case.py", line 32, in test_add
    wait.until(ec.frame_to_be_available_and_switch_to_it("iframe3"))
  File "/xxxxxx/venv/lib/python3.9/site-packages/selenium/webdriver/support/wait.py", line 81, in until
    value = method(self._driver)
  File "/xxxxxx/venv/lib/python3.9/site-packages/selenium/webdriver/support/expected_conditions.py", line 256, in _predicate
    driver.switch_to.frame(driver.find_element(*locator))
TypeError: find_element() takes from 1 to 3 positional arguments but 8 were given
```

### Operating System

Mac

### Selenium version

python

all

all

all

mdmintz:
Fix ``frame_to_be_available_and_switch_to_it()`` for string inputs.
* This resolves https://github.com/SeleniumHQ/selenium/issues/10963

### Description
Fix ``frame_to_be_available_and_switch_to_it()`` for string inputs. Currently, it only works with numeric inputs (see https://github.com/SeleniumHQ/selenium/issues/10963 for details)

### Types of changes

### Checklist

- [X] All new tests passed. Existing tests that were passing are still passing.


py/CHANGES
```diff
@@ -1,3 +1,6 @@
+Selenium 4.4.4
```
symonk: :+1: Thanks - not sure if worthy of a patch release, likely be part of `4.5.0`? will let the folks decide that one!

py/selenium/webdriver/support/expected_conditions.py
```diff
@@ -252,7 +252,7 @@ def frame_to_be_available_and_switch_to_it(locator):
 
     def _predicate(driver):
         try:
-            if hasattr(locator, '__iter__'):
+            if hasattr(locator, '__iter__') and type(locator) is not str:
```
symonk: I think we should try:

 - getting an iterator from locator using `iter(locator)` to support old style iterator protocol via `__getitem__` (if locator is not a str)

symonk: why `type(locator) is not str`, seems closed for extension, `not isinstance(locator, str)` seems more appropriate, wouldn't be unheard of for some people to subclass strings as rare as that might be we'd be unfairly limiting users and violating liskov on their behalf

py/test/selenium/webdriver/common/frame_switching_tests.py
```diff
@@ -351,6 +351,22 @@ def test_should_be_able_to_switch_to_the_top_if_the_frame_is_deleted_from_under_
     WebDriverWait(driver, 3).until(EC.presence_of_element_located((By.ID, "success")))
 
 
+def test_frame_to_be_available_and_switch_to_it_using_string_inputs(driver, pages):
```
symonk, line 354: is there existing test for non str iterables successfully unpacked to `find_element(...)` (if not could we add one)

symonk (changes requested):

py/test/selenium/webdriver/common/frame_switching_tests.py
mdmintz: The test added was the missing link to why the bug happened in the first place. Looks like the other areas are covered.

py/CHANGES
mdmintz: For a small fix like this, ``4.4.4`` could buy some time until the bigger ``4.5.0`` release is ready.

py/selenium/webdriver/support/expected_conditions.py
mdmintz: Updated in new commit.

py/BUILD.bazel
```diff
@@ -15,7 +15,7 @@ compile_pip_requirements(
     requirements_txt = ":requirements_lock.txt",
 )
 
-SE_VERSION = "4.4.3"
+SE_VERSION = "4.4.4"
```
diemol: This does not need to be bumped yet.

py/CHANGES
diemol: We are releasing 4.5.0 next week, so not sure if we will release 4.4.4

py/selenium/__init__.py
```diff
@@ -16,4 +16,4 @@
 # under the License.
 
 
-__version__ = "4.4.3"
+__version__ = "4.4.4"
```
diemol: Please do not bump the version as we do not know if this will be 4.4.4 or 4.5.0

py/selenium/webdriver/__init__.py
```diff
@@ -36,7 +36,7 @@
 from .common.proxy import Proxy  # noqa
 from .common.keys import Keys  # noqa
 
-__version__ = '4.4.3'
+__version__ = '4.4.4'
```
diemol: Please do not bump the version as we do not know if this will be 4.4.4 or 4.5.0

py/setup.py
```diff
@@ -27,7 +27,7 @@
 setup_args = {
     'cmdclass': {'install': install},
     'name': 'selenium',
-    'version': "4.4.3",
+    'version': "4.4.4",
```
diemol: Please do not bump the version as we do not know if this will be 4.4.4 or 4.5.0

diemol (changes requested):
Left some comments.

diemol (commented):
CI is passing, we just need to revert the version changes to merge.

diemol (approved):
Thank you, @mdmintz!

PR #10973, opened 2022-08-25
[ruby] Remove unnecessary and bug-creating code in platform.rb.

ch-saeki:
### Description

When calling selenium's `cygwin?` method from ruby on cygwin, regexp match results was returning different judgment result from the actual one, so it was removed.

### Motivation and Context

I ran it from cygwin, but it didn't return the correct results, so I looked into it.

### Types of changes

### Checklist


titusfortner:
Ah, yes, looks like successive rubocop updates broke this. Thanks.

PR #10985, opened 2022-08-29
[java] Set protected as ClientConfig constructor instead of private

KazuCocoa:

### Description

I'd like to define a class that inherits `ClientConfig` in Appium java client that includes a few flags for Appium Java client. Current the constructor is `private`, so could we change this to `protected`?

### Motivation and Context

We also considered defining an appium java client class that has `ClientConfig` as the member, but it could be confused for users to use both, ClientConfig and Appium java client config. So... for now, ideally, we'd like to define a subclass as AppiumClientConfig that has the ClientConfig plus a few Appium's client specific configuration.

### Types of changes

### Checklist


KazuCocoa:
cc @titusfortner

diemol (approved):
Thank you, @KazuCocoa!

PR #11009, opened 2022-09-06
fix some typos in code and documentation

asolntsev:
Just a minor code improvement, doesn't affect functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ bazel test //java/... --test_size_filters=small,medium,large --test_tag_filters=
 ```
 
 The `test_size_filters` argument takes small, medium, large. Small are akin to unit tests,
-medium is akin to integration tests, and large is akin to end to end tests.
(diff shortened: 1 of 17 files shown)


pujagani:
Thank you for your contribution. I understand it is all typos but they sit in different parts of the code. Maybe some 3rd party related code. I would appreciate it if you can break it down to small PRs. It will be easier to review and merge it in. Thank you!

asolntsev:
@pujagani Yes, I could revert changes in 3rd party code (like `dojo.js.uncompressed.js` and `third_party/dotnet/nunit-console-3.12.0/CHANGES.txt`). That seems reasonable. There was no need to fix 3rd party code. 

But I don't see how splitting this PR to multiple PRs helps to review it. These PR would contains same amount of changes and take same amount of time to review. In the end, this will create only more hassle.

titusfortner:
thanks!

PR #11010, opened 2022-09-06
remove some unneeded try/catch/fail constructs in tests

asolntsev:


pujagani (approved):
Thank you @asolntsev!

PR #11014, opened 2022-09-06
use StringBuilder instead of StringBuffer

asolntsev:
minor refactoring

### DIFF
--- a/java/src/com/thoughtworks/selenium/DefaultRemoteCommand.java
+++ b/java/src/com/thoughtworks/selenium/DefaultRemoteCommand.java
@@ -52,12 +52,12 @@ public class DefaultRemoteCommand implements RemoteCommand {
 
   @Override
   public String getCommandURLString() {
-    StringBuffer sb = new StringBuffer("cmd=");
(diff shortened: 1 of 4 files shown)


titusfortner:
Isn't `StringBuffer` preferred for thread safety?

pujagani:
> Isn't `StringBuffer` preferred for thread safety?

You are right, Titus. However, the code changes made in this pull request are in the methods that are not marked synchronized nor do they use locks so it is safe to assume that multiple threads do not access them simultaneously. `StringBuilder` is recommended performance-wise in such cases.

pujagani (approved):
Thank you, @asolntsev!

asolntsev:
@titusfortner @pujagani No, in most cases `StringBuilder` is preferred over `StringBuffer` because it doesn't use unneeded synchronisation. Almost always it's used as a local variable inside of a method, thus not needing any synchronisation.

`StringBuffer` does synchronisation, thus it's slower, and it's needed only if it's used in a multithreaded context. Which means: almost never.

PR #11015, opened 2022-09-06
Throw message with Status if response value is empty

RafTim:

### Description
Empty message exceptions aren't helpful, so at least give the user a clue about the returned status which caused the error

### Motivation and Context
in my case I had a 401 Unauthorized error which led to confusion. 

Better error handling, giving a clue about an occured exception instead of

selenium.common.exceptions.WebDriverException: Message: 

### Types of changes

### Checklist


py/selenium/webdriver/remote/errorhandler.py
```diff
@@ -204,6 +204,8 @@ def check_response(self, response: Dict[str, Any]) -> None:
         if not value:
             value = response['value']
         if isinstance(value, str):
+            if len(value) == 0:
```
symonk, line 207: could we use:

```python
if not value:
    ...
```

could we add a test for this please?

titusfortner (changes requested):
I agree that the current message isn't helpful, but I'm not sure this is the right fix. This needs to include one or more tests for when this condition is met so we can better determine any underlying issues. @RafTim is this something you can help with?

py/selenium/webdriver/remote/errorhandler.py
RafTim, line 207:
```

RafTim:
@titusfortner Sorry for the late reply. I have to admit that I'm really not used to this github workflows/ comments, change requests and other stuff... I just had this issue, debugged it and tried to make a minimal invasive commit. 
So if someone could tell me what needs to be done I would look into it.

titusfortner:
@RafTim first off, we appreciate you taking the time to do that. Can you explain the scenario for the error better?

diemol:
Will close this as we did not get enough information to understand the context better.

PR #11024, opened 2022-09-13
[build] Re-enabling browser-actions/setup-geckodriver@latest

whimboo:
### Description
[mozilla/geckodriver#2009](https://github.com/mozilla/geckodriver/issues/2009) got fixed a while ago and getting the latest geckodriver binary is working again.

### DIFF
--- a/.github/actions/setup-firefox/action.yml
+++ b/.github/actions/setup-firefox/action.yml
@@ -12,9 +12,5 @@ runs:
       uses: browser-actions/setup-firefox@latest
       with:
         firefox-version: ${{ inputs.version }}
-    - name: Check GeckoDriver
-      run: geckodriver --version
-      shell: bash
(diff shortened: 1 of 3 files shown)


whimboo:
@titusfortner or @diemol could you please review? Thanks

titusfortner:
Ruby & Python weren't even using this so I added it. Let me make sure it runs for everything.

whimboo:
> Ruby & Python weren't even using this so I added it. Let me make sure it runs for everything.

I just noticed that myself for my investigation on #11025. Thanks for getting these added!

titusfortner:
We've got a really cool filter to only run builds when changes are made... except it then becomes difficult to force the builds to be run. I've pushed this branch to selenium/latest_geckodriver and am forcing runs just to make sure we have everything.

titusfortner:
Verified this works, thanks!

PR #11025, opened 2022-09-13
Run fluxbox as window manager for jobs using Xvfb

whimboo:
PR for regression testing for https://github.com/mozilla/geckodriver/issues/2042.

### DIFF
--- a/.github/workflows/ci-java.yml
+++ b/.github/workflows/ci-java.yml
@@ -171,6 +171,8 @@ jobs:
             ${{ runner.os }}-bazel-build-
       - name: Setup bazelisk
         uses: ./.github/actions/setup-bazelisk
+      - name: Setup Fluxbox
+        run: sudo apt-get -y install fluxbox
       - name: Setup Java
         uses: actions/setup-java@v1
         with:
@@ -189,6 +191,8 @@ jobs:
           matrix.browser == 'chrome'
       - name: Start XVFB
         run: Xvfb :99 &
+      - name: Start Fluxbox
+        run: fluxbox -display :99 &
       - name: Run browser tests in ${{ matrix.browser }}
         uses: ./.github/actions/bazel-test
         with:
@@ -225,6 +229,8 @@ jobs:
             ${{ runner.os }}-bazel-build-
       - name: Setup bazelisk
         uses: ./.github/actions/setup-bazelisk
+      - name: Setup Fluxbox
+        run: sudo apt-get -y install fluxbox
       - name: Setup Java
         uses: actions/setup-java@v1
         with:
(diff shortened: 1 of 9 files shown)


titusfortner:
sounds like we found the culprit!

whimboo:
With the latest push the failure seem to be fixed. We are in process to decide which patches exactly need to be backed out. So once done and a Nightly build shipped I would like to push one more time to this PR for verification.

It's really interesting that this issue can only be seen in GitHub Actions.

whimboo:
@titusfortner I could actually need your help here for the failure in the [Ruby bindings which is related to the assumption that a window manager would set the `DESKTOP_SESSION`](https://github.com/SeleniumHQ/selenium/blob/trunk/rb/spec/integration/selenium/webdriver/spec_helper.rb#L56). Like [the following tests work now with a window manager running](https://github.com/SeleniumHQ/selenium/blob/b2598e7fab238b198971bebd49e5c3bfbaf275bb/rb/spec/integration/selenium/webdriver/window_spec.rb#L114-L139):

```
  1) Selenium::WebDriver::Window can maximize the current window FIXED
     Expected pending 'Test guarded; no reason given' to fail. No error was raised.
     # ./rb/spec/integration/selenium/webdriver/window_spec.rb:108

  2) Selenium::WebDriver::Window can make window full screen FIXED
     Expected pending 'Test guarded; no reason given' to fail. No error was raised.
     # ./rb/spec/integration/selenium/webdriver/window_spec.rb:119

  3) Selenium::WebDriver::Window can minimize the window FIXED
     Expected pending 'Test guarded; no reason given' to fail. No error was raised.
     # ./rb/spec/integration/selenium/webdriver/window_spec.rb:130
```

I wonder if we should simply set the `DESKTOP_SESSION=fluxbox` environment variable manually given that Fluxbox isn't doing that, or removing this particular behavior completely and always enforce that a window manager is needed to run the browser tests. Without one all the window manipulation tests aren't guaranteed to work.

Regarding the Java failures I would suggest we defer until Ruby is done. Thanks

whimboo:
> I wonder if we should simply set the `DESKTOP_SESSION=fluxbox` environment variable manually given that Fluxbox isn't doing that, or removing this particular behavior completely and always enforce that a window manager is needed to run the browser tests. Without one all the window manipulation tests aren't guaranteed to work.

If we need a way to detect a window manager a better way should be the following:

```
xprop -root -display :99
```

If it's not failing to connect to the display and no window manager is installed we should expect a couple of `_NET_*` lines to be returned.

titusfortner:
No, this is great. We only guarded them that way because I want to be able to run locally on linux and remotely on travis/github and have everything "pass" for the right reasons. If everything guarded for `window_manager` now works, we can completely remove that guard.

whimboo:
There is one Java test that now works with Firefox 105. I've created a separate PR: https://github.com/SeleniumHQ/selenium/pull/11046.

whimboo:
There are quite a few failures for Java on trunk as well. Those started between September 13th and 15th:
https://github.com/SeleniumHQ/selenium/actions/runs/3092869938/jobs/5004771191

The failures look similar to me so I don't think that this PR should be blocked on them. @titusfortner if you could have a look and review this PR I would appreciate.

PR #11029, opened 2022-09-14
[JS] Fix: Do not allow Select class to select disabled options

TamsilAmani:
### Description

Selecting disabled option will throw Error

### Motivation and Context

Fix for https://github.com/SeleniumHQ/selenium/issues/10812

### Types of changes

### Checklist

### DIFF
--- a/javascript/node/selenium-webdriver/lib/select.js
+++ b/javascript/node/selenium-webdriver/lib/select.js
@@ -37,6 +37,7 @@
 'use strict'
 
 const { By, escapeCss } = require('./by')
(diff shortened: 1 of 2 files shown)


titusfortner (changes requested):
We also want to check if the Select element is disabled as part of the `Select` constructor, now. Other than that, this looks good.

titusfortner (approved):

titusfortner:
Adding a check for disabled in constructor to match other languages would be nice, but is not essential. Thanks for this work.

PR #11049, opened 2022-09-21
[py] update outdated usage example

amwilkins:
Old comment referenced an outdated method, comment is updated to match new find_elements() format.


AutomatedTester (approved):
Thanks for your PR!

PR #11055, opened 2022-09-25
[iedriver] ignore zoom if running on Edge

titusfortner:
I can't figure out any way to change the zoom level for IE inside of Edge. So the only way to run in IE Mode is by setting ignore zoom level. At that point, Selenium should set it on behalf of the user. That said, I believe it isn't the default because it could cause problems with native mouse events not identifying the coordinates correctly? I don't know the underlying issue, so perhaps it needs to be fixed?

@bwalderman what do you think?

### DIFF
--- a/cpp/iedriver/BrowserFactory.cpp
+++ b/cpp/iedriver/BrowserFactory.cpp
@@ -120,7 +120,6 @@ std::string BrowserFactory::browser_command_line_switches(void) {
 void BrowserFactory::Initialize(BrowserFactorySettings settings) {
   LOG(TRACE) << "Entering BrowserFactory::Initialize";
   this->ignore_protected_mode_settings_ = settings.ignore_protected_mode_settings;
-  this->ignore_zoom_setting_ = settings.ignore_zoom_setting;
   this->browser_attach_timeout_ = settings.browser_attach_timeout;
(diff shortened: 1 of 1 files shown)


titusfortner:
Well, this would keep pointer actions from working, so I think we either merge this and state that pointer actions aren't supported, or figure out how to let users change zoom level...

bwalderman:
Do pointer actions still work correctly if the Edge and OS zoom levels are both 100%?

titusfortner:
Ooh, actually run the tests to check this... smart.

Yes, the coordinate tests still pass even though the driver complains about the mismatch

Any idea why `get_deviceXDPI` is 2x `get_logicalXDPI`? Or under which conditions we should ignore this?
https://github.com/SeleniumHQ/selenium/blob/trunk/cpp/iedriver/BrowserFactory.cpp#L817-L830

titusfortner:
@bwalderman so I tested this in Windows 11 & Windows 10. From my testing, regardless of whether Edge (or IE if it is installed) is set to a higher zoom level, IE Mode in Edge appears to always be set to 100% and always passes the coordinates tests. So I don't think there is an issue merging this. 

Do you have any concerns about it?

bwalderman:
@titusfortner, LGTM. Thank you for verifying.

PR #11058, opened 2022-09-26
[py] change "By" class with enum

plming:
### Description
* Refactoring "By" class with standarm Enum library
* Add "By" type hint where it needed.

### Motivation and Context
I checked history for "by.py" file. The file was first committed in 2010.
But according to [PEP 435](https://peps.python.org/pep-0435/), enum is introduced to Python after 2013.
So I think there's no reason to avoid standard enum library.
I expect this PR can enforce programmers not to use magic strings.

### Types of changes

### Checklist

- [ ] All new and existing tests passed. ;; Fortunately, there's no any piece of codes depending on magic strings.

### DIFF
--- a/py/selenium/webdriver/common/by.py
+++ b/py/selenium/webdriver/common/by.py
@@ -19,20 +19,20 @@
 The By implementation.
 """
 
-from enum import Enum, unique, auto
+from enum import Enum, unique
 
 
 @unique
-class By(Enum):
+class By(str, Enum):
     """
     Set of supported locator strategies.
     """
 
-    ID = auto()
-    XPATH = auto()
-    LINK_TEXT = auto()
-    PARTIAL_LINK_TEXT = auto()
-    NAME = auto()
-    TAG_NAME = auto()
-    CLASS_NAME = auto()
-    CSS_SELECTOR = auto()
+    ID = "id"
+    XPATH = "xpath"
+    LINK_TEXT = "link text"
+    PARTIAL_LINK_TEXT = "partial link text"
+    NAME = "name"
+    TAG_NAME = "tag name"
+    CLASS_NAME = "class name"
+    CSS_SELECTOR = "css selector"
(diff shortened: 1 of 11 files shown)


py/selenium/webdriver/common/by.py
```diff
@@ -19,17 +19,20 @@
 The By implementation.
 """
 
+from enum import Enum, unique, auto
 
-class By:
+
+@unique
+class By(Enum):
```
symonk: This is not such a trivial change here

py/selenium/webdriver/common/by.py
```diff
@@ -19,17 +19,20 @@
...
 
-    ID = "id"
-    XPATH = "xpath"
-    LINK_TEXT = "link text"
-    PARTIAL_LINK_TEXT = "partial link text"
-    NAME = "name"
-    TAG_NAME = "tag name"
-    CLASS_NAME = "class name"
-    CSS_SELECTOR = "css selector"
+    ID = auto()
```
symonk: won't this throw `using:` way off?

plming: > won't this throw `using:` way off?

I'm afraid but could you explain in detail? Do you mean that constants are used in somewhere?

symonk: yes, the old `str` value is passed into the `using: by`

plming:
I found there's a code using "json.dumps function" and that function needs constants for dumping..
So I tried to fix enum with str. How do you think about it?

Solonotix:
The failing test appears to be because when the exception for NoSuchElement is created, the effective call is to `repr()` on the resulting dictionary, thus the key looks like `{<By.ID: 'id'>: 'nonexistentid'}`. If the By class had a `__repr__` implementation that returned `self.value` instead, then it seems like it would pass the currently failing test on `test_no_such_element_is_raised_rather_than_index_error`

PR #11075, opened 2022-10-02
[java] small improvements by calling more accurate methods 

joerg1985:
### Description
Small performance improvements by calling more accurate methods to:
- avoid RegEx processing to replace a simple string, especially for huge scripts e.g. the atoms
- use the correct base64 decoder to avoid internal copying of the decoded array

### Motivation and Context
less overhead in java bindings

### Types of changes

### Checklist


java/src/org/openqa/selenium/OutputType.java
```diff
@@ -55,7 +55,7 @@ public String toString() {
   OutputType<byte[]> BYTES = new OutputType<byte[]>() {
     @Override
     public byte[] convertFromBase64Png(String base64Png) {
-      return Base64.getMimeDecoder().decode(base64Png);
+      return Base64.getDecoder().decode(base64Png);
```
pujagani, line 58: I am not familiar with the decoding details performance-wise. However, based on https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html#basic, they are to use different base64 encoding schemes. My concern is that it should not cause any unexpected errors.

joerg1985: The WebDriver specification tells the base64 encoded data should be rfc4648 compliant.
(see the "Base64 Encode" link in https://www.w3.org/TR/webdriver/#index)

The `Base64.getMimeDecoder()` is the decoder to rfc2045 encoded data.
The `Base64.getDecoder()` is the decoder to rfc4648 encoded data.

So i think the usage of `Base64.getDecoder()` should be the correct way to decode the data.

pujagani: That sounds right. Appreciate the thoroughness! Thank you so much.

pujagani (approved):
Thank you @joerg1985!

pujagani:
@joerg1985 I have reverted the changes made related to "avoid RegEx processing to replace a simple string, especially for huge scripts e.g. the atoms". It caused all the tests to break with an error message "Invalid escape sequence".

joerg1985:
Okay, i think there is something strage here.
The `test.replaceAll("\"", "\\\"");` will replace the `"` with a `"`, so i do not understand why there is a replace at all.

`System.out.println("\"".equals("\"".replaceAll("\"", "\\\"")));` will print `true`

joerg1985:
i will create seperate PRs for this two changes.
this proves you are totally right with seperate PRs

pujagani:
Thank you so much!

PR #11085, opened 2022-10-05
fix: Check that a port truly is free before using it

Artur-:
### Description

If for instance 0.0.0.0:12345 is bound then running chromedriver with '--port=12345 --allowed-ips=1.2.3.4' fails with address already in use. If ::1 port 1234 is in used, chromedriver with '--port=12345' fails

### Motivation and Context
Chromedriver fails to start in many cases with `[SEVERE]: bind() failed: Address already in use (48)`

### Types of changes

### Checklist


diemol (approved):
Thank you, @Artur-!

diemol:
@Artur- I am reverting the `isFree("::1", port)`, this is breaking in Linux and some macOS machines.

Artur-:
Is the problem that the machines do not have ipv6 support on or something else?

diemol:
I would say yes, however I didn't have enough time to check it

Artur-:
Ok, I assumed `} catch (Exception e) {` would have caught that

Artur-:
Ah, but that returns false. So then every port will be in use and nothing free..

diemol:
> Ah, but that returns false. So then every port will be in use and nothing free..

Exactly, that is the issue, so we do not know if ipv6 is available or the port being busy.

PR #11095, opened 2022-10-08
[py] Make `get_screenshot_as_file()` work with pathlib.Paths

NickCrews:
It currently expects a string (or something with .lower()), but accepting Paths seems quite reasonable.

I supposed if we wanted to be more strict and ensure the passed thing was path-like, we could instead do `str(Path(filename)).lower()`

### Types of changes

### Checklist


symonk:
Thanks @NickCrews, it's on my radar to convert everything to support pathlib

NickCrews:
Just signed the CLA.

It looks to me like the CI failure is unrelated?

symonk:
Thanks @NickCrews yep it is, some issues ongoing atm - Thanks for the contribution, much appreciated

Issue #11093, opened 2022-10-07
[🐛 Bug]: A selenium.common.exceptions.InvalidSessionIdException exception is raised when the print options are re-used from a session to another.


tvataire:
### What happened?

Current behavior :
Python language bindings for Selenium WebDriver raise a selenium.common.exceptions.InvalidSessionIdException exception when a PrintOptions instance is re-used by the print_page() method from a session to another.

Expected behavior : 
No exception should be raised.

### How can we reproduce the issue?

```shell
#!/bin/env python3
# -*- coding: utf8 -*-

import geckodriver_autoinstaller
from selenium.webdriver import Firefox
from selenium.webdriver.common.print_page_options import PrintOptions
from selenium.webdriver.firefox.options import Options

if __name__ == '__main__':
    geckodriver_autoinstaller.install()
    po = PrintOptions()
    po.margin_top = 0
    po.margin_left = 0
    po.margin_right = 0
    po.margin_bottom = 0
    do = Options()
    do.headless = True
    for idx in range(0, 2):
        driver = Firefox(options=do)
        try:
            print(driver.session_id)
            driver.print_page(po)
        finally:
            driver.close()
```

### Relevant log output

```shell
/mnt/PycharmProjects/selenium/venv/bin/python /mnt/PycharmProjects/selenium/print_issue.py 
ac74ec4c-8e63-47e9-a679-9b8c7153e0ff
228366ed-1da3-4315-8808-49efe081f160
Traceback (most recent call last):
  File "/mnt/PycharmProjects/selenium/print_issue.py", line 22, in <module>
    driver.print_page(po)
  File "/mnt/PycharmProjects/selenium/venv/lib/python3.9/site-packages/selenium/webdriver/remote/webdriver.py", line 619, in print_page
    return self.execute(Command.PRINT_PAGE, options)['value']
  File "/mnt/PycharmProjects/selenium/venv/lib/python3.9/site-packages/selenium/webdriver/remote/webdriver.py", line 434, in execute
    self.error_handler.check_response(response)
  File "/mnt/PycharmProjects/selenium/venv/lib/python3.9/site-packages/selenium/webdriver/remote/errorhandler.py", line 247, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.InvalidSessionIdException: Message: Got unexpected session id ac74ec4c-8e63-47e9-a679-9b8c7153e0ff


Process finished with exit code 1
```

### Operating System

Debian Bullseye

[...]

PR #11096, opened 2022-10-08
[py]: #11093 - The PrintOptions.to_dict() method should return a shal…

tvataire:
…low copy of the internal data of the class.

### Description
The PrintOptions.to_dict() method should return a shallow copy of the dictionary where are stored internal data of the class, instead of returns the dictionary itself.
This prevents the session id from being stored inside the PrintOptions instance when the Webdriver.print_page() method is called.

### Motivation and Context
If the session id is stored inside the PrintOptions instance, a selenium.common.exceptions.InvalidSessionIdException is raised when the PrintOptions instance is reused in a subsequent session.

### Types of changes

### Checklist

I tried to run unit tests on a 4 core cpu / 8 Go RAM machine with the command `bazel test //py:unit` but nothing happens. 
I launched the tests several hours ago and a 'bazel-5.3.0' process is running, but nothing is displayed on the standard output.
Is this the expected behavior ?
Is it normal that unit tests take so much time ?

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -184,8 +184,8 @@ crate_repositories()
 
 http_archive(
     name = "build_bazel_rules_nodejs",
-    sha256 = "b011d6206e4e76696eda8287618a2b6375ff862317847cdbe38f8d0cd206e9ce",
-    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.6.0/rules_nodejs-5.6.0.tar.gz"],
+    sha256 = "c29944ba9b0b430aadcaf3bf2570fece6fc5ebfb76df145c6cdad40d65c20811",
+    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.7.0/rules_nodejs-5.7.0.tar.gz"],
 )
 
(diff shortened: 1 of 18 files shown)


titusfortner (approved):

symonk:
while this does fix the problem, I'm not 100% sure on the full fix, should `_print_options` be storing a `session_id` internally as a result of executing initially here?

Is the better fix to make a copy of params in webdriver `.execute(...)` if it exists; else `{}` ? might stop us mutating random session state internally to multiple objects? (would need to test), as is this fix will prevent print options from having a bogus session id stored, my concern it - what other objects are suffering a similar fate, handling it in the remote webdriver execute I think makes it a non issue for pretty much everything (for now)

titusfortner:
@symonk I say we merge this to fix the specific problem now (it's a valid fix), then figure out a more comprehensive solution for the project. See discussion in #11093

symonk:
Fine by me, just missing cla here

titusfortner:
@tvataire looks like you committed with two separate emails that both need to be registered with CLA. I'm not sure if putting them on the same github account would also work.

tvataire:
Abandoned in favor of #11121.

PR #11099, opened 2022-10-09
New headless

titusfortner:
Most Selenium languages have implemented convenience methods for Headless mode in Chrome & Firefox.

Currently implemented:
* Ruby - Set
* JS - Set
* Java - Set & Unset
* Python - Set, Unset, Get
* .NET - N/A

The Chrome Headless mode we've set is not the production browser, and has many limitations (including not being able to load extensions).

Thanks to @mdmintz I learned that Chrome implemented a new Headless mode that is the full production browser:
https://chromium.googlesource.com/chromium/src/+/833543a511a970f44fb074b9f30c6b6268e0be71
It's been available since Chrome 94, but this change should be backwards compatible since it picks up anything it doesn't recognize (e.g., `--headless=xx`) as just `--headless`.

There may be slight performance benefit to using the old mode of headless (because it strips out functionality), but I believe the new mode more accurately represents what Selenium should encourage. If we are going to maintain a convenience feature (tbh, I would also be ok with deprecating/removing all of these), then I believe this new mode is the correct one.

I'm not adding Unsetters to Ruby or JS, or Getters to non-Python because I just don't see them as that important. I am implementing Headless in .NET (both Chrome & Firefox) to match Java implementation.

I've verified that the new headless mode works on a Linux box without a GUI / xvfb. I have not evaluated performance.

Which do you prefer:
1. Accept PR
2. Keep old implementation
3. Deprecate convenience method and let users decide

### DIFF
--- a/java/src/org/openqa/selenium/chromium/ChromiumOptions.java
+++ b/java/src/org/openqa/selenium/chromium/ChromiumOptions.java
@@ -186,6 +186,11 @@ public class ChromiumOptions<T extends ChromiumOptions<?>> extends AbstractDrive
     return (T) this;
   }
 
+  /**
+   * @deprecated Chrome provides two headless modes use `addArguments("--headless")`
+   * or preferably `addArguments("--headless=chrome")`
+   */
+  @Deprecated
   public T setHeadless(boolean headless) {
     args.remove("--headless");
     if (headless) {
--- a/java/src/org/openqa/selenium/firefox/FirefoxOptions.java
+++ b/java/src/org/openqa/selenium/firefox/FirefoxOptions.java
@@ -254,6 +254,10 @@ public class FirefoxOptions extends AbstractDriverOptions<FirefoxOptions> {
     return setFirefoxOption(Keys.LOG, logLevel.toJson());
   }
 
+  /**
+   * @deprecated Set this directly with `addArguments("--headless=chrome")`
+   */
+  @Deprecated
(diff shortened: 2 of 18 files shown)


harsha509:
Hi @titusfortner ,

Looks like there is an issue with https://www.w3.org/TR/webdriver/#dfn-print-page when set `headless=chrome`

The test runs in headless mode but still fails with error `WebDriverError: unknown error: PrintToPDF is only supported in headless mode`

Is this an issue with chromedriver ?

Thanks,
Sri Harsha

titusfortner:
@harsha509 yeah, looks like that functionality only works with "old" headless mode, so that's a confounding factor for our decision. I'm also not sure why JS is the only language testing it. :) Was going to look into it more.

AutomatedTester:
> @harsha509 yeah, looks like that functionality only works with "old" headless mode, so that's a confounding factor for our decision. I'm also not sure why JS is the only language testing it. :) Was going to look into it more.

It's because headless isn't being passed in the new way in the `conftest.py` file for python. I am assuming the others test setup is similar.

titusfortner:
It was decided that instead of changing this method that we would mark it as deprecated and advertise using the new mode as the recommended way to run in headless.

titusfortner:
Decided not to do this one, tracking with the new Issue

diemol:
Why? I thought we were going to eventually merge this?

titusfortner:
This PR is wrong. Created new issue to track actual requirement

PR #11100, opened 2022-10-10
[rust] Add --browser-version and --driver-version to selenium-manager

bonigarcia:
### Description
This PR includes two new configuration options for Selenium Manager (implemented as CLI arguments):

* `--browser-version` : To force a given major browser version (e.g., `105`,  `106 `, etc.).
*  `--driver-version` : To force a given driver version (e.g., `106.0.5249.61`, `0.31.0`, etc.).

### Motivation and Context
This PR is part of the development of Selenium Manager M1.

### Types of changes

### Checklist


titusfortner:
Is `--driver-version` implemented here, or is it a placeholder?

Is the idea to be able to do:
```
bazel run //rust:selenium-manager -- --browser chrome --browser-version 105 --driver-version 105.0.5195.52 
```

bonigarcia:
It is already implemented in this PR. But the idea is use one or another (`--driver-version` or `--browser-version` ), but not both, since it is redundant.

For example if you want to download an specific driver, you use `--driver-version` but not `--browser-version`, since the driver version implicitly refers to a browser version. In that case it would be:

``` 
bazel run //rust:selenium-manager -- --browser chrome --driver-version 105.0.5195.52 
``` 
If you want to force a (major) browser version, you use `--browser-version`, and Selenium Manager discover the proper driver version for that browser version. For instance:

``` 
bazel run //rust:selenium-manager -- --browser chrome --browser-version 105
```

titusfortner:
Ah, right, I didn't actually read the changes, sorry.

What errors is this giving when users send conflicting values?
I see the tests for passing values. Can we add tests for errors?

bonigarcia:
I have just committed a new patch with a test checking error conditions, and also, some logic to warn the user when both `--driver-version` and `--browser-version` are used (in that case, `--browser-version` is ignored).

titusfortner (approved):

rust/README.md
```diff
@@ -19,14 +19,24 @@ $ cargo run -- --help
 selenium-manager 1.0.0-M1
 Automated driver management for Selenium
 
-Usage: selenium-manager [OPTIONS] --browser <BROWSER>
+Usage: selenium-manager.exe [OPTIONS] --browser <BROWSER>
```
diemol: Can the `.exe` be printed only when using Windows?
Or maybe just leave it like `Usage: selenium-manager[.exe] [OPTIONS] --browser <BROWSER>`

bonigarcia: Yes, in fact the `.exe` only appears when running it in Windows. I have just committed a new patch removing it from the README.

diemol (approved):

PR #11101, opened 2022-10-10
Annotate JS

titusfortner:
Still a WIP while I figure out how to make sure everything is actually getting reported properly

Still working on Java & .NET

Essentially, when looking at logs (especially with service providers), it is hard to know what large JS blobs are doing. This provides an easy comment in the payload to see what it is doing.

See the difference in payload from
https://app.saucelabs.com/tests/b9722309476044b0b3e970beb702d178#3
to
https://app.saucelabs.com/tests/e1605700c8724a2abe63aa9949778feb#3

I'm not saying this is a good way to test things, but I created a way to turn on logs and verify the output in the logs for Python & Ruby. I might just verify manually for Java & .NET

### DIFF
--- a/dotnet/src/webdriver/JavaScriptEngine.cs
+++ b/dotnet/src/webdriver/JavaScriptEngine.cs
@@ -343,7 +343,7 @@ namespace OpenQA.Selenium
                 }
             }
 
-            return listenerScript;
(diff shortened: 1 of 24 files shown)


diemol:
What is the purpose of this PR? Didn't understand after reading the PR description.

titusfortner:
To be able to see what large block of JS from the payload. See the difference in the 2 Sauce links?

titusfortner:
Implemented with #11507

PR #11102, opened 2022-10-10
Add workflow to build Selenium Manager in CI (for Windows, Linux, and macOS)

bonigarcia:
### Description
This PR creates a GH Actions workflow to build Selenium Manager for the three main operating systems: Windows, Linux, and macOS. This workflow is manually triggered. The objective is to use the generated binaries (attached as artifacts at the end of the workflow execution) in the future releases of Selenium Manager.

### Motivation and Context
This PR is part of the development of Selenium Manager M1. Since it is non-trivial to cross-compile Rust (e.g., generate a Windows binary from a macOS machine), we can use the virtual environments of GH Actions (`windows-latest`, `ubuntu-latest`, and `macos-latest`) to that aim.

### Types of changes

### Checklist

### DIFF
--- a/.github/workflows/build-selenium-manager.yml
+++ b/.github/workflows/build-selenium-manager.yml
@@ -1,0 +1,63 @@
+name: build-selenium-manager
+
+on: workflow_dispatch
+
+jobs:
+  windows:
+    name: "[Windows] Build selenium-manager"
+    runs-on: windows-latest
+    steps:
+      - name: "Checkout project"
+        uses: actions/checkout@master
+      - name: "Build release"
+        run: |
+          cd rust
+          cargo build --release
+      - name: "Upload binary"
+        uses: actions/upload-artifact@v3
+        with:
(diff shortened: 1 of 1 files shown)


titusfortner:
The goal is to just generate the binaries? Where are they stored? I thought `upload-artifact` just allowed things to be used by other actions?

bonigarcia:
Yes, the goal is just to generate the Selenium Manager binaries, for Windows, Linux, and macOS. The binaries can be downloaded from GH, in the workflow execution page. You can see an example here:

https://github.com/bonigarcia/rust-examples/actions/runs/3220624172

We can manually get these binaries and include then in the Selenium releases, if finally we decide to do it that way.

titusfortner:
Though, downside is github doesn't support M1, yet.

titusfortner (approved):

diemol (approved):
Thank you, @bonigarcia!

PR #11107, opened 2022-10-11
[java] use the correct base64 decoder

joerg1985:
### Description
Use the correct base64 decoder (RFC4648) as defined by the WebDriver specification.

### Motivation and Context
Selenium should follow the WebDriver specification

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/OutputType.java
+++ b/java/src/org/openqa/selenium/OutputType.java
@@ -55,7 +55,7 @@ public interface OutputType<T> {
   OutputType<byte[]> BYTES = new OutputType<byte[]>() {
     @Override
     public byte[] convertFromBase64Png(String base64Png) {
-      return Base64.getMimeDecoder().decode(base64Png);
+      return Base64.getDecoder().decode(base64Png);
     }
 
     @Override


pujagani (approved):
Thank you @joerg1985 !

weikilla:
hi , why do this change? i am using appium java client, which is depend on this project. but currently, when i call screenshot method, will throw Illegal base64 character xxx, i found the root cause is this change.

joerg1985:
@weikilla the W3C WebDriver specification does specify the Base64 decoder to use, see: https://www.w3.org/TR/webdriver/#index

I think you should open a ticket at appium to get this fixed in the corresponding driver.
Alternatively they could overwrite the selenium implemenation, if they need a none W3C compliant implementation.

Just my two cents, the selenium team might have a different opinion on this 😄

mykola-mokhnach:
This change will break the backward compatibility between following java client versions and existing Appium drivers. We'd have to hack this implementation in Appium java client to workaround it.

Even if we update the thing for multiple drivers, existing versions of them stay unchanged and there would be a mess.

cc @diemol @titusfortner @valfirst @SrinivasanTarget

titusfortner:
@mykola-mokhnach can you please create an issue to make sure we track this? I don't understand the difference in implementations, so I'd vote for reverting it in the next release and figuring out a better path forward if we need to make the change.

mykola-mokhnach:
@titusfortner created https://github.com/SeleniumHQ/selenium/issues/11168

PR #11110, opened 2022-10-11
[java] removed the broken escape

joerg1985:
### Description
The escaping of javascript was broken, it replaced a quote with a identical quote to escape it.

I guess the original author tried to add a backslash to escape the quote, but the magic behind the `String.replaceAll` swallows the backslash. This will lead to a identical string before and after escaping the string.
(To get the assumed behavior it should have been `script.replaceAll("\"", "\\\\\"")` and not `script.replaceAll("\"", "\\\"")` )

The `String.replaceAll` call was dropped in this PR to get rid of this confusing code.

### Motivation and Context
Overhead processing the RegEx / get rid of confusing code.

### Types of changes

### Checklist


diemol (approved):
Thank you, @joerg1985!

Issue #11028, opened 2022-09-14
[🐛 Bug]: Firefox does not use the profile path when passed via options. Python


DollarAkshay:
### What happened?

It seems like firefox does not use the profile path when passed via options.

I have tried many different ways to set the profile.

#### Method 1 (Failing)
```python
options.set_preference('profile', profile_path)
```

#### Method 2 (Failing)
```python
profile = FirefoxProfile(profile_path)
options.profile = profile
```

#### Method 3 (Failing)
```python
profile = FirefoxProfile(profile_path)
driver = webdriver.Firefox(options=options, service=service, firefox_profile=profile)
```

All of these methods result in a temp profile being created somewhere else. This can be verified by looking at the `geckodriver.log` and seeing the profile path in the command.

The only way that does seem to work is passing it as an argument

#### Method 4 (Working)
```python
options.add_argument("-profile")
options.add_argument(profile_path)
```

Setting the profile via the arguments is the only way it seems to use the `profile_path`. This can be verified by looking at the `geckodriver.log` and seeing the profile path in the command and also looking at the `profile_path` folder which gets populated by a bunch of files and folders.

### How can we reproduce the issue?

```python
from selenium import webdriver
from selenium.webdriver.firefox.options import Options as FirefoxOptions
from selenium.webdriver.firefox.firefox_profile import FirefoxProfile
from selenium.webdriver.firefox.service import Service as FirefoxService


profile_path = "D:\\test"
driver_path = "./src/drivers/geckodriver_win64.exe"

service = FirefoxService(driver_path)
options = FirefoxOptions()
options.add_argument("--log-level=3")
options.set_preference('profile', profile_path)

driver = webdriver.Firefox(options=options, service=service)
```

### Relevant log output

```shell
1663163281111	geckodriver	INFO	Listening on 127.0.0.1:49637
1663163281617	mozrunner::runner	INFO	Running command: "C:\\Program Files\\Mozilla Firefox\\firefox.exe" "--marionette" "--log-level=3" "--remote-debugging-port" "49638" "--remote-allow-hosts" "localhost" "-no-remote" "-profile" "C:\\Users\\AKSHAY~1.ARA\\AppData\\Local\\Temp\\rust_mozprofileDsRaUY"
1663163282023	Marionette	INFO	Marionette enabled
[...]

PR #11118, opened 2022-10-13
Deprecate Firefox Profile Methods

titusfortner:
Based on #11028 I went looking at the Python implementation. The things that were Legacy Firefox related haven't been trimmed. Like I've been doing, I compared the implementations across the languages (except JS, because).

I added integration tests to figure out what does and doesn't work, and I've removed everything that no longer applies:
* Ruby still had a Chrome Profile that couldn't actually do anything so this deletes it
* Python & .NET were both setting legacy firefox specific features that were removed 
* Deleted all webdriver.json references and usage including all references to frozen and mutable preferences (frozen preferences were things that were needed for legacy firefox driver to work)
* All extensions methods removed - none of them work; you can add them to the directory, but Firefox won't actually load them
* Certificates methods removed - they didn't actually change things (plus please use the capability)
* Proxy methods removed (again, use the capability)
* I **added** a `startupUrl` method (Java & Ruby kind of had this already) to these

I actually deleted things instead of deprecating them, because none of them actually *do* anything right now. They are completely useless codes. Would we prefer to put empty methods back and mark them deprecated?

I know there's a lot here, but if you can review your primary language, I can make any relevant changes to all of the languages. Thanks

Options:
1. PR is good as-is
2. Deprecate things instead of just deleting them, even though the methods don't do anything
3. Deprecate everything, including things that work and have FirefoxProfile *only be used for zipping existing profiles; all other tweaks must use capabilities.


pujagani:
Thank you Titus for this undertaking and the work. I do not understand the nitty-gritty of Legacy Firefox. I trust you on it. Changes are mainly the removal of what is not working, which eventually seems fine. Though, I would prefer we first deprecate methods and then remove them to stick to Selenium's way of doing removals. Else it looks good to me and CI is also green for Java. 
All I am thinking is there could be a user, similar to the one who reported the issue, using the methods and it would cause breaking changes if we remove it.

titusfortner:
@pujagani the 2 main things a firefox profile can do:
1. Zip up an existing Firefox profile and use it
2. Build a basic Firefox profile from scratch and add things to it.

The user in the issue was trying to do the first item here, but couldn't because the Python code was doing more than just zipping the profile (which was required for legacy Firefox Profile.

There are certainly people using FirefoxProfiles for the second item, but really the only thing you can do at this point is to add preferences. Which, tbh, it would be much better if they would do via `prefs` in `Options` class than via `FirefoxProfile`. 

Another option we have is to *only accept existing profile paths and not let a user tweak anything via Firefox Profile... hmmm... I just updated the original post with options, I'm starting to lean toward option 3 now, in spite of all the work I just did here...

titusfortner:
it was decided that instead of deleting methods, we should mark everything as deprecated.
The only functionality that will be supported by FirefoxProfile going forward will be:
* needs Path in constructor
* copies contents from the directory to a temp directory
* zips temp directory on behalf of the Options class to send to the driver

Also, I need to make sure things work via grid as well because I had issues previously - https://github.com/mozilla/geckodriver/issues/1656

ollie-iterators:
This PR: https://github.com/SeleniumHQ/selenium/pull/11403, removed some deprecated Firefox methods.

titusfortner:
This code no longer applies. Tracking this in issue #11587

PR #11120, opened 2022-10-13
Implementation of Selenium Manager

titusfortner:
~~This can't be merge as-is because it turns out that the way we're running the Ruby tests with bazel isn't quite right somehow, so I have the bazel binaries in here multiple places...~~

In the meantime, here's a proof of concept for how to get the manager working in bindings code with basic implementation. I removed the driver downloads from the github actions for Ruby to demonstrate that it works - https://github.com/SeleniumHQ/selenium/actions/runs/3239912604

Update:
Added python implementation to this as well.

Update 2:
Binaries built with e6bd1f881f3bc18b147da146c1b1a00667ba9ae0


titusfortner:
Ok, I figured out the Ruby test issues & removed the extra binaries.
This PR includes the binaries with the `--driver` support that @bonigarcia just released.

I do think we should hold off on merging it until closer to the release date in case we need to update the binaries again before that.

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @classmethod
+    def driver_location(cls, browser: str) -> str:
+        """
+        Determines the path of the correct driver.
+        :Args:
+         - browser: which browser to get the driver path for.
+        :Returns: The driver path to use
+        """
+        path = os.path.dirname(__file__) + "/macos/selenium-manager"
```
symonk: minor: use `os.sep` here for `/`

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -0,0 +1,43 @@
...
+        args = [path, "--browser", browser]
```
symonk: minor: use a tuple here instead as its immutable

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -0,0 +1,43 @@
...
+logger = logging.getLogger(__name__)
+
+
+class SeleniumManager:
+    """
+    Wrapper for getting information from the Selenium Manager binaries.
+    """
+
+    @classmethod
+    def driver_location(cls, browser: str) -> str:
```
symonk: is a simple function suitable here rather than the class overhead? `@classmethod` seems abusive as it cares not about the `cls` itself / acting like some sort of factory, it's also only used once so making it just a function of the class normally and setting it as a `@property` would be sufficient too - if you think we might build on this in future and require internal state etc, otherwise a simple function is this module is perfect.

py/selenium/webdriver/common/service.py
```diff
@@ -183,11 +170,43 @@ def _terminate_process(self) -> None:
...
+            logger.error("Error terminating service process.", exc_info=True)
 
     def __del__(self) -> None:
         # `subprocess.Popen` doesn't send signal on `__del__`;
         # so we attempt to close the launched process when `__del__`
         # is triggered.
         with contextlib.suppress(Exception):
             self.stop()
+
+    def _start_process(self, path: str):
```
symonk: minor: return type here and docstring would be nice

py/selenium/webdriver/common/service.py
```diff
@@ -87,35 +88,21 @@ def start(self) -> None:
...
+            if "executable needs to be in PATH" in err.msg:
+                logger.info("driver not found in PATH, trying Selenium Manager")
+                browser = self.__class__.__module__.split(".")[-2]
+                try:
+                    path = SeleniumManager.driver_location(browser)
+                except OSError as new_err:
+                    if new_err.errno == errno.ENOENT:
+                        raise err
+                    else:
+                        raise new_err
```
symonk: minor: `raise new_error from None` here? do we care about the previous exc info?

py/selenium/webdriver/common/selenium_manager.py
titusfortner: I use class methods any time I have a function that doesn't require state. The instance method in the Service class has the state and is calling out for a static functional response. I'm expecting we'll also be implementing a `browser_location()` method as well at some point. If we implement a method to download the latest Selenium Manager for the given OS, though, that will probably set a state. So I'm interested in seeing what @bonigarcia thinks is a good way to do this in Java.

py/selenium/webdriver/common/service.py
titusfortner: Yeah, I want to throw the error that chromedriver is not found, not the error that selenium manager isn't found. For right now Selenium Manager isn't something the user can/should do anything about. The fix is for them to download chromedriver or use a third-party manager, not figure out selenium manager (yet). Maybe we can link to the docs for these, too.

py/selenium/webdriver/common/selenium_manager.py
titusfortner: Can you give me an example (not super familiar with tuples), the docs say: `The arguments used to launch the process. This may be a list or a string.`

py/selenium/webdriver/common/selenium_manager.py
titusfortner: aha, looks like I want to use `os.path.join` which implements `os.path.sep`.

py/selenium/webdriver/common/selenium_manager.py
symonk: we should make this a `@staticmethod` then rather than `@classmethod`, the latter should really return an instance of itself

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -0,0 +1,54 @@
...
+        :Args:
+         - browser: which browser to get the driver path for.
+        :Returns: The driver path to use
+        """
+        directory = sys.platform
+        if directory == "linux":
+            directory = "linux"
+        elif directory == "darwin":
+            directory = "macos"
+        elif directory == "win32" or directory == "cygwin":
```
symonk: minor: `directory in ("win32", "cygwin"):`

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -0,0 +1,54 @@
...
+            directory = "windows"
+
+        file = "selenium-manager.exe" if directory == "windows" else "selenium-manager"
+
+        path = os.path.join(os.path.dirname(__file__), directory, file)
```
symonk: can we use `pathlib` for this over `os`?

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -0,0 +1,54 @@
...
+        args = [path, "--browser", browser]
```
symonk: minor: make this a tuple instead of a list, should a user be able to add additional args for this tool?

py/BUILD.bazel
```diff
@@ -45,6 +45,24 @@ TEST_DEPS = [
     requirement("zipp"),
 ]
 
+copy_file(
```
symonk, line 48: what size are these? edit: couple of mb didnt know they are also in the diff :)

py/selenium/webdriver/common/service.py
```diff
@@ -89,33 +90,21 @@ def start(self) -> None:
...
-                raise
-        except Exception as e:
-            raise WebDriverException(
-                f"The executable {os.path.basename(self.path)} needs to be available in the path. {self.start_error_message}\n{str(e)}"
-            )
+            self._start_process(cmd)
+        except WebDriverException as err:
+            if "executable needs to be in PATH" in err.msg:
+                logger.info("driver not found in PATH, trying Selenium Manager")
+                browser = self.__class__.__module__.split(".")[-2]
```
symonk, line 95: what browsers does this support; is this the most reliable way to grab the browser?

py/selenium/webdriver/common/selenium_manager.py
titusfortner: Ah, yes that's what I want to do for Python. Thanks.

py/BUILD.bazel
titusfortner: yeah, they have to be committed directly for now, but only one place for all the languages. We agreed in last TLC meeting to do this. Eventually bazel will cross-compile binaries for all platforms.

py/selenium/webdriver/common/service.py
titusfortner: Chrome, Firefox & Edge (so everything but IE); but good call, should make sure it is one of those. 
We could alternately add a constant to each of the subclasses, but I figured the module was unlikely to change.
This is different from other languages because it gets the driver differently, but now that I'm thinking of it, maybe we don't want to use Selenium manager if someone is explicitly asking for a path that is invalid...

java/src/org/openqa/selenium/remote/service/DriverService.java
```diff
@@ -132,6 +188,18 @@ protected static File findExecutable(
       String exeDownload) {
     String defaultPath = new ExecutableFinder().find(exeName);
     String exePath = System.getProperty(exeProperty, defaultPath);
+
+    if (exePath == null) {
```
titusfortner, line 139: Yes, I need to ensure this is implemented in the other languages. At this point we shouldn't try to use Selenium Manager if the user specified a path.

titusfortner:
@bonigarcia is there a reason we shouldn't put SeleniumManager code in a separate class? It seems to have its own set of concerns as it were.

bonigarcia:
@titusfortner Actually there is no reason about that. But yes, it can be done better. I am on it.

bonigarcia:
@titusfortner I have just committed a new patch in which the Java code for the Selenium Manager integration is in a separate package/class. Thanks a lot for the suggestion!

titusfortner:
@bonigarcia I think that last commit needs to be a different PR, and we need to decide criteria for inclusion on that list.

java/src/org/openqa/selenium/manager/SeleniumManager.java
```diff
@@ -0,0 +1,128 @@
...
+    private static final String SELENIUM_MANAGER = "selenium-manager";
+    private static final String EXE = ".exe";
+    private static final String INFO = "INFO\t";
+
+    private static SeleniumManager manager;
+
+    private File binary;
+
+    private SeleniumManager() {
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
```
titusfortner, line 62: Why is this needed? Specifically why would SeleniumManager need this and the Drivers not?

java/src/org/openqa/selenium/manager/SeleniumManager.java
```diff
@@ -0,0 +1,128 @@
...
+                    Files.delete(binary.toPath());
+                } catch (IOException e) {
+                    LOG.warning(String.format("%s deleting temporal file: %s",
+                            e.getClass().getSimpleName(), e.getMessage()));
+                }
+            }
+        }));
+    }
+
+    public static SeleniumManager getInstance() {
```
titusfortner, line 74: Is there a reason to use a static factory method? We don't tend to do this in other parts of Selenium code, and (for now) I'm thinking of this class as "private api" / not something users will work with directly...

titusfortner: Actually, in Python & Ruby I'm implementing `getDriverPath()` as a static method directly, so they don't instantiate the object at all. Would that be limiting if we did it in Java?

java/src/org/openqa/selenium/manager/SeleniumManager.java
```diff
@@ -0,0 +1,128 @@
...
+                String folder = "linux";
+                String extension = "";
+                if (current.is(WINDOWS)) {
+                    extension = EXE;
+                    folder = "windows";
+                } else if (current.is(MAC)) {
+                    folder = "mac";
+                }
+                String binaryPath = String.format("%s/%s%s", folder, SELENIUM_MANAGER, extension);
+                try (InputStream inputStream = this.getClass().getResourceAsStream(binaryPath)) {
```
titusfortner, line 126: What is the advantage of copying the file before executing it?

java/src/org/openqa/selenium/manager/SeleniumManager.java
bonigarcia: This is an implementation of the singleton design pattern, which fits nicely in this case (see the following comment).

java/src/org/openqa/selenium/manager/SeleniumManager.java
bonigarcia: In Java, the binaries are distributed in a JAR file (i.e., they are compressed in these JAR). To access them, we need to access the JAR resources, getting an `InputStream` as a result. To execute this resource (a binary file), the `InputStream` should be serialized as an (executable) file. For this, it needs to be copied, and for that, a singleton pattern is the best alternative here (since we have a single instance in the JVM and we reuse the resulting binary for all the uses of the Selenium Manager singleton during all the execution, deleting the temporal file on shutdown).

java/src/org/openqa/selenium/manager/SeleniumManager.java
bonigarcia: See below.

Rakefile
```diff
@@ -114,6 +114,7 @@ JAVA_RELEASE_TARGETS = %w[
   //java/src/org/openqa/selenium/remote/http/jdk:jdk.publish
   //java/src/org/openqa/selenium/remote/http:http.publish
   //java/src/org/openqa/selenium/remote:remote.publish
+  //java/src/org/openqa/selenium/manager:manager.publish
```
diemol: Minor. Ideally, we have this in alphabetical order.

java/src/org/openqa/selenium/manager/SeleniumManager.java
diemol: Completely agree with @bonigarcia, makes sense. I would only ask to write a brief version of this comment as comment(s) in the code for future reference.

diemol (approved):
A couple of minor comments, the Java part looks good to me.

java/src/org/openqa/selenium/manager/SeleniumManager.java
bonigarcia: Done

Rakefile
bonigarcia: Done

java/src/org/openqa/selenium/manager/SeleniumManager.java
titusfortner: oh, this makes much more sense than what I thought this was doing.

titusfortner:
@bonigarcia I updated Ruby & Python to follow what Java did, and in the process I added some things to Java to follow what Ruby & Python are doing.

These are the changes - dfc154cfdf03c74be0527d0dcbcf2b0709d07ad3
@diemol  also let me know what you think.
Essentially, this has us throwing exceptions in SeleniumManager and then rescuing/logging in the DriverService.
Advantage is ensuring that this is drop-in/fall-back behavior, and I *think this is the behavior we want down the road?
Anyway, I've talked myself into and back out of it twice, so let me know what you think.

@symonk I split SeleniumManager class into 3 methods and added some logging and some exceptions. When you get a chance can you double check me? 

Now I just need to figure out .NET... how hard can it be?

java/src/org/openqa/selenium/manager/SeleniumManager.java
```diff
@@ -89,6 +91,10 @@ protected static String runCommand(String... command) {
             LOG.warning(String.format("%s running command %s: %s",
                     e.getClass().getSimpleName(), Arrays.toString(command), e.getMessage()));
         }
+        if (!output.startsWith(INFO)) {
+          throw new WebDriverException("Error running command: " + Arrays.toString(command));
```
bonigarcia, line 103: Maybe it deserves to include the value of `output` in the exception message, since that info can be relevant to discovered what happened.

java/src/org/openqa/selenium/manager/SeleniumManager.java
```diff
@@ -114,22 +120,23 @@ protected File getBinary() {
...
     public String getDriverPath(String driverName) {
+        if (!ImmutableList.of("geckodriver", "msedgedriver", "chromedriver").contains(driverName)) {
+            throw new WebDriverException("Unable to locate driver with name: " + driverName);
+        }
+
         String driverPath = null;
         File binaryFile = getBinary();
         if (binaryFile != null) {
             String output = runCommand(binaryFile.getAbsolutePath(),
                     "--driver", driverName.replaceAll(EXE, ""));
```
bonigarcia, line 155: I though the driver name in Windows contained the `.exe` at the of the end of `driverName`. But since this not like that (the previous condition you added is supposed to work also in Windows, right?), the `replaceAll(EXE, "")` in this line can be removed.

java/src/org/openqa/selenium/remote/service/DriverService.java
```diff
@@ -63,6 +64,7 @@
 public class DriverService implements Closeable {
 
   private static final String NAME = "Driver Service Executor";
+  private static final Logger LOG = Logger.getLogger(SeleniumManager.class.getName());
```
bonigarcia: Since this `LOG` is in the `DriverService` class, the class name in this line should be `DriverService` and not `SeleniumManager`.

java/src/org/openqa/selenium/manager/SeleniumManager.java
titusfortner: Hmm, at least in Ruby, it either has the driver info or it is an empty string. I'll see if that's a Ruby implementation detail and/or if Java is different.

java/src/org/openqa/selenium/manager/SeleniumManager.java
titusfortner: hmm, maybe I'm wrong; I'll double check this in all the languages.

java/src/org/openqa/selenium/remote/service/DriverService.java
titusfortner: copy paste strikes again. thanks.

java/src/org/openqa/selenium/manager/SeleniumManager.java
titusfortner: Yeah, if it fails it is either null because there was an exception, an empty string because console error doesn't return a value, or the version because somehow `--version` was passed in.

java/src/org/openqa/selenium/manager/SeleniumManager.java
titusfortner: Found a bug running this on Mac, so I'm going to pull this up to check Windows behavior.

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -0,0 +1,89 @@
...
+    @staticmethod
+    def driver_location(browser: str) -> str:
+        """
+        Determines the path of the correct driver.
+        :Args:
+         - browser: which browser to get the driver path for.
+        :Returns: The driver path to use
+        """
+        if browser not in ("chrome", "edge", "firefox"):
+            raise WebDriverException("Unable to locate driver associated with browser name: " + browser)
```
AutomatedTester:
```suggestion
            raise WebDriverException(f"Unable to locate driver associated with browser name: {browser}")
```

dotnet/src/webdriver/SeleniumManager.cs
```diff
@@ -0,0 +1,147 @@
...
+
+                    if (!Environment.OSVersion.Platform.ToString().StartsWith("Win"))
+                    {
+                        throw new WebDriverException("Selenium Manager only supports Windows in .NET at this time");
+                    }
+
+                    try
+                    {
+                        string name = "selenium-manager-" + folder;
+                        Stream fileStream = ResourceUtilities.GetResourceStream(name, name);
```
nvborisenko: dispose it

dotnet/src/webdriver/SeleniumManager.cs
```diff
@@ -0,0 +1,147 @@
...
+                        BinaryReader binReader = new BinaryReader(fileStream, Encoding.ASCII);
```
nvborisenko: dispose it

dotnet/src/webdriver/WebDriver.csproj
```diff
@@ -117,6 +117,10 @@
       <Visible>False</Visible>
       <LogicalName>webdriver_prefs.json</LogicalName>
     </EmbeddedResource>
+    <EmbeddedResource Include="$(ProjectDir)..\..\..\common\manager\windows\selenium-manager.exe">
```
nvborisenko, line 120: As I understand it includes exe file embedded into dll. Given that how many target frameworks you support, total size of nuget package will increase significantly:

image

There is easier way. Include exe file anewhere in nuget package, and then copy this file to anywhere while user compiles his project (using msbuild tagets).

dotnet/src/webdriver/SeleniumManager.cs
titusfortner: do you have an example of this? I'm not sure what you mean. :)

dotnet/src/webdriver/WebDriver.csproj
titusfortner: Yeah; the real question is whether we can do that with Bazel...

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -0,0 +1,89 @@
...
+import sys
+from pathlib import Path
+from typing import Tuple
+
+from selenium.common.exceptions import WebDriverException
+
+logger = logging.getLogger(__name__)
+
+
+class SeleniumManager:
```
symonk, line 29: whats the future of this in terms of internal state for this?  A class with just static methods is likely better just having functions in a module instead?

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -0,0 +1,89 @@
...
+            directory = "macos"
+        elif directory in ("win32", "cygwin"):
+            directory = "windows"
+
+        file = "selenium-manager.exe" if directory == "windows" else "selenium-manager"
+
+        path = Path(__file__).parent.joinpath(directory, file)
+
+        if not path.is_file():
+            raise WebDriverException("Unable to obtain Selenium Manager")
```
symonk, line 51: can we improve the error here around files not found, is `WebDriverException` appropriate here vs a `FileNotFoundError` or similar?

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -0,0 +1,89 @@
...
+
+    @staticmethod
+    def get_binary() -> Path:
+        """
+        Determines the path of the correct Selenium Manager binary.
+        :Returns: The Selenium Manager executable location
+        """
+        directory = sys.platform
+        if directory == "linux":
+            directory = "linux"
```
symonk: minor: this looks redundant

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        Determines the path of the correct driver.
+        :Args:
+         - browser: which browser to get the driver path for.
+        :Returns: The driver path to use
+        """
+        if browser not in ("chrome", "firefox"):
+            raise WebDriverException("Unable to locate driver associated with browser name: " + browser)
+
+        args = (str(SeleniumManager.get_binary()), "--browser", browser)
```
symonk, line 66: should these be static, arguably `self.get_binary()` here instead and instantiate the `SeleniumManager()` if we are planning to do more in future?

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -0,0 +1,89 @@
...
+        result = SeleniumManager.run(args)
+        command = result.split("\t")[-1].strip()
+        logger.info("using driver at: " + command)
```
symonk: I would keep these to debug, would probably also use `f""` (the interpolation/performance hit when logging is negligible, this is only ran once typically)

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -0,0 +1,89 @@
...
+        :Args:
+         - args: the components of the command being executed.
+        :Returns: The log string containing the driver location.
+        """
+        logger.debug("executing selenium manager with: " + " ".join(args))
+        result = subprocess.run(args, stdout=subprocess.PIPE).stdout.decode("utf-8")
+
+        pattern = re.compile("^INFO\t")
+        if not pattern.match(result):
+            raise WebDriverException("Unsuccessful command executed: " + " ".join(args))
```
symonk: lets just `f"Unsuccessful command executed: {args}"` here?

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        Executes the Selenium Manager Binary.
+        :Args:
+         - args: the components of the command being executed.
+        :Returns: The log string containing the driver location.
+        """
+        logger.debug("executing selenium manager with: " + " ".join(args))
+        result = subprocess.run(args, stdout=subprocess.PIPE).stdout.decode("utf-8")
+
+        pattern = re.compile("^INFO\t")
```
symonk: v.minor: could just use `re.match(...)` since the pattern is used only once

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -0,0 +1,89 @@
...
+
+    @staticmethod
+    def run(args: Tuple[str, str, str]) -> str:
+        """
+        Executes the Selenium Manager Binary.
+        :Args:
+         - args: the components of the command being executed.
+        :Returns: The log string containing the driver location.
+        """
+        logger.debug("executing selenium manager with: " + " ".join(args))
```
symonk: minor: consider `f""` and uppercase `Executing` for consistency with other logger calls

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -0,0 +1,89 @@
...
+        args = (str(SeleniumManager.get_binary()), "--browser", browser)
+        result = SeleniumManager.run(args)
+        command = result.split("\t")[-1].strip()
+        logger.info("using driver at: " + command)
+        return command
+
+    @staticmethod
+    def run(args: Tuple[str, str, str]) -> str:
+        """
+        Executes the Selenium Manager Binary.
```
symonk, line 75: unrelated to this diff but we have a mix of `sphinx` _and_ `google` styled docstrings throughout, I was looking at a tool that can autorewrite them to be the same must get back to looking at that.  More of a question - are these docstrings used at all in `selenium.dev` doc generation?

dotnet/src/webdriver/SeleniumManager.cs
nvborisenko:
```csharp
using (Stream fileStream = ResourceUtilities.GetResourceStream(name, name))
{
  // do something
} // here fileStream.Dispose() is implicitly invoked even if doing something throws an exception 
```

titusfortner: Ah, ok, the examples I was looking at made me think that I just needed `fileStream.Close()` Thanks

py/selenium/webdriver/common/selenium_manager.py
titusfortner: I think the future is going to include state. If all we support is "browser name" there's no reason for state.
But if we want to support a bunch of optional parameters (versions, locations, etc), makes more sense to have state.

py/selenium/webdriver/common/selenium_manager.py
titusfortner: The issue for this is we don't want to actually get the binary multiple times. We want to get it once per execution not on each class instantiation.

py/selenium/webdriver/common/selenium_manager.py
titusfortner: lol, I have no idea the difference between those, I just copy and paste from whatever is closest. 😂 
docs are coming from `tox -c py/tox.ini -e docs`

py/selenium/webdriver/common/selenium_manager.py
titusfortner: Maybe, but that seems weird since the code is what is trying to generate the location of the file. (as in "unable to create file" vs "unable to find file"). Everything is getting rescued right now. When/if these get exposed to the user we should re-think them.

py/selenium/webdriver/common/selenium_manager.py
titusfortner: I I think think it it is is just just fine fine..

dotnet/src/webdriver/SeleniumManager.cs
```diff
@@ -0,0 +1,146 @@
...
+                    {
+                        string name = "selenium-manager-" + folder;
+                        using (Stream fileStream = ResourceUtilities.GetResourceStream(name, name))
+                        {
+                            using (BinaryReader binReader = new BinaryReader(fileStream, Encoding.ASCII))
+                            {
+                                byte[] fileBytes = binReader.ReadBytes((int)fileStream.Length);
+                                string directoryName = string.Format(CultureInfo.InvariantCulture, "webdriver.{0}",
+                                    Guid.NewGuid().ToString("N"));
+                                var path = Path.Combine(Path.GetTempPath(), directoryName + "/" + folder);
```
nvborisenko, line 89: I don't think that writing to %temp% directory is good. Given that each time it creates new folder, disk space is used irrationally. Executing 100 tests produces 100 files which will be unpredictably cleaned up by OS. Supposably it's better to copy binary file to user's app space like `Environment.CurrentDirectory`, and make copying process thread-safe.

dotnet/src/webdriver/SeleniumManager.cs
```diff
@@ -0,0 +1,146 @@
...
+                                Directory.CreateDirectory(path);
+                                var filePath = Path.Combine(path, "selenium-manager" + extension);
+
+                                using (BinaryWriter binWriter = new BinaryWriter(File.Open(filePath, FileMode.Create)))
+                                {
+                                    binWriter.Flush();
```
nvborisenko, line 95: Redundant invocation, `Flush` is automatically used when disposing.

dotnet/src/webdriver/SeleniumManager.cs
```diff
@@ -0,0 +1,146 @@
...
+            }
+            catch (Exception ex)
+            {
+                throw new WebDriverException("Error starting process: " + process, ex);
+            }
+
+            String output = process.StandardOutput.ReadToEnd();
+
+            if (!output.StartsWith("INFO")) {
+                throw new WebDriverException("Invalid response from process: " + process);
```
nvborisenko, line 140: Include `output` into exception message to see what exactly the process wrote.

dotnet/src/webdriver/WebDriver.csproj
nvborisenko: I think yes. Bazel uses `nuspec` definition file to create nuget package. Here we can include any files like image is already included `<file src="../icon.png" target="images\" />`. The same we can use to include `*.exe` and `*.targets` files.

If we can resolve it, then it will resolve my other [concern](https://github.com/SeleniumHQ/selenium/pull/11120#discussion_r1007164747) completely.

titusfortner: So, I built the nupkg and it went from 5MB to 15MB, so not storing it for each framework, thankfully.

dotnet/src/webdriver/SeleniumManager.cs
titusfortner: well, it should be once per process, not once per test at least, since it is static method.
I think the next step is putting it in $Home/.cache/manager/ (similar to how we're storing the drivers themselves)
but then we'd need logic to make sure we have the latest version.

dotnet/src/webdriver/SeleniumManager.cs
titusfortner: but the flush is before the writing not after (with the disposal). tbh I'm not sure why you need it, I was going off of what I saw someone else doing...

dotnet/src/webdriver/SeleniumManager.cs
titusfortner: right now output is always an empty string if there's a problem. If/when we start outputting more things (like errors), we can update.

common/manager/BUILD.bazel
```diff
@@ -0,0 +1,30 @@
...
+        "macos/selenium-manager",
+        "windows/selenium-manager.exe",
+    ],
+    visibility = [
+        "//java/src/org/openqa/selenium/manager:__pkg__",
+        "//java/test/org/openqa/selenium/chrome:__pkg__",
+        "//java/test/org/openqa/selenium/edge:__pkg__",
+        "//java/test/org/openqa/selenium/firefox:__pkg__",
+        "//py:__pkg__",
+        "//rb:__pkg__",
```
harsha509, line 28: can we add this too for JS. 
` //javascript/node/selenium-webdriver:__pkg__`

the implementation goes in here https://github.com/SeleniumHQ/selenium/pull/11189

titusfortner: Yes, let's add in your PR

harsha509: Updated the PR!

5ryde:
#11214 #11201 #11213 #11215

PR #11123, opened 2022-10-13
[rust] Include --driver flag to Selenium Manager

bonigarcia:
### Description
This PR includes a new argument to Selenium Manager called  `--driver`, which is equivalent to `--browser` (i.e., it allows to select the driver to be managed), but using the driver name (i.e., chromedriver, geckodriver, or msedgedriver) instead of the browser name (chrome, firefox, edge).

### Motivation and Context
This is part of the development of Selenium Manager M1, to ease the integration with the Selenium bindings (e.g., in the `DriverService` class of the Java bindings, we have the driver name instead the browser name).

### Types of changes

### Checklist

### DIFF
--- a/rust/README.md
+++ b/rust/README.md
@@ -19,17 +19,19 @@ $ cargo run -- --help
 selenium-manager 1.0.0-M1
 Automated driver management for Selenium
 
-Usage: selenium-manager [OPTIONS] --browser <BROWSER>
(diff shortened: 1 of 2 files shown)


titusfortner (approved):
Yes! this would simplify the code I'm writing quite a bit
Depending on where the project goes we might want to do more with the errors, but at the point we're using it within the bindings, I think this is good.

PR #11124, opened 2022-10-14
Revert commits preventing Select List class from working with Disabled Elements

titusfortner:
Some unhappy people on #10812 
This should make them happier again.

Need to make sure the tests are all passing before I merge...

### DIFF
--- a/py/selenium/webdriver/support/select.py
+++ b/py/selenium/webdriver/support/select.py
@@ -34,9 +34,7 @@ class Select:
             Select(driver.find_element(By.TAG_NAME, "select")).select_by_index(2)
         """
         if webelement.tag_name.lower() != "select":
-            raise UnexpectedTagNameException(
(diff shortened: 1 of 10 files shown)


py/selenium/webdriver/support/select.py
```diff
@@ -35,10 +35,8 @@ def __init__(self, webelement) -> None:
         """
         if webelement.tag_name.lower() != "select":
             raise UnexpectedTagNameException(
-                "Select only works on <select> elements, not on <%s>" % webelement.tag_name
-            )
-        if not webelement.is_enabled():
-            raise NotImplementedError("Select element is disabled and may not be used.")
+                "Select only works on <select> elements, not on <%s>" %
```
symonk: minor: can we use `f"... not on {webelement.tag_name}"` strings here?

asolntsev (approved):

PR #11128, opened 2022-10-14
Don't kill chromedriver on CTRL+C

cristiklein:
### Description

When the user presses CTRL+C, Python gets KeyboardInterrupt -- which is expected -- while chromedriver gets SIGINT and closes -- which is not expected.

This patch makes sure chromedriver does not close on CTRL+C by placing it in a new process group.

### Motivation and Context

Python's CTRL+C handling is a bit surprising, in that it forwards a SIGINT to the child process. In the context of Selenium, this means that chromedriver is closed, which is a [common user complaint](https://stackoverflow.com/questions/27495376/catching-keyboardinterrupt-without-closing-selenium-webdriver-sessions-in-pyth).

#### Steps to reproduce

1. Open an interactive Python session: `python3`
2. At the Python prompt type the following:

    ```
    driver = webdriver.Chrome()
    ```

3. Press CTRL+C.

#### Expected result:

Python displays `KeyboardInterrupt`, Chrome window **stays open**.

#### Actual result

Python displays `KeyboardInterrupt`, Chrome window **closes**.

### Types of changes

### Checklist

### Notes to Review

* I'm unsure if this should be classified as a "bug" or "improved behavior". I couldn't find any specification on CTRL+C handling of Selenium.
* I'm unsure how to write a test for this, given that it would somehow require emulating CTRL+C key press at terminal.


AutomatedTester:
At the moment, the code is working as intended. Could you elaborate on why we would want to leave the browsers open when we have killed the overall test run?

cristiklein:
When I run Python interactively, e.g., from a Jupyter Notebook or Python console, I might do something like:

```
>>> driver = webdriver.Chrome()
>>> elems = driver.find_element(By.XPATH, '//a')
>>> for elem in elems: elems.click()
```

Then I notice that that wasn't smart, e.g., my XPath was incorrect, so I type CTRL+C to correct the XPath.

Ideally, I would want the browser window to say open, so I can get myself in a quick write-observe-learn loop.

Makes sense?

titusfortner:
Checked to see what Ruby does...
https://gist.github.com/titusfortner/a5d27c901c765af25f0081bcbb3bd394

Ruby doesn't kill the driver, but even when I wait for the page to finish loading and the driver to "reply" with the success command, the next command always fails. Back to normal after that.

`driver.quit` and ctrl+d both close driver + browser still.

cristiklein:
Happy New (Gregorian) Year everyone! :tada: 

Is there any way we can reach consensus here? For reminder, the core question is around the user pressing CTRL+C in an interactive terminal. Should the browser be killed (current behavior) or should the browser continue running (as proposed by the PR).

p0deje:
This PR would bring the code on par with Ruby bindings where we also start a new process group by default
https://github.com/SeleniumHQ/selenium/blob/36c788c46532f99c1d671ab524d97178a7195b2f/rb/lib/selenium/webdriver/common/child_process.rb#L54
On the other hand, I don't think any other bindings implement this feature. Given Ruby and Python can often be used via an interactive REPL session, it could be handy indeed to keep Chrome running.

diemol:
@cristiklein thank you for this PR, and we apologize that it took so long to get to a resolution.

In the 02.02.23 [TLC meeting](https://www.selenium.dev/meetings/2023/tlc-02-02/#proposalsdecisions), it was decided to not accept this change because the wide Python community would have their usual behavior changed.

cristiklein:
Thanks @diemol . I agree with the decision.

When I find the time, I'll work on a PR to keep the default behavior, but allow to optionally disable "kill browser on CTRL+C".

diemol:
@titusfortner @AutomatedTester @symonk isn't there a way to pass a parameter in Python and leave the browser open?

cristiklein:
If anyone bumps into the PR, I found a workaround:

```python
# Don't kill the browser on CTRL+C
import os
import subprocess

old_Popen = subprocess.Popen
def Popen_force_new_pgrp_for_chromedriver(*args, **kwargs):
    executable = args[0][0]
    if executable == 'chromedriver':
        kwargs['preexec_fn'] = os.setpgrp
        print(f'Forcing new pgrp for {executable}')
    return old_Popen(*args, **kwargs)
subprocess.Popen = Popen_force_new_pgrp_for_chromedriver
```

This code essentially intercepts the `Popen` call of selenium and forces the browser to start in a new process group. Hence, the browser does not get "the CTLR+C".

PR #11137, opened 2022-10-17
[py] WebElement parameter supported in some of expected_condition methods & test

Qnner:
> This is a new PR about #10802, I need submit a new PR because of my wrong operation

### Description
WebElement parameter supported in some of expected_conditions methods

### Motivation and Context
when you located an element already, you can just use it now;  

### Types of changes

### Checklist


titusfortner:
@symonk would it make sense to have element & locator be different kwargs for the same method rather than a single parameter that can be either? I think that would also simplify the code (if locator, element = find..., then use element, etc)

Qnner:
> Update expected_conditions.py

locator is used to locate an element, WebElement can locate an element too. So can we define WebElement is one of locator? locator is a tuple(contains `by` & `path`), or a specific WebElement object

PR #11143, opened 2022-10-18
[dotnet] Use new implementation of HttpClient in DriverService for communication

### LINKED ISSUE #11111: [dotnet] Would be better to migrate DriverService to use new HttpClient
### Feature and motivation

Selenium v4 started usage of `HttpClient` for all http requests.

But `DriverService` is skipped by mistake, it's still using old `HttpWebRequest`.

https://github.com/SeleniumHQ/selenium/blob/9e9156072ca1634a227d6b361a7d6d5d61ca917c/dotnet/src/webdriver/DriverService.cs#L223

### Usage example

-

nvborisenko:
Fixes #11111


dotnet/src/webdriver/DriverService.cs
```diff
@@ -219,20 +220,23 @@ protected virtual bool IsInitialized
...
-                    request.Timeout = 5000;
-                    HttpWebResponse response = request.GetResponse() as HttpWebResponse;
-
-                    // Checking the response from the 'status' end point. Note that we are simply checking
-                    // that the HTTP status returned is a 200 status, and that the resposne has the correct
-                    // Content-Type header. A more sophisticated check would parse the JSON response and
-                    // validate its values. At the moment we do not do this more sophisticated check.
-                    isInitialized = response.StatusCode == HttpStatusCode.OK && response.ContentType.StartsWith("application/json", StringComparison.OrdinalIgnoreCase);
-                    response.Close();
+                    using (var httpClient = new HttpClient())
```
nvborisenko, line 239: Probably would be better to keep this instance as private field in class.

dotnet/src/webdriver/DriverService.cs
```diff
@@ -363,23 +367,29 @@ private void Stop()
...
-                            // for a failed HTTP request due to a closed socket is particularly
-                            // expensive.
-                            HttpWebRequest request = HttpWebRequest.Create(shutdownUrl) as HttpWebRequest;
-                            request.KeepAlive = false;
-                            HttpWebResponse response = request.GetResponse() as HttpWebResponse;
-                            response.Close();
-                            this.driverServiceProcess.WaitForExit(3000);
-                        }
-                        catch (WebException)
+                        httpClient.DefaultRequestHeaders.ConnectionClose = true;
```
nvborisenko, line 403: What if we set Timeout (5 seconds?) explicitly?

dotnet/src/webdriver/DriverService.cs
```diff
@@ -363,23 +367,29 @@ private void Stop()
                 {
                     Uri shutdownUrl = new Uri(this.ServiceUrl, "/shutdown");
                     DateTime timeout = DateTime.Now.Add(this.TerminationTimeout);
-                    while (this.IsRunning && DateTime.Now < timeout)
+                    using (var httpClient = new HttpClient())
```
nvborisenko, line 401: Reuse the instance on class level?

titusfortner:
Do we need to update this as well?
https://github.com/SeleniumHQ/selenium/blob/9d55d474dfb25ca8bf26ab683c95f5d94c07c21f/dotnet/src/webdriver/Safari/SafariDriverService.cs#L121

nvborisenko:
> Do we need to update this as well?
> 
> https://github.com/SeleniumHQ/selenium/blob/9d55d474dfb25ca8bf26ab683c95f5d94c07c21f/dotnet/src/webdriver/Safari/SafariDriverService.cs#L121

Yes, will do

nvborisenko:
> Do we need to update this as well?
> 
> https://github.com/SeleniumHQ/selenium/blob/9d55d474dfb25ca8bf26ab683c95f5d94c07c21f/dotnet/src/webdriver/Safari/SafariDriverService.cs#L121

Done. Please pay attention on my inline comments, really waiting your feedback.

dotnet/src/webdriver/DriverService.cs
titusfortner: `this.TerminationTimeout` is set to 10 seconds and is configurable. Why explicit 5 seconds?

dotnet/src/webdriver/DriverService.cs
titusfortner: Because we are using 2 different HttpClient instances to start & stop the service? Not sure it saves much, but we could.

dotnet/src/webdriver/DriverService.cs
titusfortner: do you mean so it can be reused to stop the service?

nvborisenko: Let's keep it simple for now, we don't expect a lot of http requests to reuse client.

dotnet/src/webdriver/DriverService.cs
nvborisenko: Let's keep it simple for now.

dotnet/src/webdriver/DriverService.cs
nvborisenko: So, leaving the logic as it was before.

tqk2811:
You used HttpClient the wrong way  
True way is: make static `HttpClientHandler` and pass to every where `HttpClient` constructor  
```
static HttpClientHandler handler = new HttpClientHandler();

using HttpClient httpclient = new HttpClient(handler,false);//false for don't dispose handler
```

You can search everywhere why don't create new and dispose `HttpClientHandler`   
(Note: HttpClient default constructor [auto implement](https://github.com/microsoft/referencesource/blob/dae14279dd0672adead5de00ac8f117dcf74c184/System/net/System/Net/Http/HttpClient.cs#L132) HttpClientHandler)

nvborisenko:
This piece of code is not big pain.

> True way is: make static HttpClientHandler and pass to every where HttpClient constructor

Not only: https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient-guidelines

tqk2811:
In that guidelines 
> So if the rate of requests is high, the operating system limit of available ports might be exhausted. To avoid port exhaustion problems, reusing [HttpClient](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient) instance for as many HTTP requests as possible is recommended. This can be achieved by following [the recommended usage](https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient-guidelines#recommended-use).

I agree it's rare in this case. But we need to be careful for every scenario.

PR #11146, opened 2022-10-18
[java] Allow to customize target constructor args for EventFiringDecorator

mykola-mokhnach:
### Description

Potentially addresses https://github.com/appium/java-client/issues/1694#issuecomment-1281998055

I don't know how good the idea itself is, maybe it still needs some refactoring.
cc @valfirst 

### Motivation and Context
https://github.com/appium/java-client/issues/1694

### Types of changes


java/src/org/openqa/selenium/support/decorators/WebDriverDecorator.java
```diff
@@ -336,9 +346,22 @@ protected final <Z> Z createProxy(final Decorated<Z> decorated, Class<Z> clazz)
...
+    }
+
     try {
-      return proxy.newInstance();
-    } catch (ReflectiveOperationException e) {
+      Constructor<?> dstConstructor = proxy
+        .getConstructor(Arrays.stream(targetWebDriverConstructorArgs)
+          .map(Object::getClass)
+          .toArray(Class<?>[]::new)
+        );
```
valfirst: That might not work in 100% cases, for example: constructor arg type is `List`, but argument can be `ArrayList`. And for such cases fully automated search for applicable constructor is not a trivial task. So I would recommend to overload `WebDriverDecorator` constructor and add one more field:

```java
  private final Class<T> targetWebDriverClass;
  private final Class<?>[] targetWebDriverConstructorArgsTypes;
  private final Object[] targetWebDriverConstructorArgs;

  private Decorated<T> decorated;

...

  public WebDriverDecorator(Class<T> targetClass) {
    this(targetClass, new Object[0]);
  }

  public WebDriverDecorator(Class<T> targetClass, Object[] targetConstructorArgs) {
    this.targetWebDriverClass = targetClass;
    this.targetWebDriverConstructorArgs = targetConstructorArgs;
    this.targetWebDriverConstructorArgsTypes = Arrays.stream(targetWebDriverConstructorArgs)
          .map(Object::getClass)
          .toArray(Class<?>[]::new);
  }

  public WebDriverDecorator(Class<T> targetClass, Class<?>[] targetWebDriverConstructorArgsTypes, Object[] targetConstructorArgs) {
    this.targetWebDriverClass = targetClass;
    this.targetWebDriverConstructorArgsTypes = targetWebDriverConstructorArgsTypes;
    this.targetWebDriverConstructorArgs = targetConstructorArgs;
  ...
```

java/src/org/openqa/selenium/support/events/EventFiringDecorator.java
```diff
@@ -162,6 +162,7 @@
   private final List<WebDriverListener> listeners;
 
   public EventFiringDecorator(WebDriverListener... listeners) {
+    super();
```
valfirst:
```suggestion
```
not needed, it's always invoked implicitly

java/src/org/openqa/selenium/support/decorators/WebDriverDecorator.java
```diff
@@ -336,9 +356,22 @@ protected final <Z> Z createProxy(final Decorated<Z> decorated, Class<Z> clazz)
       .getLoaded()
       .asSubclass(clazz);
 
+    if (clazz.isInterface()
+        || this.targetWebDriverConstructorArgs.length == 0
+        || this.targetWebDriverConstructorArgTypes.length == 0
+        ) {
```
valfirst:
```suggestion
    if (clazz.isInterface()) {
```
the code below should work perfectly for constructors without args, so no need to add these checks

valfirst (commented):
LGTM,
but also I believe there is some rule such bug fixes should be covered by unit tests.

mykola-mokhnach:
> but also I believe there is some rule such bug fixes should be covered by unit tests.

I would be happy to do that if I knew how to run these tests locally. Is there some tutorial available on how to configure the IDE to be able to build/run the java selenium project and unit tests?

pujagani:
@mykola-mokhnach Please have a look at https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md#step-5-test

pujagani:
Thank you @mykola-mokhnach. The changes look good to me. Please add relevant tests and then I will merge the changes. Thank you!

valfirst (approved):

mykola-mokhnach:
Unfortunately I have miserably failed to run neither the project nor tests locally. It looks like bazel is trying to download half of the internet and even after that it's not happy by something else. I was not able to also configure IntelliJ or VsCode with Bazel plugin installed to load the project and to build it properly, so I could have proper source autocomplete. Maybe I am still missing some docs...

Eventually, most of the code has been added in a completely "blind" manner with the hope CI could do its job.

pujagani:
I appreciate your effort. Sorry to hear that setting it up was such a task. Usually, the bazel plugin makes it easy to import bazel project into Iintellij by importing the workspace file present in the folder https://github.com/SeleniumHQ/selenium. 

Thank you for adding the test! If there is an issue, I will take a look and help fix it.

diemol:
I was sitting with @mykola-mokhnach a moment ago, and now he is able to run tests. Thanks for checking, @pujagani

mykola-mokhnach:
I was able to resolve my issue with IDEA setup thanks to @diemol 
the trick was to use a separate bazel import flow rather than standard project import thing. Still not quite sure what was missing in VSCode, but I'm happy with what I have anyway. 

I was able to do more through local testing and it looks like several assumptions I did about ByteBuddy are wrong and the whole idea must be refined. I need to close the PR for further investigation.

mykola-mokhnach:
Just to provide more context on this issue:

One of the main limitations with ByteBuddy is that it cannot patch an existing instance. It is necessary to create a proxy class first and then use its instance to invoke any driver calls. For example:

```

IOSDriver driver = new IOSDriver(caps);
...
WebDriverListener webDriverListener = new WebDriverListener();
IOSDriver proxy = (IOSDriver) new EventFiringDecorator(IOSDriver.class, webDriverListener).decorate(driver);
```

would create a `second` IOSDriver instance using the current implementation. 

This approach works for interfaces though, because by calling `decorate` we basically create a vanilla Object instance and then tell it to implement all methods (they all are public) of supported interfaces (e.g. have a dummy there). When any method of this proxy is being invoked we redirect the call to the original instance and trigger callbacks when necessary.

I need to verify if the above approach would be possible for classes as well

PR #11162, opened 2022-10-24
Updated IllegalArgumentException to NullPointerException for Null Values in Require Class

vinoth959:

Update IllegalArgumentException to NullPointerException for Null Values.
### Description
Updated IllegalArgumentException to NullPointerException for Null Values in Require.java file.
Changed MustBeSet to MustBePositive for one condition based on the action in the Require class.

### Motivation and Context
Misleading Exception and cause issues in debugging.

### Types of changes

### Checklist

Updated IllegalArgumentException to NullPointerException for Null Values.
Changed MustBeSet to MustBePositive for one condition based on the action in Require Class.

### DIFF
--- a/.github/workflows/ci-ruby.yml
+++ b/.github/workflows/ci-ruby.yml
@@ -83,6 +83,7 @@ jobs:
       - uses: ./.github/actions/bazel
         with:
           command: test --test_output=all //rb:${{ matrix.target }}
+          attempts: 3
         env:
           DISPLAY: :99
 
@@ -121,6 +122,7 @@ jobs:
       - uses: ./.github/actions/bazel
         with:
           command: test --test_output=all //rb:${{ matrix.target }}
+          attempts: 3
         env:
           DISPLAY: :99
 
--- a/java/CHANGELOG
+++ b/java/CHANGELOG
@@ -1,3 +1,8 @@
+v4.5.3
+======
+* Supported CDP versions: 85, 104, 105, 106
+* Removing check "::1" in PortProber (Fixes #11159)
+
 v4.5.2
 ======
 * Supported CDP versions: 85, 104, 105, 106
--- a/java/src/org/openqa/selenium/internal/Require.java
+++ b/java/src/org/openqa/selenium/internal/Require.java
@@ -56,14 +56,14 @@ public final class Require {
 
   public static <T> T nonNull(String argName, T arg) {
     if (arg == null) {
-      throw new NullPointerException(String.format(MUST_BE_SET, argName));
+      throw new IllegalArgumentException(String.format(MUST_BE_SET, argName));
     }
     return arg;
   }
 
   public static <T> T nonNull(String argName, T arg, String message, Object... args) {
     if (arg == null) {
-      throw new NullPointerException(
+      throw new IllegalArgumentException(
         String.join(" ", argName, String.format(message, args)));
     }
     return arg;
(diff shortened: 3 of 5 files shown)


diemol (commented):
Can you please share the motivation for this pull request? The goal of `Require` is to check if the arguments passed are valid or not, and checking for nulls is a subset of those validations.

vinoth959:
Hi,

I was getting error as Input should be Set for the nonNull method and could not identify the root cause of the issue and took a while for figuring out that the value is null. Method name says nonNull and it throws illegalargumentexception. Atleast the exception should be null or the message should say the value is null. That is the reason for this pull request

Thanks,
Vinoth

diemol:
> Hi,
> 
> I was getting error as Input should be Set for the nonNull method and could not identify the root cause of the issue and took a while for figuring out that the value is null. Method name says nonNull and it throws illegalargumentexception. Atleast the exception should be null or the message should say the value is null. That is the reason for this pull request
> 
> Thanks, Vinoth

Can you share the code that caused you to get that exception?

vinoth959:
i was doing creating new webdriverwait but did not realize the driver passed in the class was null. When i run the code i got illegalargumentexception with Input must be set error and did not have clue on what's happening. On further investigation i found the mistake but that error was confusing. This is just a cosmetic change i'm proposing..

I was passing the wait as below.
 waitforElementtoLoad(element, new WebDriverWait(driver, Duration.ofSeconds(3)));

titusfortner:
looks like everything that doesn't include `MUST_BE_SET` can swallow the fact that the value was null:

```
  public static Duration nonNegative(String argName, Duration arg) {
    if (arg == null) {
      throw new IllegalArgumentException(String.format( , argName));
    }
...
```

It would make sense for some of these to be more clear what the problem is, but I don't think everything should be changed to `NullPointerException`

vinoth959:
Got it.. just my thoughts.. if not needed also fine.. I know the problem now.

vinoth959:
I strongly beleive this method should be fixed atleast.

        throw new IllegalArgumentException(String.format(MUST_BE_POSITIVE, argName)); should be        throw new IllegalArgumentException(String.format(MUST_BE_SET, message));


public static double positive(String argName, Double number, String message) {
    if (number <= 0) {
      if (message == null) {
        throw new IllegalArgumentException(String.format(MUST_BE_POSITIVE, argName));
      } else {
        throw new IllegalArgumentException(message);
      }
    }
    return number;
  }

titusfortner:
I was agreeing with you, then I realized I was using a bad example....

The fix that needs to happen is that the logic for
```
public static double positive(String argName, Double number, String message)
```
needs to match the logic for
```
public static int positive(String argName, Integer number, String message)
```

i.e., the `if (number == null)` conditional got skipped for some reason.

vinoth959:
> I was agreeing with you, then I realized I was using a bad example....
> 
> The fix that needs to happen is that the logic for
> 
> ```
> public static double positive(String argName, Double number, String message)
> ```
> 
> needs to match the logic for
> 
> ```
> public static int positive(String argName, Integer number, String message)
> ```
> 
> i.e., the `if (number == null)` conditional got skipped for some reason.

I actually understand the logic now.. i was confused.

vinoth959:
Updated the above 

> > I was agreeing with you, then I realized I was using a bad example....
> > The fix that needs to happen is that the logic for
> > ```
> > public static double positive(String argName, Double number, String message)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > needs to match the logic for
> > ```
> > public static int positive(String argName, Integer number, String message)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > i.e., the `if (number == null)` conditional got skipped for some reason.
> 
> I actually understand the logic now.. i was confused.

Updated the above logic in the pull request...

What can i do to make this more readable for null values?? We can leave it as it is?

titusfortner:
I think we just have to know that `"myValue must be set"` indicates that it was `null`

diemol (approved):
Thank you, @vinoth959!

PR #11171, opened 2022-10-26
[rust] Update Selenium Manager roadmap in README

bonigarcia:
### Description
This PR updated the roadmap (i.e., milestone planed) for the initial development of Selenium Manager.

### Motivation and Context
Part of the Selenium Manager development.

### Types of changes

### Checklist


rust/README.md
```diff
@@ -70,15 +70,11 @@ Alternatively, you can replace `cargo run` with `bazel run //rust:selenium-manag
...
+| M1: Driver management                   | <ul><li>Beta version of the Selenium Manager.</li> <li>Focused on driver management for Chrome, Firefox, and Edge.</li> <li>Selenium Manager compiled for Windows, Linux, and macOS (in GH Actions).</li> <li>Available in Selenium binding languages (Java, JavaScript, Python, Ruby, and C#).</li> <li>Used as a fallback for language bindings, when the driver is not found.</li> <li>Selenium Manager binaries bundled within the binding languages.</li></ul> |
+| M2: Driver management for IEDriver      | <ul><li>Include driver support for IExplorer.</li> <li>Consider msedgedriver in IE Mode.</li></ul>                                                                                                                                                                                                                                                                                                                                                                  |
```
titusfortner: msdgedriver in IE Mode uses the IEDriver classes and driver, so there's nothing that Manager needs to do.

rust/README.md
```diff
@@ -70,15 +70,11 @@ Alternatively, you can replace `cargo run` with `bazel run //rust:selenium-manag
...
+|-----------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| M1: Driver management                   | <ul><li>Beta version of the Selenium Manager.</li> <li>Focused on driver management for Chrome, Firefox, and Edge.</li> <li>Selenium Manager compiled for Windows, Linux, and macOS (in GH Actions).</li> <li>Available in Selenium binding languages (Java, JavaScript, Python, Ruby, and C#).</li> <li>Used as a fallback for language bindings, when the driver is not found.</li> <li>Selenium Manager binaries bundled within the binding languages.</li></ul> |
```
titusfortner, line 75: msedgedriver isn't working right now; I've set it to ignore in the bindings, I say we push it off to M2, but depends on how hard is the fix.

rust/README.md
```diff
@@ -70,15 +70,11 @@ Alternatively, you can replace `cargo run` with `bazel run //rust:selenium-manag
...
+| M2: Driver management for IEDriver      | <ul><li>Include driver support for IExplorer.</li> <li>Consider msedgedriver in IE Mode.</li></ul>                                                                                                                                                                                                                                                                                                                                                                  |
+| M3: Rich configuration                  | <ul><li>Proxy support in Selenium Manager.</li> <li>Extra configuration capabilities from binding languages to Selenium Manager (e.g., force to use a given driver version, etc.).</li></ul>                                                                                                                                                                                                                                                                        |
```
titusfortner, line 77: Add ability to specify browser location (Se Mgr will check that location for version and grab the right driver for it).
Do we support beta/dev/canary right now?

titusfortner (changes requested):
Also need to be able to support M1 architecture; cross-compilation?
How are we determining 32 vs 64 right now? Should that be configurable?
Eventually need to figure out if we can solve the snap issue for people.

rust/README.md
bonigarcia: Updated

rust/README.md
bonigarcia: This is the first notice I have about that. What is the problem? In any case, I think this PR is not the best place to report a issue like that. If I know the problem, I can try to solve it for M1.

rust/README.md
bonigarcia: Not now, but it should be easy to implement. That can be part of the extra configuration milestone.

bonigarcia:
Regarding ARM (or other architectures), I think that is not priority. I tried the Selenium Manager macOS binary (generated in GH Actions) in a macOS M1 and it works. Regarding 32-64 arch, I am not sure. It would be interesting to test the current binaries (Windows, macOS, Linux) generated in GH Actions in different platforms.

titusfortner:
> I think that is not priority

I'm not saying it is a priority, but it should be on a list of possible features. Maybe not part of "roadmap" per se.

titusfortner (approved):

PR #11189, opened 2022-10-28
[JS] add support for selenium manager

harsha509:
### Description
* Implements selenium-manager in js bindings
* if browser driver is not in path `selenium-manager` does the work for chrome and firefox

depends on :
https://github.com/SeleniumHQ/selenium/pull/11120 -> common/manager/bazel.build for visibility of files during build

### Types of changes

### Checklist


javascript/node/selenium-webdriver/chrome.js
```diff
@@ -150,19 +151,23 @@ class ServiceBuilder extends chromium.ServiceBuilder {
...
-          `http://chromedriver.storage.googleapis.com/index.html and ensure ` +
-          `it can be found on your PATH.`
+      console.log(
+        ` The ChromeDriver could not be found on the current PATH, trying Selenium Manager`
       )
+      exe = driverLocation(Browser.CHROME)
+    }
+
+    if (!exe) {
+      throw Error(`Unable to obtain driver using Selenium Manager`)
```
titusfortner: There's nothing a user can do to fix a Selenium Manager problem, but they can fix their driver problem by adding chromdriver to PATH.. We want to log Selenium Manager problems in Selenium manager code, but the driver classes should throw the same error they've always thrown.

javascript/node/selenium-webdriver/chrome.js
```diff
@@ -150,19 +151,23 @@ class ServiceBuilder extends chromium.ServiceBuilder {
...
     if (!exe) {
-      throw Error(
-        `The ChromeDriver could not be found on the current PATH. Please ` +
-          `download the latest version of the ChromeDriver from ` +
-          `http://chromedriver.storage.googleapis.com/index.html and ensure ` +
-          `it can be found on your PATH.`
+      console.log(
+        ` The ChromeDriver could not be found on the current PATH, trying Selenium Manager`
       )
+      exe = driverLocation(Browser.CHROME)
```
titusfortner: we want to throw/catch this error so that people don't even know we tried using the Selenium Manager.

javascript/node/selenium-webdriver/common/seleniumManager.js
```diff
@@ -0,0 +1,87 @@
...
+  let result
+
+  try {
+    result = execSync(args.join(' ')).toString()
+  } catch (e) {
+    throw new Error(
+      `Error executing command with ${args} : Error ${e.stderr.toString()}`
+    )
+  }
+
```
titusfortner, line 82: We also want to error if result doesn't start with `INFO\t`

javascript/node/selenium-webdriver/firefox.js
```diff
@@ -459,15 +460,16 @@ function locateSynchronously() {
...
-      'The ' +
-        GECKO_DRIVER_EXE +
-        ' executable could not be found on the current ' +
-        'PATH. Please download the latest version from ' +
-        'https://github.com/mozilla/geckodriver/releases/ ' +
-        'and ensure it can be found on your PATH.'
+    console.log(
+      `The ${GECKO_DRIVER_EXE} executable could not be found on the current PATH, trying Selenium Manager`
     )
+    exe = driverLocation(Browser.FIREFOX)
```
titusfortner: try/catch this

javascript/node/selenium-webdriver/firefox.js
```diff
@@ -459,15 +460,16 @@ function locateSynchronously() {
...
+  }
+
+  if (!exe) {
+    throw Error(`Unable to obtain driver using Selenium Manager`)
```
titusfortner: use the old error still

titusfortner (changes requested):
Essentially we don't want users to know that there's a Selenium Manager trying to do things for them, because if it works they don't care, and if it doesn't work they can't do anything about it.

javascript/node/selenium-webdriver/chrome.js
```diff
@@ -150,20 +151,31 @@ class ServiceBuilder extends chromium.ServiceBuilder {
...
-          `http://chromedriver.storage.googleapis.com/index.html and ensure ` +
-          `it can be found on your PATH.`
+      console.log(
+        ` The ChromeDriver could not be found on the current PATH, trying Selenium Manager`
       )
+
+      try {
+        exe = driverLocation(Browser.CHROME)
+      } catch (e) {
+        throw Error(
```
titusfortner: it's possible that this does not throw an error, but also that `!exe` is still `true`. Catch & console log error, then throw if `!exe`.

javascript/node/selenium-webdriver/firefox.js
```diff
@@ -459,16 +460,23 @@ function locateSynchronously() {
...
-        'https://github.com/mozilla/geckodriver/releases/ ' +
-        'and ensure it can be found on your PATH.'
+    console.log(
+      `The ${GECKO_DRIVER_EXE} executable could not be found on the current PATH, trying Selenium Manager`
     )
+
+    try {
+      exe = driverLocation(Browser.FIREFOX)
+    } catch (e) {
+      throw Error(
```
titusfortner: same issue as with chrome

PR #11191, opened 2022-10-28
[dotnet] Reworked bundling of selenium manager into nuget package

nvborisenko:
- Selenium Manager binaries (win/linux/macos) are included into package once rather that into each dll (total package size before: 14.5MB, after: 18.5MB)
- All SM binaries are copied to users project output during his build process (for all platforms). Improvement: we can determine here on which platform user builds a project and copy only specific binary; but user might build and execute project on different platforms.
- Reworked wrapper for SM process, properly output details if something went wrong
- If driver service couldn't locate webdriver then details are also included from SM
- Tested the final package for Chrome/Firefox/Edge on Win and Linux.


titusfortner (changes requested):
We don't have access to Environment.OSVersion.Platform
It requires System.Runtime.InteropServices which we will need to add as a dependency

Pull that out and let's just get windows working for now.

nvborisenko:
> It requires System.Runtime.InteropServices which we will need to add as a dependency

It requires `System.Runtime` dependency.

image

I have tested the package on Win and Linux, works good.

titusfortner:
As an update, I cherry-picked a couple of these commits, but still want to revisit the remaining suggestions here. 

The linux/mac code I want to wait on getting the new package bundled to get actual OS

Some of this should be implemented across all the bindings.

diemol:
What needs to be done in this PR now? Seems we need to clean it up to decide how we move forward?

titusfortner:
@nvborisenko I'm closing this one. 

Specifically outstanding:
* I want to wait until we have actual error outputs before adding stderr stuff, which needs to be done on all the bindings. 
* The check for "INFO" should be moved out of the `GetBinary()` method in all of the bindings. 
* I'm not sure I understand the difference in how the binary is getting closed with your proposal. Does it fix #11226?

nvborisenko:
Thanks for your attention, good to see it was useful via cherry-picking some commits to trunk.

> Does it fix https://github.com/SeleniumHQ/selenium/issues/11226

No, it doesn't. A fix is easy.

PR #11195, opened 2022-10-29
Adding instructions for a shallow clone in the CONTRIBUTING page

joshmgrant:
### Description
For faster development, it's recommended to use a shallow clone (`--depth 1`) instead of a full clone. I have updated the CONTRIBUTING doc to mention this.

### Motivation and Context
Every little bit helps. If a new or existing contributor can get the Selenium code base in a clone operation more quickly, they may see a better chance to make a change and open a PR. This change helps these new contributors get up and started quickly.

### Types of changes

### Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -68,9 +68,10 @@ Fork the project [on Github](https://github.com/seleniumhq/selenium)
 and check out your copy locally.
 
 ```shell
-% git clone git@github.com:username/selenium.git
+% git clone --depth 1 git@github.com:username/selenium.git # shallow clone recommended
 % cd selenium
 % git remote add upstream git://github.com/seleniumhq/selenium.git
+% git remote update
 ```
 
 #### Dependencies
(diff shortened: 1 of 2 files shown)


AutomatedTester:
I’ve not gone through the readme but there are a lot of changes in this PR for a readme change. Can you check and rebase accordingly

David

joshmgrant:
@AutomatedTester sorry about that, PR is now in better shape

CONTRIBUTING.md
```diff
@@ -68,9 +68,10 @@ Fork the project [on Github](https://github.com/seleniumhq/selenium)
 and check out your copy locally.
 
 ```shell
-% git clone git@github.com:username/selenium.git
+% git clone --depth 1 git@github.com:username/selenium.git # shallow clone recommended
```
titusfortner, line 71: I think the default should be to clone the whole thing, but it does make sense to add a note after this code snippet that if you do not need commit history, using `--depth 1` will be much faster.

diemol: @joshmgrant I think what Titus says makes sense, feel free to make the changes and then we can merge.

joshmgrant:
After some consideration, I don't think this PR is meaningful. Thanks!

PR #11197, opened 2022-10-30
[grid][ui] Add search field for running sessions

### LINKED ISSUE #10632: [🚀 Feature]: Having Search box added in Selenium 4 Grid GUI
### Feature and motivation

**Scenario:** When we start our selenium automation test script execution on Selenium4 Grid GUI (http://localhost:4444/ui/index.html), we can see the sessions gets located to required TestCases. We do have sorting option or can see 15 VNC sessions at time but it's difficult when the TestCases number being executed are more (~80 I am currently executing).

**Issue:** We cannot directly search with session ID or name (if we set it using setCapability("se:name","")) in VNC Sessions UI.

**Request:** Request to add new feature of adding search box in Sessions Grid 4 VNC GUI to making the search easy.

**Original Grid-4** 

Original Grid4

**Modified Grid-4** 

Modified Grid 4

### Usage example

I am going to use this search box feature for Sessions Grid 4 GUI to search particular sessionId or the name of VNC for the session I provided.

mhnaeem:
### Description
Adding a search field for running sessions in the grid UI so we can filter out sessions. Using the `SessionData` and it's properties to filter items.

Changes Include:

1) Modified the EnhancedTableToolbar to accept children elements as props, also fixed its styling using Box elements
2) Added a new React component called RunningSessionsSearchBar, this component contains three things: search input field, info/help icon, help dialog
3) Added search filters during the render step for the rows, results are filter right after they are sorted for rendering - the ordering should still work
4) Updated the tests for RunningSessions to check for some of the use cases using search field

### Motivation and Context
Fixes #10632

### Types of changes

### Checklist

### DIFF
--- a/javascript/grid-ui/src/components/RunningSessions/RunningSessions.tsx
+++ b/javascript/grid-ui/src/components/RunningSessions/RunningSessions.tsx
@@ -295,9 +295,17 @@ function RunningSessions (props) {
                   {stableSort(rows, getComparator(order, orderBy))
                     .filter((session) => {
                       if (searchFilter === '') {
-                        // don't filter anything
+                        // don't filter anything on empty search field
                         return true
                       }
+
+                      if (!searchFilter.includes('=')) {
+                        // filter on the entire session if users don't use `=` symbol
+                        return JSON.stringify(session)
+                          .toLowerCase()
+                          .includes(searchFilter.toLowerCase())
+                      }
+
                       const [filterField, filterItem] = searchFilter.split('=')
                       if (filterField.startsWith('capabilities,')) {
                         const capabilityID = filterField.split(',')[1]
--- a/javascript/grid-ui/src/components/RunningSessions/RunningSessions.tsx
+++ b/javascript/grid-ui/src/components/RunningSessions/RunningSessions.tsx
@@ -46,6 +46,7 @@ import EnhancedTableToolbar from '../EnhancedTableToolbar'
 import prettyMilliseconds from 'pretty-ms'
 import BrowserLogo from '../common/BrowserLogo'
 import OsLogo from '../common/OsLogo'
+import RunningSessionsSearchBar from './RunningSessionsSearchBar'
 import { Size } from '../../models/size'
 import LiveView from '../LiveView/LiveView'
 import SessionData, { createSessionData } from '../../models/session-data'
@@ -174,6 +175,8 @@ function RunningSessions (props) {
   const [page, setPage] = useState(0)
   const [dense, setDense] = useState(false)
   const [rowsPerPage, setRowsPerPage] = useState(5)
+  const [searchFilter, setSearchFilter] = useState('')
+  const [searchBarHelpOpen, setSearchBarHelpOpen] = useState(false)
 
   const handleRequestSort = (event: React.MouseEvent<unknown>,
     property: keyof SessionData) => {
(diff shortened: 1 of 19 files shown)


javascript/grid-ui/src/components/RunningSessions/RunningSessions.tsx
```diff
@@ -283,6 +293,18 @@ function RunningSessions (props) {
...
+                        // don't filter anything
+                        return true
+                      }
+                      const [filterField, filterItem] = searchFilter.split('=')
+                      if (filterField.startsWith('capabilities,')) {
+                        const capabilityID = filterField.split(',')[1]
+                        return (JSON.parse(session.capabilities as string) as object)[capabilityID] === filterItem
+                      }
+                      return session[filterField] === filterItem
+                    })
```
diemol:
```suggestion
                    .filter((session) => {
                      if (searchFilter === '') {
                        // don't filter anything
                        return true
                      }
                      return (session.capabilities as string).toLowerCase()
                               .indexOf(searchFilter.toLowerCase()) > -1
                    })
```

diemol: I found it misleading when typing in the search a valid keyword, such as `browserName`, then all rows get filtered and nothing is shown. That is why I find the filter logic a bit too complex. It is not really necessary to introduce a query language for a quick search for now. 

Hence my suggestion in the code change, a simple filter based on the entered string and all available capabilities. After that we can wait for feedback from the users.

javascript/grid-ui/src/components/RunningSessions/RunningSessionsSearchBar.tsx
```diff
@@ -0,0 +1,157 @@
...
+                <TableCell>Capability</TableCell>
+                <TableCell><pre>capabilities,platformName=windows</pre></TableCell>
+              </TableRow>
+            </TableBody>
+          </Table>
+        </TableContainer>
+        <p>The basic syntax for searching is <strong><i>key=value</i></strong> or <strong><i>capabilities,key=value</i></strong>.
+          All properties under <strong><i>SessionData</i></strong> are available for search and most capabilities are also searchable
+        </p>
+      </DialogContent>
```
diemol, line 146: Perhaps we can change these instructions to ones where we say that the user needs to enter a string and it will be used to filter the session by comparing it to all available fields.

diemol (changes requested):
Thank you for this, @mhnaeem!

I left a commend on the filter logic. Please have a look and then we can merge this so it is part of 4.6.0.

javascript/grid-ui/src/components/RunningSessions/RunningSessions.tsx
mhnaeem: @diemol Thank you for reviewing it. I understand that this might be a little complex however I was hoping we could keep some of the logic. My use case of the grid has 100s of tests running at the same time and they all have lots of capabilties such as `se:name` and other custom capabilities `se:some_id` that might not work great with a simpler search.

 Here's an example:
```json
{
  // capability 
  {
    "se:name": "Don't test Chrome - Sample Test"
    "se:some_id": "no-chrome"
  },
  "browserName": "edge"
}
```
If I were to search for `chrome` in the search it would show me this example from above which doesn't really help me but having the abiltiy to do `browserName=chrome` I can search specific capabilties.


As an alternative, maybe we can do something like this which will allow users to just type random stuff and filter as long as it doesn't have an `=` symbol and for more complex use cases we can do the query-ish syntax.
```js
.filter((session) => {
  if (searchFilter === '') {
    // don't filter anything
    return true
  }

  // NEW CODE
  if (!searchFilter.includes('=')) {
    return (session.capabilities as string).toLowerCase()
                               .indexOf(searchFilter.toLowerCase()) > -1
  }
  // NEW CODE - END

  const [filterField, filterItem] = searchFilter.split('=')
  if (filterField.startsWith('capabilities,')) {
    const capabilityID = filterField.split(',')[1]
    return (JSON.parse(session.capabilities as string) as object)[capabilityID] === filterItem
  }
  return session[filterField] === filterItem
})
```


Please let me know what you think and if this alternative is not desired I can remove my code.

diemol: OK, that sounds like a good alternative. Looking forward to the changes!

diemol (approved):

diemol (approved):
Thank you, @mhnaeem!

PR #11198, opened 2022-10-30
[grid] & [java] jdk http client - avoid chunking without buffering to memory

joerg1985:
### Description
Some drivers do not implement transfer-encoding chunked e.g. [Chrome](https://bugs.chromium.org/p/chromedriver/issues/detail?id=1850) 
The jdk http client implementation did solve this by buffering the request body into a byte array.
The new implementation will only buffer into memory if the content-length is not know.

@pujagani I think this was the reason the `BodyPublishers.ofInputStream` did not work, while `BodyPublishers.ofByteArray` did work with Edge and Chrome drivers.

### Motivation and Context
The implementation of `ProtocolHandshake#createSession` does use a `FileBackedOutputStream` to keep the memory
consumption low for big requests. The jdk http client did not handle this and did buffer the complete request to memory.
This change will help not to buffer the new session request.

When the jdk http client is used for the grid, the size of the payload is usually known.
This change will also help not to buffer these request while sending them to the driver.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/remote/ProtocolHandshake.java
+++ b/java/src/org/openqa/selenium/remote/ProtocolHandshake.java
@@ -97,8 +97,7 @@ public class ProtocolHandshake {
          Writer writer = new OutputStreamWriter(counter, UTF_8)) {
       payload.writeTo(writer);
 
-      try (InputStream rawIn = os.asByteSource().openBufferedStream();
-           BufferedInputStream contentStream = new BufferedInputStream(rawIn)) {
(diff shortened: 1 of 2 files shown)


pujagani (approved):
Thank you @joerg1985 for looking into this. It does solve the chunking problem seen earlier with InputStream. I ran multiple scenarios locally with the changes and confirmed the chunking problem is solved. I have also tested that changes in ProtocolHandshake do not effect either of the HTTP clients.  Appreciate the efficient solution

PR #11201, opened 2022-10-31
feat: Add generic proxy implementation

mykola-mokhnach:
### Description
This is my proposal to solve https://github.com/appium/java-client/issues/1694#issuecomment-1281998055

### Motivation and Context
There is no easy way in Java to patch/instrument a class if it is already loaded. Java Agent mechanism allows that, but it requires users to change the way their jar is being executed and is pretty complicated.
This implementation allows to create a generic proxy for any class and add one or more listeners to public method invocations. The current decorators implementation stays untouched though. I was not able to find an acceptable way to modify it in a non-breaking manner to support the current proxying strategy.

### Types of changes

cc @valfirst @pujagani

### DIFF
--- a/java/src/org/openqa/selenium/support/BUILD.bazel
+++ b/java/src/org/openqa/selenium/support/BUILD.bazel
@@ -18,6 +18,7 @@ java_export(
         ":page-factory",
         "//java/src/org/openqa/selenium/support/events",
         "//java/src/org/openqa/selenium/support/locators",
+        "//java/src/org/openqa/selenium/support/proxy",
         "//java/src/org/openqa/selenium/support/ui:clock",
         "//java/src/org/openqa/selenium/support/ui:components",
         "//java/src/org/openqa/selenium/support/ui:elements",
--- a/java/src/org/openqa/selenium/support/proxy/BUILD.bazel
+++ b/java/src/org/openqa/selenium/support/proxy/BUILD.bazel
@@ -1,0 +1,16 @@
+load("@rules_jvm_external//:defs.bzl", "artifact")
+load("//java:defs.bzl", "java_library")
+
+java_library(
+    name = "proxy",
+    srcs = glob(["*.java"]),
+    visibility = [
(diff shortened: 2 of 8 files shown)


titusfortner:
I'm concerned about adding things to the Support package, especially things that are specific to one language. Would it be an option for this to get bundled into a separate jar in a separate repo?

mykola-mokhnach:
> I'm concerned about adding things to the Support package, especially things that are specific to one language. Would it be an option for this to get bundled into a separate jar in a separate repo?

No problem. I'll more the pr to Appium java client

titusfortner:
Well, mine is just one opinion, I was curious if it might make sense to do it independently. :)

@diemol what do you think about this in Selenium vs Appium (vs separate package)?

pujagani:
@mykola-mokhnach Thank you for your contribution. Can you help me understand how will this proxy be used to solve https://github.com/appium/java-client/issues/1694#issuecomment-1281998055? How can it be used by Selenium? Just trying to understand where the code should sit. Do you have any follow-up PRs in mind that touches on the current decorators implementation?

@titusfortner We have webdriver decorator in the support package for Java. That functionality is specific to Java. So on that point, I don't see an issue. However, I want to first understand how this proxy will be used since it seems to address a broader Java concern and for that separating out is a good idea.

mykola-mokhnach:
> @mykola-mokhnach Thank you for your contribution. Can you help me understand how will this proxy be used to solve [appium/java-client#1694 (comment)](https://github.com/appium/java-client/issues/1694#issuecomment-1281998055)? How can it be used by Selenium? Just trying to understand where the code should sit. Do you have any follow-up PRs in mind that touches on the current decorators implementation?

@pujagani Please check https://github.com/appium/java-client/pull/1790
I've also added some more documentation and examples there. I am not going to further develop any changes to the existing decorators implementation as it is a dead-end. It will not work for class types without breaking changes in the internal and external logic.

pujagani:
Thank you for sharing that PR. It has helped answer my questions.

PR #11217, opened 2022-11-03
Switch to our own hermetic Ruby rules for Bazel

p0deje:
### Description

Replaces https://github.com/bazelruby/rules_ruby with https://github.com/p0deje/rules_ruby_simple which provide similar functionality for building Ruby gems, running tests and other commands. The main difference is that the new rules use fully hermetic Ruby distribution which works on every major OS including Windows.

The CI is now 3-4 minutes slower than it was before, I will need to check what happens with caching to make sure it's as fast as it can be.

### Checklist


.github/workflows/ci-ruby.yml
```diff
@@ -218,6 +194,8 @@ jobs:
           key: ${{ runner.os }}-bazel-ruby-unit-test-${{ matrix.ruby }}-${{ hashFiles('**/BUILD.bazel') }}
           restore-keys: |
             ${{ runner.os }}-bazel-ruby-unit-test-${{ matrix.ruby }}-
+      - name: Patch WORKSPACE to select Ruby ${{ matrix.ruby }}
+        run: sed -i -E "s/(rb_download\(version = \")(.+)(\"\))/\1${{ matrix.ruby }}\3/" WORKSPACE
```
titusfortner, line 204: We can't set this by ENV or something "easier?"

titusfortner (approved):

.github/workflows/ci-ruby.yml
p0deje: I don't think we can, but I would like to make it "proper" later and add support for multiple toolchains. I think this is hacky but works for now.

p0deje: @shs96c Do you know how I can define multiple `rb_download()` for different versions of Ruby in `WORKSPACE` and then select them on CI?

diemol (commented):
Could we host the rules in the SeleniumHQ org? I am not saying that I do not trust you, @p0deje :) It is just that I have seen maintainers in the past add a tool they maintain, then they switch to another focus and the tool remains unmaintained in their GitHub user. Most recent example is the tool to generate JS docs.

.github/workflows/ci-ruby.yml
shs96c: The simplest thing would to `load` the ruby version from another `bzl` file, and just modify that. eg `//rb:ruby_version.bzl` could declare a single value (`rb_version`), and then we just rewrite that file in the build matrix

rb/BUILD.bazel
```diff
@@ -139,15 +165,11 @@ ruby_library(
         ":manager-macos",
         ":manager-windows",
         ":license",
-    ],
-    deps = [
```
shs96c, line 143: Do we really do away with all these dependencies?

shs96c (approved):
LGTM

titusfortner:
@diemol we still depend on Jari's child process gem that was in his personal org then transferred to someone else (and has been technically unmaintained for several years now). 

Precedent is separate org. Plus I don't think Selenium project overall should be signing up to maintain this in perpetuity.

rb/BUILD.bazel
p0deje: Those were actually not necessary in the first place.

.github/workflows/ci-ruby.yml
p0deje: I suppose it should be possible by using [`config_setting`](https://bazel.build/docs/configurable-attributes#select-and-dependencies), but that's for the next commits.

p0deje:
@diemol I can definitely host it in the @SeleniumHQ organization if you prefer, though a longer-term plan would be to work with folks on https://github.com/bazelruby/rules_ruby to upstream the changes and/or make it official Ruby rules for Bazel. I tend to agree with @titusfortner that Selenium doesn't have to sign up for maintaining the rules considering. I'd rather do it myself until they are mature enough to be used in other Ruby projects.

diemol:
> @diemol I can definitely host it in the @SeleniumHQ organization if you prefer, though a longer-term plan would be to work with folks on https://github.com/bazelruby/rules_ruby to upstream the changes and/or make it official Ruby rules for Bazel. I tend to agree with @titusfortner that Selenium doesn't have to sign up for maintaining the rules considering. I'd rather do it myself until they are mature enough to be used in other Ruby projects.

@p0deje sounds good to me

PR #11255, opened 2022-11-11
[grid] fixed reading the Connection header

joerg1985:
### Description
The value of the `Connection` header should be handled as case-insensitive value, see [Reading the Client's Opening Handshake of rfc6455](https://www.rfc-editor.org/rfc/rfc6455#section-4.2.1).

### Motivation and Context
A client handshake was differently handled if `Upgrade` was used instead of the expected `upgrade`.

### Types of changes

### Checklist


diemol (approved):
Thank you, @joerg1985!

PR #11257, opened 2022-11-12
[py] [run python] Upgrade GitHub Actions

cclauss:

### Description

[py]
[run python]

Upgrades to match:
* https://github.com/actions/cache/releases
* https://github.com/actions/checkout/releases
* https://github.com/actions/setup-java/releases
* https://github.com/actions/setup-python/releases

See Node.js 12 actions deprecation warnings at https://github.com/SeleniumHQ/selenium/actions/runs/3449658980

### Motivation and Context

Improved performance and compatibility.  See release notes above.
### Types of changes

- [x] Infrastructure upgrade

### Checklist

### DIFF
--- a/.github/workflows/ci-python.yml
+++ b/.github/workflows/ci-python.yml
@@ -26,9 +26,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout source tree
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Set up Python 3.7
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: 3.7.10
       - name: Install dependencies
@@ -46,9 +46,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout source tree
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Set up Python 3.7
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: 3.7.10
       - name: Install dependencies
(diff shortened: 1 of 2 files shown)


cclauss:
Python-related GitHub Actions do not run when Python files are modified.

diemol:
@cclauss we have a script to determine which build targets to run, as we use bazel. Our script has room for improvement, it determines the build targets based on the changed code files. Since the GitHub Actions files are not Python code, the workflow does not run.

To force the execution, make a temporal change to one of the `.py` files, and after it runs, you can revert it.

cclauss:
Yes.  Both the Python files that were modified in  #11258 were modified in this PR and neither one of them woke up bazel.

cclauss:
GitHub Actions [paths](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) already has [filter patterns like `**.py`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#patterns-to-match-file-paths) so perhaps that would be a safer, simpler approach.

diemol:
> GitHub Actions [paths](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) already has [filter patterns like `**.py`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#patterns-to-match-file-paths) so perhaps that would be a safer, simpler approach.

@cclauss yes, but this would execute all build and test targets for the code, which is less than ideal. With the script we have, we build and test only what gets affected by the code change. It is not perfect as it ignores changes like the one you are submitting.

PR #11267, opened 2022-11-16
Stop using system-wide Python for Bazel

p0deje:
Instead, we rely on the hermetic Python toolchain installed via https://github.com/bazelbuild/rules_python. Most of the py_* targets already work fine with it.

This commit also replaces `--workspace_status_command` so that it does not require system-wide Python installation and instead uses a simple Bash script on Linux/macOS and a PowerShell script on Windows.

The wheel can now be built on Windows without the need to install Python (I only need to comment out Rules bits in WORKSPACE because they don't work on Windows ARM).

Screenshot 2022-11-15 at 15 58 15


diemol (approved):
Thank you for this, @p0deje!

LGTM.

AutomatedTester (approved):

AutomatedTester:
Looks good to me, merge when you're happy

PR #11276, opened 2022-11-17
[java][bidi] Add LogInspector

pujagani:

### Description

Add LogInspector contains the w3c Bidi spec related APIs to listen to log events.

### Motivation and Context

Refer to https://docs.google.com/document/d/1dCd8Y2PYaR5mOGSmNTwllEHNmFqegfoGkP-TCKvPzSU/edit?usp=sharing about the details. 

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/bidi/LogInspector.java
+++ b/java/src/org/openqa/selenium/bidi/LogInspector.java
@@ -7,7 +7,7 @@ import org.openqa.selenium.internal.Require;
 import java.util.*;
 import java.util.function.Consumer;
 
-public class LogInspector {
+public class LogInspector implements AutoCloseable {
 
   private final List<Consumer<ConsoleLogEntry>> consoleLogListeners = new LinkedList<>();
   private final List<Consumer<JavascriptLogEntry>> jsExceptionLogListeners = new LinkedList<>();
@@ -59,10 +59,9 @@ public class LogInspector {
           if (jsLogEntry.getLevel() == BaseLogEntry.LogLevel.ERROR) {
             jsExceptionLogListeners.forEach(
               consumer -> consumer.accept(jsLogEntry));
-          } else {
-            jsLogListeners.forEach(
-              consumer -> consumer.accept(jsLogEntry));
           }
+          jsLogListeners.forEach(
+            consumer -> consumer.accept(jsLogEntry));
         }
       );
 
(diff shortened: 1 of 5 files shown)


java/src/org/openqa/selenium/bidi/LogInspector.java
```diff
@@ -0,0 +1,114 @@
...
+  private final List<Consumer<GenericLogEntry>> genericLogListeners = new LinkedList<>();
+  private final Set<String> browsingContextIds;
+
+  private final BiDi bidi;
+
+  public LogInspector(WebDriver driver) {
+    this(new HashSet<>(), driver);
+  }
+
+  public LogInspector(String browsingContextId, WebDriver driver) {
```
krmahadevan, line 51: Maybe we could just do `this(Collections.singleton(browsingContextId), driver);`

java/src/org/openqa/selenium/bidi/LogInspector.java
```diff
@@ -0,0 +1,114 @@
+package org.openqa.selenium.bidi;
```
krmahadevan, line 18: Also I think the license header needs to be applied to both the classes (I am not sure if it gets auto applied in some way. If yes, please let me know and I will use that same mechanism as well)

titusfortner: we periodically run this to make sure everything is included: `./go copyright:update`

java/src/org/openqa/selenium/bidi/LogInspector.java
pujagani: We can call the constructor but it does not allow me to do null checks in that case. Which needs to happen before we call the constructor but when calling using `this(Collections.singleton(browsingContextId), driver);` it has to be the first statement.

krmahadevan: @pujagani - The null check for the contents of the collection would already be done by `Collections.singleton()` so we don't need to be doing it again. If we need it to be done explicitly we can create a private static method 

```java
private static Set<String> asSet(String text) {
    Require.nonNull("Browsing context id", text);
    return Collections.singleton(text);
}
```

and now within the constructor you can do 

```java
this(asSet(browsingContext), driver);
```

What do you think ?

krmahadevan: Ok. So `Collections.singleton()` doesn't do a null check.My bad. But my suggestion to do it via a static method should still do the trick and allow us to chain constructor calls.

pujagani: Thank you for your suggestion. The `Require.nonNull` does that already, it returns a value. I somehow missed that. Now we are able to chain constructors.

diemol (approved):
Thank you, @pujagani!

PR #11277, opened 2022-11-17
[grid] Support file downloads on the node

### LINKED ISSUE #9218: Selenium 4 dynamic grid - add support for get downloaded files to your host path
## 🚀 Feature Proposal

Selenium 4 Dynamic Grid - get downloaded files from eg. /home/seluser/Downloads to your host path. It should be same as get video file when you have to mount your host path (/path/on/your/host/machine:/opt/selenium/assets)

krmahadevan:
Fixes: #9218

On the node, set the directory to where the browser  will download files to via:

a. The CLI argument `--downloads-path` (or)
b. via the Toml syntax of 

```
[node]
downloads-path = "/path/to/dir/goes/here"
```
The GET end-points that will support file 
download would be (The file will ALWAYS BE DOWNLOADED AS A ZIP FILE):

* `/session/:sessionId:/se/file?filename=`

### Types of changes

### Checklist

- [X]  All new and existing tests passed.

### DIFF
--- a/java/src/org/openqa/selenium/grid/node/Node.java
+++ b/java/src/org/openqa/selenium/grid/node/Node.java
@@ -150,6 +150,12 @@ public abstract class Node implements HasReadyState, Routable {
       post("/session/{sessionId}/se/file")
         .to(params -> new UploadFile(this, sessionIdFrom(params)))
         .with(spanDecorator("node.upload_file")),
+      get("/session/{sessionId}/file")
+        .to(params -> new DownloadFile(this, sessionIdFrom(params)))
+        .with(spanDecorator("node.download_file")),
+      get("/session/{sessionId}/se/file")
+        .to(params -> new DownloadFile(this, sessionIdFrom(params)))
+        .with(spanDecorator("node.download_file")),
       get("/se/grid/node/owner/{sessionId}")
         .to(params -> new IsSessionOwner(this, sessionIdFrom(params)))
         .with(spanDecorator("node.is_session_owner").andThen(requiresSecret)),
@@ -215,6 +221,8 @@ public abstract class Node implements HasReadyState, Routable {
   }
 
   public abstract HttpResponse uploadFile(HttpRequest req, SessionId id);
+
+  public abstract HttpResponse downloadFile(HttpRequest req, SessionId id);
 
   public abstract void stop(SessionId id) throws NoSuchSessionException;
 
--- a/java/src/org/openqa/selenium/grid/node/Node.java
+++ b/java/src/org/openqa/selenium/grid/node/Node.java
@@ -150,9 +150,6 @@ public abstract class Node implements HasReadyState, Routable {
       post("/session/{sessionId}/se/file")
         .to(params -> new UploadFile(this, sessionIdFrom(params)))
         .with(spanDecorator("node.upload_file")),
-      get("/session/{sessionId}/file")
-        .to(params -> new DownloadFile(this, sessionIdFrom(params)))
-        .with(spanDecorator("node.download_file")),
       get("/session/{sessionId}/se/file")
         .to(params -> new DownloadFile(this, sessionIdFrom(params)))
         .with(spanDecorator("node.download_file")),
--- a/java/src/org/openqa/selenium/grid/node/config/NodeFlags.java
+++ b/java/src/org/openqa/selenium/grid/node/config/NodeFlags.java
@@ -222,12 +222,12 @@ public class NodeFlags implements HasRoles {
   private String nodeImplementation = DEFAULT_NODE_IMPLEMENTATION;
 
   @Parameter(
-    names = {"--downloads-dir"},
+    names = {"--downloads-path"},
     description = "The default location wherein all browser triggered file downloads would be "
       + "available to be retrieved from. This is usually the directory that you configure in "
       + "your browser as the default location for storing downloaded files.")
-  @ConfigValue(section = NODE_SECTION, name = "downloads-dir", example = "")
-  private String downloadsDir = "";
+  @ConfigValue(section = NODE_SECTION, name = "downloads-path", example = "")
+  private String downloadsPath = "";
 
   @Override
   public Set<Role> getRoles() {
--- a/java/src/org/openqa/selenium/grid/node/config/NodeFlags.java
+++ b/java/src/org/openqa/selenium/grid/node/config/NodeFlags.java
@@ -221,6 +221,14 @@ public class NodeFlags implements HasRoles {
     example = DEFAULT_NODE_IMPLEMENTATION)
   private String nodeImplementation = DEFAULT_NODE_IMPLEMENTATION;
 
+  @Parameter(
+    names = {"--downloads-dir"},
+    description = "The default location wherein all browser triggered file downloads would be "
+      + "available to be retrieved from. This is usually the directory that you configure in "
+      + "your browser as the default location for storing downloaded files.")
+  @ConfigValue(section = NODE_SECTION, name = "downloads-dir", example = "")
+  private String downloadsDir = "";
+
   @Override
   public Set<Role> getRoles() {
     return Collections.singleton(NODE_ROLE);
--- a/java/src/org/openqa/selenium/grid/node/config/NodeOptions.java
+++ b/java/src/org/openqa/selenium/grid/node/config/NodeOptions.java
@@ -148,7 +148,7 @@ public class NodeOptions {
     }
   }
 
-  public Optional<String> getDownloadsDirectory() {
+  public Optional<String> getDownloadsPath() {
     return config.get(NODE_SECTION, "downloads-path");
   }
 
--- a/java/src/org/openqa/selenium/grid/node/config/NodeOptions.java
+++ b/java/src/org/openqa/selenium/grid/node/config/NodeOptions.java
@@ -149,7 +149,7 @@ public class NodeOptions {
   }
 
   public Optional<String> getDownloadsDirectory() {
-    return config.get(NODE_SECTION, "downloads-dir");
+    return config.get(NODE_SECTION, "downloads-path");
   }
 
   public Node getNode() {
--- a/java/src/org/openqa/selenium/grid/node/config/NodeOptions.java
+++ b/java/src/org/openqa/selenium/grid/node/config/NodeOptions.java
@@ -148,6 +148,10 @@ public class NodeOptions {
     }
   }
 
+  public Optional<String> getDownloadsDirectory() {
+    return config.get(NODE_SECTION, "downloads-dir");
+  }
+
   public Node getNode() {
     return config.getClass(NODE_SECTION, "implementation", Node.class, DEFAULT_NODE_IMPLEMENTATION);
   }
--- a/java/src/org/openqa/selenium/grid/node/local/LocalNode.java
+++ b/java/src/org/openqa/selenium/grid/node/local/LocalNode.java
@@ -748,7 +748,7 @@ public class LocalNode extends Node {
     private Duration sessionTimeout = Duration.ofSeconds(NodeOptions.DEFAULT_SESSION_TIMEOUT);
     private HealthCheck healthCheck;
     private Duration heartbeatPeriod = Duration.ofSeconds(NodeOptions.DEFAULT_HEARTBEAT_PERIOD);
-    private String downloadsDir = "";
+    private String downloadsPath = "";
 
     private Builder(
       Tracer tracer,
@@ -803,8 +803,8 @@ public class LocalNode extends Node {
       return this;
     }
 
-    public Builder downloadsDirectory(String dir) {
-      this.downloadsDir = dir;
+    public Builder downloadsPath(String path) {
+      this.downloadsPath = path;
       return this;
     }
 
@@ -824,7 +824,7 @@ public class LocalNode extends Node {
         heartbeatPeriod,
         factories.build(),
         registrationSecret,
-        downloadsDir);
+        downloadsPath);
     }
 
     public Advanced advanced() {
--- a/java/src/org/openqa/selenium/grid/node/local/LocalNode.java
+++ b/java/src/org/openqa/selenium/grid/node/local/LocalNode.java
@@ -63,7 +63,6 @@ import org.openqa.selenium.io.TemporaryFilesystem;
 import org.openqa.selenium.io.Zip;
 import org.openqa.selenium.json.Json;
 import org.openqa.selenium.remote.SessionId;
-import org.openqa.selenium.remote.http.Contents;
 import org.openqa.selenium.remote.http.HttpRequest;
 import org.openqa.selenium.remote.http.HttpResponse;
 import org.openqa.selenium.remote.tracing.AttributeKey;
@@ -123,7 +122,7 @@ public class LocalNode extends Node {
   private final int maxSessionCount;
   private final int configuredSessionCount;
   private final boolean cdpEnabled;
-  private final String downloadsDir;
+  private final String downloadsPath;
 
   private final boolean bidiEnabled;
   private final AtomicBoolean drainAfterSessions = new AtomicBoolean();
@@ -148,7 +147,7 @@ public class LocalNode extends Node {
     Duration heartbeatPeriod,
     List<SessionSlot> factories,
     Secret registrationSecret,
-    String downloadsDir) {
+    String downloadsPath) {
     super(tracer, new NodeId(UUID.randomUUID()), uri, registrationSecret);
 
     this.bus = Require.nonNull("Event bus", bus);
@@ -165,7 +164,7 @@ public class LocalNode extends Node {
     this.sessionCount.set(drainAfterSessionCount);
     this.cdpEnabled = cdpEnabled;
     this.bidiEnabled = bidiEnabled;
-    this.downloadsDir = Optional.ofNullable(downloadsDir).orElse("");
+    this.downloadsPath = Optional.ofNullable(downloadsPath).orElse("");
 
     this.healthCheck = healthCheck == null ?
                        () -> {
@@ -180,10 +179,8 @@ public class LocalNode extends Node {
       .ticker(ticker)
       .removalListener((RemovalListener<SessionId, TemporaryFilesystem>) notification -> {
         TemporaryFilesystem tempFS = notification.getValue();
-        if (tempFS != null) {
-          tempFS.deleteTemporaryFiles();
-          tempFS.deleteBaseDir();
-        }
+        tempFS.deleteTemporaryFiles();
+        tempFS.deleteBaseDir();
       })
       .build();
 
@@ -485,17 +482,18 @@ public class LocalNode extends Node {
     if (slot != null && slot.getSession() instanceof DockerSession) {
       return executeWebDriverCommand(req);
     }
-    if (this.downloadsDir.isEmpty()) {
-      throw new WebDriverException(
-        "Please specify the directory that would contain downloaded files and restart the node.");
+    if (this.downloadsPath.isEmpty()) {
+      String msg = "Please specify the path wherein the files downloaded using the browser "
+        + "would be available via the command line arg [--downloads-path] and restart the node";
+      throw new WebDriverException(msg);
     }
-    File dir = new File(this.downloadsDir);
+    File dir = new File(this.downloadsPath);
     if (!dir.exists()) {
       throw new WebDriverException(
-        String.format("Cannot locate downloads directory %s.", downloadsDir));
+        String.format("Cannot locate downloads directory %s.", downloadsPath));
     }
     if (!dir.isDirectory()) {
-      throw new WebDriverException(String.format("Invalid directory: %s.",downloadsDir));
+      throw new WebDriverException(String.format("Invalid directory: %s.", downloadsPath));
     }
     String filename = req.getQueryParameter("filename");
     try {
@@ -504,13 +502,13 @@ public class LocalNode extends Node {
       ).orElse(new File[]{});
       if (allFiles.length == 0) {
         throw new WebDriverException(
-          String.format("Cannot find file [%s] in directory %s.", filename, downloadsDir));
+          String.format("Cannot find file [%s] in directory %s.", filename, downloadsPath));
       }
       if (allFiles.length != 1) {
         throw new WebDriverException(
           String.format("Expected there to be only 1 file. There were: %s.", allFiles.length));
       }
-      String content = Contents.string(allFiles[0]);
+      String content = Zip.zip(allFiles[0]);
       ImmutableMap<String, Object> result = ImmutableMap.of(
         "filename", filename,
         "contents", content);
--- a/java/src/org/openqa/selenium/grid/node/local/LocalNode.java
+++ b/java/src/org/openqa/selenium/grid/node/local/LocalNode.java
@@ -63,6 +63,7 @@ import org.openqa.selenium.io.TemporaryFilesystem;
 import org.openqa.selenium.io.Zip;
 import org.openqa.selenium.json.Json;
 import org.openqa.selenium.remote.SessionId;
+import org.openqa.selenium.remote.http.Contents;
 import org.openqa.selenium.remote.http.HttpRequest;
 import org.openqa.selenium.remote.http.HttpResponse;
 import org.openqa.selenium.remote.tracing.AttributeKey;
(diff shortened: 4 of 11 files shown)


java/src/org/openqa/selenium/grid/node/Node.java
```diff
@@ -150,6 +150,12 @@ protected Node(Tracer tracer, NodeId id, URI uri, Secret registrationSecret) {
       post("/session/{sessionId}/se/file")
         .to(params -> new UploadFile(this, sessionIdFrom(params)))
         .with(spanDecorator("node.upload_file")),
+      get("/session/{sessionId}/file")
+        .to(params -> new DownloadFile(this, sessionIdFrom(params)))
```
diemol: We only would like to have a prefixed `se` endpoint because this particular functionality is not part of the W3C standard. The upload one is kept in both ways for backwards compatibility reasons.

java/src/org/openqa/selenium/grid/node/config/NodeFlags.java
```diff
@@ -221,6 +221,14 @@ public class NodeFlags implements HasRoles {
     example = DEFAULT_NODE_IMPLEMENTATION)
   private String nodeImplementation = DEFAULT_NODE_IMPLEMENTATION;
 
+  @Parameter(
+    names = {"--downloads-dir"},
```
diemol:
```suggestion
    names = {"--downloads-path"},
```

java/src/org/openqa/selenium/grid/node/config/NodeFlags.java
```diff
@@ -221,6 +221,14 @@ public class NodeFlags implements HasRoles {
...
+    description = "The default location wherein all browser triggered file downloads would be "
+      + "available to be retrieved from. This is usually the directory that you configure in "
+      + "your browser as the default location for storing downloaded files.")
+  @ConfigValue(section = NODE_SECTION, name = "downloads-dir", example = "")
```
diemol:
```suggestion
  @ConfigValue(section = NODE_SECTION, name = "downloads-path", example = "")
```

java/src/org/openqa/selenium/grid/node/config/NodeFlags.java
```diff
@@ -221,6 +221,14 @@ public class NodeFlags implements HasRoles {
...
+  private String downloadsDir = "";
```
diemol:
```suggestion
  private String downloadsPath = "";
```

java/src/org/openqa/selenium/grid/node/config/NodeOptions.java
```diff
@@ -148,6 +148,10 @@ public Optional<URI> getPublicGridUri() {
     }
   }
 
+  public Optional<String> getDownloadsDirectory() {
+    return config.get(NODE_SECTION, "downloads-dir");
```
diemol:
```suggestion
    return config.get(NODE_SECTION, "downloads-path");
```

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
```diff
@@ -145,7 +147,8 @@ private LocalNode(
     Duration sessionTimeout,
     Duration heartbeatPeriod,
     List<SessionSlot> factories,
-    Secret registrationSecret) {
+    Secret registrationSecret,
+    String downloadsDir) {
```
diemol:
```suggestion
    String downloadsPath) {
```

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
```diff
@@ -122,6 +123,7 @@ public class LocalNode extends Node {
   private final int maxSessionCount;
   private final int configuredSessionCount;
   private final boolean cdpEnabled;
+  private final String downloadsDir;
```
diemol:
```suggestion
  private final String downloadsPath;
```

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
```diff
@@ -176,8 +180,10 @@ private LocalNode(
       .ticker(ticker)
       .removalListener((RemovalListener<SessionId, TemporaryFilesystem>) notification -> {
         TemporaryFilesystem tempFS = notification.getValue();
-        tempFS.deleteTemporaryFiles();
-        tempFS.deleteBaseDir();
+        if (tempFS != null) {
+          tempFS.deleteTemporaryFiles();
+          tempFS.deleteBaseDir();
+        }
```
diemol: Could you please revert this? It is not related to the PR.

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
```diff
@@ -471,6 +477,49 @@ public HttpResponse executeWebDriverCommand(HttpRequest req) {
...
+  public HttpResponse downloadFile(HttpRequest req, SessionId id) {
+    // When the session is running in a Docker container, the download file command
+    // needs to be forwarded to the container as well.
+    SessionSlot slot = currentSessions.getIfPresent(id);
+    if (slot != null && slot.getSession() instanceof DockerSession) {
+      return executeWebDriverCommand(req);
+    }
+    if (this.downloadsDir.isEmpty()) {
+      throw new WebDriverException(
+        "Please specify the directory that would contain downloaded files and restart the node.");
```
diemol: It is better to tell the user how to do that, ideally to point them to the docs link.

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
```diff
@@ -471,6 +477,49 @@ public HttpResponse executeWebDriverCommand(HttpRequest req) {
...
+      }
+      if (allFiles.length != 1) {
+        throw new WebDriverException(
+          String.format("Expected there to be only 1 file. There were: %s.", allFiles.length));
+      }
+      String content = Contents.string(allFiles[0]);
+      ImmutableMap<String, Object> result = ImmutableMap.of(
+        "filename", filename,
+        "contents", content);
+      return new HttpResponse().setContent(asJson(result));
```
diemol: Wouldn't it be better to zip the file and send it back?

java/src/org/openqa/selenium/remote/http/Contents.java
```diff
@@ -138,6 +143,18 @@ public static Supplier<InputStream> memoize(Supplier<InputStream> delegate) {
...
+      InputStream isr = Files.newInputStream(input.toPath())) {
+      int len;
+      byte[] buffer = new byte[4096];
+      while ((len = isr.read(buffer)) != -1) {
+        bos.write(buffer, 0, len);
+      }
+      return Base64.getEncoder().encodeToString(bos.toByteArray());
+    }
+  }
+
```
diemol, line 157: Not sure if this is a good idea, I'd prefer we send back a zipped file.

diemol (changes requested):
Thanks @krmahadevan! I left some comments.

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
krmahadevan: I did think of that as the first implementation but then again, the user wouldn't know that they are going to get back a zip file which they would need to unzip, even though they are downloading only 1 file at a time. We can perhaps provide an additional query parameter to indicate that the user needs the file in zip format? What do you think ?

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
krmahadevan: The docs link right now does not exist because I haven't yet created the documentation for it. So maybe I can come back with a different PR, that updates this with a doc link ? Alternatively I can just include some verbose text here that exactly shows how to do it here itself ? What do you think ?

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
diemol: It is about the traffic the Grid needs to handle, sending several MB files across the Node, Router, etc... vs compressing them to keep the traffic as light as possible. We can always document how they get the file and provide examples on different languages showing how to unzip a file.

java/src/org/openqa/selenium/grid/node/Node.java
krmahadevan: Fixed

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
diemol: Perhaps a brief version of the instructions is fine. Later the link can be added.

java/src/org/openqa/selenium/grid/node/config/NodeFlags.java
krmahadevan: Fixed

java/src/org/openqa/selenium/grid/node/config/NodeFlags.java
krmahadevan: Fixed

java/src/org/openqa/selenium/grid/node/Node.java
diemol: No need to comment on every review. Resolving the conversation is enough.

java/src/org/openqa/selenium/grid/node/config/NodeOptions.java
krmahadevan: Fixed

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
krmahadevan: Fixed

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
krmahadevan: Fixed

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
krmahadevan: Fixed

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
krmahadevan: Cool. Updating it now.

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
krmahadevan: Ok. Cool. I will change this to accommodate a zip file being sent back.

krmahadevan:
@diemol - I have fixed all the review comments, rebased off of `trunk` and added the changes as an additional commit for ease of review. Please help take a look

java/test/org/openqa/selenium/grid/distributor/AddingNodesTest.java
```diff
@@ -384,6 +384,11 @@ public HttpResponse uploadFile(HttpRequest req, SessionId id) {
       throw new UnsupportedOperationException("uploadFile");
     }
 
+    @Override
+    public HttpResponse downloadFile(HttpRequest req, SessionId id) {
+      throw new UnsupportedOperationException("uploadFile");
```
diemol:
```suggestion
      throw new UnsupportedOperationException("downloadFile");
```

java/src/org/openqa/selenium/grid/node/config/NodeOptions.java
```diff
@@ -148,6 +148,10 @@ public Optional<URI> getPublicGridUri() {
     }
   }
 
+  public Optional<String> getDownloadsDirectory() {
+    return config.get(NODE_SECTION, "downloads-path");
+  }
+
```
diemol: This code is not invoked anywhere. I believe it should be called from the `create` method in `LocalNodeFactory`.

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
```diff
@@ -755,6 +803,11 @@ public Builder heartbeatPeriod(Duration heartbeatPeriod) {
       return this;
     }
 
+    public Builder downloadsDirectory(String dir) {
+      this.downloadsDir = dir;
+      return this;
+    }
+
```
diemol: This also needs to be invoked.

diemol (changes requested):
I checked out this PR to run a complete test but it did not work. Probably because the get method for `downloadsPath` configuration is not being used.

The client code I used to perform a test was:

```java
package org.example;

import org.openqa.selenium.WebDriver;
import org.openqa.selenium.WebElement;
import org.openqa.selenium.firefox.FirefoxOptions;
import org.openqa.selenium.io.Zip;
import org.openqa.selenium.json.Json;
import org.openqa.selenium.remote.RemoteWebDriver;
import org.openqa.selenium.remote.http.HttpClient;
import org.openqa.selenium.remote.http.HttpMethod;
import org.openqa.selenium.remote.http.HttpRequest;
import org.openqa.selenium.remote.http.HttpResponse;

import java.io.File;
import java.io.IOException;
import java.net.MalformedURLException;
import java.net.URL;
import java.util.Map;

import static org.openqa.selenium.remote.http.Contents.string;

public class Main {

  public static void main(String[] args) throws InterruptedException, IOException {
    FirefoxOptions options = new FirefoxOptions();
    options.addPreference("browser.download.manager.showWhenStarting", false);
    options.addPreference("browser.helperApps.neverAsk.saveToDisk", "images/jpeg, application/pdf, application/octet-stream");
    options.addPreference("pdfjs.disabled", true);

    URL gridUrl = new URL("http://localhost:4444");
    RemoteWebDriver driver =  new RemoteWebDriver(gridUrl, options);
    driver.get("http://the-internet.herokuapp.com/download");
    WebElement element = driver.findElement(By.cssSelector(".example a"));
    element.click();

    Thread.sleep(10 * 1000);

    HttpRequest request = new HttpRequest(
      HttpMethod.GET,
      String.format("/session/%s/se/file", driver.getSessionId()));
    request.addQueryParameter("filename", "my_appointments-1.pdf");
    try (HttpClient client = HttpClient.Factory.createDefault().createClient(gridUrl)) {
      Map<String, Object> map = new Json().toType(string(response), Json.MAP_TYPE);
      String encodedContents = map.get("contents").toString();
      Zip.unzip(encodedContents, new File("/Users/diegomolina/Downloads"));
    }

    driver.quit();
  }
}
```

krmahadevan:
@diemol - Thanks for catching that bug! Fixed the review comments, rebased off of trunk and force pushed changes back to this branch. Please check

diemol (approved):
Thank you, @krmahadevan!

diemol:
@krmahadevan would be great if you can help us with the docs for this new feature 💪

krmahadevan:
> @krmahadevan would be great if you can help us with the docs for this new feature 💪

@diemol - I have raised a PR for this

PR: https://github.com/SeleniumHQ/seleniumhq.github.io/pull/1262

Doc links: 

* https://deploy-preview-1262--jovial-austin-42fe02.netlify.app/documentation/grid/configuration/cli_options/#specifying-path-from-where-downloaded-files-can-be-retrieved
* https://deploy-preview-1262--jovial-austin-42fe02.netlify.app/documentation/grid/configuration/toml_options/#retrieving-downloaded-files-from-node

iammac2:
Downloaded 4.8.0 and verified this feature. Working well👍🏼. Will the API also include an endpoint to clean up the downloaded files?

titusfortner:
imo it shouldn't have its own endpoint, it should be part of the cleanup process of ending the session. Once the session is finished, the files can't be accessed again.

iammac2:
Thank you @titusfortner and agree re session clean up. Have tried verifying this behaviour with a local Grid `4.8.0` config to verify what happens across sessions:

- Grid with one Node is running locally with the new `--downloads-path` CLI Node option is set i.e. `java -jar selenium-server-4.8.0.jar node --downloads-path C:\tmp`
- Chrome pref `download.default_directory` also set to same path when `RemoteWebDriver` initialized.
- Run a test on Node (noting session ID in Node log) that clicks to download a file from UI at the remote end and then retrieves the file via the new GET request.
- The test passes first time through with the file downloaded to Node download directory and GET response contains correct `filename` and `contents`.
- Wait for Node debug to state `Stopping session`
- Run the test again (i.e. new session ID in Node log) but WITHOUT the step that clicks to download the file and ONLY sends the GET request. (Would expect this scenario to fail when endpoint hit due to session cleardown). 
- In this scenario the GET response **still contains** the same `filename` and `contents`. 
- Physically checking the Node download path after Node log states `Stopping session`, confirms that the file is still present in the directory.

titusfortner:
@iammac2 I just split out the work i'd like to see us do to get downloads working. The way I want to see us support things will look like this in the bindings: https://github.com/SeleniumHQ/selenium/issues/11657

iammac2:
I see, thank you for the extra info. Will keep a watch in this further development.
So existing implementation of `GET /session/:sessionId:/se/files` endpoint will need to change as download path will need to be [generated](https://github.com/SeleniumHQ/selenium/issues/11656) and not set by user e.g `--downloads-path` and browser prefs `download.default_directory`? Grid Node will then hold internal reference by `uuid` to the download directories so it can clean them up (ref the [DELETE](https://github.com/SeleniumHQ/selenium/issues/11457))

titusfortner:
That's about right. We'll see what trouble we run into (if any) in the implementation.

vijaysp77:
@krmahadevan Can I use the preference settings for this download path inside the code (download.default_directory) instead of passing the download path while setting-up node? Because we are using docker container provisioned by aws ECS? Thanks

krmahadevan:
@vijaysp77 This feature is going through a revamp. 

The functionality will change once my PR https://github.com/SeleniumHQ/selenium/issues/11702 gets merged. 

So I would suggest that you follow the above PR and wait for this functionality to get streamlined.

patilsha:
@krmahadevan I saw your PR [1702 ](https://github.com/SeleniumHQ/selenium/pull/11702) has been merged.
Where can I find the latest **Selenium Hub Image** to see these changes in Selenium Grid.
Thanks for providing a solution to most awaited feature.

diemol:
@patilsha it was merged yesterday, it will be part of the next release.

xZero707:
I'm assuming this is not yet available in Selenium docker images? 

Setting SE_OPTS on either node or even hub yields `Was passed main parameter '--downloads-path' but no main parameter was defined in your arg class`. 
I tried the same with PR #11702 
 `Was passed main parameter '--enable-managed-download' but no main parameter was defined in your arg class`. 

Chrome node
```
GET /session/:sessionId:/se/files
= Unknown command
```

Firefox node
```
GET /session/:sessionId:/se/files
= Unsupported method - expect POST for file upload
```



I'm in dark. This feature is documented, yet cannot be used. Am I doing something wrong?

xZero707:
I'm assuming this is not yet available in Selenium docker images (4.8.1)? 

Chrome node
```
GET /session/:sessionId:/se/files
curl -X GET "http://172.17.53.2:4444/session/98473aa8661fc256bf7c0de98861005e/se/file?filename=100MB.bin"
= Unknown command
```

Firefox node
```
GET /session/:sessionId:/se/files
curl -X GET "http://172.17.53.2:4444/session/98473aa8661fc256bf7c0de98861005e/se/file?filename=100MB.bin"
= Unsupported method - expect POST for file upload
```

I also tried running this using PHP driver:
```php
$file = $driver->executeCustomCommand('/session/:sessionId/se/file?filename=100MB.bin');
```

Result:
```
Facebook\WebDriver\Exception\UnknownCommandException : unknown command: unknown command: session/41d161b598c0fd3f85a4e09b1f5f5a1f/se/file?filename=100MB.bin`
```

I'm in dark. This feature is documented, yet cannot be used. Am I doing something wrong?

I defined `--downloads-path /usr/downloads` using env `SE_OPTS` and nothing. Chrome is still downloading to `/home/seluser/Downloads` and ignoring the path I defined. 


Version: Selenium 4.8.1 (revision 8ebccac989)

krmahadevan:
@xZero707 - For this to be available in the docker selenium project, I think this should be released first. The feature has gone through some revamp.  So please create an issue on the docker-selenium project that tracks this requirement and you can also cross link this issue there. 
The documentation on this also needs to be changed. I haven't found time to get to this yet. I should hopefully be able to wrap the documentation before this weekend.

xZero707:
@krmahadevan Thank you for clarification. It makes sense. I will do as you suggested.

patilsha:
@xZero707 have you able to implement this new functionality in your project? If yes can you please share the steps over here?
I tried but still no luck and the documentation link provided above is broken. Thanks

naveenmadd:
> I checked out this PR to run a complete test but it did not work. Probably because the get method for `downloadsPath` configuration is not being used.
> 
> The client code I used to perform a test was:
> 
> ```java
> package org.example;
> 
> import org.openqa.selenium.By;
> import org.openqa.selenium.WebDriver;
> import org.openqa.selenium.WebElement;
> import org.openqa.selenium.firefox.FirefoxOptions;
> import org.openqa.selenium.io.Zip;
> import org.openqa.selenium.json.Json;
> import org.openqa.selenium.remote.RemoteWebDriver;
> import org.openqa.selenium.remote.http.HttpClient;
> import org.openqa.selenium.remote.http.HttpMethod;
> import org.openqa.selenium.remote.http.HttpRequest;
> import org.openqa.selenium.remote.http.HttpResponse;
> 
> import java.io.File;
> import java.io.IOException;
> import java.net.MalformedURLException;
> import java.net.URL;
> import java.util.Map;
> 
> import static org.openqa.selenium.remote.http.Contents.string;
> 
> public class Main {
> 
>   public static void main(String[] args) throws InterruptedException, IOException {
>     FirefoxOptions options = new FirefoxOptions();
>     options.addPreference("browser.download.manager.showWhenStarting", false);
>     options.addPreference("browser.helperApps.neverAsk.saveToDisk", "images/jpeg, application/pdf, application/octet-stream");
>     options.addPreference("pdfjs.disabled", true);
> 
>     URL gridUrl = new URL("http://localhost:4444");
>     RemoteWebDriver driver =  new RemoteWebDriver(gridUrl, options);
>     driver.get("http://the-internet.herokuapp.com/download");
>     WebElement element = driver.findElement(By.cssSelector(".example a"));
>     element.click();
> 
>     Thread.sleep(10 * 1000);
> 
>     HttpRequest request = new HttpRequest(
>       HttpMethod.GET,
>       String.format("/session/%s/se/file", driver.getSessionId()));
>     request.addQueryParameter("filename", "my_appointments-1.pdf");
>     try (HttpClient client = HttpClient.Factory.createDefault().createClient(gridUrl)) {
>       HttpResponse response = client.execute(request);
>       Map<String, Object> map = new Json().toType(string(response), Json.MAP_TYPE);
>       String encodedContents = map.get("contents").toString();
>       Zip.unzip(encodedContents, new File("/Users/diegomolina/Downloads"));
>     }
> 
>     driver.quit();
>   }
> }
> ```

@diemol  Can you pls help me here, I'm trying to test File Downloads on Jenkins Node but
**Getting [ERROR] COMPILATION ERROR : at Map<String, Object> map = new Json().toType(string(response), Json.MAP_TYPE); as** 
no suitable method found for string(org.openqa.selenium.remote.http.HttpResponse)
[ERROR] method org.openqa.selenium.remote.http.Contents.string(java.io.File) is not applicable
[ERROR] (argument mismatch; org.openqa.selenium.remote.http.HttpResponse cannot be converted to java.io.File)
[ERROR] method org.openqa.selenium.remote.http.Contents.string(org.openqa.selenium.remote.http.HttpMessage<?>) is not applicable
[ERROR] (argument mismatch; org.openqa.selenium.remote.http.HttpResponse cannot be converted to org.openqa.selenium.remote.http.HttpMessage<?>)

krmahadevan:
@naveenmadd - If you wanted to build/experiment something on top of this changes, I would suggest that you do it from `trunk`. The feature has undergone changes after it was merged.

naveenmadd:
> @naveenmadd - If you wanted to build/experiment something on top of this changes, I would suggest that you do it from `trunk`. The feature has undergone changes after it was merged.

Even I verified selenium-trunk repository, I see the same implementation [ package org.openqa.selenium.grid.node > class NodeTest > @Test @DisplayName("DownloadsTestCase") Map<String, Object> map = new Json().toType(string(response), Json.MAP_TYPE);]  
Don't know how it's accepting in that test and not for mine? suggest if you have any insights or try from your end too once..

Thanks @krmahadevan  @diemol

diemol:
This is the test case https://github.com/SeleniumHQ/selenium/blob/trunk/java/test/org/openqa/selenium/grid/router/RemoteWebDriverDownloadTest.java#L134

Issue #11270, opened 2022-11-16
[🐛 Bug]: Memory leak by creating multiple HttpClient-*-SelectorManager  threads when using jdk11 http client in loop with selenium grid


pogrebnyaksa:
### What happened?

I experience memory leak in my app - selenium spawns multiple threads.

### How can we reproduce the issue?

```shell
<properties>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
        <selenium.version>4.6.0</selenium.version>
        <lombok.version>1.18.24</lombok.version>
        <spring-web.version>5.3.23</spring-web.version>
        <jackson-databind.version>2.14.0</jackson-databind.version>
    </properties>

    <dependencies>
        <dependency>
            <groupId>org.seleniumhq.selenium</groupId>
            <artifactId>selenium-chrome-driver</artifactId>
            <version>${selenium.version}</version>
        </dependency>
        <dependency>
            <groupId>org.seleniumhq.selenium</groupId>
            <artifactId>selenium-devtools-v107</artifactId>
            <version>${selenium.version}</version>
        </dependency>
        <dependency>
            <groupId>org.seleniumhq.selenium</groupId>
            <artifactId>selenium-support</artifactId>
            <version>${selenium.version}</version>
        </dependency>
        <dependency>
            <groupId>org.seleniumhq.selenium</groupId>
            <artifactId>selenium-http-jdk-client</artifactId>
            <version>${selenium.version}</version>
        </dependency>

        <!-- https://mvnrepository.com/artifact/org.springframework/spring-web -->
        <dependency>
            <groupId>org.springframework</groupId>
            <artifactId>spring-web</artifactId>
            <version>${spring-web.version}</version>
        </dependency>
        <dependency>
            <groupId>com.fasterxml.jackson.core</groupId>
            <artifactId>jackson-databind</artifactId>
            <version>${jackson-databind.version}</version>
        </dependency>



        <dependency>
            <groupId>org.projectlombok</groupId>
            <artifactId>lombok</artifactId>
            <version>${lombok.version}</version>
            <scope>provided</scope>
        </dependency>
    </dependencies>



    
    @SneakyThrows
    public static void main(String[] args) {
        System.setProperty("webdriver.http.factory", "jdk-http-client");

        while (true) {
            WebDriver driver = null;
            try {
                ChromeOptions chromeOptions = new ChromeOptions();
                chromeOptions.setHeadless(false);
                chromeOptions.addArguments(
[...]

PR #11283, opened 2022-11-19
[java] timely get rid of the SelectorManager threads

joerg1985:
### Description
In cases where a lot of WebDrivers are created, timely closed and there is no GC done, a lot of SelectorManager threads are running. Calling the GC explicit is not best practice, but it is the only way to stop these threads. 

Has been reported in #11270

### Types of changes

### Checklist


pujagani:
The approach suggested here is not ideal since it manually calls GC. Closing in favor of https://github.com/SeleniumHQ/selenium/pull/11345 which helps fix the same issue.

PR #11286, opened 2022-11-20
[grid] improved exception handling when no cause is set

joerg1985:
### Description
In cases where the exception thrown has no cause there was no helpful stack trace to localize the issue.

### Motivation and Context
I was playing with another PR and added a `ImmutableMap.of("value", null)` call, this lead to an exception without cause set. I was not able to get the origin in code of the exception due to the missing infos in the server answer.

### Types of changes

### Checklist


diemol (approved):
Thank you, @joerg1985!

Issue #11272, opened 2022-11-16
[🐛 Bug]: "Near" relative locator fails to find elements under certain conditions


alpatron:
### What happened?

The "near" relative locator fails to locate elements in certain cases, like when a narrower object is directly above a wider object (see provided screenshot).
![obrazek](https://user-images.githubusercontent.com/1508117/202175226-2d45f55e-e250-4f48-8734-1fcabd65d185.png)

In this case, trying to locate " 'Rectangle 1' near 'Rectangle 2' " results in a `NoSuchElementException` exception even though the rectangles certainly look like they are within 50 px of each other.

I actually located the source of the issue:

https://github.com/SeleniumHQ/selenium/blob/abde4ce306d81ad913b2d9353b4f25588ec9c317/javascript/atoms/locators/relative.js#L169-L211

The algorithm to determine the proximity of object boundaries seems to perform some naive calculations, whereas what probably is desired is to calculate the shortest distance between the two rectangles' boundaries (or centres if they are overlapping).

So the code probably ought to look something like the example in this StackOverflow answer (https://stackoverflow.com/a/26178015).

I wrote that this issue is present in Selenium 4.6.0 for Python, but since the issue seems to stem from some JavaScript atoms, the issue is probably present in all versions of Selenium, but I did not check, since I only use Selenium for Python.

### How can we reproduce the issue?

```shell
https://github.com/alpatron/SeleniumRelativeLocatorBug
```

### Relevant log output

```shell
DevTools listening on ws://127.0.0.1:12320/devtools/browser/9ef7f89b-4fcb-40f3-9789-dbff1edade6b
Traceback (most recent call last):
  File "C:\Users\Vikto\Desktop\SeleniumTests Big Boy\SeleniumRelativeLocatorBug\bug.py", line 8, in <module>
    driver.find_element(
  File "C:\Users\Vikto\Desktop\SeleniumTests Big Boy\Lib\site-packages\selenium\webdriver\remote\webdriver.py", line 848, in find_element
    raise NoSuchElementException(f"Cannot locate relative element with: {by.root}")
selenium.common.exceptions.NoSuchElementException: Message: Cannot locate relative element with: {'id': 'rect1'}
```

### Operating System

Windows 10 21H1

### Selenium version

Python 4.6.0 (but should be present in all versions; see "What happened" section)

### What are the browser(s) and version(s) where you see this issue?

Chrome 107

### What are the browser driver(s) and version(s) where you see this issue?

[...]

PR #11290, opened 2022-11-21
improve "near" relative locator behaviour

alpatron:
### Description
I changed the implementation and behaviour of the "near" relative locator.

The documentation states that the near locator finds elements that "are at most 50px away from the provided locator".[^sel_doc]
Currently, the code accomplishes this by first checking some kinds of edge-to-edge distances to determine one way of closeness. Then, if this check fails, it checks to see if the centres of the bounding rectangles are within the specified distance. If this check fails as well, the elements are deemed not close. This approach breaks in certain cases as described in issue #11272.

I replaced and simplified this algorithm like so:

1. One element's bounding rectangle is scaled up, adding the specified look-up distance to its width and height, keeping the centre of the rectangle the same.
2. We check if the scaled-up bounding box intersects with the non-scaled-up bounding box of the second element.
The intersection checking is performed by the built-in `ìntersects` method, since the code uses `goog.math.Rect` objects that have this feature.[^goog_rect]

This diagram might better illustrate how this approach works:

ilust

It should be noted that this thus changes the behaviour of the near locator, and existing code that expects precisely the old behaviour of the near locator might break, but since the behaviour of the near locator was always just specified as finding elements "at most 50px away from the provided locator", I don't believe that much code exists that utilises the at-times-odd behaviour of the current near locator. On the other hand, this new implementation locates elements that seemingly should be located by this locator but aren't, as described in issue #11272.

Another noteworthy mention is that the way I implemented the locator doesn't use Euclidean distance to determine "nearness", but rather Chebyshev distance. A diagram probably best describes the difference. Shown is what a proper Euclidean near area would look like.

ilust2

If proper Euclidean distance is deemed crucial to be used in Selenium, it's possible to implement it by using the algorithm for example on this StackOverflow post (<https://stackoverflow.com/a/26178015>). But this would require more complex and longer code, and I think that the rectangular Chebyshev approach is sufficient for use and probably more natural to use than the proper Euclidean way, as I'd believe most developers think in regular rectangles rather than rectangles with rounded corners.

Technically, the documentation need not change, since the code still looks for elements that "are at most 50px away from the provided locator". What has changed is how that definition is interpreted. But given this definition is not unambiguous, the documentation probably ought to be updated in any case—whether this PR is accepted or not. I suppose if this PR is deemed acceptable, I could update the relevant documentation.

I modified the `relative_locators.html` file to trigger the edge case behaviour in issue  #11272.

I had some trouble with tests: I developed this PR using gitpod, as was recommended in the README, but some tests were failing even on a fresh repo with the latest commit from the trunk branch with no changes made. I don't know if this a problem with gitpod or whether this is expected behaviour. But after performing the fix, no extra failures were recorded. All the relative-locator tests are passing.

### Motivation and Context
Fixes issue #11272. The near relative behaviour sometimes doesn't find elements that seem like they should be within 50px of the referenced element.

### Types of changes

### Checklist

- [?] My change requires a change to the documentation.
- [?] I have added tests to cover my changes.
- [?] All new and existing tests passed.

[^sel_doc]: https://www.selenium.dev/documentation/webdriver/elements/locators/#near
[^goog_rect]: https://google.github.io/closure-library/api/goog.math.Rect.html


diemol (commented):
Thank you for this, @alpatron.

I believe we would need a blog post to describe the changes, which is basically the description of the PR you are sending.

Also, have you checked the relative locators tests? Don't mind the unrelated failing ones. This should probably break most of them across bindings.

alpatron:
Hi, @diemol.

I checked the relative-locator tests (at least the ones that I was able to run—Python, JavaScript, and Java), and they all pass. Well, they pass now: I had modified the `relative_locators.html` file to trigger the edge case behaviour in issue https://github.com/SeleniumHQ/selenium/issues/11272, but since some tests were using that file for the testing of relative-locator result ordering (the order of elements returned by `driver.find_elements`), my modification broke those tests. So I rolled back the change to the `relative_locators.html` file and now this PR only includes the change to the relative-locator implementation.

This brings me to another point: Since this is a change to the Selenium atoms, I assume that it's needed to write tests that verify the fix for all language bindings. Given I can get only the Python, JavaScript, and Java tests to run on the provided gitpod environment, this could be difficult. Or are there atom-specific tests?

Then I have a question about some code in the atoms: The [Selenium documentation](https://www.selenium.dev/documentation/webdriver/elements/locators/#near) says that, by default, the "near" relative locator finds elements that are "at most 50px away from the provided locator". However, it seems that the code is using a default distance of 100 px. See the code excerpt from the unmodified `relative.js` file.

https://github.com/SeleniumHQ/selenium/blob/6d947064806ecc18284973dcf923c63a74e55070/javascript/atoms/locators/relative.js#L142-L153

It seems that the code is setting the default distance to 100 px, not 50 px. And I don't think this has to do with weird geometry rules either. The `relative.js` code was doing some edge-to-edge and centre-distance comparisons, nothing where some geometrical compensation would be necessary. See the code excerpt for the current "nearness" check.

https://github.com/SeleniumHQ/selenium/blob/6d947064806ecc18284973dcf923c63a74e55070/javascript/atoms/locators/relative.js#L166-L211

I believe that the `distance = 100` assignment is done in error and that the "near" relative locator is using a default distance of 100 px instead of 50 px. But I might be wrong, and there might be something I'm not aware of. Either way, if the locator should find elements within 50 px, then I believe I'd need to change the default `distance = 100` assignment to `distance = 50` because my changed code definitely expects `distance` to be, well, the desired distance from the source element. However, if Selenium has actually been looking for elements in a 100-px distance for all these years, then I suppose it could be better to keep the 100-px default distance for better backward compatibility and update the documentation with info about the bug.

And finally, speaking of the current way of the "near" locator, I also created a simple illustration to illustrate how the current way works. It'd be nice to get a second pair of eyes to look at it so I could check whether I understand the current way of how the "near" relative locator works. The red R2 rectangles are examples of "near" elements that the current way deems as not near, but my changed code would deem as near.

ilust2

diemol:
I was doing a quick check and it does seem a couple of languages do pass the distance as `50px`, but others end up with the default. 

I was not able to understand what breaks and what does not from your comment. I will let our tests run and see what breaks. I guess we need to modify `relative_locators.html` so the case can be tested and then adjust tests. Plus, also make all bindings use the 50px distance.

diemol:
After checking the CI, I believe we need to make changes in the `relative_locators.html` to properly test the change. If you want, please add a test in the language you prefer and I can help replicate it across the other languages.

alpatron:
Hi, @diemol.

I added tests for the Python bindings. I changed the `relative_locators.html` file by adding the buggy case I described in issue #11272. I added two tests: One test that should find a rectangle near the other. And the other than should not, because the rectangles are more than 50 px from each other.

I also changed the default search distance in `relative.js` to 50. As I said, I don't know whether previously the default value of 100 was a bug or whether it was needed for some weird reason, but my new code needs this value to be 50 to look within 50 px of the anchor.

diemol (approved):
Thank you, @alpatron!

PR #11295, opened 2022-11-22
Setup Ruby tests CI on Windows

p0deje:
Adds support for running Ruby tests on Windows. This includes Chrome, Firefox, Edge, and unit tests.

Running tests on the WIndows is much slower (e.g. https://github.com/SeleniumHQ/selenium/actions/runs/3524631700/usage) partially because the cache cannot be easily re-used. Hopefully, setting up Bazel remote caching and remote execution would allow us to overcome this.

It's also worth mentioning that running tests on the Windows is more flaky (remote Firefox tests in particular) but I suggest we iron this in trunk already.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,4 +1,5 @@
 try-import .bazelrc.local
+try-import .bazelrc.windows.local
 
 # Ensure Windows support is accurate.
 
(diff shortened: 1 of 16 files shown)


titusfortner (approved):

p0deje:
@diemol I'd appreciate if you could take a look at the changes :)

Issue #11291, opened 2022-11-22
[🐛 Bug]: Using selenium 4.6 in Ubuntu not able to download the required driver (for chrome) instead gives IllegalStateException 


PriyomSaha:
### What happened?

- I am running the Ubuntu in Virtual Box.
- I am running a simple browser opening visiting youtube.com.
- When I am using WebdriverManager it works fine but when I am commenting that particular setup() it throws java.lang.IllegalStateException.
- Using JAVA. 
- Attached the small 1 line sample code and error.

- **As I am using a Virtual Machine where can I expect the driver to physically occur ?**
- **Where in linux machine does the drivers gets downloaded ? Actual path if someone can provide will try once by physically placing the driver**

![image](https://user-images.githubusercontent.com/44141916/203241010-7bfec6cb-22d3-43de-ac39-ce0f87948b17.png)
![image](https://user-images.githubusercontent.com/44141916/203241267-1e66144c-3b31-4d83-a055-ceef636f83ce.png)

### How can we reproduce the issue?

```shell
-Installed VM Virtual Box.
-Loaded the Ubuntu 22.10 iso into it.
-Installed java.
-Installed an preferred IDE (IntelliJ)
-Created a maven project.
-Added Selenium 4.6 as one of the dependency
```

### Relevant log output

```shell
![image](https://user-images.githubusercontent.com/44141916/203241010-7bfec6cb-22d3-43de-ac39-ce0f87948b17.png)
![image](https://user-images.githubusercontent.com/44141916/203241267-1e66144c-3b31-4d83-a055-ceef636f83ce.png)
```

### Operating System

Ubuntu

### Selenium version

4.6

### What are the browser(s) and version(s) where you see this issue?

Chrome 107

### What are the browser driver(s) and version(s) where you see this issue?

Not able to fetch the driver

### Are you using Selenium Grid?

No

bonigarcia:
Thanks for reporting @PriyomSaha. I reproduced the issue in the same environment (Ubuntu 22.10 with VirtualBox). The underlying problem is that Selenium Manager (which is a CLI application) cannot be executed in this distro:

```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.10
Release:	22.10
Codename:	kinetic

$ ./selenium-manager 
./selenium-manager: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory
```

[...]

PR #11300, opened 2022-11-23
[java] Include selenium-manager output in Java exception

bonigarcia:
### Description
This PR changes the exception message when some unexpected error happens executing Selenium Manager, including the command output instead of the command itself (which is already logged as a warning trace).

### Motivation and Context
The Selenium Mangear output will allow to debug future problems (e.g. #11291) easily.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/manager/SeleniumManager.java
+++ b/java/src/org/openqa/selenium/manager/SeleniumManager.java
@@ -91,7 +91,7 @@ public class SeleniumManager {
         String output = "";
         try {
             Process process = new ProcessBuilder(command)
-                    .redirectErrorStream(false).start();
+                    .redirectErrorStream(true).start();
             process.waitFor();
             output = CharStreams.toString(new InputStreamReader(
                     process.getInputStream(), StandardCharsets.UTF_8));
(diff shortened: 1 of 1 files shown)


titusfortner:
With the currently released selenium-manager, if there is a problem, the output is always an empty string, so this change won't help.

Have the changes to the selenium manager error handling changed this?
Do we need to be looking at stdout or stderr or both?

bonigarcia:
When some unexpected error happens, the error message goes to _stderr_. So we need to capture both _stdout_ and _stderr_. I updated this PR to capture also _stderr_, which was missing previously (thanks for pointing it out). So now, the variable `output` in the Java code is not empty when some error happens (this will be very useful for debugging in the future). I think this deserves to be done also in the rest of bindings.

titusfortner:
Can we still log the full command being executed, or does that show up in stderr?

bonigarcia:
We can include the command as well (I have just appended it in the current PR). Currently, the command itself does not add many value (since the selenium-manager binary is executed in a temporal folder which is later deleted). But if we decide to store the binary the the cache, it will help to reproducibility (i.e., the user will be able to invoke selenium-manager manually).

titusfortner:
Yeah, I was more thinking about going forward. Like, if you send arguments `--driver-version 102` it will fail because it needs to be the full driver version, and we want to be able to have visibility into that.

diemol (approved):

PR #11308, opened 2022-11-24
Replace response.getheader() with response.headers.get()

mdmintz:
### Description

Replace response.getheader() with response.headers.get() - (deprecated in urllib3 1.26.13)

### Motivation and Context

This will prevent lots of deprecation warnings when using selenium with the latest ``urllib3``

### Types of changes

### Checklist


py/CHANGES
```diff
@@ -1,3 +1,6 @@
+Selenium (TBA version)
```
symonk, line 1: minor: I think we are gonna go with a `4.6.1` here, can you update that? could I also ask you to add a line about the `__del__` changes too, would be greatly appreciated

py/selenium/webdriver/remote/remote_connection.py
```diff
@@ -319,23 +319,17 @@ def _request(self, method, url, body=None):
...
             if 300 <= statuscode < 304:
-                return self._request("GET", response.getheader("location"))
+                return self._request("GET", response.headers.get("location", None))
             if 399 < statuscode <= 500:
                 return {"status": statuscode, "value": data}
             content_type = []
-            if response.getheader("Content-Type"):
-                content_type = response.getheader("Content-Type").split(";")
+            if response.headers.get("Content-Type", None):
+                content_type = response.headers.get("Content-Type", None).split(";")
```
symonk, line 332: need to handle the `None` case here I think, will raise an `AttributeError` trying to split `NoneType` on some cases? edit: its already checked above ignore, no walrus allowed with py3.7+ support unfortunately which makes this kinda stuff much cleaner to code up.

symonk (approved):

symonk:
Thank you @mdmintz

Crocmagnon:
I hope we'll be able to have a release with this fix soon :)

Issue #11301, opened 2022-11-23
[🚀 Feature]: Selenium Manager Error Handling


titusfortner:
### Feature and motivation

All languages need to be pulling info from stderr during selenium manager execution to show failures along the lines of #11300.

### Usage example

N/A

titusfortner:
@nvborisenko can you help with the .NET for this?
@harsha509 can you help with JS?
@symonk can you help with Python?

*should be straightforward what to do, but we can discuss in `#selenium-tlc` if needed.

titusfortner:
Ruby - 148e0d5229511d267e41516e979a307c5137d840

titusfortner:
Did this for JS - 7a79429d159fb2a5b495331ee9fde65bb937b976 because I don't know how error codes are handled. Can update it later if needed.

PR #11329, opened 2022-11-28
[py]: Propagate `stderr` to exceptions when selenium manager fails

symonk:
Capture `stderr` when calling selenium manager and propogate the error to the user.  This propagates `stderr` to the user (and `stdout` for now).  When selenium manager exits `0` we pass on the stripped path returned from its stdout.

adds some unit tests for the manager aswell as we currently have none.  I've made the `SeleniunManager` class require an instance to use as its a single place we use it and I think we may want to build some internal state soon, rather than have a bunch of static methods in a class calling each other from the class directly.

Thanks!

Relates to #11301


symonk:
Did CI change? I'd definitely expect full python tests to run against this diff?

py/selenium/webdriver/common/service.py
```diff
@@ -93,7 +93,7 @@ def start(self) -> None:
                 logger.debug("driver not found in PATH, trying Selenium Manager")
                 browser = self.__class__.__module__.split(".")[-2]
                 try:
-                    path = SeleniumManager.driver_location(browser)
+                    path = SeleniumManager().driver_location(browser)
                 except WebDriverException as new_err:
                     logger.debug("Unable to obtain driver using Selenium Manager: " + new_err.msg)
```
symonk, line 98: I _think_ this is sufficient to propagate here from the changes, should include the message tho unit tests do not currently test an actual `service` based flow; only exercising the manager directly at the moment.

symonk:
`selenium-manager` binary is not `.gitignored` in `/py/*` so i've accidentally committed here, will fix this PR when I have free time after work

titusfortner:
Hah, yes, we need to add them to .gitignore. Sorry & thanks.

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -79,9 +82,10 @@ def run(args: Tuple[str, str, str]) -> str:
         :Returns: The log string containing the driver location.
         """
         logger.debug(f"Executing selenium manager with: {args}")
-        result = subprocess.run(args, stdout=subprocess.PIPE).stdout.decode("utf-8")
-
-        if not re.match("^INFO\t", result):
-            raise WebDriverException(f"Unsuccessful command executed: {args}")
-
-        return result
+        completed_proc = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
titusfortner, line 86: I think we need to also rescue this with something like: 
```
raise SeleniumManagerException(f"Unsuccessful command executed. {args}; {error}")
```

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -79,9 +82,10 @@ def run(args: Tuple[str, str, str]) -> str:
...
+        stdout, stderr = completed_proc.stdout.decode("utf-8"), completed_proc.stderr.decode("utf-8")
+        if completed_proc.returncode:
+            raise SeleniumManagerException(f"Selenium Manager exited non zero.  {stdout}{stderr}")
```
titusfortner: want to keep what command was executed in the error message...

py/selenium/webdriver/common/service.py
titusfortner: yeah, we're going to move to instantiating this class for next set of features, I think.

titusfortner (changes requested):

symonk:
important to squash here

Issue #11239, opened 2022-11-08
[🚀 Feature]: Support for browser beta/canary/dev versions in Selenium Manager


bonigarcia:
### Feature and motivation

Include support for browser version detection, considering not only stable versions, but also beta/canary/dev for Chrome, Firefox, Edge.

### Usage example

If for example Chrome beta is installed in the system, Selenium Manager should be able to detect its versions.

titusfortner:
Is this what we're looking to support with this Issue?

```
selenium-manager --browser chrome --browser-version beta
```

bonigarcia:
My intention for this milestone was only to consider the known commands to discover the browser version, not only for stable browsers, but also for beta/dev/etc. This way, Selenium Manager tries to discover (in order) the version of the stable browser, if not found, then the beta, etc. For M3 (about configuration) we can consider to force the use of beta/dev/etc. In addition, also in M3, we can implement a flag to specify a custom location for the browser (I did something similar in WebDriverManager).

titusfortner:
I agree we need to implement the flag to specify a custom location for the browser.

I was thinking this feature would be a way to ask selenium-manager to auto-locate bea/dev/canary without having to pass in the location directly.

I don't think selenium-manager should use non-production browseer unless a user specifies it.

titusfortner:
Discussion in TLC meeting was not to automatically use a non-prod version of a browser unless explicitly requested. Detection of non-prod browsers will be a nice feature for users to specify instead of specific versions.

bonigarcia:
Completed with #11334.

PR #11279, opened 2022-11-18
[rust] Include support for IEDriverServer in Selenium Manager

bonigarcia:
### Description
This PR implements the support for IEDriverServer (i.e., the driver required to use Internet Explorer) with Selenium Manager.

### Motivation and Context
This feature is part of the development of Selenium Manager M2.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist

- [X] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

PR #11334, opened 2022-11-29
[rust] Support for beta/dev/canary browser version detection with Selenium Manager (#11239)

bonigarcia:
### Description
This PR include support for browser version detection, considering not only stable versions, but also beta/canary/dev for Chrome, Firefox, Edge.

### Motivation and Context
As discussed in TLC, the detection of non-prod browsers will be a feature that needs to be explicitly requested, e.g.:

```
./selenium-manager --browser chrome --browser-version beta
./selenium-manager --browser firefox --browser-version dev
./selenium-manager --browser edge --browser-version canary
```

### Types of changes

### Checklist

### DIFF
--- a/rust/README.md
+++ b/rust/README.md
@@ -22,13 +22,13 @@ Automated driver management for Selenium
 Usage: selenium-manager [OPTIONS]
 Options:
   -b, --browser <BROWSER>
-          Browser name (chrome, firefox, or edge) [default: ]
+          Browser name (chrome, firefox, edge, or iexplorer) [default: ]
   -d, --driver <DRIVER>
-          Driver name (chromedriver, geckodriver, or msedgedriver) [default: ]
+          Driver name (chromedriver, geckodriver, msedgedriver, or IEDriverServer) [default: ]
   -v, --driver-version <DRIVER_VERSION>
           Driver version (e.g., 106.0.5249.61, 0.31.0, etc.) [default: ]
   -B, --browser-version <BROWSER_VERSION>
-          Major browser version (e.g., 105, 106, etc.) [default: ]
+          Major browser version (e.g., 105, 106, etc. Also: beta, dev, canary -or nightly- is accepted) [default: ]
   -D, --debug
           Display DEBUG messages
   -T, --trace
--- a/rust/src/chrome.rs
+++ b/rust/src/chrome.rs
@@ -15,14 +15,20 @@
 // specific language governing permissions and limitations
 // under the License.
 
+use std::collections::HashMap;
 use std::error::Error;
 use std::path::PathBuf;
 
 use crate::downloads::read_content_from_link;
 use crate::files::compose_driver_path_in_cache;
+use crate::is_unstable;
 use crate::manager::ARCH::ARM64;
-use crate::manager::OS::{MACOS, WINDOWS};
-use crate::manager::{detect_browser_version, get_major_version, BrowserManager};
+use crate::manager::OS::{LINUX, MACOS, WINDOWS};
+use crate::manager::{
+    detect_browser_version, format_one_arg, format_two_args, get_major_version, BrowserManager,
+    BrowserPath, BETA, DASH_DASH_VERSION, DEV, ENV_LOCALAPPDATA, ENV_PROGRAM_FILES,
+    ENV_PROGRAM_FILES_X86, NIGHTLY, REG_QUERY, STABLE, WMIC_COMMAND,
+};
 use crate::metadata::{
     create_driver_metadata, get_driver_version_from_metadata, get_metadata, write_metadata,
 };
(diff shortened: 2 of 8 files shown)


titusfortner:
I'm not getting what I expect from this:

```
SL-1495:se-mgr titusfortner$ ./selenium-manager --browser edge --browser-version canary
INFO	/Users/titusfortner/.cache/selenium/msedgedriver/mac64/107.0.1418.62/msedgedriver
SL-1495:se-mgr titusfortner$ ./selenium-manager --browser firefox --browser-version dev
INFO	/Users/titusfortner/.cache/selenium/geckodriver/mac64/0.32.0/geckodriver
SL-1495:se-mgr titusfortner$ ./selenium-manager --browser chrome --browser-version beta
INFO	/Users/titusfortner/.cache/selenium/chromedriver/mac64/108.0.5359.71/chromedriver
SL-1495:se-mgr titusfortner$ ./selenium-manager --browser chrome --browser-version dev
ERROR	Wrong browser/driver version
SL-1495:se-mgr titusfortner$ ./selenium-manager --browser chrome --browser-version canary
INFO	/Users/titusfortner/.cache/selenium/chromedriver/mac64/108.0.5359.71/chromedriver
SL-1495:se-mgr titusfortner$ ./selenium-manager --browser firefox --browser-version nightly
INFO	/Users/titusfortner/.cache/selenium/geckodriver/mac64/0.32.0/geckodriver
SL-1495:se-mgr titusfortner$ ./selenium-manager --browser edge --browser-version beta
INFO	/Users/titusfortner/.cache/selenium/msedgedriver/mac64/107.0.1418.62/msedgedriver
SL-1495:se-mgr titusfortner$ ./selenium-manager --browser edge
INFO	/Users/titusfortner/.cache/selenium/msedgedriver/mac64/107.0.1418.62/msedgedriver
```

bonigarcia:
I'm not sure what you were expecting, but this PR includes extra commands to discover the version of non-prod browsers. A good way to understand a bit better what is happening is running the commands using `--debug`.

titusfortner:
I'm expecting the driver version to be the correct one for the browser version I'm requesting. 
I have Chrome Canary installed on my machine, it is not v108. Is it not supposed to get a compliant driver? Also it does not support dev channel.

titusfortner:
Oh, except I renamed it so manager wouldn't be able to find it, lol. 
Also looks like beta channel hasn't updated to 109. Ok

1. If I'm specifying a version and it doesn't exist, I expect to get an error, not a fall-back.
2. Should support dev channel?
3. This is what I'm getting with Canary in the right place:
```
L-1495:macos titusfortner$ ./selenium-manager --browser chrome --browser-version canary --debug
DEBUG	Using shell command to find out chrome version
DEBUG	Running sh command: "/Applications/Google\\ Chrome\\ Canary.app/Contents/MacOS/Google\\ Chrome\\ Canary --version"
DEBUG	Output { status: ExitStatus(unix_wait_status(0)), stdout: "Google Chrome 110.0.5451.0 canary\n", stderr: "" }
DEBUG	The version of chrome is 110.0.5451.0
DEBUG	Detected browser: chrome 110
DEBUG	Reading chromedriver version from https://chromedriver.storage.googleapis.com/LATEST_RELEASE_110
DEBUG	starting new connection: https://chromedriver.storage.googleapis.com/
DEBUG	No cached session for DnsName(DnsName(DnsName("chromedriver.storage.googleapis.com")))
DEBUG	Not resuming any session
DEBUG	Using ciphersuite TLS13_AES_256_GCM_SHA384
DEBUG	Not resuming
DEBUG	TLS1.3 encrypted extensions: [Protocols([6832])]
DEBUG	ALPN protocol is Some(b"h2")
DEBUG	Ticket saved
DEBUG	Ticket saved
ERROR	Wrong browser/driver version
```
chromedriver storage will never have canary, you need to get the driver from chromium storage: https://chromedriver.chromium.org/chromedriver-canary

bonigarcia:
Regarding 1, in my opinion, Selenium Manager should make the best effort to provide the best driver possible. Throwing an error because the browser detection is failing is not the best choice, IMO, since the `LATEST` driver is likely to work with the underlying browser. In the worst case, the test will fail due to browser-driver incompatibility, but that's the fail I would expect (since the important thing is the users want to driver browsers, not the driver version itself). What we can do to improve this is to write a `WARN` log with Selenium Manager (e.g., _the browser version cannot be detected, trying with the latest driver version_ or similar), and capture this log in the bindings and re-log it to the user.

Regarding 2, I don't know if it is worth it, since driving a canary browser is something very few people will do.

Regarding 3, Selenium Manager is throwing an error since chromedriver 110 (stable) has yet to be released. But to be coherent with 1), I think this case should fallback to `LATEST`, instead of throwing that error. I can implement this for the next release.

diemol:
> Regarding 1, in my opinion, Selenium Manager should make the best effort to provide the best driver possible. Throwing an error because the browser detection is failing is not the best choice, IMO, since the `LATEST` driver is likely to work with the underlying browser. In the worst case, the test will fail due to browser-driver incompatibility, but that's the fail I would expect (since the important thing is the users want to driver browsers, not the driver version itself). What we can do to improve this is to write a `WARN` log with Selenium Manager (e.g., _the browser version cannot be detected, trying with the latest driver version_ or similar), and capture this log in the bindings and re-log it to the user.
> 
> Regarding 2, I don't know if it is worth it, since driving a canary browser is something very few people will do.
> 
> Regarding 3, Selenium Manager is throwing an error since chromedriver 110 (stable) has yet to be released. But to be coherent with 1), I think this case should fallback to `LATEST`, instead of throwing that error. I can implement this for the next release.

I agree on points 1 and 3. But maybe we should look into the Canary thing. I see more people wanting to test with unreleased versions to detect bugs early.

bonigarcia:
> I agree on points 1 and 3. But maybe we should look into the Canary thing. I see more people wanting to test with unreleased versions to detect bugs early.

No problem, I can take a look on how to support the chromedriver canary for the next release.

titusfortner:
> Selenium Manager should make the best effort to provide the best driver possible

Only by default. If a user *specifies something and we can't provide that, it needs to error. This is just like with Capabilities class. If a person thinks they are getting one thing but are actually getting another, you end up with Type II errors, which are a bigger deal.

As far as Canary goes, I don't think it ever gets released, just beta & dev channels. Canary requires using a nightly build. I think we probably want to support beta and dev and not canary right now.

diemol:
Chrome Canary can be downloaded https://www.google.com/chrome/canary/, and IIRC, it updates itself.

diemol:
Another option, to reduce the pain of figuring out which build of the ChromeDriver is needed to work with the Canary browser, we could pass `--disable-build-check` to the latest ChromeDriver. I guess this needs to be done in the bindings?

titusfortner:
Yeah, Canary download is easy. Canary *drivers are hard. Canary browser with Dev channel drivers works, but I'm not sure we want to just do that.

`--disable-build-check` would need to be set by the user in service options, when using the driver.

If Selenium Manager knew it was being set, though, it could do things like:
* Download & use Dev driver when Canary requested (if we don't support canary natively)
* Use closest driver found without updating it. (though this might also not do what the user is thinking)

Might just make more sense to require user to be explicit in the code:

```
DriverService ds = new ChromeDriverService.Builder()
  .withBrowserVersion("canary")
  .withDriverVersion("dev")
  .build();
Chromedriver driver = new ChromeDriver(ds, new ChromeOptions());
```
or
```
ChromeOptions() options = new ChromeOptions() ;
options.addArgument("--disable-build-check");

DriverService ds = new ChromeDriverService.Builder()
  .withDriverVersion("104")
  .build();
Chromedriver driver = new ChromeDriver(ds, options);
```

titusfortner:
I take it back, Selenium Manager should throw the error if user provides incompatible versions *unless `--disable-build-check` is passed. I think that makes the implementation more consistent as well?

bonigarcia:
If we decide to error when browser detection does not find the browser version (e.g., `beta`), it is not difficult to do it in Rust. Maybe it can be decided in the next TLC meeting, together with the rest of issues that needs a decision.

In any case, to progress in the development, I would merge this PR (and #11371, which has been created over this one).

titusfortner (changes requested):
When I execute this line:
```
./selenium-manager --browser chrome --browser-version dev
```

If dev doesn't exist on the machine, it returns the location of prod driver.
If dev does exist on the machine, it errors because there isn't a v110 driver released.

In the first case it should error because the browser version doesn't exist.

In the second case, I actually think it should include a warning and return v109 driver. That would mean specifically checking to see if we get `No such object: chromedriver/LATEST_RELEASE_110`.

diemol:
Why does the PR have changes that mention IEDriver too? Wasn't that already merged?

bonigarcia:
> Why does the PR have changes that mention IEDriver too? Wasn't that already merged?

The changes related to IEDriverServer is just in the README, since I forgot to include it in #11279 (and the current PR updates a bit the README).

diemol:
> > Why does the PR have changes that mention IEDriver too? Wasn't that already merged?
> 
> The changes related to IEDriverServer is just in the README, since I forgot to include it in #11279 (and the current PR updates a bit the README).

But this is making more changes than the README https://github.com/SeleniumHQ/selenium/pull/11334/files#diff-2166be5cf60b67e7f0945ccac14fbe86ae137fe7fb71914482031de897dc70f3

diemol:
> When I execute this line:
> 
> ```
> ./selenium-manager --browser chrome --browser-version dev
> ```
> 
> If dev doesn't exist on the machine, it returns the location of prod driver. If dev does exist on the machine, it errors because there isn't a v110 driver released.
> 
> In the first case it should error because the browser version doesn't exist.
> 
> In the second case, I actually think it should include a warning and return v109 driver. That would mean specifically checking to see if we get `No such object: chromedriver/LATEST_RELEASE_110`.

@bonigarcia this makes sense. So, either we work on supporting `dev` properly or we take it out of this PR and do it in a future PR.

bonigarcia:
> In the second case, I actually think it should include a warning and return v109 driver. That would mean specifically checking to see if we get `No such object: chromedriver/LATEST_RELEASE_110`.

Yes, it is a good idea.

To simplify the development on my side, I would prefer to merge this PR as it is now and #11371 (which has been created over this one). Then, I will make the new changes over #11371 (since it includes a relevant refactoring in the Rust code).

diemol:
OK, then let's merge this and follow up in the next one.

diemol (approved):

titusfortner:
Can we open a new issue to track what needs to be done?

bonigarcia:
> Can we open a new issue to track what needs to be done?

I have just created #11382, #11383, and #11384.

PR #11368, opened 2022-12-05
[ruby][bidi] Add LogInspector

TamsilAmani:
### Description

Add LogInspector contains the w3c Bidi spec related APIs to listen to log events.

### Motivation and Context

Refer to [BiDi Spec Doc](https://docs.google.com/document/d/1dCd8Y2PYaR5mOGSmNTwllEHNmFqegfoGkP-TCKvPzSU/edit?usp=sharing) about the details.

### Types of changes

### Checklist


rb/lib/selenium/webdriver/bidi/log.rb
```diff
@@ -0,0 +1,101 @@
...
+            end
+          end
+        end
+
+
+        private
+
+        def on(event, &block)
+          event = EVENTS[event] if event.is_a?(Symbol)
+          @bidi.callbacks["log.#{event}"] << block
```
TamsilAmani: @p0deje As you can see I am able to listen to the `log.entryAdded` event even though I am not subscribing it anywhere. My tests are passing. Any suggestions?

TamsilAmani: **UPDATE** : I was testing on chrome initially. But after changing to firefox, `subscribe` and `unsubscribe` are working as expected.

p0deje: Maybe Chrome always sends the messages by default?

rb/lib/selenium/webdriver/bidi/log_inspector.rb
```diff
@@ -0,0 +1,140 @@
...
+          end
+        end
+
+        def on_javascript_exception(&block)
+          enabled = log_listeners[:js_exception].any?
+          log_listeners[:js_exception] << block
+          log_listeners[:javascript] << block
+          return if enabled
+
+          on(:entry_added) do |params|
```
p0deje: I suppose we can use `on_log` here as well:

```suggestion
          on_log do |params|
```

rb/lib/selenium/webdriver/bidi/log_inspector.rb
```diff
@@ -0,0 +1,140 @@
...
+            __send__(:console_log_events, params) if type.eql?("console")
+          end
+        end
+
+        def on_javascript_log(&block)
+          enabled = log_listeners[:javascript].any?
+          log_listeners[:javascript] << block
+          return if enabled
+
+          on(:entry_added) do |params|
```
p0deje: I suppose we can use `on_log` here as well:

```suggestion
          on_log do |params|
```

rb/lib/selenium/webdriver/bidi/log_inspector.rb
```diff
@@ -0,0 +1,140 @@
...
+          @bidi = driver.bidi
+          @bidi.session.subscribe("log.entryAdded", browsing_context_ids)
+        end
+
+        def on_console_log(&block)
+          enabled = log_listeners[:console].any?
+          log_listeners[:console] << block
+          return if enabled
+
+          on(:entry_added) do |params|
```
p0deje: I suppose we can use `on_log` here as well:

```suggestion
          on_log do |params|
```

rb/lib/selenium/webdriver/bidi/log_inspector.rb
```diff
@@ -0,0 +1,140 @@
...
+            type = params["type"]
+            __send__(:console_log_events, params) if type.eql?("console")
```
p0deje: There is no need for `__send__` here:

```suggestion
            console_log_events(params) if type.eql?("console")
```

rb/lib/selenium/webdriver/bidi/log_inspector.rb
```diff
@@ -0,0 +1,140 @@
...
+        end
+
+        def on_javascript_log(&block)
+          enabled = log_listeners[:javascript].any?
+          log_listeners[:javascript] << block
+          return if enabled
+
+          on(:entry_added) do |params|
+            type = params["type"]
+            __send__(:javascript_log_events, params) if type.eql?("javascript")
```
p0deje: There is no need for `__send__` here:

```suggestion
            javascript_log_events(params) if type.eql?("console")
```

rb/lib/selenium/webdriver/bidi/log_inspector.rb
```diff
@@ -0,0 +1,140 @@
...
+          enabled = log_listeners[:js_exception].any?
+          log_listeners[:js_exception] << block
+          log_listeners[:javascript] << block
+          return if enabled
+
+          on(:entry_added) do |params|
+            type = params["type"]
+            level = params["level"]
+
+            __send__(:javascript_log_events, params) if type.eql?("javascript") && level.eql?(LOG_LEVEL[:ERROR])
```
p0deje: There is no need for `__send__` here:

```suggestion
            javascript_log_events(params) if type.eql?("console")
```

rb/lib/selenium/webdriver/bidi/log_inspector.rb
```diff
@@ -0,0 +1,140 @@
...
+            text: params['text'],
+            timestamp: params['timestamp'],
+            type: params['type'],
+            stack_trace: params['stackTrace']
+          )
+          log_listeners[:javascript].each do |listener|
+            listener.call(event)
+          end
+
+          return unless params['level'].eql?(LOG_LEVEL[:ERROR])
```
p0deje, line 130: This whole method is only called if `level.eql?(LOG_LEVEL[:ERROR])` (line 83) so you don't need to check it once again.

rb/lib/selenium/webdriver/bidi/session.rb
```diff
@@ -32,6 +32,28 @@ def status
           Status.new(status['ready'], status['message'])
         end
 
+        def subscribe(events, browsing_contexts = nil)
+          events_list = events.instance_of?(Array) ? events : [events]
```
p0deje: You can write it in a more compact way:

```suggestion
          events_list = Array(events)
```

rb/lib/selenium/webdriver/bidi/session.rb
```diff
@@ -32,6 +32,28 @@ def status
...
+                                     nil
+                                   else
+                                     browsing_contexts.instance_of?(Array) ? browsing_contexts : [browsing_contexts]
+                                   end
+
+          @bidi.send_cmd("session.subscribe", events: events_list, contexts: browsing_contexts_list)
+        end
+
+        def unsubscribe(events, browsing_contexts = nil)
+          events_list = events.instance_of?(Array) ? events : [events]
```
p0deje: You can write it in a more compact way:

```suggestion
          events_list = Array(events)
```

rb/lib/selenium/webdriver/bidi/session.rb
```diff
@@ -32,6 +32,28 @@ def status
           Status.new(status['ready'], status['message'])
         end
 
+        def subscribe(events, browsing_contexts = nil)
+          events_list = events.instance_of?(Array) ? events : [events]
+          browsing_contexts_list = if browsing_contexts.nil?
+                                     nil
+                                   else
+                                     browsing_contexts.instance_of?(Array) ? browsing_contexts : [browsing_contexts]
```
p0deje: You can write it in a more compact way:

```suggestion
          Array(browsing_contexts)
```

rb/lib/selenium/webdriver/bidi/session.rb
```diff
@@ -32,6 +32,28 @@ def status
...
+
+          @bidi.send_cmd("session.subscribe", events: events_list, contexts: browsing_contexts_list)
+        end
+
+        def unsubscribe(events, browsing_contexts = nil)
+          events_list = events.instance_of?(Array) ? events : [events]
+          browsing_contexts_list = if browsing_contexts.nil?
+                                     nil
+                                   else
+                                     browsing_contexts.instance_of?(Array) ? browsing_contexts : [browsing_contexts]
```
p0deje: You can write it in a more compact way:

```suggestion
          Array(browsing_contexts)
```

rb/spec/integration/selenium/webdriver/bidi/log_inspector_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+require_relative '../spec_helper'
+
+module Selenium
+  module WebDriver
+    class BiDi
+      describe LogInspector, exclusive: {browser: %i[chrome firefox]} do
```
p0deje, line 25: Doesn't `edge` also support it?

rb/spec/integration/selenium/webdriver/bidi/log_inspector_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+
+        it 'can listen to console log' do
+          reset_driver!(web_socket_url: true) do |driver|
+            log_entry = nil
+            log_inspector = LogInspector.new(driver)
+            log_inspector.on_console_log { |log| log_entry = log }
+
+            driver.navigate.to url_for(@page)
+            driver.find_element(id: 'consoleLog').click
+            sleep 0.5
```
p0deje: Can we instead wait until `log_entry` is not `nil`?

rb/spec/integration/selenium/webdriver/bidi/log_inspector_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+
+        it 'can listen to javascript log' do
+          reset_driver!(web_socket_url: true) do |driver|
+            log_entry = nil
+            log_inspector = LogInspector.new(driver)
+            log_inspector.on_javascript_log { |log| log_entry = log }
+
+            driver.navigate.to url_for(@page)
+            driver.find_element(id: 'jsException').click
+            sleep 0.5
```
p0deje: Can we instead wait until `log_entry` is not `nil`?

rb/spec/integration/selenium/webdriver/bidi/log_inspector_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+
+        it 'can listen to javascript error log' do
+          reset_driver!(web_socket_url: true) do |driver|
+            log_entry = nil
+            log_inspector = LogInspector.new(driver)
+            log_inspector.on_javascript_exception { |log| log_entry = log }
+
+            driver.navigate.to url_for(@page)
+            driver.find_element(id: 'jsException').click
+            sleep 0.5
```
p0deje: Can we instead wait until `log_entry` is not `nil`?

rb/spec/integration/selenium/webdriver/bidi/log_inspector_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+
+        it 'can retrieve stack trace for a log' do
+          reset_driver!(web_socket_url: true) do |driver|
+            log_entry = nil
+            log_inspector = LogInspector.new(driver)
+            log_inspector.on_javascript_log { |log| log_entry = log }
+
+            driver.navigate.to url_for(@page)
+            driver.find_element(id: 'jsException').click
+            sleep 0.5
```
p0deje: Can we instead wait until `log_entry` is not `nil`?

p0deje (approved):
LGTM - left few small suggestions.

rb/lib/selenium/webdriver/bidi/log_inspector.rb
TamsilAmani: The method `javascript_log_events()` is called two times. One in line **83** (when LOG_LEVEL[ERROR] is **true**) and another in line **69** (when LOG_LEVEL[ERROR] **need not be true**).

My question is, will a "`javascript`" log always be an error? If yes, then your suggestion is valid. But if not, then I think my logic will work. I tried to see how to produce a javascript log which is not an error but couldn't find anything.

People suggested `console.error` but that is a console log with level=error, not a javascript log.

rb/spec/integration/selenium/webdriver/bidi/log_inspector_spec.rb
TamsilAmani: Chrome and Edge don't fully support it. You can see the [WPT Tests](https://wpt.fyi/results/webdriver/tests/bidi/log/entry_added?label=master&label=experimental&aligned&view=subtest) for the browsers. In fact my tests are passing on chrome (when they shouldn't have) but firefox behaved as expected.

If you suggest I can remove chrome too from the spec.

rb/lib/selenium/webdriver/bidi/log_inspector.rb
p0deje: I suppose the javascript error is supposed to be an exception?

```javascript
throw new Error("hello!")
```

If I understand the intention correctly, then:

* `on_javascript_log` should track all console log entries (log/error)
* `on_javascript_exception` should track only exceptions

rb/spec/integration/selenium/webdriver/bidi/log_inspector_spec.rb
titusfortner: It's less intuitive than I want, but was the best I could come up with for guards:

* `except` - bug is present on specified configurations (runs non-specified configurations and fails if they pass)
* `only` -  bug is present on non-specified configurations (runs specified configurations and fails if they pass)
* `exclusive` - feature only applies to specified configurations
* `exclude` - feature is valid for specified configurations, but bug is present on it and it causes test suite to crash

The way I use these, if it is a DriverExtension and the methods are only imported to certain driver classes, use `exclusive` on top level `describe`, then `except`/`only` on sub-levels (`exclude` if it breaks something when using the others)

This feature is not an extension, the method is available on all drivers. So use `only` with firefox, and add reason that only firefox has implemented the feature so far.

rb/lib/selenium/webdriver/bidi/log_inspector.rb
TamsilAmani: Yes,
* `on_javascript_log` tracks all "javascript" log entries, whether log or error. (This is different from `on_console_log` which tracks all console logs like `console.log` and `console.error`)
* `on_javascript_exception` tracks all "javascript" logs which are only error like `throw new Error("hello!")`

p0deje: Sorry, but what are "javascript" log entries and how they are different from "console" log entries?

TamsilAmani: Here is the definition of both https://w3c.github.io/webdriver-bidi/#types-log-logentry.

The `type` property for console is `"console"` while for javascript it is `"javascript".`

Though for testing purpose I'm using `throw new Error("hello!")` for both `on_javascript_log` and `on_javascript_exception` as I don't know how to produce a javascript log with level != error

titusfortner: https://github.com/w3c/webdriver-bidi/pull/73#issuecomment-737491240
Looks like it is "the same for now" but I'm not sure where they are going with that...

TamsilAmani: Though for testing purpose I'm using `throw new Error("hello!")` for both `on_javascript_log` and `on_javascript_exception` as I don't know how to produce a javascript log with level != error

titusfortner: here - https://github.com/w3c/webdriver-bidi/pull/73#discussion_r542283994

titusfortner: looks like it is just for warnings and errors? Though I have no idea how to generate either. Are they showing up both in console log and JS log? Presumably should only be in one or the other?

TamsilAmani: @titusfortner 
* `throw new Error("hello!")` will generate Javascript Log:
        - type = "javascript"
        - level = "error"

* `console.log("hello")` will generate Console log:
        - type = "console"
        - level = "info"

* `console.error("hello")` will generate Console log
        - type = "console"
        - level = "error"
        
That's the difference I can see from the values of `type` and `level`. I have no knowledge how console error is different from javascript error.

titusfortner: The answer to the original question, imo is that we should assume JS log is always an error until someone complains that we're missing something. :-D

TamsilAmani: Alright. So if we go by that, I suppose we remove `on_javascript_log()` and only keep `on_javascript_exception()`

AutomatedTester: I think `on_javascript_log` should return everything that isn't an error and then have `on_javascript_error` return the errors. My worry is having everything seen as an exception is going to cause problems later on. The mixing of the two feels like we're heading down the same mistake of conflating `getProperty` and `getAttribute`. I think it's better to have the methods here

titusfortner: As I'm reading the spec, there are no Javascript logs that aren't errors. But this doesn't really matter to me.

AutomatedTester: after speaking to @pujagani I think I understand the issue better here.

My concern long term is we have a client API that conflates things and that will lead to problems in the client API further down the line with long winded deprecate -> remove steps.

We can do what is in the PR (except I have a suggestion which I will put near the bottom of this comment) and have `on_javascript_log` and if needed a nice convenience method for just the errors as I think that is what people really care about. This could have a "filter" to allow people to pick what they want.

I think we should also see about moving `on_console_log` to `on_console_msg` with a way for users to get a specific type of "filter" so people can get `error` only if they wanted. @pujagani points out that having the filter on both would be a similar pattern.

TamsilAmani: After discussing with David, here's what he agrees to:

* `on_javascript_exception` - to be kept as it. It will filter JS error logs.

* `on_javascript_log(filter = nil)` - it will filter all JS logs based on filter parameter (info/error/etc). If no filter is provided, we will give all JS logs and also show user a warning that it can be very large amount of data.

* `on_console_event(filter = nil)` - it will filter all 'console' logs based on the filter parameter. If no filter is provided, we will give all console logs and also show user a warning that it can be very large amount of data.

@titusfortner @p0deje any opinions on this?

p0deje: I still don't quite understand what is the difference between JavaScript and Console logs and whether the public interface should be that verbose. I understand that the API is dictated by the spec, but I suppose from the user perspective there are two things I would like to do:

1. Catch and log all JavaScript errors (exceptions).
2. Catch and log all Console messages.

Do I miss something? If no, maybe we should stick to just these two messages and let people leverage the language for filtering and storing events as they want:

```ruby
errors = []

# gather runtime JS errors
driver.on_javascript_exception do |event|
  errors << event
end

# gather console.error()
driver.on_console_event do |event|
  next unless event.error?
  errors << error
end
```

AutomatedTester: After speaking to @jgraham only  exceptions have been handled at the moment but `on_javascript_event` will do more than exceptions in the future. See https://matrix.to/#/!zvtoFTlOBeOwHrBlEK:matrix.org/$qB6_aZnVrIFu5taUXw6-Gy-YVo8gKHLfFyrnjwG0nEY?via=matrix.org&via=mozilla.org&via=igalia.com

TamsilAmani: Pointing to [his comment](https://matrix.to/#/!zvtoFTlOBeOwHrBlEK:matrix.org/$Dh6dQ0hhaho1B5yZ5P7ncw9Y_vigFFGrd7tugk-hB20?via=matrix.org&via=mozilla.org&via=igalia.com) in the same thread as the one @AutomatedTester  mentioned above, @jgraham is of the opinion that JS logs can have more levels later apart from `error` currently. 

So IMO, we should not limit JS log capturing to only `exception` else a future change will require much complexities to be addressed in case other levels for JS Logs also land.

p0deje: Okay then, I suppose we just follow what is designed in https://docs.google.com/document/d/1dCd8Y2PYaR5mOGSmNTwllEHNmFqegfoGkP-TCKvPzSU/edit#heading=h.jw2vsnry15j7, including both `on_javascript_log` and `on_javascript_exception`.

p0deje (approved):

rb/lib/selenium/webdriver/bidi/log_inspector.rb
TamsilAmani: Very well. I think we have a consent now. I will mark this PR as resolved and as discussed with @pujagani will add the `filter` parameter to the functions in another PR, just to keep things clean.

PR #11369, opened 2022-12-05
[java] Merge capabilities of slot with the new session request capabilities

pujagani:

### Description

Merge capabilities of slot with the new session request capabilities

### Motivation and Context

Node config allows the Grid user to set driver configuration. 
Example: `java -jar selenium-server-4.6.0.jar standalone --port 50859 --detect-drivers false --driver-configuration display-name="Chrome Custom" stereotype='{"browserName":"chrome","goog:chromeOptions":{"args":[\"incognito\",\"window-size=500,500\"]},"pageLoadStrategy":"normal"}'`

The configuration is applied to the session slots.
This intends to provide the configuration that all matching incoming session requests should ideally consider and pass it along to the respective web driver.
Currently, the new session request capabilities are passed as it is without considering if the session slot has additional capabilities that need to be considered.

The changes help fix this by giving merging the capabilities of the slot and the incoming session request. The capabilities of the incoming session request override the capabilities of the slot case of custom driver-configuration.

### Types of changes

### Checklist

### DIFF
--- a/py/selenium/webdriver/common/actions/action_builder.py
+++ b/py/selenium/webdriver/common/actions/action_builder.py
@@ -94,5 +94,5 @@ class ActionBuilder:
         """Clears actions that are already stored on the remote end."""
         self.driver.execute(Command.W3C_CLEAR_ACTIONS)
 
-    def _add_input(self, input) -> None:
-        self.devices.append(input)
+    def _add_input(self, new_input) -> None:
+        self.devices.append(new_input)
--- a/py/selenium/webdriver/support/expected_conditions.py
+++ b/py/selenium/webdriver/support/expected_conditions.py
@@ -431,6 +431,9 @@ def new_window_is_opened(current_handles):
 
 
 def alert_is_present():
+    """An expectation for checking if an alert is currently present and
+    switching to it."""
+
     def _predicate(driver):
         try:
             return driver.switch_to.alert
--- a/rust/Cargo.Bazel.lock
+++ b/rust/Cargo.Bazel.lock
@@ -1,5 +1,5 @@
 {
-  "checksum": "17a3d42d3eebdf25a24ac0e2ec7215f03a1f6c281abe8c491d0026247b1e6782",
+  "checksum": "cfb8c431183818c041b63dcfc880980393ff487769338055bc00a99b75ae4ffa",
   "crates": {
     "adler 1.0.2": {
       "name": "adler",
@@ -137,6 +137,115 @@
       },
       "license": "Unlicense OR MIT"
     },
+    "anyhow 1.0.68": {
+      "name": "anyhow",
+      "version": "1.0.68",
+      "repository": {
+        "Http": {
+          "url": "https://crates.io/api/v1/crates/anyhow/1.0.68/download",
+          "sha256": "2cb2f989d18dd141ab8ae82f64d1a8cdd37e0840f73a406896cf5e99502fab61"
+        }
+      },
+      "targets": [
+        {
+          "Library": {
+            "crate_name": "anyhow",
+            "crate_root": "src/lib.rs",
+            "srcs": {
+              "include": [
+                "**/*.rs"
+              ],
+              "exclude": []
+            }
+          }
+        },
+        {
+          "BuildScript": {
+            "crate_name": "build_script_build",
+            "crate_root": "build.rs",
+            "srcs": {
+              "include": [
+                "**/*.rs"
+              ],
+              "exclude": []
+            }
+          }
+        }
+      ],
+      "library_target_name": "anyhow",
+      "common_attrs": {
+        "compile_data_glob": [
+          "**"
+        ],
+        "crate_features": [
+          "default",
+          "std"
+        ],
+        "deps": {
+          "common": [
+            {
+              "id": "anyhow 1.0.68",
+              "target": "build_script_build"
+            }
+          ],
+          "selects": {}
+        },
+        "edition": "2018",
+        "version": "1.0.68"
+      },
+      "build_script_attrs": {
+        "data_glob": [
+          "**"
+        ]
+      },
+      "license": "MIT OR Apache-2.0"
+    },
+    "assert-json-diff 2.0.2": {
+      "name": "assert-json-diff",
+      "version": "2.0.2",
+      "repository": {
+        "Http": {
+          "url": "https://crates.io/api/v1/crates/assert-json-diff/2.0.2/download",
+          "sha256": "47e4f2b81832e72834d7518d8487a0396a28cc408186a2e8854c0f98011faf12"
+        }
+      },
+      "targets": [
+        {
+          "Library": {
+            "crate_name": "assert_json_diff",
+            "crate_root": "src/lib.rs",
+            "srcs": {
+              "include": [
+                "**/*.rs"
+              ],
+              "exclude": []
+            }
+          }
+        }
+      ],
+      "library_target_name": "assert_json_diff",
+      "common_attrs": {
+        "compile_data_glob": [
+          "**"
+        ],
+        "deps": {
+          "common": [
+            {
+              "id": "serde 1.0.152",
+              "target": "serde"
+            },
+            {
+              "id": "serde_json 1.0.91",
+              "target": "serde_json"
+            }
+          ],
+          "selects": {}
+        },
+        "edition": "2018",
+        "version": "2.0.2"
+      },
+      "license": "MIT"
+    },
     "assert_cmd 2.0.8": {
       "name": "assert_cmd",
       "version": "2.0.8",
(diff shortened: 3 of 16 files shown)


diemol:
We do something similar on the [`LocalNode`](https://github.com/SeleniumHQ/selenium/blob/6a8bf1ca7ed2c22eb736abb2ebb5cb58bc5d87bd/java/src/org/openqa/selenium/grid/node/local/LocalNode.java#L538-L539). Does this PR improve that code? Should we centralize that logic and do it in a single place?

pujagani:
Thank you for helping take a look. The [LocalNode](https://github.com/SeleniumHQ/selenium/blob/6a8bf1ca7ed2c22eb736abb2ebb5cb58bc5d87bd/java/src/org/openqa/selenium/grid/node/local/LocalNode.java#L538-L539) logic, is for sending the result back to the user. The PR does something similar but to send to the driver. Mainly, if the sessionSlot is configured with custom capabilities, the PR merges that with incoming session capabilities and sends it to the driver factory.  My thought process was more around the session slot being the owner of the stereotype, the logic should sit there. LocaNode is more of a meditator and is aware of the session slots.  What do you think?

java/src/org/openqa/selenium/grid/node/local/SessionSlot.java
```diff
@@ -144,7 +144,16 @@ public Either<WebDriverException, ActiveSession> apply(CreateSessionRequest sess
     }
 
     try {
-      Either<WebDriverException, ActiveSession> possibleSession = factory.apply(sessionRequest);
+      Capabilities mergedCapabilities =
+        sessionRequest.getDesiredCapabilities().merge(this.getStereotype());
```
diemol: Would this overwrite the browser options passed by the user? Like, if the custom stereotype has a `goog:chromeOptions`, would it overwrite what the user has passed?

pujagani: I think so. The capabilities of the slot override the capabilities of the incoming session request in case of custom driver-configuration.

diemol: I think that is not ideal, because we would not honor the preferences set by the user. If we want to do this, probable this needs a deep merge of browser specific options. What do you think?

pujagani: So we are trying to honour the preference settings on the Grid. My thinking was that if the Grid slots have preferences, they are meant to be universally applied to all incoming sessions that match, so ideally, they would take precedence.

In case of a deeper merge, what do you suggest we do differently that the current merge function?

diemol: Right, but if a user wants to run headless, and the stereotype has a setting to set the browser language. Both things would live inside `goog:chromeOptions`. I think having the stereotype ones take precedence is unexpected and would break the behavior users currently have in Grid. 

The current merge logic replaces the value at the top level, if we want to do this, we should follow the same logic for browser preferences, like `goog:chromeOptions`, `moz:firefoxOptions`, etc...

pujagani: Ahh okay, I think I see your point. Let me see how we can achieve that or what it will take to make it possible.

pujagani:
@diemol After giving your comments a thought, it made sense to move the logic to SessionCapabilitiesMutator since that is where we make some checks on the capabilities and add some capabilities as required. Additionally, as you correctly pointed out, I think it makes sense to give the user-sent capabilities priority over the stereotype capability to avoid any breaking changes. I have updated the changes accordingly and reverted the previous ones. Will appreciate your review. Thank you so much.

java/src/org/openqa/selenium/grid/node/config/SessionCapabilitiesMutator.java
```diff
@@ -56,33 +57,137 @@ public Capabilities apply(Capabilities capabilities) {
...
+      Map<String, Object> capsOptions = (Map<String, Object>) capabilities.asMap().get(options);
+
+      // Merge top level capabilities, excluding browser specific options.
+      // This will overwrite the browser options too, but it does not matter since we tackle it separately just after this.
+      Map<String, Object> toReturn = new HashMap<>(slotStereotype.merge(capabilities).asMap());
+
+      // Merge browser specific stereotype and capabilities options
+      switch (browserName) {
+        case "chrome":
+        case "microsoftedge":
```
diemol, line 78: In some places in the Java code I see we consider `MicrosoftEdge` and `msedge`, maybe consider it here as well?

java/src/org/openqa/selenium/grid/node/config/SessionCapabilitiesMutator.java
```diff
@@ -56,33 +57,137 @@ public Capabilities apply(Capabilities capabilities) {
...
+        String rawProfile =
+          (String) capsOptions.get("profile");
+
+        toReturn.put("profile", rawProfile);
+      }
+
+      if (name.equals("binary")) {
+        Object binary = capsOptions.get("binary");
+        toReturn.put("binary", binary);
+      }
```
diemol: Same comment as the Chromium case.

java/src/org/openqa/selenium/grid/node/config/SessionCapabilitiesMutator.java
```diff
@@ -56,33 +57,137 @@ public Capabilities apply(Capabilities capabilities) {
...
+
+      if (name.equals("log")) {
+        Map<String, Object> logLevelMap =
+          (Map<String, Object>) capsOptions.get("log");
+        toReturn.put("log", logLevelMap);
+      }
+    }
```
diemol: Same as the `binary` case.

diemol (changes requested):
Thank you, I left a few comments.

java/src/org/openqa/selenium/grid/node/config/SessionCapabilitiesMutator.java
pujagani: I am afraid I don't understand. Which Chromium case is been referred to here?

java/src/org/openqa/selenium/grid/node/config/SessionCapabilitiesMutator.java
```diff
@@ -56,33 +57,137 @@ public Capabilities apply(Capabilities capabilities) {
...
+          }
+        });
+
+        toReturn.put("extensions", stereotypeExtensions);
+      }
+
+      if (name.equals("binary")) {
+        Object binary = capsOptions.get("binary");
+        toReturn.put("binary", binary);
+      }
```
diemol: Somehow this comment was lost...

If I understand the code correctly, this would overwrite the `binary` location in the stereotype (if it exists), which is probably not ideal because the person who setup the Grid knows the browser paths.

diemol (commented):
Added missing comment for the Chromium case.

java/src/org/openqa/selenium/grid/node/config/SessionCapabilitiesMutator.java
pujagani: Ahh okay, I understand now. Thank you for helping clarify this. I will fix it.

java/src/org/openqa/selenium/grid/node/config/SessionCapabilitiesMutator.java
pujagani: This is log-level property. For this, we should consider the user preference here right?

java/src/org/openqa/selenium/grid/node/config/SessionCapabilitiesMutator.java
```diff
@@ -126,9 +127,11 @@ private Map<String, Object> mergeChromiumOptions(Map<String, Object> stereotypeO
...
       }
 
-      if (name.equals("binary")) {
-        Object binary = capsOptions.get("binary");
-        toReturn.put("binary", binary);
+      // Grid will know the browser paths so preference is given to stereotype value in this case
+      // If the value is not present in stereotype but present in capabilities only then use capabilities value
+      if (name.equals("binary") && stereotypeOptions.get("binary") == null) {
+          Object binary = capsOptions.get("binary");
+          toReturn.put("binary", binary);
```
diemol: I was thinking about this again and I think we should completely ignore `binary` when it comes in the session request capabilities. Ideally, the person who setup the Grid already has left all paths configured. 

This could also affect cloud vendors, probably BrowserStack sets the binary path where the browser is, IIRC Sauce does something like that. If a user sets the binary in a place where the browser is not present on the remote machine, session creation will fail.

java/src/org/openqa/selenium/grid/node/config/SessionCapabilitiesMutator.java
diemol: Ah OK. Log level should be fine.

diemol (changes requested):

java/src/org/openqa/selenium/grid/node/config/SessionCapabilitiesMutator.java
pujagani: I will make the changes. I see your point. Grid should be the source of truth in such cases.

diemol (approved):
Thank you, @pujagani!

Issue #9917, opened 2021-10-14
4.0.0 release is missing sdist tarball on pypi


mgorny:
## 🐛 Bug Report

It seems that the 4.0.0 release on pypi includes only wheels but missed the .tar.gz archive needed to build from source.

## To Reproduce

Visit https://pypi.org/project/selenium/#files

## Expected behavior

.tar.gz archive being present

## Test script or set of commands reproducing this issue

n/a

## Environment

OS: Gentoo Linux
Browser: n/a
Browser version: n/a
Browser Driver version: n/a
Language Bindings version: Python (any)
Selenium Grid version (if applicable): n/a

AutomatedTester:
I will need to look into this with Bazel.

I honestly thought everyone had moved to wheels for distribution of packages these days. What is the main use case for you for needing the sdist version?

mgorny:
Distributions still build their packages from sources.

chenrui333:
Yes, Homebrew would need the pypi artifact to build the formula from sources.

chenrui333:
Any update on this thread? Thanks!

bollwyvl:
> everyone had moved to wheels

Can confirm this is also blocking on conda-forge, and is indeed still documented in `py/README.rst`:

> Alternately, you can download the source distribution from PyPI (e.g. selenium-4.1.0.tar.gz), unarchive it, and run:

In the meantime, I'm investigating a full `bazel` build... but humorously, invoking `bazel build //py:selenium` naively fails because `pytest-trio`... doesn't distribute a `whl` :rofl: 

If the contents of `py` really _are_ enough to build a distribution, that would get us going, and will be investigating further, but a PyPI sdist would indeed be preferred!

pombredanne:
gentle ping: everyone expects to have source distributions available. Wheels are great, and wanted but not enough.

apteryks:
For the record, the only two files need to run the test suite which are not in a git checkout are these:

- selenium/webdriver/remote/getAttribute.js
- selenium/webdriver/remote/isDisplayed.js

I have no idea how they come into existence (I don't use or know about Bazel, and it's been too messy for anyone to successfully package yet on GNU Guix).

symonk:
bump; we should release a source distribution on pypi; I'm not sure where those `*.js` files are being pulled in from; but editable installs etc are a no-go atm for me locally

pombredanne:
> For the record, the only two files need to run the test suite which are not in a git checkout are these:
[...]

PR #11375, opened 2022-12-06
[py] fix sdist publish instructions

jameshilliard:
It appears that only wheels are being published to pypi, this should fix the publishing procedure to include sdists.

### Description
Update publish procedure to include sdists.

### Motivation and Context
Sdists on pypi are missing as reported in #9917

### Types of changes

### Checklist

- [N/A] I have added tests to cover my changes.
- [N/A] All new and existing tests passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,8 +195,8 @@ pip install bazel-bin/py/selenium-*.whl
 To publish run:
 
 ```sh
-bazel build //py:selenium-wheel
-twine upload bazel-bin/py/selenium-*.whl
+bazel build //py:selenium-wheel //py:selenium-sdist
+twine upload bazel-bin/py/selenium-*.whl bazel-bin/py/selenium-*.tar.gz
 ```
 </details>
 


titusfortner:
As described in the issue you referenced, our issue is that Bazel is not properly creating the sdist in the first place. We can't publish what isn't getting built. We probably should pull the "publishing" instructions from the README entirely because that really isn't relevant to users.

jameshilliard:
> As described in the issue you referenced, our issue is that Bazel is not properly creating the sdist in the first place. We can't publish what isn't getting built.

I hit one bug which should be fixed by [#647](https://github.com/bazelbuild/rules_pkg/pull/647), is that the issue you are referring to?

AutomatedTester:
The issue with publishing is that pypi is expecting a `PKG_INFO` which we're not generating. I haven't had time to do it but we could probably copy the implementation from `py_wheel` and create a new rule that maybe we can upstream to `rules_python`

titusfortner:
@jameshilliard thanks for digging into this!

jameshilliard:
> The issue with publishing is that pypi is expecting a `PKG_INFO` which we're not generating.

Should be fixed in #11379.

titusfortner:
I'll merge this now, thanks!

It's already in our releasing documents, which we manage here: https://github.com/SeleniumHQ/selenium/wiki/Releasing-Selenium

Issue #11372, opened 2022-12-05
[🚀 Feature]: Selenium Manager get Browser Version from Options classes


titusfortner:
### Feature and motivation

~Java already supports passing in Options to Service class for Chrome to get logging level — 
We'll want to expand on that for all service classes, and then implement it in each of the other languages.~

We decided not to expand on what was currently being done in Java and to implement a Driver Finder class that will consume the browser options instance and update the service instance.

In Progress:
Ruby - #11523 
Java - #11491

diemol:
JS -> https://github.com/SeleniumHQ/selenium/commit/2c1bd8eb164a80e382616951bf56bede57b576b2

diemol:
Implemented in all bindings, closing.

Issue #11360, opened 2022-12-03
[🚀 Feature]: Selenium Manager Bindings API


titusfortner:
### Feature and motivation

As I see it, we have 3 basic routes we can go for using Selenium Manager going forward.

1. Add relevant methods to Service classes
2. Create instance of SeleniumManager and pass it as argument into Service class
3. Create instance of SeleniumManager and pass it as argument into Driver class

My opinions
1. This seems like the easiest thing for users to work with. They may already be using these classes and can now add to them. They don't need to know how it's being done; they are requesting it, Selenium is doing it. Easy to do this with Service class builder pattern, but the constructor(s) will get weird each time we add a new parameter.
2. If we want to put all these concerns in one place, have the user explicitly opt-in to "Selenium Manager," or just highlight better the Selenium Manager feature, this might be good. Builder pattern of Service class would make this a straightforward add, and only one additional constructor required.
3. The reason to pick this one is because Selenium Manager at some point will be downloading browsers as well as drivers (or the grid?). If that's the case do the concerns actually all fit in the "Service" class? The huge down side of this is that adding all the constructors required will be messy.

Examples below are Java with its builder class, but the principle is the same in all languages. 

## Option 1 — Add everything directly in the service class

```java
DriverService ds = new ChromeDriverService.Builder()
  .withBrowserVersion("103")
  .withDriverVersion("105.0.5195.19")
  .withBrowserTTL(Duration.ofHours(3))
  .withClearManagerCache()
  .build();
Chromedriver driver = new ChromeDriver(ds, new ChromeOptions());
```

## Option 2 — Selenium Manager instance into service class

```java
SeleniumManager sm = new SeleniumManager.Builder();
  .withBrowserVersion("103")
  .withDriverVersion("105.0.5195.19")
  .withBrowserTTL(Duration.ofHours(3))
  .withClearManagerCache()
  .build();

DriverService ds = new ChromeDriverService.Builder()
  .withManager(sm)
  .build();
Chromedriver driver = new ChromeDriver(ds, new ChromeOptions());
```

## Option 3 — Selenium Manager instance into driver class

```java
SeleniumManager sm = new SeleniumManager.Builder();
  .withBrowserVersion("103")
  .withDriverVersion("105.0.5195.19")
  .withBrowserTTL(Duration.ofHours(3))
  .withClearManagerCache()
  .build();

Chromedriver driver = new ChromeDriver(new ChromeOptions(), sm);
```

[...]

Issue #11353, opened 2022-12-02
[🚀 Feature]: Selenium Manager - allow to ignore build matching


titusfortner:
### Feature and motivation

User should be able to specify any version of browser and driver they want, and if the driver would otherwise complain, they need to specify to ignore the complaint.

This should include the ability to set separate driver and browser versions, and therefore allow setting the driver version with just the major version number.

This will require bindings to parse the `goog:chromeOptions` capability for an arg with `disable-build-check`.

### Usage example

```
selenium-manager --browser chrome --browser-version 105 --driver-version 108 --disable-build-check
```
or
```
selenium-manager --browser chrome --browser-version canary --driver-version dev --disable-build-check
```

I *don't think you need to check `--disable-build-check` if a user does something that the driver itself will allow:
```
selenium-manager --browser chrome --browser-version 108 --driver-version 107
```

diemol:
I am not sure if Selenium Manager needs to be aware of `--disable-build-check`. It should probably be just the bindings, like if the browser and browser version are different, add the `--disable-build-check` flag. However, this flexibility opens the door for user errors, which indeed are user created but they could end up in issue reports (given that users try all different combinations). 

Also, this is Chromium only.

titusfortner:
Scenario:
```
DriverService ds = new ChromeDriverService.Builder()
  .withBrowserVersion("106")
  .withDriverVersion("103")
  .build();
Chromedriver driver = new ChromeDriver(ds, new ChromeOptions());
```
Bindings will translate to:
```
selenium-manager --browser chrome --browser-version 106 --driver-version 103
```

If Selenium Manager just returns the v103 of the chromedriver, then the driver will error when the code tries to use it.

I think it makes sense for Selenium Manager to be the one that throws the error where we can give a better message about what the user needs to do (use a matching version or add `--disable-build-check` to the ChromeOptions).

And yes, Chrome & Edge only.

titusfortner:
Ok, new scenario...

Maybe we don't let the user specify driver version in code, and user puts an older version of driver on PATH that they want to use. Normally we'd throw a warning saying your driver version is old, we're giving you the updated one...
If a user passes `--disable-build-check` then we throw the warning but still use it?

diemol:
Makes sense. I think letting users specify the driver version is not ideal because it opens the door to many unsupported use cases. If they want to have a special environment, they can set it up by putting it on the PATH.

titusfortner:
[...]

Issue #11351, opened 2022-12-02
[🚀 Feature]: Selenium Manager use binary from Browser Options


titusfortner:
### Feature and motivation

Evaluate the browser version at a specific location in order to know which driver to get. I think it would make sense to support this with `--browser` instead of creating a new flag. 

From a bindings implementation standpoint we'll want to inspect the `binary` key in the `goog:chromeOptions` capability to get this and pass it to the selenium-manager binary.

### Usage example

```
selenium-manager --browser "/Applications/Google Chrome Dev.app/Contents/MacOS/"
```
and
```
selenium-manager --browser "/Applications/Google Chrome Dev.app/Contents/MacOS/Google Chrome Dev"
```

bonigarcia:
I would not reuse the flag `--browser` for the browser path, since `--browser` is supposed to be the browser name (`chrome`, `edge`, etc.). Otherwise you need to infer the browser name from that path. I would create another flag, such as `--browser-path`.

titusfortner:
You're right. Keep things more obvious

bonigarcia:
This is already implemented in the Rust side (#11496). Now the bindings can start using it. 

Is `C-java` supposed to be also in the labels?

diemol:
Implemented in all bindings. Closing.

PR #11376, opened 2022-12-06
[java] WIP - proof of concept for using options with Selenium Manager

titusfortner:
### Description

I think this is the best way to get needed information from Options class to Selenium Manager.

* Ensure driver constructor uses `createServiceWithConfig` by default.
* new private `withOptions()` method that stores protected static options instance on DriverService
* Browser DriverService class can pull browser-specific functionality (`getBinary()`); though, I guess we *could implement this in all driver service classes and essentially avoid all subclasses for now.
* Couple things get cleaned up when we implement this in all browsers, but I wanted to show the idea with one to get feedback.

**Questions**:
* Does storing a protected static options instance make the most sense?
* Should we create `getBinary()` in all options classes and avoid subclassing requirements? (fwiw, Firefox's current implementation might complicate using Selenium Manager; Chrome was much easier to show the idea)
* Is there anything we can think of that we might be doing with Selenium Manager in the future that this implementation would complicate?

### Motivation

* Provide support for #11372 
* Provide support for #11351 when it is available
* It might be relevant to how we do #11353 

Previously, I was thinking about the need to specifically wrap `--driver-version`, which is why I brought up #11360 
As of now, though, I think everything that we need to be supporting in Selenium Manager we can pull from the Options class, and this is the best way I came up with for doing that.

### DIFF
--- a/java/src/org/openqa/selenium/manager/SeleniumManager.java
+++ b/java/src/org/openqa/selenium/manager/SeleniumManager.java
@@ -16,7 +16,6 @@
 // under the License.
 package org.openqa.selenium.manager;
 
-import com.google.common.collect.ImmutableList;
 import com.google.common.io.CharStreams;
 import org.openqa.selenium.Beta;
 import org.openqa.selenium.Platform;
@@ -29,7 +28,9 @@ import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.Arrays;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
 import java.util.logging.Logger;
 
 import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
@@ -87,7 +88,7 @@ public class SeleniumManager {
    * @param command the file and arguments to execute.
    * @return the standard output of the execution.
    */
-    private static String runCommand(String... command) {
+    private static String runCommand(List<String> command) {
         String output = "";
         int code = 0;
         try {
@@ -99,14 +100,14 @@ public class SeleniumManager {
                     process.getInputStream(), StandardCharsets.UTF_8));
         } catch (InterruptedException e) {
             LOG.warning(String.format("Interrupted exception running command %s: %s",
-                    Arrays.toString(command), e.getMessage()));
+                    command, e.getMessage()));
             Thread.currentThread().interrupt();
         } catch (Exception e) {
             LOG.warning(String.format("%s running command %s: %s",
-                    e.getClass().getSimpleName(), Arrays.toString(command), e.getMessage()));
+                    e.getClass().getSimpleName(), command, e.getMessage()));
         }
         if (code > 0) {
-          throw new WebDriverException("Unsuccessful command executed: " + Arrays.toString(command) +
+          throw new WebDriverException("Unsuccessful command executed: " + command +
                                         "\n" + output);
         }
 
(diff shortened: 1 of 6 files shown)


titusfortner:
(This is to address #11372)

diemol:
I am OK when the PR template is ignored, but I think we should always provide the context of a PR (unless the code is extremely explicit). I understand why this PR was created and how the context was build through Slack discussions, but leaving this PR without a context will make future maintainers life very hard. This is why we have so much code in Java today that is doing magical stuff and we do not understand why. 

As for the contents of the PR, they seem fine to me, but I do not recall all the details of the conversation.

titusfortner:
This isn't production-ready code, so I moved it to draft status and added context with the specific questions I'm interested in getting feedback on. Thanks.

titusfortner:
Some issues... 

I think we need to deprecate either `public static ChromeDriverService createDefaultService()` or  `public ChromeDriver(ChromeDriverService service, ChromeOptions options)` (make it private).

Since this code won't work as I have it implemented:
```
ChromeDriverService service = ChromeDriverService.createDefaultService();
options.setBrowserVersion("100");
options.setBinary("/path/to/old/chrome");

ChromeDriver driver = new ChromeDriver(service, options);
```

java/src/org/openqa/selenium/manager/SeleniumManager.java
```diff
@@ -99,14 +100,14 @@ private static String runCommand(String... command) {
                     process.getInputStream(), StandardCharsets.UTF_8));
         } catch (InterruptedException e) {
             LOG.warning(String.format("Interrupted exception running command %s: %s",
-                    Arrays.toString(command), e.getMessage()));
+                    command, e.getMessage()));
```
shs96c, line 103: This expansion won't be the same, so people won't be able to copy-and-paste it.

shs96c (changes requested):
The coupling here is wrong: to integrate the SeleniumManager you shouldn't need to make changes to the DriverService.

Roles and responsibilities:
* SeleniumManager: finding the driver and browser
* DriverService: managing the lifetime of the driver and browser

The DriverService doesn't care _how_ the driver and browser are found, just that it has them.

titusfortner:
> The DriverService doesn't care how the driver and browser are found

Except that is exactly what it is currently concerning itself with. If the location isn't passed in by the user (very rarely the case), it has logic for locating it. 

We're currently putting the constraint on the Manager to only be used if the existing process would otherwise error. 

We can:
1. Change that constraint and have the manager always execute and pass in the driver value to the DriverService
2. Change the existing DriverService class to insert logic between the finding and erroring steps (this implementation)
3. Separate out the Finding logic of existing DriverService code and insert Selenium Manager steps
4. Something else I'm not seeing?

@shs96c which of these are you suggesting ^^

titusfortner:
Ok, I spent a lot more time inside the DriverService classes trying to figure out a better approach... We can definitely pull out the finding pieces for better separation of concerns.

I implemented a `DriverFinder` class, with a `setDriverPath(options)` method, that will set the driver path property that will then get used by `DriverService`.  I hacked a way to get options to execute it, without the DriverService ever having to touch options. But there's no good way to ensure that the user didn't build their own Service before creating the options.

I think the Service class has to get passed in the options instance (`withOptions(options)`), but it can call out to the `DriverFinder` without needing to change anything in DriverService class directly.

User will need to have done one of these:
1. Not specified their own Service instance
2. Set the System Property before building their own Service instance
3. Built `DriverService` and used `usingDriverExecutable()`
4. Constructed `DriverService` with hard-coded path to driver
5. Explicitly called the new `withOptions()` to use SeleniumManager

So, we throw a warning if `build()` is called and none of those things have happened.

That is probably a better approach, but it's a lot to put together... 
Let me try to polish it up again after having thought through it once more....

titusfortner:
ok, I used my previous comment to rethink things and improve on where I was going with it.

I'm pretty sure #11411 is the superior implementation, so I'm going to close this.

PR #11396, opened 2022-12-09
[java] Ensure options presents in capabilities are merged as expected

pujagani:

### Description

Ensure options present in capabilities are merged as expected

### Motivation and Context

```
  MutableCapabilities browserCaps = new MutableCapabilities();
    browserCaps.setCapability("binary", "/path/to/browser");
    MutableCapabilities addCaps = new MutableCapabilities();
    addCaps.setCapability("goog:chromeOptions", browserCaps);
    ChromeOptions options = existingOptions.merge(addCaps);

```
The above example is a valid use-case that was not accounted for earlier. The changes address the same and fix it.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/chromium/ChromiumOptions.java
+++ b/java/src/org/openqa/selenium/chromium/ChromiumOptions.java
@@ -280,7 +280,46 @@ public class ChromiumOptions<T extends ChromiumOptions<?>> extends AbstractDrive
   protected void mergeInPlace(Capabilities capabilities) {
     Require.nonNull("Capabilities to merge", capabilities);
 
-    capabilities.getCapabilityNames().forEach(name -> setCapability(name, capabilities.getCapability(name)));
+    capabilities.getCapabilityNames().forEach(name -> {
+      if (!name.equals("binary") && !name.equals("extensions") && !name.equals("args")) {
+        setCapability(name, capabilities.getCapability(name));
+        return;
+      }
+
+      if (name.equals("args") && capabilities.getCapability(name) != null) {
+        List<String> arguments = (List<String>) (capabilities.getCapability(("args")));
+        arguments.forEach(arg -> {
+          if (!args.contains(arg)) {
+            addArguments(arg);
+          }
+        });
+        return;
+      }
+
+      if (name.equals("extensions") && capabilities.getCapability(name) != null) {
+        List<Object> extensionList = (List<Object>) (capabilities.getCapability(("extensions")));
+        extensionList.forEach(extension -> {
+          if (!extensions.contains(extension)) {
+            if (extension instanceof File) {
+              addExtensions((File) extension);
+            } else if (extension instanceof String) {
+              addEncodedExtensions((String) extension);
+            }
+          }
+        });
+        return;
+      }
+
+      if (name.equals("binary") && capabilities.getCapability(name) != null) {
+        Object binary = capabilities.getCapability("binary");
+        if (binary instanceof String) {
+          setBinary((String) binary);
+        } else if (binary instanceof File) {
+          setBinary((File) binary);
+        }
+      }
+    });
+
     if (capabilities instanceof ChromiumOptions) {
       ChromiumOptions<?> options = (ChromiumOptions<?>) capabilities;
       for (String arg : options.args) {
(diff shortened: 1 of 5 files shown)


java/src/org/openqa/selenium/chromium/ChromiumOptions.java
```diff
@@ -274,7 +274,46 @@ protected Object getExtraCapability(String capabilityName) {
   protected void mergeInPlace(Capabilities capabilities) {
     Require.nonNull("Capabilities to merge", capabilities);
 
-    capabilities.getCapabilityNames().forEach(name -> setCapability(name, capabilities.getCapability(name)));
+    capabilities.getCapabilityNames().forEach(name -> {
```
krmahadevan: Maybe we can use the traditional `for..each` because this lambda that we are passing into `forEach()` is multi lined. The conventional `for..each` helps with readability.

java/test/org/openqa/selenium/chrome/ChromeOptionsTest.java
```diff
@@ -264,6 +265,67 @@ void mergingOptionsWithMutableCapabilities() {
...
+  void mergingOptionsWithOptionsAsMutableCapabilities() {
+    File ext1 = TestUtilities.createTmpFile("ext1");
+    String ext1Encoded = Base64.getEncoder().encodeToString("ext1".getBytes());
+    String ext2 = Base64.getEncoder().encodeToString("ext2".getBytes());
+    ArrayList<Object> extensions = new ArrayList<>();
+    extensions.add(ext1);
+    extensions.add(ext2);
+
+    ArrayList<Object> args = new ArrayList<>();
+    args.add("silent");
```
krmahadevan: Maybe we can use `Arrays.asList()`

java/src/org/openqa/selenium/firefox/FirefoxOptions.java
```diff
@@ -372,8 +372,62 @@ public FirefoxOptions merge(Capabilities capabilities) {
     Require.nonNull("Capabilities to merge", capabilities);
     FirefoxOptions newInstance = new FirefoxOptions();
     getCapabilityNames().forEach(name -> newInstance.setCapability(name, getCapability(name)));
-    capabilities.getCapabilityNames().forEach(name -> newInstance.setCapability(name, capabilities.getCapability(name)));
     newInstance.mirror(this);
+    capabilities.getCapabilityNames().forEach(name -> {
+      if (name.equals("args") && capabilities.getCapability(name) != null) {
```
krmahadevan: If we use `Optional.ofNullable(capabilities.getCapability(name)).orElse()` (for all the if conditions) we can remove the null check.

java/test/org/openqa/selenium/chrome/ChromeOptionsTest.java
```diff
@@ -264,6 +265,67 @@ void mergingOptionsWithMutableCapabilities() {
       .isEqualTo(binary.getPath());
   }
 
+  @Test
+  void mergingOptionsWithOptionsAsMutableCapabilities() {
+    File ext1 = TestUtilities.createTmpFile("ext1");
+    String ext1Encoded = Base64.getEncoder().encodeToString("ext1".getBytes());
+    String ext2 = Base64.getEncoder().encodeToString("ext2".getBytes());
+    ArrayList<Object> extensions = new ArrayList<>();
```
krmahadevan: Here also we can use `Arrays.asList()`

java/src/org/openqa/selenium/firefox/FirefoxOptions.java
pujagani: I understand your reasoning but even if that case for some values (depending on the values, I did go over the code again to check), it will have to return a null object or a null string and that might need a check. Keeping it simple in this case should be fine.

java/src/org/openqa/selenium/chromium/ChromiumOptions.java
pujagani: I appreciate the concern. However, even after making changes, I did not see an improvement since the loop logic will be in a code block in the next line and the defining the loop remains in the line above. So I think it is okay to keep it as it is.

krmahadevan: I was not suggesting the use of a traditional for loop to improve performance, but its a convention that is followed when dealing with lamdas (atleast that's what I have learnt from the sessions that I have watched run by Venkat Subramanian) that lamdas should be a single line, if they can't be, then they perhaps are not lambdas.

pujagani: I understand, I was talking about readability itself. I don't see any improvement in how it looks but maybe I am missing something. This is something new I learned today. Thank you!

PR #11403, opened 2022-12-10
[java] remove deprecated Firefox capabilities and consolidate tests

titusfortner:
These 3 capabilities have been throwing the non-w3c deprecation warning since we added it, so they should be ok to remove now

This cleans the code where they was used and removes the relevant tests.
While I was updating tests, I also moved all relevant tests from MarionetteTest to FirefoxTest; there was a lot of duplication.


diemol (approved):
This is great, thank you @titusfortner!

I hope I checked everything correctly, I believe all removed methods were either not public or the super class will take over.

I'll wait for CI to finish and then I will merge.

PR #11404, opened 2022-12-10
[java] allow Edge to set driver log level as well

titusfortner:
Chrome allows setting log level in Options class, and consumes it in the Service class.

I want to provide the same functionality to Edge. This PR just deals with moving the things to the super class so that Edge has access to them.

This PR does not wire in the functionality in the Edge Service class. Trying to do this in smaller, easier to parse chunks. Is this the right way to do this?

Marking it as Draft since it can't be used as-is. I have other changes in mind for Service classes if this is ok.


diemol (commented):
So far the changes look good. I can check again when it is ready for review.

titusfortner:
Well, actually, I probably should back up and see if we even want this since it means things behave differently between local and remote.

titusfortner:
We decided in TLC meeting we do not want to do this, we want to do the opposite of this.

PR #11409, opened 2022-12-10
[java] change chromium service names from whitelisted to allowed

titusfortner:
In Chrome v87 Google changed the name of the parameter in the documentation from `whitelisted-ips` to `allowed-ips`, but both are still supported under the hood. 

EdgeDriverService was already using "allowed" naming convention. This PR changes the ChromeDriverService implementation to match it. The two questions I have:

1. Do we want to change what we're sending to the driver if it means that it won't be recognized by this Chrome < v87?
2. How do you communicate the name change of a System property? Marking the public constant as deprecated, doesn't seem very useful. Do we want to raise a warning?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,7 +101,6 @@ GPATH
 GRTAGS
 GSYMS
 GTAGS
-Gemfile.lock
 /rb/.bundle
 /rb/.idea
 .sonar/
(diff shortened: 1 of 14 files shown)


diemol (approved):
Thank you, @titusfortner!

PR #11411, opened 2022-12-11
Use Driver finder for sending Options to Selenium Manager

titusfortner:
This is an alternative implementation to #11376 for using Options class to work with SeleniumManager.

Instead of storing options instance in the `DriverService` class and using it in the `findExecutable()` method, this PR moves the driver finding pieces out of service classes. There are currently 2 methods, one with options parameter that can call SeleniumManager and one without options. The goal is to get everything to go through that method, and the trick is going to be figuring out how to make that happen (I think it can be done with a combination of deprecations and warning messages).

I think this one can be merged as-is, but at the least it needs more javadocs and someone(s) to make sure I've thought everything through.


java/src/org/openqa/selenium/safari/SafariDriverService.java
```diff
@@ -97,6 +103,10 @@ public int score(Capabilities capabilities) {
       return score;
     }
 
+    public Builder withOptions(SafariOptions options) {
```
pujagani, line 106: Just curious, I observed the options are passed but not used. Is that intentional?

java/src/org/openqa/selenium/safari/SafariTechPreviewDriverService.java
```diff
@@ -98,6 +104,10 @@ public int score(Capabilities capabilities) {
       return score;
     }
 
+    public Builder withOptions(SafariOptions options) {
+      return this;
```
pujagani, line 108: Same observation here

java/src/org/openqa/selenium/manager/SeleniumManager.java
```diff
@@ -147,20 +148,15 @@ private File getBinary() {
...
+   * @param options Browser Options instance.
    * @return the location of the driver.
    */
-    public String getDriverPath(String driverName) {
-        if (!ImmutableList.of("geckodriver", "chromedriver", "msedgedriver", "IEDriverServer").contains(driverName)) {
-            throw new WebDriverException("Unable to locate driver with name: " + driverName);
-        }
-
-        String driverPath = null;
+    public String getDriverPath(Capabilities options) {
```
pujagani, line 154: The public method's signature is being changed. Should there be a deprecate instead and a new method? I might be missing the context/usage here.

pujagani: Nevermind, figured out the internal usage.

pujagani:
LGTM

java/src/org/openqa/selenium/safari/SafariDriverService.java
titusfortner: To make it consistent with the others. It's just to avoid telling everyone they need to use the `withOptions` method if they want to start a service and then say "except here". I'm fine to remove it if we want to go this route.

java/src/org/openqa/selenium/chrome/ChromeDriverService.java
```diff
@@ -137,29 +137,43 @@ public ChromeDriverService(
...
   public static ChromeDriverService createDefaultService() {
-    return new Builder().build();
+    return createDefaultService(new ChromeOptions());
   }
 
   /**
    * Configures and returns a new {@link ChromeDriverService} using the supplied configuration. In
    * this configuration, the service will use the ChromeDriver executable identified by the
-   * {@link #CHROME_DRIVER_EXE_PROPERTY} system property. Each service created by this method will
+   * {@link ChromeDriverFinder} system property. Each service created by this method will
```
diemol, line 152: Not sure if that is correct? There is no `ChromeDriverFinder` system property.

java/src/org/openqa/selenium/chrome/ChromeDriverFinder.java
```diff
@@ -0,0 +1,25 @@
...
+
+  protected static File findExecutable(ChromeOptions options) {
+    return findExecutable(
+      "chromedriver",
+      CHROME_DRIVER_EXE_PROPERTY,
+      "https://chromedriver.chromium.org/",
+      "https://chromedriver.chromium.org/downloads",
+      options);
+  }
+}
```
diemol, line 25: I am not following why we need this class. Why don't we call `DriverFinder.findExecutable` in `ChromeDriver`? We could put all hard coded values in variables, make the call from there and avoid this extra class. 

This comment also applies for the rest of the drivers.

diemol (changes requested):
I am not sure if we need the `*DriverFinder.java` classes. I think we can call `DriverFinder.findExecutable` from the driver class. 

Nevertheless, I think we do not want to follow this approach which implies sending options to the driver service? So far #11491 looks better to me.

titusfortner:
We decided to go with #11491

PR #11434, opened 2022-12-15
Upgrade Ruby Capabilities

titusfortner:
**UPDATE** - I committed all the basic refactoring things and all the obvious bits we want to do. What is left is the various deprecations and logic changes. Updated the contents below accordingly.

-----
I've been avoiding this one for a while because I feel like we (I) keep making Ruby capabilities code harder.
Each commit is pretty self contained with explanations; if there is anything I'm doing in here that we definitely don't like, let me know.

I'm going to write a blog post explaining all of this stuff before it gets released. Also planning to test it with Watir and Capybara to see how much work this would create for those projects.

## Motivation
Selenium Manager is going to need Options information in/around the Service class to get the right browser/driver going forward. So this PR improves the inheritance a bit and differentiates logic allowed for local and remote drivers so the logic for getting the required info is much easier.
Plus, this brings the Ruby implementation in line with all the other bindings.

## Contents

1. `Options#add_option` is sugar that I've never loved because this is what accessors are for, and we support everything in constructor as well. We could do validation here instead of deprecation, but I prefer to keep `add_option` to be just for vendor name-spaced options
2. Require all options in constructor to be snake cased symbols in the `CAPABILITIES` list for the respective browser. Any unrecognized capability is deprecated
3. Deprecate allowing `:capabilities` argument for local drivers. It is still allowed in Remote drivers
4. Deprecate browser class methods for Capabilities (e.g., `Remote::Capabilities.chrome`) just like Java removed `DesiredCapabilities.chrome()` in 4.0 to encourage peole to switch to Options. Using `Remote::Capabilities.new(browser_name: 'chrome')` is also an easy alternative if you don't want to switch to Options.

### DIFF
--- a/rb/lib/selenium/webdriver/common/driver.rb
+++ b/rb/lib/selenium/webdriver/common/driver.rb
@@ -314,19 +314,19 @@ module Selenium
         end
       end
 
-      def process_options(options, capabilities, klass = nil)
+      def process_options(options, capabilities)
         if options && capabilities
           msg = "Don't use both :options and :capabilities when initializing #{self.class}, prefer :options"
           raise ArgumentError, msg
         end
 
-        options ? options.as_json : deprecate_capabilities(capabilities, klass)
+        options ? options.as_json : deprecate_capabilities(capabilities)
       end
 
-      def deprecate_capabilities(capabilities, klass)
-        if klass
+      def deprecate_capabilities(capabilities)
+        unless is_a?(Remote::Driver)
           WebDriver.logger.deprecate("The :capabilities parameter for #{self.class}",
-                                     ":options argument with an instance of #{klass}",
+                                     ":options argument with an instance of #{self.class}",
                                      id: :capabilities)
         end
         generate_capabilities(capabilities)
--- a/rb/lib/selenium/webdriver/common/driver.rb
+++ b/rb/lib/selenium/webdriver/common/driver.rb
@@ -314,13 +314,22 @@ module Selenium
         end
       end
 
-      def process_options(options, capabilities)
+      def process_options(options, capabilities, klass = nil)
         if options && capabilities
           msg = "Don't use both :options and :capabilities when initializing #{self.class}, prefer :options"
           raise ArgumentError, msg
         end
 
-        options ? options.as_json : generate_capabilities(capabilities)
+        options ? options.as_json : deprecate_capabilities(capabilities, klass)
+      end
+
+      def deprecate_capabilities(capabilities, klass)
+        if klass
+          WebDriver.logger.deprecate("The :capabilities parameter for #{self.class}",
+                                     ":options argument with an instance of #{klass}",
+                                     id: :capabilities)
+        end
+        generate_capabilities(capabilities)
       end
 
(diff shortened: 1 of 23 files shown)


rb/lib/selenium/webdriver/common/driver.rb
```diff
@@ -314,13 +314,22 @@ def create_bridge(caps:, url:, http_client: nil)
         end
       end
 
-      def process_options(options, capabilities)
+      def process_options(options, capabilities, klass = nil)
```
p0deje: I suppose you don't have explicitly pass this and you can try to deprecate based on `self.class` being a kind of `Chrome::Driver`, `Firefox::Driver`, etc.

rb/lib/selenium/webdriver/remote/capabilities.rb
```diff
@@ -83,38 +83,46 @@ def platform=(value)
...
           def firefox(opts = {})
+            WebDriver.logger.deprecate('Remote::Capabilities.firefox', 'Options.firefox', id: :caps_browsers)
             new({
               browser_name: 'firefox'
             }.merge(opts))
           end
           alias ff firefox
 
           def safari(opts = {})
+            WebDriver.logger.deprecate('Remote::Capabilities.safari', 'Options.safari', id: :caps_browsers)
```
p0deje, line 109: Looks like we have a loading bug in `Options::Safari` caching the first value of browser name:

```ruby
[1] pry(main)> Selenium::WebDriver::Options.safari
=> #<Selenium::WebDriver::Safari::Options:0x0000000109d8b3a0 @options={:browser_name=>"safari"}>
[2] pry(main)> Selenium::WebDriver::Safari.technology_preview!
=> true
[3] pry(main)> Selenium::WebDriver::Options.safari
=> #<Selenium::WebDriver::Safari::Options:0x0000000109d3a6a8 @options={:browser_name=>"safari"}>
```

It might be minor but it doesn't happen in `Capabilities.safari`.

p0deje (approved):
Left a couple of minor comments, but other than that - looks good to me.

rb/lib/selenium/webdriver/common/driver.rb
titusfortner: Hmm, ok, yeah I'm using it to differentiate whether it is a local or remote driver, and I can do that more easily since we aren't subclassed any more.

rb/lib/selenium/webdriver/remote/capabilities.rb
titusfortner: Yeah, because of that, in `Options::Safari#as_json` I do this:
```
@options[:browser_name] = Safari.technology_preview? ? 'Safari Technology Preview' : 'safari'
```

titusfortner: We changed browser name to use `BROWSER_NAME` constant at class load, which means it won't change when `#technology_preview!` is set. :-/ 
Found it when running tests over break, and fixed here: https://github.com/SeleniumHQ/selenium/commit/04d571231191e516480ae8acd1340878210e1861

Issue #11406, opened 2022-12-10
[🐛 Bug]: ERROR	error sending request for url (https://chromedriver.storage.googleapis.com/LATEST_RELEASE_108): error trying to connect: invalid peer certificate contents: invalid peer certificate: UnknownIssuer


vicozohili:
### What happened?

Dec 10, 2022 1:30:28 PM org.openqa.selenium.remote.service.DriverService findExecutable
WARNING: Unable to obtain driver using Selenium Manager: Unsuccessful command executed: [C:\Users\<user>\AppData\Local\Temp\selenium-manager17283176393723006992415082342136659\selenium-manager.exe, --driver, chromedriver]
WARN	Sending fatal alert BadCertificate
ERROR	error sending request for url (https://chromedriver.storage.googleapis.com/LATEST_RELEASE_108): error trying to connect: invalid peer certificate contents: invalid peer certificate: UnknownIssuer

Build info: version: '4.7.1', revision: 'c6795baf1a3'
System info: os.name: 'Windows 10', os.arch: 'x86', os.version: '10.0', java.version: '1.8.0_341'
Driver info: driver.version: unknown

### How can we reproduce the issue?

```shell
Using Selenium Manager 4.7.1 to launch chrome v108
```

### Relevant log output

```shell
Dec 10, 2022 1:30:28 PM org.openqa.selenium.remote.service.DriverService findExecutable
WARNING: Unable to obtain driver using Selenium Manager: Unsuccessful command executed: [C:\Users\<user>\AppData\Local\Temp\selenium-manager17283176393723006992415082342136659\selenium-manager.exe, --driver, chromedriver]
WARN	Sending fatal alert BadCertificate
ERROR	error sending request for url (https://chromedriver.storage.googleapis.com/LATEST_RELEASE_108): error trying to connect: invalid peer certificate contents: invalid peer certificate: UnknownIssuer

Build info: version: '4.7.1', revision: 'c6795baf1a3'
System info: os.name: 'Windows 10', os.arch: 'x86', os.version: '10.0', java.version: '1.8.0_341'
Driver info: driver.version: unknown
```

### Operating System

Windows

### Selenium version

4.7.1

### What are the browser(s) and version(s) where you see this issue?

chrome 108

### What are the browser driver(s) and version(s) where you see this issue?

NA

### Are you using Selenium Grid?

4.7.1

vicozohili:
[...]

PR #11449, opened 2022-12-20
[rust] Reuse common http client in Selenium Manager

bonigarcia:
### Description
This PR creates a shared instance of the HTTP client used in the Rust logic of Selenium Manager, using `rusttls` instead of native TLS (to prevent #11291), and accepting self-signed certificates (to prevent #11406).

### Motivation and Context
This PR will prevent issues like #11406 in the upcoming releases of Selenium Manager. Moreover, the shared HTTP client will be helpful in Selenium Manager M3 (e.g., for proxy setup).

### Types of changes

### Checklist

### DIFF
--- a/rust/src/chrome.rs
+++ b/rust/src/chrome.rs
@@ -170,6 +170,9 @@ impl SeleniumManager for ChromeManager {
                     } else {
                         format!("{}{}_{}", DRIVER_URL, LATEST_RELEASE, browser_version_int)
                     };
+                    if !browser_version.is_empty() && browser_version_int <= 0 {
+                        break;
+                    }
                     log::debug!("Reading {} version from {}", &self.driver_name, driver_url);
                     let content = read_content_from_link(self.get_http_client(), driver_url);
                     match content {
--- a/rust/src/chrome.rs
+++ b/rust/src/chrome.rs
@@ -16,6 +16,7 @@
 // under the License.
 
 use crate::config::ManagerConfig;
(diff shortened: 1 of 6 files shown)


rust/src/chrome.rs
```diff
@@ -156,31 +164,29 @@ impl SeleniumManager for ChromeManager {
             _ => {
                 let mut driver_version = "".to_string();
                 let mut browser_version_int = browser_version.parse::<i32>().unwrap_or_default();
-                if browser_version_int > 0 {
```
diemol, line 159: Why is this `if` removed?

diemol (commented):
Just a comment.

rust/src/chrome.rs
bonigarcia: That condition was actually a bug, since when `browser_version` is empty (which is a pertinent value, as checked in line 161) when the browser version cannot be discovered, and in that case, the _latest_ driver version is used as a fallback. I discover it after merging the PR about the browser path, since that PR contained a new test that revealed the issue.

bonigarcia: I have just committed a new patch to consider the original motivation of that condition, i.e., to avoid checking negative driver versions (although it is basically an error case that happens when a user forces a very low Chrome version, like `2`).

diemol: 👍

diemol (approved):

Issue #11383, opened 2022-12-08
[🚀 Feature]: Selenium-Manager falls back to the latest driver version if a given version still needs to be released


bonigarcia:
### Feature and motivation

Continuation of #11334. When a given driver version is unavailable in its repository (e.g., chromedriver `110`), Selenium Manager should fall back to the `latest` version.

### Usage example

```
./selenium-manager --browser chrome --browser-version dev
```

... and `dev` exists on the machine, but the required driver version (e.g., chromedriver `110`) has not been released yet.

titusfortner:
Just to be clear.

Assume versions:
Production: v100
Beta: v101
Dev: v102
Canary: v102

When Dev exists on the machine, but chromedriver v102 hasn't been released yet through regular channel, then Selenium Manager should get v101.

bonigarcia:
Completed with #11454.

PR #11454, opened 2022-12-20
[rust] Fallback for chromedriver versions in Selenium Manager (#11383)

bonigarcia:
### Description
This PR implements a fallback mechanism (max 5 retries) when reading the versions of chromedriver. For instance:

```
> cargo run -- --browser chrome --browser-version dev
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
     Running `target\debug\selenium-manager.exe --browser chrome --browser-version dev`
WARN    Error getting version of chromedriver 110. Retrying with chromedriver 109 (attempt 1/5)
INFO    C:\Users\boni\.cache\selenium\chromedriver\win32\109.0.5414.25\chromedriver.exe
```

```
> cargo run -- --browser chrome --browser-version canary
    Finished dev [unoptimized + debuginfo] target(s) in 0.20s
     Running `target\debug\selenium-manager.exe --browser chrome --browser-version canary`
WARN    Error getting version of chromedriver 111. Retrying with chromedriver 110 (attempt 1/5)
WARN    Error getting version of chromedriver 110. Retrying with chromedriver 109 (attempt 2/5)
INFO    C:\Users\boni\.cache\selenium\chromedriver\win32\109.0.5414.25\chromedriver.exe
```

### Motivation and Context
This PR solves #11383.

### Types of changes

### Checklist

### DIFF
--- a/rust/src/chrome.rs
+++ b/rust/src/chrome.rs
@@ -29,7 +29,7 @@ use crate::metadata::{
 };
 use crate::{
     SeleniumManager, BETA, DASH_DASH_VERSION, DEV, ENV_LOCALAPPDATA, ENV_PROGRAM_FILES,
-    ENV_PROGRAM_FILES_X86, NIGHTLY, REG_QUERY, STABLE, WMIC_COMMAND,
+    ENV_PROGRAM_FILES_X86, FALLBACK_RETRIES, NIGHTLY, REG_QUERY, STABLE, WMIC_COMMAND,
 };
 
 const BROWSER_NAME: &str = "chrome";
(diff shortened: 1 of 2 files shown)


titusfortner:
Hmm, I think there should just be one or two fallbacks instead of 5? The specific problem at hand is that it looks like Chrome only releases drivers for production and beta, and not for dev and canary, so we should use the beta version of the driver for dev, and theoretically we figure out how to get the most recent nightly build of the driver for Canary. Microsoft makes everything much easier for Edge, and this isn't an issue for Firefox or Safari.

bonigarcia:
@titusfortner Having a retries count of 5 for this logic is not too much, IMO.
See, for example, the provided example for the canary version. In that case, the proper driver version is achieved on the third attempt since the browser version was 111, but chromedriver 111 is not published (first try), nor chromedriver 110 (second try), and the latest published version is 109 (third and final try, in that case).

All in all, the minimum retries number should be three. And having a guard of two more tries, just in case, does not hurt, IMO.

Issue #11365, opened 2022-12-03
[🚀 Feature]: Selenium Manager change outputs


titusfortner:
### Feature and motivation

@nvborisenko proposes:
> if success - exit code should be 0, and the latest line in output should be exact the path to the binary".

@symonk proposes:
> never output any logging level to stdout or stderr, just write errors to stderr and non errors to stdout, that way it's easier to pipe to other tools etc and not require every consumer who cares about the path output have to do parsing to get what they care about

My biggest take is that we shouldn't be putting error messages in stdout. I don't like having to remove `INFO\t` from results, especially since it is inconsistent with behavior of `--version`.

Maybe a compromise is to log the level when `--debug` is turned on, and leave it off otherwise?

### Current behavior

This is stdout with error code 0
```
$ ~/selenium-manager --browser firefox
INFO	/Users/titusfortner/.cache/selenium/geckodriver/mac64/0.32.0/geckodriver
```

This is also stdout with error code 0
```
$ ~/selenium-manager --version
selenium-manager 1.0.0-M2
```

This is stdout with error code > 0

```
$ ~/selenium-manager --browser firefoxes
Error: "Invalid browser/driver name"
```

This is stderr with error code > 0
```
$ ~/selenium-manager --what
error: Found argument '--what' which wasn't expected, or isn't valid in this context

  If you tried to supply '--what' as a value rather than a flag, use '-- --what'

Usage: selenium-manager [OPTIONS]

For more information try '--help'
```

stderr with error code > 0
```
$ ~/selenium-manager --browser
error: The argument '--browser <BROWSER>' requires a value but none was supplied

For more information try '--help'
```

bonigarcia:
In my opinion, the log level (`INFO`, `WARN`. etc.) help to understand the log info, also when selenium-manager is executed directly from the shell (e.g. as an standalone app), so I would not remove this level. In the case of driver management, which is the big feature supported so far, the driver path is contained in a `INFO` log (which is unique, i.e., there is no more `INFO` logs). But even in that case, there could be `WARN` logs (see #11373) that contain warning messages that could be processed in the bindings to warn the user about some undesired behaviour (not and error, but something that can be potentially improved) to the final user.

AutomatedTester:
Could we make the output computer consumable? E.g. Just print the messages within a JSON blob?

titusfortner:
I was going to say that JSON seems like overkill, but then I got to thinking about how the bindings should convey potential additional information to the user.

[...]

PR #11462, opened 2022-12-21
Updated a variable, Replace str in SeleniumManager

balaji2711:
Changed the variable from output to executablePath and replaced an "INFO\t" with an empty string immediately after trimming.

### DIFF
--- a/dotnet/src/webdriver/SeleniumManager.cs
+++ b/dotnet/src/webdriver/SeleniumManager.cs
@@ -143,14 +143,14 @@ namespace OpenQA.Selenium
                 process.ErrorDataReceived -= outputHandler;
             }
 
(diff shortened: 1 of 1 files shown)


titusfortner:
Thanks for looking at this code and trying to make it better.

The "problem" is that what we're getting from StringBuilder is not always the executablePath, sometimes it is an error. If it's an error I just want to log it / throw it as it comes out. If it's not an error, then it is probably the path, and it's ok to remove INFO. In reality, based on what is currently implemented in Selenium Manager, this won't make a difference either way.

Big picture, though, what we have right now is confusing and the right thing is to change the format of the response entirely (#11365)

balaji2711:
Thanks for your comments. I will look into the feature (https://github.com/SeleniumHQ/selenium/issues/11365)

PR #11477, opened 2022-12-24
[java] Overload methods creating interactions to accept `Point`

valfirst:
### Description
Add overloaded methods creating interactions to accept `Point` instead of (`x`, `y`) pair.

### Motivation and Context
New methods are more convenient for users how work with point objects.
Invocations like
```
..., point.getX(), point.getY(), ...
```
can be replaced with simple
```
..., point, ...
```

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/interactions/PointerInput.java
+++ b/java/src/org/openqa/selenium/interactions/PointerInput.java
@@ -17,6 +17,7 @@
 
 package org.openqa.selenium.interactions;
 
+import org.openqa.selenium.Point;
 import org.openqa.selenium.WebElement;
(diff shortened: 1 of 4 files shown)


pujagani:
Implementation wise it seems fine. Though this adds an additional method, I am not sure it is consistent with other languages. @titusfortner I would like your thoughts here. Especially, I remember effort was put in to ensure all languages have a consistent API usage-wise. https://www.selenium.dev/documentation/webdriver/actions_api/wheel/

diemol (approved):
I think this is fine because the Java bindings already had this `Point` class implemented long before.

PR #11478, opened 2022-12-25
[rb] update logger with link on how to use it

titusfortner:
### Description

* Add a description with a link to info about Selenium logger when the logger gets used for the first time.

### Motivation and Context

We periodically get complaints about the Ruby Selenium logger and how chatty it gets. Especially with deprecation warnings. The Selenium Logger code lets you ignore any or all messages, but it isn't clearly documented anywhere. This PR points to all of the info on what it does and how to update it.

This code shouldn't be merged until it's actually added in the documentation — https://github.com/SeleniumHQ/seleniumhq.github.io/pull/1271

Let me know any issues with wording.

### DIFF
--- a/rb/lib/selenium/webdriver/common/logger.rb
+++ b/rb/lib/selenium/webdriver/common/logger.rb
@@ -48,9 +48,10 @@ module Selenium
       #
       # @param [String] progname Allow child projects to use Selenium's Logger pattern
       #
-      def initialize(progname = 'Selenium')
+      def initialize(progname = 'Selenium', ignored: nil)
         @logger = create_logger(progname)
-        @ignored = []
+        @ignored = Array(ignored)
+        @first_warning = false
       end
 
       #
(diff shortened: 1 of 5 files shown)


rb/lib/selenium/webdriver/common/logger.rb
```diff
@@ -48,9 +48,13 @@ class Logger
       #
       # @param [String] progname Allow child projects to use Selenium's Logger pattern
       #
-      def initialize(progname = 'Selenium')
+      def initialize(progname = 'Selenium', ignored: nil)
         @logger = create_logger(progname)
-        @ignored = []
+        @ignored = Array(ignored)
+
+        warn("Details on how to use and modify Selenium logger:\n", id: [:logger_info]) do
```
p0deje: This will print the line on the first initialization of the logger, even if there are no warnings later in the code. Maybe we should instead print this once in the first call of `deprecate`?

titusfortner: We don't initialize the logger by default, though, so it should only print if the logger actually prints a warning.

Hmm, though I guess it still prints if you initialize it and immediately set it to trace.

I'll rework.

p0deje: We initialize it all over the Selenium codebase too, so it is likely going to be printed as soon as you instantiate the driver.

titusfortner: oh, right, the logger is the conditional... yup.

titusfortner: ok, this should be better. :)

p0deje (approved):

PR #11491, opened 2022-12-27
[java] implement DriverFinder completely independent of Service classes

titusfortner:
### Description
This is an alternate implementation of #11411 

Advantages: 
* better separation of concerns; it keeps the finding logic completely separate from Service classes
* backwards compatible; user doesn't need to change anything in their code for this to work

Disadvantages: 
* The implementation defeats the purpose of the builder pattern; the value can be null in the constructor, and requires a setter/getter
* Exceptions in finding the driver happen in Driver constructor, not the Service constructor

### Motivation and Context

I got the idea from trying to see how I would implement this functionality in Ruby.

The underlying problem is that a user can create a service instance before the options instance. We can either:
* Force the user to pass in options to the Service class if they are creating their own (#11411), but we agreed last TLC meeting that we didn't like that idea in general, so we want to deprecate that rather than extend it.
* Somehow find and update the executable path on behalf of the user regardless of how they are constructing.

Is there a better Java pattern here? Like, I guess we can avoid the setter/getter by building a new Service instance from the Driver constructor with a builder method that takes an existing Service instance and the executable path from the DriverFinder, but that doesn't prevent a Service instance from having a null executable value when constructed, and errors still happen in driver constructor rather than the service construction.

Any feedback appreciated.

### DIFF
--- a/java/src/org/openqa/selenium/chrome/ChromeDriverService.java
+++ b/java/src/org/openqa/selenium/chrome/ChromeDriverService.java
@@ -21,9 +21,9 @@ import com.google.auto.service.AutoService;
 
 import org.openqa.selenium.Capabilities;
 import org.openqa.selenium.WebDriverException;
-import org.openqa.selenium.remote.service.DriverServiceInfo;
 import org.openqa.selenium.chromium.ChromiumDriverLogLevel;
 import org.openqa.selenium.remote.service.DriverService;
+import org.openqa.selenium.remote.service.DriverServiceInfo;
 
 import java.io.File;
 import java.io.IOException;
@@ -184,7 +184,7 @@ public class ChromeDriverService extends DriverService {
    * @return Whether the browser driver path was found.
    */
   static boolean isPresent() {
-    return findExePath("chromedriver", CHROME_DRIVER_EXE_PROPERTY) != null;
+    return findExePath(CHROME_DRIVER_NAME, CHROME_DRIVER_EXE_PROPERTY) != null;
   }
 
   /**
--- a/java/src/org/openqa/selenium/chrome/ChromeDriverService.java
+++ b/java/src/org/openqa/selenium/chrome/ChromeDriverService.java
@@ -21,6 +21,7 @@ import com.google.auto.service.AutoService;
 
 import org.openqa.selenium.Capabilities;
 import org.openqa.selenium.WebDriverException;
+import org.openqa.selenium.remote.service.DriverServiceInfo;
 import org.openqa.selenium.chromium.ChromiumDriverLogLevel;
 import org.openqa.selenium.remote.service.DriverService;
 
@@ -40,6 +41,14 @@ import static org.openqa.selenium.remote.Browser.CHROME;
  * Manages the life and death of a ChromeDriver server.
  */
 public class ChromeDriverService extends DriverService {
+
+  public static final String CHROME_DRIVER_NAME = "chromedriver";
+
+  /**
+   * System property that defines the location of the ChromeDriver executable that will be used by
+   * the {@link #createDefaultService() default service}.
+   */
+  public static final String CHROME_DRIVER_EXE_PROPERTY = "webdriver.chrome.driver";
 
   /**
    * System property that toggles the formatting of the timestamps of the logs
@@ -129,10 +138,18 @@ public class ChromeDriverService extends DriverService {
       unmodifiableMap(new HashMap<>(environment)));
   }
 
+  public String getDriverName() {
+    return CHROME_DRIVER_NAME;
+  }
+
+  public String getDriverProperty() {
+    return CHROME_DRIVER_EXE_PROPERTY;
+  }
+
   /**
    * Configures and returns a new {@link ChromeDriverService} using the default configuration. In
    * this configuration, the service will use the ChromeDriver executable identified by
-   * {@link org.openqa.selenium.remote.service.DriverFinder#getPath(Capabilities, String, String)} )}.
+   * {@link org.openqa.selenium.remote.service.DriverFinder#getPath(DriverServiceInfo, Capabilities)}.
    * Each service created by this method will be configured to use a free port on the current system.
    *
    * @return A new ChromeDriverService using the default configuration.
@@ -144,7 +161,7 @@ public class ChromeDriverService extends DriverService {
   /**
    * Configures and returns a new {@link ChromeDriverService} using the supplied configuration. In
    * this configuration, the service will use the ChromeDriver executable identified by
-   * {@link org.openqa.selenium.remote.service.DriverFinder#getPath(Capabilities, String, String)}.
+   * {@link org.openqa.selenium.remote.service.DriverFinder#getPath(DriverServiceInfo, Capabilities)}.
    * Each service created by this method will be configured to use a free port on the current system.
    *
    * @return A new ChromeDriverService using the supplied configuration from {@link ChromeOptions}.
--- a/java/src/org/openqa/selenium/chrome/ChromeDriverService.java
+++ b/java/src/org/openqa/selenium/chrome/ChromeDriverService.java
@@ -170,10 +170,21 @@ public class ChromeDriverService extends DriverService {
   @Deprecated
   public static ChromeDriverService createServiceWithConfig(ChromeOptions options) {
     ChromeDriverLogLevel oldLevel = options.getLogLevel();
-    ChromiumDriverLogLevel level = (oldLevel == null) ? null : ChromiumDriverLogLevel.fromString(oldLevel.toString());
+    ChromiumDriverLogLevel level = (oldLevel == null) ? null :
+                                   ChromiumDriverLogLevel.fromString(oldLevel.toString());
     return new Builder()
       .withLogLevel(level)
       .build();
+  }
+
+  /**
+   * Checks if the browser driver binary is already present. Grid uses this method to show
+   * the available browsers and drivers, hence its visibility.
+   *
+   * @return Whether the browser driver path was found.
+   */
+  static boolean isPresent() {
+    return findExePath("chromedriver", CHROME_DRIVER_EXE_PROPERTY) != null;
   }
 
   /**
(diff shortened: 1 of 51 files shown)


java/src/org/openqa/selenium/remote/service/DriverService.java
```diff
@@ -110,56 +110,22 @@ protected DriverService(
      Duration timeout,
      List<String> args,
      Map<String, String> environment) throws IOException {
-   this.executable = executable.getCanonicalPath();
+   if (this.executable != null) {
```
pujagani: I think this should be `(executable != null)`.  Because  `this.executable` will always be null before it is assigned in this check.

java/src/org/openqa/selenium/manager/SeleniumManager.java
```diff
@@ -147,20 +148,15 @@ private File getBinary() {
...
-            throw new WebDriverException("Unable to locate driver with name: " + driverName);
-        }
-
-        String driverPath = null;
+    public String getDriverPath(Capabilities options) {
         File binaryFile = getBinary();
-        if (binaryFile != null) {
-          driverPath = runCommand(binaryFile.getAbsolutePath(),
-                    "--driver", driverName.replaceAll(EXE, ""));
+        List<String> commandList = Arrays.asList(binaryFile.getAbsolutePath(), "--browser", options.getBrowserName());
```
pujagani, line 163: Just curious, why is the null check removed for the `binaryFile`?

pujagani:
Can we not have a `Require.nonNull` check in the DriverService class constructors? https://github.com/SeleniumHQ/selenium/blob/436db51c33ccdf861f47311395af63858b776a96/java/src/org/openqa/selenium/chrome/ChromeDriverService.java#L126 this calls the DriverService class constructor. 
If executable is passed explicitly, it should not be null.

Since https://github.com/SeleniumHQ/selenium/blob/436db51c33ccdf861f47311395af63858b776a96/java/src/org/openqa/selenium/chrome/ChromeDriverService.java#L145 will use the DriverFinder methods to identify executable because user chooses to use the default mechanism. 

Just a suggestion, I might be missing something here.

java/src/org/openqa/selenium/chrome/ChromeDriverFinder.java
```diff
@@ -0,0 +1,45 @@
+package org.openqa.selenium.chrome;
+
+import org.openqa.selenium.internal.Require;
+import org.openqa.selenium.manager.SeleniumManager;
+import org.openqa.selenium.os.ExecutableFinder;
+
+import java.io.File;
+import java.util.logging.Logger;
+
+public class ChromeDriverFinder {
```
krmahadevan: Can't we not create a generic driver finder, which accepts the variables as args (binary name, JVM argument, error message details) and all the browser specific finders just calls into that generic driver finder ? That way we could avoid duplicating the logic across all the driver finders.

pujagani: +1 We could abstract DriverFinder

titusfortner:
> can we not have a Require.nonNull check in the DriverService class constructors

Driver finder needs the options instance. 
If service instance was always created in driver constructor, then we'd always have the options class for driver finder. But user can create it and pass it into the driver constructor, so we can't use finder. Service constructor needs to allow null executable because the builder is no longer locating the driver. It's getting set on an existing service instance in the driver constructor.
I don't see a good alternative that doesn't require the user to change code to initialize service class with options instance.

java/src/org/openqa/selenium/chrome/ChromeDriverFinder.java
titusfortner: Yes, we can definitely optimize this solution if we think it's the right approach.

java/src/org/openqa/selenium/manager/SeleniumManager.java
titusfortner: Not sure. Probably oversight after changing things too many times.

java/src/org/openqa/selenium/remote/service/DriverService.java
titusfortner: Yes, this is true

diemol (changes requested):
I think this PR looks better than #11411, thanks @titusfortner!

However, I agree with the comments, it'd be nice to have a single `DriverFinder` class.

titusfortner:
So, this moves all the `DriverFinder` classes into one, except for `SafariDriverFinder` because the logic is different. We could probably get rid of that one too if we overloaded a couple methods in `SafariDriver` (2 valid service classes) and could push the logic for checking OS version etc to the Selenium Manager, but I'm not sure that's better.

Really, though, combining the builder pattern with the one-off setter/getter feels really weird to me. It's nice that it just works without the user needing to do anything, but I was hoping there might be a better way to do this... 😄

titusfortner:
Oh, and I changed the default error message to link to our documentation instead of directly to the driver URLs, which I think is preferable.

java/src/org/openqa/selenium/chrome/ChromeDriverService.java
```diff
@@ -40,12 +40,6 @@
  */
 public class ChromeDriverService extends DriverService {
 
-  /**
-   * System property that defines the location of the ChromeDriver executable that will be used by
-   * the {@link #createDefaultService() default service}.
-   */
-  public static final String CHROME_DRIVER_EXE_PROPERTY = "webdriver.chrome.driver";
-
```
diemol: I know the system property is still being used, now in the driver class, but I'd prefer we leave a named static variable to contain it because it helps for code readability. So maybe move the removed statement above to `ChromeDriver`?

Same applies to the rest of the drivers.

diemol (changes requested):
A small change requested to help code readability. Everything else looks 👍 

Thank you!

java/src/org/openqa/selenium/chrome/ChromeDriverService.java
titusfortner: move it to driver or reference it in Service class? Gets weird with Firefox since the constructor takes a `FirefoxDriverService` not a `GeckoDriverService` argument.

java/src/org/openqa/selenium/remote/service/DriverFinder.java
```diff
@@ -0,0 +1,40 @@
...
+  private static final Logger LOG = Logger.getLogger(DriverFinder.class.getName());
+
+  public static String getPath(Capabilities options, String driverName, String driverProperty) {
+    String exePath = new ExecutableFinder().find(driverName);
+    if (exePath == null) {
+      exePath = System.getProperty(driverProperty);
+    }
+    if (exePath == null) {
+      try {
+        exePath = SeleniumManager.getInstance().getDriverPath(options);
```
krmahadevan: Maybe we can reduce this to 

```
exePath = System.getProperty(driverProperty, SeleniumManager.getInstance().getDriverPath(options));
```

pujagani:
Looks good to me overall.

java/src/org/openqa/selenium/remote/service/DriverFinder.java
titusfortner: Soon the Selenium manager will be handling the path logic as well... And soon we'll want to output the Selenium manager error instead of catching/hiding. So this code will get optimized again soon, but want to figure that out separately.

java/src/org/openqa/selenium/chrome/ChromeDriverService.java
titusfortner: My intermediary implementation was to implement a `DriverServiceInfo` Interface that each of the driver service classes implemented — https://github.com/SeleniumHQ/selenium/pull/11491/commits/2e91ebf5c2185cd9b665621ab8a4ab7b41d1a15e

I removed because I decided it was overkill, but it is probably a better way to do this if you think we need to keep the constants?

titusfortner:
I'm moving this to draft because there is a selenium manager issue that needs to get resolved first, and it won't get released until at least 4.9

titusfortner:
I figured out a simple way to remove the need for the `SafariDriverFinder` class by overloading `DriverFinder`'s `getPath()`

diemol:
> I'm moving this to draft because there is a selenium manager issue that needs to get resolved first, and it won't get released until at least 4.9

Which one?

titusfortner:
@diemol all the issues I ran into with the selenium manager binary have been sorted now.

java/src/org/openqa/selenium/chrome/ChromeDriverService.java
diemol: I meant move it to the driver service class.

pujagani:
There seems to be a linting error in the GeckoDriverServiceTest. Can you please help fix it? Overall, the PR LGTM.

titusfortner:
Diego wanted me to keep the constants in the Service class; instead of calling the service class values from the driver, I created an interface and methods to get the values in the drivers.

Except, I can't get that class to load with Bazel. (please help)

Also, I updated the DriverInfo classes because `isAvailable()` will no longer error when just building the service class. To make that easier, I created a basic version for getting path from DriverFinder via SeleniumManager that ignores Options class.

java/src/org/openqa/selenium/edge/EdgeDriverService.java
```diff
@@ -41,6 +42,8 @@
  */
 public class EdgeDriverService extends DriverService {
 
+  public static final String EDGE_DRIVER_NAME = "chromedriver";
+
```
diemol:
```suggestion
  public static final String EDGE_DRIVER_NAME = "msedgedriver";

```

diemol (changes requested):

diemol:
I did not notice this PR being ready and I caused the conflict. I will review, solve the conflict, and if everything OK, I will merge it.

java/src/org/openqa/selenium/remote/service/DriverService.java
```diff
@@ -456,10 +420,6 @@ public DS build() {
         port = PortProber.findFreePort();
       }
 
-      if (exe == null) {
-        exe = findDefaultExecutable();
-      }
-
```
diemol, line 462: Seems this bit is missing when using the Grid, `DriverServiceSessionFactory` uses the service directly. I need to check that part. Some tests are failing because of that.

diemol (approved):
Thank you, @titusfortner!

CI failures are not related to this change.

Issue #11483, opened 2022-12-26
[🚀 Feature]: Expose browser specific functionalities for remote webdrivers in Python


j3soon:
### Feature and motivation

As mentioned in saucelabs/sauce-docs#1621, workarounds are required to use browser specific functionalities in remote webdrivers. The current workaround for Python is much shorter than [those in C#](https://docs.saucelabs.com/web-apps/automated-testing/selenium/selenium4/#install-and-uninstall-add-ons), but much longer than [using `Augmenter` in Java](https://docs.saucelabs.com/web-apps/automated-testing/selenium/selenium4/#install-and-uninstall-add-ons).

I have noticed that the workarounds can be greatly simplified with some minor modifications in the Selenium codebase. Since Python is dynamically typed, implementing this feature in Python is much easier than in Java/C#. If the potential modifications are appropriate, I will open a PR.

The following are workarounds for four browser specific functionalities:
- Network Conditions
- Full-Page Screenshots
- Install and Uninstall Add-ons
- Change Preferences During Session

<hr>

#### Network Conditions

The entries for network conditions are [set during the initialization of `ChromiumRemoteConnection`](https://github.com/SeleniumHQ/selenium/blob/8d1fcf0506b1b01484401e3d381680e4eb6eec75/py/selenium/webdriver/chromium/remote_connection.py#L35-L37), which aren't set for remote webdrivers. Therefore, we need to set the three entries manually.

- `set_network_conditions` uses `_commands["setNetworkConditions"]`,
- `get_network_conditions` uses `_commands["getNetworkConditions"]`, and
- `delete_network_conditions` uses `_commands["deleteNetworkConditions"]`.
- See the [source code](https://github.com/SeleniumHQ/selenium/blob/8d1fcf0506b1b01484401e3d381680e4eb6eec75/py/selenium/webdriver/chromium/webdriver.py#L123-L154) for further info.

```py
from selenium import webdriver

driver = webdriver.Remote(
    command_executor='http://localhost:4444/wd/hub',
    options=webdriver.ChromeOptions(),
)

driver.command_executor._commands["setNetworkConditions"] = ("POST", "/session/$sessionId/chromium/network_conditions")
driver.command_executor._commands["getNetworkConditions"] = ("GET", "/session/$sessionId/chromium/network_conditions")
driver.command_executor._commands["deleteNetworkConditions"] = ("DELETE", "/session/$sessionId/chromium/network_conditions")

[...]

PR #11500, opened 2022-12-31
[wip][py] Expose browser specific functionalities for remote webdrivers

j3soon:
### Description

- [x] Expose browser specific functionalities for remote webdrivers in Python
- [ ] Add corresponding tests

### Motivation and Context

Implements #11483.

### Types of changes

### Checklist

### DIFF
--- a/py/selenium/webdriver/chromium/remote_connection.py
+++ b/py/selenium/webdriver/chromium/remote_connection.py
@@ -32,9 +32,6 @@ class ChromiumRemoteConnection(RemoteConnection):
         self.browser_name = browser_name
         self._commands["launchApp"] = ("POST", "/session/$sessionId/chromium/launch_app")
         self._commands["setPermissions"] = ("POST", "/session/$sessionId/permissions")
-        self._commands["setNetworkConditions"] = ("POST", "/session/$sessionId/chromium/network_conditions")
-        self._commands["getNetworkConditions"] = ("GET", "/session/$sessionId/chromium/network_conditions")
-        self._commands["deleteNetworkConditions"] = ("DELETE", "/session/$sessionId/chromium/network_conditions")
         self._commands["executeCdpCommand"] = ("POST", f"/session/$sessionId/{vendor_prefix}/cdp/execute")
         self._commands["getSinks"] = ("GET", f"/session/$sessionId/{vendor_prefix}/cast/get_sinks")
         self._commands["getIssueMessage"] = ("GET", f"/session/$sessionId/{vendor_prefix}/cast/get_issue_message")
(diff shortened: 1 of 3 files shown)


titusfortner:
@j3soon are you still working on this? What is the status?

j3soon:
@titusfortner, thanks for asking!

I have finished implementing the functionalities in this PR (commit aff181f817640585c6442698a05d770932fc3891). The corresponding user API can be found in the `Usage example` section in #11483, and have been tested manually (by myself) a while ago.

However, I didn't have time to convert those examples to automated Selenium test cases, which is required in [the contributing guideline](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md#step-5-test).

I would appreciate if someone could help convert those examples into Selenium test cases. Otherwise, I plan to write the test cases myself in the far future (maybe in 2024 or 2025).

titusfortner:
hah, yes, everyone is so busy at this point. Thanks for the update. We'll see what we can do for it. I'd definitely like to see this functionality added.

isaulv:
I don't like this code, and would not want to merge it in. But the current situation does bother me a lot, and there's is one hack that tries to load an appropriate browser specific remoteConnection. But your PR seems to show that more work needs to be done (which is something I was always aware of, when I first encountered a subclassed WebElement in the form of FirefoxWebelement).

j3soon:
Hi @isaulv , I acknowledge that this patch isn't ready to be merged yet, as indicated by the `[wip]` prefix in the title.

At the moment, the patch only serves as a proof-of-concept, showing that these features could be easily implemented using python-specific hacks, as discussed in issue #11483.

If you've managed to come up with a better implementation, please feel free to replace the current one.

It's worth noting that not merging this patch will not prevent the practical use of these functionalities. Users have the option to apply the python-specific hacks on their end, based on the [remote driver documentation](https://www.selenium.dev/documentation/webdriver/drivers/remote_webdriver/#browser-specific-functionalities). This is one of the reasons why I haven't prioritized continuing to work on this patch.

titusfortner:
So, the underlying feature here, is that Java, Ruby & .NET all have ways to implement browser specific functionality in Remote Driver.

https://docs.saucelabs.com/web-apps/automated-testing/selenium/selenium4/#network-conditions

We need to figure out some way, some how to provide this functionality in Python, and I have no idea the right way to do that. I know David doesn't like the idea of mixins, which is what Ruby is doing. Java is doing pure magic, and .NET is kind of magic that you have to be explicit about. 😂

isaulv:
@j3soon  I want to thank you for pointing out possible solutions. It's a big help for other users who need to handle this issue. I didn't mean to sound like a negative downer, I just want to have a solution that makes sense for a majority of Python programmers.

@titusfortner I agree with David, that mixins are a bad idea. I would like an automatic solution since Python is dynamic enough for this to happen. I don't like mixins for many reasons, but the main argument is that multiple inheritance can add a lot of complexity to debugging.
My goal is to have a solution that works with simple OO constructs but doesn't require monkeypatching or weird code. To that end, I'll study what Java, Ruby and .Net does and see what approach makes sense. Seems like Java relies on the Augmenter, which makes sense from a Java perspective, but makes the user deal with all the details. I aim for something that is automatic but also configurable.

j3soon:
Hi @isaulv, thanks for rekindling this discussion.

Studying the implementations in other languages before proceeding seems like a promising approach. If you came across any new insights or potentially better implementations, please share them with us. I'm looking forward to your findings and would be happy to learn how to implement this in the best way possible. :)

PR #11503, opened 2023-01-01
Rubocop updates

titusfortner:
@p0deje — any issues with these style / syntax changes?


rb/lib/selenium/webdriver/safari.rb
```diff
@@ -29,7 +29,7 @@ class << self
         attr_accessor :use_technology_preview
 
         def technology_preview
-          "/Applications/Safari\ Technology\ Preview.app/Contents/MacOS/safaridriver"
+          '/Applications/Safari Technology Preview.app/Contents/MacOS/safaridriver'
```
p0deje, line 32: Does it still work?

p0deje (approved):
I left on comment, but other than that I think it's safe to merge.

rb/lib/selenium/webdriver/safari.rb
titusfortner: Yup, it's still working.

PR #11507, opened 2023-01-03
Add annotations to large JS executions

### LINKED ISSUE #11038: [🚀 Feature]: Annotate JS atoms
### Feature and motivation

We've moved several common features that used to be endpoints into atoms that get sent as large JS strings to the execute endpoint. We should prepend a comment in all the bindings for these JS executions to label what they are.

### Usage example

Right now it looks like this:

```
"script": "return (function(){return (function(){var k=this||self;function aa(a){return\"string\"==typeof a}function.... <snip 45kb>
```

would be much better for it to look like:
```
"script": "/* isDisplayed */ return (function(){return (function(){var k=this||self;function aa(a){return\"string\"==typeof a}function.... <snip 45kb>
```

titusfortner:
### Description

Adds a small comment to the executing script with the name for the atom / large JS string getting executed. fixes #11038 

### Motivation and Context

When looking at logs of payloads to/from server/driver it is difficult to see right off what is getting executed, this will make it more obvious.

@harsha509 the tests are passing for this, but wanted to double check that I did it the right way for JS; thanks.

### DIFF
--- a/dotnet/src/webdriver/RelativeBy.cs
+++ b/dotnet/src/webdriver/RelativeBy.cs
@@ -48,7 +48,7 @@ namespace OpenQA.Selenium
                 }
             }
 
-            wrappedAtom = string.Format(CultureInfo.InvariantCulture, "return ({0}).apply(null, arguments);", atom);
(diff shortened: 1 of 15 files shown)


harsha509 (approved):
Hi @titusfortner ,

Yes it is correct for JS!

titusfortner:
merged

PR #11508, opened 2023-01-03
[ruby][bidi] Add filtering capability to LogInspector

TamsilAmani:
### Description

Add filtering capability to LogInspector

### Motivation and Context

Allow users to filter logs based on log level

### Types of changes

### Checklist

### DIFF
--- a/rb/spec/integration/selenium/webdriver/bidi/log_inspector_spec.rb
+++ b/rb/spec/integration/selenium/webdriver/bidi/log_inspector_spec.rb
@@ -25,7 +25,7 @@ module Selenium
       describe LogInspector, only: {browser: %i[chrome edge firefox]} do
         let(:page) { '/bidi/logEntryAdded.html' }
 
-        it 'can listen to console log', except: {browser: %i[chrome edge]} do
+        it 'can listen to console log' do
           reset_driver!(web_socket_url: true) do |driver|
             log_entry = nil
             log_inspector = described_class.new(driver)
@@ -40,14 +40,13 @@ module Selenium
               realm: nil,
               type: 'console',
               level: LogInspector::LOG_LEVEL[:INFO],
-              method: 'log',
-              stack_trace: nil
+              method: 'log'
             )
             expect(log_entry.args.size).to eq(1)
           end
         end
 
-        it 'can listen to console log with different consumers', except: {browser: %i[chrome edge]} do
+        it 'can listen to console log with different consumers' do
           reset_driver!(web_socket_url: true) do |driver|
             log_entry = nil
             log_inspector = described_class.new(driver)
@@ -65,15 +64,14 @@ module Selenium
               realm: nil,
               type: 'console',
               level: LogInspector::LOG_LEVEL[:INFO],
-              method: 'log',
-              stack_trace: nil
+              method: 'log'
             )
             expect(log_entry.args.size).to eq(1)
             expect(log_entry_text).to eq('Hello, world!')
           end
         end
 
-        it 'can filter console info level log', except: {browser: %i[chrome edge]} do
+        it 'can filter console info level log' do
           reset_driver!(web_socket_url: true) do |driver|
             log_entry = nil
             log_inspector = described_class.new(driver)
@@ -88,14 +86,13 @@ module Selenium
               realm: nil,
               type: 'console',
               level: LogInspector::LOG_LEVEL[:INFO],
-              method: 'log',
-              stack_trace: nil
+              method: 'log'
             )
             expect(log_entry.args.size).to eq(1)
           end
         end
 
-        it 'can filter console log', except: {browser: %i[edge]} do
+        it 'can filter console log' do
           reset_driver!(web_socket_url: true) do |driver|
             log_entry = nil
             log_inspector = described_class.new(driver)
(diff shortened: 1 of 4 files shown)


rb/spec/integration/selenium/webdriver/bidi/log_inspector_spec.rb
```diff
@@ -21,8 +21,8 @@
 
 module Selenium
   module WebDriver
-    class BiDi
-      describe LogInspector, exclusive: {browser: %i[chrome firefox]} do
+    class BiDi # rubocop:disable Metrics/ClassLength
+      describe LogInspector, exclusive: {browser: %i[firefox]} do
```
titusfortner: Use `only` instead of `exclusive`

TamsilAmani: Changes committed

rb/spec/integration/selenium/webdriver/bidi/log_inspector_spec.rb
```diff
@@ -21,13 +21,13 @@
 
 module Selenium
   module WebDriver
-    class BiDi
-      describe LogInspector, exclusive: {browser: %i[chrome firefox]} do
+    class BiDi # rubocop:disable Metrics/ClassLength
```
titusfortner: let's add this file path to the exclude list in .rubocop.yml instead of in the code here.

rb/spec/integration/selenium/webdriver/bidi/log_inspector_spec.rb
```diff
@@ -21,13 +21,13 @@
...
+      describe LogInspector, only: {browser: %i[chrome edge firefox]} do
         before do
           @page = '/bidi/logEntryAdded.html'
```
titusfortner: for this, we should prefer `let` instead of setting an instance variable in a before hook: `let(:page) { '/bidi/logEntryAdded.html' }`

titusfortner (changes requested):
When I merged the PR into trunk, I might have done the guards incorrectly? Either way a bunch of tests aren't passing. Do we just need to remove the guards that don't need to be there?

TamsilAmani:
Yes @titusfortner the current guards are not working as expected. For example, I have stopped all tests from being run on chrome using the `except` hook. But one of the tests failed in chrome instead of being pending. Here's the error.

```
Expected example to fail since it is pending, but it passed.

  0) Selenium::WebDriver::BiDi::LogInspector can filter console log FIXED
     Expected pending 'Test guarded; no reason given' to fail. No error was raised.
     # ./spec/integration/selenium/webdriver/bidi/log_inspector_spec.rb:98
```

That's why I preferred to use the `exclusive` hook. This problem doesn't happen with it.

titusfortner:
If it's passing as pending that means that the feature being tested works in Chrome. Is it implemented or does the test not test the right thing?

TamsilAmani:
@titusfortner This is long but insightful.

Alright, so I got to know why some tests are failing in chrome. I'll brief it out in points.

1. It seems that Chrome has implemented BiDi to some extent. So, the tests which were supposed to fail in chrome earlier are now passing. [**RESULT** = 8/12 PASS, 4/12 PENDING]

2. The 4 tests which are pending in chrome are all related to the **`console logs`** generated from the browser. Let's take an example for the response returned when there's a console log in both chrome and firefox.

In Chrome,
`#<Selenium::WebDriver::BiDi::ConsoleLogEntry:0x0000000111a5ed78 @level="info", @text="Hello, world!", @timestamp=1673896631113.0, @stack_trace={"callFrames"=>[{"columnNumber"=>20, "functionName"=>"helloWorld", "lineNumber"=>16, "url"=>"http://localhost:8180//bidi/logEntryAdded.html"}, {"columnNumber"=>51, "functionName"=>"onclick", "lineNumber"=>6, "url"=>"http://localhost:8180//bidi/logEntryAdded.html"}]}, @type="console", @method="log", @realm=nil, @args=[{"type"=>"string", "value"=>"Hello, world!"}]>`

In Firefox,
`#<Selenium::WebDriver::BiDi::ConsoleLogEntry:0x000000012e0e3060 @level="info", @text="Hello, world!", @timestamp=1673896425144, @stack_trace=nil, @type="console", @method="log", @realm=nil, @args=[{"type"=>"string", "value"=>"Hello, world!"}]>`

As you can see, chrome returns **@stack_trace** which is **NOT NULL** while firefox returns @stack_trace which is **NULL**. Hence, a test to check `stack_trace == empty` will pass in firefox but fail in chrome.

This is what is causing the 4 tests to be pending. The response for console log is different in chrome and firefox. It is mainly due to the `stack_trace` property but there's another change in which **_chrome is returning a @realm and firefox isn't_** but this is not causing any trouble as of now.

**NOTE:**
- I couldn't find the right response which must be returned in case of console log (mainly the `stack_trace` property) on the w3c page. https://w3c.github.io/webdriver-bidi/#intro
- The CI will pass now most probably with 8 PASSING and 4 PENDING in chrome. There's no failure now. 

What do you suggest? :)

TamsilAmani:
Based on my understanding of the [WPT tests](https://github.com/web-platform-tests/wpt/blob/e81901f6539c378f6f6010f988358a84b5abea52/webdriver/tests/bidi/log/entry_added/stacktrace.py#L12-L13) which @whimboo pointed out in tlc group, I think that when `log level = info`, we don't expect a stack trace and when `log level = error`, we expect it.
This condition is correctly implemented in firefox but not in chrome as it gives a stack trace with level = info too. Hence, I've modified the tests to exclude chrome. Any further insights on this would be helpful.

titusfortner (approved):

Issue #10724, opened 2022-06-01
[🚀 Feature]: Consolidate Actions class features


titusfortner:
### Dependencies

* #11513 
* #11509 

### Feature and motivation

Did a comparative analysis of Actions class functionality between languages: 
https://www.selenium.dev/documentation/webdriver/actions_api/

#### Differences of note

* Python & Ruby can change default durations for scrolling & movement; Java & .NET can't
* Python & Ruby have a "move to Location" method, Java & .NET don't
* I just added support for Python & Ruby to specify what button to click on click and hold, release and click methods; Java & .NET don't have this
* I added support for Java to & Ruby to set an "active" pointer (e.g. use a pen with Actions instead of needing ActionBuilder) Python & .NET can't do this
* Pointer Event Properties require the Action Builder approach in Java/Python/.NET;
* Java requires using Sequences instead of having a single ActionBuilder class like Python & .NET

#### Proposal 1: Allow setting default pointer in Python & .NET

This saves the step of needing to define the action and add to the input and then perform.
Note: this example is Python, I haven't looked to see how difficult in .NET

```python
pen = ActionChains(driver)\
    .set_pointer(PointerInput(POINTER_PEN, "default pen"))\
    .move_to(element)\
    .perform()
```

#### Proposal 2: Move to Location method in Java & .NET
Adds one method to Actions class
Currently no option for moving based on Viewport origin, only Element origin
The parameter could be x, y integers or a single  Coordinate/Point object instance

```java
new Actions(driver)
        .moveToLocation(100, 200)
        .perform()
```

#### Proposal 3: Allow changing of default duration in Java & .NET (tracking in #12118)

Adds one constructor to Actions class
Appium wants a default of 50ms instead of 250ms
Putting it in Actions constructor is much lower impact than adding the parameter to each method
**Discuss with Appium Devs First**

```java
new Actions(driver, Duration.ofMillis(50))
        .moveByOffset(13, 15)
        .perform();
```

#### Proposal 4: Provide a better way to perform action sequences in Java 

The `Collections.singletonList` requirement just feels clunky and unintuitive to me.
It is extra overhead, and might be worth waiting to implement until after we do #10808 since that's where we're going to want to add multiple sequences.

instead of:
```java
((RemoteWebDriver) driver).perform(Collections.singletonList(actions));
```

Wrap it with this:
```java
new ActionsBuilder(driver)
  .addSequence(actions)
  .perform();
```

[...]

PR #11509, opened 2023-01-04
[java] [dotnet] Add move to location method to Actions

pujagani:

### Description

Add move to location method to Actions class

### Motivation and Context

Explained in #10724. The change implement proposal 2.

### Types of changes

### Checklist

### DIFF
--- a/dotnet/src/webdriver/Interactions/Actions.cs
+++ b/dotnet/src/webdriver/Interactions/Actions.cs
@@ -315,6 +315,18 @@ namespace OpenQA.Selenium.Interactions
         }
 
         /// <summary>
+        /// Moves the mouse from the upper left corner of the current viewport by the provided offset.
+        /// </summary>
+        /// <param name="offsetX">The horizontal offset to which to move the mouse.</param>
+        /// <param name="offsetY">The vertical offset to which to move the mouse.</param>
+        /// <returns>A self-reference to this <see cref="Actions"/>.</returns>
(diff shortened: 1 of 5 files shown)


titusfortner:
force pushing a rebased update; the new Java formatting updates made it easier to do it this way. Evaluating now.

titusfortner:
@diemol which of these does it make more sense to support? It feels like we should be picking the "right" one and have people use it rather than supporting two options? `Point` isn't being used anywhere else, but everything else is offsets not something that represents coordinates on the viewport, so it might be more correct...

```java
    Action moveAndClick = getBuilder(driver).moveToLocation(70, 60).click().build();
```
and
```java
    Action moveAndClick = getBuilder(driver).moveToLocation(new Point(70, 60)).click().build();
```

Also, my default for multiple constructors or methods that do the same thing with different parameters is that one should call the other rather than be a re-implementation. So one way or the other I think this PR needs to be tweaked slightly.

.NET looks good; don't have .NET environment working on new laptop yet, though.

titusfortner:
Hmm .NET test is failing in CI... Passing in my fresh .NET dev environment; probably a timing issue; trying the equivalent test from Ruby.

diemol:
This one might be "easier"?

```java
Action moveAndClick = getBuilder(driver).moveToLocation(70, 60).click().build();
```

titusfortner:
Java looks good, but .NET + Windows + Firefox isn't happy and I'll have to load up a VM to troubleshoot it.

PR #11513, opened 2023-01-05
[dotnet] Allow setting a different pointer, keyboard, or wheel on input device

pujagani:

### Description

Allow setting a different pointer, keyboard, or wheel on input device without using Action builder

### Motivation and Context

Related to proposal 1 described in #10724

### Types of changes

### Checklist

### DIFF
--- a/common/src/web/login.html
+++ b/common/src/web/login.html
@@ -1,0 +1,39 @@
+<html>
+
+<head> 
+  <title>Login</title>
+</head>
+
+<body>
+  <main id="main-holder">
+    <h1 id="login-header">Login</h1> 
+    <form id="login-form">
+      <input type="text" name="username" id="username-field" class="login-form-field" placeholder="Username">
+      <input type="password" name="password" id="password-field" class="login-form-field" placeholder="Password">
+      <input type="submit" value="Login" id="login-form-submit">
+    </form>
+  
+  </main>
+</body>
+
+<script type="text/javascript">
+const loginForm = document.getElementById("login-form");
+const loginButton = document.getElementById("login-form-submit");
+
+loginButton.addEventListener("click", (e) => {
+    e.preventDefault();
+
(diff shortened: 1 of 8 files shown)


TamsilAmani:
@pujagani, As mentioned in the issue https://github.com/SeleniumHQ/selenium/issues/10724, proposal 1 is stated as...
`Proposal 1: Allow setting default pointer in Python & .NET` 

I read about pointers and found there are only three types of pointers in selenium as of now. Mouse, pen and touch devices. https://github.com/SeleniumHQ/selenium/blob/trunk/dotnet/src/webdriver/Interactions/PointerInputDevice.cs#L53-L69

Your commit has `setActivePointer(PointerKind kind, String name)`, which seems correct. But in addition to that it also has `setActiveKeyboard(String name)` and `setActiveWheel(String name)`. Do we need to set default keyboard and wheel too? I see there's only one kind for either of them. Correct me if I'm wrong.

titusfortner:
One interesting side effect of this implementation is that devices aren't created by default when class is loaded, but lazily generated (which is what Ruby does and I like it)... but I can't remember how it works to create a new device from the last time I was in the .NET actions code. If we've taken keyboard actions and then create a new pointer device, does it create an array of the size of the keyboard device actions list and add pauses, or will the first pointer action happen at the same time as the first keyboard action?

I'm not sure if we have a test covering this use case, and if not we need to add them.

pujagani:
Thank you for taking a look. I am not very familiar with the actions code. I am digging into it and will provide more information. I don't think the test exists for the use case mentioned, I think adding it will help me answer your question.

pujagani:
"If we've taken keyboard actions and then create a new pointer device, does it create an array of the size of the keyboard device actions list and add pauses" - Based on debugging and my understanding, this is what is happening. I have also added a test to confirm the same. Is that the expected behavior? @titusfortner

PR #11515, opened 2023-01-05
Use `p0deje/setup-bazel` GitHub Action on CI

p0deje:
It provides cross-platform configuration of Bazel and fine-grained caching for bazelisk, disk_cache, repository_cache and external repositories resulting in faster CI builds.

Read more at https://github.com/p0deje/setup-bazel.


titusfortner:
Ok, so the savings on this looks real. There's a lot of variation between runs, so this is just a snapshot.
Java had that weird extra 25 minute timeout we're hitting with sonarqube that we need to figure out so I included expected numbers for when we get that sorted. Ruby runs so many tests that it is more impacted by when the last job can start than how quickly anything takes to run. 

| Language | No Cache | Current Cache | New Cache
| --------------- | --------------- | --------------- | --------------- |
| .NET | 37:26 | 33:13 |  28:18 |
| Java | 1:38:20 (1:13:20) | 50:09 |  1:00:43 (45:03) |
| JS | 20:48 | 14:52 |  8:19 |
| Python | 37:43 | 34:07 |  30:12 |
| Ruby | 35:33 | 21:04 |  25:43 |

Anyway, looks good.

titusfortner (approved):

diemol (approved):
We should rule out whatever SonarQube is doing to slowdown the build, I plan to remove it and use something across the languages. 

This PR looks nice, thanks @p0deje!

What is the long term plan for the  `p0deje/setup-bazel` action?

titusfortner:
Have we been looking at the sonarqube results? It looks like failures are ignored, and I didn't even know they were there.

Can we just remove it now and add what we're going to pay attention to down the road? 

I'm not sure how I feel about a common solution for all languages; the tools do different things in different languages and Ruby & Python already have solutions in place.

p0deje:
> What is the long term plan for the `p0deje/setup-bazel` action?

@diemol I plan to keep maintaining it - I also use it in https://github.com/p0deje/rules_ruby. However, we might fully drop it as soon as we start using remote Bazel build execution since the whole process will be run on a different machine, not on CI node.

> We should rule out whatever SonarQube is doing to slowdown the build, I plan to remove it and use something across the languages.
> Can we just remove it now and add what we're going to pay attention to down the road?

I'll try to figure it out, but if it doesn't work - I'll disable for now.

p0deje:
When I run sonar-scanner manually, it also takes 20+ minutes with `JavaScript heap out of memory` errors all over. I'm going to disable this for now - we can revisit it later.

p0deje:
Merged in 80819a8efb5656219fa8b32e504e04d5c9a74c4a

diemol:
I think we can remove Sonar, we do not have access to the project anyway. I will remove it next week.

PR #11521, opened 2023-01-07
[py] Allow setting a different pointer, keyboard, or wheel on input device

TamsilAmani:
### Description

Allow setting a different pointer, keyboard, or wheel on input device without using Action builder

### Motivation and Context

Related to proposal 1 described in https://github.com/SeleniumHQ/selenium/issues/10724

### Types of changes

### Checklist

### DIFF
--- a/py/selenium/webdriver/common/action_chains.py
+++ b/py/selenium/webdriver/common/action_chains.py
@@ -21,7 +21,10 @@ import warnings
 from selenium.webdriver.remote.webelement import WebElement
 
 from .actions.action_builder import ActionBuilder
+from .actions.key_input import KeyInput
+from .actions.pointer_input import PointerInput
 from .actions.wheel_input import ScrollOrigin
+from .actions.wheel_input import WheelInput
 from .utils import keys_to_typing
 
 
@@ -58,7 +61,7 @@ class ActionChains:
     another.
     """
 
-    def __init__(self, driver, duration=250):
+    def __init__(self, driver, duration=250, devices=None):
         """Creates a new ActionChains.
 
         :Args:
@@ -66,7 +69,18 @@ class ActionChains:
          - duration: override the default 250 msecs of DEFAULT_MOVE_DURATION in PointerInput
         """
         self._driver = driver
-        self.w3c_actions = ActionBuilder(driver, duration=duration)
+        mouse = None
+        keyboard = None
+        wheel = None
+        if devices is not None and isinstance(devices, list):
+            for device in devices:
+                if isinstance(device, PointerInput):
+                    mouse = device
+                if isinstance(device, KeyInput):
+                    keyboard = device
+                if isinstance(device, WheelInput):
+                    wheel = device
+        self.w3c_actions = ActionBuilder(driver, mouse=mouse, keyboard=keyboard, wheel=wheel, duration=duration)
 
     def perform(self):
         """Performs all stored actions."""
--- a/py/test/selenium/webdriver/common/interactions_with_device_tests.py
+++ b/py/test/selenium/webdriver/common/interactions_with_device_tests.py
@@ -1,0 +1,374 @@
+# Licensed to the Software Freedom Conservancy (SFC) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The SFC licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+"""Tests for advanced user interactions."""
+import pytest
+
+from selenium.common.exceptions import MoveTargetOutOfBoundsException
+from selenium.webdriver.common.action_chains import ActionChains
+from selenium.webdriver.common.actions import interaction
+from selenium.webdriver.common.actions.key_input import KeyInput
+from selenium.webdriver.common.actions.pointer_input import PointerInput
+from selenium.webdriver.common.actions.wheel_input import ScrollOrigin
+from selenium.webdriver.common.actions.wheel_input import WheelInput
+from selenium.webdriver.common.by import By
+from selenium.webdriver.common.keys import Keys
+from selenium.webdriver.support.ui import WebDriverWait
+
+
+def _is_element_available(driver, id):
+    """Copied from org.openqa.selenium.interactions.TestBasicMouseInterface."""
+    try:
+        driver.find_element(By.ID, id)
+        return True
+    except Exception:
+        return False
+
+
+@pytest.mark.xfail_safari
+@pytest.mark.xfail_chrome
+def test_drag_and_drop_with_pointer(driver, pages):
+    """Copied from org.openqa.selenium.interactions.TestBasicMouseInterface."""
+    element_available_timeout = 15
+    wait = WebDriverWait(driver, element_available_timeout)
+    pages.load("droppableItems.html")
+    wait.until(lambda dr: _is_element_available(driver, "draggable"))
+
+    if not _is_element_available(driver, "draggable"):
+        raise AssertionError("Could not find draggable element after 15 seconds.")
+
+    toDrag = driver.find_element(By.ID, "draggable")
+    dropInto = driver.find_element(By.ID, "droppable")
+
+    mouse = PointerInput(interaction.POINTER_MOUSE, "test mouse")
+
+    holdDrag = ActionChains(driver, devices=[mouse]).click_and_hold(toDrag)
+    move = ActionChains(driver, devices=[mouse]).move_to_element(dropInto)
+    drop = ActionChains(driver, devices=[mouse]).release(dropInto)
+
+    holdDrag.perform()
+    move.perform()
+    drop.perform()
+
+    dropInto = driver.find_element(By.ID, "droppable")
+    text = dropInto.find_element(By.TAG_NAME, "p").text
+    assert "Dropped!" == text
+
+
+@pytest.mark.xfail_safari
+def test_double_click_with_pointer(driver, pages):
+    """Copied from org.openqa.selenium.interactions.TestBasicMouseInterface."""
+    pages.load("javascriptPage.html")
+    toDoubleClick = driver.find_element(By.ID, "doubleClickField")
+
+    mouse = PointerInput(interaction.POINTER_MOUSE, "test mouse")
+
+    dblClick = ActionChains(driver, devices=[mouse]).double_click(toDoubleClick)
+    dblClick.perform()
+    assert "DoubleClicked" == toDoubleClick.get_attribute("value")
+
+
+def test_context_click_with_pointer(driver, pages):
+    """Copied from org.openqa.selenium.interactions.TestBasicMouseInterface."""
+    pages.load("javascriptPage.html")
+    toContextClick = driver.find_element(By.ID, "doubleClickField")
+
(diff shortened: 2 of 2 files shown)


titusfortner:
First, thanks for taking a look at this, I've been dreading digging back into this code because I'm not sure I did the right thing in java last year... :)

The primary reason for this functionality is that the original implementation of Actions was one mouse one keyboard. In the move to w3c actions, you can have more devices, but the code we have doesn't support that concept. You can't easily use a pen or a finger type.

I think we may just want to keep it simple for now and not think about creating multiple devices and switching among existing devices, etc.

I'm looking at what I originally wrote in Java a year ago, and what I'm seeing doesn't make sense. We're looking for an existing name. But I don't think an existing name can exist because there is no place that we can add a device name. @pujagani am I missing something? (besides writing a test last year that would demonstrate it worked as I expected).

Like in the Python code here, we never add the new active pointer we are setting to the list of devices...

So I think we might be best off with the user doing:
```
ActionChains(driver).set_active_pointer(interaction.POINTER_PEN, "pen").move_by_offset(1, 1)
```

and the implementation:
```
    def set_active_pointer(self, kind, name):
            pointer_input = self.w3c_actions.add_pointer_input(kind, name)
            self.w3c_actions.pointer_action = PointerActions(pointer_input)

        return self
```

Do you think the additional functionality is important? Do we need to discuss more?


Oh and just because lambdas are cool. The way to avoid the loop you have in Python is:
```
pointer_input = next(filter(lambda x: x.name == name, self.w3c_actions.devices), None)
```

TamsilAmani:
@titusfortner I understand what you are saying. The addition functionality will be important if the user needs to switch between multiple devices. For example,

```
action_provider = ActionChains(driver).set_active_pointer(PointerInput(interaction.POINTER_MOUSE, "test mouse"))
{ do some action }
action_provider = ActionChains(driver).set_active_pointer(PointerInput(interaction. POINTER_PEN, "pen"))
{ do some action }
action_provider = ActionChains(driver).set_active_pointer(PointerInput(interaction.POINTER_MOUSE, "test mouse"))
{ do some action }

```
If we don't use the addition functionality, the above code will create POINTER_MOUSE objects twice. That's how I understand it.

pujagani:
@titusfortner You are right. I think we need additional functionality.  @TamsilAmani 's point sounds fair.  And for python, it looks like it is possible to implement it, so we can add it. What do you think?

titusfortner:
So, all the languages (except Ruby, for *reasons) have a simple API and an advanced API for working with Actions. It is possible to do all the things with the advanced API, it's just not intuitive and often a huge pain.

The discussion is about what features we want to make possible for users to do with the simple API:

1. Using a pen/finger instead of a mouse
2. Switching between different pointer/key/wheel devices
3. Using multiple fingers at the same time

The first I think is important, and a simple version of this PR can accomplish it.

The second I actually don't think is important (I can't think of a valid use case for people using it in normal testing), and I don't think the existing code in Java is actually doing this (we can test to make sure, but I'm pretty sure it is creating new devices rather than using existing ones, and I'm pretty sure advanced usage of it won't work as expected). 

More importantly, the current implementation will not help with multiple simultaneous actions. I think the correct solution for number 2 might be a trivial use case of number 3.

So, my current thought is we should remove the functionality for searching for existing devices for now, and figure out the advanced features in the future. What do you think?

pujagani:
Thank you for the detailed context. I now have a better understanding of the concern here. I agree with you, this PR can achieve "Using a pen/finger instead of a mouse" mouse bit for now. 
For advanced feature, let's look at it in the future and also check how it is possible in different languages.

TamsilAmani:
I also agree with @titusfortner. So, wrt to point 1 ... Shall I change the implementation code to.... ?

```
    def set_active_pointer(self, kind, name):
            pointer_input = self.w3c_actions.add_pointer_input(kind, name)
            self.w3c_actions.pointer_action = PointerActions(pointer_input)

        return self
```
to make it even simpler.

pujagani:
I think for the scope of this PR, @TamsilAmani yes!

titusfortner:
So actually, I realized the only reason I used "setActive" in Java is because I didn't want to create a bunch of new constructors. Can we do what Ruby does and add optional parameters to each of the mouse methods?:

```
    pen = PointerInput(interaction.POINTER_PEN, "pen")
    hoverable = driver.find_element(By.ID, "hover")
    ActionChains(driver)\
        .move_to_element(hoverable, device=pen)\
        .perform()
```

@pujagani I think we can technically do this in .NET as well? Though maybe better to match Java as much as possible.

py/selenium/webdriver/common/action_chains.py
```diff
@@ -370,6 +373,69 @@ def scroll(self, x: int, y: int, delta_x: int, delta_y: int, duration: int = 0,
...
+                break
+
+        if wheel is None:
+            self.w3c_actions.wheel_action = WheelActions(wheel_input)
+        else:
+            self.w3c_actions.wheel_action = WheelActions(wheel)
+
+        return self
+
+    def get_active_pointer(self):
```
symonk, line 427: This is very java-esque, can we just allow optional args to be passed in when instantiating?

symonk: these don't seem to offer more than a `@property` could, can we make them properties?

py/selenium/webdriver/common/action_chains.py
```diff
@@ -370,6 +373,69 @@ def scroll(self, x: int, y: int, delta_x: int, delta_y: int, duration: int = 0,
         )
         return self
 
+    def set_active_pointer(self, pointer_input):
```
symonk: This is very java-esque, can we just allow optional args to be passed in when instantiating?  Theres also alot of duplication in the logic here throughout all of them

py/selenium/webdriver/common/action_chains.py
```diff
@@ -21,6 +21,9 @@
 from selenium.webdriver.remote.webelement import WebElement
```
symonk, line 21: We have quite a lot of typing tech debt, could we type this new code?

py/selenium/webdriver/common/action_chains.py
TamsilAmani: I'll make them properties.

py/selenium/webdriver/common/action_chains.py
TamsilAmani: Do you mean that the logic of the 3 methods (`set_active_pointer, set_active_keyboard and set_active_wheel`) are the same and I should combine them in a single function with optional arguments?

I can do that but the three methods are kept for 3 different devices. Any suggestions for that?

pujagani:
@titusfortner  I will have to check if we can do this in C#. I have tried to match it to Java actually https://github.com/SeleniumHQ/selenium/pull/11513. Let me know if that looks fine. Probably, will need to remove the logic of finding a device exists or not based on the discussion done here.

py/selenium/webdriver/common/action_chains.py
titusfortner: We shouldn't need to do "set active" at all. "Active" can be changed to whatever device is passed into the specific method.

titusfortner: you can see what the Ruby bindings do for an example (though Ruby does other things differently like not creating default devices, etc).

titusfortner (approved):

py/selenium/webdriver/common/action_chains.py
TamsilAmani: @symonk I don't understand the term _'typing tech debt'_ Can you please tell me what it means? Also, I don't think I changed that code. Let me know if it still needs to be corrected after the latest commit.

TamsilAmani:
There are still few methods which need to implement these `device=None` parameter. I will do it later as a TODO or someone else can pick up.

titusfortner:
@TamsilAmani if you can't do it now, please create an issue so we can track it, thanks!

py/selenium/webdriver/common/action_chains.py
```diff
@@ -79,7 +79,7 @@ def reset_actions(self):
         for device in self.w3c_actions.devices:
             device.clear_actions()
 
-    def click(self, on_element=None):
+    def click(self, on_element=None, device=None):
```
TamsilAmani: @titusfortner Here's a tricky case which I just realised.

I have added `device` to click() and in lines 93 & 94 that device is passed to `key_action.pause()` signifying that device is `KeyInput`. This works with `send_keys_to_element()` (line 290) which calls `click()`. The tests pass.

Now what if there's a case where the user does 
```
mouse = PointerInput(interaction.POINTER_MOUSE, "test mouse")
ActionChains(driver).click(device=mouse).perform()
```
here the click() receives a `PointerInput` device and logically it should pass this device to `pointer_action.click()` (line 92) and **NOT** in `key_action.pause()` (lines 93 and 94)

Will the user ever do `ActionChains(driver).click(device=mouse).perform()` ?

Because the below code works as we pass pointer device in `move_to_element()` which sets the pointer and `click()` uses that. 
```
ActionChains(driver).move_to_element(toClick, device=mouse).click()
```

Am I thinking right? Should `click()` be able to differentiate between `PointerInput` and `KeyInput`.
If yes, shall I mention this in the new issue which I am creating?

TamsilAmani:
I have created a separate test file _"interactions_with_device_tests.py"_ which is a replica of the already present _"interactions_tests.py"_ with the modification that the new test file uses `device` parameters in the functions to test for the feature being implemented in this PR.

If there's a better place/approach to place the tests elsewhere, let me know :)

titusfortner:
@TamsilAmani showed me code last week that made me realize there is another complication in Python.

When there's only ever one pointer, you can create as many instances of the class as you want. If you need a non-default pointer, now things get weird. 

We need to take a look at the spec and see what the driver is supposed to remember of pointers and their state. The driver does maintain state between calls to the endpoint (`perform()` method calls), so we need to make sure that Python implementation maintains state the way the driver does.

TamsilAmani:
@AutomatedTester I have added the latest changes according to what we finalised. Now regarding the question that "Different instances of ActionChains pointing to the same device", which you said is not possible, here according to  @titusfortner (my discussion with him),
_The question is what does the driver "remember" between calls
I'm pretty sure the pointer doesn't reset after each call.
but does it share the info between the different types? or just within a type?_ 

Can you throw some light on this?

AutomatedTester:
The discussion I had with @TamsilAmani was about the follow code block.The following is psuedo-ish code so don't worry about the syntax too much.


```python

hold = ActionChains(driver).click_and_hold(element)
move = ActionChains(driver).move_to(element)
release = ActionChains(driver).release()
```

This would create 3 different action chains in python all with their own default. If you by chance got things to work in the way you intended with a mouse there is no guarantee that it would work if you only put a new `device` on any of the commands. 

`variable = SomeObject()` in python is equivalent to `variable = new SomeObject()` in Java.

So if we did the same example but with [defaults](https://w3c.github.io/webdriver/#dfn-default-pointer-parameters) annotated for ease and had 1 set as pen we would get 

```python

hold = ActionChains(driver).click_and_hold(element, device=Pen) #  device == Pen
move = ActionChains(driver).move_to(element) # device == Mouse
release = ActionChains(driver).release() # device == Mouse
```

Again, this is 3 separate chains. We can form one by doing the follow, and it will then keep the history in the client which is sent to the browser.

```python

hold = ActionChains(driver).click_and_hold(element, device=Pen) #  device == Pen
move = hold.move_to(element) # device == Pen
release = move.release() # device == Pen
```

Since the first code block was sending 3 different chains the browser it doesn't know that they are related. It makes the assumption that you know what you are sending over the wire is correct. It will just process them as 3 items. If things work out... awesome but that's not guaranteed as it's 3 different chains.

The [Perform Actions](https://w3c.github.io/webdriver/#perform-actions) command does not make any allowances for storing the action chain in the driver after executing them. It _may_ store some `global key state` between actions chains but again, not guaranteed.

If you were to call [Release Actions](https://w3c.github.io/webdriver/#release-actions) then it would process the 3 chains separately, again causing undefined behaviour as there is no guarantee that when the event loop fired it would give the desired affect.

Having had a cursory glance over the current code I think it moved to having a way of storing the device to be persistent if people want without having to remember to use it on all interaction

titusfortner:
I squashed things locally and rebased and am trying to fix the linting issues. The code itself is good and can be merged as soon as the CI is happy with it.

PR #11531, opened 2023-01-10
[rust] Support different output types (logger, json, shell) in Selenium Manager (#11365)

bonigarcia:
### Description
This PR implements the flag `--output` for Selenium Manager. The accepted values for this flag are:

- `LOGGER` : Usual log traces (`INFO`, `WARN`, etc.) used in the CLI. Default value.
- `JSON` : Custom JSON notation (see below).
- `SHELL` : Minimal Unix-like output (i.e., only full path when the driver is resolved or an error message if not).

To simplify things, the implemented JSON notation in this PR is simpler than proposed in #11365. The JSON basically contains the logs (level, timestamp, and message) displayed by Selenium Manager.

The bindings can invoke Selenium Manager using `--output json` and try parse its output. Then, the protocol to interpret the parsed JSON is as follows:

- When the return code is `0`: The last (and only) `INFO` message contains the driver path (when in future releases Selenium Manager also manages browsers, this message can be the browser path).
- When the return code is different than `0`: The last (and only) `ERROR` message contains the error message to throw an exception to the user.

In both cases, the `WARN` messages can be displayed to the user, since they contain some warning to the users (e.g., when a fallback is used to resolve the driver). In principle, the `DEBUG` and `TRACE` logs can safely ignored in the bindings.

Does it make sense?

An example of the JSON output is as follows (running Selenium Manager from the source code with Cargo):

```
> cargo run -- --browser chrome --output json
{
  "logs": [
    {
      "level": "TRACE",
      "timestamp": 1673373731,
      "message": "Reading metadata from C:\\Users\\boni\\.cache\\selenium\\selenium-manager.json"
    },
    {
      "level": "DEBUG",
      "timestamp": 1673373731,
      "message": "Using shell command to find out chrome version"
    },
    {
      "level": "DEBUG",
      "timestamp": 1673373731,
      "message": "Running cmd command: \"wmic datafile where name='%PROGRAMFILES:\\\\=\\\\\\\\%\\\\\\\\Google\\\\\\\\Chrome\\\\\\\\Application\\\\\\\\chrome.exe' get Version /value\""
    },
    {
      "level": "DEBUG",
      "timestamp": 1673373731,
      "message": "Output { status: ExitStatus(ExitStatus(0)), stdout: \"\\r\\r\\n\\r\\r\\nVersion=108.0.5359.125\\r\\r\\n\\r\\r\\n\\r\\r\\n\\r\\r\\n\", stderr: \"\" }"
    },
    {
      "level": "DEBUG",
      "timestamp": 1673373731,
      "message": "The version of chrome is 108.0.5359.125"
    },
    {
      "level": "TRACE",
      "timestamp": 1673373731,
      "message": "Writing metadata to C:\\Users\\boni\\.cache\\selenium\\selenium-manager.json"
    },
    {
      "level": "DEBUG",
      "timestamp": 1673373731,
      "message": "Detected browser: chrome 108"
    },
    {
      "level": "TRACE",
      "timestamp": 1673373731,
      "message": "Reading metadata from C:\\Users\\boni\\.cache\\selenium\\selenium-manager.json"
    },
    {
      "level": "TRACE",
      "timestamp": 1673373731,
      "message": "Driver TTL is valid. Getting chromedriver version from metadata"
    },
    {
      "level": "DEBUG",
      "timestamp": 1673373731,
      "message": "Required driver: chromedriver 108.0.5359.71"
    },
    {
      "level": "DEBUG",
      "timestamp": 1673373731,
      "message": "chromedriver 108.0.5359.71 already in the cache"
    },
    {
      "level": "INFO",
      "timestamp": 1673373731,
      "message": "C:\\Users\\boni\\.cache\\selenium\\chromedriver\\win32\\108.0.5359.71\\chromedriver.exe"
    }
  ]
}
```

### Motivation and Context
Feature requested on #11365. Discussed at the [Selenium TLC Meeting](https://www.selenium.dev/meetings/2022/tlc-12-08/#proposalsdecisions).

### Types of changes

### Checklist

### DIFF
--- a/rust/src/logger.rs
+++ b/rust/src/logger.rs
@@ -33,6 +33,8 @@ enum OutputType {
 }
 
 pub struct Logger {
+    debug: bool,
+    trace: bool,
     output: OutputType,
     json: RefCell<JsonOutput>,
 }
@@ -103,6 +105,8 @@ impl Logger {
         }
 
         Logger {
+            debug,
+            trace,
             output: output_type,
             json: RefCell::new(JsonOutput { logs: Vec::new() }),
         }
@@ -131,10 +135,15 @@ impl Logger {
     fn logger(&self, message: String, level: Level) {
         match self.output {
             OutputType::Json => {
-                self.json
-                    .borrow_mut()
-                    .logs
-                    .push(self.create_json_log(message, level));
+                let trace = level <= Level::Trace && self.trace;
+                let debug = level <= Level::Debug && self.debug;
+                let other = level <= Level::Info;
+                if trace || debug || other {
+                    self.json
+                        .borrow_mut()
+                        .logs
+                        .push(self.create_json_log(message, level));
+                }
             }
             OutputType::Shell => {
                 if level == Level::Info {
--- a/rust/src/logger.rs
+++ b/rust/src/logger.rs
@@ -81,9 +81,8 @@ impl Logger {
                 if trace {
                     filter = Trace
                 }
-
                 env_logger::Builder::new()
-                    .filter_module("selenium_manager", filter)
+                    .filter_module(env!("CARGO_CRATE_NAME"), filter)
                     .target(Stdout)
                     .format(|buf, record| {
                         let mut level_style = buf.style();
--- a/rust/src/logger.rs
+++ b/rust/src/logger.rs
@@ -163,9 +163,9 @@ impl Logger {
             }
             OutputType::Shell => {
                 if level == Level::Info {
-                    println!("{}", message);
+                    print!("{}", message);
                 } else if level == Level::Error {
-                    eprintln!("{}", message);
+                    eprint!("{}", message);
                 }
             }
             _ => {
@@ -190,7 +190,7 @@ impl Logger {
         let json_output = &self.json.borrow();
         let json = json_output.deref();
         if !json.logs.is_empty() {
-            println!("{}", serde_json::to_string_pretty(json.deref()).unwrap());
+            print!("{}", serde_json::to_string_pretty(json.deref()).unwrap());
         }
     }
 }
--- a/rust/src/logger.rs
+++ b/rust/src/logger.rs
@@ -47,8 +47,15 @@ pub struct Logs {
 }
 
 #[derive(Serialize, Deserialize)]
+pub struct Result {
+    pub code: i32,
+    pub message: String,
+}
+
+#[derive(Serialize, Deserialize)]
 pub struct JsonOutput {
     pub logs: Vec<Logs>,
+    pub result: Result,
 }
 
 impl Logger {
@@ -108,7 +115,13 @@ impl Logger {
             debug,
             trace,
             output: output_type,
-            json: RefCell::new(JsonOutput { logs: Vec::new() }),
+            json: RefCell::new(JsonOutput {
+                logs: Vec::new(),
+                result: Result {
+                    code: 0,
+                    message: "".to_string(),
+                },
+            }),
         }
     }
 
@@ -142,7 +155,10 @@ impl Logger {
                     self.json
                         .borrow_mut()
                         .logs
-                        .push(self.create_json_log(message, level));
+                        .push(self.create_json_log(message.to_string(), level));
+                }
+                if level == Level::Info || level <= Level::Error {
+                    self.json.borrow_mut().result.message = message;
                 }
             }
             OutputType::Shell => {
@@ -164,6 +180,10 @@ impl Logger {
             timestamp: now_unix_timestamp(),
             message,
         }
+    }
+
+    pub fn set_code(&self, code: i32) {
+        self.json.borrow_mut().result.code = code;
     }
 
     pub fn flush(&self) {
--- a/rust/src/logger.rs
+++ b/rust/src/logger.rs
@@ -1,0 +1,167 @@
+// Licensed to the Software Freedom Conservancy (SFC) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The SFC licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use crate::metadata::now_unix_timestamp;
+use env_logger::fmt::Color;
+use env_logger::Target::Stdout;
+use log::Level;
+use log::LevelFilter::{Debug, Info, Trace};
+use serde::{Deserialize, Serialize};
+use std::cell::RefCell;
+use std::io::Write;
+use std::ops::Deref;
+use Color::{Blue, Cyan, Green, Red, Yellow};
+
+enum OutputType {
+    Logger,
+    Json,
+    Shell,
+}
+
+pub struct Logger {
+    output: OutputType,
+    json: RefCell<JsonOutput>,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct Logs {
+    pub level: String,
+    pub timestamp: u64,
+    pub message: String,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct JsonOutput {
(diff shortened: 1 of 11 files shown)


titusfortner:
The consumer of the `json` output should be the bindings. As such, we should be able to get the information we want to provide the user without iteration or heavy lifting outside of the binary (the point of having a separate binary is so we don't have to have to create identical logic in each of the bindings). 

You mentioned that `DEBUG` and `TRACE` can be ignored in bindings, but then the example shows them. Perhaps we can do similar to what we do now and only show them if `--debug` is passed?

This is the kind of logic the bindings need:
```
if logging.debug
  args.add('--debug')
result = execute(selenium-manager, args)
if result.code > 0
  throw result.error

log result.logs
return result.value
```

I also think logs should be formatted as a string (similar to how webdriver spec does [stack traces](https://w3c.github.io/webdriver/#dfn-send-an-error) instead of a nested array.

So I'd prefer to see this structure:
```
{"value": "/path/to/driver",
"code": 0,
"logs": "Error getting version of chromedriver 110. Retrying with chromedriver 109 (attempt 1/5)"
}
```
or
```
{"value": "Error Message",
"code": 1,
}
```
Then if `--debug` is turned on, then the `logs` would include each line of DEBUG info (do not need the level or the timestamp, just the message):
```
"Line 1\nLine 2\nLine 3\nLine 4"
```

diemol:
>You mentioned that DEBUG and TRACE can be ignored in bindings, but then the example shows them. Perhaps we can do similar to what we do now and only show them if --debug is passed?

I assume that was merely informative, as the bindings will get that information back. Right, @bonigarcia?

bonigarcia:
> I assume that was merely informative, as the bindings will get that information back. Right, @bonigarcia?

@diemol Yes, that info is informative. I am not sure if the bindings are going to require that info. As @titusfortner suggested, I have just committed  a second [patch](https://github.com/SeleniumHQ/selenium/pull/11531/commits/ce4e280ae9b49107b54c972a5042839f0093e591) to this PR to flush these logs (`DEBUG` and `TRACE`) in the JSON only when the flags `--debug` and `--trace` are specified.

rust/src/logger.rs
```diff
@@ -0,0 +1,176 @@
...
+use std::ops::Deref;
+use Color::{Blue, Cyan, Green, Red, Yellow};
+
+enum OutputType {
+    Logger,
+    Json,
+    Shell,
+}
+
+pub struct Logger {
```
diemol, line 35: Is there no generic log utility in Rust? Just wondering if one needs to code the format. Minor thing from my side, though.

diemol (approved):
A minor comment from my side.

rust/src/logger.rs
bonigarcia: Selenium Manager uses a crate called [env_logger](https://docs.rs/env_logger/latest/env_logger/), which is a popular Rust logger that can be configured via environment variables. The implementation of the custom `Logger` structure in Selenium Manager wraps that crate (for the `LOGGER` option selected in the CLI, which is the default one) and also, it gathers these logs to be flushed as the end as JSON format (for the `JSON` option selected in the CLI).

bonigarcia:
@titusfortner To simplify things in the bindings, I have just added another patch to include the result code and message. See some examples:

```
% cargo run -- --browser chrome --output json

{
  "logs": [
    {
      "level": "INFO",
      "timestamp": 1673435421,
      "message": "/Users/bonigarcia/.cache/selenium/chromedriver/mac-arm64/108.0.5359.71/chromedriver"
    }
  ],
  "result": {
    "code": 0,
    "message": "/Users/bonigarcia/.cache/selenium/chromedriver/mac-arm64/108.0.5359.71/chromedriver"
  }
}
```

```
% cargo run -- --browser chrome --output json --debug

{
  "logs": [
    {
      "level": "DEBUG",
      "timestamp": 1673435452,
      "message": "Using shell command to find out chrome version"
    },
    {
      "level": "DEBUG",
      "timestamp": 1673435452,
      "message": "Running sh command: \"/Applications/Google\\\\ Chrome.app/Contents/MacOS/Google\\\\ Chrome --version\""
    },
    {
      "level": "DEBUG",
      "timestamp": 1673435452,
      "message": "Output { status: ExitStatus(unix_wait_status(0)), stdout: \"Google Chrome 108.0.5359.124 \\n\", stderr: \"\" }"
    },
    {
      "level": "DEBUG",
      "timestamp": 1673435452,
      "message": "The version of chrome is 108.0.5359.124"
    },
    {
      "level": "DEBUG",
      "timestamp": 1673435452,
      "message": "Detected browser: chrome 108"
    },
    {
      "level": "DEBUG",
      "timestamp": 1673435452,
      "message": "Required driver: chromedriver 108.0.5359.71"
    },
    {
      "level": "DEBUG",
      "timestamp": 1673435452,
      "message": "chromedriver 108.0.5359.71 already in the cache"
    },
    {
      "level": "INFO",
      "timestamp": 1673435452,
      "message": "/Users/bonigarcia/.cache/selenium/chromedriver/mac-arm64/108.0.5359.71/chromedriver"
    }
  ],
  "result": {
    "code": 0,
    "message": "/Users/bonigarcia/.cache/selenium/chromedriver/mac-arm64/108.0.5359.71/chromedriver"
  }
}
```

```
% cargo run -- --browser chrome --output json --browser-version canary

{
  "logs": [
    {
      "level": "WARN",
      "timestamp": 1673435485,
      "message": "Error getting version of chromedriver 110. Retrying with chromedriver 109 (attempt 1/5)"
    },
    {
      "level": "INFO",
      "timestamp": 1673435487,
      "message": "/Users/bonigarcia/.cache/selenium/chromedriver/mac-arm64/109.0.5414.74/chromedriver"
    }
  ],
  "result": {
    "code": 0,
    "message": "/Users/bonigarcia/.cache/selenium/chromedriver/mac-arm64/109.0.5414.74/chromedriver"
  }
}
```

```
% cargo run -- --browser wrong --output json

{
  "logs": [
    {
      "level": "ERROR",
      "timestamp": 1673436121,
      "message": "Invalid browser name: wrong"
    }
  ],
  "result": {
    "code": 65,
    "message": "Invalid browser name: wrong"
  }
}
```

bonigarcia:
This PR contains a couple of [tests](https://github.com/SeleniumHQ/selenium/pull/11531/files#diff-8843f17a35e855620b36bcb89317ac2f5bdaf288ae9df752478654dc3d41e643) to assess this feature. I think something similar to this should be done in the bindings (i.e., checking `result.code` and `result.message` from the parsed JSON).

titusfortner:
I get we shouldn't optimize for what we haven't implemented yet, but if we know we're going to be returning browser location at some point, then we know just having a single message line is insufficient. In logger mode it can just be another info line, and for shell most likely it's going to actually need to be two subsequent calls, once for browser and once for driver. In json mode, though we can use a different key name instead of a common message key. 

The logging focus still makes it hard to parse. The user doesn't care about timestamps or levels, they just want the result and to see any important caveats (warnings). We shouldn't need to write the same method in 5 different languages in order to parse an array of objects to pull out and reformat the info the user wants to see. 

How about by default we have an array of warning strings without the duplicated info string and without the timestamps? 
```
{
  "logs": ["Error getting version of chromedriver 110. Retrying with chromedriver 109 (attempt 1/5)"],
  "result": {
    "code": 0,
    "driver_path": "/Users/bonigarcia/.cache/selenium/chromedriver/mac-arm64/109.0.5414.74/chromedriver"
  }
}
```

rust/src/logger.rs
diemol: 👍

diemol:
I think the timestamp is important since Selenium Manager is used in the Grid as well. And eventually it could help to debug issues on tests executed in CI or similar. 

Which duplicated info string do you mean?

titusfortner:
Grid should be logging it with Java timestamps, and it's not like each microsecond step needs its own stamp here.
The only thing logged as info shows up in log and in output. Making the log not usable without a lot of parsing.

bonigarcia:
Removing the timestamp in Rust is straightforward. Ignoring that info from the bindings is also straightforward, IMO. If there is a chance that that info can be helpful in some scenario (e.g., Grid or other third-party), my bet is to leave the JSON as it is now. It is a tiny detail, IMO.

diemol:
I think if we want Selenium Manager to be used in the future by third parties, we should have a structured log (which includes a time stamp), and discarding information is easier than adding it when a wide audience is using the tool.

titusfortner:
Selenium has a history of trying to support lots of theoretical use cases instead instead of making the primary known case straightforward. I guess I'm not going to be the one implementing it, so I'll leave the parsing logic to you.

diemol:
Sounds good, so we can leave it there and a client using Selenium Manager can ignore the values if desired.

PR #11532, opened 2023-01-10
[java] have overloaded constructors to have configurable connect and read timeouts while instantiating WebDriver object

code-with-abdullah:
Currently the code only has 10 seconds set as connect timeout and 3 minutes set as read timeout.

This fix will allow this to configurable at constructor level i.e you now have a separate constructor for ChromeDriver that supports read timeout and connect timeout as well.


titusfortner:
I don't love the idea of creating new constructors for each of the drivers, and if we do go that route, we should implement the full ClientConfig rather than just these 2 values. Also, this is missing Edge & IE support, and for this kind of implementation we'd want to set default values and have the old constructors consume the new constructor using those values. (important because of the upcoming changes to driver classes with `DriverFinder` implementation).

diemol:
+1 to what @titusfortner says. Timeouts can be set already through the builder, we are figuring out how to avoid the user needing to set a default value.

code-with-abdullah:
@titusfortner @diemol First of all thanks for your feedback.

As rightly pointed out this was missing Edge and IE support initially. That has now been updated.
For the point that we should have default values and consume those in old constructors, java does not support default value constructors, can you please guide on how this can be addressed?
Timeouts can be set already through the builder but I could not find a way that would allow end user to create an instance of ChromeDriver with these configurations. This implementation can be used to further enhance the builder method in all classes to provide support for timeout at the time of driver object instantiation.

titusfortner:
I generally agree that even though it *can be done with the builder, you aren't working with a `ChromeDriver` instance, but an augmented `RemoteWebDriver` instance, which isn't as straightforward, and it would be nice to support it directly.

But, I don't think we just want to tack on extra timeout values. Instead of:
```
  public ChromeDriver(ChromeDriverService service, ChromeOptions options, int connectTimeoutSeconds, int readTimeoutSeconds) {  }
```
we should figure out how to implement:
```
  public ChromeDriver(ChromeDriverService service, ChromeOptions options, ClientConfig clientConfig) {  }
```
It actually looks like that might be easier to implement? I'm not following some of the changes being made here, so I'd need to pull it and test it to make sure it is actually doing what we want.

What I meant by using defaults and consume is something like:

```
DEFAULT_CONNECT_TIMEOUT = Duration.ofSeconds(10);
DEFAULT_READ_TIMEOUT = Duration.ofMinutes(3);

  public ChromeDriver(ChromeDriverService service, ChromeOptions options) {
this(service, options, DEFAULT_CONNECT_TIMEOUT, DEFAULT_READ_TIMEOUT);
}

  public ChromeDriver(ChromeDriverService service, ChromeOptions options, int connectTimeoutSeconds, int readTimeoutSeconds) {
    super(new ChromeDriverCommandExecutor(service), Require.nonNull("Driver options", options), ChromeOptions.CAPABILITY, connectTimeoutSeconds, readTimeoutSeconds);
    casting = new AddHasCasting().getImplementation(getCapabilities(), getExecuteMethod());
    cdp = new AddHasCdp().getImplementation(getCapabilities(), getExecuteMethod());
  }
```
One constructor uses the other with a default instead of duplicating logic. That way when I go to make a change in the constructor logic, I only have to do it in one place.

This is actually easier if we use `ClientConfig`, though, because we don't need default constants, we can just new up an instance:
```
  public ChromeDriver(ChromeDriverService service, ChromeOptions options) {
this(service, options, ClientConfig.defaultConfig());
}
```

diemol (changes requested):
We have been chatting internally and we think it makes sense to add the extra constructor to define the timeout locally. However, we'd prefer if a `ClientConfig` is passed instead of the two variables. @krmahadevan has helped to illustrate this in #11545, perhaps you can get inspiration from there and get this PR in shape. If you need more help, please reach out to us in the Slack channel #selenium-tlc, https://www.selenium.dev/support/. 

Thank you for putting this together.

code-with-abdullah:
Hi @diemol

Thanks for your feedback! 
I shall update this PR.

diemol:
> Hi @diemol
> 
> Thanks for your feedback! I shall update this PR.

Perfect, let us know if you need any help.

code-with-abdullah:
Dear @diemol.
I have updated the PR as per your request. Please review and let me know if there are any changes needed and I will integrate those as soon as possible.

java/src/org/openqa/selenium/remote/RemoteWebDriver.java
```diff
@@ -125,23 +125,34 @@ protected RemoteWebDriver() {
...
+    this(Require.nonNull("Capabilities", capabilities), true);
   }
 
   public RemoteWebDriver(Capabilities capabilities, boolean enableTracing) {
+    this(Require.nonNull("Capabilities", capabilities),
+      enableTracing,
+      ClientConfig.defaultConfig());
+  }
+
+  public RemoteWebDriver(Capabilities capabilities, boolean enableTracing, ClientConfig clientConfig) {
```
krmahadevan: I dont think we need to enhance the RemoteWebDriver because :
* it's not representing a local driver 
* you can use the RemoteWebDriverBuilder to build a remotewebdriver object

java/src/org/openqa/selenium/grid/node/CustomLocatorHandler.java
```diff
@@ -260,6 +260,12 @@ public CustomWebDriver(CommandExecutor executor, String sessionId) {
       setSessionId(sessionId);
     }
 
+    public CustomWebDriver(CommandExecutor executor, String sessionId, int connectTimeoutSeconds, int readTimeoutSeconds) {
```
krmahadevan: This change can be reverted as well

java/src/org/openqa/selenium/remote/RemoteWebDriverBuilder.java
```diff
@@ -135,6 +135,10 @@ public class RemoteWebDriverBuilder {
     // Access through RemoteWebDriver.builder
   }
 
+  RemoteWebDriverBuilder(ClientConfig clientConfig) {
```
krmahadevan: I think we can revert this as well because we already have `RemoteWebDriver.builder().config(ClientConfig.defaultConfig())`

java/src/org/openqa/selenium/remote/http/ClientConfig.java
```diff
@@ -64,6 +64,16 @@ public static ClientConfig defaultConfig() {
       null);
   }
 
+  public static ClientConfig configWithTimeouts(int connectTimeoutSeconds, int readTimeoutSeconds) {
```
krmahadevan: This also would not be needed because the current builder lets you set these values `      ClientConfig.defaultConfig().readTimeout(Duration.ofSeconds(1))
              .connectionTimeout(Duration.ofSeconds(1));  `

java/src/org/openqa/selenium/remote/service/DriverCommandExecutor.java
```diff
@@ -77,7 +78,12 @@ public DriverCommandExecutor(DriverService service) {
    */
   protected DriverCommandExecutor(
       DriverService service, Map<String, CommandInfo> additionalCommands) {
-    super(additionalCommands, service.getUrl());
+    this(service, additionalCommands, ClientConfig.defaultConfig());
+  }
+
+  protected DriverCommandExecutor(
+    DriverService service, Map<String, CommandInfo> additionalCommands, ClientConfig clientConfig) {
+    super(additionalCommands, service.getUrl(), clientConfig);
```
krmahadevan: We would need to add an edit check that ensures that the `ClientConfig.baseUri` is set and if its not set, we would need to read the value from `DriverService` and then set it.

java/src/org/openqa/selenium/chromium/ChromiumDriver.java
```diff
@@ -135,10 +135,10 @@ protected ChromiumDriver(CommandExecutor commandExecutor, Capabilities capabilit
 
     this.capabilities = cdpUri.map(uri -> new ImmutableCapabilities(
         new PersistentCapabilities(originalCapabilities)
```
krmahadevan: The changes in this class can be reverted since they are only whitespace changes

java/src/org/openqa/selenium/ie/InternetExplorerDriver.java
```diff
@@ -173,6 +174,11 @@ public static RemoteWebDriverBuilder builder() {
     return RemoteWebDriver.builder().oneOf(new InternetExplorerOptions());
   }
 
+  @Beta
+  public static RemoteWebDriverBuilder builder(ClientConfig clientConfig) {
```
krmahadevan: I think this wouldn't work because there's an edit check in the base class builder that vets out for a combination and then throws an exception. We should probably avoid this and instead just resort to adding the overloaded constructors as we have done for the remaining webdriver implementations so that we are consistent across the board.

java/src/org/openqa/selenium/chrome/ChromeDriver.java
```diff
@@ -91,7 +96,11 @@ public static RemoteWebDriverBuilder builder() {
 
   private static class ChromeDriverCommandExecutor extends ChromiumDriverCommandExecutor {
     public ChromeDriverCommandExecutor(DriverService service) {
-      super(service, getExtraCommands());
+      this(service, ClientConfig.ClientConfig.defaultConfig());
```
diemol: Is this correct? Not sure if it compiles. Probably one `ClientConfig.` less?

Please be sure to do something like `bazel build grid` locally to verify the code compiles.

java/src/org/openqa/selenium/firefox/FirefoxDriver.java
```diff
@@ -151,10 +159,10 @@ private FirefoxDriver(FirefoxDriverCommandExecutor executor, FirefoxOptions opti
...
     this.cdpUri = cdpUri;
     this.capabilities = cdpUri.map(uri ->
-                                     new ImmutableCapabilities(
-                                       new PersistentCapabilities(capabilities)
-                                         .setCapability("se:cdp", uri.toString())
-                                         .setCapability("se:cdpVersion", "85.0")))
+        new ImmutableCapabilities(
+          new PersistentCapabilities(capabilities)
+            .setCapability("se:cdp", uri.toString())
+            .setCapability("se:cdpVersion", "85.0")))
```
diemol, line 177: Please revert this formatting change.

java/src/org/openqa/selenium/ie/InternetExplorerDriver.java
diemol: Yes, we should follow the same proposed format as in the other driver classes.

java/src/org/openqa/selenium/remote/RemoteWebDriver.java
diemol: Exactly, this is already solved through the `RemoteWebDriverBuilder`

diemol (changes requested):
Left some additional comments, @code-with-abdullah.

code-with-abdullah:
Great. Bundle of thanks to both of you for your precious feedback. I shall integrate all changes mentioned above and build the code as suggested before my next commit.

code-with-abdullah:
@diemol and @krmahadevan 

Thanks for your input on the PR. All the changes you requested have been integrated. Please review and let me know if any further changes are required.

java/src/org/openqa/selenium/edge/EdgeDriver.java
```diff
@@ -62,8 +68,12 @@ public static RemoteWebDriverBuilder builder() {
   }
 
   private static class EdgeDriverCommandExecutor extends ChromiumDriverCommandExecutor {
-    public EdgeDriverCommandExecutor(DriverService service) {
-      super(service, getExtraCommands());
+    public EdgeDriverCommandExecutor(DriverService service) throws URISyntaxException {
+      this(service, ClientConfig.defaultConfig());
+    }
+
+    public EdgeDriverCommandExecutor(DriverService service, ClientConfig clientConfig) throws URISyntaxException {
```
krmahadevan: Please remove the `throws clause` here and also kindly help clean up the same from all the constructors as well. It's not needed.

java/src/org/openqa/selenium/chrome/ChromeDriver.java
```diff
@@ -90,8 +96,12 @@ public static RemoteWebDriverBuilder builder() {
   }
 
   private static class ChromeDriverCommandExecutor extends ChromiumDriverCommandExecutor {
-    public ChromeDriverCommandExecutor(DriverService service) {
-      super(service, getExtraCommands());
+    public ChromeDriverCommandExecutor(DriverService service) throws URISyntaxException {
```
krmahadevan: Please remove the `throws clause` here and also kindly help clean up the same from all the constructors as well. It's not needed.

java/src/org/openqa/selenium/ie/InternetExplorerDriver.java
```diff
@@ -110,15 +111,19 @@ public class InternetExplorerDriver extends RemoteWebDriver {
...
   }
 
   public InternetExplorerDriver(InternetExplorerOptions options) {
-    this(null, options);
+    this(null, options, null);
   }
 
   public InternetExplorerDriver(InternetExplorerDriverService service) {
-    this(service, null);
+    this(service, null, null);
```
krmahadevan: Can we please change this to pass the InternetExplorerOptions and the `ClientConfig` object here directly into the 3 args new constructor instead of sending in null values? The same would hold true for the other constructors in this class that are doing the same.

java/src/org/openqa/selenium/ie/InternetExplorerDriver.java
```diff
@@ -130,22 +135,27 @@ public InternetExplorerDriver(InternetExplorerDriverService service) {
...
   @Beta
   public static RemoteWebDriverBuilder builder() {
     return RemoteWebDriver.builder().oneOf(new InternetExplorerOptions());
   }
 
-  private void run(InternetExplorerDriverService service, Capabilities capabilities) {
+  private void run(InternetExplorerDriverService service, Capabilities capabilities, ClientConfig clientConfig) {
     assertOnWindows();
 
     setCommandExecutor(new DriverCommandExecutor(service));
```
krmahadevan: Even though we are getting a `ClientConfig` object, we are not using it. Please help add an overloaded constructor into `DriverCommandExecutor` that takes in a DriverService and a ClientConfig object and then internally invokes the other relevant constructors and here we should be calling that new constructor wherein we pass the client Config object.

java/src/org/openqa/selenium/chrome/ChromeDriver.java
krmahadevan: Also I think we can remove this now un-used constructor

java/src/org/openqa/selenium/edge/EdgeDriver.java
```diff
@@ -62,8 +68,12 @@ public static RemoteWebDriverBuilder builder() {
   }
 
   private static class EdgeDriverCommandExecutor extends ChromiumDriverCommandExecutor {
-    public EdgeDriverCommandExecutor(DriverService service) {
-      super(service, getExtraCommands());
+    public EdgeDriverCommandExecutor(DriverService service) throws URISyntaxException {
```
krmahadevan: This is an un-used constructor. We can remove it.

java/src/org/openqa/selenium/remote/RemoteWebDriver.java
```diff
@@ -175,16 +175,20 @@ private static URL getDefaultServerURL() {
   }
 
   private static CommandExecutor createExecutor(URL remoteAddress, boolean enableTracing) {
```
krmahadevan: we shouldn't be needing any change in RemoteWebDriver. Can you please help revert all the changes in this class ?

java/src/org/openqa/selenium/safari/SafariDriver.java
```diff
@@ -113,7 +118,11 @@ public void setFileDetector(FileDetector detector) {
 
   private static class SafariDriverCommandExecutor extends DriverCommandExecutor {
     public SafariDriverCommandExecutor(DriverService service) {
```
krmahadevan: This is now an un-used constructor. Please have it removed

krmahadevan:
@code-with-abdullah -  I have logged some additional comments which would need to be addressed.
Also we would need to add tests for each of the browser flavours that would demonstrate this capability working. I was able to add a [test for chrome easily in my PR](https://github.com/SeleniumHQ/selenium/pull/11545/files#diff-7cefce0686c9a114a0d7f44552c1f4bedc2beaae19ca58b0722dc37431556095). But when I was attempting to do the same for Firefox I was facing issues (Its mainly due to my lack of knowledge on the JS world). 

Maybe @diemol can help guide here in terms of how to go about adding functional tests that will test timeouts for the browsers and for which all browsers would we need 'em.

code-with-abdullah:
> @code-with-abdullah - I have logged some additional comments which would need to be addressed. Also we would need to add tests for each of the browser flavours that would demonstrate this capability working. I was able to add a [test for chrome easily in my PR](https://github.com/SeleniumHQ/selenium/pull/11545/files#diff-7cefce0686c9a114a0d7f44552c1f4bedc2beaae19ca58b0722dc37431556095). But when I was attempting to do the same for Firefox I was facing issues (Its mainly due to my lack of knowledge on the JS world).
> 
> Maybe @diemol can help guide here in terms of how to go about adding functional tests that will test timeouts for the browsers and for which all browsers would we need 'em.

Great. I shall integrate these changes and update you.

java/src/org/openqa/selenium/remote/RemoteWebDriver.java
diemol: @code-with-abdullah I agree with @krmahadevan, please revert the changes to this class.

diemol (changes requested):

code-with-abdullah:
@krmahadevan @diemol please review as all the requested changes till now have been integrated.

java/src/org/openqa/selenium/ie/InternetExplorerDriver.java
```diff
@@ -130,25 +135,30 @@ public InternetExplorerDriver(InternetExplorerDriverService service) {
...
   public static RemoteWebDriverBuilder builder() {
     return RemoteWebDriver.builder().oneOf(new InternetExplorerOptions());
   }
 
-  private void run(InternetExplorerDriverService service, Capabilities capabilities) {
+  private void run(InternetExplorerDriverService service, Capabilities capabilities, ClientConfig clientConfig) {
     assertOnWindows();
 
-    setCommandExecutor(new DriverCommandExecutor(service));
+    setCommandExecutor(new DriverCommandExecutor(service, capabilities, clientConfig));
```
krmahadevan: I see compilation errors here because we are invoking a protected constructor which is visible only to child classes. Please fix this.

krmahadevan: We should have added a new 2 arg constructor in `DriverCommandExecutor` that accepts a `DriverService` and a `ClientConfig` object which should be called from here.

krmahadevan:
@code-with-abdullah - There are compilation errors in this PR now. It would be good if you could please run `bazel build grid` locally in your machine before pushing the changes. Also please help add tests that would ensure that the changes are tested out as well.

code-with-abdullah:
Great. I will update you when the changes are ready.

titusfortner:
Updated with trunk and re-running CI

diemol:
Removing the label as in the TLC [meeting from January 19, 2023](https://www.selenium.dev/meetings/2023/tlc-01-19/) was decided that we want to move forward with this.

BilalKamran:
@krmahadevan @diemol added tests and merged trunk, pr please.

diemol:
Seems one of the users in the commits has not signed the CLA, please check.

BilalKamran:
> Seems one of the users in the commits has not signed the CLA, please check.

@diemol Resolved

java/test/org/openqa/selenium/chrome/ChromeDriverFunctionalTest.java
```diff
@@ -71,6 +71,18 @@ public void builderOverridesDefaultChromeOptions() {
     assertThat(localDriver.manage().timeouts().getImplicitWaitTimeout()).isEqualTo(Duration.ofMillis(1));
   }
 
+  @Test
+  @NoDriverBeforeTest
+  public void driverOverridesDefaultClientConfig() {
```
krmahadevan, line 76: @code-with-abdullah - Can we please refactor the test to look something like this ? 

```java
    org.assertj.core.api.Assertions.assertThatThrownBy(() -> {
      ClientConfig clientConfig = ClientConfig.defaultConfig().readTimeout(Duration.ofSeconds(0));
      localDriver = new ChromeDriver(ChromeDriverService.createDefaultService(),
        new ChromeOptions(), clientConfig);
    }).isInstanceOf(SessionNotCreatedException.class);
```

This is how the other tests in this class that expect exceptions have been written. So we will stick to the same approach. Please have the other tests also refactored on the same lines (Also please check for un-used imports in the files that you changed and kindly help get them removed as well)

Apart from this test case related observation everything else looks good.

krmahadevan (changes requested):

java/test/org/openqa/selenium/chrome/ChromeDriverFunctionalTest.java
code-with-abdullah: Great observation @krmahadevan. Will update and push as soon as possible.

BilalKamran:
@krmahadevan @diemol requested changes are done, pr please

krmahadevan (approved):
@diemol the changes look good to me. Please help take a look incase, I missed something.

BilalKamran:
@diemol please let us know if any change needed, we really hope to get this resolved and released with 4.9.

diemol (approved):
Thank you, @code-with-abdullah!

Thanks for you help, @krmahadevan!

code-with-abdullah:
Thanks a lot @diemol
Stay blessed :)

PR #11535, opened 2023-01-11
Split `rb/BUILD` into smaller packages

p0deje:
This fits more into Bazel approach because smaller packages with a proper dependency tree would allow to use Bazel caching more efficiently. For example, only few tests would be re-run considering you ran all integration tests for Chrome before and only changed code in `chrome/driver.rb`. This is because just a few tests depend on `chrome/driver.rb`. Per-test file packages also allow us to use Bazel flaky retry flags efficiently - previously it would re-run all the tests when one of them fails.

This commit also moves certain files around to make Ruby codebase more reasonable:
* CDP client code generator lives in the `rb/lib/selenium/devtools`
* Nightly version generator lives in a new `rb/lib/selenium/support`

Since now integration tests are being in parallel by default, few changes had to be done to make support code more concurrency-safe. Notably, Grid and Rack servers are being started on randomized ports. Some other tests were adapted/guarded to work better when run in parallel.

The test configuration is now controlled by a combination of `--define <name>=<value>` flags, supporting `browser/remote/headless` as names. See README.md changes for more details.

I've also recorded a small demo showing it working locally https://youtu.be/lqqXHEBvU0Y.


README.md
```diff
@@ -205,79 +205,83 @@ twine upload bazel-bin/py/selenium-*.whl bazel-bin/py/selenium-*.tar.gz
...
+|---------------------------------------------|---------------------------------------------------|
+| `bazel build //rb:selenium-devtools`        | Build selenium-devtools Ruby gem                  |
+| `bazel build //rb:selenium-webdriver`       | Build selenium-webdriver Ruby gem                 |
+| `bazel run //rb:selenium-devtools-release`  | Build and push selenium-devtools gem to RubyGems  |
+| `bazel run //rb:selenium-webdriver-release` | Build and push selenium-webdriver gem to RubyGems |
+| `bazel run //rb:console`                    | Start Pry REPL with all gems loaded               |
+| `bazel run //rb:docs`                       | Generate YARD docs                                |
+| `bazel run //rb:lint`                       | Run RuboCop linter                                |
+
+Test targets:
```
titusfortner, line 220: I don't love the duplication of bazel testing details in Contributing & Readme. Do you think it would make sense to put all testing details in Contributing, and link to those from the README?

WORKSPACE
```diff
@@ -329,9 +329,9 @@ pin_browsers()
 
 http_archive(
     name = "rules_ruby",
-    sha256 = "115e0d6af6fd4bc937507cf74a8090ddcec64aae0a88357b8cff93decd1a2e7c",
-    strip_prefix = "rules_ruby-5820e0e7dc4be38ec0b514450f4e942446022a28",
-    url = "https://github.com/p0deje/rules_ruby/archive/5820e0e7dc4be38ec0b514450f4e942446022a28.zip",
+    sha256 = "28fd318beb3761e0e0f38563aa5344fc13907c373d2f7785ac0a6b948a6a437a",
```
titusfortner, line 332: For my own edification, how is this generated?

rb/spec/integration/selenium/webdriver/bidi/browsing_context_spec.rb
```diff
@@ -53,7 +53,7 @@ class BiDi
           expect(browsing_context.id).not_to be_nil
         end
 
-        it 'can navigate to a url' do
+        it 'can navigate to a url', except: {browser: :chrome, reason: 'navigation_id is not nil'} do
```
titusfortner, line 56: Can we create a context `when navigating` for the 4 navigation methods so we only guard it in one place?

rb/spec/integration/selenium/webdriver/chrome/driver_spec.rb
```diff
@@ -52,15 +52,9 @@ module Chrome
...
-            @headless = ENV.delete('HEADLESS')
-            reset_driver!(args: ['--headless'])
-          end
+          before { reset_driver!(args: ['--headless']) }
 
-          after(:all) do
-            quit_driver
-            ENV['HEADLESS'] = @headless
-          end
+          after { reset_driver! }
```
titusfortner: I think we should keep this as `quit_driver` in case the next `before_all` creates its own custom driver. No reason to start a default only for it to be quit right away...

rb/spec/integration/selenium/webdriver/chrome/driver_spec.rb
```diff
@@ -91,16 +85,13 @@ module Chrome
         end
 
         describe '#logs' do
-          before(:all) do
+          before do
             reset_driver!(logging_prefs: {browser: 'ALL',
```
titusfortner, line 95: Do we need to start a new driver each time? Does navigation reset the logs and/or can we clear the logs instead? (same as other drivers).

rb/spec/integration/selenium/webdriver/spec_support/test_environment.rb
```diff
@@ -27,9 +27,9 @@ def initialize
           @create_driver_error = nil
           @create_driver_error_count = 0
 
-          populate_from_bazel_target
+          WebDriver.logger
```
titusfortner: this is currently in `before(:suite)` in `spec_helper` if it belongs here we should delete it there.

rb/spec/integration/selenium/webdriver/spec_support/test_environment.rb
```diff
@@ -152,6 +158,11 @@ def create_driver!(listener: nil, **opts, &block)
         private
 
         def build_options(args: [], **opts)
+          browser = self.browser
+          if browser.to_s == 'Safari Technology Preview'
```
titusfortner: Is driver not going to be `safari_preview` here? If so it should be set properly in `#safari_preview_options` called from line 168. Also, we really should move this logic to `SafariOptions` class like Java has it. Setting it on a class method seems kind of weird at this point.

titusfortner (changes requested):
All my suggestions are for non-bazel related changes in this PR. Overall this looks great, thanks for putting the work in.

README.md
p0deje: I'd do it the other way around (link to README from CONTRIBUTING). I suppose README is a good place to start for everyone.

WORKSPACE
p0deje: 1. Comment out `sha256`.
2. Run `bazel ...`
3. It would print SHA for you at the top of the output.
4. Use new value.

rb/spec/integration/selenium/webdriver/bidi/browsing_context_spec.rb
p0deje: The failures are different so I'm not sure they are related.
@TamsilAmani Any chance you know what these failure in Chrome are?

rb/spec/integration/selenium/webdriver/spec_support/test_environment.rb
p0deje: I'll try, maybe it's not needed anymore.

rb/spec/integration/selenium/webdriver/spec_support/test_environment.rb
p0deje: We read the browser name from the ENV variable, and it's expected to be `Safari Technology Preview`. I can use something else, but that's what https://github.com/SeleniumHQ/selenium/blob/ci-merge/rb/lib/selenium/webdriver/safari/options.rb expects.

rb/spec/integration/selenium/webdriver/bidi/browsing_context_spec.rb
titusfortner: I just noticed that all 4 of them called the navigation method

rb/spec/integration/selenium/webdriver/spec_support/test_environment.rb
titusfortner: Right, but we also have `WD_SPEC_DRIVER` which should already be toggling this.

rb/spec/integration/selenium/webdriver/spec_support/test_environment.rb
titusfortner: No, we're using the logger in TestEnvironment which is called before `before(:suite)` is called, so we want to make sure the logger is already initialized and has `ignore: :logger_info` set. You have it in the right place, we should move it over.

titusfortner: Oh, that's an uncommitted change I have locally. Looks like I deleted it at some point and didn't put it back. Yes, keep it here, but change it to: `WebDriver.logger.ignore(:logger_info)`

rb/spec/integration/selenium/webdriver/bidi/browsing_context_spec.rb
p0deje: I'd rather fix this in a separate commit - I suppose it could be useful to review the rest of BiDi specs too.

titusfortner: yeah, that's fine. All the non-bazel stuff probably could have been separate. :)

rb/spec/integration/selenium/webdriver/spec_support/test_environment.rb
p0deje: You are right, it's not needed anymore.

p0deje: It works a bit weird when used with `remote`, but I'd like to later make a separate configuration for a channel, something like this:

```
bazel test --define browser=safari --define channel=preview rb/spec/...
bazel test --define browser=chrome --define channel=beta rb/spec/...
bazel test --define browser=firefox --define channel=nightly rb/spec/...
```

titusfortner: that would require pinning browsers?

p0deje: Not necessarily, I need to try out first.

rb/spec/integration/selenium/webdriver/chrome/driver_spec.rb
p0deje: I've reverted, we don't need this.

rb/spec/integration/selenium/webdriver/chrome/driver_spec.rb
p0deje: Done!

p0deje:
@titusfortner Please take another look, I think I've addressed all your comments.

rb/spec/integration/selenium/webdriver/spec_support/test_environment.rb
titusfortner: what is weird with remote? isn't it just:
```
bazel test --define browser=safari-preview --define remote=true
```
Or do you mean the implementation where we're setting the value on the class?

titusfortner (approved):
Looks great, thanks!

ollie-iterators:
> Looks great, thanks!

Will all the errors be fixed by merging this?

rb/spec/integration/selenium/webdriver/spec_support/test_environment.rb
p0deje: I don't think the current implementation for `remote-safari-preview-test` works, but in this PR it does.

titusfortner:
It seems like the firefox tests are still running the edge/driver_spec logging tests, in spite of this line?
https://github.com/SeleniumHQ/selenium/blob/rb-split-build/rb/spec/integration/selenium/webdriver/edge/driver_spec.rb#L25

p0deje:
> It seems like the firefox tests are still running the edge/driver_spec logging tests, in spite of this line?
https://github.com/SeleniumHQ/selenium/blob/rb-split-build/rb/spec/integration/selenium/webdriver/edge/driver_spec.rb#L25

There is a problem in how `before(:all)` hooks are used together with guards - they are simply ignored. For now, I've switched the spec to use `before(:each)`.

p0deje:
Ruby CI is green - https://github.com/SeleniumHQ/selenium/actions/runs/3905159213

PR #11544, opened 2023-01-12
Overhaul CI [dotnet][java][js][py][rb][rust]

p0deje:
This PR introduces few major changes to the Selenium CI, overhauling
the current implementation to make it a bit more usable and maintainable.
1. Introduces high-level orchestrating "CI" workflow that is run on every push to the main branch, every 12 hours, every pull request, and manually. This workflow checks which bindings are affected by the pushed changes and calls individual CI workflows, presenting all of the results on a single page - https://github.com/SeleniumHQ/selenium/actions/workflows/ci.yml.
2. Reworks every `"CI - <Bindings>"` workflow to be callable from other workflows or run manually. The former is important for the "CI" workflow, the latter is matter of convenience so that individual bindings CI workflow can be run.
3. Individual bindings CI workflow are also implemented in a similar manner with the following pipeline: Build -> Documentation / Lint / Unit Tests -> Integration Tests This allows to better balance the load across the CI runners queue, preventing an immediate burst of queued runs.
4. Almost every CI pipeline step is implemented via a callable bazel.yml workflow. It is responsible for setting up the whole environment (browsers, Bazel caches) and greatly simplifies the individual bindings workflow. The only exceptions not using it are few JavaScript/Python tasks not migrated to Bazel and Rust pipeline.
6. Removes unnecessary installation of JDK because it's already taken care of by Bazel.
7. Allows to start interacting SSH session inside CI upon failure. This can be done by re-running a failed build with "Enable debug logging" on.
8. Allows to force-run CI on pull requests if the title includes `[<bindings>]` text. We could also fetch commits on this PR, but it needs extra code and can be done later.


AutomatedTester (approved):
LGTM

.github/workflows/bazel.yml
```diff
@@ -0,0 +1,106 @@
+name: Bazel
+
+on:
+  workflow_call:
+    inputs:
+      name:
+        description: Name of workflow
+        required: false
```
diemol, line 8: What happens when the name is not passed?

diemol (approved):
Just a minor comment, but everything else looks good. Thank you, @p0deje!

.github/workflows/bazel.yml
p0deje: It would use the default name of the workflow: `bazel`.

ollie-iterators:
I think that the Pull Request https://github.com/SeleniumHQ/selenium/pull/11546 would solve this issue as the "CI - Rust" action failed when it was first introduced.

titusfortner (approved):

PR #11563, opened 2023-01-17
[rb] Close BiDi session on closing the last top-level browsing context

TamsilAmani:
### Description

Close BiDi session on closing the last top-level browsing context

### Motivation and Context

Related to [7a469e0#commitcomment-95580594](https://github.com/SeleniumHQ/selenium/commit/7a469e02e14e44b0173b31b3b9baa1bfe57591ba#commitcomment-95616822)

### Types of changes

### Checklist

### DIFF
--- a/rb/lib/selenium/webdriver/common/driver.rb
+++ b/rb/lib/selenium/webdriver/common/driver.rb
@@ -187,7 +187,7 @@ module Selenium
         # If no top-level browsing contexts are open after calling close,
         # it indicates that the WebDriver session is closed.
         # If the WebDriver session is closed, the BiDi session also needs to be closed.
-        @bidi&.close if open_window_handles.empty?
+        @bidi&.close if open_window_handles&.empty?
         open_window_handles
       end
 
--- a/rb/lib/selenium/webdriver/common/driver.rb
+++ b/rb/lib/selenium/webdriver/common/driver.rb
@@ -188,6 +188,7 @@ module Selenium
         # it indicates that the WebDriver session is closed.
         # If the WebDriver session is closed, the BiDi session also needs to be closed.
         @bidi&.close if open_window_handles.empty?
+        open_window_handles
       end
 
       #
--- a/rb/lib/selenium/webdriver/common/driver.rb
+++ b/rb/lib/selenium/webdriver/common/driver.rb
@@ -71,6 +71,7 @@ module Selenium
       def initialize(bridge: nil, listener: nil, **opts)
         @service_manager = nil
         @devtools = nil
+        @bidi = nil
         bridge ||= create_bridge(**opts)
         add_extensions(bridge.browser)
         @bridge = listener ? Support::EventFiringBridge.new(bridge, listener) : bridge
@@ -174,6 +175,7 @@ module Selenium
       ensure
         @service_manager&.stop
         @devtools&.close
+        @bidi&.close
       end
 
       #
@@ -181,7 +183,11 @@ module Selenium
       #
 
       def close
-        bridge.close
+        open_window_handles = bridge.close
+        # If no top-level browsing contexts are open after calling close,
+        # it indicates that the WebDriver session is closed.
+        # If the WebDriver session is closed, the BiDi session also needs to be closed.
+        @bidi&.close if open_window_handles.empty?
       end
 
       #
--- a/rb/lib/selenium/webdriver/common/driver.rb
+++ b/rb/lib/selenium/webdriver/common/driver.rb
@@ -183,12 +183,10 @@ module Selenium
       #
 
       def close
-        open_window_handles = bridge.close
         # If no top-level browsing contexts are open after calling close,
         # it indicates that the WebDriver session is closed.
         # If the WebDriver session is closed, the BiDi session also needs to be closed.
-        @bidi&.close if open_window_handles&.empty?
-        open_window_handles
+        bridge.close.tap { |handles| @bidi&.close if handles&.empty? }
       end
 
       #
--- a/rb/spec/integration/selenium/webdriver/bidi_spec.rb
+++ b/rb/spec/integration/selenium/webdriver/bidi_spec.rb
@@ -23,7 +23,12 @@ module Selenium
   module WebDriver
     describe BiDi, only: {browser: %i[chrome edge firefox]} do
       before { reset_driver!(web_socket_url: true) }
-      after { quit_driver }
+
+      after do
+        quit_driver
+      rescue Selenium::WebDriver::Error::InvalidSessionIdError
+        # do nothing
+      end
 
       it 'gets session status' do
         status = driver.bidi.session.status
--- a/rb/spec/integration/selenium/webdriver/bidi_spec.rb
+++ b/rb/spec/integration/selenium/webdriver/bidi_spec.rb
@@ -52,6 +52,32 @@ module Selenium
           level: BiDi::LogInspector::LOG_LEVEL[:ERROR]
         )
       end
+
+      it 'does not close BiDi session if one window is closed' do
+        status = driver.bidi.session.status
+        expect(status).not_to be_nil
+        expect(status.message).to eq 'already connected'
+
+        driver.switch_to.new_window(:window)
+        driver.switch_to.new_window(:tab)
+        driver.switch_to.new_window(:tab)
+
+        driver.close
+
+        status_after_closing = driver.bidi.session.status
+        expect(status_after_closing).not_to be_nil
+        expect(status_after_closing.message).to eq 'already connected'
+      end
+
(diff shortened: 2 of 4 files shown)


rb/spec/integration/selenium/webdriver/bidi_spec.rb
```diff
@@ -52,6 +52,32 @@ module WebDriver
...
+        driver.switch_to.new_window(:tab)
+
+        driver.close
+
+        status_after_closing = driver.bidi.session.status
+        expect(status_after_closing.ready).to be false
+        expect(status_after_closing.message).to be_a String
+      end
+
+      it 'closes BiDi session if last window is closed' do
```
TamsilAmani, line 77: @titusfortner this tests works in chrome but gives error in firefox. And the error is not in the `it 'closes BiDi session if last window is closed'` block but in the `after` block where it calls quit_driver.

This is the error:
```
Selenium::WebDriver::Error::InvalidSessionIdError:
       Tried to run command without establishing a connection 
```
This is not happening in chrome. The after block works fine in it. But in firefox it gives the error. Can you suggest why?

p0deje: I would assume that Firefox closes the connection and the follow-up attempts to quit driver raise exceptions. Chrome might be more at ease and not raise anything at all.

It would help if you run the spec with `DEBUG=1` and paste the output.

TamsilAmani:
```
2023-01-21 14:30:17 DEBUG Selenium WebSocket -> {:method=>"session.status", :params=>{}, :id=>1}
2023-01-21 14:30:17 DEBUG Selenium WebSocket <- {"id"=>1, "result"=>{"ready"=>false, "message"=>"Session already started"}}
2023-01-21 14:30:17 INFO Selenium -> DELETE session/77e7158e-b1df-44f5-9f18-8ba0f7b55e3c/window
1674311417155	Marionette	INFO	Stopped listening on port 61474
console.error: Region.jsm: "Error fetching region" (new TypeError("NetworkError when attempting to fetch resource.", ""))
console.error: Region.jsm: "Failed to fetch region" (new Error("NO_RESULT", "resource://gre/modules/Region.jsm", 419))
2023-01-21 14:30:17 DEBUG Selenium    <<<  {"content-type"=>["application/json; charset=utf-8"], "cache-control"=>["no-cache"], "content-length"=>["12"], "date"=>["Sat, 21 Jan 2023 14:30:17 GMT"]}
2023-01-21 14:30:17 INFO Selenium <- {"value":[]}
2023-01-21 14:30:17 DEBUG Selenium WebSocket -> {:method=>"session.status", :params=>{}, :id=>2}
2023-01-21 14:30:17 INFO Selenium -> DELETE session/77e7158e-b1df-44f5-9f18-8ba0f7b55e3c
2023-01-21 14:30:17 DEBUG Selenium    <<<  {"content-type"=>["application/json; charset=utf-8"], "cache-control"=>["no-cache"], "content-length"=>["123"], "date"=>["Sat, 21 Jan 2023 14:30:17 GMT"]}
2023-01-21 14:30:17 INFO Selenium <- {"value":{"error":"invalid session id","message":"Tried to run command without establishing a connection","stacktrace":""}}
```

Here's what I think happens. When I call `driver.close` (line 77) with only one window open, firefox closes the window and calls quit to close the session. That's where we get `Stopped listening on port 61474` in the above log.

But after this, firefox again tries to `quit_driver` (irrespective if I put after block or not, maybe it does in the spec_helper) and that gives the last error in the log.

Chrome might be easy on this check as you said. In the sense that if it calls `quit_driver` when the session is already closed, it doesn't throw error.

A solution that I found is adding `create_driver!` after my expect statement (line 79). This creates a new driver and thus `quit_driver` works, both in firefox and chrome. 

Shall I add that?? @p0deje

p0deje: I think the simplest thing you could do is to rescue `Selenium::WebDriver::Error::InvalidSessionIdError` and ignore it in `after` hook.

rb/lib/selenium/webdriver/common/driver.rb
```diff
@@ -174,14 +175,20 @@ def quit
...
+        @bidi&.close
       end
 
       #
       # Close the current window, or the browser if no windows are left.
       #
 
       def close
-        bridge.close
+        open_window_handles = bridge.close
```
luke-hill: If you rewrite this as.
```
bridge.close.tap do |open_handles|
  @bidi&.close if open_handles&.empty?
end
```

It'll run slightly more optimally and be less LOC

rb/spec/integration/selenium/webdriver/bidi_spec.rb
```diff
@@ -52,6 +57,32 @@ module WebDriver
           level: BiDi::LogInspector::LOG_LEVEL[:ERROR]
         )
       end
+
+      it 'does not close BiDi session if one window is closed' do
```
luke-hill: do we want to hint the position here like the counter test does?

rb/spec/integration/selenium/webdriver/bidi_spec.rb
```diff
@@ -52,6 +57,32 @@ module WebDriver
...
+
+        driver.close
+
+        status_after_closing = driver.bidi.session.status
+        expect(status_after_closing.ready).to be false
+        expect(status_after_closing.message).to be_a String
+      end
+
+      it 'closes BiDi session if last window is closed' do
+        status = driver.bidi.session.status
```
luke-hill, line 78: Optional. Could maybe isolate all these status calls to a let block at the top.

luke-hill (commented):
Looks good for me locally. I've not done much testing here for a while so I'll let others approve.

rb/lib/selenium/webdriver/common/driver.rb
titusfortner: Return value is always an array (or it will error);
And Rubocop will want it with braces on one line:
```
bridge.close.tap { |handles| @bidi&.close if handles.empty? }
```

rb/spec/integration/selenium/webdriver/bidi_spec.rb
titusfortner: The real question is whether session should be memoized here: https://github.com/SeleniumHQ/selenium/blob/5ce55558cbadcbc6d4172dc51be98bf21287c1e4/rb/lib/selenium/webdriver/bidi.rb#L39-L40
@p0deje is there a reason we create a new `Session` instance for each call?

p0deje: Hm, I suppose we can memoize it, why not?

titusfortner: @TamsilAmani So, let's change line 40 there to `@session ||= Session.new(self)`
If that's memoized, then there's no advantage to using a `let` statement in the specs, multiple calls execute the desired method on the same instance.

rb/lib/selenium/webdriver/common/driver.rb
TamsilAmani: Thank you @luke-hill and @titusfortner for the suggestion. The suggested change works.

Also, @titusfortner as you say "_Return value is always an array (or it will error)_", that approach makes a unit test fail, which is...
https://github.com/SeleniumHQ/selenium/blob/trunk/rb/spec/unit/selenium/webdriver/support/event_firing_spec.rb#L148

with the error
```
NoMethodError:
       undefined method `empty?' for nil:NilClass
```
That's why I think nil value can be returned. And hence I should use `handles&.empty?`. Let me know if there's a better approach.

rb/spec/integration/selenium/webdriver/bidi_spec.rb
TamsilAmani: I don't get it @luke-hill. Can you please explain what _'hint the position'_ means?

rb/spec/integration/selenium/webdriver/bidi_spec.rb
TamsilAmani: Changes work as expected. Thanks!

rb/spec/integration/selenium/webdriver/bidi_spec.rb
luke-hill: The following spec hints at the position of the window `last window` should we / do we need to hint the position here? Does it matter? Just seems inconsistent.

titusfortner: Yeah, I don't think this matters.

titusfortner (approved):

Issue #11294, opened 2022-11-22
[🐛 Bug]: Selenium Manager use Proxy info from Browser Options


rupaagra:
### What happened?

**I have tried using Selenium manager feature in selenium version 4.6. On my company network selenium manager is unable to download browser driver.** 

Earlier I was using webdriver manager on my company network to download drivers, for that I have set proxy in webdriver manager like this :
WebDriverManager.chromedriver().proxy("proxy").setup(); and it was working fine with web driver manager.

**In selenium manager is there any way to set the proxy or how selenium manager going to work on networks behind proxy?** 

I have tested  same on my personal network which does not have any firewall and proxy restriction. Selenium manager is able to download browser drivers here and I am able to execute my script without setting browser exe path.

### How can we reproduce the issue?

```shell
I have tried with below codes on network behind proxy

1. WebDriver driver = new ChromeDriver()
   driver.get("https://google.com") 

2. Tried to set proxy in chrome options
ChromeOptions options = new ChromeOptions().addArguments("--proxy-server=<Proxy url>");
WebDriver driver = new ChromeDriver(options)
driver.get("https://google.com")
```

### Relevant log output

```shell
WARNING: Unable to obtain driver using Selenium Manager: Error running command: [C:\Users\Rupal\AppData\Local\Temp\1\selenium-manager1724451228000609043917808946591\selenium-manager.exe, --driver, chromedriver]
Build info: version: '4.6.0', revision: '79f1c02ae20'
System info: os.name: 'Windows 10', os.arch: 'amd64', os.version: '10.0', java.version: '1.8.0_172'
Driver info: driver.version: Driver
java.lang.IllegalStateException: The path to the driver executable The path to the driver executable must be set by the webdriver.chrome.driver system property; for more information, see https://chromedriver.chromium.org/. The latest version can be downloaded from https://chromedriver.chromium.org/downloads
                at org.openqa.selenium.internal.Require$StateChecker.nonNull(Require.java:314)
                at org.openqa.selenium.remote.service.DriverService.findExecutable(DriverService.java:148)
                at org.openqa.selenium.chrome.ChromeDriverService.access$000(ChromeDriverService.java:39)
                at org.openqa.selenium.chrome.ChromeDriverService$Builder.findDefaultExecutable(ChromeDriverService.java:233)
                at org.openqa.selenium.remote.service.DriverService$Builder.build(DriverService.java:450)
```

### Operating System

Windows 10

### Selenium version

4.6

[...]

PR #11575, opened 2023-01-20
[rust] Support for web proxy in Selenium Manager

bonigarcia:
### Description
This PR implements a couple of new flags for Selenium Manager:

- `--proxy <PROXY>`: HTTP(S) proxy for network connection (e.g., https://myproxy.net:8080)
- `--timeout <TIMEOUT>`: Timeout for network requests (in seconds). This timeout is applied from when the request starts connecting until the response body has finished [default: 120 seconds]

### Motivation and Context
This PR is part of the development of Selenium Manager M3 (requested in #11294).

### Types of changes

### Checklist


diemol (commented):
Not sure if this has been discussed elsewhere, but if a user is behind a proxy, do we need to invoke Selenium Manager in the bindings with the `--proxy` and `--timeout` flags? (Seems there is a default value for timeout, so the flag might not be needed).

If the bindings do need to invoke Selenium Manager with those flags, how will the bindings know if there is a proxy set? Should they detect it? Wouldn't it be easier if Selenium Manager detects that it is being a proxy so we do not write the same code across bindings?

bonigarcia:
The `--timeout` flag is not required for the proxy setup. I included it in this PR since I realized that this option deserves to be configurable during the implementation of this PR (e.g., for one of the implemented tests). But it could also be a separate PR.

Regarding proxy, as I imagine it (or, for example, in the way in which the proxy setup works in WebDrivermanager), the user is the one who configures the proxy URL. I don't know how to make this discovery automatic. But there are some cases where that is not possible, for the best of my knowledge, for example, for authenticated web proxies (e.g., http://user:pass@my-proxy.net).

Therefore, IMO the bindings should provide a way to configure the proxy URL, and when that info is specified by the user, the bindings invoke Selenium Manager using `--proxy <url>`.

diemol:
Right, authenticated proxies are tricky. I am not sure how we can set an authenticated proxy in the bindings these days. Probably we need to review that, and if a proxy is set, then pass it along?

bonigarcia:
I think that having the flag `--proxy` in Selenium Manager can be helpful in some cases. In addition, if an automatic proxy setup can be done (in the Rust side), it will help for sure. I don't know if it is possible in Rust, I am starting to search now.

diemol:
Yes, we should have the flag to set the proxy, I was just thinking if there is a way to centralize most of the code in Selenium Manager. Bindings have a way to set a proxy for the browser, not the HTTP connection, which is why I wonder how each binding is doing that bit. @titusfortner, do you know?

titusfortner:
I think we can/(need to?) use the same info that gets passed in to the driver with the Options instance, and we're already going to be parsing Options instance.

bonigarcia:
Yes, I think the optimal choice is to reuse the proxy specified by the user as options (i.e., `http://proxy:port` or `https://proxy:port`) to invoke Selenium Manager with the `--proxy URL` from the bindings.

For instance, in the Java bindings, when the user does the following:

```
        Proxy proxy = new Proxy();
        proxy.setHttpProxy("proxy:port");

        options.setProxy(proxy)
```

... internally, the Java bindings invoke Selenium Manager with `--proxy http://proxy:port`.

And when the proxy is secure:

```
        Proxy proxy = new Proxy();
        proxy.setSslProxy("proxy:port");

        options.setProxy(proxy)
```

... the call to Selenium Manager should be `--proxy https://proxy:port`. Is that right?

titusfortner:
This generally seems correct. I'm guessing we prefer ssl over http. 

Actually, I'm not a proxy person. Would http even work for downloading drivers since everything redirects to https these days?

diemol (approved):
Thank you, @bonigarcia!

PR #11586, opened 2023-01-23
Evaluate the whole PR when checking CI

p0deje:
Testing if this fixes CI.


p0deje:
Screenshot 2023-01-23 at 07 40 27

Seems to work correctly now.

Issue #11373, opened 2022-12-06
[🚀 Feature]: Process PATH by Selenium Manager


bonigarcia:
### Feature and motivation

Selenium Manager can discover if a driver is in the path and if the version is suitable to drive the browser. If the driver is not compatible (typically, since it is an old version), Selenium Manager should fix the problem (i.e., resolve the proper driver and ignore the driver in the `PATH`).

We can also log a `WARN` in Selenium Manager about it (something similar to mention, e.g., "_you have an old version of ... etc._"). That `WARN` trace should also be parsed in the bindings and logged to the final user again.

### Usage example

```
$ cargo run -- --browser chrome
WARN    Incompatible release of chromedriver (version 104.0.5112.79) detected in PATH (/home/boni/drivers/chromedriver)
INFO    /home/boni/.cache/selenium/chromedriver/linux64/106.0.5249.61/chromedriver
```

titusfortner:
fwiw, this is kind of a duplicate of #11356

titusfortner:
If we continue in the "fallback-only" mode, then we should only get a new version of the driver if it would otherwise fail (browserVersion - driverVersion > 1). We would also want to parse `--ignore-build-check` for chromium.

Alternately, we can move out of fallback mode and proactively update users to the latest version if any older version is found on PATH. I don't think we're ready for this, yet, though, and we should define what criteria need to be met to change modes.

titusfortner:
So, I'm going to close this since the Rust portion was already completed.

Also, fwiw, I've changed my mind about needing to process `--ignore-build-check`. If a user wants to use a specific, outdated version, they can pass it in directly and be responsible for it working or not working.

Issue #11356, opened 2022-12-03
[🚀 Feature]: Selenium Manager check drivers on PATH


titusfortner:
### Dependencies

* Rust:
    * #12177
    * #11639
* Grid: https://github.com/SeleniumHQ/selenium/commit/0013140fd587310ad38cffef9150bd8fc96b89ed
* Python: #12356
* Java: #12352 
* Ruby: #12345 
* JS: https://github.com/SeleniumHQ/selenium/commit/778cf32787097fba3b545b96a431bcf85e7fae31
* .NET: #12344

### Feature and motivation

Right now each of the bindings are processing drivers in this order:
1. User provided specific location
2. If driver isn't specified, it is looked for in each directory listed in PATH environment variable
3. If not provided and not found in PATH, tries to get Selenium Manager to find it

We've told users to try Selenium Manager by deleting drivers and stop using 3rd party managers. One issue we've seen is that there are users with old versions of drivers in directories on their PATH that they do not know about (#11269).

We can fix this, as well as dramatically increase the usage of Selenium Manager by having Selenium Manager check if a driver is on PATH, and checking its version; if the version is compatible, use that one; if it isn't, use Selenium Manager to download the right one (and raise a warning that an old one exists).

Bindings: because not everyone can use Selenium Manager, yet, these will still also need to check PATH, but will do so *after Selenium Manager errors, not before like now.

Note: I would love for Selenium bindings not to have to ever deal with PATH again, but we're a long way out from being able to use Selenium Manager for everything by itself.

### Considerations

1. Outdated driver is found on PATH
    * Raise Error — this was my initial suggestion for this feature in beta, and now I don't like it
    * Log Warning and use it — Based on "don't remediate" I think this is the right answer for now
    * Ignore it and download new one — this is what we are moving to
2. Selenium Manager Error
    * Bindings catch error and  try locating on PATH — This is safe, but probably overkill
    * Bindings no longer do PATH and users are directed to other fixes — This is easier

titusfortner:
Important to note that this will depend on #11351 so we know for sure which browser we need to be matching driver versions for.

titusfortner:
Ok, so all of the prereqs I discussed have been implemented...

### Current Behavior:

1. If user has specified location of driver, or driver exists on PATH - use it & ignore SM
2. Otherwise use Selenium Manager
3. Catch any exceptions
    * Throw Exception with the "old" error message to put on PATH

### Desired Behavior:

[...]

Issue #11357, opened 2022-12-03
[🚀 Feature]: Selenium Manager support more architectures


titusfortner:
### Feature and motivation

Quick check of what architectures are supported by each driver...
Which of these should we support?
Do we need to wait until we can get rust and/or bazel to cross compile?
It looks like cross-compilation with `cross` might be straightforward - https://kerkour.com/rust-cross-compilation 

| Architecture | chromedriver       | edgedriver            | geckodriver               | 
|--------------|--------------------|-----------------------|---------------------------|
| mac64        | :white_check_mark: | :white_check_mark:    | :white_check_mark:  10%   |
| linux64      | :white_check_mark: | :white_check_mark:    | :white_check_mark:  48%   |
| win32        | :white_check_mark: | :white_check_mark:    | :white_check_mark:   5%   |
| mac-arm64    | :white_check_mark: | :white_check_mark:    | :white_check_mark:  10%   | 
| linux-arm64  | :x:                | :x:                   | :black_square_button:  2% | 
| linux-32     | :x:                | :x:                   | :black_square_button:  1% | 
| win-arm64    | :x:                | :white_check_mark: | :white_check_mark:  2% |  
| win64        | :x:                | :white_check_mark:    | :white_check_mark:  28%   |

titusfortner:
Well, played around a bit, and I could do:
https://github.com/titusfortner/selenium/actions/runs/3609798451

Note this builds `x86_64-pc-windows-gnu` but not `x86_64-pc-windows-msvc`

per the [cross repo](https://github.com/cross-rs/cross):
> Additional Dockerfiles for other targets can be found in [cross-toolchains](https://github.com/cross-rs/cross-toolchains). These include MSVC and Apple Darwin targets, which we cannot ship pre-built images of.

I could not figure out how to get those working, though; someone who knows more about docker than how to spell it might get further.

ollie-iterators:
Is this related to the process of consolidating Selenium and Appium by making sure that Selenium can support the same architecture as Appium?

titusfortner:
No, this is just making sure that people using Selenium on these architectures are automatically obtaining the correct drivers if publicly available.

titusfortner:
@bonigarcia so this is essentially a rust question. Which of the drivers that get published is Selenium Manager looking for based on architecture? (e.g. does it always get win32 driver instead of the win64 geckodriver or edgedriver?)

I'm trying to figure out the different scenarios and what kinds of errors.

[...]

PR #11597, opened 2023-01-30
[rust] Selenium Manager processes PATH

bonigarcia:
### Description
With this PR, Selenium process the `PATH`, i.e., it tries to find the required driver (e.g., chromedriver 109) in the set of local directories where executable programs are located. If the driver is found but is not compatible, it shows a `WARN` trace (which can be captured by the bindings using the JSON output format). If found and compatible, it uses it. Here a couple of examples:

1. Found but incompatible (chromedriver 107 in the `PATH` in a computer with Chrome 109):
```
> cargo run -- --browser chrome
	 
WARN    Incompatible release of chromedriver (version 107.0.5304.62) detected in PATH: C:\Users\boni\Documents\bat\chromedriver.exe
INFO    C:\Users\boni\.cache\selenium\chromedriver\win32\109.0.5414.74\chromedriver.exe
```

2. Found and compatible (the same situation than before, but forcing to use Chrome 107):
```
> cargo run -- --browser chrome --browser-version 107

INFO    C:\Users\boni\Documents\bat\chromedriver.exe
```

### Motivation and Context
This PR implements #11356 (and #11373).

### Types of changes

### Checklist

### DIFF
--- a/rust/src/chrome.rs
+++ b/rust/src/chrome.rs
@@ -139,14 +139,11 @@ impl SeleniumManager for ChromeManager {
         } else {
             commands = vec![self.format_one_arg(WMIC_COMMAND, browser_path)];
         }
-        let (shell, flag, args) = if WINDOWS.is(self.get_os()) {
-            ("cmd", "/C", commands)
+        let (shell, flag) = self.get_shell_command();
+        let args = if WINDOWS.is(self.get_os()) {
+            commands
         } else {
-            (
-                "sh",
-                "-c",
-                vec![self.format_one_arg(DASH_DASH_VERSION, browser_path)],
-            )
+            vec![self.format_one_arg(DASH_DASH_VERSION, browser_path)]
         };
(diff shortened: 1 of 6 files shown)


titusfortner:
Excellent, this is going to make a big difference for users.

Right now in the bindings we are skipping calling the selenium manager if users pass in the location of the driver, set it with system property (Java) or on class variable (Ruby), or if the bindings find it on PATH. So with current bindings implementation, this feature won't be used. 

Ideally the bindings will be able to offload finding the driver on PATH to the manager entirely. We need to be cautious because this would transition the manager to an "opt-out" solution instead of "opt-in." The main feature we need to implement before we can make this transition is #11599. Though, we may want to avoid making this transition until #11357 even.

Note: none of this conversation prevents merging this code, just commenting that we have to more carefully think through how/when we implement it in bindings.

diemol (approved):

PR #11603, opened 2023-01-31
[rb] Fix autoload of `WebDriver::Remote::Bridge::COMMANDS`

etiennebarrie:
`COMMANDS` is defined on `WebDriver::Remote::Bridge`, but is defined as an autoload on `WebDriver::Remote`. Since it's used in `WebDriver::Remote::Bridge#commands` (with all the namespaces nested), Ruby will first look for the constant in `WebDriver::Remote::Bridge`, then `WebDriver::Remote`, where it finds the autoload. It requires the file, which defines `WebDriver::Remote::Bridge::COMMANDS`. Then Ruby emits a warning, because it expected `selenium/webdriver/remote/commands` to define `WebDriver::Remote::COMMANDS`, since that's what the autoload declares. After that, the constant can be resolved so it still works.

This properly declares the constant as autoloaded under `WebDriver::Remote::Bridge`.

### Description

A new warning was added to Ruby 3.2 which covers the issue where an autoload declares a constant, but upon loading the file, it doesn't get added.

### Motivation and Context

Remove warnings.

### Types of changes

### Checklist


p0deje:
Thank you @etiennebarrie !

Issue #11522, opened 2023-01-07
[🚀 Feature]: Selenium Manager support Safari & Technology Preview


titusfortner:
### Feature and motivation

Ok, I get this is kind of dumb, but looking at the Selenium implementations, it would be much easier not to have to treat Safari special. Should be pretty simple to hard code the locations and error if the driver isn't in those locations

### Usage example

```
selenium-manager --browser safari
selenium-manager --driver safaridriver
==> /usr/bin/safaridriver
```
and
```
selenium-manager --browser "Safari\ Technology\ Preview"
selenium-manager --browser SafariTechnologyPreview
selenium-manager --browser SafariTP
==> /Applications/Safari Technology Preview.app/Contents/MacOS/safaridriver
```

titusfortner:
Hmm, just noticed that Ruby doesn't hard code `/usr/bin/safaridriver` and just looks on PATH, and then only stores the location for Technology Preview

diemol:
I do not understand, what would do Selenium Manager for Safari? Can it be enabled through the command line without any user interaction? (I believe you need to enter your password to enable it, right?)

titusfortner:
My thought was that Selenium Manager could store the hard coded locations of the binaries so that we could avoid having 2 constructors — https://github.com/SeleniumHQ/selenium/pull/11491/files#diff-c9e7b1e932fcdfafa7fd1c615d09d8b411f2eabdff0fd1410c397b9aa5ed88f7R18

It's a minor thing, but allows us to treat all drivers the same. /shrug

diemol:
I see. Would then the bindings query Selenium Manager to get the binaries location? Maybe I do not see the whole picture but seems an over optimization.

titusfortner:
My thoughts for desired DriverFinder behavior:
1. If the user explicitly passes in a path for the driver in Service class (constructor/builder/setter depending on language) before driver constructor, then DriverFinder is never called
2. If DriverFinder is called it first checks for Property (Java only)
3. If no property is set/all other languages Selenium Manager:
    * Determines target browser/version (based on Options or defaults)
    * Checks PATH for valid driver (#11373); uses if valid, ignores if invalid
    * Gets list of valid versions & downloads correct version for use
    * If download unsuccessful and out of date version is present, it gets used as backup
 
[...]

PR #11609, opened 2023-02-01
[rust] Include support for Safari in Selenium Manager

bonigarcia:
### Description
This PR implements the support of Safari in Selenium Manager, as requested in #11522.

In addition to the stable release of Safari, and according to its [resources page](https://developer.apple.com/safari/resources/), there are two more distributions of Safari: beta, and _Technology Preview_. For symmetry with the rest of the browser (Chrome, Firefox, etc.), the "technology preview" is considered in Selenium Manager using the `dev` label (for browser version). The beta version is also considered.

Here it is some example executed in macOS (with Safari stable and Technology Preview installed):

```
bonigarcia@SL-2030 rust % cargo run -- --browser safari --debug     
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
     Running `target/debug/selenium-manager --browser safari --debug`
DEBUG	Using shell command to find out safari version
DEBUG	Running sh command: "/usr/libexec/PlistBuddy -c \"print :CFBundleShortVersionString\" /Applications/Safari.app/Contents/Info.plist"
DEBUG	Output { status: ExitStatus(unix_wait_status(0)), stdout: "16.1\n", stderr: "" }
DEBUG	The version of safari is 16.1
DEBUG	Required driver: safaridriver (local)
INFO	/usr/bin/safaridriver
```

```
bonigarcia@SL-2030 rust % cargo run -- --browser safari --debug --browser-version dev
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/selenium-manager --browser safari --debug --browser-version dev`
DEBUG	Using shell command to find out safari version
DEBUG	Running sh command: "/usr/libexec/PlistBuddy -c \"print :CFBundleShortVersionString\" /Applications/Safari\\ Technology\\ Preview.app/Contents/Info.plist"
DEBUG	Output { status: ExitStatus(unix_wait_status(0)), stdout: "16.4\n", stderr: "" }
DEBUG	The version of safari is 16.4
DEBUG	Required driver: safaridriver (local)
INFO	/Applications/Safari Technology Preview.app/Contents/MacOS/safaridriver
```

### Motivation and Context
Part of Selenium Manager M3, requested in #11522.

### Types of changes

### Checklist

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -38,6 +38,7 @@ use crate::logger::Logger;
 use crate::metadata::{
     create_browser_metadata, get_browser_version_from_metadata, get_metadata, write_metadata,
 };
+use crate::safaritp::SafariTPManager;
 
 pub mod chrome;
 pub mod config;
@@ -49,6 +50,7 @@ pub mod iexplorer;
 pub mod logger;
 pub mod metadata;
 pub mod safari;
+pub mod safaritp;
 
 pub const REQUEST_TIMEOUT_SEC: u64 = 120; // The timeout is applied from when the request starts connecting until the response body has finished
 pub const STABLE: &str = "stable";
@@ -255,7 +257,7 @@ pub trait SeleniumManager {
     }
 
     fn is_safari(&self) -> bool {
-        self.get_browser_name().eq(SAFARI)
+        self.get_browser_name().contains(SAFARI)
     }
 
     fn is_browser_version_unstable(&self) -> bool {
@@ -264,11 +266,6 @@ pub trait SeleniumManager {
             || browser_version.eq_ignore_ascii_case(DEV)
             || browser_version.eq_ignore_ascii_case(NIGHTLY)
             || browser_version.eq_ignore_ascii_case(CANARY)
-    }
-
-    fn is_browser_version_dev(&self) -> bool {
-        let browser_version = self.get_browser_version();
-        browser_version.eq_ignore_ascii_case(DEV)
     }
 
     fn resolve_driver(&mut self) -> Result<PathBuf, Box<dyn Error>> {
@@ -456,6 +453,15 @@ pub fn get_manager_by_browser(browser_name: String) -> Result<Box<dyn SeleniumMa
         Ok(IExplorerManager::new())
     } else if browser_name.eq("safari") {
         Ok(SafariManager::new())
+    } else if vec![
+        "safari technology preview",
+        r#"safari\ technology\ preview"#,
+        "safaritechnologypreview",
+        "safaritp",
+    ]
+    .contains(&browser_name_lower_case.as_str())
+    {
+        Ok(SafariTPManager::new())
     } else {
         Err(format!("Invalid browser name: {browser_name}"))
     }
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -20,6 +20,7 @@ use crate::edge::EdgeManager;
 use crate::files::compose_cache_folder;
 use crate::firefox::FirefoxManager;
 use crate::iexplorer::IExplorerManager;
+use crate::safari::{SafariManager, SAFARI};
 use std::fs;
 
 use crate::config::OS::WINDOWS;
@@ -47,6 +48,7 @@ pub mod firefox;
 pub mod iexplorer;
 pub mod logger;
 pub mod metadata;
+pub mod safari;
 
 pub const REQUEST_TIMEOUT_SEC: u64 = 120; // The timeout is applied from when the request starts connecting until the response body has finished
 pub const STABLE: &str = "stable";
@@ -57,6 +59,8 @@ pub const NIGHTLY: &str = "nightly";
 pub const WMIC_COMMAND: &str = r#"wmic datafile where name='{}' get Version /value"#;
 pub const WMIC_COMMAND_ENV: &str = r#"wmic datafile where name='%{}:\=\\%{}' get Version /value"#;
 pub const REG_QUERY: &str = r#"REG QUERY {} /v version"#;
+pub const PLIST_COMMAND: &str =
+    r#"/usr/libexec/PlistBuddy -c "print :CFBundleShortVersionString" {}/Contents/Info.plist"#;
 pub const DASH_VERSION: &str = "{} -v";
 pub const DASH_DASH_VERSION: &str = "{} --version";
 pub const ENV_PROGRAM_FILES: &str = "PROGRAMFILES";
@@ -160,11 +164,12 @@ pub trait SeleniumManager {
                     }
                     break;
                 }
-
-                metadata
-                    .browsers
-                    .push(create_browser_metadata(browser_name, &browser_version));
-                write_metadata(&metadata, self.get_logger());
+                if !self.is_safari() {
+                    metadata
+                        .browsers
+                        .push(create_browser_metadata(browser_name, &browser_version));
+                    write_metadata(&metadata, self.get_logger());
+                }
                 if !browser_version.is_empty() {
                     Some(browser_version)
                 } else {
@@ -179,12 +184,14 @@ pub trait SeleniumManager {
         if browser_version.is_empty() || self.is_browser_version_unstable() {
             match self.discover_browser_version() {
                 Some(version) => {
-                    self.get_logger().debug(format!(
-                        "Detected browser: {} {}",
-                        self.get_browser_name(),
-                        version
-                    ));
-                    self.set_browser_version(version);
+                    if !self.is_safari() {
+                        self.get_logger().debug(format!(
+                            "Detected browser: {} {}",
+                            self.get_browser_name(),
+                            version
+                        ));
+                        self.set_browser_version(version);
+                    }
                 }
                 None => {
                     if self.is_browser_version_unstable() {
@@ -247,6 +254,10 @@ pub trait SeleniumManager {
         }
     }
 
+    fn is_safari(&self) -> bool {
+        self.get_browser_name().eq(SAFARI)
+    }
+
     fn is_browser_version_unstable(&self) -> bool {
         let browser_version = self.get_browser_version();
         browser_version.eq_ignore_ascii_case(BETA)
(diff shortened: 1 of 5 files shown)


titusfortner:
1. Does beta come with its own driver? If so bindings don't support that right now, and should. Also would need to figure out what browser name that driver expects. I can check later when back at my laptop. 
2. Does manager support using technology preview when the browser name is "Safari Technology Preview?" that's how Apple differentiates it right now.

bonigarcia:
> * Does beta come with its own driver? If so bindings don't support that right now, and should. Also would need to figure out what browser name that driver expects. I can check later when back at my laptop.

No. I cannot install the current Safari beta in my macOS (since the beta is for Monterrey and mine is Ventura), but I investigated the content of the package of the current Safari beta (called `Safari16.3MoterreySeed.pkg`), and there is no `safaridriver` in it. I found only a different `safaridriver` the Technology Preview package.

> * Does manager support using technology preview when the browser name is "Safari Technology Preview?" that's how Apple differentiates it right now.

No. As I said, in Selenium Manager, for symmetry with the rest of the browsers, the "technology preview" is considered in Selenium Manager using the browser version `dev`. I don't like having a browser name called _Safari Technology Preview_ in Selenium Manager, at the same level of other browsers (Chrome, Firefox, etc.), independently that Apple calls it like that. It would broke the design of Selenium Manager. It does not worth it. And the only impact for the Selenium Manager users (the bindings, I suppose) is calling it through `--browser-version dev`.

titusfortner:
Except that is how selenium bindings differentiate it. Either the logic goes in the manager or it has to be created in each of the bindings, and removing the need to have special safari logic in the bindings was the whole point of adding this functionality to Selenium Manager in the first place.

titusfortner:
Hmm, maybe we can override the browser version in bindings as part of setting preview...

diemol:
Yes, the bindings differentiate Safari and Safari Technology preview, but I am not sure if that was a correct decision. The driver has the same name, just in a different location. 

I think it is fine to have it in Selenium Manager as `dev`, and we set that version in the bindings when the user wants Safari Technology Preview.

titusfortner:
Only Java has two separate service classes, the rest just toggle the location based on the options. 

Whatever we do, the goal is not to have special code for safari in any of the bindings. I think we need to understand how it will be implemented before merging this.

bonigarcia:
@titusfortner In your original request you mentioned the following for Safari TP:

```
selenium-manager --browser "Safari\ Technology\ Preview"
selenium-manager --browser SafariTechnologyPreview
selenium-manager --browser SafariTP
```

Instead of that, I am proposing to use the following to get the same result:

```
selenium-manager --browser safari --browser-version dev
```

titusfortner:
Right, I'm not sure we can make that work in all of the bindings without adding the custom code I want to avoid, but we can see.

diemol:
Thinking more about this, I am not even sure I agree on merging this. 

I understand the idea is to avoid the bindings hard coding the SafariDriver path location, but now we need to hard code `dev` somewhere in the bindings so Selenium Manager returns the path.

I remember the main idea of Selenium Manager was to download the drivers and in the future download the browsers. I perceive this type of work is not adding much value to the user and it is slowing us down in terms of reaching the point where we want to start working on browser download.

titusfortner:
Was just looking at how the Grid chooses Technology Preview, and it also does it based on Browser name:

```
07:00:32.012 INFO [NodeOptions.report] - Adding Chrome for {"browserName": "chrome"} 12 times
07:00:32.013 INFO [NodeOptions.report] - Adding Firefox for {"browserName": "firefox"} 12 times
07:00:32.014 INFO [NodeOptions.report] - Adding Safari for {"browserName": "safari"} 1 times
07:00:32.015 INFO [NodeOptions.report] - Adding Safari Technology Preview for {"browserName": "Safari Technology Preview"} 1 times
07:00:32.016 INFO [NodeOptions.report] - Adding Edge for {"browserName": "MicrosoftEdge"} 12 times
```

I could change the bindings code to set "dev" as the browser version when someone specifies STP, but we'd also have to muck about in the grid code.

If we want to support Safari Technology Preview, Selenium Manager should treat STP as a separate browser the same way that Grid does.

diemol:
My point was around having Selenium Manager doing this. The user still needs to download and install the browsers. And it is not that the bindings are having an extremely complicated logic to handle this. 

Selenium Manager is supposed to stop the user from doing manual setups, but for STP they will still need to do it anyway.

titusfortner:
The reason for my request was so that the common logic for managing all driver locations exists in the Selenium Manager instead of in the bindings. The bindings don't need [conditionals](https://github.com/SeleniumHQ/selenium/pull/11491/files#diff-d80f780d144d7576486e0cd17e33c72868bd8bc6b30fe904d2fe12f8c75ca735R89-R99) or to [store constants](https://github.com/SeleniumHQ/selenium/blob/trunk/rb/lib/selenium/webdriver/safari.rb#L31-L33), the DriverFinder pings Selenium Manager with browser name, and Selenium Manager verifies the driver exists and provides the location.

diemol:
OK, I guess we can go back and forth for a while with this. We need to find a way to merge this so we can move forward. 

@bonigarcia would it be too complicated to add STP as a browser in Selenium Manager so we centralize this logic there?

After this, let's check https://github.com/orgs/SeleniumHQ/projects/5/views/1 and see which ones are a "must" before we dig into browser downloads.

bonigarcia:
@diemol: As you know, Safari differs from the other browsers since its driver is shipped with the browser itself. This difference impacts the internal design of the Selenium Manager. You can see in the current PR that I needed to include several conditions to do differently for Safari (`if !self.is_safari()`).

Moreover, the current API exposed by Selenium Manager (which will be used as a crate, theoretically) has the methods `get_manager_by_browser` and `get_manager_by_driver` to get a manager by browser and driver name, respectively. My concern was the second method since both Safari and SafariTP have the same driver name (`safaridriver`).  Besides, I don't particularly like to have two top-level Safari browsers (meaning, at the same level that Chrome-Edge-Firefox) since I said before, Safari is an anomaly w.r.t. to the rest of browsers (for Selenium Manager).

For all these reasons, and because SafariTP is basically a dev version of Safari (currently SafariTP is 16.4 and Safari is 16.1, and there is a [beta](https://developer.apple.com/safari/resources/) in the middle), I decided to create the PR like it is now, i.e., reusing the `--browser-version dev` flag for SafariTP.

Being all this said, and answering your question, yes, it is perfectly possible to add another browser (SafariTP) in addition to Safari (i.e., a new browser with a different name). And it is not particularly difficult regarding the required code. But even including the SafariTP as a browser, the `get_manager_by_driver` will not work for SafariTP (i.e., the name `safaridriver` will be used only to identify Safari, and not SafariTP).

I can submit a new patch to this PR, and decide which option is better  (i.e., two Safari browsers, or only one, using the `dev` version for SafariTP).

bonigarcia:
Some output examples with the last patch:

```
bonigarcia@SL-2030 rust % cargo run -- --browser safari --debug                       
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/selenium-manager --browser safari --debug`
DEBUG	Using shell command to find out safari version
DEBUG	Running command: "/usr/libexec/PlistBuddy -c \"print :CFBundleShortVersionString\" /Applications/Safari.app/Contents/Info.plist"
DEBUG	Output: "16.1"
DEBUG	The version of safari is 16.1
DEBUG	Required driver: safaridriver (local)
INFO	/usr/bin/safaridriver
```

```
bonigarcia@SL-2030 rust % cargo run -- --browser SafariTP --debug
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/selenium-manager --browser SafariTP --debug`
DEBUG	Using shell command to find out safaritp version
DEBUG	Running command: "/usr/libexec/PlistBuddy -c \"print :CFBundleShortVersionString\" /Applications/Safari\\ Technology\\ Preview.app/Contents/Info.plist"
DEBUG	Output: "16.4"
DEBUG	The version of safaritp is 16.4
DEBUG	Required driver: safaridriver (local)
INFO	/Applications/Safari Technology Preview.app/Contents/MacOS/safaridriver
```

```
bonigarcia@SL-2030 rust % cargo run -- --browser SafariTechnologyPreview --debug 
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/selenium-manager --browser SafariTechnologyPreview --debug`
DEBUG	Using shell command to find out safaritp version
DEBUG	Running command: "/usr/libexec/PlistBuddy -c \"print :CFBundleShortVersionString\" /Applications/Safari\\ Technology\\ Preview.app/Contents/Info.plist"
DEBUG	Output: "16.4"
DEBUG	The version of safaritp is 16.4
DEBUG	Required driver: safaridriver (local)
INFO	/Applications/Safari Technology Preview.app/Contents/MacOS/safaridriver
```

```
bonigarcia@SL-2030 rust % cargo run -- --browser "Safari\ Technology\ Preview" --debug
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/selenium-manager --browser 'Safari\ Technology\ Preview' --debug`
DEBUG	Using shell command to find out safaritp version
DEBUG	Running command: "/usr/libexec/PlistBuddy -c \"print :CFBundleShortVersionString\" /Applications/Safari\\ Technology\\ Preview.app/Contents/Info.plist"
DEBUG	Output: "16.4"
DEBUG	The version of safaritp is 16.4
DEBUG	Required driver: safaridriver (local)
INFO	/Applications/Safari Technology Preview.app/Contents/MacOS/safaridriver
```

titusfortner:
That output above is exactly what will help the selenium code.

So, fyi, Selenium bindings are going to stop using `get_manager_by_driver` approach entirely. Everything is going to come from Browser Options including: browserName, browserVersion, proxy, `goog:chromeOptions: binary`, `moz:firefoxOptions: binary`.

I kind of wish from the Selenium/Apple side we managed Safari Technology Preview as a `browserVersion` toggle instead of a `browserName` difference, but someone made a decision many years ago and we're still stuck with it. 😄

diemol (approved):

PR #11623, opened 2023-02-07
更新 options.py

71n9:

```python
options.timeouts = {"pageLoad": 60}
# ValueError: Timeout keys can only be one of the following: implicit, pageLoad, script
```
I want to set the pageLoad timeout
When I execute the above code, I receive an error

### Description

A small change

### Motivation and Context

You can set successfully after modifying the logic

### Types of changes

### Checklist


diemol:
You set it in the driver after creating it: `driver.set_page_load_timeout(30)`, for example.

71n9:
> You set it in the driver after creating it: `driver.set_page_load_timeout(30)`, for example.

```python
# Part of the code of my program

options.add_experimental_option("debuggerAddress", f"127.0.0.1:{self.port}")
options.timeouts = {"pageLoad": 60000}
try:
   self.chromeDriver = webdriver.Chrome(service=Service(ChromedriverPath), options=options)
except Exception as err:
    print(err) # Message: timeout: Timed out receiving message from renderer: 300.000

```
image

I'm trying to solve this problem
I must set the timeout only before receiving this error, or I will have to wait 300 seconds to continue execution

diemol:
That is different, you seem to be having issues creating the session. First try without this
`options.add_experimental_option("debuggerAddress", f"127.0.0.1:{self.port}")`, and then debug from that point.

If you have more questions, please use the following channels https://www.selenium.dev/support/

71n9:
> 
Running for a long time will cause chrome to get stuck


> 那是不一样的，你好像在创建会话时遇到问题。首先尝试不使用this `options.add_experimental_option("debuggerAddress", f"127.0.0.1:{self.port}")`，然后从那一点开始调试。
> 
> 如果您有更多问题，请使用以下渠道https://www.selenium.dev/support/

The session has been running for a long time Running for a long time will cause chrome to get stuck
All GUI functions of Chrome are invalid The manual shutdown procedure also fails
I have to force the end of chrome and restart the remote binding
Each re-link must wait 300 seconds

titusfortner:
@diemol we should encourage setting timeout values in options instead of after session starts.

@symonk can you check this logic? I can't tell what we support now vs what this code allows.

diemol:
I see. I was checking the docs and saw something else. Good to know.

titusfortner:
It's one of the things added with w3c that probably needs to be updated in our docs in multiple places.

71n9:
The matter involved in the end should be very complex
Should lose a few hairs
thank you for all you do Thank you for open source software

py/selenium/webdriver/common/options.py
```diff
@@ -127,7 +127,7 @@ def timeouts(self, timeouts: dict) -> None:
 
         :param timeouts: values in milliseconds for implicit wait, page load and script timeout
         """
-        if all(x in timeouts.keys() for x in ("implicit", "pageLoad", "script")):
+        if all(x in ("implicit", "pageLoad", "script") for x in timeouts.keys()):
```
symonk, line 130: Thanks, I think this is a slightly better implementation than what existed before (Enforcing every timeout was provided else giving an ambiguous error) to now making sure all keys are in the set list, this is ok for now.  In future I think we should probably extract the 'known' keys and throwaway the rest without causing an error).

symonk (approved):

symonk:
Thanks, I think this is a slightly better implementation than what existed before (Enforcing every timeout was provided else giving an ambiguous error) to now making sure all keys are in the allowed sequence, this is ok for now.  In future I think we should probably extract the 'known' keys and throwaway the rest without causing an error).

PR #11629, opened 2023-02-07
Set Docker container locale env vars using capability

jsa34:

### Description

Set locale environment variable using capability "se:locale" (necessary for Linux environments due to Chrome's inconsistent setting of this across platforms - see https://developer.chrome.com/docs/extensions/reference/i18n/#how-to-set-browsers-locale)

### Motivation and Context

Allows setting this variable consistently in the dynamic grid upon a new session request.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/grid/node/docker/DockerSessionFactory.java
+++ b/java/src/org/openqa/selenium/grid/node/docker/DockerSessionFactory.java
@@ -309,6 +309,14 @@ public class DockerSessionFactory implements SessionFactory {
       envVars.put("SE_SCREEN_WIDTH", String.valueOf(screenResolution.get().getWidth()));
       envVars.put("SE_SCREEN_HEIGHT", String.valueOf(screenResolution.get().getHeight()));
     }
+    Optional<String> locale = ofNullable(getLocale(sessionRequestCapabilities));
+    locale.ifPresent(l -> envVars.putAll(
+      Map.of(
+        "LANG", String.format("%s.UTF-8", l),
+        "LANGUAGE", l,
+        "LC_ALL", String.format("%s.UTF-8", l)
+      )
+    ));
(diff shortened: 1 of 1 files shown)


jsa34:
@diemol - my assumed fix didn't work for setting the language env variable in the binary wrapper of the selenium docker project - it set the value in the node container, but the dynamically created containers still have the default locale (en-US).

My gut is telling me this is a better, more reliable approach and is akin to the way the timezone is able to be set/overridden.

Any feedback would be much appreciated as this is very new to me.

jsa34:
Also, I couldn't see if we had tests for this area, so didn't add any tests

diemol:
OK, so it worked when you used it in standalone mode but not in Dynamic. How did you start the Standalone mode and how did you start the test?

jsa34:
So... I realised after re-testing why it wasn't working after the release that it was because I was setting environment variables in the ansible script I was using for deploying the node containers for LANGUAGE (and LANG and LC_ALL) (hence the env variables were as expected!), so made a rookie error QA-ing, despite being a QA-er!  Apologies.

diemol:
> So... I realised after re-testing why it wasn't working after the release that it was because I was setting environment variables in the ansible script I was using for deploying the node containers for LANGUAGE (and LANG and LC_ALL) (hence the env variables were as expected!), so made a rookie error QA-ing, despite being a QA-er! Apologies.

Sorry, I am not following. Does it mean it is working now?

jsa34:
No - in short, my previous fix didn't work.

Looking at how the dynamic grid creates the node containers, though, I thought this would be a more suitable place to change how the environment variables are set.  (I *think* this is how the dynamic node containers are generated)

diemol:
@jsa34 I made the change where you suggested it, thank you for digging into it and giving pointers where to fix it. Now the parent container will use its `LANGUAGE` env var to set the language in the child ones. 

I thought about having it as a capability but this use case is not common and creates more maintenance.

PR #11640, opened 2023-02-09
[java] Synchronize method to get Selenium Manager binary (fix #11620)

### LINKED ISSUE #11620: [🐛 Bug]: Selenium Manager sometimes fails in parallel test run
### What happened?

Have noticed that sometimes  Selenium Manager fails to configure driver executable path if tests are run in parallel.

Environment: 
Java: Oracle OpenJDK 17.0.2
Gradle: 7.4

Was able to reproduce it both with jUnit and TestNG, please see the code in the repository below.

Before running tests, make sure neither specific driver locations provided, nor PATH environment variable set for drivers.
Run the tests until you `get java.lang.IllegalStateException: The path to the driver executable...` exception.

```shell
https://github.com/attila-fazekas/selenium-manager-parallel-run
```

### Relevant log output

```shell
https://github.com/attila-fazekas/selenium-manager-parallel-run/tree/main/logs
```

### Operating System

macOS Mojave

### Selenium version

Java 4.8

Chrome  109.0.5414.119, Firefox 108.0.1, Edge 109.0.1518.70

ChromeDriver 109.0.5414.74, GeckoDriver 0.32.0, EdgeDriver 109.0.1518.70

No

bonigarcia:
### Description
The method to get the Selenium Manager binary in the Java bindings is not synchronized, which leading to concurrent problems to get Selenium Manager in Java, as reported in #11620. I have confirmed the bug using the provided [example repo](https://github.com/attila-fazekas/selenium-manager-parallel-run). I used that repo as well to test the fix.

### Motivation and Context
This PR fixes #11620.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/manager/SeleniumManager.java
+++ b/java/src/org/openqa/selenium/manager/SeleniumManager.java
@@ -117,7 +117,7 @@ public class SeleniumManager {
    * Determines the correct Selenium Manager binary to use.
    * @return the path to the Selenium Manager binary.
    */
-    private File getBinary() {
+    private synchronized File getBinary() {
         if (binary == null) {
             try {
                 Platform current = Platform.getCurrent();


java/src/org/openqa/selenium/manager/SeleniumManager.java
```diff
@@ -117,7 +117,7 @@ private static String runCommand(String... command) {
    * Determines the correct Selenium Manager binary to use.
    * @return the path to the Selenium Manager binary.
    */
-    private File getBinary() {
+    private synchronized File getBinary() {
```
krmahadevan, line 120: @bonigarcia - This would cause all threads to go through the process of `obtain lock, check binary variable, release lock`. We could perhaps instead consider implement a doubly checked locking mechanism ?

Something like this ?

```java
  private File getBinary() {
    if (binary != null) {
      return binary;
    }
    synchronized (this) {
      if (binary != null) {
        return binary;
      }
      try {
        Platform current = Platform.getCurrent();
        String folder = "linux";
        String extension = "";
        if (current.is(WINDOWS)) {
          extension = EXE;
          folder = "windows";
        } else if (current.is(MAC)) {
          folder = "macos";
        }
        String binaryPath = String.format("%s/%s%s", folder, SELENIUM_MANAGER, extension);
        try (InputStream inputStream = this.getClass().getResourceAsStream(binaryPath)) {
          Path tmpPath = Files.createTempDirectory(SELENIUM_MANAGER + System.nanoTime());
          File tmpFolder = tmpPath.toFile();
          tmpFolder.deleteOnExit();
          binary = new File(tmpFolder, SELENIUM_MANAGER + extension);
          Files.copy(inputStream, binary.toPath(), REPLACE_EXISTING);
        }
        binary.setExecutable(true);
      } catch (Exception e) {
        throw new WebDriverException("Unable to obtain Selenium Manager", e);
      }
    }
    return binary;
  }
```

krmahadevan:
@bonigarcia - I think there's a larger problem here.

* `org.openqa.selenium.manager.SeleniumManager` has been built to be a singleton.
* The current implementation (as in `4.8.0`) right now is such that, it attempts to locate the `binary` data member ONLY once. This means that lets say I have 2 test methods say `firefoxTest()` and `chromeTest()` then the following will happen if I run them in parallel

**Assumptions** (Only this will kick in the SeleniumManager into action)

* `chromedriver` and `geckodriver` are NOT available in my `PATH`
* I am not specifying the location of `chromedriver` and `geckodriver` via the JVM arguments.

* `firefoxTest()` will cause the [binary](https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/manager/SeleniumManager.java#L60) data member in the `SeleniumManager` singleton to be set.
* Now because of [this edit check](https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/manager/SeleniumManager.java#L121) we will end up returning the path of `geckodriver` to `chromeTest` This will be wrong.

I think the fix should be to do something like below (I have changed `SeleniumManager` to be an enum since enums are first class singletons)

```java

public enum SeleniumManager {
    manager;
    private static final Logger LOG = Logger.getLogger(SeleniumManager.class.getName());

    private static final String SELENIUM_MANAGER = "selenium-manager";
    private static final String EXE = ".exe";
    private static final String INFO = "INFO\t";

    //We cache the binary path on a per driver flavor basis.
    private final Map<String, File> binaries = new ConcurrentHashMap<>();

    SeleniumManager() {
        Runtime.getRuntime().addShutdownHook(new Thread(() -> binaries.values()
          .stream()
          .filter(Objects::nonNull)
          .filter(File::exists)
          .forEach(SeleniumManager::deleteQuietly)));
    }

  private static void deleteQuietly(File binary) {
    try {
      Files.delete(binary.toPath());
    } catch (IOException e) {
      LOG.warning(String.format("%s deleting temporal file: %s",
        e.getClass().getSimpleName(), e.getMessage()));
    }

  }

    public SeleniumManager getInstance() {
        return manager;
    }

  private File getBinary(String driverName) {
    return this.binaries.computeIfAbsent(driverName, flavor -> {
      try {
        Platform current = Platform.getCurrent();
        String folder = "linux";
        String extension = "";
        if (current.is(WINDOWS)) {
          extension = EXE;
          folder = "windows";
        } else if (current.is(MAC)) {
          folder = "macos";
        }
        String binaryPath = String.format("%s/%s%s", folder, SELENIUM_MANAGER, extension);
        try (InputStream inputStream = this.getClass().getResourceAsStream(binaryPath)) {
          Path tmpPath = Files.createTempDirectory(SELENIUM_MANAGER + System.nanoTime());
          File tmpFolder = tmpPath.toFile();
          tmpFolder.deleteOnExit();
          File binary = new File(tmpFolder, SELENIUM_MANAGER + extension);
          Files.copy(inputStream, binary.toPath(), REPLACE_EXISTING);
          binary.setExecutable(true);
          return binary;
        }
      } catch (Exception e) {
        throw new WebDriverException("Unable to obtain Selenium Manager", e);
      }
    });
  }

    public String getDriverPath(String driverName) {
        if (!ImmutableList.of("geckodriver", "chromedriver", "msedgedriver", "IEDriverServer").contains(driverName)) {
            throw new WebDriverException("Unable to locate driver with name: " + driverName);
        }

        String driverPath = null;
        File binaryFile = getBinary(driverName);
        if (binaryFile != null) {
          driverPath = runCommand(binaryFile.getAbsolutePath(),
                    "--driver", driverName.replaceAll(EXE, ""));
        }
        return driverPath;
    }
}

```

bonigarcia:
> * Now because of [this edit check](https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/manager/SeleniumManager.java#L121) we will end up returning the path of `geckodriver` to `chromeTest` This will be wrong.

That is wrong. That binary path is the Selenium Manager binary, not the driver. The Selenium Manager binary is used to manage (i.e., download, cache, etc.) the drivers (geckodriver, chromedriver, etc.).

krmahadevan:
> > * Now because of [this edit check](https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/manager/SeleniumManager.java#L121) we will end up returning the path of `geckodriver` to `chromeTest` This will be wrong.
> 
> That is wrong. That binary path is the Selenium Manager binary, not the driver. The Selenium Manager binary is used to manage (i.e., download, cache, etc.) the drivers (geckodriver, chromedriver, etc.).

Ah! Yeah now I realise my blunder. The `SeleniumManager` internally relies on the rust binary to do the actual download. Sorry for the confusion @bonigarcia

java/src/org/openqa/selenium/manager/SeleniumManager.java
bonigarcia: Synchronization at the method level will do the trick as well.

krmahadevan: No. Its different. Method level synchronisation will cause threads to obtain lock to enter into the `getBinary()` method. A doubly checked lock backed implementation will ensure that only those threads that are going to initialise the value of `binary` will take the lock. All the subsequent threads will return without having to take the lock.

bonigarcia: I agree it's different, but the difference is minimum, IMO. What I mean with my comment is that the corrurency problems detected are solved with the synchronization at the method level.

titusfortner:
Is the issue getting the binary or executing the binary? I'm a little confused because of the singleton; does synchronizing that method actually prevent two threads from attempting to execute the binary at the same time? When does the first thread relinquish the lock?

diemol (approved):
LGTM!

Failing tests are not related to this change.

PR #11661, opened 2023-02-15
[java] Selenium Manager use binary from browser options (#11351)

bonigarcia:
### Description
This PR updates the Java binding to reuse the browser binary path from options (e.g. `ChromeOptions`) to call Selenium Manager using the flag `--browser-path`.

### Motivation and Context
This PR is the Java implementation of #11351.

NOTE: To test this feature, the Selenium Manager binary (located in `common\manager` should be updated).

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/chrome/ChromeDriverService.java
+++ b/java/src/org/openqa/selenium/chrome/ChromeDriverService.java
@@ -110,10 +110,12 @@ public class ChromeDriverService extends DriverService {
     File executable,
     int port,
     List<String> args,
-    Map<String, String> environment) throws IOException {
+    Map<String, String> environment,
+    List<String> seleniumManagerArgs) throws IOException {
     super(executable, port, DEFAULT_TIMEOUT,
       unmodifiableList(new ArrayList<>(args)),
-      unmodifiableMap(new HashMap<>(environment)));
+      unmodifiableMap(new HashMap<>(environment)),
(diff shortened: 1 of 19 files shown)


titusfortner:
This is not the approach we want.
I already have an implementation that will support this — https://github.com/SeleniumHQ/selenium/pull/11491

We just need to figure out how to get Bazel happy.

bonigarcia:
I was not aware of that draft. I did this PR since @diemol asked me to help with #11351.

titusfortner:
Sorry for not having better communication. Diego probably didn't realize how much the implementation will be based on the DriverFinder. My first implementation looked similar to yours, but I got a lot of feedback and had to change it a few times (I still don't love where we ended up because the setter feels hacky, but everyone else seems to think it's fine 😄).

The way Firefox implements `getBinary()` doesn't really work for what we need, though because it was written based on the legacy firefox driver, so I think we're going to need a new method for this in Firefox & Chromium (maybe `getBrowserBinary()`), then I think we're going to need a new Interface for it so we can make sure the options class actually implements it before trying to call it (since IE & Safari can't use it and we don't want to have to write separate implementations).

It's very frustrating I have all these implementations stored in my brain, and I'm effectively unable to work on Selenium code right now. 😦

diemol:
Sorry about this! 

Closing since it is already done.

PR #11663, opened 2023-02-16
[java] Process Selenium Manager output as JSON

bonigarcia:
### Description
This PR makes the Java bindings calls Selenium Manager using the flag `--output json` (implemented in #11531). Then, the output is parsed using with GSON. From the parsed output, the logs messages from Selenium Manager with `WARN` level are displayed to the user (using `LOG.warning` in Java).

### Motivation and Context
This PR is the Java implementation for #11365.

NOTE: To test this feature, the Selenium Manager binary (located in `common\manager`) should be updated.

### Types of changes

### Checklist


java/src/org/openqa/selenium/manager/SeleniumManager.java
```diff
@@ -18,6 +18,7 @@
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.io.CharStreams;
+import com.google.gson.GsonBuilder;
```
diemol: I believe we can use this class implemented by Simon to handle JSON payloads. 

https://github.com/SeleniumHQ/selenium/blob/572434ee4988ba64d57c0fc6bb491c2a8224747d/java/src/org/openqa/selenium/json/Json.java#L30

diemol (changes requested):
I made changes so [Java uses Selenium Manager](https://github.com/SeleniumHQ/selenium/commit/8b4b818bcd8b9b507e4c43e899ad9ab95d3a6c62) to run tests, do you think we can add a test for this? It can be just with the INFO.

java/src/org/openqa/selenium/manager/SeleniumManager.java
bonigarcia: I have just committed a second patch to this PR to use that internal JSON parser instead of GSON. I tested locally and it works in the same way than GSON.

bonigarcia:
> I made changes so [Java uses Selenium Manager](https://github.com/SeleniumHQ/selenium/commit/8b4b818bcd8b9b507e4c43e899ad9ab95d3a6c62) to run tests, do you think we can add a test for this? It can be just with the INFO.

Any Java test using Selenium Manager is going to use this JSON parsing.

What do you mean by "just with the INFO"?

bonigarcia:
@diemol This PR had conflicts, but I have just resolved it. Can it be merged?

diemol (approved):
Yes, thank you!

java/src/org/openqa/selenium/manager/SeleniumManager.java
```diff
@@ -112,9 +113,11 @@ private static String runCommand(String... command) {
             throw new WebDriverException("Unsuccessful command executed: " + Arrays.toString(command) +
                     "\n" + output);
         }
-
-        String[] lines = output.split("\n");
-        return lines[lines.length -1].replace(INFO, "").trim();
+        SeleniumManagerJsonOutput jsonOutput = new Json().toType(output,
+                SeleniumManagerJsonOutput.class);
+        jsonOutput.logs.stream().filter(log -> log.level.equalsIgnoreCase(WARN))
+                .forEach(log -> LOG.warning(log.message));
```
titusfortner, line 119: @bonigarcia / @diemol can we log the debug information to `LOG.FINE` here? Is this the right place?

diemol: Yes, we can do that.

diemol:
Seems we missed to update the Selenium Manager binary before merging this.

PR #11668, opened 2023-02-17
Allow search for regular expression in expected condition

robertschulze:
This also returns the match(es) to be used further e.g. by evaluating match groups.

### Description
This allows to work with regular expressions in matching expected conditions, specifically the text within an element. It could clearly be extended to further cases. 

### Motivation and Context
By allowing to use regular expressions, 1. the matching of text becomes much more versatile and 2. it allows to work with the content of the matched groups after the match was successful. In this way e.g. status messages on websites can be matched and checked.

### Types of changes

### Checklist

### DIFF
--- a/py/selenium/webdriver/support/expected_conditions.py
+++ b/py/selenium/webdriver/support/expected_conditions.py
@@ -265,6 +265,22 @@ def text_to_be_present_in_element_attribute(locator, attribute_, text_):
     return _predicate
 
 
+def regex_to_be_present_in_element(locator, regexp):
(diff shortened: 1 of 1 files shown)


diemol:
Thank you for this PR, however we tend to not add new logic to the expected conditions class, rather we recommend to add those to a framework built on top. #10589 is an example of this case. Exceptions of this is when we accept PRs [extending existing logic](https://github.com/SeleniumHQ/selenium/pull/9881) or [implementing a condition that exists in another language binding already](https://github.com/SeleniumHQ/selenium/pull/7406).

PR #11718, opened 2023-03-01
Allow LocalNode to be extended for customisation

krmahadevan:
LocalNode currently has a private constructor 
Which prevents it from being extended for customisation. Exposing the constructor for classes extending it.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/grid/node/local/LocalNode.java
+++ b/java/src/org/openqa/selenium/grid/node/local/LocalNode.java
@@ -134,7 +134,7 @@ public class LocalNode extends Node {
   private final AtomicInteger pendingSessions = new AtomicInteger();
   private final AtomicInteger sessionCount = new AtomicInteger();
 
-  private LocalNode(
+  protected LocalNode(
     Tracer tracer,
     EventBus bus,
     URI uri,


diemol (commented):
What is the advantage? Why not keep it the way it is? Folks can still extend `Node`, right?

krmahadevan:
> What is the advantage? Why not keep it the way it is? Folks can still extend `Node`, right?

@diemol for most of the scenarios, a user should be perfectly fine in just extending `Node` and then wrapping a `LocalNode` within this extended class (like what we show in the documentation as well) but I noticed a user does not get calls when the session is quit using this approach because:

1. A decorated node (A class that essentially extends `Node` but internally delegates calls to a `LocalNode` object) when it encounters a `delete session` request via `executeWebDriverCommand` will `executeWebDriverCommand()` on the local node object that it has (as seen [here](https://github.com/krmahadevan/demo-node/blob/main/src/main/java/com/rationaleemotions/DecoratedLoggingNode.java#L61) )
2. The implementation in `LocalNode.executeWebDriverCommand()` is such that, it checks if the command is a delete session and it internally calls `LocalNode.stop(SessionId id)`. See [here](https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/grid/node/local/LocalNode.java#L473-L475)
3. Now this call from (2) never invokes the `stop(SessionId id)` on our Decorated node.

Owing to this, a user would never be able to eavesdrop on `stop(Session)` via a decorated node approach. So they will have to duplicate code from `LocalNode` completely to be able to tap into stop session events.

If we allowed users to extend `LocalNode` then they can easily just override the `stop(Session)` method in their child class and then include their custom logic as well.

Please let me know if that answers your question.

diemol (approved):
Makes sense, thank you!

Issue #11690, opened 2023-02-20
[🚀 Feature]: Different ways to configure Selenium Manager


bonigarcia:
### Feature and motivation

If the Selenium Manager configuration file is implemented (#11689), there will be two ways to configure Selenium Manager. In addition to those, an additional way to specify some common configuration parameters (e.g., the proxy server configuration) can be using *environment variables*. 

In case of implementing the three ways, a order preference should be defined. For instance, as follows:

1. CLI parameters. The configuration specified as CLI parameters are honored in the first place.
2. Configuration file. As defined in #11689.
3. Environment variables. To name these variables, the CLI parameter names can be converted using each parameter name to uppercase and replacing the dash symbol (`-`) with an underscore (`_`) and prepening `SE_`. For example, `browser-ttl` becomes `SE_BROWSER_TTL`.

### Usage example

n/a

diemol:
Implemented via #11810 
Thank you, @bonigarcia!

PR #11722, opened 2023-03-01
[rust] Selenium Manager parse mirror to get latest IEDriverServer version (#11672)

### LINKED ISSUE #11672: [🚀 Feature]: Selenium Manager uses proper IEDriverSever version
### Feature and motivation

Currently, Selenium Manager uses the latest version of IEDriverServer when using Internet Explorer. Besides, the internal Rust logic assumes that the latest [release of Selenium](https://github.com/SeleniumHQ/selenium/releases/latest) contains a version of IEDriverServer. This can pose a potential problem if a Selenium release does not ship the IEDriverServer binaries. To avoid this issue, there is an open issue in Selenium (#11280) to publish this info. When this information is available, the Selenium Manager logic should read it to use the most suitable IEDriverServer version.

### Usage example

The use of Selenium Manager will be the same to resolve IEDriverServer:

```
selenium-manager --browser ie
```

bonigarcia:
### Description
Currently, Selenium Manager assumes that the latest [release of Selenium](https://github.com/SeleniumHQ/selenium/releases/latest) contains a version of IEDriverServer. This can pose a potential problem if a Selenium release does not ship the IEDriverServer binaries. To avoid this potential problem, this PR uses the [mirrored responses of the GH API](https://raw.githubusercontent.com/SeleniumHQ/selenium/trunk/common/mirror/selenium) to discover the latest IEDriverServer version.

### Motivation and Context
This PR closes #11672.

### Types of changes

### Checklist

### DIFF
--- a/rust/src/mirror.rs
+++ b/rust/src/mirror.rs
@@ -1,0 +1,41 @@
+// Licensed to the Software Freedom Conservancy (SFC) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The SFC licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use crate::downloads::read_content_from_link;
+use reqwest::Client;
+use serde::{Deserialize, Serialize};
+use std::error::Error;
+
+pub const MIRROR_URL: &str =
+    "https://raw.githubusercontent.com/SeleniumHQ/selenium/trunk/common/mirror/selenium";
+
+#[derive(Serialize, Deserialize)]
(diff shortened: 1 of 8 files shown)


rust/src/mirror.rs
```diff
@@ -0,0 +1,41 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+use crate::downloads::read_content_from_link;
+use reqwest::Client;
+use serde::{Deserialize, Serialize};
+use std::error::Error;
+
+pub const MIRROR_URL: &str =
+    "https://raw.githubusercontent.com/SeleniumHQ/selenium/trunk/common/mirror/selenium";
```
diemol, line 24: Is there a way to configure this URL? I can imagine some folks who cannot access this URL, so they might find a way to "mirror the mirror" and consume the data from there.

bonigarcia: In the current PR, no, that value is hardcoded and cannot be changed by the user. 

But I agree that would be desirable. My [proposal for rich configuration](https://github.com/SeleniumHQ/selenium/issues/11690) is three levels: CLI parameters, configuration file, and environment variable. 

For this specific URL, it could be easy to include a CLI flag to configure that URL (e.g., `--mirror-url <url>`) in Rust. But that would imply changing the bindings to pass that URL from options.

IMO, implementing a configuration file/envs mechanism in Rust could ease global configuration capabilities.

diemol: Well, we also have the URLs for ChromeDriver and others in the same way. We should probably leave it as it is and revisit later.

diemol (approved):

Issue #11689, opened 2023-02-20
[🚀 Feature]: Configuration file for Selenium Manager


bonigarcia:
### Feature and motivation

To centralize all the Selenium Manager configuration values, maybe creating a global configuration file.

TOML can be a simple format for the configuration file. Selenium Manager will look for this file at `~/.cache/selenium/selenium-manager-config.toml`. An example of the content of this file is as follows:

```
browser = "chrome"
driver = "chromedriver"
browser-version = "106"
driver-version = "106.0.5249.61"

browser-path = "/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome"
os = "macos"
arch = "x64"
proxy = "myproxy:8080"
browser-ttl = 0
driver-ttl = 86400
driver-mirror = "http://my-chromedriver-mirror.net"
cache-path = "/my/custom/cache"

debug = false
```

The default location of this file could be change using a new flag, e.g. `--config-file <FILE_PATH>`.

### Usage example

```
selenium-manager --browser chrome --config-file /my/custom/sm-config.toml
```

titusfortner:
I don't see a good way for the bindings to use this feature. 

Maybe if we cache the manager in a specific directory, the manager can look in that directory for the config file and we can direct users to put it there?

bonigarcia:
The way I see it, the config file is a way to specify a common setup for all the calls to Selenium Manager. It is a way to avoid using the same configuration again and again per test (e.g., the proxy is set up once and reused in every single call to Selenium Manager). But of course, it is not mandatory.

titusfortner:
Right now we're only pulling things out of the Options class, so there's no way for a user to pass in a location of a config file. Maybe it makes sense to start adding custom capabilities to the Options classes to support Selenium Manager.

diemol:
I do not think this is needed because Selenium Manager needs to be transparent to the bindings user, which is the main use case. This might be useful for Grid users, but again, that is a smaller segment of people and still not sure how many would actually use it. 

I will close this as I believe we do not want this, and if we see use cases for it, we can definitely reopen.

titusfortner:
Actually, I do see potential value here. I suspect that people might want more control than what we're going to give them when we make Selenium Manager opt-out, and either we implement special `se:xxx` Options for Options classes or we support putting a config file on the system somewhere. I guess we can re-open the issue when people ask for it. 😄

PR #11810, opened 2023-03-23
[rust] Support different ways to configure Selenium Manager (config file, envs)

bonigarcia:
### Description
This PR implements #11690 and #11689. I know the latter was closed because it was not unwanted at this point, but for a similar effort, and because I honestly think it might help some users out there, I have implemented both.

So far, the Selenium Manager configuration capabilities have been done through CLI arguments. For instance, the flag `--proxy <PROXY>` allows configuring the proxy for the network connection. Nevertheless, this has the inconvenience that to make this configuration possible, all the bindings languages should implement a mechanism (e.g., through browser options) to make this feature available for users. For instance, the configuration before (i.e., the proxy) is not possible in Selenium Manager at the time of this writing since it has yet to be implemented in the bindings.

This PR enhances the configuration capabilities by providing the procedure proposed in #11690:

1. CLI parameters. 
2. Configuration file.
3. Environment variables.

IMO, option two can be interesting for global configuration in a local machine (e.g., a developer laptop). Option three can be interesting for global configuration in build servers (e.g., Jenkins, GitHub Actions, etc.). Both of them allow a rapid configuration (since it bypasses the need to implement it in the bindings).

Here are some examples of configuration:

a) For specifying a fixed version of geckodriver:
a.1) With config file:
```
boni@slimbook:~/.cache/selenium$ cat selenium-manager-config.toml 
geckodriver-version = "0.30.0"


boni@slimbook:~/dev/selenium/rust$ cargo run -- --browser firefox --debug
    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
     Running `target/debug/selenium-manager --browser firefox --debug`
DEBUG	Running command: "geckodriver --version"
DEBUG	Output: ""
DEBUG	Driver URL: https://github.com/mozilla/geckodriver/releases/download/v0.30.0/geckodriver-v0.30.0-linux64.tar.gz
INFO	/home/boni/.cache/selenium/geckodriver/linux64/0.30.0/geckodriver
```

a.2) With envs:
```
boni@slimbook:~/dev/selenium/rust$ SE_GECKODRIVER_VERSION=0.30.0 cargo run -- --browser firefox --debug
    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
     Running `target/debug/selenium-manager --browser firefox --debug`
DEBUG	Running command: "geckodriver --version"
DEBUG	Output: ""
DEBUG	Driver URL: https://github.com/mozilla/geckodriver/releases/download/v0.30.0/geckodriver-v0.30.0-linux64.tar.gz
INFO	/home/boni/.cache/selenium/geckodriver/linux64/0.30.0/geckodriver
```

b) For specifying a proxy:

b.1) With config file:
```
boni@slimbook:~/.cache/selenium$ cat selenium-manager-config.toml 
proxy = "myproxy:8080"
```

b.2) With envs:
```
SE_PROXY=myproxy:8080 ./selenium-manager --browser chrome
```

### Motivation and Context
This PR is part of M3 (rich configuration) in the Selenium Manager roadmap.

### Types of changes

### Checklist

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -28,6 +28,7 @@ tar = "0.4.38"
 infer = "0.13.0"
 exitcode = "1.1.2"
 is_executable = "1.0.1"
+toml = "0.7.3"
 
 [dev-dependencies]
 assert_cmd = "2.0.8"
--- a/rust/src/chrome.rs
+++ b/rust/src/chrome.rs
@@ -49,17 +49,17 @@ pub struct ChromeManager {
 }
 
 impl ChromeManager {
-    pub fn new() -> Box<Self> {
+    pub fn new() -> Result<Box<Self>, String> {
         let default_config = ManagerConfig::default();
         let default_timeout = default_config.timeout.to_owned();
         let default_proxy = default_config.proxy.to_owned();
-        Box::new(ChromeManager {
+        Ok(Box::new(ChromeManager {
             browser_name: CHROME_NAME,
             driver_name: CHROMEDRIVER_NAME,
             config: default_config,
-            http_client: create_http_client(default_timeout, default_proxy),
+            http_client: create_http_client(default_timeout, default_proxy)?,
             log: Logger::default(),
-        })
+        }))
     }
 }
 
--- a/rust/src/chrome.rs
+++ b/rust/src/chrome.rs
@@ -50,13 +50,15 @@ pub struct ChromeManager {
 
 impl ChromeManager {
     pub fn new() -> Result<Box<Self>, String> {
-        let default_config = ManagerConfig::default();
-        let default_timeout = default_config.timeout.to_owned();
-        let default_proxy = default_config.proxy.to_owned();
+        let browser_name = CHROME_NAME;
+        let driver_name = CHROMEDRIVER_NAME;
+        let config = ManagerConfig::default(browser_name, driver_name);
+        let default_timeout = config.timeout.to_owned();
+        let default_proxy = config.proxy.to_owned();
         Ok(Box::new(ChromeManager {
-            browser_name: CHROME_NAME,
-            driver_name: CHROMEDRIVER_NAME,
-            config: default_config,
+            browser_name,
+            driver_name,
+            config,
             http_client: create_http_client(default_timeout, default_proxy)?,
             log: Logger::default(),
         }))
--- a/rust/src/chrome.rs
+++ b/rust/src/chrome.rs
@@ -30,9 +30,9 @@ use crate::metadata::{
     create_driver_metadata, get_driver_version_from_metadata, get_metadata, write_metadata,
 };
 use crate::{
-    create_default_http_client, format_one_arg, format_two_args, SeleniumManager, BETA,
-    DASH_DASH_VERSION, DEV, ENV_LOCALAPPDATA, ENV_PROGRAM_FILES, ENV_PROGRAM_FILES_X86,
-    FALLBACK_RETRIES, NIGHTLY, REG_QUERY, STABLE, WMIC_COMMAND, WMIC_COMMAND_ENV,
+    create_http_client, format_one_arg, format_two_args, SeleniumManager, BETA, DASH_DASH_VERSION,
+    DEV, ENV_LOCALAPPDATA, ENV_PROGRAM_FILES, ENV_PROGRAM_FILES_X86, FALLBACK_RETRIES, NIGHTLY,
+    REG_QUERY, STABLE, WMIC_COMMAND, WMIC_COMMAND_ENV,
 };
 
 pub const CHROME_NAME: &str = "chrome";
@@ -50,11 +50,14 @@ pub struct ChromeManager {
 
 impl ChromeManager {
     pub fn new() -> Box<Self> {
+        let default_config = ManagerConfig::default();
+        let default_timeout = default_config.timeout.to_owned();
+        let default_proxy = default_config.proxy.to_owned();
         Box::new(ChromeManager {
             browser_name: CHROME_NAME,
             driver_name: CHROMEDRIVER_NAME,
-            config: ManagerConfig::default(),
-            http_client: create_default_http_client(),
+            config: default_config,
+            http_client: create_http_client(default_timeout, default_proxy),
             log: Logger::default(),
         })
     }
--- a/rust/src/config.rs
+++ b/rust/src/config.rs
@@ -33,13 +33,15 @@ use toml::Table;
 pub const ARM64_ARCH: &str = "arm64";
 pub const CONFIG_FILE: &str = "selenium-manager-config.toml";
 pub const ENV_PREFIX: &str = "SE_";
+pub const VERSION_PREFIX: &str = "-version";
+pub const PATH_PREFIX: &str = "-path";
 
 pub struct ManagerConfig {
     pub browser_version: String,
     pub driver_version: String,
+    pub browser_path: String,
     pub os: String,
     pub arch: String,
-    pub browser_path: String,
     pub proxy: String,
     pub timeout: u64,
     pub browser_ttl: u64,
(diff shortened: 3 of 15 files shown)


bonigarcia:
Just a note to clarify the naming of the configuration keys.

In the configuration file (`~/.cache/selenium/selenium-manager-config.toml`), each configuration key has the same name as the CLI arguments, such as `browser-version`, `driver-version`, `browser-path`, `driver-ttl`, `browser-ttl`, `proxy`, etc. Moreover, there are specific configuration keys for browser versions, driver versions, and browser paths, such as:

- `chrome-version`, `chromedriver-version`, and `chrome-path`
- `firefox-version`, `geckodriver-version`, and `firefox-path`
- `edge-version`, `msedgedriver-version`, and `edge-path`

All of these configuration keys have their equivalence in environmental variables by converting each parameter name to uppercase, replacing the dash symbol (`-`) with an underscore (`_`), and adding the prefix`SE_`. Some examples:

`SE_DRIVER_TTL`, `SE_GECKODRIVER_VERSION`, `SE_FIREFOX_PATH`, `SE_PROXY`, `SE_DEBUG`, etc.

rust/Cargo.toml
```diff
@@ -28,6 +28,7 @@ tar = "0.4.38"
 infer = "0.13.0"
 exitcode = "1.1.2"
 is_executable = "1.0.1"
+toml = "0.7.3"
```
diemol, line 31: Does the binary grow too much with this new dependency?

rust/src/chrome.rs
```diff
@@ -49,14 +49,19 @@ pub struct ChromeManager {
...
-            browser_name: CHROME_NAME,
-            driver_name: CHROMEDRIVER_NAME,
-            config: ManagerConfig::default(),
-            http_client: create_default_http_client(),
+    pub fn new() -> Result<Box<Self>, String> {
+        let browser_name = CHROME_NAME;
+        let driver_name = CHROMEDRIVER_NAME;
+        let config = ManagerConfig::default(browser_name, driver_name);
+        let default_timeout = config.timeout.to_owned();
+        let default_proxy = config.proxy.to_owned();
```
diemol, line 57: Just checking, are these things related to the PR description? Should they be part of a different PR?

rust/src/config.rs
```diff
@@ -142,3 +158,74 @@ impl ARCH {
...
+    }
+}
+
+fn get_env_name(suffix: &str) -> String {
+    let suffix_uppercase: String = suffix.replace('-', "_").to_uppercase();
+    concat(ENV_PREFIX, suffix_uppercase.as_str())
+}
+
+fn get_config() -> Result<Table, Box<dyn Error>> {
+    let config_path = get_cache_folder().join(CONFIG_FILE);
```
diemol, line 223: This means that if the file is present, it will be used, right?

What has priority, env vars or file?

diemol (commented):
Left some comments.

rust/Cargo.toml
bonigarcia: The crate `toml` is quite small. I have just built the binaries using this PR branch (see [here](https://github.com/SeleniumHQ/selenium/actions/runs/4588075402)), and the resulting binaries are:

selenium-manager_linux-x64	4.31 MB
selenium-manager_macos-x64	3.59 MB
selenium-manager_windows-x64	3.6 MB

The current binaries are (see [here](https://github.com/SeleniumHQ/selenium/actions/runs/4545218225)):

selenium-manager_linux-x64	4.14 MB	
selenium-manager_macos-x64	3.45 MB	
selenium-manager_windows-x64	3.46 MB

... so the difference is around 0.15 MB per binary.

rust/src/chrome.rs
bonigarcia: These changes are required since with this PR, the proxy setup can be different because of the config in the files/envs.

rust/src/config.rs
bonigarcia: The config file is used when is present and the corresponding CLI parameter is not honored.

In other words, the order of preference is as follows:

1. CLI parameters.
2. Configuration file.
3. Environment variables.

bonigarcia:
I have just rebased this branch with trunk. If this PR is accepted, I recommend to merge it before the others (e.g. #11840), since it is going to create conflicts otherwise.

rust/src/config.rs
```diff
@@ -150,17 +159,24 @@ impl ARCH {
...
 
 impl StringKey<'_> {
     fn get_value(&self) -> String {
         let config = get_config().unwrap_or_default();
-        let key = self.0;
-        if config.contains_key(key) {
-            config[key].as_str().unwrap().to_string()
-        } else {
-            env::var(get_env_name(key)).unwrap_or(self.1.to_owned())
+        let keys = self.0.to_owned();
```
oriongonza, line 167: You don't need to clone the keys:
```rust
 let keys = &self.0;
 let mut result: String;
 for &key in keys {
```

Also what's the purpose of `.as_str().unwrap().to_string()` instead of `to_string()` directly?

diemol: What is the alternative to cloning the keys? Maybe you can suggest the code change?

oriongonza: I just did, in the comment block. 
I'm new to contributing to github, I messed something up: I wanted to highlight the block but only highlighted the line. I should have read [the docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request?tool=webui). My bad!

Anyways the change is to just take a reference to the `Vec`,  `let keys = &self.0;` and then dereference the key using pattern matching `for &key in keys`

same effect as 
```rust
        let keys = &self.0;
        for key in keys {
            if config.contains_key(*key) {
                result = config[*key].as_str().unwrap().to_string();
```

rust/src/config.rs
```diff
@@ -150,17 +159,24 @@ impl ARCH {
...
+        let mut result;
+        for key in keys {
+            if config.contains_key(key) {
+                result = config[key].as_str().unwrap().to_string()
```
oriongonza, line 171:
```suggestion
        let keys = &self.0;
        let mut result;
        for &key in keys {
            if config.contains_key(key) {
                result = config[key].to_string()
```

oriongonza (changes requested):

rust/src/config.rs
```diff
@@ -142,3 +158,74 @@ impl ARCH {
             .contains(&arch.to_ascii_lowercase().as_str())
     }
 }
+
+pub struct StringKey<'a>(pub Vec<&'a str>, pub &'a str);
+
+impl StringKey<'_> {
```
oriongonza, line 164: Every time StringKey is used it's just used to call get_key() on it. It might make more sense to just convert it to a function eg

```rust
pub struct Keys; 
impl Keys {
  get_string_key(key: &str, default: &str){...}
  }

**Comment by bonigarcia** [comment]
@dev-ardi Can you please send a separate PR for these improvements you mentioned? Thanks a lot!

**Review by diemol — APPROVED** [review]

File: rust/src/lib.rs [hunk 7]
```diff
@@ -508,17 +508,22 @@ pub fn clear_cache(log: &Logger) {
...
         .danger_accept_invalid_certs(true)
         .use_rustls_tls()
+        .timeout(Duration::from_secs(timeout));
+    if !proxy.is_empty() {
+        match Proxy::all(proxy) {
+            Ok(p) => {
+                client_builder = client_builder.proxy(p);
+            }
+            Err(err) => {
+                return Err(err.to_string());
```
oriongonza, line 522: @bonigarcia  What was the reason to return a String instead of an Error?

bonigarcia: That's because the final error messages shown as a result are string. But perhaps there is a more idiomatic way to handle this error. Please send a PR if you find a better way to do it.

Issue #11798, opened 2023-03-22
[🐛 Bug]: UnreachableBrowserException caused by "connection closed locally" exactly 3 minutes after session start when using JDK HttpClient


robbytx:
### What happened?

I recently started using `selenium-http-jdk-client` as a workaround for https://github.com/SeleniumHQ/selenium/issues/11750#issuecomment-1460580132, and I've noticed that tests which run longer than 3 minutes often fail with the following exception:
```
Exception in thread "main" org.openqa.selenium.remote.UnreachableBrowserException: Error communicating with the remote browser. It may have died.
Build info: version: '4.8.1', revision: '8ebccac989'
System info: os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '12.6.3', java.version: '11.0.15'
Driver info: org.openqa.selenium.remote.RemoteWebDriver
Command: [60c7ca2458c8472f824c2c43dbf1ac2a, sendKeysToElement {id=ca596158-ff3e-4df8-b4ac-2c00e60ca6d7, value=[Ljava.lang.CharSequence;@474e34e4}]
Capabilities {acceptInsecureCerts: false, browserName: chrome, browserVersion: 111.0.5563.65, chrome: {chromedriverVersion: 111.0.5563.41 (976ef12907ef..., userDataDir: C:\Program Files (x86)\scop...}, goog:chromeOptions: {debuggerAddress: localhost:49762}, networkConnectionEnabled: false, pageLoadStrategy: normal, platformName: WINDOWS, proxy: Proxy(), setWindowRect: true, strictFileInteractability: false, timeouts: {implicit: 0, pageLoad: 300000, script: 30000}, unhandledPromptBehavior: dismiss and notify, webauthn:extension:credBlob: true, webauthn:extension:largeBlob: true, webauthn:extension:minPinLength: true, webauthn:extension:prf: true, webauthn:virtualAuthenticators: true}
Element: [[RemoteWebDriver: chrome on WINDOWS (60c7ca2458c8472f824c2c43dbf1ac2a)] -> css selector: input[name='q']]
Session ID: 60c7ca2458c8472f824c2c43dbf1ac2a
	at org.openqa.selenium.remote.RemoteWebDriver.execute(RemoteWebDriver.java:567)
	at org.openqa.selenium.remote.RemoteWebElement.execute(RemoteWebElement.java:257)
	at org.openqa.selenium.remote.RemoteWebElement.sendKeys(RemoteWebElement.java:113)
[...]

PR #11816, opened 2023-03-25
[java] Handle redirects inside the JdkHttpClient

joerg1985:
This PR is related to #11798 UnreachableBrowserException caused by "connection closed locally" exactly 3 minutes after session start when using JDK HttpClient.

### Description
There seems to be a bug in the JDK related to redirects, see https://bugs.openjdk.org/browse/JDK-8304701
This PR will disable the redirects of the JDK HttpClient and add a custom handling of redirects.

Additionally this PR will:
 - ensure there is no downgrade of redirected POST requests to GET requests, see javadoc of java.net.http.HttpClient.Redirect
 - ensure there is no downgrade from a secure URL to an insecure URL

I am not sure if this will fix the original issue described in #11798, so there is additional testing needed.

### Types of changes

### Checklist


java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java
```diff
@@ -289,7 +288,42 @@ public HttpResponse execute(HttpRequest req) throws UncheckedIOException {
 
     BodyHandler<byte[]> byteHandler = BodyHandlers.ofByteArray();
     try {
-      return messages.createResponse(client.send(messages.createRequest(req), byteHandler));
+      URI rawUri = messages.getRawUri(req);
+
+      // This is a workaround to https://bugs.openjdk.org/browse/JDK-8304701
+      // Warning: When this workaround is removed and the j.n.h.HttpClient.Redirect.NORMAL is used
+      // the 301 and 302 redirects will downgrade a POST to a GET request. See the javadoc of
```
diemol: So this means that even if the bug is fixed in Java we need to keep the check to avoid request downgrades?

java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java
```diff
@@ -289,7 +288,42 @@ public HttpResponse execute(HttpRequest req) throws UncheckedIOException {
...
+      // j.n.h.HttpClient.Redirect for details about the downgrade.
+      for (int i = 0; i < 20; i++) {
```
diemol: Do you think that leaving this max 20 retry is fine? I am trying to think if there is a better way.

diemol (commented):
Thank you, left some comments.

java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java
joerg1985: I could not find any info what kind of redirects are used from providers like saucelabs.com.
If they (or others) are using 301 or 302 redirects the code is needed, if not the code can be removed.

java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java
joerg1985: The old netty client has a redirect limit set to 100 with a notice to the cloud providers (see commit 829e476). So i could set this to 100 to have the same value than the old netty client, the default limit of the http client is 5 so i think this is another reason why this code is needed.
I would suggest to update the comment why the custom redirect handing is needed and set limit to 100?

java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java
titusfortner: Sauce uses 302 if the user does not have sufficient sessions available and they request an additional one. Retries are sent until an available session is found or until a timeout.

titusfortner:
fyi, a mysterious @joerg1985 got a shout out in the State of the Union address at the Selenium Conference last week...
Thanks for your work!

java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java
diemol: You are 100% right, thank you.

diemol (approved):
Thank you, @joerg1985!

Artur-:
This commit causes our several of our Saucelabs based builds to fail when going from Selenium 4.8.3 to 4.9.0. Running with 4.9.0 + this reverted makes everything work again. 

The errors are timeouts creating drivers or `org.openqa.selenium.WebDriverException: Invalid message: ERROR Internal Server Error` or `  Caused by: org.openqa.selenium.SessionNotCreatedException: Could not start a new session. Handshake response does not match any supported protocol` (which might be that same but obfuscated because of #11927)

diemol:
Can you create an issue about this, please?

Artur-:
#12069

PR #11821, opened 2023-03-27
Remove "shadow_root" assertion in Python bindings for Firefox

whimboo:
The "Get Element ShadowRoot" command has been implemented in Firefox 96 (a year ago) and it should no longer cause an assertion.

### DIFF
--- a/common/mirror/selenium
+++ b/common/mirror/selenium
@@ -42,6 +42,9 @@
         "browser_download_url": "https://github.com/SeleniumHQ/selenium/releases/download/selenium-4.8.0/selenium-java-4.8.2.zip"
       },
       {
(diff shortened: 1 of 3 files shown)


whimboo:
This PR is ready for review. Thanks!

whimboo:
CC'ing @patrickangle and @gsnedders given that the same assertion still exists for Safari and as it looks like there might have been some ongoing work here on experimental.

diemol (approved):
Thank you, @whimboo!

gsnedders:
> CC'ing @patrickangle and @gsnedders given that the same assertion still exists for Safari and as it looks like there might have been some ongoing work here on experimental.

actually shipped in Safari 16.4 on Tuesday: https://wpt.fyi/results/webdriver/tests/get_element_shadow_root/get.py?run_id=5118756219518976&run_id=6320265787604992&run_id=6292883592708096

Issue #10949, opened 2022-08-16
[🐛 Bug]: Re-add logging endpoints to Grid


titusfortner:
### What happened?

When running on Grid, the logging endpoints return an error (all browsers except for Chrome).

```
org.openqa.selenium.UnsupportedCommandException: The command 'GET /session/167E1FE8-9D49-4B7A-94FB-D36B77AC51D1/se/log/types' was not found.
```

We should add those endpoints to Grid, and add anything received from the driver.

Adding the Session logs for grid would be a nice bonus, but this shouldn't error.

### How can we reproduce the issue?

```shell
Capabilities options = new SafariOptions();
driver = new RemoteWebDriver(new URL("http://localhost:4444/wd/hub"), options);
driver.manage().logs().getAvailableLogTypes();
```

### Operating System

n/a

### Selenium version

4.4.0

### What are the browser(s) and version(s) where you see this issue?

Firefox, Safari

### What are the browser driver(s) and version(s) where you see this issue?

n/a

### Are you using Selenium Grid?

4.4.0

TamsilAmani:
I am looking into this issue

AutomatedTester:
@titusfortner I have a few questions that I hope you can clarify.

* Does Chrome endpoints still exist when `{w3c:true}`?
* While this is good for Chrome support, do we want to be adding this back just for one browser? Wouldn't it be better to wait for the similar feature set that is coming in WebDriver Bidi
* What value do we see this having over getting people to move over to our OpenTelemetry support? (yes it doesn't give us browser internals but see question 2)

TamsilAmani:
@AutomatedTester I tested the issue with `{w3c:true}`, the end point exists in chrome.

diemol:
What I understand is that Chrome works in Grid and probably others work too but the endpoint does not exist, is that right, @titusfortner?

AutomatedTester:
Firefox doesn't support this endpoint, and I don't think that safari has support. This is going to only benefit Chrome users. I'm not fully against only supporting just chrome but we need to make it a deliberate choice

diemol:
Now remembering, I think this has always been a Chrome/Chromium only feature. I assume that Grid 3 responds in the same way. 

Should we leave it as it is and wait for BiDi?

titusfortner:
At the very least, we need for it `getAvailableLogTypes()` to return an empty list not throw an exception. Or to remove the method if we don't want to support it.

TamsilAmani:
[...]

PR #11822, opened 2023-03-27
Add support for getting logs

kant2002:
### Description
To have consistency with local drivers, add support for getting logs. Unfortunately I was not able to setup dev environment and test these changes locally. But seems to be infra is here.

### Motivation and Context

I was asked in the chat why this snippet does not working. And troubleshooting lead me here.
```
var chromeOptions = new ChromeOptions();
RemoteWebDriver driver = new RemoteWebDriver(new Uri("http://localhost:4444"), chromeOptions);
driver.Navigate().GoToUrl("https://stackoverflow.com/questions/5639473/using-selenium-remotewebdriver-in-c-sharp");
var logs = driver.Manage().Logs;
var av = logs.AvailableLogTypes; // empty
var browserLogs = logs.GetLog(LogType.Browser); // empty
```

### Types of changes

### Checklist


kant2002:
@diemol are you are right person to summon, so I can have feedback on this PR?

titusfortner:
So, our plan is to get browser logs to users with the new BiDi functionality (https://www.selenium.dev/documentation/webdriver/bidirectional/bidirectional_w3c/log/) instead of implementing it separately in Grid.
 
Right now this is only working in Java, JS & Ruby, though, and not .NET, yet

titusfortner:
This PR won't do anything because those endpoint will need to be wired up in the rest of the code.
I guess the related issue to discuss this on is #10949 
I just made a comment there about the missing grid log access, but in general we're waiting for BiDi logging for the rest of this support.

PR #11823, opened 2023-03-27
[java] Remove Json Wire Protocol support

shs96c:
This PR removes JWP support from the Java bindings, though there are undoubtedly places where supporting classes (eg. those that convert capability names from the JWP to W3C formats) have been left over.

Note that it is likely this will break older clients, as it looks like `DriverCommand` was mis-encoding the `capabilities` parameter as a `Map` rather than a collection of `Map`s. Before this lands, we should check for that.

Best reviewed on a commit-by-commit basis rather than in one go, but this PR is designed to be squashed before merging.


titusfortner:
Should we deprecate `DesiredCapabilities` along with this PR and tell people to use `MutableCapabilities` if they have to set things dynamically?

diemol:
> Should we deprecate `DesiredCapabilities` along with this PR and tell people to use `MutableCapabilities` if they have to set things dynamically?

I think we should do it apart... do we need to write some reasoning behind it or even check with other TLC members?

diemol (approved):
Failing tests are unrelated to this change.

Thank you, @shs96c!

PR #11827, opened 2023-03-28
[ruby] Try to load older devtools version if current browser version doesn't match

### LINKED ISSUE #11651: [🚀 Feature]: Ruby support for older CDP versions
### Feature and motivation

Instead of supporting a fall-back option for older CDP versions like other languages, Ruby implements new version support with a separate gem. The idea is that we would be able to keep up with adding CDP support even if we can't keep up with doing a full Selenium release. Currently we're struggling to do even that. So we should allow for users to use the latest available devtools version support (e.g. devtools for v109 with Chrome v110, etc)

### Usage example

n/a

arielj:
### Description

This is an attempt to fix https://github.com/SeleniumHQ/selenium/issues/11651.

This change makes it so, if the devtools version based on the current browser doesn't match, it tries to load an older version of devtools instead.

I added some logs/warnings to alert a user that this is happening, so they know there's something to address.

I only added 2 versions back fallback, I imagine that's plenty enough time until a new devtools release is done matching the proper version but let me know if that's not desired.

I would be happy to add tests for this but I really don't know how to do a test for something like this.

I also have to mention that I was not able to use the gem from a local clone of the repo to really really confirm this is working as expected, I CAN confirm though that the fallback is working to an extent because I can see the messages so it is trying to require older versions if the current one failed:

```
Could not load selenium-devtools v111. Trying older versions.

# and then the failure

LoadError:
       cannot load such file -- selenium/devtools/v109
```

### Motivation and Context

There's a recurring problem that when chrome gets a new major version update, requiring devtools fails until a new devtools version is released and updated in the Gemfile of the app.

### Types of changes

### Checklist


diemol (approved):
Thank you, @arielj!

rogerkk:
From a random user: Thanks, @arielj! :star:

robotdana:
hi could this be release perhaps as a 0.111.1? thank you :)

diemol:
This has already been released.

robotdana:
i'm not sure it has. i think the selenium-webdriver gem has had a recent release, but the selenium-devtools gem has not had a release since this was merged. 

i have the latest version of selenium-webdriver (4.8.6) and the latest version of selenium-devtools (0.111.0) 
and see a `LoadError: cannot load such file -- selenium/devtools/v112` when trying to do call a method that requires devtools, which i wouldn't have if this code was in it.

diemol:
You are right, seems the person who released 4.8.6 did not pull in this change. We are doing a release next week in any case.

rogerkk:
> We are doing a release next week in any case.

Is that release still on for this week? Our build has been breaking all week and I'm wondering whether I should start some mitigation or just wait a little bit longer. :)

diemol:
Release is delayed by one week, Eastern holidays happened. 

However, I just pushed https://rubygems.org/gems/selenium-devtools/versions/0.112.0

PR #11832, opened 2023-03-29
[Rust] Refactors

oriongonza:
### Description
General refactors in the selenium manager

### Motivation and Context
I was going to add a new functionality and found some code that could get updated.

### Checklist
- [x] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

### Changes

- Added string interpolation, introduced in version 1.58 https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html
- ran cargo fmt
- refactored the main pipeline. This could be changed to use `if let Err` syntax if this change is not accepted

Issue #11688, opened 2023-02-20
[🚀 Feature]: Change default folder for Selenium Manager cache


bonigarcia:
### Feature and motivation

Currently, the local folder used for cache in Selenium Manager is `~/.cache/selenium` by default. To change this path, a new flag like `--cache-path <CACHE_PATH>` is required.

### Usage example

```
selenium-manger --browser chrome --cache-path /my/custom/cache
```

iampopovich:
Am i right the custom cache path must be applied for the specified browser ?
The way that Chrome has cache_path1 and FF has cache_path2 is legal right?

bonigarcia:
Completed with #12514.

PR #11837, opened 2023-03-29
[rust] Save to directory

oriongonza:
Allows the user to download the webdriver to a specific directory.

### Description

Adds the `--save-path` flag, which allows the user to download the selected driver to a folder.

Also a few refactors that I didn't see in the other PR

### Motivation and Context

I had the need for this so I implemented it here.

### Types of changes

### Checklist

### DIFF
--- a/rust/src/config.rs
+++ b/rust/src/config.rs
@@ -25,9 +25,11 @@ use crate::{
 };
 use std::env;
 use std::env::consts::OS;
+use std::path::PathBuf;
 
 pub const ARM64_ARCH: &str = "arm64";
 
+#[derive(Clone)]
 pub struct ManagerConfig {
     pub browser_version: String,
     pub driver_version: String,
@@ -38,7 +40,7 @@ pub struct ManagerConfig {
     pub timeout: u64,
     pub browser_ttl: u64,
     pub driver_ttl: u64,
-    pub custom_save_path: Option<String>,
+    pub custom_save_path: Option<PathBuf>,
 }
 
 impl ManagerConfig {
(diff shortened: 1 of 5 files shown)


oriongonza:
@bonigarcia you might want to take a look at this two PRs #11832

bonigarcia (commented):
This PR should not be merged. The flag `--save-path` is not necessary, since the feature to change the binary path will be implemented through the flag `--cache-path` (planned in #11688, which has a current PR under review: #11840).

oriongonza:
This might still be useful (it is at least for me) since the webdriver will be saved under a specific version folder eg `~/mycache/111.xxx.xxx/chromedriver`.

oriongonza:
Still, even if this gets pushed back there are a few refactors worth considering, especially using `#[Derive(Clone)]` attribute macro.
By the way, why does the get_config clone the configuration struct instead of returning a mutable reference to it? That way you can simplify (and speed up but that doesn't really matter) the addding of a new configuration.

Also it might be worth considering using optionals instead of empty strings for the configurations that don't exist since rust provides really powerful tools for working with options,

bonigarcia:
> This might still be useful (it is at least for me) since the webdriver will be saved under a specific version folder eg `~/mycache/111.xxx.xxx/chromedriver`.

I see your point now. Still, I think having both flags `--cache-path` and `--save-path` can be confusing. Instead, I propose having only `--cache-path` (which it has already another PR), and created another extra flag e.g. called `--avoid-driver-tree` to store the driver without using a tree folder for the drivers. @dev-ardi What do you think?

bonigarcia:
> Still, even if this gets pushed back there are a few refactors worth considering, especially using `#[Derive(Clone)]` attribute macro. By the way, why does the get_config clone the configuration struct instead of returning a mutable reference to it? That way you can simplify (and speed up but that doesn't really matter) the addding of a new configuration.

Yes, I have just seen your PR about it and it seems a handy way to avoid the cloning method, thanks a lot. I think we should merge it, thanks.

> Also it might be worth considering using optionals instead of empty strings for the configurations that don't exist since rust provides really powerful tools for working with options,

Yes, that is a good idea too.

diemol:
I got lost a bit. What should we do with this PR? I believe it needs to be split to handle the refactor apart?

oriongonza:
> I see your point now. Still, I think having both flags `--cache-path` and `--save-path` can be confusing. Instead, I propose having only `--cache-path` (which it has already another PR), and created another extra flag e.g. called `--avoid-driver-tree` to store the driver without using a tree folder for the drivers. @dev-ardi What do you think?

That looks like the correct approach

oriongonza:
> I got lost a bit. What should we do with this PR? I believe it needs to be split to handle the refactor apart?

I guess so. I don't know how to select only some of the changes from each commit, I'm very good with git. I could use some help with this :)

diemol:
You can create a complete new branch from `trunk` and only apply the refactoring changes to send a PR.

PR #11847, opened 2023-03-31
[JS][BiDi] Add Script module commands and types

TamsilAmani:
### Description

Add script module commands and types

### Motivation and Context

Implement script module commands and types as described in https://w3c.github.io/webdriver-bidi/#module-script.
The majority of classes and interfaces are mapping the common data types as described in https://w3c.github.io/webdriver-bidi/#data-types.

### Types of changes

### Checklist


TamsilAmani:
I suggest to add getter functions where ever possible instead of directly accessing the object's attributes. Is that okay?

javascript/node/selenium-webdriver/bidi/evaluateResult.js
```diff
@@ -0,0 +1,60 @@
...
+// under the License.
+
+const EvaluateResultType = {
+  SUCCESS: 'success',
+  EXCEPTION: 'exception',
+}
+
+class EvaluateResultSuccess {
+  constructor(type, realmId, value) {
+    this.resultType = type
```
harsha509: if this class only is for success , can we do `this.resultType = EvaluateResultType.SUCCESS;` right ?

javascript/node/selenium-webdriver/bidi/evaluateResult.js
```diff
@@ -0,0 +1,60 @@
...
+    this.realmId = realmId
+    this.result = value
+  }
+}
+
+class EvaluateResultException {
+  constructor(type, realmId, exceptionDetails) {
+    this.resultType = type
```
harsha509: same here `this.resultType = EvaluateResultType.EXCEPTION;` we can reduce one param on usage

javascript/node/selenium-webdriver/bidi/protocolType.js
```diff
@@ -0,0 +1,109 @@
...
+const PrimitiveType = {
+  UNDEFINED: 'undefined',
+  NULL: 'null',
+  STRING: 'string',
+  NUMBER: 'number',
+  SPECIAL_NUMBER: 'number',
+  BOOLEAN: 'boolean',
+  BIGINT: 'bigint',
+
+  findByName: function (name) {
```
harsha509: another way (only if we want to reduce code)

  `findByName(name) {
    return Object.values(this).find((type) => {
      return typeof type === 'string' && name.toLowerCase() === type.toLowerCase();
    }) || null;
  },`

javascript/node/selenium-webdriver/bidi/protocolValue.js
```diff
@@ -0,0 +1,210 @@
...
+      this.sharedId = remoteValue['sharedId']
+    }
+
+    if (this.value != null) {
+      this.value = this.deserializeValue(this.value, this.type)
+    }
+  }
+
+  deserializeValue(value, type) {
+    if (type === NonPrimitiveType.MAP || type === NonPrimitiveType.OBJECT) {
```
harsha509: can we use like below

  `if ([NonPrimitiveType.MAP, NonPrimitiveType.OBJECT].includes(type)) {
    return Object.fromEntries(value);
  } 
  else if ()...`

javascript/node/selenium-webdriver/bidi/protocolValue.js
```diff
@@ -0,0 +1,210 @@
...
+    if (handle === RemoteReferenceType.HANDLE) {
+      this.handle = shareId
+    } else {
+      this.handle = handle
+      this.shareId = shareId
+    }
+  }
+
+  asMap() {
+    let toReturn = {}
```
harsha509: const

TamsilAmani:
Thank you for the review @harsha509. I have pushed the changes.

harsha509 (approved):
LGTM!

PR #11852, opened 2023-03-31
Use Bazel for running Rust tests

p0deje:
Brings Rust bits on par with the rest, allowing to run tests using Bazel:

```
$ bazel test -t- //rust/...
INFO: Analyzed 9 targets (0 packages loaded, 1640 targets configured).
INFO: Found 2 targets and 7 test targets...
INFO: Elapsed time: 22.218s, Critical Path: 20.78s
INFO: 8 processes: 2 darwin-sandbox, 12 local.
INFO: Build completed successfully, 8 total actions
//rust:tests/cli_tests                                                   PASSED in 20.7s
//rust:tests/iexplorer_tests                                             PASSED in 0.5s
//rust:tests/output_tests                                                PASSED in 0.1s
//rust:tests/proxy_tests                                                 PASSED in 0.1s
//rust:tests/safari_tests                                                PASSED in 0.0s
//rust:tests/timeout_tests                                               PASSED in 0.1s
//rust:unit                                                              PASSED in 0.1s
```

It seems like proxy tests cannot be easily run in parallel to others so I ended up just using the `exclusive` tag to force all the tests sequentially.

### DIFF
--- a/.github/workflows/ci-rust.yml
+++ b/.github/workflows/ci-rust.yml
@@ -5,28 +5,27 @@ on:
   workflow_dispatch:
 
 jobs:
-  test:
-    name: Test
+  build:
+    name: Build
+    uses: ./.github/workflows/bazel.yml
+    with:
+      name: Build
+      cache-key: rust-build
+      run: bazel build //rust:selenium-manager
+
+  tests:
+    name: Tests
+    needs: build
+    uses: ./.github/workflows/bazel.yml
     strategy:
+      fail-fast: false
       matrix:
-        os:
-          - ubuntu-latest
-          - windows-latest
-          - macos-latest
-        rust:
-          - stable
-    runs-on: ${{ matrix.os }}
-    env:
-      RUST_BACKTRACE: 1
-    steps:
-    - name: Checkout repository
-      uses: actions/checkout@v3
-    - name: Install Rust toolchain
-      uses: actions-rs/toolchain@v1
-      with:
-        toolchain: ${{ matrix.rust }}
-    - name: Run tests
-      run: |
-        cd rust
-        cargo build
-        cargo test -- --nocapture
+        include:
+          - os: macos
+          - os: ubuntu
+          - os: windows
+    with:
+      name: Tests (${{ matrix.os }})
+      cache-key: rust-test
+      os: ${{ matrix.os }}
+      run: bazel test --test_env=RUST_BACKTRACE=1 --flaky_test_attempts=3 //rust/...
--- a/README.md
+++ b/README.md
@@ -298,6 +298,16 @@ More information about running Selenium's .NET tests can be found in this [READM
 
 </details>
 
+#### Rust
+<details>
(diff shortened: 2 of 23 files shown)


bonigarcia (commented):
I am not sure about this PR. What is the benefit of running the Rust test with Bazel instead of cargo?

In any case, the current PR is not exactly the same as `cargo test`. If I am not wrong, the tests executed are those included in the `tests` folders:

```
for file in glob(["tests/*.rs"])
```

Nevertheless, those are only the integration tests. There are also unit tests within the `src` folder (which is the typical way to do it with Rust). See, for example, [firefox.rs](https://github.com/SeleniumHQ/selenium/blob/trunk/rust/src/firefox.rs).

p0deje:
> I am not sure about this PR. What is the benefit of running the Rust test with Bazel instead of cargo?

The main benefits are the same as for any other code in the repository - use a single tool that automatically installs the Rust toolchain, compiles the binary, and hermetically runs the tests without requiring a developer to do all of that manually. Furthermore, once we start using remote cache and remote build, it will be possible to build Linux versions from macOS machines for example.

> Nevertheless, those are only the integration tests. There are also unit tests within the src folder (which is the typical way to do it with Rust). See, for example, [firefox.rs](https://github.com/SeleniumHQ/selenium/blob/trunk/rust/src/firefox.rs).

Thanks for pointing that out, I was not aware - will add this too.

diemol:
I wonder why tests are failing if they are passing on `trunk`, the only difference is how they are executed. Any ideas @p0deje @bonigarcia ?

p0deje:
@diemol I think the unit tests are failing because of the sandbox and can be fixed by adding a `no-sandbox` tag to the target. I am not sure what is wrong with proxy tests on Windows though.

diemol:
Strange... Why is the proxy test failing on Windows on this PR but passing on `trunk`?

p0deje:
@diemol It's actually failing on trunk too if you run it separately `cargo test --test proxy_tests`. I'm trying to figure out what's going on with @bonigarcia.

bonigarcia:
@p0deje It seems this issue has been fixed after merging #11865 by @dev-ardi (thanks about that!). Now the output in Windows is as follows:

```
C:\Users\boni\Documents\dev\selenium\rust>cargo test --test proxy_tests
   Compiling selenium-manager v1.0.0-M3 (C:\Users\boni\Documents\dev\selenium\rust)
    Finished test [unoptimized + debuginfo] target(s) in 2.43s
     Running tests\proxy_tests.rs (target\debug\deps\proxy_tests-56cacd1d69678e3c.exe)

running 3 tests
test wrong_port_proxy_test ... ok
test wrong_protocol_proxy_test ... ok
test ok_proxy_test ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.29s
```

bonigarcia (approved):

PR #11863, opened 2023-04-04
[dotnet] Don't listen to error output stream from selenium manager to avoid messing up output

nvborisenko:
### Description
There is a race condition when output of error stream from selenium manager may be messed with output from standard stream. Since we expect standard output in json format, then it might break json structure. I played around it and discovered it might happen 1 out of 200 times (created synthetic test against selenium manager).

### Motivation and Context
Using selenium manager in dotnet sometimes fails without error. My test fails sporadically. This PR fixes one of the possible "hidden" errors.

### Types of changes

### Checklist

### Notes for future
If we will need to listen error output, then we can apply the same mechanism, but store output in different object which is not shared with standard output.

### Details
This is my synthetic test: it outputs the webdriver path given from selenium manager.

image

Sometimes the path is incorrect.


diemol (approved):
Thank you, @nvborisenko!

nvborisenko:
Forgot to mention my synthetic test passes now, which invoked selenium manager 1K times. Output from SM handled correctly.

PR #11866, opened 2023-04-05
Passing options to Selenium Manager

diemol:

This passes the browser options all the way to the driver service. 
It lays down the work to implement #11372 and #11351

### Types of changes

### Checklist


py/selenium/webdriver/chrome/webdriver.py
```diff
@@ -74,8 +75,11 @@ def __init__(
             )
         else:
             keep_alive = True
+        if not options:
+            options = self.create_options()
         if not service:
             service = Service(executable_path, port, service_args, service_log_path)
+        service.path = DriverFinder.get_path(service, options)
```
titusfortner, line 82: The user may have already set path in the provided service class and we do not want to overwrite it. Also, the service class doesn't have a path setter, so this shouldn't work? And if it does work, I don't think it's "correct?"

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -61,18 +62,21 @@ def get_binary() -> Path:
...
-    def driver_location(self, browser: str) -> str:
+    def driver_location(self, options: BaseOptions) -> str:
         """
         Determines the path of the correct driver.
         :Args:
          - browser: which browser to get the driver path for.
         :Returns: The driver path to use
         """
 
+        browser = options.capabilities["browserName"]
```
titusfortner, line 73: We shouldn't need any of this any more. If the options is an options class, it'll have a browser name Selenium Manager will recognize, and if it doesn't, I think it's ok for Selenium Manager to give that error? I think the more code we can give to Selenium Manager, the better. 😃

py/selenium/webdriver/common/service.py
```diff
@@ -89,17 +88,7 @@ def start(self) -> None:
...
-                browser = self.__class__.__module__.split(".")[-2]
-
-                try:
-                    path = SeleniumManager().driver_location(browser)
-                except WebDriverException as new_err:
-                    logger.debug("Unable to obtain driver using Selenium Manager: " + new_err.msg)
-                    raise err
-
-                self._start_process(path)
+            raise err
```
titusfortner: We should be able to remove this try/catch entirely because the error handling is all done in the `_start_process()` method now. Alternately, we can move all of the `_start_process()` error handling back into this `start()` method, which is where it was before I needed to be able to rescue it and try Selenium Manager before.

titusfortner (changes requested):
A few comments; and actually, we want to add this to Safari as well.
Python does not appear to be doing the right thing for Technology Preview...

Theoretically this should work:
```
    options = SafariOptions()
    options.use_technology_preview = True
    driver = webdriver.Safari(options=options)
```
but right now you actually need to add the service with the hard coded path to the preview driver:
```
    service = SafariService(executable_path = '/Applications/Safari Technology Preview.app/Contents/MacOS/safaridriver')
    driver = webdriver.Safari(options=options, service=service)
```

py/selenium/webdriver/chrome/webdriver.py
titusfortner: hmm, except in Python we assign a default if one isn't passed in, so we wouldn't know the difference. And oddly "executable_path" in Python is actually "driver name"

titusfortner: Well, we need to figure out how to let a user pass in a specific location for a driver without overriding it. I think this might mean an `@abstractmethod` method in the superclass for getting path and calling methods in the the subclass when executing `start()`.

diemol: If the user sets the path, we do not overwrite it unless it does not work. That is done inside `DriverFinder`. I'll change it so we use a method to set the value.

py/selenium/webdriver/common/selenium_manager.py
diemol: True, but this is out of the scope of this PR. When I make the change across bindings to support Safari, then I will remove this code.

py/selenium/webdriver/common/service.py
diemol: I'll just remove the try/catch, thank you for pointing it out.

py/selenium/webdriver/chrome/webdriver.py
diemol: I made the change anyway, checking if there is a path set before doing anything. This aligns it with the code in the other bindings.

titusfortner: Oh, I was thinking the default value in the Service class constructor would break this, but we want to support both user provided and what is on PATH, which the previous implementation would address. Would be really nice to have unit tests for these... :-D Might need @symonk help?

diemol: Oh, that is a good catch. I always run tests before pushing, and they passed because I had copied a ChromeDriver to my path to test the changes to JavaScript locally and I forgot to remove it.
I now reverted the changes and it works fine. 
I am not sure how to test that as a unit or small test, because we need to actually interact with a file on the PATH. I'd prefer to create a new workflow where we remove the drivers from the PATH.

titusfortner: Yeah, it would need a unit test with some kind of mocking, which I'm not as familiar with in pytest

titusfortner: And in a future version I would like for Selenium Manager to process PATH instead of the bindings. Would love to pull out all the logic that searches all the directories for things. In which case we will need to differentiate what is passed in from the default that locates on PATH. But we can do that later.

diemol: Processing PATH is already tracked in a different issue.

PR #11869, opened 2023-04-06
[ruby] Fix devtools version fallback

robotdana:
### Description

When using the feature added in #11827 that loads devtools methods possibly a version or two earlier than was necessary asked for, i was seeing this error:

```
NoMethodError: undefined method `get_targets' for nil:NilClass
./rb/lib/selenium/webdriver/devtools.rb:75:in `start_session'
./rb/lib/selenium/webdriver/devtools.rb:34:in `initialize'
./rb/lib/selenium/webdriver/common/driver_extensions/has_devtools.rb:35:in `new'
./rb/lib/selenium/webdriver/common/driver_extensions/has_devtools.rb:35:in `devtools'
```
because `target` was nil because it was trying to load a class named:
```
Selenium::DevTools::V#{Selenium::DevTools.version}::#{method.capitalize}
```
which resolved to:
```
Selenium::DevTools::V112::Target
```
even though the falling-back code had already given up on 112 and loaded 111 instead.

The simplest way to fix this seems to be ensure that `Selenium::DevTools.version` matches what was loaded.

Also now there's a test

### Types of changes

### Checklist

### DIFF
--- a/rb/lib/selenium/devtools.rb
+++ b/rb/lib/selenium/devtools.rb
@@ -40,6 +40,7 @@ module Selenium
 
       def load_old_version(version)
         require "selenium/devtools/v#{version}"
+        self.version = version
         Kernel.warn "Using selenium-devtools version v#{version}, some features may not work as expected."
       end
     end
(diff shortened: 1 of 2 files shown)


titusfortner:
Ah, yes, we should make sure we are only using fallbacks if the version is greater than what is available.

titusfortner:
Ok, I see what this is now. Thanks for the code and the test!

Issue #11649, opened 2023-02-13
[🐛 Bug]: Selenium manager doesn't work in paths that contain spaces


cfarmerga:
### What happened?

I am using Windows 11 and have a simple js script that tries to build a driver using selenium manager. I see that it works fine when I have the script in a path which contains no spaces, but it fails when spaces exist in the path with this error:

```
c:\dev\selenium path with space>node simple.js                                                                                                                                                   
 The ChromeDriver could not be found on the current PATH, trying Selenium Manager                                                                                                                
'c:\dev\selenium' is not recognized as an internal or external command,                                                                                                                          
operable program or batch file.                                                                                                                                                                  
Unable to obtain driver using Selenium Manager: Error: Error executing command with c:\dev\selenium path with space\node_modules\selenium-webdriver\bin\windows\selenium-manager.exe,--browser,ch
rome                                                                                                                                                                                             
'c:\dev\selenium' is not recognized as an internal or external command,                                                                                                                          
operable program or batch file.                                                                                                                                                                  
                                                                                                                                                                                                 
c:\dev\selenium path with space\node_modules\selenium-webdriver\chrome.js:175                                                                                                                    
      throw Error(                                                                                                                                                                               
            ^        
```                                                                                                                                                                            
                                                                                                                                                                                                 
I see that the cause seems to be the code in selenium-webdriver/common/seleniumManager.js (line 76), where the driverLocation function tries to use the result of getBinary() directly in execSync.  It seems like this path should be quoted before being passed to execSync.

Also posted to https://stackoverflow.com/questions/75438848/selenium-builder-failing-with-spaces-in-path



### How can we reproduce the issue?

```shell
package.json:
{
  "type": "commonjs",
  "dependencies": {
    "selenium-webdriver": "4.8.0"
  }
}


Here's the relevant code:

const { Builder, Browser } = require("selenium-webdriver");

(async function example() {
  const driver = await new Builder().forBrowser(Browser.CHROME).build();
  driver.get("https://stackoverflow.com");
})();
```


### Relevant log output

```shell
c:\dev\selenium path with space>node simple.js                                                                                                                                                   
 The ChromeDriver could not be found on the current PATH, trying Selenium Manager                                                                                                                
'c:\dev\selenium' is not recognized as an internal or external command,                                                                                                                          
operable program or batch file.                                                                                                                                                                  
[...]

PR #11873, opened 2023-04-07
[js] Replace `execSync` with `spawnSync` in `seleniumManager.js` (#11649)

atrnh:
### Description

Replaced `execSync` with [`spawnSync`](https://nodejs.org/api/child_process.html#child_processspawnsynccommand-args-options) in `javascript/node/selenium-webdriver/common/seleniumManager.js`.

I also had to change the way errors are handled a bit, because `spawnSync` doesn't raise exceptions. Instead, it returns an object with an `error` property. I was able to retain the existing error-handling constructs, but it feels a bit indirect due to how `spawnSync` works. Not sure if this is a big deal or not!

### Motivation and Context

Since `execSync` does not escape shell metacharacters, attempting to execute a command with a space in it will fail. This is the case described in #11649, when the path to `selenium-manager` contains a space. This commit replaces `execSync` with `spawnSync`, which does escape shell metacharacters.

### Types of changes

### Checklist


diemol (approved):
Thank you, @atrnh!

I made some refinements because I got a few errors when I tried it locally. But it is indeed a good idea to use `spawnSync` instead of `execSync`, seems that command and arguments, plus return streams are easier to handle.

PR #11883, opened 2023-04-11
[js] fix: return statement in submit()

TamsilAmani:
### Description

The value evaluated by the submit() was not being returned.

### Motivation and Context

Help developers who use selenium to get correct result.

### Types of changes

### Checklist


harsha509:
Makes sense!

Thank you @TamsilAmani !

harsha509 (dismissed):
Thank you @TamsilAmani !

titusfortner:
What result are we expecting from submit? It's an action on the browser, so the return should be `null`

swrdfish:
@titusfortner @harsha509 
Thank you for looking into this.

I stumbled upon this bug while working on [Nightwatch](https://github.com/nightwatchjs/nightwatch). The problem is that `.submit()` is supposed to return a promise, which can be used to wait for the command to complete. Without the promise we cannot ensure that subsequent actions behave deterministically. It is not the value we are interested in but the promise to let us know when the command has finished.

PR #11887, opened 2023-04-12
Configure remote builds to use EngFlow

shs96c:
This allows us to use EngFlow's remote build grid to execute our builds. Once we hook this up into our CI actions, this will lead to significantly faster CI runs.


.bazelrc
```diff
@@ -69,3 +71,46 @@ test --test_env=HOME
...
+
+build:remote --crosstool_top=//common/remote-build/cc:toolchain
+build:remote --cpu=k8
+build:remote --extra_execution_platforms=//common/remote-build:platform
+build:remote --extra_toolchains=//common/remote-build:cc-toolchain
+build:remote --host_platform=//common/remote-build:platform
+build:remote --platforms=//common/remote-build:platform
+
+# The Docker images are running Linux
+build:remote --cpu=k8
```
p0deje, line 102: This duplicates line 96

p0deje (commented):
I see there is a whole new `common/remote-build` directory with many autogenerated files. Is there any way to shrink those? For example, I assumed we don't have to manually configure C++ toolchains because I thought it's built-in to Bazel.

.bazelrc
shs96c: Removed.

shs96c:
The files are auto-generated by (effectively, if not in reality) running bazel on the remote machine and then copying the generated files into the repo. It's not a great idea to modify them.

We don't have a hermetic C++ toolchain, so we still need to manage that ourselves. Our remote build images have a C++ toolchain installed, so we need to tell Bazel about it. Because the build is remote, we can't figure this stuff out at run time (since we need to know everything _before_ sending the actions to the remote machine), and this is the way to configure it.

p0deje:
> The files are auto-generated by (effectively, if not in reality) running bazel on the remote machine and then copying the generated files into the repo. It's not a great idea to modify them.

Got it. I think as long as it's documented how to re-generate these files should we need - we are fine.

> We don't have a hermetic C++ toolchain, so we still need to manage that ourselves. Our remote build images have a C++ toolchain installed, so we need to tell Bazel about it. Because the build is remote, we can't figure this stuff out at run time (since we need to know everything before sending the actions to the remote machine), and this is the way to configure it.

Ok, so Bazel doesn't ship with a hermetic C++ toolchain. Would using something like https://github.com/uber/hermetic_cc_toolchain help us? Probably not now, but eventually.

shs96c:
> Got it. I think as long as it's documented how to re-generate these files should we need - we are fine.

We just ran https://github.com/bazelbuild/bazel-toolchains

> Ok, so Bazel doesn't ship with a hermetic C++ toolchain. Would using something like https://github.com/uber/hermetic_cc_toolchain help us? Probably not now, but eventually.

I spent an ungodly amount of time attempting to get the zig-cc toolchain working and gave up. If all we wanted to do was support Linux, it'd be easier, but we want to support the matrix of compiling for and on Windows, Mac OS, and Linux. For  that we need various sysroots, and the legality of distributing these is unknown. That makes it a really difficult problem to solve. It's also not something that we can solve in this PR :)

Issue #11906, opened 2023-04-15
[🐛 Bug]:  cargo test fails in my machine


oriongonza:
### What happened?

I ran `cargo test` and it failed

### How can we reproduce the issue?

```shell
git switch e79b2f2c89316632e065774d948d3972b48435bf
cargo test
```
```


### Relevant log output

```shell
Running tests/proxy_tests.rs (target/debug/deps/proxy_tests-346a6ea17328a77f)

running 3 tests
test wrong_port_proxy_test ... ok
test wrong_protocol_proxy_test ... ok
test ok_proxy_test ... FAILED

failures:

---- ok_proxy_test stdout ----
thread 'ok_proxy_test' panicked at 'Unexpected failure.
code-65
stderr=""```
command=`"/home/ardi/repos/selenium/rust/target/debug/selenium-manager" "--browser" "chrome" "--proxy" "http://127.0.0.1:39559"`
code=65
stdout="ERROR\terror sending request for url (https://chromedriver.storage.googleapis.com/LATEST_RELEASE): error trying to connect: unsuccessful tunnel\n"
stderr=""
', /rustc/511364e7874dba9649a264100407e4bffe7b5425/library/core/src/ops/function.rs:250:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    ok_proxy_test

test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s

error: test failed, to rerun pass `--test proxy_tests`
```

### Operating System

Arch linux

bonigarcia:
This issue is related to #11852. I am looking into this now. It is a strange behaviour when running the proxy tests individually:

```
cargo test --test proxy_tests
```

bonigarcia:
It seems this issue has been fixed after merging your PR https://github.com/SeleniumHQ/selenium/pull/11865. Thanks again!

PR #11907, opened 2023-04-15
scrapped refactors from old PR

oriongonza:

Apply refactors from https://github.com/SeleniumHQ/selenium/pull/11837

- I delete clone from config as it's not needed anymore 
- I change some owned types that don't need to be to references
- Pattern matching for str_to_os

### Types of changes

### Checklist

- [ ] All new and existing tests passed. https://github.com/SeleniumHQ/selenium/issues/11906


rust/src/config.rs
```diff
@@ -129,12 +114,11 @@ impl OS {
 }
 
 pub fn str_to_os(os: &str) -> OS {
-    if WINDOWS.is(os) {
-        WINDOWS
-    } else if MACOS.is(os) {
-        MACOS
-    } else {
-        LINUX
+    match os {
```
bonigarcia: Please revert these changes. Currently, Selenium Manager is only compiled for Windows, macOS, and Linux. Moreover, panicking in not a good option, since all errors should properly logged through the logger (since the output of Selenium Manager is done in several options, such as JSON).

bonigarcia (changes requested):

rust/src/config.rs
oriongonza: I agree that we should not panic with user input, however I still believe it to be an error to pass any value other than the permitted ones. We can still add a fallback (` _ => LINUX`), however I don't think this should be allowed.

I leave it up to you to decide.

bonigarcia: My decision was clear: "Please revert these changes". And I gave you the reason why: Selenium Manager is only compiled for these three operating systems: Windows, macOS, and Linux.

diemol:
@dev-ardi would you like to continue working on this PR? We are about to do a release soon and wanted to know if we could have this as part of the release.

oriongonza:
@diemol sure, it's a very minor PR but I can try to get some time and finish it today :)

bonigarcia (approved):

bonigarcia (approved):

diemol (approved):
Thank you, @dev-ardi!

PR #11913, opened 2023-04-18
[dotnet] Use ChangeType method to convert any object to boolean in WebElement propoerties

nvborisenko:
This approach brings us a possibility to track what exactly we try to convert.

### Description
```csharp
var a = (object)"true";
var b = (bool)a;
```
It always throws `InvalidCastException`.

There is safer method:
```csharp
var a = (object)"true";
var b = (bool)Convert.ChageType(a, typeof(bool));
```

Even if input is not formatted well to be boolean, then we will get more clear exception:
```
System.FormatException: String 'false 66' was not recognized as a valid Boolean.
```
where `false 66` is incorrect input.

### Motivation and Context
Just to get more clear understanding what is going wrong, and highlight a value returned from server side.

### Types of changes

### Checklist


diemol (approved):
Thank you, @nvborisenko!

PR #11918, opened 2023-04-19
[java] Optimize `equals` and `hashCode` methods of `Rectangle` class

valfirst:

### Description
Optimize `equals` and `hashCode` methods of `Rectangle` class

### Motivation and Context
Avoid creation of additional objects when invoking `equals` and `hashCode` methods of `Rectangle` class

### Types of changes

### Checklist


java/src/org/openqa/selenium/Rectangle.java
```diff
@@ -105,17 +107,12 @@ public boolean equals(Object o) {
...
+    return x == rectangle.x && y == rectangle.y && height == rectangle.height
+           && width == rectangle.width;
   }
 
   @Override
   public int hashCode() {
-    int result = getPoint().hashCode();
-    result = 31 * result + getDimension().hashCode();
-    return result;
+    return Objects.hash(x, y, height, width);
```
joerg1985, line 84: the varargs of `Objects.hash` will create an array.

valfirst: @joerg1985 sorry I don't get your proposal

joerg1985: @valfirst 
Java will put all the arguments passed to an varargs argument  into a new array.
In this case every time `Objects.hash` is called an array is created for the `Object... values)` argument.
See https://www.baeldung.com/java-varargs#1-safe-usage for details.

valfirst: I see, what is your recommendation?

joerg1985: I am not able to speak for the people at selenium, this was just the hint if the code must be really fast (for what ever reason) it might fail this goal. NetBeans does generate the code below, but the `Objects.hash` one liner from IntelliJ is better to read.

```
    @Override
    public int hashCode() {
        int hash = 7;
        hash = 73 * hash + this.x;
        hash = 73 * hash + this.y;
        hash = 73 * hash + this.height;
        hash = 73 * hash + this.width;
        return hash;
    }
```

diemol: @valfirst any thoughts?

valfirst: IMHO prime-numbers-based approach is an old-fashioned and used before `Objects.hash` was introduced, however I haven't done any benchmarking or so

krmahadevan: Found this SO post : https://stackoverflow.com/q/45832458

joerg1985: I think the conversation can be closed, e.g. Point is using the `Objects.hash` too.
And @valfirst did not tell a concrete use case where `Rectangle.hashCode` must be __really__ fast.
For what ever reason i could only open the conversation, but i am not able to close it.

valfirst: I'm not sure why the focus is moved to performance, my initial goal was to make equals/hashcode clean and concise.

PS: probably word "optimize" in title confuses

diemol (approved):
Thank you, @valfirst!

Issue #11915, opened 2023-04-19
[🚀 Feature]: py: make trio an optional package dependency


blueyed:
### Feature and motivation

I think the dependency on trio in the Python package could be made optional: it appears to only be necessary for bidi support (via https://github.com/SeleniumHQ/selenium/commits/395a92669255cbf8d2f4efb5bcffa359b1507b3c/py/selenium/webdriver/common/bidi/cdp.py).  I am not sure though what this is exactly (used for/with).

It could be moved into an extra, to be installable using `selenium[trio]` or `selenium[bidi]`.

trio itself has quite a few deps, as of 0.22.0:
```
async-generator = ">=1.9"
attrs = ">=19.2.0"
cffi = {version = ">=1.14", markers = "os_name == \"nt\" and implementation_name != \"pypy\""}
idna = "*"
outcome = "*"
sniffio = "*"
sortedcontainers = "*"
```

Related commits: https://github.com/SeleniumHQ/selenium/commit/587106bf243da214b95e163f9cc3670df9c9fc56, https://github.com/SeleniumHQ/selenium/commit/6e59a16570be0635330b52f3bd2b70060ed78b16

### Usage example

This would allow to have fewer dev dependencies with Python projects, where selenium is used.

titusfortner:
@blueyed can you make a PR for this?

blueyed:
Done: https://github.com/SeleniumHQ/selenium/pull/11920

Is the name "bidi" good for this?
Do you have a reference to read up on what it is for?  That would also be the place to update docs with the PR then.

diemol:
The conversation in the PR was not continued. This looks to be a "nice to have, "but as @AutomatedTester says, BiDi will be required in the future in all cases.

PR #11920, opened 2023-04-20
py(setup.py): add bidi extra requirement for trio

blueyed:
ref: https://github.com/SeleniumHQ/selenium/issues/11915

TODO:
- [ ] doc

### DIFF
--- a/py/setup.py
+++ b/py/setup.py
@@ -72,10 +72,14 @@ setup_args = {
     'include_package_data': True,
     'install_requires': [
         "urllib3[socks]~=1.26",
-        "trio~=0.17",
-        "trio-websocket~=0.9",
         "certifi>=2021.10.8",
     ],
+    "extras_require": {
+        "bidi": [
+            "trio~=0.17",
+            "trio-websocket~=0.9",
+        ],
+    },
     'zip_safe': False
 }
 


py/setup.py
```diff
@@ -72,10 +72,14 @@
...
-        "trio~=0.17",
-        "trio-websocket~=0.9",
         "certifi>=2021.10.8",
     ],
+    'extras_require': {
+        "bidi": [
+            "trio~=0.17",
+            "trio-websocket~=0.9",
+        ],
+    },
```
blueyed, line 82: Quoting style is mixed, but I've tried to follow the existing pattern.  Should I just use double quotes everywhere?

titusfortner: @symonk what is our preference here?

blueyed: black uses double quotes, which I also prefer.

AutomatedTester: It would be double quotes but let's do that in a separate PR

AutomatedTester:
How much of an issue is this in reality? We're going to be making BiDi the main way to do things in the future.

py/setup.py
blueyed: I can use just double quotes here already.

diemol:
Closing as we never got feedback.

PR #11930, opened 2023-04-23
[dotnet] Fallback result parsing of script execution to expected type

### LINKED ISSUE #11801: [🐛 Bug]: ExecuteJavaScript<double> fails if the number has no decimal part
### What happened?

In Chrome `Browser.ExecuteJavaScript<double>("{return performance.timeOrigin ;}");` fails 1 in 10 times because it returns a double truncated to 1 decimal. However when the decimal part is 0 Selenium sees this as a long, which is not `IsInstanceOfType` double, so it crashes with `WebDriverException("Script returned a value, but the result could not be cast to the desired type");`.

This means that it is in general not possible to use `double` for `ExecuteJavaScript`, even though it often seems correct.

```shell
`Browser.ExecuteJavaScript<double>("{return 123.0;}");`
```

### Relevant log output

```shell
OpenQA.Selenium.WebDriverException : Script returned a value, but the result could not be cast to the desired type
   at OpenQA.Selenium.Support.Extensions.WebDriverExtensions.ExecuteJavaScript[T](IWebDriver driver, String script, Object[] args)
```

### Operating System

Windows 10

### Selenium version

.Net 4.8.1

Chrome 111

ChromeDriver 111

_No response_

nvborisenko:
### Description
Executing JS in browser may return result of different types. I have added new conditional thread to try casting of result to expected type.

It is not a breaking change, this change can be considered as "now trying one more way".

### Motivation and Context
Fixes #11801 

### Types of changes

### Checklist

### DIFF
--- a/dotnet/src/support/Extensions/WebDriverExtensions.cs
+++ b/dotnet/src/support/Extensions/WebDriverExtensions.cs
@@ -101,13 +101,20 @@ namespace OpenQA.Selenium.Support.Extensions
                     throw new WebDriverException("Script returned null, but desired type is a value type");
                 }
             }
-            else if (!type.IsInstanceOfType(value))
+            else if (type.IsInstanceOfType(value))
             {
-                throw new WebDriverException("Script returned a value, but the result could not be cast to the desired type");
+                result = (T)value;
             }
             else
             {
-                result = (T)value;
+                try
+                {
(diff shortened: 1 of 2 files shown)


diemol (approved):
Thank you, @nvborisenko!

Issue #11642, opened 2023-02-10
[🚀 Feature]: Selenium Manager log additional information at FINE level


titusfortner:
### Feature and motivation

Currently we're only logging warnings, we also want to understand what it is doing when it is "working properly"

Ruby & Python are doing this currently.

### Usage example

n/a

diemol:
Can you share the link of the code to see how Python/Ruby are doing it?

titusfortner:
We probably want to add:
* "Unable to locate Drivers on System, Using Selenium Manager"
* "Executing command line: <command>"
* "Driver located at: <path>"

There will be more things to log upcoming :-D 

Heck, looking at how much Java logs, maybe we even want to turn on `--debug` and put all the info into `FINER` after all. 🤷‍♂️

PR #11959, opened 2023-04-27
[java] add more logging to Selenium Manager

titusfortner:
### Description

* Changes log name to `log` which we want to make the standard in the rest of the project (future PR)
* Add `INFO` that about to use Selenium Manager
* Send `--verbose` to Selenium Manager if logging set to `FINE` or higher
* Add `FINE` for: 
    * Command executed
    * Driver Path being used
    * Any DEBUG lines of log 

### Motivation and Context

We need to be able to get more details about what problems users are having 
Also (at least while in Beta) we want users to know that Selenium Manager is being used.
We may want to change that from `INFO` to `FINE` after beta?

Ideally we have a section on Selenium Manager in our documentation that we can point to for explaining what it is doing and how to get more logging info, etc. But that's a future PR.


titusfortner:
Ah found the issue for this PR - #11642

java/src/org/openqa/selenium/manager/SeleniumManager.java
```diff
@@ -187,24 +196,34 @@ private String getBrowserBinary(Capabilities options) {
...
-          Arrays.asList(binaryFile.getAbsolutePath(),
-                        "--browser",
-                        options.getBrowserName(),
-                        "--output", "json"));
+        List<String> commandList = new ArrayList<>();
+        commandList.add(binaryFile.getAbsolutePath());
+        commandList.add("--browser");
+        commandList.add(options.getBrowserName());
+        commandList.add("--output");
+        commandList.add("--json");
```
diemol:
```suggestion
        commandList.add("json");
```

diemol: That is a value for the parameter not a parameter itself.

diemol (changes requested):
Left a suggestion

java/src/org/openqa/selenium/manager/SeleniumManager.java
```diff
@@ -54,11 +55,12 @@
 @Beta
 public class SeleniumManager {
 
-    private static final Logger LOG = Logger.getLogger(SeleniumManager.class.getName());
+    private static final Logger log = Logger.getLogger(SeleniumManager.class.getName());
```
titusfortner: We have a lot more `LOG` than `log`, so I think we do want to use `LOG` here, I'll change it back

diemol (approved):

PR #11966, opened 2023-04-28
Upgrade minimum Ruby version to 3.0 since 2.7 is EOL

p0deje:
Since Ruby 2.7 is EOL, upgrade the minimum required Ruby version to 3.0 in Ruby bindings.

### DIFF
--- a/.github/workflows/ci-ruby.yml
+++ b/.github/workflows/ci-ruby.yml
@@ -39,9 +39,9 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - ruby-version: 2.7.6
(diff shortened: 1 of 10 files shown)


titusfortner (approved):
Can we also update rubocop target?

p0deje:
> Can we also update rubocop target?

Do you mean https://github.com/SeleniumHQ/selenium/pull/11966/files#diff-010ef45e983d9d8b2649f7c517543a64e66362de0d8ef845df7527c634496e0f?

titusfortner:
yes 😆

PR #11967, opened 2023-04-28
Change Ruby Logging

titusfortner:
### Description

Change default logging level to `:info` and provide more granular filtering options

### Motivation and Context

I'm looking at how logging is being done in other languages, and as part of putting this documentation together - https://www.selenium.dev/documentation/webdriver/troubleshooting/logging/ I thought we should change our implementation to match the other languages.

### Details:

To better manage the flood of details now in `:debug` mode, I've added `#select` and `#reject` methods where you can opt in or opt out of specific types of logs.

```rb
Selenium::WebDriver.logger.select(:selenium_manager)
```

```rb
Selenium::WebDriver.logger.reject(%i[deprecations logger_info])
```

Let me know if all of these choices make sense.

### DIFF
--- a/rb/lib/selenium/webdriver/common/driver_finder.rb
+++ b/rb/lib/selenium/webdriver/common/driver_finder.rb
@@ -28,7 +28,8 @@ module Selenium
         path ||= begin
           SeleniumManager.driver_path(options)
         rescue StandardError => e
-          WebDriver.logger.warn("Unable obtain driver using Selenium Manager\n #{e.message}")
+          WebDriver.logger.warn("Unable to obtain driver using Selenium Manager\n #{e.message}",
+                                id: :selenium_manager)
           nil
         end
         msg = "Unable to locate the #{klass::EXECUTABLE} executable; for more information on how to install drivers, " \
--- a/rb/lib/selenium/webdriver/common/logger.rb
+++ b/rb/lib/selenium/webdriver/common/logger.rb
@@ -48,14 +48,17 @@ module Selenium
       #
       # @param [String] progname Allow child projects to use Selenium's Logger pattern
       #
-      def initialize(progname = 'Selenium', ignored: nil)
+      def initialize(progname = 'Selenium', ignored: nil, allowed: nil)
         @logger = create_logger(progname)
         @ignored = Array(ignored)
+        @allowed = Array(allowed)
         @first_warning = false
       end
 
       def level=(level)
-        info(':info is now the default log level, to see additional logging, set log level to :debug') if level == :info
+        if level == :info && @logger.level == :info
+          info(':info is now the default log level, to see additional logging, set log level to :debug')
+        end
 
         @logger.level = level
       end
@@ -87,10 +90,19 @@ module Selenium
       #
       # Will not log the provided ID.
       #
-      # @param [Array, Symbol] id
+      # @param [Array, Symbol] ids
       #
-      def ignore(id)
-        Array(id).each { |ignore| @ignored << ignore }
+      def ignore(*ids)
+        @ignored += Array(ids).flatten
+      end
+
+      #
+      # Will only log the provided ID.
+      #
+      # @param [Array, Symbol] ids
+      #
+      def allow(ids)
+        @allowed += Array(ids).flatten
       end
 
       #
@@ -144,7 +156,10 @@ module Selenium
       # @yield appends additional message to end of provided template
       #
       def deprecate(old, new = nil, id: [], reference: '', &block)
+        id = Array(id)
         return if @ignored.include?(:deprecations)
+
+        id << :deprecations if @allowed.include?(:deprecations)
 
         message = +"[DEPRECATION] #{old} is deprecated"
         message << if new
@@ -177,6 +192,7 @@ module Selenium
       def discard_or_log(level, message, id)
         id = Array(id)
         return if (@ignored & id).any?
+        return if @allowed.any? && (@allowed & id).none?
 
         msg = id.empty? ? message : "[#{id.map(&:inspect).join(', ')}] #{message} "
         msg += " #{yield}" if block_given?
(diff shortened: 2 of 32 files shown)


p0deje:
I'm looking through https://www.selenium.dev/documentation/webdriver/troubleshooting/logging/, but I don't see what exactly we should match in Ruby. Can you point to Java/Python/etc examples?

titusfortner:
I meant the levels.

* Actionable items
* Useful information
* Debugging details

Right now `:warn` is a mix of info and actionable
And `:info` is essentially another level of `:debug` details

I wish Ruby had more "debug" levels like the other languages, but alas.

rb/lib/selenium/webdriver/common/driver_finder.rb
```diff
@@ -27,8 +27,7 @@ def self.path(options, klass)
 
         path ||= begin
           SeleniumManager.driver_path(options)
-        rescue StandardError => e
-          WebDriver.logger.warn("Unable obtain driver using Selenium Manager\n #{e.message}")
```
p0deje, line 31: Why do we remove it?

rb/lib/selenium/webdriver/common/logger.rb
```diff
@@ -83,32 +91,59 @@ def io
       #
       # @param [Array, Symbol] id
       #
-      def ignore(id)
-        Array(id).each { |ignore| @ignored << ignore }
+      def reject(id)
+        Array(id).each { |key| @reject << key }
+      end
+      alias ignore reject
```
p0deje: What do you think about using splat arguments for those?

```suggestion
      def reject(*ids)
        @reject += ids
      end
      alias ignore reject
```

rb/lib/selenium/webdriver/common/logger.rb
```diff
@@ -83,32 +91,59 @@ def io
...
+
+      #
+      # Will only log the provided ID.
+      #
+      # @param [Array, Symbol] id
+      #
+      def select(id)
```
p0deje: Ditto.

rb/lib/selenium/webdriver/common/logger.rb
```diff
@@ -83,32 +91,59 @@ def io
       #
       # @param [Array, Symbol] id
       #
-      def ignore(id)
-        Array(id).each { |ignore| @ignored << ignore }
+      def reject(id)
```
p0deje: Frankly, having `select/reject` methods look strange to me on logger because those are typically methods on enumerable. I can see why they are there, but I would need to consult the documentation to understand their purpose. What do you think about using more discoverable names such as `allow/ignore`?

rb/lib/selenium/webdriver/common/logger.rb
```diff
@@ -133,8 +169,9 @@ def deprecate(old, new = nil, id: [], reference: '', &block)
                      ' and will be removed in a future release.'
                    end
         message << " See explanation for this deprecation: #{reference}." unless reference.empty?
+        message += " #{yield}" if block_given?
 
-        warn message, &block
+        @logger.warn { message }
```
p0deje: The code is very similar to `#discard_or_log`. I suggest you refactor it so that it accepts `@reject/@select` as the parameters. Then you can simply call this method with an extra `:deprecations` id.

rb/lib/selenium/webdriver/common/logger.rb
```diff
@@ -38,20 +38,28 @@ class Logger
...
+                     :level
 
       #
       # @param [String] progname Allow child projects to use Selenium's Logger pattern
       #
-      def initialize(progname = 'Selenium', ignored: nil)
+      def initialize(progname = 'Selenium', ignored: nil, select: nil, reject: nil)
         @logger = create_logger(progname)
-        @ignored = Array(ignored)
+        @reject = Array(Array(reject) + Array(ignored))
```
p0deje: I suggest you use sets for `@reject/@select`. This way you would have cleaner APIs (https://ruby-doc.org/stdlib-2.7.1/libdoc/set/rdoc/Set.html) when checking ids and won't have to worry about uniqueness and order.

rb/lib/selenium/webdriver/common/driver_finder.rb
titusfortner: The idea was we don't have to log this if we're raising the error message, but I don't think we have the right error message yet, so better to keep this for now.

rb/lib/selenium/webdriver/common/logger.rb
titusfortner: yeah, I wasn't sure about these names, and I think they are confusing since they are different from how they are used elsewhere. I couldn't think of a good "opposite" for `ignore`, but `allow` makes sense.

rb/lib/selenium/webdriver/common/logger.rb
titusfortner: I guess I would prefer to use Arrays because there isn't exactly a "problem" with duplication, and Ruby makes arrays so much easier/flexible with splats, etc.

rb/lib/selenium/webdriver/common/logger.rb
titusfortner: Hmm, would be nice to support `ignore(:foo, :bar, :baz)` as well as `ignore(%i[foo bar baz])`

titusfortner: but I guess if choosing one, arguments are better than Array?

p0deje: I think it's the perfect case for splat arguments.

titusfortner:
```
        @ignored += Array(ids).flatten
```

titusfortner:
I fixed the things you suggested, including changing the names. I'd prefer Arrays to Sets still just because that's what people are more used to... Methods now accept one or more params and Arrays. 

I also added logic for setting the default level in constructor. Webdrivers gem and Watir both use the Selenium Logger, and don't want to force the change on them?

p0deje:
> I fixed the things you suggested, including changing the names. I'd prefer Arrays to Sets still just because that's what people are more used to... Methods now accept one or more params and Arrays.

My proposal was to convert arrays to sets internally, not to force people into using sets.

p0deje (approved):
LGTM but CI is failing.

Issue #11643, opened 2023-02-10
[🚀 Feature]: Allow Driver Services to set Service Output


titusfortner:
### Feature and motivation

Geckodriver has this great feature allowing users to specify the output stream of the service process.

https://github.com/SeleniumHQ/selenium/blob/selenium-4.8.0/java/src/org/openqa/selenium/firefox/GeckoDriverService.java#L203-L225

It should be pretty trivial to support this in the other Driver Service classes as well.

For Chrome/Edge we are currently using the driver itself to send logged information to a file:
```
        args.add(String.format("--log-path=%s", getLogFile().getAbsolutePath()));
```

I think porting all of the process output is a better option, though we probably need some tests to verify there isn't some other oddness as a result.

One big difference is that if we are sending the whole stream to another place, the user won't have to see this in their console:
```
Starting ChromeDriver 109.0.5414.74 (e7c5703604daa9cc128ccf5a5d3e993513758913-refs/branch-heads/5414@{#1172}) on port 26243
Only local connections are allowed.
Please see https://chromedriver.chromium.org/security-considerations for suggestions on keeping ChromeDriver safe.
ChromeDriver was started successfully.
```

### Usage example

```
System.setProperty("webdriver.chrome.logfile", "/dev/null")
```

titusfortner:
Few more things:
I would expect this to output the default info to the log, but it still sends the above 4 lines to stderr
```
ChromeDriverService service = new ChromeDriverService.Builder().withLogFile(new File("chromedriver.log")).withLogLevel(ChromiumDriverLogLevel.INFO).build();
```

Also, `ChromeDriverService` should evaluate property files at runtime instead of compile time.

titusfortner:
For Geckodriver logs, why does the system property take precedence over a file passed in? 
Maybe something is passing in a log file by default that shouldn't be and giving priority to the system property was a way to ensure that you could turn off the logging?

diemol:
Closed via #11973

Issue #11410, opened 2022-12-10
[🚀 Feature]: Ensure All Service Classes Support Log Levels


titusfortner:
### Feature and motivation

As I'm working on alternative ways to implement #11376, I'm seeing a lot of differences in how the drivers are getting handled between browsers & languages.

### Language Differences
Nothing shocking here.
* Ruby & Python let you pass in whatever args you want in a list; everything is in user's hands.
* .NET requires adding explicit support via Properties in Service classes; straightforward, but obviously we're now missing things as stuff changes.
* Java is even more fun because of the Grid. Each argument can be specified as a System Property or explicitly in the `DriverService.Builder()`.

### Browser Differences
Here's where it gets a little weird. Most of the things you set via Service classes are logging related, so that's what this is focusing on. I'm focusing on Java because it's the hardest use case, but most of this stuff applies in all the bindings.

* Firefox - you can set `{"moz:firefoxOptions": {"log": {"level": "trace"}}}`, but this only tells the driver to turn on logging. If you are running locally, this gets sent to stdout in your executing code, but if you are running remotely, it gets sent to stdout in the grid console (which the client can't access). The above capabilities output the *exact* same thing locally as starting geckodriver with `--log trace`. You can't set a logfile in the driver arguments directly, but Java has an extra method on the `DriverService.Builder()` class that will pipe stdout to a file or dev/null to simulate this; not all of the other languages do this right now, and we probably want to ensure this is possible.

* Chrome - Java added an implementation so log level can be set in `ChromeOptions` the same as in `FirefoxOptions`, and then if you pass that to the DriverService with `ChromeDriverService service = ChromeDriverService.createServiceWithConfig(options)`, and pass that `service` variable as an argument to `ChromeDriver` it sets the argument on the driver, and essentially works the same as Firefox.

* Edge doesn't do what Chrome does, setting log level requires using a `EdgeDriverService.Builder()` method. I created #11404 thinking to add the same functionality to Edge that Chrome has, but... is this the right way to do this? It feels weird to add methods to an Options class that only get consumed by a Service class and are only applicable locally. 

* Safari - we don't have it implemented anywhere (yet), but you can pass `--diagnose` to the safaridriver, and the results end up in `~/Library/Logs/com.apple.WebDriver/`. 🤯 

* IE - also has ability to pass in `--log-level` to Service class, haven't spent any time with it, yet, because VMs are a pain.

### Now What?
Alternatives:
[...]

PR #11973, opened 2023-05-01
Make Driver Services consistent

titusfortner:
### Description

Different drivers currently have different behaviors. This brings everything into agreement.

### Motivation and Context

Addresses #11643 
This is also in accordance with what we discussed in #11410 as well as:
* Remove / Deprecate undesired behavior
* Improve logging ability

### Specifics:

1. Deprecates the constructors that do not include the timeout (users should be using builder anyway; IE has has even kept the constructor private for this reason)
2. Deprecates the special Firefox behaviors that came from Legacy driver (`FirefoxBinary` related)
3. Adds support for turning on Safari driver logging, including preventing users from calling superclass method to set log file location (which isn't supported)
4. Adds missing features to GeckoDriver service
    * log level
    * log truncation
    * setting alternate profile directory ([required for snap](https://github.com/mozilla/geckodriver/issues/2010))
5. Allows Chrome, Edge & IE to send all output from the process to the location of user's choosing.

The last item is the biggest one. By default geckodriver logs everything at an INFO level to stderr. This is annoying, so it was allowed to use the System Property `webdriver.firefox.logLevel` to set the `"/dev/null"` for it to discard the output. But only being able to do this by using a system property is bad, so I added methods and support for it in all browsers.

For comparison, Ruby has always sent everything from driver process to null output as a default.

This PR allows either of these:
```java
    ChromeDriverService service = new ChromeDriverService.Builder()
        .withLogOutput(DriverService.LOG_NULL)
        .build();
```
or
```java
    System.setProperty(ChromeDriverService.CHROME_DRIVER_LOG_PROPERTY,
        DriverService.LOG_NULL);
```


java/src/org/openqa/selenium/edge/EdgeDriverService.java
```diff
@@ -248,16 +231,24 @@ public Builder withLoglevel(ChromiumDriverLogLevel logLevel) {
      * @return A self reference.
      */
     public Builder withSilent(boolean silent) {
-      if (silent) {
-        this.logLevel = ChromiumDriverLogLevel.OFF;
-      }
```
diemol: I think you need to assign the value to `this.silent`, which is what you did in the `ChromeDriverService`.

java/src/org/openqa/selenium/edge/EdgeDriverService.java
```diff
@@ -248,16 +231,24 @@ public Builder withLoglevel(ChromiumDriverLogLevel logLevel) {
...
     /**
-     * Configures the comma-separated list of remote IPv4 addresses which are allowed to connect
-     * to the driver server.
+     * Configures the driver server verbosity.
+     *
+     * @param verbose whether verbose output is used
+     * @return A self reference.
+     */
+    public Builder withVerbose(boolean verbose) {
+      this.verbose = false;
```
diemol, line 258: I think you need to assign the value to `this.verbose`, otherwise it gets ignored.

diemol (changes requested):
Thanks!

A couple of comments.

diemol (approved):

diemol (approved):

PR #12007, opened 2023-05-08
[java] make all logging variable names consistent

titusfortner:
### Description
All logging uses:
```java
private static final Logger LOG
```

### Motivation and Context
Make convention the same throughout the codebase


diemol (approved):
Thank you!

PR #12011, opened 2023-05-08
Add select_all feature to select module (select.py)

### LINKED ISSUE #12010: [🚀 Feature]: add select_all
### Feature and motivation

In the [select module](https://github.com/SeleniumHQ/selenium/blob/trunk/py/selenium/webdriver/support/select.py) we find `deselect_all` which simplifies deselecting all options. This proposal is to add `select_all` to simplify selecting all options. 

### Usage example

So we can write

```
Select(driver.find_element_by_id('select_id')).select_all() # assuming tag supports "multiple"
```

instead of getting the options, looping through each, and trying to call one of the `select_by...` calls (like this [example](https://www.lambdatest.com/blog/handling-dropdown-in-selenium-webdriver-python/))

ShahJahanIshaq:
### Description
The select_all feature selects all the options in a select element that either has multiple select enabled or has only one option. It serves as a counterpart to the already existing feature deselect_all.

### Motivation and Context
Fixes #12010. 
Simplifies selecting all options in a multi-select SELECT element. 

### Types of changes

### Checklist


titusfortner:
I'm not going to weigh in on whether this PR should be merged or not, other than to point out that none of the other bindings have this feature.

I do want to reiterate that I think this whole class is "out of scope" for Selenium and should be put in another repo or at least packaged separately.

ShahJahanIshaq:
> I'm not going to weigh in on whether this PR should be merged or not, other than to point out that none of the other bindings have this feature.
> 
> I do want to reiterate that I think this whole class is "out of scope" for Selenium and should be put in another repo or at least packaged separately.

Thank you for your feedback on my pull request, @titusfortner. I appreciate your concern about the scope of the Select class in Selenium and the suggestion to package it separately. However, I believe that the select_all feature is a valuable addition to the Select class, as it solves a common issue that users may encounter when working with select elements on web pages.

If it would be more appropriate, I would be happy to consider implementing the feature in a separate module or package, but I believe that it would still be a useful addition to the Selenium library. Please let me know if you have any further feedback or suggestions on how to proceed.

titusfortner:
Sorry for being unclear, that wasn't intended as feedback on your PR. I was using this PR as an excuse to further a conversation with other Selenium Devs about the direction of the project.

Regardless, thanks for finding something that you think can be made better and submitting code for it.

diemol:
During the meeting on [25.05.2023](https://www.selenium.dev/meetings/2023/tlc-05-25/) it was decided to not continue adding features to Support classes:
- That is the direction we should go because it would set the project to be a base library, rather than something in between. A more clear concept. Also, with the people we have at the moment, we cannot maintain more stuff.
- The standard should be bug fixes only. For new features: please copy & paste/subclass/fork and change whatever you want. 


@ShahJahanIshaq thank you for taking the time to raise this PR, but given the decision taken on the linked meeting, we won't be accepting this PR. If you are interested in contributing more, [please reach out to us](https://www.selenium.dev/support/), and I am sure there are more opportunities to contribute.

Issue #11881, opened 2023-04-11
[🐛 Bug]: On grid, session is started but never used


bhecquet:
### What happened?

I encounter an issue when trying to execute more tests than a Selenium Grid can handle

My setup:
- a grid Hub: `java -jar selenium-server-4.8.3.jar hub --session-request-timeout 30 --port 4448`
- a grid node: `java -jar selenium-server-4.8.3.jar node --max-sessions 1 --port 6666 --grid-url http://selenium-server.company.com:4448  --publish-events tcp://selenium-server.company.com:4442 --subscribe-events tcp://selenium-server.company.com:4443`
- 4 instances of a test (see Java+TestNG test script below)

The script adds a wait to force new sessions to be queued
So when starting the 4 instances (methods test1, test2, test3 and test4), I can see them run, or queued, but at some point, a session is created but not used: on the node, browser is started, but URL not loaded.
In that case, we need to wait 5 mins (default session timeout) for next test to be started

From what I observed, I would say that there is a race condition between the session request timeout and the session creation thread, as the "stalled" session is logged as timed out in test script logs: `Cannot create driver: Could not start a new session. Response code 500. Message: Could not start a new session. New session request timed out `

Thanks for your help

### How can we reproduce the issue?

```shell
package company;

import org.openqa.selenium.By;
import org.openqa.selenium.Capabilities;
import org.openqa.selenium.MutableCapabilities;
import org.openqa.selenium.WebDriver;
import org.openqa.selenium.chrome.ChromeDriver;
import org.openqa.selenium.chrome.ChromeOptions;
import org.openqa.selenium.remote.RemoteWebDriver;

import java.net.MalformedURLException;
import java.net.URL;

public class Test {

    @org.testng.annotations.Test
    public void testMethod(String name) throws MalformedURLException, InterruptedException {
        for (int i = 0; i < 10; i++) {
            MutableCapabilities caps = new ChromeOptions();
            caps.setCapability("cs:test", String.format(name + "-" + i));
            try {
                System.out.println(String.format("Starting session %s: %d", name, i));
                WebDriver driver = new RemoteWebDriver(new URL("http://selenium-server.company.com:4448/wd/hub"), caps);
                System.out.println(String.format("Started session %s: %d", name, i));
                try {
                    int wait = (int) (Math.random() * 30000);
                    driver.get("https://www.selenium.dev/");
                    driver.manage().window().maximize();
                    driver.findElement(By.linkText("Documentation")).click();
[...]

PR #12014, opened 2023-05-09
Prevent grid from creating sessions that are about to timeout in queue (corrects issue #11881)

bhecquet:
### Description
As stated in issue #11881, when Selenium grid receives more session requests than it can handle, some sessions may be created whereas the client that requests it receives a timeout response.
This bug only shows up when `--session-request-timeout` option is set

### Motivation and Context
On our grid setup, for some reasons, many grid nodes may become unavailable (upgrade, electrical failure, network failure, ...)
In this case the remaining nodes are not able to handle all the incoming session requests. These requests goes to queue, but we observed that at some point, a browser may be created on node and never used (it remains in this state 5 minutes) before session is killed by grid.

This is due to the fact (details and logs can be found in mentionned issue: https://github.com/SeleniumHQ/selenium/issues/11881#issuecomment-1532552648)
- LocalNewSessionQueue waits at most 30 seconds (the configured session request timeout) before sending timeout to client
- nothing prevents the distributor to create the requested session just before that (e.g: at 29 secs)
- session creation takes about 2 seconds so when session si created, client has already received the timeout exception and will never use the created session

Correction takes place in LocalNewSessionQueue. When session request timeout expires, we first check if the session is still in queue or not (session not in queue means distributor is about to create it). In the latter case we wait a bit more so that distributor has time to create the session.

### Types of changes

### Checklist

Remark: added tests are a bit long to run to test timeout behaviour


diemol (changes requested):
Thank you for this PR!

I think you have a good point, where the current logic allows for a window where the timeout is triggered but still the Distributor can create a session that will end up being orphaned. 

However, I do not think the approach in the PR is completely correct, or clean at least. This PR is basically waiting 5 seconds more for the session request to resolve.

A nicer approach, I think, is to put the logic you are putting in `isRequestInQueue`as a filter in `timeoutSessions`. With that, we timeout only session requests that are present in the `queue`.

Something around the lines of:

```java
  private void timeoutSessions() {
    Instant now = Instant.now();

    Lock readLock = lock.readLock();
    readLock.lock();
    Set<RequestId> ids;
    try {
      ids =
          requests.entrySet().stream()
              .filter(
                  entry ->
                      queue.stream()
                          .anyMatch(
                              sessionRequest ->
                                  sessionRequest.getRequestId().equals(entry.getKey())))
              .filter(entry -> isTimedOut(now, entry.getValue()))
              .map(Map.Entry::getKey)
              .collect(ImmutableSet.toImmutableSet());
    } finally {
      readLock.unlock();
    }
    ids.forEach(this::failDueToTimeout);
  }
```

(Might need testing, I just quickly tried this).

What do you think?

bhecquet:
Hello

I think that my PR needed to be improved but, correct me if I'm wrong, the timeoutSessions method is only called once every 10 seconds by default and would not prevent the current problematic behaviour
Maybe we should (Or I misunderstand your exemple) combine both approaches:
- keep my change in 'addToQueue' method because there is no way to tell the latch not to timeout if session is being created
- take your suggestion for 'timeoutSessions' method which is way better than increasing the whole timeout

bhecquet:
I changed the code as suggested

diemol:
> Hello
> 
> I think that my PR needed to be improved but, correct me if I'm wrong, the timeoutSessions method is only called once every 10 seconds by default and would not prevent the current problematic behaviour Maybe we should (Or I misunderstand your exemple) combine both approaches:
> 
>     * keep my change in 'addToQueue' method because there is no way to tell the latch not to timeout if session is being created
> 
>     * take your suggestion for 'timeoutSessions' method which is way better than increasing the whole timeout

I think you are right to point out that we were timing out session requests that were being processed by the Distributor. Now that this PR is taking that into account, it should not happen anymore. Thank you.

diemol (approved):

diemol:
Oh wait, somehow when I reviewed the change set I only saw the filter added, not the initial code. I will revert the piece that adds the 5 seconds and go with the filter only. If the bug is still present, we can consider the other part of the PR.

diemol:
Well, the tests you added are now failing. So maybe all the code you added is needed 😄 

I will leave it as it is, but hopefully you can be around in case bugs are raised.

bhecquet:
Yes, I'll be happy to contribute improve it if necessary

PR #12023, opened 2023-05-10
Automatically format the code

shs96c:
This PR features two diffs that should be landed separately. The first adds a script to reformat code within the project. The second reformats the code.

While this currently only handles Java, the plan is to extend this to handle other languages too.


titusfortner:
Ruby is already taken care of very nicely with rubocop.

I know @symonk was doing a lot of work with Python, and I think there are some automated tools to address issues, but we weren't ready for them or something. :) 

No idea on .NET & JavaScript

PR #12024, opened 2023-05-10
[java][dotnet] Removing RC and WebDriverBackedSelenium

diemol:
This PR removes a large amount of code we have not been releasing for years, users are not requesting new releases, and it will help current and future contributors navigate through the code base. 

I've run all tests locally and they pass, but double checking with CI.


dotnet/src/webdriver/WebDriver.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFrameworks>net6.0;net5.0;net48;net47;net46;net45;netstandard2.1;netstandard2.0</TargetFrameworks>
     <AssemblyName>WebDriver</AssemblyName>
     <RootNamespace>OpenQA.Selenium</RootNamespace>
+    <Configurations>Release;Debug</Configurations>
```
nvborisenko, line 7: Any particular reason to specify the supported configurations per project? I supposed we can even reduce this file, but you add more lines.

diemol: My bad, I think the IDE added that by default. I noticed that and I was going to revert the change and then I just forgot.
Reverted through https://github.com/SeleniumHQ/selenium/commit/b9ba020876bfa5222cca30e696e2644ee6fcb580

PR #12029, opened 2023-05-12
[py] fix proxy options

titusfortner:
### Description
Allow Proxy to be used in any browser

### Motivation and Context
I added proxy property to Base Options class right before Se 4.0, but I didn't add the part that would actually add it to the capabilities. So right now I think proxy only works as intended for Firefox.

I'm removing `add_to_capabilities()` because it's a weird way to serialize a class, and also because capabilities is deprecated so no one should be passing it around. If we're just using options class, then we just need a way to serialize it, so I'm renaming the logic `to_capabilities()`. When we serialize it in options, we don't have to add logic in remote webdriver class.

### DIFF
--- a/py/selenium/webdriver/common/proxy.py
+++ b/py/selenium/webdriver/common/proxy.py
@@ -284,13 +284,8 @@ class Proxy:
                 f"Specified proxy type ({compatible_proxy}) not compatible with current setting ({self.proxyType})"
             )
 
-    def add_to_capabilities(self, capabilities):
-        """Adds proxy information as capability in specified capabilities.
-
-        :Args:
-         - capabilities: The capabilities to which proxy will be added.
-        """
-        proxy_caps = {"proxyType": self.proxyType["string"]}
+    def to_capabilities(self):
+        proxy_caps = {"proxyType": self.proxyType["string"].lower()}
         if self.autodetect:
             proxy_caps["autodetect"] = self.autodetect
         if self.ftpProxy:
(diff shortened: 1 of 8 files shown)


py/selenium/webdriver/common/proxy.py
```diff
@@ -284,13 +284,8 @@ def _verify_proxy_type_compatibility(self, compatible_proxy):
                 f"Specified proxy type ({compatible_proxy}) not compatible with current setting ({self.proxyType})"
             )
 
-    def add_to_capabilities(self, capabilities):
```
diemol, line 287: Minor comment, wouldn't this break existing implementations?

diemol (commented):
This looks fine to me, but wondering what @AutomatedTester and @symonk think.

py/selenium/webdriver/common/proxy.py
titusfortner: Yeah, this is the one contentious piece, which is why I explained my reasoning in the Motivation section above. I'm pretty sure it only breaks if people are using deprecated classes. This is the hard bit of trying to juggle changes in too many places; I was hoping to get #12030 in the same release and it wouldn't matter.

diemol: I think the code is correct. I believe we should merge this, otherwise we will stall for longer on this.

zjtalia:
Would you be open to a PR that re-implements `add_to_capabilities()` with a deprecation warning? This change breaks backward compatibility without specific mention of the removed method in the change log. I understand that it was a deprecated feature, but it caught us by surprise.

PR #12030, opened 2023-05-12
[py] Remove deprecated code in driver classes

titusfortner:
### Description

* `get_remote_connection()` is kind of a clever hack to be able to use browser specific methods in remote webdriver class. I would love if we can get a mixin or something to better support these. Regardless, what is there currently does not work for Chrome or Edge because the `browser_name` of the `ChromiumRemoteConnection` is `None`.
* Removed the keep_alive deprecation notice. It doesn't belong in Service class, so it is better off where it is still. Probably what we want is to have a "ClientConfig" class like Java where user can specify keep_alive, proxy, timeouts (and number of retries?). It would also let us better handle `_ignore_local_proxy` which is currently getting set in Options. Anyway, future PR.
* The driver class constructors have defaults so I removed some of the "if None" type protections that I don't think are needed. I'm pretty sure this logic covers everything.
* Removed w3c option warnings. The driver can error if it has a problem with anything w3c compliant that gets passed along.

### Motivation and Context
It's really difficult to see what is going on in Python code with deprecations everywhere
All of this has been deprecated since 4.0 or earlier.

I'm honestly not sure when to set things to self and when not to, and when to reference things as self, so this code can probably be cleaned up even more.


py/selenium/webdriver/safari/webdriver.py
```diff
@@ -35,79 +34,43 @@ class WebDriver(RemoteWebDriver):
...
-                stacklevel=2,
-            )
+        self.service = service if service else Service()
+        self.options = options if options else Options()
+        self.keep_alive = keep_alive
 
-        self._reuse_service = reuse_service
-        self.service = service or Service(executable_path, port=port, quiet=quiet, service_args=service_args)
-        if not reuse_service:
+        self._reuse_service = reuse_service and self.service.reuse_service
```
AutomatedTester, line 63: Should this not be 
```suggestion
        self._reuse_service = reuse_service or self.service.reuse_service
```

AutomatedTester:
Initial review and looks ok

diemol (approved):

AutomatedTester (approved):

titusfortner:
I tried a few things to get the lint test passing on CI, and nothing is working. Not sure how to proceed since things are passing just fine for me locally.

PR #12033, opened 2023-05-14
[py] Use Config class for http client settings

titusfortner:
### Description
* New `ClientConfig` class with:
  * keep-alive
  * proxy
  * timeout
  * certificate path
* `keep_alive` deprecation changed (it was never supported in Service class)
* `ignore_proxy` deprecated in Options (not an option) 

### Motivation and Context
Python implementation of `command_executor` / `RemoteConnection`  is similar to Java. A user can create the class themselves, but can only use it as a parameter for `RemoteWebDriver`. Local drivers are subclassed and have their `RemoteConnection` classes created for them. 

Java addressed this with the `ClientConfig` class. (Ruby has always had a separate http_client, and I haven't looked at what JS or .NET do).

The only way to use a proxy right now is with environment variables. If a user wants to ignore the environment variables they use `ignore_local_proxy()` on `Options` class. `ClientConfig` takes a `Proxy` instance so the user does not need to set environment variables. Instead of having the user pass `ignore_proxy: true`, the user would pass `proxy: Proxy({"ProxyType": ProxyType.DIRECT})`. If that seems too odd, I can add back the ignore_proxy parameter, but only having the "right" proxy object seems more "correct."

Example usage:
```py
proxy = Proxy()
proxy.http = "http://localhost:8080"
client_config = ClientConfig(proxy=proxy, keep_alive=False, timeout=180)
driver = webdriver.Chrome(client_config=client_config)
```

### DIFF
--- a/py/selenium/webdriver/remote/remote_connection.py
+++ b/py/selenium/webdriver/remote/remote_connection.py
@@ -27,8 +27,10 @@ import certifi
 import urllib3
 
 from selenium import __version__
+from selenium.webdriver.common.proxy import ProxyType
 
 from . import utils
+from .client_config import ClientConfig
 from .command import Command
 from .errorhandler import ErrorCode
 
@@ -38,8 +40,7 @@ LOGGER = logging.getLogger(__name__)
 class RemoteConnection:
     """A connection with the Remote WebDriver server.
 
-    Communicates with the server using the WebDriver wire protocol:
-    https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol
+    Communicates with the server using the w3c WebDriver protocol:
     """
 
     browser_name = None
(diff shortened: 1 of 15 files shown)


AutomatedTester:
Initial pass through and it looks ok

titusfortner:
This needs rebasing after other PR merged

py/selenium/webdriver/remote/remote_connection.py
```diff
@@ -152,10 +159,17 @@ def _get_connection_manager(self):
 
         return urllib3.PoolManager(**pool_manager_init_args)
 
-    def __init__(self, remote_server_addr: str, keep_alive: bool = False, ignore_proxy: bool = False):
-        self.keep_alive = keep_alive
+    def __init__(self, remote_server_addr: str, client_config: ClientConfig = ClientConfig()):
```
isaulv, line 162: So in Python, any object that is a function default is "cached", unlike Ruby, where `client_config` will always get a new instance of `ClientConfig`. 

This may affect scripts that run multiple browsers in one session, if they don't set a different default client config.

isaulv:
I like this idea of a config class. I would like an easier interface to swap out `RemoteConnection` for Python as well. I think it's kind of there but it's a process.

py/selenium/webdriver/remote/remote_connection.py
titusfortner: So the better pattern is to use None as default and then set it in the constructor body?

titusfortner:
Well, I'm still not sure the right way to get browser specific features in a remote session. Someone did some magic at one point, but then I think there was an issue. I never followed up.

titusfortner:
I'm breaking this up into pieces for my sanity and to make sure I'm doing this properly.
See: https://github.com/SeleniumHQ/selenium/pull/12364

Issue #11868, opened 2023-04-05
[🐛 Bug]: Constantly hanging connections and threads created by dev tools, holding connections on jdk11 http client.


pogrebnyaksa:
### What happened?

Constantly hanging connections and threads created by dev tools, holding connections on jdk11 http client.

### How can we reproduce the issue?

```shell
public static void main(String[] args) throws InterruptedException {
        ChromeOptions chromeOptions = new ChromeOptions();
        chromeOptions.addArguments("--ignore-certificate-errors",
                "--start-maximized",
                "--no-sandbox",
                "--dns-prefetch-disable",
                "--disable-gpu");

        System.setProperty("webdriver.http.factory", "jdk-http-client");
        var start = Instant.now();
        Duration loopDuration = Duration.ofMinutes(3);
        while (Instant.now().minus(loopDuration.toMillis(), ChronoUnit.MILLIS).isBefore(start)) {
            WebDriver driver = null;
            try {
                driver = RemoteWebDriver.builder()
                        .augmentUsing(new Augmenter())
                        .oneOf(chromeOptions)
                        .address(System.getProperty("GRID_URL"))
                        .build();
                try (DevTools devTools = ((HasDevTools) driver).getDevTools()) {
                    devTools.createSession();
                    devTools.send(Network.enable(Optional.empty(), Optional.empty(), Optional.empty()));
                    devTools.send(Network.setExtraHTTPHeaders(new Headers(Map.of("test", Boolean.TRUE.toString()))));
                    devTools.addListener(Network.requestWillBeSent(), event -> {

                    });
                    driver.get("https://cnn.com");
                }
            } finally {
                if (driver != null) {
                    driver.close();
                }
            }
            TimeUnit.SECONDS.sleep(3);
        }
        while (true) {
            TimeUnit.SECONDS.sleep(10);
        }
    }

run this code for 5 mins, then using jconsole perform full GC and in thread list you can see left hanging threads even after webdriver session was closed and enough time passed to finish all threads. In production env I can see such threads hanging for days. Same issue is observable on selenium 4.8.3, devtools-v111, chromedriver 111.0.5563.64 Stacktrace of thread is:
Name: HttpClient-20-SelectorManager
State: RUNNABLE
Total blocked: 0  Total waited: 0

Stack trace: 
java.base@17.0.5/sun.nio.ch.epoll.wait(Native Method)
[...]

PR #12035, opened 2023-05-14
[java] JdkHttpClient - Close all websockets before shutting down the executor

joerg1985:
### Description
This PR will change the following things:
- Close all websockets not only the last created
- Close the websockets before shutting down the executor, to avoid [JDK-8277969](https://bugs.openjdk.org/browse/JDK-8277969)
- close() can be called multiple times, like the javadoc of Closeable is demanding

### Motivation and Context
Should help with #11868

### Types of changes

### Checklist


diemol (commented):
Thank you for this PR!

I think this makes sense. But I do not understand how we are closing the `underlyingSocket` one, or how the `close` method is actually closing it. Probably I am missing something.

joerg1985:
@diemol sending the close message initiates an orderly closure of the websocket.
from the javadoc: `The sendClose method does not close this WebSocket's input. It merely closes this WebSocket's output by sending a Close message. `

diemol (approved):

PR #12036, opened 2023-05-14
[java] support Selenium Manager using proxy info from options

titusfortner:
This is for Java solution to https://github.com/SeleniumHQ/selenium/issues/11294

I tested it and made sure it worked locally.

My primary question is how to handle http vs ssl. Does this logic look correct?


bonigarcia (approved):
LGTM.

diemol (approved):
Thank you!

Issue #11657, opened 2023-02-14
[🚀 Feature]: Add Download Support to Bindings


titusfortner:
### Status

|                      | Java   | Python | Ruby   | .NET   | JavaScript |
|----------------------|--------|--------|--------|--------|------------|
| PR                   | #12979 | #13023 | #12037 | #12981 | #13102          |
| Capability Method    | :white_check_mark:      | :white_check_mark:      | :white_check_mark:      | :white_check_mark:      | :white_check_mark:          |
| Capability Protected | :white_check_mark:      | :white_check_mark:      | :white_check_mark:      | :white_check_mark:      | :white_check_mark:          |
| Name List            | :white_check_mark:      | :white_check_mark:      | :white_check_mark:      | :white_check_mark:      | :white_check_mark:          |
| Directory Download   | :white_check_mark:      | :white_check_mark:      | :white_check_mark:      | :white_check_mark:      | ?          |
| Create Directory   | :white_check_mark:      | :white_check_mark:      |  :white_check_mark:      | :white_check_mark:      | :white_check_mark:          |
| Delete               | :white_check_mark:      | :white_check_mark:      | :white_check_mark:      | :white_check_mark:      | :white_check_mark:          |

### Feature and motivation

For the bindings we want to require the user to both enable downloads since we don't want that to be default behavior, and then have the ability to get the list of available files (#11458), and to download a specific file (#11466).

This is going to rely on Grid implementation in #11656 

My Java syntax proposal is below, feedback appreciated.

### Usage example

```
chromeOptions.enableDownloads(true);
driver = new RemoteWebDriver(url, chromeOptions);

driver.get(siteUrl);
driver.findElement(By.id("download-this")).click();
List<Map<String, String>> downloadedFiles = ((HasDownloads) driver).getDownloads();
String fileToDownload = downloadedFiles.get(0).get("name");
((HasDownloads) driver).getDownload(fileToDownload, outputStream);
```

titusfortner:
@harsha509 can you help me with the PR I made and tell me if it should work?

harsha509:
Closing as completed!

PR #12037, opened 2023-05-15
[rb] implement remote downloads

titusfortner:
### Status
Ruby delete fails because of https://github.com/SeleniumHQ/selenium/pull/13025
I rebased this PR on that fix to make sure it passes

### Description
* creates a new `HasFileDownloads` module with `#downloadable_files` and `#download_file`
* Implemented `Options#enableDownloads` to allow downloading
* Only `RemoteWebDriver` includes this new module
* Rather than put the implementation of these methods into `Remote::Bridge`, I created `Remote::Features` similar to how we've done the browser features files
* Moved the upload implementation into that new feature file
* In order to support both browser & remote command list, a new `#command_list` created and `Remote::Features#add_commands` for `Remote::Driver` to use
* Had to do custom processing of `se:downloadsEnabled`

### Motivation and Context

This implements #11657


p0deje (approved):

titusfortner:
I want to hold off on this until after we releases 4.10 to keep pace with other bindings. Wanted a proof of concept in Ruby to make it easier to do it in other languages (maybe)

diemol (commented):
Maybe I missed this because I am not very fluent in Ruby... But I think we should return an error when the capability `se:downloadsEnabled` is not present.

titusfortner:
Won't the node do this? We're not currently doing this elsewhere in ruby code, though I guess we could...

diemol:
I guess I was thinking of disabling the method in some way or throwing an error. People will try this with any Grid they have available because most users do not read the prerequisites. Also, it won't work with most cloud vendors for now, so trying to avoid throwing work to their customer support teams.

titusfortner:
Makes sense. We should do the same with bidi capability, then.

Oddly, ruby used to do this with the javascript enabled capability and was the only one, and this is what prevented selenium 2 ruby tests from working with Selenium 3 grid.

giladshanan:
@titusfortner any insight on https://github.com/SeleniumHQ/selenium/issues/13307?

PR #12038, opened 2023-05-15
Replace GetCallingAssembly with GetExecutingAssembly

MichalStrehovsky:
`GetCallingAssembly` is not supported with .NET 7 PublishAot.

It's not clear why `GetCallingAssembly` was chosen - `GetCallingAssembly` returns the assembly of the method that is calling this method. If two different assemblies were to call `ResourceUtilities.AssemblyVersion`, this code would cache the assembly of the first caller and produce what is essentially an incorrect result for the second caller.

Given the only caller of `ResourceUtilities.AssemblyVersion` is within the same assembly, it's likely `Assembly.GetExecutedAssembly()` was intended (it's also indicated by the name of the local variable).

Fixes dotnet/runtime#86202.
Fixes #8549 (that was won't fixed but it's the same problem).

Issue #8549, opened 2020-07-19
CoreRT AOT Execution Incompatibility


josephmoresena:
## 🐛 Bug Report

It is not possible to run a CoreRT compiled program using Selenium + FirefoxDriver. An exception occurs when trying to create instances of FirefoxOptions and FirefoxDriver.

## To Reproduce

Simply follow the instructions to create a console executable from the CoreRT repo and call some constructor for the FirefoxProfile and FirefoxDriver classes.

Compiling the program using CoreRT and when executing, an exception will appear in the lines of the call of said constructors due to the use of System.Reflection.Assembly.GetCallingAssembly () since it is not supported in the AOT environment of CoreRT.

Detailed steps to reproduce the behavior:

## Expected behavior

It was expected to be able to instantiate objects using the constructors of the FirefoxDriver and FirefoxProfile classes without getting any kind of exception or at least any solvable from the rd.xml file.

## Test script or set of commands reproducing this issue

	using System;
	using System.Runtime.InteropServices;
	using System.Text;
	using OpenQA.Selenium;
	using OpenQA.Selenium.Firefox;
	using System.Linq;
	using System.Threading;

	namespace ConsoleApp1
	{
		class Program
		{
			static Boolean isEncodingRegistered = !RuntimeInformation.FrameworkDescription.Contains("Core");

			static FirefoxDriverService service;
			static IWebDriver webDriver;

			private static void RegisterEncoding()
			{
				if (!isEncodingRegistered)
				{
					Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);
					isEncodingRegistered = true;
				}
			}

			static Program()
			{
				RegisterEncoding();
				string driverPath = $@"{AppDomain.CurrentDomain.BaseDirectory}";

				service = FirefoxDriverService.CreateDefaultService(driverPath);
				Console.WriteLine("SERVICE");
				FirefoxOptions options = new FirefoxOptions();
				Console.WriteLine("OPTIONS");
				options.AcceptInsecureCertificates = true;
				options.UseLegacyImplementation = true;
				options.Profile = new FirefoxProfile();
				Console.WriteLine("PROFILE");
				webDriver = new FirefoxDriver(service, options);
				Console.WriteLine("DRIVER");
				webDriver.Navigate().GoToUrl("https://www.selenium.dev/");			
				webDriver.FindElements(By.ClassName("nav-item")).FirstOrDefault().Click();
[...]

PR #11871, opened 2023-04-06
[dotnet] Use selenium manager in tests

nvborisenko:
**Thanks for contributing to Selenium!**
Related to #11304 

Now it's not necessary to store driver binaries on machine where tests are executing, Selenium Manager does it.

To tests these changes: remove `c:\Projects\WebDriverServers\` directory and tests should be passed.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist

- [ ] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #12045, opened 2023-05-15
[dotnet] Be more friendly for single file publishing and native compilation

nvborisenko:
### Description
No major changes, just use APIs  which are supported in native compilation.

### Motivation and Context
Selenium is a library, and I don't see reasons why this library cannot be friendly for native compilation.

Addresses:
- https://github.com/dotnet/runtime/issues/86202
- https://github.com/SeleniumHQ/selenium/pull/12038
- https://github.com/SeleniumHQ/selenium/issues/8549

### Area to be improved
"selenium-manager.exe" is packed in nuget package. When user consumes nuget package, then binaries (for all platforms) are copied to user's output. In case of single file publishing the situation is resolved in this PR - binaries are copied to output as they are expected to be there.

In case of native compilation (`PublishAOT` build property we need to be smarter:
- Include selenium-manager as embedded resource into user's assembly
- Try to look up selenium-manager from embedded resources if there are no binaries on the disk


dotnet/src/webdriver/SeleniumManager.cs
```diff
@@ -38,8 +39,45 @@ namespace OpenQA.Selenium
     /// </summary>
     public static class SeleniumManager
     {
-        private static string basePath = System.IO.Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
-        private static string binary;
+        private readonly static string binaryFullPath;
+
+        static SeleniumManager()
+        {
+#if !NET45
```
titusfortner: Since it's only one argument, I think we should switch these and use `if NET45`.

dotnet/src/webdriver/SeleniumManager.cs
```diff
@@ -50,7 +88,7 @@ public static class SeleniumManager
         /// </returns>
         public static string DriverPath(DriverOptions options)
         {
-            var binaryFile = Binary;
+            var binaryFile = binaryFullPath;
```
titusfortner: Do we need this assignment? Can't we just use `binaryFullPath` everywhere once it is defined?

titusfortner (commented):
Were you able to build this to test that Selenium Manager works with the changes? I just created #12047 which could help with that.

nvborisenko:
CI for .net building already uses selenium manager (cannot put a link to PR cauze I am on mobile).

nvborisenko:
Ah, and yes, I have tested it locally.

dotnet/src/webdriver/SeleniumManager.cs
nvborisenko: Done, thanks for the feedback.

dotnet/src/webdriver/SeleniumManager.cs
nvborisenko: Good catch, fixed.

nvborisenko:
Link to Pr where selenium manager is used for internal testing by default: https://github.com/SeleniumHQ/selenium/pull/11871

diemol (approved):

fake119:
Hello. When would you deploy new binary with this fix?

diemol:
End of May we plan to do a new release.

NCLnclNCL:
Fix ?? Bro

NCLnclNCL:
> End of May we plan to do a new release.

When update bro

titusfortner:
soon. we'll announce it on our blog when it is ready — https://www.selenium.dev/blog/

PR #12051, opened 2023-05-16
[dotnet] Avoid try/catch while working with known types in Selenium Manager

nvborisenko:
### Description
This approach is more convenient little bit.

### Motivation and Context
When debugging unit tests, debugger always stops in `try/catch` block.

### Types of changes

### Checklist

### DIFF
--- a/common/mirror/selenium
+++ b/common/mirror/selenium
@@ -102,9 +102,6 @@
       },
       {
         "browser_download_url": "https://github.com/SeleniumHQ/selenium/releases/download/selenium-4.8.0/selenium-server-4.8.3.zip"
-      },
-      {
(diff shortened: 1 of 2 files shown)


titusfortner:
The reason for the try/catch is that Selenium Manager is in beta and not fully featured, so we don't want users to get errors related to Selenium Manager, yet, because they can't do anything to fix it. We want the error message to include what they can do something about: "put drivers on path"

I think the next step is a custom Selenium Manager Error and link to how to debug problems (logging, submitting issues, etc)

nvborisenko:
I just reworked small piece of code, logically it still the same, but debugger is happy now. Ah, and performance of this piece became better if it's relevant.

titusfortner:
Where are potential errors getting caught, then?

diemol (approved):
Thank you, this is a better solution than the one I coded.

PR #12061, opened 2023-05-18
[dotnet] Correct product version in http user agent

nvborisenko:
### Description
It puts correct version in http request header like `4.9.1` rather than `4.0.0`. Assembly version is still `4.0.0` which is correct. It is achieved via `AssemblyInformationalVersion` attribute set in bazel.

### Motivation and Context
Highlight correct product version to track usage of selenium.

### Types of changes

### Checklist


dotnet/src/webdriver/Internal/ResourceUtilities.cs
nvborisenko, line 1: Should we remove unused `AssemblyVersion` method?

diemol: Yes, if it is not used, we should remove it.

nvborisenko: Removed.

diemol (approved):
Thank you, @nvborisenko!

PR #12063, opened 2023-05-18
[dotnet] Fix building errors of the solution in VS via isolating output path per project

nvborisenko:
### Description
This PR fixes errors like

image

### Motivation and Context
`WebDriver.NET` solution consists of many projects, and each of them has one shared folder to built into it (`/build/dotnet/cli`). When we build solution, there is a chance to get inconsistent build process like:
- files are always overridden
- especially when building projects concurrently

[Microsoft paid an attention on this situation](https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/7.0/solution-level-output-no-longer-valid)

### Types of changes

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,6 +80,7 @@ selenium.egg-info/
 third_party/java/jetty/jetty-repacked.jar
 *.user
 *.cache
+bin/
 obj/
 *.opensdf
 *.suo
--- a/.gitignore
+++ b/.gitignore
@@ -80,7 +80,7 @@ selenium.egg-info/
 third_party/java/jetty/jetty-repacked.jar
 *.user
 *.cache
-bin/
+dotnet/**/bin/
 obj/
 *.opensdf
 *.suo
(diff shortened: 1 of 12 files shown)


.gitignore
```diff
@@ -80,6 +80,7 @@ selenium.egg-info/
 third_party/java/jetty/jetty-repacked.jar
 *.user
 *.cache
+bin/
```
nvborisenko: This is dangerous, this `bin` folder might be used by some other language specific project. I may propose to move this entry into `.gitignore` especially for `dotnet` (creating another gitignore file? or be more specific here)

titusfortner: Since this is a different issue from fixing build errors, I think you should remove the change from this PR and make a new PR with what you think is correct along with an explanation why. Then we can maintain documentation for each of these things.

nvborisenko: Fixed, now it looks safer. Using `dotnet/**/bin/` in gitignore

diemol (approved):
Thank you, @nvborisenko!

PR #12070, opened 2023-05-19
[java] JdkHttpClient - Downgrade the method for a 303 redirect

### LINKED ISSUE #12069: [🐛 Bug]: Timeout when running Saucelabs tests on Selenium 4.9.0+ using JdkHttpClient
### What happened?

When running multiple Saucelabs tests in parallel we see a lot of timeouts when trying to upgrade from Selenium 4.8.3 to 4.9.0. 

The errors are timeouts creating drivers or `org.openqa.selenium.WebDriverException: Invalid message: ERROR Internal Server Error` or `Caused by: org.openqa.selenium.SessionNotCreatedException: Could not start a new session. Handshake response does not match any supported protocol` (which might be that same but obfuscated because of #11927)

After bisecting the issue, we see that this is caused by #11816. It seems a bit like Saucelabs uses redirects when waiting for a session to be available and Selenium no longer correctly follows those

```shell
https://github.com/vaadin/testbench/pull/1645 is our test PR which shows the problem when run in our CI. Not sure how to reproduce it elsewhere as it does not reproduce locally. Potentially by running more tests in parallel than available Saucelabs slots, so the tests have to wait for a driver for more than 30-60s.
```

### Relevant log output

```shell
org.openqa.selenium.SessionNotCreatedException: Could not start a new session. Handshake response does not match any supported protocol 

or

org.openqa.selenium.WebDriverException: Invalid message: ERROR Internal Server Error
```

### Operating System

Linux, version 4.14.231-173.361.amzn2.x86_64

### Selenium version

4.9.1

Unrelated

Unrelated

_No response_

joerg1985:
### Description
The method downgrade in case of a 303 redirect was not performed. 

### Motivation and Context
Will fix #12069

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java
+++ b/java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java
@@ -55,6 +55,7 @@ import org.openqa.selenium.remote.http.ClientConfig;
 import org.openqa.selenium.remote.http.CloseMessage;
 import org.openqa.selenium.remote.http.HttpClient;
 import org.openqa.selenium.remote.http.HttpClientName;
+import org.openqa.selenium.remote.http.HttpMethod;
 import org.openqa.selenium.remote.http.HttpRequest;
 import org.openqa.selenium.remote.http.HttpResponse;
 import org.openqa.selenium.remote.http.Message;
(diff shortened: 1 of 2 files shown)


diemol (approved):
Thank you, @joerg1985!

PR #12072, opened 2023-05-19
[java] Implements binding volumes for dynamic nodes.

### LINKED ISSUE #1642: [🚀 Feature]: Allows  to configure volume mappings for dynamically generated containers of Dynamic Grid
### Feature and motivation

I want volume mapping to upload files and deliver camera video streams

### Usage example

```yml
[docker]

configs = [
    "selenium/standalone-firefox:4.3.0-20220726", "{\"browserName\": \"firefox\",\"volumes\":[\"/data:/data\"]}",
]
```

```python
options.add_argument('--allow-file-access-from-files')
options.add_argument('--use-fake-ui-for-media-stream')
options.add_argument('--use-fake-device-for-media-stream')
options.add_argument('--use-file-for-fake-video-capture=/data/test.y4m')
options.add_argument('--use-file-for-fake-audio-capture=/data/test.wav')
```

damencho:
### Description
Adds option to mount volumes for dynamically loaded nodes. Fixes SeleniumHQ/docker-selenium#1642. 

### Motivation and Context

While testing a webrtc app we need to provide audio and video files to be used as mic and camera source for chrome via `--use-file-for-fake-audio-capture` and `--use-file-for-fake-video-capture`.

### Types of changes
Adding array of volumes to bind. Changes based on the existing devices functionality. 

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/grid/node/docker/DockerFlags.java
+++ b/java/src/org/openqa/selenium/grid/node/docker/DockerFlags.java
@@ -85,6 +85,21 @@ public class DockerFlags implements HasRoles {
   private List<String> devices;
 
   @Parameter(
+    names = {"--docker-volumes"},
+    description =
+      "Exposes volumes to a container. Each volume mapping declaration must have the"
+      + " path of the volume in both host and container separated by a colon like in this"
+      + " example: /path/in/host:/path/in/container",
+    arity = 1,
+    variableArity = true,
+    splitter = NonSplittingSplitter.class)
+  @ConfigValue(
+    section = DockerOptions.DOCKER_SECTION,
+    name = "volumes",
+    example = "[\"/mnt/disk:/mnt/custom-files\"]")
+  private List<String> volumes;
+
+  @Parameter(
       names = {"--docker-video-image"},
       description = "Docker image to be used when video recording is enabled")
   @ConfigValue(
--- a/java/src/org/openqa/selenium/grid/node/docker/DockerOptions.java
+++ b/java/src/org/openqa/selenium/grid/node/docker/DockerOptions.java
@@ -29,6 +29,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -144,6 +145,8 @@ public class DockerOptions {
 
     List<Device> devicesMapping = getDevicesMapping();
 
+    Map<String, String> volumesMapping = getVolumesMapping();
+
     // If Selenium Server is running inside a Docker container, we can inspect that container
     // to get the information from it.
     // Since Docker 1.12, the env var HOSTNAME has the container id (unless the user overwrites it)
@@ -178,6 +181,7 @@ public class DockerOptions {
                     image,
                     caps,
                     devicesMapping,
+                    volumesMapping,
                     videoImage,
                     assetsPath,
                     networkName,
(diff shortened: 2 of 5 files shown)


diemol:
Thank you for this PR.

However, how would this work in Kubernetes?

damencho:
> Thank you for this PR.
> 
> However, how would this work in Kubernetes?

I'm not so familiar with Kubernetes, but just having the files on the vms running the nodes will be enough.

diemol:
The "problem" is that this probably would not work in Kubernetes and we will end up having two implementations. One for Docker and one for Kubernetes. As I mentioned in the related issue, a solution that allows you to upload files for testing to the Node or run an executable before everything is launched is better because it would work in all environments.

damencho:
But what I see, Kubernetes already can define shared drives with pods, so this is Kubernetes native way of doing it. 
And this is just docker implementation. As the dynamic docker implementation is internal for selenium this binding of volumes need to be added just for docker implementation impl. 
Isn't this correct?

damencho:
There is a native Kubernetes way of pushing files to pods. If we would do it that way, again because of the dynamic docker impl, it will be a docker only impl.
Downloading the files on boot up is not very useful, as video files are not small. Every session booting to download over http more than 400 MB of data everytime doesn't sound good.

damencho:
Actually I don't see the pushing files at runtime a possible solution for docker or Kubernetes. 

So the only way of doing this at runtime is this with the volumes or we can add a config option which is URL and in the https://github.com/SeleniumHQ/docker-selenium/blob/trunk/NodeDocker/start-selenium-grid-docker.sh we will query that URL, download a script, make the script executable and execute it just before starting the selenium binary ... which is not so clean as the volumes, but will work, with some delay for downloading big files locally in the node before every start ...

aaronkvanmeerten:
> The "problem" is that this probably would not work in Kubernetes and we will end up having two implementations. One for Docker and one for Kubernetes. As I mentioned in the related issue, a solution that allows you to upload files for testing to the Node or run an executable before everything is launched is better because it would work in all environments.

I believe this could still work in a k8s cluster if you run on nodes which start from a custom image or mount/copy specific files locally at boot time.  Same idea applies with a nomad cluster.  This feature seems to give flexibility to use the docker volume feature for those that wish to customize their cluster for this specific type of testing.

diemol:
> But what I see, Kubernetes already can define shared drives with pods, so this is Kubernetes native way of doing it. And this is just docker implementation. As the dynamic docker implementation is internal for selenium this binding of volumes need to be added just for docker implementation impl. Isn't this correct?

@damencho yes, that is correct. However, this is not for dynamic Grids (which we probably would do only with [KEDA](https://www.selenium.dev/blog/2022/scaling-grid-with-keda/)). Therefore, people would need to do X for Docker and Y for Kubernetes, and we as a project need to support it.

>There is a native Kubernetes way of pushing files to pods. If we would do it that way, again because of the dynamic docker impl, it will be a docker only impl.
Downloading the files on boot up is not very useful, as video files are not small. Every session booting to download over http more than 400 MB of data everytime doesn't sound good.

@damencho We added the dynamic Grid because it was a feature already existing in other projects. When I worked with others to implement [Zalenium](https://github.com/zalando/zalenium), we supported the approach to mount volumes in both environments. This led to tons of maintenance, basically because there were a lot of customization users wanted to do to the way the volume was mounted. Plus you now have more use cases like OpenShift or podman. I understand the simplicity of mounting a volume to access a 400MB file, but if the file is always the same one, why don't you create a docker image that extends the Docker Selenium ones and add the file? The idea is to prepare the infrastructure to enable tests to run as fast as possible. 

>Actually I don't see the pushing files at runtime a possible solution for docker or Kubernetes. 
>So the only way of doing this at runtime is this with the volumes or we can add a config option which is URL and in the https://github.com/SeleniumHQ/docker-selenium/blob/trunk/NodeDocker/start-selenium-grid-docker.sh we will query that URL, download a script, make the script executable and execute it just before starting the selenium binary ... which is not so clean as the volumes, but will work, with some delay for downloading big files locally in the node before every start ... 

I believe this is a better solution because it would work for all environments. For example, users who have Windows VMs can also benefit from this. I believe a "clean" solution depends on the use case. But having a solution that works in all environments is much better, I think. This is how many cloud providers (Grid as a service) work.

>I believe this could still work in a k8s cluster if you run on nodes which start from a custom image or mount/copy specific files locally at boot time. Same idea applies with a nomad cluster. This feature seems to give flexibility to use the docker volume feature for those that wish to customize their cluster for this specific type of testing.

@aaronkvanmeerten true, but this involves work on the user side to customize their environment. Which needs to be documented and is also prone to errors (hence more GitHub issues).

damencho:
> I understand the simplicity of mounting a volume to access a 400MB file, but if the file is always the same one, why don't you create a docker image that extends the Docker Selenium ones and add the file? The idea is to prepare the infrastructure to enable tests to run as fast as possible.

That makes sense, thanks.

diemol:
If you are still open to work in a PR, I'd be happy to help by pointing the places where changes can be done so we get implemented the approach we discussed above.

damencho:
I don't think we want to pursue the downloading the file every time a browser is started, seems an expensive approach. We currently run one session per browser(to avoid interference between sessions and the webrtc bandwidth estimation and cpu usage when sending an HD video), and these are 400-450 sessions per test run and running those test tens of times a day imagine the time spent per day in downloading a 500MB file ... and the bytes ...

diemol:
Seems that building your own docker image with the file(s) is a much better approach.

PR #12082, opened 2023-05-22
[rb][BiDi] Add Script module commands and types 

TamsilAmani:
### Description

Add script module commands and types

### Motivation and Context

Implement script module commands and types as described in https://w3c.github.io/webdriver-bidi/#module-script.
The majority of classes and interfaces are mapping the common data types as described in https://w3c.github.io/webdriver-bidi/#data-types.

### Types of changes

### Checklist


TamsilAmani:
I have to add one more test for `this` parameter. I will add it shortly. Also, the work for `add/remove preload script` will be pushed in a separate PR.

p0deje (commented):
That's quite a lot of code, any chance we could avoid all the type mapping and just rely on native Ruby types? I understand that not every type exists, but I am not sure when those would be needed as-is? For instance, is there any reason why someone would prefer a returned `EvaluateResultException` object instead of raised Ruby exception? The latter allows leveraging regular Ruby `being ... rescue .. ned` to handle exceptions, rather than check for the returned object class.

titusfortner:
Yeah, this is a lot of strict typing for Ruby. There is no compile time error checking, so there's less need for all the "enums" and type wrappers. Unless it needs to be processed, it's probably better to just return things as Hashes.

TamsilAmani:
Ok. So I'll try to remove custom-defined classes/objects where possible. I am inclined to go with what is mentioned in w3c spec hence creating those classes. I will push the commit with relevant changes soon. Thank you @titusfortner and @p0deje

TamsilAmani:
@p0deje @titusfortner 
Changes made as per the suggestions. Now I am giving the response to the user as it is. No wrapping into object for now.

titusfortner (approved):

titusfortner (changes requested):
Ruby shouldn't be enforcing types this way. We need a better approach for it

diemol:
Are we ever merging this?

PR #12100, opened 2023-05-26
[dotnet] have support proxies for Selenium Manager

titusfortner:
### Description
sends proxy argument to selenium manager if found in options

### Motivation and Context
implements #11294 for .NET


nvborisenko:
I am testing internal web app using selenium, which is accessible internally without proxy. My internet access (to download browser binaries) is behind of corporate proxy. Conflict?

titusfortner:
It shouldn't. This use case is like for when the CI tool is not given direct access to the internet. To get internet access, you have to explicitly direct requests to a designated proxy server, and that proxy server has rules that will allow the specific requests to be routed to the internet.

Issue #12016, opened 2023-05-09
[🚀 Feature]: Default All Driver Log Output to dev/null


titusfortner:
### Feature and motivation

This has always been the default behavior for Ruby.

I think a lot of people are annoyed with how much extra stuff gets sent to the console (especially Firefox — https://github.com/mozilla/geckodriver/issues/619).

Once we document how to turn on logging for the drivers (I'm working on it), I think we should swallow the driver output by default and (maybe) direct people to the docs for how to turn it on/save it.

### Usage example

n/a

titusfortner:
.NET - defaults to no output
Ruby - defaults to no output
Python - defaults to no output, except for Firefox, which defaults to `geckodriver.log`
Java - defaults to stderr
JS  - defaults to no output

diemol:
During the meeting on [25.05.2023](https://www.selenium.dev/meetings/2023/tlc-05-25/) it was decided to move browser driver output to `dev/null`.

diemol:
Reopening because Python is still missing.

diemol:
Implemented in Python via #12103

PR #12103, opened 2023-05-27
[py] improve driver logging

titusfortner:
### Description
* Deprecate the current `log_file` or `log_path` Service class constructor parameters, which only accepts strings as filenames
* Create new `log_output` (which matches Java's `withLogOutput()`) constructor parameter. This parameter that accepts:
    * String for name of file to create (so just like the current methods)
    * TextIOWrapper as an already open file (addresses #10703)
    * int (for explicitly specifying `subprocess.STDOUT` or `subprocess.DEVNULL`)
* Default for Firefox is now null output instead of `geckodriver.log` (#12016)
* If log file selected, Chromium and IE now pipe the whole process output to a file rather than using driver argument (this matches what Java and Ruby do, and captures all driver related output which is nice)
    * EXCEPT - if `--append-log` or `--readable-timestamp` are passed (chromium) because those only work with `log-file` driver argument
* Safari - quiet parameter does not do anything in Apple's driver; it was never removed from prior implementation 

### Motivation and Context
Make things consistent between drivers and with Java & Ruby

One thing I'm not sure of....
This currently deprecates and warns about the change in default Firefox behavior; I think I'd prefer just changing it and maybe logging a message about the change with a link to the docs (https://www.selenium.dev/documentation/webdriver/drivers/service/#setting-log-output)

Also, I'm not sure I have the typing exactly right? I'm not sure what `typing.IO[typing.Any]` actually includes.


diemol (approved):
Thank you, @titusfortner!

vringar:
Thanks for implementing this!
As this got merged in the 4.11 merge window, will it get released on July 21? And are there any beta/preview/nightly artifacts published somewhere to use this immediately?

titusfortner:
@vringar take a look at #12207

py/selenium/webdriver/common/service.py
```diff
@@ -51,14 +52,27 @@ def __init__(
...
         env: typing.Optional[typing.Mapping[typing.Any, typing.Any]] = None,
         start_error_message: typing.Optional[str] = None,
         **kwargs,
     ) -> None:
+        if isinstance(log_output, str):
+            self.log_output = open(log_output, "a+", encoding="utf-8")
+        elif log_output is subprocess.STDOUT:
+            self.log_output = None
+        elif log_output is None or log_output is subprocess.DEVNULL:
+            self.log_output = open(os.devnull, "wb")
```
asottile-sentry, line 66: these file descriptors are never closed leading to ResourceWarnings when this is garbage collected (and leaked fds)

PR #12109, opened 2023-05-29
[grid] Set savefile.default_directory for Chrome

### LINKED ISSUE #12092: [🚀 Feature]: Managed Downloads: Set "savefile.default_directory" in Chrome
### Use Case

In our site, clicking a button triggers the print preview modal with a custom file to print. Because it's a separate file (generated/fetched on-the-fly), we can't just use regular mechanisms to print the webpage.

Using Chrome's flag for `--kiosk-printing` will automatically confirm the print modal (and it's defaulting to "Save as PDF" since there are no other printers in the container). However, this is downloaded to the directory specified by the `savefile.default_directory` preference (by default - more information below).

### Request

Since we are using managed downloads, Selenium sets `download.default_directory` for us ([I think this is the file](https://github.com/SeleniumHQ/selenium/blob/6222ac0d2573657bf85141b4bfad8e0ad9160541/java/src/org/openqa/selenium/grid/node/local/LocalNode.java#L535)). Could we also set `savefile.default_directory` to the same value please?

### Additional Information
- `--kiosk-printing` will not prompt to choose a directory when saving a PDF
- I checked the value via `cat /tmp/.org.chromium.Chromium.7wzSS0/Default/Preferences | jq '.savefile'` in the Selenium Node (Chrome) container
  - I got the directory via chrome://profile-internals
- There is no `savefile` key (`jq` shows `null`) when Chrome is first launched
- Attempting to print (and thus save to PDF) will set it by default to `/home/seluser/Downloads` (which doesn't exist)
- If the `savefile.default_directory` value doesn't exist, then no file will be saved
  - This is kind of a bug I guess?
- If the user right-clicks and chooses "Save As" on a webpage or image, this will prompt for a directory, and update the value of `savefile.default_directory` if the page is saved (but not if the user cancels the dialogue box)
  - This is only when using VNC, we can't interact with the OS from Selenium

ShadowLNC:
This allows PDF printing to the managed downloads directory when using Chrome with the --kiosk-printing flag.

Fixes #12092

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/grid/node/local/LocalNode.java
+++ b/java/src/org/openqa/selenium/grid/node/local/LocalNode.java
@@ -533,6 +533,8 @@ public class LocalNode extends Node {
               "download.prompt_for_download",
               false,
               "download.default_directory",
+              tempDir.getAbsolutePath(),
+              "savefile.default_directory",
               tempDir.getAbsolutePath());
       String optionsKey = Browser.CHROME.is(caps) ? "goog:chromeOptions" : "ms:edgeOptions";
       return appendPrefs(caps, optionsKey, map);
--- a/java/test/org/openqa/selenium/grid/gridui/BUILD.bazel
+++ b/java/test/org/openqa/selenium/grid/gridui/BUILD.bazel
@@ -8,9 +8,6 @@ java_selenium_test_suite(
     browsers = [
         "chrome",
     ],
-    tags = [
-        "selenium-remote",
-    ],
     deps = [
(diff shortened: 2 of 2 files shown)


ShadowLNC:
I've done some preliminary testing by building `selenium_server_deploy.jar` and copying that over the `selenium-server.jar` into my Docker container for Selenium Node (Chrome) - looks like it's working so far.

diemol (approved):
Thank you, @ShadowLNC!

diemol:
Failing tests are not related to this change.

PR #12122, opened 2023-05-31
[py] Removing browser filter when checking shadow root.

diemol:
Only Python had it, fixes #11893

During the meeting on [25.05.2023](https://www.selenium.dev/meetings/2023/tlc-05-25/) it was decided add support for everything in spec and let the drivers error when things do not work.

### Types of changes

### Checklist

### DIFF
--- a/py/selenium/webdriver/remote/webelement.py
+++ b/py/selenium/webdriver/remote/webelement.py
@@ -236,7 +236,6 @@ class WebElement(BaseWebElement):
     def shadow_root(self) -> ShadowRoot:
         """Returns a shadow root of the element if there is one or an error.
         Only works from Chromium 96, Firefox 96, and Safari 16.4 onwards.
-        Previous versions of browsers will throw an assertion exception.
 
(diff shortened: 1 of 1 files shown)


py/selenium/webdriver/remote/webelement.py
```diff
@@ -235,20 +235,13 @@ def send_keys(self, *value) -> None:
     @property
     def shadow_root(self) -> ShadowRoot:
         """Returns a shadow root of the element if there is one or an error.
-        Only works from Chromium 96 and Firefox 96 onwards. Previous versions
-        of browsers will throw an assertion exception.
+        Only works from Chromium 96, Firefox 96, and Safari 16.4 onwards.
+        Previous versions of browsers will throw an assertion exception.
```
titusfortner: This line can be removed as it is no longer true.

diemol: "Previous versions of browsers will throw an assertion exception"?

PR #12125, opened 2023-06-01
[rb][bidi] Added print command with tests

TamsilAmani:
### Description

Print command for BiDi Browsing Context module along with tests.

### Motivation and Context

https://w3c.github.io/webdriver-bidi/#command-browsingContext-print

### Types of changes

### Checklist

### DIFF
--- a/rb/lib/selenium/webdriver/bidi/browsing_context.rb
+++ b/rb/lib/selenium/webdriver/bidi/browsing_context.rb
@@ -91,10 +91,9 @@ module Selenium
           print_result = @bidi.send_cmd('browsingContext.print',
                                         context: @id,
                                         background: background,
-                                        margin: {bottom: margin.bottom, top: margin.top, left: margin.left,
-                                                 right: margin.right},
+                                        margin: margin.to_h,
                                         orientation: orientation,
-                                        page: {height: page.height, width: page.width},
+                                        page: page.to_h,
                                         pageRanges: page_ranges,
                                         scale: scale,
                                         shrinkToFit: shrink_to_fit)
--- a/rb/lib/selenium/webdriver/bidi/browsing_context.rb
+++ b/rb/lib/selenium/webdriver/bidi/browsing_context.rb
@@ -73,33 +73,35 @@ module Selenium
           )
         end
 
-        def print_page(**options)
-          print_result = @bidi.send_cmd('browsingContext.print',
-                                        context: @id,
-                                        background: false,
-                                        margin: {
-                                          bottom: options['bottom'] || 1.0,
-                                          left: options['left'] || 1.0,
-                                          right: options['right'] || 1.0,
-                                          top: options['top'] || 1.0
-                                        },
-                                        orientation: 'portrait',
-                                        page: {
-                                          height: options.dig(:page, :height) || 27.94,
-                                          width: options.dig(:page, :width) || 21.59
-                                        },
-                                        pageRanges: options['page_ranges'] || [],
-                                        scale: options['scale'] || 1.0,
-                                        shrinkToFit: options['shrink_to_fit'] || true)
-
-          print_result['data']
+        # Organising margin and page parameters in a struct for print command
+        Margin = Struct.new(:bottom, :top, :left, :right, keyword_init: true) do
+          def initialize(bottom: 1.0, top: 1.0, left: 1.0, right: 1.0)
+            super
+          end
         end
 
-        def save_print_page(path, **options)
-          File.open(path, 'wb') do |file|
-            content = Base64.decode64 print_page(**options)
-            file << content
+        Page = Struct.new(:height, :width, keyword_init: true) do
+          def initialize(height: 27.94, width: 21.59)
+            super
           end
+        end
+
+        def print_page(background: false, margin: Margin.new, orientation: 'portrait', page: Page.new, page_ranges: [],
+                       scale: 1.0, shrink_to_fit: true, path: nil)
+          print_result = @bidi.send_cmd('browsingContext.print',
+                                        context: @id,
+                                        background: background,
+                                        margin: {bottom: margin.bottom, top: margin.top, left: margin.left,
+                                                 right: margin.right},
+                                        orientation: orientation,
+                                        page: {height: page.height, width: page.width},
+                                        pageRanges: page_ranges,
+                                        scale: scale,
+                                        shrinkToFit: shrink_to_fit)
+
+          return File.write(path, Base64.decode64(print_result['data']), mode: 'wb') unless path.nil?
+
+          print_result['data']
         end
 
         def close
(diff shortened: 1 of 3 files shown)


rb/spec/integration/selenium/webdriver/bidi/browsing_context_spec.rb
```diff
@@ -23,6 +23,8 @@ module Selenium
   module WebDriver
     class BiDi
       describe BrowsingContext, only: {browser: %i[chrome edge firefox]} do
+        let(:magic_number) { 'JVBER' }
```
p0deje, line 26: What is this magic number?

rb/lib/selenium/webdriver/bidi/browsing_context.rb
```diff
@@ -73,6 +73,35 @@ def get_tree(max_depth: nil)
           )
         end
 
+        def print_page(**options)
```
p0deje: I suggest we explicitly list out all available options in the method signature and then just refer to them

```ruby
def print_page(
  bottom: 1.0,
  top: 1.0,
  left: 1.0,
  right: 1.0,
  height: 27.94, 
  width: 21.59,
  page_ranages: [],
  scale: 1.0,
  shrink_to_fit: true
)
  print_result = @bidi.send_cmd(
    'browsingContext.print',
    context: @id,
    background: false,
    margin: {bottom: bottom, top: top, left: left, right: right},
    orientation: 'portrait',
    page: {height: height, width: width},
    pageRanges: page_ranges,
    scale: scale,
    shrinkToFit: shrink_to_fit
  )

  print_result['data']
end
```

rb/lib/selenium/webdriver/bidi/browsing_context.rb
```diff
@@ -73,6 +73,35 @@ def get_tree(max_depth: nil)
...
+                                        background: false,
+                                        margin: {
+                                          bottom: options['bottom'] || 1.0,
+                                          left: options['left'] || 1.0,
+                                          right: options['right'] || 1.0,
+                                          top: options['top'] || 1.0
+                                        },
+                                        orientation: 'portrait',
+                                        page: {
+                                          height: options.dig(:page, :height) || 27.94,
```
p0deje: Why so precise default values? Is it something from the spec?

rb/lib/selenium/webdriver/bidi/browsing_context.rb
```diff
@@ -73,6 +73,35 @@ def get_tree(max_depth: nil)
...
+                                        },
+                                        pageRanges: options['page_ranges'] || [],
+                                        scale: options['scale'] || 1.0,
+                                        shrinkToFit: options['shrink_to_fit'] || true)
+
+          print_result['data']
+        end
+
+        def save_print_page(path, **options)
+          File.open(path, 'wb') do |file|
```
p0deje: You can simplify this:

```ruby
File.write(path, Base64.decode64(print_page(**options)), mode: 'wb')
```

rb/spec/integration/selenium/webdriver/bidi/browsing_context_spec.rb
titusfortner: We use it in our current print tests. I thought you knew what it was. I must have copied it from Java. /shrug

rb/lib/selenium/webdriver/bidi/browsing_context.rb
titusfortner: Should we create a basic PORO/Struct for options so we don't have so many optional parameters in this method? We can also use it for WebDriver classic print method.

p0deje: That would be nice actually. Structs would be enough:

```ruby
Margin = Struct.new(:bottom, :top, :left, :right, keyword_init: true) do 
  def initialize(bottom: 1.0, top: 1.0, left: 1.0, right: 1.0)
    super
  end
end

def print_page(margin: Margin.new, ...)
```

rb/spec/integration/selenium/webdriver/bidi/browsing_context_spec.rb
TamsilAmani: _Magic number_ is the static base64 encoded bytes which appear at the start of files. For PDF it's value is 'JVBER'. That's why we use it to test here.
More info: 
https://bitofhex.com/2019/02/19/base64-encoded-signatures/

rb/lib/selenium/webdriver/bidi/browsing_context.rb
TamsilAmani: Yes, the default values are mentioned in the spec.
https://w3c.github.io/webdriver-bidi/#command-browsingContext-print

TamsilAmani:
@titusfortner should I combine `save_print_page` with `print_page`? Since the former only needs one extra `path` parameter.

rb/lib/selenium/webdriver/bidi/browsing_context.rb
TamsilAmani: Done

rb/lib/selenium/webdriver/bidi/browsing_context.rb
TamsilAmani: Done

rb/lib/selenium/webdriver/bidi/browsing_context.rb
```diff
@@ -73,6 +73,37 @@ def get_tree(max_depth: nil)
...
+          end
+        end
+
+        def print_page(background: false, margin: Margin.new, orientation: 'portrait', page: Page.new, page_ranges: [],
+                       scale: 1.0, shrink_to_fit: true, path: nil)
+          print_result = @bidi.send_cmd('browsingContext.print',
+                                        context: @id,
+                                        background: background,
+                                        margin: {bottom: margin.bottom, top: margin.top, left: margin.left,
+                                                 right: margin.right},
```
p0deje:
```suggestion
                                        margin: margin.to_h,
```

rb/lib/selenium/webdriver/bidi/browsing_context.rb
```diff
@@ -73,6 +73,37 @@ def get_tree(max_depth: nil)
...
+                                        orientation: orientation,
+                                        page: {height: page.height, width: page.width},
```
p0deje:
```suggestion
                                        page: page.to_h,
```

rb/spec/integration/selenium/webdriver/bidi/browsing_context_spec.rb
p0deje: TIL

p0deje (approved):
Left a couple of suggestion, but looks great overall

rb/lib/selenium/webdriver/bidi/browsing_context.rb
TamsilAmani: Done

rb/lib/selenium/webdriver/bidi/browsing_context.rb
TamsilAmani: Done

PR #12131, opened 2023-06-02
[rb] have RubyMine to use artifacts built by bazel if exist

titusfortner:
### Description
This has RubyMine look in `bazel-bin` directories first. This will use artifacts in that directory even if the same files exist in `selenium/rb`.

### Motivation and Context
Can't easily use Selenium Manager with RubyMine

@p0deje 
1. ~Is this a bad use of `$LOAD_PATH`?~
2. ~This should work in a gem? I think? Do we need to test it?~
3. ~Should we check in the `selenium/rb/selenium.iml` file into git which should cause RubyMine to default to the ruby binary in bazel-bin?~
4. ~Should we go one step further and have `test_environment.rb` run the bazel command in shell?~

Ok, take 2... 
This PR will find the selenium manager binaries that were generated by bazel when running tests in RubyMine...

The question remains...
I think adding `bazel-bin/rb/lib` to `$LOAD_PATH` is the same as adding `-I ../bazel-bin/rb/lib` to config arguments? And I think there's no down side to having it there, even when bazel is running the tests and not RubyMine?

### DIFF
--- a/rb/lib/selenium/webdriver/common/selenium_manager.rb
+++ b/rb/lib/selenium/webdriver/common/selenium_manager.rb
@@ -27,9 +27,13 @@ module Selenium
     # @api private
     #
     class SeleniumManager
-      BIN_PATH = '../../../../../bin'
-
       class << self
+        attr_writer :bin_path
+
+        def bin_path
+          @bin_path ||= '../../../../../bin'
+        end
+
         # @param [Options] options browser options.
         # @return [String] the path to the correct driver.
         def driver_path(options)
@@ -72,7 +76,7 @@ module Selenium
         # @return [String] the path to the correct selenium manager
         def binary
           @binary ||= begin
-            path = File.expand_path(BIN_PATH, __FILE__)
+            path = File.expand_path(bin_path, __FILE__)
             path << if Platform.windows?
                       '/windows/selenium-manager.exe'
                     elsif Platform.mac?
--- a/README.md
+++ b/README.md
@@ -285,9 +285,8 @@ If you want to use RubyMine for development, a bit of extra configuration is nec
 
 1. Run `bazel build @bundle//:bundle //rb:selenium-devtools //rb:selenium-webdriver` before configuring IDE.
 2. Open `rb/` as a main project directory.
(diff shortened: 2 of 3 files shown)


rb/lib/selenium/webdriver/common/selenium_manager.rb
```diff
@@ -72,8 +70,8 @@ def generate_command(binary, options)
         # @return [String] the path to the correct selenium manager
         def binary
           @binary ||= begin
-            path = File.expand_path(BIN_PATH, __FILE__)
-            path << if Platform.windows?
+            path = $LOAD_PATH.find { |path| path.match(/\/bin$/) }
```
p0deje: $LOAD_PATH is one for all Ruby gems - I don't think it's going to work like that.

rb/spec/integration/selenium/webdriver/spec_helper.rb
```diff
@@ -17,14 +17,6 @@
 # specific language governing permissions and limitations
 # under the License.
 
-begin
```
p0deje: Just pushed https://github.com/SeleniumHQ/selenium/commit/7e198932c15a8b0f8ef672a49acf82025603d22c so there should be no more conflicts.

p0deje:
> Is this a bad use of $LOAD_PATH?

Yeah, I think so - it's meant to be used for require/load in Ruby, not for internal manipulations.

> This should work in a gem? I think? Do we need to test it?

I am not sure it will still work. You can test by building the gem, then installing in your system and trying to use it.

> Should we check in the selenium/rb/selenium.iml file into git which should cause RubyMine to default to the ruby binary in bazel-bin?

I think it would be useful but since I don't use RubyMine - I never bothered to check it.

> Should we go one step further and have test_environment.rb run the bazel command in shell?

I'd rather find a way to have this executed in selenium.iml.

rb/lib/selenium/webdriver/common/selenium_manager.rb
titusfortner: I'll stop being clever and do the explicit thing.

p0deje (approved):

PR #12132, opened 2023-06-02
[ci] remove drivers from GitHub Actions Runners

### LINKED ISSUE #12047: [🚀 Feature]: CI should use Selenium Manager for all tests
### Feature and motivation

I think our CI should delete the drivers that exist on the runners and always use Selenium Manager. I know that this means if we break Selenium Manager that none of the tests will pass, but it's essentially dog-fooding our own product, and we need something that ensures it works. 

I think @diemol has an alternate proposal.

### Usage example

n/a

titusfortner:
### Description
If tests would use the drivers already on the system, this PR will have them use Selenium Manager instead.
I just copied the code that Nikolay  added to .NET tests

### Motivation and Context
Fixes #12047

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -55,7 +55,6 @@ build --test_env=GITHUB_ACTIONS
 build --test_env=MOZ_HEADLESS
 build --test_env=PATH  # Remove once browser pinning works
 build --test_env=SELENIUM_BROWSER
-build --test_env=SELENIUM_MANAGER_BASE_PATH
 build --test_env=TRAVIS
 build --test_env=PYTHON_VERSION
 
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -55,6 +55,16 @@ jobs:
       - name: Free space
         if: inputs.os != 'windows'
         run: ./scripts/github-actions/free-disk-space.sh
+      - name: Remove driver directories Windows
+        if: inputs.os == 'windows'
+        run: |
+          rm "$env:ChromeWebDriver" -r -v
+          rm "$env:EdgeWebDriver" -r -v
+          rm "$env:GeckoWebDriver" -r -v
+      - name: Remove driver directories Non-Windows
+        if: inputs.os != 'windows'
+        run: |
+          sudo rm -rf $CHROMEWEBDRIVER $EDGEWEBDRIVER $GECKOWEBDRIVER
       - name: Set Ruby version
         if: inputs.ruby-version != ''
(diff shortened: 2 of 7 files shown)


titusfortner:
And... this is windows only, so I need to fix.

titusfortner:
Ok, this seems to work now.

diemol:
Looks fine but I am triggering the workflow manually to see how it goes.

diemol:
JS is the one that needs some help. Some weeks ago I created this environment variable https://github.com/SeleniumHQ/selenium/blob/trunk/.bazelrc#L58. If you are OK with that approach, we need to set the path where the Selenium Manager binaries are and then JS should work.

titusfortner:
I don't like the environment variable approach. Let's see if we can just add the copy commands to the testing target. 🤞

titusfortner:
Ok, I thought I did something weird in this PR that broke .NET tests, but no, that's what we're seeing on trunk as well with the change to the Windows 2019 Github Actions Runner (https://github.com/SeleniumHQ/selenium/actions/runs/5166344256/jobs/9306454093)

So I think this is good to merge.

titusfortner:
Nope. Looks like these Grid node tests assume there's a driver on PATH?
https://github.com/SeleniumHQ/selenium/actions/runs/5166561591/jobs/9307066780#step:14:676

titusfortner:
Ok, manually kicked off a complete run — https://github.com/SeleniumHQ/selenium/actions/runs/5167196139
* Those .NET tests have been failing since we updated from 2019 to 2022
* Those Java Browser tests are frequent intermittent failures; unrelated to this PR

The problem is the "Medium" Java tests.

For some reason the Node cannot access the Selenium Manager:
```
WARNING: Unable to obtain chromedriver using Selenium Manager: Unable to parse: thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 30, kind: ReadOnlyFilesystem, message: "Read-only file system" }', src/files.rs:58:34
```

This is the same error that we were getting in Rust tests when @p0deje removed the `no-sandbox` tag. Is this an environmental/permissions issue? Do we need to run these tests differently?

p0deje:
@titusfortner Try adding `no-sandbox` tag to the test targets that are failing.

titusfortner:
Yes, that fixed that error. I debugged those NodeOptions tests, and it turns out that they can only pass is there is a driver on PATH (they work on Mac because safaridriver). There has to be a way to mock this out? @diemol?

diemol:
It is a bit tricky because they should actually check the PATH, but I believe the changes I did should be fine.

PR #12136, opened 2023-06-03
[java] do not send driver logs to console by default

### LINKED ISSUE #12016: [🚀 Feature]: Default All Driver Log Output to dev/null

titusfortner:
### Description
* Sends logs to DEVNULL instead of STDERR unless otherwise specified
* Adds a log line to indicate where the logs went (do we need to log something? maybe a better message?)

### Motivation and Context
Fixes #12016 for Java

### DIFF
--- a/java/src/org/openqa/selenium/remote/service/DriverService.java
+++ b/java/src/org/openqa/selenium/remote/service/DriverService.java
@@ -465,7 +465,9 @@ public class DriverService implements Closeable {
         }
 
         if (logLocation == null) {
-          return System.err;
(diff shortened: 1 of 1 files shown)


diemol (approved):
Thank you, @titusfortner!

PR #12141, opened 2023-06-04
[rb] Initial extensibility points for Appium

p0deje:
This PR is the first draft that implements extensibility points that would be useful in Appium per appium/ruby_lib_core#429. With this PR, we can try to see if that would be enough to avoid monkey-patching and hopefully make Selenium break Appium less often. If this is enough, we can repeat the same for other bindings too.

@KazuCocoa Please let me know what do you think about these changes.

UPD: The Appium side is complete (https://github.com/appium/ruby_lib_core/pull/480) and greatly simplified with this PR in mind. I propose we merge this and note down what needs to be done in other bindings. Python is likely going to be similar, but I am not sure about other languages.

### DIFF
--- a/rb/lib/selenium/webdriver/remote/bridge.rb
+++ b/rb/lib/selenium/webdriver/remote/bridge.rb
@@ -33,7 +33,7 @@ module Selenium
 
         class << self
           attr_reader :extra_commands
-          attr_writer :locator_converter
+          attr_writer :element_class, :locator_converter
 
           def add_command(name, verb, url, &block)
             @extra_commands ||= {}
@@ -43,6 +43,10 @@ module Selenium
 
           def locator_converter
             @locator_converter ||= LocatorConverter.new
+          end
+
+          def element_class
+            @element_class ||= Element
           end
         end
 
@@ -432,7 +436,7 @@ module Selenium
                    "e.initEvent('submit', true, true);\n" \
                    "if (form.dispatchEvent(e)) { HTMLFormElement.prototype.submit.call(form) }\n"
 
-          execute_script(script, Element::ELEMENT_KEY => element)
+          execute_script(script, Bridge.element_class::ELEMENT_KEY => element)
         rescue Error::JavascriptError
           raise Error::UnsupportedOperationError, 'To submit an element, it must be nested inside a form element'
         end
@@ -519,7 +523,7 @@ module Selenium
         #
 
         def active_element
-          Element.new self, element_id_from(execute(:get_active_element))
+          Bridge.element_class.new self, element_id_from(execute(:get_active_element))
         end
 
         alias switch_to_active_element active_element
@@ -539,7 +543,7 @@ module Selenium
                  execute :find_element, {}, {using: how, value: what.to_s}
                end
 
-          Element.new self, element_id_from(id)
+          Bridge.element_class.new self, element_id_from(id)
         end
 
         def find_elements_by(how, what, parent_ref = [])
@@ -557,7 +561,7 @@ module Selenium
                   execute :find_elements, {}, {using: how, value: what.to_s}
                 end
 
-          ids.map { |id| Element.new self, element_id_from(id) }
+          ids.map { |id| Bridge.element_class.new self, element_id_from(id) }
         end
 
         def shadow_root(element)
@@ -631,7 +635,7 @@ module Selenium
         end
 
         def commands(command)
-          command_list[command]|| Bridge.extra_commands[command]
+          command_list[command] || Bridge.extra_commands[command]
         end
 
         def unwrap_script_result(arg)
@@ -640,7 +644,7 @@ module Selenium
             arg.map { |e| unwrap_script_result(e) }
           when Hash
             element_id = element_id_from(arg)
-            return Element.new(self, element_id) if element_id
+            return Bridge.element_class.new(self, element_id) if element_id
 
             shadow_root_id = shadow_root_id_from(arg)
             return ShadowRoot.new self, shadow_root_id if shadow_root_id
@@ -652,7 +656,7 @@ module Selenium
         end
 
         def element_id_from(id)
-          id['ELEMENT'] || id[Element::ELEMENT_KEY]
+          id['ELEMENT'] || id[Bridge.element_class::ELEMENT_KEY]
         end
 
         def shadow_root_id_from(id)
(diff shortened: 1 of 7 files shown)


KazuCocoa:
Thank you, the code seems good enough for me (at a quick glance). Let me write a code in the ruby_lib_core with this branch

rb/lib/selenium/webdriver/remote/http/common.rb
```diff
@@ -26,10 +26,18 @@ class Common
...
-            'Content-Type' => "#{CONTENT_TYPE}; charset=UTF-8",
-            'User-Agent' => "selenium/#{WebDriver::VERSION} (ruby #{Platform.os})"
+            'Content-Type' => "#{CONTENT_TYPE}; charset=UTF-8"
           }.freeze
 
+          class << self
+            attr_accessor :extra_headers
+            attr_writer :user_agent
+
+            def user_agent
```
titusfortner, line 36: Hmm, do we need all this? Watir is just subclassing the Default client and making a tweak here. https://github.com/watir/watir/blob/main/lib/watir/http_client.rb

What is Appium doing now?

rb/lib/selenium/webdriver/remote/bridge.rb
```diff
@@ -22,13 +22,30 @@ module WebDriver
...
         PORT = 4444
 
         attr_accessor :http, :file_detector
         attr_reader :capabilities
 
+        class << self
+          attr_accessor :extra_commands
+          attr_writer :locator_converter
+
+          def add_command(name, verb, url, &block)
```
titusfortner, line 38: Would we want Selenium to use this as part of `Features` modules, so we aren't subclassing `#commands`?

titusfortner: Which is to say, we probably should be doing the same thing with custom browser features as Appium does, either way.

p0deje: Yeah, I would refactor Selenium after Appium approves the changes.

rb/lib/selenium/webdriver/remote/http/common.rb
p0deje: @KazuCocoa Would implementing a custom HTTP client in Appium be better than this?

KazuCocoa: Oh, the watir way seems good. Appium overrides DEFAULT_HEADERS, b ut the watir way is sufficient for the usage. Then, this pr can drop the user_agent related change actually

titusfortner: To be clear, I'm not saying the Watir approach is "correct," if it makes more sense to do something else, Watir will do that way. 😄  Presumably we are supposed to avoid using `Http::Default` in case someone wants to use `Http::Curb`? I don't really understand why we have that one. 😂

KazuCocoa: Yea. Actually, the Watir way is better than the current Appium's one at least :P

Hmm, I like this pr approach, to provide a way to extend headers explicitly, than overriding the parent class and calling super

diemol:
I think we should have a place where we can track this, a project or a single issue? We need to do this for all bindings.

rb/lib/selenium/webdriver/remote/http/common.rb
titusfortner: Hmm, yes, I agree. With the approach in this PR we can just inject what the user sends regardless. Would be better to minimize subclassing. 

@twalpole does capybara do anything to the user agent?

titusfortner:
At the very least we can start explicitly listing the pain points in each of the bindings.

rb/lib/selenium/webdriver/remote/http/common.rb
twalpole: No it doesn't

p0deje:
> At the very least we can start explicitly listing the pain points in each of the bindings.

I think we'd need Appium maintainers to help list them out as we did in https://github.com/appium/ruby_lib_core/issues/429. Of course, we can always read the whole Appium bindings codebase, but collaborating will probably be faster. Let's see how this PR goes and consider it a proof-of-concept - we can document other bindings afterwards.

KazuCocoa:
Sorry so late, but at least https://github.com/appium/ruby_lib_core/pull/480 works with this. UA and additional finders/converter and add_commands

p0deje:
@KazuCocoa No worries, do you think there is anything else on the Appium side you could simplify if Selenium provides extension points? If yes, I'll try to hack on it in this PR. If no, I'll bring this PR to the Selenium TLC for a discussion and merge.

KazuCocoa:
I think this PR change, modifying headers, adding a new command, and adding custom locators, is sufficient. Especially adding customer locators (and avoiding converters) is the most helpful for other languages as well.


One small thing is Appium clients define its own element class like:
https://github.com/appium/ruby_lib_core/blob/master/lib/appium_lib_core/element.rb
https://github.com/appium/python-client/blob/92583ce39003b740bb5e57bbf2114d283d884d22/appium/webdriver/webelement.py#L26
to add a few methods, and currently find_element/s return the element such as 
https://github.com/appium/ruby_lib_core/pull/480/files#diff-83f0e0b8947c612afa05fae0398d0b342b434c8e42dde86d15610098ea991883R256

This could be only for language-dependent things but it might be helpful to give a custom element class for find_element/s like:
```
        def find_element_by(how, what, parent_ref = [])
          super(how, what, parent_ref, ::Appium::Core::Element)
        end
```
and it returns the given element (or default selenium's one if not provided)

p0deje:
> This could be only for language-dependent things but it might be helpful to give a custom element class for find_element/s 

Shall we provide a setter for a default class in the driver?

```ruby
Selenium::WebDriver.element_class = Appium::Core::Element # defaults to Selenium::WebDriver::Element
```

KazuCocoa:
Sounds good 👍

p0deje:
@KazuCocoa I've rebased the branch and pushed d6e2b03e00ac9e736d2aa4a2c962b85f0c1efdfc which allows you to override the returned element class, you can use it to point to Appium::Core::Element. Let me know if that's enough and I'll move on with preparation for merging this PR.

KazuCocoa:
Thank you, it made the appium client less override and simper. looks good

p0deje:
@titusfortner Would you mind taking another look, if it's good we can discuss this on a TLC meeting and talk about extending to other bindings.

diemol:
I am moving this to the next milestone as it needs a review from @titusfortner

p0deje:
Per discussion on DevSummit 2024, we are going to proceed with this, merge the Ruby part, and create an issue to track what needs to be done for other bindings.

PR #12150, opened 2023-06-05
[py] Fix BrowserError issues

bastimeyer:
### Description

This fixes two issues with `selenium.webdriver.common.bidi.cdp.BrowserError`:

1. A `KeyError` can get raised when creating a new `BrowserError` instance when the passed `obj` dict/mapping doesn't have the `code` or `message` keys.
2. When the browser has sent an invalid session ID, a string incorrectly gets passed to the `BrowserError` constructor, which would make it raise a `TypeError`.

### Motivation and Context

1. I noticed that issue while experimenting with Firefox's CDP v85 implementation. Apparently FF doesn't properly support `devtools.fetch.enable` and it returns a response with the `error` key (error message irrelevant):
   https://github.com/SeleniumHQ/selenium/blob/38e658a3d3c58f294f5525836703af451ead0b0d/py/selenium/webdriver/common/bidi/cdp.py#L270-L273
   That `error` dict however is missing the `code` key, hence the `KeyError` being raised while trying to initialize the exception.
   Since both `code` and `message` are now "optional", do you want me to add default values instead, e.g. `"unknown"`?
2. Simple error that could've been caught earlier if there were any typing information (which I didn't add). The error code I've added is copied from the error a few lines above. I don't know if that's correct though.
   https://github.com/SeleniumHQ/selenium/blob/38e658a3d3c58f294f5525836703af451ead0b0d/py/selenium/webdriver/common/bidi/cdp.py#L438

No updated tests, because I couldn't find any tests which affect these parts of the code.

### Types of changes

### Checklist


diemol (approved):
Thank you, @bastimeyer!

PR #12153, opened 2023-06-05
[rb] add rake tasks to wrap bazel commands

titusfortner:
### Description
A few wrappers

### Motivation and Context
Just convenience methods so I'm not constantly looking at the README. We can get more complicated if we want, but figured I'd see what you think of the idea in general. Or should we just be learning the bazel? 😁


rb/Rakefile
```diff
@@ -0,0 +1,18 @@
...
+require 'rspec/core/rake_task'
+
+desc 'Setup everything to run tests in RubyMine'
+task :setup do
+  system 'bazel build @bundle//:bundle //rb:selenium-devtools //rb:selenium-webdriver'
+end
+
+desc 'Run unit tests'
+task :unit do
+  system 'bazel test //rb/spec/unit/...'
```
p0deje: I recently started using `bazel test --test_size_filters small //rb/...` instead - works pretty fine.

rb/Rakefile
```diff
@@ -0,0 +1,18 @@
...
+end
+
+desc 'Run all integration tests in chrome'
+task :spec do
+  system "bazel test //rb/spec/integration/..."
```
p0deje: Also, `bazel test --test_size_filters large //rb/...`.

titusfortner:
@p0deje does this look better? I renamed `:setup` to `:update` since it needs to be run whenever assets change (new devtools release, etc)

README.md
```diff
@@ -277,7 +277,7 @@ bazel-selenium/external/bundle/bin/rdbg -A
 
 If you want to use RubyMine for development, a bit of extra configuration is necessary to let the IDE know about Bazel toolchain and artifacts:
 
-1. Run `bazel build @bundle//:bundle //rb:selenium-devtools //rb:selenium-webdriver` before configuring IDE.
+1. Run `bundle exec rake update` as necessary to update generated artifacts.
```
p0deje, line 280: I still think you don't need to include rake in the dev dependencies, but feel free to merge if you don't want to both bother updating.

p0deje: Also, this should be run from `rb/` directory, not the root.

p0deje (approved):

README.md
titusfortner: Hmm, this could go below step 2 where we open `rb/` as main project directory. And yeah, I guess I kind of miss the old `./go` wrapper. 😂

PR #12154, opened 2023-06-05
Selenium Manager & Error things

titusfortner:
### Description
* Logger logic - `@first_use` should be when anything actually prints to screen
* Error documentation - We log an additional message if one exists for it
* Selenium Manager - Better error management; User has the best info we can give them and is directed to where to go for more info
* Tests had to change because we now check "file exists" which we weren't before

### Motivation and Context
The error stuff all matches Java logic

### DIFF
--- a/rb/Gemfile.lock
+++ b/rb/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    selenium-devtools (0.113.0)
+    selenium-devtools (0.114.0)
       selenium-webdriver (~> 4.2)
     selenium-webdriver (4.9.1)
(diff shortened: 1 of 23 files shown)


rb/Gemfile.lock
```diff
@@ -25,7 +25,6 @@ GEM
     irb (1.6.4)
       reline (>= 0.3.0)
     json (2.6.3)
-    json (2.6.3-java)
```
p0deje: Please undo this - I'm pretty sure it's necessary either by JRuby or TruffleRuby.

p0deje (approved):

PR #12169, opened 2023-06-07
[py] removed redundant attributes `capabilities` and `set_capability` in wpewebkit/options.py

sandeepsuryaprasad:

### Description
Removed redundant methods `capabilities` and `set_capability` in `wpewebkit/options.py`
Since `Options` class in `wpewebkit/options.py` inherits from `ArgOptions`, I have removed redundant methods which sets and gets capabilities. 

Replaced `set_capabilities` method with  a setter method to add capabilities to `self._caps`.

### Motivation and Context
Since both methods  `capabilities` and `set_capability` are already implemented in parent class 
`selenium.webdriver.common.options.ArgOptions`, I found that same attributes are being implemented once again in  `wpewebkit/options.py`. 

To get the capabilities from internal dictionary `self._caps` we do have a getter method in
`selenium.webdriver.common.options.Options` class 

`@property`
`def capabilities(self):`
	`return _caps`

And the capabilities are being set through `set_capabilities` method. For the purpose of consistency I have 
repalced  `set_capabilities()` method in `selenium.webdriver.common.options.Options` class to a setter method.

`@property.setter`
`def  capabilities(self, values):`
	  `name, value = values`
	  `self._caps[name] = value`

so now the capabilities can be set through setter,

`self.capabilities` `=` `("pageLoadStrategy", "normal")`

### Types of changes

### Checklist


py/selenium/webdriver/common/options.py
```diff
@@ -29,15 +29,16 @@ def __init__(self) -> None:
...
         self._proxy = None
-        self.set_capability("pageLoadStrategy", "normal")
+        self.capabilities = ("pageLoadStrategy", "normal")
         self.mobile_options = None
 
     @property
     def capabilities(self):
         return self._caps
 
-    def set_capability(self, name, value) -> None:
```
diemol: Why are you renaming this? I thought the scope was to remove this from the `wpwebkit` ones.

diemol: This is a breaking change, btw.

diemol (changes requested):

py/selenium/webdriver/common/options.py
sandeepsuryaprasad: Yes, the initial scope was to remove the duplicate methods in `wpwebkit`. After that I thought  converting `set_capability` to a setter method `capabilities` would be right choice. Since both the methods are used internally in `Options` class, I did not had to do changes in other places. If it's too risky change, you can ignore this change and probably merge the commit where only the duplicate attributes were removed.

diemol: Well, it is not in the scope and breaks the API for current users. It also makes it confusing because `capabilities` is a very ambiguous name across the project and code base.

It'd be easier if you revert the changes and I am happy to merge what was considered in the initial scope.

sandeepsuryaprasad: Sure.. No problem. I will revert the changes and re-commit.

sandeepsuryaprasad: @diemol can you please check now..

diemol (approved):
Thank you, @sandeepsuryaprasad!

PR #12174, opened 2023-06-07
Fix docstring for removed arg (browser_profile)

mdmintz:
``browser_profile`` was removed from ``start_session()`` args in ``remote/webdriver.py``, but it's still in the docstring.
It should be removed from the docstring to match the remaining args.

### Types of changes

- [X] Fix docstring

### Checklist

- [X] My change is to the documentation / docstrings.


titusfortner:
Thanks!

PR #12183, opened 2023-06-08
[py] all assignments to instance attributes in `Timeouts` class are now being set through setter method.

sandeepsuryaprasad:

### Description
In `Timeouts` class in `common/timeouts.py`, values of 3 instance attributes are being saved in `self._implicit_wait`, `self._page_load` and `self._script` through `__init__` method. The thing is `self._convert` function is being called from `__init__` function and also from every setter methods as well. 

Instead of directly assigning values to internal variables `self._implicit_wait`, `self._page_load` and `self._script` in constructor which can by-pass setter method (while creating instance), we can remove leading `_` so that every attribute assignment goes through corresponding setter methods, which makes the code more efficient. Now the function `self._convert` will be called only from setter function and it is not necessary to call it from constructor. 

### Motivation and Context
In `Timeouts` class in `common/timeouts.py`, values of 3 instance attributes are being saved in `self._implicit_wait`, `self._page_load` and `self._script` through `__init__` method. The thing is `self._convert` function is being called from `__init__` function and also from every setter methods as well. 

Instead of directly assigning values to internal variables `self._implicit_wait`, `self._page_load` and `self._script` in constructor which can by-pass setter method (while creating instance), we can remove leading `_` so that every attribute assignment goes through corresponding setter methods, which makes the code more efficient. Now the function `self._convert` will be called only from setter function and it is not necessary to call it from constructor. 

### Types of changes

### Checklist


diemol (commented):
Can you please clean up your branch? It is not easy when all those commits are present.

In addition, what's the motivation of this PR? Please note that those change might break current user's code. If you have features or improvements in mind like these ones, it is better to open a feature request first to allow a conversation about them.

PR #12186, opened 2023-06-09
[py] instance attributes in Timeout class are now being set through setter…

sandeepsuryaprasad:

### Description
[py] all assignments to instance attributes in `Timeouts` class are now being set through setter method.

### Motivation and Context
1. In class `Timeouts` under `common/timeouts.py`, three instance attributes are being set in the `__init__` method.
`self._implicit_wait`, `self._page_load` and `self._script`.

2. The problem with this approach is that we are directly assigning converted values from `self._convert` method to internal variables `self._implicit_wait`, `self._page_load` and `self._script` by by-passing setter methods ( `@implicit_wait.setter`, `@page_load.setter`, `@script.setter`) while creating of the object.

3. The setter methods comes into play only when we are directly assigning values to instance attributes, e.g. `self.implicit_wait = 10` or `self.page_load = 30`. Now this assignment goes through setter methods. 

4. Ultimately the setter methods are setting values to internal variables `_implicit_wait`, `_page_load` and `_script`.

5. So the change that I am suggesting is that, instead of directly assigning internal attributes in `__init__` method, let the assignment go through setter method. 

`self.implicit_wait = implicit_wait`  
`self.page_load = page_load`
`self.script = script`

the above statements would call corresponding setter methods, which in-turn sets internal attributes `_implicit_wait`, `_page_load` and `_script`. In this case we are doing an assignment to a property descriptor (data descriptor) which goes through `@implicit_wait.setter`, `@page_load.setter`, `@script.setter` methods.

6. I believe this change will not break any of the existing code base, since we are not renaming any attributes.

### Types of changes

### Checklist

### DIFF
--- a/py/selenium/webdriver/common/timeouts.py
+++ b/py/selenium/webdriver/common/timeouts.py
@@ -49,9 +49,9 @@ class Timeouts:
          - script - Either an int or a float. The number passed in needs to how many
             seconds the driver will wait.
         """
-        self._implicit_wait = self._convert(implicit_wait)
-        self._page_load = self._convert(page_load)
-        self._script = self._convert(script)
+        self.implicit_wait = implicit_wait
+        self.page_load = page_load
+        self.script = script
 
     @property
     def implicit_wait(self) -> float:


diemol (approved):
Thank you, @sandeepsuryaprasad!

PR #12210, opened 2023-06-16
[py] optimised `Options` object in `ie/options.py` by replacing 34 different methods with one single descriptor

sandeepsuryaprasad:

### Description
Optimised `Options` object in `ie/options.py` by replacing 34 different `getters` and `setters` by one single custom descriptor `IeOptions`.
### Motivation and Context
- In `Options` class in `ie/options.py` we are setting around 17 different browser attributes to an internal dictionary `_options` using `@properties`. Each attribute has to have one `getter` and one `setter` method.  for all 17 attributes we have 34 methods, which does the common action of updating internal dictionary `_options`

- I have replace all 34 methods with one single descriptor data `IeOptions`.

### Types of changes

### Checklist

### DIFF
--- a/java/BUILD.bazel
+++ b/java/BUILD.bazel
@@ -17,9 +17,7 @@ java_plugin(
     processor_class = "com.google.auto.service.processor.AutoServiceProcessor",
     visibility = ["//visibility:private"],
     deps = [
-        artifact("com.google.auto:auto-common"),
         artifact("com.google.auto.service:auto-service"),
-        artifact("com.google.guava:guava"),
     ],
 )
 
@@ -32,7 +30,6 @@ java_library(
         "//java:__subpackages__",
     ],
     exports = [
-        artifact("com.google.auto.service:auto-service"),
         artifact("com.google.auto.service:auto-service-annotations"),
     ],
 )


titusfortner:
How is this backwards compatible if we are removing the property setters?

Does this still work? 
```
options.initial_browser_url = "https://www.selenium.dev";
```

sandeepsuryaprasad:
> How is this backwards compatible if we are removing the property setters?
> 
> Does this still work?
> 
> ```
> options.initial_browser_url = "https://www.selenium.dev";
> ```

Yes, it works.

sandeepsuryaprasad:
> > How is this backwards compatible if we are removing the property setters?
> > Does this still work?
> > ```
> > options.initial_browser_url = "https://www.selenium.dev";
> > ```
> 
> Yes, it works.

we can get the information about the descriptor object in IDE while creating descriptor object in `PrintOptions` class. You need to include the docstring in corresponding descriptor class.

Screenshot 2023-06-17 at 9 18 24 AM

Screenshot 2023-06-17 at 9 14 09 AM

Screenshot 2023-06-17 at 9 17 10 AM

isaulv:
I think this is good. How do you feel about using `_set_name` hook so that's less repetitive to write `attribute = DescriptiorObject("name")`?

sandeepsuryaprasad:
> I think this is good. How do you feel about using `_set_name` hook so that's less repetitive to write `attribute = DescriptiorObject("name")`?

Yes, we can do that, but normally in python if the name of the `property` is let's say `browser_version`, the actual data is hidden in internal variable `_browser_version` (this is the standards). But in `BaseOptions` class, the name of the `property` is `browser_version` but the actual data is stored in `browserVersion`. 

so when I use the hook `__set_name__`, `browser_version` will be passed to `name` argument in `__set_name__`. 

`Class BaseOptionsDescriptor:`
      `def __set_name__(self, cls, name):`  # name will have  "browser_version"
             `self.name = name`

`Class BaseOptions:`
     `browser_version = BaseOptionsDescriptor()` # "browser_version" is implicitly passed to "name" argument

`browser_version` will be passed as an argument to `name` parameter in `__set_name__` method. But in the current code, the browser version is not stored in `_browser_version`, rather it is stored in `browserVersion` class variable. 

I personally feel that they made it little complicated by having separate name for property and separate name for actual variable where the data is stored. It would have been very simple had they just used `_browser_version` to store actual version instead of `browserVersion`. If this was the case the name of the descriptor object is `browser_version` and the actual data will be in `_browser_version`. I could have easily used `__set_name__` Hook like below

`Class BaseOptionsDescriptor:`
 `def __set_name__(self, cls, name):` 
             `self.private_name = "_"+name`

`Class BaseOptions`
`browser_version = BaseOptionsDescriptor()`

so in the above code in, `self.private_name` I would have had `_browser_version` and name of descriptor or property is `browser_verison`

To avoid code from breaking,  I have explicitly informed `__init__` in `descriptor` that please go-ahed and store actual value of browser version in `browserVersion` variable. So this is the reason I am using `__init__` method instead of hook `__set_name__`.

Otherwise as you rightly said I could have used `__set_name__` to avoid  `attribute = DescriptiorObject("name")` repeatedly.

Python documentation (just in case if someone wants to refer)
https://docs.python.org/3/howto/descriptor.html

PR #12211, opened 2023-06-16
[java] Use netty as http server for unit testing

joerg1985:
### Description
Use a netty based http server for unit testing and removed the old com.sun.net.httpserver.HttpServer based JreServer.
This PR will also unify the proxy tests by using the netty based http server and dropping some not needed dependencies.

### Motivation and Context
The old JreServer was based in java/src, not in java/test this added a `requires jdk.httpserver;` into the generated java-modules.class of the client jars. Additionally there are now less dependencies needed in self-tests.

### Types of changes

### Checklist


titusfortner:
I think we're getting rid of the netty dependency in October when we drop Java 8.

Not sure the options, would we be able to do something with default jdk http client?

I guess either way this doesn't increase dependences.

diemol:
This is something I wanted to do, because we use the Jre server only for tests. 

@titusfortner this is about the Netty server, not the HTTP client.

diemol (approved):
Thank you, @joerg1985!

asolntsev:
@joerg1985 Sorry, but I don't understand the need for this PR. 
JreServer was used only in tests, right? 

>  this added a `requires jdk.httpserver;` into the generated java-modules.class of the client jars. 

Since JreServer was used only in tests, then we should just move JreServer from `src/main` to `src/test`. Then client jars would know nothing about `requires jdk.httpserver;`.

> Additionally there are now less dependencies needed in self-tests.

I don't understand. Before this PR, we used only Java for tests. Now we additionally use Netty server for tests. I see **more dependencies**, not less.

diemol:
Not following...

The JreServer was being used to run tests with Selenium RC, now that we deleted Selenium RC, the Jre Server was a left over. We usually have used the same server for testing, so this was some sort of tech debt. With this PR we are using Netty for everything, which makes things a little more simple.

Why do you say that there are more dependencies, @asolntsev?

asolntsev:
》 Why do you say that there are more dependencies, @asolntsev?

Because "Java" is one dependency,  and "Java+Netty server" is two dependencies.

P.S. I know that Netty is used in Selenium project anyway, but it's netty client, not server. So I consider "Netty server" as an additional dependency.

krmahadevan:
@asolntsev - i think we use netty server for powering the server component of Grid no ?

asolntsev:
Probably. 
But nevertheless,  "jre" is less than "jre+netty".

joerg1985:
@asolntsev from my point of view it was more like `jre+netty+jdk.httpserver` before and now it is `jre+netty`. As mentioned above, netty is already used by the grid server.

PR #12214, opened 2023-06-17
[py] added type hints to `common/action_chains.py`, `alert.py` and `support/relative_locator.py`

sandeepsuryaprasad:

### Description
Added type hints to `common/action_chains.py`, `alert.py` and `support/relative_locator.py`.

### Motivation and Context
Type hint support to all methods of `ActionChains`, `Alert` objects in `common/action_chains.py` and `common/alert.py` and `RelativeBy` object in `support/relative_locator.py`
### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/remote/RemoteWebDriver.java
+++ b/java/src/org/openqa/selenium/remote/RemoteWebDriver.java
@@ -363,24 +363,6 @@ public class RemoteWebDriver
     return elementLocation.findElements(this, context, findCommand, locator);
   }
 
-  /**
-   * @deprecated Rely on using {@link By.Remotable} instead
-   */
-  @Deprecated
-  protected WebElement findElement(String by, String using) {
-    throw new UnsupportedOperationException(
-        "`findElement` has been replaced by usages of " + By.Remotable.class);
-  }
-
-  /**
-   * @deprecated Rely on using {@link By.Remotable} instead
-   */
-  @Deprecated
-  protected List<WebElement> findElements(String by, String using) {
-    throw new UnsupportedOperationException(
-        "`findElement` has been replaced by usages of " + By.Remotable.class);
-  }
-
   protected void setFoundBy(SearchContext context, WebElement element, String by, String using) {
     if (element instanceof RemoteWebElement) {
       RemoteWebElement remoteElement = (RemoteWebElement) element;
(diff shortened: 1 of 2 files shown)


py/selenium/webdriver/common/action_chains.py
```diff
@@ -27,6 +31,11 @@
 from .actions.wheel_input import WheelInput
 from .utils import keys_to_typing
 
+any_device = Union[KeyInput, PointerInput, WheelInput]
+
+if TYPE_CHECKING:
+    from typing import Self
```
isaulv: Self is a recent feature, so you may need to update the dependency to include `typing_extension`
Also put all the other types in here.
and maybe put `from __future__ import annotations`.

py/selenium/webdriver/common/action_chains.py
```diff
@@ -61,7 +70,7 @@ class ActionChains:
     another.
     """
 
-    def __init__(self, driver, duration=250, devices=None):
+    def __init__(self, driver, duration: int = 250, devices: Optional[List[any_device]] = None):
```
isaulv: And if you use typing extensions , you may be able to do `List[any_device] | None`

py/selenium/webdriver/common/timeouts.py
```diff
@@ -37,7 +37,33 @@ class JSONTimeouts(TypedDict, total=False):
     JSONTimeouts = Dict[str, int]
 
 
+class TimeoutDescriptor:
```
isaulv: Your pull request should stick to one topic, like type hints only.

py/selenium/webdriver/common/action_chains.py
titusfortner: Recent feature as in not applicable to Python 3.8 which we still have to support?

isaulv: @titusfortner Python 3.8 doesn't have `Self` in the `typing` module but we should probably generalize the typing code so that it just uses typing_extensions by default. Probably not a bad idea really since typing extensions can be updated during a Python version's lifetime (like at this time, it could have fixes for `Self` that's not in Python 3.11.

titusfortner: Ah, so pull it in as a dependency instead of using the one in the Python standard library?

isaulv: As far as I know it's not a bad practice.

titusfortner: in general we try to minimize dependencies, but it seems like a good way to provide more flexibility.

AutomatedTester:
Please remove the previous PR commits from this PR

py/selenium/webdriver/common/action_chains.py
sandeepsuryaprasad: > Self is a recent feature, so you may need to update the dependency to include `typing_extension`
> Also put all the other types in here.
> and maybe put `from __future__ import annotations`.

Instead of `Self` I can use `TypeVar` as well. I separate out the type hints in this PR and raise a new one.

PR #12216, opened 2023-06-17
[java] Create less HttpClient instances while creating a CDP connection

joerg1985:
### Description
The `CdpEndpointFinder` created a short living instance of a `HttpClient`, now the code uses this `HttpClient` if the endpoint was found to create the connection.

### Motivation and Context
There are complains about the number of threads created combined with a slow cleanup strategy of the new `JdkHttpClient`. Creating less short living `HttpClient` instances will create less threads waiting for cleanup.

### Types of changes

### Checklist

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -126,11 +126,6 @@ build:remote --experimental_credential_helper=%workspace%/scripts/credential-hel
 # Use pinned browsers when running remotely
 build:remote --//common:pin_browsers
 
-# Extend the remote config for CI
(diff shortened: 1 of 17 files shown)


diemol (approved):
Thank you, @joerg1985!

We were creating clients like this because the [async-http-client](https://github.com/AsyncHttpClient/async-http-client) recommended to have a single instance per JVM. But as we are slowly moving users to the new Java 11 HTTP client, paying more attention to this makes a lot of sense.

Issue #12217, opened 2023-06-17
[🐛 Bug]: Expected to read a START_MAP but instead have: END. Last 0 characters


bchhabra:
### What happened?

 private FirefoxDriver prepareFirefoxWebDriver () {
        FirefoxOptions options = new FirefoxOptions();
        options.addArguments("--headless=new");
        return new FirefoxDriver(options);
    }

<selenium.version>4.10.0</selenium.version> 
Issue happens only in jenkins while launching the browser in headless mode.

not able to reproduce in macos Ventura 13.4 or Windows.

### How can we reproduce the issue?

```shell
launch Firefox as headless on RHEL 7.9

NAME="Red Hat Enterprise Linux Server"
VERSION="7.9 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VARIANT="Server"
VARIANT_ID="server"
VERSION_ID="7.9"
PRETTY_NAME="Red Hat Enterprise Linux Server 7.9 (Maipo)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.9:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.9
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.9"
```

### Relevant log output

```shell
Error Message

Unable to obtain: Capabilities {acceptInsecureCerts: true, browserName: firefox, moz:debuggerAddress: true, moz:firefoxOptions: {args: [--headless=new]}}

Stacktrace

org.openqa.selenium.remote.NoSuchDriverException: Unable to obtain: Capabilities {acceptInsecureCerts: true, browserName: firefox, moz:debuggerAddress: true, moz:firefoxOptions: {args: [--headless=new]}}
	at com.sample.test.ratePayTest(PPRTest.java:41)
Caused by: org.openqa.selenium.json.JsonException: Unable to parse: 
	at com.sample.test.ratePayTest(PPRTest.java:41)
Caused by: org.openqa.selenium.json.JsonException: 
Expected to read a START_MAP but instead have: END. Last 0 characters read: 
Build info: version: '4.10.0', revision: 'c14d967899'
[...]

PR #12219, opened 2023-06-18
[java] Improved the exit code and error handling

joerg1985:
### Description
This change will improve the handling of errors while calling the selenium manager.
The behavior is now more like the dotnet implementation, which looks better to me.

These are the changes in detail:
- first check the exit code, parse the output only if successful
- include the exit code and the raw output in the exception
- replaced the exit code check `code > 0` with ` code != 0`  (dotnet does it the same way).
- raise a exception when there was an exception calling the selenium manager (dotnet does it the same way).
- logging was removed, as we have the exception now with the full stack and more informations

### Motivation and Context
To have a better to read exception in cases like #12217 with more information to debug the issue.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/manager/SeleniumManager.java
+++ b/java/src/org/openqa/selenium/manager/SeleniumManager.java
@@ -40,6 +40,7 @@ import org.openqa.selenium.Platform;
 import org.openqa.selenium.Proxy;
 import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.json.Json;
+import org.openqa.selenium.json.JsonException;
 
 /**
  * This implementation is still in beta, and may change.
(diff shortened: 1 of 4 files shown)


java/src/org/openqa/selenium/manager/SeleniumManager.java
```diff
@@ -111,26 +111,28 @@ private static String runCommand(String... command) {
...
-              "%s running command %s: %s",
-              e.getClass().getSimpleName(), Arrays.toString(command), e.getMessage()));
+      throw new WebDriverException(
+        "Failed to run command: "
+          + Arrays.toString(command),
+          e);
     }
-    SeleniumManagerJsonOutput jsonOutput =
-        new Json().toType(output, SeleniumManagerJsonOutput.class);
-    if (code > 0) {
```
diemol, line 127: Selenium Manager can return a code different than 0 but it will still return the valid error in a JSON payload. The new code will output the whole payload which is not necessary.

diemol (changes requested):

java/src/org/openqa/selenium/manager/SeleniumManager.java
joerg1985: The new version will now try to parse the response in any way.
I handled the empty string to avoid the misleading exception we hand in other cases.

diemol: Thank you

diemol (approved):

PR #12226, opened 2023-06-19
Made network interception threads fail silently

### LINKED ISSUE #12220: [🐛 Bug]: Network interception threads bypass exception handling blocks and may terminate the application at any time
### What happened?

The problem is the same as in #10856. If you repeatedly navigate to pages, a `Selenium::WebDriver::Error::WebDriverError` exception will be raised eventually.

I assume that this is happening: 
- when the driver navigates to page A, it creates a new thread for each request made by the browser, as specified in `attach_socket_listener` method of `selenium/webdriver/common/websocket_connection.rb`.
- Then, when it navigates to page B, if there are still incomplete requests being fetched that belonged to page A, they will eventually lead to the existence of invalid InterceptionIds.

I tried to set `Thread.current.abort_on_exception` to false, so that these threads may fail silently. It seems to me that this doesn't affect the driver negatively, since the threads that are failing will be invalid and therefore they will be inacessible anyway. I belive this is the solution to this problem.

I don't know if the browser itself deletes these InterceptionIds or if they are somehow generated by `selenium-devtools` tools and for some reason need to be ignored. I'm assuming that these ids are managed solely by the browser, while devtools can only ask for the existing ids.

It seems that by using `Thread.current.abort_on_exception=true`, there's no way to safely catch errors when driver is using network interception. If you wrap your code with a `begin...rescue` block, if multiple threads fail, it will bypass the block and terminate the program. This can be simulated with the following code:

```ruby
def test
  # If true -> bypass the begin..rescue block (not intended)
  # If false -> program ends normally (intended)
  Thread.abort_on_exception = true

  begin
    (0..1).each do |thread|
      Thread.new{ (9..10).each{|i| sleep(1); 10/(10-i) } }
    end

    sleep(2)
   rescue Exception => e
     # Wait just a bit more, so that the two
     # threads may raise exception at once.
     # e.g.:
     # 'puts e.message' or
     # 'puts e.backtrace' or
     # 'puts Logger.info(e.message)' or just
     # 'sleep(0.1)'
     puts "Exception caught :)"
     retry
   end

   puts "Program finished normally"
rescue Exception => ex
  # This rescue case should never need to be
  # executed if thread exceptions were safely
  # captured by the inner begin..rescue block.
  #
  # In fact, depending on how you adjust the
  # sleep() time, even this block will be bypassed
  puts "Inner block was bypassed"
  puts e.message
end

10.times{test}
```

Given this example, it seems to me that threads should really be aborted silently, so that any application using selenium with network interception may be safe regarding exception handling.

```ruby
# Using titusfortner's example in the mentioned issue,
# this code could be used to verify the behaviour:
require 'selenium-webdriver'

driver = Selenium::WebDriver.for :chrome

ids = []
driver.intercept do |request, &continue|
    continue.call{|response| puts "response will be continued"}
end

10.times { driver.get "https://youtube.com" }
```

### Relevant log output

```shell
#<Thread:0x00007fa59ed3c4a8 /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/common/websocket_connection.rb:121 run> terminated with exception (report_on_exception is true):
/home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools.rb:49:in `send_cmd': -32602: Invalid InterceptionId.:  (Selenium::WebDriver::Error::WebDriverError)
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-devtools-0.109.0/lib/selenium/devtools/v109/fetch.rb:66:in `continue_request'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:128:in `continue_request'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:121:in `intercept_response'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:84:in `block in request_paused'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:162:in `with_cancellable_request'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:81:in `request_paused'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:48:in `block in intercept'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/common/websocket_connection.rb:131:in `block in callback_thread'
#<Thread:0x00007fa5a3360dd0 /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/common/websocket_connection.rb:121 run> terminated with exception (report_on_exception is true):
/home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools.rb:49:in `send_cmd': -32602: Invalid InterceptionId.:  (Selenium::WebDriver::Error::WebDriverError)
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-devtools-0.109.0/lib/selenium/devtools/v109/fetch.rb:66:in `continue_request'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:128:in `continue_request'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:121:in `intercept_response'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:84:in `block in request_paused'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:162:in `with_cancellable_request'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:81:in `request_paused'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:48:in `block in intercept'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/common/websocket_connection.rb:131:in `block in callback_thread'
/home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools.rb:49:in `send_cmd': -32602: Invalid InterceptionId.:  (Selenium::WebDriver::Error::WebDriverError)
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-devtools-0.109.0/lib/selenium/devtools/v109/fetch.rb:66:in `continue_request'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:128:in `continue_request'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:121:in `intercept_response'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:84:in `block in request_paused'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:162:in `with_cancellable_request'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:81:in `request_paused'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:48:in `block in intercept'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/common/websocket_connection.rb:131:in `block in callback_thread'
/home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools.rb:49:in `send_cmd': -32602: Invalid InterceptionId.:  (Selenium::WebDriver::Error::WebDriverError)
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-devtools-0.109.0/lib/selenium/devtools/v109/fetch.rb:66:in `continue_request'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:128:in `continue_request'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:121:in `intercept_response'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:84:in `block in request_paused'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:162:in `with_cancellable_request'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:81:in `request_paused'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/devtools/network_interceptor.rb:48:in `block in intercept'
        from /home/luca/.rvm/gems/ruby-3.2.0/gems/selenium-webdriver-4.10.0/lib/selenium/webdriver/common/websocket_connection.rb:131:in `block in callback_thread'
```

### Operating System

Ubuntu 22.10

### Selenium version

selenium-webdriver 4.7.1 and 4.10.0, selenium-devtools 0.109.0, ruby 3.2.0 (2022-12-25 revision a528908271) [x86_64-linux]

Chrome version 109.0.5414.74 (Official Build) (64-bit)

Not sure

No

jlucartc:
When `Thread.abort_on_exception is true`, multiple threads may cause the application using Selenium's network interception to terminate without any means of dealing with the exception, which may leave the application in an inconsistent state. Therefore, this option is being set to false.

This change will affect only `callback_thread` method in `websocket_connection.rb` since the socket thread created in `attach_socket_listener`, being only one thread, causes no harm to exception handling, and thus can safely raise exceptions. Also, the rescue clause was removed from `callback_thread` since it won't be raising exceptions anymore.

Fixes #12220

### Types of changes

### Checklist


rb/lib/selenium/webdriver/common/websocket_connection.rb
```diff
@@ -129,8 +129,6 @@ def callback_thread(params)
           Thread.current.report_on_exception = true
 
           yield params
-        rescue *CONNECTION_ERRORS
```
p0deje, line 132: Can you please revert CONNECTION_ERRORS change? We still want to rescue it even if it no longer crashes the whole thread.

p0deje: Still there, did you forget to push?

jlucartc: Sorry, I forgot to push the changes. I've just sent them now.

p0deje:
Ok, I can now [see](https://github.com/SeleniumHQ/selenium/actions/runs/5522507285/jobs/10072473001) why we needed to abort the thread on exception - because we have a case where we'd like to raise from event:

https://github.com/SeleniumHQ/selenium/blob/a04a22f4490da1b9576f251a5f16fa34e8898645/rb/spec/integration/selenium/webdriver/devtools_spec.rb#L48-L56

Any ideas how do we suppress internal CDP errors while preserving this behavior?

jlucartc:
I believe that something like this could solve this case: 

```ruby
def callback_thread(params)
  Thread.new do
    # Threads will not abort silently, so that the exception may be handled
    Thread.current.abort_on_exception = true

    # We might end up blocked forever when we have an error in event.
    # For example, if network interception event raises error,
    # the browser will keep waiting for the request to be proceeded
    # before returning back to the original thread. In this case,
    # we should at least print the error.
    Thread.current.report_on_exception = true

    yield params
 # We're gonna need a wider clause for exception handling, or else
 # the IntercepID error will not be captured and will be raised in the main thread.
  rescue Exception => e
    Thread.stop
    # No exception can be raised here, because it will lead to
    # unsafe thread exception handling. Therefore, the only way
    # to make the exception known is by informing the main thread directly.

    # If the error is a connection error, set a flag that signals the main thread.
    # Else, ignore the exception
    lock.synchronize{
        kill_connection(e.class)  if CONNECTION_ERRORS.include?(e.class) and is_connection_alive?
    }
  end
end
```

```ruby
# Since the error capture will be syncronized, I don't think that these two methods need to be inside
# a 'lock.syncronize' block also.
def kill_connection(error_class)
    @keep_connection = false
    @connection_error_class = error_class
end

def is_connection_alive?
    @keep_connection
end
```
Then, the websocket_connection may try to verify `@keep_connection` right before each incoming message received in `attach_socket_listener`.

p0deje:
Hm, maybe we simply need to do the following

```ruby
      def callback_thread(params)
        Thread.new do
          Thread.current.abort_on_exception = true

          # We might end up blocked forever when we have an error in event.
          # For example, if network interception event raises error,
          # the browser will keep waiting for the request to be proceeded
          # before returning back to the original thread. In this case,
          # we should at least print the error.
          Thread.current.report_on_exception = true

          yield params
        rescue Error::WebDriverError, *CONNECTION_ERRORS
          Thread.stop
        end
      end
```

jlucartc:
But how is the main thread gonna know that the exception was raised? Doesn't the test case need a raised exception?

p0deje:
In this case, it won't know for WebDriver errors, but that's exactly the point of this PR, isn't it?

jlucartc:
If all the tests need is a RuntimeError, then your suggestion would work. I was thinking that somehow the main thread would need to know when *CONNECTION_ERRORS were raised.

p0deje:
>I was thinking that somehow the main thread would need to know when *CONNECTION_ERRORS were raised.

No, that's why we rescue them.

jlucartc:
I got it. Then I'm gonna send the commit with this version:

```ruby
      def callback_thread(params)
        Thread.new do
          Thread.current.abort_on_exception = true

          # We might end up blocked forever when we have an error in event.
          # For example, if network interception event raises error,
          # the browser will keep waiting for the request to be proceeded
          # before returning back to the original thread. In this case,
          # we should at least print the error.
          Thread.current.report_on_exception = true

          yield params
        rescue Error::WebDriverError, *CONNECTION_ERRORS
          Thread.stop
        end
      end
```

PR #12235, opened 2023-06-21
[dotnet] Make WebDriver.csproj file more friendly with native pack command

nvborisenko:
### Description
Now I am able to create local nuget package with `dotnet pack` command.

### Motivation and Context
This is one step further to be more friendly for .net devs.

### Types of changes

### Checklist


nvborisenko:
Build is failing, but the changes are not related to it and should not break build process. Seems it was failing before, or tools were upgraded recently.

I see it is failing because nuget tool cannot find `pdb` files in place? Not sure. In any case at this moment we don't need symbols because symbols even are not pushed to the nuget.org

If we cannot fix it quickly, then I can propose to turn off creation of symbols: https://github.com/SeleniumHQ/selenium/blob/a04a22f4490da1b9576f251a5f16fa34e8898645/dotnet/src/support/BUILD.bazel#LL129C31-L129C31.

And return it back after @shs96c refactors the whole process of ci for .net.

diemol (approved):
Thank you, @nvborisenko!

PR #12249, opened 2023-06-24
[bazel][java] Remove TestNG dependency

joerg1985:
### Description
This PR will remove TestNG, there is no change for the jars generated by bazel. I had to add `log4j-core:2.20.0` otherwise 3.0.0-alpha1 would be picked.

### Motivation and Context
Selenium switched from TestNG to JUnit in 2012, it is time to remove it.

### Types of changes

### Checklist


titusfortner (approved):

java/maven_deps.bzl
```diff
@@ -97,7 +97,7 @@ def selenium_java_deps():
             "org.mockito:mockito-core:4.11.0",
             "org.slf4j:slf4j-api:2.0.7",
             "org.slf4j:slf4j-jdk14:2.0.7",
-            "org.testng:testng:7.7.1",
+            "org.apache.logging.log4j:log4j-core:2.20.0",
```
diemol, line 100: Why do we need log4j? Are we using that transitively?

joerg1985: `com.github.spotbugs:spotbugs` does use it and so it is added transitively.

titusfortner: Is this required for our linting tests? Why do we have it as a runtime dependency?

diemol: We should also get rid of spotbugs and use something across the project.

Also, those are all the dependencies, when the packages are created then only the runtime are packed.

diemol (approved):
Thank you, @joerg1985!

PR #12255, opened 2023-06-25
[py] simplify error check in check_response for w3c

KazuCocoa:

In Appium dev, I observed Python client did not handle exceptions properly. https://github.com/appium/python-client/pull/868/files#r1199231346 For example when a response had `unknown method` value in the error key, it raised a general `WebDriverException`. It looks like python client had a case that did not handle w3c response properly.

After a few investigation, probably the main reason was this python binding gives extra `value` for `check_response`. Maybe we can simplify it.

### Description

in this pr, I have parsed error response by checking the `error` key as same as Appium Python client does after the above error case. This is should be the same as what currently Ruby binding also does.

So,

1. get `value` from the response
2. get `error` from the 1
3. get the class mapping with the `error` value in 2
4. Raise the error

With this PR, python client will not refer to `status` in the response. Ruby binding also does this already, and selenium v4 no longer support old protocol, I  think this behavior should be ok for you.

I also noticed the input of `check_response` got nested `value` such as 

```
{'status': 500, 'value': '{"value":{"error":"session not created","message":"session not created: This version of ChromeDriver only supports Chrome version 97\\nCurrent browser version is 114.0.5735.133 with binary path /Applications/Google Chrome.app/Contents/MacOS/Google Chrome","stacktrace":"0   chromedriver                        0x000000010c995e69 chromedriver + 5160553\\n1   chromedriver                        0x000000010c920593 chromedriver + 4679059\\n2   chromedriver                        0x000000010c4d42c8 chromedriver + 172744\\n3   chromedriver                        0x000000010c4faa93 chromedriver + 330387\\n4   chromedriver                        0x000000010c4f6636 chromedriver + 312886\\n5   chromedriver                        0x000000010c4f2cf6 chromedriver + 298230\\n6   chromedriver                        0x000000010c52d423 chromedriver + 537635\\n7   chromedriver                        0x000000010c527123 chromedriver + 512291\\n8   chromedriver                        0x000000010c4fcdce chromedriver + 339406\\n9   chromedriver                        0x000000010c4fe105 chromedriver + 344325\\n10  chromedriver                        0x000000010c95123e chromedriver + 4878910\\n11  chromedriver                        0x000000010c968d17 chromedriver + 4975895\\n12  chromedriver                        0x000000010c96ea3f chromedriver + 4999743\\n13  chromedriver                        0x000000010c96961a chromedriver + 4978202\\n14  chromedriver                        0x000000010c945bb1 chromedriver + 4832177\\n15  chromedriver                        0x000000010c985fd8 chromedriver + 5095384\\n16  chromedriver                        0x000000010c986161 chromedriver + 5095777\\n17  chromedriver                        0x000000010c99d2a8 chromedriver + 5190312\\n18  libsystem_pthread.dylib             0x00007ff8198a24e1 _pthread_start + 125\\n19  libsystem_pthread.dylib             0x00007ff81989df6b thread_start + 15\\n"}}'}
```
 for probably backward compatibility. So, the `response` value has extra `value` for some reason. I guess for backward compatibility in the past.

In this PR, I have changed the behavior to pass only one `value` key to the `check_response` so that the `check_response` can expect only W3C error format as the argument. (`test_handle_errors_better` test code addresses this point.)

I also removed unused exception for w3c such as IME stuff.

### Motivation and Context

I'd like to cleanup error handling for W3C format to help Appium Python client's implementation to reduce duplication code for both.

### Types of changes

### Checklist


py/selenium/webdriver/remote/errorhandler.py
```diff
@@ -141,62 +121,41 @@ def check_response(self, response: Dict[str, Any]) -> None:
...
-           object.
+           object. The format is expected to be {'value': {message body}}
 
         :Raises: If the response contains an error message.
         """
-        status = response.get("status", None)
-        if not status or status == ErrorCode.SUCCESS:
+
+        payload_dict = response
+        if type(response) != dict:
```
mykola-mokhnach: prefer using isinstance

py/selenium/webdriver/remote/errorhandler.py
```diff
@@ -18,13 +18,14 @@
...
 from typing import Type
+from typing import Union
+from typing import Sequence
+from typing import List
 
 from selenium.common.exceptions import ElementClickInterceptedException
 from selenium.common.exceptions import ElementNotInteractableException
 from selenium.common.exceptions import ElementNotSelectableException
 from selenium.common.exceptions import ElementNotVisibleException
-from selenium.common.exceptions import ImeActivationFailedException
```
mykola-mokhnach, line 26: this change makes it breaking

py/selenium/webdriver/remote/remote_connection.py
```diff
@@ -316,31 +315,27 @@ def _request(self, method, url, body=None):
...
             if not any([x.startswith("image/png") for x in content_type]):
                 try:
                     data = utils.load_json(data.strip())
                 except ValueError:
-                    if 199 < statuscode < 300:
-                        status = ErrorCode.SUCCESS
-                    else:
-                        status = ErrorCode.UNKNOWN_ERROR
-                    return {"status": status, "value": data.strip()}
+                    return {"value": data.strip()}
```
mykola-mokhnach: not sure how the upstream code would understand if this is an erroneous response or a successful one after the status field is removed

py/selenium/webdriver/remote/errorhandler.py
```diff
@@ -49,87 +50,66 @@
 from selenium.common.exceptions import UnknownMethodException
 from selenium.common.exceptions import WebDriverException
 
-
-class ExceptionMapping:
```
mykola-mokhnach, line 53: By removing this logic we make the client 100% incompatible with the legacy JWP protocol. 

In scope of Appium I can guarantee that the server only supports W3C one and we do not care about JWP support anymore. I cannot guarantee this is the case in Selenium world though. Selenium java client, for example, still supports both JWP and W3C protocols, for example. 

Leaving this decision up to client maintainers.

py/selenium/webdriver/remote/remote_connection.py
```diff
@@ -316,31 +315,27 @@ def _request(self, method, url, body=None):
                 response = http.request(method, url, body=body, headers=headers)
             statuscode = response.status
         data = response.data.decode("UTF-8")
-        LOGGER.debug(f"Remote response: status={response.status} | data={data} | headers={response.headers}")
+        LOGGER.debug(f"Remote response: data={data} | headers={response.headers}")
         try:
             if 300 <= statuscode < 304:
                 return self._request("GET", response.headers.get("location", None))
```
mykola-mokhnach, line 325: this code looks suspicious to me as it might potentially fall into an endless recursion. Usually clients (including urllib3, which is this code base on) handle such stuff automatically with some default number of allowed redirect retries to avoid the above recursion issue.

py/selenium/webdriver/remote/errorhandler.py
KazuCocoa: Yes, need feedback at the point. I saw W3C flag removals in the past so I thought this client was removing old protocol behaviors but could be wrong

titusfortner: Selenium does not support JWP, but there are still various pieces of JWP-only code left that haven't been removed. If it doesn't apply to w3c we want it gone. :)

py/selenium/webdriver/remote/remote_connection.py
titusfortner: I was actually testing this out earlier this week for Sauce-related reasons... The default number of allowed redirects is 3, and that is actually what happens. You can set remote URL to http://demo.cyotek.com/features/redirectlooptest.php and see that you only get 3. I'm not sure what this loop does, I just know it didn't result in infinite retries.

titusfortner:
Somehow you broke the cookie_test and the linter is also being particular.

py/selenium/webdriver/remote/remote_connection.py
KazuCocoa: Maybe they can be removed, but removed from this pr to reduce diff

KazuCocoa:
The cookies test returned non-w3c error response (according to the error result). I have added the pattern as a test case https://github.com/SeleniumHQ/selenium/pull/12255/files#diff-16a7f3559108027005862f77d542ca48d5dc188f7325ca4df1b73984f7de86c5R246-R256 for now to handle the case as well.

py/selenium/webdriver/remote/errorhandler.py
```diff
@@ -49,154 +51,136 @@
...
-    NO_SUCH_COOKIE = [62, "no such cookie"]
-    UNABLE_TO_CAPTURE_SCREEN = [63, "unable to capture screen"]
-    ELEMENT_CLICK_INTERCEPTED = [64, "element click intercepted"]
-    UNKNOWN_METHOD = ["unknown method exception"]
-
-    METHOD_NOT_ALLOWED = [405, "unsupported operation"]
+    UNKNOWN_ERROR = 13
+
+
+def _get_stacktrace(payload_dict):
```
AutomatedTester: we can simplify this to only handle the W3C keys

py/selenium/webdriver/remote/errorhandler.py
```diff
@@ -49,154 +51,136 @@
...
+        message = response_dict.get("message")
+        if message is None:
+            # w3c
+            message = payload_dict.get('message', error)
+
+        exception_class: Type[WebDriverException] = ERROR_TO_EXC_MAPPING.get(
+            error, WebDriverException
+        )
+
+        # backtrace from remote in Ruby
```
AutomatedTester: What does this comment mean?

py/selenium/webdriver/remote/errorhandler.py
```diff
@@ -49,154 +51,136 @@
...
+    'element not selectable': ElementNotSelectableException,
+    'invalid coordinates': InvalidCoordinatesException,
+}
 
 
 class ErrorCode:
     """Error codes defined in the WebDriver wire protocol."""
 
     # Keep in sync with org.openqa.selenium.remote.ErrorCodes and errorcodes.h
     SUCCESS = 0
```
AutomatedTester: We don't need to use the Error codes at all anymore

AutomatedTester (changes requested):

py/selenium/webdriver/remote/errorhandler.py
KazuCocoa: Removed the ErrorCode class from errorhandler.py. in py/selenium/webdriver/remote/remote_connection.py, i have left 0 and 15 instead. The `status` probably safe to remove, but I'm not so confident to include the change in this PR since the `value` also has `value` (vanilla webdriver response) key. That may need to tweak properly at the same time, I think.

KazuCocoa:
Updated. It would be appreciated to run the CI, thanks

KazuCocoa:
fixed the lint issue after applying the trunk

KazuCocoa:
Hm.., tox seems good in Codespace:
```
@KazuCocoa ➜ /workspaces/selenium (py-w3c-error) $ tox -c py/tox.ini
docs: commands[0] /workspaces/selenium/py> sphinx-build -b html -d ../build/doctrees docs/source ../build/docs/api/py
Running Sphinx v1.8.2
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 0 source files that are out of date
updating environment: [] 0 added, 4 changed, 0 removed
reading sources... [100%] webdriver_remote/selenium.webdriver.remote.webelement                                                                                                                                                                                                                                                                                                                   
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] webdriver_remote/selenium.webdriver.remote.webelement                                                                                                                                                                                                                                                                                                                    
generating indices... genindex py-modindex
highlighting module code... [100%] typing                                                                                                                                                                                                                                                                                                                                                         
writing additional pages... search
copying static files... done
copying extra files... done
dumping search index in English (code: en) ... done
dumping object inventory... done
build succeeded.

The HTML pages are in ../build/docs/api/py.
docs: OK ✔ in 5.62 seconds
.pkg: _optional_hooks /workspaces/selenium/py> python /usr/local/python/3.10.8/lib/python3.10/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: get_requires_for_build_sdist /workspaces/selenium/py> python /usr/local/python/3.10.8/lib/python3.10/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: get_requires_for_build_wheel /workspaces/selenium/py> python /usr/local/python/3.10.8/lib/python3.10/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: prepare_metadata_for_build_wheel /workspaces/selenium/py> python /usr/local/python/3.10.8/lib/python3.10/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: build_sdist /workspaces/selenium/py> python /usr/local/python/3.10.8/lib/python3.10/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
flake8: install_package /workspaces/selenium/py> python -I -m pip install --force-reinstall --no-deps /workspaces/selenium/py/.tox/.tmp/package/7/selenium-4.10.0.tar.gz
flake8: OK ✔ in 3.16 seconds
isort: install_package /workspaces/selenium/py> python -I -m pip install --force-reinstall --no-deps /workspaces/selenium/py/.tox/.tmp/package/8/selenium-4.10.0.tar.gz
.pkg: _exit /workspaces/selenium/py> python /usr/local/python/3.10.8/lib/python3.10/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
  docs: OK (5.62=setup[0.19]+cmd[5.43] seconds)
  flake8: OK (3.16 seconds)
  isort: OK (1.99 seconds)
  congratulations :) (12.34 seconds)
@KazuCocoa ➜ /workspaces/selenium (py-w3c-error) $ 
```

PR #12256, opened 2023-06-26
[py] moved `ConnectionType` class outside `Mobile` class to avoid nesting of classes

sandeepsuryaprasad:
… in remote/mobile.py

### Description
I have moved `ConnectionType` class outside `Mobile` class in `remote/mobile.py` to avoid nesting of classes and for better readability.

### Motivation and Context
The class `ConnectionType` is inside `Mobile` class, which is I felt bit confusing and hard to read. So I have moved the inner class `ConnectionType` out of `Mobile`  class. 
### Types of changes

### Checklist


diemol:
It is a rather short class. Is this because the current state is not aligned with a Python style guide? I do not see what the benefit is.

sandeepsuryaprasad:
> It is a rather short class. Is this because the current state is not aligned with a Python style guide? I do not see what the benefit is.

Normally we do have nested functions or classes only in case of decorators or in case of fixtures. But this is a normal class, so I have moved it out of `Mobile` class to keep `ConnectionType` object separate.

diemol (approved):
Thank you for clarifying!

PR #12263, opened 2023-06-26
[java] Deprecate all 'lift style' code

joerg1985:
### Description
As suggested in the tlc Slack channel, this is the proposal to deprecate all 'lift style' related code.
I think this should be removed in Selenium 5 if the community is okay with it. 

### Motivation and Context
- This code has not been updated for one decade, except for refactoring.
- I could not find anything online, only the selenium wiki entry from 2015
- The wiki states: `This is very much work in progress at the moment, ...`
- The code is based on the JUnit4 dependency and the TestCase style is based on JUnit3
- There is an extra jar generated which seens [not to be in use](https://mvnrepository.com/artifact/org.seleniumhq.selenium/lift)

### Types of changes

### Checklist

### DIFF
--- a/dotnet/private/nuget.bzl
+++ b/dotnet/private/nuget.bzl
@@ -72,9 +72,10 @@ def _stage_files_for_packaging(ctx, staging_dir):
         relative_dest_file = _get_relative_destination_file(src_file)
         src_list.append((src_file, relative_dest_file))
         if (ctx.attr.create_symbol_package):
-            symbol_file = dep[DefaultInfo].files.to_list()[0]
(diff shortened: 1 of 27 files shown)


titusfortner:
So yeah, this code is from WebDriver before it merged with Selenium.
Looks like they were created and added to by one person in 2008 & 2009, and have only been moved and had various formatting changes since then.
I don't even think we need to wait for Selenium 5; should be fine with the normal 2 releases then remove. If someone complains we can spin it off. 😆

diemol (approved):
This makes sense, thank you @joerg1985!

PR #12269, opened 2023-06-27
few tweaks to driver finding logic

titusfortner:
### Description
1. Don't execute Java `ExecutableFinder()` if a System Property has been defined
2. Python log results even if error code isn't 0

### Motivation and Context
Ensured logic the same between Java/Python/Ruby/.NET

### DIFF
--- a/java/src/org/openqa/selenium/remote/service/DriverFinder.java
+++ b/java/src/org/openqa/selenium/remote/service/DriverFinder.java
@@ -14,11 +14,14 @@ public class DriverFinder {
 
   public static String getPath(DriverService service, Capabilities options) {
     Require.nonNull("Browser options", options);
-    String defaultPath = new ExecutableFinder().find(service.getDriverName());
-    String exePath = System.getProperty(service.getDriverProperty(), defaultPath);
+    String exePath = System.getProperty(service.getDriverProperty());
+
+    if (exePath == null) {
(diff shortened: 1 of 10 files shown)


diemol (commented):
Does this mean we are letting Selenium Manager find the driver on the PATH in Java?

titusfortner:
No, not yet. The default parameter for getProperty always evaluates, even when there is a property, which is less performant and unnecessary. So very minor performance improvement; and now we just remove those 3 lines when we do let SM get from PATH.

java/src/org/openqa/selenium/remote/service/DriverFinder.java
```diff
@@ -14,11 +14,14 @@ public class DriverFinder {
 
   public static String getPath(DriverService service, Capabilities options) {
     Require.nonNull("Browser options", options);
-    String defaultPath = new ExecutableFinder().find(service.getDriverName());
```
diemol, line 17: This line is the one that finds the driver on the PATH, and it is getting removed. Right?

titusfortner: no, it moved to line 20. I moved it after `getProperty()` with a conditional to only execute if there was no property found.

diemol: You are right.

diemol (approved):

PR #12294, opened 2023-06-29
[java] make external modules static

joerg1985:
### Motivation and Context
In the past without the java module system you could exclude jars from the dependencies in your pom.xml.
This helped to reduce the overhead of dependencies which where not used by your code.
e.g. if you do not use the new JDK HttpClient you could exclude the netty async client.

With the module system enabled this was not possible, the netty async client was a required module.
When compiling the code a `module not found` error is raised.

### Description
This change will use the static modifier on external² modules, making them [optional at runtime](https://nipafx.dev/java-modules-optional-dependencies/). The user can now remove them and might risk to get an exception at runtime.

² not selenium and not java modules

### Types of changes

### Checklist


joerg1985:
@titusfortner this is what we wrote about in slack.

joerg1985:
The browser driver jars should also be static, so i need to rework this.

joerg1985:
Looks like the browser jars are not referenced in the other jars so there is no need to make them static.

@shs96c I am not sure if the flags are really set in the final jars, e.g. the transitive flag is set for `java.logging` in `org.seleniumhq.selenium.api`, but the module-info.class has this flag not set as far as i can tell. I see a change in the outJar when changing the flags. So i assume the flags are lost in the bazel-jar-magic but did not find the cause for this.

diemol (approved):
Thank you, @joerg1985!

PR #12298, opened 2023-07-01
[rb] move logging and rspec matchers to top namespace

titusfortner:
### Description
* move custom logger RSpec matchers to same place as logger
* matchers must match both ID and log level/type
* matchers can work with logs that aren't `Selenium::WebDriver.logger` (requires forwarding `progname`)
* allow matching multiple log ids
* matcher log level names are past-tense

### Motivation and Context
* Would like to be able to use the matchers in other projects (Watir)
* Previously added support for matching any log level but didn't enforce it, only the ids
* Contemplating moving these files into a separate package for Selenium 5

### Considerations

Should this file be namespaced?
Is there a better alternative to passing in a custom logger with each method?

### Types of changes


titusfortner:
Also, should we move the `self.logger` method out of `selenium/webdriver.rb` and into `selenium/webdriver/common/logger.rb`? If a class like `Selenium::Server` were to use it, it wouldn't need to pull in all of Selenium to make that work.

rb/lib/selenium/logger/logger_matchers.rb
```diff
@@ -0,0 +1,78 @@
...
+LEVELS = {errored: 'ERROR',
+          warned: 'WARN',
+          waring: 'WARN',
+          deprecated: 'WARN',
+          info: 'INFO',
+          informed: 'INFO',
+          debugged: 'DEBUG'}.freeze
+
+LEVELS.each_key do |level|
+  RSpec::Matchers.define :"have_#{level}" do |expected_ids, logger = nil|
```
p0deje: This will be loaded in runtime where RSpec might be unavailable. It would be better to keep it in `rb/spec/` so it's not part of the gem itself.

titusfortner:
I put this at top level in case we want to release a separate `selenium-logger.gem`, or, even, split this code into a separate repo/project since it isn't part of "core" Selenium. It will also make it easier to have separate loggers for devtools gem or server class or selenium-manager if we make that its own gem as well.

p0deje:
> I put this at top level in case we want to release a separate `selenium-logger.gem`, or, even, split this code into a separate repo/project since it isn't part of "core" Selenium. It will also make it easier to have separate loggers for devtools gem or server class or selenium-manager if we make that its own gem as well.

You would then need to make sure that matchers are not loaded unless RSpec is available.

titusfortner:
Yeah, this needed several pieces of tidying still.
* everything namespaced `Selenium::Logger`
* rescuing LoadError if RSpec not installed
* removed hard coded reference to `Selenium::WebDriver.logger`, which means users have to explicitly pass in the logger being used

p0deje (approved):

PR #12299, opened 2023-07-02
[rb] implement downloading grid with Selenium Manager

titusfortner:
### Status
I'm moving this to draft for us to discuss whether we want to implement this in other bindings and synchronize the versions.

Also, play-by-play is here - https://www.youtube.com/watch?v=0rL6Dtoo8As

### Description
* Created method for downloading grid in Selenium Manager
* Executes by default with old method as backup

### Motivation and Context
* Remove network command executions from Selenium code and let the binary manage it

### Types of changes

### DIFF
--- a/rb/lib/selenium/server.rb
+++ b/rb/lib/selenium/server.rb
@@ -79,25 +79,14 @@ module Selenium
       #
 
       def download(required_version = :latest)
-        required_version = latest if required_version == :latest
-        download_file_name = "selenium-server-#{required_version}.jar"
-
-        return download_file_name if File.exist? download_file_name
-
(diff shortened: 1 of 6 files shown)


p0deje (approved):
`selenium/server.rb` always seemed to much for me - I like SM approach.

titusfortner:
I think ruby is the only one that implemented it in the first place, though. It's a great feature, and would be nice to have in all the bindings.

But yes, this is exactly the kind of overhead and duplication that Selenium manager is great at removing. (thanks again @bonigarcia)

diemol:
I believe we still want to implement this.

PR #12320, opened 2023-07-06
[java] less memory allocation

joerg1985:
### Description
Optimized atoms and the grid server to create less garbage objects.

### Motivation and Context
Will reduce the gc needed while polling e.g. for an element to get visble. Helps also when a grid server handles alot of sessions, usually returning only some bytes to the client.

### Types of changes

### Checklist


joerg1985:
PS: i just added one change, to not create a stack trace if the logger will dispose this log message.

diemol (approved):

PR #12328, opened 2023-07-09
[py] added type checks in `setter` methods of different browser options

sandeepsuryaprasad:

### Description
Added type checks in `setter` methods of different browser options.

### Motivation and Context
The type check was not being done before setting the values of some browser options (e.g `binary_location`) in setter methods of different browsers. I have added these validations in `setter` methods of below options.
- `chromium/options.py`
- `firefox/options.py`
- `safari/options.py`
- `wpewebkit/options.py`
- `safari/services.py`

### Types of changes

### Checklist


py/selenium/webdriver/chromium/options.py
```diff
@@ -51,6 +51,8 @@ def binary_location(self, value: str) -> None:
         :Args:
          - value: path to the Chromium binary
         """
+        if not isinstance(value, str):
+            raise TypeError("Binary Location must be a string")
```
AutomatedTester: Since we're repeating the message all over, it might be worth added the error messages to variables and use that? Since we have the base class the error messages could live there and then we know each subclass has the same error message.

```suggestion
            raise TypeError(self.BINARY_LOCATION_ERROR)
```

sandeepsuryaprasad: Yes. We can include all common error messages in base class so that we can make use of it in sub-classes. let me give a try.

sandeepsuryaprasad: > Since we're repeating the message all over, it might be worth added the error messages to variables and use that? Since we have the base class the error messages could live there and then we know each subclass has the same error message.

Added class variable `BINARY_LOCATION_ERROR` in `ArgOptions` class to hold common binary error message.

AutomatedTester (approved):

PR #12335, opened 2023-07-10
[java] fixed partial matches for UrlTemplate

joerg1985:
### Description
Only match a url when it is a full match and not a partial match. e.g. 
the template `/element` should not match for `/something-with-element-in-it`

### Motivation and Context
This might lead to requests taking the wrong route, e.g. the CustomLocatorHandlerTest did pass because of this.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/remote/http/UrlTemplate.java
+++ b/java/src/org/openqa/selenium/remote/http/UrlTemplate.java
@@ -24,37 +24,48 @@ import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-/** An incredibly bad implementation of URL Templates, but enough for our needs. */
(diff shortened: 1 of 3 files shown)


java/src/org/openqa/selenium/remote/http/UrlTemplate.java
```diff
@@ -24,37 +24,48 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-/** An incredibly bad implementation of URL Templates, but enough for our needs. */
+/** A bad implementation of URL Templates, but enough for our needs. */
```
titusfortner, line 27: This is my favorite update.

joerg1985: I couldn't resist :D

PR #12336, opened 2023-07-10
[grid] reject requests only if it has been in the queue before trying…

joerg1985:
### Description
Reject requests only if it has been in the queue before trying to create the sessions.

### Motivation and Context
This fixes a race condition, when `--reject-unsupported-caps` is set to true and the request is added to the queue while the `NewSessionRunnable` is executed.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
+++ b/java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
@@ -753,7 +753,17 @@ public class LocalDistributor extends Distributor implements AutoCloseable {
 
     @Override
     public void run() {
-      boolean loop = true;
+      Set<RequestId> inQueue;
+      boolean loop;
+      if (rejectUnsupportedCaps) {
+        inQueue = sessionQueue.getQueueContents().stream()
+          .map(SessionRequestCapability::getRequestId)
+          .collect(Collectors.toSet());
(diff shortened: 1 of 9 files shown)


java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
```diff
@@ -753,7 +753,10 @@ private class NewSessionRunnable implements Runnable {
 
     @Override
     public void run() {
-      boolean loop = true;
+      Set<RequestId> inQueue = sessionQueue.getQueueContents().stream()
+        .map(SessionRequestCapability::getRequestId)
+        .collect(Collectors.toSet());
+      boolean loop = !inQueue.isEmpty();
```
diemol: This is querying the whole queue on each execution, I would prefer to avoid that.

diemol (changes requested):

java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
joerg1985: I update the code to read the queue only when the `rejectUnsupportedCaps` flag is set, when the flag is not set there is no need to read it.

Note: There is another change in the last push to avoid creating copies of the `ImmutableCapabilities`.

diemol (approved):

Issue #1839, opened 2016-03-18
Python client keys_to_typing no longer handles non-string iterable


joshbruning:
https://github.com/SeleniumHQ/selenium/commit/d095cd2a410cbf8ffd95cac9aed364221151f722
regressed keys_to_typing in the Python client.

## Expected Behavior -
Send list [Keys.ENTER], results in enter key press.

## Actual Behavior -
Send list [Keys.ENTER], results in "[u'\ue007']" sent to a text box.

PR #12340, opened 2023-07-11
[py] fixed a bug in function `keys_to_typing` in `common/utils`

sandeepsuryaprasad:

### Description
The argument `value` in `keys_to_typing` function in `common/utils`, can never be an instance of `Keys` class. The `value` will be either a `str` or `int` or `float`.

### Motivation and Context
- One of the conditional statement never gets executed in function `keys_to_typing` due to the fact that the value of `value` argument can never be an instance of `Keys` class. 
- The values for `value` argument could be either a string `str` or `int` or `float`
- If the value is `int` or `float`, we are explicitly casting it to a `str` and appending each item to a list `_typing`
- Also, there are 2 `for` loops which was kind of redundant. I have fixed this redundancy by eliminating one of the loop.

### Types of changes

### Checklist

### DIFF
--- a/py/selenium/webdriver/common/action_chains.py
+++ b/py/selenium/webdriver/common/action_chains.py
@@ -18,7 +18,6 @@
 """The ActionChains implementation."""
 from __future__ import annotations
 
-import warnings
 from typing import TYPE_CHECKING
 from typing import List
 from typing import Optional
--- a/py/selenium/webdriver/common/service.py
+++ b/py/selenium/webdriver/common/service.py
@@ -218,5 +218,7 @@ class Service(ABC):
             raise
         except OSError as err:
             if err.errno == errno.EACCES:
-                raise WebDriverException(f"'{os.path.basename(self._path)}' executable may have wrong permissions.") from err
+                raise WebDriverException(
+                    f"'{os.path.basename(self._path)}' executable may have wrong permissions."
+                ) from err
             raise
--- a/py/selenium/webdriver/common/utils.py
+++ b/py/selenium/webdriver/common/utils.py
@@ -24,7 +24,6 @@ from typing import Optional
 from typing import Union
 
 from selenium.types import AnyKey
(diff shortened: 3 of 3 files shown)


titusfortner:
The code you removed was to fix this bug - https://github.com/SeleniumHQ/selenium/issues/1839
We need to verify we didn't cause the problem again.

sandeepsuryaprasad:
No, the change that I have made is not breaking the fix https://github.com/SeleniumHQ/selenium/issues/1839 
I just reverified by running all typing related tests once again in my local. 
when you say `element.send_keys(Keys.TAB)`, `Keys.TAB` gets evaluated to unicode `\ue004` string. Now a unicode string is being passed to `keys_typing_to`, which is not an `int` or `float`. So I am directly appending the unicode string to `_typing` list.

The problem with the current code is that with the line `if isinstance(val, Keys)` the argument `val` can never be an instance of class `Keys`. This code never gets executed and it always jumps to the  `else` block. In `else` block the unicode string is getting appended to the list `_typing`

AutomatedTester (approved):

AutomatedTester:
There are some conflict. Could you look and correct and I will get it merged

sandeepsuryaprasad:
> There are some conflict. Could you look and correct and I will get it merged

@AutomatedTester Please do not merge this PR since it is conflicting with one more commit https://github.com/SeleniumHQ/selenium/commit/a0e569b0825e19bbd1e931cdaffa48ae98efe090

isaulv:
I think the intention of checking for `Keys` is that it's suppose to check for control/modifier keys. 
Can you update your pull request? Can you check if there are tests for action chains that test things like ctrl-c ctrl-v?

diemol:
@sandeepsuryaprasad did you want to continue with this PR?

sandeepsuryaprasad:
> @sandeepsuryaprasad did you want to continue with this PR?

@diemol you can close this for now.

Issue #12301, opened 2023-07-02
[🐛 Bug]: Selenium Manager and.NET


titusfortner:
### What happened?

Unlike other languages, .NET is doing the first part of the location outside of the "driver finder." So the Driver verification class is not distinguishing between the default behavior and what a user might be passing in.

If the user is passing in something incorrect, we need to raise an exception for that, rather than try to remediate it the way we would if the default isn't working.

Also, unlike other languages .NET has you specify the directory the driver is in, not the full path to the file itself. We may want to be more accommodating of this possibility by checking if it is a full path or if it is the directory.

### How can we reproduce the issue?

```shell
var service = ChromeDriverService.CreateDefaultService();
service.DriverServicePath = "definitely/not/a/driver/path.exe";

driver = new ChromeDriver(service);
```

### Relevant log output

oh .NET and logging...

### Operating System

any

### Selenium version

trunk

### What are the browser(s) and version(s) where you see this issue?

any

### What are the browser driver(s) and version(s) where you see this issue?

any

### Are you using Selenium Grid?

no

NCLnclNCL:
use this will working   ChromeDriverService service = ChromeDriverService.CreateDefaultService(@"D:\chromedriver new", "chromedriver.exe");

titusfortner:
Yes, but right now if the driver doesn't exist in that directory, Selenium will ignore it and use the Manager to download a new one, when it should error instead.

NCLnclNCL:
> Yes, but right now if the driver doesn't exist in that directory, Selenium will ignore it and use the Manager to download a new one, when it should error instead.

Yes it is bad

titusfortner:
Fixed with referenced PRs

PR #12344, opened 2023-07-12
[dotnet] Use Selenium Manager to locate drivers on PATH

titusfortner:
### Dependencies
~Don't evaluate this PR until the following have been merged:~
~* #12342~
~* #12343~
~* Also I think this breaks Appium so I need to redo it without changing all the method parameters.~

### Description
* Driver Finder is only used in Default service constructor
* Removed DriverFinder usage in Driver constructor; it only needs to happen when a default Service class is created because that is the only place a user does not have to specify it, and we do not want to remediate.
* No longer need special Safari Technology Preview handling in Service; SM handles it based on Options
* I made the driver url parameter optional because we no longer use it, and deleted those constants
* Allows user to pass full path to driver in the Service class constructor to match other bindings; should reduce the number of backwards compatibility issues (see #12301)

### Motivation and Context
* Implements #11356 for .NET

### Considerations

This is not backwards compatible from the standpoint that if a user started the service with a random path and it wasn't valid, the code would ignore it and look on PATH or use Selenium Manager. This was a bug (#12301)

### DIFF
--- a/dotnet/src/webdriver/Chrome/ChromeDriverService.cs
+++ b/dotnet/src/webdriver/Chrome/ChromeDriverService.cs
@@ -67,6 +67,11 @@ namespace OpenQA.Selenium.Chrome
         /// <returns>A ChromeDriverService using a random port.</returns>
         public static ChromeDriverService CreateDefaultService(string driverPath)
         {
+            if (Path.GetFileName(driverPath).Contains(DefaultChromeDriverServiceExecutableName))
+            {
+                driverPath = Path.GetDirectoryName(driverPath);
+            }
+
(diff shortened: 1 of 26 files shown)


diemol:
Do we need to fix the Firefox tests before merging this? Seems Chrome usually passes after a retry.

titusfortner:
The issue is on trunk as well. Somehow we're losing the geckodriver connection and I don't know why. It passes locally, and usually passes on retry.

diemol:
Yes, I was running them locally and they all pass. Strange.

diemol (approved):

PR #12345, opened 2023-07-12
[rb] have Selenium Manager binary locate drivers on PATH

titusfortner:
### Description
* Remove the use of `Platform#find_binary` in `DriverFinder`
* Remove `#find_binary` from `Platform`
* Fixed unit tests

### Motivation and Context
* Ruby implementation of #11356 

### Considerations
* It is possible for this to be a breaking change since the `find_binary` logic may differ slightly from the binary, but it is unlikely and warnings will be thrown in the console.
* Are there any of these methods in this API private class that I'm deleting that we want to keep?


p0deje (approved):
Looks great!

Issue #11676, opened 2023-02-20
[🚀 Feature]: Selenium Manager downloads latest version of Chrome browser


bonigarcia:
### Status
* Rust - #12353 
* Ruby - #12398 
* Python - #12404 
* .NET - #12405 
* JS - #12411 
* Java - #12424

### Feature and motivation

The next major feature to be implemented in Selenium Manager is the automated management (download, cache) of the principal browsers nowadays: Chrome, Firefox, and Edge.

This issue gets track of the implementation of this feature for downloading the latest version of **Chrome** on the Rust side.

### Usage example

```
selenium-manager --browser chrome --resolve-browser
```

bonigarcia:
Maybe the flag ` --resolve-browser` (specified in the issue description) is not required. When Selenium Manager is called as follows:

```
selenium-manager --browser chrome
```

If the browser (Chrome, in this case) is locally installed, it is detected (and ultimately, used by Selenium). When it is not detected, it is downloaded (using [CfT endpoints](https://github.com/GoogleChromeLabs/chrome-for-testing#json-api-endpoints) for discovery).

titusfortner:
I'm updating this issue to track all the bindings.

diemol:
All PRs got merged!

Issue #11678, opened 2023-02-20
[🚀 Feature]: Selenium Manager downloads fixed version of Chrome browser (Rust)


bonigarcia:
### Feature and motivation

The next major feature to be implemented in Selenium Manager is the automated management (download, cache) of the principal browsers nowadays: Chrome, Firefox, and Edge.

This issue gets track of the implementation of this feature for managing a given version of **Chrome** (e.g., 101) on the Rust side.

### Usage example

```
selenium-manager --browser chrome --browser-version 101 --resolve-browser
```

diemol:
I missed this issue before. The problem with Chrome is that you cannot download old versions. You can only download the current, beta, dev, and canary.

bonigarcia:
That's true. The alternative to install former versions of Chrome is installing [Chromium](https://www.chromium.org/getting-involved/download-chromium/).

titusfortner:
ooc, How different is Chrome from Chromium?

bonigarcia:
A Chromium build is equivalent to Chrome but doesn't include proprietary parts, such as proprietary media formats such as AAC, H.264, or MP3. Also, Chrome is evergreen (i.e., it updates itself automatically), while Chromium is not. You can check a detailed comparison [here](https://www.browserstack.com/guide/difference-between-chrome-and-chromium) or [here](https://www.lambdatest.com/blog/difference-between-chrome-and-chromium/).

I would say that for most automated testing use cases, Chromium and Chrome are equivalent. Of course, there can be some cases in which Chrome (and not Chromium) is mandatory (e.g., videoconferencing web applications), but I believe it is a minority.

titusfortner:
Do we even have a good place to download chromium binaries from?

I think we're ok using equivalent Chromium if people want historical versions

bonigarcia:
> Do we even have a good place to download chromium binaries from?

Yes, it is explained in the link I posted above ([Chromium](https://www.chromium.org/getting-involved/download-chromium/), see "Downloading old builds of Chrome / Chromium"). All the Chromium releases are hosted in the official [repository](https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html). The idea is that Selenium Manager will do that process in an automated manner when the user requests an old version of Chrome.

titusfortner:
Right, but we need to be able to match up the version number with the build number, and I haven't seen anywhere that maintains a good list of that; would we need to create our own?

bonigarcia:
[...]

PR #12353, opened 2023-07-12
[rust] Download latest version of Chrome for Testing (#11676)

bonigarcia:
### Description
This PR implements the first feature for automated browser management in Selenium Manager, concretely for the latest version of [Chrome for Testing](https://googlechromelabs.github.io/chrome-for-testing/) (CfT). 

The general procedure of this feature is as follows. When the user (e.g., the bindings) invokes Selenium Manager, currently, only the driver is managed (i.e., downloaded and cached). With this PR, when managing Chrome, the browser binary will also be managed (if it is not present in the system). Moreover, this PR includes the flag `--force-browser-download` to force downloading of the browser (even in the -usual- case that Chrome is already installed). I created a [Gist document](https://gist.github.com/bonigarcia/167434561a5c6235ef77972b156337ad) to showcase the most representative cases.

Different things have changed with PR and deserve attention and discussion. Comments are welcome. Maybe the most relevant change is the output of Selenium Manager. So far, when using the JSON output, the field `result.message` contained the driver path. With this PR, this is no longer convenient since Selenium Manager can return two things: the driver path and (optionally) the browser path. Therefore, with this PR, there are two new JSON fields: `result.driver_path` and `result.browser_path`. These values should be read in the bindings. See the previous Gist snippets for examples.

A second change has to do with the browser TTL. So far, the browser TTL was used to cache the browser versions discovered by running shell command(s). But by default, the `browser_TTL` was `0`, so in practice, this caching mechanism (stored in the local metadata file, i.e., `~/.cache/selenium/selenium-manager.json`) is not used (although the logic is present in the Rust code).

With this PR, network calls to the CfT endpoints are used to discover the latest version of Chrome (very similar to what happens with chromedriver and other drivers). I believe these discovered versions deserve better to be cached in the local metadata. Therefore, I made two changes related to this:

1. Stop caching the discovered local browser versions (by running shell commands). Therefore, shell commands are always used to discover the browser version.
2. Use the existing mechanism to cache the discovered browser versions to be downloaded (e.g., CfT releases). This way, the second time a browser needs to be downloaded, the version can be already cached in the metadata.
3. Change the default value of `browser_TTL`, from  `0` to `3600` seconds (i.e., 1 hour, in the same way than `driver_TTL`. As a possible idea for the future, we can unify both TTLs in a single.

Finally, the overall timeout for network requests has been increased from `180` (in seconds) to `300` since browser download takes longer than driver download.

### Motivation and Context
This is the first PR related to milestone 4 in the original [Selenium Manager roadmap](https://github.com/SeleniumHQ/selenium/tree/trunk/rust), i.e., "M4: Browser management: Chrome/Chromium".

I have tested this feature both in Windows and Linux. Moreover, the tests are executed in Windows, Linux, and macOS on GitHub actions. Unfortunately, I no longer have a macOS, so I did have the chance to test it locally on a macOS machine. It would be great if someone could try this PR on macOS.

### Types of changes

### Checklist


titusfortner:
I'm not sure I understand the `--force-browser-download` approach.
I only want to download it if I need it...

* If the browser isn't found in expected directory, on PATH and user didn't specify, then presumably the code would fail, and Selenium Manager should always download it.
* If the browser is found but it isn't the version specified by the user, Selenium Manager should download the right version.

And it doesn't look like we always get the browser location back. I think we should always report the browser location for what we are using to match the driver. 

When running command line without `--output json` is it just outputting driver location?

What is the difference now between using `--browser` and `--driver`?

diemol:
>A second change has to do with the browser TTL. So far, the browser TTL was used to cache the browser versions discovered by running shell command(s). But by default, the browser_TTL was 0, so in practice, this caching mechanism (stored in the local metadata file, i.e., ~/.cache/selenium/selenium-manager.json) is not used (although the logic is present in the Rust code).

Do we really need this on this first iteration? In short, we want to download the browser if the user does not have it. And starting to cache versions from CfT would mean that we want to update the browser as well?

bonigarcia:
> I'm not sure I understand the `--force-browser-download` approach. I only want to download it if I need it...

From the bindings perspective, you can safely ignore this flag. I added it for two reasons. First, I need a way to test this feature in CI (CfT is only downloaded when Chrome is unavailable, and in GH Actions, Chrome is already installed). Second, it can potentially be used by someone to download CfT even when a complete Chrome release is already in the system.

> * If the browser isn't found in expected directory, on PATH and user didn't specify, then presumably the code would fail, and Selenium Manager should always download it.

That is what this feature does. When Chrome is not found, the latest version of CfT is downloaded.

> * If the browser is found but it isn't the version specified by the user, Selenium Manager should download the right version.

That is not part of this PR. It will be implemented for #11678.
 
> And it doesn't look like we always get the browser location back. I think we should always report the browser location for what we are using to match the driver.

Currently, the browser path is only returned when CfT is downloaded. But indeed, Selenium Manager can always return the browser path (the found one or the downloaded one). Do you want to implement it that way?

> When running command line without `--output json` is it just outputting driver location?

See "Resolving chrome (forced) for the first time (regular output)" on the  [Gist document](https://gist.github.com/bonigarcia/167434561a5c6235ef77972b156337ad).

> What is the difference now between using `--browser` and `--driver`?

Selenium Manager works in the same way for both flags. It is a way to select a "manager" (i.e., a module for managing driver/browser for different vendors, such as Chrome, Firefox, or Edge) by browser name (chrome, firefox, edge) or driver name (chromedriver, geckodriver, msedgedriver). But once the manager is defined, the behavior is the same.

bonigarcia:
> Do we really need this on this first iteration? In short, we want to download the browser if the user does not have it. And starting to cache versions from CfT would mean that we want to update the browser as well?

The TTL mechanism is a way to improve the overall performance of the browser/driver management. So strictly, we don't need it for this iteration. But since this is already implemented for drivers, I decided to use it for browser version discovery like it is already used for driver version discovery (which, in both cases, it requires network requests to remote endpoints). And as usual, we can put it to `0` to turn it off.

diemol:
> > Do we really need this on this first iteration? In short, we want to download the browser if the user does not have it. And starting to cache versions from CfT would mean that we want to update the browser as well?
> 
> The TTL mechanism is a way to improve the overall performance of the browser/driver management. So strictly, we don't need it for this iteration. But since this is already implemented for drivers, I decided to use it for browser version discovery like it is already used for driver version discovery (which, in both cases, it requires network requests to remote endpoints). And as usual, we can put it to `0` to turn it off.

That is the question, do we need to do browser version discovery always or only when we do not find the Chrome browser in the system?

bonigarcia:
> That is the question, do we need to do browser version discovery always or only when we do not find the Chrome browser in the system?

Browser version discovery using shell commands is always done. And now, when the browser is not discovered in the system, remote (for download) browser version discovery (using CfT endpoints) is used. And that found version is written in the metadata (considered fresh using the `TTL_browsers`).

titusfortner:
> Currently, the browser path is only returned when CfT is downloaded. But indeed, Selenium Manager can always return the browser path (the found one or the downloaded one). Do you want to implement it that way?

Yes, please. Right now the bindings can't even make use of the feature you implemented to reference beta/dev versions because we have no way to tell the driver to use them.

Currently we can get the browser binary path from the options class, and we next need to implement setting it. If we always have the browser value returned we can always just set it.

bonigarcia:
I managed to run this PR in a macOS VM and found a relevant difference. In macOS, the CtF binary package is called `Google Chrome for Testing.app`, which is something that the Selenium Manager logic needs to consider. I have just committed a second patch to this PR to fix this difference.

Also, I have updated the [Gist document](https://gist.github.com/bonigarcia/167434561a5c6235ef77972b156337ad), including the Selenium Manager output in macOS (see the two last snippets).

bonigarcia:
> > Currently, the browser path is only returned when CfT is downloaded. But indeed, Selenium Manager can always return the browser path (the found one or the downloaded one). Do you want to implement it that way?
> 
> Yes, please. Right now the bindings can't even make use of the feature you implemented to reference beta/dev versions because we have no way to tell the driver to use them.
> 
> Currently we can get the browser binary path from the options class, and we next need to implement setting it. If we always have the browser value returned we can always just set it.

No problem, I will do that. I inspected what needs to be done in the code, and it requires different changes. So I prefer to do it in a separate PR since this one is already quite big.

bonigarcia:
I have implemented the requested feature, i.e., always returning the browser path (the found one or the downloaded one). For that, I have a local commit on top of the commits of this PR.

Should I commit these new changes as part of this PR, or should I wait until this PR is accepted and merged?

nir-tal-talkspace:
this PR is important - 
this is the log when trying to run SM with Chrome v115 currently SM downgrades chromedriver to v114 instead of using v115

```
https://www.selenium.dev/documentation/webdriver/drivers/service/#setting-log-output
Jul 19, 2023 3:38:25 PM org.openqa.selenium.manager.SeleniumManager lambda$runCommand$1
WARNING: Error getting version of chromedriver 115. Retrying with chromedriver 114 (attempt 1/5)
```

@titusfortner @diemol @bonigarcia

bonigarcia:
@nir-tal-talkspace In fact, the important PR for discovering chromedriver 115+ is #12208. It is already merged, but unfortunately, it is unavailable in Selenium 4.10.0.

bonigarcia:
I believe we don't want to merge this PR until Selenium 4.11.0 is released, so I have just sent two new commits to this PR. With these new commits:

1. Selenium Manager always returns the browser path: the found one (already installed) or the downloaded one (CfT). 
2. Selenium Manager tries to find the browser in the PATH when it is not found in the usual locations (e.g., `/Applications/Google Chrome.app/Contents/MacOS/Google Chrome` in macOS, etc.). This case (i.e., the browser is in the PATH and not in the known locations) is not usual, but it makes Selenium Manager a bit more robust in finding browser paths.

Here it is some examples of the output:

```
C:\Users\boni\Documents\dev\selenium\rust>cargo run -- --browser chrome --debug
DEBUG   Checking chromedriver in PATH
DEBUG   Running command: chromedriver --version
DEBUG   Output: ""
DEBUG   chromedriver not found in PATH
DEBUG   chrome detected at C:\Program Files\Google\Chrome\Application\chrome.exe
DEBUG   Using shell command to find out chrome version
DEBUG   Running command: wmic datafile where name='C:\\Program Files\\Google\\Chrome\\Application\\chrome.exe' get Version /value
DEBUG   Output: "\r\r\n\r\r\nVersion=115.0.5790.99\r\r\n\r\r\n\r\r\n\r"
DEBUG   Detected browser: chrome 115.0.5790.99
DEBUG   Reading metadata from https://googlechromelabs.github.io/chrome-for-testing/known-good-versions-with-downloads.json
DEBUG   Required driver: chromedriver 115.0.5790.98
DEBUG   Driver URL: https://edgedl.me.gvt1.com/edgedl/chrome/chrome-for-testing/115.0.5790.98/win64/chromedriver-win64.zip
INFO    Driver path: C:\Users\boni\.cache\selenium\chromedriver\win64\115.0.5790.98\chromedriver.exe
INFO    Browser path: C:\Program Files\Google\Chrome\Application\chrome.exe
```

```
C:\Users\boni\Documents\dev\selenium\rust>cargo run -- --browser chrome --output json
{
  "logs": [
    {
      "level": "INFO",
      "timestamp": 1689852905,
      "message": "Driver path: C:\\Users\\boni\\.cache\\selenium\\chromedriver\\win64\\115.0.5790.98\\chromedriver.exe"
    },
    {
      "level": "INFO",
      "timestamp": 1689852905,
      "message": "Browser path: C:\\Program Files\\Google\\Chrome\\Application\\chrome.exe"
    }
  ],
  "result": {
    "code": 0,
    "message": "C:\\Users\\boni\\.cache\\selenium\\chromedriver\\win64\\115.0.5790.98\\chromedriver.exe",
    "driver_path": "C:\\Users\\boni\\.cache\\selenium\\chromedriver\\win64\\115.0.5790.98\\chromedriver.exe",
    "browser_path": "C:\\Program Files\\Google\\Chrome\\Application\\chrome.exe"
  }
}
```

For the bindings, the JSON output is what matters. Notice that now, there are two new fields: `result.driver_path` and `result.browser_path`. I left the driver path in `result.message`, but maybe it is time to make a breaking change and stop using `result.message` for driver's path, and leave that field for error messages only.

@diemol @titusfortner What do you think?

titusfortner:
@bonigarcia can you rebase this one? If Google isn't going to fix their bug right away, maybe we should knock out the fix in Selenium after all since none of the 3rd party managers can do what we can do. 😆

titusfortner:
Shouldn't this get v116? Or are we only working on stable right now?
```
2023-07-21 16:28:20 DEBUG Selenium [:selenium_manager] Executing Process ["/Users/titusfortner/code/selenium/bazel-bin/rb/bin/macos/selenium-manager", "--browser", "chrome", "--output", "json", "--browser-version", "116", "--debug"] 
2023-07-21 16:28:23 DEBUG Selenium [:selenium_manager] Checking chromedriver in PATH 
2023-07-21 16:28:23 DEBUG Selenium [:selenium_manager] Running command: chromedriver --version 
2023-07-21 16:28:23 DEBUG Selenium [:selenium_manager] Output: "" 
2023-07-21 16:28:23 DEBUG Selenium [:selenium_manager] chromedriver not found in PATH 
2023-07-21 16:28:23 DEBUG Selenium [:selenium_manager] Reading metadata from https://googlechromelabs.github.io/chrome-for-testing/known-good-versions-with-downloads.json 
2023-07-21 16:28:23 DEBUG Selenium [:selenium_manager] Required driver: chromedriver 116.0.5845.42 
2023-07-21 16:28:23 DEBUG Selenium [:selenium_manager] Driver URL: https://edgedl.me.gvt1.com/edgedl/chrome/chrome-for-testing/116.0.5845.42/mac-x64/chromedriver-mac-x64.zip 
2023-07-21 16:28:23 DEBUG Selenium [:selenium_manager] Driver path: /Users/titusfortner/.cache/selenium/chromedriver/mac-x64/116.0.5845.42/chromedriver 
```

bonigarcia:
> @bonigarcia can you rebase this one?

Done

> Shouldn't this get v116? Or are we only working on stable right now?

Not yet, but it will be shortly (I want to do it next week). As explained before, this PR implements #11676, which is about the latest Chrome. Managing different versions than stable will be another PR (implementing #11678).

titusfortner:
I think the non-rust stuff in CI failures is ok, but some rust unit tests are failing, please double check. I'll rebase #12398 on this and see if everything makes sense integration wise from Ruby bindings.

titusfortner:
When we pass in `--browser-path` can that value still be output in result?

titusfortner:
and I think there's been a regression with the escaping of things (might not be in this PR, I'm not testing against trunk)
when setting `options.binary = path_to_chrome` in Ruby

These work:
```
"'/Applications/Google Chrome Dev.app/Contents/MacOS/Google Chrome Dev'"
'"/Applications/Google Chrome Dev.app/Contents/MacOS/Google Chrome Dev"'
```
But not these:
```
"/Applications/Google Chrome Dev.app/Contents/MacOS/Google Chrome Dev"
"/Applications/Google\ Chrome\ Dev.app/Contents/MacOS/Google\ Chrome\ Dev"
'/Applications/Google Chrome Dev.app/Contents/MacOS/Google Chrome Dev'
'/Applications/Google\ Chrome\ Dev.app/Contents/MacOS/Google\ Chrome\ Dev'
```
I can correct it in Ruby code, but would rather not have to...


```
2023-07-22 14:56:20 DEBUG Selenium [:selenium_manager] Selenium Manager binary found at /Users/titusfortner/code/selenium/bazel-bin/rb/bin/macos/selenium-manager 
2023-07-22 14:56:20 DEBUG Selenium [:selenium_manager] Executing Process ["/Users/titusfortner/code/selenium/bazel-bin/rb/bin/macos/selenium-manager", "--browser", "chrome", "--browser-path", "/Applications/Google Chrome Dev.app/Contents/MacOS/Google Chrome Dev", "--output", "json", "--debug"] 
2023-07-22 14:56:20 DEBUG Selenium [:selenium_manager] Checking chromedriver in PATH 
2023-07-22 14:56:20 DEBUG Selenium [:selenium_manager] Running command: chromedriver --version 
2023-07-22 14:56:20 DEBUG Selenium [:selenium_manager] Output: "" 
2023-07-22 14:56:20 DEBUG Selenium [:selenium_manager] chromedriver not found in PATH 
2023-07-22 14:56:20 DEBUG Selenium [:selenium_manager] Using shell command to find out chrome version 
2023-07-22 14:56:20 DEBUG Selenium [:selenium_manager] Running command: /Applications/Google Chrome Dev.app/Contents/MacOS/Google Chrome Dev --version 
2023-07-22 14:56:20 DEBUG Selenium [:selenium_manager] Output: "" 
2023-07-22 14:56:20 DEBUG Selenium [:selenium_manager] chrome has not been discovered in the system 
```

titusfortner:
I wrote a suite of integration tests for the behaviors I'm expecting to see in future releases. https://gist.github.com/titusfortner/4da1670b5ecaf5ca328c03bde7e80a80

I can't commit it because it is based on my having installed everything (or not) on my system, but I have something to help evaluate future updates.

With the one exception above, everything I'm expecting to see passing right now is passing.

I'm pleased that with this code the bindings can now use the special browser-version values that SM implemented last year.
Chrome & Edge:  `beta`, `dev` and `canary`
Firefox: `dev` and `nightly`

Also, I like that we are throwing an error when a bad browser version is specified

titusfortner:
Hmm, Windows might not be happy, though. Investigating

bonigarcia:
> I think the non-rust stuff in CI failures is ok, but some rust unit tests are failing, please double check.

Now it's fixed.

> When we pass in `--browser-path` can that value still be output in result?

With the latest changes, yes.

> and I think there's been a regression with the escaping of things

I included a couple of new commits in this PR to make more robust the path manipulation (escaped or not).

> I wrote a suite of integration tests for the behaviors I'm expecting to see in future releases.

Great, let me know if everything works as expected with the latest changes.

titusfortner:
Yes, everything works exactly as expected on MacOS and Windows!
Including the weird windows parsing bug so I can even remove that extra fix in Ruby.
Excellent.

titusfortner (approved):

titusfortner:
Oh, one thing... I think if "beta" "dev" "canary" "nightly" etc aren't found on the system it should error and not download the latest stable version of the browser.
Minor point, though, and we can fix it later if we agree.

diemol (approved):

PR #12354, opened 2023-07-12
[grid] removed a deprecated json wire endpoint

joerg1985:
### Description
Dropped the `/session/{sessionId}/file` endpoint used by json wire protocol and updated a test still using it.

### Types of changes

### Checklist


joerg1985:
There is the `get("/se/grid/node/status")` endpoint which is not mentioned in the [docs](https://www.selenium.dev/documentation/grid/advanced_features/endpoints/#status). Either the docs are incomplete or this endpoint might be not needed?
https://github.com/SeleniumHQ/selenium/blob/067f10a3a3b0f6f0c2b26f4e12a2be264085b2b1/java/src/org/openqa/selenium/grid/node/Node.java#L183-L185

diemol:
> There is the `get("/se/grid/node/status")` endpoint which is not mentioned in the [docs](https://www.selenium.dev/documentation/grid/advanced_features/endpoints/#status). Either the docs are incomplete or this endpoint might be not needed?
> 
> https://github.com/SeleniumHQ/selenium/blob/067f10a3a3b0f6f0c2b26f4e12a2be264085b2b1/java/src/org/openqa/selenium/grid/node/Node.java#L183-L185

This seems a duplicate, I thought it was being used internally but it is just not used. We can remove it in a separate PR.

diemol (commented):
While this is correct, my concern is that this will break Selenium 3.x clients, and this would force them to upgrade to Selenium 4. 

Latest Selenium 3.x can do W3C already, but this part with the prefixed `file` endpoint is not there. I've seen several people using Grid 4 with 3.x clients.

I would say we need to leave this for Selenium 5.

titusfortner:
I don't understand the downside. If they are stuck on Selenium 3 client, they can "only" use up to Selenium 4.11?
I guess I don't see a huge value in supporting 5 year old code still. Is this the only thing we're choosing to keep around?

diemol:
I would need to check again, but if recall correctly, over 50% downloads were still latest 3.x, not sure if we want to break them.

titusfortner:
How many of those 50% are using Grid 4.x? (Least problems happen when keeping Grid & bindings in sync)
Why not encourage those 50% to upgrade to 4.x bindings?

diemol:
What is the need to remove this now when the maintenance cost is very low or none?

titusfortner:
We can say that about any one thing we choose to remove. /shrug
I'm not going to die on this hill, but I don't think we should be waiting for a major version change to remove it.

diemol:
OK, I gave my opinion but it seems to not be important even when some data was added to back it up. Let's merge it then.

PR #12369, opened 2023-07-14
Create Client config class for http client values

titusfortner:
### Status

* Dependent on #12364 — putting code here as placeholder; do not need to review until that one is merged.

### Description
* Creates a Client config class like Java has to store configuration values related to Http Client
* Have users set "keep alive" in client config instead of in Options since it is not related to capabilities
* Have users set "ignore proxy" in client config instead of webdriver directly; this is done by creating a proxy of type "Direct" instead of default "System"

Browser subclasses for `RemoteConnection` and `WebDriver` set deprecated value defaults to `None`
Remote superclass for those throw the deprecation warning if they are not `None` so subclasses still need to send them.

@isaulv told me to be careful creating objects in constructor parameters, so I'm just using `None` and setting defaults in first line.

Also, I separated out the code for whether to use a proxy into a new private method 

### Motivation and Context
This is to match what Java is doing and better organize the constructor parameters
also sets us up to make more values configurable in a future PR.

### DIFF
--- a/py/selenium/webdriver/chrome/remote_connection.py
+++ b/py/selenium/webdriver/chrome/remote_connection.py
@@ -1,0 +1,29 @@
+# Licensed to the Software Freedom Conservancy (SFC) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The SFC licenses this file
+# to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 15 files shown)


diemol:
I believe this was already implemented by https://github.com/SeleniumHQ/selenium/commit/7f8a53207e41dfeeab7d4bade285b38b4e82f040

PR #12377, opened 2023-07-18
[py] added locator validation in `locate_with` method in `relative_locator.py`

sandeepsuryaprasad:
**Thanks for contributing to Selenium!**
**A PR well described will help maintainers to quickly review and merge it**

Before submitting your PR, please check our [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) guidelines.
Avoid large PRs, help reviewers by making them as simple and short as possible.

### Description
Added locator type validation in `locate_with` function in `relative_locator.py` to validate for one of 8 locator types.
### Motivation and Context
- Locator type is not getting validated in `locate_with` function. I have added that validation. 
- If the user provides locator type other than the 8 available one's, I am raising exception.
- Validating using `assert` is good in test methods, but not in a library function.  

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [] New feature (non-breaking change which adds functionality)
- [] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist

- [x] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) document.
- [] My change requires a change to the documentation.
- [] I have updated the documentation accordingly.
- [] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #12378, opened 2023-07-18
[py] removed `with_tag_name` function from `relative_locator.py`

sandeepsuryaprasad:

### Description
This PR is linked to the changes made in https://github.com/SeleniumHQ/selenium/pull/12377
`with_tag_name` function is removed from relative_locator.
### Motivation and Context
- Since `locate_with` function takes care of finding relative element using `tagname`, `with_tag_name` can be eliminated.
### Types of changes

### Checklist


titusfortner:
But `with_tag_name()` is a public method, so removing it can break user code. The way relative locators were implemented earlier made this pattern more important, other languages moved away from it and Python didn't. But, we would need to deprecate this first.

Regardless, we currently have a moratorium on changes to code in support modules/directories for now until we know how we are going to maintain them going forward.

PR #12385, opened 2023-07-19
[grid] improved the new session handling when the queue is empty

joerg1985:
### Description
This PR implements two strategies to reduce the overhead of the new session runnable:
- peek into the a local queue to avoid collecting stereotypes when the queue is empty.
- ensure all slots share the same stereotype instance to hit the fast-path in .equals

### Motivation and Context
A idle selenium server will check the session queue each 10ms, therefore the stereotypes of the slots need to be collected. This will make the idle server not really idle, this can be observed when profiling the GC or looking at the power consumption of an VM. The more cores/threads the system has available, the more slots need to be collected, the more garage is generated.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/ImmutableCapabilities.java
+++ b/java/src/org/openqa/selenium/ImmutableCapabilities.java
@@ -135,22 +135,7 @@ public class ImmutableCapabilities implements Capabilities {
   }
 
   public ImmutableCapabilities(Capabilities other) {
-    Require.nonNull("Capabilities", other);
-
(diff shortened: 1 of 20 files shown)


joerg1985:
@diemol this PR will avoid the lookup in a remote queue and still improve the situation with a local queue.
Is this better than in the original #12336 where the queue was always queried?

diemol (approved):
Yes, this solution is a very good idea. Thank you, @joerg1985!

alokeutpal (approved):

PR #12398, opened 2023-07-21
[rb] add browser output from selenium manager to options

titusfortner:
### Status
Draft because:
* ~proof of concept — looks like it'll be relatively easy to wire these in if we want to try to do it before 4.11?~
* ~waiting on #12353~
* ~waiting on binaries getting committed to trunk - https://github.com/SeleniumHQ/selenium/actions/runs/5646633604~
* ~This corrects for what looks to be a change in how SM binary escapes characters, not sure we want that~
* ~needs tests~

### Description
Sets browser location on Options class if supported (Chrome, Edge, Firefox only)
* Errors on Chrome & Edge if browser version invalid (Not backwards compatible, but GOOD!)
* Uses Dev/Beta/Canary/Nightly if found on system and specified by name as `browser_version`
* Downloads Stable Chrome For Testing if Chrome not installed on System.

### Motivation and Context
* This ensures that the browser Selenium Manager is checking against is the one actually getting used by the driver

### DIFF
--- a/rb/lib/selenium/webdriver/common/selenium_manager.rb
+++ b/rb/lib/selenium/webdriver/common/selenium_manager.rb
@@ -39,17 +39,24 @@ module Selenium
         def driver_path(options)
           command = generate_command(binary, options)
 
-          location = run(*command)
-          WebDriver.logger.debug("Driver found at #{location}", id: :selenium_manager)
-          Platform.assert_executable location
+          output = run(*command)
 
-          location
+          browser_path = output['browser_path']
+          driver_path = output['driver_path']
+          Platform.assert_executable driver_path
+
+          if options.respond_to? :binary
+            options.binary = browser_path
+            options.browser_version = nil
+          end
+
+          driver_path
         end
 
         private
 
         def generate_command(binary, options)
-          command = [binary, '--browser', options.browser_name, '--output', 'json']
+          command = [binary, '--browser', options.browser_name]
           if options.browser_version
             command << '--browser-version'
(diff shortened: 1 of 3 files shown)


titusfortner:
@diemol / @bonigarcia yeah, I have no idea how to do this in Java because of the crazy `SeleniumManagerJsonOutput` implementation. Also, I'm not sure how Java can do `setBinary()` in a reasonable way if the Type is `Capabilities`

rb/lib/selenium/webdriver/common/selenium_manager.rb
```diff
@@ -39,30 +39,36 @@ def bin_path
...
         def generate_command(binary, options)
-          command = [binary, '--browser', options.browser_name, '--output', 'json']
+          command = [binary, '--browser', options.browser_name]
           if options.browser_version
             command << '--browser-version'
             command << options.browser_version
           end
           if options.respond_to?(:binary) && !options.binary.nil?
             command << '--browser-path'
-            command << options.binary.gsub('\\', '\\\\\\')
```
p0deje: Is this really fixed in SM now?

p0deje (approved):

rb/lib/selenium/webdriver/common/selenium_manager.rb
titusfortner: I ran without it and it worked with spaces, but now that I think about it, it was there for if the user backspaced their own spaces? I'll put it back.

PR #12407, opened 2023-07-24
[java] Filter related fixed in the NettyClient and the JdkHttpClient

joerg1985:
### Description
The NettyClient had a `with` implementation different to the default implementation of the interface. A caller would propably expect the interface behaviour. The PR will fix this and also fix the filters are applied twice for the NettyClient and not applied for the JdkHttpClient.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java
+++ b/java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java
@@ -51,22 +51,12 @@ import org.openqa.selenium.Credentials;
 import org.openqa.selenium.TimeoutException;
 import org.openqa.selenium.UsernameAndPassword;
 import org.openqa.selenium.WebDriverException;
-import org.openqa.selenium.remote.http.BinaryMessage;
-import org.openqa.selenium.remote.http.ClientConfig;
-import org.openqa.selenium.remote.http.CloseMessage;
-import org.openqa.selenium.remote.http.ConnectionFailedException;
-import org.openqa.selenium.remote.http.HttpClient;
-import org.openqa.selenium.remote.http.HttpClientName;
-import org.openqa.selenium.remote.http.HttpMethod;
-import org.openqa.selenium.remote.http.HttpRequest;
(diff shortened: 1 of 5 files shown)


titusfortner:
Somewhat off-topic, but what's an example usage of a filter? I couldn't figure it out from looking at the code.

joerg1985:
The filters are used by the RemoteWebdriverBuilder for several reasons. E.g. logging the http traffic, closing the http client, exception handling. This all might be refactored in Selenium 5, because i think for all uses there are better places in the code. I only was duplicate logging of the http traffic, so i investigated what was wrong and fixed it for now.

java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java
```diff
@@ -51,22 +51,12 @@
...
-import org.openqa.selenium.remote.http.ConnectionFailedException;
-import org.openqa.selenium.remote.http.HttpClient;
-import org.openqa.selenium.remote.http.HttpClientName;
-import org.openqa.selenium.remote.http.HttpMethod;
-import org.openqa.selenium.remote.http.HttpRequest;
-import org.openqa.selenium.remote.http.HttpResponse;
-import org.openqa.selenium.remote.http.Message;
-import org.openqa.selenium.remote.http.TextMessage;
-import org.openqa.selenium.remote.http.WebSocket;
+ import org.openqa.selenium.remote.http.*;
```
diemol, line 54: We try to have all imports declared individually.

joerg1985: Sorry, intellij did refactor it on save. Should i restore the imports in a new PR?

diemol: Please restore them and push directly to `trunk`. No need for a PR.

joerg1985:
Finally managed it to push it to the trunk in https://github.com/SeleniumHQ/selenium/commit/842f413b2c2d5d2e40af7d4f4e6f654641fe7a75

PR #12409, opened 2023-07-24
[py]: Initial mechanism for better deprecations

symonk:
This is a stab at an initial deprecations mechanism to give better fine grained control over deprecations for users and it will be easier to understand what exactly should be removed in a given release.  It should also help us to aggressively chase after the deprecated code (some of which has been sitting for a while).

If you like the concept, I will implement it further, but the idea being we can give fine grained, reusable deprecation messages that are stored in one central place so easy to see the current state of deprecated code.

Let me know your thoughts, thanks

### DIFF
--- a/py/BUILD.bazel
+++ b/py/BUILD.bazel
@@ -16,7 +16,7 @@ compile_pip_requirements(
     requirements_txt = ":requirements_lock.txt",
 )
 
-SE_VERSION = "4.10.0"
+SE_VERSION = "4.11.0"
 
 BROWSER_VERSIONS = [
     "v85",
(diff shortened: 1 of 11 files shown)


AutomatedTester:
I'm not too keen on the Selenium<Version>Warning as it feels like it could cause issues for people remembering to update the version in the warning.

The overall idea is sound though

titusfortner:
Ruby does it like this - https://github.com/SeleniumHQ/selenium/blob/trunk/rb/lib/selenium/webdriver/common/logger.rb#L164-L176

So an alternative would be to have a method with these strings pasted into the boilerplate, and not need different classes or constants.
```py
SeleniumWarning().deprecate("desired_capabilities", "an options instance", "4.12")
```

I like specifying planned removal release. I should add that in Ruby.

symonk:
Closing this in favour of #12432 - lets discuss over there

PR #12417, opened 2023-07-25
[rb] remove deprecated code

titusfortner:
### Description
* Removes `#headless!` methods
* Enforces w3c compliance in Options classes
* finally removes the `Service#driver_opts` code; a poorly conceived idea to convert key/value pairs into args
* removed the class methods for browsers for `Remote::Capabilities`. Want to encourage people to move to `Options` as much as possible. (next release we remove ability for users to pass Capabilities instance to local drivers)
* removed the aliases for JWP capabilities

### Motivation and Context
This is everything that was deprecated for Selenium 4.8 in January 2023.
Removes a lot of extra code.


p0deje (approved):

compwron:
FYI (for anyone getting here later) this is a breaking change for using selenium for IE testing, oops.

titusfortner:
Yes, we only deprecate for breaking changes, so removing a deprecation is always a breaking change.
Selenium does not do semantic versioning.

PR #12432, opened 2023-07-27
[py] added decorator for handling deprecations.

sandeepsuryaprasad:

### Description
The motivation for this PR is https://github.com/SeleniumHQ/selenium/pull/12409
In this PR, I have created a decorator to handle deprecations in a much better and maintainable way.

### Motivation and Context
- I have created two decorators `deprecated_function` and `deprecated_attributes` in `selenium/deprecated.py`
- Any deprecated function would be decorated with `deprecated_function` decorator which takes care of logging warning message and from what version of selenium the function would be deprecated.
- Any deprecated instance variable would be decorated with `deprecated_attributes` decorator, which takes care of logging appropriate warning message for corresponding deprecated variable.
### Types of changes

### Checklist

### DIFF
--- a/py/selenium/deprecated.py
+++ b/py/selenium/deprecated.py
@@ -24,7 +24,7 @@ def deprecated_attributes(*attrs, message):
                 _attr = getattr(args[0], attr, None)
                 if _attr:
                     warn(f"'{message}': will be removed from {removed_from_version}", DeprecationWarning, stacklevel=2)
-                    return result
+            return result
 
         return wrapper
 
--- a/py/selenium/deprecated.py
+++ b/py/selenium/deprecated.py
@@ -41,7 +41,7 @@ def deprecated_attribute(message, **dep_attr):
         def wrapper(*args, **kwargs):
             func(*args, **kwargs)
             dep_attr_name = list(dep_attr.keys())[0]  # getting the name of deprecated attr
-            dep_attr_type = list(dep_attr.values())[0]  # getting should the value be truthy or falsy?
+            dep_attr_type = list(dep_attr.values())[0]  # should the value be truthy or falsy?
             # get the value of the deprecated attributes
             dep_attr_value = getattr(args[0], dep_attr_name)  # getting actual value passed to depattr
             # check the truthiness of the deprecated attribute
--- a/py/selenium/deprecated.py
+++ b/py/selenium/deprecated.py
@@ -1,0 +1,31 @@
+from warnings import warn
+
+removed_from_version = "4.12"
+
+
+def deprecated_function(message):
+    """decorator to log deprecation warning messgaes for deprecated methods."""
+
+    def _deprecated_function(func):
+        def wrapper(*args, **kwargs):
+            warn(f"{message}: will be removed from {removed_from_version}", DeprecationWarning, stacklevel=2)
+            return func(*args, **kwargs)
+
+        return wrapper
+
+    return _deprecated_function
+
+
+def deprecated_attributes(*attrs, message):
+    def _deprecated_attributes(func):
+        def wrapper(*args, **kwargs):
+            result = func(*args, **kwargs)
+            for attr in attrs:
+                _attr = getattr(args[0], attr, None)
+                if _attr:
+                    warn(f"'{message}': will be removed from {removed_from_version}", DeprecationWarning, stacklevel=2)
+                    return result
+
+        return wrapper
+
+    return _deprecated_attributes
--- a/py/selenium/deprecated.py
+++ b/py/selenium/deprecated.py
@@ -1,14 +1,33 @@
+# Licensed to the Software Freedom Conservancy (SFC) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The SFC licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from functools import wraps
 from warnings import warn
 
-removed_from_version = "4.12"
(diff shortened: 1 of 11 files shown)


titusfortner:
This is a cool idea, but I'm not sure the advantage of this over a more straightforward method? Does it show up differently in the IDE with this annotation? Does the wrapper know the name of the function so we don't have to put it in the string?

py/selenium/deprecated.py
```diff
@@ -0,0 +1,31 @@
+from warnings import warn
```
symonk, line 19: Thanks @sandeepsuryaprasad will review more shortly, I put together a somewhat similar idea so i'm not too biased, open to having just have a reusable mechanism to handle it in general that works well.

one minor thing is new files should be prefixed with the licensing meta data

py/selenium/deprecated.py
```diff
@@ -0,0 +1,31 @@
...
+
+removed_from_version = "4.12"
```
symonk: I would rather we provided a \`version=\` flag maybe, what do you think? We are not great and getting deprecated things out atm, I think due to also not following semantic versioning makes it a tad harder (imo), but allowing us to define this vs hard coding would be useful

py/selenium/deprecated.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def deprecated_function(message):
+    """decorator to log deprecation warning messgaes for deprecated methods."""
+
+    def _deprecated_function(func):
+        def wrapper(*args, **kwargs):
```
symonk, line 29: `@functools.wraps(func) here`

py/selenium/deprecated.py
```diff
@@ -0,0 +1,31 @@
...
+            return func(*args, **kwargs)
+
+        return wrapper
+
+    return _deprecated_function
+
+
+def deprecated_attributes(*attrs, message):
+    def _deprecated_attributes(func):
+        def wrapper(*args, **kwargs):
```
symonk, line 41: same here re: functools.wraps

py/selenium/deprecated.py
```diff
@@ -0,0 +1,31 @@
...
+            result = func(*args, **kwargs)
+            for attr in attrs:
+                _attr = getattr(args[0], attr, None)
+                if _attr:
+                    warn(f"'{message}': will be removed from {removed_from_version}", DeprecationWarning, stacklevel=2)
```
symonk: does `stacklevel=2` play nice with a decorator (especially one with args?) in terms of the stack in the output? would need to double check

py/selenium/deprecated.py
```diff
@@ -0,0 +1,31 @@
...
+    return _deprecated_function
+
+
+def deprecated_attributes(*attrs, message):
+    def _deprecated_attributes(func):
+        def wrapper(*args, **kwargs):
+            result = func(*args, **kwargs)
+            for attr in attrs:
+                _attr = getattr(args[0], attr, None)
+                if _attr:
```
symonk: not sure this work as you are expecting, e.g a deprecated attribute that been assigned `falsy`, e.g `log_path=False` would probably not warn here?

symonk:
this stuff is quite trick with edge cases, for example a `@staticmethod` wouldn't have an instance to pass as `args[0]` if we ever needed to decorate it, is it simpler if we define a function and call that in places we are currently deprecating things? ie.

```python
from selenium.deprecations import deprecated

deprecated(message=..., version=4.12")  
```
Theres possibly alot of edge cases with the decorator approach.
Either way, i'll close my draft PR since we can colloborate as part of this work here.

py/selenium/deprecated.py
sandeepsuryaprasad: > Thanks @sandeepsuryaprasad will review more shortly, I put together a somewhat similar idea so i'm not too biased, open to having just have a reusable mechanism to handle it in general that works well.
> 
> one minor thing is new files should be prefixed with the licensing meta data

Yes, I have added license stuff.

py/selenium/deprecated.py
sandeepsuryaprasad: I have changed it to `version` now. The reason I had named it as `removed_from_version` is that, it should not be confused with current version of selenium. It is the version (future) from which something will be removed.

py/selenium/deprecated.py
sandeepsuryaprasad: Yes this is added.

py/selenium/deprecated.py
sandeepsuryaprasad: I have added here as well

py/selenium/deprecated.py
sandeepsuryaprasad: I have tweaked the logic. Now we can pass arbitrary keyword argument to the decorator where we can specify "should the user be warned if the deprecated attribute is falsy or should the user be warned if the he tries to pass some value to deprecated attribute which evaluates to truthy"

sandeepsuryaprasad:
> This is a cool idea, but I'm not sure the advantage of this over a more straightforward method? Does it show up differently in the IDE with this annotation? Does the wrapper know the name of the function so we don't have to put it in the string?

Yes wrappers know the name of the function. `func.__name__` will give the name of the function which we have decorated.

sandeepsuryaprasad:
> this stuff is quite trick with edge cases, for example a `@staticmethod` wouldn't have an instance to pass as `args[0]` if we ever needed to decorate it, is it simpler if we define a function and call that in places we are currently deprecating things? ie.
> 
> ```python
> from selenium.deprecations import deprecated
> 
> deprecated(message=..., version=4.12")  
> ```
> 
> Theres possibly alot of edge cases with the decorator approach. Either way, i'll close my draft PR since we can colloborate as part of this work here.

I have checked for `static` and `class` methods, `deprecated_function` decorator works fine in both cases. But if we want to make a function instead of decorator, that is also perfectly fine..

sandeepsuryaprasad:
Getting `ModuleNotFound` error while trying to import the decorator from `deprecated.py` module. This is the file that is newly added to `selenium` folder.

sandeepsuryaprasad:
Actually we don't have to write custom decorator. In `typing_extensions` module we do have `deprecated` decorator.
and in python 3.12 (yet to be released) there is a PEP to include this in `typing` module
https://peps.python.org/pep-0702/#example

titusfortner:
aha, this is the PR I was looking for last week and couldn't find when I raised https://github.com/SeleniumHQ/selenium/pull/13402

I think we decided we want to use the annotation that is in `typing_extensions ` instead of this.

Issue #12427, opened 2023-07-27
[🐛 Bug]: Selenium-manager.exe console window briefly popping up


SenZmaKi:
### Selenium-manager.exe console window briefly popping up

When you build the script below into an executable using Pyinstaller by running `pyinstaller --windowed my_script.py`  a console window named selenium-manager.exe briefly pops up. This only occurs if you run the executable and not the script as a python file, I'm assuming this cause there's already a console window so windows doesn't create a new one.

I've narrowed it down to the following method `selenium.webdriver.common.selenium_manager.SeleniumManager.run`. I believe this is caused by the call to `subprocess.run`  which on windows if made without the flag`creationflags=subprocess.CREATE_NO_WINDOW` a console window is created if one is unavailable.

Possible fix:
```
if sys.platform == "win32":
     completed_proc = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, creationflags=subprocess.CREATE_NO_WINDOW)
else: 
     completed_proc = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```

### For the code below run: pyinstaller --windowed my_script.py

```python
from subprocess import CREATE_NO_WINDOW
from selenium.webdriver.chrome.service import Service
from selenium.webdriver import Chrome, ChromeOptions

def setup_options(options: ChromeOptions) -> ChromeOptions :
    options.add_argument('--headless=new')
    options.add_argument('--disable-extensions')
    options.add_argument('--disable-infobars')
    options.add_argument('--no-sandbox')
    return options

def setup_driver() -> Chrome:
    service_chrome = Service()
    service_chrome.creation_flags = CREATE_NO_WINDOW
    options = setup_options(ChromeOptions())
    return Chrome(service=service_chrome, options=options)

driver = setup_driver()
# Do some automation
driver.quit()

```

### Relevant log output

```shell
None
```

### Operating System

Windows-10-10.0.19045-SP0

### Selenium version

Python 4.10.0

### What are the browser(s) and version(s) where you see this issue?

Chrome 115.0.5790.110

### What are the browser driver(s) and version(s) where you see this issue?

ChromeDriver, MsedgeDriver

### Are you using Selenium Grid?

No

titusfortner:
Can you create a PR with that code? Thanks.

symonk:
Thanks for the write up @SenZmaKi, would happily take a PR for the fix

[...]

PR #12435, opened 2023-07-28
[py] Add creation flag for windows in selenium_manager

SenZmaKi:
Addresses Issue: #12427 

The call to `subprocess.run` is made without the flag `creationflags=CREATE_NO_WINDOW` for windows platform which results in a brief console window popping up if the script is built into an executable that doesn't already have a console window e.g for a GUI app. The following pull request looks to address the aforementioned issue.

Bug 
```python
completed_proc = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False)
```

Fix
```python
if sys.platform == "win32":
     completed_proc = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, creationflags=subprocess.CREATE_NO_WINDOW)
else: 
     completed_proc = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```

### DIFF
--- a/py/selenium/webdriver/common/actions/action_builder.py
+++ b/py/selenium/webdriver/common/actions/action_builder.py
@@ -39,19 +39,16 @@ class ActionBuilder:
         keyboard: Optional[KeyInput] = None,
         duration: int = 250,
     ) -> None:
-        if not mouse:
-            mouse = PointerInput(interaction.POINTER_MOUSE, "mouse")
-        if not keyboard:
-            keyboard = KeyInput(interaction.KEY)
-        if not wheel:
-            wheel = WheelInput(interaction.WHEEL)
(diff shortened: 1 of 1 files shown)


symonk (approved):

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -117,7 +117,10 @@ def run(args: List[str]) -> dict:
         command = " ".join(args)
         logger.debug(f"Executing process: {command}")
         try:
-            completed_proc = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False)
+            if sys.platform == "win32":
```
symonk, line 120: only query would be, any special case for `cygwin`?

symonk:
closes #12427

PR #12439, opened 2023-07-28
Rename methods in Actions class to be more closer to .net rules

nvborisenko:
EOM


titusfortner:
Oof, when Java devs write/review c# code.

To be honest I'm guessing no one is using this feature, but we probably should maintain backwards compatibility with the public methods.

nvborisenko:
We can mark existing methods as `Obsolete`.

nvborisenko:
I am so lazy :) Actually I think it's better just to rename it and do not make a panic that `setActivePointer` now is deprecated and user should use `SetActivePointer` instead. Yeah, I vote for silent breaking change.

PR #12444, opened 2023-07-30
Add explicit delimiters to node configs list

sbabcoc:
### Description
This PR adds explicit delimiters to the driver configuration settings returned by the implementations of `Config.getAll()`, rather than relying on the `display-name` setting as the delimiter. The existing strategy is fragile, because the addition of a new setting with an earlier ASCII-betical sort position would either be lost or grouped with the wrong collection of settings. 

### Motivation and Context
The objectives of this PR are to eliminate the fragility of the existing driver configuration partitioning strategy and to optimize the extraction of settings from the configuration list. (The previous implementation iterates the settings list multiple times to process the list of settings into Capabilities objects.)  
I revised a few variable names and added code comments to make it easier for human readers to follow the flow of execution through the code.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/grid/config/Config.java
+++ b/java/src/org/openqa/selenium/grid/config/Config.java
@@ -71,7 +71,6 @@ public interface Config {
         entry -> {
           return String.format("%s=%s", entry.getKey(), toJson(entry.getValue()));
         })
-      .sorted()
       .collect(Collectors.toList());
     // add record separator
     entryList.add(DELIMITER);
--- a/java/src/org/openqa/selenium/grid/config/Config.java
+++ b/java/src/org/openqa/selenium/grid/config/Config.java
@@ -23,6 +23,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
 import org.openqa.selenium.json.Json;
 
 public interface Config {
@@ -59,14 +61,22 @@ public interface Config {
     }
   }
 
+  String DELIM_KEY = "\u001E";
+  String DELIMITER = DELIM_KEY + "=\"record-separator\"";
+
   default List<String> toEntryList(Map<String, Object> mapItem) {
-    return mapItem.entrySet().stream()
-        .map(
-            entry -> {
-              return String.format("%s=%s", entry.getKey(), toJson(entry.getValue()));
-            })
-        .sorted()
-        .collect(ImmutableList.toImmutableList());
+    // transform config settings map into list of key/value strings
+    List<String> entryList = mapItem.entrySet().stream()
+      .map(
+        entry -> {
+          return String.format("%s=%s", entry.getKey(), toJson(entry.getValue()));
+        })
+      .sorted()
+      .collect(Collectors.toList());
+    // add record separator
+    entryList.add(DELIMITER);
+    // return immutable config settings list
+    return ImmutableList.<String>builder().addAll(entryList).build();
   }
 
   default String toJson(Object value) {
--- a/java/src/org/openqa/selenium/grid/node/config/NodeOptions.java
+++ b/java/src/org/openqa/selenium/grid/node/config/NodeOptions.java
@@ -344,143 +344,124 @@ public class NodeOptions {
   private void addDriverConfigs(
       Function<ImmutableCapabilities, Collection<SessionFactory>> factoryFactory,
       ImmutableMultimap.Builder<Capabilities, SessionFactory> sessionFactories) {
+
     Multimap<WebDriverInfo, SessionFactory> driverConfigs = HashMultimap.create();
-    config
-        .getAll(NODE_SECTION, "driver-configuration")
-        .ifPresent(
-            drivers -> {
-              /*
-               The four accepted keys are: display-name, max-sessions, stereotype, webdriver-executable.
-               The mandatory keys are display-name and stereotype. When configs are read, they keys always
-               come alphabetically ordered. This means that we know a new config is present when we find
-               the "display-name" key again.
-              */
 
-              if (drivers.size() == 0) {
-                throw new ConfigException("No driver configs were found!");
+    // get all driver configuration settings
+    config.getAll(NODE_SECTION, "driver-configuration")
+      // if settings exist
+      .ifPresent(drivers -> {
+        Map<String, String> configMap = new HashMap<>();
+        List<Map<String, String>> configList = new ArrayList<>();
+
+        // iterate over driver settings
+        for (String setting : drivers) {
+          // split this setting into key/value pair
+          String[] values = setting.split("=", 2);
+          // if format is invalid
+          if (values.length != 2) {
+            throw new ConfigException("Driver setting '" + setting
+              + "' does not adhere to the required 'key=value' format!");
+          }
+          // if this is a record separator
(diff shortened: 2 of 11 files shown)


java/src/org/openqa/selenium/grid/config/Config.java
```diff
@@ -59,14 +61,22 @@ default <X> X getClass(String section, String option, Class<X> typeOfClass, Stri
...
-            })
-        .sorted()
-        .collect(ImmutableList.toImmutableList());
+    // transform config settings map into list of key/value strings
+    List<String> entryList = mapItem.entrySet().stream()
+      .map(
+        entry -> {
+          return String.format("%s=%s", entry.getKey(), toJson(entry.getValue()));
+        })
+      .sorted()
```
sbabcoc: With the addition of an explicit delimiter, sorting is probably no longer necessary.

java/src/org/openqa/selenium/grid/node/config/NodeOptions.java
```diff
@@ -344,143 +344,124 @@ private SessionFactory createSessionFactory(String clazz, Capabilities stereotyp
...
+              throw new ConfigException("Found config with no 'stereotype' setting! " + configMap);
+            }
+
+            // add config to list
+            configList.add(configMap);
+            // prepare for next config
+            configMap = new HashMap<>();
+          } else {
+            // add setting to config
+            configMap.put(values[0], unquote(values[1]));
```
sbabcoc, line 389: It would be quite easy to ensure that only supported key names are used, and that each key is only specified once.

java/src/org/openqa/selenium/grid/node/config/NodeOptions.java
```diff
@@ -344,143 +344,124 @@ private SessionFactory createSessionFactory(String clazz, Capabilities stereotyp
...
+              LOG.warning("Driver file exists but does not seem to be a executable! " + webDriverExecutablePath);
+            }
+
+            // add specified driver executable path to capabilities
+            confStereotype = new PersistentCapabilities(confStereotype)
+              .setCapability("se:webDriverExecutable", webDriverExecutablePath);
+          }
+
+          Capabilities stereotype = enhanceStereotype(confStereotype);
+          String configName = thisConfig.getOrDefault("display-name", "Custom Slot Config");
```
sbabcoc, line 431: I'm pretty sure that there was no way to get to this point in the prior implementation without `display-name` being defined. The new implementation explicitly checks for the presence of this setting and bails out if it's absent. Was there a plan for a "custom slot config" feature? If so, I can remove the check for the setting and allow for this possibility. Perhaps the intent was to allow specification of the name in the stereotype?

java/src/org/openqa/selenium/grid/config/Config.java
diemol: Yes, it would not be needed.

java/src/org/openqa/selenium/grid/config/Config.java
```diff
@@ -59,14 +61,22 @@ default <X> X getClass(String section, String option, Class<X> typeOfClass, Stri
     }
   }
 
+  String DELIM_KEY = "\u001E";
+  String DELIMITER = DELIM_KEY + "=\"record-separator\"";
```
diemol, line 65: How do you use this in a toml file?

sbabcoc: Delimiters aren't specified in client requests; these are inserted by the API. Clients have the ability to define lists of maps, and the API flattens these into lists of key/value strings. The delimiters enable the API to determine where each group of associated key/value strings ends.
If clients ever call the `getAll()` method directly, they'll need to be aware of these break points as well, but I strongly doubt that anyone will ever want/need to do so.

sbabcoc: However, TOML actually supports Unicode character constants in the same format that Java and JSON use.

java/src/org/openqa/selenium/grid/config/Config.java
sbabcoc: Shall I remove the sorting, then? The only effects of this would be on the unit tests, which currently expect the list items to be sorted.

diemol: Sure, it can be removed.

java/src/org/openqa/selenium/grid/config/Config.java
diemol: Maybe I can understand better through an example. How would a TOML file look like with this delimiter?

java/src/org/openqa/selenium/grid/config/Config.java
sbabcoc: :heavy_check_mark: I removed the sorting and revised the affected unit tests. (**TomlConfigTest** didn't require any revisions, as the actual values returned by the **JToml** parser are sorted.)

java/src/org/openqa/selenium/grid/config/Config.java
sbabcoc:
```
[example]
"\u001E" = "record-separator"
```
Note that while this is possible in TOML, it would be a bad idea with this interface, as the API would treat this explicitly added separator as an end-of-record marker with predicably undesirable results.

The only record separators the API should find are the ones inserted by the `toEntryList()` method.

sbabcoc: Perhaps I should mark these DELIMITER constants as **INTERNAL**. Do we have an established pattern for marking and documenting such constants?

java/src/org/openqa/selenium/grid/node/config/NodeOptions.java
sbabcoc: If we determine that there's a use case for "custom slot config", adding support for this should be easy.

java/src/org/openqa/selenium/grid/config/Config.java
diemol: I still do not get it. Apologies. 

Can you share a real TOML file you have used to test this? I need help to understand how the delimiter looks like.

sbabcoc: The delimiter will not be used in TOML files. It's only used internally by the Config API to indicate the end of an associated group of settings. This eliminates reliance in the `display-name` setting for this purpose and the need to sort the settings list.

sbabcoc: I used an ASCII control code as the key name so it's highly unlikely that anyone will ever add a new setting that will be mistaken for this delimiter. The current strategy of using `display-name` as the delimiter requires that the list of key/value strings returned by `getAll()` must be sorted and is at risk for breakage if someone adds a new setting whose sort position precedes `display-name` (e.g. - `debug-port`).  
Delimiters are added by `Config.toEntryList()` and removed by `NodeOptions.addDriverConfigs()`. Their sole purpose is to mark the end of each group of driver configuration settings. No one will ever explicitly include these delimiters in their driver configuration files.

sbabcoc: @diemol: At this point, the only references to the delimiter (other than the methods mentioned above) are in the unit tests for the `getAll()` method.

diemol (approved):
Thank you, @sbabcoc!

diemol:
I tested this locally, no need to wait for the CI.

PR #12431, opened 2023-07-27
[dotnet] Fix error when we send non-base64 data for fetch command

nvborisenko:
### Description
I see a lot of exceptions while working with CDP. This PR fixes only one exception.

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist

- [ ] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #12445, opened 2023-07-30
[dotnet] Fix continueResponse method in CDP

nvborisenko:
Looks like this code was wrote by a mistake.

### Types of changes

### Checklist

### DIFF
--- a/dotnet/src/webdriver/DevTools/DevToolsSession.cs
+++ b/dotnet/src/webdriver/DevTools/DevToolsSession.cs
@@ -582,7 +582,10 @@ namespace OpenQA.Selenium.DevTools
                 var eventData = messageObject["params"];
 
                 LogTrace("Recieved Event {0}: {1}", method, eventData.ToString());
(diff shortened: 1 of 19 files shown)


titusfortner:
Is there a test we can add to show it failing/passing?

nvborisenko:
Most likely only for cdp v115 (as the latest, I didn't find tests for others).

titusfortner:
Yes yes, we only test the version that is on the CI...

nvborisenko:
No, it is supposed to be used by https://github.com/SeleniumHQ/selenium/blob/trunk/dotnet/src/webdriver/NetworkManager.cs which is not even covered by unit tests.

Given that BiDi is our primary goal, I have no motivation to create tests for features nobody is using (or usage of it is blocked by https://github.com/SeleniumHQ/selenium/pull/12431)

diemol (approved):
Thank you, @nvborisenko!

PR #12455, opened 2023-07-31
Build universal macOS Selenium-Manager on CI

p0deje:
Here is an example of a built binary - https://github.com/SeleniumHQ/selenium/actions/runs/5720210958. It's double-size as we'd expect.

```bash
$ lipo -archs ./selenium-manager
x86_64 arm64
$ du -sh selenium-manager
7.3M	selenium-manager
```


diemol (approved):
Seems like a good solution.

bonigarcia (approved):
LGTM, thanks!

PR #12463, opened 2023-08-01
[bazel] Get Python tests running on RBE CI

shs96c:


diemol:
The are already running on RBE
https://github.com/SeleniumHQ/selenium/actions/runs/8380065425/job/22948414365#step:16:2919

PR #12497, opened 2023-08-04
[dotnet] Small performance improvement for DriverFactory

jerone:
### Description
I noticed in https://github.com/SeleniumHQ/selenium/commit/962a34bfcfd6b5368f9f5df9e5e188b75771f891 that an `if`-statement was used, while probably an `else-if` should have been used, to skip all other conditions when the Chrome driver is used.

### Types of changes

### Checklist


titusfortner (approved):
Thanks!

PR #12501, opened 2023-08-06
[Grid] Add delete files method

### LINKED ISSUE #11986: [🚀 Feature]: Implement DELETE request that deletes all present files in a session in /session/{sessionId}/se/files in grid
### Feature and motivation

I understand that files get automatically deleted after a session but let's say I have a test case (which I really do) that does 10 file downloads and those files don't have unique names. Now with the existing implementation when a POST request is made to "/session/{sessionId}/se/files" it will fetch all the files again and there will be a discrepancy between what file I need and what file is being given to me by the endpoint. To avoid this, if a DELETE request is supported, after fetching each file we could delete it and not worry about what file is being provided to us because other files just got deleted. 

Code example below - 

### Usage example

```java
protected String getDownloadedFileByGrid( String expectedFile ) throws Exception {

		 String home = System.getProperty("user.home");
		 SessionId sessionId = ( ( RemoteWebDriver ) driver ).getSessionId();
		 File dirToCopy = new File(home + File.separator + "Downloads" + File.separator +  "gridnodes" + File.separator + sessionId);
		 String uri = String.format("/session/%s/se/files", sessionId );

		 String fileToDownload = null;

		 try ( HttpClient client = HttpClient.Factory.createDefault().createClient(new URL( "http://localhost:4444/" ) )) {
			
			 HttpRequest request = new HttpRequest( org.openqa.selenium.remote.http.HttpMethod.GET, uri);
			 String text = string(response);
			 Type responseType = new TypeToken<Map<String, Map<String, List<String>>>>() {
			 }.getType();
			 Map<String, Map<String, List<String>>> map = new Json().toType(text, responseType);
			 Map<String, List<String>> parsedResponse = map.get("value");
			 for (String eachFile : parsedResponse.get("names")) {
				 if( eachFile.contains( expectedFile ) ) {
					 fileToDownload = eachFile;
					 break;
				 }
			 }
		 }

		 try (HttpClient client = HttpClient.Factory.createDefault().createClient(new URL( "http://localhost:4444/" ) )) {

			 HttpRequest request = new HttpRequest( org.openqa.selenium.remote.http.HttpMethod.POST, uri);
			 request.setContent( Contents.asJson(singletonMap("name", fileToDownload)));
			 String text = string(response);
			 Type responseType = new TypeToken<Map<String, Map<String, String>>>() {
			 }.getType();

			 Map<String, Map<String, String>> map = new Json().toType(text, responseType);
			 Map<String, String> parsedResponse = map.get("value");
			 String encodedContents = parsedResponse.get("contents");
			 Zip.unzip(encodedContents, dirToCopy );
			 System.out.println("The file which was "
					 + "downloaded in the node is now available in the directory: "
					 + dirToCopy.getAbsolutePath());
		 }

		 return dirToCopy + File.separator + fileToDownload;
```

This is where the problem is -
```
for (String eachFile : parsedResponse.get("names")) {
				 if( eachFile.contains( expectedFile ) ) {
					 fileToDownload = eachFile;
					 break;
				 }
			 }
```

Let's say I download 10 files and they don't have unique names and I say give me file that contains ".xls" or ".pdf". First file we get is good but when It downloads the second file, the list of files from node is fetched again. It will check ok this first file contains ".xls" or ".pdf and it will return back the same file.

aldobrynin:
Implement delete files method: `DELETE /session/{sessionId}/se/files`

Fixes #11986

### Description
Implemented method to delete all files in current session's downloads directory

### Motivation and Context
Sometimes it is required to cleanup downloads directory. For example, in order to use pooled session developers need to be able to reset session state before next test run. Downloads directory is of the session features that one might want to cleanup during session lifecycle

### Types of changes

### Checklist

### DIFF
--- a/dotnet/test/common/Environment/DriverFactory.cs
+++ b/dotnet/test/common/Environment/DriverFactory.cs
@@ -46,7 +46,7 @@ namespace OpenQA.Selenium.Environment
                 browser = Browser.Chrome;
                 options = GetDriverOptions<ChromeOptions>(driverType, driverOptions);
             }
-            if (typeof(EdgeDriver).IsAssignableFrom(driverType))
+            else if (typeof(EdgeDriver).IsAssignableFrom(driverType))
             {
                 browser = Browser.Edge;
                 options = GetDriverOptions<EdgeOptions>(driverType, driverOptions);
--- a/java/src/org/openqa/selenium/grid/node/Node.java
+++ b/java/src/org/openqa/selenium/grid/node/Node.java
@@ -165,6 +165,9 @@ public abstract class Node implements HasReadyState, Routable {
             post("/session/{sessionId}/se/files")
                 .to(params -> new DownloadFile(this, sessionIdFrom(params)))
                 .with(spanDecorator("node.download_file")),
+            delete("/session/{sessionId}/se/files")
+                .to(params -> new DownloadFile(this, sessionIdFrom(params)))
+                .with(spanDecorator("node.download_file")),
             get("/se/grid/node/owner/{sessionId}")
                 .to(params -> new IsSessionOwner(this, sessionIdFrom(params)))
                 .with(spanDecorator("node.is_session_owner").andThen(requiresSecret)),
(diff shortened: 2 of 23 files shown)


karanjeetbirdeye:
If a file is downloaded and gets deleted from session, when the next downloaded file is present in session, it would give us only 1 file (last downloaded file) or all of them?

aldobrynin:
> If a file is downloaded and gets deleted from session, when the next downloaded file is present in session, it would give us only 1 file (last downloaded file) or all of them?

it would give us only 1 file

krmahadevan (approved):

krmahadevan:
@diemol - I have gone through the changes and they look good to me.

diemol (approved):
Thank you, @aldobrynin!

Thank you for reviewing, @krmahadevan!

PR #12502, opened 2023-08-06
[grid] no need for CDP version dependencies in the server

joerg1985:
### Description
This PR will remove the CDP version dependencies from the grid and only keep the devtools dependency.
The grid is only interested in the WebDriverInfo implementations of all the browsers, they do not rely on the concrete versions. If this would be the case the server version would be strong bound to the browser version. I think this is a not desired behavior and should be avoided in any case.

Or am i wrong and the CDP version dependencies are somewhere needed in the server?

### Motivation and Context
Reduces the size of the standalone server by ~3 mb

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/grid/BUILD.bazel
+++ b/java/src/org/openqa/selenium/grid/BUILD.bazel
@@ -154,9 +154,10 @@ java_export(
         "//java/src/org/openqa/selenium/grid/sessionmap/httpd",
         "//java/src/org/openqa/selenium/grid/sessionqueue/httpd",
         "//java/src/org/openqa/selenium/ie",
+        "//java/src/org/openqa/selenium/remote",
         "//java/src/org/openqa/selenium/safari",
         "//javascript/grid-ui:react_jar",
-    ] + CDP_DEPS,
+    ],
     deps = [
         ":base-command",
         "//java/src/org/openqa/selenium/cli",


krmahadevan:
@joerg1985 - Since the node is going to be responsible for spawning the actual browser, wouldn't this cause failures with respect to CDP functionality if the uber jar is removed off of it ?

joerg1985:
@krmahadevan thanks for the feedback.

These parts are not browser version dependend, as far as i can tell.
The `org.openqa.selenium.devtools.*`, `org.openqa.selenium.devtools.events.*`, `org.openqa.selenium.devtools.idealized.*` and `org.openqa.selenium.devtools.noop.*` used by the `DriverServiceSessionFactory`, `ProxyNodeWebsockets`, `ChromiumDriver` and `FirefoxDriver` will be still there in the uber jar.
The PR is adding a new runtime_dep to `"//java/src/org/openqa/selenium/devtools"` to ensure this.

I might have missed something, therefore i created this PR to discuss.

krmahadevan:
@joerg1985 - I may not be the best person to comment on this, but I was just curious

* My knowledge on bazel is very basic. When we specify  a runtime dependency, doesn't it mean that the jar will be expected to be in the CLASSPATH at runtime ? Something like `<provided>` in maven? If that's the case, then these classes will not be in the Grid's fat jar.
* So when a test case, makes use of a CDP related functionality but via RemoteWebDriver, then these classes would need to be present in the fat jar at the node end, else the functionality will not work.

For e.g., when a user does `RemoteWebDriver driver = new RemoteWebDriver(url, new ChromeOptions());` I think the call would go back to node, wherein we would perhaps be using reflection and invoke `new ChromeDriver()` which will call https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/chromium/ChromiumDriver.java#L160 which internally uses service loader to find the devtools version and then instantiate an implementation of `org.openqa.selenium.devtools.CdpInfo` and these implementations I think need these classes.

* If that's the case, then wouldnt this break the functionality ?

Does that sound valid?

joerg1985:
>     * My knowledge on bazel is very basic. When we specify  a runtime dependency, doesn't it mean that the jar will be expected to be in the CLASSPATH at runtime ? Something like `<provided>` in maven? If that's the case, then these classes will not be in the Grid's fat jar.

Bazel's `runtime_dep` are in the fat jar at the end, the PR replaces one `runtime_dep` with another, so this should be okay.


>     * So when a test case, makes use of a CDP related functionality but via RemoteWebDriver, then these classes would need to be present in the fat jar at the node end, else the functionality will not work.
> 
> 
> For e.g., when a user does `RemoteWebDriver driver = new RemoteWebDriver(url, new ChromeOptions());` I think the call would go back to node, wherein we would perhaps be using reflection and invoke `new ChromeDriver()` which will call https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/chromium/ChromiumDriver.java#L160 which internally uses service loader to find the devtools version and then instantiate an implementation of `org.openqa.selenium.devtools.CdpInfo` and these implementations I think need these classes.
> 
>     * If that's the case, then wouldnt this break the functionality ?

The grid is not working with reflection, it is more like a proxy to the webdriver process.
Therefore the grid only needs to be aware of the connections used, the CDP / BiDi is using websockets so they need to be exposed too. 

But let's see what others say to this PR.

diemol:
I think this should work. But I am not 100% sure. What do you think, @shs96c?

java/src/org/openqa/selenium/devtools/BUILD.bazel
```diff
@@ -65,6 +65,7 @@ java_library(
         ":mutation-listener",
     ],
     visibility = [
+        "//java/src/org/openqa/selenium/grid:__pkg__",
```
shs96c: This will be a problem, causing the classes in the `org.openqa.selenium.grid` package to not only be in the `org.seleniumhq.selenium:selenium-remote-driver` maven artifact, but also in `org.seleniumhq.selenium:selenium-grid` maven artifact. This is one of the reasons why we try and keep `visibility` as close to empty as possible :) 

If we really want to avoid having the CDP classes in the server, then we can create a new `java_export` target for `//java/src/org/openqa/selenium/devtools`, mark that as being publicly visible, and then have both the remote and grid packages depend upon it.

joerg1985: And how about just adding a `//java/src/org/openqa/selenium/remote` runtime_dep in `//java/src/org/openqa/selenium/grid`?
We have it in the server anyway due to transitive dependencies e.g. by the `//java/src/org/openqa/selenium/chrome`.

So there is no need to modify the visibility of `//java/src/org/openqa/selenium/devtools`?

joerg1985:
@shs96c I just update the PR to avoid the issue with the packages in multiple maven artefacts.

joerg1985:
@diemol i think this could be merged without side effects now or do you have concerns about this?

diemol (approved):
I just hadn't seen the PR before. Thanks, @joerg1985!

PR #12507, opened 2023-08-07
[java] PageSize width and height to match PrintOption default orientation.

HyoungUkJJang:
Change the values of PageSize.class width and height to match the default orientation in PrintOption.class.
image

### Description
* The picture below is part of the PrintOption.class. It is set to PORTRAIT by default.
image

* **However, the default constructor of the PageSize.class is set to the value LANDSCAPE.**
image

* the picture below shows the standard size of an A4 sheet on Google. I noticed that some default values for width and height in PageSize.class were wrong, so I fixed them to match the correct measurements.
* If the width and height values have been specified in detail for a specific reason, we will restore them to their original values. Thank you.
image

### Motivation and Context
* It seems necessary to make changes because using the default values of the Selenium library for A4-sized PDF output may confuse users.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/print/PageSize.java
+++ b/java/src/org/openqa/selenium/print/PageSize.java
@@ -24,8 +24,8 @@ public class PageSize {
 
   public PageSize() {
     // Initialize with defaults. A4 paper size defaults in cms.
-    this.height = 21.59;
-    this.width = 27.94;
+    this.height = 29.7;
+    this.width = 21;
   }
 
   public PageSize(double height, double width) {


titusfortner:
The bigger question is why we have defaults at all here, because the spec should be handling defaults. (I'm guessing because our Java code is allergic to anything being null). [Per the spec](https://w3c.github.io/webdriver/#dfn-print-page), if nothing is sent in, the defaults are:

> Let pageWidth be the result of [getting a property with default](https://w3c.github.io/webdriver/#dfn-getting-the-property-with-default) named width and with a default of 21.59 from page.

> Let pageHeight be the result of [getting a property with default](https://w3c.github.io/webdriver/#dfn-getting-the-property-with-default) named height and with a default of 27.94 from page.

So it looks like what we have got height/width backwards, which we should change. As for the difference between what is in the spec and A4, I think needs to be addressed [by w3c ](https://github.com/w3c/webdriver/issues)

This isn't the most used feature, but it would be cool to have a 3rd party library that wrapped this to allow users to change paper sizes by format, etc.

HyoungUkJJang:
Thank you for your kind response. It might have been better to submit this issue as a bug report. I appreciate your understanding as this is my first time providing feedback on an open-source project. 😄

Currently, it seems that not many users are utilizing this feature, and it doesn't appear to have critical bugs. I think it would be great if an update could be considered in the future!

I will close this Pull Request. It has been a positive experience. Thank you.

titusfortner:
If you want to create a pull request to change the defaults to what is in the spec I can raise the issue at our next meeting and make sure that's what we want to do

PR #12508, opened 2023-08-07
Allow external uri to be configurable for components that support server functionality - #12491

### LINKED ISSUE #12491: [🚀 Feature]: Make external URI configurable
### Feature
Add `--external-uri` configuration parameter to `server` section. This will allow nodes to provide a URI to hub (or distributor) which is route-able when SE node is running behind some sort proxy.

### Motivation
There are cases when nodes cannot report self URI that is route-able from hub during node registration process in which case the node get's stuck in `Sending registration event...` loop, while the hub logs will show connection timeouts or unknown host exceptions.

I believe this can be easily fixed in `org.openqa.selenium.grid.server.BaseServerOptions#getExternalUri` by first reading this newly added `external-uri` parameter. When empty, implementation can fall back to the current one otherwise it could just use the provided value.

### Our concrete use case:
We run ~5000 selenium tests. For performance reasons we deploy ~60 instances of our app each bundled with a SE node across ~20 single node k8s clusters (so each cluster runs 3 apps each with a chrome SE node). One extra k8s cluster runs on the CI agent itself where SE hub is running. For various reasons - which for brevity I don't want to detail - these clusters communicate with each other via node ports (on private IPs). 

So I end up in a situation where I can configure SE node so it can reach SE hub via (let's say 192.168.1.2:30027-30030) and I can configure it to report its own IP correctly (let's say 192.168.1.3, via `--host`), but I get into a problem with `--port`.
If I leave it on it's default (5555), then hub will try to connect to it on 192.168.1.3:5555 (but that is not route-able). One solution that works is picking a port from node port range and use that port both for node port and start SE node on that port inside the container but this has an unwanted administration overhead as well as adding non deterministic behavior as static node port range is only available on k8s version 1.27 in a feature that is not generally available yet.

It is a lot simpler if this feature is added.

### Usage example

Until docker part is updated we could use `SE_OPTS` to supply `--external-uri=http://<node-ip>:<node-port>`.

utamas:
SE nodes might be behind some sort of proxy exposed to hub on a different hostname(/ip) and/or port than component would by default report themselves (e.g.: hub and nodes are in different k8s clusters
 and services are exposed via node ports).

Fixes #12491

Relevant update to docs is at https://github.com/SeleniumHQ/seleniumhq.github.io/pull/1448

### Types of changes

### Checklist


diemol:
I left a comment on the issue. Have you checked it?

utamas:
@diemol yes (although, I haven't tagged you)

joerg1985 (commented):
I think the URI should be an URL, so external-url might be a better parameter. Currently i don't have access to a dev system for the next days, so i am not able to make a real review, sorry.

utamas:
@joerg1985  I'll make changes later today/tomorrow

titusfortner:
@utamas thanks for your work on this code. I'm trying to figure out what all we want to get into 4.14, does this have much more work to do?

utamas:
Hi @titusfortner I think a simple renaming is the only thing that needs to be done. I'm starting that now.

utamas:
@joerg1985 as I'm making my changes I can now see why I picked `external-uri`. The property I needed 
to make a slight adjustment is called: `org.openqa.selenium.grid.server.BaseServerOptions#getExternalUri`.

Despite this I think URL is a more accurate representation, hence made the changes.

utamas:
@titusfortner and @joerg1985  please let me know if anything else needs to get done.

titusfortner:
@utamas yeah, looks like we have some failures in our Remote Build Java tests:

```
Executing tests from //java/test/org/openqa/selenium/grid/server:BaseServerOptionsTest
-----------------------------------------------------------------------------
Failures: 1
1) externalUriFailsForNonUriStrings() (org.openqa.selenium.grid.server.BaseServerOptionsTest)
org.opentest4j.AssertionFailedError: [External URI must be parseable as URI.] 
expected: "Supplied external URI is invalid: Illegal character in path at index 3: not a uri"
 but was: "Supplied external URI is invalid: Illegal character in path at index 3: not a URL"
```

utamas:
@titusfortner my bad. Run the test after rebase but not after changing the name.

utamas:
@titusfortner how can I make sure this makes it into next release?

titusfortner:
@joerg1985 can you take a look at this one?

java/src/org/openqa/selenium/grid/server/BaseServerOptions.java
```diff
@@ -84,28 +84,48 @@ public int getMaxServerThreads() {
...
+        .get(SERVER_SECTION, "external-url")
+        .map(
+            url -> {
+              try {
+                return new URI(url);
+              } catch (URISyntaxException e) {
+                throw new RuntimeException(
+                    "Supplied external URI is invalid: " + e.getMessage(), e);
+              }
+            })
```
joerg1985, line 97: could be simplified to .map(URI::create)

joerg1985:
it look good to me, but i never approved a PR. so i am not totally sure what i have to check in this case.

diemol (approved):

utamas:
Thank you guys!

PR #12520, opened 2023-08-08
[py+ci] Get closer to getting all the python tests running on EngFlow

shs96c:
See what happens when we set the tags and data for each of the chrome and firefox tests. Should be a no-op for the existing test runs.


py/BUILD.bazel
```diff
@@ -14,6 +14,9 @@ compile_pip_requirements(
     name = "requirements",
```
captain-hoity-toity, line 14: It's not a bug, it's a... feature?

diemol:
The are already running on RBE
https://github.com/SeleniumHQ/selenium/actions/runs/8380065425/job/22948414365#step:16:2919

Issue #11695, opened 2023-02-20
[🚀 Feature]: Selenium Manager documentation


bonigarcia:
### Feature and motivation

If #11694 is done, Selenium Manager needs its own documentation.

### Usage example

n/a

bsquizz:
+1 I'd like to see some improvements related to how the binary is released and the documentation around that. I see several issues open here related to enhancing this and I'm not sure which ticket to place this upvote on... but let me highlight a problem we have recently run into:

As of Selenium version 4.6.0 it appears that the selenium-manager binary that is released with the python `selenium` package requires glibc 2.29. This is a problem for systems running RHEL8 which is based on glibc 2.28. If we had the opportunity to install the python package with a different selenium-manager binary that would help in this kind of situation.

It's not clear to me how/where this binary gets built and released so documenting how it appears in the python package would be helpful as well.

For the time being, I think our quick fix is to simply pin to selenium 4.5.0. It's not clear to me what the recommended approach would be to using a different 'selenium-manager' binary. I suppose I could compile it myself and then overwrite the one shipped with the python package, or update the PATH so that my custom selenium-manager is executed before the "packaged" one. I have to try some of these options out.

titusfortner:
I'm against releasing it separately, it would only be (more) confusing to Selenium users who already don't realize they don't need to download it. It's still beta and we're still working out bugs.

@bsquizz can you update to latest 4.8, that bug should have been fixed.

bsquizz:
Thanks, will give the latest 4.8 a try.

As far as releasing binaries within the python package goes, I can understand that releasing it as part of the package makes people's lives easier -- I am just not sure how I would "swap out" selenium-manager to use my own if it was necessary. It looks like the path is hard-coded, correct? https://github.com/SeleniumHQ/selenium/blob/trunk/py/selenium/webdriver/common/selenium_manager.py#L51

So at the very least I think an env var allowing users to change the binary location might be needed.

[...]

PR #12526, opened 2023-08-09
[rust] Unify browser_ttl and driver_ttl in a single config key

bonigarcia:
### Description
This PR unifies the two existing configuration keys in SM related to time-to-live (i.e., `driver-ttl` and `browser-ttl`) in a single one called `ttl`.

### Motivation and Context
As of PR [#12353](https://github.com/SeleniumHQ/selenium/pull/12353), `browser-ttl` refers to the browser version discovered by making network requests to CfT, not by finding browser versions by doing shell commands. As of this change, that value is 3600 seconds (1 hour).

As of PR [#12394](https://github.com/SeleniumHQ/selenium/pull/12394), `driver-ttl` is 1 hour (and not 1 day, as previously) since we detected that the rate change of chromedriver is quicker than it used to be.

All in all, `browser-ttl` and `driver-ttl` now refer to the same concept (i.e., the discovered versions using network requests, and their value is the same. Therefore, I propose to unify both configuration keys in a single one.

The context of this PR is the [documentation](https://github.com/SeleniumHQ/selenium/issues/11695), to simplify a bit the explanation of the concept of TTL on the Selenium Manager page.

### Types of changes

### Checklist

### DIFF
--- a/rust/README.md
+++ b/rust/README.md
@@ -43,10 +43,8 @@ Options:
           HTTP proxy for network connection (e.g., https://myproxy.net:8080)
       --timeout <TIMEOUT>
           Timeout for network requests (in seconds) [default: 300]
-      --driver-ttl <DRIVER_TTL>
-          Driver TTL (time-to-live) [default: 3600]
-      --browser-ttl <BROWSER_TTL>
-          Browser TTL (time-to-live) [default: 3600]
+      --ttl <TTL>
+          TTL (time-to-live) for discovered versions (online) of drivers and browsers [default: 3600]
       --cache-path <CACHE_PATH>
           Local folder used to store downloaded assets (drivers and browsers), local metadata, and configuration file [default: ~/.cache/selenium]
       --clear-cache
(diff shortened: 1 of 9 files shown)


titusfortner:
What did we find that needed to be checked once an hour to ensure up to date? 24 hours still seems more reasonable here. So long as the major versions match which I think the logic ensures already?

titusfortner (approved):

bonigarcia:
> What did we find that needed to be checked once an hour to ensure up to date?

Since the CfT endpoints are used to discover the chromedriver versions, I saw that the recommended chromedriver version changes more often than it used to. In normal conditions, twenty-four hours is a reasonable TTL. But 1 hour is safer. For instance, when the chromedriver team released the fix for chromedriver 115 on macOS, a TTL of 1 hour allows getting a newer version sooner (in the case the old one is cached in the metadata).

titusfortner:
I guess a couple extra connection requests every hour is pretty minor 😂

PR #12531, opened 2023-08-10
[rust] Rename metadata file to selenium-manager-metadata.json

bonigarcia:
### Description
This PR changes the metadata file handled by SM from `selenium-manager.json` to `selenium-manager-metadata.json`.

### Motivation and Context
In addition to the metadata file, the SM cache (`~/.cache/selenium`) can store another file, named `selenium-manager-config.toml`. When having these two files, the current metadata file name (`selenium-manager.json`) is not very expressive. Thus, I think `selenium-manager-metadata.json` is a bit better.

The context of this PR is the [documentation](https://github.com/SeleniumHQ/selenium/issues/11695).

### Types of changes

### Checklist

### DIFF
--- a/rust/src/metadata.rs
+++ b/rust/src/metadata.rs
@@ -25,7 +25,8 @@ use serde::{Deserialize, Serialize};
 
 use crate::Logger;
 
-const METADATA_FILE: &str = "selenium-manager.json";
+const METADATA_FILE_OLD: &str = "selenium-manager.json";
+const METADATA_FILE: &str = "selenium-manager-metadata.json";
 
 #[derive(Serialize, Deserialize)]
 pub struct Browser {
@@ -50,6 +51,10 @@ pub struct Metadata {
 }
 
 fn get_metadata_path(cache_path: PathBuf) -> PathBuf {
+    let old_metadata = cache_path.join(METADATA_FILE_OLD);
+    if old_metadata.exists() {
+        fs::remove_file(old_metadata).unwrap_or_default();
+    }
     cache_path.join(METADATA_FILE)
 }
 
--- a/rust/src/metadata.rs
+++ b/rust/src/metadata.rs
@@ -26,7 +26,7 @@ use serde::{Deserialize, Serialize};
 use crate::Logger;
 
 const METADATA_FILE_OLD: &str = "selenium-manager.json";
-const METADATA_FILE: &str = "selenium-manager-metadata.json";
+const METADATA_FILE: &str = "se-metadata.json";
 
 #[derive(Serialize, Deserialize)]
 pub struct Browser {


titusfortner:
I'd probably go the other way and just call it `config.yml`
```
~/.cache/selenium/config.yml
```

Further, would it make sense for this to be a toml file and merge it with what Grid uses?
Might be nice for Grid to automatically look in `~/.cache/selenium/config.toml` by default?

```
[server]
port = 4449

[sessionqueue]
session-request-timeout = 500

[manager]
proxy = "user@pass:myproxy:8080"
```

titusfortner:
and I didn't read your whole description before commenting. Are there 2 files, or just 2 options for files? What do the 2 different files store?

bonigarcia:
In the cache, in addition to downloaded drivers and browser, there are two files that SM reads:

- `selenium-manager-metadata.json` (former to this PR, called `selenium-manager.json`): This file contains the metadata, i.e., the versions discovered by SM making network requests, and the TTL in which they are valid. SM automatically maintains this file.
- `selenium-manager-config.toml` (optional): Configuration values for SM. This file is maintained by the end-user and read by SM.

All of this info will be adequately explained in the Selenium Manager doc page, which I am working on.

titusfortner:
My 2 thoughts are:

A file that is always in a Selenium directory doesn't need to include selenium in the name.

I think it would be nice to have one file for all selenium configuration and a default place for it to go. @diemol does that sound useful for grid?

titusfortner:
Sorry tapped the wrong button on my phone. 👍

diemol:
Shorter names are better since they are already inside the `selenium` directory. 

I would not mix Grid for now because SM is still beta, and most users do not really know about this directory.

bonigarcia:
> Shorter names are better since they are already inside the `selenium` directory.

What about `sm-metadata.json` and `sm-config.toml`? The second one is a small break change w.r.t. to the previous version (since some people can already be using `selenium-manager-config.toml`, but since we are on beta, I suppose it is not a big deal.

titusfortner:
I don't think there's an inherent reason for these to be limited to manager? There might be other things we want to put in there without having everything named differently? `.cache/selenium/config.toml` & `.cache/selenium/metadata.json`?

But I don't feel strongly about it, so if you think the full name is better...

bonigarcia:
Something in the middle could be `se-metadata.json` and `se-config.toml`. Also, the prefix `se-` is coherent with the prefix used for envs (`SE_`).

titusfortner:
Oh, because users can specify a different directory, then it makes sense to have a more descriptive file name...

Yes, I like se-*

titusfortner (approved):
I think change to se-* and go ahead and merge. I'm not worried about backwards compatibility yet. We'll highlight it in release notes.

rust/src/metadata.rs
```diff
@@ -25,7 +25,8 @@ use serde::{Deserialize, Serialize};
 
 use crate::Logger;
 
-const METADATA_FILE: &str = "selenium-manager.json";
+const METADATA_FILE_OLD: &str = "selenium-manager.json";
```
titusfortner, line 28: I think we should just change this or throw a warning saying it is changed and remove it next release.

bonigarcia: The metadata file is entirely transparent for the user since it is handled automatically by SM. This change does not support any impact on the use of SM, so we should not display any warning about it. A warning trace will alert to something that does not deserve the user's attention. On the other hand, yes, we can remove the logic for `METADATA_FILE_OLD` in the next release.

titusfortner: I'm confused if it isn't for backwards compatibility for the user, why do we need it for this release? 😄

bonigarcia: The motivation and context of this change is explained at the beginning, as usual.

titusfortner: I get the motivation for changing the name, I don't get the motivation for copying things over from the old file. Building a new metadata file for a new release seems minimal difficulty? I get there could end up being an old metadata file in the directory for current users, but if it isn't for users that doesn't seem like a big issue, either. Dunno, just trying to minimize the amount of code that we know will get removed later.

bonigarcia: Nothing is copied from the old file. Three lines of code check the existence of the old metadata file, and if it exists, it is deleted. It is a way to get the cache cleaned (i.e., without the former metadata file, which is not used anymore). As I said before, these three lines can be deleted in upcoming releases.

PR #12532, opened 2023-08-10
[JS] [atoms] add get-element-from-cache as a fragment

jlipps:
While potentially considered an implementation detail this was an essential fragment for Appium. Can we pretty please have it back?


titusfortner:
Still not super familiar with Bazel files or atoms. Does this need visibility by `chrome-driver`?

jlipps:
Yeah probably not. I just pushed again with nothing in the visibility array. It still seemed to build for my purposes.

PR #12533, opened 2023-08-10
[py] fixed safari tests

sandeepsuryaprasad:

### Description
Fixed Select class and Launcher tests for Safari.

### Motivation and Context
- Chrome raises `NotImplementedError` while trying to select item from a list box which is disabled. But Safari doesn't. I have marked all those tests as `xfail_safari`.
- Fixed Safari launcher test by passing `service` object to `WebDriver`
### Types of changes

### Checklist

### DIFF
--- a/py/test/selenium/webdriver/common/select_class_tests.py
+++ b/py/test/selenium/webdriver/common/select_class_tests.py
@@ -48,7 +48,7 @@ def test_select_by_index_single(driver, pages):
             assert sel.first_selected_option.text == select["values"][x]
 
 
-@pytest.mark.xfail_safari(reason="Unlike chrome, safari does not raise NotImplemented error")
+@pytest.mark.xfail_safari(reason="options incorrectly reported as enabled")
 def test_raises_exception_select_by_index_single_disabled(driver, pages):
     pages.load("formPage.html")
     sel = Select(driver.find_element(By.NAME, disabledSingleSelect["name"]))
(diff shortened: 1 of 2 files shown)


titusfortner:
What is raising `NotImplemented` errors? I'm wondering if that is the right behavior.

sandeepsuryaprasad:
@titusfortner , when we try to operate on a list box which is disabled, in chrome below exception is raised.

```
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/selenium/webdriver/support/select.py", line 98, in select_by_index
    self._set_selected(opt)
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/selenium/webdriver/support/select.py", line 213, in _set_selected
    raise NotImplementedError("You may not select a disabled option")
NotImplementedError: You may not select a disabled option
```
In case of Safari, no exception is raised, program execution completes without any exceptions or any messages. But yes, since the list box is disabled nothing gets selected. I think it is better to raise `ElementNotInteractableException` instead of `NotImplementedError`

titusfortner:
Oh, support classes are special, but we should know why it isn't getting that error before just skipping the tests

sandeepsuryaprasad:
My understanding is that the error should be raised by safaridriver itself and nothing to do with python bindings.

titusfortner:
These tests are all about describing intended vs actual behavior. The xfail message needs to reflect how actual behavior differs from intended.

This is the support module where we are working with opinionated behavior. The opinion we are enforcing is that if the user wants to select a disabled option with the Select class, it should throw an error instead of ignoring it. So we check the status and throw the error in the bindings.

Is Safari reporting that the element is not disabled when it is? Why isn't it getting the error that we want it to have?

sandeepsuryaprasad:
I was debugging this issue, and found that in function `_set_selected` in `select.py`,
```
    def _set_selected(self, option) -> None:
        if not option.is_selected():
            if not option.is_enabled():
                raise NotImplementedError("You may not select a disabled option")
            option.click()
```
`option.is_enabled()` returns `True` (Incorrect)for Safari and `False` for Chrome (Correct). So that's the reason why `NotImplementedError` is raised in case of Chrome and not in Safari

titusfortner:
So that needs to be the message in xfail, "options incorrectly reported as enabled"

sandeepsuryaprasad:
Yes, I have changed it and pushed commit.

PR #12534, opened 2023-08-10
[dotnet] options do not belong in the service class

titusfortner:
### Description
Follow Java approach and use Driver Finder in Driver constructor and not in the Service constructor.

### Motivation and Context
If a user creates the service instance independently it will use DriverFinder for the default options instead of specified options. Generally not an issue since users have no reason to create a Service class in .NET if they aren't going to specify the location of the driver. But users need to be able to toggle things like logging without having options. and let DriverFinder still set location based on actual options.

### DIFF
--- a/dotnet/src/webdriver/Chrome/ChromeDriverService.cs
+++ b/dotnet/src/webdriver/Chrome/ChromeDriverService.cs
@@ -16,6 +16,7 @@
 // limitations under the License.
 // </copyright>
 
+using System;
 using System.IO;
 using OpenQA.Selenium.Chromium;
 using OpenQA.Selenium.Internal;
@@ -46,7 +47,7 @@ namespace OpenQA.Selenium.Chrome
         /// <returns>A ChromeDriverService that implements default settings.</returns>
         public static ChromeDriverService CreateDefaultService()
         {
-            return CreateDefaultService(new ChromeOptions());
+            return new ChromeDriverService(null, null, PortUtilities.FindFreePort());
         }
 
         /// <summary>
@@ -54,6 +55,7 @@ namespace OpenQA.Selenium.Chrome
         /// </summary>
         /// /// <param name="options">Browser options used to find the correct ChromeDriver binary.</param>
         /// <returns>A ChromeDriverService that implements default settings.</returns>
+        [Obsolete("CreateDefaultService() now evaluates options in Driver constructor")]
         public static ChromeDriverService CreateDefaultService(ChromeOptions options)
         {
             string fullServicePath = DriverFinder.FullPath(options);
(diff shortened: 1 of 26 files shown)


titusfortner:
@nvborisenko any .NET conventions to fix in my code?

I kind of want to make all Driver location constructors in Driver classes obsolete since Selenium Manager should handle everything now (and there's the service class if it does not).

dotnet/test/common/CustomDriverConfigs/StableChannelEdgeDriver.cs
```diff
@@ -14,9 +14,13 @@ public StableChannelEdgeDriver(EdgeOptions options)
...
         }
 
+        public StableChannelEdgeDriver(EdgeDriverService service, EdgeOptions options)
+            : base(service, options)
+        {
+        }
         public static EdgeOptions DefaultOptions
         {
-            get { return new EdgeOptions(); }
+            get { return new EdgeOptions() { AcceptInsecureCertificates = true }; }
```
nvborisenko, line 23: I don't think we should set it by default. I am returning from vacation soon, will look through deeper.

titusfortner: I added it to match Chrome, and these are just the tests, so I'm not sure it matters either way. But if it doesn't matter we should remove it from both. Test should specify anything that doesn't match defaults.

dotnet/test/common/Environment/EnvironmentManager.cs
```diff
@@ -231,9 +236,15 @@ public void CloseCurrentDriver()
         {
             if (driver != null)
             {
-                driver.Quit();
+                try
```
nvborisenko, line 232: We are ignoring potential issues in `Quit()` method here.

dotnet/src/webdriver/Chrome/ChromeDriverService.cs
```diff
@@ -46,14 +47,15 @@ private ChromeDriverService(string executablePath, string executableFileName, in
         /// <returns>A ChromeDriverService that implements default settings.</returns>
         public static ChromeDriverService CreateDefaultService()
         {
-            return CreateDefaultService(new ChromeOptions());
+            return new ChromeDriverService(null, null, PortUtilities.FindFreePort());
```
nvborisenko, line 50: I don't think `null` is allowed as a service path. I quickly looked through further usages of it and, seems, `null` is not valid argument here: https://github.com/SeleniumHQ/selenium/blob/4eeed646fdb3ef31ad3f712f7ebd852b2ca849ee/dotnet/src/webdriver/DriverService.cs#L267

titusfortner: The idea of this is that if the user passed something in to use it. Otherwise we need to set it from the options. We only know both in driver constructor, so we always make sure it isn't null before it gets used.

dotnet/test/common/Environment/EnvironmentManager.cs
titusfortner: Yes, this needed to catch/ignore as well. Thanks.

Issue #5600, opened 2018-03-12
Chromedriver leaves forked Chrome instances hanging in background after test run (with large CPU load)


garrymann:
## Meta
OS:  
Microsoft Windows [Version 10.0.16299.248] x64
Selenium Version:  
2.53.1
Browser:  
Chrome 65 x64
Browser Version:  
Version 65.0.3325.146 (Official Build) (64-bit)
Additional rerequisites:
- ChromeDriver (2.36) 
- NodeJS (v8.10.0)
- cucumber: ^1.3.0
- protractor: ^5.1.1

## Expected Behavior -
2. The chromedriver.exe should stop and its Chrome instances as well.

## Actual Behavior -
2. The chromedriver.exe stops 4-5 secs after the tests completed. Then the forked Chrome instances stay in the background as running processes with huge CPU usage.

## Steps to reproduce -
1. Run at least 1 failing test with CucumberJS and Protractor. 
note: I'm running e2e tests against an Angular 1.6 application
2. Examine the processes in Task manager.

## Note:
This issue was introduced by the latest Chrome and ChromeDriver update. No symptomps were present in previous versions.

## Screenshot:
![Actual result](https://user-images.githubusercontent.com/17498617/37300130-8604fc2a-2625-11e8-8503-6cc57450ed6d.png)

cgoldberg:
> Run at least 1 failing test with CucumberJS and Protractor.

does your code always call quit() on the driver instance when a test fails? (usually done as a teardown/cleanup fixture in your framework)

garrymann:
@cgoldberg Hi!
The .quit() is handled by Protractor as far as I know. So it is not necessary since runner [teardownEnv ](https://github.com/angular/protractor/blob/35dcea01bf/lib/runner.js#L418) will take care of that for you.

barancev:
If you (or Protractor) called quit, but chromedriver could not stop the browser -- report this issue to chromedriver: https://sites.google.com/a/chromium.org/chromedriver/help

garrymann:
@barancev Done. Thanks.
https://bugs.chromium.org/p/chromedriver/issues/detail?id=2311

Things I've tried but did not resolve the issue:
reinstall chrome (v65 x64)
disable and remove extensions
run it without a single chrome instance open in the background
delete npm-cache
redo npm i
update to npm 5.7  (latest)
update to node 8 (latest LTS)
update windows 10: Version 10.0.16299.248

Any other ideas?

[...]

PR #12539, opened 2023-08-11
[java] Copy SM binary to cache folder and use it from there (#11359)

### LINKED ISSUE #11359: [🚀 Feature]: Cache Selenium Manager

bonigarcia:
### Description
This PR copies the SM binary from the distribution (jar) to the cache folder (`~/.cache/selenium), which is invoked from then.

This PR has two commits:

1. The first commit assumes that the cache path is always the path (`~/.cache/selenium), which makes things easier.
2. The "problem" is that the cache folder can be configured in Selenium Manager with other mechanisms: the configuration file (`se-config.toml` -if we finally renamed that way-) or environmental variables (in particular, `SE_CACHE_PATH` is the one for the cache path). The second commit allows to read also the cache path from the configuration file (first) and the env (second).

### Motivation and Context
Resolves #11359.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/manager/BUILD.bazel
+++ b/java/src/org/openqa/selenium/manager/BUILD.bazel
@@ -21,7 +21,6 @@ java_export(
         "//java/src/org/openqa/selenium/json",
         "//java/src/org/openqa/selenium/os",
         artifact("com.google.guava:guava"),
-        artifact("io.ous:jtoml"),
     ],
 )
 
--- a/java/src/org/openqa/selenium/manager/BUILD.bazel
+++ b/java/src/org/openqa/selenium/manager/BUILD.bazel
@@ -21,6 +21,7 @@ java_export(
         "//java/src/org/openqa/selenium/json",
         "//java/src/org/openqa/selenium/os",
         artifact("com.google.guava:guava"),
+        artifact("io.ous:jtoml"),
     ],
 )
 
--- a/java/src/org/openqa/selenium/manager/SeleniumManager.java
+++ b/java/src/org/openqa/selenium/manager/SeleniumManager.java
@@ -71,6 +71,7 @@ public class SeleniumManager {
   private final String managerPath = System.getenv("SE_MANAGER_PATH");
   private Path binary = managerPath == null ? null : Paths.get(managerPath);
   private String seleniumManagerVersion;
+  private boolean binaryInTemporalFolder = false;
 
   /** Wrapper for the Selenium Manager binary. */
   private SeleniumManager() {
@@ -78,6 +79,23 @@ public class SeleniumManager {
     String releaseLabel = info.getReleaseLabel();
     int lastDot = releaseLabel.lastIndexOf(".");
     seleniumManagerVersion = BETA_PREFIX + releaseLabel.substring(0, lastDot);
+    if (managerPath == null) {
+      Runtime.getRuntime()
+          .addShutdownHook(
+              new Thread(
+                  () -> {
+                    if (binaryInTemporalFolder && binary != null && Files.exists(binary)) {
+                      try {
+                        Files.delete(binary);
+                      } catch (IOException e) {
+                        LOG.warning(
+                            String.format(
+                                "%s deleting temporal file: %s",
+                                e.getClass().getSimpleName(), e.getMessage()));
+                      }
+                    }
+                  }));
+    }
   }
 
   public static SeleniumManager getInstance() {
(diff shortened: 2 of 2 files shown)


titusfortner:
I really don't want the bindings to process the toml file. The goal is to move the extra processing out of the bindings, not create more.

bonigarcia:
For that reason, I sent two commits in this PR. If the bindings do not want to process the config file (or the envs), the first commit should be enough.

diemol:
I agree, only env vars I would say. No TOML config read for now. 

Also, this class should give you the Selenium version https://github.com/SeleniumHQ/selenium/blob/43eb1e5477c9ef30bf6e2097ef984dec3f25077f/java/src/org/openqa/selenium/BuildInfo.java#L26

bonigarcia:
I have just sent two additional commits to this PR with the requested changes (read env but not config file and use of `BuildInfo` to get the Selenium version instead of a constant).

titusfortner:
The Java & Ruby implementations are slightly different and we need to make sure we have what we want, and then implement in the others.

Let's prioritize this as primary goal for 4.13

Andreito95:
@titusfortner hello, I just wanted to let you know that after the update to selenium new version 4.12 I no longer have the CPU resources usage problems that i reported in my pull request #12642 
 
 Hope this is helpful :)
 
 
image

titusfortner:
I have no idea what we did that fixed it, but thanks for letting us know.

Andreito95:
> I have no idea what we did that fixed it, but thanks for letting us know.

Hello, after 2 days of work...
Same processes, same environment (21 parallel/equal processes with 21 headless browsers)

My scripts quits and reopen drivers frequently, it may be something related to that or something else...
But version 4.11 was like this right after starting the processes, 4.12 is around 50/60% cpu for 1/2 days then goes 100%

image

diemol:
It seems that browsers are left open, which could be the consequence of a bug in your code. 

Can you please open a new issue and fill out all the information required in the template? With that, we can reproduce the issue.

Andreito95:
> It seems that browsers are left open, which could be the consequence of a bug in your code.
> 
> Can you please open a new issue and fill out all the information required in the template? With that, we can reproduce the issue.

hello and thanks for the answer
i am not so expert on github, i hope this is not the wrong place where we should talk about this

i won't open a a new request since if this is a bug i will study it... but... i use ,quit to close drivers...
it isn't enough? you have a better way to close drivers without leaving zombie processes?

Thanks

diemol:
We can only troubleshoot your issue if there is a way to reproduce it.

Andreito95:
@diemol @titusfortner 

Hello guys, after days of research i found the problem
Removing the option "no-sandbox" solved it, so actually the high CPU usage problems is caused by the option "no-sandbox"

Thank you for your time :)

Reference: https://github.com/SeleniumHQ/selenium/issues/5600
Comment: https://github.com/SeleniumHQ/selenium/issues/5600#issuecomment-383107970

image

titusfortner:
How does this work with when the cache location is not writeable? Do we need to fall back to streaming it?

bonigarcia:
Yes, that case was not considered, so it will fail when the cache is not writable in the current implementation. Let me review this PR to consider that case.

bonigarcia:
I've added a new patch to this PR to consider when the cache path is not writable. In that case, the Selenium Manager binary will be extracted in a temporal folder and used from there. That binary is not deleted using the original shutdown hook to avoid the problem related to grid (see #12802).

titusfortner:
Will that potentially result in a bunch of Selenium Manager binaries in temp files? (but only if cache directory isn't writeable)

bonigarcia:
No problem. I added a new commit to restore the shutdown hook to delete the SM binary only when it is in the temporal folder.

diemol (approved):

PR #12545, opened 2023-08-14
[ci] Run the RBE workflow on PRs and commits to trunk

shs96c:


shs96c:
This seems to be working as expected. Landing, and if we need to add a filter for PRs to forked repos, we can add that later.

PR #12547, opened 2023-08-14
[rb] copy selenium manager binary to cache and use from there

titusfortner:
### Description
* gets the minor selenium version from `Selenium::WebDriver::VERSION` to know what folder to create
* creates version folder in `~/.cache/selenium` to put selenium manager in
* validates selenium manager and uses from that location

### Motivation and Context
Implements #11359


titusfortner:
I think we've decided not to do this for Ruby. May revisit later.

PR #12554, opened 2023-08-15
[py] moved static method `get_path` to `SeleniumManager` class

sandeepsuryaprasad:

### Description
Moved the static method `get_path` from `DriverFinder` class in `common/driver_finder.py` to `SeleniumManager` class in `common/selenium_manager.py`.

### Motivation and Context
- Its not common/standard practice (at least in Python) to have one single static method in a separate class.
- Instead of having a single static method in a separate class definition, it is better to have it as a standalone function.
- In this PR, I have moved static method `get_path` to `SeleniumManager` class itself. So that `driver_finder.py` and `DriverFinder` class could be eliminated.
- Now all the methods related to SM binary and browser driver path are present in one single class, `SeleniumManager`.

### Types of changes

### Checklist


diemol:
I believe we want to leave it as it is because we want to have more things in DriverFinder. cc @titusfortner 

One thing I would appreciate is that an issue is raised before a PR like this is created. Changes like this should be discussed and therefore avoid unnecessary work. We are also in Slack if you want to discuss things you'd like to work on.

titusfortner:
I'm seeing some value in the Ruby side to making SeleniumManager class a lighter wrapper of the binary, and putting more of the driver finding logic into the DriverFinder class, like this - https://github.com/SeleniumHQ/selenium/pull/12429

Not sure if it is the right direction; we should discuss

isaulv:
While I would agree with the sentiment that having a class with one static method is better written as a function, I think it's totally fine if we want to namespace these functions into a class as opposed to a separate module. But either way is fine, and the class way gives use more flexibility in the future if we want to add state, or other functionality.

AutomatedTester (approved):

titusfortner:
I'm closing this in favor of #13387

PR #12557, opened 2023-08-15
[JS] [atoms] update fragment wrapper to apply window object to fragments

### LINKED ISSUE #12549: [🐛 Bug]: find-elements-fragment has an issue with JSON
### What happened?

I generated `find-element-fragment.js` and pasted it into a browser. One line (minified it looks like the only line in the fragment that includes `ba.JSON.stringify`) seems to host this bug: `ba` is undefined.

Reading the minified code as best I can, it looks like `ba` is either `document` or `window.navigator`. Neither of these host the `JSON` global (it is on `window`, I think).

Removing `ba.` from the line so that it references `JSON` as a pure global fixes the issue.

I'm not sure how to navigate the whole Google Closure Compiler situation. A quick grep of the codebase for a fixable line only showed up `goog.global.JSON`. Perhaps this issue is internal to the Compiler?

```shell
1. Generate the fragment with `bazel build //javascript/webdriver/atoms/inject/...`
2. Copy the contents of the built `find-element-fragment.js` file into a browser console
3. Invoke the pasted contents as an IIFE with some params (like `'css selector', 'body'`)
4. Observe the error
```

### Relevant log output

```shell
VM368:1631 Uncaught TypeError: Cannot read properties of undefined (reading 'stringify')
    at Object.<anonymous> (<anonymous>:1631:26)
    at findElement (<anonymous>:2633:8)
    at <anonymous>:1:1
```

### Operating System

macOS

### Selenium version

4.11.0

Chrome, Safari

N/A

N/A

jlipps:
fix #12549

went ahead and made a PR in case this works. All the atoms built with this change have passed Appium's mobile safari e2e test suite. It doesn't appear to break too many things. Let's see what Selenium CI says.

### DIFF
--- a/javascript/node/selenium-webdriver/test/lib/capabilities_test.js
+++ b/javascript/node/selenium-webdriver/test/lib/capabilities_test.js
@@ -212,7 +212,7 @@ test.suite(function (env) {
         await driver.switchTo().frame(frame)
         assert.strictEqual(
           await driver.findElement(By.css('body')).getText(),
-          LOREM_IPSUM_TEXT
+          fp.split('/').pop()
         )
 
         if (driver) {
--- a/javascript/private/fragment.bzl
+++ b/javascript/private/fragment.bzl
@@ -51,18 +51,11 @@ def closure_fragment(
 
     # Wrap the output in two functions. The outer function ensures the
     # compiled fragment never pollutes the global scope by using its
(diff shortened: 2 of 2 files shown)


AutomatedTester:
The CI failures seem to be fall out from test updates I did the other day.

AutomatedTester (approved):

Issue #12560, opened 2023-08-16
[🐛 Bug]: [rb] - SM doesn't play nicely with array of `capabilities`


luke-hill:
### What happened?

When using the SM instead of `webdrivers` gem, it does not play nicely when using a 2-itemed array of `capabilities`. This is still needed (I believe), because locally / remotely you may want to provide both options and capabilities.

Some examples are.... (From my gem): https://github.com/site-prism/automation_helpers/blob/main/lib/automation_helpers/drivers/v4/local.rb#L39-L51

### How can we reproduce the issue?

```shell
Generate a driver (I used capybara), whereby the capabilities is a 2-object array. This then passes `options` to SM as follows


[#<Selenium::WebDriver::Remote::Capabilities:0x00007ff68d454d50 @capabilities={}>, #<Selenium::WebDriver::Chrome::Options:0x00007ff68d408770 @options={:args=>[], :prefs=>{}, :emulation=>{}, :local_state=>{}, :exclude_switches=>[], :perf_logging_prefs=>{}, :window_types=>[], :browser_name=>"chrome"}, @profile=nil, @logging_prefs={}, @encoded_extensions=[], @extensions=[]>]
```
```


### Relevant log output

```shell
Solution is to add a small "shim" here on line 59 of `selenium_manager.rb`: https://github.com/SeleniumHQ/selenium/blob/trunk/rb/lib/selenium/webdriver/common/selenium_manager.rb#L59

        def generate_command(binary, options)
          options = options.last
```

Obviously this won't work for everyone, so maybe some detection logic is needed?
```


### Operating System

Ubuntu 20

### Selenium version

Ruby 4.11

### What are the browser(s) and version(s) where you see this issue?

chrome v115

### What are the browser driver(s) and version(s) where you see this issue?

v115 chrome

### Are you using Selenium Grid?

_No response_

diemol:
What about a PR that has the fix and a test?

titusfortner:
Ruby Selenium intentionally no longer supports using capabilities class in local drivers. It can still be used for remote drivers, but it is recommended to use Options instance.

PR #12561, opened 2023-08-16
[rb] - Possible fix for SM issue 12560

luke-hill:
### Description
This will check if the passed in options is directly a type of Options class or if it's an array it'll detect the item which is an Option class

### Motivation and Context
Currently there is an issue #12560 - However there may be a better way to solve this, or solving it further up the invocation chain. Happy to tweak if required

NB: My current code works with `webdrivers` gem, so this might be a use case for others.

### Types of changes

### Checklist


titusfortner:
We used to parse this in the DriverFinder and throw a warning that it wasn't supported. 
I don't think you are getting the capabilities you think you are getting at this point.

Issue #12273, opened 2023-06-28
[🚀 Feature]: Update .NET service classes


titusfortner:
### Feature and motivation

Java, Ruby & Python have all implemented various features for Service classes that can not be done in .NET right now.

- [x] Firefox can not be set to log to file ➡️ #16081 
- [x] No driver can be set to log to the console. ➡️ #16097
- [x] Chrome/Edge can not set log level (in Service class) ➡️ #16098
- [x] Chrome/Edge can not set `readableTimeStamp` ➡️ #16133
- [x] Firefox can not set truncated logs or profile root ➡️ #16131
- [x] ~~Safari can not enable logging~~ 

### Usage example

            var service = EdgeDriverService.CreateDefaultService();
            service.LogToConsole = true
            service.LogLevel = ChromiumDriverLogLevel.Debug 

            driver = new EdgeDriver(service);

diemol:
I struck out Safari logs because I did not see any output while trying it out.

diemol:
All PRs got merged. Closing this.

PR #12594, opened 2023-08-22
[dotnet] Hide standard drivers output

### LINKED ISSUE #11772: [🚀 Feature]: Add Option to FirefoxDriver to log geckodriver.exe output into a file or null in C#
### Feature and motivation

With an additional property like "LogfileName" it would be possible to put the marionette output into a file or away at all.

**firefoxService.LogfileName = "marionette.log";** or
**firefoxService.LogfileName = "null";**

Seleniumtest for Chrome, Edge and Firefox in Azure Devops run very fine. You also get a good logging while the test is running and at the end for all the scenarios.

**But:** Firefox logfiles are very inflated, because the geckodriver speaks to marionette, which itself controls the firefox browser session.

This inflated logfiles makes it difficult to see real problems, a lot is unimportant stuff like:
```
console.warn: SearchSettings: "get: No settings file exists, new profile?" (new Not ...
console.error: (new TypeError("NetworkError: Network request failed", "resource://services-s
```

### Usage example

It is possible to run geckodriver already before the Selenium tests are starting, like:
`geckodriver.exe --binary "C:\Program Files\Mozilla Firefox\firefox.exe" --log fatal >geckolog.txt`
And, as you see, you can redirect the output into a file.

**So if you put " > {{LogfileName}} at the end of the geckodriver.exe commandline**, it would work the same.
In the moment there is no way to do this (I checked the WebDriver.dll via JustDecompile)

```
FirefoxDriverService firefoxService = FirefoxDriverService.CreateDefaultService();
firefoxService.FirefoxBinaryPath = @"C:\Program Files\Mozilla Firefox\firefox.exe";
firefoxService.LogfileName = "geckolog.txt";
```

**So it looks for Firefox:**
Anmerkung 2023-03-14 211538

**And so clean for Chrome (and also Edge):**
Anmerkung 2023-03-14 212000

nvborisenko:
### Description
All process output from drivers comes to user's console. It's messing, especially when user starts drivers concurrently. Let's hide output, leaving a room for us to improve logging in .net (we can pop up all output from drivers to user later, optionally).

### Motivation and Context
Fixes #11772

### Types of changes

### Checklist

### DIFF
--- a/dotnet/src/webdriver/DriverService.cs
+++ b/dotnet/src/webdriver/DriverService.cs
@@ -268,6 +268,8 @@ namespace OpenQA.Selenium
             this.driverServiceProcess.StartInfo.Arguments = this.CommandLineArguments;
             this.driverServiceProcess.StartInfo.UseShellExecute = false;
             this.driverServiceProcess.StartInfo.CreateNoWindow = this.hideCommandPromptWindow;
+            // do not show process output, we can redirect it to logging mechanism when we will have it in place
+            this.driverServiceProcess.StartInfo.RedirectStandardOutput = true;
 
             DriverProcessStartingEventArgs eventArgs = new DriverProcessStartingEventArgs(this.driverServiceProcess.StartInfo);
             this.OnDriverProcessStarting(eventArgs);


titusfortner:
But then you *can't get output at all for Firefox, and this won't do anything: https://github.com/SeleniumHQ/selenium/blob/trunk/dotnet/src/webdriver/Firefox/FirefoxDriverService.cs#L193

I put it on my plate to work on this since we fixed things in the other bindings (#12273)

Essentially we want to do what we did in Java and allow users to direct the process output to where they want it — https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/remote/service/DriverService.java#L440-L472

But lots of things to add, because .NET doesn't support most of the logging parameters that other bindings do: 
 because .NET doesn't implement most of the logging things in most of the browsers
https://www.selenium.dev/documentation/webdriver/browsers/chrome/#log-output
https://www.selenium.dev/documentation/webdriver/browsers/firefox/#log-output

diemol:
So what is the plan here? Do you both work on #12273? Or will @titusfortner describe what needs to be done, and maybe @nvborisenko helps us?

titusfortner:
If @nvborisenko has time to help with it, would be so appreciated. .NET stuff should be so much less complicated than Java because we don't have to deal with System Properties. I have enough work and personal responsibilities over the next few weeks that I can't commit to much.

Looking at next release (https://github.com/SeleniumHQ/selenium/milestone/14). It is (probably) the last release before we remove support for Java 8, so I think the one essential thing I think we need is #11359 

Likely that will be the only thing I can commit to over the next month and everything else will get pushed back again.

nvborisenko:
I can commit to the following implementation:
- Driver process output will be completely hidden from user (Titus can continue his development regarding logging levels for driver binary, we are not intersected)
- `DriverService` class will provide a possibility to use to listen to driver output via C# events like
```csharp
driverService.OnStandardOutput =+ (sender, e) => File.Append("C:\\my_logs.txt", e.Data);
```

If this approach looks good to us (which is implemented widely in Selenium/C#) then I will implement it here in scope of this PR. Otherwise, I will close PR.

Issue #12576, opened 2023-08-19
[🐛 Bug]: Deadlock in Selenium Java bindings


asolntsev:
### What happened?

My Java test opens a browser and **hangs forever**. 

### How can we reproduce the issue?

```shell
I managed to create a simple reproducible example: https://github.com/asolntsev/selenium-deadlock

When running `mvn test`, it hangs immediately.

A sample thread dump is stored [here](https://github.com/asolntsev/selenium-deadlock/blob/main/thread-dump.txt).
It contains two threads blocking each other: "main" #1 and "CDP Connection" #57.
```

### Relevant log output

```shell
https://github.com/asolntsev/selenium-deadlock/blob/main/thread-dump.txt
```

### Operating System

MacOS (no matter)

### Selenium version

4.11.0 - 4.9.0 (probably older versions too)

### What are the browser(s) and version(s) where you see this issue?

Chrome 116, Edge 115 (versions don't really matter)

### What are the browser driver(s) and version(s) where you see this issue?

automatically downloaded by SeleniumManager

### Are you using Selenium Grid?

no

joerg1985:
I had a look at this and i could not find the root cause for this on the first look.
But i have done a fix to the code, the listeners where added and called multiple times for each event.

https://github.com/SeleniumHQ/selenium/commit/4641ff189d96d285428459f14fd0694069b37f6e

diemol:
@asolntsev does this fix the issue? You can test it with the Nightly release. https://www.selenium.dev/downloads/#nightly

asolntsev:
@diemol @joerg1985 No, it doesn't help. I updated my sample project to run on latest Selenium 4.12.0-SNAPSHOT:
https://github.com/asolntsev/selenium-deadlock

Just clone it and run `mvn test`. The deadlock still happens. :(

joerg1985:
Sorry i did not post an update here... i had to undo the change as the CI was failing.
I had a closer look at the implementation and there are multiple things to fix:
1. `NetworkInterceptor.close` does set a flag in `Network`, all following uses of `NetworkInterceptor` will not longer log errors because of this.
2. `NetworkInterceptor.prepareToInterceptTraffic` does register the callbacks multiple times.
     The events are processed multiple times and raise a lot of warning logs.
3. The `Network` will block the thread handling the request until the response is comming to replace the response.
    This is not needed we should be able to instantly send fullfilRequest instead.
PS:
[...]

PR #12603, opened 2023-08-23
[java] NetworkInterceptor related fixes

joerg1985:
### Description / Motivation and Context
Will fix the issues described in https://github.com/SeleniumHQ/selenium/issues/12576#issuecomment-1690208820 and will most likely fix the issue #12576. The interceptor will now instantly fullfil the request without waiting while holding the read lock. 

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/devtools/idealized/Network.java
+++ b/java/src/org/openqa/selenium/devtools/idealized/Network.java
@@ -19,7 +19,6 @@ package org.openqa.selenium.devtools.idealized;
 
 import static java.net.HttpURLConnection.HTTP_OK;
 import static java.nio.charset.StandardCharsets.UTF_8;
-import static java.util.logging.Level.WARNING;
 
 import java.io.ByteArrayInputStream;
 import java.io.InputStream;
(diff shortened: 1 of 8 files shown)


java/src/org/openqa/selenium/devtools/idealized/Network.java
```diff
@@ -62,7 +58,7 @@ public abstract class Network<AUTHREQUIRED, REQUESTPAUSED> {
   private volatile Filter filter = defaultFilter;
   protected final DevTools devTools;
 
-  private final AtomicBoolean networkInterceptorClosed = new AtomicBoolean();
+  private final AtomicBoolean perpared = new AtomicBoolean();
```
diemol:
```suggestion
  private final AtomicBoolean prepared = new AtomicBoolean();
```

diemol (commented):
To be honest, I've not touched this code so the whole flow change is not straight forward for me. 

Could you please walk me briefly through the changes? Especially in `Network.java`. An alternative is to add a test to verify this? Open to ideas.

java/src/org/openqa/selenium/devtools/idealized/Network.java
joerg1985: thanks for the hint

joerg1985:
There are tests for the `NetworkInterceptor`, but some were disabled due to issues with GitHub Actions.
So i re enabled them and lets see what happens, they might work now.

If they are still failing, i will try to fix them first and then guide through the changes.

joerg1985:
After looking into the disabled unit tests is see some hidden features of the `NetworkInterceptor` this PR would break.
Closing this for now

Issue #12207, opened 2023-06-15
[🚀 Feature]: Upload nightly python builds


bollwyvl:
### Feature and motivation

Expand the nightly CI to upload built python `.tar.gz` and `.whl` artifacts _somewhere_.

Downstreams would better be able to react to breaking changes before they are released.

The alternative, of every downstream properly configuring `bazel` and supporting to tools seems untenable for the general case.

### Usage example

Given:
- `pip`-compatible python packages [aren't coming](https://github.com/github/roadmap/issues/94) to GitHub any time soon
- selenium's nightly CI job includes an `actions/upload-artifact` for the python distributions
- a python downstream such as [SeleniumLibrary](https://github.com/robotframework/SeleniumLibrary) wants to test against the latest selenium

During downstream CI:
- a job would would find, download, extract, and cache the artifacts
  - using either the GitHub API or a [third-party tool](https://nightly.link)
- a job would install the cached artifacts and run the test suite
  - in the case of a fail, an issue in the downstream's issue tracker would be created or updated, with a pointer to the build logs
  - the downstream issue would be resolved by a downstream PR such as:
    - fixing a newly-introduced incompatibility to keep the current compatibility window 
    - pinning to the current version of `selenium`
  - the downstream would release a compatible version before the next selenium release 

End users:
- would `pip install -U {downstream}` and get a compatible version of selenium

bollwyvl:
Not sure precisely how to proceed: perhaps `workflows/bazel.yml` could offer a general-purpose `archive-path` (and optional `archive-name`) input. This might look something like this:

```diff
diff --git a/.github/workflows/bazel.yml b/.github/workflows/bazel.yml
index 27a64df592..aa3b4b92c5 100644
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -41,6 +41,16 @@ on:
         required: false
         type: string
         default: ''
+      archive-name:
+        description: Name of an artifact to upload
+        required: false
+        default: ''
+        type: string
+      archive-path:
+        description: Paths to upload as archive
+        required: false
+        default: ''
+        type: string
 
 jobs:
   bazel:
@@ -115,6 +125,12 @@ jobs:
           edge-version: ${{ inputs.browser-version || 'stable' }}
       - name: Run Bazel
         run: ${{ inputs.run }}
+      - name: Upload Artifacts
+        if: inputs.archive-path
+        uses: actions/upload-artifact@v3
+        with:
[...]

PR #12605, opened 2023-08-24
[py] Add python nightly #12207

dev-velo:

### Description

attempts to accomplish #12207 
### Types of changes

### Checklist


.github/workflows/bazel.yml
```diff
@@ -119,8 +123,38 @@ jobs:
...
+          pip install --upgrade twine
+          echo "[distutils]" > ~/.pypirc
+          echo "index-servers =" >> ~/.pypirc
+          echo "    testpypi" >> ~/.pypirc
+          echo "[testpypi]" >> ~/.pypirc
+          echo "username:$PYPI_TOKEN" >> ~/.pypirc
+          echo "password:$PYPI_PASS" >> ~/.pypirc
+        env:
+          PYPI_TOKEN: ${{ secrets.PYPI_TOKEN }}
+          PYPI_PASS: ${{ secrets.PYPI_PASS }}
```
dev-velo: This requires someone to create 2 github secrets to be used for twine uploading

.github/workflows/bazel.yml
```diff
@@ -119,8 +123,38 @@ jobs:
...
       - name: Run Bazel
         run: ${{ inputs.run }}
+      - name: Upload Python Artifacts
```
dev-velo: Whoops, this needs to be changed to use the twine upload command. a sample upload command would be great

dev-velo:
think this is now ready to be checked, made some small changes

.github/workflows/bazel.yml
```diff
@@ -119,8 +128,43 @@ jobs:
...
+          echo "NIGHTLY_VERSION=$NIGHTLY_VERSION" >> $GITHUB_ENV
+      - name: Update Version
+        if: inputs.language-release-setup == 'python'
+        run: |
+          ./scripts/github-actions/update-version.sh ${{ github.env.NIGHTLY_VERSION }}
+      - name: Install Python
+        if: inputs.language-release-setup == 'python'
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.9'
```
diemol: Can we bump this in the same way it is done for Ruby? 
https://github.com/SeleniumHQ/selenium/blob/dd10ff8f89df8ee1dd02664f0832042d7f50458f/.github/workflows/nightly.yml#L36C1-L39

.github/workflows/bazel.yml
diemol: This is particular to the Python Nightly. Could we have this in that workflow file? It is just a few commands that need to be run.

.github/workflows/bazel.yml
```diff
@@ -119,8 +128,43 @@ jobs:
...
+          echo "password:$PYPI_PASS" >> ~/.pypirc
+        env:
+          PYPI_TOKEN: ${{ secrets.PYPI_TOKEN }}
+          PYPI_PASS: ${{ secrets.PYPI_PASS }}
       - name: Run Bazel
         run: ${{ inputs.run }}
+      - name: Upload Python Artifacts
+        if: inputs.language-release-setup == 'python'
+        run: |
+            twine upload bazel-bin/py/selenium-${{ github.env.NIGHTLY_VERSION }}-py3-none-any.whl bazel-bin/py/selenium-${{ github.env.NIGHTLY_VERSION }}.tar.gz
```
diemol: This can go in the Nightly release flow.

.github/workflows/bazel.yml
```diff
@@ -41,12 +41,21 @@ on:
...
         description: Files to upload to the Nightly release tah
         required: false
         type: string
         default: ''
-
+      get-nightly-version:
+        description: Whether to get the nightly version
+        required: false
+        type: boolean
+        default: false
```
diemol: If we do it like Ruby (see below), I don't think this is needed.

.github/workflows/nightly.yml
```diff
@@ -60,3 +60,17 @@ jobs:
...
       nightly-release-files: build/dist/*.*
+
+  python:
+    if: inputs.language == 'python' || github.event_name == 'schedule'
+    name: Python
+    uses: ./.github/workflows/bazel.yml
+    with:
+      name: Release
+      cache-key: python-nightly
+      run: bazel build --stamp //py:selenium-wheel //py:selenium-sdist
```
diemol: I think all commands placed in the `bazel.yml` can go here.

diemol (changes requested):
Thank you for this, left some comments.

dev-velo:
@diemol 
Made some changes in the workflow, no longer deriving the github tag and just using the version thats in the local version.py. Did this to match what Ruby and Java do. A good next step would be to automate all version bumps and not just do it for one binding. Leaving that as a future task.
The only portion I can't test locally is the twine upload but I believe I followed the steps [right from here on twine upload](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#create-an-account) correctly. 

Steps for local testing
1. edit scripts/github-actions/update-python-version.sh, comment out line 45. This has to be done since its only applicable to github terminals.
```#echo "NIGHTLY_VERSION=$NIGHTLY_VERSION" >> $GITHUB_ENV```
2. give perms ```chmod +x scripts/github-actions/update-python-version.sh```
3. run the script ```./scripts/github-actions/update-python-version.sh```
4. Check changed files to verify version updated correctlly 
```
py/BUILD.bazel
py/docs/source/index.rst
py/version.py
```
5. bazel build ```bazel build --stamp //py:selenium-wheel //py:selenium-sdist```
this should output the proper nightly artifacts
Screenshot 2023-08-26 at 1 02 25 AM

File: .github/workflows/bazel.yml (hunk 1 above)
Comment (dev-velo) [reply .github/workflows/bazel.yml]: done

File: .github/workflows/bazel.yml (hunk 3 above)
Comment (dev-velo) [reply .github/workflows/bazel.yml]: removed the extra stuff of getting tags from git to simplify, did try to get to work with a bazel function but the main issue I ran into was ruby never referenced the version in the build file which python did multiple times. Not well versed in bazel either which made it more difficult lol I assume it can be done but I would defer it to someone more familiar with bazel

File: py/version.py [hunk 7]
```diff
@@ -0,0 +1 @@
+SE_VERSION = "4.11.2"
```
captain-hoity-toity, line 1: This method feels like it had five cups of coffee.

dev-velo:
@titusfortner apologies for the delay, finally was able to get a new laptop so I picked the work back up. moved the logic out of the python code and now handle it all at the scripting side of things. Ran the tests on my fork https://github.com/RevealOscar/selenium/actions/runs/6990764802/job/19020609283 and the only things that fail are some unrelated integration tests. Everything else passed. This is from the summary

//py:test-remote-test/selenium/webdriver/common/api_example_tests.py     FAILED in 3 out of 3 in 22.6s
//py:test-remote-test/selenium/webdriver/common/window_tests.py          FAILED in 3 out of 3 in 35.6s
Executed 50 out of 50 tests: 48 tests pass and 2 fail locally.

titusfortner:
Oh man, I really liked where you were going before this update.

1. Can we still consolidate the python versions into one place instead of creating yet another place to pull from? 😂 
2. Can we do this logic in Python? We should be able to do like Ruby is doing and straight execute a file.

Ruby is pulling the version dynamically here: https://github.com/SeleniumHQ/selenium/blob/selenium-4.15.0/rb/lib/selenium/support/nightly_version_generator.rb#L33

Python should actually be less complicated because Ruby is creating nightly versions of selenium-webdriver and selenium-devtools

titusfortner:
Also, welcome back!

titusfortner:
This was the original nightly one for Ruby — https://github.com/SeleniumHQ/selenium/commit/538c11af6d87bf09e7d06d1441b04c64a3b4aaa2

We can help get the bazel piece sorted out if we have the right logic in place.

dev-velo:
@titusfortner oh I agree, it was a better approach to consolidating files but Im not the most versed in python lol I can give it another go. @mdmintz would you mind helping? Ran into an issue of \_\_name\_\_ is not in globals'. A nightly selenium artifact could help downstream frameworks like seleniumbase

mdmintz:
@RevealOscar, can you be more specific as to what you’re looking for? SeleniumBase does run several checks to see if the latest selenium version is fully working before releasing a new SB version with the newer selenium version pinned in the dependencies.

dev-velo:
@mdmintz can you provide me with some pointers on how to go about consolidating the various version files selenium has in python to point to 1 source of truth? The goal is to produce a nightly python artifact. I have a separate branch https://github.com/SeleniumHQ/selenium/compare/trunk...RevealOscar:selenium:python_nightly_consolidation where I just need help with correcting the import statements. Dont really program at all in python, It seems using relative paths isnt really a thing across packages, and i havent fully understood [sys.path imports](https://stackoverflow.com/questions/30669474/beyond-top-level-package-error-in-relative-import). I only ask since I have seen you mention the help a selenium nightly artifact could provide on slack

mdmintz:
@RevealOscar Things work quite a bit differently in SeleniumBase, even though it uses Selenium Python. I’m probably not the one you’re looking for for this task.

titusfortner:
>  Im not the most versed in python

Ah, no worries, let me see what I can help with.

.github/workflows/nightly.yml
```diff
@@ -59,3 +59,35 @@ jobs:
...
+          PYPI_TOKEN: ${{ secrets.PYPI_TOKEN }}
+          PYPI_PASS: ${{ secrets.PYPI_PASS }}
+      - name: Update Version
+        run: |
+          chmod +x scripts/github-actions/update-python-version.sh
+          ./scripts/github-actions/update-python-version.sh
+      - name: Build and Upload
+        run: |
+          bazel build --stamp //py:selenium-wheel //py:selenium-sdist
+          twine upload bazel-bin/py/selenium-$NIGHTLY_VERSION-py3-none-any.whl bazel-bin/py/selenium-$NIGHTLY_VERSION.tar.gz
```
diemol, line 93: Don't we need to point to `https://test.pypi.org/project/selenium/` when we do this upload?

dev-velo: Followed the steps outlined in the wiki https://github.com/SeleniumHQ/selenium/wiki/Releasing-Selenium#python which totally could be out of date

diemol:
Thank you for your efforts, @RevealOscar! Thanks to the scripts @titusfortner did to automate releases, I completed this work and got some inspiration from this PR.

We are now [publishing](https://github.com/SeleniumHQ/selenium/actions/runs/7949128770/job/21699936354#step:16:110) nightly builds to [TestPyPi](https://test.pypi.org/project/selenium/).

The Selenium [Downloads](https://www.selenium.dev/downloads/#nightly) page has also been updated.

PR #12608, opened 2023-08-24
[dotnet] Simplfy to only target .NET Standard 2.0

jimevans:
### Description
Simplifies the .NET bindings to produce assembly artifacts only targeting .NET Standard 2.0, instead of multi-targeting many different versions of the .NET Frameworks.

### Motivation and Context
Rather than targeting multiple individual target framework versions, simplify the project targeting to only target .NET Standard 2.0. Doins so, the resulting assemblies will be usable by projects targeting any of the following versions:

* .NET Framework: 4.6.1, 4.6.2, 4.7, 4.7.1, 4.7.2, 4.8, 4.8.1
* .NET Core: 2.0, 2.1, 2.2, 3.0, 3.1
* .NET versions: 5.0, 6.0, 7.0

Additionally, they will be usable from the Mono and Xamarin frameworks.

This will simplify the build process and maintenance going forward, and will allow the project to greatly reduce the complexity of the Bazel builds.

As a note, this does remove support for .NET 4.5.x, the last version of which (4.5.2) was released in 2014, and reached end-of-life in April 2022.

### Types of changes

### Checklist


titusfortner:
So this effectively removes users of net45
Which I think is deprecated?

I'm a little delete happy these days, so I'm not going to complain, just wanted to verify.

Also, I had someone I trust say:
> net472 (fixes an issue where you support both .NET FX and .NET Standard 2.0)

Is this something we should include as well?

Regardless, thank you! This will make building things much easier at least 😂

nvborisenko:
This PR makes everybody's live so much easier. Two hands up for being it merged.

diemol (approved):
🎉

Issue #11536, opened 2023-01-11
[🐛 Bug]: Multiple Async/Fetch Errors Being Thrown when Trying to Transform Network Requests and Add Headers


ascendantofrain:
### What happened?

When trying to set up network interception using the NetworkRequestHandler, I am receiving multiple different errors being thrown as a website is loaded.

Errors include but are not limited to:
```System.InvalidOperationException: A command response was not received: Fetch.getResponseBody```
```System.InvalidOperationException: There is already one outstanding 'SendAsync' call for this WebSocket instance. ReceiveAsync and SendAsync can be called simultaneously, but at most one outstanding operation for each of them is allowed at the same time.```

### How can we reproduce the issue?

```shell
var requestHandler = new NetworkRequestHandler
{
    RequestMatcher = request => true,
    RequestTransformer = req =>
    {
        if (!req.Headers.ContainsKey("xxx"))
            req.Headers.Add("xxx", "xxx");
                
        return req;
    }
};

var driver = new ChromeDriver();

var networkInterceptor = driver.Manage().Network;
networkInterceptor.AddRequestHandler(requestHandler);
        
await networkInterceptor.StartMonitoring();

// load a site and do stuff

await networkInterceptor.StopMonitoring()
```

### Relevant log output

```shell
System.InvalidOperationException: A command response was not received: Fetch.continueRequest
   at OpenQA.Selenium.DevTools.DevToolsSession.<SendCommand>d__35.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at OpenQA.Selenium.DevTools.DevToolsSession.<SendCommand>d__34`2.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at OpenQA.Selenium.DevTools.V108.Fetch.FetchAdapter.<ContinueRequest>d__16.MoveNext()
[...]

Issue #10054, opened 2021-11-18
[🐛 Bug]: Selenium 4 WebSocket exception during performing basic authorization C#


lopukhDA:
### What happened?

Hello, currently we are under investigation of the possibility to use Selenium 4 on our project and we are faced with the next issue:
During performing the basic authorization, authorization is passing successfully but at the same time we have the next exception

`There is already one outstanding 'SendAsync' call for this WebSocket instance. ReceiveAsync and SendAsync can be called simultaneously, but at most one outstanding operation for each of them is allowed at the same time.`

Could you please take a look at this issue?

### How can we reproduce the issue?

```shell
static void Main(string[] args)
        {
            var options = new ChromeOptions();

            var driver = new ChromeDriver(ChromeDriverService.CreateDefaultService(), options, TimeSpan.FromMinutes(2));


            NetworkAuthenticationHandler handler = new NetworkAuthenticationHandler()
            {
                UriMatcher = d => d.Host.Contains("jigsaw.w3.org"),
                Credentials = new PasswordCredentials("guest", "guest")
            };

            var networkInterceptor = driver.Manage().Network;
            networkInterceptor.AddAuthenticationHandler(handler);
            networkInterceptor.StartMonitoring();

            driver.Navigate().GoToUrl("https://jigsaw.w3.org/HTTP/Basic/");

            networkInterceptor.StopMonitoring();
            driver.Quit();
        }
```

### Relevant log output

```shell
Starting ChromeDriver 95.0.4638.69 (6a1600ed572fedecd573b6c2b90a22fe6392a410-refs/branch-heads/4638@{#984}) on port 61026
Only local connections are allowed.
Please see https://chromedriver.chromium.org/security-considerations for suggestions on keeping ChromeDriver safe.
ChromeDriver was started successfully.

DevTools listening on ws://127.0.0.1:61029/devtools/browser/046fbea2-493a-4b0e-89a4-e0390b253469
[15732:36144:1118/170820.587:ERROR:chrome_browser_main_extra_parts_metrics.cc(230)] crbug.com/1216328: Checking Bluetooth availability started. Please report if there is no report that this ends.
[15732:37544:1118/170820.587:ERROR:device_event_log_impl.cc(214)] [17:08:20.587] USB: usb_device_handle_win.cc:1048 Failed to read descriptor from node connection: A device attached to the system is not functioning. (0x1F)
[...]

PR #12614, opened 2023-08-25
[dotnet] Refactor WebSocket communication for BiDi

jimevans:
### Description
Replaces the existing WebSocket communication mechanism with one more robust. Rather than immediately relying on event handlers to react to events and command reponses, it writes the incoming data to a queue which is read from a different thread. This eliminates the issue where the user might have multiple simultaneous sends or receives to the WebSocket while their event handler is running. It also dispatches incoming events on different threads for the same reason. This should eliminate at least some of the issues surrounding socket communication with bidirectional features, whether implemented using CDP or the WebDriver BiDi protocol.

### Motivation and Context
In the .NET WebSocket client implementation, one can have simultaneous sends and receives, but only one of each type. This can lead to unexpected errors for users attempting to use BiDI-enabled features like network interception, where it's common to respond to an event using an event handler, and issuing another command from within the event handler body. Specific instance of a reported issue this should solve is #11536 and #10054.

### Types of changes

### Checklist


dotnet/src/webdriver/DevTools/DevToolsSession.cs
```diff
@@ -437,116 +413,38 @@ private void OnTargetDetached(object sender, TargetDetachedEventArgs e)
...
-            }
+            LogTrace("Creating WebSocket");
+            this.connection = new WebSocketConnection();
+            connection.DataReceived += OnConnectionDataReceived;
+            await connection.Start(this.websocketAddress);
+            LogTrace("WebSocket created; starting message listener");
         }
 
-        private void CancelReceiveTask()
+        private async void TerminateSocketConnection()
```
nvborisenko: Avoid `async void`, better is `async Task`.

dotnet/src/webdriver/DevTools/DevToolsSession.cs
```diff
@@ -437,116 +413,38 @@ private void OnTargetDetached(object sender, TargetDetachedEventArgs e)
...
-            {
-            }
-            catch (WebSocketException)
-            {
-            }
-            finally
-            {
-                this.sessionSocket.Dispose();
-                this.sessionSocket = null;
+                await Task.Delay(TimeSpan.FromMilliseconds(10));
```
nvborisenko, line 442: Seems algorithm is contiguously pinging something instead being event based.

dotnet/src/webdriver/DevTools/DevToolsSession.cs
```diff
@@ -437,116 +413,38 @@ private void OnTargetDetached(object sender, TargetDetachedEventArgs e)
...
             }
+
+            this.messageQueue.Writer.Complete();
+            this.messageQueueMonitorTask.Wait();
```
nvborisenko: `Wait` may introduce potential deadlock. Better `GetAwaiter().GetResult()` or even `await`.

dotnet/src/webdriver/DevTools/DevToolsSession.cs
```diff
@@ -46,14 +47,15 @@ public class DevToolsSession : IDevToolsSession
         private bool isDisposed = false;
         private string attachedTargetId;
 
-        private ClientWebSocket sessionSocket;
+        private WebSocketConnection connection;
         private ConcurrentDictionary<long, DevToolsCommandData> pendingCommands = new ConcurrentDictionary<long, DevToolsCommandData>();
+        private readonly Channel<string> messageQueue;
```
nvborisenko: Seems `System.Threading.Channels` package is needed only for queue. I think we can use any other queue built in BCL, and remove external dependency.

dotnet/test/common/NetworkInterceptionTests.cs
```diff
@@ -0,0 +1,36 @@
...
+                INetwork network = driver.Manage().Network;
+                NetworkResponseHandler handler = new NetworkResponseHandler();
+                handler.ResponseMatcher = (responseData) => responseData.Url.Contains("simpleTest.html");
+                handler.ResponseTransformer = (responseData) =>
+                {
+                    responseData.Body = "<html><body><p>I intercepted you</p></body></html>";
+                    return responseData;
+                };
+                network.AddResponseHandler(handler);
+                network.StartMonitoring();
```
nvborisenko: async method not awaited

dotnet/src/webdriver/DevTools/DevToolsSession.cs
jimevans: Of course, you’re right here. Thanks for catching the oversight.

dotnet/src/webdriver/DevTools/DevToolsSession.cs
jimevans: Given how and when this code is called, I believe the chances of deadlock are quite minimal, but sure, this doesn’t really fit the idiomatic pattern, so I’m happy to change it. Using `await` here is not the best choice because we want to retain the synchronous API (and yes, I’ve read all the literature about not mixing sync/async APIs;, that ship has sailed as the library API is largely synchronous by design).

dotnet/src/webdriver/DevTools/DevToolsSession.cs
jimevans: @nvborisenko Polling for a channel reader to be empty is the only way to ensure that no messages sent before the WebSocket connection was closed are lost. Since the socket connection is always closed before this method is called, the channel reader is guaranteed to be empty at some point. The comment before the loop attempts to explain this but is unclear about it. I'll update the comment.

dotnet/test/common/NetworkInterceptionTests.cs
jimevans: Thanks for the catch.

dotnet/src/webdriver/DevTools/DevToolsSession.cs
jimevans: `Channel<T>` provides a far cleaner API for the type of producer/consumer pattern that we are enabling here, due to its "blocking read until data available" API, but I'll look into a change using a `ConcurrentQueue<T>` instead.

nvborisenko: I also like [BlockingCollection](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.blockingcollection-1?view=net-7.0) with very easy way of collection consumption:
```
foreach (var item in bc.GetConsumingEnumerable())
{
    Console.WriteLine($"Consuming: {item}");
}
```

jimevans: Far less performant than `ConcurrentQueue<T>`, but probably not an issue given where the bottlenecks might be for this process. I'll explore that too.

nvborisenko (approved):
Seems we don't need to put binaries to `third_party/dotnet/nuget`, please double check.

Looks good to me according current level of communication understanding. Thanks @jimevans for refactoring it.

jimevans:
@nvborisenko One was a new addition, which I forgot to add after removing the Channel-based implementation. The other was an update of a version of one already added. I've reverted that version upgrade. I believe we are ready to go with this one.

PR #12637, opened 2023-08-30
Closing out logging on quit

Sean-Gomez:

### Description

The line I added closes out the log file connections before the service performs a full shutdown

### Motivation and Context

A project I'm working on uses the Selenium webdriver and I noticed unclosed sockets remained in some of our unit tests, which could potentially affect performance. This addition should now properly close the open sockets of the log output file process.

### Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)

### Checklist

### DIFF
--- a/py/selenium/webdriver/common/service.py
+++ b/py/selenium/webdriver/common/service.py
@@ -23,9 +23,9 @@ import warnings
 from abc import ABC
 from abc import abstractmethod
 from platform import system
-from subprocess import DEVNULL
 from subprocess import PIPE
 from time import sleep
+from typing import TextIO
 from urllib import request
 from urllib.error import URLError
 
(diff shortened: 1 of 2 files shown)


titusfortner (changes requested):
The biggest problem is that not all `log_output` instances respond to `close()`. It can be used to set STDOUT or STDERR which are actually `int`

What if we move it to the Service class's `close()` method and put in a conditional so that it works for log file, stdout or null?

Sean-Gomez:
Screenshot from 2023-08-30 13-49-30

Hi Titus, I have attached a screenshot of the modified quit function with a conditional to close log_output whether it's an int or a file-like object. I'm not sure what you meant by move it to the service class's close() function

Sean-Gomez:
Hi Titus, my latest commit should close log_output if it's set to an int like you said. Sorry, but I'm not sure what you meant when you said move it to the server class' close method, could you clarify that if there's still an issue with the commit? Thanks!

titusfortner:
oops, sorry, I meant the `stop()` method: https://github.com/Sean-Gomez/selenium/blob/ed632c22efccab6abb88de25622d9f40a1de389c/py/selenium/webdriver/common/service.py#L142

This logic would apply to all the service classes, not just Firefox.

titusfortner:
Ok, I moved your solution to the stop method. There was already logic there, but I don't think it was doing the right thing.

@isaulv can you double check us on this change?

I think the issue is that devnull needs to be closed and we were ignoring it before. Also not sure what previous code was rescuing; should I keep that just in case?

isaulv:
@titusfortner I think the proposed code is correct. It will close open files (TextIO) and if it's a file descriptor like stdout or stdin, os.close should handle that file descriptor, and even dev/null. 
It shouldn't throw if the resource is already closed.

isaulv (approved):
LGTM

titusfortner:
Thanks @Sean-Gomez

PR #12641, opened 2023-08-30
[js] do not need browser version if Selenium is managing it

titusfortner:
JS fix equivalent to #12639

### DIFF
--- a/javascript/node/selenium-webdriver/common/seleniumManager.js
+++ b/javascript/node/selenium-webdriver/common/seleniumManager.js
@@ -70,7 +70,8 @@ function driverLocation(options) {
   let args = ['--browser', options.getBrowserName(), '--output', 'json']
 
   if (options.getBrowserVersion() && options.getBrowserVersion() !== '') {
-    args.push('--browser-version', options.getBrowserVersion())
+    args.push('--browser-version', options.getBrowserVersion());
+    options.setBrowserVersion("");
   }
 
   const vendorOptions =


javascript/node/selenium-webdriver/common/seleniumManager.js
```diff
@@ -70,7 +70,8 @@ function driverLocation(options) {
   let args = ['--browser', options.getBrowserName(), '--output', 'json']
 
   if (options.getBrowserVersion() && options.getBrowserVersion() !== '') {
-    args.push('--browser-version', options.getBrowserVersion())
+    args.push('--browser-version', options.getBrowserVersion());
+    options.setBrowserVersion("");
```
diemol, line 74: GeckoDriver still rejects an empty browser version. Have you checked if the key is still sent?

harsha509: Hi @titusfortner , @diemol ,

Looking into the implementation of other bindings,  it appears that we should nullify the browser version once we have determined the driver path right?

If yes below should work!

```
  // Once driverPath is available, set browserVersion to null
  if (output.result.driver_path) {
    options.setBrowserVersion(null);
  }
```

harsha509:
Updated the above with commit https://github.com/SeleniumHQ/selenium/commit/1c036abf7975819f9cf5449f26494162edaa09c7!

closing this PR as suggested by @titusfortner !

titusfortner:
Thanks @harsha509 !

Issue #12659, opened 2023-09-01
[🐛 Bug]: selenium seems to override _, breaking another package


stefanoborini:
### What happened?

I am facing a very difficult debugging situation, and I'll gather as many information as I can.

We are using a package called shinyMatrix, together with an R shiny application. Everything is quite dynamic, with matrices widgets generated on the fly. We know that shinyMatrix uses, under the hood, Vue.

What I see is that if I run the test under selenium, and only under selenium, I get the following error in Chrome console

![image](https://github.com/SeleniumHQ/selenium/assets/132967/771fa5a7-2fba-4e05-96b6-6bdf22670ae4)

After some debugging and watch of the _ variable, I found that the following snippet of code seem to remap _ to a different function

![image](https://github.com/SeleniumHQ/selenium/assets/132967/7b40d5eb-5609-485f-af03-a6002b6bc71c)

It's quite hard to say what's going on, because it's all minified. I was able to trace down that file to verbatim code part of the ruby bindings (I am using the python bindings):

```
grep -r "allowfullscreen allowpaymentrequest" *
rb/lib/selenium/webdriver/atoms/getAttribute.js:<SNIP> {"class":"className",readonly:"readOnly"},dd="allowfullscreen allowpaymentrequest allowusermedia async autofocus autoplay checked compact complete controls declare default defaultchecked defaultselected defer disabled ended formnovalidate hidden indeterminate iscontenteditable ismap itemscope loop multiple muted nohref nomodule noresize noshade novalidate nowrap open paused playsinline pubdate readonly required reversed scoped seamless seeking selected truespeed typemustmatch willvalidate".split(" ");ba("_",function(a,b){var<SNIP>
```

The resulting effect of the overwrite is that this function in shinymatrix is no longer able to find the lodash utility function cloneDeep:

![image](https://github.com/SeleniumHQ/selenium/assets/132967/13f3dfad-19e8-4c0c-98a9-15acf15e70da)

I don't know if this is a selenium problem, a Vue problem, a shiny problem, or a shinyMatrix problem. For now the most likely culprit seems to be selenium, but I might have to investigate further if you have more information about what's going on.

### How can we reproduce the issue?

Hard to reproduce without my full code, but I am using a very simple script:

```
import time
from selenium import webdriver
from uidriver.widgets.shiny import Selector, Tabset

options = webdriver.ChromeOptions()
driver = webdriver.Chrome(options=options)
driver.implicitly_wait(2)
[...]

PR #12662, opened 2023-09-01
use property instead of atom to get index of option element

titusfortner:
### Description
I've said I don't want to update code in the support module, but this was changed in Ruby & Java, but not Python, .NET or JS.

### Motivation and Context
1. Much more performant
2. Probably fixes whatever is going on in #12659

### Types of changes

### DIFF
--- a/dotnet/src/support/UI/SelectElement.cs
+++ b/dotnet/src/support/UI/SelectElement.cs
@@ -247,7 +247,7 @@ namespace OpenQA.Selenium.Support.UI
 
             foreach (IWebElement option in this.Options)
             {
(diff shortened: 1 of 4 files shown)


nvborisenko (approved):

titusfortner:
Fun fact, the existing JavaScript code does not allow you to select by index with a number, only with a string.

titusfortner:
This should also change `multiple` to use getDomAttribute instead of getAttribute. I'm putting into draft. Low priority

titusfortner:
This didn't solve the issue I thought it did, and I don't really want to deal with it.

PR #12667, opened 2023-09-02
[dotnet] Cache atoms

nvborisenko:
### Description
Cache atoms instead of reading it from resources.

### Motivation and Context
Performance improvement.

### Types of changes

### Checklist

### DIFF
--- a/dotnet/src/webdriver/WebElement.cs
+++ b/dotnet/src/webdriver/WebElement.cs
@@ -17,6 +17,7 @@
 // </copyright>
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


nvborisenko:
This is topic for discussion. From one point of view it's not a big deal to read atoms from resources (used rarely), but from another point view why not to cache it in memory (~100KB).

jimevans:
This feels like a micro-optimization to me without a lot of benefit. Reading a resource is pretty fast, and I'm not sure it's worth the tradeoff of constantly keeping these in memory. I'm not dead-set against it, but I'm not sure we need this particular change.

nvborisenko:
I was looking for opinion, I have have the same. Thanks.

Issue #12225, opened 2023-06-19
[🐛 Bug]: Performance issue in case of performance logs enabled in chrome


sparsh00:
### What happened?

After the changes made in https://github.com/SeleniumHQ/selenium/pull/11973 jar spawning is not adding the `--log-path=` flag to Chrome in the normal flow. Need to enable the `readableTimestamp` or `appendLog` flag to get it working   https://github.com/SeleniumHQ/selenium/pull/11973/files#diff-405c68a6b743ea8f4b340b5ea0c92c802a24109b4c38818335f6ca39511e2bc8R385.
What is has caused is that the chromedriver logs are append to file while going through jar causing performance issues.

### How can we reproduce the issue?

```shell
Run session on Chrome while adding logging params and opening heavy websites like Pinterest
"goog:loggingPrefs": {
  "driver": "ALL",
  "browser": "SEVERE",
  "performance": "ALL",
  "server": "ALL"
}
```

### Relevant log output

```shell
No direct logging of chromedriver flags in selenium logs found by looking around after facing performance issue.
```

### Operating System

Windows 10

### Selenium version

4.9.1

### What are the browser(s) and version(s) where you see this issue?

Chrome latest

### What are the browser driver(s) and version(s) where you see this issue?

Chromedriver related to the chrome version

titusfortner:
What is happening is that the output of the entire process is being sent to the file requested. Everything going to the console is put into the log file. 
Appending to File and Readable Time Stamp do not work with console output, so if either of those features are requested, we use the driver output instead of porting the console output to the file.

I only evaluated this with parameters being sent to start the driver, not parameters being sent to the driver to start the browser, which shouldn't be affected, but I'll take a look.

titusfortner:
Sure enough, the `goog:loggingPrefs` value requires the driver to have logging turned on, and the current behavior doesn't allow that without using `--log-path` to driver.

### Selenium 4.9.0
* `--log-path=FILE` turns on logging, respects the log levels in loggingPrefs map, sends it to that file
* `--log-level=LEVEL` without log-path turns on logging, overrides the log levels in loggingPrefs map, sends it to the console

### Selenium 4.10.0
* `--log-path=FILE` turns on logging, does not respect anything in loggingPrefs map
* `--log-level=LEVEL` without log-path still has default logging set to dev/null and does not respect anything in loggingPrefs map

### Options:
[...]

PR #12674, opened 2023-09-03
[java] parse log output to support streams and file location in system properties

titusfortner:
### Description
The `DriverService.build()` order steps are:
1.  Process System Properties
2. Create args
3. Create Service instance
4. Set process output

Changes by above step:
1. (except Safari) parse (in superclass) the property for log file location; ensures either `logFile` or `logOutput` is set
2. (Chromium) if `logFile` is set use `--log-path` since required for performance logging; also unsets `logFile` for Step 4
3. (Safari) sets `logOutput` to `LOG_NULL` since step 1 not applicable; (doesn't affect actual logging anyway)
4. Process streams output based on `logFile` or `logOutput`

### Motivation and Context
* The fix to #12225 broke setting `LOG_STDOUT`, `LOG_NULL` and `LOG_STDERR` in System Property
* There was duplication in how system properties were processed for log file

### DIFF
--- a/java/src/org/openqa/selenium/remote/service/DriverService.java
+++ b/java/src/org/openqa/selenium/remote/service/DriverService.java
@@ -437,37 +437,33 @@ public class DriverService implements Closeable {
       return DEFAULT_TIMEOUT;
     }
 
-    protected OutputStream getLogOutput(String logProperty) {
-      if (logOutputStream != null) {
-        return logOutputStream;
-      }
-
+    protected OutputStream getLogOutput() {
       try {
-        File logFileLocation = getLogFile();
-        String logLocation;
-
-        if (logFileLocation == null) {
-          logLocation = System.getProperty(logProperty);
-        } else {
-          logLocation = logFileLocation.getAbsolutePath();
-        }
-
-        if (logLocation == null) {
-          return ByteStreams.nullOutputStream();
-        }
-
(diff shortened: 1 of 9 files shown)


java/src/org/openqa/selenium/remote/service/DriverService.java
```diff
@@ -437,40 +437,32 @@ protected Duration getDefaultTimeout() {
...
-          case LOG_STDOUT:
-            return System.out;
-          case LOG_STDERR:
-            return System.err;
-          case LOG_NULL:
-            return ByteStreams.nullOutputStream();
-          default:
-            return new FileOutputStream(logLocation);
-        }
+        return logFile != null ? new FileOutputStream(logFile) : logOutputStream;
```
diemol: nit: Use `getLogFile` to get the value.

diemol (approved):
Thanks!

titusfortner:
I moved it to draft when I saw I needed to fix the switch statement (no breaks) and to swap the file and output precedence. Fixed now.

java/src/org/openqa/selenium/remote/service/DriverService.java
titusfortner: I guess I "fixed" this because I need to toggle off output stream not file. 😂

diemol (approved):

PR #12684, opened 2023-09-04
[java] replaced the EventAttribute(Value) with an AttributeMap

joerg1985:
### Description
Replaced the EventAttribute & EventAttributeValue with an AttributeMap interface.

### Motivation and Context
- makes the code better readable, there are no `EventAttribute.setValue` calls needed anymore
- a little less overhead for tracing `EventAttributeValue` instances are gone
- a little less overhead when tracing is disabled, the `AttributeMap` is not filled when disabled

### Types of changes

### Checklist


pujagani (approved):
I have pulled the changes, build it, and tried the whole observability flow. It works as expected. @joerg1985 Thank you for the optimization and the changes :)

PR #12700, opened 2023-09-07
[java] Fix NewSession Runner

KrishnaSuravarapu:

### Description

We have started consuming sessionrequests in batches with https://github.com/SeleniumHQ/selenium/pull/10987. Since we get a list of request and not a single one, while loop is not at all needed. Infact, having while loop would unnecessarily cause a retry in some race conditions (when batchsize < available Nodes). This anyway is ran via new session executor, so retries will happen after a specified interval

### Types of changes

### Checklist

### DIFF
--- a/dotnet/src/webdriver/DevTools/DevToolsSession.cs
+++ b/dotnet/src/webdriver/DevTools/DevToolsSession.cs
@@ -66,14 +66,13 @@ namespace OpenQA.Selenium.DevTools
                 throw new ArgumentNullException(nameof(endpointAddress));
             }
 
-            this.CommandTimeout = TimeSpan.FromSeconds(5);
+            this.CommandTimeout = TimeSpan.FromSeconds(30);
             this.debuggerEndpoint = endpointAddress;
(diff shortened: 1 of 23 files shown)


diemol (commented):
Can you explain better the race condition you mention? Otherwise, I do not see why we should change this.

Also, if the user has configured a long interval between queue processing, their overall testing time might slowdown.

KrishnaSuravarapu:
Lets say i can run 8 sessions (max-sessions) on chrome, firefox and edge. I've started 10 sessions on chrome. For the first iteration of while loop, it'll fetch first 8 sessions and try to start those in threads. In the next iteration (Since the stereotypes are still not empty), if the first set of sessions are not yet started, the stereotype map would be recreated and would try to start next 2 sessions which are bound to fail (since we can only run 8 sessions and the next 2 sessions will fail and will be readded to queue). Agreed on the interval part for queue processing. The documentation itself says this right `If all slots are busy, new session request will be retried after the given interval.`

diemol (approved):
👍

PR #12704, opened 2023-09-07
Update fragment.bzl to fix a bug where "_" was overwritten

### LINKED ISSUE #12659: [🐛 Bug]: selenium seems to override _, breaking another package

mdmintz:
Update fragment.bzl to fix a bug where ``_`` was overwritten

This resolves https://github.com/SeleniumHQ/selenium/issues/12659

This reverts https://github.com/SeleniumHQ/selenium/pull/12557

### Types of changes

### Checklist

### DIFF
--- a/javascript/private/fragment.bzl
+++ b/javascript/private/fragment.bzl
@@ -51,7 +51,12 @@ def closure_fragment(
 
     # Wrap the output in two functions. The outer function ensures the
     # compiled fragment never pollutes the global scope by using its
-    # own scope on each invocation.
+    # own scope on each invocation. We must import window.navigator into
+    # this unique scope since Closure's goog.userAgent package assumes
+    # navigator and document are defined on goog.global. Normally, this
+    # would be window, but we are explicitly defining the fragment so that
+    # goog.global is _not_ window.
+    #     See http://code.google.com/p/selenium/issues/detail?id=1333
     wrapper = (
         "function(){" +
         "return (function(){%output%; return this._.apply(null,arguments);}).apply({" +
--- a/javascript/private/fragment.bzl
+++ b/javascript/private/fragment.bzl
@@ -54,8 +54,10 @@ def closure_fragment(
     # own scope on each invocation.
     wrapper = (
         "function(){" +
-        "return (function(){%output%; return this._.apply(null,arguments);}).apply(" +
-        "window, arguments);}"
+        "return (function(){%output%; return this._.apply(null,arguments);}).apply({" +
+        "navigator:typeof window!='undefined'?window.navigator:null," +
+        "document:typeof window!='undefined'?window.document:null" +
+        "}, arguments);}"
     )
 
     browser_defs = {


titusfortner (changes requested):
We should keep the text that was deleted as well. Can you run git revert on it?

asolntsev (approved):

Jabbar2010:
Hi, @mdmintz @titusfortner @jarib @KazuCocoa there is an error on my side,
image
 I found the root cause is from atoms, as this PR, you just add navigator and document from window, but no JSON
image
image
Besides above, I also found some other apis not passed into the inner function, such as window.Range...
So do you know how to resolve my issue? and you can take a look at my another linked issue: https://github.com/appium/appium-remote-debugger/issues/383, I also have the "_" override because of this.Is there a good way to fix it?

Jabbar2010:
Sorry, updated: https://github.com/appium/appium-remote-debugger/issues/383#issuecomment-2165253340

PR #12705, opened 2023-09-07
Update visibility_tests.py to verify that "_" is not overwritten when calling "is_displayed()"

mdmintz:
Update visibility_tests.py to verify that ``_`` is not overwritten when calling ``is_displayed()``.

This is a test for https://github.com/SeleniumHQ/selenium/pull/12704, but in a separate PR to follow TDD principles. Currently, there is a bug in `4.12.0` (https://github.com/SeleniumHQ/selenium/issues/12659), and this test should catch that (and currently fail because of it). Once https://github.com/SeleniumHQ/selenium/pull/12704 is merged, this test should pass.

### Types of changes

- [X] Add a test (TDD-style)

### Checklist

- [X] I have added a test.

### DIFF
--- a/py/test/selenium/webdriver/common/visibility_tests.py
+++ b/py/test/selenium/webdriver/common/visibility_tests.py
@@ -127,6 +127,17 @@ def test_should_say_element_with_zero_transform_is_visible(driver, pages):
     assert zero_tranform.is_displayed() is True
 
 
+def test_checking_for_visibility_does_not_overwrite_underscore(driver, pages):
+    pages.load("cssTransform.html")
+    driver.execute_script("_ = 'abc123';")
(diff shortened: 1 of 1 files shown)


titusfortner (changes requested):
I don't believe in running integration tests that verify something unusual doesn't happen just because there was once a bug. Testing for a generalized problem in JS unit tests would be preferred, but I think that would require a better understanding of what we want to do there and how to do it correctly and how to avoid doing it incorrectly, which I'm not sure we know.

isaulv (approved):

diemol:
The bug was fixed, but this test is failing. I am not sure what the intention was. I will close this for now and you can send another PR if you still want to have this test in `trunk`.

PR #12708, opened 2023-09-08
[dotnet] Make cdp domains with lazy initialization

nvborisenko:
### Description
- Just make CDP domains as lazy objects
- Avoid creation an unnecessary Task when calling event handlers

### Motivation and Context
Each time, when browser makes a request, client side creates new cdp domain object. Let's initialize it once and reuse.

### Types of changes

### Checklist

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -119,6 +119,9 @@ jobs:
         uses: browser-actions/setup-edge@latest
         with:
           edge-version: ${{ inputs.browser-version || 'stable' }}
+      - name: Setup Safari
+        if: inputs.browser == 'safari'
+        run: sudo safaridriver --enable
       - name: Run Bazel
         run: ${{ inputs.run }}
       - name: Start SSH session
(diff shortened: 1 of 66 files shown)


titusfortner:
@nvborisenko these have test failures, can you verify?

nvborisenko:
Reproduced it locally - if we execute the following tests in shared driver:
1. TestCanInterceptNetworkCalls
2. InitializedWebDriverDoesNotThrowException
Then we get `TimeoutException`.

I should investigate what is a reason. This scenario is not reproducible in `trunk`. Thanks for pointing me.

nvborisenko:
I fixed it, but this [fix](https://github.com/SeleniumHQ/selenium/pull/12708/commits/9296a3901b9ae91f2a2e8ee7706066fb3db308f8) looks strange. I don't believe this is true fix. @titusfortner I wonder if other bindings do the same when stopping network monitoring.

Please re-run CI.

titusfortner:
@pujagani I'm still clueless about the bidi implementations. I think the .NET approach is very different from the Java approach, but do you know the answer to this one?

nvborisenko:
Fully ignore this PR for now, I converted it to draft.

nvborisenko:
Closing it, I don't want to break something what is working even if I could not understand how.

PR #12715, opened 2023-09-09
disabling console appearing on windows when running in pythonw

Q1Dian:

### Description

Hiding console when running in pythonw

### Motivation and Context

When running in pythonw, there will be a console popup that annoys the user (me as i have OCD). I found out that there is a argument STARTUPINFO in windows that makes subprocess.popen not display the console.

### Types of changes

- [ x] New feature (non-breaking change which adds functionality)

### Checklist

- [x ] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) document.

### DIFF
--- a/py/selenium/webdriver/common/service.py
+++ b/py/selenium/webdriver/common/service.py
@@ -203,16 +203,32 @@ class Service(ABC):
         cmd.extend(self.command_line_args())
         close_file_descriptors = self.popen_kw.pop("close_fds", system() != "Windows")
         try:
-            self.process = subprocess.Popen(
-                cmd,
-                env=self.env,
-                close_fds=close_file_descriptors,
-                stdout=self.log_output,
-                stderr=self.log_output,
-                stdin=PIPE,
-                creationflags=self.creation_flags,
-                **self.popen_kw,
-            )
+            if system() == "Windows":
+                si = subprocess.STARTUPINFO()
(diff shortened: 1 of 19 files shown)


titusfortner:
This is not very elegant. I'm not familiar with the constructor can we set properties on it after creating the instance?

isaulv:
@titusfortner This is correct behavior but I wonder if we should make this configurable.

titusfortner:
I think it is correct behavior to not open a separate console on Windows, so I don't think it needs to be configurable. I was hoping the solution would be a little more succinct. Maybe we just split it into 2 different methods for windows/non-windows?

isaulv:
The current method already exists. I'm fine with it really.

titusfortner:
Fair enough. If it looks good, approve the review?

py/selenium/webdriver/common/service.py
```diff
@@ -203,16 +203,32 @@ def _start_process(self, path: str) -> None:
         cmd.extend(self.command_line_args())
```
captain-hoity-toity, line 203: I think there's a simpler way to do this.

titusfortner:
Thanks @JefferyVin

Q1Dian:
np, glad to help @titusfortner

PR #12752, opened 2023-09-15
Look for Selenium Manager in path defined by Environment Variable

titusfortner:
### Description
* each of the bindings look for `SE_MANAGER_PATH` environment variable for Selenium Manager first before looking in the standard location

### Motivation and Context
* Allows users to override the bundled copy of Selenium Manager which could be useful for:
    - debugging with unreleased binary
    - using custom build on non-supported architecture
    - designating the location from a package manager (e.g., Conda)
    - putting in a directory with known permissions
* This is an alternate solution to #11359
    - Java is the only language that uses SM in a temp directory for each process and would benefit from a standard location; no good reason to move the binary for other languages
    - Using an environment variable is a more flexible solution for package managers (see conversation in #12536)
    - The cached directory would change per Selenium release, where SM in a directory with environment variable won't need to be moved when Selenium is updated (useful for custom builds and package managers)
    - This code is much more straightforward and easier to maintain than #12547 & #12539

### DIFF
--- a/dotnet/src/webdriver/SeleniumManager.cs
+++ b/dotnet/src/webdriver/SeleniumManager.cs
@@ -34,35 +34,36 @@ namespace OpenQA.Selenium
     /// </summary>
     public static class SeleniumManager
     {
-        private readonly static string binaryFullPath;
+        private static readonly string BinaryFullPath = Environment.GetEnvironmentVariable("SE_MANAGER_PATH");
 
         static SeleniumManager()
         {
-            var currentDirectory = AppContext.BaseDirectory;
 
-            string binary;
-            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            if (BinaryFullPath == null)
             {
-                binary = "selenium-manager/windows/selenium-manager.exe";
-            }
-            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
-            {
-                binary = "selenium-manager/linux/selenium-manager";
-            }
-            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
-            {
-                binary = "selenium-manager/macos/selenium-manager";
-            }
-            else
-            {
-                throw new WebDriverException("Selenium Manager did not find supported operating system");
+                var currentDirectory = AppContext.BaseDirectory;
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    BinaryFullPath = Path.Combine(currentDirectory, "selenium-manager\\windows\\selenium-manager.exe");
+                }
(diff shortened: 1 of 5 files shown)


titusfortner:
I guess I should say, this doesn't *have to be an alternative. We could do both.

I think if we want to get 4.13 out soon, this one is sufficient for that need instead of waiting to implement #11359 in all the languages.

titusfortner:
Hmm, but looks like I broke some tests. 😂

dotnet/src/webdriver/SeleniumManager.cs
```diff
@@ -41,24 +41,33 @@ static SeleniumManager()
...
-                binary = "selenium-manager/macos/selenium-manager";
+                binary = $"selenium-manager/macos/{file}";
             }
             else
             {
                 throw new WebDriverException("Selenium Manager did not find supported operating system");
             }
 
-            binaryFullPath = Path.Combine(currentDirectory, binary);
+            if (Environment.GetEnvironmentVariable("SE_MANAGER_PATH") != null)
```
nvborisenko: I believe `SE_MANAGER_PATH` is full path to the sm binary. In this case let's restructure logic:
```csharp
if (env var exists)
{
    binaryFullPath = env var
}
else
{
    binaryFullPath = // use old logic
}
```

We need it to support the following requirement:
> using custom build on non-supported architecture

If we run "old logic", then we will get *Selenium Manager did not find supported operating system*.

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -52,7 +52,11 @@ def get_binary() -> Path:
 
         file = "selenium-manager.exe" if directory == "windows" else "selenium-manager"
 
-        path = Path(__file__).parent.joinpath(directory, file)
+        if os.getenv("SE_MANAGER_PATH"):
+            directory = os.getenv("SE_MANAGER_PATH")
+            path = f'{directory}/{file}'
```
isaulv: I would use `Path(directory) / file`. It would be more windows friendly.

titusfortner:
Hah, busted for not using Java 8 syntax. Hopefully last time that ever happens.

dotnet/src/webdriver/SeleniumManager.cs
titusfortner: My thought was that it would be to the directory, not to the file (probably because that was the easiest way to do it in Ruby) 😂 
Let me think which is better. Good point on checking for OS support.

titusfortner:
Ok, so I made a few important changes.

1. It expects the filename to be part of environment variable, not just the directory.
2. It defaults to using linux even if code doesn't know what it is; (this matches Java implementation)

nvborisenko (approved):
.net code reads env variable twice, can be improved.

Issue #12691, opened 2023-09-05
[🐛 Bug]: Selenium Manager evaluating TTL too early


titusfortner:
### What happened?

This may be a bug, or this may be an issue of differing order of principles, so I'll recap mine in case that's the issue:

1. If a user tells Selenium to do something specific, it has to do it (they pass in a specific version of browser or a location of a driver or a browser, and while in beta Selenium treats having a driver on PATH as a user telling it to do something specific). If it can't do it, it errors.
2. If user doesn't specify, then Selenium does whatever it takes to execute. Even if it isn't the ideal driver version, even if they haven't run a session in forever, even if they are offline or behind a proxy. Selenium can throw a warning message explaining why something isn't ideal, but it runs the session. (Note: if the driver complains, that's a separate issue).
3. Selenium ensures that the "right" browser and driver versions are used by locating or downloading as necessary
4. Selenium improves the performance of step 3 by only looking for updates when necessary or after a certain period of time has transpired. But this is performance enhancement only, not something to prohibit functionality..

The log is below, but extra info:

```
titusfortner@SL-1837 code % ls -al ~/.cache/selenium/chromedriver/mac-x64 
drwxr-xr-x@ 4 titusfortner  staff  128 Sep  3 07:29 .
drwxr-xr-x@ 3 titusfortner  staff   96 Aug 30 07:58 ..
drwxr-xr-x@ 3 titusfortner  staff   96 Aug 30 07:58 116.0.5845.96
```

It's just that the ttl is expired:
```
titusfortner@SL-1837 code % cat ~/.cache/selenium/se-metadata.json 
{
  "browsers": [],
  "drivers": [
    {
      "major_browser_version": "116",
      "driver_name": "chromedriver",
      "driver_version": "116.0.5845.96",
      "driver_ttl": 1693797814
    }
  ]
}
```

### How can we reproduce the issue?

```shell
set ttl to a value in the past and run with `--browser chrome --output json --offline`
```

### Relevant log output

```shell
2023-09-05 08:48:43 FINE - [SeleniumManager getBinary] Selenium Manager binary found at: /var/folders/rh/25cw4qq10_157fjpztk0psx80000gn/T/selenium-manager18251076474367938167165495520198471/selenium-manager 
2023-09-05 08:48:46 FINE - [SeleniumManager runCommand] Executing Process: [--browser, chrome, --output, json, --debug, --offline] 
[...]

PR #12753, opened 2023-09-15
[rust] Search better driver possible in the cache

bonigarcia:
### Description
This PR  add some Rust logic to get the _best_ possible driver from the cache when the regular algorithm does not work (e.g., in offline mode or other error). It throws a `WARN` when this logic is reached. This _best_ driver is found in two different ways:

1. It tries to find a driver that matches the major browser version. This should for Chrome and Edge, but not for Firefox (since geckodriver follows a different versioning schema than Firefox).
2. It tries to find the latest driver in the cache.

Example (offline mode but the driver was already there):

```
boni@slimbook:~/dev/selenium/rust$ cargo run -- --browser chrome --debug --offline
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/selenium-manager --browser chrome --debug --offline --clear-metadata`
WARN	Offline flag set, but also asked not to avoid browser downloads. Honouring offline flag
DEBUG	Using Selenium Manager in offline mode
DEBUG	Deleting metadata file /home/boni/.cache/selenium/se-metadata.json
DEBUG	chromedriver not found in PATH
DEBUG	chrome detected at /opt/google/chrome/google-chrome
DEBUG	Running command: /opt/google/chrome/google-chrome --version
DEBUG	Output: "Google Chrome 116.0.5845.140 "
DEBUG	Detected browser: chrome 116.0.5845.140
WARN	There was an error managing chrome (Unable to discover proper chromedriver version in offline mode); using driver found in the cache
INFO	Driver path: /home/boni/.cache/selenium/chromedriver/linux64/116.0.5845.96/chromedriver
INFO	Browser path: /opt/google/chrome/google-chrome
```

### Motivation and Context
This will allow to prevent problems explained in #12691.

### Types of changes

### Checklist

### DIFF
--- a/rust/Cargo.Bazel.lock
+++ b/rust/Cargo.Bazel.lock
@@ -1,5 +1,5 @@
 {
-  "checksum": "0c722a74f5cadc8bb75a62763e298e0698afcb48a51d9f7bf6754e285caa644b",
+  "checksum": "1f16f4cfa433faeb0bd8242b382da4d71e41405dc8a07363d8c1f410ee0d651e",
   "crates": {
     "addr2line 0.19.0": {
       "name": "addr2line",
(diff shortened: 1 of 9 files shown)


titusfortner:
Just curious, if 116 isn't in cache, will it look for 115?

I'm not as worried about DEBUG messages, but I think we need to be more clear with WARN messages.

> WARN	Offline flag set, but also asked not to avoid browser downloads. Honouring offline flag

This message is confusing. We shouldn't refer to "offline" unless the user actually is offline, but right now we're using "offline mode" as the "non opted-in" mechanism for the Grid. (We can't disable SM because we need it to check on path, but we aren't accessing the network because we want users to opt-in to that). 

How about:
> WARN 	 downloads disabled, to enable set --selenium-manager true

or 
> WARN 	 network calls disabled, to enable set --selenium-manager true

bonigarcia:
> Just curious, if 116 isn't in cache, will it look for 115?

If 116 is not in the cache, it will look for latest (which might be 117, 115, or whatever latest).

> I'm not as worried about DEBUG messages, but I think we need to be more clear with WARN messages.

I agree. That `WARN` message was introduced by Simon in their last change to fix the RBE build. But for me, it is more a `DEBUG`  than a `WARN`. But I decided not to change it in this PR since I want to review it carefully in a future change.

titusfortner:
No, I think `WARN` is right. If the user doesn't get what we think is "right" we should warn them, but the message needs to be informative and it should explain how to fix it or link them to a place they can get that info. More people will opt-in to using Selenium Manager if SM warns them they aren't getting what they should be using.

titusfortner:
Oh, you're right, these should be debug.

PR #12754, opened 2023-09-15
[rust] update log messages

titusfortner:
### Description
* change a logging from warn to debug
* add message to an error

### Motivation and Context
* since "avoid browser downloads" is the default, we should not throw warning
* when we throw a warning we should give the suggestion for the fix

This is why I'm wondering if it makes sense to call this something else other than offline mode, because I'm not sure this message always applies. What do you think?


rust/src/lib.rs
```diff
@@ -98,7 +98,7 @@ pub const TTL_SEC: u64 = 3600;
 pub const UNAME_COMMAND: &str = "uname -{}";
 pub const ESCAPE_COMMAND: &str = "printf %q \"{}\"";
 pub const SNAPSHOT: &str = "SNAPSHOT";
-pub const OFFLINE_REQUEST_ERR_MSG: &str = "Unable to discover proper {} version in offline mode";
+pub const OFFLINE_REQUEST_ERR_MSG: &str = "Unable to discover proper {} version in offline mode; run grid node with --selenium-manager true";
```
diemol, line 101: This is already displayed in the Grid. No need to change it here.

rust/src/main.rs
```diff
@@ -163,7 +163,7 @@ fn main() {
         }
         cli.force_browser_download = false;
         if !cli.avoid_browser_download {
-            log.warn("Offline flag set, but also asked not to avoid browser downloads. Honouring offline flag");
+            log.debug("Offline flag set, but also asked not to avoid browser downloads. Honouring offline flag");
```
diemol, line 166: Yes, this is what I wanted to change a while ago.

diemol (changes requested):

titusfortner:
Cool, I just committed the one change, and I'll close this

PR #12761, opened 2023-09-17
Better way to handle SM env variable and unsupported platform

nvborisenko:
### Description
This is just a better way how to handle env var for sm path. And react appropriately if it is not specified.

### Motivation and Context
Make Selenium better, improve https://github.com/SeleniumHQ/selenium/pull/12752

### Types of changes

### Checklist

### DIFF
--- a/dotnet/src/webdriver/SeleniumManager.cs
+++ b/dotnet/src/webdriver/SeleniumManager.cs
@@ -38,28 +38,28 @@ namespace OpenQA.Selenium
 
         static SeleniumManager()
         {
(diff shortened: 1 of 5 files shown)


titusfortner:
Thanks for this. For right now I decided to go with minimal change to existing code that ensures platform is not evaluated if ENV is set. I'm not making any changes as to whether we should or shouldn't throw when platform is not recognized, though I think we probably should come to some agreement on that for all languages in a future release. I'm not entirely sure how UNIX platforms should be handled for instance.

PR #12764, opened 2023-09-18
Roll pinned browsers

shs96c:


titusfortner:
this has been updated since.

Issue #12760, opened 2023-09-17
[🚀 Feature]: Python Service class Properties


titusfortner:
### Feature and motivation

Right now Python Driver Services can only have parameters set in the constructor.

I think those arguments should also have property setter/getters, or whatever equivalent we are moving to.

### Usage example

```python
service = webdriver.ChromeService()
service.service_args = ['--log', 'debug']
```

PR #12767, opened 2023-09-18
[py] implementation of new feature-12760, setting `service_args`  through `getters/setters`

sandeepsuryaprasad:

### Description
This PR is an implementation of new feature request https://github.com/SeleniumHQ/selenium/issues/12760, for setting `service_args` not only through constructor, but also through `setter/getter` methods using properties.

### Motivation and Context
New feature request [12760](https://github.com/SeleniumHQ/selenium/issues/12760)
### Types of changes

### Checklist

### DIFF
--- a/py/python.iml
+++ b/py/python.iml
@@ -15,7 +15,7 @@
       <sourceFolder url="file://$MODULE_DIR$/test" isTestSource="true" />
       <excludeFolder url="file://$MODULE_DIR$/venv" />
     </content>
-    <orderEntry type="jdk" jdkName="Python 3.9 (py)" jdkType="Python SDK" />
+    <orderEntry type="jdk" jdkName="Python 3.11" jdkType="Python SDK" />
     <orderEntry type="sourceFolder" forTests="false" />
   </component>
   <component name="sonarModuleSettings">
--- a/py/python.iml
+++ b/py/python.iml
@@ -15,7 +15,7 @@
       <sourceFolder url="file://$MODULE_DIR$/test" isTestSource="true" />
       <excludeFolder url="file://$MODULE_DIR$/venv" />
     </content>
-    <orderEntry type="jdk" jdkName="Python 3.11" jdkType="Python SDK" />
+    <orderEntry type="jdk" jdkName="Python 3.9 (py)" jdkType="Python SDK" />
     <orderEntry type="sourceFolder" forTests="false" />
   </component>
   <component name="sonarModuleSettings">
--- a/py/selenium/webdriver/chromium/service.py
+++ b/py/selenium/webdriver/chromium/service.py
@@ -40,10 +40,12 @@ class ChromiumService(service.Service):
         env: typing.Optional[typing.Mapping[str, str]] = None,
         **kwargs,
     ) -> None:
-        self.service_args = service_args or []
+        if service_args is None:
+            service_args = []
+        self._service_args = service_args
 
         if isinstance(log_output, str):
-            self.service_args.append(f"--log-path={log_output}")
+            self._service_args.append(f"--log-path={log_output}")
             self.log_output = None
         else:
             self.log_output = log_output
@@ -56,5 +58,15 @@ class ChromiumService(service.Service):
             **kwargs,
         )
 
+    @property
+    def service_args(self):
+        return self._service_args
+
+    @service_args.setter
+    def service_args(self, value):
+        if not isinstance(value, list):
+            raise TypeError("service args must be a list")
+        self._service_args.extend(value)
+
     def command_line_args(self) -> typing.List[str]:
-        return [f"--port={self.port}"] + self.service_args
+        return [f"--port={self.port}"] + self._service_args
--- a/py/selenium/webdriver/chromium/service.py
+++ b/py/selenium/webdriver/chromium/service.py
@@ -35,14 +35,14 @@ class ChromiumService(service.Service):
         self,
         executable_path: str = None,
         port: int = 0,
-        service_args: typing.Optional[typing.List[str]] = None,
+        service_args: typing.Sequence[str] = None,
         log_output: SubprocessStdAlias = None,
         env: typing.Optional[typing.Mapping[str, str]] = None,
         **kwargs,
     ) -> None:
         if service_args is None:
             service_args = []
-        self._service_args = service_args
+        self.service_args = service_args
 
         if isinstance(log_output, str):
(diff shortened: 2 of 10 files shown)


py/selenium/webdriver/chromium/service.py
```diff
@@ -40,10 +40,12 @@ def __init__(
         env: typing.Optional[typing.Mapping[str, str]] = None,
         **kwargs,
     ) -> None:
-        self.service_args = service_args or []
```
titusfortner, line 43: just curious, is this not a good convention?

py/selenium/webdriver/ie/service.py
```diff
@@ -40,16 +39,20 @@ def __init__(
          - executable_path : Path to the IEDriver
          - port : Port the service is running on
          - host : IP address the service port is bound
-         - log_level : Level of logging of service, may be "FATAL", "ERROR", "WARN", "INFO", "DEBUG", "TRACE".
-           Default is "FATAL".
+         - log_level : Level of logging of service, may be "DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL".
```
titusfortner: These are not the IE Driver log levels

py/selenium/webdriver/ie/service.py
```diff
@@ -40,16 +39,20 @@ def __init__(
...
+           Default is "WARNING".
```
titusfortner: default is determined by driver

py/python.iml
```diff
@@ -15,7 +15,7 @@
       <sourceFolder url="file://$MODULE_DIR$/test" isTestSource="true" />
       <excludeFolder url="file://$MODULE_DIR$/venv" />
     </content>
-    <orderEntry type="jdk" jdkName="Python 3.9 (py)" jdkType="Python SDK" />
+    <orderEntry type="jdk" jdkName="Python 3.11" jdkType="Python SDK" />
```
titusfortner: I think we want to keep minimum supported here, which is 3.9

titusfortner (changes requested):
The move to `_service_args` looks correct. Otherwise I think there are a few issues that need to be updated. Thanks.

py/selenium/webdriver/chromium/service.py
sandeepsuryaprasad: It is good one-liner assignment, kind of short-cut. But most of the times I see explicit check for truthiness checking against `None`.  I think both ways are fine.

py/selenium/webdriver/ie/service.py
sandeepsuryaprasad: ohh sorry.. kind of confused with regular python log levels..

py/python.iml
sandeepsuryaprasad: I am not sure how this got changed... never edited this file... it needs to be reverted back to 3.9

titusfortner:
Also, can you rebase with trunk, I think we have conflicts

sandeepsuryaprasad:
I have cleaned it up..

py/selenium/webdriver/safari/service.py
```diff
@@ -42,20 +42,30 @@ def __init__(
...
-        return ["-p", f"{self.port}"] + self.service_args
+    @property
+    def service_args(self):
+        return self._service_args
+
+    @service_args.setter
+    def service_args(self, value):
+        if not isinstance(value, list):
+            raise TypeError("service args must be a list")
+        self._service_args.extend(value)
```
jsa34: It might be just me, but the setter seems more like an append rather than replace/set the value.

Is this the intended behaviour?  Setters for me usually replace the existing value

sandeepsuryaprasad: every time when assignment happens to `service_args` I am not replacing the old args, rather I am updating the existing contents of the list. 
e.g. 
```
service = Service(service_args = ["--log", "debug"])
service.service_args = ["--port", 1234]
print(service.service_args)
# prints 
["--log", "debug", "--port", 1234]
```
@titusfortner , should this be the intended behaviour? or should the old list be replaced with new one every time an assignment happens to `service_args`?

titusfortner: I don't know the Python convention, but it does feel a little weird to append to something existing with an assignment operator. (They way Ruby does it is so much better 😉). I lean toward replacing instead of adding. @isaulv what do you think?

AutomatedTester: This should definitely not be a `setter` as that creates the expectation that it is setting the value that you pass in. It is not intuitive to users that it will extend. We either need a different method name or we should see about closing this PR.

py/selenium/webdriver/chromium/service.py
```diff
@@ -56,5 +58,15 @@ def __init__(
             **kwargs,
         )
 
+    @property
+    def service_args(self):
+        return self._service_args
+
+    @service_args.setter
+    def service_args(self, value):
+        if not isinstance(value, list):
```
symonk: should probably accept a `Sequence` here instead of explicitly requiring a `list` check

py/selenium/webdriver/chromium/service.py
```diff
@@ -56,5 +58,15 @@ def __init__(
...
+            raise TypeError("service args must be a list")
+        self._service_args.extend(value)
```
symonk: a `@setter` should probably replace the entire sequence, not extend to it - this will create confusing behaviour

symonk:
Thanks @sandeepsuryaprasad! These seem to be 'placeholder' getter/setters for things at the moment which python typically doesn't really care about (its not as painful to add them in when required compared to other statically typed languages like java, but I don't really mind if we have them, can help with some validation around setting the service args

py/selenium/webdriver/chromium/service.py
```diff
@@ -56,5 +58,15 @@ def __init__(
             **kwargs,
         )
 
+    @property
+    def service_args(self):
```
symonk, line 62: be nice to add typing to things we are adding, feel free to ignore tho as theres plenty of general typing work to do

sandeepsuryaprasad: @symonk thanks for your review comments. I have updated the code.

titusfortner:
Sorry this one has languished, can you de-conflict it and I'll review again? Thanks!

diemol:
I am closing this because there are several file conflicts, but also, there was no more activity from the OP.

cgoldberg:
this was done in #15889

PR #12772, opened 2023-09-18
[dotnet] Use the lower version of Newtonsoft.Json as dependency

nvborisenko:
### Description
Any library should be dependent on lower version of dependency as possible. Selenium can work with `Newtonsoft.Json v5.0.1`.

### Motivation and Context
Just making selenium better.

### Types of changes

### Checklist


nvborisenko:
It should not even build successfully. I am surprised.

Issue #12692, opened 2023-09-05
[🐛 Bug]: User.Identity.IsAuthenticated and User.Identity.Name not working anymore since webDriver version 4.12


fangwangse:
### What happened?

I have OIDC authentication enabled in my API. When user gets authenticated via the swagger UI (or web client), I get correct values for 
_httpContextAccessor.HttpContext.User.Identity.Name
_httpContextAccessor.HttpContext.User.Identity.IsAuthenticated (=True)
in my services.

When I add the following packages:
    <PackageReference Include="Selenium.Support" Version="4.11.0" />
    <PackageReference Include="Selenium.WebDriver" Version="4.11.0" />
It still works fine.

But, if I upgrade above packages to the latest version 4.12.x, i.e.:
    <PackageReference Include="Selenium.Support" Version="4.12.0" />
    <PackageReference Include="Selenium.WebDriver" Version="4.12.0" />

_httpContextAccessor.HttpContext.User.Identity.Name
_httpContextAccessor.HttpContext.User.Identity.IsAuthenticated
stop returning correct values:
 ...Identity.Name is always null, ...Identity.IsAuthenticated is always False.

If I remove these 2 packages, or downgraded back to 4.11.0, correct values are back as well.

### How can we reproduce the issue?

```shell
In Program.cs:

services.AddHttpContextAccessor();
services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme).AddJwtBearer(...);
services.AddScoped<IHttpClientFactory, HttpClientFactory>();
services.AddScoped<IUserAuthenticationService, UserAuthenticationService>();
services.AddSwaggerGen(options =>
        {
            var oidcAddress = "your jwt token server address";
            options.AddSecurityDefinition(AuthorizationSystems.Oidc.ToString(), new OpenApiSecurityScheme
            {
                Type = SecuritySchemeType.OAuth2,
                Flows = new OpenApiOAuthFlows
                {
                    AuthorizationCode = new OpenApiOAuthFlow
                    {
                        AuthorizationUrl = new Uri($"{oidcAddress}/auth/realms/master/protocol/openid-connect/auth"),
                        TokenUrl = new Uri($"{oidcAddress}/auth/realms/master/protocol/openid-connect/token")
                    }
                }
            })
            .AddSecurityRequirement(new OpenApiSecurityRequirement
            {
                {
                    new OpenApiSecurityScheme
                    {
                        Reference = new OpenApiReference
                        {
[...]

PR #12777, opened 2023-09-19
[dotnet] Remove Microsoft.IdentityModel.Tokens as dependency

nvborisenko:
### Description
We can remove `Microsoft.IdentityModel.Tokens` dependency.

What is done:
- Adopted small piece code what we need
- Remove everything related to `Microsoft.IdentityModel.Tokens`

Original source: https://raw.githubusercontent.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/6.19.0/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs

### Motivation and Context
Ideally any library should not have dependencies.
Related to https://github.com/SeleniumHQ/selenium/issues/12692

### Types of changes

### Checklist


dotnet/src/webdriver/Internal/Base64UrlEncoder.cs
```diff
@@ -0,0 +1,159 @@
+// Thanks to Microsoft.IdentityModel.Tokens package for this code.
```
titusfortner: We need to put the Selenium Copyright at the top. I think we put the links to the methods we are drawing from in the comments to the specific methods (see below).

dotnet/src/webdriver/Internal/Base64UrlEncoder.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// Converts a subset of an array of 8-bit unsigned integers to its equivalent string representation which is encoded with base-64-url digits. Parameters specify
+        /// the subset as an offset in the input array, and the number of elements in the array to convert.
+        /// </summary>
+        /// <param name="inArray">An array of 8-bit unsigned integers.</param>
+        /// <param name="length">An offset in inArray.</param>
+        /// <param name="offset">The number of elements of inArray to convert.</param>
+        /// <returns>The string representation in base 64 url encoding of length elements of inArray, starting at position offset.</returns>
+        /// <exception cref="ArgumentNullException">'inArray' is null.</exception>
+        /// <exception cref="ArgumentOutOfRangeException">offset or length is negative OR offset plus length is greater than the length of inArray.</exception>
+        public static string Encode(byte[] inArray, int offset, int length)
```
titusfortner: Selenium doesn't call this method signature directly, and we end up checking for null twice. Since offset is always 0, we should be able to combine this method into the one below.

dotnet/src/webdriver/Internal/Base64UrlEncoder.cs
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        /// <summary>
+        /// Converts a subset of an array of 8-bit unsigned integers to its equivalent string representation which is encoded with base-64-url digits.
+        /// </summary>
+        /// <param name="inArray">An array of 8-bit unsigned integers.</param>
+        /// <returns>The string representation in base 64 url encoding of length elements of inArray, starting at position offset.</returns>
+        /// <exception cref="ArgumentNullException">'inArray' is null.</exception>
+        /// <exception cref="ArgumentOutOfRangeException">offset or length is negative OR offset plus length is greater than the length of inArray.</exception>
+        public static string Encode(byte[] inArray)
```
titusfortner: Add link to source code: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/6.19.0/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs#L85

(I think we should use github.com not githubusercontent, and link to the line of the method).

dotnet/src/webdriver/Internal/Base64UrlEncoder.cs
```diff
@@ -0,0 +1,159 @@
...
+            _ = inArray ?? throw new ArgumentNullException(nameof(inArray));
+
+            return Encode(inArray, 0, inArray.Length);
+        }
+
+        /// <summary>
+        ///  Converts the specified string, which encodes binary data as base-64-url digits, to an equivalent 8-bit unsigned integer array.</summary>
+        /// <param name="str">base64Url encoded string.</param>
+        /// <returns>UTF8 bytes.</returns>
+        public static byte[] DecodeBytes(string str)
```
titusfortner, line 123: Link to put in comments (as described above)- https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/6.19.0/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs#L179

titusfortner (changes requested):
This looks good to me, will be nice to remove the dependency.

PR #12781, opened 2023-09-19
[dotnet] Remove System.Drawing.Common as package dependency

nvborisenko:
This is continuation of https://github.com/SeleniumHQ/selenium/pull/12777, please review that first.

### Description
I think selenium is not responsible to convert images from one format to another. It is already enough for users to deal with images in `png` format, even it is possible to save image as file.

If user wants to convert image, then he is able to get image as bytes and use any other 3rd-party library especially designed to work with graphics.

### Motivation and Context
Continue my way to reduce package dependencies.

### Types of changes

### Checklist


titusfortner:
I think where this is coming from is that Java supports:
* Base64
* Bytes
* File (stored as png)

If .NET doesn't allow saving as png in default libraries...

We can't just remove this. We can deprecate the method.
We should have a different method to return it as a Stream so a user knows they need to do something else with it.

Another option that might be nice is to move the ability to save a stream as a PNG to the support package.

nvborisenko:
Now all users will get a warning. And those, who are really converting image to different formats, will get exception at runtime.

titusfortner:
Please rebase on trunk and fix conflicts so I can better see what this is doing. :)

titusfortner:
> those, who are really converting image to different formats, will get exception

I think we want to do this in stages. If we are removing functionality we should mark it as obsolete for a couple releases, then remove it.

nvborisenko:
Agree, I will prepare separate PR just to mark methods as obsolete. Keeping this one as open until we decide completely remove it.

dotnet/src/webdriver/Screenshot.cs
nvborisenko, line 1: As a reminder for myself: we need adjust tests accordingly.

nvborisenko: Done.

nvborisenko:
I think we are ready to land it.

ckuetbach:
Two weeks for a breaking change in a minor release is somewhat annoying.

The Obsolete message was merged 22nd October and the breaking change was introduced Nov 10. (https://github.com/SeleniumHQ/selenium/pull/12878)

nvborisenko:
The obsolete message shipped with v4.14 (25 Sep) and breaking change shipped with v4.16 (06 Dec). Selenium project doesn't follow semantic versioning, sorry for inconvenience.

PR #12784, opened 2023-09-19
[bazel] Update `rules_dotnet` to the latest version

shs96c:


dotnet/test/common/WebDriver.Common.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net48</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0,net7.0</TargetFrameworks>
```
nvborisenko: Project with tests should not be targeted to `netstandard`. If it is not so easy to target `net48`, then we can stay on `net7.0` only, and add more targets later.

shs96c: Done

dotnet/src/webdriver/WebDriver.nuspec
```diff
@@ -36,12 +36,6 @@
...
   </metadata>
   <files>
-    <file src="**\WebDriver.dll" target="lib" />
-    <file src="**\WebDriver.pdb" target="lib" />
-    <file src="**\WebDriver.xml" target="lib" />
-    <file src="..\icon.png" target="images" />
-    <file src="..\manager\**" target="manager" />
-    <file src="..\Selenium.WebDriver.targets" target="build" />
-    <file src="..\Selenium.WebDriver.targets" target="buildTransitive" />
+    <file src="**" target="." exclude="obj\**;*.csproj;*.config"/>
```
nvborisenko, line 39: Better to declare package's content as it was before, in more precise way. Now it's something unmanageable, unpredictable content.

dotnet/src/webdriver/Internal/Base64UrlEncoder.cs
```diff
@@ -23,7 +23,7 @@ namespace OpenQA.Selenium.Internal
     /// <summary>
     /// Encodes and Decodes strings as Base64Url encoding.
     /// </summary>
-    internal static class Base64UrlEncoder
+    public static class Base64UrlEncoder
```
nvborisenko, line 26: Keep it internal, already implemented in `trunk`.

dotnet/src/webdriver/Internal/PortUtilities.cs
```diff
@@ -24,7 +24,7 @@ namespace OpenQA.Selenium.Internal
     /// <summary>
     /// Encapsulates methods for working with ports.
     /// </summary>
-    internal static class PortUtilities
+    public static class PortUtilities
```
nvborisenko, line 27: Keep it internal, already implemented in `trunk`.

dotnet/src/tools/BUILD.bazel
nvborisenko, line 1: Seems it is temporary, should we clean up?

dotnet/paket.lock
nvborisenko, line 1: Is `paket` really great? We are becoming far and far from "normal" .net dev procedures. If it is really OK, then it would be nice to create a simple doc how to update some dependency with transitive dependencies, and it will work locally and in bazel.

dotnet/0001-Include-more-of-the-SDK.patch
nvborisenko, line 1: Seems extra files for this repo.

shs96c: Yeah. These are the patches we apply to `rules_dotnet` to make it work properly for us.

dotnet/paket.lock
shs96c: I'd prefer to use straight `nuget`, as I think that's probably what other people are familiar with, but there are primitives in `rules_dotnet` that allow us to easily manage third party dependencies with `paket` and none for straight `nuget`. I'd rather we have something that allows us to work easily with dependencies than nothing at all.

dotnet/src/tools/BUILD.bazel
shs96c: This was the only way I had of inspecting the DLLs generated by the build. The other tooling was either out of date, targeted a platform other than the version of .Net that runs on macOS, or are Windows only. Happy to delete it if no-one else will find it useful.

dotnet/src/webdriver/Internal/Base64UrlEncoder.cs
shs96c: I'll rebase on trunk

dotnet/src/webdriver/Internal/PortUtilities.cs
shs96c: I'll rebase on trunk

dotnet/src/webdriver/WebDriver.nuspec
shs96c: The `nupsec` files are written to work with the bazel build, and aren't designed to be used with the regular .Net tooling. The content is completely predictable, since we construct the contents hermetically.

nvborisenko:
```
<file src="**" target="." exclude="obj\**;*.csproj;*.config"/>
```

I don't think it's predictable content. Everything what is on disk now will be packed, it is not something we expect. Just explore nupkg file, you might find extra files there.

PR #12830, opened 2023-09-27
[java] support non-standard commands in RemoteWebDriverBuilder

titusfortner:
Builder does magic to create instances of HttpCommandExecutor inside a lambda that I'm not sure I completely follow, but I think this works.

### Description
HttpCommandExecutor has a [Constructor for setting additional commands](https://github.com/SeleniumHQ/selenium/blob/selenium-4.13.0/java/src/org/openqa/selenium/remote/HttpCommandExecutor.java#L82-L83)

Then when getting the commandCodec, it [loops through to defines those commands](https://github.com/SeleniumHQ/selenium/blob/selenium-4.13.0/java/src/org/openqa/selenium/remote/HttpCommandExecutor.java#L166-L168)

I think this is the right way to do this,

### Motivation and Context
This allows Appium users to use RemoteWebDriverBuilder


java/src/org/openqa/selenium/remote/RemoteWebDriverBuilder.java
```diff
@@ -438,7 +452,13 @@ private DriverService startDriverServiceIfNecessary() {
 
   private CommandExecutor createExecutor(HttpHandler handler, ProtocolHandshake.Result result) {
     Dialect dialect = result.getDialect();
-    Function<Command, HttpRequest> commandEncoder = dialect.getCommandCodec()::encode;
+    CommandCodec<HttpRequest> commandCodec = dialect.getCommandCodec();
+    for (Map.Entry<String, CommandInfo> entry : additionalCommands.entrySet()) {
+      commandCodec.defineCommand(
+          entry.getKey(), entry.getValue().getMethod(), entry.getValue().getUrl());
+    }
```
pujagani: This is nit. The change does not need to be incorporate. Just a suggestion, this can use forEach: 
```
 additionalCommands.forEach(
        (name, commandInfo) ->
            commandCodec.defineCommand(name, commandInfo.getMethod(), commandInfo.getUrl()));
```

pujagani:
LGTM to me logically.

pujagani:
Is there a way we can add tests?

pujagani:
If we were to document the functionality this  change adds, it might be good to call out that it can also be done via https://github.com/SeleniumHQ/selenium/blob/9a6947ea4fd685b5f997415565bc5dc875a71321/java/src/org/openqa/selenium/remote/AdditionalHttpCommands.java#L27.

java/src/org/openqa/selenium/remote/RemoteWebDriverBuilder.java
titusfortner: I copied straight from HttpCommandExecutor class, but probably it does that because it is using two steps, I think to test for nulls in entry set? At any rate, an easy change to update.

titusfortner:
I'm not sure what you mean as far as the documentation comment. ` public interface AdditionalHttpCommands ` only provides `getAdditionalCommands()` which we can't support because our implementation is all wrapped up in a lambda and not actually a class. 

I'll see if I can figure out how to add a test.

java/src/org/openqa/selenium/remote/RemoteWebDriverBuilder.java
pujagani: I think both do the same thing.

pujagani:
I should have been clear. From what I understand, after adding the changes in this PR. There are 2 ways to add additional commands, one via the APIs defined in this change while using the RemoteWebDriverBuilder and one via using the Java service Loader that the `AdditionalHttpCommand` interface provides. The code hits https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/remote/Dialect.java#L57 which picks up additional commands using the Java service loader. Example: https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/chrome/AddHasCdp.java. I might be missing something here if that's not the case. I think it might be good for us to document both ways whenever we document it. I can also help document the second way.

titusfortner:
Good call on the tests. I'm not sure this actually works without the `AutoService` magic.

The use case is for Appium. 
Appium extends RemoteWebDriver and adds a bunch of its own methods:

So like this:
```
    class CheeseDriver extends RemoteWebDriver {
      public String getCheese() {
        return (String) getExecuteMethod().execute("cheese", null);
      }
    }
```

The idea for Builder is to do this:
```
    CommandInfo cheese = new CommandInfo("session/:sessionId/cheese", HttpMethod.GET);
    Map<String, CommandInfo> commands = ImmutableMap.of("getCheese", cheese);

    WebDriver driver =
        RemoteWebDriver.builder()
            .address(uri.toString())
            .oneOf(new FirefoxOptions())
            .additionalCommands(commands)
            .build();
```

But then you have to be able to cast this to the subclass, which isn't working.

I can do:
    RemoteWebDriver rwd = (RemoteWebDriver) driver;

But I can't do
    CheeseDriver cd = (CheeseDriver) driver;


I added the failing test code in the PR to show what I'm trying.

titusfortner:
Does `AutoService` only work for things in the same package, or can people have subclasses in other projects that can make use of it? Maybe it would be easier to work with Appium team to pull things in automatically?

pujagani:
It should work in other projects if the service provider i.e.  the additional command interface implementation has properly defined services file in the META-INF folder. I tried this out locally. I used a local Java project to implement the AddAdditionalCommands interface and added the appropriate services file. I built the jar locally and used it as a dependency our examples Java repo. Ran a quick test and debugged to see if the [code](https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/remote/Dialect.java#L59) picks it up while service loading and it does. We can work with the Appium team if they want to go this route.

titusfortner:
Well, this code won't work, so I'm closing it. We'll move the discussion to https://github.com/appium/java-client/issues/2030

PR #12834, opened 2023-09-28
[dotnet] Remove IdentityModel nuget from bazel workspace

nvborisenko:
### Description
We didn't clean IdentityModel dependency completely.

### Motivation and Context
Fix .net CI build.

### Types of changes

### Checklist


dotnet/src/webdriver/Internal/Base64UrlEncoder.cs
```diff
@@ -23,7 +23,7 @@ namespace OpenQA.Selenium.Internal
     /// <summary>
     /// Encodes and Decodes strings as Base64Url encoding.
     /// </summary>
-    internal static class Base64UrlEncoder
+    public static class Base64UrlEncoder
```
nvborisenko: I don't like making it public, will redo it via `InteralsVisibleTo` test assembly.

Issue #11606, opened 2023-02-01
[🐛 Bug]: Cannot create DevTools session twice


apetrivs:
### What happened?

Since 4.7.1 selenium version DevTools session cannot be created after DevTools session was opened and closed.
TimeoutException is thrown.

Steps to reproduce:

1. Create DevTools session
2. Enable Network.
3. Close session.
4. Try to create new session with the same driver.

### How can we reproduce the issue?

```shell
import io.github.bonigarcia.wdm.WebDriverManager;
import org.openqa.selenium.chrome.ChromeDriver;
import org.openqa.selenium.devtools.DevTools;
import org.openqa.selenium.devtools.v108.network.Network;

import java.util.Objects;
import java.util.Optional;

public class Main {

    public static void main(String[] args) {
        WebDriverManager.chromedriver().setup();
        ChromeDriver chromeDriver = null;
        try {
            chromeDriver = new ChromeDriver();
            DevTools devTools = chromeDriver.getDevTools();
            devTools.createSession();
            devTools.send(Network.enable(Optional.empty(), Optional.empty(), Optional.empty()));

            devTools.clearListeners();
            devTools.close();

            devTools = chromeDriver.getDevTools();
            devTools.createSession(); // StackTrace starts here!
            devTools.send(Network.enable(Optional.empty(), Optional.empty(), Optional.empty()));

            devTools.clearListeners();
            devTools.close();
        } finally {
            Objects.requireNonNull(chromeDriver).quit();
        }
    }
}
```

### Relevant log output

```shell
SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
Starting ChromeDriver 109.0.5414.74 (e7c5703604daa9cc128ccf5a5d3e993513758913-refs/branch-heads/5414@{#1172}) on port 15746
Only local connections are allowed.
Please see https://chromedriver.chromium.org/security-considerations for suggestions on keeping ChromeDriver safe.
ChromeDriver was started successfully.
Feb 01, 2023 11:52:10 AM org.openqa.selenium.devtools.CdpVersionFinder findNearestMatch
WARNING: Unable to find an exact match for CDP version 109, so returning the closest version found: 108
Exception in thread "main" org.openqa.selenium.TimeoutException: java.util.concurrent.TimeoutException
Build info: version: '4.7.1', revision: 'c6795baf1a3'
[...]

PR #11345, opened 2022-12-01
[java] Add close method to JDK 11 client. Ensure close methods for Http client is called.

pujagani:
**Thanks for contributing to Selenium!**
**A PR well described will help maintainers to quickly review and merge it**

Before submitting your PR, please check our [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) guidelines.
Avoid large PRs, help reviewers by making them as simple and short as possible.

### Description

Fixes #11270.

### Motivation and Context

1. The memory leak is prominently seen when Augmenter is used which in turn creates a WebSocket connection using the HTTP Client to create a CDP connection. Though webdriver implementation supporting CDP does the same. The same client instance needs to be closed to ensure no resources are being held when the driver quits. The changes add the required code to achieve this.

2. Additionally, we supply an executor to the JDK 11 Http client for the worker threads, which is shutdown when the client is closed along with the underlying websocket. We also ensure there is no reference to the JDK 11 Http client instance after closing the client. This has helped fix the ever-increasing threads related to the JDK 11 Http client. HttpClientImpl (Java 11's implementation) does a timely check internally to see if no references to the client are there then closes the Selector Manager threads which are later collected when the GC kicks in. 

The combination of the above 2 changes has helped ensure that there is no leak from JDK 11 Http client in different scenarios.

Before the changes: 
Screenshot 2022-12-01 at 12 58 39 PM
 
After the changes: 
Screenshot 2022-11-30 at 8 18 37 PM

### Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist

- [X] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [X] All new and existing tests passed.

PR #12843, opened 2023-09-30
Require Java 11

### LINKED ISSUE #11526: [🚀 Feature]: Java 11 as the minimum supported version for Selenium Java bindings
### Feature and motivation

Currently, Java 8 is the minimum supported version. The goal is to bump this to Java 11. 
This will help take advantage of JDK 11 HTTP client as the default client (which is currently optional, read more on the usage https://www.selenium.dev/blog/2022/using-java11-httpclient/) and remove dependency on AsyncHttpClient. This change will help resolve major issues like https://github.com/SeleniumHQ/selenium/issues/9528 for all Selenium clients.
Additionally, it will help remove the Guava dependency for Immutable collections as well (this change can be done in iterations once we move to Java 11). 

### Usage example

Usage will remain as it is. Users will not need to add explicit dependency on Guava library in their test setups.

titusfortner:
I'm sure there's a lot more that can be cleaned up that I'm missing, but I think this is the first step.

### Description
* Requires Java 11
* Removes Netty Http client
* Defaults to Native Java 

### Motivation and Context
Fixes #11526


java/src/org/openqa/selenium/remote/http/HttpClient.java
joerg1985, line 1: I would keep the system property, to allow to use custom client implementations. If it is not wished to support other clients it does not make sense to have the HttpClient interfaces in my mind.

joerg1985:
This test is using the NettyClient directly, this should be fixed:
https://github.com/SeleniumHQ/selenium/blob/0b5cf2d5a21e0c8f26cf34b4d619670a9bb888ba/java/test/org/openqa/selenium/remote/http/RetryRequestTest.java#L50

Does it make sense to have a bazel file in the jdk folder or could this just be part of the parent "../remote/http"?

java/src/org/openqa/selenium/remote/http/HttpClient.java
titusfortner: Good point. No reason not to just change the default, at least for now. I think this option has only been available since 4.5, though, and I'm guessing no one has implemented their own since then? I kind of think we should put it back to where it was 4.4, but of course I haven't actually looked at that code.

titusfortner:
> Does it make sense to have a bazel file in the jdk folder

I don't know? I mostly copied what I saw Netty had done in the places that seemed to make sense. If you think we should change something, we probably should because I don't really know what I'm changing. 😆

joerg1985:
> > Does it make sense to have a bazel file in the jdk folder
> 
> I don't know? I mostly copied what I saw Netty had done in the places that seemed to make sense. If you think we should change something, we probably should because I don't really know what I'm changing. 😆

I am no bazel expert too, there might be a reason for it :D

joerg1985:
I did some changes to make the bazel build work, i hope this is okay for you

java/src/org/openqa/selenium/remote/http/HttpClient.java
shs96c: We've implemented at least three versions of the `HttpClient` (based on Apache HttpClient, the current async model, and the native JDK) Having a mechanism to switch implementations without needing to change much is a really nice feature.

shs96c:
> Does it make sense to have a bazel file in the jdk folder or could this just be part of the parent "../remote/http"?

The "bazel way" is a build file per java package, so the `jdk` folder should definitely have its own `BUILD.bazel` file.

shs96c:
The `NettyDomainSocketClient` needs to stay, as it's used for the Docker integration because the other `HttpClient` implementations can't use unix domain sockets.

java/src/org/openqa/selenium/remote/http/HttpClient.java
joerg1985: I restored it, there are also unit tests for this

joerg1985:
I had to remove the htmlunit-driver dependency to get rid of the asynchttpclient, this added a circular dependency what should be avoided. Beside some helper classes i could not find any use of the htmlunit-driver so i hope this is fine. Lets wait what the CI will tell us.

titusfortner:
> the other HttpClient implementations can't use unix domain sockets

Looks like we can transition those things when we require Java 17 next year: https://openjdk.org/jeps/380

titusfortner:
@rbri is htmlunit-driver also moving to Java 11 requirement for the net release? Will that address the circular dependencies?

diemol (commented):
I don't understand why we need to make so many changes in a PR for this. We have always mentioned that moving completely to Java 11 is an iteration process.

Why don't we change the default http client, target Java 11 and that's it. The rest can be iterated little by little. 

Long PRs are just too hard to review.

joerg1985:
> @rbri is htmlunit-driver also moving to Java 11 requirement for the net release? Will that address the circular dependencies?

The circular dependency is now gone due to removal of the htmlunit-driver dependency.

joerg1985:
@diemol you are right about large PRs, but i think this PR is now ready. The CI is failing but this would also happen if only the default client has been changed. There is only one additional change here, the removal of the old netty client and it's dependencies (by removing the htmlunit-driver support).

diemol:
However, almost all CI workflows in Java are failing. We cannot merge like that.

titusfortner:
Yeah, I didn't want to keep distributing a library with a vulnerability, which is why I removed the netty client instead of just changing the default. 

Thanks @joerg1985 for doing all the work to clean up after me. 😁 

I'll take a look at the failures tomorrow

joerg1985:
Okay then there should be single PRs created for all of the changes.

java/src/org/openqa/selenium/chrome/ChromeDriverInfo.java
```diff
@@ -43,17 +41,6 @@ public String getDisplayName() {
...
-      // Allowing any origin "*" through remote-allow-origins might sound risky but an attacker
-      // would need to know the port used to start DevTools to establish a connection. Given
-      // these sessions are relatively short-lived, the risk is reduced. Only set when the Java
-      // 11 client is not used.
-      return new ImmutableCapabilities(
-          CapabilityType.BROWSER_NAME,
-          CHROME.browserName(),
-          ChromeOptions.CAPABILITY,
-          ImmutableMap.of("args", ImmutableList.of("--remote-allow-origins=*")));
-    }
```
diemol, line 56: This is also present on Edge.

java/src/org/openqa/selenium/remote/http/netty/NettyDomainSocketClient.java
```diff
@@ -1,234 +0,0 @@
...
-import org.openqa.selenium.internal.Require;
-import org.openqa.selenium.remote.http.AddSeleniumUserAgent;
-import org.openqa.selenium.remote.http.ClientConfig;
-import org.openqa.selenium.remote.http.HttpClient;
-import org.openqa.selenium.remote.http.HttpRequest;
-import org.openqa.selenium.remote.http.HttpResponse;
-import org.openqa.selenium.remote.http.RemoteCall;
-import org.openqa.selenium.remote.http.WebSocket;
-
-class NettyDomainSocketClient extends RemoteCall implements HttpClient {
```
diemol: Why is this being removed? This is not using [AsyncHttpClient](https://github.com/AsyncHttpClient/async-http-client)

java/test/org/openqa/selenium/testing/TestUtilities.java
```diff
@@ -30,7 +30,6 @@
 import org.openqa.selenium.JavascriptExecutor;
 import org.openqa.selenium.Platform;
 import org.openqa.selenium.WebDriver;
-import org.openqa.selenium.htmlunit.HtmlUnitDriver;
```
diemol: Why do we need to remove this?

diemol (changes requested):

java/test/org/openqa/selenium/testing/TestUtilities.java
joerg1985: The bazel repin will not remove the asynchttpclient from the maven_install.json when the htmlunit-driver is referenced. I could not fond any use of the htmlunit-driver beside helpers. But yes this is not directly related to this PR.

@titusfortner I will create a new PR for removing the htmlunit-driver tomorrow to see if this is used anywhere.

joerg1985: PS: the htmlunit-driver does reference selenium as a dependency and selenium does reference htmlunit-driver as a dependency, what is usually not a good idea

titusfortner: I think Selenium should be a dependency and not the other way around. But might make sense to get @rbri input. I'm not sure why things are the way they are now.

joerg1985: #12849 it the PR for removing this, the CI on my fork is still passing without.
I think this has been used in the past and has been replaced with real browser tests.

java/src/org/openqa/selenium/chrome/ChromeDriverInfo.java
pujagani: Also present in ChromiumOptions.

java/test/org/openqa/selenium/remote/http/netty/NettyDomainSocketTest.java
```diff
@@ -1,29 +0,0 @@
...
-package org.openqa.selenium.remote.http.netty;
-
-import org.openqa.selenium.remote.http.HttpClient;
-import org.openqa.selenium.remote.internal.DomainSocketsTestBase;
-
-class NettyDomainSocketTest extends DomainSocketsTestBase {
-
-  @Override
-  protected HttpClient.Factory createFactory() {
-    return new NettyClient.Factory();
```
pujagani, line 27: Since NettyDomainSocket needs to stay. This test will need to be kept or updated accordingly.

pujagani (changes requested):

pujagani:
This also impacts the changes done as part of https://github.com/SeleniumHQ/selenium/issues/11606. Hence the failing "DevToolsReuseTest" test. Earlier, this was not a problem since we used one instance of AsyncHttpClient and it was only closed when the JVM shuts down. With the JdkHttpClient, we use a new instance as required and close it too. Changes in https://github.com/SeleniumHQ/selenium/pull/11345, which assigns the client to null is while closing causing the test failure and preventing re-use. It was done to prevent memory leaks. I am not sure about the right approach to address it. We might need to discuss/brainstorm over it.

diemol:
> This also impacts the changes done as part of #11606. Hence the failing "DevToolsReuseTest" test. Earlier, this was not a problem since we used one instance of AsyncHttpClient and it was only closed when the JVM shuts down. With the JdkHttpClient, we use a new instance as required and close it too. Changes in #11345, which assigns the client to null is while closing causing the test failure and preventing re-use. It was done to prevent memory leaks. I am not sure about the right approach to address it. We might need to discuss/brainstorm over it.

@pujagani is there a bug report for that? I can imagine that someone might bump into that situation when using the Java 11 client.

java/src/org/openqa/selenium/chrome/ChromeDriverInfo.java
titusfortner: Removed.

java/src/org/openqa/selenium/remote/http/netty/NettyDomainSocketClient.java
titusfortner: Because I thought this directory was all AHC, so I deleted the whole directory and left alone the grid/netty server & redis directories. I've put it back. :)

java/test/org/openqa/selenium/remote/http/netty/NettyDomainSocketTest.java
titusfortner: I put it back.

titusfortner:
Ok, the `NettyDomainSocketTest` calls `NettyClient.Factory()` which pretty much needs all of Netty plus AsyncHttpClient.

So I think we need a different test if we need a test for it.

pujagani:
@diemol I don't think there is one. Will double check else I will create it and link it here.

titusfortner:
This is passing all the tests and looks right to me. 

The only issue is that I removed `NettyDomainSocketTest` because it required AHC. If we need a test for that class, can someone please help add it?

pujagani:
https://github.com/SeleniumHQ/selenium/issues/12882 -  Created an issue to track the regression

pujagani:
> This is passing all the tests and looks right to me.
> 
> The only issue is that I removed `NettyDomainSocketTest` because it required AHC. If we need a test for that class, can someone please help add it?

I looked into it, but I don't think the current test can be modified without having another client library. Since Java HTTP Client does not support UNIX sockets.

rbri:
> > @rbri is htmlunit-driver also moving to Java 11 requirement for the net release? Will that address the circular dependencies?
> 
> The circular dependency is now gone due to removal of the htmlunit-driver dependency.

@joerg1985
first of all thanks for removing the outdated stuff. I think getting rid of the circular dependency is a step forward for selenium (see my comment on https://github.com/SeleniumHQ/selenium/pull/12849).

Regarding the htmlunit-driver i think i have no chance - i have to move to Java 11. I like to stay with the support of the latest selenium version.
For HtmlUnit itself i like to stay at java8 if possible but i will see.
From my point of view we can go ahead with this making a compatible htmlunit-driver is a task on my desk like all the other htmlunit related stuff (of course any help is welcome....)

titusfortner:
> Since Java HTTP Client does not support UNIX sockets.

But we're testing with java 17 and I think Java 17 does support it? Can we create an implementation in test code only?

pujagani:
> > Since Java HTTP Client does not support UNIX sockets.
> 
> But we're testing with java 17 and I think Java 17 does support it? Can we create an implementation in test code only?

Ahh yes! I think that can be done. I am not sure how can we create an implementation in test code only but will be happy to take a took.

PR #12847, opened 2023-10-01
Ensure Selenium Manager returns an exit code

shs96c:
Without this, the java process API will wait for the Selenium Manager to exit indefinitely.


shs96c:
I want to see whether or not this fixes some of the timeouts we're seeing in the RBE builds. If so, I think this is safe to land.

titusfortner:
Looks like this didn't do what you hoped? Still want to merge?

diemol:
This can probably be closed.

PR #12848, opened 2023-10-02
Prevent browser to be created and unused in case a short session-request-timeout is set

### LINKED ISSUE #12832: [🐛 Bug]: browser created but never used (same as #11881)
### What happened?

Hello, 

In april, I raised the issue #11881 about browser being created on selenium grid node but never used when request timeout is set.
This issue was corrected (I thought my correction was enough) but I still see the issue in an heavy loaded grid.
The point is that when requesting a new session, we do :

https://github.com/SeleniumHQ/selenium/blob/9a6947ea4fd685b5f997415565bc5dc875a71321/java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java#L216

which waits for X secs (the session request timeout)

Then (this was the original bug), if the session is being created on node, we wait a bit more because session request is not in the queue anymore, but session may not be created when timeout expires

So, to correct it, I added an other wait  of 5 seconds
https://github.com/SeleniumHQ/selenium/blob/9a6947ea4fd685b5f997415565bc5dc875a71321/java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java#L218

These 5 seconds were arbitrary choosen but adding some logs today, I've seen that driver creation takes more time
```
15:16:32.199 INFO [LocalDistributor.newSession] - 530153fd-a618-4418-8198-535a6c54117f: new session => start creating
15:16:38.766 INFO [LocalDistributor.newSession] - 530153fd-a618-4418-8198-535a6c54117f: new session => created
``` 
or even more
```
15:09:42.391 INFO [LocalDistributor.newSession] - 769d66eb-178a-4e45-8a8c-7f06c3f9177f: new session => start creating
15:09:51.892 INFO [LocalDistributor.newSession] - 769d66eb-178a-4e45-8a8c-7f06c3f9177f: new session => created
```
So 5 seconds is definitely not enough

A quick solution would be to increase the time on line (e.g: 15 seconds)
https://github.com/SeleniumHQ/selenium/blob/9a6947ea4fd685b5f997415565bc5dc875a71321/java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java#L218

But do you think there is an other way to correct the issue ? (fixed time waits are never the best solution 😕  )
Looking at the logs (added logs can be seen here: https://github.com/bhecquet/selenium/tree/driver_sessions_staled), I think that wait time should be increased, but also, we need to prevent distributor from taking the session request once first timeout expires

```shell
Set a grid with 1 hub, 1 node and try to create 4 browsers at the time
After some retries, a browser will be created without being used
```

### Relevant log output

```shell
>>> Client node
  WARN  2023-09-28 15:39:17,389 [TestNG-tests-1] SeleniumGridDriverFactory: Error creating driver on hub

>>> Hub
15:38:42.358 INFO [LocalNewSessionQueue.addToQueue] - bd2be918-88a3-41ef-809a-5e0c1f66ef5f: start waiting session creation
15:38:42.867 INFO [LocalDistributor.newSession] - bd2be918-88a3-41ef-809a-5e0c1f66ef5f: new session
15:38:44.400 INFO [LocalDistributor.reserveSlot] - No slots found for request bd2be918-88a3-41ef-809a-5e0c1f66ef5f and capabilities Capabilities ...
15:38:44.400 INFO [LocalDistributor.newSession] - Unable to find a free slot for request bd2be918-88a3-41ef-809a-5e0c1f66ef5f. 
...
15:39:11.001 INFO [LocalDistributor.reserveSlot] - No slots found for request bd2be918-88a3-41ef-809a-5e0c1f66ef5f and capabilities Capabilities ...
15:39:11.001 INFO [LocalDistributor.newSession] - Unable to find a free slot for request bd2be918-88a3-41ef-809a-5e0c1f66ef5f. 
15:39:11.002 WARN [SeleniumSpanExporter$1.lambda$export$1] - Will retry session bd2be918-88a3-41ef-809a-5e0c1f66ef5f
15:39:11.516 INFO [LocalDistributor.newSession] - bd2be918-88a3-41ef-809a-5e0c1f66ef5f: new session
15:39:12.361 INFO [LocalNewSessionQueue.addToQueue] - bd2be918-88a3-41ef-809a-5e0c1f66ef5f: a bit more time
15:39:12.548 INFO [LocalDistributor.reserveSlot] - No slots found for request bd2be918-88a3-41ef-809a-5e0c1f66ef5f and capabilities Capabilities {a ....
15:39:12.548 INFO [LocalDistributor.newSession] - Unable to find a free slot for request bd2be918-88a3-41ef-809a-5e0c1f66ef5f. 
15:39:12.549 WARN [SeleniumSpanExporter$1.lambda$export$1] - Will retry session bd2be918-88a3-41ef-809a-5e0c1f66ef5f
15:39:12.551 INFO [LocalDistributor.newSession] - bd2be918-88a3-41ef-809a-5e0c1f66ef5f: new session
15:39:13.580 INFO [LocalDistributor.newSession] - bd2be918-88a3-41ef-809a-5e0c1f66ef5f: new session => start creating
15:39:17.362 INFO [LocalNewSessionQueue.addToQueue] - bd2be918-88a3-41ef-809a-5e0c1f66ef5f: session not created
15:39:20.706 INFO [LocalDistributor.newSession] - bd2be918-88a3-41ef-809a-5e0c1f66ef5f: new session => created



>>> Node
15:39:13.585 INFO [LocalNode.newSession] - new session => start creating
15:39:20.695 INFO [LocalNode.newSession] - Session created by the Node. Id: 183f4330244d3d4a34a68313e7434dd9, Caps: Capabilities {acceptInsecureCerts: true, browserName: chrome, browserVersion: 114.0.5735.199, chrome: {chromedriverVersion: 114.0.5735.90 (386bc09e8f4f..., userDataDir: C:\Users\xxx\AppData\Lo...}, goog:chromeOptions: {debuggerAddress: localhost:63241}, networkConnectionEnabled: false, pageLoadStrategy: normal, platformName: any, proxy: Proxy(autodetect), se:bidiEnabled: false, se:cdp: ws://host.comp.any..., se:cdpVersion: 114.0.5735.199, setWindowRect: true, sr:beta: false, sr:testName: login, sr:try: 0, strictFileInteractability: false, timeouts: {implicit: 0, pageLoad: 300000, script: 30000}, unhandledPromptBehavior: dismiss and notify, webauthn:extension:credBlob: true, webauthn:extension:largeBlob: true, webauthn:extension:minPinLength: true, webauthn:extension:prf: true, webauthn:virtualAuthenticators: true}
```

### Operating System

Windows / Linux

### Selenium version

4.13.0

not related to browser

not related to browser

4.13.0

bhecquet:
### Description
This PR corrects the issue #12832 where driver sessions may be created, but unused if driver takes long time to create
In out case, session request timeout is set to 30 secs, as we have several hubs for redundancy, and we do not want to be stuck with a single hub on driver creation.

### Motivation and Context
When hubs are loaded, slots take some time to free, so that a new session can be created, and in this case, request session timeout may be raised during driver creation. As a result, driver is created on node, but client has received a SessionNotCreatedException

In this PR, after a driver is successfuly created, we check the session queue to see if the session request has expired or not. If it has expired (not in requests list or has reached timeout), then, we notify the node to stop the session immediately to release the slot

### Types of changes

### Checklist

I would have liked to add a unit test for LocalDistributor part, but could not find a proper way to do this. If you have any idea ...

Fixes #12832


joerg1985:
I think there might be an easier way to handle this cases, just return a boolean in `NewSessionQueue.complete`?

```
  /**
   * @return true if the queue accepted the session/exception, false if the session was not accepted and should be terminated
   * (usually when the session request hit the timeout before and was rejected)
   */
  public abstract boolean complete(
      RequestId reqId, Either<SessionNotCreatedException, CreateSessionResponse> result);
```

The `LocalDistributor` could then handle it this way:

```
        boolean accepted = sessionQueue.complete(reqId, response);
        attributeMap.put("request.accepted", accepted);

        if (!accepted && response.isRight()) {
          terminateSession(response.right().getSession().getId());
        }
```

Only a idea, i am not telling the current PR way is bad :D

bhecquet:
I agree with you, there will be less changes in code
Will change my PR accordingly

java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
```diff
@@ -290,9 +278,15 @@ public boolean retryAddToQueue(SessionRequest request) {
...
+
         if (!requests.containsKey(request.getRequestId())) {
           return false;
-        }
+        } else if (isTimedOut(Instant.now(), requests.get(request.getRequestId()))) {
+          // as we try to re-add a session request that has already expired, force session timeout
+          complete(request.getRequestId(), Either.left(new SessionNotCreatedException("Timed out creating session")));
+          return false;
+        } 
+
```
joerg1985: I don't think this check is needed, the regular timeout handling will pick this up when it is again in the queue.

java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
```diff
@@ -233,15 +230,6 @@ public HttpResponse addToQueue(SessionRequest request) {
...
 
-      Lock writeLock = this.lock.writeLock();
-      writeLock.lock();
-      try {
-        requests.remove(request.getRequestId());
-        queue.remove(request);
-      } finally {
-        writeLock.unlock();
-      }
-
```
joerg1985: I think this should not have been removed?

java/src/org/openqa/selenium/grid/sessionqueue/remote/RemoteNewSessionQueue.java
```diff
@@ -171,7 +171,12 @@ public void complete(
     }
 
     HttpTracing.inject(tracer, tracer.getCurrentContext(), upstream);
-    client.with(addSecret).execute(upstream);
+    HttpResponse response = client.with(addSecret).execute(upstream);
+    if (response.isSuccessful()) {
+      return Values.get(response, Boolean.class);
+    } else {
+      return false;
```
joerg1985: `return Values.get(response, Boolean.class);` should be enough here, there is an internal error handling in `Values.get`

java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
```diff
@@ -838,8 +840,32 @@ private void handleNewSessionRequest(SessionRequest sessionRequest) {
...
         }
 
-        sessionQueue.complete(reqId, response);
+        boolean isSessionValid = sessionQueue.complete(reqId, response);
+        // is session request has timed out, tell the node to remove the session, so that it's not staled
+        if (!isSessionValid && response.isRight()) {
+          LOG.info("Stopping expired session");
+          URI nodeURI = response.right().getSession().getUri();
+          Node node = getNodeFromURI(nodeURI);
+          node.stop(response.right().getSession().getId());
```
joerg1985: Sending the DELETE session request should close the browser e.g.:

https://github.com/SeleniumHQ/selenium/blob/2ffb772261814d36375f210033bd5ec9f78663e0/java/src/org/openqa/selenium/grid/node/local/LocalNode.java#L304-L307

java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
```diff
@@ -214,9 +214,6 @@ public HttpResponse addToQueue(SessionRequest request) {
       try {
 
         boolean sessionCreated = data.latch.await(requestTimeout.toMillis(), MILLISECONDS);
-        if (!(sessionCreated || isRequestInQueue(request.getRequestId()))) {
```
joerg1985, line 217: `isRequestInQueue` should be unused now an could be removed

java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
```diff
@@ -388,14 +385,17 @@ public void complete(
...
         data = requests.get(reqId);
       } finally {
         readLock.unlock();
       }
 
       if (data == null) {
-        return;
+        return false;
+      } else {
+        isSessionTimedOut = isTimedOut(Instant.now(), data);
```
joerg1985, line 392: The regular timeout handling should already have rejected the session so it is not in the `requests` any more.
So we will never hit this check and it can be removed.

java/src/org/openqa/selenium/grid/sessionqueue/SessionNotCreated.java
```diff
@@ -54,7 +56,7 @@ public HttpResponse execute(HttpRequest req) throws UncheckedIOException {
       SessionNotCreatedException exception = new SessionNotCreatedException(message);
       queue.complete(requestId, Either.left(exception));
 
-      HttpResponse res = new HttpResponse();
+      HttpResponse res = new HttpResponse().setContent(asJson(singletonMap("value", false)));
```
joerg1985: It might be better to pass the return value of `queue.complete` to the response to avoid having a hard coded value.

java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
bhecquet: This check is needed because the "timeoutSessions" method only runs every 10 secs (per configuration).
If we remove it, LocalDistributor may try to recreate the session while it's really timed out

java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
bhecquet: I removed it because, on success, or failure, the "complete" method will be called (or the timeoutSessions), and so, queue and requests will be cleaned.
Moreover, here, it seems `contexts.remove(reqId)` is missing

java/src/org/openqa/selenium/grid/sessionqueue/remote/RemoteNewSessionQueue.java
bhecquet: Ok, was not sure about it, I will change it

java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
bhecquet: So, I don't understand you well, node.stop() does the same thing (at least, I see the browser close :) )

java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
bhecquet: Yes, removed it

java/src/org/openqa/selenium/grid/sessionqueue/SessionNotCreated.java
bhecquet: I put a hard coded value because the value did not matter
But, You're right, better to pass the method reply

java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
bhecquet: Same comment as above, regular timeout handling may not have run here (every 10 secs by default)

java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
joerg1985: Okay, sorry for the confusion.

java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
joerg1985: The `contexts.remove(reqId)` is intentionally missing here to allow `retryAddToQueue` reading it.
If it is safe to remove the other code i am not sure, someone else has to review.

java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
joerg1985: I think the default value is 1s so and it should be fine*, anyways i don't think this should not be done here.
This is a "atomic" check, when there is 1ms time left before hitting the timeout it will pass and you have to wait for the regular timeout logic to happen.
Just trying to keep the code here as simple as possible as it is already complex, but there might be other opinions. @diemol what do you think about this?

\* Why did you set this timeout so hight in your configuration? 
I am trying to understand / find the root cause for this, to find a good solution to this.

java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
```diff
@@ -560,6 +561,7 @@ public Either<SessionNotCreatedException, CreateSessionResponse> newSession(
             new CreateSessionRequest(request.getDownstreamDialects(), caps, request.getMetadata());
 
         try {
+
```
diemol: Please remove this extra line. It just adds to the changes and demands more time to the PR review.

java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
```diff
@@ -838,8 +840,32 @@ private void handleNewSessionRequest(SessionRequest sessionRequest) {
           }
         }
 
-        sessionQueue.complete(reqId, response);
+        boolean isSessionValid = sessionQueue.complete(reqId, response);
```
diemol, line 844: `isSessionValid` is not perhaps an accurate name. This is when a session gets created on the Grid but has no client, so probably `orphaned` is better.

diemol: Or `isSessionStalled`

java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
```diff
@@ -838,8 +840,32 @@ private void handleNewSessionRequest(SessionRequest sessionRequest) {
...
+        // is session request has timed out, tell the node to remove the session, so that it's not staled
```
diemol:
```suggestion
        // If the session request has timed out, tell the Node to remove the session so it does not stall.
```

java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
```diff
@@ -838,8 +840,32 @@ private void handleNewSessionRequest(SessionRequest sessionRequest) {
...
+        if (!isSessionValid && response.isRight()) {
+          LOG.info("Stopping expired session");
```
diemol: Please add something more descriptive to this message. 
At least a brief explanation of what this means. 

Also, do you know if the user needs to know this? Is `INFO` the right logging level?

java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
```diff
@@ -838,8 +840,32 @@ private void handleNewSessionRequest(SessionRequest sessionRequest) {
...
+          URI nodeURI = response.right().getSession().getUri();
+          Node node = getNodeFromURI(nodeURI);
```
diemol, line 854: What happens if `node` is `null`?

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
```diff
@@ -401,7 +401,6 @@ public boolean isSupporting(Capabilities capabilities) {
   public Either<WebDriverException, CreateSessionResponse> newSession(
       CreateSessionRequest sessionRequest) {
     Require.nonNull("Session request", sessionRequest);
-
```
diemol: Please add back this line. It just adds to the changes and demands more time to the PR review.

java/src/org/openqa/selenium/grid/sessionqueue/SessionCreated.java
```diff
@@ -50,9 +52,10 @@ public HttpResponse execute(HttpRequest req) throws UncheckedIOException {
       HTTP_REQUEST.accept(span, req);
 
       CreateSessionResponse response = Contents.fromJson(req, CreateSessionResponse.class);
-      queue.complete(requestId, Either.right(response));
+      boolean isSessionValid = queue.complete(requestId, Either.right(response));
+
+      HttpResponse res = new HttpResponse().setContent(asJson(singletonMap("value", isSessionValid)));
```
diemol: Same comment as above. I don't think `isSessionValid` is the correct name.

java/src/org/openqa/selenium/grid/sessionqueue/SessionNotCreated.java
```diff
@@ -52,9 +54,9 @@ public HttpResponse execute(HttpRequest req) throws UncheckedIOException {
 
       String message = Contents.fromJson(req, String.class);
       SessionNotCreatedException exception = new SessionNotCreatedException(message);
-      queue.complete(requestId, Either.left(exception));
+      boolean isSessionValid = queue.complete(requestId, Either.left(exception));
 
-      HttpResponse res = new HttpResponse();
+      HttpResponse res = new HttpResponse().setContent(asJson(singletonMap("value", isSessionValid)));
```
diemol: Same comment as above. I don't think `isSessionValid` is the correct name.

java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
```diff
@@ -93,7 +94,6 @@
     objectName = "org.seleniumhq.grid:type=SessionQueue,name=LocalSessionQueue",
     description = "New session queue")
 public class LocalNewSessionQueue extends NewSessionQueue implements Closeable {
-
```
diemol: Please add back this line. It just adds to the changes and demands more time to the PR review.

java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
diemol: @joerg1985 is correct. This cannot be removed, otherwise it will duplicate requests when retrying the session creation.

java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
```diff
@@ -290,9 +278,15 @@ public boolean retryAddToQueue(SessionRequest request) {
       Lock writeLock = lock.writeLock();
       writeLock.lock();
       try {
+
```
diemol: Please remove this extra line. It just adds to the changes and demands more time to the PR review.

java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
diemol: The check runs every 10 seconds, so there is a chance of retying a timed-out request.

It does make sense to check if the request has timed out before retrying it. 

However, if you return `true`, `handleNewSessionRequest` in the `LocalDistributor` will invoke `complete` again, which should not cause any issues, but it is just an unneeded call. So maybe return `true` and add a comment about it. 

There is no need for an `else if`, and `if` is enough. Finally, you can reuse the existing `failDueToTimeout` method.

java/test/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueueTest.java
```diff
@@ -166,6 +167,97 @@ private void waitUntilAddedToQueue(SessionRequest request) {
...
+
+              isCompleted.set(queue.complete(sessionRequest.getRequestId(), Either.right(sessionResponse)));
+            })
+        .start();
+
+    HttpResponse httpResponse = queue.addToQueue(sessionRequest);
+    try {
+      Thread.sleep(400); // wait for thread to complete
+    } catch (InterruptedException ignore) {
+    }
```
diemol: Please use a `CountDownLatch` instead of a `Thread.sleep`. Other tests can serve you as example.

java/test/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueueTest.java
```diff
@@ -166,6 +167,97 @@ private void waitUntilAddedToQueue(SessionRequest request) {
...
+
+              isCompleted.set(queue.complete(sessionRequest.getRequestId(), Either.left(new SessionNotCreatedException("not created"))));
+            })
+        .start();
+
+    HttpResponse httpResponse = queue.addToQueue(sessionRequest);
+    try {
+      Thread.sleep(400); // wait for thread to complete
+    } catch (InterruptedException ignore) {
+    }
```
diemol: Please use a `CountDownLatch` instead of a `Thread.sleep`. Other tests can serve you as an example.

diemol (changes requested):
Thank you for your PR. 

I am missing a test that also checks that the session is removed from the Node. You could add it in `DistributorTest`.

java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
bhecquet: ok

java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
bhecquet: ok, done as you suggested

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
bhecquet: done

java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
bhecquet: ok

java/src/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueue.java
bhecquet: ok, you are the expert :)

java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
bhecquet: In this case, I should invert the logic, as for now, complete returns true is session creation has been completed.
With such name, saying that complete returns false when everything runs fine bothers me.

Would `isNotStalled` or `isNotTimedout` be ok for you ?

java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
bhecquet: ok, done

java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
bhecquet: ok, level changed to debug and message improved

java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
bhecquet: a beautiful NPE ...
Ok, I correct it

java/test/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueueTest.java
bhecquet: ok, thanks

java/test/org/openqa/selenium/grid/sessionqueue/local/LocalNewSessionQueueTest.java
bhecquet: corrected

java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
diemol: Good point. Then let's leave `isSessionValid`. But please add a comment above the variable indicating what a "valid session" means in this context.

diemol: Feel free to resolve the conversation when the comment is added.

java/src/org/openqa/selenium/grid/sessionqueue/SessionCreated.java
diemol: Feel free to resolve the conversation when the comment explaining what a valid session is added.

java/src/org/openqa/selenium/grid/sessionqueue/SessionNotCreated.java
diemol: Feel free to resolve the conversation when the comment explaining what a valid session is added.

diemol (changes requested):

java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java
bhecquet: ok done

java/src/org/openqa/selenium/grid/sessionqueue/SessionNotCreated.java
bhecquet: done

pujagani:
@bhecquet  Probably to verify the Distributor changes via tests, essentially to check if this works end to end as expected, have a look at the setup in https://github.com/SeleniumHQ/selenium/blob/trunk/java/test/org/openqa/selenium/grid/router/SessionQueueGridTest.java. The end-to-end tests there try to valid different scenarios involving the new session queue. Maybe add to the same test or create another test class with a similar setup to emulate the motivation the changes solve.

diemol (approved):
Thank you, @bhecquet!

PR #12849, opened 2023-10-02
[java] remove the unused htmlunit-driver from the unit tests

joerg1985:
### Description
This change will remove the htmlunit-driver dependency and all references in the unit tests to the htmlunit-driver.
There is functional no change, only testing related changes.

### Motivation and Context
The CI is not running any htmlunit-driver tests, it might have done this in the past.
A lot of tests were disabled with NotYetImplemented/Ignore annotations for the htmlunit-driver. 

There are a lot of unused dependencies from the htmlunit-driver in the build pipeline.
Even worse, the htmlunit-driver has a dependency back to selenium what should be avoided.

### Types of changes

### Checklist


diemol (approved):
Thank you, @joerg1985!

rbri:
Thanks @joerg1985,
as discussed last year in Berlin with @diemol it's fine to do that.
A branch of the test suite was used to test the htmlunit-driver in the past, but this branch was not updated since years. And because i have no time to maintain this all i decide some time ago to use a simpler test suite for the htmlunit-driver. And also HtmlUnit itself uses a similar approach to (with a circular dependency between HtmlUnit and the htmlunit-driver :-)). This is also some kind of test so there is no real need to have HtmlUnit in the selenium test suite.

PR #12853, opened 2023-10-02
[java] ensure proper error message gets logged

titusfortner:
Java Throwable class (at least for Java 11) does not set the detailMessage in the `initCause` method the same way it does in the way it does in the constructor:

```
    public Throwable(Throwable cause) {
        fillInStackTrace();
        detailMessage = (cause==null ? null : cause.toString());
        this.cause = cause;
    }
```
vs

```
    public synchronized Throwable initCause(Throwable cause) {
        if (this.cause != this)
            throw new IllegalStateException("Can't overwrite cause with " +
                                            Objects.toString(cause, "a null"), this);
        if (cause == this)
            throw new IllegalArgumentException("Self-causation not permitted", this);
        this.cause = cause;
        return this;
    }
```

### DIFF
--- a/java/src/org/openqa/selenium/remote/http/WebSocket.java
+++ b/java/src/org/openqa/selenium/remote/http/WebSocket.java
@@ -65,7 +65,7 @@ public interface WebSocket extends Closeable {
 
     default void onError(Throwable cause) {
       String message = cause.getMessage();
-      if (message == null && cause.getCause().getMessage() != null) {
+      if (message == null && cause.getCause() != null) {
         message = cause.getCause().getMessage();
       }
       LOG.log(WARNING, message, cause);
--- a/java/src/org/openqa/selenium/remote/http/WebSocket.java
+++ b/java/src/org/openqa/selenium/remote/http/WebSocket.java
@@ -64,7 +64,11 @@ public interface WebSocket extends Closeable {
     }
 
     default void onError(Throwable cause) {
-      LOG.log(WARNING, cause.getMessage(), cause);
+      String message = cause.getMessage();
+      if (message == null && cause.getCause().getMessage() != null) {
+        message = cause.getCause().getMessage();
+      }
+      LOG.log(WARNING, message, cause);
     }
   }
 }


java/src/org/openqa/selenium/remote/http/WebSocket.java
```diff
@@ -64,7 +64,11 @@ default void onText(CharSequence data) {
     }
 
     default void onError(Throwable cause) {
-      LOG.log(WARNING, cause.getMessage(), cause);
+      String message = cause.getMessage();
+      if(message == null && cause.getCause().getMessage() != null) {
```
diemol:
```suggestion
      if (message == null && cause.getCause().getMessage() != null) {
```

diemol (changes requested):

java/src/org/openqa/selenium/remote/http/WebSocket.java
diemol: Can `cause.getCause()` be `null`?

titusfortner: Maybe? It doesn't error, it just doesn't log anything.

titusfortner: oh. `getCause()`, misunderstood because was reading from my phone. Might as well protect against it.

pujagani:
Good observation. This is a good start. I think similar changes might be required in other places as well. Example: https://github.com/SeleniumHQ/selenium/blob/c64ed7d067f8314b2d3fbe3600dcc1ed47b866a4/java/src/org/openqa/selenium/devtools/DevToolsException.java#L25, https://github.com/SeleniumHQ/selenium/blob/c64ed7d067f8314b2d3fbe3600dcc1ed47b866a4/java/src/org/openqa/selenium/bidi/BiDiException.java#L25

titusfortner:
@diemol I updated the code to make sure `getCause()` isn't null. It doesn't matter if `getMessage()` is null, it will just end up in the same place, but if `getCause()` were null for some reason, then... NPE.

titusfortner:
@pujagani I wanted to keep the changes limited to the one place I saw a specific problem (missing info in the logs I'm looking at). The code is still the same in Java 21, which leads me to believe it is intentional? I raised a ticket with Oracle to get their response.

> We will review your report and have assigned it an internal review ID : 9076067. Depending upon the completeness of the report and our ability to reproduce the problem, either a new bug will be posted, or we will contact you for further information.

🤷

diemol (approved):

PR #12866, opened 2023-10-04
[java] reduce log noise at FINE level

titusfortner:
### Description
* Change `W3CHttpResponseCodec.decode()` & `DumpHttpExchangeFilter` log level to FINER
* Add PrintPage & fullPageScreenshot to the list of commands that does not log the response (since they are all base64 encoded responses)
* `NEW_SESSION` responses are parsed with Capabilities.toString() 

### Motivation and Context
* RemoteWebDriver logs most of the important information and intelligently removes responses with Base64 encoded values; `W3CHttpResponseCodec.decode` & `DumpHttpExchangeFilter` do not add much information and they destroy the logs.
* `NEW_SESSION` response can have Base64 values; the `Capabilities` classes have a complicated parser to truncate nested values that are > 100 characters. We should use that in the responses as well. I was going to put this behind `webdriver.remote.shorten_log_messages` system property, but if you want the full values, people can switch to `FINEST`.

So, this is the first time I realized we have `Debug.getDebugLogLevel()`. 
1. I don't think I like magically toggling `FINER` messages to `INFO` [when anyone runs anything in debug mode](https://github.com/SeleniumHQ/selenium/blob/selenium-4.13.0/java/src/org/openqa/selenium/internal/Debug.java#L28-L36). Really feels like we should let users do the more Java thing here. Who is this for? Why not just set the logging output for the user if they request it? Or set this in our tests instead of all of Selenium?
2. How about a system property of `selenium.webdriver.logger.level`?
3. If we do use this, we should be using it everywhere we are otherwise using `FINER`

### DIFF
--- a/java/src/org/openqa/selenium/remote/RemoteWebDriver.java
+++ b/java/src/org/openqa/selenium/remote/RemoteWebDriver.java
@@ -750,11 +750,12 @@ public class RemoteWebDriver
     }
     // No need to log a screenshot response.
     if ((commandName.equals(DriverCommand.SCREENSHOT)
-            || commandName.equals(DriverCommand.ELEMENT_SCREENSHOT))
+            || commandName.equals(DriverCommand.ELEMENT_SCREENSHOT)
+            || commandName.equals(DriverCommand.PRINT_PAGE))
         && toLog instanceof Response) {
       Response responseToLog = (Response) toLog;
       Response copyToLog = new Response(new SessionId((responseToLog).getSessionId()));
-      copyToLog.setValue("*Screenshot response suppressed*");
+      copyToLog.setValue(String.format("*%s response suppressed*", commandName));
       copyToLog.setStatus(responseToLog.getStatus());
       copyToLog.setState(responseToLog.getState());
       text = String.valueOf(copyToLog);
--- a/java/src/org/openqa/selenium/remote/RemoteWebDriver.java
+++ b/java/src/org/openqa/selenium/remote/RemoteWebDriver.java
@@ -742,21 +742,33 @@ public class RemoteWebDriver
       return;
     }
     String text = String.valueOf(toLog);
-    if (commandName.equals(DriverCommand.EXECUTE_SCRIPT)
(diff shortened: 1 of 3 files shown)


pujagani:
The changes made here, based on the motivation look good to me. 
I don't think I understand "I don't think I like magically toggling FINER messages to INFO". How are we doing that? If we are doing this, then I agree with you that it should be in user's control.

pujagani (approved):

titusfortner:
So, for instance, we use `getDebugLogLevel()` in many places in the code (e.g., [BiDi Connection](https://github.com/SeleniumHQ/selenium/blob/selenium-4.13.0/java/src/org/openqa/selenium/bidi/Connection.java#L145))

This looks to see if the code is being run in debug mode (e.g. IntelliJ debugger), or if system properties are set — https://github.com/SeleniumHQ/selenium/blob/selenium-4.13.0/java/src/org/openqa/selenium/internal/Debug.java#L35

If those conditions are true, it automatically assigns the level for the logs to `INFO` instead of `FINE` — https://github.com/SeleniumHQ/selenium/blob/selenium-4.13.0/java/src/org/openqa/selenium/internal/Debug.java#L47


In some respects I like that it makes it easy to get logs. 

Telling users to add:
```
System.setProperty("selenium.debug", "true")
```
Is easier than asking them to follow: https://www.selenium.dev/documentation/webdriver/troubleshooting/logging/

But it's ... unexpected and a little creepy. I knew we were auto turning on more logging when the tests were run, I thought it was in our test code, not in Selenium itself. I can see wanting to debug my test code without having all the Selenium noise (especially when running via Server). But also, most Java devs aren't used to JUL, so we aren't doing the most common thing, either.

Really we need to use `getDebugLogLevel()` everywhere or nowhere instead of mixing like we have now.

pujagani:
Thank you for elaborating. I appreciate that. I understand this better and agree with you. Maybe we can track this as a separate issue.

pujagani:
Thank you for explaining it in detail. I agree with your point here. Using it only for the test good makes sense I think.

PR #12873, opened 2023-10-05
[dotnet] Improve nuget packages metadata

### LINKED ISSUE #12709: [🚀 Feature]: Publish release notes in Nuget packages
### Feature and motivation

There are no release notes for the nuget packages [Selenium.WebDriver](https://www.nuget.org/packages/Selenium.WebDriver) and [Selenium.Support](https://www.nuget.org/packages/Selenium.Support)

This requires a <PackageReleaseNotes> property in the .csproj files, which most likely can be automated to:
`
<PackageReleaseNotes>$([System.IO.File]::ReadAllText("$(MSBuildProjectDirectory)/../../CHANGELOG"))</PackageReleaseNotes>
`

nvborisenko:
### Description
- add `readme.md`
- add release notes
- change bazel targets dependencies (on behalf of @shs96c)

Fixes https://github.com/SeleniumHQ/selenium/issues/12709

### Types of changes

### Checklist


titusfortner:
why the switch from windows to linux?

nvborisenko:
> why the switch from windows to linux?

Because we can, why not. I thought linux is "cheaper".

nvborisenko:
On linux we are getting `System.IO.IOException: Read-only file system : '/home/runner/.local/share/NuGet'`. So reverting this change back to run on Windows. I think we can fix it later.

titusfortner:
If we're running rbe on Linux, I think it's good to have this run on windows anyway.

dotnet/src/support/WebDriver.Support.StrongNamed.nuspec
```diff
@@ -4,35 +4,28 @@
...
     <projectUrl>https://selenium.dev</projectUrl>
-    <repository url="https://github.com/SeleniumHQ/selenium" />
+    <repository type="GitHub" url="https://github.com/SeleniumHQ/selenium" />
     <license type="expression">Apache-2.0</license>
     <iconUrl>https://selenium.dev/images/selenium_logo_square_green.png</iconUrl>
-    <icon>images\icon.png</icon>
+    <icon>icon.png</icon>
     <tags>selenium webdriver support browser automation</tags>
+    <releaseNotes>
+      See the latest changes, included in this release, at https://raw.githubusercontent.com/SeleniumHQ/selenium/trunk/dotnet/CHANGELOG.
```
titusfortner: is there a reason to prefer githubusercontent over github?
Also, can we use `$version$` here? We could use the tag instead of the branch in case we change the default branch at some point. The tag won't change and it will show the exact updates for the release:
e.g., https://github.com/SeleniumHQ/selenium/tree/selenium-4.13.0/dotnet/CHANGELOG

titusfortner:
Do we need to update this section?
https://github.com/SeleniumHQ/selenium/wiki/Releasing-Selenium#net

Should I try to build a nuget on my machine from this PR before we merge it? 😄

dotnet/src/support/WebDriver.Support.StrongNamed.nuspec
nvborisenko: > is there a reason to prefer githubusercontent over github?

To get plain text without GitHub UI. I have no preferences here.

Using `version` is very good idea. It is not clear how to deal with patches. This is list of tags:
```
selenium-4.12.0
selenium-4.12.1-dotnet
selenium-4.12.1-java
selenium-4.12.2-dotnet
selenium-4.12.3-dotnet
selenium-4.12.4-dotnet
```

I think we can mark release commit with `selenium-4.12.0-dotnet` tag to have consistent url.

nvborisenko:
Releasing process should stay the same. Eventually bazel targets were renamed from `//dotnet/src/webdriver:package` to `//dotnet/src/webdriver:webdriver-pack`. @shs96c is there any reason why we did it? I's more important - can we return target names back?

titusfortner:
Hmm, .NET tests are failing for this, @nvborisenko can you look when you get a chance?

nvborisenko:
It was failing immediately after we applied new bazel rules for dotnet. I remember about tests, hoping @shs96c will help us, in next stage.

titusfortner:
Things are passing on trunk. Something in this PR is an issue 😦 
Can we break it up?

nvborisenko:
Things are not passing on `trunk`:
```
ERROR: Skipping '//dotnet/test/common:chrome': no such target '//dotnet/test/common:chrome': target 'chrome' not declared in package 'dotnet/test/common' defined by D:/a/selenium/selenium/dotnet/test/common/BUILD.bazel (Tip: use `query "//dotnet/test/common:*"` to see all the targets in that package)
WARNING: Target pattern parsing failed.
ERROR: no such target '//dotnet/test/common:chrome': target 'chrome' not declared in package 'dotnet/test/common' defined by D:/a/selenium/selenium/dotnet/test/common/BUILD.bazel (Tip: use `query "//dotnet/test/common:*"` to see all the targets in that package)
INFO: Elapsed time: 21.481s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (1 packages loaded)
```

titusfortner:
The latest RBE on trunk is passing — https://github.com/SeleniumHQ/selenium/actions/runs/6436104331/job/17478770284
The latest CI with .NET on trunk is passing — https://github.com/SeleniumHQ/selenium/actions/runs/6435011324

nvborisenko:
CI - RBE / Test / All RBE tests is passing now.
The latest CI with .NET on trunk is always failing, even if shown green check mark. If you follow by the link you provided, you will see:
```
ERROR: Skipping '//dotnet/test/common:chrome': no such target '//dotnet/test/common:chrome': target 'chrome' not declared in package 'dotnet/test/common' defined by D:/a/selenium/selenium/dotnet/test/common/BUILD.bazel (Tip: use `query "//dotnet/test/common:*"` to see all the targets in that package)
```

Executing tests on github actions is different topic, not related to this PR.

titusfortner:
> The latest CI with .NET on trunk is always failing

That's suboptimal

dotnet/src/webdriver/assets/Selenium.WebDriver.targets
nvborisenko, line 1: don't forget to take the latest from trunk

titusfortner: Oh, I forgot about this one. It might not work with the change I made yesterday, let me see.

nvborisenko: It was a comment for myself :) I am going revert bazel related stuff in this PR, and keep here changes truly related to the topic. PR is not ready for review yet until CI build is green. Will notify you.

titusfortner: Thank you.

dotnet/src/support/WebDriver.Support.StrongNamed.nuspec
nvborisenko: @titusfortner your opinion is needed here.

dotnet/src/webdriver/assets/Selenium.WebDriver.targets
nvborisenko: Perfect, git is smart enough to determine the content of final file. Resolved.

nvborisenko:
Ok, CI build is failing, I guess something related to cache, and I have no experience in bazel even to try fixing it. In any case it is out of scope for this PR.

Last open question to move further: what is url to release notes should be (see Unresolved conversations). Thank you.

dotnet/src/support/WebDriver.Support.StrongNamed.nuspec
titusfortner: minor versions are when we release all bindings at the same time. The patch versions are when we need to do language specific versions. We used not to tag them at all, but we were going 4-5 months between releases for a bit so it was more important to tag when we made updates to things like CDP versions in each language.

titusfortner: I would prefer we use github directly so that people have context and can navigate around if the need to for any reason. Raw github is better for copying code directly, which I don't think is what we want to direct people to do.

nvborisenko: We decided to use static link to trunk, at least for now.

PR #12878, opened 2023-10-05
[dotnet] Deprecate saving screenshots with different image formats

nvborisenko:
### Description
This is the 1st phase of out journey to remove `System.Drawing.Common` dependency.

### Motivation and Context
Any great library should keep number of dependencies minimized as possible.

### Types of changes

### Checklist

### DIFF
--- a/dotnet/src/webdriver/Screenshot.cs
+++ b/dotnet/src/webdriver/Screenshot.cs
@@ -26,6 +26,7 @@ namespace OpenQA.Selenium
     /// <summary>
     /// File format for saving screenshots.
     /// </summary>
+    [Obsolete("Selenium doesn't support decoding/encoding between different image formats.")]
     public enum ScreenshotImageFormat
     {
         /// <summary>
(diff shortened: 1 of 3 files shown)


nvborisenko:
Enabler for https://github.com/SeleniumHQ/selenium/pull/12781

dotnet/src/webdriver/Screenshot.cs
```diff
@@ -26,6 +26,7 @@ namespace OpenQA.Selenium
     /// <summary>
     /// File format for saving screenshots.
     /// </summary>
+    [Obsolete("Selenium doesn't support decoding/encoding between different image formats.")]
```
titusfortner:
```suggestion
    [Obsolete("Support for decoding/encoding between different image formats is deprecated")]
```

dotnet/src/webdriver/Screenshot.cs
```diff
@@ -84,6 +85,7 @@ public override void SaveAsFile(string fileName)
         /// <param name="fileName">The full path and file name to save the screenshot to.</param>
         /// <param name="format">A <see cref="ScreenshotImageFormat"/> value indicating the format
         /// to save the image to.</param>
+        [Obsolete("Selenium doesn't support decoding/encoding between different image formats. Use \"SaveAsFile(string fileName)\" method instead.")]
```
titusfortner:
```suggestion
        [Obsolete("Support for decoding/encoding between different image formats is deprecated; use SaveAsFile(string fileName) method instead")]
```

titusfortner (approved):
I offered suggestions that match how we say it in Ruby, feel free to keep or change.

PR #12879, opened 2023-10-05
[dotnet] Format dotnet code according to default config

nvborisenko:
### Description
Fix small linter issues in `dotnet/src` folder.

### Motivation and Context
Keep code "clean".

### Types of changes

### Checklist

### DIFF
--- a/dotnet/src/support/Events/FindElementEventArgs.cs
+++ b/dotnet/src/support/Events/FindElementEventArgs.cs
@@ -1,4 +1,4 @@
-﻿// <copyright file="FindElementEventArgs.cs" company="WebDriver Committers">
+// <copyright file="FindElementEventArgs.cs" company="WebDriver Committers">
 // Licensed to the Software Freedom Conservancy (SFC) under one
 // or more contributor license agreements. See the NOTICE file
 // distributed with this work for additional information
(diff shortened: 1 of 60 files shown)


dotnet/src/support/Events/FindElementEventArgs.cs
```diff
@@ -1,4 +1,4 @@
-﻿// <copyright file="FindElementEventArgs.cs" company="WebDriver Committers">
+// <copyright file="FindElementEventArgs.cs" company="WebDriver Committers">
```
titusfortner, line 1: what is different in these lines?

titusfortner:
Looks good. Is there a linting tool for this?

dotnet/src/support/Events/FindElementEventArgs.cs
nvborisenko: Invisible character:

image

nvborisenko:
https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-format

> `dotnet-format` is a code formatter for dotnet that applies style preferences to a project or solution. Preferences will be read from an *.editorconfig* file, if present, otherwise a default set of preferences will be used. At this time dotnet-format is able to format C# and Visual Basic projects with a subset of [supported .editorconfig options](https://github.com/dotnet/format/blob/main/docs/Supported-.editorconfig-options.md).

titusfortner:
Hmm, it isn't working on my machine, does it need Windows?
Can we add a linting job to our CI like we have in other languages?

nvborisenko:
It doesn't require Windows.

Many things, which are supposed to work, don't work on this solution out of the box...

titusfortner:
What did you have to do to get it working?

PR #12893, opened 2023-10-06
Refactor common logic for setting up driver services and options into…

manuelsblanco:
… parent class

- Add a new protected static method `setupServiceAndOptions` in the parent class `ChromiumDriver`. This method handles the common logic for setting up driver services and options.

- Update `generateExecutor` methods in both `ChromeDriver` and `EdgeDriver` classes to call the new `setupServiceAndOptions` method from `ChromiumDriver`.

This change aims to reduce code duplication and improve maintainability.

### Description
This pull request addresses the issue of duplicated code across the ChromeDriver and EdgeDriver classes for setting up driver services and options.

Changes

Introduced a new protected static method called setupServiceAndOptions in the parent class ChromiumDriver. This method encapsulates the common logic for setting up driver services and options, including null checks and executable path setup.
Modified the generateExecutor methods in both ChromeDriver and EdgeDriver classes to call the new setupServiceAndOptions method from ChromiumDriver. This eliminates the need for duplicated code in these methods.

### Motivation and Context
The primary motivation for this change is to improve code maintainability and reduce duplication. Prior to this change, both ChromeDriver and EdgeDriver classes had their own implementations of the generateExecutor method, which contained nearly identical code for setting up driver services and options. This redundancy makes the codebase harder to manage and increases the risk of inconsistencies.

### Types of changes

### Checklist


titusfortner:
@manuelsblanco looks like it is failing the build. Can you take a look?

java/src/org/openqa/selenium/chromium/ChromiumDriver.java
```diff
@@ -41,13 +41,15 @@
...
 import org.openqa.selenium.bidi.HasBiDi;
+import org.openqa.selenium.chrome.ChromeOptions;
 import org.openqa.selenium.devtools.CdpEndpointFinder;
 import org.openqa.selenium.devtools.CdpInfo;
 import org.openqa.selenium.devtools.CdpVersionFinder;
 import org.openqa.selenium.devtools.Connection;
 import org.openqa.selenium.devtools.DevTools;
 import org.openqa.selenium.devtools.HasDevTools;
 import org.openqa.selenium.devtools.noop.NoOpCdpInfo;
+import org.openqa.selenium.edge.EdgeOptions;
```
pujagani, line 52: I understand your motivation for this change. But this is might not best approach for it. Ideally, the EdgeDriver and ChromeDriver extend the ChromiumDriver. The superclass should not be dependent on the details of the implementing subclasses.

pujagani (changes requested):

java/src/org/openqa/selenium/chromium/ChromiumDriver.java
manuelsblanco: @pujagani Any suggestions to avoid repeated code, or should we leave it as it is for now?

titusfortner: When I refactored the Services code to make Edge & Chrome match I started to look at abstracting out a superclass, and it didn't look simple, so I backed away slowly. It would be good to figure it out, but it's not a priority.

manuelsblanco: Well if we are all on the same page, we can close this PR for now without the merge.

PR #12898, opened 2023-10-07
[java] use the java process builder to run external processes

joerg1985:
### Description
Replaced commons-exec with the java process builder to get rid of the dependency.

### Motivation and Context
With Java 11 we now have all we need to implement the same functionality without a dependency.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/os/CommandLine.java
+++ b/java/src/org/openqa/selenium/os/CommandLine.java
@@ -27,6 +27,7 @@ import java.util.concurrent.TimeUnit;
 import org.openqa.selenium.Platform;
 import org.openqa.selenium.WebDriverException;
 
+@Deprecated
 public class CommandLine {
 
   private final OsProcess process;
--- a/java/src/org/openqa/selenium/os/OsProcess.java
+++ b/java/src/org/openqa/selenium/os/OsProcess.java
@@ -44,6 +44,7 @@ import org.openqa.selenium.internal.Require;
 import org.openqa.selenium.io.CircularOutputStream;
 import org.openqa.selenium.io.MultiOutputStream;
 
+@Deprecated
 class OsProcess {
   private static final Logger LOG = Logger.getLogger(OsProcess.class.getName());
 
--- a/java/src/org/openqa/selenium/manager/SeleniumManager.java
+++ b/java/src/org/openqa/selenium/manager/SeleniumManager.java
@@ -28,6 +28,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -43,7 +44,7 @@ import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.json.Json;
 import org.openqa.selenium.json.JsonException;
 import org.openqa.selenium.manager.SeleniumManagerOutput.Result;
-import org.openqa.selenium.os.CommandLine;
+import org.openqa.selenium.os.ExternalProcess;
 
 /**
  * This implementation is still in beta, and may change.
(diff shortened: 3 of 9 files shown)


java/src/org/openqa/selenium/os/CommandLine.java
```diff
@@ -1,174 +0,0 @@
...
-import static org.openqa.selenium.Platform.WINDOWS;
-
-import java.io.File;
-import java.io.OutputStream;
-import java.util.Map;
-import java.util.concurrent.TimeUnit;
-import org.openqa.selenium.Platform;
-import org.openqa.selenium.WebDriverException;
-
-public class CommandLine {
```
pujagani, line 30: Why are we removing this class entirely? I see this is a public class and we export it as a Maven dependency https://github.com/SeleniumHQ/selenium/blob/bceee4e33098ed063dbadf8e94c881993bed0ccd/java/src/org/openqa/selenium/os/BUILD.bazel#L8. Though I don't know how many users use this. I am afraid this will be a breaking change. If we want to remove it, we can do it how we typically do it in Selenium, by deprecating it and warning the users for 2 versions at least before removing it.

joerg1985: I missed this point and will rework this point to have a none breaking change.
The old classes will be deprecated and i will create a new one e.g. `ExternalProcess`.

titusfortner: If we had a Project Manager, we'd have planned this out 2 releases ago so we could have done it all at the same time. 😆

joerg1985: @pujagani this makes the PR smaller and better to review too, thanks for the hint.

java/src/org/openqa/selenium/os/OsProcess.java
```diff
@@ -44,6 +44,7 @@
 import org.openqa.selenium.io.CircularOutputStream;
 import org.openqa.selenium.io.MultiOutputStream;
 
+@Deprecated
 class OsProcess {
   private static final Logger LOG = Logger.getLogger(OsProcess.class.getName());
```
pujagani, line 49: I appreciate your effort here. I have not looked into it in depth so this is purely out of curiosity. Since I understand that using Java 11 APIs may not necessarily be lift and shift. Why was there a need for a new class ExternalProcess? Was it not possible to modify OsProcess methods to use the Java 11 APIs?

titusfortner:
Oh shoot, I was going to merge this before 4.14. :(

java/src/org/openqa/selenium/os/OsProcess.java
joerg1985: @pujagani The OsProcess is the internal class, so keeping the API would mean to keep the `CommandLine` alive.
I don't think the `CommandLine` API should be preserved, it mixes the methods to start a new process with controlling a existing process.
We would also need to implement ways to async start a process, what is not supported by the java process builder. These methods would be unused in the current code base, but should be at least be tested.

joerg1985: PS: If there are better names than `ExternalProcess` i could rename it.

e.g. I could rename `OsProcess` to `LegacyOsProcess` and `ExternalProcess` to `OsProcess`.
For me as not native english speaker, `OsProcess` sounds somehow operating system related.

pujagani: Thank you for explaining that in depth. I appreciate it. I guess even the package name might need renaming at a later point.

diemol (approved):
Thank you, @joerg1985!

@mykola-mokhnach @SrinivasanTarget @valfirst, does this affect Appium?

mykola-mokhnach:
> Thank you, @joerg1985!
> 
> @mykola-mokhnach @SrinivasanTarget @valfirst, does this affect Appium?

yes it does: https://github.com/appium/java-client/pull/2036#issuecomment-1756017744

That is why I have requested this feature

diemol:
> > Thank you, @joerg1985!
> > @mykola-mokhnach @SrinivasanTarget @valfirst, does this affect Appium?
> 
> yes it does: [appium/java-client#2036 (comment)](https://github.com/appium/java-client/pull/2036#issuecomment-1756017744)
> 
> That is why I have requested this feature

OK, so you are already checking this. We should be able to merge this anytime, @joerg1985.

valfirst:
@mykola-mokhnach it seems you mixed up PRs, this one about process management

@diemol it affects Appium, but I'm not sure how much, anyway once SNAPSHOT is published, it can be tested

PR #12907, opened 2023-10-09
Sim down CI runs

shs96c:


titusfortner:
@shs96c I think we did this better in London? Can we close?

diemol:
I think we can close this one.

Issue #12918, opened 2023-10-10
[🚀 Feature]: Need a way to enforce HTTP1.1 usage with the Java HTTP client


mykola-mokhnach:
### Feature and motivation

Currently it is not possible to enforce HTTP1.1 usage in the client. This feature is needed because Appium server only supports HTTP1.1 so far (SPDY support has been added [recently](https://github.com/appium/appium/pull/19105), but it's optional), so we'd like to default to HTTP1.1 for good.

### Usage example

Not sure yet where exactly this needs to live.

titusfortner:
I started working on this, looks straightforward, hope to have something we can put in nightly

joerg1985:
I think this is nothing the client should do.

The inital request is done with http 1.1 in any case and will ony upgrade to http/2 when the server will respond with switch protocol.

The server should remove the http Upgrade header (not the one for WebSockets ;)) to avoid the http/2 upgrade.

mykola-mokhnach:
> I think this is nothing the client should do.
> 
> The inital request is done with http 1.1 in any case and will ony upgrade to http/2 when the server will respond with switch protocol.
> 
> The server should remove the http Upgrade header (not the one for WebSockets ;)) to avoid the http/2 upgrade.

maybe. What I observe for now is that we did not see connection issues with previous client versions using netty @asolntsev got a similar results with their [tests](https://github.com/appium/java-client/pull/2036#issuecomment-1756859338). And enforcing HTTP version was the only workaround I was able to google. 
@joerg1985 I would be happy to accept any suggestions to avoid the aforementioned connection errors that don't require this particular change.

joerg1985:
@mykola-mokhnach I just had a look at the initial request and it is like expected.
You can enable the logging by using `System.setProperty("jdk.httpclient.HttpClient.log", "headers,content");`

The only difference in the request are the 'Connection', 'HTTP2-Settings', 'Upgrade' headers.

HTTP2 enabled client:

```
POST /session HTTP/1.1
Connection: Upgrade, HTTP2-Settings
Content-Length: 215
Host: 127.0.0.1:5555
HTTP2-Settings: AAEAAEAAAAIAAAABAAMAAABkAAQBAAAAAAUAAEAA
Upgrade: h2c
Cache-Control: no-cache
Content-Type: application/json; charset=utf-8
User-Agent: selenium/4.14.0 (java windows)
```

HTTP2 disabled client:

```
POST /session HTTP/1.1
[...]

PR #12919, opened 2023-10-10
[java] allow setting version in the Http Client Config

titusfortner:
### Description
Adds version to the constructor and all the methods for Http Client Config
I didn't add tests because our current tests don't seem to be testing the other parameters in a way I can manage. Would need some mocks that I don't have time to figure out.

### Motivation and Context
Addresses #12918 since the default is HTTP 2

### DIFF
--- a/java/src/org/openqa/selenium/remote/http/ClientConfig.java
+++ b/java/src/org/openqa/selenium/remote/http/ClientConfig.java
@@ -23,7 +23,6 @@ import java.net.Proxy;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
-import java.net.http.HttpClient;
 import java.time.Duration;
 import javax.net.ssl.SSLContext;
 import org.openqa.selenium.Credentials;
@@ -40,7 +39,7 @@ public class ClientConfig {
   private final Proxy proxy;
   private final Credentials credentials;
   private final SSLContext sslContext;
-  private HttpClient.Version version;
+  private String version;
 
   protected ClientConfig(
       URI baseUri,
@@ -50,7 +49,7 @@ public class ClientConfig {
       Proxy proxy,
       Credentials credentials,
       SSLContext sslContext,
-      HttpClient.Version version) {
+      String version) {
     this.baseUri = baseUri;
     this.connectionTimeout = Require.nonNegative("Connection timeout", connectionTimeout);
     this.readTimeout = Require.nonNegative("Read timeout", readTimeout);
@@ -214,7 +213,7 @@ public class ClientConfig {
     return sslContext;
   }
 
-  public ClientConfig version(HttpClient.Version version) {
+  public ClientConfig version(String version) {
     return new ClientConfig(
         baseUri,
         connectionTimeout,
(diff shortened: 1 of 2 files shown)


java/src/org/openqa/selenium/remote/http/ClientConfig.java
```diff
@@ -48,19 +49,28 @@ protected ClientConfig(
       Filter filters,
       Proxy proxy,
       Credentials credentials,
-      SSLContext sslContext) {
+      SSLContext sslContext,
+      HttpClient.Version version) {
```
diemol: Wouldn't this couple the `ClientConfig` with a particular client implementation? In this case, the current Java 11 HTTP implementation. 

Is `HttpClient.Version` accepted in other Java HTTP clients?

titusfortner: You're right, this should be a String. No idea other HTTP clients. If they want it, they can use it

java/src/org/openqa/selenium/remote/http/ClientConfig.java
```diff
@@ -40,7 +39,7 @@ public class ClientConfig {
   private final Proxy proxy;
   private final Credentials credentials;
   private final SSLContext sslContext;
-  private HttpClient.Version version;
+  private String version;
```
diemol, line 42: It would be nice to have a more descriptive name for the variable name since we do not have any JavaDocs that could explain more. 

Something like `httpVersion` should do.

mykola-mokhnach: Could we make it of java.net.http.HttpClient.Version type? String might be confusing as it is not clear which value to provide unless it is going to fail in runtime at valueOf call

java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java
```diff
@@ -150,6 +151,11 @@ public void connectFailed(URI uri, SocketAddress sa, IOException ioe) {
       builder.sslContext(sslContext);
     }
 
+    String version = config.version();
```
mykola-mokhnach, line 154: could be simplified to 
```ofNullable(config.version()).map(Version::valueOf).ifPresent(builder::version)```

java/src/org/openqa/selenium/remote/http/ClientConfig.java
diemol: If we use a type the config would be coupled to a specific implementation, right?

titusfortner: could subclass the config, create a public setter/getter that takes a HttpClient.Version but stores it as a string.

java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java
titusfortner: Needs to be consistent with the rest of the file. Just trying to get a quick fix here.

java/src/org/openqa/selenium/remote/http/ClientConfig.java
titusfortner: Yes, you already have `AppiumClientConfig`, that can have a setter/getter that takes HTTPClient.Version

PR #12933, opened 2023-10-11
[Bugfix] Refactored code in the WebDriverException class:

manuelsblanco:
- Replaced the filter expression from checking if a string is not null or empty using s == null || s.equals("") to s == null || s.isEmpty().

Updated a comment in the public enum WindowType:
- Changed the comment from "Represents the type of a new browser window that may be created." to "Represents the type of new browser window that may be created."

### Description
Corrected a typo and used the IsEmpty() method instead of comparing with an empty string

### Motivation and Context
Improve code quality by using the appropriate type of comparison

### Types of changes

### Checklist


pujagani (approved):
Thank you @manuelsblanco!

PR #12937, opened 2023-10-12
Remote webdriver session

sazi-y:

### Description

Add assignable Session ID to Webdriver. Added an extra parameter to the constructor of the WebDriver.
The session Id property is then assigned the already existing session, so it can point to it.
This means that for every session that is created by the webdriver, one can
return the session Id or cache it so that if a future call wants to
continue on the same webdriver session on Selenium Grid, it can do so by passing
in the preferred session Id which will then points it to an already existing webdriver
and automation can continue.

### Motivation and Context

A Webdriver normally generates a new Session Id everytime it is a new instance.
In a case where you want to do multiple tests, with selenium grid, if you have a stream of automation steps that wish to continue on the same Driver/Session, you would have to cache  an instance of that webdriver locally and wait for the next call. Once a continuation call has been made, you would have to fetch the instance of the webdriver, 
with its Session Id intact. This is obviously not quick and also not a good practice
for memory as you would have to cache multiple driver instances for multiple tests, running at the same time.

### Types of changes

### Checklist

### DIFF
--- a/dotnet/src/webdriver/Remote/RemoteWebDriver.cs
+++ b/dotnet/src/webdriver/Remote/RemoteWebDriver.cs
@@ -103,12 +103,34 @@ namespace OpenQA.Selenium.Remote
         }
 
         /// <summary>
+        /// Initializes a new instance of the <see cref="RemoteWebDriver"/> class. This constructor defaults proxy to http://127.0.0.1:4444/wd/hub
+        /// </summary>
+        /// <param name="remoteAddress">URI containing the address of the WebDriver remote server (e.g. http://127.0.0.1:4444/wd/hub).</param>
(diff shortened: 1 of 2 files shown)


titusfortner:
This has been requested a number of times over the years. The main problem is that there is no way to guarantee that the capabilities reported by the driver object are the ones being used by the driver. The project does not want to manage this.

There is a way to reuse a Chrome session based on the remote debug address, which has the advantage of allowing two different drivers to manage the same browser instance with potentially different capabilities, which means that the capabilities in the driver object are always correct for the action being taken. If your use case requires you to maintain a string in memory instead of an object in memory, that would be the preferred solution.

PR #12940, opened 2023-10-12
[java] Allow setting proxy for all http client instances

pujagani:

### Description

Allow setting proxy for all HTTP client instances via system property

### Motivation and Context

The default AsyncHttpClient used in the previous version of Selenium allowed the use of a global proxy set via the system properties. https://github.com/SeleniumHQ/selenium/blob/selenium-4.13.0/java/src/org/openqa/selenium/remote/http/netty/NettyClient.java#L88C3-L88C3

Extending the same functionality for the current default JDK 11 Http Client. 

However, we have the option to use client config to set proxy in the RemoteWebDriverBuilder. It is used by the underlying HttpClient of RemoteWebDriver only. The other HttpClients created along the flow, do not use it. Example: 
https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/devtools/CdpEndpointFinder.java#L48

Similarly, for the Grid, we create different instances of HttpClient to be used in different classes which cannot be configured to use the proxy.

The system property approach aims to provide a convenient way to set a global proxy.

This
### Types of changes

### Checklist


titusfortner:
This is a System property not environment variable, right?

Do we need something similar for ssl or is http sufficient?

pujagani:
Yes! System property is the correct description. I think HTTP is sufficient but I will need to check.

diemol (approved):
This looks good, it should be merged!

java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java
```diff
@@ -123,28 +125,34 @@ protected PasswordAuthentication getPasswordAuthentication() {
...
-                return List.of(proxy);
-              }
+    String proxyHost = System.getProperty("http.proxyHost");
+    String proxyPort = System.getProperty("http.proxyPort");
+
+    Proxy proxy =
+        (proxyHost != null && proxyPort != null)
+            ? new Proxy(
+                Proxy.Type.HTTP, new InetSocketAddress(proxyHost, Integer.parseInt(proxyPort)))
+            : config.proxy();
```
joerg1985, line 135: @pujagani I think an explicit set `config.proxy()` should beat the global system property?
And is it needed to implement this, this is just the [standard behaviour](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/doc-files/net-properties.html) of the HttpClient and should work out of the box?

joerg1985: PS: I use `java.net.useSystemProxies=true` to pickup the system settings

pujagani: I had the same thought process before I made the code change. I tried using https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/doc-files/net-properties.html, and also tried digging in the HttpClient client at the part where the proxy is set, it didn't seem to use the standard system properties.

Issue #12737, opened 2023-09-13
[🚀 Feature]: remove guava from client bindings


joerg1985:
### Feature and motivation

When using Java 11 we should not need the guava library in the client bindings not realy, we can do some small changes to remove them. Removing them from the dependencies should be done with Selenium 5, as client code could relay on it on the class path. We might be able to remove it from the server too, but this must be evaluated and could be done later.

### Usage example

Less dependencies are in general better.

titusfortner:
Why would we wait for Selenium 5? Once we require Java 11, if we can update syntax and methods and dependencies we should just do it.

diemol:
I understand the reasoning, a user might be relying on Selenium providing Guava as a dependency. But I think we can tell them what to do in the release notes.

joerg1985:
Okay, this will be a breaking change:
https://github.com/SeleniumHQ/selenium/blob/e5182732e95bddbaf91a042c6a70bacf20a5ac85/java/src/org/openqa/selenium/support/ui/ExpectedCondition.java#L34-L36

titusfortner:
Do we allow the native Java interface, or would we need to create a separate implementation for it. I'm hoping to pull it out of core Selenium code entirely for Selenium 5, so maybe we don't need to touch it and can spin it off at that time.

joerg1985:
It should be working fine with the native java interface.

joerg1985:
I just pushed a commit (42cc35585b8a60f722c67d34fdc877e3b4c2c89b) to remove the usage of guava from the browser bindings.

These are left areas are left:
- devtools -> #12943 
- bidi -> same changes as in the dev tools connection, i will wait for the devtools PR to get merged
- support -> open
- remote -> open
- os -> only referenced in build file without usage

titusfortner:
@joerg1985 I'm going to remove the milestone on this one since there's no need for a specific timeframe on it. Are you planning to continue working on this? Thanks.

diemol:
Having a single issue for this is helpful but not so actionable. 

@joerg1985, if you have identified where we can make code changes, I suggest creating an issue for each place you found and adding instructions for the implementation details. With that, anyone could jump in and help. Then, closing this issue.

joerg1985:
Yes i have planned to work on this as soon as possible. I think i will continue after christmas. I will split this in multiple tasks the next days.

joerg1985:
[...]

PR #12943, opened 2023-10-12
[java] removed use of guava from devtools

joerg1985:
### Description
This PR removes all usages of guava from devtools, usually this is an easy replacement with Java 11 methods.

The changes in `CdpClientGenerator.java` are not so easy because of the used `com.google.common.reflect.TypeToken`.
I have replaced it with a new method `JsonInput.readArray` to read an array of a specific type. 

There was also a bug in the code generated in the past, i have fixed too, e.g. see the cast in `Rectangle.fromJson` from `List<Number>` to `Rectangle` what will probably fail:
```
package org.openqa.selenium.devtools.v116.domsnapshot.model;

** cropped **

public class Rectangle {

    ** cropped **

    private static Rectangle fromJson(JsonInput input) {
        return (Rectangle)input.read((new TypeToken<List<Number>>() {
        }).getType());
    }

   ** cropped **
}
```

### Motivation and Context
This is one step in direction of #12737.

The CdpClientGenerator does still use guava to keep this PR small.

### Types of changes

### Checklist


joerg1985:
Looks like some tests fail, i will check this.

joerg1985:
Tests are working and the PR is ready for review.

java/src/org/openqa/selenium/devtools/CdpClientGenerator.java
```diff
@@ -40,6 +40,7 @@
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
```
krmahadevan, line 43: @joerg1985 - Are we only selectively removing references to guava library? I ask because I see references to Immutable map in this class https://github.com/SeleniumHQ/selenium/blob/8393fed02a81b3dabfe4a20f83ee978449b40c22/java/src/org/openqa/selenium/devtools/CdpClientGenerator.java#L31C34-L31C46

joerg1985: This is one step in direction of #12737, removing guava from the java client.

The CdpClientGenerator does still use guava to keep this PR small and is not part of the client.

pujagani:
Thank you @joerg1985 for this! Good effort. The changes look good to me.

diemol (approved):
Thank you, @joerg1985!

PR #12948, opened 2023-10-13
[dotnet] store the mutated element for the user not just the target id

titusfortner:
### Description
When an element is mutated it returns the ID of the element, this stores that element in the representative Data class instead of just the target ID. 

### Motivation and Context
* The other languages do it.
* It's not obvious to the user how to get the information

Required now:
```cs
            var mutations = new List<IWebElement>();
            using IJavaScriptEngine monitor = new JavaScriptEngine(driver);
            monitor.DomMutated += (sender, e) =>
            {
                var locator = By.CssSelector($"*[data-__webdriver_id='{e.AttributeData.TargetId}']");
                mutations.Add(driver.FindElement(locator));
            };
```

Desired:
```cs
            var mutations = new List<IWebElement>();
            using IJavaScriptEngine monitor = new JavaScriptEngine(driver);
            monitor.DomMutated += (sender, e) =>
            {
                mutations.Add(e.Element);
            };
```

Java:
```java
    CopyOnWriteArrayList<WebElement> mutations = new CopyOnWriteArrayList<>();
    ((HasLogEvents) driver).onLogEvent(domMutation(e -> mutations.add(e.getElement())));
```

Ruby:
```rb
    mutations = []
    driver.on_log_event(:mutation) { |mutation| mutations << mutation.element }
```


dotnet/src/webdriver/JavaScriptEngine.cs
```diff
@@ -380,6 +380,9 @@ private void OnScriptBindingCalled(object sender, BindingCalledEventArgs e)
             if (e.Name == MonitorBindingName)
             {
                 DomMutationData valueChangeData = JsonConvert.DeserializeObject<DomMutationData>(e.Payload);
+                var locator = By.CssSelector($"*[data-__webdriver_id='{valueChangeData.TargetId}']");
+                valueChangeData.Element = driver.FindElements(locator)[0];
```
nvborisenko: It throws exception in case of no elements were found. Should we be safe here?

titusfortner: Cool, I'll use `FirstOrDefault()` then

titusfortner:
There are a few places where the Args template isn't storing the useful information but I do not want to spend time on any of this if we can start pushing the actual BiDi spec. Actually, it probably is forward compatible the way it is written. We can add the more useful things with the correct classes in the next implementation. Just frustrating right now. :)

PR #12951, opened 2023-10-16
[py] filter which bidi logs are recorded based on event type

titusfortner:
### Description
Changed the logic for updating the console message based on whether the message is an error or a log type.

### Motivation and Context
I updated the tests to show what doesn't work in current code.

@AutomatedTester I don't really understand how this is implemented, I'm just looking at the inputs/outputs. It appears that only one message can be processed, though, so a second execute script never gets triggered. I think that means it is ok to match off of `messages.value.type_`, but let me know if I'm misunderstanding this.

### DIFF
--- a/py/selenium/webdriver/common/log.py
+++ b/py/selenium/webdriver/common/log.py
@@ -138,9 +138,6 @@ class Log:
         async with session.wait_for(self.devtools.runtime.ConsoleAPICalled) as messages:
             yield console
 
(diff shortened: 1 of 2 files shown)


titusfortner:
I think the code needs to be able to record more than just the first event, or only one event at a time, but I don't know how to implement it, so I'm backing away slowly. This change is now minor, so I'm merging it.

Issue #12887, opened 2023-10-06
[🚀 Feature]: Create Selenium Manager binaries in GH Actions with debug symbols


bonigarcia:
### Feature and motivation

To troubleshoot SM in complex error cases, it is interesting to capture the backtrace. But to do that, the SM binaries have to be created, including the debug symbols. Since these binaries are much larger than the default release artifacts, we can generate them on demand with GH Actions.

### Usage example

These binaries will be used to troubleshoot problems that users report, e.g., as in #12828.

bonigarcia:
Completed via #12974.

PR #12974, opened 2023-10-20
[rust] Include checkbox in SM workflow to generate binaries with debug symbols

bonigarcia:
### Description
This PR modifies the existing workflow to generate the SM binaries in CI to include the following checkbox when triggering the [SM build](https://github.com/SeleniumHQ/selenium/actions/workflows/build-selenium-manager.yml):
image

This checkbox allows generating the SM binaries including the debug symbols, required to get a stack trace in case of an error. This will be used only on demand to troubleshoot complicated problems.

****Example of release SM binaries (checkbox disabled):**
_selenium-manager_linux-x64 5.22 MB	
selenium-manager_macos-universal 8 MB	
selenium-manager_windows-x32 4.06 MB_
https://github.com/SeleniumHQ/selenium/actions/runs/6585759070

****Example of debug SM binaries (checkbox enabled):**
_selenium-manager_linux-x64-debug 158 MB	
selenium-manager_macos-universal-debug 75.2 MB	
selenium-manager_windows-x32-debug 181 MB_
https://github.com/SeleniumHQ/selenium/actions/runs/6585702295

### Motivation and Context
Implementation for  #12887.

### Types of changes

### Checklist

### DIFF
--- a/.github/workflows/build-selenium-manager.yml
+++ b/.github/workflows/build-selenium-manager.yml
@@ -18,8 +18,8 @@ jobs:
       - name: "Update Rust"
         run: |
           rustup update
-          rustup toolchain install stable-i686-pc-windows-msvc
-          rustup default stable-i686-pc-windows-msvc
+          rustup toolchain install stable-i686-pc-windows-gnu
+          rustup default stable-i686-pc-windows-gnu
           rustc -vV
       - name: "Build release binary"
         if: ${{ inputs.debug == false }}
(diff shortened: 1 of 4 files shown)


titusfortner:
Why not just have it always build both?

bonigarcia:
I did that since the debug artifacts are different from the usual use case. It will be generated on demand, just when a complex troubleshooting case happens. Typically, these artifacts will be downloaded by the user who experienced some problem. Since the artifacts built with GH have a retention time (6 days currently), we likely need to regenerate the build artifacts on purpose. For that, the checkbox is disabled by default (i.e., for the release), and we will trigger it with the checkbox enabled for troubleshooting.

diemol (approved):

PR #12977, opened 2023-10-20
script to automate part of CDP updates

titusfortner:
### Description
* checks if latest stable version is present in code and exits if not
* downloads latest stable pdls
* updates the Java & .NET directories and files
* updates all references in the files themselves

### Motivation and Context
Chrome is releasing more often, we need this to be easier.

This should do pretty much everything. I did a comparison of how it would handle adding 118 and it was equivalent.

Want to run this one before still (is it too aggressive to combine them?):
```bash
bazel run //scripts:pinned_browsers > common/repositories.bzl.foo
mv common/repositories.bzl.foo common/repositories.bzl
buildifier --mode fix --lint fix -r common/repositories.bzl
```

and run this one after:
```
bazel run //rb:selenium-devtools-release
```

### DIFF
--- a/scripts/update-devtools.sh
+++ b/scripts/update-devtools.sh
@@ -1,0 +1,129 @@
+#!/bin/bash
+
+if [ -z "$1" ]; then
+    echo "Please provide a language argument: java, dotnet, ruby, python, javascript, or all."
+    exit 1
+fi
+
+LANGUAGE="$1"
+
+FULL_VERSION=$(curl -s https://googlechromelabs.github.io/chrome-for-testing/last-known-good-versions.json | jq -r '.channels.Stable.version')
+
+MAJOR_VERSION=$(echo "$FULL_VERSION" | cut -d'.' -f1)
+PREVIOUS_VERSION=$((MAJOR_VERSION - 1))
+REMOVING_VERSION=$((MAJOR_VERSION - 3))
+
+# Update pdl files if target directory does not exist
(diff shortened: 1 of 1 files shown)


titusfortner:
I'm not certain this will grab the early stable version, so we'll have to see when it comes out.

scripts/update-devtools.sh
```diff
@@ -0,0 +1,129 @@
...
+
+    if [ -d "$SOURCE_DIR" ]; then
+        cp -r "$SOURCE_DIR/"* "$TARGET_DIR/"
+    fi
+
+    mv js_protocol.pdl "$TARGET_DIR"
+    mv browser_protocol.pdl "$TARGET_DIR"
+    echo "Contents from $SOURCE_DIR have been copied to $TARGET_DIR!"
+
+    # Fix known syntax issue with downloaded pdl
```
diemol, line 42: I would add to the comment the why. This is Javadocs breaking when they see a `<script>` tag, as it interprets it as a security vulnerability.

diemol (approved):
This is great, thank you, @titusfortner!

We need to add this to the release docs. And create a GitHub workflow with it? Well, I think we should merge this and fix any issues while we use the script.

scripts/update-devtools.sh
titusfortner: The goal is to rewrite this in Python & integrate with pinning the browsers. I'm going to merge this for the upcoming release and we can fix it up next week.

Issue #12975, opened 2023-10-20
[🐛 Bug]: Retry of hub on `TimeoutException`  does perform Command multiple times


joerg1985:
### What happened?

The retry handling of the hub might execute a command twice, after the client already failed.
The hub will retry sending a command to the node on `TimeoutException`, although the driver did execute the command the first time and is only waiting for the page to load. See #12558 for a potential real world occurrence of this issue.

I think the `TimeoutException` should not be handled by the retry filter of the hub, as the command has allready been send to the driver. The other retrys are okay in my mind, e.g. on `ConnectionTimeout`.

The default timeouts should also be updated, the default w3c page load timeout is 300s and the default read timeout of the client is 180s. In general it might make sense to have a slightly higher read time on the client as on hub the to ensure the error generated by the hub has some time to be send to the client. The same should also be considered for the timeout of the node to driver communication.

### How can we reproduce the issue?

```shell
** this will only work with a hub/node setup, the standalone does not use http internally to handle requests **
** run the node with -Djdk.httpclient.HttpClient.log=headers to see the requests send to the driver **

This dummy code will provoke the issue after ~5 min, the chromedriver will wait after the click up to 300s for the page to load. The http client on the client will timeout after 180s and raise a TimeoutException. The http client on the hub will also raise a TimeoutException after 180s and retry to send the command to the driver.
The click is than performed again, look at second call to '../click' in output of the node.


import java.net.URL;
import java.time.Duration;
import java.time.LocalTime;
import java.time.temporal.ChronoUnit;
import java.util.concurrent.CountDownLatch;
import java.util.concurrent.TimeUnit;
import java.util.logging.Level;
import java.util.logging.Logger;
import org.openqa.selenium.By;
import org.openqa.selenium.remote.RemoteWebDriver;
import org.apache.hc.core5.http.impl.bootstrap.HttpServer;
import org.apache.hc.core5.http.impl.bootstrap.ServerBootstrap;
import org.apache.hc.core5.http.io.SocketConfig;
import org.apache.hc.core5.http.io.entity.HttpEntities;
import org.apache.hc.core5.util.TimeValue;
import org.openqa.selenium.TimeoutException;
import org.openqa.selenium.chrome.ChromeOptions;


public class RetryClick {

    private static final Logger LOG = Logger.getLogger(TEts.class.getName());

    public static void main(String[] args) throws Exception {
        final SocketConfig socketConfig = SocketConfig.custom()
[...]

Issue #12558, opened 2023-08-15
[🐛 Bug]: Timeout exception thrown randomly with RemoteWebDriver


Abhijith2893:
### What happened?

When executing tests at scale there are some random failures with read timeouts (seems to wait for 3 mins). We are observing this primarily when clicking an element but have also observed it during navigate to page. Seems to be a reoccurrence of https://github.com/SeleniumHQ/selenium/issues/9528

We are using dockerized grid v4.10 with the hub/node model and we confirmed that the docker process starts with `jdk-http-client`

`java -Dwebdriver.http.factory=jdk-http-client -jar /opt/selenium/selenium-server.jar --ext /opt/selenium/selenium-http-jdk-client.jar hub --session-request-timeout 3600 --session-retry-interval 15 --relax-checks true --bind-host false --config /opt/selenium/config.toml`

### How can we reproduce the issue?

```shell
Unfortunately similar to https://github.com/SeleniumHQ/selenium/issues/9528 we are unable to find a consistent repro for this. Seems flaky and rerunning the test passes but we are still seeing this happen daily.
```

### Relevant log output

```shell
We use an internal framework to run selenium tests and the DEBUG logs indicate that a 500 was returned during the click event. We check to see if the element is enabled and clickable (which passes) but the subsequent click seems to fail.


DEBUG    2023-08-15 19:24:02.928  selenium.webdriver.remote.remote_connection:remote_connection.py:303 -> POST http://selenium-four-hub-nomad.service.eng.consul:4444/session/ab89759c17aac47ed60f6d6b9a8b6882/element {"using": "xpath", "value": "//*[@class='extension-name'][text()='JSON']"}
DEBUG    2023-08-15 19:24:02.943  urllib3.connectionpool:connectionpool.py:473 -> http://selenium-four-hub-nomad.service.eng.consul:4444 "POST /session/ab89759c17aac47ed60f6d6b9a8b6882/element HTTP/1.1" 200 95
DEBUG    2023-08-15 19:24:02.943  selenium.webdriver.remote.remote_connection:remote_connection.py:319 -> Remote response: status=200 | data={"value":{"element-6066-11e4-a52e-4f735466cecf":"3AFE27AB794812F985C270183415B5A4_element_21"}} | headers=HTTPHeaderDict({'content-length': '95', 'content-type': 'application/json; charset=utf-8', 'cache-control': 'no-cache'})
[...]

PR #12978, opened 2023-10-20
[grid] keep HttpClient alive until unused #12558 

joerg1985:
### Description
The old implementation did close the HttpClient some minutes after access and did not check if the HttpClient has request still waiting for a response. 

### Motivation and Context
This did break long running requests, especially when the client has a high read timeout set. 

**Note:** This PR might make #12975 even worse, as it will not close the http client and prevent the retry to send the command more often.  

**Note:** This PR will also fix the broken get session url endpoint and uses it where ever it makes sense.
This might lead to issues when people are using a grid of mixed versions, but is this supported at all?

~~**Note:** This PR will also add more output when the JSON parser failed, as it was useful to debug the broken get session url endpoint and might be useful for other issues too.~~ has been removed from the PR as requested.

### Types of changes

### Checklist


joerg1985:
An alternative way to handle this might be to use a WeakReference in the cache and use a Cleaner to call `.close` there.
This could also be used in other areas like the `ProxyWebsocketsIntoGrid` where it is not easy to call `.close` after use.

java/src/org/openqa/selenium/json/Input.java
```diff
@@ -97,7 +97,18 @@ public String toString() {
       length = MEMORY_SIZE;
     }
 
-    return "Last " + length + " characters read: " + new String(buffer, offset, length);
+    String last = "Last " + length + " characters read: " + new String(buffer, offset, length);
```
pujagani: This fix can be probably a separate PR.

joerg1985: Okay, i will remove this from the PR.
It was helpfull to debug the broken session url endpoint.

diemol (approved):
Can you please solve the conflict so we merge this?

joerg1985:
@diemol i have fixed it, the PR should have no conflicts left.

diemol:
There are some failures in CI, do you know if they are related?

titusfortner:
latest on trunk broke the CI so that might be causing these failures.

joerg1985:
i am currently looking into it

PR #12979, opened 2023-10-20
[java] implement file downloads

titusfortner:
### Status

Update: I re-implemented this because I realized it shouldn't use AutoService; no bazel changes required. 

This code works, but it is in draft because:
* ~I think I'm putting the interface and the implementation in the right directories, but bazel complains about AutoService, and I'm not sure I've added that dependency in the right place~
* ~I think the tests might need to get moved somewhere else, but to be honest, I'm not quite sure how to add this one to the test files~ The test I updated here is more of a unit test for the grid, so I think it should stay the same and I need to add this test someplace else, but I'm not sure where that someplace else needs to be. Advice appreciated.
* I'd like to get Python & .NET implemented before these get merged.

### Description
* Implement `getDownloadableFiles()`, `downloadFile(filename)` and `deleteDownloadableFiles()`

### Motivation and Context
Implements #11657


diemol (approved):
I think the updated test is enough, you are testing both the bindings and the Grid. 

If you wish, you could add unit tests for the bindings and then mock all the requests and responses, which I guess should go where the other `RemoteWebDriver` tests are.

pujagani (approved):

krmahadevan (approved):

rbri:
Hi @titusfortner,
just saw this mentioned in the release note and i'm thinking about making this possible with the HtmlUnit driver also.
But i'm a bit confused - i can't find a description of how the file list you get by get(/se/files) is constructed. Does this mean the browser checks all links in the page for something that might point to a file?
Any hint will be great!
Thanks 
  ronald

titusfortner:
It works with the grid, not with the individual drivers. If you are local you can set whatever directory you want in the options.

PR #12981, opened 2023-10-21
[dotnet] implement file downloads

titusfortner:
### Description
* Implements driver methods: `GetDownloadableFiles`, `DownloadFile` and `DeleteDownloadableFiles`
* Implements option property: `EnableDownloads`

I can't test this because I can't figure out how to run it in remote with bazel, or if that's possible right now. But I think this is right.

### Motivation and Context
Implements https://github.com/SeleniumHQ/selenium/issues/11657

### DIFF
--- a/dotnet/src/webdriver/CapabilityType.cs
+++ b/dotnet/src/webdriver/CapabilityType.cs
@@ -161,6 +161,11 @@ namespace OpenQA.Selenium
         /// </summary>
         public static readonly string WebSocketUrl = "webSocketUrl";
 
+        /// <summary>
+        /// Capability name used to get a value indicating whether files may be downloaded from remote end.
+        /// </summary>
+        public static readonly string EnableDownloads = "se:downloadsEnabled";
+
         private static readonly List<string> KnownSpecCompliantCapabilityNames = new List<string>() {
             BrowserName,
             BrowserVersion,
--- a/dotnet/src/webdriver/DriverCommand.cs
+++ b/dotnet/src/webdriver/DriverCommand.cs
@@ -394,6 +394,21 @@ namespace OpenQA.Selenium
         public static readonly string SetUserVerified = "setUserVerified";
 
         /// <summary>
+        /// Represents the GetDownloadableFiles command.
+        /// </summary>
+        public static readonly string GetDownloadableFiles = "getDownloadableFiles";
+
+        /// <summary>
+        /// Represents the DownloadFile command.
+        /// </summary>
+        public static readonly string DownloadFile = "downloadFile";
+
+        /// <summary>
+        /// Represents the DeleteDownloadableFiles command.
+        /// </summary>
+        public static readonly string DeleteDownloadableFiles = "deleteDownloadableFiles";
+
+        /// <summary>
         /// Lists the set of known commands valid for the Selenium library.
         /// </summary>
         public static readonly IList<string> KnownCommands = new List<string>()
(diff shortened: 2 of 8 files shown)


dotnet/src/webdriver/CapabilityType.cs
```diff
@@ -161,6 +161,11 @@ public static class CapabilityType
         /// </summary>
         public static readonly string WebSocketUrl = "webSocketUrl";
 
+        /// <summary>
+        /// Capability name used to get a value indicating whether files may be downloaded from remote end.
+        /// </summary>
+        public static readonly string EnableDownloads = "se:downloadsEnabled";
```
nvborisenko, line 167: Should we ad it to `KnownSpecCompliantCapabilityNames`? I don't know, just to double check.

dotnet/src/webdriver/DriverCommand.cs
```diff
@@ -393,6 +393,21 @@ public static class DriverCommand
         /// </summary>
         public static readonly string SetUserVerified = "setUserVerified";
 
+        /// <summary>
+        /// Represents the GetDownloadableFiles command.
+        /// </summary>
+        public static readonly string GetDownloadableFiles = "detDownloadableFiles";
```
nvborisenko: Seems mis typo: **d**etDownloadableFiles

dotnet/src/webdriver/CapabilityType.cs
titusfortner: Don't need to, the code also allows anything with a colon as valid by default.

dotnet/src/webdriver/IHasDownloadscs.cs
YevgeniyShunevych, line 1: A typo in the file name: `IHasDownloadscs.cs` -> `IHasDownloads.cs`

dotnet/src/webdriver/WebDriver.cs
```diff
@@ -30,7 +30,7 @@ namespace OpenQA.Selenium
     /// <summary>
     /// A base class representing a driver for a web browser.
     /// </summary>
-    public class WebDriver : IWebDriver, ISearchContext, IJavaScriptExecutor, IFindsElement, ITakesScreenshot, ISupportsPrint, IActionExecutor, IAllowsFileDetection, IHasCapabilities, IHasCommandExecutor, IHasSessionId, ICustomDriverCommandExecutor, IHasVirtualAuthenticator
+    public class WebDriver : IWebDriver, ISearchContext, IJavaScriptExecutor, IFindsElement, ITakesScreenshot, ISupportsPrint, IActionExecutor, IAllowsFileDetection, IHasCapabilities, IHasCommandExecutor, IHasSessionId, ICustomDriverCommandExecutor, IHasVirtualAuthenticator, IHasDownloads
```
YevgeniyShunevych: Is it correct to add the downloads functionality to `WebDriver` class rather than `RemoteWebDriver`? Looks like the functionality should work with Selenium Grid, not local drivers, so `RemoteWebDriver` might be a better place for it. Or I'm missing something?

titusfortner: Maybe, yes, this isn't implemented correctly anyway, so I'm overhauling. .NET interface implementation is a little different than Java's. But RemoteWebDriver probably makes more sense than the superclass.

dotnet/src/webdriver/DriverCommand.cs
titusfortner: fixed, thanks

dotnet/src/webdriver/WebDriver.cs
titusfortner: Yes, you were right on this, I fixed it. Thanks.

dotnet/src/webdriver/IHasDownloadscs.cs
titusfortner: good catch, thank you.

PR #13023, opened 2023-10-22
[py] implement file downloads

titusfortner:
### Status
The tests work when I run it locally against a grid I've started; I can't seem to get the remote tests to work with bazel locally, so it's in draft until I see if they pass on the CI. I can't figure out how to get the `test/selenium/webdriver/remote/**/*.py` tests to pass in the `--driver`. What's there "should" work. So I skipping them. But they work against 4.14.1 grid.

### Description
* Implements `get_downloadable_files()`, `download_file()` and `delete_downloadable_files()`
* implements `options.enable_downloads = True`
* It turns on downloads for all tests because I couldn't figure out how to make it conditional
* It adds empty `moz:firefoxOptions` because of https://github.com/SeleniumHQ/selenium/pull/13024

### Motivation and Context
Implements https://github.com/SeleniumHQ/selenium/issues/11657


titusfortner:
@AutomatedTester / @symonk I tried to add testing in the remote directory, but for some reason it wasn't recognizing the driver argument from the bazel file. /shrug

PR #13024, opened 2023-10-22
[grid] fix bug in downloading files by ensuring browser vendor option…

titusfortner:
### Description
If browser vendor options missing, merge in mutable capabilities with the capabilities set. 
I think this is better than casting the capabilities to `MutableCapabilities` to add it?

### Motivation and Context

Not all bindings create an empty browser vendor options map by default


diemol (commented):
Where is the bug? In the absence of browser options key or the casting? In theory `ofNullable` should be handling it.

titusfortner:
If there isn't a moz:firefoxOptions key, then we're just mutating a new hash that isn't attached to anything. Python downloading doesn't work unless another geckodriver option happens to be set.

diemol (approved):
Thank you for the explanation

titusfortner:
What, the description made complete sense in my mind... 😂 

I'll fix the formatting.

PR #13025, opened 2023-10-23
[grid] add value key to downloadFile response per w3c convention

titusfortner:
### Description
Add a value key to the delete downloadFile response

### Motivation and Context
Per the W3C [Processing Model](https://w3c.github.io/webdriver/#processing-model)
> Let response’s [body](https://w3c.github.io/webdriver/#dfn-body) be the [UTF-8 encoded](https://w3c.github.io/webdriver/#dfn-utf-8-encode) [JSON serialization](https://w3c.github.io/webdriver/#dfn-json-serialization) of a JSON [Object](https://w3c.github.io/webdriver/#dfn-object) with a key "value" set to data.

Ruby expects to see a value key from all responses, even if the data value is `null`

### DIFF
--- a/.github/workflows/build-selenium-manager.yml
+++ b/.github/workflows/build-selenium-manager.yml
@@ -1,31 +1,49 @@
 name: Build selenium-manager
 
-on: workflow_dispatch
+on:
+  workflow_dispatch:
+    inputs:
+      debug:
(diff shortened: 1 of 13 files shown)


diemol (approved):
Good catch, thank you, @titusfortner!

PR #13026, opened 2023-10-23
[dotnet] remove non-w3c compliant capability code

titusfortner:
### Description
* Remove invalid capabilities
* Update invalid capabilities to valid capabilities

### Motivation and Context
As best as I can tell, if the capability isn't in [this list](https://github.com/SeleniumHQ/selenium/blob/selenium-4.14.0/dotnet/src/webdriver/CapabilityType.cs#L164-L176) it gets ignored when starting a session — https://github.com/SeleniumHQ/selenium/blob/selenium-4.14.0/dotnet/src/webdriver/WebDriver.cs#L652-L655

I don't know if the browser version and platform name code is necessary, but if it is looking for the old values it won't work.

Whatever is going on here, it can be cleaned up more, I just made some quick changes while I noticed it.

### DIFF
--- a/dotnet/src/webdriver/Remote/DesiredCapabilities.cs
+++ b/dotnet/src/webdriver/Remote/DesiredCapabilities.cs
@@ -34,19 +34,6 @@ namespace OpenQA.Selenium.Remote
         /// <summary>
         /// Initializes a new instance of the <see cref="DesiredCapabilities"/> class
         /// </summary>
-        /// <param name="browser">Name of the browser e.g. firefox, internet explorer, safari</param>
-        /// <param name="version">Version of the browser</param>
-        /// <param name="platform">The platform it works on</param>
-        public DesiredCapabilities(string browser, string version, Platform platform)
-        {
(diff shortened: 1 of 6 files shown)


dotnet/src/webdriver/Remote/DesiredCapabilities.cs
```diff
@@ -124,28 +79,28 @@ public string BrowserName
         /// <summary>
         /// Gets or sets the platform
         /// </summary>
-        public Platform Platform
+        public Platform PlatformName
```
nvborisenko, line 82: CI Build fails.

```
dotnet\src\webdriver\Remote\ReadOnlyDesiredCapabilities.cs(213,148): error CS1061: 'ReadOnlyDesiredCapabilities' does not contain a definition for 'Platform' and no accessible extension method 'Platform' accepting a first argument of type 'ReadOnlyDesiredCapabilities' could be found (are you missing a using directive or an assembly reference?)
```

nvborisenko:
I think it is still actual. Up.

PR #13053, opened 2023-10-25
[bot] Determine more reliably if node has shadow DOM

dlanza1:
### Description
node.shadowRoot might have an object assigned (some framework working with ShadyDOM does assign an object to node.shadowRoot), but it is no necessarily used by the browser as it must be constructed by ShadowRoot, otherwise is ignored.

### Motivation and Context

Mechanisms like bot.dom.isShown() would determine an element is not shown because it believes it does not have parent.

### Types of changes

### Checklist


isaulv:
So there are libraries using the shadow property to store an object only to be reified as a proper shadow object later? that's interesting.

diemol (changes requested):
I'm sorry for the late review.

This breaks things in different places. Could you have a look at the builds?

PR #13071, opened 2023-10-30
[py] Fix test for CONDA_PREFIX in selenium_manager

mweinelt:
Fix test for CONDA_PREFIX in selenium manager, which regressed in 4.12.0.

### Description
Accessing CONDA_PREFIX in os.environ fails with a KeyError when it is not set. The condition should instead check for the existence of the key, before accessing it in the following codepath.

Fixes: d4285d1 ("fix for conda install of selenium-manager (#12536)")

### Motivation and Context

```pytb
 File "/nix/store/jdm5chqkmh3z0wbip7r165fas3bfg1y4-python3-3.11.5-env/lib/python3.11/site-packages/selenium/webdriver/common/driver_finder.py", line 38, in get_path
   path = SeleniumManager().driver_location(options) if path is None else path
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 File "/nix/store/jdm5chqkmh3z0wbip7r165fas3bfg1y4-python3-3.11.5-env/lib/python3.11/site-packages/selenium/webdriver/common/selenium_manager.py", line 78, in driver_location
   args = [str(self.get_binary()), "--browser", browser]
               ^^^^^^^^^^^^^^^^^
 File "/nix/store/jdm5chqkmh3z0wbip7r165fas3bfg1y4-python3-3.11.5-env/lib/python3.11/site-packages/selenium/webdriver/common/selenium_manager.py", line 57, in get_binary
   if not path.is_file() and os.environ["CONDA_PREFIX"]:
                             ~~~~~~~~~~^^^^^^^^^^^^^^^^
 File "<frozen os>", line 679, in __getitem__
KeyError: 'CONDA_PREFIX'
```

### Types of changes

### Checklist

### DIFF
--- a/py/selenium/webdriver/common/selenium_manager.py
+++ b/py/selenium/webdriver/common/selenium_manager.py
@@ -57,7 +57,7 @@ class SeleniumManager:
 
             path = Path(__file__).parent.joinpath(directory, file)
 
-        if not path.is_file() and os.environ["CONDA_PREFIX"]:
+        if not path.is_file() and "CONDA_PREFIX" in os.environ:
             # conda has a separate package selenium-manager, installs in bin
             path = Path(os.path.join(os.environ["CONDA_PREFIX"], "bin", file))
             logger.debug("Conda environment detected, using `%s`", path)


titusfortner:
So this never actually worked?

I think the plan is to remove this functionality entirely now that people can set `SE_MANAGER_PATH` to point to the location of the selenium manager. Either the user can set it or the package manager can set it.

titusfortner:
Thank you for the PR.

Between this and #13076 it appears this was not working, so I'm removing it.

The correct fix now is for the location of the Selenium Manager binary to be specified in `SE_MANAGER_PATH` environment variable. Ideally this can be added by the anaconda project by default.

PR #13082, opened 2023-11-01
py: fix RelativeBy#near to take 2 parameters

pinterior:
### Description

`RelativeBy#near` should take 2 parameters to specify target *element (or locator)* and *distance*.

also:
- add more strict type hints on relative_locator.py
- ~remove default arguments raises Exception~
- add some tests

### Motivation and Context

want to find far element

### Types of changes

### Checklist

   - on my side, some tests already fails on trunk, but newly created tests passed.


titusfortner:
1. We can't change a public method without warning. We typically mark methods or parameters as obsolete for 2 releases before removing something.
2. We aren't accepting changes in the support classes right now. We're meeting next week to decide how to approach this section of the code going forward.

pinterior:
> 1. We can't change a public method without warning.

resurrected default `None` argument.
old `locate_with(...).near(elem)` style call still works with default `distance` argument,  but need to use different name for `near(element_or_locator, distance)` method?

> 2. We aren't accepting changes in the support classes right now.

sure.
hope `near` locating will be able to use with custom distance in the future.

diemol:
The "issue" is that Java and C# already accept distance as a parameter. JavaScript and Python don't, and I did not understand the Ruby code to know whether it accepts it.

pinterior:
the main issue is following `near` method looks like to take single `int` parameter, but it doesn't work.

```python3
    def near(self, element_or_locator_distance: Union[WebElement, Dict, int] = None) -> "RelativeBy":
```

diemol:
> the main issue is following `near` method looks like to take single `int` parameter, but it doesn't work.
> 
> ```python
>     def near(self, element_or_locator_distance: Union[WebElement, Dict, int] = None) -> "RelativeBy":
> ```

Good point. Where can I see the failed test that shows the bug on the current implementation?

pinterior:
running test on faddc1b0ec0d453815e3cfffce95d1b345249bc3 clearly shows `near(123)` fails with `InvalidArgumentException`

diemol:
@pinterior can you please fix the linting so I can merge this?

diemol (approved):
Thank you!

PR #13113, opened 2023-11-08
[java] fix null request in case of chunked request

nikhlagrwl:

### Description
Selenium throws a NullPointerException in case of chunked request. The request object is initialised after sending 100 response for chunked requests and hence for the last chunk of the request, the request object is null.

### Types of changes

### Checklist


joerg1985:
@nikhlagrwl thanks for opening the PR, what client are you using to have chunked requests?
Some drivers do not support chunked requests, therefore you might run into simil ar issues with your client when directly connecting to a driver. So there might also be a client issue to fix, beside the server issue.

nikhlagrwl:
> @nikhlagrwl thanks for opening the PR, what client are you using to have chunked requests? Some drivers do not support chunked requests, therefore you might run into simil ar issues with your client when directly connecting to a driver. So there might also be a client issue to fix, beside the server issue.

@joerg1985  I tried with a ruby client and even with curl command, it's failing on both the case.
Getting the below error on server in both the scenario
`07:12:57.574 WARN [DefaultChannelPipeline.onUnhandledInboundException] - An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
java.lang.NullPointerException: Cannot invoke "org.openqa.selenium.remote.http.HttpRequest.setContent(java.util.function.Supplier)" because "this.request" is null`

joerg1985:
@nikhlagrwl could you share the curl call?

nikhlagrwl:
> @nikhlagrwl could you share the curl call?

echo '{"desiredCapabilities":{"browserName": "firefox"},"requiredCapabilities":{}}' | curl -N -X POST --connect-timeout 20 --data-binary @- -H "Transfer-Encoding: chunked" -H 'Content-Type: application/json; charset=utf-8' -H 'Expect: 100-continue' http://localhost:4444/wd/hub/session -vvv

diemol:
> > @nikhlagrwl could you share the curl call?
> 
> echo '{"desiredCapabilities":{"browserName": "firefox"},"requiredCapabilities":{}}' | curl -N -X POST --connect-timeout 20 --data-binary @- -H "Transfer-Encoding: chunked" -H 'Content-Type: application/json; charset=utf-8' -H 'Expect: 100-continue' http://localhost:4444/wd/hub/session -vvv

Please note, that payload is not valid anymore as it is JWP, not W3C WebDriver.

java/src/org/openqa/selenium/netty/server/RequestConverter.java
```diff
@@ -67,6 +67,7 @@ protected void channelRead0(ChannelHandlerContext ctx, HttpObject msg) throws Ex
       io.netty.handler.codec.http.HttpRequest nettyRequest =
           (io.netty.handler.codec.http.HttpRequest) msg;
 
+      request = createRequest(ctx, nettyRequest);
```
joerg1985: this should be at it's original place

java/src/org/openqa/selenium/netty/server/RequestConverter.java
```diff
@@ -67,6 +67,7 @@ protected void channelRead0(ChannelHandlerContext ctx, HttpObject msg) throws Ex
...
       if (HttpUtil.is100ContinueExpected(nettyRequest)) {
         ctx.write(new HttpResponse().setStatus(100));
         return;
```
joerg1985: This `if` block should be moved below the `if (request == null) {` block to ensure we do set the request in case of a protocol upgrade.

joerg1985 (changes requested):

java/src/org/openqa/selenium/netty/server/RequestConverter.java
nikhlagrwl: Please re-review, resolved comment

joerg1985 (approved):

titusfortner:
please run `./scripts/format.sh` to fix the linting issues

nikhlagrwl:
@titusfortner Resolved linting issues

PR #13126, opened 2023-11-09
[rb] implement toggle to use BiDi implementation

titusfortner:
### Description
If capabilities comes back with a populated `web_socket_url`, extends `Bridge` class with `BiDiBridge` module, to override the classic method and use the bidi method. The existing BiDi classes that are constructed with driver parameters should all accept bridge arguments, which allows this to work.

### Motivation and Context
And so begins the transition from WebDriver-Classic to WebDriver-BiDi


p0deje (approved):
Lets go!

titusfortner:
closing in favor of https://github.com/SeleniumHQ/selenium/pull/14092

PR #13136, opened 2023-11-10
Use Runfiles library as an NuGet package

purkhusid:
### Description
The Runfiles library is now available on NuGet so I switched to that.
Also updated the csprojs to use the same TFMs as the Bazel build

### Motivation and Context
Step toward making the .Net build debuggable via IDEs


titusfortner:
I broke it up, added to it, renamed it and merged separately. Thanks @purkhusid looks good.

PR #13140, opened 2023-11-10
[dotnet] Internal logging

nvborisenko:
We need logging for .net.

### Description

There is single entry point `Log`, but `Context` aware.

Usage:

```csharp
public class Tests
{
    ChromeDriver driver;

    [OneTimeSetUp]
    public void OneTimeSetup()
    {
        // configure global logging
        Log.SetLevel(LogEventLevel.Trace);
    }

    [SetUp]
    public void Setup()
    {
        driver = new ChromeDriver();
    }

    [TearDown]
    public void Tear()
    {
        driver?.Dispose();
    }

    [Test]
    public async Task Test1()
    {
        // adjust logging per test
        using Log.CreateContext(LogEventLevel.Info).Handlers.Add(new ConsoleLogHandler());

        driver.Url = "https://www.yahoo.com/";
    }

    [Test]
    public async Task Test2()
    {
        // It uses globally configured logging
        driver.Url = "https://www.yahoo.com/";
    }
}
```

### Motivation and Context
We need logging for .net.

### Types of changes

### Checklist


YevgeniyShunevych:
Thank you for this feature implementation, @nvborisenko. Let me please review this PR a bit.

Overall implementation is fine, as for me, "ambient context" looks like a good way to inject logging in the current state of WebDriver without breaking changes. I believe in the future WebDriver versions we'll be able to have it injected/configured through `WebDriver` property or constructor parameter.

Anyway, I have a couple of recommendations regarding the logging API.

I would merge `Log` and `LogContext` classes into one public non-static `LogContext` (with few static members) and remove `ILogContext`. Both `Log` and `LogContext` contain similar methods `SetLevel` and `AddHandler`. In case of merge the API can be changed the following way.

### Configure global logging

```cs
Log.SetLevel(Level.Trace);
// ->
LogContext.Global.SetLevel(Level.Trace);
```

Where `LogContext.Global` property and `SetLevel` return the same instance of `LogContext`.

### Adjust logging per test

```cs
Log.ForContext.SetLevel(Level.Trace).AddHandler(new ConsoleLogHandler());
// ->
LogContext.ConfigureAmbient().SetLevel(Level.Trace).AddHandler(new ConsoleLogHandler());
```

Currently, if you call twice `Log.ForContext.AddHandler(new ConsoleLogHandler());` within the same thread, you'll register `ConsoleLogHandler` twice. This definitely can happen for users, and will produce duplicated log entries.

So I would replace `Log.ForContext` property with `LogContext.ConfigureAmbient()`. Calling `LogContext.ConfigureAmbient()` should reset current `_ambientLogContext.Value` with clone of `_globalLogContext` as it is now.

### Logger initialization

```cs
private readonly ILogger _logger = Log.ForContext.GetLogger<HttpCommandExecutor>();
// ->
private readonly ILogger _logger = LogContext.ResolveLogger<HttpCommandExecutor>();
```

`ResolveLogger` under the hood, if `_ambientLogContext.Value` is `null`, should not initialize it, but use `_globalLogContext`.
If someone doesn't use internal logging at all or use only global logging, there is no need to clone `LogContext` (initialize ambient context) for each thread.

dotnet/src/webdriver/Internal/Logging/LogEvent.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+            IssuedBy = issuedBy;
+            TimeStamp = timeStamp;
+            Level = level;
+            Message = message;
+        }
+
+        public Type IssuedBy { get; }
+
+        public DateTime TimeStamp { get; }
```
YevgeniyShunevych: Maybe rename it to `Timestamp` with lower "s", as in old `LogEntry` class.

dotnet/src/webdriver/Internal/Logging/LogLevel.cs
```diff
@@ -0,0 +1,12 @@
+namespace OpenQA.Selenium.Internal.Logging
+{
+    public enum Level
```
YevgeniyShunevych: Enum named `Level`, but file is `LogLevel.cs`. I understand that we already have `LogLevel` in the library and we have a name conflict. I propose to name this enum as `LogEventLevel`. For example, Serilog uses such name for analogous log level enum.

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
```diff
@@ -162,6 +166,8 @@ public virtual Response Execute(Command commandToExecute)
                 throw new ArgumentNullException(nameof(commandToExecute), "commandToExecute cannot be null");
             }
 
+            _logger.Trace($"Executing http command: {commandToExecute}");
```
YevgeniyShunevych:
```suggestion
            _logger.Trace($"Executing HTTP command: {commandToExecute}");
```

dotnet/src/webdriver/Internal/Logging/LogContext.cs
```diff
@@ -0,0 +1,79 @@
...
+            }
+
+            _handlers.Add(handler);
+
+            return this;
+        }
+
+        public object Clone()
+        {
+            return new LogContext(_level, _handlers?.ToList());
```
YevgeniyShunevych: During cloning you create a new list of handlers by `_handlers?.ToList()`, which is OK. But then inside `LogContext` another `List` instance is created:
```cs
_handlers = new List<ILogHandler>(handlers);
```
So 2 new lists instantiated instead of 1 during cloning.

dotnet/src/webdriver/SeleniumManager.cs
```diff
@@ -34,6 +35,8 @@ namespace OpenQA.Selenium
     /// </summary>
     public static class SeleniumManager
     {
+        private static readonly ILogger _logger = Log.ForContext.GetLogger(typeof(SeleniumManager));
```
YevgeniyShunevych: In static classes I think it is better not to store `ILogger` in a field, but get logger dynamically right in the place where you do logging (before `_logger.Trace(...)`). Because currently only the first occurred `LogContext`'s logger will be written to the field and used in all cases.

dotnet/src/webdriver/Internal/Logging/LogLevel.cs
nvborisenko: Renamed to `LogEventLevel`

dotnet/src/webdriver/Internal/Logging/ILogContext.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public interface ILogContext : ICloneable
+    {
+        ILogger GetLogger<T>();
+
+        ILogger GetLogger(Type type);
+
+        void LogMessage(ILogger logger, LogEventLevel level, string message);
+
+        ILogContext SetLevel(LogEventLevel level);
```
nvborisenko, line 68: `SetLevel()` seems non-native for C#. Using property would be better, but we need think about providing "one-line" configuration (Builder? and in future?).

dotnet/src/webdriver/Internal/Logging/ILogger.cs
```diff
@@ -0,0 +1,21 @@
...
+
+        void Debug(string message);
+
+        void Info(string message);
+
+        void Warn(string message);
+
+        void Error(string message);
+
+        LogEventLevel Level { get; set; }
```
nvborisenko, line 61: As discussed with @YevgeniyShunevych we don't need it, at least at this phase of implementation.

dotnet/src/webdriver/Internal/Logging/LogEvent.cs
nvborisenko: Done, now it is `Timestamp`.

dotnet/src/webdriver/Internal/Logging/LogContext.cs
nvborisenko: Reworked cloning mechanism. `LogContext` object should have own list of nested loggers/handlers.

dotnet/src/webdriver/Internal/Logging/ILogContext.cs
nvborisenko: I had an opportunity to work on it and now we  can use it like
```csharp
Log.SetMinimumLevel(LogEventLevel.Debug)).WithHandler(new ConsoleLogHandler());
```

dotnet/src/webdriver/SeleniumManager.cs
nvborisenko: I reworked it, in general we want to use `private static readonly` in all cases. It guarantees that class captures `ILogger` one time. Then log's emission happens later, based on current log context.

titusfortner:
Can you update this documentation below with the correct info? I can't quite tell looking at the code. Also, should we have unit tests to make sure the logging works as expected?

### Getting a Logger

.NET logger is managed with a static class, so all access to logging is managed simply by referencing `Log`

### Logger Level
.NET has 6 logger levels: `Error`, `Warn`, `Info`, `Debug`, `Trace` and `None`. The default level is `Info`.

To change the level of the logger:
```csharp
using OpenQA.Selenium.Internal.Logging;

Log.SetLevel(LogEventLevel.Trace);
```

### Actionable Items
.NET logs actionable content at logger level `Warn`
Example:
```
11:04:40.986  WARN LoggingTest: this is a warning
```

### Useful Information
.NET logs useful information at logger level `Info`
Example:
```
11:04:40.986  INFO LoggingTest: this is useful information
```

### Debug Information
.NET logs most debug content at logger level `Debug`
Example:
```
11:04:40.986 DEBUG LoggingTest: this is detailed debug information
```

### Logger Output
By default all logs are sent to `Console.Error` output.

### Logger filtering
To set the level you want to use on a per-class basis:
```csharp
Log.SetLevel(typeof(LoggingTest), LogEventLevel.Trace);
```

Or per execution context:
```csharp
using (var ctx = Log.CreateContext(LogEventLevel.Trace).SetMinimumLevel(typeof(LoggingTest), LogEventLevel.Info))
{
  // everything here emits log messages according configured context
}
```

dotnet/src/webdriver/Internal/Logging/ILogger.cs
nvborisenko: I figured it out and now we configure log event level per logger. I should still make `ILogger` as internal.

nvborisenko:
Unit tests are mandatory, working on it.

dotnet/src/webdriver/Internal/Logging/ConsoleLogHandler.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// </summary>
+    public class ConsoleLogHandler : ILogHandler
+    {
+        private static readonly Dictionary<LogEventLevel, string> _levelMap = new Dictionary<LogEventLevel, string>
+        {
+            { LogEventLevel.Trace, "TRC" },
+            { LogEventLevel.Debug, "DBG" },
+            { LogEventLevel.Info, "INF" },
+            { LogEventLevel.Warn, "WRN" },
+            { LogEventLevel.Error, "ERR" }
```
YevgeniyShunevych: This is subjective. But anyway, I would prefer log levels this way.

```suggestion
            { LogEventLevel.Trace, "TRACE" },
            { LogEventLevel.Debug, "DEBUG" },
            { LogEventLevel.Info, " INFO" },
            { LogEventLevel.Warn, " WARN" },
            { LogEventLevel.Error, "ERROR" }
```

nvborisenko: Idea here is to make output prefix beeing length-fixed:
```
16:36:10.763 TRC SeleniumManager: Driver path: C:\Users\nikol\.cache\selenium\chromedriver\win64\119.0.6045.105\chromedriver.exe
16:36:10.836 DBG HttpCommandExecutor: Executing command: []: newSession {"capabilities":{"firstMatch":[{"browserName":"chrome","
```

YevgeniyShunevych: `" INFO"` and `" WARN"`, which are shorter than 5 characters, can be prefixed with a space character.

nvborisenko: Also works. both are lenght-fixed. @titusfortner what do you prefer?

titusfortner: Yeah, harder to read the abbreviations I think

nvborisenko: From the other hand why not `Information` rather than `Info`, or `Warning` rather that `Warn`.

titusfortner: Java & Python do `Warning`, Ruby does `Warn`, so either is fine. I'm not worried about the spacing, either. Most of this is convention for the given language, but Info, Warn, Debug are all English words.

nvborisenko: Did it like @YevgeniyShunevych proposed.

```
"TRACE", "DEBUG", " INFO", " WARN", "ERROR"
```

dotnet/src/webdriver/Internal/Logging/ILogger.cs
nvborisenko: Made `ILogger` as internal, so we don't care about its API.

titusfortner:
> To direct output to a console, you need to add a handler:

1. Where are logs sent by default?
2. How do we send logs to a file?

titusfortner:
What is the thought behind naming it `SetMinimumLevel` instead of just `SetLevel`?

nvborisenko:
> Where are logs sent by default?

No logs by default at all. User should explicitly turn it on.

> How do we send logs to a file?

Should we implement it as well?

> What is the thought behind naming it `SetMinimumLevel` instead of just `SetLevel`?

It is clearer that SetMinimumLevel(Trace) will emit logs with `Debug` level also, not only with `Trace` level. I think we can rename it to `SetLevel` just to be aligned with others bindings.

titusfortner:
I really think we should turn it on by default, and just not log anything at the default level unless it is important.

Yes, ideally we have a Handler implementation that can stream to stdout, stderr or file

nvborisenko:
I returned back `SetLevel()`. By default now we use `ConsoleLogHandler` which outputs to strerr.

dotnet/src/webdriver/Internal/Logging/ILogContext.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <param name="level">The minimum log level.</param>
+        /// <returns>The current instance of <see cref="ILogContext"/> with the minimum log level set for the specified type.</returns>
+        ILogContext SetLevel(Type issuer, LogEventLevel level);
+
+        /// <summary>
+        /// Adds a log handler to the current context.
+        /// </summary>
+        /// <param name="handler">The log handler to add.</param>
+        /// <returns>The current instance of <see cref="ILogContext"/> with the log handler added.</returns>
+        ILogContext AddHandler(ILogHandler handler);
```
nvborisenko: I will redesign it, we have to have a possibility to remove handlers (since `ConsoleLogHandler` is registered by default)

titusfortner: Since this is just for Selenium, I think we're ok just supporting one? (e.g., setting a new handler replaces the old handler)

nvborisenko: We support many handlers. I need just expose some API to remove some of them. Not hard change, but it should be done before release.

nvborisenko: Resolved. Now we can use `Log.Handlers.Add(new Handler())`

titusfortner: I guess I mean that for the other languages setting to output to a file doesn't also output to console

nvborisenko: Hidden stones! I (user) would like to manage all handlers.

```csharp
Log.Handlers
  .Clear()
  .Add(new ConsoleLogHandler())
  .Add(new SqlLogHandler())
  .Add(new ReportPortalLogHandler());
```

titusfortner: But I don't see a good use case for having multiple handlers, so it becomes just an extra step.

That said, I don't have strong feelings on it so long as it is well documented.

nvborisenko: Logging is event-based. Event may have subsriber**s**. Log handler is a subsriber.

Moreover reading docs for java impl:
image

titusfortner:
Do you know what the test failures are from?

nvborisenko:
This PR is ready for review. API is stabilized, user is able to do whatever he wants. We still don't have the implementation of `FileLogHandler`, let's leave it for next releases (writing to a file is not easy task in terms of efficiency).

Build is failing, I have no ideas why.

YevgeniyShunevych (approved):

titusfortner:
Thanks for all of your work on this @YevgeniyShunevych & @nvborisenko !!

PR #13142, opened 2023-11-11
Set a name for thread started to forward output of external processes

centic9:
The thread started to forward output of external processes currently does not have a name. 

In some of our applications, we sometimes see some threads being left over.

In order to make it easier to analyze the root-cause, it would be good to have a name set also for this thread.

Thus this PR simply sets a more useful thread-name for this thread.


diemol (commented):
This makes sense. Can you improve this by adding the binary that the class is trying to launch? Otherwise you get more than one thread and they all have the same name.

centic9:
> This makes sense. Can you improve this by adding the binary that the class is trying to launch? Otherwise you get more than one thread and they all have the same name.
Sure, I added the first part of the command to include the name of the executable that is called.

pujagani:
Please run the `./scripts/format.sh` script on your branch to help fix the failing CI. Thank you!

diemol (approved):

PR #13162, opened 2023-11-16
don't leak a file descriptor to os.devnull by default

asottile:
since this is passed along to subprocess directly we can use the subprocess constants still

### Description

avoid one `open` call without a close -- this doesn't fix the other one in the file but I am less concerned with that one since I am only using the default behaviour

I also adjusted `is` to `==` for the subprocess constants -- they are not guaranteed to be singletons so they should be compared with equality

### Motivation and Context

regression in #12103

### Types of changes

### Checklist

- [x] My change requires a change to the documentation. **N/A**
- [x] I have updated the documentation accordingly. **N/A**
- [x] I have added tests to cover my changes. **N/A**

### DIFF
--- a/py/selenium/webdriver/common/service.py
+++ b/py/selenium/webdriver/common/service.py
@@ -135,7 +135,7 @@ class Service(ABC):
     def stop(self) -> None:
         """Stops the service."""
 
-        if self.log_output != PIPE:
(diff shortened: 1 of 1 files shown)


titusfortner:
Thank you!

PR #13173, opened 2023-11-20
[rust] Tracking Selenium Manager usage through Plausible (#11211)

### LINKED ISSUE #11211: [🚀 Feature]: Implement download tracking in selenium manager
### Feature and motivation

The idea is to get basic usage information about who is using Selenium Manager. 

We could create a basic page on `https://www.selenium.dev/selenium/manager`. The Selenium Manager would route all requests for downloads through that page.

Something like:
```
https://selenium.dev/selenium/download/manager?language=ruby&os=linux&driver=geckodriver&driver_version=v0.32.0
```

Which would redirect and return the correct file, but would allow us to see what language/os/browser/browser versions are being used.

bonigarcia:
### Description
This PR implements a mechanism to track the usage of Selenium Manager through [Plausible](https://plausible.io/). 

To that, the Rust code sends a `pageview` request to plausible using the events API to track visitors. As requested in #11211, the data we want to collect is the following:

- Selenium version
- Language binding 
- Browser
- Browser version
- Operating system
- Architecture

For "language binding", a new flag called `--language-binding` has been introduced in this PR. Bindings should use this property to report it (e.g., `--language-binding java`, `--language-binding ruby`, etc.).

All of this information is gathered through custom properties (called `props` in the [Plausible Event API](https://plausible.io/docs/events-api)). The (fake) page to track these visitors will be the following:

https://selenium.dev/sm-stats

An example of the equivalent POST request to track a page view is the following:

```
curl -i -X POST https://plausible.io/api/event \
  -H 'User-Agent: Selenium Manager 0.4.16' \
  -H 'Content-Type: application/json' \
  --data '{"name":"pageview","url":"https://selenium.dev/sm-stats","domain":"selenium.dev","props":{"browser":"chrome","browser_version":"119.0.6045.105","os":"linux","arch":"x86_64","lang":"java","selenium_version":"4.16"}}'
```

The data gathered by Plausible will be available on:

https://plausible.io/selenium.dev

To configure the custom properties (e.g., for filtering), these properties should be enabled:

https://plausible.io/selenium.dev/settings/properties

I tested this feature using a free Plausible account and my personal domain  (bonigarcia.dev):

image

This PR is already pointing to selenium.dev. @diemol @titusfortner If you want to check how data is gathered, you can trigger the [CI Rust workflow](https://github.com/SeleniumHQ/selenium/actions/workflows/ci-rust.yml) using the branch `sm_plausible`.

NOTE: I was concerned about the impact of this feature on the overall performance of Selenium Manager. To get some numbers, I calculated the average time of running the Selenium Manager tests using the `trunk` branch (i.e., without this feature) compared to the branch `sm_plausible`. 

For that, I executed the following command in Linux:

```
time cargo test
```

I executed this command (with every asset in the cache) 10 times for each branch, and I calculated the average total time.

average_total_time_trunk = 11.1399 seconds
average_total_time_plausible = 12.4713 seconds

Therefore, the overhead introduced to the requests made to Plausible averages 0.7341 seconds. Since 23 page views were requested during the tests, each `pageview` request added an overhead of around 32 ms, which seems reasonable.

### Motivation and Context
This PR resolves #11211.

### Types of changes

### Checklist

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -829,7 +829,7 @@ pub trait SeleniumManager {
         Ok(driver_path)
     }
 
-    fn stats(&self) {
+    fn stats(&self) -> Result<(), Error> {
         let props = Props {
             browser: self.get_browser_name().to_string(),
             browser_version: self.get_browser_version().to_string(),
@@ -839,6 +839,7 @@ pub trait SeleniumManager {
             selenium_version: self.get_selenium_version().to_string(),
         };
         send_stats_to_plausible(self.get_http_client(), props, self.get_logger());
+        Ok(())
     }
 
     fn check_error_with_driver_in_path(
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -837,12 +837,12 @@ pub trait SeleniumManager {
     fn stats(&self) -> Result<(), Error> {
         if !self.is_avoid_stats() && !self.is_offline() {
             let props = Props {
-                browser: self.get_browser_name().to_string(),
-                browser_version: self.get_browser_version().to_string(),
-                os: self.get_os().to_string(),
-                arch: self.get_arch().to_string(),
-                lang: self.get_language_binding().to_string(),
-                selenium_version: self.get_selenium_version().to_string(),
+                browser: self.get_browser_name().to_ascii_lowercase(),
+                browser_version: self.get_browser_version().to_ascii_lowercase(),
+                os: self.get_os().to_ascii_lowercase(),
+                arch: self.get_arch().to_ascii_lowercase(),
+                lang: self.get_language_binding().to_ascii_lowercase(),
+                selenium_version: self.get_selenium_version().to_ascii_lowercase(),
             };
             let http_client = self.get_http_client().to_owned();
             let sender = self.get_sender().to_owned();
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -45,6 +45,7 @@ use is_executable::IsExecutable;
 use reqwest::{Client, Proxy};
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
+use std::sync::mpsc::{Receiver, Sender};
 use std::time::Duration;
 use std::{env, fs, thread};
 use walkdir::DirEntry;
@@ -152,6 +153,10 @@ pub trait SeleniumManager {
     fn get_logger(&self) -> &Logger;
 
     fn set_logger(&mut self, log: Logger);
+
+    fn get_sender(&self) -> &Sender<String>;
+
+    fn get_receiver(&self) -> &Receiver<String>;
 
     fn get_platform_label(&self) -> &str;
 
@@ -839,9 +844,10 @@ pub trait SeleniumManager {
                 lang: self.get_language_binding().to_string(),
                 selenium_version: self.get_selenium_version().to_string(),
             };
-            let http_client = self.get_http_client().clone();
+            let http_client = self.get_http_client().to_owned();
+            let sender = self.get_sender().to_owned();
             thread::spawn(move || {
-                send_stats_to_plausible(http_client, props);
+                send_stats_to_plausible(http_client, props, sender);
             });
         }
         Ok(())
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -830,15 +830,13 @@ pub trait SeleniumManager {
     }
 
     fn stats(&self) {
-        let sm_version = env!("CARGO_PKG_VERSION");
-        let selenium_version = sm_version.strip_prefix(SM_BETA_LABEL).unwrap_or(sm_version);
         let props = Props {
             browser: self.get_browser_name().to_string(),
             browser_version: self.get_browser_version().to_string(),
             os: self.get_os().to_string(),
             arch: self.get_arch().to_string(),
             lang: self.get_language_binding().to_string(),
-            selenium_version: selenium_version.to_string(),
+            selenium_version: self.get_selenium_version().to_string(),
         };
         send_stats_to_plausible(self.get_http_client(), props, self.get_logger());
     }
@@ -1452,6 +1450,16 @@ pub trait SeleniumManager {
     fn set_language_binding(&mut self, language_binding: String) {
         if !language_binding.is_empty() {
             self.get_config_mut().language_binding = language_binding;
+        }
+    }
+
+    fn get_selenium_version(&self) -> &str {
+        self.get_config().selenium_version.as_str()
+    }
+
+    fn set_selenium_version(&mut self, selenium_version: String) {
+        if !selenium_version.is_empty() {
+            self.get_config_mut().selenium_version = selenium_version;
         }
     }
 }
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -830,15 +830,17 @@ pub trait SeleniumManager {
     }
 
     fn stats(&self) -> Result<(), Error> {
-        let props = Props {
-            browser: self.get_browser_name().to_string(),
-            browser_version: self.get_browser_version().to_string(),
-            os: self.get_os().to_string(),
-            arch: self.get_arch().to_string(),
-            lang: self.get_language_binding().to_string(),
-            selenium_version: self.get_selenium_version().to_string(),
-        };
-        send_stats_to_plausible(self.get_http_client(), props, self.get_logger());
+        if !self.is_avoid_stats() {
+            let props = Props {
+                browser: self.get_browser_name().to_string(),
+                browser_version: self.get_browser_version().to_string(),
+                os: self.get_os().to_string(),
+                arch: self.get_arch().to_string(),
+                lang: self.get_language_binding().to_string(),
+                selenium_version: self.get_selenium_version().to_string(),
+            };
+            send_stats_to_plausible(self.get_http_client(), props, self.get_logger());
+        }
         Ok(())
     }
 
@@ -1461,6 +1463,16 @@ pub trait SeleniumManager {
     fn set_selenium_version(&mut self, selenium_version: String) {
         if !selenium_version.is_empty() {
             self.get_config_mut().selenium_version = selenium_version;
+        }
+    }
+
+    fn is_avoid_stats(&self) -> bool {
+        self.get_config().avoid_stats
+    }
+
+    fn set_avoid_stats(&mut self, avoid_stats: bool) {
+        if avoid_stats {
+            self.get_config_mut().avoid_stats = true;
         }
     }
 }
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -835,7 +835,7 @@ pub trait SeleniumManager {
     }
 
     fn stats(&self) -> Result<(), Error> {
-        if !self.is_avoid_stats() {
+        if !self.is_avoid_stats() && !self.is_offline() {
             let props = Props {
                 browser: self.get_browser_name().to_string(),
                 browser_version: self.get_browser_version().to_string(),
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -46,7 +46,7 @@ use reqwest::{Client, Proxy};
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::time::Duration;
-use std::{env, fs};
+use std::{env, fs, thread};
 use walkdir::DirEntry;
 
 pub mod chrome;
@@ -839,7 +839,10 @@ pub trait SeleniumManager {
                 lang: self.get_language_binding().to_string(),
                 selenium_version: self.get_selenium_version().to_string(),
             };
-            send_stats_to_plausible(self.get_http_client(), props, self.get_logger());
+            let http_client = self.get_http_client().clone();
+            thread::spawn(move || {
+                send_stats_to_plausible(http_client, props);
+            });
         }
         Ok(())
     }
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -38,6 +38,7 @@ use crate::safaritp::{SafariTPManager, SAFARITP_NAMES};
 use crate::shell::{
     run_shell_command, run_shell_command_by_os, run_shell_command_with_log, split_lines, Command,
 };
+use crate::stats::{send_stats_to_plausible, Props};
 use anyhow::anyhow;
 use anyhow::Error;
 use is_executable::IsExecutable;
@@ -62,6 +63,7 @@ pub mod mirror;
 pub mod safari;
 pub mod safaritp;
 pub mod shell;
+pub mod stats;
 
 pub const REQUEST_TIMEOUT_SEC: u64 = 300; // The timeout is applied from when the request starts connecting until the response body has finished
 pub const STABLE: &str = "stable";
@@ -112,6 +114,7 @@ pub const NOT_ADMIN_FOR_EDGE_INSTALLER_ERR_MSG: &str =
     "{} can only be installed in Windows with administrator permissions";
 pub const ONLINE_DISCOVERY_ERROR_MESSAGE: &str = "Unable to discover {}{} in online repository";
 pub const UNC_PREFIX: &str = r"\\?\";
+pub const SM_BETA_LABEL: &str = "0.";
 
 pub trait SeleniumManager {
     // ----------------------------------------------------------
@@ -826,6 +829,20 @@ pub trait SeleniumManager {
         Ok(driver_path)
     }
 
+    fn stats(&self) {
+        let sm_version = env!("CARGO_PKG_VERSION");
+        let selenium_version = sm_version.strip_prefix(SM_BETA_LABEL).unwrap_or(sm_version);
+        let props = Props {
+            browser: self.get_browser_name().to_string(),
+            browser_version: self.get_browser_version().to_string(),
+            os: self.get_os().to_string(),
+            arch: self.get_arch().to_string(),
+            lang: self.get_language_binding().to_string(),
+            selenium_version: selenium_version.to_string(),
+        };
+        send_stats_to_plausible(self.get_http_client(), props, self.get_logger());
+    }
+
     fn check_error_with_driver_in_path(
         &mut self,
         is_driver_in_path: &bool,
@@ -1425,6 +1442,16 @@ pub trait SeleniumManager {
     fn set_cache_path(&mut self, cache_path: String) {
         if !cache_path.is_empty() {
             self.get_config_mut().cache_path = cache_path;
+        }
+    }
+
+    fn get_language_binding(&self) -> &str {
+        self.get_config().language_binding.as_str()
+    }
+
+    fn set_language_binding(&mut self, language_binding: String) {
+        if !language_binding.is_empty() {
+            self.get_config_mut().language_binding = language_binding;
         }
     }
 }
--- a/rust/src/main.rs
+++ b/rust/src/main.rs
@@ -24,11 +24,11 @@ use selenium_manager::config::{BooleanKey, StringKey, CACHE_PATH_KEY};
 use selenium_manager::grid::GridManager;
 use selenium_manager::logger::{Logger, BROWSER_PATH, DRIVER_PATH};
 use selenium_manager::metadata::clear_metadata;
-use selenium_manager::REQUEST_TIMEOUT_SEC;
 use selenium_manager::TTL_SEC;
 use selenium_manager::{
     clear_cache, get_manager_by_browser, get_manager_by_driver, SeleniumManager,
 };
+use selenium_manager::{REQUEST_TIMEOUT_SEC, SM_BETA_LABEL};
 use std::backtrace::{Backtrace, BacktraceStatus};
 use std::path::Path;
 use std::process::exit;
@@ -203,6 +203,9 @@ fn main() {
     selenium_manager.set_cache_path(cache_path.clone());
     selenium_manager.set_offline(cli.offline);
     selenium_manager.set_language_binding(cli.language_binding.unwrap_or_default());
+    let sm_version = clap::crate_version!();
+    let selenium_version = sm_version.strip_prefix(SM_BETA_LABEL).unwrap_or(sm_version);
+    selenium_manager.set_selenium_version(selenium_version.to_string());
 
     if cli.clear_cache || BooleanKey("clear-cache", false).get_value() {
         clear_cache(selenium_manager.get_logger(), &cache_path);
--- a/rust/src/main.rs
+++ b/rust/src/main.rs
@@ -217,6 +217,7 @@ fn main() {
     selenium_manager
         .set_timeout(cli.timeout)
         .and_then(|_| selenium_manager.set_proxy(cli.proxy.unwrap_or_default()))
+        .and_then(|_| selenium_manager.stats())
         .and_then(|_| selenium_manager.setup())
         .map(|driver_path| {
             let log = selenium_manager.get_logger();
@@ -225,7 +226,6 @@ fn main() {
                 &driver_path,
                 &selenium_manager.get_browser_path_or_latest_from_cache(),
             );
-            selenium_manager.stats();
             flush_and_exit(OK, log, None);
         })
         .unwrap_or_else(|err| {
--- a/rust/src/main.rs
+++ b/rust/src/main.rs
@@ -137,6 +137,10 @@ struct Cli {
     /// DotNet, Ruby)
     #[clap(long)]
     language_binding: Option<String>,
+
+    /// Avoid sends usage statistics to plausible.io
+    #[clap(long)]
+    avoid_stats: bool,
 }
 
 fn main() {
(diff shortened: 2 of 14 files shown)


rust/src/lib.rs
```diff
@@ -825,6 +828,18 @@ pub trait SeleniumManager {
...
+    fn stats(&self) {
+        let props = Props {
+            browser: self.get_browser_name().to_string(),
+            browser_version: self.get_browser_version().to_string(),
+            os: self.get_os().to_string(),
+            arch: self.get_arch().to_string(),
+            lang: self.get_language_binding().to_string(),
+            selenium_version: self.get_selenium_version().to_string(),
+        };
+        send_stats_to_plausible(self.get_http_client(), props, self.get_logger());
```
diemol: Is this call blocking, or is it sent in a different thread? We do not want to block the main thread to send this. 

Also, we need an environment variable to allow users to opt-out.

bonigarcia: > Is this call blocking, or is it sent in a different thread? We do not want to block the main thread to send this.

It's a blocking call. Everything happens in the same thread in Selenium Manager. But I can explore how to spawn a new thread for this call.

> Also, we need an environment variable to allow users to opt-out.

I was not aware it was a requirement. But I can implement it.

diemol: >I was not aware it was a requirement. But I can implement it.

Yeah. We did not finish the conversation on #11211 before this PR was created. When implementing tracking, we need to give users the possibility to opt-out. 

>It's a blocking call. Everything happens in the same thread in Selenium Manager. But I can explore how to spawn a new thread for this call.

We need to make it non-blocking because it should not affect the user if this call fails.

bonigarcia: I have just added two commits to this PR to implement the requested features. Now, the POST request to plausible.io is done asynchronously. Also, the argument `--avoid-stats` is accepted by Selenium Manager for not sending the statistics to plausible.io. As usual, this argument can be configured using the configuration file or an env (`SE_AVOID_STATS=true`).

titusfortner:
Excellent. Have we tested the output? What does it look like on Plausible?

titusfortner:
@bonigarcia is this ready?
@diemol do we see what the output looks like in Plausible?

bonigarcia:
> is this ready?

Yes, it is.

diemol:
@bonigarcia 

>NOTE: I was concerned about the impact of this feature on the overall performance of Selenium Manager. To get some numbers, I calculated the average time of running the Selenium Manager tests using the trunk branch (i.e., without this feature) compared to the branch sm_plausible.

Is all still the same after you have placed the call in a non blocking thread?

bonigarcia:
> Is all still the same after you have placed the call in a non blocking thread?

The call to Plausible is now executed in a different thread, so the time overhead is null.

rust/src/main.rs
```diff
@@ -208,10 +223,16 @@ fn main() {
     selenium_manager
         .set_timeout(cli.timeout)
         .and_then(|_| selenium_manager.set_proxy(cli.proxy.unwrap_or_default()))
+        .and_then(|_| selenium_manager.stats())
```
diemol, line 226: Could we move the stats creation and upload at the end? The "primary" tasks of SM should go first and these type of things at the end, ideally to prevent any issues. 

I mean, if there is an issue with the `stats()` code, would the execution continue?

bonigarcia: The call to Plausible (`.stats()`) is done in a different thread. This way, even if Plausible is offline or any other unexpected error happens, the regular behavior of Selenium Manager will not be affected since the `.setup()` logic is executed in the main thread. I have tested it.

titusfortner:
@bonigarcia can you resolve conflicts here?

bonigarcia:
> can you resolve conflicts here?

Done.

rust/src/main.rs
diemol: 👍

rust/src/stats.rs
```diff
@@ -0,0 +1,76 @@
...
+use reqwest::Client;
+use serde::{Deserialize, Serialize};
+use std::sync::mpsc::Sender;
+use std::time::Duration;
+
+const PLAUSIBLE_URL: &str = "https://plausible.io/api/event";
+const SM_USER_AGENT: &str = "Selenium Manager {}";
+const APP_JSON: &str = "application/json";
+const PAGE_VIEW: &str = "pageview";
+const SELENIUM_DOMAIN: &str = "selenium.dev";
```
diemol:
```suggestion
const SELENIUM_DOMAIN: &str = "manager.selenium.dev";
```

diemol (commented):
I changed the domain because I created one to only have the SM stats and avoid mixing them with the website.

image

bonigarcia:
> I changed the domain because I created one to only have the SM stats and avoid mixing them with the website.

Great.

rust/src/stats.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+#[derive(Default, Serialize, Deserialize)]
+pub struct Props {
+    pub browser: String,
+    pub browser_version: String,
+    pub os: String,
+    pub arch: String,
+    pub lang: String,
+    pub selenium_version: String,
```
diemol, line 49: I know we did not mention this before, but... should we add the driver version? What do you think @bonigarcia?

bonigarcia: I suppose it is not very relevant since most users would use the browser version (`browserVersion` set through options) but not the driver version.

BTW, the browser version sent to plausible is that value, i.e., what the user specifies through options (e.g., `"beta"`, `"120"`, `""`) and not the discovered browser version (e.g., `120.0.6099.71`, etc.). Is that what we want, right?

diemol: I think it makes sense to avoid the driver version then.

Regarding the browser version... I think we should collect both. Does `requested_browser_version` sound good?

bonigarcia: It can be done (gather both `requested_browser_version` and `browser_version`). But if we do that, we should send the stats after executing the driver/browser management. In other words, the call to `.stats()` is done after `.setup()`.

In that case, and since after the `.setup()` SM is almost finished (it only needs to log the output), we should wait for the `.stats()` to be finished (since now it is executed in a separate thread). Alternatively, we should execute both `.stats()` and `.setup()` in the same thread (as it was done in the original PR). In any case, the call to Plausible will include some ms overhead (as reported in the description of this PR).

Do we want to do that?

diemol: That is a very good point. 

This also means that, with the current implementation, most of the time the `browser_version` value is empty because users normally don't specify a browser version.

To avoid loosing data, what if we execute `.stats()` in the same thread but shorten the read timeout to 1 second? In theory, we should not need to wait for a response. 

One last thing I noticed is that executing `.stats()` is not guarded by `selenium_manager.is_offline()`. I mean, if the user wants an offline execution, we should not send stats either.

titusfortner: Can't we create the second process at the end of the setup() and not worry about the timeout?

diemol: The problem is that the binary will exit and we might not send the POST completely.

titusfortner: ah, so we're just taking advantage of the fact that setup() takes longer than stats() with the current implementation?

diemol: I would say so.

titusfortner: I actually think it is ok not to have discovered version. All that tells us is what the user happened to have installed on their machine? I don't think it is worth the effort to capture it...

diemol: Why not? I think it is not a big effort, we are just trying to be careful to avoid blocking the user. I think the data is worth to have.

titusfortner: My point is that if it adds a second to each test it's not worth the tradeoff. I'm more interested in what people request than what they have installed. Most likely what we'll see installed is the latest version of chrome because it auto updates.

titusfortner: Actually, this is a good point. Enabling Selenium Manager on Grid is going to look different than using it in client. Can we differentiate when grid is requesting and when the Java client is rquesting?

diemol: It does not add a second to each test, it should be a few ms as Boni pointed out. Worst case, it would add 1 second. We could even shorten that to 500ms.

bonigarcia: > One last thing I noticed is that executing `.stats()` is not guarded by `selenium_manager.is_offline()`. I mean, if the user wants an offline execution, we should not send stats either.

Good catch. I would add that.

> ah, so we're just taking advantage of the fact that setup() takes longer than stats() with the current implementation?

Exactly.

> Actually, this is a good point. Enabling Selenium Manager on Grid is going to look different than using it in client. Can we differentiate when grid is requesting and when the Java client is rquesting?

I would say that the Grid implementation could use the flag `--language-binding` to report it. Perhaps: `--language-binding grid`.

titusfortner: Can we just fire the event and not wait for the response since we aren't going to do anything with it, anyway?

bonigarcia: I like the current implementation because it does not impact the overall performance. Moving the `.stats()` call to the end of the SM logic will affect some ms each time SM is executed. It is worth mentioning that in my tests (i.e., using the GHA infrastructure), this overhead is an average of 32 ms. But for any user, this figure will depend on the underlying network latency.

I tried to implement a *fire and forget* approach for the POST requests to Plausible in the Rust logic. But if the client doesn't wait for the response, Plausible does not count it. So, for gathering the discovered browser version (I can also include the found driver version for the same prize), I'm afraid we need to wait for the response, adding some MS overhead per test.

bonigarcia: I have just added a commit to check the offline mode.

Please let me know your final decision about the current discussion. There are two alternatives:

a) Move the call to `stats()` at the end of SM logic and add the discovered browser version (and found driver version if you want) to the data sent to Plausible. This call should be blocking, and so, it will add a little overhead.
b) Leave it in the current way, in which the browser version is the one reported by the user and `stats()` is done at the beginning (concurrently).

diemol: I think we could go with option a) and iterate. Maybe we do not need a perfect solution now, just something that helps us understand usage.

titusfortner: My vote is to keep existing implementation and just get the browser-version passed by the user because that's the least impact.

titusfortner: "usage" comes from what the user passes in. Knowing what a user has on their system seems extra. I think we should start with the least-impact and add to it if we think we need more details.

diemol: 🤦 

I meant option `b)`, so as it currently is.

bonigarcia: My vote, if I have one, is b. For me, it is better if this feature is not impacting the overall performance, even if it is a few ms.

diemol: @bonigarcia,  of course you have a say!

titusfortner: And we all agree, lol. We can definitely revisit after we start seeing data.

diemol, line 49: Can we lower case all these properties, please? I just tried it and we get different values if they are sent with different casing.

diemol (changes requested):

rust/src/stats.rs
diemol: e.g.

image

bonigarcia: > Can we lower case all these properties, please?

I have committed a patch to send them in lowercase.

diemol: Great! Let's merge this and I will implement the extra flag in the different languages.

diemol (approved):
Thank you, @bonigarcia!

zfouts:
What's the config option to opt out?  I don't like the idea of this doing an API call every time I run selenium. I've blocked it at my edge, but would prefer if we can just straight up disable the API calls.

titusfortner:
Oh, I never posted the blog announcing the release
https://github.com/SeleniumHQ/seleniumhq.github.io/pull/1564/files

Environment variable SE_AVOID_STATS or with the config file

titusfortner:
And for what it's worth it only reports usage once a day by default based on the ttl setting.

bonigarcia:
> And for what it's worth it only reports usage once a day by default based on the ttl setting.

That is not correct. TTL is used only to give validity to the metadata. The metadata contains the knowledge of the proper driver version for a given browser version based on the response of online endpoints (e.g., CfT). The idea of using the TTL period for sending the stats to Plausible was proposed somewhere, but that decision was not made.

rjhancock:
1) I wish this was communicated more publicly before it was put in place. I do understand the desire and need to know the information, but make it opt-in vs opt-out.

2) The method described for opting out appears to be a no-op unless it is applied in a library that is not included within the entire SeleniumHQ organization [as it only appears thrice](https://github.com/search?q=org%3ASeleniumHQ+SE_AVOID_STATS&type=code) and not a single place to disable sending actually occurs within the code.

3) It sends those reports multiple times. Just within my test suite it has attempted to send the info at least a dozen times across 1800 tests. Since it is also used in production, it would be reporting thousands of times A DAY just from our use case.

titusfortner:
We're going to reduce the frequency of reports in the next version, but I'm not sure I understand your underlying concern.

rjhancock:
The underlying concern is some of us use Selenium in more sensitive networks and now we have more work to protect the privacy of what we do to disable this "feature" that was opt-out vs opt-in.

I understand the desire to collect such metrics but a more diplomatic way of doing it, in the face of ever growing invasive privacy and tracking issues, would have been "we've added telemetry collection on an opt-in basis. It would help us to know where to focus our efforts if you'd turn it on" vs "<no mention except a single line in a readme file>"

Currently downgraded the version used until I can put in a place to permanently block all telemetry data as the above mentioned environment variable doesn't seem to be checked within the code anywhere and doesn't appear to be a known one for other systems either.

diemol:
Regarding privacy, it is worth mentioning that we use https://plausible.io/, which is privacy first and GDPR compliant.

titusfortner:
Thanks for clarifying. I was a little confused because the number of times the same information is sent seemed irrelevant to the point you were making.

Three potential concerns as I see it.

1. Performance — shouldn't be affected the http call made by a separate thread
2. Privacy — there is nothing personally identifiable, browser name, version, os, architecture, calling language. Plausible itself gathers IP addresses for locations, but it is only aggregated: https://plausible.io/privacy-focused-web-analytics
3. Security — If there is a proxy or other network restrictions, the call won't go anywhere. Also we're still working on getting these signed

We talked about a several different ways to implement this, some more complex and some more simple, and this was the best compromise we came up with. The binary itself has not been opt-in for a while now. The new behavior (automatic downloads) is currently opt-in, but this will change soon as well.

rjhancock:
It seems my concerns weren't clear enough. Regardless of Plausible's stance on privacy, that does not mean all of us wish to share such data for a variety of reasons and this change caused several alarms to flag on our end during testing due to unexpected requests being made that we did not authorize.

I don't deny that the data you request is valuable for the project as a whole, but such analytics should be Opt In vs Opt Out. Ask the community as a whole to opt-in to new analytics and tell them why, don't just turn it on with barely a mention in the readme.

This issue caused additional billable time to my client to ensure the data is not sent so this decision is having monetary damages to individuals and companies.

walksonair:
Thank you for providing visibility into this metric gathering. Due to some sensitive work/contracts, I can't allow any data to escape. Could you let me know if there's a way to turn this off for such projects that require it? At the moment, we get `Error sending stats to Plausible` on some networks. Again, it would help if we could opt out of this metric collection for some projects. Thank you again!

titusfortner:
I understand your concerns, and at the very least it could have been communicated better. Opt-in is typically for personally identifiable info or tracking data; we view this as essentially the same GDPR compliant data we collect by default from all of our website visitors. Everything collected is publicly available. If we had more resources and more assistance we might have found other ways of implementing or communicating things, but we're doing the best we can do with the resources we have.

titusfortner:
Yes, you can disable it by setting environment variable SE_AVOID_STATS to "true" or with the [config file](https://www.selenium.dev/documentation/selenium_manager/#configuration)

rjhancock:
We've found that using the environment variable does somehow work despite not being checked in code and are using that IN ADDITION to other methods.

I'm not dismissing the work you're doing, we do appreciate it, but even with it being GDPR compliant collection it is still something that should have been opt-in.

titusfortner:
> despite not being checked in code

It is checked by the binary (generated from the Rust code). 
We decided that it was only worth collecting the data if it was opt-out. 
That there would be concerns and frustrations was foreseeable and we could have communicated it better. 

@diemol / @bonigarcia do either of you have time to write a blog post and update our docs? I can add it to my list if not.

zfouts:
@titusfortner FYI on the release blog post, it was categorized as 2023 by mistake, may wish to move that to the 2024 folder so it's more obvious and visible. 

I agree with others that it should have been communicated more widely in advance, and also that the data collected is valuable. However, an opt-in approach is always best when collecting data. But primarily, communication on the intent should have been made before the release.

rjhancock:
> > despite not being checked in code
> 
> It is checked by the binary (generated from the Rust code).

Using GitHub search only shows it in [three places](https://github.com/search?q=org%3ASeleniumHQ+SE_AVOID_STATS&type=code) and looking through what some would consider to be obvious places does not show that environment variable being checked anywhere.

Even extending to all of GitHub still does not show where it is being checked.

Again, I agree the data is valuable in general and for the future of the project, and I do appreciate the hard work y'all put into this.

In a world where people don't like being tracked and reported on, especially developers, opt-in is ALWAYS the best approach EVEN when the data is compliant with all laws. The data being shared via a browser making a request is expected, that same data being harvested and sent to a third-party we didn't agree to, even if benign, is not.

titusfortner:
I'm not sure why it matters where in the code it is set. 
it parses environment variables with `SE_*_*` and cli values with `--*-*` and config file values, etc
https://github.com/SeleniumHQ/selenium/blob/selenium-4.17.0/rust/src/lib.rs#L1478-L1480
Everything we see is publicly available on plausible's site - https://plausible.io/manager.selenium.dev

I'm not disagreeing with you, your viewpoint is valid, but all software is compromises and tradeoffs based on resources and requirements and the best I can say is that this implementation was at the confluence of our consensus evaluation.

rjhancock:
The location matters to those of us who may now have to do additional auditing to ensure the dependencies we include in our projects don't introduce undesired effects such as this feature.

Might I suggest changing the default to be `true` until the next major version and re-communicate to the community why the change was made and why you want the data?

You still get your data, eventually, and can do a re-do on the communication front.

Every decision is a compromise and a set of trade offs regardless of it is software or IRL. Having this be opt-out for tracking was a bad decision.

titusfortner:
Alternately, if you specify the location of the driver you want in the service class, the binary won't be used at all.

rjhancock:
For those using the Ruby gem, configure a custom browser and pass an options package to it with this argument.

`options.add_argument('--avoid-stats=true')`

Still suggest putting in extra measure on your network to block the tracking as a backup however.

titusfortner:
> pass an options package to it with this argument

That... shouldn't work. Selenium Manager needs the args and those options arguments get passed to the driver when starting the browser.

rjhancock:
Oddly it does. Would love a more elegant solution to ensure this feature is disabled on deployments. Currently have a custom hosts file, environment variables set, and this to disable sending of data using the selenium-webdrivers gem.

titusfortner:
Hmm, really think is has to be something else and/or this behavior will get changed unexpectedly for you. Can you turn on logging (`Selenium::WebDriver.logger.level = :debug`) and link to a gist with the results?

rjhancock:
I saw it report it as unrecognized but it does not attempt to send the analytics.

https://gist.github.com/rjhancock/718994edf7756392336f07ae1a05d950

This is only 8 of 1800+ tests.

andrew-azarov:
> Regarding privacy, it is worth mentioning that we use https://plausible.io/, which is privacy first and GDPR compliant.

It's not compliant, because when you install the package consent is not triggered. Thus when someone installs without reading README but by using standard Python tools they don't see a warning nor consent on stealing sensitive data (and just having an app receive IP address of client without explicit consent is enough for breaking PI laws).

As long as you can install selenium without triggering EXPLICIT consent - it is not legal.

Moreover - there are use cases when subdomain (non public) can be a privately identifiable (and remember in GDPR Article 4 is identifiable) information.
For example you setup a catchall subdomain ala jack-coch-zx0vu80.clients.website.tld for ease of access to a private panel for customer and limit that to 1 ip address or subnet - this is by all definitions private space and private identifiable information as nobody but the script runner can see it and access it. And now this goes to plausible without ANY consent.

Besides that neither manual nor README includes a statement about telemetry.
EDIT: not even website has consent or Privacy policy. This must be reported asap to national PII protection services.
Any company which uses selenium - could have inadvertently compromised their own privacy policies and must report breach (https://commission.europa.eu/law/law-topic/data-protection/reform/rules-business-and-organisations/obligations/what-data-breach-and-what-do-we-have-do-case-data-breach_en)

I see lawsuits written all over this

diemol:
@andrew-azarov I am not a lawyer, but SFC approved this tool mainly because of its focus on data privacy.

> Built with privacy of your visitors in mind. No need to annoy them with a cookie/GDPR consent banner. 

Check their content about it. https://plausible.io/privacy-focused-web-analytics

andrew-azarov:
> @andrew-azarov I am not a lawyer, but SFC approved this tool mainly because of its focus on data privacy.
> 
> > Built with privacy of your visitors in mind. No need to annoy them with a cookie/GDPR consent banner.
> 
> Check their content about it. https://plausible.io/privacy-focused-web-analytics

The issue is that USING the PI collecting tool makes you liable for the data you collect. So there should be liability disclosure ie privacy policy, BEFORE you collect the data, doesn't matter how or for what, it can be a local app collecting mouse movement on your PC without sending it to a third party - it still must disclose that this data is collected and saved so the user is aware that this happens. Otherwise it is by all definition a malware.
Just a note about data collection deep in the help docs is not enough, especially when you can completely skip the docs when installing via packages.

This is the first breach of GDPR.
Second one is that they SEND the data to plausible, inadvertently exposing IP addresses, library and OS data (which can include hostname in some cases, even hidden hostnames like Kernel compilation hostname of internal VM in case of BSD for example).
As there is no detailed description of what data is sent I can only speculate, but per GDPR if the bulk of data is enough for personal identification - it is not compliant, it must be anonymized beforehand.
Plausible in this case receives IP (because selenium connects to it) and browser, OS, whatever else there. It's enough to identify a person even without explicit personal information available like in case of hostname I wrote about previously.

So what should have been done.
Selenium MUST have a privacy policy with clearly described and written PI it uses. Also this document must disclose all of the third parties processing that data. The consent must be explicitly taken upon installation of software, if no consent and selenium wants to keep telemetry on by default - then selenium MUST NOT be able to install.

There is a reason Google switched to consent 2.0 recently. That is because GDPR was written by lawyers and not technical people, and lawyers don't understand that even ping command exposes your IP address.

diemol:
We looked at https://plausible.io/data-policy, and that gave us an understanding of what is done with the data.

We have also placed this in the docs: https://www.selenium.dev/documentation/selenium_manager/#data-collection, but from what you are saying, is it not enough?

Again, I am not a lawyer, and we rely on the advice of others at SFC. However, would you be willing to help us improve based on your comments? We just want to make Selenium better by understanding global usage; we are not interested in individual data. I am just confused because the tone you are using is kind of threatening, so I wonder if you would like to help us improve.

Schnurres:
So you implemented and launched a data collection feature without ever consulting a lawyer or data protection expert, but claim it is GDPR compliant without ever having verified this?

That seems very negligent to me. Simply saying 'I am not a lawyer' is not a solution either. When you implement such critical features, you have to ensure compliance before making any claims.

diemol:
I did not say that; I said the tool was reviewed by [SFC](https://sfconservancy.org/), which owns Selenium and has a legal team. Could you tell me why you suggested that when I did not say that explicitly?

I am having trouble understanding the threatening tone. If there is something that needs improvement, we can work it out with your help. This is the type of behavior that burns down OSS maintainers; instead of collaborating, we get harsh comments.

rjhancock:
As I stated previously, the issue is that data started being collected and sent without informed consent from us developers and the only notice that was given was a line item on the release notes.

It was not clearly stated what was going on or what was going to happen. Having it be an opt-in situation vs opt-out would have solved this problem AS WELL AS a plea to the community for why you want the data and requesting they turn it on.

I recently had this discussion with another OS project about adding Sentry monitoring. Everyone was on board for including the package as it will help, but we decided to very strictly have it off by default. We've already started notifying users MONTHS before we implement it, why we want to collect the data, and requesting they turn it on.

In a world where data is king, and is generally taken without consent, be the bigger firm and specifically REQUEST it BEFORE taking it. I've wasted several hours ensuring all of my systems have extra protections in place because I've yet to find a suitable replacement.

titusfortner:
No PII is retained, and we make all information we collect public — https://plausible.io/manager.selenium.dev
If you can find lawyers that think they can win an argument that this constitutes a violation of personal data retention, they can talk to our lawyers.

I do understand your frustration; I also wish there were a better option, but there is no way to get meaningful information and have it be opt-in. So the options were either not to collect any information or to do so the way we have. I get that you would make another decision than we did, that's fine, but it doesn't mean we didn't sufficiently consider the options or that we would have chosen differently if only you explained it louder/harder at us.

This conversation is not being helpful to anyone at this point, so I'm locking it.

PR #13174, opened 2023-11-20
[dotnet] Use .GetType() instead of typeof() for determining the response types for DevTools commands

EdwinVanVliet:
### Description
The CommandResponseTypeMap.cs TryGetCommandResponseType function uses typeof(T) syntax for retrieving the correct response type of a DevTools command. This only works when the inputted command is known compile time. When using the predefined *-Adapter.cs classes this won't be an issue, as compile it is known which types are being inputted. When using DevToolsSession.cs SendCommand function directly while only knowing the type of the command runtime. The typeof(T) function will yield ICommand as a result, eventually causing an exception to be thrown because no response type could be found for ICommand. 

### Motivation and Context
We run Selenium while creating some DevTools commands dynamically which now causes exceptions. 

### Types of changes

### Checklist

- [ ] All new and existing tests passed. => having issues to run the tests...


diemol:
@nvborisenko what do you think?

nvborisenko:
All commands are well known, and types can be (I would say "should be") resolved at compile time. To me it is hard to imagine why we should move it to runtime.

> The typeof(T) function will yield ICommand as a result

@EdwinVanVliet how do you use it?

EdwinVanVliet:
@nvborisenko 

We used to run all the commands we issue to DevTools through a BufferBlock (https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.dataflow.bufferblock-1?view=net-8.0) in order to prevent threading issues when sending out lot's of commands from multiple threads. When using the Network domain RequestPaused and AuthRequired events to apply all sorts of filtering on HTTP requests in the browser and responding to authentication popups, we ran into various issues with threading because the underlying Selenium part was not handling this well. This is a historical reason however, because your changes on async-await part resolved this as far as I'm aware. The Bufferblock has since been removed, invoking the commands is still working using the DevToolsSession.SendCommand function.

We have an abstract class which contains lot's of functionality for interacting with DevTools; handling screencast, browser authentication, tracing network traffic, tracing browser metrics, handling browser tab switching and so on. All actual version specific stuff is delegated to version specific implementations of that abstract class. That mainly supplies the version specific commands and let the abstract class do the logic based on that. Because this abstract-implementation distinction the compiler doesn't know which type of event is being used. We have ways to work around this of course; but this requires quite some rework while this fix is minor and have been running it for a couple of years in our own Selenium repo. 

Second reason is that DevToolsSession.cs SendCommand function is public and accepts ICommand as an argument. The interface definition allows others to use it this way, but unexpectedly breaks in some usage patterns. Either this function should be made internal or it shouldn't break unexpectedly. 

Edit: making this function internal is not a feasible solution from my side. The adapter classes from Selenium don't support sessionless commands which are actually needed in some cases, for instance when wanting to use the Trace domain functions for all tabs. Trace domain (https://chromedevtools.github.io/devtools-protocol/tot/Tracing/#method-start) start only works for the session you run the command against, when leaving the sessionId empty it works for all sessions at the same time. This was the only way to get it to trace for multiple tabs at the same time.

nvborisenko:
> Second reason is that DevToolsSession.cs SendCommand function is public and accepts ICommand as an argument. The interface definition allows others to use it this way, but unexpectedly breaks in some usage patterns. Either this function should be made internal or it shouldn't break unexpectedly.

We don't want to break somebody else.

I am proposing the following:
1. If you are OK with your workaround, please do it on your side. Honestly we don't want to invest so much into DevTools area, given that BiDi implementation is coming. I hope this workaround is not tricky. Most likely it would be around how you call the function, probably using [MethodInfo.MakeGenericMethod](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.methodinfo.makegenericmethod) to invoke this method might be useful in your case.

2. Given that it is risky to change the implementation, what if we will add overloaded method?
```
void SomeMethod<T>();
void SomeMethod(Type t);
```
Will it be working solution in your case?

I hope the first option will be applied.

EdwinVanVliet:
@nvborisenko 

> 2\. Given that it is risky to change the implementation, what if we will add overloaded method?

That would require an overload of the SendCommand function, and also adding a duplication of TryGetCommandResponseType function. That would work for us, but also introduces unneeded complexity.

nvborisenko:
@EdwinVanVliet what about the first option?

EdwinVanVliet:
@nvborisenko 

Using MethodInfo.MakeGenericMethod would eventually result into the same problem because the result of MakeGenericMethod is only known runtime. 

In the project I work on we have our own private Selenium repo, including our own build pipeline. On that repo we have accumulated various fixes and changes over the years, including this one. From our perspective we want our private repo to get closer to the public Selenium repo in order to prevent merge conflicts and so on. We can leave this fix intact on our private Selenium repo, therefore having no need to do any other work arounds.

nvborisenko:
True! Then let's be more specific here. Please create very simple code to highlight the issue. I still believe selenium should resolve types at compilation phase, everything else (which is known only at runtime), should be resolved at runtime. We really don't want to introduce unnecessary code in this codebase. As you said, it would introduce complexity (from your point of view), and increase efforts for supporting (from our point of view).

nvborisenko:
Somebody may say "what is a problem, just merge it". But it would be a precedent... why we cannot use generic methods. We want to use generic methods, why not. We should be crystally clear here "why not".

EdwinVanVliet:
@nvborisenko 

I've created a code example demonstrating the issue. 


```
// See https://aka.ms/new-console-template for more information
using OpenQA.Selenium.Chrome;
using OpenQA.Selenium.DevTools;

Console.WriteLine("Hello, World!");

var input = Console.ReadLine();

var service = ChromeDriverService.CreateDefaultService("C:\\Git\\selenium-webdriver\\dotnet\\WebDriver.Console\\bin\\Debug\\net6.0\\chromedriver.exe");
var driver = new ChromeDriver(service);
var devtools = driver.GetDevToolsSession();

ICommand command;
if (input == "a")
{
    command = new OpenQA.Selenium.DevTools.V119.Target.GetTargetsCommandSettings();
}
else
{
    command = new OpenQA.Selenium.DevTools.V120.Target.GetTargetsCommandSettings();
}

try
{
    // this breaks because actual implementation of ICommand is only known runtime
    var response = devtools.SendCommand(command).GetAwaiter().GetResult();
}
catch (Exception ex)
{
    Console.WriteLine(ex);
}


driver.CloseDevToolsSession();
driver.Close();

```

Running this will output:

```
System.InvalidOperationException: Type OpenQA.Selenium.DevTools.ICommand does not correspond to a known command response type.
   at OpenQA.Selenium.DevTools.DevToolsSession.SendCommand[TCommand](TCommand command, CancellationToken cancellationToken, Nullable`1 millisecondsTimeout, Boolean throwExceptionIfResponseNotReceived) in C:\Git\selenium-webdriver\dotnet\src\webdriver\DevTools\DevToolsSession.cs:line 161
   at Program.<Main>$(String[] args) in C:\Git\selenium-webdriver\dotnet\WebDriver.Console\Program.cs:line 26
```

nvborisenko:
@EdwinVanVliet It works:

```csharp
var dt = typeof(DevToolsSession);
var mi = dt.GetMethods().First(x => x.Name == nameof(DevToolsSession.SendCommand));
var mi2 = mi.MakeGenericMethod(command.GetType());

var  res = mi2.Invoke(devtools, new object[] { command, default(CancellationToken), null, true });
```

The idea is pretty easy: if types are known in runtime, then use methods invocation in runtime. This is definitely can be considered as a workaround, not a solution.

Given that how painful it is, and we don't want to "redesign" this stuff, I don't mind to introduce overloaded methods:
```csharp
Method<T>(T command);
Method(Type commandType);
```

Up to you, please suggest which one is better in your case.

EdwinVanVliet:
@nvborisenko 

I understand your point of view not wanting to introduce breaking changes. The proposed changes are workarounds however, both requiring additional changes on our side in downstream code. As we have been running the proposed changes in the PR for over a year in our production systems we prefer to keep using those, as the solution is cleaner and doesn't require us changing downstream code. For those reasons I will abandon the PR and keep using the proposed changes at our private Selenium repo.

nvborisenko:
I thought our negotiation would be method overloading.

PR #13175, opened 2023-11-20
[dotnet] Added TargetAttached datastructure to DevTools Target.

EdwinVanVliet:
### Description

This PR adds the TargetAttached event which allows consumers to be informed about new targets being opened (new tab for instance) by Chromium. 

Recently we had a discussion about attaching the browser to a new tab still being an issue; https://github.com/SeleniumHQ/selenium/pull/10603

This PR is the first in a series to resolve the issue from the previous PR.

### Motivation and Context
The end goal is that we can pause new targets to allow the consumer to enable the required DevTools domains on the target. For instance to be able to trace all the network traffic on that target, the Network domain should be enabled directly after the target is created.

With this PR only the TargetAttached event is being added. In followup PRs I will add;
- Setting up DevToolsSession.cs to enable or disable automatically pausing new targets until Runtime.runIfWaitingForDebugger is invoked.
- Add a function to DevToolsSession.cs which allows the consumer to invoke multiple requests in one batch, to prevent locking issues when issueing the Runtime.runIfWaitingForDebugger command. 

### Types of changes

### Checklist

- [ ] All new and existing tests passed. => have an issue running tests.


dotnet/src/webdriver/DevTools/TargetAttachedEventArgs.cs
```diff
@@ -0,0 +1,41 @@
+// <copyright file="DevToolsEventData.cs" company="WebDriver Committers">
```
nvborisenko: Incorrect file name.

dotnet/src/webdriver/DevTools/TargetAttachedEventArgs.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    /// <summary>
+    /// Event arguments present when the TargetAttached event is raised.
+    /// </summary>
+    public class TargetAttachedEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Gets the ID of the session of the target attached.
+        /// </summary>
+        public string SessionId { get; set; }
```
nvborisenko: Please use `internal` modifier for setters.

nvborisenko (changes requested):
Thank you for contribution. Just fix minor issues.

dotnet/src/webdriver/DevTools/TargetAttachedEventArgs.cs
EdwinVanVliet: @nvborisenko Fixed!

dotnet/src/webdriver/DevTools/TargetAttachedEventArgs.cs
EdwinVanVliet: @nvborisenko Fixed!

nvborisenko (approved):

diemol:
If it looks good for you, @nvborisenko, feel free to trigger CI and then merge.

PR #13190, opened 2023-11-23
[java] POC: Porting WebDriver classic command to BiDi

pujagani:

### Description

This POC demonstrates the changes in porting the WebDriver Classic command to use BiDi under the hood. 
The changes port GET and PRINT command. 

### Motivation and Context

The next big step for Selenium towards Selenium 5 is porting WebDriver Classic commands to use WebDriver BiDi protocol. This is the first step towards it. The changes are not trivial, a lot of pieces need to be considered. Instead of moving the mountain, trying to make progress towards it in decent-sized chunks of work.

Reason for choosing GET command: demonstrate the need for requiring driver instance since capabilities like pageLoadStrategy are required

Reason for choosing the PRINT command: Required additional step to serialize parameters into PrintOptions. That piece of code can be moved to PrintOptions but the build kept breaking due to cycling dependency so didn't dig deep into it.

Includes the items that need to be considered/included:
TODO: 
- Handling commands if the browser supports BiDi but not the command 
- Error handling - correct mapping to error codes and throwing respective exceptions to avoid any breaking changes to the user
- The current approach should work for most commands but atoms will need special handling (I have an idea of what is needed and how to do this)
- Actions, JavascriptCommandExecutor etc - they all pass the commands through the RemoteWebDriver, so the current approach will work seamlessly as long we add the commands to BiDiCommandExecutor
- Add a mapper to BiDi LocalValue to map argument based on type to LocalValue instance (I will be implementing this next)

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/remote/BiDiCommandExecutor.java
+++ b/java/src/org/openqa/selenium/remote/BiDiCommandExecutor.java
@@ -17,179 +17,20 @@
 
 package org.openqa.selenium.remote;
 
-import static org.openqa.selenium.remote.DriverCommand.GET;
-import static org.openqa.selenium.remote.DriverCommand.PRINT_PAGE;
-
 import java.io.IOException;
-import java.io.StringReader;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicReference;
-import java.util.function.BiFunction;
-import org.openqa.selenium.UnsupportedCommandException;
-import org.openqa.selenium.bidi.Network;
-import org.openqa.selenium.bidi.Script;
-import org.openqa.selenium.bidi.browsingcontext.BrowsingContext;
-import org.openqa.selenium.bidi.browsingcontext.ReadinessState;
-import org.openqa.selenium.json.Json;
-import org.openqa.selenium.json.JsonInput;
-import org.openqa.selenium.json.TypeToken;
-import org.openqa.selenium.print.PageMargin;
-import org.openqa.selenium.print.PageSize;
-import org.openqa.selenium.print.PrintOptions;
 
 public class BiDiCommandExecutor implements CommandExecutor {
 
-  private static Json JSON = new Json();
-
   private final RemoteWebDriver driver;
-
-  private final AtomicReference<BrowsingContext> currentContext = new AtomicReference<>();
-  private final AtomicReference<Script> script = new AtomicReference<>();
-  private final AtomicReference<Network> network = new AtomicReference<>();
-
-  private final Map<String, BiFunction<Command, RemoteWebDriver, Response>> commandHandlerMap =
-      new HashMap<>();
-
-  // Each browsing context has an associated id, maintains state
-  // We will need to maintain a map of all the browsingContext to run further commands
-  // Switching between tabs etc might be tricky
-  private final Map<String, BrowsingContext> browsingContextMap = new ConcurrentHashMap<>();
+  private final BiDiCommandMapper commandMapper;
 
   public BiDiCommandExecutor(RemoteWebDriver driver) {
     this.driver = driver;
-    init(this.driver);
-  }
-
-  private void init(RemoteWebDriver driver) {
-    script.set(new Script(driver));
-    network.set(new Network(driver));
-
-    BrowsingContext parentContext = new BrowsingContext(driver, driver.getWindowHandle());
-    browsingContextMap.put(parentContext.getId(), parentContext);
-    currentContext.set(parentContext);
-
-    commandHandlerMap.put(
-        GET,
-        (command, webDriver) -> {
-          String pageLoadStrategy =
-              (String) webDriver.getCapabilities().getCapability("pageLoadStrategy");
-          currentContext
-              .get()
-              .navigate(
-                  (String) command.getParameters().get("url"),
-                  ReadinessState.getReadinessState(pageLoadStrategy));
-
-          // Required because W3C Classic sets the current browsing context to current top-level
-          // context on navigation
-          // This is crucial for tests that:
-          // Switch to frame -> find element -> navigate url -> find element (in BiDi it will try to
-          // find an element in the frame)
-          // But in WebDriver Classic it will try to find the element on the page navigated to
-          // So to avoid breaking changes, we need to add this step
-          // Refer: Pt 9 of https://www.w3.org/TR/webdriver2/#navigate-to
-          // Refer: https://w3c.github.io/webdriver-bidi/#command-browsingContext-navigate
-          driver.switchTo().window(currentContext.get().getId());
-          return new Response(webDriver.getSessionId());
-        });
-
-    commandHandlerMap.put(
-        PRINT_PAGE,
-        (command, webDriver) -> {
-          String result = currentContext.get().print(serializePrintOptions(command));
-          Response response = new Response(webDriver.getSessionId());
-          response.setValue(result);
-          return response;
-        });
+    commandMapper = new BiDiCommandMapper(driver);
   }
 
   @Override
   public Response execute(Command command) throws IOException {
-    // This is not optimal
-    // But each time Classic calls switchTo, we need to identify the current browsing context to run
-    // commands
-    // Maybe we need a way where every time switchTo commands are called, only then we update the
-    // current browsing context
-
-    String currentWindowHandle = driver.getWindowHandle();
-    BrowsingContext browsingContext =
-        browsingContextMap.computeIfAbsent(
-            currentWindowHandle, windowHandle -> new BrowsingContext(driver, windowHandle));
-
-    browsingContextMap.putIfAbsent(currentWindowHandle, browsingContext);
-    currentContext.set(browsingContext);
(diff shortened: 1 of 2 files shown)


java/src/org/openqa/selenium/bidi/browsingcontext/ReadinessState.java
```diff
@@ -17,19 +17,38 @@
...
 package org.openqa.selenium.bidi.browsingcontext;
 
+
 public enum ReadinessState {
-  NONE("none"),
-  INTERACTIVE("interactive"),
-  COMPLETE("complete");
+  NONE("none", "none"),
+  INTERACTIVE("interactive", "eager"),
+  COMPLETE("complete", "normal");
```
diemol, line 26: Perhaps add a comment indicating the reasoning for this mapping.

java/src/org/openqa/selenium/remote/BiDiCommandExecutor.java
```diff
@@ -0,0 +1,167 @@
...
+        (command, webDriver) -> {
+          String pageLoadStrategy =
+              (String) webDriver.getCapabilities().getCapability("pageLoadStrategy");
+          currentContext
+              .get()
+              .navigate(
+                  (String) command.getParameters().get("url"),
+                  ReadinessState.getReadinessState(pageLoadStrategy));
+          return new Response(webDriver.getSessionId());
+        });
```
diemol: Should we define all these parameters somewhere else? I assume we don't want to declare them all here, as this is a PoC.

java/src/org/openqa/selenium/remote/Delegator.java
```diff
@@ -0,0 +1,77 @@
...
+import java.io.IOException;
+import java.net.URL;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import org.openqa.selenium.internal.Require;
+import org.openqa.selenium.remote.http.ClientConfig;
+import org.openqa.selenium.remote.http.HttpClient;
+
+public class Delegator extends HttpCommandExecutor implements CommandExecutor {
```
diemol: Maybe `BiDiDelegator`, not sure about the name. Also, a quick comment noting the purpose of the class would be nice.

java/test/org/openqa/selenium/bidi/PortClassicToBiDiTest.java
```diff
@@ -0,0 +1,62 @@
...
+    server = new NettyAppServer();
+    server.start();
+  }
+
+  @Test
+  @NotYetImplemented(SAFARI)
+  @NotYetImplemented(IE)
+  @NotYetImplemented(EDGE)
+  void canNavigate() {
+    driver.get("http://google.com");
```
diemol:
```suggestion
    driver.get("https://selenium.dev");
```

java/test/org/openqa/selenium/bidi/PortClassicToBiDiTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+    assertThat(driver.getTitle()).isEqualTo("Google");
```
diemol:
```suggestion
    assertThat(driver.getTitle()).isEqualTo("Selenium");
```

diemol (changes requested):
Thank you, @pujagani!

I left a few simple comments. Looks good to me. 

Would be nice if @shs96c shares his opinions.

java/src/org/openqa/selenium/remote/BiDiCommandExecutor.java
pujagani: I did think of using a codec but that did not fit properly here. I do want to populate the command map in some other place. I will look into it. You are right, executor is not the proper place for it.

java/src/org/openqa/selenium/remote/Delegator.java
pujagani: I think since even the DriverCommandExecutor uses it, I was not sure what name to give. I think BiDiDelegator might make sense.

java/src/org/openqa/selenium/remote/BiDiCommandExecutor.java
diemol: That is fine. As a first step I do not mind having it here, but it is good we want to have it somewhere else as we move forward.

diemol (approved):

titusfortner:
This is great. I love seeing all that extra code removed from the driver classes.

What is the separation of concerns between `BiDiDelegator` and `BiDiCommandExecutor`? Can't `BiDiCommandExecutor` just extend `HttpCommandExecutor` directly? It looks like it is essentially only storing the list of commands?

Oh, this is when the user passes an existing `HttpCommandExecutor` in to the `RemoteWebDriver` constructor... Would it make sense not to use BiDi at all for that, though? Do we want to override a user's custom command executor even if we could? At least in the short term? Then toggle which Executor to use in `createExecutor()`? Or does `TracedCommandExecutor` make that approach more complicated than it is worth?

java/src/org/openqa/selenium/remote/BiDiCommandExecutor.java
pujagani: +1 I have started working on it too.

pujagani:
BiDiDelegator knows where to delegate the commands and that's it. BiDiExecutor will have reference to methods/class that handles converting the Classic command to BiDiCommand and also maintain browsing context to identify where to run the command since almost all BiDi commands need the browsing context id to be sent to run commands. For atoms, it will have additional methods to convert arguments to how BiDi expects it to run scripts. 

I think we don't want to override the user's custom executor. I have considered that case. So we do a check that if the user passes an executor that is an instance of BiDiDelegator then we set the BiDiCommandExecutor for it. Otherwise, the executor passed in will be used as it is.
So essentially, the default behaviour for Selenium would be to use BiDi unless the user wants to pass their own CommandExecutor then we run with that. I think we need to allow users to be able to do that since there could be a possibility that web sockets are not supported by their company proxy etc.

pujagani:
Details or gaps found due to failing tests:
1. AlertsTest - fails because the page has alert when the page loads. BiDi doesn't finish loading the page completely due to the alert and hence times out when page load strategy is normal. However, WebDriver Classic, doesn't get blocked due to the alert and moves on. 
2. Page loading tests - Any tests that make use of page load timeouts fail because BiDi does not take into consideration the page load timeout values set. I think action item here is creating a github issue for the same in the w3c BiDi repo.

java/src/org/openqa/selenium/remote/BiDiCommandExecutor.java
pujagani: I added a BiDiCommandMapper. I think that might help.

joerg1985:
@pujagani is there a long therm plan to support classic w3c commands or is the plan to shift to BiDi at a certain time?

I am currently playing with BiDi and noticed it is relying on a stable tcp-ip connection, what seems to be not the case in the company i am working at. Looks like our datacenters sometimes have short connection outtakes or at least some connections are dropped (might be the VPN, who knows).

We did not notice them till now, as HTTP clients do automatically reconnect when a connection is stale.
Is there a way to reconnect the Websocket connection when it breaks? This is propably no easy task especially when the driver is sending a event to a broken connection.

pujagani:
Thank you @joerg1985 for sharing this. The plan is to move to BiDi under the hood Selenium 5 onwards (whenever it happens). 
I don't know if there is a way to reconnect when the WebSocket connection breaks. We might need to track this as a separate issue to ensure this move to BiDi is resilient to such network changes.

java/src/org/openqa/selenium/remote/BiDiDelegator.java
```diff
@@ -0,0 +1,82 @@
...
+
+// Delegates the command to the respective command executor.
+// If the commands is supported in BiDi, it is delegated to BiDiCommandExecutor
+// Otherwise, it falls back on HttpCommandExecutor
+// The reason both are required is that, HttpCommandExecutor route will always be used to start and
+// stop the session.
+public class BiDiDelegator extends HttpCommandExecutor implements CommandExecutor {
+
+  private BiDiCommandExecutor biDiCommandExecutor = null;
+  private static final Set<String> biDiCommands = new HashSet<>();
```
shs96c, line 39: You could use `Set.of` here now we're using Java 11+

java/src/org/openqa/selenium/remote/BiDiDelegator.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  public void setBiDiCommandExecutor(BiDiCommandExecutor biDiCommandExecutor) {
+    this.biDiCommandExecutor = biDiCommandExecutor;
+  }
+
+  @Override
+  public Response execute(Command command) throws IOException {
+    if (biDiCommandExecutor != null && biDiCommands.contains(command.getName())) {
+      return biDiCommandExecutor.execute(command);
```
shs96c, line 75: If you sniff the response to see if it's an `unknown command` here, you could build the list of supported commands dynamically. This would allow the Bidi implementation to not be fully complete, and we'd fall back to Classic commands where possible.

diemol:
I bumped into this and I think we can close this since the underlaying code has moved too much. When #17232 is done we can probably start this again. 

@pujagani please reopen if I missed something.

PR #12844, opened 2023-09-30
Add RBS files to Ruby

aguspe:
### Description
I created RBS files for the following classes and modules so far:

- AbstractEventListener
- ActionBuilder
- Atoms
- Bridge
- Chrome
- Chromium::Features
- Driver
- Element
- Edge
- Edge::Features
- Firefox
- Firefox::Features
- HasCasting
- HasLaunching
- InputDevice
- Interaction
- Logger
- KeyActions
- KeyInput
- Manager
- Navigation
- PointerActions
- PointerInput
- Remote
- Response
- Safari
- Safari::Features
- ScrollOrigin
- SearchContext
- ShadowRoot
- TakesScreenshot
- TargetLocator
- Timeouts
- TypingInteracton
- Version
- VirtualAuthenticator
- Webdriver
- WheelInput

Also, I added support for steep: https://github.com/soutaro/steep

You can run the following to do the type check: 
`steep check`

This PR is the first one of multiple PRs adding support for RBS and type check to the ruby library.

### Motivation and Context
Ruby Selenium does a lot of meta-programming which makes it more difficult to use with an IDE like RubyMine. 
Adding rbs files, *should make this much easier.

More information and further discussion can be found here: https://github.com/SeleniumHQ/selenium/issues/10943

### Types of changes
- [x] New feature (non-breaking change which adds functionality)

### Checklist
- [x] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #10943, opened 2022-08-12
[🚀 Feature]: Add rbs files to Ruby


titusfortner:
### Feature and motivation

We're adding Type hints to Python & looking at TypeScript for JS, the equivalent for Ruby is to add rbs files.

Ruby Selenium does a lot of meta-programming which makes it more difficult to use with an IDE like RubyMine. Adding rbs files, *should make this much easier. 

### Usage example

I think they'll look like this?
```ruby
module Selenium
  module WebDriver
    module Edge
      class Options < Selenium::WebDriver::Chrome::Options
        KEY:String
        BROWSER:String

        def enable_logging: (browser_options: Hash) -> Hash

        def binary_path: () -> String
      end
    end
  end
end
```

and then RubyMine and other IDEs will be able to do type hints, etc

aguspe:
Hi I would like to help out with this, I forked the repo and created a new PR to start adding RBS files

titusfortner:
That would be great, we'd appreciate the help with it.

Hah, Sparkling Watir, I like it! You didn't have to fork it, though, you could have just taken over Tap Watir. 😁 
I wish I'd had the time to work on it back when people were asking for it.

aguspe:
Haha, it's good to know, I assumed people were really busy, so I thought the fork was the easiest way to have something working fast.
 
But if it's okay, I could take over, I have a couple of PRs open in Tap Watir, which is just the extra stuff I added for Sparkling Watir, so I can update and merge them.

And use Sparkling Watir for Desktop automation with Appium, which is something I have been playing with on another gem.

titusfortner:
Oh, I saw there were PRs at one point, and, well, life...
I added you as maintainer to the tap watir repo
There's a Watir channel on slack if you want to discuss things there - https://www.selenium.dev/support/#ChatRoom

titusfortner:
We're making some great progress on this. Thanks to @aguspe / @Pexpe 
Do we need to track what we have and what we still need?

aguspe:
> We're making some great progress on this. Thanks to @aguspe / @Pexpe Do we need to track what we have and still need?

Hi @titusfortner thank you so much, right now I'm working on this draft: https://github.com/SeleniumHQ/selenium/pull/13234, I added all the generated RBS files for all classes and modules, and then I'm fixing all the steep issues that originated, I added two screenshots there but hopefully after the holidays I will be able to wrap this PR up and put it for review

titusfortner:
[...]

PR #13192, opened 2023-11-23
[rb] Extend RBS support for logger and log entry

aguspe:
### Description
This PR expands RBS support for the following classes:

* Selenium::Webdriver::LogEntry
* Selenium::Webdriver::Logger

It also adds support for the standard library 'Forwardable' on the Steepfile

After the update on the RBS files for the LogEntry class this is the result:

Screenshot 2023-11-23 at 15 11 32

The logger class keeps producing an error due to the proc signature coming from logger.formatted = not being identified, even though is part of the standard library:

Screenshot 2023-11-23 at 15 43 07

### Motivation and Context
This change expands the work done on: https://github.com/SeleniumHQ/selenium/pull/12844

The main motivation behind it comes from: https://github.com/SeleniumHQ/selenium/issues/10943

### Types of changes

### Checklist

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -81,7 +81,7 @@ jobs:
           java-version: ${{ inputs.java-version }}
           distribution: 'temurin'
       - name: Setup Bazel
-        uses: p0deje/setup-bazel@0.3.0
+        uses: p0deje/setup-bazel@0.3.2
         with:
           bazelisk-cache: true
           disk-cache: ${{ inputs.cache-key }}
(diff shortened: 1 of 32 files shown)


titusfortner:
The Logger class does not actually use the LogEntry class, so these are actually unrelated. That said, if this fixes an error, we might as well get it merged, thanks!

aguspe:
> The Logger class does not actually use the LogEntry class, so these are actually unrelated. That said, if this fixes an error, we might as well get it merged, thanks!

So the way I'm expanding RBS support is just by looking at the result of the Steep check and fixed the issues, 
I did these two classes together just because they appeared in alphabetical order

My idea is to continue updating the type support in small PRs, containing two or three classes, so eventually when 'steep check' is run, there are no type issues

PR #13194, opened 2023-11-23
[rust] Update ring and required crates for riscv support

jsmolic:
Cargo.lock: Update ring and required crates for riscv support

### Description
Selenium-webdriver fails to compile on RISC-V due to outdated version of ring used (https://bugs.gentoo.org/918098).
Ring has added RISC-V support in
https://github.com/briansmith/ring/pull/1627, so update ring and additional necessary crates to allow us to build on RISC-V. This change was compile tested inside a virtual machine.

### Types of changes

### Checklist

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1380,17 +1380,16 @@ dependencies = [
 
 [[package]]
 name = "ring"
-version = "0.16.20"
+version = "0.17.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3053cf52e236a3ed746dfc745aa9cacf1b791d846bdaf412f60a8d7d6e17c8fc"
+checksum = "9babe80d5c16becf6594aa32ad2be8fe08498e7ae60b77de8df700e67f191d7e"
 dependencies = [
  "cc",
+ "getrandom",
  "libc",
- "once_cell",
  "spin",
  "untrusted",
- "web-sys",
- "winapi",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


titusfortner:
The lockfile should be auto generated, don't we need to change something in another file to ensure this is used?

bonigarcia (changes requested):
Indeed, the `Cargo.lock` is automatically updated by Cargo when there is some change in `Cargo.toml`. So, I don't see the point of updating the lock file individually. In summary, I think this PR is incomplete or incorrect.

jsmolic:
I understand now, since ring is indirect dependency here it needs to be fixed in the source of package which is actually depending on ring. Apologies for jumping to conclusion that it should be changed in selenium, I will close this PR.
Thank you for your input!

titusfortner:
I suspect we can require a different version of a transitive dependency, depending on how it is set up? It just needs to be in a different file.

jsmolic:
The dependency chain goes like this: selenium -> reqwest -> hyper-rustls -> rustls -> ring and the culprit seems to be hyper-rustls; it doesn't use the latest version of rustls which pulls in the correct version of ring that compiles on risc-v.
Is there any way we can pin the rustls version?

titusfortner:
@AutomatedTester or @bonigarcia can you comment on how transitive rust dependencies work?

bonigarcia:
There is plenty of online information about that:

https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html
https://stackoverflow.com/questions/27770031/how-do-i-pin-indirect-dependencies-of-a-crate

PR #13235, opened 2023-12-02
[rb] Update rules_ruby to the latest version

p0deje:
Notably, it now allows to read version from `.ruby-version` file and use a more convenient `run_binary()` rule instead of a `genrule()`.

The commit also bumps supported Ruby versions on CI.


rb/.ruby-version
```diff
@@ -0,0 +1 @@
+3.0.6
```
p0deje, line 1: @titusfortner We can also have it in the root folder, does it matter for RubyMine?

titusfortner: No, Rubymine looks there.

PR #13237, opened 2023-12-04
[dotnet] Removed {{separator}} token from C# filename codegen

EdwinVanVliet:
### Description

On our own build environment we noticed that the intermediate files being generated when running the codegen contain the {{separator}} token. When looking into the Utility.cs ReplaceTokensInPath function there is no replacement being done on the {{separator}} token. This PR removes the {{separator}} token from the template within CodeGenerationDefinitionTemplateSettings.cs causing filenames to be outputted without {{separator}}. 

Also this issue caused some files to end up in the wrong directory. The AXPropertyName.cs should be located in the Accesibility domain, but it ended up in WebAuthn. 

Before fix.
SeparatorIssue

After fix.
SeparatorIssueFixed

### Types of changes

### Checklist

### DIFF
--- a/third_party/dotnet/devtools/src/generator/CodeGen/CodeGenerationDefinitionTemplateSettings.cs
+++ b/third_party/dotnet/devtools/src/generator/CodeGen/CodeGenerationDefinitionTemplateSettings.cs
@@ -42,7 +42,7 @@
             TypeEnumTemplate = new CodeGenerationTemplateSettings
             {
                 TemplatePath = "type-enum.hbs",
-                OutputPath = "{{domainName}}{{separator}}{{className}}.cs",
+                OutputPath = "{{domainName}}\\{{className}}.cs",
             };
         }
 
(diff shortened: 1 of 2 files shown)


nvborisenko:
Somebody introduced `{{separator}}` token. I don't know why, but let's not break this logic.

I propose very simple fix:
Utility.cs
```csharp
path = path.Replace("{{separator}}", System.IO.Path.DirectorySeparatorChar.ToString());
```

If it works, then we also can remove the following line:
```csharp
path = path.Replace('\\', System.IO.Path.DirectorySeparatorChar);
```

Seems somebody also tried to fix the same thing around path separator token.

nvborisenko (changes requested):
We still need to support `{{separator}}` token. Let's replace it in runtime.

nvborisenko:
Seems it works, please doble-check carefully:

```diff
- path = path.Replace('\\', System.IO.Path.DirectorySeparatorChar);
+ path = path.Replace("{{separator}}", System.IO.Path.DirectorySeparatorChar.ToString());
+ //path = path.Replace('\\', System.IO.Path.DirectorySeparatorChar);
```

EdwinVanVliet:
@nvborisenko 

That also works, confirmed with our build environment. Committed the changes.

We both need the `{{separator}}` and `\\` replacements because CodeGenerationDefinitionTemplateSettings.cs contains both forms.

nvborisenko (approved):

nvborisenko:
Thanks @EdwinVanVliet for your contribution!

PR #13249, opened 2023-12-06
[dotnet] Possibility to output internal log messages to file

nvborisenko:
Continue improvement of internal logging.

### Description
- Added `FileLogHandler` class which listens to internal log events
- Fixed issue when copying of log handlers across contexts

### Motivation and Context
To be aligned with others bindings, and providing file log handler out of the box.

### Types of changes

### Checklist


nvborisenko:
This change introduces a breaking change to `ILogHandler`. I even don't mind to land this feature as `v4.16.2`. Then docs will be aligned across bindings.

Why now rather than later?
Stabilize public API and avoid breaking changes in future.

nvborisenko:
Thanks @nvborisenko for your contribution!

dotnet/test/common/Internal/Logging/FileLogHandlerTest.cs
```diff
@@ -0,0 +1,43 @@
+using NUnit.Framework;
+using System;
+using System.IO;
+
+namespace OpenQA.Selenium.Internal.Logging
+{
+    public class FileLogHandlerTest
+    {
+        [Test]
```
YevgeniyShunevych, line 9: `[Test]` is redundant when there are `[TestCase]`'s.

dotnet/test/common/Internal/Logging/FileLogHandlerTest.cs
```diff
@@ -0,0 +1,43 @@
...
+                {
+                    fileLogHandler.Handle(new LogEvent(typeof(FileLogHandlerTest), DateTimeOffset.Now, LogEventLevel.Info, "test message"));
+                }
+
+                Assert.That(File.ReadAllText(tempFile), Does.Contain("test message"));
+            }
+            catch (Exception)
+            {
+                throw;
+            }
```
YevgeniyShunevych, line 36: `catch` block can be removed.

dotnet/src/webdriver/Internal/Logging/LogContext.cs
```diff
@@ -60,19 +60,16 @@ public ILogContext CreateContext(LogEventLevel minimumLevel)
                 loggers = new ConcurrentDictionary<Type, ILogger>(_loggers.Select(l => new KeyValuePair<Type, ILogger>(l.Key, new Logger(l.Value.Issuer, minimumLevel))));
             }
 
-            IList<ILogHandler> handlers = null;
+            var context = new LogContext(minimumLevel, this, loggers, null);
```
YevgeniyShunevych, line 63: I think that last `ILogHandlerList handlers` parameter can be removed from `LogContext` constructor. Currently in both usage places `null` is passed.

dotnet/src/webdriver/Internal/Logging/FileLogHandler.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        /// <summary>
+        /// Handles a log event by writing it to the log file.
+        /// </summary>
+        /// <param name="logEvent">The log event to handle.</param>
+        public void Handle(LogEvent logEvent)
+        {
+            lock (_lockObj)
+            {
+                _streamWriter.WriteLine($"{logEvent.Timestamp:r} {_levels[(int)logEvent.Level]} {logEvent.IssuedBy.Name}: {logEvent.Message}");
```
YevgeniyShunevych, line 50: Instead of `r` timestamp format maybe better use `yyyy-MM-dd HH:mm:ss.fff`?

dotnet/src/webdriver/Internal/Logging/FileLogHandler.cs
```diff
@@ -0,0 +1,94 @@
...
+        private StreamWriter _streamWriter;
+
+        private readonly object _lockObj = new object();
+        private bool _isDisposed;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FileLogHandler"/> class with the specified file path.
+        /// </summary>
+        /// <param name="path">The path of the log file.</param>
+        public FileLogHandler(string path)
```
YevgeniyShunevych, line 24: Maybe rename `path` parameter to `filePath`? Just to be certain that it is not a directory path.

dotnet/src/webdriver/Internal/Logging/LogContext.cs
```diff
@@ -137,6 +134,19 @@ public ILogContext SetLevel(Type issuer, LogEventLevel level)
...
+            if (Handlers != null && _parentLogContext != null && _parentLogContext.Handlers != null)
+            {
+                foreach (var logHandler in Handlers)
+                {
+                    if (!_parentLogContext.Handlers.Contains(logHandler))
+                    {
+                        (logHandler as IDisposable)?.Dispose();
+                    }
+                }
+            }
```
YevgeniyShunevych, line 148: I tried to refactor the highlighted code block.

```cs
foreach (var logHandler in Handlers)
{
    if (logHandler is IDisposable disposableLogHandler &&
        (_parentLogContext == null || !_parentLogContext.Handlers.Contains(logHandler)))
    {
        disposableLogHandler.Dispose();
    }
}
```

dotnet/src/webdriver/Internal/Logging/FileLogHandler.cs
nvborisenko: Done.

dotnet/src/webdriver/Internal/Logging/FileLogHandler.cs
nvborisenko: Done.

dotnet/src/webdriver/Internal/Logging/LogContext.cs
nvborisenko: Yes, this is weak place in design. In ctor want to take IList and always create new List. We should always create new list (inherited from parent), because of we want to modify this list and don't affect parent's. I will look at how to make it more confident.

dotnet/src/webdriver/Internal/Logging/LogContext.cs
nvborisenko: `Handlers` might be null. Then we should `Dispose` object when it is not handled by parent. I wrote a code from performance point of view, like checking conditions step by step before doing actual disposing.

dotnet/test/common/Internal/Logging/FileLogHandlerTest.cs
nvborisenko: Fixed in trunk.

dotnet/test/common/Internal/Logging/FileLogHandlerTest.cs
nvborisenko: Fixed in trunk.

dotnet/src/webdriver/Internal/Logging/LogContext.cs
YevgeniyShunevych: As far as I saw `public ILogHandlerList Handlers { get; }` is read-only and cannot be null, because in constructor it is always initialized with an object: `Handlers = handlers ?? new LogHandlerList(this);`. That's why I omitted null checks.

nvborisenko: Ah, right, we want to allow be `null` to avoid allocation.

nvborisenko: My code is more clear, it verifies things for `null` before acting, considering the order (to be more performant). Resolving it.

dotnet/src/webdriver/Internal/Logging/LogContext.cs
nvborisenko: Resolved by https://github.com/SeleniumHQ/selenium/commit/00b579d6e3254e282315bd40e86e37a28acf769e

PR #13258, opened 2023-12-07
[py] Add page load strategy enum

### LINKED ISSUE #13236: [🚀 Feature]: Add enum for `_PageLoadStrategyDescriptor`
### Feature and motivation

Hi!

In my project, I've encountered a situation where I need to change the default page load strategy, but the only way to find the names of the available strategies is by searching the documentation or looking at the descriptor implementation.

I think it would be convenient to have an enum for these names. This would help with hints of available strategy names and save us from having to remember their full names. It would also be great to add info about difference between these strategies in the docstring of this class, as [is done in the documentation](https://www.selenium.dev/documentation/webdriver/drivers/options/#pageloadstrategy).

Thanks!

### Usage example

If enum is named `PageLoadStrategy`, I can easily use strategy name as follows:

```python
from selenium.webdriver.chrome.options import Options, PageLoadStrategy

options = Options()
options.page_load_strategy = PageLoadStrategy.EAGER
```

M1troll:
### Description
Add `PageLoadStrategy` to get rid of writing the webdriver's page load strategy manually.

### Motivation and Context
Fixes [issue#13236](https://github.com/SeleniumHQ/selenium/issues/13236):

> Hi!
> 
> In my project, I've encountered a situation where I need to change the default page load strategy, but the only way to find the names of the available strategies is by searching the documentation or looking at the descriptor implementation.
> 
> I think it would be convenient to have an enum for these names. This would help with hints of available strategy names and save us from having to remember their full names. It would also be great to add info about difference between these strategies in the docstring of this class, as [is done in the documentation](https://www.selenium.dev/documentation/webdriver/drivers/options/#pageloadstrategy).
> 
> Thanks!
> 
> ### Usage example
> If enum is named `PageLoadStrategy`, I can easily use strategy name as follows:
> 
> ```python
> from selenium import webdriver
> from selenium.webdriver.chrome.options import Options, PageLoadStrategy
> 
> options = Options()
> options.page_load_strategy = PageLoadStrategy.EAGER
> ```

### Types of changes

### Checklist

- [x] My change requires a change to the documentation. -> [normal](https://www.selenium.dev/documentation/webdriver/drivers/options/#normal-default), [eager](https://www.selenium.dev/documentation/webdriver/drivers/options/#eager), [none](https://www.selenium.dev/documentation/webdriver/drivers/options/#none) for Python

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -56,6 +56,18 @@ def java_version
   end
 end
 
+def dotnet_version
+  File.foreach('dotnet/selenium-dotnet-version.bzl') do |line|
+    return line.split('=').last.strip.tr('"', '') if line.include?('SE_VERSION')
+  end
+end
+
+def python_version
+  File.foreach('py/BUILD.bazel') do |line|
+    return line.split('=').last.strip.tr('"', '') if line.include?('SE_VERSION')
+  end
+end
+
 # The build system used by webdriver is layered on top of rake, and we call it
 # "crazy fun" for no readily apparent reason.
 
@@ -127,7 +139,7 @@ task all: [
   :"selenium-java",
   '//java/test/org/openqa/selenium/environment:webserver'
 ]
-task all_zip: [:'prep-release-zip']
+task all_zip: [:'java-release-zip', :'dotnet-release-zip']
 task tests: [
   '//java/test/org/openqa/selenium/htmlunit:htmlunit',
   '//java/test/org/openqa/selenium/firefox:test-synthesized',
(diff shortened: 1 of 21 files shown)


isaulv:
One concern I have is that enum string objects remain Enums until a json library parses it as a string object. But if tests pass, this is fine.

diemol (approved):
Thank you, @M1troll!

Thank you for reviewing, @isaulv!

titusfortner:
@isaulv do you have specific code recommendation for that?

isaulv:
@titusfortner the enum object is fine so long it gets ran through a JSON converter. If it doesn't, and that caps object is used in some other context it may not work right.

PR #13259, opened 2023-12-07
[bidi][java] Add support for Input module (Actions)

pujagani:

### Description

Provide support for Input module in BiDi https://w3c.github.io/webdriver-bidi/#command-input-performActions.
The input required by BiDi and Classic is the same except when the origin is as an element, hence reused existing Actions class APIs.
The tests are the same as what we have currently for WebDriver classic except that to perform actions it uses BiDi. This will make porting easier in the future.

### Motivation and Context

To start porting WebDriver Classic to WebDriver BiDi for Actions. This is the long-term motivation. Before we can port, we need to add corresponding BiDi APIs and the changes address that.

### Types of changes

### Checklist


pujagani:
The failing test is due to a timeout exception when "get" command is called. I am unable to produce it locally and not sure why is it failing only on CI.

PR #13265, opened 2023-12-07
[dotnet] Change a list of downloadable files to IReadOnlyList

### LINKED ISSUE #13264: [dotnet] Change signature of GetDownloadableFiles to IEnumerable<string>?
https://github.com/SeleniumHQ/selenium/blob/1bcb948894767cf4525b565f0c5dac73cd3e15e0/dotnet/src/webdriver/IHasDownloads.cs#L32C9-L32C9

`List<string> GetDownloadableFiles();` allows add new items to the list.

I think it would be better to use `IReadonlyList<string>` or `IEnumerable<string>` (or any other immutable interface friendly with `Linq`). It is breaking change, but would not be widely spread.

nvborisenko:
### Description
Retrieved list should be immutable.

### Motivation and Context
Fixes #13264 

### Types of changes

### Checklist

### DIFF
--- a/dotnet/src/webdriver/IHasDownloads.cs
+++ b/dotnet/src/webdriver/IHasDownloads.cs
@@ -28,8 +28,8 @@ namespace OpenQA.Selenium
         /// <summary>
         /// Retrieves the downloadable files.
         /// </summary>
-        /// <returns>A list of file names available for download.</returns>
-        List<string> GetDownloadableFiles();
+        /// <returns>A read-only list of file names available for download.</returns>
+        IReadOnlyList<string> GetDownloadableFiles();
 
         /// <summary>
         /// Downloads a file with the specified file name and returns a dictionary containing the downloaded file's data.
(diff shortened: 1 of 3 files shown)


titusfortner:
Is this change worth breaking backwards compatibility without warning?

nvborisenko:
There are 2 types of breaking changes: compilation compatibility and binary compatibility. This change is related to binary compatibility (even not sure), and is not related to compilation compatibility. I suppose only 0.01% of users might be impacted.

titusfortner (approved):

nvborisenko:
I lied, it is compilation level breaking change. It breaks users who wrote:
```
List<string> myVar = driver.GetDownloadableFiles();
```

titusfortner:
I see how it is. Still want to do it, or put it back?

nvborisenko:
Now or later... Given that selenium doesn't follow semver...

PR #13271, opened 2023-12-08
[rb] Fully support Chrome 120+ old headless mode

### LINKED ISSUE #13112: [🐛 Bug]: Ruby: undefined method `logs' for #<Selenium::WebDriver::Chrome::Driver> with Chrome Beta (120) + headless mode
### What happened?

As below, when I create Selenium::WebDriver instance with Chrome Beta (v120) and the headless mode option and call the logs method, I get a NoMethodError.

```ruby
require 'selenium-webdriver'

options = Selenium::WebDriver::Options.chrome(browser_version: 'beta')
options.args << '--headless'
driver = Selenium::WebDriver.for(:chrome, options: options)

driver.logs
#=> undefined method `logs' for #<Selenium::WebDriver::Chrome::Driver:0x487c57d05253aecc browser=:"chrome-headless-shell"> (NoMethodError)
```

In contrast, with the stable version (v119) of Chrome there is no problem.

```ruby
require 'selenium-webdriver'

options = Selenium::WebDriver::Options.chrome(browser_version: 'stable')
options.args << '--headless'
driver = Selenium::WebDriver.for(:chrome, options: options)

driver.logs
# => works fine
```

After investigating, I found that `driver.bridge.browser` is `:"chrome-headless-shell"` in Chrome Beta + headless mode, and the following code is not properly handled.

- https://github.com/SeleniumHQ/selenium/blob/0c86525184355bddc44b6193ae7236f11a7fb129/rb/lib/selenium/webdriver/common/driver.rb#L330-L344

---

Investigation code
```ruby
require 'selenium-webdriver'

browser_versions = %w(stable beta)

browser_versions.each do |browser_version|
  options = Selenium::WebDriver::Options.chrome(browser_version: browser_version)
  options.args << '--headless'
  driver = Selenium::WebDriver.for(:chrome, options: options)

  puts "browser_version: #{browser_version}"
  puts "driver.respond_to?(:logs): #{driver.respond_to?(:logs)}"
  puts "driver.bridge.browser: #{driver.send(:bridge).browser}"

  puts
end
```

Output
```
browser_version: stable
driver.respond_to?(:logs): true
driver.bridge.browser: chrome

browser_version: beta
driver.respond_to?(:logs): false
driver.bridge.browser: chrome-headless-shell

```

```shell
require 'selenium-webdriver'

options = Selenium::WebDriver::Options.chrome(browser_version: 'beta')
options.args << '--headless'
driver = Selenium::WebDriver.for(:chrome, options: options)

driver.logs
```
```


### Relevant log output

```shell
2023-11-08 13:16:42 DEBUG Selenium [:selenium_manager] Selenium Manager binary found at /Users/hoshino_ta/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/selenium-webdriver-4.15.0/bin/macos/selenium-manager
2023-11-08 13:16:42 DEBUG Selenium [:selenium_manager] Executing Process ["/Users/hoshino_ta/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/selenium-webdriver-4.15.0/bin/macos/selenium-manager", "--browser", "chrome", "--browser-version", "beta", "--output", "json", "--debug"]
2023-11-08 13:16:42 DEBUG Selenium [:selenium_manager] chromedriver not found in PATH
2023-11-08 13:16:42 DEBUG Selenium [:selenium_manager] chrome not found in PATH
2023-11-08 13:16:42 DEBUG Selenium [:selenium_manager] chrome beta not found in the system
2023-11-08 13:16:42 DEBUG Selenium [:selenium_manager] Required browser: chrome 120.0.6099.5
2023-11-08 13:16:42 DEBUG Selenium [:selenium_manager] chrome 120.0.6099.5 already exists
2023-11-08 13:16:42 DEBUG Selenium [:selenium_manager] chrome 120.0.6099.5 is available at /Users/hoshino_ta/.cache/selenium/chrome/mac-x64/120.0.6099.5/Google Chrome for Testing.app/Contents/MacOS/Google Chrome for Testing
2023-11-08 13:16:42 DEBUG Selenium [:selenium_manager] Required driver: chromedriver 120.0.6099.5
2023-11-08 13:16:42 DEBUG Selenium [:selenium_manager] chromedriver 120.0.6099.5 already in the cache
2023-11-08 13:16:42 DEBUG Selenium [:selenium_manager] Driver path: /Users/hoshino_ta/.cache/selenium/chromedriver/mac-x64/120.0.6099.5/chromedriver
2023-11-08 13:16:42 DEBUG Selenium [:selenium_manager] Browser path: /Users/hoshino_ta/.cache/selenium/chrome/mac-x64/120.0.6099.5/Google Chrome for Testing.app/Contents/MacOS/Google Chrome for Testing
Starting ChromeDriver 120.0.6099.5 (dc7563588833ccad6482a3aa5848108c4387e7e5-refs/branch-heads/6099@{#21}) on port 9515
Please see https://chromedriver.chromium.org/security-considerations for suggestions on keeping ChromeDriver safe.
ChromeDriver was started successfully.
selenium.rb:10:in `<main>': undefined method `logs' for #<Selenium::WebDriver::Chrome::Driver:0x29e4ce086854de18 browser=:"chrome-headless-shell"> (NoMethodError)

driver.logs
      ^^^^^
```

### Operating System

macOS Monterey

### Selenium version

Ruby 4.15.0 (Ruby: 3.1.0, selenium-webdriver: 4.15.0)

Chrome Beta (120.0.6099.5)

chromedriver 120.0.6099.5

_No response_

neilvcarvalho:
### Motivation and Context

When Chrome's new headless mode was announced, it was mentioned that they intended to remove the old headless mode from the Chrome binary. Some people still want to use the old headless mode, as it's more lightweight, performant, and has fewer dependencies - at the expense of having fewer features.

Chrome 120+ has a standalone `chrome-headless-shell` binary to keep the old headless mode working for those who wish to use it. If your Ruby code sets up the Chrome driver with `--headless` (which currently defaults to `old`) or `--headless=old` explicitly, you will lose capabilities such as setting permissions, fetching the logs, and using the CDP API.

This change is explained on the Chrome for Developers blog: [Download old Headless Chrome as chrome-headless-shell](https://developer.chrome.com/blog/chrome-headless-shell?hl=en)

That happens because the new binary has a browser name of `chrome-headless-shell`, instead of `chrome`. That makes it be recognized as something other than Chrome, not setting its capabilities.

### Description

This commit introduces `chrome-headless-shell` as an alternative Chrome browser name, fixing this issue.

### Types of changes

### Checklist

Fixes #13112

### DIFF
--- a/rb/lib/selenium/webdriver/remote/bridge.rb
+++ b/rb/lib/selenium/webdriver/remote/bridge.rb
@@ -60,7 +60,7 @@ module Selenium
           @capabilities = Capabilities.json_create(capabilities)
 
           case @capabilities[:browser_name]
-          when 'chrome'
+          when 'chrome', 'chrome-headless-shell'
             extend(WebDriver::Chrome::Features)
           when 'firefox'
             extend(WebDriver::Firefox::Features)
@@ -82,7 +82,7 @@ module Selenium
         def browser
           @browser ||= begin
             name = @capabilities.browser_name
-            name ? name.tr(' ', '_').downcase.to_sym : 'unknown'
+            name ? name.tr(' -', '_').downcase.to_sym : 'unknown'
           end
         end
 
--- a/rb/lib/selenium/webdriver/common/driver.rb
+++ b/rb/lib/selenium/webdriver/common/driver.rb
@@ -43,7 +43,7 @@ module Selenium
 
         def for(browser, opts = {})
           case browser
-          when :chrome
+          when :chrome, :chrome_headless_shell
             Chrome::Driver.new(**opts)
           when :internet_explorer, :ie
             IE::Driver.new(**opts)
@@ -329,7 +329,7 @@ module Selenium
 
       def add_extensions(browser)
         extensions = case browser
-                     when :chrome, :msedge
+                     when :chrome, :chrome_headless_shell, :msedge
                        Chromium::Driver::EXTENSIONS
                      when :firefox
                        Firefox::Driver::EXTENSIONS


rb/CHANGES
```diff
@@ -13,6 +13,7 @@ BiDi:
 
 Chrome:
 * Fix http proxy configuration for chrome (#13093)
+* Fully support Chrome 120+ old headless mode (#13271)
```
titusfortner: Don't add to the changelog, we have an automated process that this will get in the way of.

rb/lib/selenium/webdriver/remote/bridge.rb
```diff
@@ -82,7 +82,7 @@ def session_id
         def browser
           @browser ||= begin
             name = @capabilities.browser_name
-            name ? name.tr(' ', '_').downcase.to_sym : 'unknown'
+            name ? name.tr(' -', '_').downcase.to_sym : 'unknown'
```
titusfortner, line 85: doesn't `#tr` only take one character? I think we need to chain the `#tr` or use `#gsub`.

titusfortner (changes requested):
Thanks for this PR. 

Let's leave the test files alone, those get toggled off of the values we pass in, not what comes back from the browser. (also we just aren't going to be running tests with the old headless).

rb/lib/selenium/webdriver/remote/bridge.rb
neilvcarvalho: It can take more than one character. One example on the documentation is:

```rb
'hello'.tr('aeiou', '-') # => "h-ll-"
```
> If replacements is shorter than selector, it is implicitly padded with its own last character:

Docs: https://ruby-doc.org/3.2.2/String.html#method-i-tr

rb/CHANGES
neilvcarvalho: Oh, great!

neilvcarvalho: Done, I removed the commit that introduced this.

neilvcarvalho:
> Let's leave the test files alone, those get toggled off of the values we pass in, not what comes back from the browser. (also we just aren't going to be running tests with the old headless).

@titusfortner I removed the specs changes, amending on the same commit. Thank you for the explanation!

rb/lib/selenium/webdriver/remote/bridge.rb
titusfortner: Well... today I learned... lol
I wonder how many times I've chained `#tr` unnecessarily...
Thanks.

titusfortner (approved):

ryanb:
I'm running into this issue and having trouble with the new headless. Do you know when the next release will be that contains this fix?

titusfortner:
The current plan is to release about as often as Google does https://chromiumdash.appspot.com/schedule
The recommendation is to use `--headless=new`

ryanb:
Thanks. I'm having trouble migrating to `--headless=new` so for now I'll stick to Chrome v119 until this PR is released.

For those who may be facing the same issue, you can set the version like this.

```rb
Selenium::WebDriver::Chrome::Options.new(
  args: ["headless=old"],
  browser_version: "119",
),
```

jeppester:
> The current plan is to release about as often as Google does https://chromiumdash.appspot.com/schedule The recommendation is to use `--headless=new`

I think a lot of projects will need to implement this temporary workaround in order to get their CI to pass.
Many of hours will be spent investigating the problem and eventually implementing the workaround, and then after the next release, the workaround will become obsolete code in those projects.

@titusfortner
Is it completely out of question to make a bugfix release containing this fix?

titusfortner:
Using `--headless=new` is not a temporary workaround. It is the recommendation going forward. And maybe on a patch release, I've got other things on my plate right now, so haven't looked at what's feasible.

ryanb:
@titusfortner a couple questions that might help those migrating to the new headless.

1. Should we file a bug if we find an inconsistency between old and new headless?
2. If old and new headless has intentional differences, are those differences documented somewhere?

Thanks for your time.

titusfortner:
> Should we file a bug if we find an inconsistency between old and new headless?

Probably. It depends on if it was actually a bug in the old headless. :)
Old Headless was essentially a stripped down fake browser.
New Headless is an actual headless implementation of Real Chrome.

ryanb:
@titusfortner thanks for the reply.

> Probably. It depends on if it was actually a bug in the old headless. :)

I guess a better check is if it works in non headless chrome but not in new headless chrome. In my case drag & drop isn't working in new headless but it does work in old headless and non headless Chrome. I'll look into it further and file a bug once I've isolated the issue.

titusfortner:
Oh that's interesting. Drag and Drop with HTML 5 draggable / droppable elements hasn't actually ever worked with the  drivers as far as I know. (perpetual complaints as you may have guessed).

runephilosof-abtion:
> Using `--headless=new` is not a temporary workaround. It is the recommendation going forward.

It is a temporary workaround in projects that use selenium, because it is only necessary until this commit is released.

Just see the linked pull requests
Shopify https://github.com/Shopify/maintenance_tasks/pull/936
> Instead of waiting for a new release of selenium-webdriver that includes the fix to continue using the old headless, or configuring Selenium to use an older version of Chrome, this moves to the new headless Chrome.

Openproject https://github.com/opf/openproject/pull/14426
> This will be fixed in next selenium-webdriver with this PR: https://github.com/SeleniumHQ/selenium/pull/13271
> 
> In the meantime, using the new headless mode with --headless=new fixes the issue.

They are working around the problem that this pull request fixes. It seems they would have preferred to just update to a selenium-webdriver version that included this fix.

If it is the recommendation going forward to add `--headless=new`, then that should be more prominent in the documentation. The only place it is mentioned is https://www.selenium.dev/documentation/webdriver/browsers/edge/#arguments

There are probably also a lot of frameworks that need to catch up and add `--headless=new` to their templates, so it would be a nice Christmas gift for a lot of people that they would not have to use a lot of time to figure out that they need to add `--headless=new`.

If you release a patch version, everyone can upgrade to `--headless=new` when they are ready, or when it becomes the default.

titusfortner:
> It is a temporary workaround in projects that use selenium, because it is only necessary until this commit is released.

We're arguing semantics. Encouraging people to use the recommended settings shouldn't include the baggage of the term "workaround," even if it entails "extra work," but I think we each get the other's point here.

Google will soon(?) be switching `--headless` to use `--headless=new` instead of `--headless=old` as it is now. I wish they had done this at the same time as making a backwards incompatible name change, or at least given us a heads up about it, but they didn't.

There is code in trunk that for /reasons/ cannot be released, yet, so doing a patch release is no longer an option. The next minor release will be for all languages when Chrome 121 comes out in January.

neilvcarvalho:
For folks who were following this and want to use the old headless mode: this was released this week 🎉

PR #13278, opened 2023-12-11
[py] Update WPEWebKit support code

lauromoura:
### Description

This PR changes the support code around WPEWebKit helper modules after the last update, a couple of years ago.

### Motivation and Context

In the recent Selenium versions, some changes in the way the driver classes are named and the move towards options instead of raw capabilities dictionaries, alongside other minor issues, made the existing WPEWebkit module to fail to create the driver correctly. For example, failing either to find the right service executable or failing the capability verification.

The WPEWebKit is a WebKit port focused on embedded devices. Developers creating WebApps for these devices can take advantage of the remote webdriver (used by these helper classes) to automate testing directly on the device.

### Types of changes

### Checklist

  - There's a change in a `wpewebkit.Service` parameter, but they're not yet covered by the API docs. Should I add them in a separate PR?
  - Ditto
  - The changes should be covered by the existing tests, so no new tests added.
  - I didn't see any failure in the unit tests and most of the `//py:test-wpewebkit` passes, with some specific failures that will be dealt in a separate PR.

### DIFF
--- a/py/selenium/webdriver/wpewebkit/webdriver.py
+++ b/py/selenium/webdriver/wpewebkit/webdriver.py
@@ -48,18 +48,18 @@ class WebDriver(RemoteWebDriver):
          - desired_capabilities : Dictionary object with desired capabilities
          - service_log_path : Path to write service stdout and stderr output.
         """
-        if options:
-            capabilities = options.to_capabilities()
-            capabilities.update(desired_capabilities)
-            desired_capabilities = capabilities
-        else:
+        if not options:
             options = Options()
+
+        if desired_capabilities:
+            for name, value in desired_capabilities.items():
+                options.set_capability(name, value)
 
         self.service = Service(executable_path, port=port, log_path=service_log_path)
         self.service.path = DriverFinder.get_path(self.service, options)
         self.service.start()
 
-        super().__init__(command_executor=self.service.service_url, desired_capabilities=desired_capabilities)
+        super().__init__(command_executor=self.service.service_url, options=options)
         self._is_remote = False
 
     def quit(self):
--- a/py/selenium/webdriver/wpewebkit/webdriver.py
+++ b/py/selenium/webdriver/wpewebkit/webdriver.py
@@ -55,7 +55,7 @@ class WebDriver(RemoteWebDriver):
             for name, value in desired_capabilities.items():
                 options.set_capability(name, value)
 
-        self.service = Service(executable_path, port=port, log_path=service_log_path)
+        self.service = Service(executable_path, port=port, log_output=service_log_path)
         self.service.path = DriverFinder.get_path(self.service, options)
         self.service.start()
 
--- a/py/selenium/webdriver/wpewebkit/webdriver.py
+++ b/py/selenium/webdriver/wpewebkit/webdriver.py
@@ -17,12 +17,10 @@
 
 import http.client as http_client
 
-from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
 from selenium.webdriver.common.driver_finder import DriverFinder
 from selenium.webdriver.remote.webdriver import WebDriver as RemoteWebDriver
 
 from .options import Options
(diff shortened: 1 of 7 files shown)


py/selenium/webdriver/common/driver_finder.py
```diff
@@ -40,7 +40,7 @@ def get_path(service: Service, options: BaseOptions) -> str:
             msg = f"Unable to obtain driver for {options.capabilities['browserName']} using Selenium Manager."
             raise NoSuchDriverException(msg) from err
 
-        if path is None or not Path(path).is_file():
```
diemol: Why is this needed?

py/selenium/webdriver/wpewebkit/webdriver.py
```diff
@@ -48,18 +48,18 @@ def __init__(
...
-        if options:
-            capabilities = options.to_capabilities()
-            capabilities.update(desired_capabilities)
-            desired_capabilities = capabilities
-        else:
+        if not options:
             options = Options()
 
-        self.service = Service(executable_path, port=port, log_path=service_log_path)
+        if desired_capabilities:
```
titusfortner: We shouldn't be supporting desired_capabilities in driver classes.

py/selenium/webdriver/common/driver_finder.py
```diff
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 import logging
-from pathlib import Path
+import shutil
```
titusfortner: I think we should prefer pathlib to shutil. @isaulv / @AutomatedTester opinions?

py/selenium/webdriver/wpewebkit/webdriver.py
```diff
@@ -48,18 +48,18 @@ def __init__(
...
+            for name, value in desired_capabilities.items():
+                options.set_capability(name, value)
+
+        self.service = Service(executable_path, port=port, log_output=service_log_path)
```
titusfortner: A user should be able to pass in their own Service class, but it makes sense to require executable_path to be present for this class.

titusfortner (changes requested):
To be honest, I think these classes should be in a separate library from Selenium. You should be able to subclass what is there and have it work properly. (If you can't, we should fix that). If it's in Selenium, it should match what the other drivers are doing.

py/selenium/webdriver/wpewebkit/webdriver.py
lauromoura: Right. I'll remove in favor of only the newer `options` scheme.

py/selenium/webdriver/wpewebkit/webdriver.py
lauromoura: Right. I'll update the `wpewebkit.WebDriver` API to match the Safari one, receiving a Service instead of the individual service details.

py/selenium/webdriver/common/driver_finder.py
lauromoura: I'll revert this change and make `wpewebkit.Service` use `shutil.which` internally for getting the right `WPEWebDriver` path as the default executable path.

py/selenium/webdriver/common/driver_finder.py
lauromoura: It should not be needed, I'm moving the path expansion into the driver-specific code.

py/selenium/webdriver/common/driver_finder.py
sandeepsuryaprasad: I think pathlib should be fine..
https://discuss.python.org/t/incrementally-move-high-level-path-operations-from-shutil-to-pathlib/19208

lauromoura:
(Updated after missign a tox fix and rebase)

diemol (approved):

lauromoura:
Pinging @titusfortner for review

PR #13285, opened 2023-12-11
update proxyType values to match w3c spec

titusfortner:
### Description
* Update comments; there are no defaults in spec (https://w3c.github.io/webdriver/#ref-for-dfn-own-properties-2)
* Remove outdated Firefox integer values
* Remove invalid reserved option
* Require proxyType to be set before using

### Motivation and Context
Making the libraries consistent between languages and updating to match w3c behavior.

### Notes
* Looking at the JS code, I don't think it needs any changes to match.
* I'm not sure how much of this can/should be marked deprecated instead of just making these changes to conform to the spec.
* I was going to remove "Unspecified" as an enum, but it makes sense to keep it I think in Java & .NET because strict, but not as much in Python, so removed to match Ruby & JS implementations


p0deje (approved):

diemol (approved):

dotnet/src/webdriver/Proxy.cs
```diff
@@ -51,15 +47,15 @@ public enum ProxyKind
...
-        /// Use the system values for proxy settings (default on Linux).
+        /// Use the system values for proxy settings.
         /// </summary>
         System,
 
         /// <summary>
-        /// No proxy type is specified.
+        /// No proxy type is specified. This must be changed before use
         /// </summary>
         Unspecified
```
nvborisenko, line 63: I think we can easily remove it. Right now user may write something like:
```csharp
var proxy = new Proxy();
// and then it was expected user should specify proxy kind
```

If we make proxy kind as required in constructor, then everything is resolved at compilation level:
```csharp
var proxy = new Proxy(ProxyKind.Direct);
```

I think it can be done in scope of other PR, seems there is additional API how to change proxy kind. Need to walk through it deeper.

titusfortner: Well, that's definitely a backwards incompatible change. Plus I figured it would make sense to keep the implementation close to what Java is doing, still.

nvborisenko: That's why we need to look at this deeper later. Looks not bad to me.

titusfortner:
I feel like I read something completely different every time I look at the spec:

> Otherwise
[Set a property](https://w3c.github.io/webdriver/#dfn-set-a-property) of capabilities with name "proxy" and a value that is a new JSON [Object](https://w3c.github.io/webdriver/#dfn-object).

Grid is deserializing the empty Proxy and failing, so I need to remove the errors I'm throwing here. 😂

pujagani (approved):

diemol:
@titusfortner what is left to do here?

PR #13297, opened 2023-12-12
update to use bazel 7

titusfortner:
Bazel 7 was released yesterday — https://blog.bazel.build/2023/12/11/bazel-7-release.html

It seems to be a drop-in replacement even if we don't update from `WORKSPACE` to `MODULE.bazel`

Let's see what breaks...


titusfortner:
looks like updating rules_jvm made it worse 😂 

Obviously I do not know what I'm doing and will wait for someone with more experience (@shs96c / @p0deje) to take a look at this.

titusfortner:
Simon said he'd work on this after getting the JS things sorted. Do not need this one.

PR #13308, opened 2023-12-14
[java] removed usage of FileBackedOutputStream in the client 

joerg1985:
### Description
This will remove the usage of `FileBackedOutputStream` in the client, i don't think it is needed to cache on disk in the client for the following reasons:
- The code did already encode the given Capabilities to a string, so everything was in memory before writing it to disk, see old implementation of `NewSessionPayload.create(Capabilities caps)`. This might has been different in the days before the W3C protocol was implemented.
- JEP 254: Compact Strings, a string has usually only 50% of the size like in Java 8
- Memory size has increased the last years

The server will still use a `FileBackedOutputStream` to keep memory usage low, see `RequestConverter` as this is out of focus from #12737.

**Please notice:**
There is a new interface created in this PR named `Contents.Supplier`. In all areas we can easily ensure the InputStream can be created multiple times and we already know the length of the content. So the `Contents.memoize` can be dropped which used a `FileBackedOutputStream` to ensure the InputStream can be created multiple times. 
Providing the length does help the HttpClient not to start chunking requests, some drivers do not support this.

This new interface is used by the `HttpMessage` too, the old setter has been deprecated and delegates to the new setter.

### Motivation and Context
This is part of #12737
 
### Types of changes

### Checklist


diemol (approved):
This looks good to me, thanks @joerg1985!

I think the idea of putting the content to disk is to help session creation request when the user sends a huge Firefox profile. But this is probably fine these days for the reasons you mention in the PR description. 

However, I would prefer a review from @shs96c, as he wrote most of this code.

Issue #13019, opened 2023-10-22
[🚀 Feature]: Release Selenium Manager with Bazel


titusfortner:
### Feature and motivation

Right now bazel is built and released with Cargo. It would be much better for our build if what is built by bazel matches exactly for that architecture what is built for release. This may not be possible since I'm not sure bazel supports doing different things depending on architecture, which means it would actually require cross-compilation that we've determined we can't do. Would be nice to document this either way.

### Usage example

```
cargo build --release
```
should do the same thing as
```
bazel build //rust:selenium-manager
```

diemol:
Didn't @shs96c try this and could not do it due to the cross-compile issues?

titusfortner:
I wasn't sure if it was an absolute impossibility or a roadblock that was more trouble than it was worth that day. Either way I want to have documented what doesn't work. :)

titusfortner:
I'm going to close this one. It seems like if we can't cross-compile, the current solution is good enough, and we are probably more likely to re-implement SM in Go than to improve this.

Issue #12620, opened 2023-08-27
[🚀 Feature]: Use Bazel for Selenium Manager


titusfortner:
### Feature and motivation

What we are testing with Bazel is not what we are building with Cargo.

Let's get this working the right way with bazel, so we don't have to troubleshoot issues when GitHub runners changing system settings.

### Usage example

`bazel build //rust:selenium-manager`

titusfortner:
Now that we are using released artifacts, we are using the same thing for tests as for releases, which should be good enough for now. If we decide to change things, it will be part of a bigger effort and we do not need this issue to track it.

PR #13314, opened 2023-12-15
[build] Archive Selenium Manager Binaries

### LINKED ISSUE #13150: [🚀 Feature]: Stop storing Selenium Manager binaries in Repo
### Feature and motivation

Bazel cannot build all of the binaries because of limitations with cross compiling in Rust.
Bazel cannot access the current location of binaries in the GitHub Actions assets list because it requires an authentication key.

Bazel needs a static URL to reference in order to pull the latest built version of the binaries in a location that is not checked into the selenium repo.

We should have some way to:
1. Build Regular & Debug versions of Selenium Manager whenever the Rust code changes
2. Store these files /somewhere/ that isn't in Selenium repo
3. Have a static URL to reference the latest version

One solution is to create a `seleniumhq/selenium_manager_debug` repo with a single commit, and have github actions in selenium repo publish all of the Selenium Manager assets (production & Debug) to a "latest" tag/release in that repo (similar to what we're doing in the Selenium repo with the nightly tag/release). So long as we keep the names the same (no versioning) and overwrite the previous files, the URL should be static.

### Usage example

n/a

titusfortner:
### Status

This is in Draft because it is using my fork to demo the proof of concept.
This Git Hub Actions Run — https://github.com/titusfortner/selenium/actions/runs/7219076550
Made this commit — https://github.com/titusfortner/selenium_manager_debug/commit/fab2d26a53a006f4e5a58f483abc8dee2915d7bd
And did this release — https://github.com/titusfortner/selenium_manager_debug/releases/tag/selenium-manager-fab2d26

### Description
* On every commit to trunk in the rust directory (too much? daily seemed too little and also too much?)
    * Builds all Selenium Manager binaries (including debug)
    * Creates a commit in a separate repo with the hash values of the files
    * Creates a tag from that commit and does a release, and uploads all selenium manager binaries
* Toggle Selenium Manager behavior in build based on `pinned_browsers` instead of `stamp` (the way I'm skipping downloading is definitely hacky, there is probably a better way)
* Deletes Selenium Manager binaries from repo
* `selenium_manager.py` file very similar to `pinned_browsers.py` to update the url and hash for the latest version of binaries

### Motivation and Context
* Fixes #13150 
* Removes the need for #12620 & #13019
* If running tests with Selenium Manager it will use the exact binary that the user will have when released
* Less Windows Dev Pain
* Separate Repo makes it clear that these are archives not production releases
* Still it's an easy way for someone to grab a debug version of the binary and to try it locally to help troubleshoot problems

### Future Considerations
* Remove all Bazel Rust things as Lost Cause
* ~Make `build-selenium-manager.yml` part of `ci-rust.yml`~
* Would probably be better to store all the files in the releases zipped, but I got `http_file` working before I could figure out how to get `http_archive` working


diemol (commented):
This is pretty good, thank you @titusfortner!

I think building on every commit is fine, given that it does not change a lot.

titusfortner:
Ok, I've updated the repo names to point to https://github.com/SeleniumHQ/selenium_manager_snapshots/ (the releases there were automatically generated by Selenium's CI).

I've moved the CI code from `build-selenium-manager.yml` to `ci-rust.yml`. Binaries are only built/released when rust directory has a commit and all tests are passing (or triggered manually).

titusfortner:
I had this working, then I tweaked something and now it isn't deploying. At least it is passing RBE now. Once I figure out why it isn't releasing the binaries any more, I'll merge this.

titusfortner:
Ok, this makes sense. It wasn't working as a PR because of tokens, but when I run this code as a Selenium branch, it works — https://github.com/SeleniumHQ/selenium/actions/runs/7331186361
So I'm merging it.

PR #13330, opened 2023-12-18
[dotnet] Added function to enable halting targets until runtime.runIfWaitingForDebugger is invoked

EdwinVanVliet:
### Description
This PR introduces `EnableWaitForDebuggerOnStart` function to `DevToolsSession.cs` which enables halting new targets (tabs for instance) until the `runtime.runIfWaitingForDebugger` command is invoked. For instance when wanting to capture the network traffic on a new target, the network domain should be instructed to do so before the target is doing any network traffic. 

In PR https://github.com/SeleniumHQ/selenium/pull/13175 the datastructure for listening to the TargetAttached events was introduced. 

In a follow up pull request I will introduce the ability to sent multiple DevTools commands in a single SendCommand function. This will allow proper handling of the TargetAttached event. The singular SendCommand function will get blocked because the commands running against that target will be blocked until `runtime.runIfWaitingForDebugger` is invoked. By bundling the commands we can get around that issue.

Also see; https://github.com/SeleniumHQ/selenium/pull/10603

### Motivation and Context
When a new target (new tab for instance) is created DevTools continues execution based on the WaitForDebugger property set in the SetAutoAttach feature. When the WaitForDebugger property is set to false DevTools continues without waiting the consumer to properly enable all domains on the target. For instance when being interested in network events from the network domain we need to invoke Network.enable for that specific target. The current behavior causes some of the network events to be missing due to Selenium enabling the domains after the target already performed network requests.

By allowing consumers to enable the WaitForDebugger property we can enable all domains and tell DevTools whenever we are ready by using the Run.runIfWaitingForDebugger command.

### Types of changes

### Checklist


dotnet/src/webdriver/DevTools/IDevToolsSession.cs
```diff
@@ -40,6 +40,13 @@ public interface IDevToolsSession : IDisposable
         /// </summary>
         event EventHandler<DevToolsEventReceivedEventArgs> DevToolsEventReceived;
 
+        /// <summary>
+        /// Enables waiting for the debugger when opening a new target (tab for instance).
+        /// The target will be halted until the Runtime.runIfWaitingForDebugger command is invoked.
+        /// </summary>
+        /// <returns></returns>
+        Task EnableWaitForDebuggerOnStart();
```
nvborisenko: Running this way will infinitely increase interface. Today it is `EnableSomething()`, tomorrow - `DisableSomething()`, the day after tomorrow `EnableSomethingElse()`. I think of that probably it would be better to introduce some kind of `DevToolsOptions`. This object represents desired options in the given context, optionally provided by user in class constructor, or in method call.

Example (pseudo-code):
```csharp
var session = new DevToolsSession(..., new DevToolsOptions() { WaitForDebugger = true });
```

This above is just an example how we can introduce new API without breaking changes in future, `Options` can be extended. I am still not sure it is the best suggestion, need to dive deeper into issue.

PS: The biggest issue is that we cannot cover all scenarios provided by CDP in "simplified" API.

EdwinVanVliet: Agreed, using an options class seems like a better approached. I've committed some changes to remove EnableWaitForDebuggerOnStart function from the IDevToolsSession interface. Introduced DevToolsSessionOptions.cs to enable this functionality. By default it will be turned off.

nvborisenko: Looks better, thanks. Seems `protocolVersion` argument can be a part of new Options class. If you can, please make these contructors/methos as obsolete and point users to use new Options class.

EdwinVanVliet: Committed some changes to make the DevToolsOptions more prominent in the API. Kept the old version and marked them as obsolete to prevent introducing issues for consumers who these specific functions. 

Setting the AutoDetectDevToolsProtocolVersion option is moved downstream (DevToolsSession.cs) instead of on the specific Driver implementations.

dotnet/src/webdriver/DevTools/DevToolsOptions.cs
```diff
@@ -0,0 +1,37 @@
+// <copyright file="DevToolsSession.cs" company="WebDriver Committers">
```
nvborisenko: Incorrect file name.

dotnet/src/webdriver/DevTools/DevToolsOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace OpenQA.Selenium.DevTools
+{
+    /// <summary>
+    /// Contains options configuring the DevTools session.
+    /// </summary>
+    public class DevToolsOptions
+    {
+        /// <summary>
+        /// Enables or disables waiting for the debugger when creating a new target.
+        /// If enabled, all targets will be halted until the runtime.runIfWaitingForDebugger is invoked.
```
nvborisenko: Let's mention what is default value for this option.

dotnet/src/webdriver/DevTools/IDevToolsSession.cs
nvborisenko: Thanks, looks good.

nvborisenko:
Looks good to me, thank you @EdwinVanVliet . Last ask: can we create at least one test which demonstrates how to use this new feature in area of capturing network requests in newly opened browser tab?

EdwinVanVliet:
@nvborisenko 

Sounds like a good idea, however with current setup that's not possible yet. When halting the tab an event is raised indicating a new target is created. To continue that target at least two commands need to be issued (runtime.runIfWaitingForDebugger and Network.enable). The problem here is that command will be blocked until runtime.runIfWaitingForDebugger is invoked.

To workaround that we sent the commands in one batch without blocking each other. That will be introduced in the following PR and I will try to add a test case demonstrating the behavior with that.

dotnet/src/webdriver/DevTools/DevToolsOptions.cs
EdwinVanVliet: Fixed

dotnet/src/webdriver/DevTools/DevToolsOptions.cs
EdwinVanVliet: Fixed

nvborisenko: Forgot to push?

nvborisenko:
Thanks @EdwinVanVliet , looking forward your contribution.

nvborisenko (approved):

PR #13356, opened 2023-12-23
[build] update script for change logs

titusfortner:
Making this a PR just to get feedback...

This script limits what is getting added to changelogs to the src or lib directories, so changes to root or test directories won't get added to changelogs. Is this what we want. I typically don't add to changelogs anything that doesn't affect users, but maybe that's not what we want? Let me know.

Also, because this took a lot of work... this works for both minor & patch releases, and is automatically called from the versions task.

For instance, this bumps patch version and adds java commits to the readme since previous patch version
```
./go java:version["4.16.2"]
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -624,7 +624,7 @@ namespace :py do
   desc 'Update Python changelog'
   task :changelog do
     header = "Selenium #{python_version}"
-    update_changelog(python_version, 'py', 'py/', 'py/CHANGES', header)
+    update_changelog(python_version, 'py', 'py/selenium/webdriver', 'py/CHANGES', header)
   end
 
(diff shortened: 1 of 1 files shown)


diemol (commented):
Does this commit the changes or leaves changes to be reviewed before committing and pushing?

titusfortner:
The automated build process I've been working on always prompts before commits, it doesn't prompt for tagging because we can fix that if it's wrong.

I ended up already merging most of what was originally in this PR just so I could work with the whole build process more easily. 

This PR is now just asking how much to limit what goes in the changelog

diemol (approved):
Then it looks good to me.

PR #13372, opened 2023-12-29
[build] Remove features that are not w3c compliant

titusfortner:
### Description
* Remove non-compliant commands that are not being used in Selenium code anywhere
    * GET_ALL_SESSIONS
    * ELEMENT_EQUALS
    * SET_ALERT_CREDENTIALS
    * GET_APP_CACHE_STATUS
    * IS_BROWSER_ONLINE
    * SET_BROWSER_ONLINE
    * GET_SESSION_LOGS
* Remove commands that are implemented by the single subclass (`W3CHttpCommandCodec`)
    * GET_AVAILABLE_LOG_TYPES
    * GET_ELEMENT_LOCATION
    * GET_ELEMENT_SIZE
    * GET_LOG
* Deprecate features that shouldn't work because the endpoints are no longer valid
    * Getting & Setting location (geo-location)
    * Getting & Setting network connection

### Motivation and Context
* Slowly removing things that don't work or at least are not compliant

### Future Considerations
* Need to contact Appium about things they need to implement so we can get rid of them (alternately we can change the namespacing on them to be w3c compliant; kind of depends on what we want to live where in the two code bases)
    * GET_SCREEN_ORIENTATION
    * SET_SCREEN_ORIENTATION
    * GET_SCREEN_ROTATION
    * SET_SCREEN_ROTATION
    * SWITCH_TO_CONTEXT
    * GET_CURRENT_CONTEXT_HANDLE
    * GET_CONTEXT_HANDLES
* Can we merge these classes since we no longer have multiple implementations? I don't think this would affect how we are moving to BiDi spec.

### DIFF
--- a/java/src/org/openqa/selenium/chromium/ChromiumDriver.java
+++ b/java/src/org/openqa/selenium/chromium/ChromiumDriver.java
@@ -302,11 +302,13 @@ public class ChromiumDriver extends RemoteWebDriver
   }
 
   @Override
+  @Deprecated
   public Location location() {
     return locationContext.location();
   }
 
   @Override
+  @Deprecated
   public void setLocation(Location location) {
     Require.nonNull("Location", location);
     locationContext.setLocation(location);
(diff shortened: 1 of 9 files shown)


diemol (approved):
This makes sense, if the build passes, you should merge it.

Issue #10397, opened 2022-02-24
Remove HTML 5 functionality helper code


titusfortner:
### Feature and motivation

This is ApplicationCache, WebStorage, & LocationContext

### Usage example

#10344

sc-mk:
Hi Titus,
any suggestions on how to solve the problem with the MoveToElement method, please?

titusfortner:
I think what needs to be decided here is how we are handling all of the HTML5 functionality across all the languages. 

Next step is essentially an inventory of which of the storage/cache/location things are being supported by JS and which have been removed and decide which direction we want for all of the bindings.

symonk:
@titusfortner what do we do in the short term? `driver.application_cache` simply raises and exception right now and will not work in the python side - do we remove it (it's already deprecated) until we come up with the above and possibly implement something else?

titusfortner:
Next step was to figure out which languages marked as deprecated, which removed and which implemented them in JS.

Then we figure out how they should be done in all of them.

titusfortner:
More accurately this should be discussing the HTML functionality that was added back in 2015

ApplicationCache, WebStorage, & LocationContext

Need to inventory which languages have methods pointing to outdated endpoints, which ones have re-implemented with JS and which have deleted them.

titusfortner:
Per [TLC Meeting](https://www.selenium.dev/meetings/2022/tlc-11-10/):

> @diemol will list the changes that need to be done in the bindings and then deprecate/remove them

iampopovich:
Hi 👋
I'm planning to assist with this task. Please let me know if there has been any progress or new discussions about where to start. For now, I intend to familiarize myself with the implemented methods and follow this plan: "Need to inventory which languages have methods pointing to outdated endpoints, which ones have re-implemented with JS and which have deleted them."

titusfortner:
This has been lower priority/mostly forgotten about. Any help would be greatly appreciated!

aguspe:
Hi I would like to help with the ruby bindings for this @titusfortner I will start working on it now and push a PR over the weekend

aguspe:
After the merge of #14276 is there anything we can continue working on regarding this for the ruby bindings? or should we wait until a general decision on all the bindings is done? @titusfortner

aguspe:
Would it be possible to remove the rb label here @p0deje ? 

I'm just looking for something new to pick up and it will help me filter, thank you!

juddgaddie:
[...]

PR #13373, opened 2023-12-29
[java] deprecate all html5 offline storage implementations See #10397

titusfortner:
### Description
Deprecates all LocalStorage and WebStorage implementations
Also deprecates `AppCacheStatus` enum which isn't being used anywhere and we probably could just delete

### Motivation and Context
See #10397

Here is the JavaScript equivalent of the methods being deprecated:
https://github.com/SeleniumHQ/selenium/blob/22dcb1796d5c791c1f6d5dde947b11fe5bd03748/java/src/org/openqa/selenium/remote/codec/w3c/W3CHttpCommandCodec.java#L220-L266

Do we want to put this in our docs somewhere and point people to it, or just tell them to find what they need online?

### DIFF
--- a/java/src/org/openqa/selenium/chromium/ChromiumDriver.java
+++ b/java/src/org/openqa/selenium/chromium/ChromiumDriver.java
@@ -292,11 +292,13 @@ public class ChromiumDriver extends RemoteWebDriver
   }
 
   @Override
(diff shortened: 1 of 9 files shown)


diemol (approved):
Thank you, @titusfortner!

Issue #11599, opened 2023-01-30
[🚀 Feature]: Error when Selenium Manager used on unsupported architecture


titusfortner:
### Feature and motivation

Right now, Selenium Manager is called when a driver isn't specified or found on PATH. We only support 3 architectures, though. We should add code to explicitly check the user's architecture in preparation for additional additional architectures and throw an error until the architecture is actually supported. 

This isn't as much of a problem while the Manager is in a "fall back" mode, but it can cause unexpected/confusing errors especially once we stop hiding them.

titusfortner:
Turns out that Mac ARM architecture works just fine with AMD Selenium Manager, so that one isn't an issue. 

So, I'm going off of this list here: https://github.com/SeleniumHQ/selenium/issues/11357

There are drivers available in various places for each of these architectures, but our current selenium manager binaries won't even run on them:
linux-32
win32
linux-arm64
win-arm64
linux-armv7

If we checked in a 32 bit version of Selenium Manager for Windows & Linux, it should also work for 64 bit OS, right?

titusfortner:
Here are some rough numbers for [geckodriver downloads](https://tooomm.github.io/github-release-stats/?username=mozilla&repository=geckodriver):


| Architecture  | Percent |  Currently Supported |
|---------------|---------|-----------|
| linux64       |     51% | &#10004;   |
| win64         |     24% | &#10004;   |
| macos         |      9% | &#10004;   |
| macos-aarch64 |      5% | &#10004;   |
| win32         |      5% |    |
| win-aarch64   |      2% |    |
| linux-aarch64 |      2% |   |
| linux32       |      1% |    |

So presumably we are currently supporting "only" ~90% of usage

bonigarcia:
To clarify these figures a bit, the Selenium Manager Rust code already supports every platform in geckodriver (including win32, linux32, etc.). But it is true that the current compilation is only for win64, linux64, and mac64. So win32 and linux32 cannot be downloaded with Selenium Manager since the binary cannot be executed. But I have doubts ARM64 in Windows and Linux. Maybe the x64 binary can be executed in Windows and Linux (like in macOS). In that case, the `win-aarch64` and `linux-aarch64` are also supported. Unfortunately, I don't have an ARM64 machine for Windows or Linux to check it.

titusfortner:
I'm almost certain arm64 will need to be supported separately for Linux/Windows.

[...]

PR #13381, opened 2024-01-01
SeleniumManager python wrapper should check if architecture/platform combination is supported

seidnerj:
### Description
SeleniumManager's get_binary() now checks if architecture/platform combination is supported (and not only if the platform is supported).

### Motivation and Context
Currently, when running selenium manager on unsupported architectures, best case scenario the user gets an OS level error message that's not necessarily clear, worst case scenario - the binary is completely missing and then the user gets a "Unable to obtain working Selenium Manager binary" message (for example, this happens on RP, since all RPG are set to install wheels from "piwheel" repo which seems to discard all executables that are not supported by RPs architectures).

### Types of changes

### Checklist

### DIFF
--- a/py/selenium/webdriver/common/selenium_manager.py
+++ b/py/selenium/webdriver/common/selenium_manager.py
@@ -17,9 +17,9 @@
 import json
 import logging
 import os
+import platform
 import subprocess
 import sys
-import platform
 from pathlib import Path
 from typing import List
 
--- a/py/selenium/webdriver/common/selenium_manager.py
+++ b/py/selenium/webdriver/common/selenium_manager.py
@@ -19,6 +19,7 @@ import logging
 import os
 import subprocess
 import sys
+import platform
 from pathlib import Path
 from typing import List
 
@@ -47,17 +48,19 @@ class SeleniumManager:
             return Path(path)
 
         dirs = {
-            "darwin": "macos",
-            "win32": "windows",
-            "cygwin": "windows",
-            "linux": "linux",
-            "freebsd": "linux",
-            "openbsd": "linux",
+            ("darwin", "any"): "macos",
+            ("win32", "any"): "windows",
+            ("cygwin", "any"): "windows",
+            ("linux", "x86_64"): "linux",
+            ("freebsd", "x86_64"): "linux",
+            ("openbsd", "x86_64"): "linux",
         }
 
-        directory = dirs.get(sys.platform)
+        arch = platform.machine() if sys.platform in ("linux", "freebsd", "openbsd") else "any"
+
+        directory = dirs.get((sys.platform, arch))
         if directory is None:
-            raise WebDriverException(f"Unsupported platform: {sys.platform}")
+            raise WebDriverException(f"Unsupported platform/architecture combination: {sys.platform}/{arch}")
 
         if sys.platform in ["freebsd", "openbsd"]:
             logger.warning("Selenium Manager binary may not be compatible with %s; verify settings", sys.platform)


titusfortner:
Aha, found the issue where I was originally discussing this, with this resolution: https://github.com/SeleniumHQ/selenium/issues/11599#issuecomment-1518253608

I'm a little concerned about a strict allow list, but I guess if there are problems, we can update it again.

seidnerj:
I'm not in-love with this solution but I think it's better than what we currently have, which I guess is also a kind of strict allow list as it is. We can implement a more comprehensive solution by checking the file format of the executable included or embedding this information during the build process but I don't think the ROI for that is worth it.

seidnerj:
Just read everything, glad I could pitch in :)

titusfortner:
Thanks!

viniciusd:
Our code worked on previous versions on Docker running on Mac (hence Linux ARM) up to version 4.16. After this change, it won't initialize the browser anymore, so it looks like this was a breaking change

Is the `SeleniumManager` incompatible with `linux/aarch64`? How did things work on the previous versions? Should the `WebDriver` initializer fall back out of `SeleniumManager` if the architecture isn't allow-listed?

__Edit__

Allow-listing linux/aarch64 does the trick, i.e., adding `("linux", "x86_64"): "linux",` to `dirs` enables my existing code to run again. Is this an acceptable solution? Should I make a PR with it?

seidnerj:
> Our code worked on previous versions on Docker running on Mac (hence Linux ARM) up to version 4.16. After this change, it won't initialize the browser anymore, so it looks like this was a breaking change
> 
> 
> 
> Is the `SeleniumManager` incompatible with `linux/aarch64`? How did things work on the previous versions? Should the `WebDriver` initializer fall back out of `SeleniumManager` if the architecture isn't allow-listed?
> 
> 
> 
> __Edit__
> 
> 
> 
> Allow-listing linux/aarch64 does the trick, i.e., adding `("linux", "x86_64"): "linux",` to `dirs` enables my existing code to run again. Is this an acceptable solution? Should I make a PR with it?

You mean adding `("linux", "aarch64"): "linux",` to `dirs`, right?

I was under the impression that binaries for aarch64 did not exist yet. @titusfortner?

viniciusd:
> You mean adding `("linux", "aarch64"): "linux",` to `dirs`, right?
 
Yes, but it probably isn't the correct way of approaching it as it might not be supported by the Selenium manager.

I am looking into the code to understand what is happening. I have the driver installed in my PATH, I do not rely on the Selenium Manager, yet this check prevents Selenium from finding the driver as if relying on the Manager became mandatory.

I manage the binaries myself and this shouldn't be prevented. A failure on the Manager should still prompt a PATH search

seidnerj:
I don't know in what order does Selenium resolve the binaries, if either first use Selenium Manager and fallback to binaries found in path or vice versa, but either way it doesn't sound like this has to do with this specific PR. 

What you're saying really is that if Selenium Manager is not available no fallback to binaries potentially found in path occurs. That's a different and separate issue. 

I am unfortunately not familiar with that code but you're of course welcome to fix this and create a PR for the maintainers' review (I am not a maintainer of Selenium).

diemol:
https://www.selenium.dev/documentation/selenium_manager/#alternative-architectures

seidnerj:
Thanks for this @diemol !

The relevant section: "If you are running Linux on arm64/aarch64, 32-bit architecture, or a Raspberry Pi, Selenium Manager will not work for you. The biggest issue for people is that they used to get custom-built drivers and put them on PATH and have them work. Now that Selenium Manager is responsible for locating drivers on PATH, this approach no longer works, and users need to use a Service class and set the location directly. There are a number of advantages to having Selenium Manager look for drivers on PATH instead of managing that logic in each of the bindings, so that’s currently a trade-off we are comfortable with."
@viniciusd

viniciusd:
> I don't know in what order does Selenium resolve the binaries, if either first use Selenium Manager and fallback to binaries found in path or vice versa, but either way it doesn't sound like this has to do with this specific PR.
> 
> What you're saying really is that if Selenium Manager is not available no fallback to binaries potentially found in path occurs. That's a different and separate issue.
> 
> I am unfortunately not familiar with that code but you're of course welcome to fix this and create a PR for the maintainers' review (I am not a maintainer of Selenium).

Thanks, and sorry for rushing into commenting on this PR as it was when Selenium starting breaking on my scenario.

Now I understand it wasn't caused by this change, it is a breaking change that was exposed on this PR because previously there was no architecture distinction (on 4.16.0, it just does an OS lookup).

And thanks @diemol for the reference

I will keep looking into it to see if I can propose an automatic fallback that the community would be comfortable with.

I find it particularly interesting/weird how 4.16.0 works fine even though it is resolving to the x86 selenium manager, which is supposed to be incompatible already

seidnerj:
No problem 🙂

viniciusd:
You two might find it interesting that the x86 manager is indeed running and returning the correct driver path on my arm64 (Docker on Mac):
```
# /usr/local/lib/python3.9/site-packages/selenium/webdriver/common/linux/selenium-manager --browser chrome --browser-path /usr/bin/chromium
INFO	Driver path: /usr/bin/chromedriver
INFO	Browser path: /usr/bin/chromium
# file /usr/local/lib/python3.9/site-packages/selenium/webdriver/common/linux/selenium-manager
/usr/local/lib/python3.9/site-packages/selenium/webdriver/common/linux/selenium-manager: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, stripped
# uname -a
Linux 6.6.12-linuxkit #1 aarch64 GNU/Linux
```

Probably some Debian magic happening behind the curtains, I couldn't find an explanation as I don't have [Debian multiarch](https://wiki.debian.org/Multiarch/HOWTO) or even qemu installed. Well 🤷‍♂️

seidnerj:
That's strange. I am not an expert on docker but my bet is that this has to do with the fact that at the end of the day you're really running on an arm processor (arm64 and aarch64 are just different names for the same arch). 

Would love to know how this works though.

Issue #13022, opened 2023-10-22
[🚀 Feature]: Make Selenium Manager classes thin wrappers


titusfortner:
### Status
* Java — #13385 
* Ruby — #13386 
* .NET — #13833
* Python — #13387
* JS — #13853

### Feature and motivation

Make Selenium Manager class only responsible for locating the correct binary, and executing that binary with the provided arguments.

The DriverFinder class parses any options and service class details required and sends them to Selenium Manager.

This allows:
* better separation of concerns between DriverFinder and Selenium Manager (see https://github.com/SeleniumHQ/selenium/pull/12554)
* creating a Server class to get the grid with Selenium Manager (see https://github.com/SeleniumHQ/selenium/pull/12299)
* [releasing SM as a separate package](https://github.com/SeleniumHQ/selenium/issues/13021) (would be an alternative to https://github.com/SeleniumHQ/selenium/issues/11694, and would provide support for projects to use [older versions of Selenium](https://github.com/titusfortner/webdrivers/#update-future-of-this-project).

diemol:
Do you mean having something like `DriverService` but for SM?

titusfortner:
Yes, I guess it's essentially that.

diemol:
Implemented in all bindings.

PR #13385, opened 2024-01-02
[java] Selenium Manager generates output from argument list 

titusfortner:
### Description
* Have `DriverFinder` convert the `Capabilities` into the arguments required by Selenium Manager binary
* Deprecate existing `getPath()` method

### Motivation and Context
* Addresses Java portion of #13022 
* The method name `getPath()` is no longer an accurate one

### Updates
Updates:

* Changes `getPath()` to `getResult()` in `DriverFinder` as well. The name isn't great, but it is the name of the current class and I can't think of anything better. (`getPaths()`? `SeleniumManagerOutput()`?) 
* `browserVersion` is set to `null` in the 3 applicable driver classes instead of DriverFinder, and only if there is a browser path.
* Removing instead of deprecating the methods that are no longer used
* File validations are done in the Results class
* No longer a specific "Selenium Manager" error, the specific errors in SM class will bubble up anyway, and it makes the try/catch easier.

### Status
* Do we want a different name than `getResult()`?
* I'm going to add tests still


titusfortner:
Do I need to deprecate, or can I just remove because the class is annotated as `@beta`?

titusfortner:
and should we also rename `DriverFinder`'s `getPath()` method? I think `DriverFinder` still makes sense as a class name, since the browser information is related to driver finding, but the return value is no longer a path.

diemol (approved):
I would delete since it is market as beta.

My only concern is that we do not have enough tests to properly verify this refactor. Haven't we had the situation in the past where moving some arguments around ended up in some bugs?

titusfortner:
Tests are a good idea. 😂

java/src/org/openqa/selenium/remote/service/DriverFinder.java
```diff
@@ -37,7 +46,9 @@ public static Result getPath(DriverService service, Capabilities options, boolea
 
     if (result.getDriverPath() == null) {
       try {
-        result = SeleniumManager.getInstance().getDriverPath(options, offline);
+        List<String> arguments = toArguments(options, offline);
+        result = SeleniumManager.getInstance().getResult(arguments);
+        ((MutableCapabilities) options).setCapability("browserVersion", (String) null);
```
bonigarcia: Are these capabilities always mutable? Maybe it is, I'm asking just to ensure.

titusfortner: Yes, all the browser option classes inherit from mutable capabilities class.

titusfortner: That said, looking at other code, it looks like we create a new presistent capabilities object rather than mutating. So I'm thinking I'll refactor this.

titusfortner: Yeah, if this is parsed where DriverFinder is called, it doesn't need and cast and it can be limited it to the places it actually applies.

java/src/org/openqa/selenium/manager/SeleniumManagerOutput.java
```diff
@@ -139,6 +143,29 @@ public String getBrowserPath() {
...
+        Path filePath = Paths.get(entry.getValue());
+        if (!Files.exists(filePath)) {
+          throw new IllegalArgumentException(
+              "The path for '" + entry.getKey() + "' is not a valid file: " + entry.getValue());
+        } else if (!Files.isExecutable(filePath)) {
+          throw new IllegalArgumentException(
+              "The file at '" + entry.getKey() + "' is not executable: " + entry.getValue());
+        }
+      }
+    }
```
diemol: Why do we need this logic here? If we need it, shouldn't it be inside Selenium Manager?

diemol (changes requested):

java/src/org/openqa/selenium/manager/SeleniumManagerOutput.java
titusfortner: I'm pulling in checks for what user is passing in as well so they get the same logging and error messaging.

titusfortner:
Ok, refactored along the lines of Simon's suggestions, added tests, they pass, I'm generally happy with the naming I think right now, and it's matching Ruby & Python PRs. This can be merged once approved.

java/src/org/openqa/selenium/chrome/ChromeDriver.java
```diff
@@ -96,10 +95,11 @@ private static ChromeDriverCommandExecutor generateExecutor(
...
     Require.nonNull("Driver clientConfig", clientConfig);
-    Result result = DriverFinder.getPath(service, options);
-    service.setExecutable(result.getDriverPath());
-    if (result.getBrowserPath() != null && !result.getBrowserPath().isEmpty()) {
-      options.setBinary(result.getBrowserPath());
+    DriverFinder finder = new DriverFinder(service, options);
+    service.setExecutable(finder.getDriverPath());
+    if (finder.hasBrowserPath()) {
+      options.setBinary(finder.getBrowserPath());
+      options.setCapability("browserVersion", (Object) null);
```
diemol, line 102: But why do we want to remove the `browserVersion`?

titusfortner: It's necessary in Firefox because Firefox does the compliant thing and fails if it doesn't match (e.g., "stable" will fail). Chrome it isn't technically necessary, but if Selenium Manager has matched the version and found the exact browser to use, sending the driver the version number is unnecessary.

titusfortner: Like, per the spec, Chrome *should fail if the browser versions do not match, but it does not, but I figured the code should handle it in case it became spec compliant.

diemol: Understood. We already do this in the Grid, and it was missing here. Thanks.

diemol (approved):
👍

java/src/org/openqa/selenium/chrome/ChromeDriver.java
titusfortner: It was being done in Driver Finder, this moves it up a level to the place that actually needs it.

diemol (approved):

PR #13386, opened 2024-01-02
[rb] update SOC for driver finder and selenium manager classes

titusfortner:
### Description
* Selenium Manager does not know about Options/Capabilities
* DriverFinder returns browser & driver paths
* Options mutation for binary location happens in `LocalDriver` module
* More reasonable method names

### Motivation and Context
* More straightforward implementation of #13022 
* `DriverFinder` is not labeled private API, so deprecating instead of removing, not the case with `SeleniumManager`


rb/lib/selenium/webdriver/common/selenium_manager.rb
```diff
@@ -77,14 +74,12 @@ def binary
...
+
           begin
-            Platform.assert_file(location)
             Platform.assert_executable(location)
-          rescue TypeError
-            raise Error::WebDriverError,
-                  "Unable to locate or obtain Selenium Manager binary; #{location} is not a valid file object"
           rescue Error::WebDriverError => e
-            raise Error::WebDriverError, "Selenium Manager binary located, but #{e.message}"
+            raise Error::WebDriverError, "#{exe} located, but has problems: #{e.message}"
```
p0deje: Do we have `exe` variable in this context?

titusfortner: Nope, bad copy paste.

p0deje (approved):

titusfortner:
I realized how inefficient some of the code has gotten, and missing a bunch of errors and logging opportunities and had to update all the specs accordingly. At this rate maybe I won't get this feature out for all the bindings before next release. 😂

diemol (approved):

rb/lib/selenium/webdriver/common/selenium_manager.rb
```diff
@@ -34,88 +34,37 @@ def bin_path
...
-                "#{directory}/linux/selenium-manager"
-              end
-            rescue Error::WebDriverError => e
-              raise Error::WebDriverError, "Unable to obtain Selenium Manager binary for #{e.message}"
-            end)
+            if (location = ENV.fetch('SE_MANAGER_PATH', nil))
+              WebDriver.logger.debug("Selenium Manager set by ENV['SE_MANAGER_PATH']: #{location}")
+            else
+              location = platform_location
+              WebDriver.logger.debug("Looking for Selenium Manager at: #{location}")
```
titusfortner: this isn't needed, it will either fail or be found and both of those log an output.

rb/lib/selenium/webdriver/common/driver_finder.rb
```diff
@@ -20,25 +20,55 @@
...
+          path = klass.driver_path
+          path = path.call if path.is_a?(Proc)
+          exe = klass::EXECUTABLE
 
-        path ||= begin
-          SeleniumManager.driver_path(options) unless options.is_a?(Remote::Capabilities)
-        rescue StandardError => e
-          raise Error::NoSuchDriverError, "Unable to obtain #{klass::EXECUTABLE} using Selenium Manager; #{e.message}"
+          if path
+            validate_files(driver_path: path)
```
titusfortner: This needs to raise a NoSuchDriverError as well

titusfortner: Also, stop being lazy with the return value; an extra line won't kill you.

titusfortner:
Sorry, made a bunch more changes, would appreciate another glance over. It should match what I have PR'd for Java & Python

p0deje (approved):

diemol (approved):
Thank you, @titusfortner!

Earlopain:
Hi @titusfortner, you wrote that `DriverFinder` is not private api and should be deprecated. However the `path` method was removed which broke the rails test suite (and probably users system tests). Is that intentional? https://github.com/rails/rails/issues/51658

simi:
:information_source: this broke RubyGems.org build as well https://github.com/rubygems/rubygems.org/actions/runs/8834385973/job/24255908169?pr=4645

PR #13387, opened 2024-01-02
[py] update SOC for driver finder and selenium manager classes

titusfortner:
### Description
* Selenium Manager does not know about Options/Capabilities
* DriverFinder is not static and returns browser & driver paths in separate methods
* Browser classes change options only if applicable
* More reasonable method names

### Motivation and Context
* implementation of #13022
* an alternative to #12554 

### Comments
* Am I doing memoization correctly?
* Am I using private methods correctly?


diemol (approved):

AutomatedTester (approved):
I think this looks good enough. I am not sure we need to have the `Rakefile` stuff as that should all be handled by bazel already

titusfortner:
Thanks, I would have merged this for 4.17 but I wanted to get .NET happy first as well before merging everything else. 😦 

The `Rakefile` stuff includes tasks to make it so I can test and debug locally with PyCharm, but also each binding has its own syntax for testing, which is a pain when I'm trying to remember which are which.

Python is a good example of this.
You think you want:
```
bazel test //py:test-chrome
```
but actually it is:
```
bazel test //py:common-chrome
```
So I create task that does both and outputs results

My plan is to support a wrapper for each language with defaults like:
```
./go <language>:test:unit
./go <language>:test:<browser>
```

diemol (approved):

PR #13392, opened 2024-01-03
[java] set a timeout on joining threads after ending remote server

titusfortner:
### Description
Just adds a timeout value of 30 seconds (no idea what the right value is)

### Motivation and Context
This seems to keep the test from hanging at timing out after 30 minutes and then failing...

Running the tests to verify

### DIFF
--- a/java/src/org/openqa/selenium/os/ExternalProcess.java
+++ b/java/src/org/openqa/selenium/os/ExternalProcess.java
@@ -265,6 +265,9 @@ public class ExternalProcess {
 
     if (exited) {
       worker.join(30000);
+      if (worker.isAlive()) {
+        worker.interrupt();
+      }
     }
 
     return exited;
@@ -295,6 +298,9 @@ public class ExternalProcess {
       try {
         if (process.waitFor(timeout.toMillis(), MILLISECONDS)) {
           worker.join(30000);
+          if (worker.isAlive()) {
+            worker.interrupt();
+          }
           return;
         }
       } catch (InterruptedException ex) {
--- a/java/src/org/openqa/selenium/os/ExternalProcess.java
+++ b/java/src/org/openqa/selenium/os/ExternalProcess.java
@@ -264,7 +264,7 @@ public class ExternalProcess {
     boolean exited = process.waitFor(duration.toMillis(), TimeUnit.MILLISECONDS);
 
     if (exited) {
-      worker.join();
+      worker.join(30000);
     }
 
     return exited;
--- a/java/src/org/openqa/selenium/os/ExternalProcess.java
+++ b/java/src/org/openqa/selenium/os/ExternalProcess.java
@@ -294,7 +294,7 @@ public class ExternalProcess {
 
       try {
         if (process.waitFor(timeout.toMillis(), MILLISECONDS)) {
-          worker.join();
+          worker.join(30000);
           return;
         }
       } catch (InterruptedException ex) {


java/src/org/openqa/selenium/os/ExternalProcess.java
```diff
@@ -264,7 +264,7 @@ public boolean waitFor(Duration duration) throws InterruptedException {
     boolean exited = process.waitFor(duration.toMillis(), TimeUnit.MILLISECONDS);
 
     if (exited) {
-      worker.join();
+      worker.join(30000);
```
joerg1985, line 267: How about `worker.interrupt()` after the join, this might stop the hanging thread.

joerg1985 (changes requested):

java/src/org/openqa/selenium/os/ExternalProcess.java
titusfortner: If it hangs on join() will it ever get to interrupt()?

titusfortner:
Welp, I thought this worked locally, but definitely not working on CI 😦

java/src/org/openqa/selenium/os/ExternalProcess.java
joerg1985: The timeout should be added to the join call too.
```
worker.join(30000);
worker.interrupt();
```

joerg1985:
> Welp, I thought this worked locally, but definitely not working on CI 😦

There is another join call below in the same method, are you sure the timeout was added to the correct one? This might be different on the CI due to the if statement.

java/src/org/openqa/selenium/os/ExternalProcess.java
titusfortner: is 30 seconds the right timeout?

titusfortner:
@joerg1985 I've exhausted my abilities on this one. Do you have time to figure out the fix?

joerg1985:
@titusfortner do you have a description about the root problem and how to reproduce it?

joerg1985:
I had a look at the CI logs, looks like agumenting the driver fails, this might lead to a half initialized state.

Add a try/catch block in the RemoteSupplier (line 60 to 62) stopping the server in this case could help. Will try this later on and check if this is true.

joerg1985:
@titusfortner 
The `worker` is now a daemon to avoid issues when a process is not terminated, but this seems not to be the fix for this.
I will have to debug into this to see what is going wrong.

titusfortner:
It's windows specific, so it's 100% reproducible starting remote grid in our tests on windows.

titusfortner:
But no, I don't know how the thread joining works at all, I just know that it hangs in that particular line of code. For 30 minutes (and I have no idea where the 30 minutes is coming from, either)

joerg1985:
@titusfortner i had a deeper look at the code:
1. `supportsNormalTermination` in windows is false (at least until JDK 17), so it runs into the lower code
2. the worker does not terminate even if `process.destroyForcibly` is called in windows, for what ever reason.
3. there was a missing `process.waitFor` when `process.destroyForcibly` was called, i missed this when writing it
4. using `worker.join(8000);` did fix this for me, so i don't know what the problem was with your branch
5. i added a `worker.interrupt()` in any case to ensure the worker fails internally when not already stopped

titusfortner:
@joerg1985 
Thanks! So, it's passing the CI test that's been failing because of this, but now `DevToolsReuseTest` is failing. 😁

joerg1985:
@titusfortner i will check this tomorrow

joerg1985:
@titusfortner in this run it passed, so i think the change is okay: https://github.com/SeleniumHQ/selenium/actions/runs/7491544532/job/20393002729#step:15:1795

PR #13402, opened 2024-01-06
[py] add deprecation library requirement

titusfortner:
### Description
Add a dependency on deprecation library

### Motivation and Context
Current means of doing deprecations in Python is challenging.
Using an annotation would be much easier

### Notes
* This may not be the right fix
* There were two libraries, this one seems to have fewer features, but its also the one being maintained
* I think I saw somewhere someone  rolling our own annotation, but I'm not sure where, but I seem to remember it feeling clunky?

What do y'all think?


diemol (approved):
This should be added. Deprecating something in Python should not be as complicated as it is currently.

titusfortner:
@isaulv @AutomatedTester @symonk 
Do either of you remember seeing a custom implementation? I thought @symonk was doing something, but I can't find it; maybe @sandeepsuryaprasad did something at one point?
Anyway, would like to know if there's an easy win here, and what it is.

sandeepsuryaprasad:
@titusfortner  @symonk @isaulv @AutomatedTester This decorator is implemented in Python 3.13. PEP 702  (https://peps.python.org/pep-0702/). So going forward once python 3.13 is released, we can avoid the dependency on deprecation library since the decorator is already implemented in `warnings` module https://peps.python.org/pep-0702/#example

titusfortner:
@sandeepsuryaprasad We have to continue supporting users on 3.12 and below until ~2028.
Is there a way to back port that functionality to earlier versions of Python?

sandeepsuryaprasad:
@titusfortner yes it is. the new decorator will be added to `typing_extensions` https://peps.python.org/pep-0702/#backwards-compatibility

titusfortner:
@sandeepsuryaprasad — but this still won't be available until October?

sandeepsuryaprasad:
@titusfortner  Yes.. you are right.. this feature won't be available until October-24.

titusfortner:
There were two packages I saw:
* https://pypi.org/project/Deprecated/ — this PR; very basic and maintained
* https://pypi.org/project/deprecation/ — lots of cool bells and whistles but not actively maintained

It looks like the official deprecation implementation in 4.13 is going to be very basic as well, no bells, no whistles. In fact, the syntax looks exactly the same as the Deprecated package syntax.

@AutomatedTester what do you think about using this package, adding the annotations as necessary and then just removing this library and replacing it with the standard one in October?

AutomatedTester:
I think this should be fine to add. I am always trying to limit third party but if we feel it will add value (I don't think there is much in terms of deprecation at the moment)

titusfortner:
Well, there are a few Firefox things that @diemol didn't deprecate because it was complicated. 😂 

And I was going to use it to deprecate some Selenium Manager things, but we've decided we can just remove instead of deprecate Selenium Manager things while it is still in beta.

titusfortner:
Hmm, it looks like we might be able to use this already? `from typing_extensions import deprecated`??

titusfortner:
And I verified this works. Nice.

PR #13407, opened 2024-01-08
Node WebSocket not working with sub-path option

### LINKED ISSUE #13393: [🐛 Bug]: WebSocket Path Not Working with Selenium Grid --sub-path Option
### What happened?

**Description:**
After setting up Selenium Grid with the `--sub-path=/selenium-grid` option, I encountered an issue with the WebSocket path. Previously, I was able to access the WebSocket using the path `ws://ip:port/sessions/{ID}/se/vnc`. However, after adding the `--sub-path` option, the expected WebSocket path `ws://ip:port/selenium-grid/sessions/{ID}/se/vnc` does not work.

**Steps to Reproduce:**
1. Start Selenium Grid with the `--sub-path=/selenium-grid` option.
2. Attempt to access the WebSocket at `ws://ip:port/selenium-grid/sessions/{ID}/se/vnc`.

**Expected Behavior:**
The WebSocket should be accessible at the path `ws://ip:port/selenium-grid/sessions/{ID}/se/vnc`, reflecting the `--sub-path` configuration.

**Actual Behavior:**
The WebSocket path `ws://ip:port/selenium-grid/sessions/{ID}/se/vnc` is not functional, and there seems to be no alternative path that accommodates the `--sub-path` configuration.

**Additional Information:**
- This issue seems to specifically relate to the interaction between the WebSocket and the `--sub-path` option in Selenium Grid.
- Any guidance or workaround to address this issue would be greatly appreciated.

If there is no sub-path option available, could you please inform me of any other methods to set a sub-path specifically for WebSocket connections? I would greatly appreciate your guidance on this.

Thank you for your assistance in resolving this matter.

```shell
1. Configured and launched selenium-hub in Docker with `SE_SUB_PATH=/selenium/`.
2. Set up the Selenium Grid and confirmed manager UI accessibility at `http://ip:port/selenium/#ui`.
3. Ran a JUnit Test with RemoteWebDriver and confirmed the test was executing. For ease of testing, I included a Thread.sleep(10000); statement to allow sufficient time for observation.
4. Attempted to monitor the test using VNC:
   - The VNC request from the manager interface connects to an URL without the `/selenium` prefix.
   - Direct VNC connection to `ws://ip:port/session/{ID}/se/vnc` is successful.
   - However, attempting a VNC connection to `ws://ip:port/selenium/session/{ID}/se/vnc` fails to establish a connection.
```

### Relevant log output

```shell
Not Important
```

### Operating System

WSL2 with selenium docker

### Selenium version

Java 4.16.1

Chrome 119

Not Important

4.16.1

VietND96:

### Description
[java] Update subPath for ProxyNodeWebsockets

### Motivation and Context
Fix issue #13393 - WebSocket at `ws://ip:port/selenium-grid/sessions/{ID}/se/vnc` with `selenium-grid` is the `sub-path`

---
Some info on the fix
- In `NodeOptions` added method `getSubPath()` to extract the subPath from config `grid-url` / `hub` (in `[node]` section)
- In `ProxyNodeWebsockets`, subPath is removed from the request URI before matching UrlTemplate

### Types of changes

### Checklist


java/src/org/openqa/selenium/grid/node/config/NodeFlags.java
```diff
@@ -261,6 +262,13 @@ public class NodeFlags implements HasRoles {
   @ConfigValue(section = NODE_SECTION, name = "enable-managed-downloads", example = "false")
   public Boolean managedDownloadsEnabled;
 
+  @Parameter(
+      names = {"--sub-path"},
+      arity = 1,
+      description = "A sub-path that should be considered for all user facing routes on the Node")
+  @ConfigValue(section = NETWORK_SECTION, name = "sub-path", example = "my_company/selenium_grid")
+  public String subPath;
```
krmahadevan: @vietnd96 - Can you please let me know what would the application behaviour be in the below scenario? 

1. Hub was started with the sub path as "/foo"
2. Node1 is now being started with the sub-path as "/bar"
3. Node2 is now being started with the sub-path as "foobar"

So does this mean that a testcase would need to know the sub-path using which the node (to which the session was routed to) was spun off, in order for it to be able to access the web sockets?

diemol (changes requested):
The Node should not need changes, the Router/Hub should be in charge of forwarding the WebSocket message to the Node and that is it. The client does not have direct access to Nodes, which is why we should not change the Nodes at all.

VietND96:
> The Node should not need changes, the Router/Hub should be in charge of forwarding the WebSocket message to the Node and that is it. The client does not have direct access to Nodes, which is why we should not change the Nodes at all.

@diemol Yes, the client does not have direct access to Nodes. The change is only for Router/Hub forwarding to `ProxyNodeWebsockets`.
The Router/Hub forwarding request e.g `/subPath/session/{id}/se/vnc`. In `ProxyNodeWebsockets` there are some UrlTemplate matching to open particular Ws, and the problem is there. When matching e.g req `/subPath/session/{id}/se/vnc` vs template `/session/{id}/se/vnc` it returned not match, which caused the Ws not open correctly.
In NodeOptions, at least `grid-url` or `hub` is required always for Node register to Router/Hub, user can pass the Grid URL within subPath from the config. In Node, only there is `getGridSubPath()` to extract the subPath from Router/Hub URL and set it to `ProxyNodeWebsockets` for matching UrlTemplate as well as construct the URLs for se:vnc, cdp, bidi, fwd in session capabilities, which is used to establish Ws
I have renamed the method to `getGridSubPath`, hope that avoid the confusion that we are trying to allow users to set subPath for Nodes to access directly from outside.

diemol (approved):

diemol:
> > The Node should not need changes, the Router/Hub should be in charge of forwarding the WebSocket message to the Node and that is it. The client does not have direct access to Nodes, which is why we should not change the Nodes at all.
> 
> @diemol Yes, the client does not have direct access to Nodes. The change is only for Router/Hub forwarding to `ProxyNodeWebsockets`. The Router/Hub forwarding request e.g `/subPath/session/{id}/se/vnc`. In `ProxyNodeWebsockets` there are some UrlTemplate matching to open particular Ws, and the problem is there. When matching e.g req `/subPath/session/{id}/se/vnc` vs template `/session/{id}/se/vnc` it returned not match, which caused the Ws not open correctly. In NodeOptions, at least `grid-url` or `hub` is required always for Node register to Router/Hub, user can pass the Grid URL within subPath from the config. In Node, only there is `getGridSubPath()` to extract the subPath from Router/Hub URL and set it to `ProxyNodeWebsockets` for matching UrlTemplate as well as construct the URLs for se:vnc, cdp, bidi, fwd in session capabilities, which is used to establish Ws I have renamed the method to `getGridSubPath`, hope that avoid the confusion that we are trying to allow users to set subPath for Nodes to access directly from outside.

Understood, thank you. It makes sense.

Issue #9668, opened 2021-07-21
Webelements inside the vue.js's fragment are not visible to Selenium webdriver 


spbrantan:
## 🐛 Bug Report

Elements with attribute 'fragment' are not visible.
Impossible to interact with child elements.

Root Cause
vue.js uses a concept called Fragments to create more than one root node in a Vue component. This is done for Accessibility support which is necessary to allow assistive technology like screen readers to interpret web pages and applications, More about this in -> https://blog.logrocket.com/fragments-in-vue-js/

So, basically they are creating HTML tag that would not read as a node by the DOM and called it fragments.

Hence the webelements inside such fragments are not visible for Webdriver. This is more or less like a iframe inside html page where we do switchToIFrame and interact with it. Only that for Fragments there are no ways to make it visible to the webdriver.

PFA Screenshot of DOM snapshot containing the Fragment.

![image](https://user-images.githubusercontent.com/55972871/126429657-939efdf3-3ae7-4948-8d30-f5191b6b83a2.png)

## To Reproduce

https://jsfiddle.net/brantansp/bwdkx1tg/3/#&togetherjs=eM6k1p6Iaz

Detailed steps to reproduce the behavior:

System.out.println(webDriver.findElement(By.xpath("((//div[@class='nested-list']/div)[1]//div)[4]")).getText()); //returns null

## Expected behavior

System.out.println(webDriver.findElement(By.xpath("((//div[@class='nested-list']/div)[1]//div)[4]")).getText()); //should return value

## Test script or set of commands reproducing this issue

WebDriverManager.chromedriver().setup();
WebDriver webDriver = new ChromeDriver();
System.out.println(webDriver.findElement(By.xpath("((//div[@class='nested-list']/div)[1]//div)[4]")).getText());

## Environment

OS: Windows 10 x64
Browser: Chrome
Browser version: 89
Browser Driver version: ChromeDriver 89
Language Bindings version: Java 3.141.59

sethidden:
The article you posted is using `vue-fragment`, but I'm getting the same issue using the `vue-frag` library.

Reading `vue-frag`'s documentation I saw the below excerpt:

> #How does this work?

[...]

PR #13430, opened 2024-01-11
[atoms] use css locators in dom.js

joerg1985:
### Description
Using xpaths will need to transfer the complete `wicked-good-xpath`, this can easily be replaced by a css locator.

### Motivation and Context
This will reduce the size of the `isDisplayed.js` from ~43kb to ~17kb.

### Types of changes

### Checklist

### DIFF
--- a/javascript/atoms/BUILD.bazel
+++ b/javascript/atoms/BUILD.bazel
@@ -119,10 +119,10 @@ closure_js_library(
     deps = [
         ":bot",
         ":color",
+        ":css",
         ":domcore",
         ":json",
         ":useragent",
-        ":xpath",
         "@io_bazel_rules_closure//closure/library",
     ],
 )
@@ -239,7 +239,7 @@ closure_js_library(
     name = "locators",
     srcs = glob(
         ["locators/*.js"],
-        exclude = ["locators/xpath.js"],
+        exclude = ["locators/css.js"],
     ),
     suppress = [
         "JSC_IMPLICITLY_NULLABLE_JSDOC",
@@ -250,11 +250,12 @@ closure_js_library(
     ],
     deps = [
         ":bot",
+        ":css",
         ":dom",
         ":errors",
         ":json",
         ":useragent",
-        ":xpath",
+        "//third_party/js/wgxpath",
         "@io_bazel_rules_closure//closure/library",
     ],
 )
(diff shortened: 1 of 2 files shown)


titusfortner:
@twalpole I know you did some work on a separate isDisplayed atom for Capybara. Did you do something similar here?

twalpole:
@titusfortner I just went through all the code and removed all the stuff that was supporting legacy browsers - which took it down to 4.7k - [lib/capybara/selenium/atoms/isDisplayed.min.js](https://github.com/teamcapybara/capybara/blob/master/lib/capybara/selenium/atoms/isDisplayed.min.js) - unminified 16.1k https://github.com/teamcapybara/capybara/blob/master/lib/capybara/selenium/atoms/src/isDisplayed.js.  I did not specifically swap from xpath to css because of the added scoping functionality xpath provides, but it likely did involve getting rid of wicked-good-xpath since every modern browser supports it natively without needing a JS polyfill.

javascript/atoms/BUILD.bazel
```diff
@@ -250,11 +250,12 @@ closure_js_library(
     ],
     deps = [
         ":bot",
+        ":css",
         ":dom",
         ":errors",
         ":json",
         ":useragent",
-        ":xpath",
+        "//third_party/js/wgxpath",
```
titusfortner, line 258: So, do we need to depend on wicked good xpath here?

titusfortner:
We've also discussed [moving from closure to Typescript](https://github.com/SeleniumHQ/selenium/issues/9668#issuecomment-1072652972) for these.

javascript/atoms/BUILD.bazel
joerg1985: Yes, this is still needed here for legacy browsers to have XPath locator support.

This PR does only remove the need for wicked good xpath for is-displayed atom.

joerg1985:
> We've also discussed [moving from closure to Typescript](https://github.com/SeleniumHQ/selenium/issues/9668#issuecomment-1072652972) for these.

Is it possible to run TypeScript in IE 11 and old Android versions?

javascript/atoms/BUILD.bazel
titusfortner: Which legacy browsers require it? IE 11?

titusfortner:
@joerg1985

> Is it possible to run TypeScript in IE 11 and old Android versions

Wouldn't we transpile and minify it the same (ish) way closure does?

diemol (approved):
Thank you, @joerg1985!

PR #13444, opened 2024-01-15
Refactor argument creation in DriverService classes

manuelsblanco:
### Description
The argument creation process within EdgeDriverService, ChromeDriverService, and DriverService classes has been refactored to improve code readability and maintainability. Common argument creation code was abstracted into a new method named `createCommonArgs` in the base DriverService class. This eliminates code duplication across different classes and increases maintainability.

### Motivation and Context
The motivation behind this change was to enhance the readability and maintainability of the codebase by removing duplicated argument creation logic. The new method, `createCommonArgs`, centralizes common code, making it easier to manage and reducing the chances of inconsistencies.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/chrome/ChromeDriverService.java
+++ b/java/src/org/openqa/selenium/chrome/ChromeDriverService.java
@@ -282,32 +282,7 @@ public class ChromeDriverService extends DriverService {
     @Override
     protected List<String> createArgs() {
       List<String> args = new ArrayList<>();
-      args.add(String.format("--port=%d", getPort()));
-
-      // Readable timestamp and append logs only work if log path is specified in args
-      // Cannot use logOutput because goog:loggingPrefs requires --log-path get sent
-      if (getLogFile() != null) {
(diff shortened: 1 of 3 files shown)


pujagani (changes requested):
I appreciate the motivation for the changes. However, the createArgs() implementation is different for different browsers. For example, GeckoDriverService used for Firefox has a different set of args and hence the createArgs() implementation is different. Chromium based browsers (Edge and Chrome) have the same args and hence the same method implementation. The changes made here need to be in sync with this.

manuelsblanco:
@pujagani Thank you, I'm going to keep this in mind to improve que code of the PR.

titusfortner:
The only way to properly remove this duplication would be to create a ChromiumDriverService that extends DriverService and is the superclass for Chrome & Edge classes. But when I started down the path of creating this last year it got complicated (can't remember why) and I decided it was more trouble than it was worth. 

If you want to figure out how to get that working, but it might not be simple. In the mean time I'm going to close this PR. Thanks.

PR #13450, opened 2024-01-17
Refactor switch statements to if-else in various classes

manuelsblanco:
The switch statements in several classes such as WindowProxyProperties, Connection, CompletionCommand etc., have been refactored to if-else statements for better readability and simplicity. This also includes the handling of IllegalArgumentExceptions in the cases where the given input does not match the expected cases.

### Description
This PR contains changes to several classes, such as WindowProxyProperties, Connection, and CompletionCommand. The switch statements in these classes have been refactored to if-else statements for improved readability and simplicity. This refactor also improves the handling of IllegalArgumentExceptions.
### Motivation and Context
Simplifying control structures like switch statements into if-else statements can make the code more readable, particularly when there are numerous or complex cases to handle. Additionally, in instances where the provided input does not match the anticipated cases, it is now being managed by throwing an IllegalArgumentException.

### Types of changes

### Checklist

Huge thanks for your time reviewing this PR and I am open to any suggestions or changes required.


java/src/org/openqa/selenium/bidi/script/WindowProxyProperties.java
```diff
@@ -31,15 +31,11 @@ public static WindowProxyProperties fromJson(JsonInput input) {
...
-      switch (input.nextName()) {
-        case "context":
-          browsingContext = input.read(String.class);
-          break;
-
-        default:
-          input.skipValue();
-          break;
-      }
+        if (input.nextName().equals("context")) {
```
diemol, line 34: Please switch them to avoid a NPE. 

```java
        if ("context".equals(input.nextName())) {
```

diemol (changes requested):
Can you please run the format script?

@pujagani do you think this is ok?

pujagani (changes requested):
While I understand the reason for the change here. I find it more readable with "switch case" for cases that are implementing `fromJson()` method. The reason being checks are easy with switch cases and the Selenium Java code base, all implementation of`fromJson()` methods follow the same pattern. So it is just easy to replicate and not worry about adding additional checks or NPE.  
Example, in `WindowProxyProperties. ` `WindowProxyProperties` did not initially have the context id, it was added to support iframes and get its browsing context it.  BiDi also tends to add more properties to the class since it is work in progress, so keeping the switch statement would be better in the long-term.
We are going to be deserializing a lot of classes in future for BiDi too. This is my opinion, but its not set in stone. What do you think @diemol ?
 For cases, where `IllegalArgumentException` needs to be thrown, the changes are fine.

PR #13452, opened 2024-01-17
[build] properly get selenium manager from bazel output

titusfortner:
Looks like .NET was hard coded to using the binaries in the repo even when we built them.

This is an attempt to get the binaries from the target output.


dotnet/src/webdriver/WebDriver.csproj.prebuild.sh
```diff
@@ -1,4 +1,9 @@
 #!/bin/bash
+echo "Building Selenium Manager binaries"
+bazel build //dotnet/src/webdriver:manager-linux
+bazel build //dotnet/src/webdriver:manager-windows
+bazel build //dotnet/src/webdriver:manager-macos
+
```
titusfortner: Should these get put in conditionals like the ones below? If they are already built, bazel isn't going to rebuild them?

dotnet/src/webdriver/WebDriver.csproj.prebuild.sh
nvborisenko, line 1: Please do the same for window based building process: `WebDriver.csproj.prebuild.cmd` file.

nvborisenko (changes requested):

dotnet/src/webdriver/WebDriver.csproj.prebuild.sh
nvborisenko: In my opinion your approach is better. Bazel is smart enough to understand where the target's output is cached. I don't mind to invoke bazel build again and again given that it takes `~seconds`.

nvborisenko: 7 seconds for invoking all dependent bazel targets. Pretty good for now, at least we can even improve it (when it will become annoying) via caching input files for msbuild targets.

dotnet/src/webdriver/WebDriver.csproj.prebuild.sh
nvborisenko: Resolved.

.github/workflows/ci-dotnet.yml
```diff
@@ -23,4 +23,4 @@ jobs:
       java-version: 17
       os: windows
       run: |
-        bazel test //dotnet/test/common:ElementFindingTest-firefox //dotnet/test/common:ElementFindingTest-chrome --pin_browsers=true
+        bazel test //dotnet/test/common:ElementFindingTest-firefox //dotnet/test/common:ElementFindingTest-chrome
```
titusfortner: Tests pass when I leave `--pin_browsers=true` in here, but it shouldn't be necessary.

nvborisenko (approved):

PR #13454, opened 2024-01-17
[rb] Add `base64` gem as a runtime dependency

### LINKED ISSUE #13447: [🐛 Bug]: [rb] Either declare or remove dependency on `base64`
### What happened?

Ruby 3.3 shows a warning if the base64 gem is required without it being declared in the gemspec, Ruby 3.4 will throw an error.
 See https://github.com/ruby/ruby/blob/master/doc/NEWS/NEWS-3.3.0.md#stdlib-updates

Many gems have opted to inline the base64 implementation, however this gem makes use of the urlsafe variants which are a bit more complex and not just a wrapper over `pack`/`unpack`. You would need to inline the entirety of https://github.com/ruby/base64/blob/9669a7d3b0e3b9a739969404daf58f912c58c6b3/lib/base64.rb, similar to https://github.com/newrelic/newrelic-ruby-agent/pull/2378

```shell
Use `selenium-webdriver` on Ruby 3.3
```

### Relevant log output

```shell
base64 is not part of the default gems since Ruby 3.4.0. Add it to your Gemfile.
```

### Operating System

Irrelevant

### Selenium version

Ruby master

Irrelevant

Irrelevant

_No response_

Earlopain:
### Description

Closes #13447

Ruby 3.3 shows a warning if requiring base64 without specifying it in the gemspec
Ruby 3.4 will throw a warning.
See https://github.com/ruby/ruby/blob/master/doc/NEWS/NEWS-3.3.0.md#stdlib-updates

There is no version constraint on base64. This follows what other projects have done:
* https://github.com/rails/rails/pull/48907
* ~~https://github.com/rubocop/rubocop/pull/12094~~ Nevermind this one, I misremembered
* https://github.com/newrelic/newrelic-ruby-agent/pull/2238

There may be some issue with this in combination with older bundler versions. But since this gem requires Ruby >= 3.0 I don't think this is a concern.
* https://github.com/rubocop/rubocop/issues/12310
* https://github.com/newrelic/newrelic-ruby-agent/issues/2312

### Types of changes

### Checklist


rb/Gemfile.lock
```diff
@@ -4,6 +4,7 @@ PATH
     selenium-devtools (0.120.0)
       selenium-webdriver (~> 4.2)
     selenium-webdriver (4.17.0.nightly)
+      base64
```
titusfortner:
```suggestion
      base64 (~> 0.2)
```

titusfortner:
Excellent. Can you add the version and re-run bundler for the lockfile? Thanks.

Earlopain:
Yeah, sure. Done

titusfortner:
@Earlopain Thanks for your help!

PR #13464, opened 2024-01-19
[dotnet] Improve logging performance when it is disabled

### LINKED ISSUE #13411: [🚀 Feature]: [dotnet] Improve performance of internal logging
### Feature and motivation

Before:

```csharp
logger.Trace("abc");
```

The problem is `abc` is always allocated in memory despice that logging is turned off.

What we can do:
```csharp
if (logger.IsEnabled(LogEventLabel.Trace))
{
  logger.Trace("abc");
}
```

Interesting what else we can do to improve logging..

CC: @YevgeniyShunevych 

### Usage example

Purely internal stuff.

nvborisenko:
### Description
It adds `IsEnabled` method to check there is handler and event can be emitted. Now internally we should use the following pattern to emit a message:
```csharp
if (logger.IsEnabled(LogEventLevel.Trace))
{
  logger.Trace("my message");
}
```

### Motivation and Context
Fixes https://github.com/SeleniumHQ/selenium/issues/13411

### Benchmark

Code:
```csharp
[Benchmark]
public void EmitLogOld()
{
    var message = new string('a', 1000);

    logger.Trace(message);
}

[Benchmark]
public void EmitLogNew()
{
    if (logger.IsEnabled(LogEventLevel.Trace))
    {
        var message = new string('a', 1000);

        logger.Trace(message);
    }
}
```

Results:
```
| Method     | Mean       | Error     | StdDev    | Gen0   | Allocated |
|----------- |-----------:|----------:|----------:|-------:|----------:|
| EmitLogOld | 213.412 ns | 4.0365 ns | 3.7757 ns | 0.3226 |    2024 B |
| EmitLogNew |   9.917 ns | 0.2251 ns | 0.3369 ns |      - |         - |
```

I also tried with `MethodImpl` attribute with aggressive inlining, got `mean = 9.1 ns`. Decided to not use it, keeping code simple.

### Types of changes

### Checklist


dotnet/src/webdriver/Internal/Logging/LogContext.cs
```diff
@@ -96,9 +96,14 @@ public ILogger GetLogger(Type type)
             return _loggers.GetOrAdd(type, _ => new Logger(type, _level));
         }
 
+        public bool IsEnabled(ILogger logger, LogEventLevel level)
+        {
+            return Handlers != null && level >= _level && level >= GetLogger(logger.Issuer).Level;
```
YevgeniyShunevych: Doesn't `GetLogger(logger.Issuer).Level` equal to `logger.Level` here? Possible simplification.

nvborisenko: Fixed.

YevgeniyShunevych (approved):

PR #13466, opened 2024-01-19
[build] create workflow to PR changes to browser versions

titusfortner:
This is set to run daily and create a PR with any changes.

The PR will look like this: https://github.com/titusfortner/selenium/pull/54

(I'm not sure why it was posted as me and not the CI bot, but it was posted by this action: https://github.com/titusfortner/selenium/actions/runs/7585149002)

I did this with patch files in case we want to use this approach for other things in the future.

### DIFF
--- a/.github/workflows/pin-browsers.yml
+++ b/.github/workflows/pin-browsers.yml
@@ -1,0 +1,46 @@
+name: Pin Browsers
+on:
+  schedule:
+    - cron: 10 0 * * *
+  workflow_dispatch:
+
+jobs:
+  update:
+    name: Update Pinned Browsers
+    uses: ./.github/workflows/bazel.yml
+    with:
+      name: Pin Browsers
+      cache-key: pin-browsers
+      run: bazel run //scripts:pinned_browsers
+
+  pull-request:
+    runs-on: ubuntu-latest
+    needs: update
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/pin-browsers.yml
```diff
@@ -0,0 +1,46 @@
...
+      name: Pin Browsers
+      cache-key: pin-browsers
+      run: bazel run //scripts:pinned_browsers
+
+  pull-request:
+    runs-on: ubuntu-latest
+    needs: update
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
```
p0deje: Please use v4.

.github/workflows/pin-browsers.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Download patch
+        uses: actions/download-artifact@v2
```
p0deje: Please use v4.

p0deje (approved):
This is really cool!

titusfortner:
@p0deje you gave me the idea yesterday 😁 

And the old versions is because I was copy/pasting from chat gpt 😂 easy fix at least

diemol (approved):
Thank you! ❤️

Issue #11587, opened 2023-01-23
[🚀 Feature]: Deprecate Firefox Profile


titusfortner:
### Feature and motivation

I created a huge PR that removed a bunch of outdated Firefox Profile functionality (https://github.com/SeleniumHQ/selenium/pull/11118), but we agreed at a TLC meeting, that even if the functionality no longer works, we should go through and deprecate things.

* Several bindings still support functionality from Legacy Firefox Driver, which doesn't work at all.
* Adding extensions no longer works
* Adding preferences works, but that should be done via Options class instead

Honestly, we can probably get away with deprecating the whole class and having the FirefoxOptions class method/property for `setProfile` just take the path of the directory to zip and use. We would need to test to make sure you can use a directory and add preferences to it via options, or if we need the Firefox profile for that.

### Usage example

n/a

willdurand:
> Adding extensions no longer works

Hey, do you know why extensions no longer work?

(Disclaimer: I work on add-ons at Mozilla)

titusfortner:
@willdurand hi, thanks.

Firefox Profile originates from when Selenium managed the driver before Firefox 47 and geckodriver, so we needed to populate everything. Doing that with Profile class made sense.

Right now, the accepted way to add these to Firefox is with a method to geckodriver after the session is started - https://www.selenium.dev/documentation/webdriver/browsers/firefox/#add-ons

The Profile class generates a default profile, adds things to the places they need to be, then sends it to geckodriver for use with Firefox when it starts. Looking at a "real" profile and comparing where the Profile class was putting extensions, didn't match, and I didn't see how to put extensions the way they need to be used in the place that Firefox needs to see them. If you know where to put extensions (in what format) in what location for Firefox to use it, we can do something with that. Ideally, though, geckodriver would manage that by allowing users to put it in options  directly like Chrome - https://github.com/mozilla/geckodriver/issues/1476

If you want to discuss more, please find us on irc/matrix/slack - https://www.selenium.dev/support/#ChatRoom
Thanks!

iampopovich:
@titusfortner hi 👋 is this issue still actual? Should it be closed? 
It's not my first mention of someone from the dev team, sorry. In search of tasks I could help with, I found some obsolete tasks like this one. I guess some tasks in the issues backlog are not relevant now.

PR #13477, opened 2024-01-21
[py] Firefox Profile Fixes and Deprecations

titusfortner:
### Description
Currently when a user does this:
```py
    options = webdriver.FirefoxOptions()
    options.profile = "/path/to/profile"
```
Selenium is passing that path to FirefoxProfile class. But FirefoxProfile constructor adds a bunch of default values instead of just zipping it and using it. I was going to just put the zipping method in the options class for when the profile is set as a string, but decided it didn't make sense to automatically update any existing profile.
User can still add preferences to the profile and have them update

Also finally added the deprecations to this

### Motivation and Context
* Python portion of #11587 
* Might be part of what is going on in #11028


diemol (approved):

titusfortner:
Intermittent failure of one of the tests because the timestamp can be different for the user JS file between executions. I'll fix.

PR #13492, opened 2024-01-23
Add type to parameter of webelement.WebElement.send_keys

adamtheturtle:
### Description

I have set the type of `*value` to `AnyKey`.
I chose `AnyKey` rather than `str`, despite the method docstring saying that value is "a string", because the method uses `"".join(map(str, value))`, which tells me that maybe a `value` element may not be a string.

### Motivation and Context

Without this, `pyright` in strict mode gives me a `reportUnknownMemberType` error.

### Types of changes

### Checklist


diemol (changes requested):
The Python folks in the project tell me that using `str` is more suitable, if you want to add this.

adamtheturtle:
@diemol - sure, I made the change.

diemol (approved):
Thank you, @adamtheturtle!

Issue #13097, opened 2023-11-04
[🚀 Feature]: Replace `Newtonsoft.Json` package with `System.Text.Json`


nvborisenko:
### Feature and motivation

We use `Newtonsoft.Json` package to serialize/deserialize json. It is great library. But seems json is becoming built-in format in BCL (.net 6). In future it allow us to be not dependent on any 3rd-party libraries.

### Usage example

It is entirely internal stuff.

We will produce multi-targeted package:
- `netstandard2.0` - depends on `System.Text.Json`
- `net6.0` - has no dependencies

### Actions:
- [ ] Migrate `WebDriver` project without CDP generation
- [ ] Change generation of CDP models
- [ ] Migrate unit tests
- [ ] Introduce new target framework `net6.0` without dependencies

titusfortner:
But we can't do this until we drop support for everything below .NET 6?

jimevans:
@titusfortner Not _entirely_ accurate. We could still target .NET Standard 2.0, but that target would just have to have a dependency on the System.Text.Json assembly. Same for .NET Standard 2.1 (to handle .NET Core 2 and 3). But if we target those two platforms, plus .NET 6, 7, and 8, the latter three would not require dependencies.

NCLnclNCL:
Great

NCLnclNCL:
need do, it will reduce size for native aot

titusfortner:
So the idea is to add compile conditionals to use `System.Text.Json` in some places and `Newtonsoft.Json` in others? That seems like a lot more of a maintenance burden than just having a 3rd party dependency?

NCLnclNCL:
> So the idea is to add compile conditionals to use `System.Text.Json` in some places and `Newtonsoft.Json` in others? That seems like a lot more of a maintenance burden than just having a 3rd party dependency?

No, System.Text.Json is supported in .net standard 2.0(> .net 4.6.2), it can be installed via nuget, currently selenium only supports .net standard 2.0 and above

nvborisenko:
> need do, it will reduce size for native aot

I don't think this story is about size reduction for native AOT. I am expecting selenium package will be targeted to `net6.0` without dependencies. Size reduction via assembly trimming would be next step (if `bazel` allows).

NCLnclNCL:
> > need do, it will reduce size for native aot
> 
[...]

PR #13498, opened 2024-01-24
Use PrivateAssets=compile to avoid exposing Newtonsoft.json to downstream consumers

kmgallahan:
### Description

Use PrivateAssets=compile to avoid exposing Newtonsoft.json to downstream consumers. 

### Motivation and Context

Currently, the Newtonsoft.json public API surface is exposed to all consumers whether they want to use it or not. With this change consumers will only see the API as being available if it is referenced on purpose.

### Types of changes

**Only a breaking change in that consuming projects would need to add a direct reference to Newtonsoft.json if they want to use it and it isn't referenced elsewhere.**

### Checklist

### DIFF
--- a/dotnet/src/webdriver/WebDriver.csproj
+++ b/dotnet/src/webdriver/WebDriver.csproj
@@ -52,7 +52,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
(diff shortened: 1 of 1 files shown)


nvborisenko:
Thanks for your contribution.

Selenium assembly depends on `Newtonsoft.Json` assembly, the same for nuget package. If we remove this dependency, then users will face exceptions in run-time (till they install `Newtonsoft.Json` package directly).

https://github.com/SeleniumHQ/selenium/issues/13097 should minimize dependency tree.

PR #13512, opened 2024-01-25
[build] move copyright from rake task to a bazel target

titusfortner:
#### Status
~Rebased off latest, and relies on #13514 now to show off that it can auto correct copyright as well.
So wait for 13514 to merge before merging this.~

What #13514 looks like needs to be discussed more. This PR is useful either way, and the success of the experiment shows that it works.

### Description
* rewrite ruby code in Python
* Add target in scripts to execute it with bazel
* remove code from rake file and rake tasks
* run target as part of format_code script

### Motivation and Context
* I added copyright check to build as part of release, but it doesn't really belong there
* Copyright should get run on every RBE execution as part of the linter to make sure each commit is correct
* ~This is the first step~ Ok, it's 2 lines to add it to the linter so I did it with this PR

You can see from my little experiment here - https://github.com/SeleniumHQ/selenium/pull/13512/commits/fdf14afa5984addcd71cb52e5a0e737fa2f00e5e
that it found all 4 issues: https://github.com/SeleniumHQ/selenium/actions/runs/7660852315/job/20879036168?pr=13512


titusfortner:
Hmm, I'm not sure why the final commit doesn't have any checks?

diemol (approved):

PR #13513, opened 2024-01-25
[ci] Create pre-release workflow

titusfortner:
### Description

This CI job can only be run manually; has a text input for version number and a checkbox for whether it uses Beta channel (beta on the "Early Stable Release" day has always been the same build number as the Stable Release several days later).
workflow_dispatch

Running this on current code with the current Beta CDP (122)
looks like this: https://github.com/titusfortner/selenium/actions/runs/7661130480/job/20880319510
and results in this:
https://github.com/titusfortner/selenium/pull/58

### Motivation and Context
The previous release process didn't handle the Rust version -> SM build -> prep process correctly.
There's no good way to do that with a Rake task, so I created this github workflow to do it.

This is also nice because you can kick it off and it will automatically run the tests.
While the tests are running:
1. Verify the files all look correct
2. Pull the PR (`gh pr checkout 12345`), make the updates to the changelogs and push them back to the pr and merge it

Then all that is left is to run: `./go all:release` and it should finish doing everything


.github/workflows/ci-rust.yml
```diff
@@ -43,6 +52,8 @@ jobs:
     steps:
       - name: "Checkout project"
         uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.inputs.branch }}
```
p0deje: You can use `${{ inputs.branch }}` for this (and below).

p0deje (approved):

titusfortner:
I'm doing a few things wrong with this one, it'll be better once #13516 is merged.

titusfortner:
Ok, this has been updated to use v4 actions, and makes sure inputs are strings not booleans, because booleans confuse bash. 

I re-ran it on my fork:
https://github.com/titusfortner/selenium/actions/runs/7673663697/job/20917198122

this is the PR:
https://github.com/titusfortner/selenium/pull/58

I'm going to merge it once tests pass; let me know if there are any concerns or things I can change.

PR #13514, opened 2024-01-25
[ci] automatically update PRs with format fixes

titusfortner:
### Status
Well, now I'm trying to see if I can do something complicated and add a label to rerun tests if there has been an automatic reformatting. It's set up so that it shouldn't end up in a loop. Not sure it is worth going this far, though.

### Description
Uses what was done for automatically pinning browsers
Creates a patch file with format changes
In a new job applies the patch file and commits it
pushes the resulting commit back to the PR

### Motivation and Context
It's a pain to ask people to fix formatting in their PRs

I introduced formatting bugs in the PR so this should push to the PR. 🤞 

There we go, it works!
https://github.com/SeleniumHQ/selenium/pull/13514/commits/36ccb886d4d4a99972f8159193fed0ebf2d6e04c

### DIFF
--- a/.github/workflows/ci-rbe.yml
+++ b/.github/workflows/ci-rbe.yml
@@ -6,11 +6,13 @@ on:
     branches:
       - trunk
   workflow_dispatch:
+  issues:
+    types: [ labeled ]
 
 jobs:
   format:
     name: Format
-    if: github.repository_owner == 'seleniumhq'
+    if: github.repository_owner == 'seleniumhq' && github.event_name != 'issues'
     uses: ./.github/workflows/bazel.yml
     with:
       name: Check format script run
@@ -18,12 +20,13 @@ jobs:
       ruby-version: jruby-9.4.5.0
       run: ./scripts/github-actions/check-format.sh
       artifact-name: format-fixes
+      ignore-failure: true
 
   push_fixes:
     name: Push changes
     runs-on: ubuntu-latest
     needs: format
-    if: github.event_name != 'push' && ${{ failure() }}
+    if: github.event_name != 'push' && ${{ failure() }} && github.event_name != 'issues'
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


titusfortner:
well, this probably works. We'll see once selenium-ci gets off its butt and signs the CLA.

titusfortner:
Hmm, but I'm expecting this to re-run the tests... I wonder why it does not...

p0deje:
> I introduced formatting bugs in the PR so this should push to the PR. 🤞


This would only work on Selenium PRs, not from forks. Is this fine for you?

titusfortner:
@p0deje why wouldn't it work for forks?

p0deje:
> @p0deje why wouldn't it work for forks?

@titusfortner Because Selenium bot would not have permissions to push to the forked repository, would it?

titusfortner:
There's a checkbox for
"Allow edits from maintainers" that I believe is checked by default. But I should test something that has it turned off.

titusfortner:
Hmm, the label to rerun the tests didn't kick off new tests automatically. Does this need to be on trunk first? Maybe this shouldn't be automatic, but run manually by adding the label? I'm going to test this more on my fork...

.github/workflows/ci-rbe.yml
```diff
@@ -6,21 +6,95 @@ on:
     branches:
       - trunk
   workflow_dispatch:
+  issues:
+    types: [ labeled ]
```
diemol, line 10: Will this run when any label is applied?

titusfortner:
I spent a *lot of time exploring options here, and I'm not happy with anything.

To prevent infinite loops, github won't allow tests to rerun from the same workfkow. I can get it to label and rerun, but the tests don't get attached to the commit that was just made in the UI. Either way, the PR shows zero checks in the checks tab (what is in this PR)

I'm out of ideas and tired of working on this, so I'm giving up. 😄

PR #13522, opened 2024-01-29
Copyright dotnet

titusfortner:
### Status
I did not see any files that had other copyright info at the top or anything that should be excluded from these changes.

### Description
* Update copyright script to work for cs files:
    * Add file name and copyright tags at top and bottom (requires turning properties into methods to pass file name)
    * Change encoding to prevent "Zero Width No-Break Space" characters
    * Makes sure there is a '\n' after the copyright if one does not exist
    * Removes trailing white space on lines while we're at it
* Fixes .cs files
    * None of the test files and a few of the recently added files did not include copyright at all
    * Several files were missing the closing copyright tag
    * Several files were misnamed (copy/paste errors)

### Motivation and Context
Bring .NET in line with all the other languages


dotnet/src/support/Events/EventFiringWebDriver.cs
```diff
@@ -1,19 +1,20 @@
-// <copyright file="EventFiringWebDriver.cs" company="WebDriver Committers">
+// <copyright file="EventFiringWebDriver.cs" company="Selenium Committers">
 // Licensed to the Software Freedom Conservancy (SFC) under one
-// or more contributor license agreements. See the NOTICE file
+// or more contributor license agreements.  See the NOTICE file
```
nvborisenko, line 3: Double whitespace.

nvborisenko:
I would prefer using `.editorconfig` file. Then everybody (not only you equipped with py script), can write some code satisfying project rules.

Moreover, then we can include `dotnet format` command as a part of CI Format script.

dotnet/src/support/Events/EventFiringWebDriver.cs
titusfortner: Back in the olden days before the internet, 2 spaces was the standard. Regardless, every other file in Selenium repo has this information double spaced.

titusfortner:
There is a bazel command for it, too! 😜 

It's been added into the format script, so it will fail RBE test like everything else that isn't properly formatted, and the fix, like other things in the project will be to run `./scripts/format.sh` and commit the changes.

I'm not opposed to having editorconfig, but someone has to write it. 😄

dotnet/src/support/Events/EventFiringWebDriver.cs
nvborisenko: Why not triple spaced.. :) Since it is spread across all files in the repo, I am ok with it.

nvborisenko (approved):
Go ahead, please don't forget to resolve conflicts and verify newly added files with old copyright notice.

nvborisenko:
Ah, we already have `.editorconfig` file.

I have added the following there:
```
[*.cs]
indent_style = space
indent_size = 4
end_of_line = crlf
charset = utf-8
file_header_template = <copyright file="{fileName}" company="Selenium Committers">\nLicensed to the Software Freedom Conservancy (SFC) under one\nor more contributor license agreements.  See the NOTICE file\ndistributed with this work for additional information\nregarding copyright ownership.  The SFC licenses this file\nto you under the Apache License, Version 2.0 (the\n"License"); you may not use this file except in compliance\nwith the License.  You may obtain a copy of the License at\n\n  http://www.apache.org/licenses/LICENSE-2.0\n\nUnless required by applicable law or agreed to in writing,\nsoftware distributed under the License is distributed on an\n"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY\nKIND, either express or implied.  See the License for the\nspecific language governing permissions and limitations\nunder the License.\n</copyright>
```

Works good, but `;` symbol breaks everything. I tried escaping it in many ways - no results.

image

A1exKH (approved):
LGTM.

VietND96 (approved):

VietND96:
Merged the PR based on reviewers approval and CI passed.

Issue #13302, opened 2023-12-13
[🚀 Feature]:[bidi][grid]Remove se:bidi and rewrite webSocketUrl


diemol:
### Feature and motivation

In Grid, currently, we are enhancing the `webSocketUrl` returned after a BiDi session is created. This value is sent back to the client as `se:bidi`. 

### Usage example

Thinking this through, we do not need `se:bidi`, we can rewrite `webSocketUrl` and therefore simplify the logic implementation on the client side.

pujagani:
The approach is the above changes tries to reconstruct the websocket URL while creating websocket connection in the Node side, since it is possible to do so because it uses the session id in the URL. The only parameter that is required is the port in which the BiDi session is running, so that is added as a capability. 
The approach may not be ideal but it prevents having "se:bidi" capability to store the rewritten URL and simplifies the logic a bit..

pujagani:
Closed via https://github.com/SeleniumHQ/selenium/pull/13528

elgreco247:
Doesn't seem to work for Chrome 123.0.6312.86 (no webSocketUrl capability returned from grid).

diemol:
Please create an issue and fill out the template, thanks.

PR #13528, opened 2024-01-31
[java] Remove "se:bidi"

pujagani:

### Description

Implements #13302

### Types of changes

### Checklist


java/src/org/openqa/selenium/grid/node/ProxyNodeWebsockets.java
```diff
@@ -49,7 +49,7 @@ public class ProxyNodeWebsockets
     implements BiFunction<String, Consumer<Message>, Optional<Consumer<Message>>> {
 
   private static final UrlTemplate CDP_TEMPLATE = new UrlTemplate("/session/{sessionId}/se/cdp");
-  private static final UrlTemplate BIDI_TEMPLATE = new UrlTemplate("/session/{sessionId}/se/bidi");
+  private static final UrlTemplate BIDI_TEMPLATE = new UrlTemplate("/session/{sessionId}");
```
diemol: I think we can keep the endpoint, just set the URL we are creating in `LocalNode` in the `webSocketUrl` capability.

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
```diff
@@ -806,19 +806,27 @@ private Session createExternalSession(
 
     // Check if the user wants to use BiDi
     boolean webSocketUrl = toUse.asMap().containsKey("webSocketUrl");
-    // Add se:bidi if necessary to send the bidi url back
-    boolean bidiSupported = isSupportingBiDi || toUse.getCapability("se:bidi") != null;
+    boolean bidiSupported = isSupportingBiDi || toUse.getCapability("webSocketUrl") != null;
     if (bidiSupported && bidiEnabled && webSocketUrl) {
-      String bidiPath = String.format("/session/%s/se/bidi", other.getId());
-      toUse = new PersistentCapabilities(toUse).setCapability("se:bidi", rewrite(bidiPath));
+      String biDiUrl = (String) other.getCapabilities().getCapability("webSocketUrl");
```
diemol, line 816: Why don't we rewrite the URL and keep the endpoint? We could put the URL in `webSocketUrl`.

diemol (changes requested):

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
pujagani: https://github.com/SeleniumHQ/selenium/issues/13302#issuecomment-1918459493 The rationale behind the changes. Let me know what you think or if I am missing something. I did try the approach suggested earlier but that did not work due to not having the port information.

diemol: I am sorry, I missed those comments. 

What if we store the "local" `webSocketUrl` in a different field and then retrieve that URL when the BiDi endpoint is hit? In the same way we do it with `debuggerAddress`.

As in `se:cdp` maps to `debuggerAddress`, we could map `webSocketUrl` to `se:gridWebSocketUrl`.

Just and idea.

pujagani: Sounds good!

java/src/org/openqa/selenium/grid/node/local/LocalNode.java
```diff
@@ -805,20 +805,33 @@ private Session createExternalSession(
...
+      URI uri = null;
+      try {
+        uri = new URI(biDiUrl);
+      } catch (URISyntaxException e) {
+        throw new IllegalArgumentException("Unable to create URI from " + uri);
+      }
+      toUse =
+          new PersistentCapabilities(toUse)
+              .setCapability("se:gridWebSocketUrl", uri)
+              .setCapability("webSocketUrl", rewrite(uri.getPath()));
```
diemol: Why don't we keep using the `se/bidi` endpoint? Then the matching in `ProxyWebNodeSockets` is easier. We will still need to read the `webSocketUrl` in `DriverServiceSessionFactory` and get the internal one, but at least the client code does not need to know anything about it.

pujagani: The matching logic has not changed in `ProxyWebNodeSockets`. It is now actually following the pattern of the WebSocket URL i.e.` /session/{sessionId}`. Do you mean it will impact performance while matching?

diemol: I only meant it from the simplicity point of view, as every functionality has its own endpoint. So, leaving it the way it was:
```
private static final UrlTemplate BIDI_TEMPLATE = new UrlTemplate("/session/{sessionId}/se/bidi");
```

diemol (approved):
Thank you, @pujagani!

PR #13576, opened 2024-02-13
[dotnet] Fix network response data encoding

### LINKED ISSUE #11726: [🚀 Feature]: Direct access to the raw data of the body via stream/byte array and not exclusively as string
### Feature and motivation

Unfortunately, I have not yet found a way to get to my desired destination even via detours. Intermediary proxies have been the only option so far, but I'd like it better if there was a way to do this in Selenium itself. In this respect I'll put the topic here as a feature request. In my search I have not found a similar feature request yet. If there should be a similar feature request, we can close it right away.

But now to my actual request:
I use Selenium in C# to run automated tests. In this context, I use the possibilities available in Selenium to tap the generated network traffic in order to run further tests. However, in this context I came across something that makes me unhappy. Within the provided network functions and events to influence the same, in all objects the body is available to me exclusively as data type "String". But in my examples binary data is transferred. This results in very unpleasant consequences for me. First, I can't match the data, because I first have to convert the provided string back into a byte array. However, this may cause conversion errors, plus it is completely impractical.

For this reason, my wish would be for Selenium to be able to provide the entire body data as a standalone object. Of course, that can continue to provide the body additionally as a string for all I care, but it should also be possible to access the original data stream in binary form, either as a stream, byte array, or similar constructs. I have roughly sketched this below as an example. For the time being, this only concerns the response side. From my point of view, such a customization would of course have to be done from the request side through everything between status and on the response side.

I would be very grateful if you could provide the data in this form.

### Usage example

```
internal class Program
{
    private static void Main(string[] args)
    {
        IWebDriver driver = new ChromeDriver();
        SetupNetworkLogging(driver);
        driver.Navigate().GoToUrl("https://www.selenium.dev/");

        Console.Read();
    }

    public static void SetupNetworkLogging(IWebDriver driver)
    {
        var manager = new NetworkManager(driver);
        manager.NetworkResponseReceived += ResponseHandler;

        var monitoring = manager.StartMonitoring();
        monitoring.Wait();
    }

    private static void ResponseHandler(object sender, NetworkResponseReceivedEventArgs e)
    {
        // sample binary access
        var body = e.ResponseBody;
        using (Stream responseStream = body.GetStream())
        {
            while (responseStream.Position < responseStream.Length)
            {
                var buffer = new byte[4];
                responseStream.Read(buffer, 0,buffer.Length);
                // The buffer can be processed further here.
            }
        }

        // This way you could get the body as a string, possibly even as an overload with a special encoding.
        Console.Write(body.ToString());
    }
}
```

nvborisenko:
### Description
NetworkManager exposes HTTP response body data as `string` type. Not all responses have strings. In case if the response is bytes, selenium is doing extra convertion from `bytes` to `utf8 string`. It leads to data loosing, we cannot convert string back to bytes.

### Motivation and Context
Fix #11726 

### New approach to get data
```csharp
network.NetworkResponseReceived += (sender, e) =>
{
    if (e.ResponseResourceType == "Image")
    {
        Console.WriteLine($"{e.ResponseUrl} : {e.ResponseResourceType}");
        if (!string.IsNullOrEmpty(e.ResponseBody))
        {
            Console.WriteLine(Encoding.UTF8.GetBytes(e.ResponseBody).Length);
        }

        if (e.ResponseContent != null)
        {
            Console.WriteLine(e.ResponseContent.ReadAsByteArray().Length);
        }
    }
};

await network.StartMonitoring();

driver.Url = "https://selenium.dev";
```

And its output is:
```
https://www.selenium.dev/images/sponsors/saucelabs.png : Image
11191
6255
```

where `6255` is correct bytes of the image.

### Types of changes

### Checklist

### DIFF
--- a/dotnet/src/webdriver/DevTools/v119/V119Network.cs
+++ b/dotnet/src/webdriver/DevTools/v119/V119Network.cs
@@ -261,7 +261,7 @@ namespace OpenQA.Selenium.DevTools.V119
                     }
                     else
                     {
-                        responseData.Content = new HttpResponseContent(Encoding.UTF8.GetBytes(bodyResponse.Body));
+                        responseData.Content = new HttpResponseContent(bodyResponse.Body);
                     }
                 }
             }
--- a/dotnet/src/webdriver/DevTools/v119/V119Network.cs
+++ b/dotnet/src/webdriver/DevTools/v119/V119Network.cs
@@ -257,11 +257,11 @@ namespace OpenQA.Selenium.DevTools.V119
                 {
                     if (bodyResponse.Base64Encoded)
                     {
-                        responseData.Body = Encoding.UTF8.GetString(Convert.FromBase64String(bodyResponse.Body));
+                        responseData.Content = new HttpResponseContent(Convert.FromBase64String(bodyResponse.Body));
                     }
                     else
                     {
-                        responseData.Body = bodyResponse.Body;
+                        responseData.Content = new HttpResponseContent(Encoding.UTF8.GetBytes(bodyResponse.Body));
                     }
                 }
             }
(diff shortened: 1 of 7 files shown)


diemol (approved):
Thank you, @nvborisenko!

PR #13582, opened 2024-02-14
Use all provided options while merging them for Firefox

oleg-rd:
### Description
This fix addresses the discrepancy that we have in methods of processing browser options for Chromium(`mergeChromiumOptions`) and Firefox(`mergeFirefoxOptions`). In the case of Chromium, if the option name doesn't match any of previously processed, we add it anyway. It doesn't happen for Firefox. So in the case when one option is set in stereotype, but other is provided in capabilities this second one is missed from the result.

### Motivation and Context
We have observed different behavior with the same setup running tests for Chrome and Firefox. While for Chrome everything worked fine, in case of Firefox provided capability was missed while creating session so that we got an exceptions (in our case, missing `androidPackage` lead to error about missing Firefox binary, which wasn't expected in case of session for mobile device). Updated version of `shouldMergeFirefoxSpecificOptionsFromStereotypeAndCaps` unit test demonstrates this case.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/grid/node/config/SessionCapabilitiesMutator.java
+++ b/java/src/org/openqa/selenium/grid/node/config/SessionCapabilitiesMutator.java
@@ -171,6 +171,8 @@ public class SessionCapabilitiesMutator implements Function<Capabilities, Capabi
       Map<String, Object> stereotypeOptions, Map<String, Object> capsOptions) {
     Map<String, Object> toReturn = new HashMap<>(stereotypeOptions);
 
+    List<String> handledOptions = List.of("args", "prefs", "profile", "log", "binary");
+
     for (Map.Entry<String, Object> entry : capsOptions.entrySet()) {
       String name = entry.getKey();
       Object value = entry.getValue();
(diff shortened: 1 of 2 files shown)


java/src/org/openqa/selenium/grid/node/config/SessionCapabilitiesMutator.java
```diff
@@ -215,6 +217,14 @@ private Map<String, Object> mergeFirefoxOptions(
         Map<String, Object> logLevelMap = (Map<String, Object>) value;
         toReturn.put("log", logLevelMap);
       }
+
+      if (name.equals("binary") && !stereotypeOptions.containsKey("binary")) {
+        toReturn.put(name, value);
+      }
```
diemol, line 223: We intentionally decided to not do this one because the assumption is that the Grid is already configured and a test should not determine the browser driver binary location.

diemol (changes requested):

java/src/org/openqa/selenium/grid/node/config/SessionCapabilitiesMutator.java
oleg-rd: Hi Diego!

There reason I've added this part is mainly because there is the same one for Chromium in the existing implementation -- [link](https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/grid/node/config/SessionCapabilitiesMutator.java#L158)

Also, I think this condition covers the case when binary wasn't configured from grid to allow to set it from tests.

Anyway, I'm OK with removing it from my PR as it doesn't directly affect our case - let me know.

diemol: I missed that. OK. Good point. Thank you.

diemol (changes requested):
Could you please run this [script](https://github.com/oleg-rd/selenium/blob/merge-all-firefox-options/scripts/format.sh) and commit the changes?

diemol (approved):
Thank you, @oleg-rd!

PR #13583, opened 2024-02-14
Fixing NPE when "se:downloadsEnabled" not set

### LINKED ISSUE #13578: [🐛 Bug]: NPE in HasDownloads.requireDownloadsEnabled() when using local browser
### What happened?

First of all, thank you for the new download function. It works really well.

Even though [the documentation](https://www.selenium.dev/documentation/grid/configuration/cli_options/#enabling-managed-downloads-by-the-node) clearly indicates that a download is only supported via a Selenium Grid, I found an unfortunate NPE when trying to perform a download without Selenium Grid.

The NPE occurs when `HasDownloads.requireDownloadsEnabled()` tries to cast the return value of `capabilities.getCapability("se:downloadsEnabled")` to a boolean. In the case of a locally launched browser, `capabilities.getCapability("se:downloadsEnabled")` returns `null`, causing the following NPE:

`java.lang.NullPointerException: Cannot invoke "java.lang.Boolean.booleanValue()" because the return value of "org.openqa.selenium.Capabilities.getCapability(String)" is null`

I think there are to ways to prevent this NPE.

- Changing `(boolean) capabilities.getCapability("se:downloadsEnabled")` to `Boolean.parseBoolean(capabilities.getCapability("se:downloadsEnabled"))` since it will evaluate `null` to `false`.
- Check for `null` and throw a WebDriverException stating that downloads are only supported when running tests via Selenium Grid:
```java
Object downloadCapability = capabilities.getCapability("se:downloadsEnabled");
    if (downloadCapability == null) {
      throw new WebDriverException("Use a Selenium Grid in order to work with downloadable files.")
    }
    boolean downloadsEnabled = (boolean) downloadCapability;
    if (!downloadsEnabled) {
      throw new WebDriverException(
          "You must enable downloads in order to work with downloadable files.");
    }
```

```shell
public static void main(String[] args) throws InterruptedException {
        ChromeOptions chromeOptions = new ChromeOptions();
        chromeOptions.setEnableDownloads(true);
        WebDriver driver = new ChromeDriver(chromeOptions);
        driver.navigate().to("https://www.selenium.dev/downloads/");
        WebDriver augmentedDriver = new Augmenter().augment(driver);

        // This will cause the NPE
        List<String> downloadedFiles = ((HasDownloads) augmentedDriver).getDownloadableFiles();
}
```

### Relevant log output

```shell
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "java.lang.Boolean.booleanValue()" because the return value of "org.openqa.selenium.Capabilities.getCapability(String)" is null
	at org.openqa.selenium.HasDownloads.requireDownloadsEnabled(HasDownloads.java:36)
	at org.openqa.selenium.remote.RemoteWebDriver.getDownloadableFiles(RemoteWebDriver.java:636)
```

### Operating System

Windows 10

### Selenium version

Java 4.17.0

Chrome 121

ChromeDriver 121

_No response_

Trigtrig:
### Description
Fixes the NPE described in https://github.com/SeleniumHQ/selenium/issues/13578. I'm using `capabilities.is("se:downloadsEnabled")` because this function already checks for possible `null` values and casts to `boolean` using `Boolean.parseBoolean()`.

### Motivation and Context
Fixes the NPE described in https://github.com/SeleniumHQ/selenium/issues/13578.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/HasDownloads.java
+++ b/java/src/org/openqa/selenium/HasDownloads.java
@@ -33,7 +33,7 @@ public interface HasDownloads {
    * @throws WebDriverException if capability to enable downloads is not set
    */
   default void requireDownloadsEnabled(Capabilities capabilities) {
-    boolean downloadsEnabled = (boolean) capabilities.getCapability("se:downloadsEnabled");
+    boolean downloadsEnabled = capabilities.is("se:downloadsEnabled");
     if (!downloadsEnabled) {
       throw new WebDriverException(
           "You must enable downloads in order to work with downloadable files.");


diemol (approved):
Thank you, @Trigtrig!

PR #13608, opened 2024-02-20
[dotnet] Treat unsuccessful http responses as errored commands

nvborisenko:
### Description
Currently http responses are treated as unsuccessful only based on body content. This PR verifies http status code first, avoiding a process of parsing body content.

### Motivation and Context
Related to https://github.com/SeleniumHQ/selenium/pull/11713, according https://www.w3.org/TR/webdriver2/#errors

### Types of changes

### Checklist


dotnet/src/webdriver/Response.cs
```diff
@@ -149,6 +144,26 @@ public static Response FromJson(string value)
...
+        {
+            Dictionary<string, object> deserializedResponse = JsonConvert.DeserializeObject<Dictionary<string, object>>(value, new ResponseValueJsonConverter());
+
+            var response = new Response();
+
+            var valueDictionary = deserializedResponse["value"] as Dictionary<string, object>;
+
+            response.Value = valueDictionary;
+
+            response.Status = WebDriverError.ResultFromError(valueDictionary["error"].ToString());
```
nvborisenko: We need guard here to avoid `NullReferenceException`

nvborisenko: Done.

PR #13661, opened 2024-03-06
[rb] Support running Ruby tests in RBE for Chrome

p0deje:
Hooks Ruby integration tests into RBE. For now, only Chrome is being tested, but I'm going to follow the Java approach and generate separate targets for each browser in the coming PRs:

- //rb/spec/integration/selenium/webdriver:driver-chrome
- //rb/spec/integration/selenium/webdriver:driver-firefox
- //rb/spec/integration/selenium/webdriver:driver-edge
- //rb/spec/integration/selenium/webdriver:driver-chrome-remote
- //rb/spec/integration/selenium/webdriver:driver-firefox-remote
- etc

Successful build - https://gypsum.cluster.engflow.com/invocations/default/88e8242b-d89a-4a1e-879e-6beb468c53a7. There are service tests failing due to SM invocation, so these targets are added to `.skipped-tests`.


WORKSPACE
```diff
@@ -367,6 +367,82 @@ rb_bundle_fetch(
         "//:rb/selenium-devtools.gemspec",
         "//:rb/selenium-webdriver.gemspec",
     ],
+    gem_checksums = {
```
shs96c, line 384: Out of curiosity, how is this block of checksums generated? Any chance we can put this in a lock file somewhere that isn't the main workspace?

rb/.ruby-version
```diff
@@ -1 +1 @@
-3.0.6
+jruby-9.4.5.0
```
shs96c: Makes sense. Makes things a lot simpler and faster for most people, and people who want to look into Ruby can toggle this easily.

rb/spec/tests.bzl
```diff
@@ -0,0 +1,85 @@
...
+    "//rb/spec/integration:safari": {
+        "WD_REMOTE_BROWSER": "safari",
+        "WD_SPEC_DRIVER": "safari",
+    },
+    "//rb/spec/integration:safari-preview": {
+        "WD_REMOTE_BROWSER": "safari-preview",
+        "WD_SPEC_DRIVER": "safari-preview",
+    },
+    "//conditions:default": {},
+}) | select({
```
shs96c, line 30: `|`? Interesting. I normally use `+`. Had no idea this would work

rb/spec/tests.bzl
```diff
@@ -0,0 +1,85 @@
+load("@rules_ruby//ruby:defs.bzl", "rb_test")
+load("//common:browsers.bzl", "chrome_data")
+
+ENV = select({
+    "//rb/spec/integration:chrome": {
```
shs96c, line 5: If I'm reading this right, you'll only be able to test one browser in a single build. Is that correct, or have I missed something?

shs96c (approved):
The Bazel bits look good to me. A couple of questions, but we an always iterate :)

rb/.ruby-version
p0deje: Actually, I think I committed this by mistake :)
I think we should keep the minimum supported Ruby version here - JRuby 9.4 is a newer version targeting Ruby 3.1.

WORKSPACE
p0deje: We can extract it into its own bzl file (e.g. `rb/gem_checksums.bzl`) and then `load()` it, but it won't work with Bzlmod because load's are not allowed within `MODULE.bazel`.

Ideally, this is part of `Gemfile.lock` itself and I track the work to implement it in https://github.com/bazel-contrib/rules_ruby/issues/63, but I haven't figured out how to enable checksums in Bundler yet.

This dictionary is generated by `rb_bundle_fetch()` repository rule and printed out.

rb/spec/tests.bzl
p0deje: `+` only works for lists, while `|` works for dicts
https://github.com/bazelbuild/starlark/issues/14

rb/spec/tests.bzl
p0deje: `+` only works for lists, while `|` works for dicts
https://github.com/bazelbuild/starlark/issues/14

rb/spec/tests.bzl
p0deje: Yes, at the moment. However, I'd like to add Java-like targets for each browser in the following PRs:

- //rb/spec/integration/selenium/webdriver:driver-chrome
- //rb/spec/integration/selenium/webdriver:driver-firefox
- //rb/spec/integration/selenium/webdriver:driver-edge
- //rb/spec/integration/selenium/webdriver:driver-chrome-remote
- //rb/spec/integration/selenium/webdriver:driver-firefox-remote
- etc

PR #13673, opened 2024-03-10
[dotnet] Add dispose on constructor failure to ensure driver closes

MASACR99:
### Description

First of all I just changed the dotnet version first since is the one my project uses so I could test it out, the idea behind this change is just that if the driver fails to start a browser version (for example if there's a missmatch between the version and driver) the driver closes instead of staying open in the background. For that I just added a try catch on the WebDriver constructor and in case there's an exception I just force the disposal.

### Motivation and Context

The issue came from my work where we use selenium on a server, but the IT guy managing the server refuses to force a version for the browser so from time to time when any of the browsers updates and the driver stops working it leaves a ton of drivers open in task manager which must be closed before upgrading the driver, with this fix these rogue driver sessions should stop.

### Types of changes

### Checklist

I don't think this change requires updated tests, 5 tests failed locally but don't seem to be related to my change

___

## **Type**
Bug fix

___

## **Description**
- Added error handling in the WebDriver constructor to ensure resources are properly disposed of in case of an exception during driver initialization.
- This change aims to address issues where driver instances remain running in the background if an exception occurs, particularly useful in environments where browser versions frequently change.

___

## **Changes walkthrough**
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Error handling
</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>WebDriver.cs</strong><dd><code>Add Exception Handling to WebDriver Constructor</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
<li>Wrapped <code>StartSession</code> and subsequent initializations in a try-catch <br>block.<br> <li> In the catch block, added a call to <code>Dispose</code> before rethrowing the <br>exception.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/13673/files#diff-ef1874c09b1dca0d62ea1f0ead685355944b9cb7ae63df60ad9b32abdf90c8dc">+18/-9</a>&nbsp; &nbsp; </td>

</tr>                    

___

> ✨ **PR-Agent usage**:


diemol:
This is not the correct solution for your issue. Selenium Grid has a way to stop processes when a session fails to start, and the server is responsible for cleaning up when something fails during the creation. 

In addition, this won't work in all cases because there is no guarantee a session was created, which causes the user to see exceptions will be thrown for no reason. 

If you still want that code, please put it in your framework. On the other hand, we can help with the Grid configuration. Join us in our Slack channel: https://www.selenium.dev/support/#ChatRoom

MASACR99:
Hi, thanks diemol for the feedback but I still think there should be a way the Webdriver closes the driver if it's left open after trying to start a session, it takes a lot less effort to just check if the driver was created and dispose of it rather than having the server have to check its processes and kill the zombie drivers.

I can add a check in the exception to check if the driver was created to avoid the dispose generating errors but, as I see the current implementation of the dispose, it already checks some exceptions and supresses them.

Thanks for your time and your work on these projects!

diemol:
I just think issues should get fixed where they happen and avoid workarounds. But I am open to hearing other opinions... @nvborisenko @titusfortner

nvborisenko:
Ideally WebDriver should not perform any operations in the constructor. Ideally we want to see something like (in Selenium v5?):
```csharp
using var driver = new ChromeDriver();
await driver.StartAsync(); // or any other method name to initiate new session
```

In this way `ctor` is just an initialization of the object state, what is good. There are no any hidden activities. But we cannot just break existing logic.

This issue is about if `this.StartSession(capabilities);` throws an exception, then user has no ways to clean up all underlying disposable stuff.

For this particular case I can accept:
```csharp
try
{
  this.StartSession(capabilities);
}
catch (Exception)
{
  // !!! dispose specific resources, don't call this.Dispose()
  throw;
}
```

MASACR99:
Sorry for the delay had a few things to manage. This afternoon I'm gonna implement this changes, I just have 1 question, if we try to dispose the driver and it fails to receive the Quit command should the exception be thrown (there'd be an exception thrown inside the catch of the StartSession command) or managed and suppressed?

nvborisenko:
I haven't looked at code yet, and I expect `Quit()` method should raise an exception. If the object is constructed successfully, then we suppose `Dispose/Quit` also should be successful, we should not be silent and ignore exceptions - raise it to end user.

Looked at code: we are silent if `Quit()` throws an exception, the same in Java. Let's don't touch this behavior in scope of your pull request.

nvborisenko:
@MASACR99 thank you. Before moving further can we see exact exception (or problem) you met, which is motivated you to create this PR. I am asking because of while I was reviewing your PR I saw other issues related to disposing. And now I am not sure which case we are trying to resolve here. Am I right that if `this.StartSession()` throws an exception, then we meet resources leaking?

MASACR99:
Yes so the current issue I'm trying to fix happens when creating a new WebDriver (in my case I managed to test with Chrome and Firefox), if I specify a driver version that isn't compatible with the currently installed driver (I used in this test the chromedriver for version 114 and have chrome version 122.0.6261.112 installed) and try to create a new driver I get an output error with the message 

> "Unhandled exception. System.InvalidOperationException: session not created: This version of ChromeDriver only supports Chrome version 114"

With the stack trace being: 
> Current browser version is 122.0.6261.112 with binary path C:\Program Files\Google\Chrome\Application\chrome.exe (SessionNotCreated)
   at OpenQA.Selenium.WebDriver.UnpackAndThrowOnError(Response errorResponse, String commandToExecute)
   at OpenQA.Selenium.WebDriver.StartSession(ICapabilities capabilities)
   at OpenQA.Selenium.Chromium.ChromiumDriver..ctor(ChromiumDriverService service, ChromiumOptions options, TimeSpan commandTimeout)
   at OpenQA.Selenium.Chrome.ChromeDriver..ctor(ChromeDriverService service, ChromeOptions options, TimeSpan commandTimeout)
   at OpenQA.Selenium.Chrome.ChromeDriver..ctor(String chromeDriverDirectory, ChromeOptions options, TimeSpan commandTimeout)
   at OpenQA.Selenium.Chrome.ChromeDriver..ctor(String chromeDriverDirectory, ChromeOptions options)
   at OpenQA.Selenium.Chrome.ChromeDriver..ctor(String chromeDriverDirectory)
   at Program.<Main>$(String[] args) in C:\Users\username\source\repos\test\test\Program.cs:line 4

And after the program has crashed there will be a driver.exe kept open in the task manager (I've been able to replicate on different windows machines not sure if Linux or Mac would detect te process and close it)
   
### If you want some steps to reproduce the error I'll list them here:
1. Get a driver version not matching currently installed target browser
2. Create new driver for browser forcing incorrect driver
3. An error will happen and a "browsername"driver.exe will be kept open in the process list


### I'll also add the code I used for this test 
(code in C# .NET8 with latest version of selenium nugget installed):

````
using OpenQA.Selenium.Chrome;

var driver = new ChromeDriver("C:\\repos\\test\\test\\chromedriver.exe");

driver.Navigate().GoToUrl("https://www.google.com");
````


Keep in mind in this specific test, after the console application closes, any processes created by it will close BUT if a windows service has this issue the chromedriver will stay open on the background.

And I'll also attach some images showing proof of this happening (i've censored the user name folder in the screenshots)
selenium-proof-1
selenium-proof-2
selenium-proof-3

**Comment by nvborisenko** [comment]
And does `Quit()` method close the `chromedriver.exe` process?

File: dotnet/src/webdriver/WebDriver.cs [hunk 1]
```diff
@@ -54,7 +54,18 @@ public class WebDriver : IWebDriver, ISearchContext, IJavaScriptExecutor, IFinds
...
-            this.StartSession(capabilities);
+
+            try
+            {
+                this.StartSession(capabilities);
+            }
+            catch (Exception)
+            {
+                // Failed to start driver session, disposing of driver
+                this.DriverDispose();
```
nvborisenko: Just `this.Quit()`? Thus no need to split disposal to several methods. Yeah, I asked to dispose specific allocated resources before the exception raises, but now, when we have clearer picture, let's keep things as simple as possible (where `Quit()` looks better than `Dispose()` in ctor)

nvborisenko (changes requested):

dotnet/src/webdriver/WebDriver.cs
MASACR99: Last commit undid DriverDispose and changed the catch to call this.Quit() as requested

nvborisenko (approved):
Thank you @MASACR99 !

Issue #13930, opened 2024-05-11
[🐛 Bug]: Getting SessionNotCreatedException when the Date of the Windows is in Arabic language format


MahmoudElSharkawy:
### What happened?

When the Date of the windows is in Arabic language format, the session does not start and I get 
`org.openqa.selenium.SessionNotCreatedException: Could not start a new session. Possible causes are invalid address of the remote server or browser start-up failure. 
`

![image](https://github.com/SeleniumHQ/selenium/assets/46620378/f1a328ba-8f94-41d5-8783-c0a6eda7c5a2)

I get this when I debug the issue:
![image](https://github.com/SeleniumHQ/selenium/assets/46620378/f02263e4-b0c9-4796-861b-b6ff91618282)

### How can we reproduce the issue?

```shell
public class TestDataLanguage {
// Just switch the language of the date to Arabic and run this simple code with any browser
    @Test
    public void testDataLanguage() {
        WebDriver driver = new ChromeDriver();
        driver.get("https://www.google.com/");
    }
}
```

### Relevant log output

```shell
org.openqa.selenium.SessionNotCreatedException: Could not start a new session. Possible causes are invalid address of the remote server or browser start-up failure. 

	at org.openqa.selenium.remote.RemoteWebDriver.execute(RemoteWebDriver.java:536)
	at org.openqa.selenium.remote.RemoteWebDriver.startSession(RemoteWebDriver.java:232)
	at org.openqa.selenium.remote.RemoteWebDriver.<init>(RemoteWebDriver.java:161)
	at org.openqa.selenium.chromium.ChromiumDriver.<init>(ChromiumDriver.java:114)
	at org.openqa.selenium.chrome.ChromeDriver.<init>(ChromeDriver.java:88)
	at org.openqa.selenium.chrome.ChromeDriver.<init>(ChromeDriver.java:83)
	at org.openqa.selenium.chrome.ChromeDriver.<init>(ChromeDriver.java:51)
	at TestDataLanguage.testDataLanguage(TestDataLanguage.java:8)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #13934, opened 2024-05-13
[Java] Migrate the String.Format to String.Format(Locale.US) to handle different langauges for Edge, Chrome and Firefox

MustafaAgamy:
### Description
- This is a contributions to fix: 
https://github.com/SeleniumHQ/selenium/issues/13930

*Changes Done:
- Migrate the String.Format to Formatter.Format in :
- ChromeDriverSerivce.java
- EdgeDriverService.java
- RemoteWebDriver.java

### Motivation and Context
*Issue explaination:
- Having System Date & Language in Arabic caused with driver setup. Using String.Format for ports and some other lines caused them to be localized in the system language and in arabic case it caused the numbers and some other character to be localized in arabic, e.g. the port. And that caused an issues when appending the port to the URL in order to setup the browser

*Fix explaination:
- Now it's utilizing the Formatter.Format and forcing the localization to be in (English.US) ny passing the corresponding Locale in the Formatter Constructor

### Types of changes

### Checklist

___

### **PR Type**
Bug fix

___

### **Description**
- Updated `ChromeDriverService`, `EdgeDriverService`, and `RemoteWebDriver` to use `Formatter.format` with `Locale.US` to ensure consistent formatting across different system locales.
- This change addresses issues where system locale settings (like Arabic) affected the formatting of numbers and text in logs and port settings, potentially causing errors in browser setup.
- The use of `Locale.US` ensures that numbers and text are formatted in a standard way, avoiding issues related to localization.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix
</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>ChromeDriverService.java</strong><dd><code>Use Formatter with Locale.US in ChromeDriverService</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
java/src/org/openqa/selenium/chrome/ChromeDriverService.java

<li>Replaced <code>String.format</code> with <code>Formatter.format</code> using <code>Locale.US</code> for <br>consistent number formatting.<br> <li> Ensured log file paths and log levels are formatted without <br>localization issues.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/13934/files#diff-405c68a6b743ea8f4b340b5ea0c92c802a24109b4c38818335f6ca39511e2bc8">+6/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>EdgeDriverService.java</strong><dd><code>Standardize EdgeDriverService Formatting with Locale.US</code>&nbsp; &nbsp; </dd></summary>
<hr>
      
java/src/org/openqa/selenium/edge/EdgeDriverService.java

<li>Switched from <code>String.format</code> to <code>Formatter.format</code> with <code>Locale.US</code> to <br>avoid localization in log and port settings.<br> <li> Adjusted logging and port settings to use non-localized number <br>formats.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/13934/files#diff-62903f54af704e41bd4657395512ce9c62fa0d4966ef3f0f665fcc045e7b15c4">+6/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>RemoteWebDriver.java</strong><dd><code>Implement Formatter in RemoteWebDriver for Uniform Output</code></dd></summary>
<hr>
      
java/src/org/openqa/selenium/remote/RemoteWebDriver.java

<li>Updated string formatting to use <code>Formatter</code> with <code>Locale.US</code> for <br>consistent command and log messages.<br> <li> Modified session handling and logging to prevent localization issues.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/13934/files#diff-0cf141b01e48a733ec9037c51d8e610e7727d05a6b79aa0176def7c3a0fc311b">+8/-6</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

___

> 💡 **PR-Agent usage**:


java/src/org/openqa/selenium/chrome/ChromeDriverService.java
```diff
@@ -283,34 +285,35 @@ protected void loadSystemProperties() {
     @Override
     protected List<String> createArgs() {
       List<String> args = new ArrayList<>();
-      args.add(String.format("--port=%d", getPort()));
+      try (Formatter formatter = new Formatter(Locale.US)) {
```
pujagani: It might be a good idea to add the same implementation to [GeckoDriver](https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/firefox/GeckoDriverService.java) service as well to ensure consistency across languages.

MustafaAgamy: Will do so. 

However the fix still require some investigation for firefox, so I'm still looking into it and I'll commit the changes and Tests

MustafaAgamy: @pujagani  Added the required logic in GekoDriverService.java

java/src/org/openqa/selenium/firefox/GeckoDriverService.java
```diff
@@ -219,15 +221,15 @@ protected void loadSystemProperties() {
     @Override
     protected List<String> createArgs() {
       List<String> args = new ArrayList<>();
-      args.add(String.format("--port=%d", getPort()));
+      args.add(new Formatter(Locale.US).format("--port=%d", getPort()).toString());
```
pujagani: I think this needs to be done similarly to ChromeDriverService to ensure the `Formatter` instance is closed using try-with-resources.

java/src/org/openqa/selenium/remote/RemoteWebDriver.java
```diff
@@ -815,9 +817,9 @@ public String toString() {
       platformName = "unknown";
     }
 
-    return String.format(
-        "%s: %s on %s (%s)",
-        getClass().getSimpleName(), caps.getBrowserName(), platformName, getSessionId());
+      return new Formatter(Locale.US).format(
```
pujagani: It might be a good idea to ensure a single `Formatter` instance is used and that instance is closed too.

java/src/org/openqa/selenium/firefox/GeckoDriverService.java
MustafaAgamy: The issue is we've to flush the formatter everytime because when I triumph D using one formatter object without flushing it
It causes the strings to append

MustafaAgamy: @pujagani  I refactored the code to use String.format again with Locale.US

pujagani:
I think the changes in the PR are not necessary. I looked into this further to realize the findings below:
```
  public static String format(String format, Object... args) {
        return new Formatter().format(format, args).toString();
    } 
```

This creates a Formatter instance:
```
  public Formatter() {
        this(Locale.getDefault(Locale.Category.FORMAT), new StringBuilder());
    }
```
So setting the default using "`setDefault(Locale)`" should work out of the box without any changes.

pujagani:
Before we close this PR, can you try on your end to double-check if it works as expected using "`setDefault`"?

MustafaAgamy:
@pujagani
I just did this to minimize the scope of the affects

But setting the setting the Default to Locale.US should work

You can call it In the same create Args methods once

 I will change the fixes to that and give it a try?

MustafaAgamy:
@pujagani 

Done can you please check and let me know ? (it worked local)
Also maybe another pipeline run to make sure everything is working as expected?

java/src/org/openqa/selenium/edge/EdgeDriverService.java
```diff
@@ -276,6 +277,8 @@ protected void loadSystemProperties() {
 
     @Override
     protected List<String> createArgs() {
+      Locale.setDefault(Locale.US);
```
pujagani: This can be removed too. Just the test will suffice.

MustafaAgamy:
@pujagani  I revisited the formatting issue 

Should be good now

java/src/org/openqa/selenium/edge/EdgeDriverService.java
pujagani: I think setting locale explicitly is not required. " Locale.setDefault(Locale.US);" line can be removed from all classes.

java/test/org/openqa/selenium/edge/EdgeArabicDateTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.junit.jupiter.api.Test;
+import org.openqa.selenium.net.PortProber;
+import org.openqa.selenium.testing.JupiterTestBase;
+import org.openqa.selenium.testing.NoDriverBeforeTest;
+
+public class EdgeArabicDateTest extends JupiterTestBase {
+
+  @Test
+  @NoDriverBeforeTest
+  void shouldLaunchSuccessfullyWithArabicDate() {
```
diemol: Can this test be placed in one of the existing files?

java/test/org/openqa/selenium/chrome/ChromeArabicDateTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.junit.jupiter.api.Test;
+import org.openqa.selenium.net.PortProber;
+import org.openqa.selenium.testing.JupiterTestBase;
+import org.openqa.selenium.testing.NoDriverBeforeTest;
+
+public class ChromeArabicDateTest extends JupiterTestBase {
+
+  @Test
+  @NoDriverBeforeTest
+  void shouldLaunchSuccessfullyWithArabicDate() {
```
diemol: Can this test be placed in one of the existing files?

java/test/org/openqa/selenium/firefox/FirefoxArabicDateTest.java
```diff
@@ -0,0 +1,48 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.openqa.selenium.testing.drivers.Browser.FIREFOX;
+
+import java.util.Locale;
+import org.junit.jupiter.api.Test;
+import org.openqa.selenium.net.PortProber;
+import org.openqa.selenium.testing.JupiterTestBase;
+import org.openqa.selenium.testing.NoDriverBeforeTest;
+
+public class FirefoxArabicDateTest extends JupiterTestBase {
```
diemol: Can this test be placed in one of the existing files?

diemol (changes requested):

java/test/org/openqa/selenium/chrome/ChromeArabicDateTest.java
MustafaAgamy: @diemol Done Please Check

java/test/org/openqa/selenium/firefox/FirefoxArabicDateTest.java
MustafaAgamy: @diemol Done Please Check

java/test/org/openqa/selenium/edge/EdgeArabicDateTest.java
MustafaAgamy: @diemol Done Please Check

titusfortner:
@diemol this look ok?

pujagani:
https://github.com/SeleniumHQ/selenium/pull/13934#issuecomment-2136539473 -  I think we don't need to make any changes. Just the tests are sufficient.

MustafaAgamy:
@pujagani 

But without forcing the localization to be In English, users having their OS default language In arabic will still counter the same issue, won't they?

titusfortner:
Is this correct summary?
When a user's default language isn't English, then the code is going to fail unless they add `Locale.setDefault(Locale.US)`.

Then is the question whether we add it to the library itself or whether the user needs to add it to their test?
Is there a scenario in which adding that code can cause other problems?

MustafaAgamy:
@titusfortner
I don't see adding it to Selenium itself  causing any issues from My POV tbh

And also many of the users In the middle east are facing start up failures cuz of locale issue

pujagani:
If users are facing a startup problem, it might be good idea to add it on their end. I am not sure if Selenium should assume which default locale to set. Rather the users can themselves add the locale of their choice. What we can do instead is have a test to show how to do that (which a part of this PR does) and have a blog/documentation regarding this. Does that make sense?

MustafaAgamy:
I Tested it against some langauges like chinese and japanese, only arabic had this issue because of the url setup

- Users are getting start up failure exceptions with no indication of What went wrong at all so having a documentation might not be enough for them or at least provide a meaningful error for them

- Selenium already got Selenium Manager for better drivers handling for the end user by following the Batteries Included approach, isn't that the same? Because having the port set In Arabic will never work, so shouldn't that be handled within Selenium itself?

pujagani:
I understand your point and agree we should have better error handling for this or an informative message. My concern here is that for example if someone wants to use Chinese but now we are setting the locale to US, that is not the expected behavior. Probably, we need to find a different solution or only set the locale to US if it is set to Arabic and probably reset it after start up, not sure what is the correct way to approach this. @titusfortner @diemol

krmahadevan:
@MustafaAgamy - Just curious. Even the [JVM arguments way](https://stackoverflow.com/a/8809137/679824) of setting the locale doesnt help here is it?

MustafaAgamy:
@krmahadevan I didn't try to be honest

MustafaAgamy:
@pujagani

What if we minimize the scope by moving back to String.format with locale set to en for the required parts only, I guess this way the end user's default locale won't get affected where it's not really needed

krmahadevan:
> @krmahadevan I didn't try to be honest

@MustafaAgamy can you give it a try and post back the results?

pujagani:
> @pujagani
> 
> What if we minimize the scope by moving back to String.format with locale set to en for the required parts only, I guess this way the end user's default locale won't get affected where it's not really needed

I think this might be the best thing to do. After understanding the issue properly, my thoughts were on the same line. Keep the scope limited to start up. It might also be good to leave a comment in the code as to why it is done.

pujagani:
> @MustafaAgamy - Just curious. Even the [JVM arguments way](https://stackoverflow.com/a/8809137/679824) of setting the locale doesnt help here is it?

@MustafaAgamy Can you please try this out? If that works, then no code change require and we can document it.

MustafaAgamy:
@pujagani sorry I didnt have access to My laptop for the last couple of days


I'll try today and let you know

pujagani:
@MustafaAgamy No rush. We appreciate your contribution and time to Selenium.

joerg1985:
Just my 2 cent:
I think calling `Locale.setDefault(Locale.US)` is a bad idea, this will change propably change alot of things. 

Why using `String.format` here at all, if we do not want Java to format the String? 
```
// users might have a default local with not latin characters set, String.format would break the argument
args.add("--port=" + getPort());
```

MustafaAgamy:
@pujagani
The JVM Arguments seems to be working

I added them to the VM Arguments In intelliji and the issue seems to be fixed

But we will have to document it for the users

pujagani:
Thank you for trying it out. Can we close this PR then? Can you help document it too?

MustafaAgamy:
We won't need the tests? 
If yes, then I can Remove the changed code lines and only Keep the test files. 
Also where should I document it?

pujagani:
Yes, we can only keep the tests. You can help add the documentation by creating a PR in our website repo https://github.com/SeleniumHQ/seleniumhq.github.io.

MustafaAgamy:
@pujagani
Regarding the documentation of adding the JVM Arguments where should I add it In the documentation repository, or should this question be moved to slack for better discussion?

pujagani:
Can we move this to slack for further discussion please?

diemol:
@MustafaAgamy, please leave only the tests, and then I can merge this.

I would say you can add the docs at https://www.selenium.dev/documentation/webdriver/browsers/

MustafaAgamy:
@diemol

Noted, will do so as soon as I find some free time

MustafaAgamy:
@diemol 
@pujagani 

Sorry for the delay 

Changes removed from the class files and only the tests are kept

I'll work on making the required changes to the documentation asap

Thank you.

pujagani (approved):
Thank you @MustafaAgamy!

MustafaAgamy:
@diemol One last thing 
Regarding the documentation, do you want me to add the note and the changes required in the same tab or do you want me to add a separate file

MustafaAgamy:
@pujagani by the way removing the logic will cause the arabic date tests to fail on the pipeline since we removed the logic added in the class files
Should I add some way to ignore Exception?

diemol:
The Firefox test is failing, and it seems to be making the others fail. Can you please have a look?

Regarding the docs, the same page is fine, thanks.

MustafaAgamy:
@pujagani @diemol 

Fixes to Tests are done they should pass now

diemol (approved):
Thank you, @MustafaAgamy!

Issue #13920, opened 2024-05-08
[🐛 Bug]: OpenQA.Selenium.WebDriverException : no such execution context


MJB222398:
### What happened?

This is an intermittent issue (happening maybe once in 200 tests or so in my production solution).

Opened a RemoteWebDriver instance and attempted to navigate to a URL and got this exception:

```
OpenQA.Selenium.WebDriverException : no such execution context
  (Session info: MicrosoftEdge=123.0.2420.65)
at OpenQA.Selenium.WebDriver.UnpackAndThrowOnError(Response errorResponse, String commandToExecute)
   at OpenQA.Selenium.WebDriver.Execute(String driverCommandToExecute, Dictionary`2 parameters)
   at OpenQA.Selenium.WebDriver.set_Url(String value)
```

### How can we reproduce the issue?

```shell
Can try to make this but won't be straight forward. Wondering if logs give enough to go on?
```

### Relevant log output

```shell
Web Driver Logs:

2024-05-08 11:45:02.881 DEBUG HttpCommandExecutor: Executing command: []: newSession {"capabilities":{"firstMatch":[{"browserName":"MicrosoftEdge","ms:edgeOptions":{"args":["--start-maximized","--disable-dev-shm-usage","--user-data-dir=/home/seluser/selenium/profile","--profile-directory=Default"]}}]}}
2024-05-08 11:45:02.881 TRACE HttpCommandExecutor: >> Method: POST, RequestUri: 'http://localhost:4444/wd/hub/session', Version: 1.1, Content: System.Net.Http.ByteArrayContent, Headers:
{
  Accept: application/json; charset=utf-8
  Content-Type: application/json; charset=utf-8
}
2024-05-08 11:45:03.254 TRACE HttpCommandExecutor: << StatusCode: 200, ReasonPhrase: 'OK', Version: 1.1, Content: System.Net.Http.HttpConnectionResponseContent, Headers:
{
  Cache-Control: no-cache
  Content-Length: 1497
  Content-Type: application/json; charset=utf-8
}
2024-05-08 11:45:03.254 DEBUG HttpCommandExecutor: Response: (bbaa01d40eaea84113032366214e87e8 Success: System.Collections.Generic.Dictionary`2[System.String,System.Object])
2024-05-08 11:45:03.269 DEBUG HttpCommandExecutor: Executing command: [bbaa01d40eaea84113032366214e87e8]: get {"url":"https://mydomain.com/mypage"}
[...]

PR #13978, opened 2024-05-20
[dotnet] Log http requests/responses via internal DiagnosticsHttpHandler

ChrstnAnsl:
### Description
Added Context to support and fix BUG https://github.com/SeleniumHQ/selenium/issues/13920

### Motivation and Context
 I have encountered the same issue and wanting to debug it more but unfortunately there's not enough error logs to let me debug it

### Types of changes

### Checklist

___

### **PR Type**
Bug fix, Enhancement

___

### **Description**
- Added detailed logging for HTTP request and response bodies in `HttpCommandExecutor`.
- Enhanced trace logging to include the content of HTTP requests and responses.
- Aimed to improve debugging capabilities by providing more comprehensive error logs.

___

### **Changes walkthrough** 📝
</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>HttpCommandExecutor.cs</strong><dd><code>Add detailed logging for HTTP request and response bodies</code></dd></summary>
<hr>
      
dotnet/src/webdriver/Remote/HttpCommandExecutor.cs

<li>Added logging for request and response bodies.<br> <li> Enhanced trace logging to include request and response content.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/13978/files#diff-7e90084c9f3e8c994cbbfabdbc93ecd024159b65ca59d9f6efc53cace50e7b10">+8/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

___

> 💡 **PR-Agent usage**:


nvborisenko:
@SeleniumHQ/selenium-committers should we include http request/response body of internal wire protocol to internal logs?

ChrstnAnsl:
> @SeleniumHQ/selenium-committers should we include http request/response body of internal wire protocol to internal logs?

Hi @nvborisenko I'll wait for everyone's opinion on whether we should include HTTP request/response bodies in internal logs. Let's gather everyone's thoughts about this then later on we can decide whether to include it or not. Thanks for looking up

pujagani:
Hey! Thank you for your contribution. However, a typical concern with logging HTTP responses is the body size. For example, the response size might be huge when fetching the web page source. It might be a good idea to log the HTTP response body if the status code is not 2xx. Basically, logging in detail on error situations only.

ChrstnAnsl:
> Hey! Thank you for your contribution. However, a typical concern with logging HTTP responses is the body size. For example, the response size might be huge when fetching the web page source. It might be a good idea to log the HTTP response body if the status code is not 2xx. Basically, logging in detail on error situations only.

That's a great valid point. I will take it into consideration and update the PR accordingly. Thanks @pujagani

ChrstnAnsl:
> > Hey! Thank you for your contribution. However, a typical concern with logging HTTP responses is the body size. For example, the response size might be huge when fetching the web page source. It might be a good idea to log the HTTP response body if the status code is not 2xx. Basically, logging in detail on error situations only.
> 
> That's a great valid point. I will take it into consideration and update the PR accordingly. Thanks @pujagani



> Hey! Thank you for your contribution. However, a typical concern with logging HTTP responses is the body size. For example, the response size might be huge when fetching the web page source. It might be a good idea to log the HTTP response body if the status code is not 2xx. Basically, logging in detail on error situations only.

Hi @pujagani updated the PR. Please see the changes. Thank you!

pujagani:
LGTM. Thank you! @nvborisenko is the C# expert. Please help review it further.

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
```diff
@@ -286,13 +286,25 @@ private async Task<HttpResponseInfo> MakeHttpRequest(HttpRequestInfo requestInfo
...
+                    }
                 }
 
                 using (HttpResponseMessage responseMessage = await this.client.SendAsync(requestMessage).ConfigureAwait(false))
                 {
                     if (_logger.IsEnabled(LogEventLevel.Trace))
                     {
                         _logger.Trace($"<< {responseMessage}");
+
+                        if ((int)responseMessage.StatusCode < 200 || (int)responseMessage.StatusCode > 299)
```
diemol: Why not higher than 399? Do we care about logging redirects?

ChrstnAnsl: Hi @diemol, Great point. Please see the latest commit -- comment fixed. The updated value is now `399`

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
```diff
@@ -286,13 +286,25 @@ private async Task<HttpResponseInfo> MakeHttpRequest(HttpRequestInfo requestInfo
...
 
                 using (HttpResponseMessage responseMessage = await this.client.SendAsync(requestMessage).ConfigureAwait(false))
                 {
                     if (_logger.IsEnabled(LogEventLevel.Trace))
                     {
                         _logger.Trace($"<< {responseMessage}");
+
+                        if ((int)responseMessage.StatusCode < 200 || (int)responseMessage.StatusCode > 399)
+                        {
+                            string responseBody = await responseMessage.Content.ReadAsStringAsync();
```
nvborisenko: Please don't read from response to log something. I believe in execution flow later somebody reads it, I hope it's time to log it as already read body.

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
nvborisenko: The initial variant was correct, redirections are handled by lower `HttpClientHandler`, and are turned on by default (if I am not mistaken). Thus this code should never be executed in case of redirections.

nvborisenko: Wrote this comment and discovered another good option: if we want to log http requests/responses, we can do it in lower level... in `HttpClientHandler`. This way requires more code to be written (easy code), but it is more efficient. @ChrstnAnsl how do you feel comfortable in this topic?

ChrstnAnsl: Just a clarification should Invoke this logger inside of the `CreateHttpClient` or I can create a custom Logging for the `HttpClientHandler`

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
ChrstnAnsl: Removed the logger logic and added it to `HttpClientHandler `instead

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
ChrstnAnsl: Hi @nvborisenko I agree that logging HTTP requests/responses at the HttpClientHandler level does offer better efficiency. It's a good approach if we're looking to minimize overhead and streamline the logging process. If the extra code required for logging at the lower level isn't a concern and we prioritize performance, then implementing it in HttpClientHandler is definitely a solid option. Thanks for bringing it up!

Please see the latest updates. Thanks again!

dotnet/src/webdriver/Remote/ResponseLoggerInterceptor.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// Initializes a new instance of the ResponseLoggerInterceptor with logger.
+        /// </summary>
+        /// <param name="logger">Internal logger to be used for logging responses.</param>
+        public ResponseLoggerInterceptor(ILogger logger)
+        {
+            _logger = logger;
+        }
+
+        /// <summary>
+        /// Intercepts an HTTP response and logs the response body if the status code is not in the 2xx.
```
MJB222398: This summary is not quite correct, it will also not log if the status code is in the 3xxs

dotnet/src/webdriver/Remote/ResponseLoggerInterceptor.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// </summary>
+        /// <param name="response">The HTTP response message to intercept.</param>
+        /// <returns>A task representing the asynchronous operation.</returns>
+        public async Task InterceptAsync(HttpResponseMessage response)
+        {
+            var responseBodyStr = await response.Content.ReadAsStringAsync();
```
MJB222398: Should this line not move inside the if statement?

MJB222398 (commented):
Thanks for doing this, I really want to get to the bottom of #13920 !

dotnet/src/webdriver/Remote/ResponseLoggerInterceptor.cs
ChrstnAnsl: should we move this inside? or should we not, if that's the case I will disregard the comment. Please re-open this if needed.

dotnet/src/webdriver/Remote/ResponseLoggerInterceptor.cs
ChrstnAnsl: Updated. Resolving this comment now

dotnet/src/webdriver/Remote/ResponseLoggerInterceptor.cs
MJB222398: Unless I am missing something it should be inside because we only need the response body inside the if statement. We will be reading the response body every time with the code as it is here even if we don't need it.

nvborisenko:
Stop reviewing it please, this code even cannot be compiled. Does anybody know what is `IHttpInterceptor`? When I mention http requests delegating I mean https://learn.microsoft.com/en-us/dotnet/api/system.net.http.delegatinghandler?view=net-8.0

ChrstnAnsl:
@nvborisenko I see that there's some attributes and files are being ignored and not being pushed this is what's causing the error in the build. I will try another work around for my solution.


UPDATE:

- Apologies for the continuous build error now, I have updated my PR to handle https://learn.microsoft.com/en-us/dotnet/api/system.net.http.delegatinghandler?view=net-8.0. Thanks

piechoo:
Any update with this :) ?

nvborisenko:
Thanks, now it is something closer to be reviewed.

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
```diff
@@ -237,7 +237,9 @@ private void CreateHttpClient()
 
             httpClientHandler.Proxy = this.Proxy;
 
-            this.client = new HttpClient(httpClientHandler);
+            // Use the custom LoggingHandler
+            var loggingHandler = new ResponseLoggerInterceptor(httpClientHandler);
```
nvborisenko: Use `loggingHandler` only when `Trace` log level is enabled. Otherwise use default handler (it is regular pattern for performance).

dotnet/src/webdriver/Remote/ResponseLoggerInterceptor.cs
```diff
@@ -0,0 +1,58 @@
+// <copyright file="ResponseLoggerInterceptor.cs" company="WebDriver Committers">
```
nvborisenko: Usually all classes should be in separated files. I have no strong opinion, probably this class may live in `HttpCommandExecutor` file. It is discussable. I would put this class even into `HttpCommandExecutor` class as iner class and name it `DiagnosticsHttpHanddler`. And for sure make it as `internal`.

dotnet/src/webdriver/Remote/ResponseLoggerInterceptor.cs
```diff
@@ -0,0 +1,58 @@
...
+        public ResponseLoggerInterceptor(HttpMessageHandler innerHandler)
+            : base(innerHandler)
+        {
+        }
+
+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            if (request.Content != null)
+            {
+                string requestContent = await request.Content.ReadAsStringAsync();
```
nvborisenko: - please use `var`
- add `.ConfigureAwait(false)`
- put `cancellationoToken`

dotnet/src/webdriver/Remote/ResponseLoggerInterceptor.cs
```diff
@@ -0,0 +1,58 @@
...
+                _logger.Trace($">> Body: {requestContent}");
+            }
+
+            HttpResponseMessage response = await base.SendAsync(request, cancellationToken);
```
nvborisenko: `var` and `ConfigureAwait(false)` please

dotnet/src/webdriver/Remote/ResponseLoggerInterceptor.cs
```diff
@@ -0,0 +1,58 @@
...
+
+            if ((int)response.StatusCode < 200 || (int)response.StatusCode > 399)
```
nvborisenko: I am not sure, is there any method available like `Is Successful` rather than compare integers in the range?

dotnet/src/webdriver/Remote/ResponseLoggerInterceptor.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                string responseContent = await response.Content.ReadAsStringAsync();
```
nvborisenko: `cancellationToken` and `.ConofigureAwait(flase)`

nvborisenko (changes requested):

dotnet/src/webdriver/Remote/ResponseLoggerInterceptor.cs
ChrstnAnsl: Agree! That’s totally make sense.

dotnet/src/webdriver/Remote/ResponseLoggerInterceptor.cs
ChrstnAnsl: Let me try to check if there’s something like that and get back to you

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
nvborisenko: Reviewing approach with `HttpClientHandler`, thanks for reimplementing.

dotnet/src/webdriver/Remote/ResponseLoggerInterceptor.cs
nvborisenko: `IsSuccessStatusCode` - https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpresponsemessage.issuccessstatuscode?view=net-8.0

ChrstnAnsl: I was about to check on that but thanks for the heads up!

MJB222398: So that considers the request successful if the response code was between 200-299 so different to the logic you have above. But maybe that's what we want?

ChrstnAnsl: yes, we only want to log the responses that aren't `success`.

nvborisenko: Let's apply the easiest way, and later we always can improve if needed.

ChrstnAnsl:
Fix all comments. Please see the latest commit

ChrstnAnsl:
@nvborisenko Fix and consider all the suggestions and comments. 

Please see the latest commit

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
```diff
@@ -385,5 +393,43 @@ private class HttpResponseInfo
...
+            /// Sends the specified request and returns the associated response.
+            /// </summary>
+            /// <param name="request">The request to be sent.</param>
+            /// <param name="cancellationToken">A CancellationToken object to allow for cancellation of the request.</param>
+            /// <returns>The http response message content.</returns>
+            protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+            {
+                if (request.Content != null)
+                {
+                    var requestContent = await request.Content.ReadAsStringAsync().ConfigureAwait(false);
```
nvborisenko, line 422:
```suggestion
                    var requestContent = await request.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
```

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
```diff
@@ -385,5 +393,43 @@ private class HttpResponseInfo
...
+                    _logger.Trace($">> Body: {requestContent}");
+                }
+
+                var response = await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
+
+                if (!response.IsSuccessStatusCode)
+                {
+                    var responseContent = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
```
nvborisenko, line 437:
```suggestion
                    var responseContent = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
```

nvborisenko, line 437: `response.Content` might be `null` here.

nvborisenko:
@ChrstnAnsl this is exactly what we wanted to see and can accept, **thank you**! Please just fix minor issues.

BTW do you want to improve the logic here little bit? Now in logs we will see 2 log entries for http request: the first one is like `>> POST to https://url ...` and the second one is request body content. What if we can merge it and put http request details as a single log entry like:
```
POST to https://url ...
{
  ...
}
```
where `{ ... }` is request body content.

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
ChrstnAnsl: Hi @nvborisenko this seems to be failing in my local when running up the tests. I believe because of the targeted framework `netstandard2.0` 

Failing build (added the cancellationToken as part of the argument)
image

I don't want to update some of the versions that might affect the overall majority of the application's performance. Do we still want to add the `cancellationToken `as an argument in `ReadAsStringAsync `since `ReadAsStringAsync `doesn't have `overload`?

Passing build without the cancellationToken as an argument
image

nvborisenko: My fault, this overload is not available in `netstartaard2.0`

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
nvborisenko: This overload is not available in `netstandadrd2.0`

ChrstnAnsl:
Hi @nvborisenko adjusted the logic and fix the last minor comment. Please see the latest commit. Thank you

nvborisenko:
Let me help here slightly, I am going to push new commits if you don't mind.

ChrstnAnsl:
Sure, go ahead! Thanks

ChrstnAnsl:
Hi @nvborisenko so you did just revert my changes and just retain the `!response.IsSuccessStatusCode && response.Content != null`. It seems like we can cater your suggestion on "updating the logic" later on. But yeah, Thanks for that.

nvborisenko:
I am done, could you please review?

ChrstnAnsl:
Hi @nvborisenko LGTM!

nvborisenko (approved):

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
```diff
@@ -395,5 +392,57 @@ private class HttpResponseInfo
...
+
+                if (request.Content != null)
+                {
+                    var requestContent = await request.Content.ReadAsStringAsync().ConfigureAwait(false);
+                    requestLogMessageBuilder.AppendFormat("{0}{1}", Environment.NewLine, requestContent);
+                }
+
+                _logger.Trace(requestLogMessageBuilder.ToString());
+
+                var response = await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
```
MJB222398: Can we send the request and handle the logging of it in parallel? Don't think we need to await reading the content and logging it before then awaiting sending?

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
ChrstnAnsl: Can we resolve this one?

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
nvborisenko: Micro optimization, but OK. I will do:
- send a request
- put log message
- await response

nvborisenko: Resolved.

ChrstnAnsl: If logging were done in parallel, there is a possibility of interleaving logs from different requests, making it harder to follow the sequence of events. But @nvborisenko already have a commit for it so yeah nvm.

MJB222398: @nvborisenko Sorry I meant send the request (but don't await) and then read the request content etc. and log it. As the key part and the slow part of the method I think is sending the request so if we kick that off first and then handle the logging. And then await the response when logging is done. Maybe its not necessary, up to you.

nvborisenko: Exactly, I implemented that as you said: kick off sending request, do logging, await response.

```csharp
var responseTask = base.SendAsync(request, cancellationToken).ConfigureAwait(false); // kick off sending request

_logger.Trace(requestLogMessageBuilder.ToString()); // do logging

var response = await responseTask; // await response
```

nvborisenko:
In logs we see:

```
10:36:40.134 TRACE HttpCommandExecutor: >> Method: POST, RequestUri: 'http://localhost:37361/session/e44b09c547486f5ea4c198895969e659/url', Version: 1.1, Content: System.Net.Http.ByteArrayContent, Headers:
{
  Accept: application/json; charset=utf-8
  User-Agent: selenium/4.23.0-nightly202406201609
  User-Agent: (.net linux)
  Content-Type: application/json; charset=utf-8
}
{"url":"http://localhost:43323/common/temp/page7006338836466627400.html"}
10:36:40.197 TRACE HttpCommandExecutor: << StatusCode: 200, ReasonPhrase: 'OK', Version: 1.1, Content: System.Net.Http.HttpConnectionResponseContent, Headers:
{
  Cache-Control: no-cache
  Content-Length: 14
  Content-Type: application/json; charset=utf-8
}
```

Looks good to me.

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
MJB222398: @nvborisenko not quite, I said send the request and then read the request content and log it etc. The code as it is now has already read the request before it sends it. I would think that sending the request and awaiting response takes more time than reading the request so I thought it best to send first and then read it and log it while we are waiting. It won't be much faster so completely happy to drop it, just thought it was a good idea.

nvborisenko: Ah, another micro optimization. Sorry, didn't catch it initially. I resolved it right now in `trunk`, please review.

nvborisenko: In any case your suggestions are valid, I am happy to implement it if they don't make code base fat.

MJB222398: Thanks, no worries. I can't see any PRs open about it, maybe I am looking wrong. You can include a link perhaps?

nvborisenko: There is no PRs, I did it directly in trunk -  https://github.com/SeleniumHQ/selenium/commit/aba09e486d3c4f348a0a653d95b5bc94304c5587

MJB222398: Looks decent, thanks. I eagerly await being able to investigate my issue further with this.

PR #14010, opened 2024-05-21
[dotnet] toggle starting test server with Java if Runfiles are not available

titusfortner:
### Description
This allows users to run .NET tests with bazel (both pinned and unpinned browsers), as well as locally with Java using dotnet on Mac. @nvborisenko can you verify this works on Windows? I can spin up a VM if I need to.

This might be more of a hack than the right solution, but it's the best I know how to do. Feedback appreciated.

### Motivation and Context
Current code does not work on Mac to run tests locally

___

### **PR Type**
Enhancement, Bug fix

___

### **Description**
- Improved `Runfiles` handling in `EnvironmentManager` by adding null checks and adjusting logic for `FileNotFoundException`.
- Enhanced `TestWebServer` startup logic:
  - Added fallback to `JAVA_HOME` environment variable.
  - Refactored platform-specific executable path handling.
  - Improved error handling for missing Java executable.
  - Simplified process start logic and environment variable setup.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement
</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>EnvironmentManager.cs</strong><dd><code>Improve `Runfiles` handling in `EnvironmentManager`</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
dotnet/test/common/Environment/EnvironmentManager.cs

<li>Added null check for <code>Runfiles</code> creation.<br> <li> Adjusted logic to handle <code>FileNotFoundException</code> when <code>Runfiles</code> are not <br>available.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14010/files#diff-75436dc8c8dced10f6b96f66e3e3a5d6b31ec7e64ae11a0187b48877d82c3972">+4/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>TestWebServer.cs</strong><dd><code>Enhance `TestWebServer` startup logic and error handling</code>&nbsp; </dd></summary>
<hr>
      
dotnet/test/common/Environment/TestWebServer.cs

[...]

PR #13952, opened 2024-05-16
[dotnet] allow asynchronous command execution

titusfortner:
### Description
This surfaces the ability to allow the driver to execute commands asynchronously by returning a task object representing the async operation in a way that can be awaited in user code.

### Motivation and Context
This is in preparation for implementing an asynchronous solution in .NET. 
This PR is to demonstrate the template for the naming conventions we want to use.
The next step would be to change the async method implementation to be able to use WebDriverBiDi if it is enabled, with the current sync operation as the fallback

### Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

___

### **PR Type**
Enhancement

___

### **Description**
- Introduced asynchronous command execution methods across various classes in the WebDriver .NET bindings.
- Added `ExecuteAsync` in `ICommandExecutor`, `DriverServiceCommandExecutor`, `HttpCommandExecutor`, and `WebDriver` to support async operations.
- Implemented `BackAsync` in `INavigation` and `Navigator` for asynchronous navigation.
- Modified existing synchronous methods to utilize the new asynchronous methods ensuring backward compatibility.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement
</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>ICommandExecutor.cs</strong><dd><code>Add asynchronous command execution interface</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
dotnet/src/webdriver/ICommandExecutor.cs

<li>Added <code>ExecuteAsync</code> method to allow asynchronous command execution.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/13952/files#diff-4b9443bafd5a20a4713027f0f29bd91e6f83649cf92d3cf54f401a1a06639ef7">+9/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
[...]

PR #14012, opened 2024-05-22
[dotnet] implement WebDriver BiDi

titusfortner:
### Status
This is in Draft so we can discuss what the API could look like
The PR is overly complicated because it relies on #13952 & #14010 so those must be merged first; can tidy it up after merging those (and if you want to hold off on reviewing it until then, that's fine, too).

### Description
* Adds the WebDriver BiDi dependency from Jim
* BiDi is disabled in tests by default
* Starts a BiDi session if `webSocketUrl` is set
* Implements all the Navigation commands with conditionals for whether BiDi is present

### Motivation and Context
This should be the easiest way to have working BiDi implementation in .NET since Jim has already written all the lower level commands (and would love not to duplicate that effort if we don't have to)
The conditionals in each method are not the most elegant, but they are straightforward, and I think that's preferred.
It'll be easy enough to just remove the conditionals when we remove support for WebDriver classic (Selenium 6?)

### Considerations
We still need to implement updating `browsingContextId` when windows are switched, and figure out how navigation works if the browser context is a frame (recursively get parent?)

Figure out a better way to toggle BiDi with Bazel; I foresee us running tests with it toggled on and off, and I'm not sure an environment variable is the right way to do it. (@p0deje probably knows how to allow adding a flag for this)

@YevgeniyShunevych your input is appreciated as well


titusfortner:
Ok, I also need to figure out how to add the nuget dependency to:
https://github.com/SeleniumHQ/selenium/blob/trunk/dotnet/paket.bzl
https://github.com/SeleniumHQ/selenium/blob/trunk/dotnet/paket.nuget.bzl
https://github.com/SeleniumHQ/selenium/blob/trunk/dotnet/workspace.bzl

and maybe also some  `BUILD.bazel` files

p0deje:
> Figure out a better way to toggle BiDi with Bazel; I foresee us running tests with it toggled on and off, and I'm not sure an environment variable is the right way to do it. (@p0deje probably knows how to allow adding a flag for this)

We probably want to have both environment variable and a Bazel flag?

```
bazel test --bidi rb/...
bazel test --cdp rb/...
```

titusfortner:
@p0deje I'm more thinking about toggling between WebDriver classic and webDriver BiDi commands. I'm not very interested in toggling between CDP & BiDi, because we should be getting away from CDP. 😁

p0deje:
> Figure out a better way to toggle BiDi with Bazel; I foresee us running tests with it toggled on and off, and I'm not sure an environment variable is the right way to do it. (@p0deje probably knows how to allow adding a flag for this)

Maybe then it would be better to have individual test targets for classic and bidi? It would ensure that we properly cache test results and don't re-run unless necessary?

```
bazel test //rb/spec/integration/selenium/webdriver:driver-firefox-bidi
bazel test //rb/spec/integration/selenium/webdriver:driver-firefox-classic
```

To run all the tests, we'd use:

```
bazel test --test_tag_filters bidi rb/spec/...
bazel test --test_tag_filters classic rb/spec/...
```

titusfortner:
@p0deje I think I like that idea more. I created a separate issue to discuss this here - #14020

titusfortner:
I made a mess of this one and can't seem to fix it, closing and opening a new PR. 😁

Issue #13995, opened 2024-05-21
[🚀 Feature]: Implement Navigation Commands with BiDi


titusfortner:
### Feature and motivation

Moving forward, Selenium plans to update implementations to use WebDriver BiDi instead of WebDriver Classic if BiDi is enabled.

It was decided at the Selenium Dev Summit that we would implement the *navigation* commands with BiDi in all the bindings as the minimum for Selenium 5. This will act as a proof of concept that a user can successfully use either BiDi or Classic depending on the capabilities that are set. We do not need to implement everything in BiDi spec before releasing Selenium 5, just have a working module as a pattern to work from.

### Considerations
* let's not worry about making sure we're storing the browser context id properly everywhere for this issue
* we do need to make sure we're looking at the user's page load strategy to determine how to wait for the navigation

### Implementations
* .NET — #14052

krishnamohan-kothapalli:
Hi, I'd like to work on the **Java implementation** for this issue.

**My plan:**
- Modify the `RemoteNavigation` inner class in `RemoteWebDriver.java` to route `get()`, `back()`, `forward()`, and `refresh()` through BiDi when it's enabled, falling back to Classic otherwise
- Map the driver's `pageLoadStrategy` capability to the appropriate `ReadinessState` (`normal` → `COMPLETE`, `eager` → `INTERACTIVE`, `none` → `NONE`)
- Add integration tests covering the navigation commands with BiDi enabled

Please let me know if this approach looks right or if there's anything I should be aware of before I start. Happy to adjust based on feedback!

shossain786:
I've opened #17715 implementing the **Python** binding for this, following the merged Ruby PR (#14094) and the in-review Java PR (#17232).

It routes the Classic navigation commands through the BiDi `browsingContext` module when BiDi is enabled (`webSocketUrl` present), falling back to Classic otherwise:

| Classic | BiDi |
|---|---|
| `driver.get(url)` | `browsing_context.navigate(context, url, wait=<readiness>)` |
| `driver.back()` | `browsing_context.traverse_history(context, delta=-1)` |
| `driver.forward()` | `browsing_context.traverse_history(context, delta=1)` |
| `driver.refresh()` | `browsing_context.reload(context, wait=<readiness>)` |

`pageLoadStrategy` maps to readiness state (`normal → COMPLETE`, `eager → INTERACTIVE`, `none → NONE`).

[...]

PR #14052, opened 2024-05-29
[dotnet] implement WebDriver BiDi with WebDriverBiDi dependency

titusfortner:
This PR is to evaluate the high level API

### Status
This is in draft because:
* We haven't committed to using WebDriver BiDi package
* WebDriver BiDi package might be changing its implementation

### Description
* Adds the WebDriver BiDi package dependency
* BiDi is disabled in tests by default
* Starts a BiDi session if `webSocketUrl` is set
* Implements all the Navigation commands with conditionals for whether BiDi is present
* Implements PageLoadStrategy waits
* Bazel pieces are all properly wired up now

### Motivation and Context
.NET implementation of https://github.com/SeleniumHQ/selenium/issues/13995
The conditionals in each method are not the most elegant, but they are straightforward, and I think that's preferred.
It'll be easy enough to just remove the conditionals (and probably the Sync methods as well) when we remove support for WebDriver classic (likely Selenium 6)

### Considerations
This sidesteps updating `browsingContextId` from context changes; we'll handle that separately
Still need to sort out creating new bidi bazel targets properly

attn: @YevgeniyShunevych

### DIFF
--- a/dotnet/src/webdriver/Navigator.cs
+++ b/dotnet/src/webdriver/Navigator.cs
@@ -30,6 +30,13 @@ namespace OpenQA.Selenium
     {
         private WebDriver driver;
         private string browsingContextId;
+        private static readonly Dictionary<string, ReadinessState> PageLoadStrategyMapper = new()
+        {
+            {"normal", ReadinessState.Complete},
+            {"eager", ReadinessState.Interactive},
+            {"none", ReadinessState.None}
+        };
+        private ReadinessState readinessState;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Navigator"/> class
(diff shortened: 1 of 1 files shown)


nvborisenko (changes requested):
We should not introduce new external dependencies.

diemol:
@nvborisenko I believe this PR does not make sense anymore, correct?

nvborisenko:
This PR can be considered as actual when it starts using of our BiDi implementation. For me it is interesting how we are going to spread BiDi Driver instance across downstream classes (like `Navigator` class).

nvborisenko:
Now we have own implementation.

nvborisenko:
And ask me, together we will find the best approach.

PR #14073, opened 2024-06-04
[rb] Implement High Level Logging API with BiDi

titusfortner:
### Description
* Implements add and remove methods for websocket callers (there is definitely a better way to do this)
* Deprecates BiDi#callbacks (is there another reason to keep this method?)
* I think this is probably a replacement for LogInspector (which is not the high level API we want, and I don't think we need an intermediary implementation between the socket calls and the high level API)
* Unlike LogInspector, this implementation doesn't do any type enforcement or structs or classes; return values are plain JSON. Do we want to add more structure and/or typing from that implementation into this? 

### Motivation and Context
Partial Ruby implementation of #13992
#13964 

### Updates
Based on feedback:
* using `#object_id` of the proc to return the id instead of tracking it separately
* updated browser context implementation to avoid specifying browser context ids if they don't apply (which they don't for console logs)
* `Script` class is going to implement things from both script and log domains. It makes sense to implement the domain specific functionality in separate, lower level classes and keep `Script` for the high level API, so I created `LogHandler` class for this.
* Created a Struct subclass that will convert camel case JSON keys into snake case methods; using these Structs provides the user with a more straightforward access of declared parameters, without needing to maintain a lot of classes. If we dynamically generate classes in the future, it likely won't change the user's code
* Deprecated `Driver#script` alias of `Driver#exectue_script`
* Deprecated `LogInspector` - this should provide all of the required functionality of LogInspector

___

### **PR Type**
Enhancement, Tests, Documentation

___

### **Description**
- Introduced `LogHandler` class for managing console and JavaScript log entries.
- Deprecated `BiDi#callbacks` and `LogInspector` class.
- Added `add_callback` and `remove_callback` methods to `WebSocketConnection`.
- Created `Script` class to handle log entries and manage handlers.
- Refactored `subscribe` and `unsubscribe` methods in `Session` class to use keyword arguments.
- Added integration tests for `Script` class and BiDi error handling.
- Updated type signatures for `Driver` and `Script` classes.

___

### **Changes walkthrough** 📝
</strong></td><td><details><summary>9 files</summary><table>
<tr>
  <td>
    <details>
      <summary><strong>bidi.rb</strong><dd><code>Enhance BiDi with logging handlers and callback management</code></dd></summary>
<hr>
      
rb/lib/selenium/webdriver/bidi.rb

<li>Added <code>LogHandler</code> and <code>Struct</code> autoloads.<br> <li> Deprecated <code>BiDi#callbacks</code> method.<br> <li> Introduced <code>add_callback</code> and <code>remove_callback</code> methods.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14073/files#diff-b7660c7b53fec20ce95452b2afdbf306e9c36363e5bf8812863375b7bc96e42e">+14/-0</a>&nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>log_handler.rb</strong><dd><code>Introduce LogHandler class for managing log entries</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
rb/lib/selenium/webdriver/bidi/log_handler.rb

<li>Created <code>LogHandler</code> class for managing log entries.<br> <li> Added methods to handle console and JavaScript log entries.<br> <li> Implemented subscription management for log entries.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14073/files#diff-255281b194a6769984ac959abce6e066d00fbfc11e43f8b21363695d0dd652b8">+63/-0</a>&nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>log_inspector.rb</strong><dd><code>Deprecate LogInspector and update callback handling</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
rb/lib/selenium/webdriver/bidi/log_inspector.rb

<li>Deprecated <code>LogInspector</code> class.<br> <li> Updated method to use <code>add_callback</code> instead of direct callback <br>manipulation.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14073/files#diff-c115bc66f1e7633707e716f9070f39216a43bbc942e715d24927c4e6204b4421">+5/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>session.rb</strong><dd><code>Refactor session subscription methods with keyword arguments</code></dd></summary>
<hr>
      
rb/lib/selenium/webdriver/bidi/session.rb

<li>Refactored <code>subscribe</code> and <code>unsubscribe</code> methods to use keyword arguments.<br> <br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14073/files#diff-c5762deba70cf1701e2bef027944b0feca93904bf20a46acdb07a8a09b062530">+6/-6</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>struct.rb</strong><dd><code>Add Struct class for camel to snake case conversion</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
rb/lib/selenium/webdriver/bidi/struct.rb

<li>Created <code>Struct</code> class to convert camel case JSON keys to snake case <br>methods.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14073/files#diff-04c05a628e46aa69c7d088546e665bcdd069b1869bc82d7e0466d4915624f3b7">+40/-0</a>&nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>common.rb</strong><dd><code>Require Script class in common module</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
rb/lib/selenium/webdriver/common.rb

- Added `require 'selenium/webdriver/common/script'`.

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14073/files#diff-5bf2b6715d2e2a368c57757a030c83091c6d52e4db8175f66df8b98b12ac0d7a">+1/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>driver.rb</strong><dd><code>Deprecate Driver#script alias and add Script instance method</code></dd></summary>
<hr>
      
rb/lib/selenium/webdriver/common/driver.rb

<li>Deprecated <code>Driver#script</code> alias for <code>#execute_script</code>.<br> <li> Introduced <code>Driver#script</code> method for accessing <code>Script</code> instance.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14073/files#diff-ecf89969518aec5e73ce3aff22dc10e88185534845a8a17228952c92f1498cdd">+16/-6</a>&nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>script.rb</strong><dd><code>Add Script class for handling log entries</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
rb/lib/selenium/webdriver/common/script.rb

<li>Created <code>Script</code> class to handle console and JavaScript log entries.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14073/files#diff-72c27d09830dafa2065913fa74230f543e1d063b3f2bde3fe1c512192a569a4c">+45/-0</a>&nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>websocket_connection.rb</strong><dd><code>Add callback management methods to WebSocketConnection</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
rb/lib/selenium/webdriver/common/websocket_connection.rb

<li>Added <code>add_callback</code> and <code>remove_callback</code> methods for managing WebSocket <br>callbacks.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14073/files#diff-6675f0e8e985c3acbf5be1124a44431bc20e31f11f7c8054edc572a0e5de66da">+9/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    
</table></details></td></tr><tr><td><strong>Tests
</strong></td><td><details><summary>3 files</summary><table>
<tr>
  <td>
    <details>
      <summary><strong>script_spec.rb</strong><dd><code>Add integration tests for Script class</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
rb/spec/integration/selenium/webdriver/bidi/script_spec.rb

<li>Added integration tests for <code>Script</code> class.<br> <li> Tested console message and JavaScript error handling.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14073/files#diff-511fb8dd07c92b6e7a8767aafc8f3008f7fb18799fd027b9ee9b4e45d9cbea33">+109/-0</a>&nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>bidi_spec.rb</strong><dd><code>Add BiDi error handling and deprecation tests</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
rb/spec/integration/selenium/webdriver/bidi_spec.rb

<li>Added test for BiDi error handling when not enabled.<br> <li> Added deprecation test for <code>BiDi#callbacks</code>.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14073/files#diff-c5ef656609cbe88cee0262b4d90beb709dd351b19e77f4a46ae5b57121532734">+10/-23</a>&nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>driver_spec.rb</strong><dd><code>Add test for Driver#script method with deprecation warning</code></dd></summary>
<hr>
      
rb/spec/integration/selenium/webdriver/driver_spec.rb

- Added test for `Driver#script` method with deprecation warning.

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14073/files#diff-f99e2caaa528d675ddc6279b1f1a74dde50ea5870a14dfffcbd45e6a03140693">+10/-1</a>&nbsp; &nbsp; </td>

</tr>                    
</table></details></td></tr><tr><td><strong>Documentation
</strong></td><td><details><summary>2 files</summary><table>
<tr>
  <td>
    <details>
      <summary><strong>driver.rbs</strong><dd><code>Update type signature for Driver class</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
rb/sig/lib/selenium/webdriver/common/driver.rbs

- Added `@script` instance variable to type signature.

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14073/files#diff-91b084e4330efe2794c17d5b79d203958bd89469087f9455069256e9ec846bda">+1/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>script.rbs</strong><dd><code>Add type signatures for Script class</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
rb/sig/selenium/web_driver/script.rbs

- Added type signatures for `Script` class methods.

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14073/files#diff-6e6e3f4d0c646aad276f827328a9f265e951cba92923ec3ed48fb16dfc84ea2e">+20/-0</a>&nbsp; &nbsp; </td>

</tr>                    

___

> 💡 **PR-Agent usage**:


rb/lib/selenium/webdriver/common/script.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      # @return [int] ID of the handler
+      def add_console_message_handler
+        subscribe_log_entry unless @log_entry_subscribed
+        @bidi.add_callback('log.entryAdded') do |params|
+          yield(params) if params['type'] == 'console'
+        end
+      end
+
+      # @return [int] ID of the handler
```
pujagani: We want to follow this pattern of returning the id of the handler for other languages too right?

rb/lib/selenium/webdriver/common/script.rb
```diff
@@ -0,0 +1,59 @@
...
+      def add_javascript_error_handler
+        subscribe_log_entry unless @log_entry_subscribed
+        @bidi.add_callback('log.entryAdded') do |params|
+          yield(params) if params['type'] == 'javascript'
```
pujagani: For Java, we need to map the returned JSON result to a type so the end-user does not have to figure out how to use it or figure out the structure of the JSON. The IDE will provide suggestions with all the getter methods. I don't know if that applies to Ruby.

rb/spec/integration/selenium/webdriver/bidi/script_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+          expect {
+            driver.script
+          }.to raise_error(WebDriver::Error::WebDriverError, /Script methods requires enabling BiDi/)
+        end
+      end
+
+      it 'logs console messages' do
+        log_entries = []
+
+        driver.script.add_console_message_handler { |log| log_entries << log }
```
pujagani, line 40: I like the design we agreed on :)

rb/lib/selenium/webdriver/common/script.rb
titusfortner: We can't do it in .NET because of the way the native event handlers work.

In Ruby this approach is especially useful because the norm is to work with blocks, which can't get assigned to a variable in a way that it can be used later to remove it.

Maybe it can come back with a type in Java to make it more clear what it is? What do you think?

rb/lib/selenium/webdriver/common/script.rb
titusfortner: Probably the biggest downside of Ruby is that very little is discoverable. Ruby has so many things automatically generated at runtime that you often don't see any suggestions in the IDE. That said, no structure at all is a bit extreme, so there will likely end up being some kind of middle ground to find here. I'll discuss with Alex what is the right balance.

rb/lib/selenium/webdriver/common/driver.rb
```diff
@@ -99,6 +99,20 @@ def navigate
...
       end
 
+      #
+      # @return [Script]
+      # @see Script
+      #
+
+      def script(*args)
+        if args.any?
+          # Maintain legacy alias
```
p0deje: Let's deprecate maybe?

rb/lib/selenium/webdriver/common/driver.rb
```diff
@@ -99,6 +99,20 @@ def navigate
...
+          execute_script(*args)
+        else
+          @script ||= WebDriver::Script.new(bridge)
```
p0deje, line 114: Any reason to pass `bridge` instead of `bridge.bidi`? I don't see bridge used directly, only the bidi accessor.

rb/lib/selenium/webdriver/common/websocket_connection.rb
```diff
@@ -40,6 +40,8 @@ def initialize(url:)
 
         process_handshake
         @socket_thread = attach_socket_listener
+        @track_callbacks = Hash.new { |callbacks, event| callbacks[event] = {} }
```
p0deje: I don't think you need to track callbacks outside of `@callbacks` variable. This would only complicate implementation and double the memory size because we now store passed blocks in 2 different hashes.

What if instead we use `Proc#object_id` as a unique reference to a callback.

```ruby
id = ws.add_callback('foo', { puts _1 }) #=> returns Proc#object_id, e.g. 320
ws.remove_callback('foo', id) #=> calls Enumerable#delete_if { _1.object_id == id }
```

rb/spec/integration/selenium/webdriver/bidi/script_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+        driver.navigate.to url_for('bidi/logEntryAdded.html')
+
+        driver.find_element(id: 'jsException').click
+        driver.find_element(id: 'consoleLog').click
+
+        wait.until { log_entries.any? }
+        expect(log_entries.size).to eq(1)
+        expect(log_entries.first).to include('level' => 'info',
```
p0deje: FYI There were some flaky failures related to Edge I fixed in 062c125c062eea270c9ce8dafcb8f92f32e1bf51. This happened because Edge was emitted "Driver started" console message by itself so I had to ensure we don't just wait for ANY console messages, but wait for specific "Hello, world!" one. You probably need to do the same in these specs too.

rb/lib/selenium/webdriver/common/script.rb
```diff
@@ -0,0 +1,59 @@
...
+        end
+      end
+
+      # @param [int] id of the handler previously added
+      def remove_console_message_handler(id)
+        @bidi.remove_callback('log.entryAdded', id)
```
p0deje: Do we want to unsubscribe when last handler removed?

rb/lib/selenium/webdriver/common/script.rb
p0deje: Ideally we would generate the types for returned objects from BiDi spec, but we're far from there.

rb/lib/selenium/webdriver/common/script.rb
```diff
@@ -0,0 +1,59 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+module Selenium
+  module WebDriver
+    class Script
```
p0deje, line 22: Don't we want to namespace this under `BiDi` module?

rb/lib/selenium/webdriver/common/driver.rb
titusfortner: Bridge has always been the pattern, so seems to make sense to keep it for now even if it is overkill.
There are going to be places in some of these classes where we may need to call window_handle to get current browsing context and get capabilities; seems should have it there to be consistent
also, I'm not sure if there is a use case for `EventFiringBridge`

rb/lib/selenium/webdriver/common/script.rb
titusfortner: How much overhead is it?

rb/lib/selenium/webdriver/common/script.rb
titusfortner: As "High Level API" no
But, I'm starting to think we don't want High Level API to call directly to websocket, and should go through a BiDi namespaced class. Let me resubmit an idea

rb/lib/selenium/webdriver/common/script.rb
titusfortner: I'm going to do a quick implementation with Struct and see how it looks.

rb/lib/selenium/webdriver/common/script.rb
p0deje: No clue how it would work in hours-running tests with multiple scenarios.

rb/lib/selenium/webdriver/common/driver.rb
p0deje: > Bridge has always been the pattern, so seems to make sense to keep it for now even if it is overkill.
Sure, it's internal detail so we can change later.

rb/lib/selenium/webdriver/common/script.rb
titusfortner: Implemented a Struct. I think this is the right compromise between wordiness and flexibility; if we create new objects for these later, the user is unlikely to need to change  anything the because the accessors should remain the same.

rb/lib/selenium/webdriver/common/websocket_connection.rb
titusfortner: Perfect! I hated what I had originally 😂

rb/lib/selenium/webdriver/common/script.rb
titusfortner: ok, added code to check and remove. Easy enough.

rb/lib/selenium/webdriver/common/script.rb
titusfortner: The thing with `Script` class specifically is that it is going to hold implementations for both the `script` and the `log` domains (based on our conversation at the summit). So we do want to call down to something lower level rather than make it be the lower level class.

rb/spec/integration/selenium/webdriver/bidi/script_spec.rb
titusfortner: This is a good point, but I do want us to be as specific as we can be for these. I think if we navigate before adding a handler then we should be safe.

titusfortner:
I have some more tweaks for this, but does it look good enough for now?

p0deje (approved):

PR #14107, opened 2024-06-07
[py] Implement High Level Logging API with BiDi

p0deje:
Partial Python implementation of https://github.com/SeleniumHQ/selenium/issues/13992, closely mimics Ruby part from https://github.com/SeleniumHQ/selenium/pull/14073

### DIFF
--- a/py/conftest.py
+++ b/py/conftest.py
@@ -79,6 +79,13 @@ def pytest_addoption(parser):
         dest="use_lan_ip",
         help="Whether to start test server with lan ip instead of localhost",
     )
+    parser.addoption(
+        "--bidi",
+        action="store",
+        dest="bidi",
+        metavar="BIDI",
+        help="Whether to enable BiDi support",
+    )
 
 
 def pytest_ignore_collect(path, config):
@@ -158,6 +165,20 @@ def driver(request):
         driver_instance = getattr(webdriver, driver_class)(**kwargs)
     yield driver_instance
 
+    # Close the browser after BiDi tests. Those make event subscriptions
+    # and doesn't seems to be stable enough, causing the flakiness of the
+    # subsequent tests.
+    # Remove this when BiDi implementation and API is stable.
+    if bool(request.config.option.bidi):
+
+        def fin():
+            global driver_instance
+            if driver_instance is not None:
+                driver_instance.quit()
+            driver_instance = None
+
+        request.addfinalizer(fin)
+
     if request.node.get_closest_marker("no_driver_after_test"):
         driver_instance = None
 
@@ -166,6 +187,7 @@ def get_options(driver_class, config):
     browser_path = config.option.binary
     browser_args = config.option.args
     headless = bool(config.option.headless)
+    bidi = bool(config.option.bidi)
     options = None
 
     if browser_path or browser_args:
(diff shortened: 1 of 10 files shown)


py/conftest.py
```diff
@@ -187,6 +196,14 @@ def get_options(driver_class, config):
             options.add_argument("--headless=new")
         if driver_class == "Firefox":
             options.add_argument("-headless")
+
+    if bidi:
+        if not options:
+            options = getattr(webdriver, f"{driver_class}Options")()
+
+        if driver_class == "Chrome" or driver_class == "Edge" or driver_class == "Firefox":
```
titusfortner: Do we need to specify this? If a dev tries to run bidi tests on ie it'll give the right failure message instead of ignoring the flag.

rb/spec/integration/selenium/webdriver/bidi/script_spec.rb
```diff
@@ -65,7 +65,7 @@ module WebDriver
         expect(log_entries.size).to eq(2)
       end
 
-      it 'logs removes console message handler' do
+      it 'removes console message handler' do
```
titusfortner, line 68: oops
thanks

titusfortner (approved):
This is fine for tests, but won't be good enough for our documentation:
```
add_console_message_handler(log_entries.append)
```
We should add a test method (I can do it in Ruby) to test filtering, where we only append errors and then click both consoleLog and consoleError buttons. That way we can show how to use lambdas and/or callbacks (and we can just copy those to the docs)

titusfortner:
Yeah, this isn't really testing the code, just showing how to use it, so maybe we don't need it in specs, just in our docs. What does this look like in the python?

```
        driver.navigate.to url_for('bidi/logEntryAdded.html')
        log_entries = []
        driver.script.add_console_message_handler { |log| log_entries << log  if log.level == 'error'}
        driver.find_element(id: 'consoleLog').click
        driver.find_element(id: 'consoleError').click
        wait.until { log_entries.any? }
        expect(log_entries[0].level).to eq('error')
```

py/conftest.py
p0deje: Should be right, I'll remove

p0deje:
>

> Yeah, this isn't really testing the code, just showing how to use it, so maybe we don't need it in specs, just in our docs. What does this look like in the python?
> 
> ```
>         driver.navigate.to url_for('bidi/logEntryAdded.html')
>         log_entries = []
>         driver.script.add_console_message_handler { |log| log_entries << log  if log.level == 'error'}
>         driver.find_element(id: 'consoleLog').click
>         driver.find_element(id: 'consoleError').click
>         wait.until { log_entries.any? }
>         expect(log_entries[0].level).to eq('error')
> ```

The equivalent would one of:

```py
# simple lambda
errors = []
driver.script.add_console_message_handler(lambda log_entry: if log.level == 'error': errors.append(log_entry) }
```
```py
# function
self.errors = []

def collect_error(log_entry)
    if log_entry.level == 'error':
        self.errors.append(log_entry)

driver.script.add_console_message_handler(collect_error}
```

```py
# aggregator object
class ErrorsHandler:
    def __init__():
      self.errors = []

   def collect_error(log_entry)
      if log_entry.level == 'error':
          self.errors.append(log_entry)

errors_handler = ErrorsHandler()
driver.script.add_console_message_handler(errors_handler.collect_error}
```

titusfortner (approved):

AutomatedTester (approved):
LGTM

titusfortner:
@p0deje I didn't realize you added the options in here. I implemented `enable_bidi()` and `web_socket_url` parameters in trunk. You want me to rebase this?

Issue #14087, opened 2024-06-05
[🐛 Bug]: TypeError: WebDriver.__init__() got an unexpected keyword argument 'desired_capabilities'


tscandalitta:
### What happened?

I'm running some automation tests on a webkit based browser

Code:
```python
from selenium.webdriver import WebKitGTK

webdriver_path: str = "/usr/bin/WebKitWebDriver"
driver = WebKitGTK(executable_path=webdriver_path)
driver.get("https://google.com")
```

Error:
```
python3.11 mytest.py 
Traceback (most recent call last):
  File "/tests/mytest.py", line 4, in <module>
    driver = WebKitGTK(executable_path=webdriver_path)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tests/.venv/lib/python3.11/site-packages/selenium/webdriver/webkitgtk/webdriver.py", line 66, in __init__
    super().__init__(
TypeError: WebDriver.__init__() got an unexpected keyword argument 'desired_capabilities'
```
As you can see inspecting `selenium.webdriver.webkitgtk.webdriver.py`, the constructor of RemoteWebDriver does not accept `desired_capabilities`
![Screenshot from 2024-06-05 15-41-05](https://github.com/SeleniumHQ/selenium/assets/38343557/b40bc9ad-e4e7-4c8a-a897-b05cc052a239)

If I remove that param `desired_capabilities=desired_capabilities` I get:
```
python3.11 test_drivers.py 
Traceback (most recent call last):
  File "/tests/mytest.py", line 4, in <module>
    driver = WebKitGTK(executable_path=webdriver_path)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tests/.venv/lib/python3.11/site-packages/selenium/webdriver/webkitgtk/webdriver.py", line 66, in __init__
    super().__init__(
  File "/tests/.venv/lib/python3.11/site-packages/selenium/webdriver/remote/webdriver.py", line 188, in __init__
    capabilities = options.to_capabilities()
                   ^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'to_capabilities'
```
So I have to replace `desired_capabilities=desired_capabilities` with `options=options`

[...]

PR #14128, opened 2024-06-12
[py] remove desired capabilities argument for Webkitgtk

iampopovich:
### **User description**

according to #14087 
i removed capabilities and add default Options same as in safari.webdriver class

### Types of changes

### Checklist

___

### **PR Type**
Bug fix

___

### **Description**
- Removed the `desired_capabilities` argument from the `WebDriver` constructor in `webkitgtk` module.
- Updated the constructor to use `options` directly, simplifying the initialization process.
- Removed the logic for merging `desired_capabilities` with `options` capabilities.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix
</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>webdriver.py</strong><dd><code>Remove <code>desired_capabilities</code> argument and simplify WebDriver <br>initialization.</code></dd></summary>
<hr>
      
py/selenium/webdriver/webkitgtk/webdriver.py

<li>Removed <code>desired_capabilities</code> argument from the <code>WebDriver</code> constructor.<br> <li> Updated the constructor to use <code>options</code> directly.<br> <li> Simplified the initialization process by removing capability merging <br>logic.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14128/files#diff-0d84f899f993a1342a3a2e75a3ef749fb36e08dd022b607a7ab81beda9cfcbfc">+8/-18</a>&nbsp; &nbsp; </td>

</tr>                    

___

> 💡 **PR-Agent usage**:


titusfortner:
This does fix the problem described in the issue, but this class is problematic in several ways, still

It's hard to write this code when we can't have tests because we aren't managing the drivers. 

But I think it needs to look more like:
https://github.com/SeleniumHQ/selenium/blob/trunk/py/selenium/webdriver/wpewebkit/webdriver.py

iampopovich:
> But I think it needs to look more like

I'm trying to understand the differences between the classes if I make wpewebkit and webkitgtk identical. I removed parameters such as port, executable_path, and log path because there are default parameters in the Service class, and other driver binds similarly do not use them. However, in this case, the two classes become identical, and I'm not sure if I correctly understood what is required in the comment above.

classes in the latest commit
Screenshot 2024-06-13 at 22 32 35
classes after I applied changes described above
Screenshot 2024-06-13 at 22 51 28

titusfortner:
It makes sense to me that they are identical if they aren't actually doing anything other than starting/stopping the driver.

titusfortner:
@iampopovich the python linter is failing on some of these changes, can you take a look?

iampopovich:
@titusfortner Running the script `format.sh` did not make any corrections, but during the linting stage, `isort` showed an error. It might be a good idea to add the `isort` run to the `format.sh` script.

titusfortner:
From root run `./go py:lint`

The format script is supposed to work without any additional software installed, but we haven't gotten that figured out for python. If we add the python bit, everyone who doesn't have a python dev environment will get failures. There's an open PR for part of it, we need to dig into it again.

titusfortner (approved):

Issue #14143, opened 2024-06-17
[🐛 Bug]: ValueError: ZIP does not support timestamps before 1980


milahu:
### What happened?

selenium firefox fails to start
if the firefox profile files have timestamp zero = 1970-01-01

https://github.com/SeleniumHQ/selenium/blob/9ce4e666d5ac5c98d14de1f904e06c1a42301a93/py/selenium/webdriver/firefox/firefox_profile.py#L155

fix: add `strict_timestamps=False` ([source](https://stackoverflow.com/a/72391785/10440128))

```py
with zipfile.ZipFile(fp, "w", zipfile.ZIP_DEFLATED, strict_timestamps=False) as zipped:
```

### How can we reproduce the issue?

```shell
find firefox-profile -print0 |
xargs -0 touch -m --date=1970-01-01T00:00:00+0000
```

### Relevant log output

```shell
File "/nix/store/wjlanzvm75kck4nxq0n2mij926ksp3vv-python3.11-selenium-4.17.0/lib/python3.11/site-packages/selenium/webdriver/firefox/webdriver.py", line 69, in __init__
    super().__init__(command_executor=executor, options=options)
  File "/nix/store/wjlanzvm75kck4nxq0n2mij926ksp3vv-python3.11-selenium-4.17.0/lib/python3.11/site-packages/selenium/webdriver/remote/webdriver.py", line 188, in __init__
    capabilities = options.to_capabilities()
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/wjlanzvm75kck4nxq0n2mij926ksp3vv-python3.11-selenium-4.17.0/lib/python3.11/site-packages/selenium/webdriver/firefox/options.py", line 113, in to_capabilities
    opts["profile"] = self._profile.encoded
                      ^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/wjlanzvm75kck4nxq0n2mij926ksp3vv-python3.11-selenium-4.17.0/lib/python3.11/site-packages/selenium/webdriver/firefox/firefox_profile.py", line 160, in encoded
    zipped.write(filename, filename[path_root:])
  File "/nix/store/k3701zl6gmx3la7y4dnflcvf3xfy88kh-python3-3.11.9/lib/python3.11/zipfile.py", line 1795, in write
    zinfo = ZipInfo.from_file(filename, arcname,
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #14168, opened 2024-06-21
[py] property strict_timestamp was added

iampopovich:
### **User description**

### Description

according to #14143 property was added to configure timestamp mode

### Types of changes

### Checklist

___

### **PR Type**
Enhancement, Bug fix

___

### **Description**
- Introduced a new `_strict_timestamps` attribute to the `FirefoxProfile` class to manage timestamp behavior.
- Added getter and setter methods for the `strict_timestamps` property.
- Modified the `encoded` method to utilize the `strict_timestamps` property when creating a `ZipFile`.

___

### **Changes walkthrough** 📝
</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>firefox_profile.py</strong><dd><code>Add strict_timestamps property to FirefoxProfile class</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
py/selenium/webdriver/firefox/firefox_profile.py

<li>Added <code>_strict_timestamps</code> attribute with getter and setter methods.<br> <li> Integrated <code>strict_timestamps</code> property into <code>ZipFile</code> instantiation.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14168/files#diff-323be8fb876362e1b98205768bcd1e386d4607db6a5164932f7b34e091a22cc1">+9/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

___

> 💡 **PR-Agent usage**:


py/selenium/webdriver/firefox/firefox_profile.py
```diff
@@ -53,13 +53,13 @@ def __init__(self, profile_directory=None):
            This defaults to None and will create a new
            directory when object is created.
         """
+        self._strict_timestamps = False
```
titusfortner: This isn't a backwards compatible change, and this is the first time someone has reported an issue. Let's keep current behavior and allow users to toggle it if it causes them a problem.

py/selenium/webdriver/firefox/firefox_profile.py
```diff
@@ -53,13 +53,13 @@ def __init__(self, profile_directory=None):
...
         self._desired_preferences = {}
         if profile_directory:
-            newprof = os.path.join(tempfile.mkdtemp(), "webdriver-py-profilecopy")
+            self._profile_dir = os.path.join(tempfile.mkdtemp(), "webdriver-py-profilecopy")
             shutil.copytree(
-                profile_directory, newprof, ignore=shutil.ignore_patterns("parent.lock", "lock", ".parentlock")
+                profile_directory, self._profile_dir, ignore=shutil.ignore_patterns("parent.lock", "lock", ".parentlock")
```
titusfortner: I think we want to keep old logic here. The intent is to copy the directory then assign _profile_dir to the new location.

titusfortner (changes requested):
Thanks for adding this

py/selenium/webdriver/firefox/firefox_profile.py
iampopovich: I reverted the changes. I don't understand the need to use an intermediate variable, but it seems minor.

py/selenium/webdriver/firefox/firefox_profile.py
iampopovich: I reverted the change back to the default value of True. As I described in the bug comments, I believe that the client code should work the same for both True/False values for dates after 1980.

py/selenium/webdriver/firefox/firefox_profile.py
titusfortner: I'm not certain either, stuff like this I assume it fixed someone's bug at some point and consider it a "Chesterton's Fence."  I'd look at commit history if it was a big deal or if if it seemed to cause performance issues. In general people shouldn't be relying on Firefox Profiles any longer (the whole class/approach was deprecated at one point, but I removed that because it is still technically supported). We do need to do a lot more deprecation work in these (there's an open issue somewhere).

py/selenium/webdriver/firefox/firefox_profile.py
titusfortner: Ok, I got bad information about this change, and then didn't look deeper (and/or read the conversation in that issue carefully). I was given the impression that the change meant it wasn't guaranteeing time stamp information, which appears not to be the case.

So, the only non-backwards compatible concern is if someone for some reason is relying on the fact that they will get an exception if files before 1980 are being zipped. That does not seem like a reasonable concern for Selenium users. In which case even the toggle is excessive. So. Yeah, let's remove the toggle and just set to true. 

Thanks, and sorry for the extra hassle.

py/selenium/webdriver/firefox/firefox_profile.py
iampopovich: I suggest leaving this piece of code as it is. If the FirefoxProfile functionality is already deprecated, it may eventually be removed on its own in the future. For now, it indeed seems minor. I just noticed the long chain of assignments.

py/selenium/webdriver/firefox/firefox_profile.py
milahu: > let's remove the toggle and just set to true.

i guess you mean false in

```py
zipfile.ZipFile(fp, "w", zipfile.ZIP_DEFLATED, strict_timestamps=False)
```

> this is the first time someone has reported an issue

this happens when using firefox from the [nix](https://github.com/NixOS/nix) store
where all files have timestamp zero = 1970-01-01

py/selenium/webdriver/firefox/firefox_profile.py
```diff
@@ -53,6 +53,7 @@ def __init__(self, profile_directory=None):
            This defaults to None and will create a new
            directory when object is created.
         """
+        self._strict_timestamps = True
```
titusfortner:
```suggestion
        self._strict_timestamps = False
```

py/selenium/webdriver/firefox/firefox_profile.py
```diff
@@ -93,6 +94,13 @@ def update_preferences(self):
                 f.write(f'user_pref("{key}", {json.dumps(value)});\n')
 
     # Properties
+    @property
```
titusfortner: Let's just remove all of this, don't need to maintain an extra property for something that is backwards compatible.

titusfortner (changes requested):

py/selenium/webdriver/firefox/firefox_profile.py
titusfortner: Yes, `False`, and makes sense

py/selenium/webdriver/firefox/firefox_profile.py
iampopovich: Okay, now the parameter `strict_timestamps = False` is simply passed in the `encode` property

py/selenium/webdriver/firefox/firefox_profile.py
```diff
@@ -53,6 +53,7 @@ def __init__(self, profile_directory=None):
            This defaults to None and will create a new
            directory when object is created.
         """
+
```
milahu: remove diff noise in lines 56 and 96

milahu:
ok. squash commits as something like
`set strict_timestamps=False in zipfile.ZipFile`
or
`disable strict timestamps in ZipFile`

titusfortner:
We squash as part of the GitHub merge process. Hmm, I'm not sure why that chrome test is failing in this PR but not on trunk, but at least it is obviously unrelated.

titusfortner (approved):

titusfortner:
Thanks @milahu & @iampopovich; sorry that I made a one-line change more difficult than it needed to be.

Issue #14184, opened 2024-06-24
[🐛 Bug]: Missing dependency websocket-client in Python source package


rolim:
### What happened?

When installing the source distribution of selenium-4.22.0, the required dependency "websocket-client>=1.8.0" is missing.

### How can we reproduce the issue?

```shell
$ pip3 install selenium --no-binary selenium
$ python3 -c "from selenium import webdriver"
```

### Relevant log output

```shell
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/tmp/venv/lib/python3.12/site-packages/selenium/webdriver/__init__.py", line 20, in <module>
    from .chrome.webdriver import WebDriver as Chrome  # noqa
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/venv/lib/python3.12/site-packages/selenium/webdriver/chrome/webdriver.py", line 18, in <module>
    from selenium.webdriver.chromium.webdriver import ChromiumDriver
  File "/tmp/venv/lib/python3.12/site-packages/selenium/webdriver/chromium/webdriver.py", line 22, in <module>
    from selenium.webdriver.remote.webdriver import WebDriver as RemoteWebDriver
  File "/tmp/venv/lib/python3.12/site-packages/selenium/webdriver/remote/webdriver.py", line 67, in <module>
    from .websocket_connection import WebSocketConnection
  File "/tmp/venv/lib/python3.12/site-packages/selenium/webdriver/remote/websocket_connection.py", line 23, in <module>
    from websocket import WebSocketApp
ModuleNotFoundError: No module named 'websocket'
```

### Operating System

Ubuntu

### Selenium version

4.22.0

### What are the browser(s) and version(s) where you see this issue?

n/a

### What are the browser driver(s) and version(s) where you see this issue?

n/a

titusfortner:
Ah, we need to add it to the  install requires section of the setup.py file

Delta456:
I believe this issue can be closed because #14187 is merged.

PR #14187, opened 2024-06-25
[py] websocket-client v.1.8.0 was added to setup.py

iampopovich:
### **User description**

### Description

according to request in #14184 websocket-client was added to setup.py
version of client the same as in py/requirements.txt

### Types of changes

### Checklist

___

### **PR Type**
Bug fix, Dependencies

___

### **Description**
- Added `websocket-client` version 1.8.0 to the `install_requires` list in `setup.py`.
- This change addresses the requirement specified in issue #14184.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Dependencies
</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>setup.py</strong><dd><code>Add websocket-client dependency to setup.py</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
py/setup.py

<li>Added <code>websocket-client</code> version 1.8.0 to the <code>install_requires</code> list.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14187/files#diff-722aeeec2821793eb6adf0c529cc7439c4b27ce78937cbb8840e94d4fc1c4017">+1/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

___

> 💡 **PR-Agent usage**:


iampopovich:
@titusfortner I also noticed that the build script `py/BUILD.bazel` uses a loose requirement for the version of `websocket-client`. Should we enforce version 1.8.0 consistently across the entire project?
Screenshot 2024-06-25 at 18 26 49

titusfortner:
I agree they should all probably match. 

I don't know how python does things. In ruby if it is a transitive dependency, we leave the requirements loose to give the user more flexibility. I'm not sure if that matters here.

iampopovich:
>  I don't know how python does things. 

It depends on the build system.

With `pip`, if there's a conflict between libraries, the package manager will attempt to install the most suitable dependency that satisfies all requirements and avoids conflicts. Otherwise, developers will need to resolve dependency conflicts manually. In our case bazel uses pip to install requirements as it described [here](https://github.com/iampopovich/selenium/blob/cd96b6233d12e1d61addf8479c7d5acf1f76d1b3/py/BUILD.bazel#L55)

`poetry`, creates groups for dependencies, installing only versions that satisfy its requirements.

The more loose requirements there are, the more flexibility the build system has to assemble a "working" configuration without conflicts. Managing a strictly defined set of dependencies that are tested thoroughly during build and testing phases makes configuration management easier. Updating dependencies becomes a more manageable process.


Potentially, we might encounter version conflicts if other dependencies require `websocket-client` >= 1.8.0. I suggest trying to use version 1.8.0 and observing the build results.
@titusfortner cc

iampopovich:
I also noticed that the `setup.py` file does not include packages introduced after 2011, such as Safari (introduced in 2015). Should we add the missing packages to the `packages` section? And webdriver.support introduced twice 
Screenshot 2024-06-25 at 20 23 19

titusfortner:
Python has had a lot of different fingers in the pot over the years, it is not surprising that there are things that have been missed as different people add things. Yes, please, update the things that make sense. Thanks.

iampopovich:
@titusfortner  I made changes to match dependencies in setup.py and BUILD.bazel. I think we should freeze dependencies to specific versions separately if needed later on. Right now, there are differences in library versions: for instance, `py/BUILD.bazel `uses 'trio~=0.17' in the 'requires' section, but `py/requirements.txt` uses 'trio>=0.20.2'. Maybe we should set up a single place for dependencies to get consistent information.

titusfortner:
What do you suggest? I wish I knew more about  bazel python rules in general.

iampopovich:
@titusfortner I propose to accept the current pull request as is. Improvements for the build and dependencies will be considered later in a separate task. I also need to delve deeper into Bazel build rules

titusfortner:
@iampopovich do these happen as part of your tooling or work flow?

> [Merge branch 'trunk' into fix-14184-adding-websocket-client](https://github.com/SeleniumHQ/selenium/pull/14187/commits/4e407d02b2d6175c25264a2db5f9ba656e928bee)

I can't see the results of the tests whenever an update is made 😁

iampopovich:
@titusfortner I updated the branch to pull in the latest changes from trunk. If it's not necessary, I'll update the branch locally only.

titusfortner:
GitHub will automatically squash & rebase for us, so if there aren't any changes that are likely to cause an issue, it isn't needed to keep merging trunk. It's only an issue because I want to merge it and it shows tests haven't run. 😁 

Thanks for the code!

PR #14242, opened 2024-07-09
[dotnet] quit fails after not successful new session

Indomitable:
### **User description**
If new session creation fails a Quit() method is called which calls Dispose and then Quit command, unfortunately in this case session is still not created and it throws null reference exception and this hides the actual exception

### Types of changes

### Checklist

This PR fixes a bug which calls Quit command after trying to create new session. If session fails it calls the Quit method in order to dispose the driver and the commander, unfortunately since the session id is null the Execute Quit command fails with null reference exception. 
For reference in Java client this is solved with another try/catch when calling `quit()` but I think checking  the session id if  null is better solution.

___

### **PR Type**
Bug fix

___

### **Description**
- Added a null check for `sessionId` before executing the Quit command in `Dispose` method to prevent null reference exceptions when session creation fails.
- Ensures that the actual exception is not hidden by a subsequent null reference exception.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>WebDriver.cs</strong><dd><code>Add null check for sessionId before executing Quit command</code></dd></summary>
<hr>

<li>Added a null check for <code>sessionId</code> before executing the Quit command.<br> <li> Prevents null reference exception when session creation fails.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14242/files#diff-ef1874c09b1dca0d62ea1f0ead685355944b9cb7ae63df60ad9b32abdf90c8dc">+4/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

___

> 💡 **PR-Agent usage**:

### DIFF
--- a/dotnet/src/webdriver/WebDriver.cs
+++ b/dotnet/src/webdriver/WebDriver.cs
@@ -703,7 +703,10 @@ namespace OpenQA.Selenium
         {
             try
             {
-                this.Execute(DriverCommand.Quit, null);
+                if (this.sessionId is not null)
+                {
+                    this.Execute(DriverCommand.Quit, null);
+                }
             }
             catch (NotImplementedException)
             {
--- a/dotnet/src/webdriver/WebDriver.cs
+++ b/dotnet/src/webdriver/WebDriver.cs
@@ -62,8 +62,15 @@ namespace OpenQA.Selenium
             }
             catch (Exception)
             {
-                // Failed to start driver session, disposing of driver
-                this.Quit();
+                try
+                {
+                    // Failed to start driver session, disposing of driver
+                    this.Quit();
+                }
+                catch
+                {
+                    // Ignore the clean-up exception. We'll propagate the original failure.
+                }
                 throw;
             }
 
--- a/java/src/org/openqa/selenium/chromium/ChromiumOptions.java
+++ b/java/src/org/openqa/selenium/chromium/ChromiumOptions.java
@@ -31,7 +31,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.stream.Stream;
@@ -120,24 +119,12 @@ public class ChromiumOptions<T extends ChromiumOptions<?>>
    *       "allow-outdated-plugins");
    * </code></pre>
    *
-   * <p>Each argument may contain an option "--" prefix: "--foo" or "foo". Arguments with an
+   * <p>Each argument may contain an optional "--" prefix: "--foo" or "foo". Arguments with an
    * associated value should be delimited with an "=": "foo=bar".
    *
    * @param arguments The arguments to use when starting Chrome.
    */
   public T addArguments(List<String> arguments) {
-    /*
-     --remote-allow-origins is being added by default since Chrome 111. We need to check
-     if the argument already exists and then remove it.
-    */
-    String remoteAllowOrigins = "remote-allow-origins";
-    Optional<String> newArg =
-        arguments.stream().filter(arg -> arg.contains(remoteAllowOrigins)).findFirst();
-    Optional<String> existingArg =
-        args.stream().filter(arg -> arg.contains(remoteAllowOrigins)).findFirst();
-    if (newArg.isPresent() && existingArg.isPresent()) {
-      args.remove(existingArg.get());
-    }
     args.addAll(arguments);
     return (T) this;
   }


nvborisenko:
@Indomitable please share the exception with stacktrace you faced and you are trying to fix. Probably we may propose better solution.

Indomitable:
@nvborisenko  Here is my input:
The exception thrown is NullReferenceException. What is the case. 
The request for `StartSession` throws exception
https://github.com/SeleniumHQ/selenium/blob/6b28a8c03feb0c4121d07299a18444dcbc34b35e/dotnet/src/webdriver/WebDriver.cs#L651
this leads to calling the Quit() in the catch https://github.com/SeleniumHQ/selenium/blob/6b28a8c03feb0c4121d07299a18444dcbc34b35e/dotnet/src/webdriver/WebDriver.cs#L66 
which calls `Dispose` and then the `Quit` command
https://github.com/SeleniumHQ/selenium/blob/6b28a8c03feb0c4121d07299a18444dcbc34b35e/dotnet/src/webdriver/WebDriver.cs#L706
but since the sessionId is not initialized yet , a NullReferenceException is thrown when tries to build the URL for the quit command
https://github.com/SeleniumHQ/selenium/blob/6b28a8c03feb0c4121d07299a18444dcbc34b35e/dotnet/src/webdriver/HttpCommandInfo.cs#L128
As I wrote in the Java client this is fixed using second `try/catch` for the quit method 
https://github.com/SeleniumHQ/selenium/blob/6b28a8c03feb0c4121d07299a18444dcbc34b35e/java/src/org/openqa/selenium/remote/RemoteWebDriver.java#L160-L170

nvborisenko:
I understand the flow how it is happening. I am curious in the stracktrace (what you see in your output).

My thoughts:
- I see invoking of `Dispose` in `ctor`, it is not good
- I see `NullReferenceException` which probably can be avoided in some another way rather than `try/catch`

Indomitable:
Here is the stack trace. I'm using the Appium.NET library which uses Selenium 
Actually it looks like I'm a bit of wrong. Since the SessionId is empty it tries to make a `DELETE /session/` request. 
The result is empty body but with content type of `application/json` and then in the `FromErrorJson` it can not convert the body to Dictionary. 
```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at OpenQA.Selenium.Response.FromErrorJson(String value)
   at OpenQA.Selenium.Remote.HttpCommandExecutor.CreateResponse(HttpResponseInfo responseInfo)
   at OpenQA.Selenium.Remote.HttpCommandExecutor.Execute(Command commandToExecute)
   at OpenQA.Selenium.Appium.Service.AppiumCommandExecutor.Execute(Command commandToExecute)
   at OpenQA.Selenium.Appium.AppiumDriver.Execute(String driverCommandToExecute, Dictionary`2 parameters)
   at OpenQA.Selenium.WebDriver.Dispose(Boolean disposing)
   at OpenQA.Selenium.WebDriver.Dispose()
   at OpenQA.Selenium.WebDriver.Quit()
   at OpenQA.Selenium.Appium.AppiumDriver..ctor(ICommandExecutor commandExecutor, ICapabilities appiumOptions)
   at OpenQA.Selenium.Appium.AppiumDriver..ctor(Uri remoteAddress, ICapabilities appiumOptions, TimeSpan commandTimeout, AppiumClientConfig clientConfig)
   at OpenQA.Selenium.Appium.AppiumDriver..ctor(Uri remoteAddress, ICapabilities appiumOptions, TimeSpan commandTimeout)
   at OpenQA.Selenium.Appium.AppiumDriver..ctor(Uri remoteAddress, ICapabilities appiumOptions)
   at OpenQA.Selenium.Appium.iOS.IOSDriver..ctor(Uri remoteAddress, DriverOptions driverOptions)
```

Indomitable:
So it depends from the server. In this case the Appium server returns empty body and we get json deserialization issue. One more problem that I see is that a `DELETE /session/` call is made which may result to undefined behavior.

nvborisenko:
@Indomitable  Ok, seems I am right that `Dispose()` in `ctor` is unnecessary. But, I don't want to introduce old/new issues :D

I propose to align with java implementation, probably they know hidden stones: https://github.com/SeleniumHQ/selenium/blob/785914e555cfbb2733eb9626f439828b6469a57d/java/src/org/openqa/selenium/remote/RemoteWebDriver.java#L160

So in `ctor` let's try/catch `Quit()` invocation and that's it! Your current changes, you made, are also acceptable.

Indomitable:
PR is updated with the try/catch around the Quit.

nvborisenko (approved):
Thanks for your contribution!

Issue #14045, opened 2024-05-28
[🚀 Feature]: Use Environment Variables to set driver locations


titusfortner:
### Feature and motivation

We discussed in the Selenium Dev Summit about having the ability to set driver location system-wide using environment variables, similar to how Java currently uses System Properties.

See #13806 

### Usage example

If we match Java System Properties, it would look like:
```
os.environ["WEBDRIVER_GECKO_DRIVER"] = path_to_geckodriver
```

pujagani:
This is a nit. But why is Java label added to this issue? I might be missing something.

diemol:
The idea is to leave a path for users who want to avoid Selenium Manager completely. The decision was to implement an env var that the service would use. I guess we can keep the system properties in Java instead of implementing the env var.

titusfortner:
I included it intentionally because perhaps Java should look for the System Property *and* the Environment Variable?
Is there a scenario where you have a VM where you want to run multiple bindings and only set things one way? /shrug

titusfortner:
Actually, which convention do we want to match?

For instance in Java:
System Property: `webdriver.gecko.driver`
Java wrapper for property: `GeckoDriverService.GECKO_DRIVER_EXE_PROPERTY`

Options:
Match property convention: `WEBDRIVER_GECKO_DRIVER`
Match env convention: `SE_GECKODRIVER`

pujagani:
I am in favour of having a consistent way across bindings, if that is possible.

pujagani:
`Match env convention: SE_GECKODRIVER ` I think this. But I don't have any strong opinions about this. I would prefer that this is the convention we use for other env variables example: SE_EDGE_PATH. It is might be easier for users to identify/set Selenium related environment variables.

aguspe:
@pujagani @titusfortner I'm working on this hopefully I will put it for review this weekend :)

p0deje:
#14287 added Ruby part and the following variables:

- `SE_CHROMEDRIVER`
- `SE_EDGEDRIVER`
- `SE_GECKODRIVER`
- `SE_IEDRIVER`
- `SE_SAFARIDRIVER`

Delta456:
I can work on the Python functionality.

Delta456:
The python part has also been implemented. In https://github.com/SeleniumHQ/selenium/commit/a2cacc197ff0abd7b74b1025bff88090aafe3b37 and uses the same environment variables as Ruby.

[...]

PR #14287, opened 2024-07-20
[rb] Allow driver path to be set using ENV variables

aguspe:
### Description
I added a new method on the common service class that sets the path equal to the value of the environment variable for the respective driver or returns nil

### Motivation and Context
As explained in #14045 the goal of this PR is to provide users with an alternative option to set the path to their own drivers if they do not want to use selenium manager

### Types of changes

### Checklist


aguspe:
The only test failing is an unrelated Edge test, which I will create a PR for but I need #14433 to be merged first

rb/spec/integration/selenium/webdriver/chrome/service_spec.rb
```diff
@@ -37,6 +37,19 @@ module Chrome
         it 'can be started outside driver' do
           expect(service_manager.uri).to be_a(URI)
         end
+
+        it 'uses the path from the environment' do
+          ENV['SE_CHROMEDRIVER'] = DriverFinder.new(Options.new, described_class.new).driver_path
```
p0deje: This would leak into all subsequent specs - you need to clear `ENV` properly after the test is done.

rb/spec/integration/selenium/webdriver/chrome/service_spec.rb
```diff
@@ -37,6 +37,19 @@ module Chrome
...
+
+          expect(service_manager.uri).to be_a(URI)
```
p0deje: Can we have a more specific assertion to ensure it's read correctly?

p0deje (commented):
Few comments from my side:

1. What do you think about making the code more explicit rather than relying on class names? For example, we can make every service class have a constant specifying which environment variable can be used to override the finder behavior. This would make it easier for anyone looking at the code and we would be able to document it.

```ruby
module Selenium
  module WebDriver
    module Safari
      class Service
        # You can explicitly specify the location of the driver
        # using SE_SAFARIDRIVER="..." environment variable.
        DRIVER_PATH_ENV_KEY = "SE_SAFARIDRIVER"
```

2. Looking at the specs, I feel like they belong more to the unit tests, rather than integration ones because they don't need to start the driver. Can you update accordingly?

rb/lib/selenium/webdriver/common/service.rb
```diff
@@ -87,16 +88,25 @@ def initialize(path: nil, port: nil, log: nil, args: nil)
...
+
+      def find_driver_path
+        default_options = WebDriver.const_get("#{self.class.name&.split('::')&.[](2)}::Options").new
+        DriverFinder.new(default_options, self).driver_path
+      end
+
+      def env_path
+        class_name = self.class.name&.split('::')&.[](2)&.downcase
+        driver_name = class_name == 'firefox' ? 'gecko' : class_name
+        parsed_driver = "SE_#{driver_name&.upcase}DRIVER"
```
luke-hill: I think edge is referred to as msedgedriver. So it might need more guards. But I'm not 100% on this.

aguspe:
> ```ruby
> DRIVER_PATH_ENV_KEY = "SE_SAFARIDRIVER"
> ```

1 - That's a great idea with the constant to have the env variable, I just changed that
2 - I think this is more of an integration test also because when I try to run it as a unit test locally I have this error related to selenium manager:

```
0) Selenium::WebDriver::Chrome::Service when initializing driver with a path env variable updates the path after setting the environment variable
     Failure/Error: raise Error::NoSuchDriverError, "Unable to obtain #{exe}"

     Selenium::WebDriver::Error::NoSuchDriverError:
       Unable to obtain chromedriver; For documentation on this error, please visit: https://www.selenium.dev/documentation/webdriver/troubleshooting/errors/driver_location
     # ./lib/selenium/webdriver/common/driver_finder.rb:75:in `rescue in paths'
     # ./lib/selenium/webdriver/common/driver_finder.rb:49:in `paths'
     # ./lib/selenium/webdriver/common/driver_finder.rb:39:in `driver_path'
     # ./spec/unit/selenium/webdriver/chrome/service_spec.rb:124:in `block (4 levels) in <module:Chrome>'
     # ------------------
     # --- Caused by: ---
     # Selenium::WebDriver::Error::WebDriverError:
     #   not a file: "/Users/apeque01/Desktop/main_folder/Projects/open_source/selenium/rb/bin/macos/selenium-manager"
     ```

rb/lib/selenium/webdriver/common/service.rb
aguspe: Let's see now with the new env variables because last run the ie driver test did not fail, so I want to double check

p0deje:
> I think this is more of an integration test also because when I try to run it as a unit test locally I have this error related to selenium manager:

Yes, if you want to test Selenium Manager - it would be an integration test. However, if you only want to test the logic for fetching driver path via `ENV` - it should be enough to write a unit test.

aguspe:
> > I think this is more of an integration test also because when I try to run it as a unit test locally I have this error related to selenium manager:
> 
> Yes, if you want to test Selenium Manager - it would be an integration test. However, if you only want to test the logic for fetching driver path via `ENV` - it should be enough to write a unit test.

That's a great point, I moved everything as a unit test know because you are right the main core of this test is to fetch the ENV value

I ran all the unit tests locally and all seem to be good so it's ready for re-review @p0deje

diemol:
Are there docs for this?

p0deje:
> Are there docs for this?

Not really, we probably need to add it somewhere in https://www.selenium.dev/documentation/selenium_manager/?

aguspe:
@p0deje @diemol I can create a PR to update the docs tonight :) I will ping it here too

diemol:
@aguspe do you have time to add docs for this?

diemol:
@aguspe I wrote a comment and I did not see your reply, sorry!

aguspe:
Sorry for the delay but here is the PR for the doc update: https://github.com/SeleniumHQ/seleniumhq.github.io/pull/1950

Let me know if any changes or more info is needed! 
I did not add any translations since I can only help with Spanish, Italian and Danish

PR #14352, opened 2024-08-07
[JS]: Handle optional dependency for @bazel/runfiles

harsha509:
### Description
In an npm project, requiring `@bazel/runfiles` with `const { runfiles } = require('@bazel/runfiles')` throws an exception because `@bazel/runfiles` is not available outside of Bazel builds.

The fix safely attempts to require @bazel/runfiles and logs a message if it fails.

### Types of changes

### Checklist

___

### **PR Type**
Bug fix

___

### **Description**
- Removed direct import of `@bazel/runfiles` which caused exceptions in non-Bazel environments.
- Added a try-catch block to safely attempt requiring `@bazel/runfiles`.
- Logged informative error messages if `@bazel/runfiles` is not available.
- Set `runfiles` to null if the module is not available to prevent further issues.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>index.js</strong><dd><code>Handle optional dependency for `@bazel/runfiles` in tests</code></dd></summary>
<hr>

javascript/node/selenium-webdriver/testing/index.js

<li>Removed direct import of <code>@bazel/runfiles</code>.<br> <li> Added a try-catch block to optionally require <code>@bazel/runfiles</code>.<br> <li> Logged error messages if <code>@bazel/runfiles</code> is not available.<br> <li> Set <code>runfiles</code> to null if the module is not available.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14352/files#diff-a73b91b68f7208515f860a1e2558954d2c8cfbb516a78bf0e272082da96f079f">+12/-1</a>&nbsp; &nbsp; </td>

</tr>                    

___

> 💡 **PR-Agent usage**:


pujagani:
I have faced this few times. I think this looks good to me.

javascript/node/selenium-webdriver/testing/index.js
```diff
@@ -46,6 +45,19 @@ const { Browser } = require('../lib/capabilities')
...
+let runfiles
+try {
+  // Attempt to require @bazel/runfiles
+  runfiles = require('@bazel/runfiles').runfiles
+} catch (error) {
+  // Handle error if @bazel/runfiles is not supported by mocha
+  const errorMessage = `Error requiring @bazel/runfiles: ${error.message}\n
+   Note: If you are running tests with Mocha or Jasmine, this module is not needed.\n
+   For more details, see: https://github.com/bazelbuild/rules_nodejs/issues/3770`
+  console.error(errorMessage)
```
shs96c: We don't need to log a message. For most users, this will just be noise they don't need

javascript/node/selenium-webdriver/testing/index.js
```diff
@@ -46,6 +45,19 @@ const { Browser } = require('../lib/capabilities')
...
+  runfiles = null // Set to null if not available
```
shs96c: Do we properly wrap all usages of `runfiles` later in this file? Maybe we do a check before we use it, and exit quickly if we know it'll never succeed?

javascript/node/selenium-webdriver/testing/index.js
harsha509: Removed!

javascript/node/selenium-webdriver/testing/index.js
harsha509: Added a check for `runfiles` and implemented a safe, silent exit if not found.

javascript/node/selenium-webdriver/testing/index.js
```diff
@@ -540,6 +547,11 @@ function getTestHook(name) {
 }
 
 function locate(fileLike) {
+  if (!runfiles) {
+    // eslint-disable-next-line n/no-process-exit
+    process.exit(1)
```
shs96c: If we're not running in Bazel, we should try the `fs.existsSync` path, and if that fails then return an error. We don't want to exit the process.

The mental model I'm thinking of it:

1. In all cases, check the local disk
2. If we're running in bazel, also check bazel's runfiles

harsha509: HI Simon,
Updated as suggested!

shs96c (approved):
LGTM! Thank you!

PR #14367, opened 2024-08-09
[bidi][java] ensure session.unsubscribe is done when the last listener is removed by id

joerg1985:
### Motivation and Context / Description
When removing the last listener by `id` the `session.unsubscribe` was not done for the `event`.
When removing the last listener by `event` the `session.unsubscribe` was done.

So i guess this was a none intentional behavior. @pujagani is this assumption correct?
There might be better way to fix this so this is a PR to have an okay on this.

And there was a write lock used for read, i guess this was a copy and paste error.

### Types of changes

### Checklist

### DIFF
--- a/java/src/org/openqa/selenium/bidi/BiDi.java
+++ b/java/src/org/openqa/selenium/bidi/BiDi.java
@@ -20,7 +20,9 @@ package org.openqa.selenium.bidi;
 import java.io.Closeable;
 import java.time.Duration;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.function.Consumer;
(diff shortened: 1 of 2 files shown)


pujagani:
It was kept that way intentionally. Since both are used differently. The one where we pass the id, it is mainly used for high-level APIs. For Network interception high-level BiDi APIs, we want to subscribe on initialization and let the user control if they want to add listeners or no. Example: https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/remote/RemoteNetwork.java#L54 - subscribes to the event. Once the listeners are removed, by this PR's logic we will end up unsubscribing to the event and then each time listeners are added, we will need to subscribe again.

joerg1985:
@pujagani thanks for the explanation, in this case i will close this PR.

Issue #14427, opened 2024-08-22
[🐛 Bug]: ExpectedCondition::elementToBeClickable does not consider CSS `pointer-events: none`


rr-it:
### What happened?

The docs mention

* `static ExpectedCondition<WebElement> elementToBeClickable(By locator)`
* `static ExpectedCondition<WebElement> elementToBeClickable(WebElement element)`

> An expectation for checking an element is visible and enabled such that you can click it.
See  https://www.selenium.dev/selenium/docs/api/java/org/openqa/selenium/support/ui/ExpectedConditions.html#elementToBeClickable(org.openqa.selenium.By)

So `elementToBeClickable` should check if an element is clickable.

Elements that have CSS `pointer-events: none` set are not clickable. This should be considered in the check for clickability.

See https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events#values

### How can we reproduce the issue?

The following `<button>`-element is not clickable:

HTML:

```html
<button class="disabled">…</button>
```

CSS:

```css
.disabled {
    pointer-events: none;
}
```

But `elementToBeClickable` only checks if an element is there and enabled.
It does not check for the CSS property `pointer-event: none` - which might make an element non-clickable.

https://github.com/SeleniumHQ/selenium/blob/26f9aa2821641444c8946e484c72369d51289b7b/java/src/org/openqa/selenium/support/ui/ExpectedConditions.java#L617-L672

Adding `element.getCssValue("pointer-events") != "none"` to the condition might give the expected result.

### Relevant log output

```shell
IMO not relevant here
```

### Operating System

IMO not relevant here

### Selenium version

IMO not relevant here

### What are the browser(s) and version(s) where you see this issue?

IMO not relevant here

### What are the browser driver(s) and version(s) where you see this issue?

IMO not relevant here

shbenzer:
Just because a pointer-event is set to none doesn't mean an element isn't clickable.

To use your example of a disable button: I _can_ click it, but nothing should happen. If I were writing tests for security permissions, for example, I would want to be able to simulate clicking said button when it is enabled and when it is disabled, then assert that the result of clicking the button when enabled does not occur when the button is disabled.

rr-it:
@shbenzer AFAIK your mentioned case does already not work for buttons "disabled" via `pointer-events: none`.

> To use your example of a disable button: I can click it, but nothing should happen.

[...]

Issue #14030, opened 2024-05-24
[🐛 Bug]: Opacity : 0 is not considered as "Displayed" 


aarahman7:
### What happened?

In our React powered Web App, one of the checkboxes is getting isDisplayed() as false although everything is fine.

1. The element is available in DOM.
2. isEnabled() is true
3. element.click() is working
4. element.getSize() is having non-zero width and height.

When I checked, found that Selenium is considering opacity : 0 as invisible and hence making isDisplayed as false. This is happening in antd react library which is a common library among React Web Apps and this issue seems to be happening in all these areas.

Proposal is, we shall remove this check for opacity.

Additional Note:
https://playwright.dev/docs/actionability#visible

In Playwright also, opacity 0 is not considered as invisible.

### How can we reproduce the issue?

```shell
//As you could see here, only isDisplayed() is false, everything else is fine.

public class RadioTest {
    public static void main(String[] args) throws InterruptedException {
        WebDriver driver = new ChromeDriver();
        driver.get("https://ant.design/components/radio");
        WebElement radioButton = driver.findElement(By.xpath("(//input[@class='ant-radio-input'])[1]"));
        System.out.println("Displayed : " + radioButton.isDisplayed());
        System.out.println("Enabled : " + radioButton.isEnabled());
        System.out.println("Size : " + radioButton.getSize());
        Rectangle rect = radioButton.getRect();
        System.out.println("Rect : " + Arrays
                .stream(rect.getClass().getFields())
                .collect(Collectors
                        .toMap(
                                Field::getName,
                                field -> getValueFromField(field, rect)
                        )
                )
        );
        Thread.sleep(5000);
        driver.quit();
    }

    public static Object getValueFromField(Field field, Object rect) {
        try {
            return field.get(rect);
        } catch (IllegalAccessException e) {
            throw new RuntimeException(e);
        }
    }
}
```

### Relevant log output

```shell
NA
```

### Operating System

Mac

### Selenium version

4.20.0

### What are the browser(s) and version(s) where you see this issue?

Chrome

### What are the browser driver(s) and version(s) where you see this issue?

NA

### Are you using Selenium Grid?

No

diemol:
I thought opacity 0 meant fully transparent, so the human eye cannot see it. Is that correct?

aarahman7:
@diemol Yes. That was my understanding too. But in React libraries like **antd** they are making the opacity as 0 by default.

The element is visible here despite opacity being 0.

[...]

PR #14432, opened 2024-08-23
[java][py] Upgraded EC::toBeClickable for pointer-events: none

shbenzer:
### **User description**
Upgraded `ExpectedCondition::toBeClickable` methods to correctly handle pointer-events: none.

### Description
- Upgraded `ExpectedCondition::toBeClickable` methods to correctly handle pointer-events: none.
- Upgraded webdriverwait_tests.py::test_expected_condition_element_to_be_clickable
- added example to javascript.html for above test

### Motivation and Context
Bug #14427 

### Types of changes

### Checklist

___

### **PR Type**
Bug fix, Tests

___

### **Description**
- Enhanced `ExpectedCondition::elementToBeClickable` methods to correctly handle elements with `pointer-events: none`.
- Added a test case in `webdriverwait_tests.py` to verify that elements with `pointer-events: none` are not considered clickable.
- Updated `javascriptPage.html` to include a button with `pointer-events: none` for testing purposes.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>ExpectedConditions.java</strong><dd><code>Enhance elementToBeClickable to handle pointer-events: none</code></dd></summary>
<hr>

java/src/org/openqa/selenium/support/ui/ExpectedConditions.java

<li>Updated <code>elementToBeClickable</code> methods to check for <code>pointer-events: </code><br><code>none</code>.<br> <li> Ensures elements with <code>pointer-events: none</code> are not considered <br>clickable.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14432/files#diff-9d144815e7045eabe45652340d3517cf24a5bc09c113f6f441aca18a3c0353a5">+2/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    
<tr>
  <td>
    <details>
      <summary><strong>webdriverwait_tests.py</strong><dd><code>Add test for pointer-events: none in clickable elements</code>&nbsp; &nbsp; </dd></summary>
<hr>

py/test/selenium/webdriver/common/webdriverwait_tests.py

<li>Added test case for elements with <code>pointer-events: none</code>.<br> <li> Ensures elements with <code>pointer-events: none</code> are not clickable.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14432/files#diff-cdc0b4528a21c95e4bbede1854baf3eb1580438660127c75e9cca47693bb8691">+7/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>javascriptPage.html</strong><dd><code>Add example button with pointer-events: none</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

common/src/web/javascriptPage.html

- Added a button with `pointer-events: none` for testing.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14432/files#diff-ae4246f838b956b3a7f202c40e51048b8458bdef5e711ca02afc1d35cfeb26d8">+2/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

___

> 💡 **PR-Agent usage**:


joerg1985:
It might be better to raise a speaking WebDriverException instead of just returning false, this could help people to understand what is wrong.

Beside this, i am not sure this will work as expected. The webdriver will click the center of the element, if this would hit the [element or a child inside](https://www.w3.org/TR/webdriver2/#dfn-obscuring) it is considered as a working click and will be performed, otherwise a `ElementClickInterceptedException` is raised.

So i would expect element.click() to work in case the element has `pointer-events: none` and there is a child with a different `pointer-events` value inside it, covering the center of the element.

joerg1985:
PS: comparing strings with `!=` is a reference check in java, this must be changed to `.equals`

shbenzer:
> It might be better to raise a speaking WebDriverException instead of just returning false, this could help people to understand what is wrong.
> 
> Beside this, i am not sure this will work as expected. The webdriver will click the center of the element, if this would hit the [element or a child inside](https://www.w3.org/TR/webdriver2/#dfn-obscuring) it is considered as a working click and will be performed, otherwise a `ElementClickInterceptedException` is raised.
> 
> So i would expect element.click() to work in case the element has `pointer-events: none` and there is a child with a different `pointer-events` value inside it, covering the center of the element.

I'm totally up for suggestions and don't mind reworking the solution. What do you think would be a better way of accomplishing this? 

My thinking was that we already raise a webdriverexception from the element.click() if the element and no children are clickable, so wouldn't adding the pointer-events:none to the check if an element is clickable be sufficient? Or are you saying we should update the exception to tell you what the cause of it not being clickable is?

> PS: comparing strings with `!=` is a reference check in java, this must be changed to `.equals`

Good catch

joerg1985:
Don't get me wrong, but i think the only way to check if something is clickable is to click and check **exactly** the expected action happens. And this check does not cover cases other unexpected things are happening too. And this check is not invertable, e.g. to check the click will open a popup is possible, but it might have run javascript and modified the original page too. It might be possible in the future to implement a `clickAndExpectNoChangeToThePage` or  `clickAndExpectNoClickEventHandlerCalled` using BiDi (or now using the CDP).


People are assuming the `ExpectedConditions.elementToBeClickable` does really check everything e.g. https://github.com/SeleniumHQ/selenium/issues/14030#issuecomment-2129334529 but this is not true.
And even adding a statement like `might return incorrect results` to the javadoc will alot of people not read or ignore.


So i think it would be the best to deprecate it now and remove it in Selenium 5.
This will probably not happen, just a little dream of me 😄

shbenzer:
That was one of my original concerns with this bug ticket. At the end of the day I can move my mouse around and click anything I want, things just might not happen. 

I like the idea of `toBeClickable` being refactored into something like `toBeInteractable` or `clickHasConsequences`. The check instead being _will something happen_ if you click it, not _can_ you click it

shbenzer:
Adding an EC of `toBeInteractable` would probably be best. Would serve as a check for the element not interactable exception, as well. https://w3c.github.io/webdriver/#dfn-error-code

joerg1985:
@VietND96  I don't think this should be merged, see my concerns above in https://github.com/SeleniumHQ/selenium/pull/14432#issuecomment-2328743898

VietND96:
Ah yes, I saw. I just adding a label for filtering PRs that have the potential to merge. Of course, the PR needs to be reviewed and approved.

PR #14433, opened 2024-08-23
[rb] Fix add_cause method not being able to process an array of hashes

### LINKED ISSUE #14415: [🐛 Bug]: error from remote server causes a secondary error in Ruby library
### What happened?

An error from a remote Selenium server is not correctly reported when using the selenium-webdriver v4.23.0 library in Ruby.

It appears that a bug in the selenium-webdriver exception handling code causes another exception to be raised.

Note that this problem does not occur in the previous v4.22.0 version.

I can reproduce by deliberately creating a situation where the client cannot obtain a new session within the timeout.

In one terminal, start selenium via a Docker container:

```
docker run \
  --rm \
  -p "4444:4444" \
  --shm-size="2g" \
  -e SE_SESSION_REQUEST_TIMEOUT=1 \
  -e SE_NODE_MAX_SESSIONS=1 \
  selenium/standalone-chromium:latest
```

In another terminal, run the following Ruby script:

``` ruby
# test.rb

require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "selenium-webdriver", "4.23.0"
end

2.times.each do
  options = Selenium::WebDriver::Options.chrome
  driver = Selenium::WebDriver.for :remote, url: "http://localhost:4444/wd/hub", options: options
  driver.navigate.to "http://google.com"
end
```

### Relevant log output

```shell
/Users/jon/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver/remote/response.rb:61:in `set_backtrace': backtrace must be Array of String (TypeError)

          cause.set_backtrace(backtrace)
                              ^^^^^^^^^
	from /Users/jon/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver/remote/response.rb:61:in `add_cause'
	from /Users/jon/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver/remote/response.rb:41:in `error'
	from /Users/jon/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver/remote/response.rb:52:in `assert_ok'
	from /Users/jon/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver/remote/response.rb:34:in `initialize'
	from /Users/jon/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver/remote/http/common.rb:101:in `new'
	from /Users/jon/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver/remote/http/common.rb:101:in `create_response'
	from /Users/jon/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver/remote/http/default.rb:103:in `request'
	from /Users/jon/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver/remote/http/common.rb:67:in `call'
	from /Users/jon/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver/remote/bridge.rb:675:in `execute'
	from /Users/jon/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver/remote/bridge.rb:76:in `create_session'
	from /Users/jon/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver/common/driver.rb:323:in `block in create_bridge'
	from <internal:kernel>:90:in `tap'
	from /Users/jon/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver/common/driver.rb:322:in `create_bridge'
	from /Users/jon/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver/common/driver.rb:73:in `initialize'
	from /Users/jon/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver/remote/driver.rb:38:in `initialize'
	from /Users/jon/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver/common/driver.rb:57:in `new'
	from /Users/jon/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver/common/driver.rb:57:in `for'
	from /Users/jon/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/selenium-webdriver-4.23.0/lib/selenium/webdriver.rb:89:in `for'
	from test.rb:10:in `block in <main>'
	from test.rb:8:in `times'
	from test.rb:8:in `each'
	from test.rb:8:in `<main>'
```

### Operating System

macOS Sonoma

### Selenium version

Ruby 4.23.0

Chromium 127.0

ChromeDriver 127.0.6533.119

_No response_

aguspe:
### Description
This PR adds the backtrace from remote method again, since when an error originating from a remote server was processed, since it is not an array of strings it failed, so we need to convert the hashes into strings

### Motivation and Context
A proper error message is needed for debugging purposes, so this bugfix helps our users get that.
This PR fixes #14415
 
### Types of changes

### Checklist

___

### **PR Type**
Bug fix, Tests

___

### **Description**
- Added a new method `backtrace_from_remote` to process and convert backtrace from remote server errors into a readable format.
- Modified existing error handling to incorporate the new backtrace processing method, ensuring proper error messages are generated.
- Added integration tests to verify the correct handling of backtrace when errors originate from a remote server.
- Updated type signatures to reflect the new method and its usage.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>response.rb</strong><dd><code>Enhance error handling for remote server backtrace</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

rb/lib/selenium/webdriver/remote/response.rb

<li>Added method <code>backtrace_from_remote</code> to process backtrace from remote <br>server.<br> <li> Modified <code>add_cause</code> to handle array backtrace from remote server.<br> <li> Improved error handling for remote server errors.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14433/files#diff-83593956dcdaeb6b67dbb10105cdbf2ad7e80691729aa77ad705d4281c87553c">+21/-1</a>&nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>response.rbs</strong><dd><code>Update type signature for new backtrace method</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

rb/sig/lib/selenium/webdriver/remote/response.rbs

- Added type signature for `backtrace_from_remote` method.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14433/files#diff-3947056448035b224b9189b479ff80ccc5b49cc67d5dd28f96995ca0f6f1a622">+2/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    
<tr>
  <td>
    <details>
      <summary><strong>error_spec.rb</strong><dd><code>Add tests for remote server backtrace handling</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

rb/spec/integration/selenium/webdriver/error_spec.rb

<li>Added test for backtrace handling when using a remote server.<br> <li> Ensured <code>WebDriverError</code> is raised with a cause.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14433/files#diff-9a507396bce1fb2360f6852a8022c9df9dcaf6869be5a95feef6571adb2bcd20">+7/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

___

> 💡 **PR-Agent usage**:


aguspe:
Ready for review @p0deje @ioquatix

rb/lib/selenium/webdriver/remote/response.rb
```diff
@@ -58,12 +58,30 @@ def assert_ok
...
           ex
         end
 
+        def backtrace_from_remote(server_trace)
+          server_trace.filter_map do |frame|
+            next unless frame.is_a?(Hash)
+
+            file = frame['fileName']
+            line = frame['lineNumber']
+            meth = frame['methodName']
```
ioquatix:
```suggestion
            method = frame['methodName']
```

I personally avoid abbreviations.

aguspe: I like method too, I just wanted to preserve the original backtrace_from_remote style

rb/spec/integration/selenium/webdriver/error_spec.rb
```diff
@@ -47,6 +47,13 @@ module WebDriver
       rescue WebDriver::Error::NoSuchElementError => e
         expect(e.backtrace).not_to be_empty
       end
+
+      it 'has backtrace when using a remote server', only: {driver: :remote} do
+        options = Selenium::WebDriver::Options.chrome
+        Selenium::WebDriver.for :remote, url: 'http://localhost:4444/wd/hub', options: options
```
p0deje: Why would this code fail and raise an error?

aguspe: I updated the options line, to use an incorrect binary

The idea is to produce a session disconnection. In my old code, I tested locally by having a remote that would time out, but this would not work here, so that's why I'm using the incorrect binary now. 

The thing I was looking at was that maybe that URL would not work and produce a TCP error instead of a session error, if the server is not running in that port but I did not figure out how to change the port of the server @p0deje

p0deje: This test would run on all browsers and even non-remote ones. It would also leak the driver and not close when the test is finished. Can you please change it to;

1. Have a `only: {driver: :remote}` guard.
2. Use `create_driver!` helper.
3. some assertion for the backtrace itself (that's what the test is checking based on its title).

aguspe: Thank you for the explanation @p0deje I look more into it, and tried the test locally too and is working, ready for re-review

aguspe:
@p0deje I updated to exclude firefox after the pipelines failures, because instead of throwing a session error, firefox throws the following error:

Selenium::WebDriver::Error::InvalidArgumentError:
       binary is not a Firefox executable
 
 and unless I set the timeout of the remote server to 1 second, I tried with other options to cause the session error and they throw different types of errors, so I think excluding firefox is the best option on this test
 
 Should this be a bug btw? because I was expecting the same error across all the browsers

aguspe:
So the issue on local Mac and local Firefox is that it doesn't fail and the guard throws an error, I will look into that

However, in Edge there is this bug, which I can look into but it seems to be outside of the scope of this PR:

```
Failures:

  1) Selenium::WebDriver::Remote::Driver errors when not set
     Failure/Error:
       expect {
         driver.downloadable_files
       }.to raise_exception(Error::WebDriverError,
                            'You must enable downloads in order to work with downloadable files.')

       expected Selenium::WebDriver::Error::WebDriverError with "You must enable downloads in order to work with downloadable files.", got #<Selenium::WebDriver::Error::UnknownError: Cannot find downloads file system for session id: 0dd8f51... os.arch: 'amd64', os.version: '10.0', java.version: '17.0.11'
       Driver info: driver.version: unknown> with backtrace:
         # ./rb/lib/selenium/webdriver/remote/response.rb:63:in `add_cause'
         # ./rb/lib/selenium/webdriver/remote/response.rb:41:in `error'
         # ./rb/lib/selenium/webdriver/remote/response.rb:52:in `assert_ok'
         # ./rb/lib/selenium/webdriver/remote/response.rb:34:in `initialize'
         # ./rb/lib/selenium/webdriver/remote/http/common.rb:101:in `new'
         # ./rb/lib/selenium/webdriver/remote/http/common.rb:101:in `create_response'
         # ./rb/lib/selenium/webdriver/remote/http/default.rb:103:in `request'
         # ./rb/lib/selenium/webdriver/remote/http/common.rb:67:in `call'
         # ./rb/lib/selenium/webdriver/remote/bridge.rb:685:in `execute'
         # ./rb/lib/selenium/webdriver/remote/features.rb:62:in `downloadable_files'
         # ./rb/lib/selenium/webdriver/common/driver_extensions/has_file_downloads.rb:27:in `downloadable_files'
         # ./rb/spec/integration/selenium/webdriver/remote/driver_spec.rb:87:in `block (3 levels) in <module:Remote>'
         # ./rb/spec/integration/selenium/webdriver/remote/driver_spec.rb:86:in `block (2 levels) in <module:Remote>'
     # ./rb/spec/integration/selenium/webdriver/remote/driver_spec.rb:86:in `block (2 levels) in <module:Remote>'
     ```

p0deje:
> However, in Edge there is this bug, which I can look into but it seems to be outside of the scope of this PR:

Looks like it's failing in the trunk as well, so maybe you could take a look in a separate PR.

aguspe:
> > However, in Edge there is this bug, which I can look into but it seems to be outside of the scope of this PR:
> 
> Looks like it's failing in the trunk as well, so maybe you could take a look in a separate PR.

Great, I will do

p0deje:
There are still some failures in RBE tests. I wonder if you could find a different way to test this w/o relying on binary. Maybe try sending a command that would cause a failure (e.g. find non-existent element)?

aguspe:
> There are still some failures in RBE tests. I wonder if you could find a different way to test this w/o relying on binary. Maybe try sending a command that would cause a failure (e.g. find non-existent element)?

That's a good point, I updated the error type now @p0deje

aguspe:
Ready for re-review @p0deje

p0deje:
@aguspe Can you please double-check that the spec actually catches the regression you fixed? It's possible that the backtrace is not coming from the remote in this particular case, so we might need to find a different scenario. I know I suggested it in the first place, but I hoped you'd check that by yourself, though I haven't asked for it explicitly.

aguspe:
> @aguspe Can you please double-check that the spec actually catches the regression you fixed? It's possible that the backtrace is not coming from the remote in this particular case, so we might need to find a different scenario. I know I suggested it in the first place, but I hoped you'd check that by yourself, though I haven't asked for it explicitly.

So if we want to have a session-related scenario, we could close the driver earlier, here are 2 screenshots of an example before and after the fix on the same spec locally, I will push my changes now so we do not use the find element and we focus on the session which I think is a better spec

Could you re-run the pipeline @p0deje ?

Screenshot 2024-09-02 at 07 01 41


Screenshot 2024-09-02 at 07 02 37

aguspe:
@p0deje the failing test in firefox seems not to be related to my PR, but I will make another PR fixing the issue and also the Edge one so the pipelines work again, unless you prefer me to fix it on this one

aguspe:
@p0deje I just updated the test to cause the session error without quitting the driver, this should avoid quitting driver errors

aguspe:
Updated to use title now, I tested locally still produces the session error

p0deje:
@aguspe There are still some failures in RBE tests

aguspe:
@p0deje Tested locally on chrome, switched to window handle, works as expected, hopefully now the only issue left is the existent edge one

aguspe:
@p0deje now the test is not using any script (I cannot reproduce the error locally), but I'm modifying the session id to cause an error

aguspe:
Now the only issue left is the edge bug and with this merge, it's is going to be easier to debug @p0deje, so when this gets merged I will jumpt to work on the edge fix and fix any review comments from #14287

Issue #14461, opened 2024-09-01
[🐛 Bug]: DefaultSlotMatcher is considering extension capabilities to determine whether the slot matches the requested capabilities


sbabcoc:
### What happened?

My Selenium framework includes the ability to manage a local Grid instance, including Appium nodes. Because the Appium server needs an execution context to function, I have the ability to run the server via PM2. When this configuration is employed, the shutdown procedure is different. Consequently, I need to include a property in the capabilities object that indicates this configuration. This is done via an extension capability names `nord:options`. The expectation is that this capability will be available to query, but won't affect slot matching. Unfortunately, this isn't currently the case. Adding this extension capability causes the corresponding node to no longer match Appium session requests.

The scenario that caused my immediate issue is that both the node capabilities and the new session request define values for `nord:options`, each specifying a different informational property. The slot matcher compares the values of these capabilities and rejects the slot because the objects differ.

### How can we reproduce the issue?

```shell
Create a grid with a node whose capabilities include a `nord:options` extension capability. Try to acquire a session from this node with capabilities that specify a different value for `nord:options`.

EXPECT: The slot matcher should provide a session
ACTUAL: No matching session is provided
```

### Relevant log output

```shell
FAILED: com.nordstrom.automation.selenium.android.AndroidTest.testSearchActivity
        {
  "constName": "ANDROID",
  "description": "",
  "className": "com.nordstrom.automation.selenium.platform.TargetType",
  "class": "com.nordstrom.automation.selenium.listeners.PlatformInterceptor$PlatformIdentity"
}
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
	at org.openqa.selenium.support.ui.FluentWait.propagateIfNotIgnored(FluentWait.java:252)
	at org.openqa.selenium.support.ui.FluentWait.until(FluentWait.java:213)
	at com.nordstrom.automation.selenium.core.DriverManager.injectDriver(DriverManager.java:169)
	at com.nordstrom.automation.selenium.core.DriverManager.beforeInvocation(DriverManager.java:105)
	at com.nordstrom.automation.selenium.listeners.DriverListener.beforeInvocation(DriverListener.java:55)
	at com.nordstrom.automation.testng.AbstractListenerChain.beforeInvocation(AbstractListenerChain.java:231)
	at org.testng.internal.invokers.InvokedMethodListenerInvoker.invokeListener(InvokedMethodListenerInvoker.java:56)
[...]

PR #14485, opened 2024-09-10
[grid] Eliminate vendor-specific handling of extension capabilities 

sbabcoc:
### Description
The existing handling of extension capabilities assigns special significance to four recognized prefixes: `goog:`, `moz:`, `ms:`, and `se:`. Capabilities with these prefixes are entirely ignored by the current default slot matcher implementation, but custom prefixes are considered, as well as those for Safari and Appium. This inconsistency means that properties in extension capabilities can cause affected `newSession` requests to fail even though extension capabilities are supposed to be vendor-specific and should probably not be evaluated by the default slot matcher.

This PR retains the "special" status of the "se:" prefix and specifically filters out **all** extension capabilities with names that end with "options", "Options", "loggingPrefs", and "debuggerAddress". This maintains existing behavior while allowing node configurations and `newSession` requests to include extension capabilities that won't affect slot matching.

### Motivation and Context
Revolves #14461 

> Note that this is technically a breaking change, because it revises the default slot matcher to consider "special" extension capabilities without name suffix "options" that would previously have been ignored, and the matcher will now ignore "non-special" extension capabilities with name suffix "options" that would previously have been considered.
However, I'm unaware of any current use cases that will be adversely affected by this change.

The strategy employed by the PR is that extension capabilities which should be ignored for matching can be expressed as capabilities with name prefix "se:" or one of the "special" name suffixes. All other extension capabilities will be considered for slot matching. This is a generalization/extrapolation of existing patterns from current use cases.

### Recommended slot matcher enhancements

To reduce the need for custom slot matchers, we could extend the **WebDriverInfo** interface to add a new method that vendors could implement to evaluate their own criteria:

```
Boolean matches(Capabilities stereotype, Capabilities capabilities);
```

The default implementation would return `null` to indicate that no evaluation has been performed. If the vendor chooses to implement the `matches` method, their initial step must be to invoke their own `isSupporting` method, returning `null` if the corresponding driver doesn't support the specified capabilities. 

The implementation in **DefaultSlotMatcher** would be updated to iterate over the available **WebDriverInfo** providers, retaining only those whose `isSupporting` methods return `true`. The `matches` methods of this filtered list of providers will then be applied to the available nodes to determine which of these satisfies the criteria specified by the client request. The nodes that satisfy these evaluations (if any) would then be evaluated against the remaining common criteria.

Another potential enhancement would be to enable vendor-specific `matches` methods to return a weighted integer result instead of a simple `Boolean`. This would enable best-match behavior. Perhaps this is getting too complicated, though.

### Types of changes

### Checklist

___

### **PR Type**
Bug fix, Tests

___

### **Description**
- Removed the special status of certain extension capability prefixes in `DefaultSlotMatcher`, opting to ignore all extension capabilities with names ending in "options" (case-insensitive).
- Simplified the logic for matching capabilities by removing unnecessary checks and conditions.
- Updated `RelaySessionFactory` to streamline session creation by removing redundant capability filtering.
- Revised test cases in `DefaultSlotMatcherTest` to align with the new handling of extension capabilities.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>DefaultSlotMatcher.java</strong><dd><code>Revise extension capabilities handling in DefaultSlotMatcher</code></dd></summary>
<hr>

java/src/org/openqa/selenium/grid/data/DefaultSlotMatcher.java

<li>Removed special handling for specific extension capability prefixes.<br> <li> Updated logic to ignore all extension capabilities with names ending in "options" (case-insensitive). <li> Simplified capability matching logic.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14485/files#diff-08612f5a21ea03efdd38ead8885dd4b72be486ea76f0de8f0a4dce609f34af0d">+31/-51</a>&nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>RelaySessionFactory.java</strong><dd><code>Simplify session creation in RelaySessionFactory</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

java/src/org/openqa/selenium/grid/node/relay/RelaySessionFactory.java

<li>Removed logic to filter out browserName when appium:app is present.<br> <li> Simplified session creation process.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14485/files#diff-8fe345d7ed627b58287d8a4c5d6813fae678894be658c5c87ffc3f9c862f62ab">+0/-11</a>&nbsp; &nbsp; </td>

</tr>                    
<tr>
  <td>
    <details>
      <summary><strong>DefaultSlotMatcherTest.java</strong><dd><code>Update DefaultSlotMatcher tests for revised capability handling</code></dd></summary>
<hr>

java/test/org/openqa/selenium/grid/data/DefaultSlotMatcherTest.java

<li>Updated test cases to reflect changes in extension capability <br>handling.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14485/files#diff-efbc947cb2db3c8fe94ef12f0e57d80901e86d3e3992611966b165c1e9543445">+10/-8</a>&nbsp; &nbsp; </td>

</tr>                    

___

> 💡 **PR-Agent usage**:

### DIFF
--- a/java/src/org/openqa/selenium/grid/data/DefaultSlotMatcher.java
+++ b/java/src/org/openqa/selenium/grid/data/DefaultSlotMatcher.java
@@ -39,17 +39,27 @@ import org.openqa.selenium.Capabilities;
  *       Then the {@code stereotype} must contain the same values.
  * </ul>
  *
- * <p>One thing to note is that extension capabilities are not considered when matching slots, since
- * the matching of these is implementation-specific to each driver.
+ * <p>Note that extension capabilities are considered for slot matching, with the following exceptions:
+ *
+ * <ul>
+ *   <li>Extension capabilities with prefix "se:"
+ *   <li>Extension capabilities with these suffixes:
+ *     <ul>
+ *       <li>"options"
+ *       <li>"Options"
+ *       <li>"loggingPrefs"
+ *       <li>"debuggerAddress"
+ *     </ul>
+ * </ul>
  */
 public class DefaultSlotMatcher implements SlotMatcher, Serializable {
 
   /*
-   List of prefixed extension capabilities we never should try to match, they should be
+   List of extension capability suffixes we never should try to match, they should be
    matched in the Node or in the browser driver.
   */
-  private static final List<String> EXTENSION_CAPABILITIES_PREFIXES =
-      Arrays.asList("goog:", "moz:", "ms:", "se:");
+  private static final List<String> EXTENSION_CAPABILITY_SUFFIXES =
+      Arrays.asList("Options", "options", "loggingPrefs", "debuggerAddress");
 
   @Override
   public boolean matches(Capabilities stereotype, Capabilities capabilities) {
@@ -76,14 +86,14 @@ public class DefaultSlotMatcher implements SlotMatcher, Serializable {
 
     // At the end, a simple browser, browserVersion and platformName match
     boolean browserNameMatch =
-        (capabilities.getBrowserName() == null || capabilities.getBrowserName().isEmpty())
+        capabilities.getBrowserName() == null
+            || capabilities.getBrowserName().isEmpty()
             || Objects.equals(stereotype.getBrowserName(), capabilities.getBrowserName());
     boolean browserVersionMatch =
-        (capabilities.getBrowserVersion() == null
-                || capabilities.getBrowserVersion().isEmpty()
-                || Objects.equals(capabilities.getBrowserVersion(), "stable"))
-            || browserVersionMatch(
-                stereotype.getBrowserVersion(), capabilities.getBrowserVersion());
+        capabilities.getBrowserVersion() == null
+            || capabilities.getBrowserVersion().isEmpty()
+            || Objects.equals(capabilities.getBrowserVersion(), "stable")
+            || browserVersionMatch(stereotype.getBrowserVersion(), capabilities.getBrowserVersion());
     boolean platformNameMatch =
         capabilities.getPlatformName() == null
             || Objects.equals(stereotype.getPlatformName(), capabilities.getPlatformName())
@@ -102,21 +112,17 @@ public class DefaultSlotMatcher implements SlotMatcher, Serializable {
         .filter(name -> !name.contains(":"))
         // Platform matching is special, we do it later
         .filter(name -> !"platformName".equalsIgnoreCase(name))
-        .map(
+        .filter(name -> capabilities.getCapability(name) != null)
+        .allMatch(
             name -> {
-              if (capabilities.getCapability(name) instanceof String) {
-                return stereotype
-                    .getCapability(name)
-                    .toString()
-                    .equalsIgnoreCase(capabilities.getCapability(name).toString());
-              } else {
-                return capabilities.getCapability(name) == null
-                    || Objects.equals(
-                        stereotype.getCapability(name), capabilities.getCapability(name));
+              if (stereotype.getCapability(name) instanceof String
+                  && capabilities.getCapability(name) instanceof String) {
+                return ((String) stereotype.getCapability(name))
+                    .equalsIgnoreCase((String) capabilities.getCapability(name));
               }
-            })
-        .reduce(Boolean::logicalAnd)
-        .orElse(true);
+              return Objects.equals(
+                  stereotype.getCapability(name), capabilities.getCapability(name));
+            });
   }
 
   private Boolean managedDownloadsEnabled(Capabilities stereotype, Capabilities capabilities) {
@@ -140,39 +146,39 @@ public class DefaultSlotMatcher implements SlotMatcher, Serializable {
     */
     return capabilities.getCapabilityNames().stream()
         .filter(name -> name.contains("platformVersion"))
-        .map(
+        .allMatch(
             platformVersionCapName ->
                 Objects.equals(
                     stereotype.getCapability(platformVersionCapName),
-                    capabilities.getCapability(platformVersionCapName)))
-        .reduce(Boolean::logicalAnd)
-        .orElse(true);
+                    capabilities.getCapability(platformVersionCapName)));
   }
 
   private Boolean extensionCapabilitiesMatch(Capabilities stereotype, Capabilities capabilities) {
     /*
-     We match extension capabilities when they are not prefixed with any of the
-     EXTENSION_CAPABILITIES_PREFIXES items. Also, we match them only when the capabilities
-     of the new session request contains that specific extension capability.
+     We match extension capabilities in new session requests whose names do not have the prefix "se:" or
+     one of the reserved suffixes ("options", "Options", "loggingPrefs", or "debuggerAddress"). These are
+     forwarded to the matched node for use in configuration, but are not considered for node matching.
     */
     return stereotype.getCapabilityNames().stream()
+        // examine only extension capabilities
         .filter(name -> name.contains(":"))
-        .filter(name -> capabilities.asMap().containsKey(name))
-        .filter(name -> EXTENSION_CAPABILITIES_PREFIXES.stream().noneMatch(name::contains))
-        .map(
+        // ignore Selenium extension capabilities
+        .filter(name -> !name.startsWith("se:"))
+        // ignore special extension capability suffixes
+        .filter(name -> EXTENSION_CAPABILITY_SUFFIXES.stream().noneMatch(name::endsWith))
+        // ignore capabilities not specified in the request
+        .filter(name -> capabilities.getCapability(name) != null)
+        .allMatch(
             name -> {
-              if (capabilities.getCapability(name) instanceof String) {
-                return stereotype
-                    .getCapability(name)
-                    .toString()
-                    .equalsIgnoreCase(capabilities.getCapability(name).toString());
-              } else {
-                return capabilities.getCapability(name) == null
-                    || Objects.equals(
-                        stereotype.getCapability(name), capabilities.getCapability(name));
+              // evaluate capabilities with String values
+              if (stereotype.getCapability(name) instanceof String
+                  && capabilities.getCapability(name) instanceof String) {
+                return ((String) stereotype.getCapability(name))
+                    .equalsIgnoreCase((String) capabilities.getCapability(name));
               }
-            })
-        .reduce(Boolean::logicalAnd)
-        .orElse(true);
+              // evaluate capabilities with Number or Boolean values
+              return Objects.equals(
+                  stereotype.getCapability(name), capabilities.getCapability(name));
+            });
   }
 }
--- a/java/src/org/openqa/selenium/grid/node/relay/RelaySessionFactory.java
+++ b/java/src/org/openqa/selenium/grid/node/relay/RelaySessionFactory.java
@@ -40,7 +40,6 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.openqa.selenium.Capabilities;
 import org.openqa.selenium.ImmutableCapabilities;
-import org.openqa.selenium.MutableCapabilities;
 import org.openqa.selenium.SessionNotCreatedException;
 import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.grid.data.CreateSessionRequest;
@@ -50,7 +49,6 @@ import org.openqa.selenium.grid.node.SessionFactory;
 import org.openqa.selenium.internal.Debug;
 import org.openqa.selenium.internal.Either;
 import org.openqa.selenium.internal.Require;
-import org.openqa.selenium.remote.CapabilityType;
 import org.openqa.selenium.remote.Command;
 import org.openqa.selenium.remote.Dialect;
 import org.openqa.selenium.remote.DriverCommand;
@@ -149,15 +147,6 @@ public class RelaySessionFactory implements SessionFactory {
               "New session request capabilities do not " + "match the stereotype."));
     }
 
-    // remove browserName capability if 'appium:app' is present as it breaks appium tests when app
-    // is provided
-    // they are mutually exclusive
-    MutableCapabilities filteredStereotype = new MutableCapabilities(stereotype);
-    if (capabilities.getCapability("appium:app") != null) {
-      filteredStereotype.setCapability(CapabilityType.BROWSER_NAME, (String) null);
-    }
-
-    capabilities = capabilities.merge(filteredStereotype);
     LOG.info("Starting session for " + capabilities);
 
     try (Span span = tracer.getCurrentContext().createSpan("relay_session_factory.apply")) {
--- a/java/test/org/openqa/selenium/grid/data/DefaultSlotMatcherTest.java
+++ b/java/test/org/openqa/selenium/grid/data/DefaultSlotMatcherTest.java
@@ -19,6 +19,7 @@ package org.openqa.selenium.grid.data;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import java.util.Map;
 import org.junit.jupiter.api.Test;
 import org.openqa.selenium.Capabilities;
 import org.openqa.selenium.ImmutableCapabilities;
@@ -536,7 +537,37 @@ class DefaultSlotMatcherTest {
   }
 
   @Test
-  void vendorExtensionPrefixedCapabilitiesAreIgnoredForMatching() {
+  void seleniumExtensionCapabilitiesAreIgnoredForMatching() {
+    Capabilities stereotype =
+      new ImmutableCapabilities(
+        CapabilityType.BROWSER_NAME,
+        "chrome",
+        CapabilityType.BROWSER_VERSION,
+        "84",
+        CapabilityType.PLATFORM_NAME,
+        Platform.WINDOWS,
+        "se:cdpVersion",
+        1,
+        "se:downloadsEnabled",
+        true);
+
+    Capabilities capabilities =
+      new ImmutableCapabilities(
+        CapabilityType.BROWSER_NAME,
+        "chrome",
+        CapabilityType.BROWSER_VERSION,
+        "84",
+        CapabilityType.PLATFORM_NAME,
+        Platform.WINDOWS,
+        "se:cdpVersion",
+        2,
+        "se:downloadsEnabled",
+        false);
+    assertThat(slotMatcher.matches(stereotype, capabilities)).isTrue();
+  }
+
+  @Test
+  void vendorOptionsCapabilitiesAreIgnoredForMatching() {
     Capabilities stereotype =
         new ImmutableCapabilities(
             CapabilityType.BROWSER_NAME,
@@ -545,10 +576,10 @@ class DefaultSlotMatcherTest {
             "84",
             CapabilityType.PLATFORM_NAME,
             Platform.WINDOWS,
-            "goog:cheese",
-            "amsterdam",
-            "ms:fruit",
-            "mango");
+            "food:fruitOptions",
+            "mango",
+            "dairy:options",
+            Map.of("cheese", "amsterdam"));
 
     Capabilities capabilities =
         new ImmutableCapabilities(
@@ -558,10 +589,40 @@ class DefaultSlotMatcherTest {
             "84",
             CapabilityType.PLATFORM_NAME,
             Platform.WINDOWS,
-            "goog:cheese",
-            "gouda",
-            "ms:fruit",
-            "orange");
+            "food:fruitOptions",
+            "orange",
+            "dairy:options",
+            Map.of("cheese", "gouda"));
+    assertThat(slotMatcher.matches(stereotype, capabilities)).isTrue();
+  }
+
+  @Test
+  void specialExtensionCapabilitiesAreIgnoredForMatching() {
+    Capabilities stereotype =
+      new ImmutableCapabilities(
+        CapabilityType.BROWSER_NAME,
+        "chrome",
+        CapabilityType.BROWSER_VERSION,
+        "84",
+        CapabilityType.PLATFORM_NAME,
+        Platform.WINDOWS,
+        "food:loggingPrefs",
+        "mango",
+        "food:debuggerAddress",
+        Map.of("cheese", "amsterdam"));
+
+    Capabilities capabilities =
+      new ImmutableCapabilities(
+        CapabilityType.BROWSER_NAME,
+        "chrome",
+        CapabilityType.BROWSER_VERSION,
+        "84",
+        CapabilityType.PLATFORM_NAME,
+        Platform.WINDOWS,
+        "food:loggingPrefs",
+        "orange",
+        "food:debuggerAddress",
+        Map.of("cheese", "gouda"));
     assertThat(slotMatcher.matches(stereotype, capabilities)).isTrue();
   }
 
--- a/java/test/org/openqa/selenium/grid/node/config/NodeOptionsTest.java
+++ b/java/test/org/openqa/selenium/grid/node/config/NodeOptionsTest.java
@@ -63,6 +63,7 @@ import org.openqa.selenium.ie.InternetExplorerDriverInfo;
 import org.openqa.selenium.internal.Either;
 import org.openqa.selenium.json.Json;
 import org.openqa.selenium.net.NetworkUtils;
+import org.openqa.selenium.remote.Browser;
 import org.openqa.selenium.safari.SafariDriverInfo;
 
 @SuppressWarnings("DuplicatedCode")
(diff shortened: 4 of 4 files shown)


java/src/org/openqa/selenium/grid/data/DefaultSlotMatcher.java
```diff
@@ -93,25 +84,21 @@ public boolean matches(Capabilities stereotype, Capabilities capabilities) {
 
   private Boolean initialMatch(Capabilities stereotype, Capabilities capabilities) {
     return stereotype.getCapabilityNames().stream()
-        // Matching of extension capabilities is implementation independent. Skip them
-        .filter(name -> !name.contains(":"))
-        // Platform matching is special, we do it later
-        .filter(name -> !"platformName".equalsIgnoreCase(name))
-        .map(
-            name -> {
-              if (capabilities.getCapability(name) instanceof String) {
```
sbabcoc, line 107: The implementation of this conditional fails to consider that the corresponding stereotype capability could have a non-String value (including `null`), which would produce unexpected behavior. The revised implementation will only perform the case-insensitive comparison if the capability in the session request and the node stereotype both have String values.

java/src/org/openqa/selenium/grid/data/DefaultSlotMatcher.java
```diff
@@ -145,29 +132,22 @@ private Boolean platformVersionMatch(Capabilities stereotype, Capabilities capab
...
-                return capabilities.getCapability(name) == null
-                    || Objects.equals(
-                        stereotype.getCapability(name), capabilities.getCapability(name));
-              }
-            })
-        .reduce(Boolean::logicalAnd)
-        .orElse(true);
+      .filter(name -> name.contains(":"))
+      .filter(name -> capabilities.getCapability(name) != null)
+      .map(name -> {
```
sbabcoc: This revised comparison will only evaluate values that are strings, numbers, or Booleans; complex values (objects and arrays) are ignored. The prior "special" treatment of four recognized extension capability prefixes (`goog:`, `moz:`, `ms:`, and `se:`) has been replaced by this consistent handling of all extension capabilities.

java/src/org/openqa/selenium/grid/node/relay/RelaySessionFactory.java
```diff
@@ -149,15 +147,6 @@ public Either<WebDriverException, ActiveSession> apply(CreateSessionRequest sess
               "New session request capabilities do not " + "match the stereotype."));
     }
 
-    // remove browserName capability if 'appium:app' is present as it breaks appium tests when app
```
sbabcoc, line 152: The `merge` operation that necessitated this special-case handling has been eliminated. The session factory shouldn't be altering clients' requested capabilities.

java/test/org/openqa/selenium/grid/data/DefaultSlotMatcherTest.java
```diff
@@ -433,10 +436,10 @@ void vendorExtensionPrefixedCapabilitiesAreIgnoredForMatching() {
             "84",
             CapabilityType.PLATFORM_NAME,
             Platform.WINDOWS,
-            "goog:cheese",
-            "amsterdam",
-            "ms:fruit",
-            "mango");
+            "food:dairy",
```
sbabcoc: This test has been updated to verify the revised behavior, in which all extension capabilities with complex values are ignored for purposes of node matching.

java/test/org/openqa/selenium/grid/node/config/NodeOptionsTest.java
```diff
@@ -148,7 +149,7 @@ boolean isDownloadEnabled(WebDriverInfo driver, String customMsg) {
               reported.add(caps);
               return Collections.singleton(HelperFactory.create(config, caps));
             });
-    String expected = driver.getDisplayName();
+    String expected = "Edge".equals(driver.getDisplayName()) ? Browser.EDGE.browserName() : driver.getDisplayName();
```
sbabcoc: This test was broken when I got here. For **Microsoft Edge**, the display name is `Edge`, but the browser name is `MicrosoftEdge`. Consequently, the expected value would never be matched.

java/test/org/openqa/selenium/grid/data/DefaultSlotMatcherTest.java
pujagani: I think it is a good idea to create a new test and let the existing test be as it is to verify the changes don't cause a breaking change for the users.

pujagani (changes requested):

java/test/org/openqa/selenium/grid/data/DefaultSlotMatcherTest.java
sbabcoc: I restored the original test, altering the method name and inverting the expectation to correspond with the revised behavior.

pujagani:
@diemol @krmahadevan Please help review this when time permits!

joerg1985:
@VietND96 sorry i don't know enought to really review the changes in this PR.

From the comment in the code [here](https://github.com/SeleniumHQ/selenium/blob/751bacb6bc934436ec9dec2416a022d8d577e30a/java/src/org/openqa/selenium/grid/data/DefaultSlotMatcher.java#L148-L152) and after looking into the issue #14461 i am asking myself: should the request really contain these capabilities?

The node should be picked by the desired capability, but is able to respond with a extended set of capabilities.
To stick with the example, i request a "Appium" session and the response is this is a "Appium session started via PM2".
So the client could handle the "PM2" sessions different, but i think a intermediary node is the better place to handle this.
I don't think there is currently a extension point (like specifiying your own slot matcher) to handle the shutdown of a session.

@sbabcoc Is your custom slot matcher starting a new process and this must be terminated?

sbabcoc:
@joerg1985 As I read the W3C specification, the slot matcher should not be interpreting custom capabilities in any way. These should be entirely ignored and passed on without examination. Unfortunately, current functionality violates this specification. This PR modifies the slot matching behavior, replacing special-case handling of four specific prefixes with indifference to custom capabilities with complex values. This allows Appium custom capabilities to be considered for slot matching, while custom capabilities for Chrome, Edge, and Firefox will be ignored.

java/src/org/openqa/selenium/grid/data/DefaultSlotMatcher.java
```diff
@@ -145,27 +135,22 @@ private Boolean platformVersionMatch(Capabilities stereotype, Capabilities capab
...
+              }
+              if (capabilities.getCapability(name) instanceof Number
+                  || capabilities.getCapability(name) instanceof Boolean) {
+                return Objects.equals(
+                    stereotype.getCapability(name), capabilities.getCapability(name));
               }
+              return true;
             })
         .reduce(Boolean::logicalAnd)
         .orElse(true);
```
joerg1985: I would change this to make it more clear what is going on, filtering the element to exclude them from matching.
E.g. in pseudo code:
```
return stereotype.getCapabilityNames().stream()
   .filter(name -> name.contains(":"))
   // match only when the capabilities of the new session request contains that specific extension capability
   .filter(name -> capabilities.getCapability(name) != null)
   // {comment on why to exclude them from matching} 
   .filter(name -> !(capabilities.getCapability(name) instanceof Map))
   .filter(name -> !(capabilities.getCapability(name) instanceof List))
   .allMatch( … {same code as above in the .map of the DefaultSlotMatcher.matches method is used}
```

The `DefaultSlotMatcher.matches` method could be reworked to use `.allMatch` too, this makes it - at least for me - easier to read than `.reduce(Boolean::logicalAnd).orElse(true);`

sbabcoc: @joerg1985 Thanks for the feedback! I switched all three instances of `map`/`reduce` to `allMatch` and added a few code comments.

diemol:
> @joerg1985 As I read the W3C specification, the slot matcher should not be interpreting custom capabilities in any way. These should be entirely ignored and passed on without examination. Unfortunately, current functionality violates this specification. This PR modifies the slot matching behavior, replacing special-case handling of four specific prefixes with indifference to custom capabilities with complex values. This allows Appium custom capabilities to be considered for slot matching, while custom capabilities for Chrome, Edge, and Firefox will be ignored.

@sbabcoc This is only valid for end nodes (ChromeDriver, GeckoDriver, etc.), Grid, and all its components are an intermediate node. That part of the spec is relaxed so the code can be customized as needed, and with that, Grid and cloud vendors can adapt to the needs of their users.

sbabcoc:
@diemol The main issues with the existing implementation are that it treats the "non-special" extension capabilities different than it does the "special" ones, and it alters the requested client capabilities before forwarding them to the node. My revised implementation eliminates both of these issues while retaining all currently utilized behavior.

VietND96:
As my understanding, after this change, for example a Node set stereotypes `"se:containerName": "my-node1"`. In client code, it must set capability `se:containerName` to request for the slot matching, since it is a String cap, right?

sbabcoc:
@VietND96 Capabilities with "simple" values (Boolean/Number/String) will be considered for slot matching, even if their capability names specify a prefix. Capabilities with "complex" values (Array/Object) will be ignored for slot matching if their capability names specify a prefix.
We shouldn't give special treatment to any prefixes, because this produces inconsistent behavior. We also shouldn't be altering clients' requested capabilities in the node matching process.

VietND96:
I added few tests to DefaultSlotMatcher for some scenarios that I want to verify against this change. Let's wait for the CI.

VietND96 (changes requested):
It looks like there are tests get impacted by this change. Need to fix CI.

VietND96:
CI failed due to a case, where Node stereotype `moz:debuggerAddress: 127.0.0.1:3922`, and request capabilities `moz:debuggerAddress: true`
As you mentioned this change will consider caps with simplex value in slot matcher. And session could not be assigned to slot due to this change removing filter list of special prefixed extension caps `"goog:", "moz:", "ms:", "se:"` out.
As I understand, those prefixed caps will contain configs of browser, or Selenium function. In Node stereotypes it could be added automatically or assigned another value based on what it fetches from ENV, OS, etc.
And the same prefixed caps in client request caps, it could be just a boolean value to enable feature.

VietND96:
I will collect and convert those failure to unit test, you just need to fix and make UT can pass when running
> bazel test //java/test/org/openqa/selenium/grid/data/...

diemol:
The only change we need is to avoid matching prefixed capabilities with `options` as a name. 

As in, we would not process `nord:options`, simply pass it along. 

Can you simplify the changes made in this PR to implement that, please?

sbabcoc:
@VietND96 I'm trying to understand why the comparison in the test you added should return `true`. Not only are the values for `moz:debuggerAddress` not the same; the values are of completely different types. I understand that this comparison would return `true` in the current implementation, but this is actually one of the bugs my PR is addressing.
I strongly suspect that there are no real-world scenarios in which this sort of mismatch should be ignored. This seems like a test that's been tailored to verify an arbitrary behavior that's an unintended artifact of the current implementation. If the extension prefixes were replaced with `appium`, the current implementation returns `false`, and this is a behavior that existing clients actually depend on.

sbabcoc:
@diemol Changing the behavior to only ignore extension capabilities whose names contain "options" would produce the same behavior that is causing the test @VietND96 added to fail. The values of `moz:debuggerAddress` would be compared and found to be unequal, and the comparison would return `false`. As I stated in my previous reply, I believe this is the correct behavior - unequal "simple" values should cause the comparison to return `false`.

The approach I'm presenting in this PR provides the greatest level of control and consistency for clients.
* If the client wants the capability to be considered for matching, declare it with a "simple" value.
* If the client wants the capability to be ignored for matching, declare it with a "complex" value.
* No extension capability prefixes get "special" treatment, which avoids confusing, inconsistent behavior for custom extensions.

diemol:
>The values of moz:debuggerAddress would be compared and found to be unequal, and the comparison would return false.

No, because we will still ignore `moz:`. The only change we need is to ignore `:options`, which makes sense to be ignored when matching.

diemol:
All other changes should be reverted, please.

sbabcoc:
@diemol I understand that we currently ignore all "special" extension capabilities, but there appears to be no real-world scenarios of "special" extension capabilities with "simple" values in which this behavior is actually needed. For Appium extension capabilities, current clients rely on capabilities with "simple" values being considered for node matching.
I believe that the existing treatment of "special" extension capabilities is undesirably permissive, and the inconsistent behavior we see for "non-special" extension capabilities is also undesirable.

diemol:
>I believe that the existing treatment of "special" extension capabilities is undesirably permissive, and the inconsistent behavior we see for "non-special" extension capabilities is also undesirable.

Why?

sbabcoc:
@diemol Why is it desirable to support the scenario enforced by the test that @VietND96 added? We match nodes even when capabilities like `moz:debuggerAddress` are entirely different between requested and actual values. We don't exhibit this behavior for Appium nodes, where capabilities like `appium:automationName` are considered when matching requested and actual values.
The revised behavior you recommend (only ignoring extension capabilities whose names contain "options") would produce behavior that doesn't satisfy the condition expected by the test that @VietND96 added.

diemol:
I cannot see what @VietND96 added because you are force-pushing the whole time, and I don't know who added what. I don't know which scenario you mean.

VietND96:
I added it back to trunk https://github.com/SeleniumHQ/selenium/commit/10119a9094e290c3720f9b4f8dece0919a2cd850

diemol:
@sbabcoc is correct, this `moz:debuggerAddress` should not be considered for matching. That is an internal Firefox thing.

sbabcoc:
@diemol I force-push after rebasing to keep the commit history clean. My attempt to rebase after @VietND96 added his unit test went horribly wrong and I had to check out a prior commit to recover, which removed his test. I let him know, and he created a separate PR to add his test to trunk.

sbabcoc:
@diemol It seems like you're advocating for ignoring all extension capabilities for purposes of slot matching. This approach would mean that custom matchers would be required in all scenarios where matching based on extension capabilities is desired (e.g. - `appium:automationName`). The approach implemented by this PR is a compromise. It enables clients add slot matching criteria by adding extension capabilities with "simple" values, while extension capabilities specified as "complex" will be ignored.

The enhancements I recommended in the PR description provide the outline of a mechanism for vendors to define slot matcher extensions that evaluate their specific matching criteria. This would enable vendors to define their own custom criteria without the need to implement custom slot matchers.

diemol:
>@diemol It seems like you're advocating for ignoring all extension capabilities for purposes of slot matching.

Not really. Just the ones named `options`, as in `nord:options`.

sbabcoc:
@diemol I can revise my PR to implement this strategy. It is significantly less complex to implement and much easier to describe. I'll update the unit tests accordingly.

Is the concept of a slot matcher extension mechanism worth exploring further? This would be an additional method for the **WebDriverInfo** interface, which is already being scanned for service providers.

diemol:
You can already implement your own [SlotMatcher](https://www.selenium.dev/documentation/grid/configuration/cli_options/#distributor), see `--slot-matcher`

sbabcoc:
@diemol Yes, but you can only specify one custom matcher. If you need sessions from multiple vendors, this means running with multiple hubs. With a slot matcher extension mechanism, vendor-specific matching gets activated automatically and standing up a grid that supplies sessions from multiple vendors is much easier.

diemol:
The matcher is used both at the Distributor and the Node. You would need to decouple that logic to allow more matchers.

sbabcoc:
@diemol The extension would be implemented in **DefaultSlotMatcher**, which is already integrated with Distributor and Node.

sbabcoc:
@diemol I've revised the matcher to only ignore extension capabilities with suffix "options" (case-insensitive) and updated the affected unit test accordingly. I also updated the PR description to indicate the revised implementation.

java/src/org/openqa/selenium/grid/node/relay/RelaySessionFactory.java
diemol: Where did it get eliminated?

java/src/org/openqa/selenium/grid/data/DefaultSlotMatcher.java
```diff
@@ -44,13 +42,6 @@
  */
 public class DefaultSlotMatcher implements SlotMatcher, Serializable {
 
-  /*
-   List of prefixed extension capabilities we never should try to match, they should be
-   matched in the Node or in the browser driver.
-  */
-  private static final List<String> EXTENSION_CAPABILITIES_PREFIXES =
```
diemol, line 51: I want to keep this because the browser vendors can make changes and we don't want to change things here every time they do it. For example, `moz:debuggerAddress` is now being sent from the client side to enable CDP in Firefox.

diemol (changes requested):

java/src/org/openqa/selenium/grid/data/DefaultSlotMatcher.java
sbabcoc: The core objective of this PR is to migrate all vendor-specific capabilities that should be ignored for purposes of slot matching into "options" objects. (I had initially implemented this to ignore all extension capabilities with "complex" values.) The current strategy of ignoring all extension capabilities with a small number of "special" prefixes while considering all extension capabilities with "non-special" prefixes produces inconsistent behavior and requires vendors to implement special-case slot matchers, each of which is mutually exclusive to every other custom slot matcher.

In your example of `moz:debuggerAddress`, the desire is to ignore this capability for slot matching. What happens if Mozilla adds a new extension capability that should be considered for slot matching? The current strategy doesn't allow for this possibility, forcing Mozilla to implement a custom slot matcher, and for all affected clients to update their Grid configurations to add this matcher. For clients with Grid configurations that supply sessions of browsers from multiple vendors, this could force complete reconfiguration of their automation infrastructure, segregating their session providers into multiple grids by vendor.

With the revised behavior implemented by this PR, `moz:debuggerAddress` would be considered for slot matching. To ignore this option for slot matching, it would be migrated into `moz:firefoxOptions`, I believe this is a more sensible approach, because this capability truly is an option, not an identity specifier.

diemol: >The current strategy of ignoring all extension capabilities with a small number of "special" prefixes while considering all extension capabilities with "non-special" prefixes produces inconsistent behavior and requires vendors to implement special-case slot matchers, each of which is mutually exclusive to every other custom slot matcher

This is an assumption.


If Mozilla adds a new capability for matching, we ignore it and pass it along to GeckoDriver.

java/src/org/openqa/selenium/grid/node/relay/RelaySessionFactory.java
sbabcoc: Deleted line 160 below.

java/src/org/openqa/selenium/grid/data/DefaultSlotMatcher.java
sbabcoc: This is precisely my point. Outside of mutually exclusive custom slot matchers, vendors have no control over how their extension capabilities are handled. Capabilities with "special" prefixes will always be ignored and capabilities with "non-special" prefixes will always be considered. This PR eliminates the inconsistency and adds the ability for vendors to control whether their extension capabilities will be treated as identifying characteristics or as end-node configuration options.

diemol: Please leave the prefixes. Then we can merge the PR.

sbabcoc: I revised my implementation to restore the "special" status of extension capabilities with prefix "se:". I also added explicit filtering for extension capabilities with suffixes "options", "Options", "loggingPrefs", and "debuggerAddress". This captures all existing use cases, which avoids breaking any existing functionality.
I had failed to identify the full collection of existing extension capabilities that needed to be ignored for slot matching. Thanks for your reviews and feedback!

diemol: Please leave the prefixes. Then we can merge the PR.

diemol: @sbabcoc I won't merge the PR because the comments have not been addressed. The request is to leave the prefixes as they were and only focus on ignoring capabilities with a prefix and the name ` options`.

sbabcoc: @diemol I understand your request to reinstate the "special" handling of all extension capabilities with prefixes `goog`, `moz`, and `ms`. Doing so would reintroduce the inconsistency that I'm attempting to eliminate.

In its current form, the revisions in this PR eliminate the special status of `goog`, `moz`, and `ms`.  The only prefix that is entirely ignored is `se`. To continue ignoring the small number of existing vendor-specific extension capabilities, I added explicit checks for these suffixes: "options", "Options", "loggingPrefs", and "debuggerAddress".

There are two capability categories: identity and configuration. The former is considered for slot matching; the latter is ignored for slot matching. The existing implementation of **DefaultSlotMatcher** handles extension capabilities inconsistently, ignoring all capabilities from three "special" vendors (configuration) while considering all other capabilities (identity). Vendors (even the "special" ones) lack the ability to differentiate identity values from configuration values.

If my PR is approved and merged, vendors will have the ability to add identity values and configuration values as needed, with the guarantee that the default slot matcher will treat them as intended.

java/src/org/openqa/selenium/grid/node/relay/RelaySessionFactory.java
```diff
@@ -149,15 +147,6 @@ public Either<WebDriverException, ActiveSession> apply(CreateSessionRequest sess
...
-    // is provided
-    // they are mutually exclusive
-    MutableCapabilities filteredStereotype = new MutableCapabilities(stereotype);
-    if (capabilities.getCapability("appium:app") != null) {
-      filteredStereotype.setCapability(CapabilityType.BROWSER_NAME, (String) null);
-    }
-
-    capabilities = capabilities.merge(filteredStereotype);
```
diemol, line 160: Now I remember. This was done because the user can specify something in the stereotype that will be sent to the Node that will receive the request.

java/src/org/openqa/selenium/grid/data/DefaultSlotMatcher.java
diemol: Why is it an inconsistency? Browser vendors just need those capabilities to be passed along. They do not use the Grid and configure stereotypes to do matching.

diemol (changes requested):

java/src/org/openqa/selenium/grid/data/DefaultSlotMatcher.java
sbabcoc: Every extension capability for Google, Mozilla, and Microsoft is ignored for slot matching; every extension capability for everyone else (including Apple and Appium) is considered for slot matching. This behavior is inconsistent. It also fails to provide vendors with the ability to control which capabilities are treated as identity values and which are treated as configuration options.

diemol: Apple does not need to be there; this might change in the future.

However, Appium does need to be considered, except the `options` one.

Browser vendors just need to have their capabilities passed along. Why do you keep saying that they will use their capabilities to do matching in Grid when this is not true?

sbabcoc: Appium extension capabilities are most certainly being considered for slot matching. This is still true in the revised functionality implemented by this PR. I think the behavior of the current matcher is what necessitated the merge operation and explicit check for 'appium:app'.

sbabcoc: With my current implementation, all existing Appium extension capabilities are considered except for the four suffixes: "options", "Options", "loggingPrefs", and "debuggerAddress". There's implementation in the Appium server that merges the prefixed capabilities into their "options" collection with no thought to which capabilities are "identity" values and which are configuration options. I don't know if this gets reflected back to Grid in a way that's visible to **DefaultSlotMatcher**, but their configuration examples indicate that every option can be declared in either place. I'm attempting to engage the Appium folks to see how they feel about deprecating this pattern in favor of clear differentiation between "identity" values and configuration options.

diemol: I would not prefer to leave `loggingPrefs`, and `debuggerAddress` hardcoded. I prefer to have the 4 prefixes ignored as it was before.

I don't mean to be rude, I am taking your considerations, and I strongly prefer the approach I've been suggesting in previous comments. Please let me know if that works for you or if I should raise the change in a different PR.

sbabcoc: Thanks for remaining engaged in this conversation! I understand your hesitation to alter the established behavior. Perhaps the new behavior can be activated by an optional core Selenium capability. That way, the new behavior will only be present when it's explicitly requested.

Issue #13993, opened 2024-05-21
[🚀 Feature]: Implement high level BiDi network commands


titusfortner:
### Feature and motivation

At the Selenium Dev Summit we agreed on this API to be generally applied across the bindings; we'll keep this labeled beta while we make sure that it works for what is needed

We want the methods to be accessible from a network() method available directly from the Driver class (e.g., `driver.network.addRequestHandler()`, `driver.network().addRequestHandler()`). We can't do everything just like this in all the languages, because, for example, .NET uses events with a += and -= for adding and removing handler events so we don't went "add" and "remove" methods.

### Implementations:
| Method                                     | Java | NodeJS | Python | Ruby | .NET |
|--------------------------------------------|------|--------|--------|------|------|
| `addRequestHandler()` |      |        |  #14592  |    #14751  |      |
| `removeRequestHandler()`                                |      |        |  #14592 |  #14751    |      |
| `clearRequestHandlers()`                          |      |        | #14592 |   #14751     |    |
| `addResponseHandler()`         |      |        |        |      |     |
| `removeResponseHandler()` |      |        |        |      |     |
| `clearResponseHandlers()`            |      |        |        |      |      |
| `addAuthenticationHandler()` |   #14334   |   #14345    | #14592 |   #14523   |      |
| `removeAuthenticationHandler()` |   #14334      |  #14345        | #14592 |  #14523    |   |
| `clearAuthenticationHandlers()` |   #14334      |   #14345      | #14592 |  #14523    |   |

### Considerations:
If we can figure out how to get the "add" methods return an id that can be used by the "remove" methods, that would be a lot easier for or users. Might be too complicated to implement.

titusfortner:
I've been thinking about this, and I think the `add` methods should return an id. This is going to be much easier for languages that want to pass in lambdas without storing a reference to the location os the object in memory.

This is how `moz/addon/install` endpoint works and the id gets passed into the payload to remove the addon.

I was looking for common wisdom here and it generally seems to match — https://softwareengineering.stackexchange.com/questions/314066/restful-api-should-i-be-returning-the-object-that-was-created-updated

p0deje:
> I've been thinking about this, and I think the add methods should return an id. This is going to be much easier for languages that want to pass in lambdas without storing a reference to the location os the object in memory.

Totally fine by me!

pujagani:
"`addJavaScriptErrorHandler`()" is part of the list - I think that is a mistake. It is part of script module right.

[...]

PR #14592, opened 2024-10-13
[py] BiDi Network implementation of Intercepts and Auth in Python

shbenzer:
### **User description**
Add implementations for BiDi Network's Auth and Interception to Python

### Description
Added network.py: 

- Network class
- Event class
- Request class
- add_intercept()
- remove_intercept()
- add_request_handler()
- remove_request_handler()
- continue_request()
- add_auth_handler()
- remove_auth_handler()
- continue_with_auth()
- clear_request_handlers()

Updated websocket_connection.py
Added bidi_network_tests.py

### Motivation and Context
Issue #13993 

### Types of changes

### Checklist

___

### **PR Type**
Enhancement, Tests

___

### **Description**
- Added BiDi Network implementation for network interception and authentication.

- Introduced `Network`, `Request`, and `NetworkEvent` classes for handling network events.

- Extended WebDriver to include a `network` property for BiDi capabilities.

- Added comprehensive tests for BiDi Network functionality, including intercepts and authentication.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>network.py</strong><dd><code>Implement BiDi Network functionality for intercepts and authentication</code></dd></summary>
<hr>

py/selenium/webdriver/common/bidi/network.py

<li>Implemented <code>Network</code> class for network interception and authentication.<br> <li> Added <code>Request</code> and <code>NetworkEvent</code> classes for handling network requests <br>and events.<br> <li> Provided methods for adding/removing intercepts, request handlers, and <br>authentication handlers.<br> <li> Introduced support for continuing requests and handling <br>authentication.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14592/files#diff-d04fc5573acd6ca5bcc8167940f134f1a3ead8a5230c066c077a1319a1ffe1d4">+328/-0</a>&nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>webdriver.py</strong><dd><code>Extend WebDriver with BiDi Network integration</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

<li>Integrated <code>Network</code> class into WebDriver via a new <code>network</code> property.<br> <li> Ensured BiDi WebSocket connection is initialized for network <br>operations.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14592/files#diff-c87c150ed2a08f5293db58a06da8bf7f14c3a5582586c6270c1be661dfdd0985">+12/-0</a>&nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>bidi_network_tests.py</strong><dd><code>Add tests for BiDi Network functionality</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/test/selenium/webdriver/common/bidi_network_tests.py

<li>Added tests for BiDi Network initialization and intercept <br>functionality.<br> <li> Verified request handling, continuation, and authentication <br>mechanisms.<br> <li> Marked tests as expected to fail on Safari.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14592/files#diff-97c8d35b88402ef5de934c8fb95da5f7c7c04039a79baf20cc8fd6c91379dac7">+76/-0</a>&nbsp; &nbsp; </td>

</tr>

___

### DIFF
--- a/py/selenium/webdriver/common/bidi/network.py
+++ b/py/selenium/webdriver/common/bidi/network.py
@@ -111,7 +111,8 @@ class Network:
             If intercept is None, all intercepts will be removed.
         """
         if intercept is None:
-            for intercept_id in self.intercepts:  # remove all intercepts
+            intercepts_to_remove = self.intercepts.copy()  # create a copy before iterating
+            for intercept_id in intercepts_to_remove:  # remove all intercepts
                 self.conn.execute(self.command_builder("network.removeIntercept", {"intercept": intercept_id}))
                 self.intercepts.remove(intercept_id)
         else:


shbenzer:
was adding this to BUILD.bazel for easy testing:
```
py_test(
    name = "bidi_network_tests",
    srcs = ["test/selenium/webdriver/common/bidi_network_tests.py"],
    deps = [
        ":init-tree",
        ":selenium",
        ":webserver",
    ] + TEST_DEPS,
)
```
so i could run `bazel test //py:bidi_network_tests`

shbenzer:
Think it's ready now... @AutomatedTester could you give it a look?

p0deje (commented):
I'm not a Python expert but I ended up being the person who drafted the initial BiDi implementation. I tried to avoid using the `async/await` pattern because I assumed that it can't be used in a sync code that the majority of Python clients of the library do. Is there any reason to use the pattern in this PR?

shbenzer:
> I'm not a Python expert but I ended up being the person who drafted the initial BiDi implementation. I tried to avoid using the `async/await` pattern because I assumed that it can't be used in a sync code that the majority of Python clients of the library do. Is there any reason to use the pattern in this PR?

It has pros and cons, but I think you're right that it isn't needed and could be more trouble than it's worth. I'll remove it for the time being and if we decide later it's an easy shift.

py/selenium/webdriver/remote/command.py
```diff
@@ -25,7 +25,11 @@ class Command:
 
         https://w3c.github.io/webdriver/
     """
-
+    ADD_INTERCEPT: str = "network.addIntercept"
```
p0deje: We don't need those, do we?

py/selenium/webdriver/remote/remote_connection.py
```diff
@@ -36,6 +36,11 @@
 LOGGER = logging.getLogger(__name__)
 
 remote_commands = {
+    Command.ADD_INTERCEPT: ("POST", "/session/$sessionId/network/intercept"),
```
p0deje: We don't need those, do we?

py/test/selenium/webdriver/common/bidi_network_tests.py
```diff
@@ -0,0 +1,50 @@
...
+
+from selenium import webdriver
+from selenium.webdriver.common.bidi.network import Network
+
+
+def url_for(page):
+    return webserver.where_is(page)
+
+@pytest.fixture
+def driver():
```
p0deje: There are some existing fixtures we can use for these tests, please see an example of usage in https://github.com/SeleniumHQ/selenium/blob/d9739391458e315d2dbff99dcf068444bde5db61/py/test/selenium/webdriver/common/bidi_script_tests.py

p0deje (commented):
Can we add some high-level API examples and tests per an original issue?

- driver.network.add_request_handler(...)
- driver.network.add_response_handler(...)

This would allow us to surface the API usage and make sure it works nicely via lambdas.

py/test/selenium/webdriver/common/bidi_network_tests.py
shbenzer: You're right

py/selenium/webdriver/remote/remote_connection.py
shbenzer: I thought we might, but I'll defer to you

py/selenium/webdriver/remote/command.py
shbenzer: I thought we might, but I'll defer to you

shbenzer:
> Can we add some high-level API examples and tests per an original issue?
> 
> * driver.network.add_request_handler(...)
> * driver.network.add_response_handler(...)
> 
> This would allow us to surface the API usage and make sure it works nicely via lambdas.

Could you provide a bit more guidance here? What would those two tests entail that isn't already being covered? the tests I added are an expanded version of the .spec from the ruby implementation

p0deje:
I think the current tests expose the low-level API that users are not supposed to call. The high-level API unfortunately is not implemented in any bindings so there is no example to follow. However, the proposal is to allow the add_request_handler to accept a matcher (e.g. URL filter for what requests should be handled) and a callback that received the request to handle. This callable is responsible for continuing/failing the request optionally allowing to mutate it (e.g. add headers).

```python
def handle_request(request):
    request.headers["X-FOO"] = "bar"
    request.continue()

driver.network.add_request_handler('https://google.com', handle_request)
```

This is roughly what we had considered to be a user-facing API.

shbenzer:
> I think the current tests expose the low-level API that users are not supposed to call. The high-level API unfortunately is not implemented in any bindings so there is no example to follow. However, the proposal is to allow the add_request_handler to accept a matcher (e.g. URL filter for what requests should be handled) and a callback that received the request to handle. This callable is responsible for continuing/failing the request optionally allowing to mutate it (e.g. add headers).
> 
> ```python
> def handle_request(request):
>     request.headers["X-FOO"] = "bar"
>     request.continue()
> 
> driver.network.add_request_handler('https://google.com', handle_request)
> ```
> 
> This is roughly what we had considered to be a user-facing API.

Interesting. I think I understood - what do you think about this most recent push?

py/test/selenium/webdriver/common/bidi_network_tests.py
```diff
@@ -0,0 +1,53 @@
...
+    password = 'your_password'
+    
+    network.add_intercept(phases=[Network.PHASES['auth_required']])
+    network.on('auth_required', lambda event: network.continue_with_auth(event['requestId'], username, password))
+    
+    driver.get(url_for("basicAuth"))
+    assert driver.find_element_by_tag_name('h1').text == 'authorized'
+
+def test_add_request_handler(driver, network):
+    network.add_request_handler(callback=lambda event: network.continue_request(event['requestId'], url=url_for("basicAuth")))
```
p0deje: 1. Let's change the order of params here (url, callback). 
2. The callback should receive a `request` object that allows it to inspect its contents via methods (e.g. headers, url, etc). It's better to represent it this way than a raw dict.
3. The `network.continue_request` should be `request.continue()`. This would allow us to keep `network` API compact.

shbenzer:
tests passing locally
```
(selenium_dev_curr) (base) simonbenzer@Simons-MBP selenium % bazel test //py:bidi_network_tests                                                
WARNING: Ignoring JAVA_HOME, because it must point to a JDK, not a JRE.
Starting local Bazel server and connecting to it...
WARNING: For repository 'aspect_bazel_lib', the root module requires module version aspect_bazel_lib@2.8.1, but got aspect_bazel_lib@2.9.1 in the resolved dependency graph.
INFO: Analyzed target //py:bidi_network_tests (395 packages loaded, 10836 targets configured).
INFO: Found 1 test target...
Target //py:bidi_network_tests up-to-date:
  bazel-bin/py/bidi_network_tests
INFO: Elapsed time: 8.816s, Critical Path: 1.27s
INFO: 29 processes: 4 internal, 25 darwin-sandbox.
INFO: Build completed successfully, 29 total actions
//py:bidi_network_tests                                                  PASSED in 0.7s
```

py/selenium/webdriver/common/bidi/network.py
```diff
@@ -0,0 +1,125 @@
...
+        'before_request': 'beforeRequestSent',
+        'response_started': 'responseStarted',
+        'auth_required': 'authRequired'
+    }
+
+    def __init__(self, conn):
+        self.conn = conn
+        self.callbacks = {}
+
+    def continue_response(self, request_id, status_code, headers=None, body=None):
```
p0deje: All the methods except for the ones mentioned in https://github.com/SeleniumHQ/selenium/issues/13993 should be private.

py/test/selenium/webdriver/common/bidi_network_tests.py
```diff
@@ -0,0 +1,53 @@
+import pytest
+
+from selenium import webdriver
+from selenium.webdriver.common.bidi.network import Network
+
+
+def url_for(page):
```
p0deje: You can use `pages` fixture to avoid this method:

```
        pages.load("javascriptPage.html")
```

py/test/selenium/webdriver/common/bidi_network_tests.py
```diff
@@ -0,0 +1,53 @@
...
+    network.add_intercept(phases=[Network.PHASES['auth_required']])
+    network.on('auth_required', lambda event: network.continue_with_auth(event['requestId'], username, password))
+    
+    driver.get(url_for("basicAuth"))
+    assert driver.find_element_by_tag_name('h1').text == 'authorized'
+
+def test_add_request_handler(driver, network):
+    network.add_request_handler(callback=lambda event: network.continue_request(event['requestId'], url=url_for("basicAuth")))
+    driver.get(url_for("basicAuth"))
+    assert driver.find_element_by_tag_name('h1').text == 'authorized'
```
p0deje: We need to add some tests for interception of requests to make sure it's working. There are some examples in Ruby that use CDP, but maybe you can adopt them for these cases: https://github.com/SeleniumHQ/selenium/blob/02a2bfcb923e4e633ee78dd26ecaec3b3e4ffd12/rb/spec/integration/selenium/webdriver/devtools_spec.rb#L166-L215

p0deje:
> Interesting. I think I understood - what do you think about this most recent push?

Much closer to what we are aiming for! Let's add some abstractions (Request/Response) objects, move continuation there, and write more tests. I've left more detailed comments in the PR review.

py/test/selenium/webdriver/common/bidi_network_tests.py
shbenzer: In regards to 3, continue is reserved in python

p0deje: Let's call it `continue_request()` and `continue_response()` accordingly. Still, keep those inside `Request/Response` objects respectively.

shbenzer:
@p0deje Is this closer?

p0deje:
> @p0deje Is this closer?

Not quite, we need to make Request/Response be passed to the callbacks instead of plain `event`, see https://github.com/SeleniumHQ/selenium/pull/14592#issuecomment-2477187410 for the example.

shbenzer:
> > @p0deje Is this closer?
> 
> Not quite, we need to make Request/Response be passed to the callbacks instead of plain `event`, see [#14592 (comment)](https://github.com/SeleniumHQ/selenium/pull/14592#issuecomment-2477187410) for the example.

Hmm... Seems I misunderstood. Let me see if I got it now:

1. move add_response_handler, remove_response_handler, add_request_handler, and remove_request_handler back into Network class
2. in the callback_on_url_match() wrapper function inside each add_handler:
```
        def callback_on_url_match(data):
            if url_pattern in data['request']['url']:
                callback(data)
```
it should create a request/response object, then perform the callback on that object instead of on the event itself.
3. The response and request classes should just be simple objects with the continue method (and mark continue public). e.g.
```
class Request:
    def __init__(self, request_id, url, method, headers, body, network: Network):
        self.request_id = request_id
        self.url = url
        self.method = method
        self.headers = headers
        self.body = body
        self.network = network

    def continue_request(self):
      """Continue after sending a request."""
      params = {
          'requestId': self.request_id
      }
      if self.url is not None:
          params['url'] = url
      if self.method is not None:
          params['method'] = method
      if self.headers is not None:
          params['headers'] = headers
      if self.postData is not None:
          params['postData'] = postData
      self.network.conn.execute('network.continueRequest', params)
```

shbenzer:
@p0deje I did some updates based on my comment above. If I still haven't quite understood, I can just revert the commits lol

py/selenium/webdriver/common/bidi/network.py
```diff
@@ -0,0 +1,174 @@
...
+                request_id = data['request'].get('requestId')
+                url = data['request'].get('url')
+                method = data['request'].get('method')
+                headers = data['request'].get('headers', {})
+                body = data['request'].get('postData', None)
+                request = Request(request_id, url, method, headers, body, self)
+                callback(request)
+        self.__on('before_request', callback_on_url_match)
+
+    def remove_request_handler(self):
```
p0deje: Registering multiple intercepts is possible, so `remove_request_handler` should accept an ID of intercept that should be removed. 

Overall it would look like this:

1. User registers a new intercept via `add_request_handler`.
2. The method subscribes to the `before_request` event if it's the first registered callback.
3. The method creates a dictionary where a key is a unique ID of the callback (e.g. 1).
4. The method returns the ID to the user which they can store in a variable.
5. 1-4 can be repeated multiple times to register separate callbacks each having a unique ID. For example, one callback blocking requests to some URL, one callback logging all requests, etc. However, we don't need to subscribe to `before_request` again we have the subscription established already.
6. ... I am not sure what the behavior of event handling is going to be exactly, it's being discussed in https://seleniumhq.slack.com/archives/CBH302726/p1731689236280019
7. Eventually, when the user wants to remove a specific handler, then can pass the ID to `remove_request_handler().
8. If all callbacks are removed, the client unsubscribes from the `before_request` event.

py/test/selenium/webdriver/common/bidi_network_tests.py
```diff
@@ -0,0 +1,72 @@
...
+
+from selenium import webdriver
+from selenium.webdriver.common.bidi.network import Network
+from selenium.webdriver.common.bidi.network import Response
+from selenium.webdriver.common.bidi.network import Request
+
+
+@pytest.fixture
+def network(driver):
+    yield Network(driver)
```
p0deje: Why not access it as `driver.network` instead?

py/selenium/webdriver/common/bidi/network.py
shbenzer: Just applied the changes; I also generated a more in depth docstring for add_request_handler and add_response_handler. Let me know what you tihnk

shbenzer:
@AutomatedTester could you provide your thoughts here as well?

shbenzer:
Based on this [W3C Chart](https://wpt.fyi/results/webdriver/tests/bidi?label=master&label=stable&aligned) Safari must still be marked as xfail

shbenzer:
@p0deje @pujagani do we know why `network.addIntercept` is receiving `Unrecognized Command` errors?

```
platform linux -- Python 3.8.19, pytest-7.4.4, pluggy-1.3.0
rootdir: /mnt/engflow/worker/work/1/exec/bazel-out/k8-fastbuild/bin/py/common-chrome-bidi-test/selenium/webdriver/common/bidi_network_tests.py.runfiles/_main/py
configfile: pyproject.toml
plugins: instafail-0.5.0, trio-0.8.0, mock-3.12.0
collected 6 items
py/test/selenium/webdriver/common/bidi_network_tests.py::test_add_response_handler[chrome] FAILED [ 16%]
______________________ test_add_response_handler[chrome] _______________________
network = <selenium.webdriver.common.bidi.network.Network object at 0x7fcc289c6400>
    @pytest.mark.xfail_safari
    def test_add_response_handler(network):
        passed = [False]
    
        def callback(response):
            if response.status_code == 200:
                passed[0] = True
            response.continue_response()
    
>       network.add_response_handler(callback)
py/test/selenium/webdriver/common/bidi_network_tests.py:22: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
py/selenium/webdriver/common/bidi/network.py:158: in add_response_handler
    self.__add_intercept(phases=[self.PHASES['response_started']])
py/selenium/webdriver/common/bidi/network.py:55: in __add_intercept
    self.conn.execute('network.addIntercept', params)
py/selenium/webdriver/remote/webdriver.py:384: in execute
    response = self.command_executor.execute(driver_command, params)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <selenium.webdriver.chromium.remote_connection.ChromiumRemoteConnection object at 0x7fcc289c6100>
command = 'network.addIntercept'
params = {'contexts': None, 'phases': ['responseStarted'], 'sessionId': '563acf6031fd090c74e119bded4856de', 'urlPatterns': None}
    def execute(self, command, params):
        """Send a command to the remote server.
    
        Any path substitutions required for the URL mapped to the command should be
        included in the command parameters.
    
        :Args:
         - command - A string specifying the command to execute.
         - params - A dictionary of named parameters to send with the command as
           its JSON payload.
        """
        command_info = self._commands.get(command) or self.extra_commands.get(command)
>       assert command_info is not None, f"Unrecognised command {command}"
E       AssertionError: Unrecognised command network.addIntercept
py/selenium/webdriver/remote/remote_connection.py:393: AssertionError
```

shbenzer:
I think I figured it out, could you run the merge workflow again @pujagani ?

shbenzer:
@p0deje hoping this fix does it - if you have a better idea let me know

shbenzer:
Able to initialize connections now. However connections are getting lost because of InvalidArgumentExceptions

```
py/test/selenium/webdriver/common/bidi_network_tests.py::test_remove_authentication_handler[chrome] FAILED [100%]
__________________ test_remove_authentication_handler[chrome] __________________

driver = <selenium.webdriver.chrome.webdriver.WebDriver (session="0ad0ececb7125dcf6a0e586d8bde047c")>
pages = <conftest.pages.<locals>.Pages object at 0x1079a4a60>

    @pytest.mark.xfail_safari
    def test_remove_authentication_handler(driver, pages):
>       driver.network.add_authentication_handler('test', 'test')

py/test/selenium/webdriver/common/bidi_network_tests.py:87: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
py/selenium/webdriver/common/bidi/network.py:102: in add_authentication_handler
    self.__add_intercept(phases=[self.PHASES['auth_required']])
py/selenium/webdriver/common/bidi/network.py:57: in __add_intercept
    self.conn.execute(command)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <selenium.webdriver.remote.websocket_connection.WebSocketConnection object at 0x1079a47f0>
command = {'command': 'network.addIntercept', 'id': 2, 'params': {'contexts': None, 'phases': ['authRequired'], 'urlPatterns': None}}

    def execute(self, command):
        self._id += 1
        payload = self._serialize_command(command)
        payload["id"] = self._id
        if self.session_id:
            payload["sessionId"] = self.session_id
    
        data = json.dumps(payload)
        logger.debug(f"-> {data}"[: self._max_log_message_size])
        self._ws.send(data)
    
        self._wait_until(lambda: self._id in self._messages)
        response = self._messages.pop(self._id)
    
        if "error" in response:
>           raise Exception(response["error"])
E           Exception: invalid argument

py/selenium/webdriver/remote/websocket_connection.py:67: Exception

------------------------------ live log logreport ------------------------------
ERROR    websocket:_logging.py:77 Connection to remote host was lost. - goodbye

=========================== short test summary info ============================
FAILED py/test/selenium/webdriver/common/bidi_network_tests.py::test_add_response_handler[chrome]
FAILED py/test/selenium/webdriver/common/bidi_network_tests.py::test_remove_response_handler[chrome]
FAILED py/test/selenium/webdriver/common/bidi_network_tests.py::test_add_request_handler[chrome]
FAILED py/test/selenium/webdriver/common/bidi_network_tests.py::test_remove_request_handler[chrome]
FAILED py/test/selenium/webdriver/common/bidi_network_tests.py::test_add_authentication_handler[chrome]
FAILED py/test/selenium/webdriver/common/bidi_network_tests.py::test_remove_authentication_handler[chrome]
========================= 6 failed, 1 passed in 19.79s =========================
```

shbenzer:
@p0deje 

Current status: Only 1 test is currently passing (test_network_initialized()), however the browser now recognizes commands as legitimate seemingly. I am not sure why callbacks are not being called, as the returned request_ids from add_request/response_handler() are Null

Possible causes: 
- race-conditions (should we make implementation async again?)
- incorrect implementation of add_intercept (could be why auth handler is not working in pages.load(basicAuth))

p0deje:
@shbenzer You might need to add some waits to give the time for the callback to be called. Start by adding `sleep(5)` and see if it stabilizes the test. If yes, you can figure out a better waiting strategy. If no, there is probably something wrong with the implementation itself.

shbenzer:
@p0deje Tried some sleeps (e.g. 

```
@pytest.mark.xfail_safari
def test_add_response_handler(driver, pages):
    passed = [False]

    def callback(response):
        passed[0] = True
        response.continue_response()

    driver.network.add_response_handler(callback)
    pages.load("basicAuth")
    sleep(10)
    assert passed[0], "Callback was NOT successful"
``` 

and failures persist

shbenzer:
@aguspe you have any suggestions?

p0deje:
@shbenzer Any chance you could take a look at https://github.com/SeleniumHQ/selenium/pull/14900? The implementation @aguspe made works and you can follow it to make the Python bindings part.

aguspe:
Oh sorry @shbenzer I think I got your notification lost and I marked everything as read :)

I can give it a deeper look over the weekend, but maybe the issue is here:

`   def __on(self, event, callback):
        """Set a callback function to subscribe to a network event."""
        event = self.EVENTS.get(event, event)
        self.callbacks[event] = callback
        if len(self.subscriptions[event]) == 0:
            session_subscribe(self.conn, event, self.__handle_event)`
           
The part that confused me the most was not how to add a callback, but also to have the listener listening for the event, it looks right but maybe you can print out to the console the responses and the requests

I noticed a couple of errors while I was implementing it, and by debugging it that way I noticed that I was using the listener wrongly :)

shbenzer:
Haven't had a ton of time lately, but should hopefully be able to take a good look over the next two weeks. I'm currently getting this error when trying to test though

```
Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
In file included from external/abseil-cpp~/absl/base/log_severity.cc:15:
external/abseil-cpp~/absl/base/log_severity.h:18:10: fatal error: 'array' file not found
   18 | #include <array>
      |          ^~~~~~~
1 error generated.
Target //py:bidi_network_tests failed to build
INFO: Elapsed time: 39.731s, Critical Path: 0.64s
INFO: 440 processes: 433 internal, 7 darwin-sandbox.
ERROR: Build did NOT complete successfully
//py:bidi_network_tests                                         FAILED TO BUILD

Executed 0 out of 1 test: 1 fails to build.
```

shbenzer:
> Oh sorry @shbenzer I think I got your notification lost and I marked everything as read :)
> 
> I can give it a deeper look over the weekend, but maybe the issue is here:

No worries @aguspe ! If you can find the time though, please do take a look - it'll likely save me a good amount of time.

shbenzer:
@aguspe and @p0deje So I finally got bazel working again, and have a few thoughts on the implementation:

Bear in mind I wrote this pr 2 months ago now lol, so I'm trying to get back into this.

1. I don't like the command iterator. I added it because just doing `self.conn.execute(command)` wasn't working because it was requiring an iterator to execute. `script.py` uses `self.conn.execute(session_subscribe(LogEntryAdded.event_class))`, can we utilize session_subscribe like this for network? execute() takes in a command and serializes it itself, so I'm not sure if session_subscribe() will accomplish what we need. but maybe add_callback() would (see 2)

2. the ruby bindings use the `add_callback()` function from the bidi object, the equivalent in python is `self.conn.add_callback()` and it's used by `script.py`. Do we think we can shift to use that instead? Also, do we think we should be storing the callbacks inside the network object instead of using that functionality in the conn object? (or evend do both?)

3. The problems occur/tests fail in the communication with the bidi connection - it's not able to add, remove, continue intercepts/auth. The only thing performing correctly is the establishment of the connection. I think the communication/command problems would either be fixed or made easier to solve by points 1 and 2.

shbenzer:
Also, do we know who wrote `websocketconnection.py` and/or `script.py`? I'd love for them to chime in here

p0deje:
1. I used iterator because that's how original CDP implementation was written and I wanted to preserve the pattern. We can implement something else if you want. `session_subscribe` can definitely be re-used across all BiDi APIs.
2. Yes, let's use connection callbacks for this. You can use them in network or move them there, it's an implementation detail from my point of view.
3. :+1:

> Also, do we know who wrote websocketconnection.py and/or script.py? I'd love for them to chime in here

I wrote those two 😄

shbenzer:
Switched to using add_callback() in __on(). Getting dropped websocket connections which I'm not sure the cause of, and that might be the cause of the failures. Not sure otherwise - I think we're close though.

```
------------------------------ live log logreport ------------------------------
ERROR    websocket:_logging.py:77 Connection to remote host was lost. - goodbye

py/test/selenium/webdriver/common/bidi_network_tests.py::test_add_request_handler[chrome] FAILED [ 57%]
_______________________ test_add_request_handler[chrome] _______________________

driver = <selenium.webdriver.chrome.webdriver.WebDriver (session="58c6c81e3fbfa3a444b4a3f1611d88af")>
pages = <conftest.pages.<locals>.Pages object at 0x104e2d2e0>

    @pytest.mark.xfail_safari
    def test_add_request_handler(driver, pages):
        passed = [False]
    
        def callback(request):
            passed[0] = True
            request.continue_request()
    
        driver.network.add_request_handler(callback)
        pages.load("basicAuth")
>       assert passed[0], "Callback was NOT successful"
E       AssertionError: Callback was NOT successful
E       assert False

py/test/selenium/webdriver/common/bidi_network_tests.py:62: AssertionError

------------------------------ live log logreport ------------------------------
ERROR    websocket:_logging.py:77 Connection to remote host was lost. - goodbye

py/test/selenium/webdriver/common/bidi_network_tests.py::test_remove_request_handler[chrome] FAILED [ 71%]
_____________________ test_remove_request_handler[chrome] ______________________

driver = <selenium.webdriver.chrome.webdriver.WebDriver (session="83972e585cd798470b4bce36673f79fc")>
pages = <conftest.pages.<locals>.Pages object at 0x104c51a30>

    @pytest.mark.xfail_safari
    def test_remove_request_handler(driver, pages):
        passed = [False]
    
        def callback(request):
            passed[0] = True
            request.continue_request()
    
        test_request_id = driver.network.add_request_handler(callback)
>       driver.network.remove_request_handler(request_id=test_request_id)

py/test/selenium/webdriver/common/bidi_network_tests.py:73: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
py/selenium/webdriver/common/bidi/network.py:169: in remove_request_handler
    self.__remove_intercept(request_id=request_id)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <selenium.webdriver.common.bidi.network.Network object at 0x104c51610>
intercept = None, request_id = None

    def __remove_intercept(self, intercept=None, request_id=None):
        """Remove an intercept from the network."""
        if request_id is not None:
            command = {'method': 'network.removeIntercept', 'requestId': request_id}
            self.conn.execute(self.command_iterator(command))
        elif intercept is not None:
            command = {'method': 'network.removeIntercept', 'intercept': intercept}
            self.conn.execute(self.command_iterator(command))
        else:
>           raise ValueError('Either requestId or intercept must be specified')
E           ValueError: Either requestId or intercept must be specified

py/selenium/webdriver/common/bidi/network.py:91: ValueError

```

py/selenium/webdriver/common/bidi/network.py
```diff
@@ -0,0 +1,262 @@
...
+                # create request object to pass to callback
+                request_id = data['request'].get('requestId', None)
+                url = data['request'].get('url')
+                method = data['request'].get('method')
+                headers = data['request'].get('headers', {})
+                body = data['request'].get('body', None)
+                request = Request(request_id, url, method, headers, body, self)
+                callback(request)
+        self.__on('before_request', callback_on_url_match)
+        self.callbacks[request_id] = callback
```
p0deje: I don't think it is ever going to work - it will always end up being `{None: callback}`. This way, the callback will never really be called. It's likely a key issue why your tests are failing.

Looking at Ruby implementation, you need to return callback ID rather than request ID:
https://github.com/SeleniumHQ/selenium/blob/b2cea1debe78b6554371987357ae1c54b30ec5bf/rb/lib/selenium/webdriver/common/network.rb#L61-L71

shbenzer:
hmm... @p0deje I unified request/response_id with the callback_id returned by add_callback (by updating __on() to return it) - however the issue persists. It was a good catch, but seem there is a deeper flaw.

shbenzer:
I updated the callback_on_url() construction in add_request/response_handler() to create a unique request_id if not provided with one in data, but it might be better to throw an exception instead (or simply set it as None):

```
def callback_on_url_match(data):
  # create response object to pass to callback
  if url_pattern in data['response']['url']:
      request_id = data['request'].get('requestId', uuid.uuid4())
      ...
```

p0deje:
@shbenzer Do you not want to continue working on the PR?

shbenzer:
> @shbenzer Do you not want to continue working on the PR?

I accidentally deleted the branch when I was cleaning up my fork (didn't think it'd close the pr but it's been reopened now) - I'm still working on it. However I'd love if someone could pull and give it a look to see if they can figure out the current blockage.

shbenzer:
Adding this to Build.Bazel for easy testing:

```
[
    py_test_suite(
        name = "common-%s-bidi-network" % browser,
        size = "large",
        srcs = glob(
            [
                "test/selenium/webdriver/common/**/bidi_network_tests.py",
            ],
        ),
        args = [
            "--instafail",
            "--bidi=true",
        ] + BROWSERS[browser]["args"],
        data = BROWSERS[browser]["data"],
        env_inherit = ["DISPLAY"],
        tags = ["no-sandbox"] + BROWSERS[browser]["tags"],
        deps = [
            ":init-tree",
            ":selenium",
            ":webserver",
        ] + TEST_DEPS,
    )
    for browser in BROWSERS.keys()
]
```

py/selenium/webdriver/remote/remote_connection.py
AutomatedTester: Not sure we will need these

py/selenium/webdriver/common/bidi/network.py
```diff
@@ -0,0 +1,326 @@
...
+        params = {"requestId": self.request_id}
+        if self.url is not None:
+            params["url"] = self.url
+        if self.method is not None:
+            params["method"] = self.method
+        if self.headers is not None:
+            params["headers"] = self.headers
+        if self.body is not None:
+            params["body"] = self.body
+        command = {"method": "network.continueRequest", "params": params}
```
AutomatedTester: For a future PR, we might want to have a utility method that builds the commands to make things easier to read

AutomatedTester (approved):

shbenzer:
@AutomatedTester Thanks for the review, do you think you could provide any insight into why these tests are failing? The browser is hanging after adding an intercept it seems.

py/selenium/webdriver/common/bidi/network.py
```diff
@@ -0,0 +1,326 @@
...
+            params = {
+                "phases": phases,
+            }
+        elif contexts is None:
+            params = {"phases": phases, "urlPatterns": url_patterns}
+        elif url_patterns is None:
+            params = {"phases": phases, "contexts": contexts}
+        else:
+            params = {"phases": phases, "contexts": contexts, "urlPatterns": url_patterns}
+        command = {"method": "network.addIntercept", "params": params}
```
navin772: We should either use `Command.ADD_INTERCEPT` here in the method since we are declaring it in `command.py` (which is currently not utilized) or remove those commands from `command.py`.

shbenzer:
Just finished a full rewrite, here is the current status:

1. `add_intercept()` PASS
2. `remove_intercept()` PASS
3. `add_request_handler()` PASS
4. `remove_request_handler()` PASS
5. `continue_request()` FAIL
6. `add_auth_handler()` PASS
7. `continue_with_auth()` FAIL

I suspect the problem with continue_request() and continue_with_auth() stems from the construction/implementation of the Request object, but I am unsure.

AutomatedTester (approved):

AutomatedTester (changes requested):
The tests are failing for this

shbenzer:
> The tests are failing for this

Yep, the `continue_request()` and `continue_with_auth()` functions aren't working - I was hoping you could help me figure out why.

AutomatedTester:
> > The tests are failing for this
> 
> Yep, the `continue_request()` and `continue_with_auth()` functions aren't working - I was hoping you could help me figure out why.

Ok, let me have a look

shbenzer:
@AutomatedTester find anything?

shbenzer:
@AutomatedTester find anything? If not, I think I’ll close this and we’ll merge in #14738. We can do needed refactors (e.g., implement the api per #13993) after.

I have verified that using add_request_handler() sets a callback on event, then when a network request is intercepted, that callback is called successfully. The request objects have valid request_ids, but continue_request() is not functioning correctly.

shbenzer:
Firefox is now passing - but Chrome and Edge still fail for some reason

shbenzer:
Any thoughts @p0deje @AutomatedTester ?

navin772:
@shbenzer this seems like a known issue with Chromium BiDi - https://github.com/SeleniumHQ/selenium/issues/15553. We had a discussion on slack TLC channel regarding this (pinged you there).

shbenzer:
`//py:common-firefox-bidi-test/selenium/webdriver/common/driver_element_finding_tests.py TIMEOUT in 2 out of 2 in 600.2s`

Unrelated failure due to flaky test

AutomatedTester (approved):

AutomatedTester:
LGTM! Thanks!

cgoldberg:
congrats on landing this!

shbenzer:
> congrats on landing this!

Thank you!

gabriel-swapcard:
How do I use the url_patterns for add_request_handler? I'm trying different formats, but I always get an invalid argument exception

shbenzer:
> How do I use the url_patterns for add_request_handler? I'm trying different formats, but I always get an invalid argument exception

It should be a list of these objects:

```
network.UrlPatternString = {
    type: "string",
    pattern: text,
}
```

See: https://w3c.github.io/webdriver-bidi/#type-network-UrlPattern and https://w3c.github.io/webdriver-bidi/#command-network-addIntercept

If you find a problem (which is certainly possible given the newness of our implementations, as well as the browsers') please create an issue with a reproducible example for us to diagnose.

PR #14593, opened 2024-10-13
[dotnet] Allow setting of PageDimensions and PageMargins in PrintOptions directly

shbenzer:
### **User description**
PrintOptions.cs was missing setters for PageMargins and PageDimensions.

### Description
added setters in PrintOptions.cs for PageMargins and PageDimensions.

### Motivation and Context
fixes issue

### Types of changes

### Checklist

___

### **PR Type**
Bug fix

___

### **Description**
- Added missing setters for `PageDimensions` and `PageMargins` in the `PrintOptions` class to allow modification of these properties.
- This change addresses an issue where these properties were previously read-only, limiting their usability.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>PrintOptions.cs</strong><dd><code>Add missing setters for PageDimensions and PageMargins</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/PrintOptions.cs

<li>Added a setter for <code>PageDimensions</code> property.<br> <li> Added a setter for <code>PageMargins</code> property.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14593/files#diff-b7252358e015a773e2d453e794cf8c614fcd330985e0ddf4fe3202392f508e2c">+2/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

___


dotnet/src/webdriver/PrintOptions.cs
nvborisenko, line 1: Please don't forget to adjust inline xml documentation for the changed properties. E.g. before: `Gets ...`, after `Gets or sets ...`

dotnet/src/webdriver/PrintOptions.cs
```diff
@@ -106,6 +106,7 @@ public bool ShrinkToFit
         public PageSize PageDimensions
         {
             get { return pageSize; }
+            set { pageSize = value; }
```
nvborisenko: If we really want to allow set the value, then we need to add basic validation (for `null`).

dotnet/src/webdriver/PrintOptions.cs
shbenzer: Good catch, thank you!

dotnet/src/webdriver/PrintOptions.cs
nvborisenko: Or allow it to be `null`? Then we should allow `null` for other properties, and then in code consider everywhere that properties might be null to avoid `NullReferenceExceptioon`.

From my point of view allowing it to be `null` is good idea, it represents an optional parameter according specification.

nvborisenko (changes requested):
Allow properties to be `null` or not. Nullable properties is very good idea (see inline comment).

dotnet/src/webdriver/PrintOptions.cs
```diff
@@ -199,6 +201,11 @@ internal Dictionary<string, object> ToDictionary()
                 toReturn["shrinkToFit"] = this.shrinkToFit;
             }
 
+            if (this.pageSize == null)
```
nvborisenko: Now it is better, which makes unnecessary to keep `this.pageSize` field initialized: https://github.com/SeleniumHQ/selenium/blob/4c2491faa3207cce1e9890bac16e98dfc7e5ee2e/dotnet/src/webdriver/PrintOptions.cs#L55

nvborisenko: Ah not, it is behavior breaking change. Please keep it as is.

dotnet/src/webdriver/PrintOptions.cs
```diff
@@ -199,6 +201,11 @@ internal Dictionary<string, object> ToDictionary()
                 toReturn["shrinkToFit"] = this.shrinkToFit;
             }
 
+            if (this.pageSize == null) 
+            { 
+                this.pageSize = new PageSize(); 
+            }
+            
             if (this.pageSize.Height != DefaultPageHeight || this.pageSize.Width != DefaultPageWidth)
```
nvborisenko: It can be optimized to avoid allocation: `if (this.pageSize != null && (this.pageSize.Height != DefaultPageHeight || this.pageSize.Width != DefaultPageWidth))`

nvborisenko:
Good progress here, I am going to checksum (using CI). If pass, the I vote for removing `DefaultPageHeight` value on client side, because I hope it should be handled oon server side.

nvborisenko:
We don't have test for `Print` command. @shbenzer do you want to add new tests?

shbenzer:
@nvborisenko sure, I wrote the documentation examples - they should easily port over.

shbenzer:
Done

dotnet/test/common/PrintOptionsTest.cs
nvborisenko, line 1: Absolutely incorrect:
- We should use NUnit test framework
- There is no asserts
- There is no `Print` command invocation

I propose to take java tests as a reference to test `Print` command (hoping it exists :) )

nvborisenko, line 1: Thank you, I will help to adopt it to follow principles

nvborisenko: BTW do you use AI to write code?

nvborisenko:
I looked at java implementation, it stores default values on client side. So let's be in sync.
- Add check for `null` when setting value (do not allow private value to be null)
- Add tests that value cannot be null
- Be in sync with java implementation when serializing object to json

shbenzer:
> * Add check for `null` when setting value (do not allow private value to be null)

Do we want to throw an error when setting as null, or do we want to just set the default in that case and let it continue?

nvborisenko:
> > * Add check for `null` when setting value (do not allow private value to be null)
> 
> Do we want to throw an error when setting as null, or do we want to just set the default in that case and let it continue?

Definitely throw argument null exception.

nvborisenko (approved):
Looks good, thank you

shbenzer:
@nvborisenko Could you explain the code changes you made?

Particularly PageSizeDictionary()s changes, I can see the logic in the changes to the main sets, as well as refining the test example to just test construction instead of construction inside of the print command.

nvborisenko:
Only for you..

1.
image
The same and simpler.

2.
image
In this case value is `double` (I don't remember), which is never can be `null`.

3.
image
`Assert` is correct, not `assert` which even could not be compiled. And the goal of this test is that we cannot set property to `null`, we don't test Print command.

shbenzer:
Lol thank you

Those all make sense to me

Issue #13865, opened 2024-04-24
[🚀 Feature]: Automatically drain node after n-failed session attempts


krishtoautomate:
### Feature and motivation

User might have n number of nodes and if sessions are randomly allocated during the tests and are failing on few nodes due to issue with docker node. It will be great if this feature is implemented: Automatically drain node after n-failed session attempts.

### Usage example

when new RemoteWebDriver session is creating on n-number of nodes with 100-200 test cases in parallel, if I have issues with any docker nodes, those nodes will drain automatically after n(5-10) failed session attempts. the total impact on test execution is less and I end up re-executing only few tests which failed initially.

diemol:
Why is the Docker node failing? How are you monitoring that? Can you share more details? It sounds like we want to fix something with a workaround instead of a proper fix.

krishtoautomate:
Failed Reasons can be any, proxy or internet or chrome crashing or not launching due to driver mismatch. Provided as an example to explain use case. Node can be appium node as well.

diemol:
But that is a Node misconfiguration. Infrastructure needs to be tested before being made available for use. Implementing something like this would hide issues. It is an incomplete workaround.

krishtoautomate:
if we have too many nodes, sometimes our nodes might fall behind for different reasons and we dont want our tests to impact due to nodes which are not working. is their any way this feature can be implemented with some cli argument to make it active and default as off

diemol:
If you have too many nodes, how are you monitoring them? Why would there be a driver mismatch? Are you not testing the changes before sharing the modified infrastructure with the rest of the users?

If a Node is not working, testing the changes done to it should alert you even before you run any regular tests.

Suppose this feature request is implemented, and you have a driver mismatch or network issues with your nodes. What ends up happening is that all of them get shut down. How would you diagnose the actual problem? The result is that the infrastructure went down, and you need to figure out why.

PR #14628, opened 2024-10-21
[grid] Improvement for Node handling

### LINKED ISSUE #13769: [🐛 Bug]: org.openqa.selenium.WebDriverException: Unable to route (POST) /session/0405a24bf1d06c6ae8402b8d12027934/...
### What happened?

When executing a command on a just timedout session a `WebDriverException: Unable to route ...` is raised.

For some seconds after the session timedout there is a inconsistent state, until the `sessionCleanupNodeService` of the `LocalNode` did run. The `LocalSessionMap` still tells the session does exist, but the `Node.isSessionOwner` does return false. This will prevent the route inside the `Node` is matched and lead to the `WebDriverException: Unable to route ...`.

### Relevant log output

```shell
N/A
```

### Operating System

Win 10 x64

### Selenium version

4.17.1

N/A

N/A

N/A

VietND96:

### Motivation and Context
checkSessionCount() - to check and drain node after sessions are reached. Steps in order, so in logs, we saw something like

> 09:51:06.344 INFO [LocalNode.checkSessionCount] - Draining Node, configured sessions value (1) has been reached.
09:51:06.533 INFO [LocalNode.newSession] - Session created by the Node. Id: b47edd11098edb876c805745161ea67c, Caps:

It might confuse that the session created in a draining node.
So, moving it to block finally { }
The trade-off is the session count will include both success and unsuccess. When the session comes, the remaining count decreases.
It could be suitable for autoscaling. For example, if a faulty Node (for unknown reasons), the session cannot be created, and the node will be drained to signal the scaler to scale a new one. Instead of keeping Node running, it biases the scaler already had enough nodes running.
Also, there was a request #13865 - Automatically drain node after n-failed session attempts. By moving that check to `finally` blocks, it would be best for both.

---
Add the `drain()` method to the shutdown hook. Node will send drain event to remove Node immediately once Runtime stopped.

---
Fixes #13769 - Exception `NoSuchSessionException` should throw out when command is routed to Node that session just timed out.

### Types of changes

### Checklist


java/src/org/openqa/selenium/grid/node/Node.java
```diff
@@ -152,7 +152,10 @@ protected Node(
                     req ->
                         getSessionId(req.getUri())
                             .map(SessionId::new)
-                            .map(this::isSessionOwner)
+                            .map(
+                                sessionId ->
+                                    this.getSession(sessionId) != null
```
joerg1985: I do not understand this change here, it does not help match the route. The `this.getSession(sessionId)` should return null here, in case of #13769.
The `NodeTest.shouldOnlyRespondToWebDriverCommandsForSessionsTheNodeOwns` should fail, as soon as we enter here.

VietND96: I saw the test failure and updated something new, instead of testing against random UUID

java/src/org/openqa/selenium/grid/node/ForwardWebDriverCommand.java
```diff
@@ -17,19 +17,30 @@
...
     this.node = Require.nonNull("Node", node);
   }
 
+  @Override
+  public boolean matches(HttpRequest req) {
+    return getSessionId(req.getUri())
+        .map(id -> node.isSessionOwner(new SessionId(id)))
+        .orElse(false);
+  }
+
```
joerg1985: Is this code reached? In my mind the the `isSessionOwner` must be checked inside the `execute` before calling  `node.executeWebDriverCommand`.

VietND96: You are correct! I updated and tried to check via a unit test. Can you review one more time

java/test/org/openqa/selenium/grid/e2e/SessionTimeoutTest.java
```diff
@@ -0,0 +1,51 @@
+// Licensed to the Software Freedom Conservancy (SFC) under one
```
diemol: This test could go inside the `router` package. There are some others that do similar actions.

java/test/org/openqa/selenium/grid/e2e/SessionTimeoutTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.chrome.ChromeOptions;
+import org.openqa.selenium.grid.Bootstrap;
+import org.openqa.selenium.remote.RemoteWebDriver;
+
+public class SessionTimeoutTest {
+  @Test
+  void testSessionTimeout() throws Exception {
+    Assumptions.assumeThat(System.getProperty("webdriver.chrome.binary")).isNull();
+    Bootstrap.main(("hub --host 127.0.0.1 --port 4444").split(" "));
```
diemol: Check other tests inside the `router` package, there are nicer ways to place the parameters.

java/test/org/openqa/selenium/grid/e2e/SessionTimeoutTest.java
```diff
@@ -0,0 +1,51 @@
...
+    Bootstrap.main(
+        ("node --host 127.0.0.1 --port 5555 --session-timeout 12 --selenium-manager true")
```
diemol: This test must be added to `.skipped-tests` because it cannot be executed in engflow. 

Or, find a way to set the driver, just like the `StressTest` does it.

diemol (changes requested):

java/test/org/openqa/selenium/grid/e2e/SessionTimeoutTest.java
VietND96: I moved test to `StressTest` and handle rely on current setup() in there

java/test/org/openqa/selenium/grid/router/StressTest.java
```diff
@@ -124,4 +132,15 @@ void multipleSimultaneousSessions() throws Exception {
 
     CompletableFuture.allOf(futures).get(4, MINUTES);
   }
+
+  @Test
+  void testStopTimedOutSession() throws Exception {
+    assertThat(server.isStarted()).isTrue();
+    WebDriver driver =
```
diemol: Perhaps I am too picky, but this class has a specific purpose (Stress Test). To make sense, we either rename the class to something different or move the new test to its class.

diemol (changes requested):

java/test/org/openqa/selenium/grid/router/StressTest.java
VietND96: I updated it to `multipleSimultaneousSessionsTimedOut()`

diemol (approved):

Issue #14480, opened 2024-09-09
[🚀 Feature]: [dotnet] Road to be AOT friendly


nvborisenko:
### Feature and motivation

Any consumer of Selenium library in .NET cannot use it within publishing his application as [Native AOT Deployment](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/?tabs=net7%2Cwindows).

The detailed affected parts are coming...

**Selenium Manager** ✅ **v4.26**
- [x] Selenium Manager - #14481 

**W3C WebDriver** ✅ **v4.26**
- [x] New Session
- [x] ... - #14574

**W3C BiDi** ✅ **v4.28**
- [x] ... https://github.com/SeleniumHQ/selenium/pull/15162

**CDP (DevTools)**
- [ ] Generate JsonSerializerContext for all DTO models
- [ ] ...

**Fine tuning**
- Address all analyzer warnings
- [ ] ...

### Usage example

Create console app, publish as native AOT, it fails at runtime.

nvborisenko:
Blocked by  https://github.com/bazelbuild/rules_dotnet/issues/443

RenderMichael:
I’m interested in this feature for running Selenium. Often times, the roads to AOT compatibility leads to improvements across the board, even for non-AOT scenarios.

NCLnclNCL:
> Blocked by [bazelbuild/rules_dotnet#443](https://github.com/bazelbuild/rules_dotnet/issues/443)

when it will be fixed

nvborisenko:
> Blocked by https://github.com/bazelbuild/rules_dotnet/issues/443

Fixed, waiting release soon.

> Often times, the roads to AOT compatibility leads to improvements across the board, even for non-AOT scenarios.

Who knows potential issues... , I cannot prepare decomposed tasks to archive the goal. Our mission here is to adopt existing codebase to be ready for AOT.

The technical enabler is resolved, so the list of impacted areas will be be growing so fast soon.

nvborisenko:
@RenderMichael are you able to test the preview version of Selenium which seems is compatible with AOT (excepting CDP)? Your input would be valuable. I can share nuget package file here (built locally).

RenderMichael:
@nvborisenko I would be happy to test out the AOT compatibility of Selenium :)

nvborisenko:
[...]

PR #14639, opened 2024-10-24
[dotnet] Lazy-load Selenium manager binary location

RenderMichael:
### **User description**

### Description

The location of the selenium manager binary is calculated when it is first accessed.

### Motivation and Context

I ran into this while working on https://github.com/SeleniumHQ/selenium/issues/14480

It is considered bad practice for static constructors to throw exceptions. If an exception is throws, the type is rendered unusable for the rest of the application lifecycle. The stack-trace of the exception is also much harder to follow, and the debugging experience is worsened. The fact that it throws `TypeInitializationException` threw me off initially, because I was testing AOT at the time.

Exceptions are expected on this flow, in exceptional cases, they are even thrown directly. With this change, the Stack trace changes:

Before:
image

After:
image

### Types of changes

### Checklist

___

### **PR Type**
Enhancement

___

### **Description**
- Implemented lazy-loading for the Selenium Manager binary path to improve performance and avoid exceptions during static initialization.
- Replaced the static constructor with a method `GetBinaryFullPath` to determine the binary path dynamically.
- Enhanced error handling by providing more informative exception messages when the binary is not found or the platform is unsupported.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>SeleniumManager.cs</strong><dd><code>Implement lazy-loading for Selenium Manager binary path</code>&nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/SeleniumManager.cs

<li>Introduced lazy-loading for the Selenium Manager binary path.<br> <li> Replaced static constructor with a method to determine binary path.<br> <li> Improved error handling for unsupported platforms.<br> <li> Enhanced exception messages for missing binaries.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14639/files#diff-df9a0140c4ef310382b15de2b26d228971d00530d18f4c1bd3c74f735c3af665">+11/-9</a>&nbsp; &nbsp; </td>

</tr>                    

___

### DIFF
--- a/dotnet/src/webdriver/SeleniumManager.cs
+++ b/dotnet/src/webdriver/SeleniumManager.cs
@@ -229,12 +229,13 @@ namespace OpenQA.Selenium
     {
         public record LogEntryResponse(string Level, string Message);
 
-        public record ResultResponse(
-            [property: JsonPropertyName("driver_path")]
-            string DriverPath,
-            [property: JsonPropertyName("browser_path")]
-            string BrowserPath)
+        public record ResultResponse
         {
+            [JsonPropertyName("driver_path")]
+            public string? DriverPath { get; init; }
+
+            [JsonPropertyName("browser_path")]
+            public string? BrowserPath { get; init; }
         }
     }
 
--- a/dotnet/src/webdriver/SeleniumManager.cs
+++ b/dotnet/src/webdriver/SeleniumManager.cs
@@ -229,13 +229,12 @@ namespace OpenQA.Selenium
     {
         public record LogEntryResponse(string Level, string Message);
 
-        public record ResultResponse(string DriverPath, string BrowserPath)
+        public record ResultResponse(
+            [property: JsonPropertyName("driver_path")]
+            string DriverPath,
+            [property: JsonPropertyName("browser_path")]
+            string BrowserPath)
         {
-            [JsonPropertyName("driver_path")]
-            public string DriverPath { get; } = DriverPath;
-
-            [JsonPropertyName("browser_path")]
-            public string BrowserPath { get; } = BrowserPath;
         }
     }
 
--- a/dotnet/src/webdriver/SeleniumManager.cs
+++ b/dotnet/src/webdriver/SeleniumManager.cs
@@ -26,6 +26,8 @@ using System.Text;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
+#nullable enable
+
 namespace OpenQA.Selenium
 {
     /// <summary>
@@ -77,7 +79,7 @@ namespace OpenQA.Selenium
         /// <returns>
         /// An array with two entries, one for the driver path, and another one for the browser path.
         /// </returns>
-        public static Dictionary<string, string> BinaryPaths(string arguments)
+        public static Dictionary<string, string?> BinaryPaths(string arguments)
         {
             StringBuilder argsBuilder = new StringBuilder(arguments);
             argsBuilder.Append(" --language-binding csharp");
@@ -88,7 +90,7 @@ namespace OpenQA.Selenium
             }
 
             var smCommandResult = RunCommand(_lazyBinaryFullPath.Value, argsBuilder.ToString());
-            Dictionary<string, string> binaryPaths = new()
+            Dictionary<string, string?> binaryPaths = new()
             {
                 { "browser_path", smCommandResult.BrowserPath },
                 { "driver_path", smCommandResult.DriverPath }
@@ -182,7 +184,8 @@ namespace OpenQA.Selenium
 
             try
             {
-                jsonResponse = JsonSerializer.Deserialize<SeleniumManagerResponse>(output, _serializerOptions);
+                jsonResponse = JsonSerializer.Deserialize<SeleniumManagerResponse>(output, _serializerOptions)
+                    ?? throw new WebDriverException("SeleniumManagerResponse was null");
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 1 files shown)


dotnet/src/webdriver/SeleniumManager.cs
```diff
@@ -36,27 +36,28 @@ public static class SeleniumManager
     {
         private static readonly ILogger _logger = Log.GetLogger(typeof(SeleniumManager));
 
-        private static readonly string BinaryFullPath = Environment.GetEnvironmentVariable("SE_MANAGER_PATH");
+        private static string? _binaryFullPath;
+        private static string BinaryFullPath => _binaryFullPath ??= GetBinaryFullPath();
```
nvborisenko: Cunning line of code. If we would use `Lazy<string> _lazyBinaryFullPath`, then it would look more straightforward and simple.

```csharp
private static Lazy<string> _lazyBinaryFullPath = new Lazy<string>(...);
private static string BinaryFullPath => _lazyBinaryFullPath.Value;
```

And now seems we don't need `field` and `property` with identical purposes, just use `_lazyBinaryFullPath.Value` everywhere.

nvborisenko (changes requested):
Thank you! It can be even simplified to avoid cunning logic, please see inner comment.

dotnet/src/webdriver/SeleniumManager.cs
RenderMichael: Good call. PR has been updated.

RenderMichael:
Thanks for the feedback, the PR has been updated

dotnet/src/webdriver/SeleniumManager.cs
```diff
@@ -36,27 +36,25 @@ public static class SeleniumManager
...
-        private static readonly string BinaryFullPath = Environment.GetEnvironmentVariable("SE_MANAGER_PATH");
-
         private static readonly JsonSerializerOptions _serializerOptions = new() { PropertyNameCaseInsensitive = true, TypeInfoResolver = SeleniumManagerSerializerContext.Default };
 
-        static SeleniumManager()
+        private static readonly Lazy<string> _lazyBinaryFullPath = new(() =>
         {
-
-            if (BinaryFullPath == null)
+            string? binaryFullPath = Environment.GetEnvironmentVariable("SE_MANAGER_PATH");
```
nvborisenko, line 46: Nullable reference types are not yet enabled in this project, sorry :( I understand you, but let's not produce new warnings.

RenderMichael: Nullable reference types were invented with this scenario in mind, and can be enabled file-by-file (or even more granular) with `#nullable enable`

Would it be fine to make those kinds of changes when we touch adjacent code, as long as it is not too invasive?

nvborisenko: Let's go with `#nullable enable` for this file, it is fine. This is how I am going to introduce this feature in BiDi namespace.

nvborisenko: Introducing it for project will require fixing warnings immediately, what nobody wants to do :)

RenderMichael: I enabled NRT for this file, and fixed ~~two~~ three hypothetical null reference exceptions with webdriver exceptions

> what nobody wants to do :)

:)

nvborisenko: If you don't mind I pushed changes regarding optional properties in json model. Are you OK with these changes?

RenderMichael: Unfortuanately the change to `ResultReponse` will not work. From the docs:
> The parameter names of a parameterized constructor must match the property names and types. Matching is case-insensitive, and the constructor parameter must match the actual property name even if you use [[JsonPropertyName]](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonpropertynameattribute) to rename a property.

RenderMichael: Additionally, this does not remove any nullability issues - this goes back to the null reference exception style. Any constructor parameter that is not found is populated with `default` ie. `null` for strings. This is changing in v9 actually, behind a feature flag: https://devblogs.microsoft.com/dotnet/system-text-json-in-dotnet-9/#respecting-non-optional-constructor-parameters

RenderMichael: I made one final push - to make the properties required and to suppress the null warning - with the knowledge that this only gets instantiated by JSON serialization

nvborisenko: Thank you, good collaboration.

> The parameter names of a parameterized constructor must match the property names and types. Matching is case-insensitive, and the constructor parameter must match the actual property name even if you use [[JsonPropertyName]](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonpropertynameattribute) to rename a property.

It worked though.

nvborisenko (approved):
Thank you!

RenderMichael:
Thank you @nvborisenko !

I dug into the record syntax that worked, and you were right, it was working. There may be some special support for records. However, the moment I added `[JsonRequired]`, I got an exception:
image
image

I'm not familiar with the STJ esoterica here, but it looks like what we have now works. Maybe once v9 comes out, the internal serialization options could enable the feature flags I linked and prevent null reference exceptions even further.

Issue #14640, opened 2024-10-24
[🚀 Feature]: .NET: Support nullable reference types


RenderMichael:
### Feature and motivation

Annotate the code for nullable reference type support. This would be massively beneficial for clarity and to prevent null reference exceptions, both within Selenium and for users.

### Usage example

![image](https://github.com/user-attachments/assets/44811002-17d2-424f-921a-c9ab62fb04ee)
![image](https://github.com/user-attachments/assets/1d5d7127-1873-4d57-8ec4-fb7d70eec19f)

## Affected namespaces

- [x] `OpenQA.Selenium`
- [x] `OpenQA.Selenium.BiDi`
- [x] `OpenQA.Selenium.BiDi.Communication`
  - [x] Nested namespaces
- [x] `OpenQA.Selenium.BiDi.Modules`
  - [x] Nested namespaces
- [x] `OpenQA.Selenium.Chrome`
- [x] `OpenQA.Selenium.Chromium`
- [x] `OpenQA.Selenium.DevTools`
  - [x] Nested namespaces
- [x] `OpenQA.Selenium.Edge`
- [x] `OpenQA.Selenium.Firefox`
- [x] `OpenQA.Selenium.Firefox.Internal`
- [x] `OpenQA.Selenium.IE`
- [x] `OpenQA.Selenium.Interactions`
- [x] `OpenQA.Selenium.Interactions.Internal`
- [x] `OpenQA.Selenium.Internal`
- [x] `OpenQA.Selenium.Internal.Logging`
- [x] `OpenQA.Selenium.Remote`
- [x] `OpenQA.Selenium.Safari`
- [x] `OpenQA.Selenium.Support.UI`
- [x] `OpenQA.Selenium.VirtualAuth`
- [x] Handle nullability on CDP-generated types
- [x] Enable Nullable Reference Types in the `.csproj` and `bazel`

RenderMichael:
If Selenium is interested in accepting NRT, I'll gladly take up the task

nvborisenko:
It would be great, and it would be challenging as for you as for us :) Only one request: try keeping PRs small.

nvborisenko:
@RenderMichael do you mind to join [Selenium Slack channel](https://join.slack.com/t/seleniumhq/shared_invite/zt-2stlcmc6b-Hww~3r3yAzquhSq4riOFxA) for messaging instantly?

nvborisenko:
[...]

PR #14672, opened 2024-10-29
[dotnet] Enable NRT on exceptional types

RenderMichael:
### Description

Adds nullable reference type annotations to exceptional types, such as exception types.

This also includes the polyfill nullability attributes, so this PR may block some other nullability work.

Contributes to #14640 

### Motivation and Context

Some easy wins on types whose nullability status is fairly unambiguous.

### Types of changes

### Checklist

___

### **PR Type**
enhancement

___

### **Description**
- Enabled nullable reference types across multiple files to improve null safety.
- Updated exception classes to use nullable types for constructor parameters.
- Added polyfill for nullable attributes to ensure compatibility with older .NET versions.
- Enhanced `FirefoxDriver` with nullable reference types and added `MemberNotNullWhen` attribute.
- Updated `IFileDetector` and `DefaultFileDetector` with `NotNullWhen` attribute for method parameters.

___

PRDescriptionHeader.CHANGES_WALKTHROUGH
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>15 files</summary><table>
<tr>
  <td>
    <details>
      <summary><strong>DefaultFileDetector.cs</strong><dd><code>Enable nullable reference types and add NotNullWhen attribute</code></dd></summary>
<hr>

dotnet/src/webdriver/DefaultFileDetector.cs

<li>Enabled nullable reference types.<br> <li> Added <code>NotNullWhen</code> attribute to <code>IsFile</code> method parameter.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14672/files#diff-5985d640cf10c81f2513d99ac1a05ce4886bf714d0ccacd8c99fa5cb87ead70a">+5/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>DetachedShadowRootException.cs</strong><dd><code>Enable nullable reference types for exception class</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/DetachedShadowRootException.cs

- Enabled nullable reference types.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14672/files#diff-480b9c7189baac4635e6fabe703d2802ce3d76b526f96bddfb0594a709c35ef0">+2/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>CommandResponseException.cs</strong><dd><code>Update exception constructors with nullable parameters</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/DevTools/CommandResponseException.cs

<li>Enabled nullable reference types.<br> <li> Updated constructor parameters to nullable types.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14672/files#diff-09e5131aeff292821687495f17ab73be8a2865fc813c3c711afdbbc6318b51ae">+4/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>DriverServiceNotFoundException.cs</strong><dd><code>Update exception constructors with nullable parameters</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/DriverServiceNotFoundException.cs

<li>Enabled nullable reference types.<br> <li> Updated constructor parameters to nullable types.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14672/files#diff-84f761477aae375efa85e7a478d5054b18970ded7ac42276650427663e5d2adb">+5/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>ElementClickInterceptedException.cs</strong><dd><code>Update exception constructors with nullable parameters</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/ElementClickInterceptedException.cs

<li>Enabled nullable reference types.<br> <li> Updated constructor parameters to nullable types.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14672/files#diff-5b47d8c7b1eabd95230ca9d12dddf0fd27250e511f456e60196b31a450d6b00d">+4/-3</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>ElementNotInteractableException.cs</strong><dd><code>Update exception constructors with nullable parameters</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/ElementNotInteractableException.cs

<li>Enabled nullable reference types.<br> <li> Updated constructor parameters to nullable types.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14672/files#diff-f0bc93719028efe470988fd783d3fa6773990eed4c92ffa8d63f8b1de3ae57d7">+5/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>ElementNotSelectableException.cs</strong><dd><code>Update exception constructors with nullable parameters</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/ElementNotSelectableException.cs

<li>Enabled nullable reference types.<br> <li> Updated constructor parameters to nullable types.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14672/files#diff-38c139c39d57059d91648ce9f7cd005399f72cb9deffb13bc5ea698ed145c229">+5/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>ElementNotVisibleException.cs</strong><dd><code>Update exception constructors with nullable parameters</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/ElementNotVisibleException.cs

<li>Enabled nullable reference types.<br> <li> Updated constructor parameters to nullable types.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14672/files#diff-a270f8d08a7a2230b6f72d3634941c7f9df2c7da6047fb81e7993ab3e538cb47">+4/-3</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>ErrorResponse.cs</strong><dd><code>Update ErrorResponse with nullable fields and parameters</code>&nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/ErrorResponse.cs

<li>Enabled nullable reference types.<br> <li> Updated fields and constructor parameters to nullable types.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14672/files#diff-590fc1beb0ca79f92c02382589b0db3dc3acec08df8019f45dd48e4d9e1c0cdd">+10/-8</a>&nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>FirefoxDriver.cs</strong><dd><code>Enhance FirefoxDriver with nullable reference types</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/Firefox/FirefoxDriver.cs

<li>Enabled nullable reference types.<br> <li> Added <code>MemberNotNullWhen</code> attribute for <code>HasActiveDevToolsSession</code>.<br> <li> Updated method parameters to nullable types.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14672/files#diff-337f94531a5e0d064a23d898667e0770f6a953ff7c264c4ba55110c5e7485fd2">+22/-3</a>&nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>IFileDetector.cs</strong><dd><code>Enable nullable reference types and add NotNullWhen attribute</code></dd></summary>
<hr>

dotnet/src/webdriver/IFileDetector.cs

<li>Enabled nullable reference types.<br> <li> Added <code>NotNullWhen</code> attribute to <code>IsFile</code> method parameter.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14672/files#diff-390281df5880487409bdb1f76f82cd7dffd27a4dba5b22f8d908edc9f8163450">+5/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>InsecureCertificateException.cs</strong><dd><code>Update exception constructors with nullable parameters</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/InsecureCertificateException.cs

<li>Enabled nullable reference types.<br> <li> Updated constructor parameters to nullable types.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14672/files#diff-dc889989570e31402f77d23fc61c1588abebc2c42d0cf719eccd01b959d6f819">+5/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>NullableAttributes.cs</strong><dd><code>Add polyfill for nullable attributes</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/Internal/NullableAttributes.cs

- Added polyfill for nullable attributes for compatibility.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14672/files#diff-cfe6d372e5bd681491df8538cab8feffca78cbf36b0a3af9d6530d939de97112">+149/-0</a>&nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>InvalidCookieDomainException.cs</strong><dd><code>Update exception constructors with nullable parameters</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/InvalidCookieDomainException.cs

<li>Enabled nullable reference types.<br> <li> Updated constructor parameters to nullable types.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14672/files#diff-e9ccf39ef29e61eba169927ea88d15f396921625bdce509fb41138aaf434f1ca">+4/-3</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>InvalidElementStateException.cs</strong><dd><code>Update exception constructors with nullable parameters</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/InvalidElementStateException.cs

<li>Enabled nullable reference types.<br> <li> Updated constructor parameters to nullable types.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14672/files#diff-f49066198c2c19319db9fff029ea76f11ec19d133e701550e31493b9cc3c2f0e">+4/-3</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

___


dotnet/src/webdriver/DetachedShadowRootException.cs
```diff
@@ -17,7 +17,8 @@
 // </copyright>
 
 using System;
-using System.Runtime.Serialization;
+
+#nullable enable
```
nvborisenko, line 22: Missed nullable message in ctor

dotnet/src/webdriver/StaleElementReferenceException.cs
```diff
@@ -68,7 +69,7 @@ public StaleElementReferenceException(string message, Exception innerException)
         /// </summary>
         /// <param name="message">The original message for exception</param>
         /// <returns>The final message for exception</returns>
-        protected static string GetMessage(string message)
+        protected static string GetMessage(string? message)
         {
             return message + "; " + supportMsg + supportUrl;
```
nvborisenko: if `message` is null then exception in runtime? something like FormatException? I mean please double check that string concatenation works with `null` (AFAIR it doesn't)

dotnet/src/webdriver/StackTraceElement.cs
```diff
@@ -43,18 +45,18 @@ public StackTraceElement()
...
         /// <param name="elementAttributes">A <see cref="Dictionary{K, V}"/> containing the names and values for the properties of this <see cref="StackTraceElement"/>.</param>
-        public StackTraceElement(Dictionary<string, object> elementAttributes)
+        public StackTraceElement(Dictionary<string, object>? elementAttributes)
         {
             if (elementAttributes != null)
             {
                 if (elementAttributes.ContainsKey("className") && elementAttributes["className"] != null)
                 {
-                    this.className = elementAttributes["className"].ToString();
+                    this.className = elementAttributes["className"].ToString() ?? "";
```
nvborisenko: It is reasonable to convert it to: `TryGetValue(...)` and assign if not `null`. Otherwise do nothing, because of field is already initializaed with defauls in ctor.

dotnet/src/webdriver/NoSuchElementException.cs
```diff
@@ -69,7 +70,7 @@ public NoSuchElementException(string message, Exception innerException)
         /// </summary>
         /// <param name="message">The original message for exception</param>
         /// <returns>The final message for exception</returns>
-        protected static string GetMessage(string message)
+        protected static string GetMessage(string? message)
         {
             return message + "; " + supportMsg + supportUrl;
```
nvborisenko: `null` + "abc" - not sure it works.

dotnet/src/webdriver/NoSuchDriverException.cs
```diff
@@ -69,7 +70,7 @@ public NoSuchDriverException(string message, Exception innerException)
         /// </summary>
         /// <param name="message">The original message for exception</param>
         /// <returns>The final message for exception</returns>
-        protected static string GetMessage(string message)
+        protected static string GetMessage(string? message)
         {
             return message + "; " + supportMsg + supportUrl;
```
nvborisenko: `null` + "abc" - not sure it works.

dotnet/src/webdriver/StaleElementReferenceException.cs
nvborisenko: I checked, it works. Hm, trying to recall when I saw it, and why I always use `$"{a}{b}"` now.

dotnet/src/webdriver/Firefox/FirefoxDriver.cs
```diff
@@ -372,7 +391,7 @@ public void UninstallAddOn(string addOnId)
         public Screenshot GetFullPageScreenshot()
         {
             Response screenshotResponse = this.Execute(GetFullPageScreenshotCommand, null);
-            string base64 = screenshotResponse.Value.ToString();
+            string base64 = screenshotResponse.Value.ToString()!;
```
nvborisenko: We should double-check whether `.Value` might be `null`. I mean not here, in general. Probably `Value` always not null.

dotnet/src/webdriver/InvalidSelectorException.cs
```diff
@@ -68,7 +69,7 @@ public InvalidSelectorException(string message, Exception innerException)
         /// </summary>
         /// <param name="message">The original message for exception</param>
         /// <returns>The final message for exception</returns>
-        protected static string GetMessage(string message)
+        protected static string GetMessage(string? message)
         {
             return message + "; " + supportMsg + supportUrl;
```
nvborisenko: Flashback, just to recall why it works (see another comment)

dotnet/src/webdriver/UnhandledAlertException.cs
```diff
@@ -27,11 +28,10 @@ namespace OpenQA.Selenium
     [Serializable]
     public class UnhandledAlertException : WebDriverException
     {
-        private string alertText;
-
         /// <summary>
         /// Initializes a new instance of the <see cref="UnhandledAlertException"/> class.
         /// </summary>
+        [Obsolete("Use a constructor overload that sets alertText")]
```
nvborisenko: It is unnecessary to deprecate some method. I guess we should understand whether `AlertText` might be null or not.

dotnet/src/webdriver/Internal/NullableAttributes.cs
```diff
@@ -0,0 +1,149 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
nvborisenko: We should put a link here where we stole it.

dotnet/src/webdriver/DetachedShadowRootException.cs
RenderMichael: Fixed

dotnet/src/webdriver/StaleElementReferenceException.cs
RenderMichael: That's a valid thought, I converted all the `GetMessage` method to interpolated strings.

> why I always use `$"{a}{b}"` now

On older runtimes, string interpolation always converts to `string.Format`. Sometimes it can simplify to `string.Concat`. But starting in .NET 6, there was a huge overhaul and the compiler takes care of a lot of magic for us. I think it's always better.

dotnet/src/webdriver/StackTraceElement.cs
RenderMichael: I converted `StackTraceElement` and `ErrorResponse` to use `TryGetValue(...)`, it looks much better now.

dotnet/src/webdriver/NoSuchElementException.cs
RenderMichael: I converted this method to interpolated string as well.

dotnet/src/webdriver/Firefox/FirefoxDriver.cs
RenderMichael: More validation is music to my ears :)

There are two real boundaries in Selenium: user input and the server. In the latter case, we basically always have an understanding of the expected response. Maybe we can have a generic `Response.GetValue<T>()` method that automatically de-serializes? Here it would look like `screenshotResponse.GetValue<string>()`. 

Or perhaps the Execute method can be generic, like so? 
```csharp
Response<string> screenshotResponse = this.Execute<string>(GetFullPageScreenshotCommand, null);
```

In this particular case, we are not actually doing null checking, since it `.Value` were null, we would get a null reference on `ToString()`. The reason we suppress the null warning is because an untyped `object.ToString()` is unfortunately annotated as nullable (almost every type I have seen replaces this with a non-nullable string).

So a "better" approach might be to cast like `string base64 = (string)screenshotResponse.Value;`, since we know the response to be a string.

dotnet/src/webdriver/NoSuchDriverException.cs
RenderMichael: Converted to interpolated string.

dotnet/src/webdriver/InvalidSelectorException.cs
RenderMichael: Converted to interpolated string.

dotnet/src/webdriver/Internal/NullableAttributes.cs
RenderMichael: Resolved on Slack to avoid adding our own nullable attributes.

dotnet/src/webdriver/UnhandledAlertException.cs
RenderMichael: Fixed

RenderMichael:
@nvborisenko PR feedback addressed, PTAL

dotnet/src/webdriver/ErrorResponse.cs
```diff
@@ -42,52 +44,41 @@ public ErrorResponse()
...
         /// <param name="responseValue">A <see cref="Dictionary{K, V}"/> containing names and values of
         /// the properties of this <see cref="ErrorResponse"/>.</param>
-        public ErrorResponse(Dictionary<string, object> responseValue)
+        public ErrorResponse(Dictionary<string, object?>? responseValue)
         {
             if (responseValue != null)
             {
-                if (responseValue.ContainsKey("message"))
+                if (responseValue.TryGetValue("message", out object? messageObj)
+                    && messageObj?.ToString() is string message)
```
nvborisenko, line 48: `&& messageObj is not null` instead of `&& messageObj?.ToString() is string message`?

RenderMichael: The problem is, untyped `object.ToString()` is annotated to return a `string?`. we need to handle that potentially null value too.

We can suppress it like before, with `messageObj.ToString()!`

nvborisenko: Ah, I like `objects` in selenium.. Looked through where and how this class used, and seems it makes sense to make fields nullable.

RenderMichael: I reworked the class to allow nullable fields. Let me know if it needs any more changes

RenderMichael:
@nvborisenko I think everything's been addressed in this PR, it's ready for another look

nvborisenko:
Sorry for the delay, paid work. Thanks, will review it (90% is already reviewed). For me it is conceptual basis for further upcoming PRs (especially for attributes), thus thanks for understanding.

RenderMichael:
@nvborisenko Absolutely no worries! 

Yes the attributes are important here, and if we are serious about investing in AOT compatibility, those attributes are even more important (not realistic to suppress those warnings the same way as for nullability).

[Here is a guide from Microsoft](https://devblogs.microsoft.com/dotnet/creating-aot-compatible-libraries/#targetframeworks) that I use for AOT trimming, with an example of the two options for polyfilling attributes to netstandard (same guideline applies for nullability and AOT attributes).

I think adding the attributes ourselves is fine as long as we hide behind `#if !NET8_0_OR_GREATER` and make them `internal` (can duplicate or somehow share for the support package)

nvborisenko:
Regarding AOT compatibility, we already made a big step forward: we made classic command/responses which are `objects` properly serialized. Let's keep this progress: CDP and BiDi namespaces promise to be easier, at least it is strongly-typed.

I understand the importance of attributes for nullability, and thanks for sharing the link: now I can vote for `Approach 2: Define the attributes internally`. If you are seriously motivated, please continue - you are founder!

RenderMichael:
Changes are now implemented! Thanks for the feedback

dotnet/src/webdriver/ErrorResponse.cs
```diff
@@ -102,38 +86,22 @@ public ErrorResponse(Dictionary<string, object> responseValue)
         /// <summary>
         /// Gets or sets the message from the response
         /// </summary>
-        public string Message
-        {
-            get { return this.message; }
-            set { this.message = value; }
-        }
+        public string? Message { get; } = string.Empty;
```
YevgeniyShunevych: The `Message` property differs from its siblings. I think this property should either become non-nullable or lose its `string.Empty` default value. After the PR changes and before, as I see this property is never set to `null` and logically (according to constructor logic) should not return `null`, so I would make it non-nullable.

RenderMichael: That's a good point. It's worth noting that the `stackTrace` check doesn't work, because it comes back as a `string` instead of a list of stack frames, and there's an extra `error` entry in the scenarios I tested this against.

But I guess that's a separate issue

dotnet/src/webdriver/Internal/NullableAttributes.cs
```diff
@@ -0,0 +1,165 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+#if !NET8_0_OR_GREATER
+
+// Original code in https://github.com/dotnet/runtime/blob/419e949d258ecee4c40a460fb09c66d974229623/src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/NullableAttributes.cs
```
nvborisenko, line 24: Will be amazing to put a short reason why we stole it. The link you provided earlier is pretty good.

RenderMichael: Let me know if this message could be improved

dotnet/src/webdriver/ErrorResponse.cs
```diff
@@ -42,58 +39,45 @@ public ErrorResponse()
...
-                        this.message = "The error did not contain a message.";
-                    }
+                    this.Message = message;
                 }
-
-                if (responseValue.ContainsKey("screen") && responseValue["screen"] != null)
+                else
                 {
-                    this.screenshot = responseValue["screen"].ToString();
+                    this.Message = "The error did not contain a message.";
```
nvborisenko, line 54: No, it should be definitely server-side issue.

nvborisenko: If it `didn't contain a message`, then let's review the specification.

RenderMichael: Where can I find the spec?

This is the existing behavior, can we address it as a follow-up?

RenderMichael: It's worth noting that the current logic doesn't actually work for the stack trace property, because it is returned as a string instead of an `object[]`. It looks like there's something to fix here, with the spec I can give it a shot.

image

nvborisenko: I am just reading changes and see you introduced `The error did not contain a message.` magic message. Probably I missed something.

RenderMichael: The diffs are much easier to look at with whitespace disabled https://github.com/SeleniumHQ/selenium/pull/14672/files?w=1

dotnet/src/webdriver/IFileDetector.cs
```diff
@@ -16,6 +16,10 @@
 // limitations under the License.
 // </copyright>
 
+#nullable enable
```
nvborisenko: Let's put nullable directive after using statements.

dotnet/src/webdriver/Internal/NullableAttributes.cs
```diff
@@ -0,0 +1,166 @@
+// <copyright file="IWebElementReference.cs" company="WebDriver Committers">
```
nvborisenko: Filename is incorrect.

dotnet/src/webdriver/ErrorResponse.cs
nvborisenko: https://www.w3.org/TR/webdriver2/#errors - seems should be `stacktrace`, not `stackTrace`

nvborisenko: I don't mind we can handle `stacktrace` || `stackTrace`, but it should be `string`. Interesting, somebody did it especially, probably we can see it in git history why...

nvborisenko: In any case, in this PR let's keep things as simpler as possible, and in the following PR I would propose to align this with the specification. When `Screenshot` can be removed. And look at other bindings, how they handle `error` object.

dotnet/src/webdriver/StackTraceElement.cs
nvborisenko, line 1: Seems this class is obsolete. When we will align with spec (in future PR), seems it can be easily removed. But again, we should align with other bindings.

nvborisenko: Resolving this conversation, just let you know.

nvborisenko:
I am 99% confident with the proposed changes.

Now, the question is: what if we enable nullable context globally? Given that I can disable it for "DevTools/CDP" namespace, how many warnings will we get related to nullability?

dotnet/src/webdriver/Internal/NullableAttributes.cs
RenderMichael:
```suggestion
// <copyright file="NullableAttributes.cs" company="WebDriver Committers">
```

RenderMichael:
For that, we can add
```csharp
#nullable disable warnings
#nullable enable annotations
```
to the top of the generated CDP code. Ideally it would be nice to handle nullability on that side, as long as it’s simple enough.

RenderMichael:
Sorry, misread your comment about nullability aside from the devtools code.

After this PR, there’s still more code to enable nullability for. Some of it may even involve changing code to handle null values. After all the changes are done, we can set `<Nullable>enable<Nullable>` in the project file.

dotnet/src/webdriver/IFileDetector.cs
RenderMichael: Fixed in all files

nvborisenko:
Just want to understand how far we are from nullable context enabled globally. If we enable it right now, then do you think we can achieve the goal soon?.. within your help.

dotnet/src/webdriver/Firefox/FirefoxDriver.cs
nvborisenko: All right, this is server side. Given that old behavior is the still the same.

RenderMichael:
I’m not sure. It involves touching every file, but some files were trivial to enable. It all depends on the state of null-protection in the remaining code.

dotnet/src/webdriver/Firefox/FirefoxDriver.cs
```diff
@@ -288,6 +292,9 @@ public void SetContext(FirefoxCommandContext context)
         /// </summary>
         /// <param name="addOnDirectoryToInstall">Full path of the directory of the add-on to install.</param>
         /// <param name="temporary">Whether the add-on is temporary; required for unsigned add-ons.</param>
+        /// <returns>The add-on ID.</returns>
```
nvborisenko: `ID` could look better as `unique identifier` in user-faced docs.

nvborisenko:
OK, thank you, then let's continue enable it file by file. At the end of this story we will enable it globally, like big commit spread across all files. But please, if you feel that it would be better to enable it globally (to avoid one big boom commit), then let's do. BTW, how many warnings will we get?

dotnet/src/webdriver/Firefox/FirefoxDriver.cs
nvborisenko, line 1: We can discard all changes in this file and land the entire PR. Are you OK with it?

nvborisenko (approved):
Amazing start!

Issue #13334, opened 2023-12-19
[🚀 Feature]: deprecate WebElement.getAttribute


joerg1985:
### Feature and motivation

`WebElement.getDomAttribute` and `WebElement.getDomProperty` have been added in Selenium 4.

It might be good to deprecate `WebElement.getAttribute` to remove it at a undefined time in the future.

### Usage example

This would warn people to move to the 'new' methods when maintaining existing code.

titusfortner:
Bold.

I'm not completely against the idea, but also not sure it's the right call.

NCLnclNCL:
> ### Feature and motivation
> `WebElement.getDomAttribute` and `WebElement.getDomProperty` have been added in Selenium 4.
> 
> It might be good to deprecate `WebElement.getAttribute` to remove it at a undefined time in the future.
> 
> ### Usage example
> This would warn people to move to the 'new' methods when maintaining existing code.

Let me ask, are the 3 APIs above different?

titusfortner:
`getAttribute()` assumes that people don't know the difference between a property and an attribute and don't want to be bothered to figure it out (a decent assumption), so it uses a bunch of fancy JavaScript that does a pretty good job of guessing whether the user is looking for the property or the attribute and returning that value.  It's great convenience, but it is now implemented with a client side atom which means a large payload gets sent to the driver/server on every usage.

The ideal world people figure out if they want an attribute or property and use the correct method. But that might be asking a lot of users, and we already have the code right there...

joerg1985:
I was reading the blog post and the javadoc before opening the issue, this gave me the feeling the getAttribute method is the legacy way of reading attributes/properties.

If this is not true and there are no plans to ever remove the getAttribute method, this issue can be closed.

If this is a legacy method, it might be good to mark it deprecated or make a clear statement in the docs to ensure no new code is using this.
I agree there is alot of code using this, therefore a deprecation notice could be added, something like: 
`don't panic, this method will not be removed before selenium 6, but should not be used in new code.`

titusfortner:
It would definitely be valuable to encourage people to use the other methods.

titusfortner:
We decided in TLC Meeting 1/18 that we would go ahead and mark this feature deprecated to encourage people to move to the more precise property and attribute methods.

We need a blog post explaining everything though, and I'm still not sure we want to actually remove it any time soon because it is quite useful.

Delta456:
[...]

PR #14676, opened 2024-10-29
[dotnet] Added Deprecation to WebElement.GetAttribute() per #13334

shbenzer:
### Description
Added Deprecation to WebElement.GetAttribute() per #13334

### Motivation and Context
#13334 

### Types of changes

### Checklist

___

### **PR Type**
enhancement

___

### **Description**
- Added a deprecation warning to the `GetAttribute` method in the `IWebElement` interface.
- Suggested using the `GetDomAttribute` method as an alternative.

___

PRDescriptionHeader.CHANGES_WALKTHROUGH
<tr>
  <td>
    <details>
      <summary><strong>IWebElement.cs</strong><dd><code>Deprecate `GetAttribute` method in `IWebElement` interface</code></dd></summary>
<hr>

dotnet/src/webdriver/IWebElement.cs

<li>Added an <code>Obsolete</code> attribute to the <code>GetAttribute</code> method.<br> <li> Provided a deprecation warning message suggesting the use of <br><code>GetDomAttribute</code>.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14676/files#diff-09553d2da00fdead17001e2a7558cba418b6e926879880eedf0ed487164055cf">+1/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

___


nvborisenko:
```csharp
driver.Url = "https://nuget.org";

var input = driver.FindElement(By.Id("search"));
input.SendKeys("abc");

var val = input.GetAttribute("value");
var val2 = input.GetDomAttribute("value");
var val3 = input.GetDomProperty("value");

Console.WriteLine(val); // "abc"
Console.WriteLine(val2); // ""
Console.WriteLine(val3); // abc
```

shbenzer:
> ```cs
> driver.Url = "https://nuget.org";
> 
> var input = driver.FindElement(By.Id("search"));
> input.SendKeys("abc");
> 
> var val = input.GetAttribute("value");
> var val2 = input.GetDomAttribute("value");
> var val3 = input.GetDomProperty("value");
> 
> Console.WriteLine(val); // "abc"
> Console.WriteLine(val2); // ""
> Console.WriteLine(val3); // abc
> ```

good catch, I forgot to add GetDomProperty() to the deprecation note

nvborisenko:
We have to double check which one is correct: `GetAttribute` or `GetDomAttribute`. I though the correct one is `GetAttribute`.

shbenzer:
> We have to double check which one is correct: `GetAttribute` or `GetDomAttribute`. I though the correct one is `GetAttribute`.

According to @titusfortner in #13334 

>. getAttribute() assumes that people don't know the difference between a property and an attribute and don't want to be bothered to figure it out (a decent assumption), so it uses a bunch of fancy JavaScript that does a pretty good job of guessing whether the user is looking for the property or the attribute and returning that value. It's great convenience, but it is now implemented with a client side atom which means a large payload gets sent to the driver/server on every usage.

> We decided in TLC Meeting 1/18 that we would go ahead and mark this feature deprecated to encourage people to move to the more precise property and attribute methods.

nvborisenko:
All right, `GetAttribute()` is deprecated.

Since it is breaking change of core (most used interface) functionality, let's add info when it will be completely removed. In v5 or in v6? It should be definitely major selenium version, but which one... Usually we remove deprecated methods in 2 minor releases, but this another case, this would be real pain for all users. So major release should be considered. But if v5 will be released in 2 months, then it is not enough time range - then v6 is preferable.

shbenzer:
I agree that major releases should be considered. Since v5 is too soon, I'll put v6 in. 

Worst case we make a change later and make it v7.

dotnet/src/webdriver/IWebElement.cs
```diff
@@ -170,7 +170,7 @@ public interface IWebElement : ISearchContext
         /// </list>
         /// </remarks>
         /// <exception cref="StaleElementReferenceException">Thrown when the target element is no longer valid in the document DOM.</exception>
-        [Obsolete("Use GetDomAttribute(string attributeName) or GetDomProperty(string propertyName")]
+        [Obsolete("Use GetDomAttribute(string attributeName) or GetDomProperty(string propertyName). GetAttribute() is schedule for removal in Selenium v6"]
```
nvborisenko: `GetAttribute(string attributeName) will be removed in Selenium 6.` - full signature and typo.

shbenzer: Done

nvborisenko (approved):

nvborisenko:
@shbenzer please do not merge! Kind of code-freeze after release. Keep it open for a while.

nvborisenko:
@shbenzer please check failing CI build, easy to fix.

shbenzer:
seemed to be missing the system import as well

PR #14726, opened 2024-11-07
[dotnet] Allow `RemoteSessionSettings` to use any value for metadata

### LINKED ISSUE #14725: [🐛 Bug]: [dotnet]: `RemoteSessionSettings` should allow `JsonNode` metadata settings
### What happened?

Selenium should accept System.Text.Json types, since that's what is used internally to serialize the request.

(It would also be nice to add explicit overloads to `RemoteSessionSettings.AddMetadataSetting` so it's clear what types are accepted and what types aren't.)

```shell
using OpenQA.Selenium;
using System.Text.Json.Nodes;

var remoteSessionSettings = new RemoteSessionSettings();
remoteSessionSettings.AddMetadataSetting("example", JsonNode.Parse("true"));
```

### Relevant log output

```shell
System.ArgumentException: 'Metadata setting value must be JSON serializable. (Parameter 'settingValue')'
```

### Operating System

Windows 11

### Selenium version

.NET `Selenium.WebDriver` 4.26.1

N/A (didn't get that far)

N/A

No

RenderMichael:
### **User description**
Fixes #14725

### Description

The `RemoteSessionSettings.AddMetadataSetting` method only allows users to pass in "primitive" types.

This is not necessary, as any type that can be serialized to JSON is valid. If a type is somehow not JSON-serializable, then the proper exception will be thrown by the serializer.. It also restricts the use of JSON primitives such as `JsonNode` and `JsonElement` from being used.

### Motivation and Context

```csharp
using OpenQA.Selenium;
using System.Text.Json.Nodes;

var remoteSessionSettings = new RemoteSessionSettings();

var trueValue =JsonNode.Parse("true");

remoteSessionSettings.AddMetadataSetting("example", trueValue);
```

### Types of changes

### Checklist

___

### **PR Type**
Enhancement, Tests

___

### **Description**
- Enhanced `RemoteSessionSettings` to support `System.Text.Json.Nodes` and `System.Text.Json.Elements`.
- Replaced the `mustMatchDriverOptions` field with a property `MustMatchDriverOptions` for better encapsulation.
- Improved the `IsJsonSerializable` method to handle `JsonNode` and `JsonElement` types, and added null checks.
- Added comprehensive tests to verify the handling of metadata settings in `RemoteSessionSettings`, ensuring compatibility with various JSON serializable types.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>RemoteSessionSettings.cs</strong><dd><code>Enhance RemoteSessionSettings to support System.Text.Json types</code></dd></summary>
<hr>

dotnet/src/webdriver/Remote/RemoteSessionSettings.cs

<li>Added support for <code>System.Text.Json.Nodes</code> in <code>RemoteSessionSettings</code>.<br> <li> Replaced <code>mustMatchDriverOptions</code> with a property <br><code>MustMatchDriverOptions</code>.<br> <li> Enhanced <code>IsJsonSerializable</code> method to handle <code>JsonNode</code> and <code>JsonElement</code>.<br> <li> Improved null handling in <code>IsJsonSerializable</code>.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14726/files#diff-288dfba3f612ed5c9b4def9fa29733bca2af53aca8e3d1a265b1303235201c26">+35/-34</a>&nbsp; </td>

</tr>                    
<tr>
  <td>
    <details>
      <summary><strong>RemoteSessionCreationTests.cs</strong><dd><code>Add tests for RemoteSessionSettings JSON compatibility</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/test/remote/RemoteSessionCreationTests.cs

<li>Added tests for <code>RemoteSessionSettings</code> metadata settings.<br> <li> Verified JSON serializability of various data types.<br> <li> Ensured compatibility with <code>JsonNode</code> and <code>JsonElement</code>.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14726/files#diff-d28e50e4cbe121457f6cc57b297c09b8f95c2c9e0d8781900566c94bbc9bfd8d">+64/-0</a>&nbsp; &nbsp; </td>

</tr>                    

___


dotnet/src/webdriver/Remote/RemoteSessionSettings.cs
```diff
@@ -306,34 +301,42 @@ private List<object> GetFirstMatchOptionsAsSerializableList()
             return optionsMatches;
         }
 
-        private bool IsJsonSerializable(object arg)
+        private static bool IsJsonSerializable(object arg)
```
nvborisenko: What if we remove this method at all? Because this class doesn't know about serialization context? It cannot determine whether something can be serialized..

RenderMichael: This logic was introduced when the `RemoveSessionSettings` was created 6 years ago 0b310c111486897cc7c711855ee9dc9005fc2dc3

The original explanation was `The value to be set must be serializable to JSON for transmission across the wire to the remote end.`. However, models are perfectly JSON serializable. I don't see any potential problem with this.

On the other hand, it makes the serialization very confined, since we know all possible types... Especially since the `ToString()` implementation is JSON serializing the instance (not sure if anyone relies on this ToString behavior, hopefully not).

I think if we change the `ToString()` behavior to behave like a normal, cheap `ToString()` instead of really serializing the type, then it's safe to remove the restriction.

nvborisenko: Yes, let's do it together. While we are here in context of json serialization, then let's "break" it. We already broke it, you know.

If we are going to normalize a usage of `ToString()` methods, then we should have a suggestion for users who rely on this changed behavior.

RenderMichael: PR has been updated to allow any type to be JSON-serialized, not just the "primitives", like a reverse `ResponseValueJsonConverter`.

nvborisenko: Road to simplicity, thank you!

nvborisenko:
Can you please adjust PR description according to the latest changes? Thx.

RenderMichael:
PR description updated,

dotnet/src/webdriver/Remote/RemoteSessionSettings.cs
```diff
@@ -35,7 +36,6 @@ public class RemoteSessionSettings : ICapabilities
         private const string AlwaysMatchCapabilityName = "alwaysMatch";
 
         private readonly List<string> reservedSettingNames = new List<string>() { FirstMatchCapabilityName, AlwaysMatchCapabilityName };
-        private DriverOptions mustMatchDriverOptions;
```
nvborisenko: Keeping it as a field is better rather than a property. `internal property` is still good, but it is an indicator for us like "please understand me and refactor".

RenderMichael: I can do this.

I am always nervous about having properties which are just getters/setters for fields, because it is a minefield for typos. The infinite-recursion bug from the other day, in `ActionSequence.InputDevice`, was an example of this. If an auto-property is possible with the exact same semantics (this one has `private set` so it is still safe), then I like to do those.

nvborisenko: Please do at least in scope of this PR. Then we will understand why `internal property` is introduced, probably it is used somewhere... I don't know, but it is another story.

RenderMichael: Done.

The property is not used anywhere, maybe it was in the past.

nvborisenko (approved):
Thank you!

PR #14737, opened 2024-11-10
[dotnet] Fix `RelativeBy.Near` and empty list return, port Java tests

### LINKED ISSUE #13610: [🐛 Bug]: "Near" relative locator doesn't work for the dotNet
### What happened?

I came across the fact that the Near relative locator for dotNet does not work.
I tried to debag, and found that I did not have the element. I get null at this point - js.ExecuteScript(wrappedAtom, parameters);
https://github.com/SeleniumHQ/selenium/blob/trunk/dotnet/src/webdriver/RelativeBy.cs#L109

Message: 
System.NullReferenceException : Object reference not set to an instance of an object.

There are no tests for near relative locators in the dotNet project - https://github.com/SeleniumHQ/selenium/blob/b6b209602d544b769a0f4635022cf55a7b8f81b4/dotnet/test/common/RelativeLocatorTest.cs

At the same time, test coverage looks better for Java -  https://github.com/SeleniumHQ/selenium/blob/trunk/java/test/org/openqa/selenium/support/locators/RelativeLocatorTest.java

```shell
I tried using the web page https://github.com/SeleniumHQ/selenium/blob/trunk/common/src/web/relative_locators.html
 to test relative locators and add a test for the Near locator (By analogy with Java tests): 

[Test]
public void ShouldBeAbleToFindElementsNear()
{
 var elements = seleniumDriver.FindElements(RelativeBy
                            .WithLocator(By.TagName("td"))
                            .Near(By.Id("center")));

 Assert.That(elements, Is.EquivalentTo(new List<string> { "second", "eighth", "fourth", "sixth", "first", "third", "seventh", "ninth" }));"seventh", "ninth");
}

the elements variable is null.

```

### Relevant log output

```shell
Message: 
System.ArgumentException : Expected: IEnumerable But was: null (Parameter 'actual')
```

### Operating System

Windows 10

### Selenium version

.Net 

Chrome

chromedriverVersion, 121.0.6167.184 

No

RenderMichael:
### **User description**

### Description

Bug 1: 
The `Near` method is now fixed, with a "blind" copy of what the Java code is doing.

Bug 2:
When querying for elements using the relative locator, sometimes you are returned nothing. In that case, the response JSON is `[]`, and the de-serializer doesn't know you're getting back web elements.

As such, the response object has a runtime type of `ReadOnlyCollection<object>` and the cast `as ReadOnlyCollection<IWebElement>` fails (ie, returns null).

I can't think of a more elegant solution that doesn't require gutting the JavaScript execution process.

One more reason the `as` operator is not the right tool when the type is expected to be an exact match 🫠

### Motivation and Context

Fixes #13610 

### Types of changes

### Checklist

___

### **PR Type**
Bug fix, Tests

___

### **Description**
- Fixed the issue where relative locators returned null instead of an empty collection when no matches are found.
- Added logic to handle the deserialization quirk by returning an empty collection.
- Introduced a new test to verify that an empty list is returned when no elements are found using relative locators.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>RelativeBy.cs</strong><dd><code>Handle empty results in relative locators</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/RelativeBy.cs

<li>Added handling for empty results when using relative locators.<br> <li> Implemented conditional logic to return an empty collection when no <br>elements are found.<br> <li> Included debug assertion for unreachable code paths.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14737/files#diff-3b7446f646eaa1973fcc59f1c877e3c1456444d4cd4c4a6deaadc46d6d2c8f59">+27/-2</a>&nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>RelativeLocatorTest.cs</strong><dd><code>Add test for empty result in relative locators</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/test/common/RelativeLocatorTest.cs

<li>Added a test to verify behavior when no elements are found using <br>relative locators.<br> <li> Ensured the test checks for an empty list return.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14737/files#diff-01b475a5c22cd87cef8166aeed5106ad36264ff6964e888dff4d4228ef0f9c48">+11/-0</a>&nbsp; &nbsp; </td>

</tr>

___


dotnet/src/webdriver/RelativeBy.cs
```diff
@@ -288,7 +308,7 @@ private RelativeBy Near(object locator, int atMostDistanceInPixels)
 
             Dictionary<string, object> filter = new Dictionary<string, object>();
             filter["kind"] = "near";
-            filter["args"] = new List<object>() { GetSerializableObject(locator), "distance", atMostDistanceInPixels };
+            filter["args"] = new List<object>() { GetSerializableObject(locator), atMostDistanceInPixels };
```
RenderMichael, line 311: I'm just copying what I see in the Java version, and it's beginning to work

https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/support/locators/RelativeLocator.java#L185

RenderMichael:
Tests are a one-to-one clone of the java tests: https://github.com/SeleniumHQ/selenium/blob/trunk/java/test/org/openqa/selenium/support/locators/RelativeLocatorTest.java

With the exception of the one C#-deserializer-specific test at the bottom

dotnet/test/common/RelativeLocatorTest.cs
```diff
@@ -53,14 +75,166 @@ public void ShouldBeAbleToCombineFilters()
...
+            Assert.That(values, Is.EquivalentTo(new List<string> { "second", "eighth", "fourth", "sixth", "first", "third", "seventh", "ninth" }));
+        }
+
+        [Test]
+        public void EnsureNoRepeatedElements()
+        {
+            driver.Url = EnvironmentManager.Instance.UrlBuilder.CreateInlinePage(new InlinePage()
+               .WithTitle("Repeated Elements")
+               .WithStyles(
+                " .c {\n"
```
nvborisenko: It is better to use row string literals: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/raw-string

dotnet/test/common/RelativeLocatorTest.cs
```diff
@@ -53,14 +75,166 @@ public void ShouldBeAbleToCombineFilters()
...
-                elementIds.Add(id);
-            }
+                var rect2 = driver.FindElement(RelativeBy.WithLocator(By.Id("rect4")).Near(rect3));
 
-            Assert.That(elementIds, Is.EquivalentTo(new List<string>() { "third" }));
+            }, Throws.TypeOf<NoSuchElementException>().With.Message.EqualTo("Unable to find element; For documentation on this error, please visit: https://www.selenium.dev/documentation/webdriver/troubleshooting/errors#no-such-element-exception"));
+        }
+
+        //------------------------------------------------------------------
+        // Tests below here are not included in the Java test suite
```
nvborisenko, line 231: Do not mention Java in source code, we rely on it to increase coverage, but it is not one-to-one relationship. Thx.

RenderMichael: I was copying the existing style, should all those be removed?

image

dotnet/src/webdriver/RelativeBy.cs
```diff
@@ -108,8 +108,28 @@ public override ReadOnlyCollection<IWebElement> FindElements(ISearchContext cont
             filterParameters["filters"] = this.filters;
             parameters["relative"] = filterParameters;
             object rawElements = js.ExecuteScript(wrappedAtom, parameters);
-            ReadOnlyCollection<IWebElement> elements = rawElements as ReadOnlyCollection<IWebElement>;
-            return elements;
+
+            if (rawElements is ReadOnlyCollection<IWebElement> elements)
```
nvborisenko, line 112: Sorry, I didn't dive deeper here, seems it would be better to rely on some lower interface like `IReadOnlyCollection<...>`? Leaving this comment just to pay an attention what exactly `.ExecuteScript()` returns in terms of types - json serialization topic again.

RenderMichael: We would have to use a non-generic interface like `ICollection` or `IEnumerable` to capture the `ReadOnlyCollection<object>`. At that point we would need to do something like `.OfType<object>().Cast<IWebElement>()` or something like that.

This is a band-aid for the real need of a generic `.ExecuteScript<TReturn>()` method.

RenderMichael: If you look at the `WebDriver.ParseJavaScriptReturnValue` method, you will see that because the de-serializer tries to take a guess about the return type. This is because it has no type information, which would require a generic method.

dotnet/test/common/RelativeLocatorTest.cs
nvborisenko: LOL, no comments :D I guess there is no search entry for Ruby or Python. Java is the best :)

Seriously, keep it as it, separate topic.

dotnet/src/webdriver/RelativeBy.cs
nvborisenko: The generic method is obvious: it is only one way how an user is able to specify the expected/required type (json serializable?). For reviewers: we should understand why `rawElements is ReadOnlyCollection<IWebElement>` and no more..

RenderMichael: Whenever there is more than one element in the response, the de-serializer will recognize that we are returning elements and de-serialize to a `ReadOnlyCollection<IWebElement>`.

However, there is a special case when no elements are returned: the response is `[]` and the de-serializer will not know we are expecting elements. That is why it returns a `ReadOnlyCollection<object>`, and that is the scenario we are handling here.

nvborisenko: Now I see, that is generic problem that all commands in Selenium are not strongly-typed. Thanks for diving into this particular issue.

nvborisenko (approved):

dotnet/src/webdriver/RelativeBy.cs
nvborisenko: All right.

PR #14742, opened 2024-11-11
[dotnet] Make `FirefoxProfile` AOT-safe

RenderMichael:
### **User description**

### Description

This PR hand-rolls parsing of the `webdriver_prefs.json` file, instead of parsing into a `Dictionary<string, object>`

This PR is a lot easier to read with [whitespace disabled](https://github.com/SeleniumHQ/selenium/pull/14742/files?w=1).

### Motivation and Context

An open-typed `object` dictionary is inherently AOT-unsafe. In this scenario, we know the expected structure of the JSON, so we can manually parse the values.

### Types of changes

### Checklist

___

### **PR Type**
enhancement

___

### **Description**
- Refactored JSON parsing in `FirefoxProfile` to use `JsonObject` instead of `Dictionary<string, object>`, enhancing AOT safety.
- Updated `Preferences` class to handle preferences using `JsonNode`, improving type safety and performance.
- Added nullability annotations to improve code quality and prevent null reference exceptions.
- Enhanced exception handling for preference setting and JSON parsing.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>FirefoxProfile.cs</strong><dd><code>Refactor JSON Parsing to Use JsonObject for AOT Safety</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/Firefox/FirefoxProfile.cs

<li>Replaced <code>Dictionary<string, object></code> with <code>JsonObject</code> for JSON parsing.<br> <li> Removed custom JSON serialization options.<br> <li> Improved exception handling for JSON parsing.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14742/files#diff-1dd17ad451c78ebceb2cd417a5c8e270092e67f75798ab31b40f01265bdcd8b0">+6/-12</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>Preferences.cs</strong><dd><code>Update Preferences Handling with JsonNode and Nullability</code></dd></summary>
<hr>

dotnet/src/webdriver/Firefox/Preferences.cs

<li>Changed preference storage to use <code>JsonNode</code>.<br> <li> Updated <code>SetPreferenceValue</code> to handle <code>JsonNode</code>.<br> <li> Added nullability annotations.<br> <li> Improved exception handling for preference setting.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14742/files#diff-61c0a793a182862cb1a99b975b2440afaaaa606ddddeab2acc9b943e0c2a3125">+40/-26</a>&nbsp; </td>

</tr>

___


dotnet/src/webdriver/Firefox/Preferences.cs
```diff
@@ -64,6 +67,7 @@ public Preferences(Dictionary<string, object> defaultImmutablePreferences, Dicti
         /// <param name="value">A <see cref="string"/> value give the preference.</param>
         /// <remarks>If the preference already exists in the currently-set list of preferences,
         /// the value will be updated.</remarks>
+        /// <exception cref="ArgumentNullException">If <paramref name="key"/> is null.</exception>
```
nvborisenko: Are you writing it manually? When I see `null` here I expect it can be `<see langword="null" />`

RenderMichael: Yes I'm doing it manually, I don't know if Rider has special tooling but Visual Studio doesn't do much for this. Using the tag is better, let me make that change

dotnet/src/webdriver/Firefox/Preferences.cs
```diff
@@ -153,39 +160,46 @@ private static bool IsWrappedAsString(string value)
...
+                    return;
+                }
+                if (jValue.TryGetValue(out long longValue))
+                {
+                    this.preferences[key] = longValue.ToString(CultureInfo.InvariantCulture);
+                    return;
+                }
             }
 
             throw new WebDriverException("Value must be string, int or boolean");
```
nvborisenko: I don't believe that `double` is denied. While you are here, do you think whether it is possible to avoid any type restrictions?

RenderMichael: I was trying to do research on the `webdriver_prefs.json` file format expectations, and I didn't see any decimal values anywhere. The current behavior will not only throw on doubles, but also `long`s. This PR already fixes `long` behavior (it will need to be `long.MaxValue + 1` before we throw), I can do the same for `double`'es.

The last primitive is `null`. That is still not allowed, right?

RenderMichael: Support for `double` has been added

nvborisenko: We should understand the purpose of this class. The good news is it is `internal` (just left a note for reviewers).

RenderMichael: This functionality is accessible to users in the `FirefoxProfileManager.GetProfile` method. There are currently-disabled tests in `FirefoxProfileManagerTest`. I did not make any test changes, but those tests work with the new logic.

dotnet/src/webdriver/Firefox/Preferences.cs
```diff
@@ -153,39 +160,51 @@ private static bool IsWrappedAsString(string value)
...
+                    return;
+                }
+                if (jValue.TryGetValue(out double doubleValue))
+                {
+                    this.preferences[key] = doubleValue.ToString(CultureInfo.InvariantCulture);
+                    return;
+                }
             }
 
             throw new WebDriverException("Value must be string, int or boolean");
```
RenderMichael:
```suggestion
            throw new WebDriverException("Value must be string, number or boolean");
```

RenderMichael:
Support for floating-point preferences has been added and the error message has been changed to reflect that.

This PR is ready for another look!

dotnet/src/webdriver/Firefox/Preferences.cs
nvborisenko: It can be type agnostic. Here we have a validation that incoming value should be `int or string or something else well known`. Since incoming is `JsonNode`, then we can completely remove this restriction and just accept `JsonNode` as is. This method is called by another internal methods (strongly-typed! int, string, ...). I guess if remove this restriction, then we simplify understanding of why it cannot be `double`. Probably it can, but it is another story.

RenderMichael: I imagine this restriction is based on the possible values for Firefox profile preferences. Maybe Firefox will throw an exception if we try to set a wrong type of value?

We can accept any value here and let Firefox throw any errors.

Should we accept any JSON values like objects or arrays, or should we allow only `true`, `false`, numbers, and strings? What about `null`?

nvborisenko: See, https://github.com/SeleniumHQ/selenium/blob/03386772307dde5dd063ef1bde35d8013132df94/dotnet/src/webdriver/Firefox/FirefoxProfile.cs#L137 (the same for int and string) - this entry point for public users. We are still manage "allowed" types.

RenderMichael: The entry point for the JSON parsing is the `webdriver_pref.json` file, which users can set values on. If their file has something like this:

```json
{
  "frozen": {
    "my-pref": [1, 2, 3]
  }
}
```

We will have to handle that somehow. Today, it throws the webdriver error "must be string, int, or boolean".

nvborisenko: Right, `webdriver_pref.json` is our file, fully manageable. We can parse it as is without any restrictions.

RenderMichael: After doing some research, I found out this:

> A value may either be integer (0, 2 etc.), boolean (true or false) or string (text, such as a file path)

https://kb.mozillazine.org/User.js_file#Adding_user.js_entries

For this reason, we can assume that the `webdriver_pref.json` will only ever be those values.

nvborisenko: OK! Final concern: we don't know validation rules on client side, and it is better to delegate validation to server side. I just want to say "deserialize as is because of it is json". Rules... should we replicate "unknown" rules on client side? I don't think so.

nvborisenko: Maintenance... it is effort.

dotnet/src/webdriver/Firefox/FirefoxProfile.cs
```diff
@@ -296,22 +296,16 @@ private void UpdateUserPreferences()
...
-
             using (Stream defaultPrefsStream = ResourceUtilities.GetResourceStream("webdriver_prefs.json", "webdriver_prefs.json"))
             {
                 using (StreamReader reader = new StreamReader(defaultPrefsStream))
                 {
                     string defaultPreferences = reader.ReadToEnd();
-                    Dictionary<string, object> deserializedPreferences = JsonSerializer.Deserialize<Dictionary<string, object>>(defaultPreferences, jsonSerializerOptions);
-                    Dictionary<string, object> immutableDefaultPreferences = deserializedPreferences["frozen"] as Dictionary<string, object>;
-                    Dictionary<string, object> editableDefaultPreferences = deserializedPreferences["mutable"] as Dictionary<string, object>;
+                    JsonObject deserializedPreferences = JsonNode.Parse(defaultPreferences) as JsonObject
```
nvborisenko: Probably `JsonDocument` would be better.

RenderMichael: Fixed. It came out much simpler with the `JsonDocument`.

dotnet/src/webdriver/Firefox/Preferences.cs
RenderMichael: I agree. I always like "defense in depth", if someone has a bug (firefox, Selenium's server, misconfiguration of whoever wrote the `webdriver_pref.json` file...) I want this system to tell users clearly what went wrong.

But that cannot come at the expense of maintenance. I think the current state of this PR is simple enough, especially with the migration to `JsonDocument` in the latest commit.

dotnet/src/webdriver/Firefox/Preferences.cs
```diff
@@ -64,9 +63,31 @@ public Preferences(Dictionary<string, object> defaultImmutablePreferences, Dicti
...
+                throw new ArgumentNullException(nameof(value));
+            }
+
+            if (IsWrappedAsString(value))
+            {
+                throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, "Preference values must be plain strings: {0}: {1}", key, value));
+            }
+
+            this.ThrowIfPreferenceIsImmutable(key, value);
+            this.preferences[key] = string.Format(CultureInfo.InvariantCulture, "\"{0}\"", value);
```
RenderMichael, line 90: The existing logic wraps strings into json strings, so `abc` becomes `"abc"`. It achieves this by manually adding `"` to the front and end of the string.

This can be achieved simpler by doing this:
```suggestion
            this.preferences[key] = JsonValue.Create(value).ToJsonString();
```
 Not only will this wrap the string in double-quotes, it will also escape special characters using the format JSON rules. [Example](https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK7wCYgNQB8CwAUAAIBMAjEcQAwAEx5AdACoCmAHgowFIDOA9jEYA5fhla8A3ESrkAnAAo+ggGoQANqlaMAwslYQErBQCJAPBuAQfZMBKFv2UwAygmSwA5guvXJQA=)

Should I make this change? This method is publicly reachable by users.

RenderMichael: Looking through the available types, I found a method which does the JSON string encoding, but does not add the double-quotes. The following would make things type-safe, otherwise users can mess up the JSON.

```suggestion
            this.preferences[key] = string.Format(CultureInfo.InvariantCulture, "\"{0}\"", JsonEncodedText.Encode(value));
```

nvborisenko: Investigated. We have `webdriver.json` file, bundled as assembly resource. This file contains some default values (key/value pairs, grouped to mutable/frozen). Then at runtime we deserialize it, and provide API to modify key/values in memory. And finally all key/values are going into `*.js` file:
```js
user_pref("key1", value1);
user_pref("key2", value2);
```

where `value` is escaped like `preference.Value.Replace(@"\", @"\\")`.

So, don't apply escaping rules for **json**, I believe it is not aligned with js source code rules. Safer to not touch it.

nvborisenko:
Thank you, @RenderMichael !

PR #14755, opened 2024-11-13
[dotnet] Improve the debugging experience of tests

RenderMichael:
### **User description**

### Description

Use the new `ConfigureAwait(ConfigureAwaitOptions)` method introduced in .NET 8 to avoid first-chance exceptions in the test cleanup

### Motivation and Context

When debugging through code, .NET provides first-chance exceptions, allowing users to see exceptions as they are thrown, even if they are handled or suppressed.

Unrelated exceptions can get in the way when testing unrelated code. This change makes the debugger break less when it is not necessary.

image

### Types of changes

### Checklist

___

### **PR Type**
enhancement

___

### **Description**
- Introduced a new `QuitNoThrow` method to handle HTTP requests in a way that suppresses exceptions during test cleanup.
- Replaced existing inline HTTP client logic with the new `QuitNoThrow` method in the `Stop` method of `TestWebServer`.
- Utilized `ConfigureAwait(ConfigureAwaitOptions.SuppressThrowing)` to prevent first-chance exceptions from interrupting the debugging process.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>TestWebServer.cs</strong><dd><code>Improve test cleanup by suppressing exceptions using ConfigureAwait</code></dd></summary>
<hr>

dotnet/test/common/Environment/TestWebServer.cs

<li>Added <code>QuitNoThrow</code> method to handle HTTP requests without throwing <br>exceptions.<br> <li> Replaced inline HTTP client logic with <code>QuitNoThrow</code> in the <code>Stop</code> method.<br> <li> Utilized <code>ConfigureAwait(ConfigureAwaitOptions.SuppressThrowing)</code> to <br>suppress exceptions.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14755/files#diff-9b1522ed8b7052b3c0e172fa95173cec4340245fe34caf1ca94cf17376a10dca">+15/-14</a>&nbsp; </td>

</tr>

___


dotnet/test/common/Environment/TestWebServer.cs
nvborisenko, line 1: So smart for test project. I vote for keeping simplicity, rather than being technically perfect.

RenderMichael: I realized we don't actually need to dispose the HTTP response, since we are disposing the client. I made it a little simpler, do you think this is acceptable?

nvborisenko: We have a lot of try/catch in codebase, how this one affects your debugging experience?

RenderMichael: I am not sure if I have something mis-configured, but the webserver `Stop` method always throws an exception for me, as shown in the PR description. This is something which causes a debugger break every time I am running tests.

It would be nice to avoid the debugger breaking on other exceptions which are well-handled, such as `Runfiles.Create()`. However, those are in synchronous methods, and there is no nice way to get that to happen.

Luckily, the quit method is an asynchronous operation and .NET 8 has an `await` modifier which avoids throwing the exception.

RenderMichael: In my opinion, this is a simplification. It requires extracting a method only because `Stop` is currently synchronous. Would it be better to make `Stop` async, and propagate that out to the callers? It would not be more complex, and it would avoid doing sync-over-async where it is not necessary

nvborisenko: It throws on my end, but my debugger doesn't hit it. If making `Stop()` method asynchronous will help you, then it is better. Thanks.

nvborisenko:
Sorry, I cannot accept it. We have a lot of `try/catch` pattern in the codebase. Please understand your IDE why the debugger hits this exception, my debugger doesn't hit. Probably you have an active breakpoint in this class.

Again sorry, and thanks for understanding.

PR #14763, opened 2024-11-15
[rb] Deprecate CDP methods on Firefox

aguspe:
### Description
This PR adds deprecation warnings for all the CDP methods for Firefox

Example of deprecation warnings:

Screenshot 2024-11-21 at 15 20 12

### Motivation and Context
Based on https://fxdx.dev/deprecating-cdp-support-in-firefox-embracing-the-future-with-webdriver-bidi/

We want to deprecate CDP methods on firefox and encourage Selenium users to use the new BiDi implementation

### Types of changes

### Checklist


p0deje (commented):
We should provide a replacement implementation and mention it in deprecation. 

For example,`on_log_event` should be deprecated in favor of `script.add_console_message_handler` / `script.add_javascript_error_handler`.

If we don't have exact replacements yet, I suggest we hold off deprecating the methods (suggestion per https://seleniumhq.slack.com/archives/CBH302726/p1731796745420569?thread_ts=1731576590.042029&cid=CBH302726).

Issue #14789, opened 2024-11-22
[🐛 Bug]: Selenium 4.26.0 has memory leak.


shinobi01:
### What happened?

I'm using Java + Cucumber + testNG + Appium to do automation test on mobile devices. I got some errors as following
**Caused by: java.lang.OutOfMemoryError: Compressed class space**
So I checked with VisualVM, the Metaspace and Classes are always increasing.
<img width="1071" alt="error" src="https://github.com/user-attachments/assets/173c69ce-7a9c-422f-8859-4a052aa01314">
The loaded classs are not unloaded.
After downgrade to 4.23.0, this issue is gone
<img width="969" alt="Screenshot 2024-11-22 at 13 43 05" src="https://github.com/user-attachments/assets/3caafdb9-4387-47f2-957a-948a225dde24">

### How can we reproduce the issue?

```shell
Sorry I can't share my code, but create a cucumber test cases and rerun it several times should be able to reproduce.
```

### Relevant log output

```shell
Please refer to dump files (if any exist) [date].dump, [date]-jvmRun[N].dump and [date].dumpstream.
Exception in provider
org.apache.maven.surefire.booter.SurefireExecutionException: Exception in provider
	at org.apache.maven.plugin.surefire.InPluginVMSurefireStarter.runSuitesInProcess(InPluginVMSurefireStarter.java:91)
	at org.apache.maven.plugin.surefire.AbstractSurefireMojo.executeProvider(AbstractSurefireMojo.java:1144)
	at org.apache.maven.plugin.surefire.AbstractSurefireMojo.executeAfterPreconditionsChecked(AbstractSurefireMojo.java:1002)
	at org.apache.maven.plugin.surefire.AbstractSurefireMojo.execute(AbstractSurefireMojo.java:848)
	at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:126)
	at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2(MojoExecutor.java:328)
	at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute(MojoExecutor.java:316)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:212)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:174)
	at org.apache.maven.lifecycle.internal.MojoExecutor.access$000(MojoExecutor.java:75)
[...]

PR #14793, opened 2024-11-24
[java] reuse the classes created by the WebDriverDecorator #14789

joerg1985:
### Description
This PR will add a cache to the `WebDriverDecorator` and reuse the classes created.
To enabled reusing the handler, the target object is injected to a new field when a proxy is created.
As soon as the handler must call a method, the target object is red from the field inside the proxy.

### Motivation and Context
Each invocation of a decorated method did return a new class, this will end in a OutOfMemoryError as soon as a certain number of classes has been generated. See #14789 

### Types of changes

### Checklist

___

### **PR Type**
Bug fix, Enhancement, Tests

___

### **Description**
- Implemented a caching mechanism in `WebDriverDecorator` to reuse proxy classes, preventing excessive class creation and potential `OutOfMemoryError`.
- Introduced `Definition` and `ProxyFactory` classes to manage proxy creation and caching.
- Added `HasTarget` interface to facilitate access to target objects within proxies.
- Enhanced method implementations to utilize cached proxies, improving performance and memory usage.
- Added tests to verify that the same proxy class is reused and that the correct instance is accessed during method calls.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>WebDriverDecorator.java</strong><dd><code>Implement caching and reuse of proxy classes in WebDriverDecorator</code></dd></summary>
<hr>

java/src/org/openqa/selenium/support/decorators/WebDriverDecorator.java

<li>Introduced caching mechanism to reuse proxy classes.<br> <li> Added <code>Definition</code> and <code>ProxyFactory</code> classes for managing proxies.<br> <li> Implemented <code>HasTarget</code> interface for accessing target objects.<br> <li> Modified method implementations to utilize cached proxies.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14793/files#diff-c1eeb9788c8a7d5f29ee083e2e9fb8fd77b9fb22b0d1b7efbbdea54abeb9a650">+126/-44</a></td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>DecoratedWebDriverTest.java</strong><dd><code>Add tests for proxy class reuse and instance correctness</code>&nbsp; </dd></summary>
<hr>

java/test/org/openqa/selenium/support/decorators/DecoratedWebDriverTest.java

<li>Added test to ensure no excessive class creation.<br> <li> Added test to verify correct instance usage.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14793/files#diff-190c26600303f4ae1dac64bae6490c646c1098c6c29b1e108721738d62e56bd9">+40/-0</a>&nbsp; &nbsp; </td>

</tr>

___


pujagani:
@mykola-mokhnach Can you please help review this PR?

java/src/org/openqa/selenium/support/decorators/WebDriverDecorator.java
```diff
@@ -183,6 +190,59 @@
...
+    public int hashCode() {
+      return Objects.hash(decoratedClass, originalClass);
+    }
+  }
+
+  protected static class ProxyFactory<T> {
+    private final Class<? extends T> clazz;
+    private final Field targetLookup;
+
+    private ProxyFactory(Class<? extends T> clazz, Field targetLookup) {
```
mykola-mokhnach: I would rather limit the clazz to adhere to an interface containing setter and getter for the decorated target. 
See https://github.com/appium/java-client/blob/42480feca9845bbbb569b13d995b80980bf4ca33/src/main/java/io/appium/java_client/proxy/Helpers.java#L135 for an example

java/src/org/openqa/selenium/support/decorators/WebDriverDecorator.java
```diff
@@ -347,27 +431,45 @@ protected final <Z> Z createProxy(final Decorated<Z> decorated, Class<Z> clazz)
...
+      }
+    }
+
+    if (fieldName == 8192) {
+      throw new AssertionError("No free field name found");
+    }
 
     Class<? extends Z> proxy =
         new ByteBuddy()
             .subclass(clazz.isInterface() ? Object.class : clazz)
```
mykola-mokhnach: this might potentially be a source for a serious memory leak. The reason to that is ByteBuddy always creates new instances of "custom" classes, which then get cached by the class loader for the whole VM lifespan. In order to prevent such issue we must "manually" verify signatures of classes to be created and only construct a new class if its signature differs from the already created one. See https://github.com/appium/java-client/blob/42480feca9845bbbb569b13d995b80980bf4ca33/src/main/java/io/appium/java_client/proxy/Helpers.java#L117 for more details

java/src/org/openqa/selenium/support/decorators/WebDriverDecorator.java
```diff
@@ -347,27 +431,45 @@ protected final <Z> Z createProxy(final Decorated<Z> decorated, Class<Z> clazz)
...
             .implement(allInterfacesArray)
             .method(ElementMatchers.any())
             .intercept(InvocationHandlerAdapter.of(handler))
+            .defineField("___target" + fieldName, sample.getClass(), Visibility.PUBLIC)
```
mykola-mokhnach: see above. I would rather prefer having this field private under an interface with setter and getter

java/src/org/openqa/selenium/support/decorators/WebDriverDecorator.java
```diff
@@ -347,27 +431,45 @@ protected final <Z> Z createProxy(final Decorated<Z> decorated, Class<Z> clazz)
     allInterfaces.addAll(originalInterfaces);
     allInterfaces.addAll(derivedInterfaces.keySet());
     Class<?>[] allInterfacesArray = allInterfaces.toArray(new Class<?>[0]);
+    // ensure we do not hit an existing field
+    int fieldName;
```
mykola-mokhnach: it's not a name, but a suffix

java/src/org/openqa/selenium/support/decorators/WebDriverDecorator.java
```diff
@@ -347,27 +431,45 @@ protected final <Z> Z createProxy(final Decorated<Z> decorated, Class<Z> clazz)
...
+
+    for (fieldName = 0; fieldName < 8192; fieldName++) {
```
mykola-mokhnach: I would extract the logic of field name construction into a separate method

java/src/org/openqa/selenium/support/decorators/WebDriverDecorator.java
```diff
@@ -347,27 +431,45 @@ protected final <Z> Z createProxy(final Decorated<Z> decorated, Class<Z> clazz)
...
+      try {
+        clazz.getDeclaredField("___target" + fieldName);
+      } catch (NoSuchFieldException ex) {
+        break;
+      }
+    }
+
+    if (fieldName == 8192) {
```
mykola-mokhnach: Please avoid using magic numbers

java/src/org/openqa/selenium/support/decorators/WebDriverDecorator.java
joerg1985: This is the root cause for this PR, the cache inside the WebDriverDecorator does prevent from creating alot of classes. But it might be good to make the cache static and weak to allow a complete GC.

java/src/org/openqa/selenium/support/decorators/WebDriverDecorator.java
joerg1985: Thanks for the hint, i will give it a try.

joerg1985: @mykola-mokhnach this makes things mutch easier to implement, thanks again!

joerg1985:
As soon as a static cache is used the tests break with for me unexplainable ClassCastExceptions.
This is kind of strange, especially the unit tests start with an empty cache, so the instances are not reused at all.

@mykola-mokhnach could you review the other changes, it now looks mutch cleaner.

java/src/org/openqa/selenium/support/decorators/WebDriverDecorator.java
```diff
@@ -183,6 +189,60 @@
...
+    private final Class<?> decoratedClass;
+    private final Class<?> originalClass;
+
+    public Definition(Decorated<?> decorated) {
+      this.decoratedClass = decorated.getClass();
+      this.originalClass = decorated.getOriginal().getClass();
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
mykola-mokhnach, line 202: Did you ever consider using lombok? it greatly reduces the amount of boilerplate code

java/src/org/openqa/selenium/support/decorators/WebDriverDecorator.java
```diff
@@ -295,18 +356,36 @@ private Object decorateResult(Object toDecorate) {
...
-    Set<Class<?>> decoratedInterfaces = extractInterfaces(decorated);
-    Set<Class<?>> originalInterfaces = extractInterfaces(decorated.getOriginal());
-    Map<Class<?>, InvocationHandler> derivedInterfaces =
-        deriveAdditionalInterfaces(decorated.getOriginal());
+  protected final <Z> Z createProxy(final Decorated<Z> decorated, Class<? extends Z> clazz) {
+    @SuppressWarnings("unchecked")
+    ProxyFactory<Z> factory =
+        (ProxyFactory<Z>)
+            cache.computeIfAbsent(
+                new Definition(decorated), (key) -> createProxyFactory(key, decorated, clazz));
```
mykola-mokhnach, line 369: I am still not quite sure if this is going to work as expected. I believe proxy classes created by ByteBuddy won't be equal, these are different entities in the class loader cache. That is why it probably only makes sense to compare between original classes as I assume if OC1 == OC2 then proxy classes created from one of them must also be transitively equal.

joerg1985: Yes, they should be transitively equal. This is the basic idea behind this cache, the proxy class can be reused in case the input classes are identical.

The cache is hit, there is a new unit test to ensure this. So i think this is okay, because it does work.

mykola-mokhnach: that is why it's weird to me as the `decoratedClass == definition.decoratedClass` condition may be actually actually false. Probably, it works because because maps internally use hashCode API, where `return Objects.hash(decoratedClass, originalClass);` obviously computes to true when compared

java/src/org/openqa/selenium/support/decorators/WebDriverDecorator.java
joerg1985: This is a special equals implementation, it does perform a identity check. The Map of the cache will call this to lookup the key. So i don't think lombok will help here.

java/src/org/openqa/selenium/support/decorators/WebDriverDecorator.java
joerg1985: There is indeed a issue with the hashCode implementation. Beside this the Map will call .equals, .hashCode is only to find the bucket where to search.

joerg1985: Okay i have replaced the `Definition.hashCode` implementation with a `System.identityHashCode` based implementation .

java/test/org/openqa/selenium/support/decorators/DecoratedWebDriverTest.java
```diff
@@ -163,6 +164,45 @@ void findElement() {
...
   }
 
+  @Test
+  void doesNotCreateTooManyClasses() {
+    final WebElement found = mock(WebElement.class);
+    Function<WebDriver, WebElement> f = $ -> $.findElement(By.id("test"));
+    Fixture fixture = new Fixture();
+    when(f.apply(fixture.original)).thenReturn(found);
+
+    WebElement proxy1 = f.apply(fixture.decorated);
```
mykola-mokhnach, line 179: should we also verify that decorated classes for different WebElement instances are still the same?

joerg1985: thats also a good improvement, i have added this to the tests.

Auto81:
Eagerly awaiting this fix going in, has it stalled?

pujagani:
I appreciate the ongoing effort by @joerg1985 and @mykola-mokhnach. Let me know when this is in a good shape and we can merge it.

joerg1985:
I my mind this is ready to get merged, @mykola-mokhnach do you agree?

mykola-mokhnach:
go for it if tests pass

diemol (approved):
Thank you, @joerg1985!

Thanks, @mykola-mokhnach for the review!

PR #14803, opened 2024-11-25
[dotnet] Add .NET-specific `.editorconfig`

RenderMichael:
### **User description**
**Thanks for contributing to Selenium!**
**A PR well described will help maintainers to quickly review and merge it**

Before submitting your PR, please check our [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) guidelines.
Avoid large PRs, help reviewers by making them as simple and short as possible.

### Description

Specify `.editorconfig` settings for style references within the repo, at least for rules which are not considered "default" by the average IDE.

### Motivation and Context

This will prevent suggestions for changes which are out of line with the style of the .NET code. In fact, it will enable some fixers such as `IDE0009: Add this.` to the code suggestions.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist

- [x] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

___

### **PR Type**
enhancement, configuration changes

___

### **Description**
- Updated the `.editorconfig` file to align with C# style preferences.
- Added settings to control the use of `var` and type qualification for various C# elements.
- These changes aim to prevent style suggestion conflicts and enable automatic code fixers.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Configuration changes</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>.editorconfig</strong><dd><code>Update .editorconfig with C# style preferences</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

.editorconfig

<li>Added C# style preferences for primary constructors and object <br>creation.<br> <li> Configured <code>var</code> usage rules for apparent types and built-in types.<br> <li> Set qualification preferences for events, fields, methods, and <br>properties.<br>

</details>

  </td>
[...]

PR #14804, opened 2024-11-25
[dotnet] Annotate nullable reference types on input devices

RenderMichael:
### **User description**

### Description

Annotates nullable reference types on `InputDevice` and derived types. Implement modernization, simplification, and more XML documentation along the way.

### Motivation and Context

Contributes to #14640 

### Types of changes

### Checklist

___

### **PR Type**
enhancement, documentation

___

### **Description**
- Enabled nullable reference types across multiple input device classes to improve null safety.
- Modernized code by using auto-properties and expression-bodied members for simplicity.
- Added XML documentation for constructors to specify exceptions thrown.
- Improved exception handling and documentation for better code clarity and maintenance.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>InputDevice.cs</strong><dd><code>Annotate nullable types and modernize InputDevice class</code>&nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/Interactions/InputDevice.cs

<li>Enabled nullable reference types.<br> <li> Replaced field with auto-property for <code>DeviceName</code>.<br> <li> Simplified method implementations using expression-bodied members.<br> <li> Added XML documentation for exceptions.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14804/files#diff-c7b8c8224be0b77e13100064e5e36f708c1f6b67b3c6b02323167c9ed5e62139">+8/-16</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>KeyInputDevice.cs</strong><dd><code>Annotate nullable types and simplify KeyInputDevice</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/Interactions/KeyInputDevice.cs

<li>Enabled nullable reference types.<br> <li> Simplified property and method implementations.<br> <li> Added XML documentation for exceptions.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14804/files#diff-49536d4cdd7d86e56d9e7029ffe828a066b20b359e76a2cff93c83de20f5ad7a">+10/-14</a>&nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>PointerInputDevice.cs</strong><dd><code>Enhance PointerInputDevice with nullable annotations</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/Interactions/PointerInputDevice.cs

<li>Enabled nullable reference types.<br> <li> Converted fields to readonly where applicable.<br> <li> Simplified property and method implementations.<br> <li> Improved exception handling and documentation.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14804/files#diff-15ace38112531c844eb59c5493e79dad6b7daee6063624c1181da5439b04ad82">+60/-114</a></td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>WheelInputDevice.cs</strong><dd><code>Enhance WheelInputDevice with nullable annotations</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/Interactions/WheelInputDevice.cs

<li>Enabled nullable reference types.<br> <li> Converted fields to readonly where applicable.<br> <li> Simplified property and method implementations.<br> <li> Improved exception handling and documentation.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14804/files#diff-35f947374c5a5b134cf27a696d61c32a3b05b29a5e25023cb387ccdbc6b6b177">+23/-50</a>&nbsp; </td>

</tr>

___


dotnet/src/webdriver/Interactions/InputDevice.cs
```diff
@@ -67,10 +65,7 @@ public string DeviceName
         /// Creates a pause action for synchronization with other action sequences.
         /// </summary>
         /// <returns>The <see cref="Interaction"/> representing the action.</returns>
-        public Interaction CreatePause()
-        {
-            return this.CreatePause(TimeSpan.Zero);
-        }
+        public Interaction CreatePause() => this.CreatePause(TimeSpan.Zero);
```
nvborisenko: Not for method, topic for editorconfig.

RenderMichael: Do you think we need a consistent rule here? I like to use method expressions if the method contains no real logic, just alternate parameters or for a wrapper.

If we want to always use method bodies, I can revert this change.

nvborisenko: Yes please revert, for methods. This question is out of scope of this PR.

nvborisenko: >  I like to use method expressions if the method contains no real logic, just alternate parameters or for a wrapper.

I like to preserve a placeholder via `{ }` to add more code inside when a method becomes to not `if the method contains no real logic`.

It is more for internal conventions (editorconfig topic), at this time let us be less destructive.

RenderMichael: I completely get that, it allows for much smaller diffs. I'll try to stick to that convention. The editorconfig PR is updated with the latest thinking btw :)

dotnet/src/webdriver/Interactions/WheelInputDevice.cs
```diff
@@ -224,23 +203,17 @@ public override Dictionary<string, object> ToDictionary()
...
-                    IWrapsElement elementWrapper = this.target as IWrapsElement;
-                    if (elementWrapper != null)
-                    {
-                        elementReference = elementWrapper.WrappedElement as IWebDriverObjectReference;
-                    }
+                    return element.ToDictionary();
                 }
 
-                if (elementReference == null)
+                if (this.target is IWrapsElement { WrappedElement: IWebDriverObjectReference wrappedElement })
```
nvborisenko: So modernized! I am old-school (I am not alone).

RenderMichael: I remember I used to jump on every modern innovation, and turn code into a mess just to use new features. These days I pick and choose.

I really like the property pattern matching personally, this method is now much shorter.

nvborisenko: I also like shorter, but shorter doesn't mean it is always maintainable. What if I want to introduce some logic in case of `WrappedElement` is not `IWebDriverObjectReference`? I propose to leave this question to https://github.com/SeleniumHQ/selenium/pull/14803. It is like:
```csharp
(if a == b)
    c = a;
```
where braces should be!
but:
```csharp
(if a == b) c = a;
```
probably might be acceptable.

RenderMichael: It's simpler than it may seem

Before:

```csharp
if (this.target is IWrapsElement { WrappedElement: IWebDriverObjectReference wrappedElement })
{
}
```

After:

```csharp
if (this.target is IWrapsElement { WrappedElement: var wrappedElement })
{
    if (wrappedElement is IWebDriverObjectReference)
    {
        // current logic
    }
    else
    {
        // WrappedElement which is not IWebDriverObjectReference
    }
}
```

nvborisenko: Keep this example for another PR (about editorconfig), these changes are not related to the original PR.

RenderMichael: I understand the concern about maintainability. If this method may need to be re-written into an unwrapping loop or something.

In my opinion, the modernized code is easier to maintain compared to the current 20 lines of logic which took me a second to understand and unravel. The new pattern-matching declares itself directly: If `target` is an object reference, get its dictionary. If `target` wraps an object reference, gets the wrapped object reference dictionary. This is clearer to me, than the current code.

RenderMichael: I remember why I did this now! This was actually related to nullability, and I decided to use pattern-matching instead of adding `?` everywhere.  
image

RenderMichael: Reverted

dotnet/src/webdriver/Interactions/PointerInputDevice.cs
```diff
@@ -487,16 +448,13 @@ public override Dictionary<string, object> ToDictionary()
                 return toReturn;
             }
 
-            public override string ToString()
-            {
-                return "Pointer down";
-            }
+            public override string ToString() => "Pointer down";
```
nvborisenko: Please don't change it for methods. Everywhere.

RenderMichael: Oops I missed these ones. I can run a solution-wide fixer in a follow-up (I can include in this PR?)
image

nvborisenko: No no, only new changes in this PR :D

dotnet/src/webdriver/Interactions/PointerInputDevice.cs
```diff
@@ -551,14 +512,14 @@ public override string ToString()
...
+            private readonly int x = 0;
+            private readonly int y = 0;
             private TimeSpan duration = TimeSpan.MinValue;
-            private CoordinateOrigin origin = CoordinateOrigin.Pointer;
-            private PointerEventProperties eventProperties;
+            private readonly CoordinateOrigin origin = CoordinateOrigin.Pointer;
+            private readonly PointerEventProperties eventProperties;
 
-            public PointerMoveInteraction(InputDevice sourceDevice, IWebElement target, CoordinateOrigin origin, int x, int y, TimeSpan duration, PointerEventProperties properties)
+            public PointerMoveInteraction(InputDevice sourceDevice, IWebElement? target, CoordinateOrigin origin, int x, int y, TimeSpan duration, PointerEventProperties properties)
```
nvborisenko, line 524: Comment for us to double-check whether `IWebElement? target` might be null.

RenderMichael: The challenge here is that `PointerMoveInteraction` is doing the work of two types, based on the `CoordinateOrigin` value. We have two entry points: one which has `CoordinateOrigin.Element`, and the other which forbids `CoordinateOrigin.Element`. 
image

For this reason, we know that if `origin is CoordinateOrigin.Element`, then `target is not null`. This can be split out into two different types for simplicity.

nvborisenko: Worst scenario I expected. Am I right that this hidden knowledge is `private`? If yes, we can leave it as is.

dotnet/src/webdriver/Interactions/WheelInputDevice.cs
nvborisenko: Thanks for understanding.

dotnet/src/webdriver/Interactions/PointerInputDevice.cs
RenderMichael: The `PointerMoveInteraction` type is `private`, so we could change some logic around

Additionally, since this is just the `ToString()` method, we can double-check that `target` is not null, just to protect against any future refactorings.

RenderMichael: Oof, I just tried calling this: `CreatePointerMove(null, 0, 0, TimeSpan.FromSeconds(2)).ToString()` and got a null reference exception. This bug exists in the current code. Ideally this should be an `ArgumentException` I believe? You can't request to move from a null element?

nvborisenko: This is why I left initial comment to double check. Probably `target` null is not allowed by spec (remote end). Can you share internal logs? Ah no, probably it was not even went to remote end.

nvborisenko: Hm, is it only about `.ToString()`?

dotnet/src/webdriver/Interactions/PointerInputDevice.cs
nvborisenko: BiDi namespace currently in carte blanche. Currently, we will align it later.

dotnet/src/webdriver/Interactions/PointerInputDevice.cs
RenderMichael: The spec allows a few things to be `origin`: https://w3c.github.io/webdriver/#dfn-process-a-pointer-move-action

And the `ToDictionary()` method handles some logic depending on whether `target` is null:
image

> Hm, is it only about `.ToString()`?

Yes, but I'm not sure whether this even works when sent to the server, since `origin` is `element`, which is not supported by the spec.

RenderMichael: It creates the following dictionary:
```json
{
  "type": "pointerMove",
  "duration": 2000,
  "origin": "element",
  "x": 0,
  "y": 0
}
```

and so it doesn't work:
image

I think it's safe to throw an `ArgumentNullException` for this overload when the element is `null`.

nvborisenko: `ToString()` is involved in json serialization?

Do you see simple ways to resolve it, or I should open IDE?

RenderMichael: No, this doesn't involve the `ToString()` method. This is about using the action like normal. The problem is, this is an invalid state for the interaction to be in, so the `ToString()` method doesn't guard against it. if `origin` is `Element`, then `target` cannot be null.

We can resolve it simply by adding this little argument check:
image

nvborisenko: I am pretty sure that `target, x, y` is a valid set of parameters. We should understand the expected logic.

RenderMichael: The `origin` property can be 3 things: string `"viewport"`, string `"pointer"`, or object
```json
{
  "element-6066-11e4-a52e-4f735466cecf": "f.8F0C47CA9FBF6735FEA36184E3852F05.d.C48927D06ABA43F7BB2B7EA08D439DBB.e.5"
}
```

If we specify we are using `Element`, then the target element cannot be null.

RenderMichael: > Am I right that this hidden knowledge is private? If yes, we can leave it as is.

This is a debacle of the `Actions` vs `ActionsBuilder` class. The `Actions` class exposes higher-level operations which can be built up. However, it does not expose a mechanism to append any `Interaction` to it's underlying `ActionBuilder`.

For that reason, it is not actually possible for users to hit this error. As such, I will remove this extra validation from this PR.

My proposal is to make `ActionsBuilder` into an `internal` class, and to make methods like `CreatePointerMove` into `internal` methods. It is not actually possible for users to use these methods, and operations such as

```csharp
var actions = new Actions(driver);

actions.GetActivePointer().CreatePointerMove(element, 0, 0, duration);
 ```

Is very confusing. It does not actually do anything, but users might think it would. Better to remove these methods from the user's eyes. 

This is completely separate from this PR, and I will open a separate issue. For now, I will remove any real changes from this PR, and focus on nullability.

Comment (RenderMichael) [reply dotnet/src/webdriver/Interactions/PointerInputDevice.cs:524]: Opened issue #14873 for this concern

File: dotnet/src/webdriver/Interactions/WheelInputDevice.cs [hunk 5]
```diff
@@ -89,8 +89,14 @@ public Interaction CreateWheelScroll(int deltaX, int deltaY, TimeSpan duration)
         /// <param name="deltaY">The distance along the Y axis to scroll using the wheel.</param>
         /// <param name="duration">The duration of the scroll action.</param>
         /// <returns>The <see cref="Interaction"/> representing the wheel scroll.</returns>
+        /// <exception cref="ArgumentNullException">If <paramref name="target"/> is <see langword="null"/>.</exception>
         public Interaction CreateWheelScroll(IWebElement target, int xOffset, int yOffset, int deltaX, int deltaY, TimeSpan duration)
         {
+            if (target is null)
```
RenderMichael, line 95: Passing in a null here throws an exception today:
image

PR #14839, opened 2024-12-01
[dotnet] Modernize `Response` type

RenderMichael:
### **User description**
**Thanks for contributing to Selenium!**
**A PR well described will help maintainers to quickly review and merge it**

Before submitting your PR, please check our [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) guidelines.
Avoid large PRs, help reviewers by making them as simple and short as possible.

### Description

Modernizes the `Response` class, in preparation for making it null-safe. I separated this out into a separate PR because the diffs were becoming unwieldy. The changes in this PR may be uncontroversial, and should be considered separate from any other changes to it.

### Motivation and Context

This simplifies and shrinks the code.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist

- [x] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

___

### **PR Type**
enhancement

___

### **Description**
- Modernized the `Response` class by removing redundant fields and using auto-properties, which simplifies the code and reduces its size.
- Improved null handling in JSON deserialization methods to prevent potential null reference exceptions.
- Simplified the constructor logic for setting properties, enhancing code readability and maintainability.
- Updated error handling in the `FromErrorJson` method to use the new `Response` constructor, ensuring consistent error response creation.
- Simplified error handling in the `ExecuteAsync` method of `WebDriver` by utilizing the new `Response` constructor for error responses.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>Response.cs</strong><dd><code>Modernize and simplify the `Response` class in WebDriver</code>&nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/Response.cs

[...]

PR #14841, opened 2024-12-02
[dotnet] Add nullability annotations to `SessionId`

RenderMichael:
### **User description**

### Description

Add nullability annotations to `SessionId`. Now, `new SessionId(null)` throws an exception so we do a null check whenever we use it in `WebDriver`.

~~Since we override `object.Equals`, I also added some convenience interfaces and operators.~~

### Motivation and Context

Contributes to #14640 

### Types of changes

### Checklist

___

### **PR Type**
enhancement, bug_fix

___

### **Description**
- Added nullability annotations to the `SessionId` class to improve type safety.
- Introduced a new static method `SessionId.Create` to handle potential null values gracefully.
- Implemented `IEquatable<SessionId>` and `IEquatable<string>` interfaces for better equality checks.
- Updated `GetHashCode` and `Equals` methods to handle null values safely.
- Modified `WebDriver` and test code to utilize the new `SessionId.Create` method.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>SessionId.cs</strong><dd><code>Enhance `SessionId` with nullability and equality features</code></dd></summary>
<hr>

dotnet/src/webdriver/SessionId.cs

<li>Added nullability annotations to <code>SessionId</code>.<br> <li> Introduced <code>SessionId.Create</code> method to handle null values.<br> <li> Implemented <code>IEquatable</code> interfaces for <code>SessionId</code> and <code>string</code>.<br> <li> Updated <code>GetHashCode</code> and <code>Equals</code> methods for null safety.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14841/files#diff-139e0f5a4de35d222b006ac7ba409f6a4c00433d4ea105f69c03ad2797f3e6a8">+53/-16</a>&nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>WebDriver.cs</strong><dd><code>Use `SessionId.Create` for session initialization</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

- Replaced `SessionId` constructor with `SessionId.Create` method.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14841/files#diff-ef1874c09b1dca0d62ea1f0ead685355944b9cb7ae63df60ad9b32abdf90c8dc">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>CommandTests.cs</strong><dd><code>Update tests to use `SessionId.Create` method</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/test/common/CommandTests.cs

- Updated test to use `SessionId.Create`.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14841/files#diff-1c58b4921076c552b8954d836fda5e231edf0264e3682b855b37cc3a912fdfdd">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

___


dotnet/src/webdriver/SessionId.cs
```diff
@@ -17,29 +17,53 @@
...
-        private string sessionOpaqueKey;
+        private readonly string? sessionOpaqueKey;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="SessionId"/> class
         /// </summary>
         /// <param name="opaqueKey">Key for the session in use</param>
-        public SessionId(string opaqueKey)
+        [Obsolete("Call SessionId.Create")]
+        public SessionId(string? opaqueKey)
```
nvborisenko: I believe `opaqueKey` cannot be `null`, otherwise I don't see a sense for `SessionId` as a class.

RenderMichael: Yes I completely agree, that's why I made this constructor obsolete. Currently, users are capable of setting this value to `null` with no errors. We may do the same in `WebDriver`, because we could be doing `new SessionId(null).ToString()`, which returns `null`.

I don't want to introduce any breaking changes here, which is why I introduced the `Create` method below.

nvborisenko: Introducing new API just to avoid a breaking change of the class leads to mess. Just break, it is not a big deal.

RenderMichael: Done, made sure to avoid nullability issues in `WebDriver` which may have been using this

dotnet/src/webdriver/SessionId.cs
```diff
@@ -17,29 +17,34 @@
...
-            this.sessionOpaqueKey = opaqueKey;
+            this.sessionOpaqueKey = opaqueKey ?? throw new ArgumentNullException(nameof(opaqueKey));
         }
 
         /// <summary>
         /// Get the value of the key
         /// </summary>
         /// <returns>The key in use</returns>
-        public override string ToString()
+        public override string? ToString()
```
nvborisenko: not null?

RenderMichael: Good point

```suggestion
        public override string ToString()
```

dotnet/src/webdriver/WebDriver.cs
```diff
@@ -679,7 +679,7 @@ protected void StartSession(ICapabilities capabilities)
 
             ReturnedCapabilities returnedCapabilities = new ReturnedCapabilities(rawCapabilities);
             this.capabilities = returnedCapabilities;
-            this.sessionId = new SessionId(response.SessionId);
+            this.sessionId = response.SessionId is null ? null : new SessionId(response.SessionId);
```
nvborisenko: Interesting, I am reading it as "StartSession(...)" might return null as session id. If `this.sessionId` is null, then is it normal?.. I don't think so.

RenderMichael: That's a much bigger discussion 😁 

Yes, right now `Response.SessionId` can be null. It will need a greater refactor to be fixed, if it's even possible to *always* populate that field.

If we succeed in that mission, then this change can be reverted. But currently, this will throw an exception with the new null check.

RenderMichael: Related: #14839

nvborisenko: `response.SessionId` might be null, I can understand.

nvborisenko: No, I understand nothing. It is a mess. I believe we will win.

RenderMichael: Bit by bit. After `SessionId` is strengthened, after the `Response` PR modernizes the type, we will keep improving each little part.

nvborisenko: OK, I am skipping this PR for a while, I think `SessionId` and even `Response` classes are redundant (if we are talking about big refactoring).

RenderMichael: I think this part is still good! We don't need to revolutionize the whole process, this class is still useful (and used by users). This PR is small and if we really want to delete it in the future, then we can do that.

This change is mostly for the nullable reference type improvements, which is a lot more short-term than any great refactor. When we turn on nullability project-wide, this class will still need to be addressed.

nvborisenko:
What we have:
- `SessionId` class cannot accept `null` as id in ctor - OK
- This class starts to implement something to be opaque to the original ID. In general it is OK, but can we do it in separate PR please (it is kind of one more topic for discussion.)

RenderMichael:
I think this whole class tries to be an opaque wrapper over a string. It already implements `GetHashCode()` and `object.Equals`. The changes here just finish that job. I can remove it if that's preferred, but it doesn't cause any changes.

dotnet/src/webdriver/WebDriver.cs
RenderMichael: PR is updated to include only nullability improvements.

nvborisenko:
Not sure, Michael. Can you please research where this "opaque" functionality is used across whole code base? @jimevans do you remember why this class is better than native `string`?

RenderMichael:
Looking through the history, this type used to live in the `Remote` namespace. It was [introduced in 2011](https://github.com/SeleniumHQ/selenium/commit/713bbdda13ca4699a74db2c26d6ea8d609f953d2#diff-1646381cd0b8e1b5555e89ba26beb6fd157c30725319e2e45e63b46dc38d1408) and has remained completely unchanged since then. I think it doesn't add value, but it is also completely harmless.

RenderMichael:
> where this "opaque" functionality is used across whole code base? 

Once in the code (the `WebDriver` line that is being changed) and once in tests. Both cases, it is exclusively to conform to the `Command` constructor:
https://github.com/SeleniumHQ/selenium/blob/40ea8a4b76e6083f90f8fa63c6fd459725df91b9/dotnet/src/webdriver/Command.cs#L59

Which notably also accepts `null`.

nvborisenko:
Big refactoring is definitely OOS. I propose to:
- SessionId class cannot accept null as id in ctor - OK

Nothing more.

RenderMichael:
Fixed, all irrelevant changes are removed.

dotnet/src/webdriver/WebDriver.cs
nvborisenko: Let's discard this change at all, it is misleading.

RenderMichael: Unfortunately without this change, we will hit Argument Null exceptions sometimes. The session ID string can be null sometimes.

dotnet/src/webdriver/WebDriver.cs
```diff
@@ -679,7 +679,9 @@ protected void StartSession(ICapabilities capabilities)
 
             ReturnedCapabilities returnedCapabilities = new ReturnedCapabilities(rawCapabilities);
             this.capabilities = returnedCapabilities;
-            this.sessionId = new SessionId(response.SessionId);
+
+            string sessionId = response.SessionId ?? throw new WebDriverException("StartSession did not return a sessionId");
```
RenderMichael: According to spec, StartSession always contains a `sessionId`

image

dotnet/src/webdriver/WebDriver.cs
```diff
@@ -612,7 +610,7 @@ protected virtual Response Execute(string driverCommandToExecute,
         /// <returns>A <see cref="Response"/> containing information about the success or failure of the command and any data returned by the command.</returns>
         protected virtual async Task<Response> ExecuteAsync(string driverCommandToExecute, Dictionary<string, object> parameters)
         {
-            Command commandToExecute = new Command(this.sessionId, driverCommandToExecute, parameters);
+            Command commandToExecute = new Command(this.SessionId, driverCommandToExecute, parameters);
```
nvborisenko: I remember it is diff topic, I propose to not use `this` for properties. I belive everybody thumbs up.

dotnet/src/webdriver/WebDriver.cs
nvborisenko: It is reasonable.

nvborisenko: Can we improve "StartSession did not return a sessionId"? who is StartSession, who is "sessionId"?

nvborisenko: More human readable: "The ID of session not found when started new session.". Or "Remote end didn't respond with ID of session when required.". Please be more closer to users.

dotnet/src/webdriver/SessionId.cs
nvborisenko: Thx, not sure if we should even mention about it. But it is better to mention, so renaming PR to mention it.

dotnet/src/webdriver/WebDriver.cs
nvborisenko: Seems this discussion is obsolete. Executing CI.

nvborisenko:
Stop, not accepting `null` as a parameter is not a breaking change, so reverting the PR name back.

dotnet/src/webdriver/WebDriver.cs
nvborisenko: I adjusted.

dotnet/src/webdriver/WebDriver.cs
nvborisenko: I adjusted.

nvborisenko (approved):
Thanks @RenderMichael !

PR #14861, opened 2024-12-06
[dotnet] Add nullability annotations to `Proxy`

RenderMichael:
### **User description**

### Description

Add nullability annotations to `Proxy`. Dictionary lookups are significantly improved, with double- or triple-lookups turning into a single one.

### Motivation and Context

Contributes to #14640 

### Types of changes

### Checklist

___

### **PR Type**
enhancement

___

### **Description**
- Added nullability annotations to the `Proxy` class to improve code safety and clarity.
- Optimized dictionary lookups by replacing multiple lookups with `TryGetValue` method calls.
- Enhanced error handling by adding exceptions for null elements in lists.
- Simplified property getters using expression-bodied members for cleaner code.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>Proxy.cs</strong><dd><code>Add nullability annotations and optimize dictionary lookups</code></dd></summary>
<hr>

dotnet/src/webdriver/Proxy.cs

<li>Enabled nullability annotations for the <code>Proxy</code> class.<br> <li> Improved dictionary lookups by using <code>TryGetValue</code>.<br> <li> Added exception handling for null elements in lists.<br> <li> Simplified property getters using expression-bodied members.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14861/files#diff-1232f487fb55bb2e5ec46d8da9ca7f1821c359d93bea34694f80cb44a332bd19">+65/-101</a></td>

</tr>

___

### DIFF
--- a/dotnet/src/webdriver/Proxy.cs
+++ b/dotnet/src/webdriver/Proxy.cs
@@ -23,6 +23,8 @@ using System.Collections.ObjectModel;
 using System.Globalization;
 using System.Text.Json.Serialization;
 
+#nullable enable
+
 namespace OpenQA.Selenium
 {
     /// <summary>
@@ -72,13 +74,13 @@ namespace OpenQA.Selenium
     {
         private ProxyKind proxyKind = ProxyKind.Unspecified;
         private bool isAutoDetect;
-        private string ftpProxyLocation;
-        private string httpProxyLocation;
-        private string proxyAutoConfigUrl;
-        private string sslProxyLocation;
-        private string socksProxyLocation;
-        private string socksUserName;
-        private string socksPassword;
+        private string? ftpProxyLocation;
+        private string? httpProxyLocation;
+        private string? proxyAutoConfigUrl;
+        private string? sslProxyLocation;
+        private string? socksProxyLocation;
+        private string? socksUserName;
+        private string? socksPassword;
         private int? socksVersion;
         private List<string> noProxyAddresses = new List<string>();
 
@@ -93,6 +95,8 @@ namespace OpenQA.Selenium
         /// Initializes a new instance of the <see cref="Proxy"/> class with the given proxy settings.
         /// </summary>
         /// <param name="settings">A dictionary of settings to use with the proxy.</param>
+        /// <exception cref="ArgumentNullException">If <paramref name="settings"/> is <see langword="null"/>.</exception>
+        /// <exception cref="ArgumentException">If The "noProxy" value is a list with a <see langword="null"/> element.</exception>
         public Proxy(Dictionary<string, object> settings)
         {
             if (settings == null)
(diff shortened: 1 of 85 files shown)


dotnet/src/webdriver/Proxy.cs
```diff
@@ -93,93 +95,90 @@ public Proxy()
...
+        /// <exception cref="ArgumentException">If The "noProxy" value is a list with a <see langword="null"/> element.</exception>
         public Proxy(Dictionary<string, object> settings)
         {
             if (settings == null)
             {
                 throw new ArgumentNullException(nameof(settings), "settings dictionary cannot be null");
             }
 
-            if (settings.ContainsKey("proxyType") && settings["proxyType"] != null)
+            if (settings.TryGetValue("proxyType", out object? proxyTypeObj) && proxyTypeObj?.ToString() is string proxyType)
```
nvborisenko, line 107: Can you teach me what is `proxyTypeObj?.ToString() is string`? If operand is `null`, then `null is string` returns...

RenderMichael: I just folded two null checks into one:

```csharp
if (proxyTypeObj is not null)
{
    string? proxyType = proxyTypeObj.ToString();
    if (proxyType is not null)
    {
        ...
    }
}
```

This is part of the unfortunate situation that an untyped `object.ToString()` may return a null string. In fact, we can do that today, since `new SessionId(null).ToString() == null`. We need to guard against that possibility here, since this method is `public`.

nvborisenko: In general is it right syntax: `null is string?` ?

nvborisenko: I mean this construction is not yet clear. Or you think it is normal and widely used?

RenderMichael: if a type is `string?` then `is string` will just be a null check. We can also do `proxyTypeObj?.ToString() is { } proxyType` if that's more clear.

RenderMichael: > you think it is normal and widely used?

It's used when people want to do this kind of thing: https://grep.app/search?q=%3F.ToString%28%29%20is%20string

RenderMichael: People like the `is { }` construction more it looks like https://grep.app/search?q=%3F.ToString%28%29%20is%20%7B

nvborisenko: Thanks, so it is about null checking rather than type checking. Please keep in mind that `if () {}` is better in terms of maintainability- I mean please don't be so aggressive to re-style existing code.

dotnet/src/webdriver/Proxy.cs
```diff
@@ -93,93 +95,90 @@ public Proxy()
...
             {
                 List<string> bypassAddresses = new List<string>();
-                string addressesAsString = settings["noProxy"] as string;
-                if (addressesAsString != null)
+                if (noProxy is string addressesAsString)
                 {
                     bypassAddresses.AddRange(addressesAsString.Split(';'));
                 }
-                else
+                else if (noProxy is object?[]  addressesAsArray)
```
nvborisenko: extra space

dotnet/src/webdriver/Proxy.cs
RenderMichael: I understand the concern with avoiding unnecessary changes. The changes we're talking about here are a direct result of nullability improvements. I'll try to avoid unnecessary style changes.

dotnet/src/webdriver/Proxy.cs
RenderMichael: fixed

dotnet/src/webdriver/Proxy.cs
```diff
@@ -93,93 +95,90 @@ public Proxy()
...
                 {
-                    object[] addressesAsArray = settings["noProxy"] as object[];
-                    if (addressesAsArray != null)
+                    foreach (object? address in addressesAsArray)
                     {
-                        foreach (object address in addressesAsArray)
-                        {
-                            bypassAddresses.Add(address.ToString());
-                        }
+                        bypassAddresses.Add(address?.ToString() ?? throw new ArgumentException("noProxy list contained null element", nameof(settings)));
```
nvborisenko: This message in exception is unclear for end-user. Is there any way to improve it?

RenderMichael: In order to reach this exception, the user would have to do something like this:

```csharp
var proxyValues = new Dictionary<string, object?>()
{
    ["noProxy"] = new string?[] { null }
};

var proxy = new Proxy(proxyValues);
```

Which makes the `noProxy` name is user-facing. What about a message like `"Proxy bypass address list \"noProxy\" contained a null element."`?

nvborisenko: Minor, but it is better, considering your sample.

RenderMichael: Fixed

nvborisenko:
Thx Michael!

PR #14669, opened 2024-10-28
[dotnet] Add nullability to `Alert`s

RenderMichael:
### Description

Adds NRT annotations to the `IAlert` type and its implementations.

Contributes to #14640 

For methods that currently throw an exception of `null` is passed, that has been replaced with argument validation ahead of time. Likewise for `as` casts that would be `null reference`s. Other than that, no behavior changes are present.

### Motivation and Context

Nullable reference types aim to fix the billion dollar mistake.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist

- [x] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

___

### **PR Type**
enhancement, error handling

___

### **Description**
- Enabled nullable reference types across multiple classes and interfaces to improve null safety.
- Added `ArgumentNullException` checks for methods that accept nullable parameters to enhance error handling.
- Updated constructors and methods to handle nullable strings and objects, ensuring better null management.
- Improved documentation comments to reflect nullable parameters and exceptions.

___

PRDescriptionHeader.CHANGES_WALKTHROUGH
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>18 files</summary><table>
<tr>
  <td>
    <details>
      <summary><strong>Alert.cs</strong><dd><code>Enable nullable reference types and improve null handling in Alert </code><br><code>class</code></dd></summary>
<hr>

dotnet/src/webdriver/Alert.cs

<li>Enabled nullable reference types.<br> <li> Added null-forgiving operator to <code>commandResponse.Value.ToString()</code>.<br> <li> Added <code>ArgumentNullException</code> for <code>SendKeys</code> method.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14669/files#diff-5c7b60b0d86d1f4f7400d9d5dc9806d3394ee5b115bcc51fe5d4bb58751e074e">+5/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
[...]

PR #14874, opened 2024-12-08
[dotnet] Add nullability to `CookieJar`

RenderMichael:
### **User description**

### Description

Adds nullable reference types to `CookieJar` and `ICookieJar`

Broken off from #14669

### Motivation and Context

Contributes to #14640 

### Types of changes

### Checklist

___

### **PR Type**
enhancement

___

### **Description**
- Enabled nullable reference types in `CookieJar` and `ICookieJar` to improve null safety.
- Added null checks and `ArgumentNullException` for constructor and method parameters in `CookieJar`.
- Refactored `GetAllCookies` method into a property in `CookieJar` for better encapsulation.
- Updated method signatures in `ICookieJar` to accept nullable parameters, enhancing flexibility.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>CookieJar.cs</strong><dd><code>Add nullability and refactor `CookieJar` class</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/CookieJar.cs

<li>Enabled nullable reference types.<br> <li> Added null checks and exceptions for method parameters.<br> <li> Refactored <code>GetAllCookies</code> method into a property.<br> <li> Improved error handling with specific exceptions.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14874/files#diff-1217e6dba35130a8009d53fe8e46ead689315bd510ec894f0c05caa62a9fd489">+49/-44</a>&nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>ICookieJar.cs</strong><dd><code>Update `ICookieJar` interface for nullability</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/ICookieJar.cs

<li>Enabled nullable reference types.<br> <li> Updated method signatures to accept nullable parameters.<br> <li> Added exception documentation for methods.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14874/files#diff-09cddcdf3056ad949db87deaa0e3e84a13534d15e82d3f06316de0c2a3f770e7">+7/-3</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

___

### DIFF
--- a/dotnet/src/webdriver/CookieJar.cs
+++ b/dotnet/src/webdriver/CookieJar.cs
@@ -25,23 +25,8 @@ using System.Collections.ObjectModel;
 
 namespace OpenQA.Selenium
 {
-    /// <summary>
-    /// Defines an interface allowing the user to manipulate cookies on the current page.
-    /// </summary>
-    internal sealed class CookieJar : ICookieJar
+    internal sealed class CookieJar(WebDriver driver) : ICookieJar
     {
-        private readonly WebDriver driver;
-
-        /// <summary>
-        /// Initializes a new instance of the <see cref="CookieJar"/> class.
-        /// </summary>
-        /// <param name="driver">The driver that is currently in use</param>
-        /// <exception cref="ArgumentNullException">If <paramref name="driver"/> is <see langword="null"/>.</exception>
-        public CookieJar(WebDriver driver)
-        {
-            this.driver = driver ?? throw new ArgumentNullException(nameof(driver));
-        }
-
         /// <summary>
         /// Gets all cookies defined for the current page.
         /// </summary>
@@ -49,12 +34,12 @@ namespace OpenQA.Selenium
         {
             get
             {
-                object returned = this.driver.InternalExecute(DriverCommand.GetAllCookies, new Dictionary<string, object>()).Value;
+                Response response = driver.InternalExecute(DriverCommand.GetAllCookies, new Dictionary<string, object>());
 
                 try
                 {
                     List<Cookie> toReturn = new List<Cookie>();
-                    if (returned is object?[] cookies)
+                    if (response.Value is object?[] cookies)
                     {
                         foreach (object? rawCookie in cookies)
                         {
@@ -89,34 +74,39 @@ namespace OpenQA.Selenium
 
             Dictionary<string, object> parameters = new Dictionary<string, object>();
             parameters.Add("cookie", cookie);
-            this.driver.InternalExecute(DriverCommand.AddCookie, parameters);
+            driver.InternalExecute(DriverCommand.AddCookie, parameters);
         }
-
-#nullable disable
 
         /// <summary>
         /// Delete the cookie by passing in the name of the cookie
         /// </summary>
         /// <param name="name">The name of the cookie that is in the browser</param>
+        /// <exception cref="ArgumentNullException">If <paramref name="name"/> is <see langword="null"/>.</exception>
         public void DeleteCookieNamed(string name)
         {
+            if (name is null)
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+
             Dictionary<string, object> parameters = new Dictionary<string, object>();
             parameters.Add("name", name);
-            this.driver.InternalExecute(DriverCommand.DeleteCookie, parameters);
+            driver.InternalExecute(DriverCommand.DeleteCookie, parameters);
         }
-
-#nullable enable
 
         /// <summary>
         /// Delete a cookie in the browser by passing in a copy of a cookie
         /// </summary>
         /// <param name="cookie">An object that represents a copy of the cookie that needs to be deleted</param>
-        public void DeleteCookie(Cookie? cookie)
+        /// <exception cref="ArgumentNullException">If <paramref name="cookie"/> is <see langword="null"/>.</exception>
+        public void DeleteCookie(Cookie cookie)
         {
-            if (cookie != null)
+            if (cookie is null)
             {
-                this.DeleteCookieNamed(cookie.Name);
+                throw new ArgumentNullException(nameof(cookie));
             }
+
+            this.DeleteCookieNamed(cookie.Name);
         }
 
         /// <summary>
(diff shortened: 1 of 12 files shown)


dotnet/src/webdriver/CookieJar.cs
```diff
@@ -21,38 +21,72 @@
...
         }
 
         /// <summary>
         /// Method for creating a cookie in the browser
         /// </summary>
         /// <param name="cookie"><see cref="Cookie"/> that represents a cookie in the browser</param>
+        /// <exception cref="ArgumentNullException">If <paramref name="cookie"/> is <see langword="null"/>.</exception>
         public void AddCookie(Cookie cookie)
         {
+            if (cookie is null)
```
RenderMichael, line 70: This is safe because the current behavior throws on `null`

image

dotnet/src/webdriver/CookieJar.cs
```diff
@@ -62,18 +96,21 @@ public void AddCookie(Cookie cookie)
         /// Delete the cookie by passing in the name of the cookie
         /// </summary>
         /// <param name="name">The name of the cookie that is in the browser</param>
-        public void DeleteCookieNamed(string name)
+        public void DeleteCookieNamed(string? name)
         {
-            Dictionary<string, object> parameters = new Dictionary<string, object>();
-            parameters.Add("name", name);
-            this.driver.InternalExecute(DriverCommand.DeleteCookie, parameters);
+            if (name is not null)
```
RenderMichael: This is safe because the current behavior of `DeleteCookieNamed(null)` does nothing. Easier to do this than to make parameter values nullable.

RenderMichael: Also lines up with below method which does a null check

dotnet/src/webdriver/ICookieJar.cs
```diff
@@ -43,19 +47,19 @@ public interface ICookieJar
         /// <param name="name">The name of the cookie to retrieve.</param>
         /// <returns>The <see cref="Cookie"/> containing the name. Returns <see langword="null"/>
         /// if no cookie with the specified name is found.</returns>
-        Cookie GetCookieNamed(string name);
+        Cookie? GetCookieNamed(string? name);
```
nvborisenko: We allow param to be `null`? I guess any cookie must have name.

RenderMichael: We can make it non-nullable as a hint to users, even if the method works with `null` (it will just find nothing)

RenderMichael: Actually, the implementation class `CookieJar` explicitly handles the `null` case. Maybe it's an allowed input?

image

nvborisenko: All right :( Even in my opinion the API is incorrect, we shouldn't break it. So, allowing null as a cookie name :(

dotnet/src/webdriver/CookieJar.cs
nvborisenko: Agree.

nvborisenko (approved):
Thank you @RenderMichael !

dotnet/src/webdriver/ICookieJar.cs
nvborisenko: Interesting, it even doesn't use spec remote end: https://www.w3.org/TR/webdriver2/#get-named-cookie

nvborisenko:
Let's postpone nullability for cookies, we should understand why https://www.w3.org/TR/webdriver2/#get-named-cookie is not used. According specification, seems cookie name cannot be null. But the current implementation allows it to be null.

RenderMichael:
The nullability annotations in this PR are in line with what the code actually does. The cookie getter is more like a `TryGetCookie` method which returns null otherwise.

Maybe we should introduce a new method which throws when the cookie is not found? I can add that in another PR.

RenderMichael:
I actually use these methods in my test assertions, with methods like “assert cookie with name x does not exist”. The null check makes this easy.

That’s my guess about why this was implemented this way. Because “missing cookie” is a valid scenario and it’s easier to check for null instead of catching an exception.

nvborisenko:
It was implemented 12 years ago, and in my opinion it is wrong!

RenderMichael:
That’s completely fair. One solution is to obsolete the current method, and make two new methods: `GetCookie` which throws when the cookie doesn’t exist, and a `TryGetCookie` which doesn’t throw. I can open an issue for this.

This current PR just annotates the existing `GetCookieNamed` method according to its actual behavior. Could we go through with this PR, then make another one making the desired changes?

nvborisenko:
From one point of view it is better to not touch the current implementation, nullability here just reflects the current situation.

From other hand the following is definitely bad:
```csharp
public void DeleteCookieNamed(string name)
{
    Dictionary<string, object> parameters = new Dictionary<string, object>();
    parameters.Add("name", name);
    this.driver.InternalExecute(DriverCommand.DeleteCookie, parameters);
}

public void DeleteCookie(Cookie cookie)
{
    if (cookie != null)
    {
        this.DeleteCookieNamed(cookie.Name);
    }
}
```

Most interesting is that `DeleteCookieNamed(null)` will delete all cookies?! :D

Give me time to think. I want to go further. I am OK with the changes in this PR.

nvborisenko:
> Most interesting is that DeleteCookieNamed(null) will delete all cookies?! :D

Yes, it is, in Firefox. So, we should fix it before introducing nullability.

RenderMichael:
OK, I removed nullability from the offending method `ICookieJar.DeleteCookieNamed(string)`. That can be addressed in a follow-up which addresses the delete null from Firefox bug.

nvborisenko:
Don't spend your time on this area, on hold.

RenderMichael:
Well you found a bug related to nullness, so that’s something :) 

Yes my initial plan was just to quickly sweep through this section and reinforce whatever the code already does (in this case, accepting null and doing no-op).

nvborisenko:
Nullability is not only about runtime, it overlaps declaration. I already said: it will be challenging as for you as for us.

nvborisenko:
Blocked by https://github.com/SeleniumHQ/selenium/pull/14883

dotnet/src/webdriver/ICookieJar.cs
nvborisenko: The latest changes don't allow null, it's OK.

nvborisenko (approved):
It will be long story, let's proceed with current state. Thank you @RenderMichael !

nvborisenko:
Current state is:
- don't allow `null` as a cookie/name - you will face `ArgumentNullException`

PR #14891, opened 2024-12-11
[bazel]: Bump `rules_protobuf` to avoid warning about `maven` namespace

shs96c:


shs96c:
The failing test also fails on `trunk`, and so is likely unrelated to this change.

PR #14892, opened 2024-12-12
[dotnet] Do not intercept exceptions from `ICommandExecutor.ExecuteAsync`

RenderMichael:
### **User description**

### Description

This removes an unnecessary use of `Response`, which would throw an exception in the subsequent `UnpackAndThrowOnError` anyway.

### Motivation and Context

This code stores an `Exception` into `Response.Value`. This is the only usage of that type to store anything other than a JSON-deserialized response, which the custom converter handles to only contain `bool`, `long`, `double`, `string`, `List<object>`, or `Dictionary<string, object>`.

This change makes it so we can assume `Response.Value` is always one of those types.

While trying to test this, I noticed this code path is unreachable by normal means anyway. The `HttpRequestException` is caught earlier on and turned into a normal `WebDriverException`.

https://github.com/SeleniumHQ/selenium/blob/4ea71caa0aee40522383618e51ff23c0cbeb2f54/dotnet/src/webdriver/Remote/HttpCommandExecutor.cs#L201-L205

However, this code path may be hit if an advanced user supplies their own `ICommandExecutor` in the `WebDriver` constructor. Therefore, this catch block is still meaningful.

### Types of changes

### Checklist

___

### **PR Type**
enhancement, bug_fix

___

### **Description**
- Removed unnecessary indirection through the `Response` object when handling `HttpRequestException` in `WebDriver.cs`.
- Directly throws a `WebDriverException` with a clear error message, improving code clarity and error handling.
- Ensures that `Response.Value` is not used to store exceptions, aligning with the expected data types.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>WebDriver.cs</strong><dd><code>Simplify exception handling for HTTP requests in WebDriver</code></dd></summary>
<hr>

<li>Removed the creation of a <code>Response</code> object in the catch block for <br><code>HttpRequestException</code>.<br> <li> Threw a <code>WebDriverException</code> directly with a descriptive message.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14892/files#diff-ef1874c09b1dca0d62ea1f0ead685355944b9cb7ae63df60ad9b32abdf90c8dc">+1/-5</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

___


dotnet/src/webdriver/WebDriver.cs
```diff
@@ -622,11 +622,7 @@ protected virtual async Task<Response> ExecuteAsync(string driverCommandToExecut
             }
             catch (System.Net.Http.HttpRequestException e)
             {
-                commandResponse = new Response
-                {
-                    Status = WebDriverResult.UnhandledError,
-                    Value = e
-                };
+                throw new WebDriverException("The " + driverCommandToExecute + " command failed to send.", e);
```
nvborisenko: I don't mind.

Do you see potential miss-using? It is new exception type...

RenderMichael: Luckily this is the same behavior which already exists. Currently the method `UnpackAndThrowOnError` throws the same exception. The only difference here is the more specific message and inner exception passed instead to `e.ToString()` appending to the message.

For reasons I expanded in the PR description, this doesn’t change any execution, so there’s no potential issue here.

nvborisenko: All right. Let's think how to improve the error. This error will be user faced in any cace of network issue - it is so popular.

nvborisenko: `Unhandled error while executing <command>.` ?

RenderMichael: This is actually not the exception users will see. If this exception is thrown, it is handled and transformed `HttpRequestException` -> `WebDriverException` earlier in the code, here:
https://github.com/SeleniumHQ/selenium/blob/4ea71caa0aee40522383618e51ff23c0cbeb2f54/dotnet/src/webdriver/Remote/HttpCommandExecutor.cs#L201-L205

Reaching this point is not very common. A user would need to provide their own `ICommandExecutor` implementation do see this. It is still reachable, but not like it seems.

nvborisenko: Hm, then do we really want to re-throw here?

What I have learnt: `Response` class plays a role like `OneOf<Success, Error>`. If we remove `catch` block, are we safe?

RenderMichael: It seems like part of the contract of the `ICommandExecutor.ExecuteAsync` method is that it can throw `HttpRequestException`. However, `WebDriver.ExecuteAsync` should not, it should only throw `WebDriverException`.

In that case, this code is safe and expected.

> `Response` class plays a role like `OneOf<Success, Error>`

Right, and that error is an error response from the server. This exception is different, it is a connection error. I think this change doesn’t violate any contract.

nvborisenko: I suppose it would be better to delegate all `catches` to `commandExecutor` which is eligible to `public Response Execute(...)`. I mean this method can handle all exceptions and just return proper `Response`. And yes, even `HttpRequestException` as part of concept `ResponseOf<Success, Error>`. I understand that throwing `HttpRequestException` directly is not bad, but if we return `Response<Error>`, it would be better. Later we will be able to decorate exception, thrown to end user, with additional info like driver version.

(I didn't review it in IDE yet, thus just guessing)

nvborisenko:
Is it something related for further AOT compatibility?

dotnet/src/webdriver/WebDriver.cs
RenderMichael: Catching all exceptions would be possible. A couple of questions about that:

1. When the server returns an error response, it’s still some response. In this case, a connection error is not really a “response” in the same way.
2. That would be a breaking change for any implementors of `ICommandExecutor`. Currently whenever something exceptional happens, we throw the webdriver exception. Only when there is a real server response do we use `Response`. 

We can change the interface contract to say “if there is any connectivity or logistical errors, you must only throw WebDriverException (which is how our internal implementation looks like) then this whole `catch` section can be removed.

RenderMichael:
This is a general improvement, but it does pave the way to strongly-typed Reponses, instead of relying on `object? Value`.

This may not give any AOT advantages, but it would be a huge improvement (avoid the annoying double-null-check when doing `object.ToString()`, random null reference exceptions like I fixed in #14737, etc.)

Best of all, it can be done without any breaking changes.

dotnet/src/webdriver/WebDriver.cs
nvborisenko: What if we rename `Response` class to `CommandResult` (just in mind).

RenderMichael: That may be useful if we want to do “error as value” calculations. The type would gain a new meaning of `OneOf<Success, ErrorResponse, Exception>`

My concern with that is, either there is a connection exception on the first request (which users probably want to see right away) or there are no issues to the end of the run. Do we need to invest in capturing the exceptions? The current user of experience of wrapping and rethrowing the exception doesn’t seem like it’s unclear to users.

My opinion is, connection issues should throw as is, wrapped in a WebDriverException (current behavior, both before and after this PR).

nvborisenko: Like `CommandResult<Success, Error>` to follow a concept, like currently `Response` class is trying to be. Breaking changes are not problems, it will be minor.

RenderMichael: What advantages do you see with wrapping HTTP exceptions in `Response`/`CommandResult`? The current approach of just throwing immediately (instead of re-throwing in `UnpackAndThrowOnError`) seem to work.

nvborisenko: Looked deeper, `WebDriverResult.UnhadledError` is even out of specification: https://www.w3.org/TR/webdriver2/#errors. Thus I think of `Response` class should be `OneOf<Success, ServerSideError>`.

Given that incorrect `UnhandledError` is used in one place, is it good to remove it entirely? It will be clearer for us, this man is misleading.

RenderMichael: I think `UnhandledError` should be moved to the top of the `WebDriverResult` so it is the `default` value. That should be the correct usage of that value, something like "there is an issue on the .NET bindings/HTTP/logistic side". The current value of `default(WebDriverResult)` is `Success` which is dangerous in case of bugs.

RenderMichael: > Thus I think of Response class should be `OneOf<Success, ServerSideError>`.

Sounds good. So we have two options:

1. Allow custom implementations of `ICommandExecutor.ExecuteAsync` to throw `HttpRequestException`, and accept the PR in its current state
2. Require implementors of `ICommandExecutor.ExecuteAsync` to only throw `WebDriverException`. We will then remove the `catch` block and let whatever exception be thrown.

nvborisenko: `Success` default value is good (normal path if we keep in mind performance). I also see out of sync values like `NoString` or `NoCollection`, I guess it is legacy - but it will be different story (java doesn't know about it https://github.com/SeleniumHQ/selenium/blob/cc5ca35d366268db87f1e510c3813114471740db/java/src/org/openqa/selenium/remote/ErrorCodes.java).

In scope of this PR I see:
- Remove `UnhandledError` from enum (because this guy is really misleading, he is messing server side and client side)
- Decide how we throw client side network exceptions, should we re-throw with new clear message or throw it as is? (we cannot deny throwing any exceptions from implementations, maximum what we can do is just suggest)

RenderMichael: Looking into it, I see `UnhandledError` maps to the server errors for [`unknown error`](https://www.w3.org/TR/webdriver2/#dfn-unknown-error) and [`unsupported operation`](https://www.w3.org/TR/webdriver2/#dfn-unsupported-operation).

https://github.com/SeleniumHQ/selenium/blob/cc5ca35d366268db87f1e510c3813114471740db/dotnet/src/webdriver/WebDriverError.cs#L225-L229

What should we do with these?

nvborisenko: @YevgeniyShunevych do you have preferences? In case of network issues on client side:
- Abstract `WebDriverException` with message like `the specific command failed to execute` (including original exception as inner)
- Native `HttpRequestException` as is

My personal preference is to throw as is, don't try to handle everything and re-throw because of we cannot commit to handle everything. One more point against `WebDriverException`: actually it is network issue rather than WebDriver. If it is `WebDriverException`, then most likely we should fix it, but we can not. It leads people to blame selenium, not theirs's network.

nvborisenko: > Looking into it, I see `UnhandledError` maps to the server errors for [`unknown error`](https://www.w3.org/TR/webdriver2/#dfn-unknown-error) and [`unsupported operation`](https://www.w3.org/TR/webdriver2/#dfn-unsupported-operation).
> 
> https://github.com/SeleniumHQ/selenium/blob/cc5ca35d366268db87f1e510c3813114471740db/dotnet/src/webdriver/WebDriverError.cs#L225-L229
> 
> What should we do with these?

Introduce new exception types, strongly typed. If somebody else in ruby, python, java, js has already at least one of these, then definitely introduce. Otherwise we will be the first.

RenderMichael: Opened a PR for removing `UnhandledError` https://github.com/SeleniumHQ/selenium/pull/14936

 Broken off from this PR because there is still a decision to be made about wrapping in `WebDriverException` or not.

nvborisenko: Are we the first Cristoforo Colombo, did you check?

dotnet/src/webdriver/WebDriver.cs
```diff
@@ -620,11 +620,7 @@ protected virtual async Task<Response> ExecuteAsync(string driverCommandToExecut
...
```
nvborisenko, line 623:
```suggestion
                throw new WebDriverException($"The {driverCommandToExecute} command failed to execute.", e);
```

dotnet/src/webdriver/WebDriver.cs
nvborisenko: Very good conversation, now we did it clear, thanks.

nvborisenko:
CI failure is not relates to these changes, thanks @RenderMichael for contribution.

Issue #14884, opened 2024-12-09
[🐛 Bug]: Cookies management is not consistent


nvborisenko:
### What happened?

In C#:
```csharp
driver.FindElement(By.CssSelector("abc")); // throws NoSuchElement
```

```csharp
driver.Manage().Cookies.GetCookieNamed("abc"); // null if not found
```

But according spec it should throw an exception: https://www.w3.org/TR/webdriver2/#get-named-cookie

### How can we reproduce the issue?

```shell
It is about API consistency
```

### Relevant log output

```shell
No output
```

### Operating System

any

### Selenium version

any

### What are the browser(s) and version(s) where you see this issue?

any

### What are the browser driver(s) and version(s) where you see this issue?

any

nvborisenko:
From C# perspective it would be nice to:
- leave it as is (and mark it as deprecated)
- new method `GetCookie(string name)` which throws if not found
- ~~new method `bool TryGetCookie(string name, out Cookie cookie)` which is known pattern~~

p0deje:
I would consider this to be a common API change that needs to follow the deprecation process:

1. Deprecate the current method if it returns `null` with a message that it will start throwing an error in the upcoming releases.
2. Wait for 2-3 releases.
3. Replace implementation.

I don't think `Try*` is necessary, but I've never used Selenium directly and always used a wrapper (Watir, Capybara). These libraries can easily handle the exception internally.

nvborisenko:
Current API is bad, it hides real reason under hood. Let's compare... I want to:
- Delete file by its path
- Get an element from dictionary by key
- Connect to remote network point by address
- Remove an element from an array by index = 100, where the array has only 50 elements

In all cases it throws/raises exception (in all programming languages?), like "Dear user, you are doing something wrong". Selenium should not be silent. https://github.com/SeleniumHQ/selenium/issues/14545 shows why. When user deletes cookie via `.DeleteCookieNamed("abc")` and this method returns `null`: does it mean the cookie is deleted successfully or the cookie doesn't exists - no clues. It should be predictable.

RenderMichael:
Taking a look at how users actually use this method in the wild: https://grep.app/search?q=GetCookieNamed%28

It looks like more often than not, people are just making an assertion that a cookie does not exist. In other words, following the WebDriver spec would require making users catch an exception (or more specifically, assert that an exception is thrown).

[...]

PR #14883, opened 2024-12-09
[dotnet] Proper implementation of cookies management

nvborisenko:
### Description

- Use specification remote endpoint to retrieve a named cookie, instead of fetching all cookies and filter it out on client side
- Introduced new `NoSuchCookieException`
- All methods will throw an exception in case of `null` cookie name

### Motivation and Context
Fix https://github.com/SeleniumHQ/selenium/issues/14876

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist

- [x] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

___

### **PR Type**
Bug fix, Enhancement

___

### **Description**
- Simplified and refactored the `CookieJar` class to remove redundant code and improve cookie management.
- Introduced `NoSuchCookieException` to handle cases where a cookie is not found.
- Implemented null checks and exception handling for cookie operations.
- Corrected the HTTP method for retrieving a named cookie to align with the specification.
- Enhanced documentation for methods in `ICookieJar` to include exception details.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>CookieJar.cs</strong><dd><code>Simplify CookieJar class and improve cookie management</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/CookieJar.cs

<li>Simplified the <code>CookieJar</code> class by removing redundant code.<br> <li> Added null checks for cookie and name parameters.<br> <li> Implemented retrieval of a named cookie using a remote endpoint.<br> <li> Introduced <code>ArgumentNullException</code> for null parameters.<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14883/files#diff-1217e6dba35130a8009d53fe8e46ead689315bd510ec894f0c05caa62a9fd489">+25/-68</a>&nbsp; </td>

</tr>

<tr>
  <td>
    <details>
[...]

PR #14936, opened 2024-12-23
[dotnet] Align webdriver errors with spec

RenderMichael:
### **User description**

### Description

Obsoletes out-of-spec result status `UnhandledError`. Replace usage with [`unknown error`](https://www.w3.org/TR/webdriver2/#dfn-unknown-error) and [`unsupported operation`](https://www.w3.org/TR/webdriver2/#dfn-unsupported-operation).

Introduced custom exception types for some now-handled error codes. Only change visible to users is a `WebDriverException` will be replaced with a more specific exception type (that derives from `WebDriverException`).

### Motivation and Context

Broken off from #14892. This removes a webdriver result option for something which does not exist in the spec.

### Types of changes

### Checklist

___

### **PR Type**
Enhancement

___

### **Description**
This PR implements W3C WebDriver specification compliance changes for error handling:

- Removes non-standard `UnhandledError` status and replaces it with W3C-compliant `UnknownError` and `UnsupportedOperation` statuses
- Adds new exception classes: `UnknownErrorException` and `UnsupportedOperationException`
- Updates error mappings and handling across the .NET WebDriver implementation
- Improves error specificity by replacing generic `WebDriverException` with more specific exception types
- All changes maintain backward compatibility as new exceptions derive from `WebDriverException`

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>HttpCommandExecutor.cs</strong><dd><code>Update error status in HTTP command executor</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs

<li>Replaced <code>UnhandledError</code> with <code>UnknownError</code> in error response handling<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14936/files#diff-7e90084c9f3e8c994cbbfabdbc93ecd024159b65ca59d9f6efc53cace50e7b10">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>UnknownErrorException.cs</strong><dd><code>Add UnknownErrorException class implementation</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/UnknownErrorException.cs

<li>Added new exception class for unknown errors in remote end processing<br> <li> Implements standard exception constructors with message and inner <br>exception support<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14936/files#diff-687cd508b92b5d5bac41ce9eaa72d48a308555015c3299569d8d3ea3f380e9ee">+51/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>UnsupportedOperationException.cs</strong><dd><code>Add UnsupportedOperationException class implementation</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/UnsupportedOperationException.cs

<li>Added new exception class for unsupported operations<br> <li> Implements standard exception constructors with message and inner <br>exception support<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14936/files#diff-a15cdd443799eb35aea44e10f887af3b50065095a06c3e33ff65c92c75586dcd">+50/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>WebDriver.cs</strong><dd><code>Update WebDriver error handling implementation</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

<li>Replaced <code>UnhandledError</code> with <code>UnknownError</code> in error handling<br> <li> Added handling for <code>UnknownError</code> and <code>UnsupportedOperation</code> exceptions<br> <li> Removed obsolete <code>UnhandledError</code> case<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14936/files#diff-ef1874c09b1dca0d62ea1f0ead685355944b9cb7ae63df60ad9b32abdf90c8dc">+7/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>WebDriverError.cs</strong><dd><code>Update WebDriver error mappings</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/WebDriverError.cs

<li>Updated error mappings to use <code>UnknownError</code> and <code>UnsupportedOperation</code> <br>instead of <code>UnhandledError</code><br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14936/files#diff-dd2b762723d0295d3aa610b1a284a5308c6ac8f5b2647348f22c259c7ee81df5">+2/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>WebDriverResult.cs</strong><dd><code>Update WebDriverResult enum definitions</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/WebDriverResult.cs

<li>Replaced <code>UnhandledError</code> enum value with <code>UnknownError</code><br> <li> Added new <code>UnsupportedOperation</code> enum value<br> <li> Updated documentation comments<br>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14936/files#diff-a45ca4893bc613b4d7fec3f0ec99f24b163c9d823499b196a1626a8c847d923b">+7/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

___


dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
```diff
@@ -326,7 +326,7 @@ private Response CreateResponse(HttpResponseInfo responseInfo)
                 }
                 else
                 {
-                    response.Status = WebDriverResult.UnhandledError;
+                    response.Status = WebDriverResult.UnknownError;
```
RenderMichael, line 329: This code is a best-effort attempt to give users a decent status.

Reading through [the spec](https://www.w3.org/TR/webdriver2/#dfn-unknown-error), it looks like `unknown error` is generally used as the "something went wrong" error code. It seems fitting to use it as a fallback here as well.

dotnet/src/webdriver/WebDriverResult.cs
```diff
@@ -90,9 +90,9 @@ public enum WebDriverResult
         InvalidElementState = 12,
 
         /// <summary>
-        /// An unhandled error occurred.
+        /// An unknown error occurred in the remote end while processing the command.
```
RenderMichael: Description taken from [spec](https://www.w3.org/TR/webdriver2/#dfn-unknown-error)

nvborisenko:
I guess we can move further reviewing it: at least 2 members think of it came from obsolete https://www.selenium.dev/documentation/legacy/json_wire_protocol, and python doesn't have this error code.

nvborisenko:
@RenderMichael I propose to expand this PR to align all actual error codes, not only `UnhandledError` if you can. Seems python state is more accurate: https://github.com/SeleniumHQ/selenium/blob/7854e62162d2f23d7fc0238ee4e8d6d54afca90e/py/selenium/webdriver/remote/errorhandler.py including custom exceptions.

dotnet/src/webdriver/WebDriverResult.cs
```diff
@@ -90,9 +90,9 @@ public enum WebDriverResult
...
         /// </summary>
-        UnhandledError = 13,
+        UnknownError = 13,
```
nvborisenko, line 103: Replacing `13` with new mimical enum member is a breaking change, and even removing value too. In any case it will be a breaking change. As I understand correctly we can choose the following strategies to deprecate:
- Deprecate old members with `[Obsolete]` attribute - preferred
- Just remove (probably nobody will be broken, but we are able to be safer)

Whatever way we choose, I am going to highlight a breaking change in commit message (just rename PR title).

nvborisenko: Interesting what if:
```csharp
[Obsolete]
UnhandledError = 13,
UnknownError = 13,
```
then we can avoid breaking change now, give a chance for potential affected users to adopt, and only then announce breaking change? Sounds like a plan.

RenderMichael: I took the suggestion of making the new error value an alias for the old one. As such, this is no longer a breaking change and I removed the label from the PR title

dotnet/src/webdriver/WebDriverError.cs
```diff
@@ -27,32 +27,16 @@ namespace OpenQA.Selenium
     /// </summary>
     internal static class WebDriverError
     {
-        /// <summary>
```
RenderMichael, line 30: No changes in this file - only re-ordering to align with spec, for readability

dotnet/src/webdriver/WebDriver.cs
```diff
@@ -854,6 +851,15 @@ private static void UnpackAndThrowOnError(Response errorResponse, string command
...
 
+                        case WebDriverResult.UnknownError:
+                            throw new UnknownErrorException(errorMessage);
+
+                        case WebDriverResult.UnknownMethod:
+                            throw new UnknownMethodException(errorMessage);
+
+                        case WebDriverResult.UnsupportedOperation:
+                            throw new UnsupportedOperationException(errorMessage);
+
```
RenderMichael, line 862: There are 3 remaining (in-spec) `WebDriverResult` values which do not have custom exceptions (every single other case does, so I assume it's desirable)

- [`NoSuchCookie`](https://www.w3.org/TR/webdriver2/#dfn-no-such-cookie) - as discussed in https://github.com/SeleniumHQ/selenium/issues/14884 this error code is never returned. The breaking change PR https://github.com/SeleniumHQ/selenium/pull/14883 does introduce the new exception type, but it will not be merged anytime soon. Would it be OK to add just the `NoSuchCookieException` in this PR instead?
- [`SessionNotCreated`](https://www.w3.org/TR/webdriver2/#dfn-session-not-created) - I tried to reproduce this based on what the [spec](https://www.w3.org/TR/webdriver2/#dfn-new-sessions) says should return this, and I got a spec violation! I set `capabilities` to `null` and got `invalid argument` instead of the spec-required `session not created`. I think maybe this error doesn't matter much.
- [`UnableToCaptureScreen`](https://www.w3.org/TR/webdriver2/#dfn-unable-to-capture-screen) - struggling to reproduce this. This may be an extremely edge case scenario.

dotnet/src/webdriver/WebDriver.cs
```diff
@@ -854,6 +851,15 @@ private static void UnpackAndThrowOnError(Response errorResponse, string command
...
                         default:
                             throw new InvalidOperationException(string.Format(CultureInfo.InvariantCulture, "{0} ({1})", errorMessage, errorResponse.Status));
```
RenderMichael, line 864: Is this exception reachable? If so, should it be `WebDriverException` instead of invalid operation exception?

dotnet/src/webdriver/WebDriverResult.cs
nvborisenko: The `public enum WebDriverResult` is public. I still think it is a breaking change. And what about:
- NoStringWrapper
- NoStringLength 
- NoString 
- NoCollection 
- IndexOutOfBounds 
I guess it should be deprecated.

dotnet/src/webdriver/WebDriver.cs
nvborisenko: > [NoSuchCookie](https://www.w3.org/TR/webdriver2/#dfn-no-such-cookie) - as discussed in https://github.com/SeleniumHQ/selenium/issues/14884 this error code is never returned. The breaking change PR https://github.com/SeleniumHQ/selenium/pull/14883 does introduce the new exception type, but it will not be merged anytime soon. Would it be OK to add just the NoSuchCookieException in this PR instead?

As I understand everybody is OK to use special endpoint to get a named cookie. Then we can do it later.

nvborisenko: > [SessionNotCreated](https://www.w3.org/TR/webdriver2/#dfn-session-not-created) - I tried to reproduce this based on what the [spec](https://www.w3.org/TR/webdriver2/#dfn-new-sessions) says should return this, and I got a spec violation! I set capabilities to null and got invalid argument instead of the spec-required session not created. I think maybe this error doesn't matter much.
[UnableToCaptureScreen](https://www.w3.org/TR/webdriver2/#dfn-unable-to-capture-screen) - struggling to reproduce this. This may be an extremely edge case scenario.

Let's skip, this PR is making selenium better, enough.

dotnet/src/webdriver/WebDriver.cs
nvborisenko: It might be, we cannot always map server-side error with whatever we can handle. Usually it is related to when we send some command and get result. From my perspective it is OK to throw some default exception, even if it is out of spec. Some providers, which are spec compatible, might not follow spec.

dotnet/src/webdriver/WebDriverResult.cs
RenderMichael: I took the suggestion you mentioned:
```csharp
[Obsolete]
UnhandledError = 13,
UnknownError = 13,
```

So anyone still using `UnhandledError` is not broken.

I went through and obsoleted all the result statuses that are no longer returned by W3C.

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
nvborisenko: We introduced `UnknownError` to represent server-side errors. In my opinion if StatusCode is not success, and something else is not expected in http response, then this is a problem of client.

Example: StatusCode = 400, ContentType doesn't start with "application/json". This situation cannot be mapped to server-side `UnknownError`. It is better to throw `WebDriverException` with message "ah I got unsuccessful status from remote end with code 400 and without any json details inside" (dummy, just as an example).

From other point of view I would like to not use `if (responseInfo.ContentType != null && responseInfo.ContentType.StartsWith(JsonMimeType, StringComparison.OrdinalIgnoreCase))` constraint and just let further execution. But again, we should handle "bad" exceptions and provide "good" to end users.

dotnet/src/webdriver/WebDriverResult.cs
RenderMichael: I made the obsoletion message `"This result is not returned by W3C specifications"`. Since anyone dealing with this enum is going to be an advanced user, we can be clear with them why this is obsolete now. If a different message is preferred, I can change it.

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
RenderMichael: That makes complete sense.

In this PR, I would like to replace `UnhandledError` with the better `UnknownError`. In some follow-up PR I can try to improve this branch of the response parsing.

nvborisenko: You are also right, let's stay on something small.

dotnet/src/webdriver/WebDriverResult.cs
nvborisenko: Very good, thank you! I also see strange `This result is not returned by W3C specifications`, and I am not native English speaker. We should understand why we deprecate it.

nvborisenko: `No more longer eligible with WebDriver Specification https://www.w3.org/TR/webdriver2/#errors` ?

RenderMichael: What about a message like `WebDriver no longer returns this error status`?

RenderMichael: A reference to the spec is good. How about then: `This error status is no longer returned by the WebDriver Specification https://www.w3.org/TR/webdriver2/#errors`?

nvborisenko: I trust your grammar, and I see all info what I expected as a end user.

RenderMichael: Message is updated (since we are inside `enum` it is not simple to make a `const string`, I hope inline is OK despite repetition)

dotnet/src/webdriver/WebDriverError.cs
```diff
@@ -218,14 +213,23 @@ static WebDriverError()
...
             resultMap[UnexpectedAlertOpen] = WebDriverResult.UnexpectedAlertOpen;
             resultMap[UnknownCommand] = WebDriverResult.UnknownCommand;
-            resultMap[UnknownError] = WebDriverResult.UnhandledError;
-            resultMap[UnknownMethod] = WebDriverResult.UnknownCommand;
-            resultMap[UnsupportedOperation] = WebDriverResult.UnhandledError;
+            resultMap[UnknownError] = WebDriverResult.UnknownError;
+            resultMap[UnknownMethod] = WebDriverResult.UnknownMethod;
+            resultMap[UnsupportedOperation] = WebDriverResult.UnsupportedOperation;
+
+            // TODO: Remove these strings; it is no longer valid in the specification.
```
nvborisenko: Just remove, it is internal implementation.

RenderMichael: Deleted and obsoleted public-facing error string

RenderMichael: `WebDriverError` is `internal` so I can remove the strings. Removed.

nvborisenko:
Seems are deprecating "aggressively", I mean it will become to hard to manage when exactly we should remove deprecated code. I propose to just to append a hint when exactly it will be deprecated. Usually, as I understand deprecation policy, it is +2 minor releases.

So instead of `[Obsolete(....)]` it will be better for everybody to see `[Obsolete(... Will be removed in 4.30)]` (where 4.27 is current).

dotnet/src/webdriver/WebDriverResult.cs
```diff
@@ -32,115 +34,134 @@ public enum WebDriverResult
...
         /// </summary>
+        [Obsolete("This error status is no longer returned by the WebDriver Specification https://www.w3.org/TR/webdriver2/#errors. Will be removed in 4.30")]
         NoStringWrapper = 5,
 
         /// <summary>
-        /// No driver matching the criteria exists.
+        /// Occurs if the given <see href="https://www.w3.org/TR/webdriver2/#dfn-session-id">session id</see> is not in the list of <see href="https://www.w3.org/TR/webdriver2/#dfn-active-sessions">active sessions</see>, meaning the session either does not exist or that it's not active.
         /// </summary>
-        NoSuchDriver = 6,
+        NoSuchDriver = IndexOutOfBounds,
```
RenderMichael:
```suggestion
        NoSuchDriver = 6,
```

nvborisenko (approved):

nvborisenko:
Thanks @RenderMichael for the contribution!

PR #14972, opened 2024-12-28
[dotnet] Tolerate invalid UTF-16 strings in DevTools JSON response

### LINKED ISSUE #14903: [🐛 Bug]: IWebDriver.Manage().Network.StartMonitoring() causes file uploads to fail
### What happened?

I have a project that uploads files into the web app under test on the Chrome browser. With Selenium 4.23 this was working flawlessly. I tried to update to Selenium 4.27 and with no changes other than this NuGet package version this stopped working. 

The implementation of the web app under test is such that when files are uploaded this is done in 4MB chunks. If the file size is less than 4MB then the upload still works as it used to with 4.23. However if the size is greater than 4MB such that it is chunked then the upload will not work anymore. 

Narrowing it down I eventually find that it is a call to IWebDriver.Manage().Network.StartMonitoring() that appears to  be the cause. If I remove this line it works fine, with it it will not do the upload. 

Here is a simple example to see the issue. Note that this is not the site I was testing, I cannot give you access to that, but this one appears to show the issue too. Note here I don't know the chunking size for this site but I was observing failure using a file that was 80MB and seeing success with a file that was 1KB.

```csharp

IWebDriver driver = null;
try
{
    driver = new ChromeDriver();
    driver.Manage().Network.StartMonitoring().GetAwaiter().GetResult();
    driver.Navigate().GoToUrl("https://tus.io/demo");
    Thread.Sleep(1000);
    driver.FindElement(By.Id("P0-0")).SendKeys("C:\\Users\\YourName\\file.file");
    Thread.Sleep(22000);
}
finally
{
    driver?.Quit();
}
```

### Relevant log output

Trace level driver logs are 200MB so cannot upload or paste in here. Please download from: https://www.filemail.com/d/xbkqrdjzsxgdvms

### Operating System

Windows 10

### Selenium version

4.27.0

Chrome 131.0.6778.140

ChromeDriver 131.0.6778.140

I was but seems not necessary to reproduce the issue

RenderMichael:
### **User description**

### Description

This restores Newtonsoft behavior of tolerating invalid UTF-16 strings.

### Motivation and Context

Fixes #14903

### Types of changes

### Checklist


dotnet/src/webdriver/DevTools/DevToolsSession.cs
```diff
@@ -600,7 +600,19 @@ private void ProcessMessage(string message)
...
+                    var loggableEventData = eventData.DeepClone();
+                    var loggableRequest = loggableEventData["request"]!;
+                    loggableRequest["postData"] = "*RAW POST DATA REMOVED FROM LOGS*";
+                    loggableRequest["postDataEntries"] = new JsonArray();
+                    LogTrace("Recieved Event {0}: {1}", method, loggableEventData.ToString());
+                }
+                else
+                {
+                    LogTrace("Recieved Event {0}: {1}", method, eventData.ToString());
+                }
```
RenderMichael: Fundamentally, this error happens because the string resembles JSON, but is in fact invalid JSON 
image

dotnet/src/webdriver/DevTools/DevToolsSession.cs
```diff
@@ -600,7 +600,19 @@ private void ProcessMessage(string message)
                 var methodParts = method.Split(new char[] { '.' }, 2);
                 var eventData = messageObject["params"];
 
-                LogTrace("Recieved Event {0}: {1}", method, eventData.ToString());
```
RenderMichael: This is not a complete solution, the JSON exception is simply thrown later: 
image

The above is in generated cdp code (`\cdp\v131\Fetch\FetchAdapter.cs` in the above screenshot).

Related relevant code:
image

RenderMichael: @nvborisenko You mentioned trying out `JavaScriptEncoder.UnsafeRelaxedJsonEscaping`. I cannot edit the generated code directly but I added a spoof here, and it did not fix the exception
image

RenderMichael:
I'm not sure if this is considered a full fix, it just moves the exception down
Before: 
image
After:
image

dotnet/src/webdriver/DevTools/DevToolsSession.cs
nvborisenko: I modified generator to use `JavaScriptEncoder.UnsafeRelaxedJsonEscaping`, it didn't help. The worst scenario is that it works with `Newtonsoft.Json` (Selenium v4.23?).

RenderMichael: I do not understand how Newtonsoft accepts this nonsense input. It’s not real JSON, just some raw binary data.

Maybe we should intersept the post data and sanitize it ourselves?

nvborisenko: One more observation: in json response I see binary data as is, probably not even encoded! This is why VS Code shows json improperly. CDP specification doesn't say it should be encoded (base64). I think this is why they deprecated it.

PS: But why `Newtonsoft.Json` could parse it successfully? I think it is issue in this library. I didn't check it yet, but I guess:

1 using old selenium version (v4.23?) when we upload binary file
2 and we download it manually back
3 then the content of "original" and "downloaded" files is different

RenderMichael: I tried switching from `JsonNode` to `JsonElement` just to see what would happen.

Good news, no exceptions on `ToString()` like the logging has, but the de-serialization still fails
image

Maybe we should parse and encode the postData ourselves?

nvborisenko: I am stupid for this activity. It is easier to:
- really `Newtonsoft.Json` parses it correctly?

RenderMichael: I narrowed down the issue. It looks like this character [`U+DB40`](https://unicodeplus.com/U+DB40) appears here for some reason. It is the first part of a UTF-16 surrogate pair, but the second half is missing.
image

I opened the raw `Firefox Installer.exe` file and found this segment here:
image

Note how in this case, it is just a regular [`I`](https://unicodeplus.com/U+0049).

What is happening here? The CDP is giving us back broken JSON, this may be a bug on their end.

I will investigate whether it is possible to work around this somehow.

RenderMichael: I found a workaround, with a custom converter.

We can add to the postData property:

image

As well as to `PostDataEntry.Bytes`:
image

The following converter:

```csharp
internal sealed class InvalidUtf16Converter : JsonConverter<string>
{
    public override bool HandleNull => true;

    public override string? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
    {
        try
        {
            return reader.GetString();
        }
        catch (InvalidOperationException)
        {
            var bytes = reader.ValueSpan;
            var sb = new StringBuilder(bytes.Length);
            foreach (var b in bytes)
                sb.Append(Convert.ToChar(b));
            return sb.ToString();
        }
    }

    public override void Write(Utf8JsonWriter writer, string value, JsonSerializerOptions options) =>
        writer.WriteStringValue(value);
}
```

Since we have a good idea of which models may have arbitrary data, 

However... it only works with `JsonElement.Deserialize<T>`. When using a `JsonNode.Deserialize<T>`, the exception happens way before even the converter is reached. Even `JsonNode.ToString()` throws an exception!

For this reason, we may need to change various `JsonNode`s to `JsonElement`s, such as `DevToolsEventReceivedEventArgs.EventData` (that property is used everywhere in the cdp-generated code, but luckily the line `e.EventData.Deserialize(eventData.EventArgsType)` works exactly the same (it just hits a different overload). Also in `DevToolsCommandData.Result`

I am not familiar enough with the cdp code generation process to make generator changes. Is there any docs for this?

nvborisenko: Most likely you are interested in https://github.com/SeleniumHQ/selenium/tree/trunk/third_party/dotnet/devtools/src/generator/Templates

There is no docs, generally saying build process invokes generation of CDP classes based on `*.cdl` (I don't remember exactly, it is a definition of contract) files applying `*.hbs` templates (https://handlebarsjs.com/).

RenderMichael:
Repro test:

```csharp
[Test]
public void AAA()
{
    driver.Manage().Network.StartMonitoring().GetAwaiter().GetResult(); // this enables serialization
    driver.Navigate().GoToUrl("https://tus.io/demo");

    Thread.Sleep(10_000);
    driver.FindElement(By.Id("P0-0")).SendKeys("C:\\Users\\Nick\\Downloads\\Firefox Installer.exe");

    Thread.Sleep(60_000);
}
```

Passes both on Firefox and Chrome.

third_party/dotnet/devtools/src/generator/Templates/DevToolsSessionDomains.hbs
```diff
@@ -12,6 +16,15 @@ namespace {{rootNamespace}}
         private Lazy<{{dehumanize Name}}.{{dehumanize Name}}Adapter> m_{{dehumanize Name}};
 {{/each}}
 
+        [global::System.ComponentModel.EditorBrowsableAttribute(global::System.ComponentModel.EditorBrowsableState.Never)] // Generated code use only
+        internal static global::System.Text.Json.JsonSerializerOptions DevToolsSerializerOptions { get; } = new global::System.Text.Json.JsonSerializerOptions()
```
nvborisenko: Better to put it into solution under git rather that generate.

RenderMichael: Fixed, this turned out very nice and clean!

dotnet/src/webdriver/DevTools/DevToolsSessionDomains.cs
```diff
@@ -43,5 +60,29 @@ protected DevToolsSessionDomains()
         /// Populates the command response type map.
         /// </summary>
         protected abstract void PopulateCommandResponseTypeMap();
+
+        private sealed class InvalidUtf16Converter : JsonConverter<string>
```
nvborisenko: We have `DevTools/Json` subfolder - great candidate to put this class in.

nvborisenko: I would like to rename it to `StringConverter`.

third_party/dotnet/devtools/src/generator/Templates/domain.hbs
```diff
@@ -61,7 +61,7 @@ namespace {{rootNamespace}}.{{domain.Name}}
                 if (m_eventMap.ContainsKey(e.EventName))
                 {
                     var eventData = m_eventMap[e.EventName];
-                    var eventArgs = e.EventData.Deserialize(eventData.EventArgsType);
+                    var eventArgs = e.EventData.Deserialize(eventData.EventArgsType, OpenQA.Selenium.DevTools.DevToolsSessionDomains.DevToolsSerializerOptions);
```
nvborisenko:
```suggestion
                    var eventArgs = e.EventData.Deserialize(eventData.EventArgsType, global::OpenQA.Selenium.DevTools.DevToolsSessionDomains.DevToolsSerializerOptions);
```

dotnet/src/webdriver/DevTools/DevToolsSessionDomains.cs
RenderMichael: Sure. That will clash with `System.ComponentModel.StringConverter`, so I'll have to qualify the name.

dotnet/src/webdriver/DevTools/DevToolsSessionDomains.cs
```diff
@@ -26,6 +34,15 @@ public abstract class DevToolsSessionDomains
     {
         private CommandResponseTypeMap responseTypeMap = new CommandResponseTypeMap();
 
+        [EditorBrowsable(EditorBrowsableState.Never)] // Generated code use only
+        internal static JsonSerializerOptions DevToolsSerializerOptions { get; } = new JsonSerializerOptions()
```
nvborisenko: We should move this to somewhere else. Not sure where to. Probably to `DevToolsSession.cs`. Or even better to `Json` subfolder.

RenderMichael: Moved to new type `OpenQA.Selenium.DevTools.Json.DevToolsJsonOptions`

dotnet/src/webdriver/DevTools/Json/StringConverter.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    public override string? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+    {
+        try
+        {
+            return reader.GetString();
+        }
+        catch (InvalidOperationException)
+        {
+            // Backwards compatibility with Newtonsoft tolerating invalid UTF-16 sequences
```
nvborisenko: Let us improve this comment:
- don't mention Newtonsoft
- don't use "sometimes" word - we always should know what is happening

dotnet/src/webdriver/DevTools/Json/StringConverter.cs
```diff
@@ -0,0 +1,35 @@
...
+            // CDP sometimes sends invalid surrogate pairs on file upload
+
+            var bytes = reader.ValueSpan;
+            var sb = new StringBuilder(bytes.Length);
+            foreach (byte b in bytes)
+                sb.Append(Convert.ToChar(b));
```
nvborisenko: I remember we don't have code style rules but please!
```csharp
foraech(...)
{

}
```

dotnet/src/webdriver/DevTools/Json/DevToolsJsonOptions.cs
```diff
@@ -0,0 +1,18 @@
+using System.ComponentModel;
+using System.Text.Json;
+
+#nullable enable
+
+namespace OpenQA.Selenium.DevTools.Json;
+
+[EditorBrowsable(EditorBrowsableState.Never)] // Generated code use only
```
nvborisenko: Unnecessary.

dotnet/src/webdriver/DevTools/Json/StringConverter.cs
RenderMichael: Removing mention to Newtonsoft.

I would also like to know why this is happening! I don't know what is causing this.

dotnet/src/webdriver/DevTools/Json/DevToolsJsonOptions.cs
```diff
@@ -0,0 +1,16 @@
+using System.Text.Json;
+
+#nullable enable
+
+namespace OpenQA.Selenium.DevTools.Json;
+
+internal static class DevToolsJsonOptions
+{
+    public static JsonSerializerOptions DevToolsSerializerOptions { get; } = new JsonSerializerOptions()
```
nvborisenko: Rename to `Default` ?

RenderMichael: Done.

dotnet/src/webdriver/DevTools/Json/StringConverter.cs
nvborisenko:
```
Fallback to read the value as bytes instead of string.
System.Text.Json library throws exception when CDP remote end sends non-encoded string as binary data.
Using JavaScriptEncoder.UnsafeRelaxedJsonEscaping doesn't help because the string actually is byte[].
https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-Request - here "postData" property
is a string, which we cannot deserialize properly. This property is marked as deprecated, and new "postDataEntries"
is suggested for using, where most likely it is base64 encoded.
```

RenderMichael: Since this solution is so targeted, do you think we should try to target `Network.Request.postData` specifically here?

nvborisenko: No, we cannot manage the source for generator, it is provided by Google.
In this comment we should leave all knowledge we know about why this "strange" workaround lives.

nvborisenko: Ah, and don't forget about mentioning "missing surrogate pair".

RenderMichael: What I mean is, can we add to the generator "if property name is postData, add JSON converter attribute"

nvborisenko: I don't think so, it will add complexity and will lead to even more "strange workaround". As soon as Google will remove deprecated property, we will "easily" remove this workaround.

nvborisenko: But from other hand why not to be specific? If we will generate attribute conditionally, then will it be friendly with upcoming AOT?

nvborisenko: No, adding a complexity to stuff which will be deprecated is not good idea.
- Chromium removes "bad property" (`postData`) - we "remove" workaround
- CDP is supposed to be replaced by BiDi

RenderMichael: Speaking of removing complexity, wouldn't it be cool to convert the handlebars generator into a modern incremental source generator? That sounds like a fun little project.

RenderMichael: comment added

RenderMichael:
That's my bad for forgetting the file header. Fixed.

nvborisenko (approved):
Thanks @RenderMichael !

dotnet/src/webdriver/DevTools/Json/StringConverter.cs
nvborisenko: It will require reworking entire bazel build process. A lot of efforts, but you can try )

PR #14986, opened 2024-12-30
[grid] retry if no node does support the Capabilities

joerg1985:
### Description
This PR will allow to retry if currently no node does support the Capabilities, this can change over time when a new node registers.

### Motivation and Context
When using `--drain-after-session-count` and a single node which is automatically restarted after shutdown, this PR will allow the pending session request to retry until the node is restarted and healthy again.

### Types of changes

### Checklist

___

### **PR Type**
Bug fix

___

### **Description**
- Implements retry mechanism in LocalDistributor when no nodes support requested capabilities
- Previously, the system would fail immediately if no node supported the capabilities
- Now it will retry, allowing time for new nodes to register that may support the requested capabilities
- Particularly useful when using `--drain-after-session-count` feature with node restarts

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Error handling</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>LocalDistributor.java</strong><dd><code>Add retry mechanism for unsupported capabilities</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

java/src/org/openqa/selenium/grid/distributor/local/LocalDistributor.java

<li>Added retry logic when no node supports the requested capabilities<br> <li> Updated error message to indicate waiting for node registration

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14986/files#diff-024d7dccb0a4ff87c444dc3549032e7c6b2595f582f4f267cdf48c7f8c2f87c1">+5/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

___


joerg1985:
@VietND96 is this a common case when using Kubernetes, when only one node is used?

diemol (commented):
The nodes are not restarted after shutdown. Plus, the `rejectUnsupportedCaps` configurable value keeps the request in the queue for the configured time. I think this scenario is already covered.

joerg1985:
Some more words to the scenario, the node process is external monitored and restarted after shutdown, e.g. via container heath check.

The `rejectUnsupportedCaps` mechanism does not work here, at the time the `NewSessionRunnable` is executed the nodes might be different than later on when session is about to be started. So the result of `isNotSupported` will flip from `false` to `true` and we run into this `if`. I guess this is the reason for having the `if` statement here.

I have stress test for the scenario, but it is currently failing due to this issue and #14987. It might be best to add this test to the PR as soon as #14987 is merged, to reproduce it and than look at it.

joerg1985:
@diemol i have created the test `org.openqa.selenium.grid.distributor.DrainTest.sessionIsNotRejectedWhenNodeDrains` to provoke it. The test will have two sessions in the queue and a single node with `drain-after-session-count 1` is started in a loop. This will sooner or later (on my system 1 to 6 iterations) run into the issue.

As soon as the `retry = true` line added in this pr is removed the test will fail with: `java.util.concurrent.ExecutionException: org.openqa.selenium.SessionNotCreatedException: Could not start a new session. Response code 500. Message: Could not start a new session. Unable to find a node supporting the desired capabilities`

PS: the `/readyz` endpoint has been changed to return 200, this will allow to use the `UrlChecker` and have a consistent behaviour compared to the hub.

VietND96:
> @VietND96 is this a common case when using Kubernetes, when only one node is used?

I can confirm `--drain-after-session-count` is a use case in K8s. In K8s, Node is deployed as Job or Deployment, with behavior as below.
- Deployment: assume that `--drain-after-session-count=1`. In this type, it always go to the situation `automatically restarted after shutdown`, since there is `restartPolicy: Always`, which means the container will be restarted internally after the last run has exited (the distributor considers it is a shutdown). Whenever it is up and registered to Hub, the distributor probably considers it a NodeRestarted event since it has the same URL and different UUID.
- Job: in this type, it could be often go to the situation `automatically restarted after shutdown`. Since after the container exited, the high-level Pod terminated. A new Node running in a new Pod will be spawned (which is assigned a new IP from pool, or sometime due to the IPs pool limits the same IP could be randomly assigned). It would be possibility of 50% node shutdown and 50% node restarted after shutdown.

VietND96:
For example, a node `automatically restarted after shutdown` which has no support Caps always, might it cause the infinite loop or long delays?

joerg1985:
> For example, a node `automatically restarted after shutdown` which has no support Caps always, might it cause the infinite loop or long delays?

The session request should be rejected after `--session-request-timeout` has reached, if no node matching the capabilities is found in time.

diemol (approved):
Feel free to merge after CI is green.

Issue #13803, opened 2024-04-11
[🐛 Bug]: RemoteWebDriver - Implementation mismatch since version 4.14.0


johnny:
### What happened?

Since https://github.com/SeleniumHQ/selenium/issues/13802 was closed and cannot be reopened, I needed to create a new ticket.

As a user, I expect urls with embedded authentication to work in Java like in any other language.

http://admin:myStrongPassword@localhost:4444

### How can we reproduce the issue?

```shell
https://github.com/SeleniumHQ/selenium/issues/13802
```

### Relevant log output

```shell
https://github.com/SeleniumHQ/selenium/issues/13802
```

### Operating System

macOS

### Selenium version

Java 4.14.0-4.18.1

### What are the browser(s) and version(s) where you see this issue?

All

### What are the browser driver(s) and version(s) where you see this issue?

RemoteWebDriver

### Are you using Selenium Grid?

4.18.1

diemol:
I understand, but as I mentioned in the other issue, due to changes in the HTTP client, that is how you can connect to authenticated services.

johnny:
I understand that I can write more code when constructing the RemoteWebDriver. In our case, that wouldn't work though, since we are using fluentlenium.

There is already code in the JdkHttpClient implementation that tries to handle embedded userinfo, but it's broken.
https://github.com/SeleniumHQ/selenium/blob/2b5999625744adccf2a0ad30cdd18d503044603b/java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java#L107-L120

A fix is easy (drop the user from the url in above linked code) and then Java behaves the same again as other languages.

diemol:
Would you like to send a PR?

joerg1985:
@johnny are you using the selenium server configuration to enable basic auth or some kind of reverse proxy?

Delta456:
> A fix is easy (drop the user from the url in above linked code) and then Java behaves the same again as other languages.

I looked all over and I cannot find anything related to the URL.

Issue #13802, opened 2024-04-11
[🐛 Bug]: RemoteWebDriver: Basic Auth URLs broken in Java client since 4.14.0


johnny:
### What happened?

When using the remote web driver with a url that contains the basic auth credentials, the session cannot be created. See the log. Selenium grid was installed using the helm chart.
This is broken since 4.14.0 and we confirmed it's still broken until 4.18.1.

This is due to the behavior of the HTTP client described here: https://stackoverflow.com/questions/54208945/java-11-httpclient-not-sending-basic-authentication

It doesn't send the basic auth information without the server explicitely asking for basic auth.

We worked around the issue by setting proper client options in a custom webdriver:

```
 private static CommandExecutor createExecutor(URL remoteAddress, boolean enableTracing) {
        ClientConfig config = ClientConfig.defaultConfig().baseUrl(remoteAddress);
        String info = remoteAddress.getUserInfo();
        if (info != null && !info.trim().isEmpty()) {
            String[] parts = info.split(":", 2);
            String username = parts[0];
            String password = parts.length > 1 ? parts[1] : null;

            try {
                config = config.baseUrl(new URL(remoteAddress.getProtocol(), remoteAddress.getHost(), remoteAddress.getPort(), remoteAddress.getFile()))
                    .authenticateAs(new UsernameAndPassword(username, Objects.requireNonNull(password)));
            } catch (MalformedURLException e) {
                throw new RuntimeException(e);
            }
        } else {
            config = config.baseUrl(remoteAddress);
        }
        if (enableTracing) {
            Tracer tracer = OpenTelemetryTracer.getInstance();
            CommandExecutor executor =
                new HttpCommandExecutor(
                    Collections.emptyMap(),
                    config,
                    new TracedHttpClient.Factory(tracer, HttpClient.Factory.createDefault()));
            return new TracedCommandExecutor(executor, tracer);
        } else {
            return new HttpCommandExecutor(config);
        }
    }
```

### How can we reproduce the issue?

```shell
capabilities.setCapability("browserName", "chrome");
            capabilities.setCapability("uilicious:config", config);

            WebDriver driver;

            URL url = new URL("https://foo:bar@example.com");
            driver = new RemoteWebDriver(url, capabilities);

We used https://github.com/uilicious/selenium-java-example and upgraded selenium to the latest version.
```

### Relevant log output

```shell
org.openqa.selenium.SessionNotCreatedException: Could not start a new session. Unable to parse remote response: <html>
[...]

PR #15071, opened 2025-01-13
[java] Fixed issue with embedded authentication in URLs for JdkHttpClient

iampopovich:
### **User description**

### Description

This pull request includes several changes to the `JdkHttpClient` class to handle and test the stripping of credentials from URLs. The most significant changes include modifications to the `JdkHttpClient` class, the addition of new methods and fields, and the introduction of new tests to ensure the correct functionality.

### Updates to `JdkHttpClient` class:

* Added a `ClientConfig` field and initialized it in the constructor to store configuration settings.
* Introduced logic to strip credentials from the base URI in the `getPasswordAuthentication` method.
* Added a package-private method `getBaseUri` for testing purposes to retrieve the base URI from the configuration.
* Enhanced the `send` method to include the cause in the `WebDriverException` when interrupted.

### New tests in `JdkHttpClientTest` class:

* Added tests to verify that credentials are stripped from the URL and that other URL components are preserved. [[1]](diffhunk://#diff-2dafb043b33952f6ce060100ec76740334f47cd85b850dcfee0dcba7c4622156R20-R25) [[2]](diffhunk://#diff-2dafb043b33952f6ce060100ec76740334f47cd85b850dcfee0dcba7c4622156R35-R79)

### Motivation and Context

Fixes https://github.com/SeleniumHQ/selenium/issues/13803

### Types of changes

### Checklist

___

### **PR Type**
Bug fix, Tests

___

### **Description**
- Fixed issue with embedded authentication in URLs for `JdkHttpClient`.

- Added logic to strip credentials from base URI in `JdkHttpClient`.

- Introduced new tests to validate credential stripping and URL component preservation.

- Enhanced exception handling in `send` method to include cause details.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>JdkHttpClient.java</strong><dd><code>Implement credential stripping and enhance exception handling</code></dd></summary>
<hr>

java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java

<li>Added logic to strip credentials from base URI.<br> <li> Introduced a package-private method <code>getBaseUri</code> for testing.<br> <li> Enhanced exception handling in the <code>send</code> method.<br> <li> Updated <code>ClientConfig</code> initialization to handle credential-free URIs.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15071/files#diff-8c4bb26a4a1da292af9347bac545d0a98344057ae0373642be830b5744f828aa">+27/-1</a>&nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>JdkHttpClientTest.java</strong><dd><code>Add tests for URL credential stripping and validation</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

java/test/org/openqa/selenium/remote/http/jdk/JdkHttpClientTest.java

<li>Added tests to verify credential stripping from URLs.<br> <li> Validated preservation of other URL components.<br> <li> Tested handling of URLs without credentials.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15071/files#diff-2dafb043b33952f6ce060100ec76740334f47cd85b850dcfee0dcba7c4622156">+51/-0</a>&nbsp; &nbsp; </td>

</tr>

___

### DIFF
--- a/java/src/org/openqa/selenium/chromium/AddHasCdp.java
+++ b/java/src/org/openqa/selenium/chromium/AddHasCdp.java
@@ -47,20 +47,15 @@ public abstract class AddHasCdp implements AugmenterProvider<HasCdp>, Additional
 
   @Override
   public HasCdp getImplementation(Capabilities capabilities, ExecuteMethod executeMethod) {
-    return new HasCdp() {
-      @Override
-      public Map<String, Object> executeCdpCommand(
-          String commandName, Map<String, Object> parameters) {
-        Require.nonNull("Command name", commandName);
-        Require.nonNull("Parameters", parameters);
+    return (commandName, parameters) -> {
+      Require.nonNull("Command name", commandName);
+      Require.nonNull("Parameters", parameters);
 
-        Map<String, Object> toReturn =
-            (Map<String, Object>)
-                executeMethod.execute(
-                    EXECUTE_CDP, Map.of("cmd", commandName, "params", parameters));
+      Map<String, Object> toReturn =
+          (Map<String, Object>)
+              executeMethod.execute(EXECUTE_CDP, Map.of("cmd", commandName, "params", parameters));
 
-        return Map.copyOf(toReturn);
-      }
+      return Map.copyOf(toReturn);
     };
   }
 }
--- a/java/src/org/openqa/selenium/chromium/AddHasLaunchApp.java
+++ b/java/src/org/openqa/selenium/chromium/AddHasLaunchApp.java
@@ -57,13 +57,9 @@ public class AddHasLaunchApp implements AugmenterProvider<HasLaunchApp>, Additio
 
   @Override
   public HasLaunchApp getImplementation(Capabilities capabilities, ExecuteMethod executeMethod) {
-    return new HasLaunchApp() {
-      @Override
-      public void launchApp(String id) {
-        Require.nonNull("id of Chromium App", id);
-
-        executeMethod.execute(LAUNCH_APP, Map.of("id", id));
-      }
+    return id -> {
+      Require.nonNull("id of Chromium App", id);
+      executeMethod.execute(LAUNCH_APP, Map.of("id", id));
     };
   }
 }
--- a/java/src/org/openqa/selenium/chromium/AddHasPermissions.java
+++ b/java/src/org/openqa/selenium/chromium/AddHasPermissions.java
@@ -57,15 +57,11 @@ public class AddHasPermissions
 
   @Override
   public HasPermissions getImplementation(Capabilities capabilities, ExecuteMethod executeMethod) {
-    return new HasPermissions() {
-      @Override
-      public void setPermission(String name, String value) {
-        Require.nonNull("Permission name", name);
-        Require.nonNull("Permission value", value);
-
-        executeMethod.execute(
-            SET_PERMISSION, Map.of("descriptor", Map.of("name", name), "state", value));
-      }
+    return (name, value) -> {
+      Require.nonNull("Permission name", name);
+      Require.nonNull("Permission value", value);
+      executeMethod.execute(
+          SET_PERMISSION, Map.of("descriptor", Map.of("name", name), "state", value));
     };
   }
 }
--- a/java/src/org/openqa/selenium/safari/AddHasDebugger.java
+++ b/java/src/org/openqa/selenium/safari/AddHasDebugger.java
@@ -55,11 +55,6 @@ public class AddHasDebugger implements AugmenterProvider<HasDebugger>, Additiona
 
   @Override
   public HasDebugger getImplementation(Capabilities capabilities, ExecuteMethod executeMethod) {
-    return new HasDebugger() {
-      @Override
-      public void attachDebugger() {
-        executeMethod.execute(ATTACH_DEBUGGER, null);
-      }
-    };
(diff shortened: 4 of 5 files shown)


java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java
```diff
@@ -108,6 +110,8 @@ public class JdkHttpClient implements HttpClient {
 
     Credentials credentials = config.credentials();
     String info = config.baseUri().getUserInfo();
+    URI baseUri = config.baseUri();
```
VietND96: I see this variable is never used.

java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java
```diff
@@ -121,6 +125,23 @@ protected PasswordAuthentication getPasswordAuthentication() {
             }
           };
       builder = builder.authenticator(authenticator);
+
+      // Remove credentials from URL
+      try {
+        this.config =
+            ClientConfig.defaultConfig()
```
VietND96: From the beginning, `this.config = config`. I don't think here it should attempt to create a new ClientConfig with defaultConfig()

VietND96 (approved):

java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java
iampopovich: I think this field is minor and can be removed.
I added it to keep a consistent code style, similar to other fields from config, for example:
https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/remote/http/jdk/JdkHttpClient.java#L87-L91

However, I didn’t update the calls from config.baseUri to this.baseUri anywhere.

VietND96: Don't worry, I addressed few things myself, can you check once?

joerg1985:
@VietND96 Please do not merge this, in my mind this will not fix the issue.

diemol (commented):
@iampopovich have you tested this locally?

iampopovich:
@diemol sure
I also provided new tests for the bugfix, you can try run it locally
Unfortunately i am away from pc, so i can attach test results later today or tomorrow

diemol:
Just wondering why @joerg1985 thinks this does not fix the issue.

joerg1985:
@diemol i will write later this evening details about this and the original issue.

joerg1985:
This PR does remove the UserInfo from the URI, this is okay, but it will not fix the issue #13803 from my point of view.

I have asked inside the ticket #13803 for details about the setup, but never did get a response. In #13802 you can see a nginx response with 404, so my current assumption is the authentication is done by a reverse proxy not following the RFC for WWW-Authenticate. The Java `Authenticator` will work only in case the server does send a 401 + challenge. https://datatracker.ietf.org/doc/html/rfc7235#section-4.1

We would need to translate the UserInfo from the URI to a 'Authorization' header and send it blind to the server. This might end in other issues, e.g. we do not know the expected scheme and might send the plain credentials (Basic scheme) instead of a hashed one (Digest scheme) over a unencrypted connection.

The nginx config might be an attemp to hide the service from attackers by forcing the clients to send the 'Authorization' header without a challenge. Some of the selenium language bindings might do so by default and IF it should be supported, this could be implemented in Java too. But this should be done via opt-in and not by default to avoid unexpected issues. And we should know the usecase so we can document this, to avoid more magic lines in the code nobody is aware of the use.

I would propalby hide the service using a UUID prefix in the path, the server has a command line flag to add a prefix...

VietND96:
@iampopovich, without this fix, for example a simple get started `driver = new RemoteWebDriver("http://admin:password@localhost:4444/wd/hub", options);`, does it work as of now?

VietND96:
> @iampopovich, without this fix, for example a simple get started `driver = new RemoteWebDriver("http://admin:password@localhost:4444/wd/hub", options);`, does it work as of now?

Actually, this kind of embedded auth URL works in Python binding. In background, it also extract `user:pass` from url then Base64 encode it and set to request header `Authorization`. You can inspect code at https://github.com/SeleniumHQ/selenium/blob/7e194df3145ce12ecce931237447a9cb33da751c/py/selenium/webdriver/remote/remote_connection.py#L419
Probably, it is correct with @joerg1985 on the approach that should do.
In JdkHttpClient, how to insert the `Authorization` to request headers, I think you have to check it.

VietND96:
> @iampopovich, without this fix, for example a simple get started `driver = new RemoteWebDriver("http://admin:password@localhost:4444/wd/hub", options);`, does it work as of now?

Looks like it works without this change, since I added a test to example site https://github.com/SeleniumHQ/seleniumhq.github.io/pull/2131

iampopovich:
@VietND96 What should we do next? Clearly, this PR does not fix the bug. I tried implementing authorization similar to the Python bindings client, but it didn’t solve the issue. At the moment, I don’t have enough knowledge to propose an alternative solution. 

We can close this pull request for now, and perhaps I’ll revisit the problem later if it’s still relevant.

PR #15179, opened 2025-01-28
[grid] Add feature to go directly to a VNC session

### LINKED ISSUE #15178: [🚀 Feature]: Direct Link to Selenium Grid VNC Session
### Feature and motivation

When a session has been started on the grid there should be a way to get an URL to go to the session immediately like `https://example.com/ui/#/session/SESSION_ID/vnc`

### Usage example

A session is started and the service that started the session prints the URL to the VNC session. If a developer want to see that the session is working as intended the developer can simply click/copy the link instead of going to the grid and finding the correct session.

emilhem:
### Motivation and Context
Makes it easy to link directly to a running session.

Users can now after retrieving the session id go to `https://grid.example.com/ui/#/session/SESSION_ID` to directly launch a VNC session.

Fixes: #15178 

### Types of changes

### Checklist


javascript/grid-ui/src/screens/Sessions/Sessions.tsx
```diff
@@ -26,12 +26,25 @@ import Loading from '../../components/Loading/Loading'
...
   })
+  const navigate = useNavigate()
+
+  const { sessionId } = useParams<{ sessionId: string }>()
+
+  useEffect(() => {
+    if (data === undefined || data.sessionsInfo === undefined || data.sessionsInfo.sessions === undefined) {
+      return
+    }
+    if ((sessionId !== undefined || sessionId === '') && data.sessionsInfo.sessions.length === 0) {
```
diemol: I wonder why a navigate is needed instead of a return here. 

Also, can you please check the [suggestions from the PR review tool](https://github.com/SeleniumHQ/selenium/pull/15179#issuecomment-2619418000)? At least the first one is valid.

diemol (changes requested):

javascript/grid-ui/src/screens/Sessions/Sessions.tsx
emilhem: The reason why there's a `navigate` there is to "redirect" to `/sessions` if the URL is something like `/session/SESSIONID` and there is no other sessions running and the current session closes. It also redirects if the user navigates to `/session/SESSIONID_THAT_IS_NO_LONGER_RUNNING`.

diemol (approved):

diemol (approved):
Thank you, @emilhem!

PR #15217, opened 2025-02-02
[dotnet] Trim away CDP when publishing AOT apps

RenderMichael:
### Description
Removes unnecessary AOT warnings from the user's eyes, and trims away a lot of code unless it is directly used.

### Motivation and Context
Size and warning savings

Sample app:

```C#
Console.WriteLine("Start");

Log.SetLevel(LogEventLevel.Trace);

var options = new ChromeOptions { UseWebSocketUrl = true, BrowserVersion = "131" };

using var driver = new ChromeDriver(options);

driver.Url = "https://www.google.com";

Thread.Sleep(2500);
driver.FindElements(By.CssSelector("div"));
```

Size before PR: 8.9 MB
image

+ 460 AOT warnings
image

Size after PR: 6.5 MB
image

+ 2 warnings
image

Size savings: 2.3 MB
image

### Types of changes

### Checklist

___

### **PR Type**
Enhancement

___

### **Description**
- Optimized the `WebDriver` class to reduce application size.

- Deferred initialization of `NetworkManager` to save resources.

- Removed unused or redundant code for cleaner implementation.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>WebDriver.cs</strong><dd><code>Optimize and streamline `WebDriver` initialization</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

<li>Removed direct initialization of <code>NetworkManager</code>.<br> <li> Deferred <code>NetworkManager</code> initialization using a null-coalescing <br>assignment.<br> <li> Deleted unused <code>Network</code> property setter.<br> <li> Cleaned up redundant code for better maintainability.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15217/files#diff-ef1874c09b1dca0d62ea1f0ead685355944b9cb7ae63df60ad9b32abdf90c8dc">+1/-5</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

___


dotnet/src/webdriver/WebDriver.cs
```diff
@@ -216,10 +215,7 @@ public virtual IFileDetector FileDetector
             }
         }
 
-        internal INetwork Network
-        {
-            get { return this.network; }
-        }
+        internal INetwork Network => this.network ??= new NetworkManager(this);
```
RenderMichael, line 200: `NetworkManager` captures `IDevTools.GetDevToolsSession` in the constructor, which pulls in many CDP types.

We do not want to touch `NetworkManager` unless user calls `driver.Network`.

nvborisenko: Good catch!

nvborisenko (approved):
Nice done!

RenderMichael:
Failures not related to changes

```bazel
//javascript/node/selenium-webdriver:test-bidi-browsingcontext-inspector-test.js-chrome
//javascript/node/selenium-webdriver:test-bidi-browsingcontext-inspector-test.js-firefox
//py:common-chrome-bidi-test/selenium/webdriver/common/bidi_tests.py
//rb/spec/integration/selenium/webdriver:action_builder-chrome-remote
//rb/spec/integration/selenium/webdriver:action_builder-firefox-remote
//rb/spec/integration/selenium/webdriver:driver-chrome
//rb/spec/integration/selenium/webdriver:driver-chrome-remote
//rb/spec/integration/selenium/webdriver:select-chrome-remote
```

nvborisenko:
@RenderMichael  Big brain is right about thread-safe: if we apply `Lazy<T>` or simple `lock` - is would be AOT friendly?

RenderMichael:
@nvborisenko Unfortunately Lazy<T> is not too AOT-friendly, anything captured in the factory is stuck. It doesn’t matter too much usually (it’s usually not too complex, the Lazy is usually used somewhere anyway, etc). This was a very special case.

nvborisenko:
But in future if we really want to make some `lazy property initialization thread-safe and AOT-friendly` what coding pattern will be?

nvborisenko:
Like here in BiDi namespace: https://github.com/SeleniumHQ/selenium/blob/trunk/dotnet/src/webdriver/BiDi/BiDi.cs. I use `Lazy<T>`, do we have better options?

RenderMichael:
Hmm, that might actually be a good place to do the same trick. I would write a sample app that uses only a small part of the BiDi commands, and try replacing the Lazy field with a property that creates the value on initialization with `??=`. Maybe it will allow a good amount of code to be trimmed away

nvborisenko:
I asked big brain:

To replace Lazy<T> in AOT (Ahead-of-Time) environments like Unity IL2CPP or .NET Native, you need a solution that avoids reflection. Here's an AOT-friendly approach using manual lazy initialization:

1. Why Lazy<T> Isn't AOT-Friendly
The default constructor of Lazy<T> uses Activator.CreateInstance, which relies on reflection and is incompatible with AOT compilation.

Even when using Lazy<T> with a factory delegate, some AOT environments may still restrict its use due to internal synchronization mechanisms.

2. AOT-Friendly Replacement
Implement a simple thread-safe lazy initializer using a Func<T> delegate and explicit locking:

```csharp
public class AotFriendlyLazy<T>
{
    private readonly Func<T> _factory;
    private T _value;
    private bool _isInitialized;
    private readonly object _lock = new object();

    public AotFriendlyLazy(Func<T> factory)
    {
        _factory = factory ?? throw new ArgumentNullException(nameof(factory));
    }

    public T Value
    {
        get
        {
            if (!_isInitialized)
            {
                lock (_lock)
                {
                    if (!_isInitialized)
                    {
                        _value = _factory();
                        _isInitialized = true;
                    }
                }
            }
            return _value;
        }
    }
}
```
Usage Example:
```csharp
private readonly AotFriendlyLazy<ExpensiveResource> _resource =
    new AotFriendlyLazy<ExpensiveResource>(() => new ExpensiveResource());

public ExpensiveResource Resource => _resource.Value;
```
3. Key Features
No Reflection: Explicitly initializes objects via a provided Func<T>.

Thread Safety: Uses double-check locking to ensure thread safety.

AOT Compatibility: Avoids reflection and dynamic code generation.

4. Alternatives
Explicit Initialization with Lazy<T> (if supported):
If your AOT environment allows it, use Lazy<T> with an explicit factory delegate (avoids Activator.CreateInstance):

```csharp
// Works if Lazy<T>'s non-reflection code paths are AOT-friendly
private readonly Lazy<ExpensiveResource> _resource =
    new Lazy<ExpensiveResource>(() => new ExpensiveResource());
```
LazyInitializer:
Use LazyInitializer.EnsureInitialized for simpler scenarios (ensure it doesn’t use reflection in your target environment):

```csharp
private ExpensiveResource _resource;
public ExpensiveResource Resource => 
    LazyInitializer.EnsureInitialized(ref _resource, () => new ExpensiveResource());
```
5. When to Use This Approach
Unity/IL2CPP: Required for compatibility.

.NET Native/UWP: Avoids reflection restrictions.

High-Performance Scenarios: Manual control over initialization logic.

This pattern ensures AOT compatibility while retaining the "initialize-on-first-use" behavior of Lazy<T>.

nvborisenko:
What I know: `Lazy<T>` is thread-safe. We really want to be thread-safe.

RenderMichael:
The big brain is not wrong there, about AOT compatibility. I think the normal Lazy<T> is AOT-safe, this PR solves another issue: AOT rooting. 

Anything in the Lazy’s factory is rooted by the AOT compiler as “something which may be invoked”. This means AOT trimming cannot eliminate the NetworkManager constructor, and that constructor also roots the entire DevTools functionality. That’s megabytes of code that is never called.

Lazy<T> cannot do this. If we had a `Lazy<INetwork>` then we would not see any of those savings in size.

RenderMichael:
@nvborisenko I tried it out. Just like I thought, it didn't save anything. (`11.0MB` - `2.0kB`)
image

image

This PR was a special case where the constructor, by itself, was very heavy.

PR #15228, opened 2025-02-04
[dotnet] Increment `WebDriver` towards nullability

RenderMichael:
### **User description**
**Thanks for contributing to Selenium!**
**A PR well described will help maintainers to quickly review and merge it**

Before submitting your PR, please check our [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) guidelines.
Avoid large PRs, help reviewers by making them as simple and short as possible.

Increment `WebDriver` towards nullability

### Motivation and Context

Contributes to #14640 

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist

- [x] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

___

### **PR Type**
Enhancement, Bug fix

___

### **Description**
- Incrementally refactored `WebDriver` to support nullability annotations.

- Improved error handling with null checks and exceptions.

- Enhanced code readability by simplifying property definitions.

- Updated `WebElementFactory` methods to handle nullable dictionaries.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>WebDriver.cs</strong><dd><code>Refactored `WebDriver` for nullability and error handling</code></dd></summary>
<hr>

dotnet/src/webdriver/WebDriver.cs

<li>Added nullability annotations to properties and methods.<br> <li> Replaced manual null checks with null-coalescing operators.<br> <li> Simplified property definitions using expression-bodied members.<br> <li> Improved error handling with specific exceptions for null values.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15228/files#diff-ef1874c09b1dca0d62ea1f0ead685355944b9cb7ae63df60ad9b32abdf90c8dc">+105/-127</a></td>

</tr>

<tr>
  <td>
    <details>
[...]

Issue #14846, opened 2024-12-03
[🐛 Bug]: IWebElement.GetDomProperty("style") returns "System.Collections.Generic.Dictionary`2[System.String,System.Object]" (as string)


Mek7:
### What happened?

I am upgrading Selenium test code in .NET 8 from Selenium.Support version 4.25.0 to 4.27.0
Because IWebElement.GetAttribute() method is now marked as deprecated, I replaced it with GetDomProperty().
For example:
`webDriver.FindElement(By....whatever).GetDomProperty("style")`
should return a string like
`background: rgb(65, 131, 215); height: 8.125%;`
but returns this string:
`System.Collections.Generic.Dictionary`2[System.String,System.Object]`
Looks like a bug - that string is the string representation of calling .ToString() on a Dictionary instance.
Or is it supposed to work like this?

When I use GetDomAttribute() instead of GetDomProperty(), the correct string is returned:
`background: rgb(65, 131, 215); height: 8.125%;`

### How can we reproduce the issue?

```shell
See above in "What happened"
```

### Relevant log output

```shell
Not sure - using Selenium in Visual Studio unit test framework. Not seeing any useful stuff in debug output.
```

### Operating System

Windows 11

### Selenium version

.NET 4.27.0

### What are the browser(s) and version(s) where you see this issue?

Firefox 132.0.2

### What are the browser driver(s) and version(s) where you see this issue?

GeckoDriver 0.35.0

### Are you using Selenium Grid?

no

nvborisenko:
For maintainers:

**GetDomProperty:**

```csharp
var style = element.GetDomProperty("style");
Console.WriteLine(style); // System.Object[]
```

```
21:00:42.948 TRACE HttpCommandExecutor: >> GET RequestUri: http://localhost:57238/session/92b41016a843a63333ee03f6aa03c648/element/f.533155AB6C15ADAF23E24748C23919EA.d.A26436E61238A5F5AA2A84DDBC427528.e.12/property/style, Content: null, Headers: 3
21:00:42.955 TRACE HttpCommandExecutor: << StatusCode: 200, ReasonPhrase: OK, Content: System.Net.Http.HttpConnectionResponseContent, Headers: 1
{"value":["height","line-height","width"]}
```

**GetDomAttribute:**

```csharp
var style = element.GetDomAttribute("style");
Console.WriteLine(style); // height:20px;line-height:20px;width:20px
```

```
[...]

PR #15230, opened 2025-02-04
[dotnet] Annotate nullability on more of `WebElement`

RenderMichael:
### **User description**

Annotate nullability on more of `WebElement`

### Motivation and Context

Contributes to #14640 

### Types of changes

### Checklist

___

### **PR Type**
Enhancement, Bug fix

___

### **Description**
- Added nullability annotations to `WebElement` class for improved type safety.

- Enhanced error handling for null or unexpected responses in various methods.

- Updated method signatures to reflect nullable return types where applicable.

- Improved code readability and maintainability with structured nullability directives.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>WebElement.cs</strong><dd><code>Add nullability annotations and enhance error handling</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/WebElement.cs

<li>Enabled nullability annotations to improve type safety.<br> <li> Added error handling for null or unexpected response values.<br> <li> Updated method signatures to reflect nullable return types.<br> <li> Improved code structure with nullability directives and comments.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15230/files#diff-905b5488c93f362385670059ce14436b328e787a5964aec3f31e73266dd3ae2e">+68/-22</a>&nbsp; </td>

</tr>

___


dotnet/src/webdriver/WebElement.cs
```diff
@@ -549,10 +547,13 @@ public virtual string GetCssValue(string propertyName)
 
             Response commandResponse = this.Execute(DriverCommand.GetElementValueOfCssProperty, parameters);
 
-            return commandResponse.Value.ToString();
-        }
+            if (commandResponse.Value is null)
+            {
+                throw new WebDriverException("GetElementValueOfCssProperty command returned a successful result, but contained no data");
```
nvborisenko: We should test it carefully.

RenderMichael: I tried to be very careful with this, to change no behavior. I read the spec, it says this will always be a string.

The only reason I am comfortable with this change, is because it would previously be a `NullReferenceException` if this were `null`. This way, users have something better to report if there is a bug here.

nvborisenko: This is very good insight. Please give me a chance to revisit it.

dotnet/src/webdriver/WebElement.cs
```diff
@@ -77,7 +79,11 @@ public virtual string TagName
 
                 Response commandResponse = this.Execute(DriverCommand.GetElementTagName, parameters);
 
-                return commandResponse.Value.ToString();
+                if (commandResponse.Value is null)
+                {
+                    throw new WebDriverException("GetElementTagName command returned a successful result, but contained no data");
+                }
```
nvborisenko: Do you think `commandResponse.Value` should be non-nullable?.. I mean probably it might be, we should check. If yes, then it would be better to make a check at that layer.

If 100% no, I would suggest to introduce new method like `EnsureCommandResponseHasValue(...)` to minimize diplication.

RenderMichael: `Response.Value` can be null in many cases. Many commands send back no response content.

In this case, `TagName` never returns null. In other methods, like `GetAttribute`, `null` is a valid response.

I can minimize the code here by doing something like:

```C#
return commandResponse.Value?.ToString() ?? throw new WebDriverException("GetElementTagName command returned a successful result, but contained no data");;
```

The situation is case-by-case here, involving reading the spec and figuring out which commands may return `null` and which ones cannot. That is why it's hard to make a general solution here.

RenderMichael: Ideally, we would cast to `(string)` instead of using `.ToString()`, because we should know exactly the expected response type. Otherwise, we get issues like this one: https://github.com/SeleniumHQ/selenium/issues/14846

nvborisenko: I see, what about `EnsureCommandResponseHasValue(...)`?

nvborisenko: Even in `OpenQA.Selenium.Response` class? And consumers invoke it when necessary.

RenderMichael: There are only  4 instance of null checks on `.Value` in this PR. Searching through the repo shows no other instances of a null check like this:
image

I understand what you mean, it's the same operation in multiple places. It just feels like a null-checking method would make the code more complex instead of simpler.

nvborisenko: Today 4 duplicated entries, tomorrow greater (https://github.com/SeleniumHQ/selenium/pull/15236). I would say it is better to introduce new method. The situation is the same as `HttpResponseMessage.EnsureSuccessStatusCode()`.

RenderMichael: There are two main types which care about `.Value`: `WebElement` and `WebDriver`. 

Incremental progress on `WebDriver` here: https://github.com/SeleniumHQ/selenium/pull/15228

Other places are where we call `InternalExecute`, easily visible in this PR: https://github.com/SeleniumHQ/selenium/pull/15233

You can see that in most places, if we care about the response value at all (so many places do not) then we do not do a null check. We do a type check instead (eg. `is Dictionary<string, object?>`, `is object[]`)

I want to do a refactoring on `Response` once the obsolete members can be fully removed and the type is completely immutable. I want to expose a `JsonNode? ValueNode` so users will not have to deal with "deserializer magic" regarding which type the `Value` is. That might affect the null check here as well.

nvborisenko: I just vote for minimize code, just to introduce `internal` method to ensure the `.Value` cannot be null. Later, we will see all consumers who is aware about null. I still think it will be helpful for us.

RenderMichael: Let me add that method.

RenderMichael: Opened PR https://github.com/SeleniumHQ/selenium/pull/15243

RenderMichael: Previous PR is merged, new helper is now being used here.

Issue #15355, opened 2025-03-01
[🚀 Feature]: Publish Selenium Manager as a native shared library


nvborisenko:
### Feature and motivation

Selenium Manager is a tool, executable binary. Meaning that if I want to use it programmatically I should invoke it as new OS process.

But it can be simplified:
- Share SM as native library (very easy to do!)
- Expose rust `extern` function (easy to do)

Then I will be able to use Selenium Manager functionality without a hell of "OS process communication". I am from .NET world.

### Usage example

- In .NET it will be 20 lines of code instead of 200 (including realtime logging from SM!)
- More performant! (and memory 50KB vs 200B)
- Lightweight footprint for end users: like if my application is targeted to MacOS, then only MacOS native SM will be a part of my application (at this time all 3 targets)
- I will be able to publish my application as single file. Currently SM binaries are placed besides of my application, I don't want to see SM.
- Bindings will not include executable binaries which leads to be more friendly with Antiviruses.

titusfortner:
Here's the analysis I'm looking at: https://chatgpt.com/share/67c77ba9-1d58-800a-8480-c7406f98e9b5

My takeaway is that it isn't worth having multiple implementations, but we should discuss it more.

cgoldberg:
No comment on the PR or decision, but I just wanna say that ChatGPT analysis is freakin incredible... like... wow 🤯

I'm still adjusting to how impressive some AI can be, and that really hit me. Scary.

titusfortner:
Yeah, I still have to go back and forth with it a bit to make sure I'm actually asking the right questions, but deep research is a game changer

shs96c:
I read the analysis, and it boils down to:

1. Language bindings need to include a native component
2. Linking against a DLL may cause the runtime to crash
3. In all cases, memory usage will reduce
4. There should be less code in the language bindings since more can be delegated to the rust code
5. In some cases deployment may be simpler (the `openssl` issue mentioned in the analysis)

Taking each of these in turn:

1. We already need to do this for the Selenium Manager binary itself. A DLL isn't a problem
2. We've faced this before (the original input library we used to use on Linux and Windows), and it's not a big issue
3. This is a Good Thing :)
4. This is likely true, and making things simpler is good. Replacing congruent code with a single implementation is a win.
5. This is a Good Thing too (I can imagine that code-signing on macOS may be an issue at some point for a standalone binary)

[...]

PR #15368, opened 2025-03-03
[rust] Publish Selenium Manager as native shared library

nvborisenko:
### **User description**
This is a PR for ongoing collaboration to make Selenium Manager be shipped as native library (C-compatible dynamic library).

### Motivation and Context
Implements #15355

### Checklist
- [x] Using `cargo build` should produce 3 output types (executable, rust lib, native shared lib)
- [ ] Update `bazel` where it is required
- [ ] Update github workflow to publish native libs to github artifacts
- [ ] Export rust function for external consumers (like we do `sm.exe --browser chrome`, but now as a function)

### Types of changes

### Checklist

___

### **PR Type**
Enhancement

___

### **Description**
- Added support for building Selenium Manager as a shared library.

- Introduced a new `rust_shared_library` target in Bazel build files.

- Exported a sample Rust function (`get_test`) for external usage.

- Updated `Cargo.toml` to define library crate types.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Configuration changes</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>defs.bzl</strong><dd><code>Introduced `rust_shared_library` in Bazel definitions</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

rust/defs.bzl

<li>Added <code>rust_shared_library</code> to the list of available Rust build targets.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15368/files#diff-1e4007bf9c975c70186f0a3eb4ad20cd2667a9662764253494f1739e519e2aa4">+2/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>rustfmt_wrapper.bzl</strong><dd><code>Added `rust_shared_library` to formatting wrapper</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

rust/private/rustfmt_wrapper.bzl

<li>Added <code>rust_shared_library</code> to the wrapper functions.<br> <li> Ensured formatting tests are applied to shared libraries.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15368/files#diff-184b6bbf8144457bb05df94f32091ac975acf051e51413db06f00122397ea754">+5/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>BUILD.bazel</strong><dd><code>Defined `rust_shared_library` target in Bazel</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

rust/BUILD.bazel

<li>Added a new <code>rust_shared_library</code> target for Selenium Manager.<br> <li> Configured the shared library to include necessary dependencies.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15368/files#diff-f61e0cab3fc6c7e1843594d5ef3fb58d3bd96878630213dfcd5a94ef4fdcc079">+13/-1</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>Cargo.toml</strong><dd><code>Updated Cargo.toml for shared library support</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

rust/Cargo.toml

<li>Defined library crate types (<code>cdylib</code> and <code>rlib</code>).<br> <li> Added configuration for the Selenium Manager binary and library.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15368/files#diff-74eb46d64310c0817ed15b2389cf81183c18cdebbaa256a4198057855ecf14d7">+7/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>lib.rs</strong><dd><code>Added exported function for shared library usage</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

rust/src/lib.rs

<li>Added an exported function <code>get_test</code> for external usage.<br> <li> Demonstrated how to expose Rust functions for C compatibility.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15368/files#diff-f64796ccc388f201986ea7eaaf07bbefbd8ac1a56bb995fb505da80879872a39">+13/-0</a>&nbsp; &nbsp; </td>

</tr>

___

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -42,12 +42,6 @@ use crate::shell::{
 use crate::stats::{send_stats_to_plausible, Props};
 use anyhow::anyhow;
 use anyhow::Error;
-// Re-export FFI functionality
-pub use ffi::{
-    WebDriverPathResult,
-    get_dummy_webdriver_path,
-    free_webdriver_path_result
-};
 use reqwest::{Client, Proxy};
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
@@ -63,7 +57,6 @@ pub mod downloads;
 pub mod edge;
 pub mod files;
 pub mod firefox;
-pub mod ffi;
 pub mod grid;
 pub mod iexplorer;
 pub mod lock;
@@ -74,6 +67,8 @@ pub mod safari;
 pub mod safaritp;
 pub mod shell;
 pub mod stats;
+
+mod ffi;
 
 pub const REQUEST_TIMEOUT_SEC: u64 = 300; // The timeout is applied from when the request starts connecting until the response body has finished
 pub const STABLE: &str = "stable";


rust/src/lib.rs
```diff
@@ -1674,3 +1674,16 @@ fn get_index_version(full_version: &str, index: usize) -> Result<String, Error>
         .ok_or(anyhow!(format!("Wrong version: {}", full_version)))?
         .to_string())
 }
+
+// ----------------------------------------------------------
+// Exported functions
+// ----------------------------------------------------------
+
+// this just an example how to expose function for external usage
```
nvborisenko: @bonigarcia can you please help to implement this function in rust? This function should be equivalent of `selenium-manager.exe --browser chrome --browser-version 130 --browser-path /some/path --proxy some_value`.

bonigarcia: This is a major change, so there should be a project consensus on this feature first. Personally, I am not sure about it. Are the other bindings (in addition to .Net) following this approach?

titusfortner:
@nvborisenko since this will be easiest for .NET, what would those bindings need to do to use this? What would be next steps/timeframe?

nvborisenko:
I have prepared a checklist:
- Using cargo build should produce 3 output types (executable, rust lib, native shared lib) - **DONE?**
- Update `bazel` where it is required - **I am not sure it is even required. Probably if Boni wants to create new unit test**
- Update github workflow to publish native libs to github artifacts - **Here you Titus can help. We expect `sm.so`, `sm.dll,` and `sm.dylib`** published as a release to https://github.com/SeleniumHQ/selenium_manager_artifacts besides regular `sm`, `sm.exe`. Naming convention should be the same as for binary (excepting file extension), reflecting OS architecture where it can be ran (literally as is).
- Export rust function for external consumers (like we do sm.exe --browser chrome, but now as a function) - **And here it is big question. When we call SM, what we want to get? I see only `driver_path` and `browser_path`.** In any case @bonigarcia might help or not, we should: 1) understand how to avoid memory leaking (when rust allocates memory, and who is release). 2) Agree on method signature: input/output parameters.

nvborisenko:
> @nvborisenko since this will be easiest for .NET, what would those bindings need to do to use this? What would be next steps/timeframe?

I asked big brain, seems it is easy to do for all bindings. But, yes, .NET is the easiest.

titusfortner:
I think this is what it would look like, and probably pretty straightforward:

src/
├── main.rs      # CLI entry point (minimal, calls into lib.rs)
├── lib.rs       # Core logic (shared between CLI and FFI)
├── ffi.rs       # FFI functions (pub extern "C")

Move the implementation details from main.rs to lib.rs
Put the external functions in ffi.rs
Both main.rs and ffi.rs manage their respective inputs, execute the functions in lib.rs and manage their respective outputs

I'm not sure how .NET handles streaming the logs via ffi, and I think we're also supposed to call in to free the memory once we've gotten the paths and stuff, but should all be doable.

titusfortner:
All the logistics for building/storing should be just updating this one file - https://github.com/SeleniumHQ/selenium/blob/trunk/.github/workflows/ci-rust.yml

FFI needs a get_paths function that accepts a JSON object and returns a JSON object. We need both browser path and driver path to send back to the bindings. Service class starts the driver at the provided path, then the driver needs to reference the matching browser location when sending capabilities to the driver

nvborisenko:
> FFI needs a get_paths function that accepts a JSON object and returns a JSON object.

No json objects in native functions, please. We have primitive strings, numbers, booleans, delegates (callback functions).

> I'm not sure how .NET handles streaming the logs via ffi

To "stream" logs in runtime, rust will just call this delegate (pointer to function), provided by bindings.

titusfortner:
We need to send key value pairs back and forth, there's no primitive for a map, and everything works with json

nvborisenko:
Structures.

nvborisenko:
Streaming log messages is resolved (I am not care about memory leaking yet). Verified, works good.

About returning of `driver_path` and `browser_path`. Do we really need it? Can it be 2 separate functions? I am asking because of I remember the logic in bindings when SM might return empty `browser_path`, and then binding decide something. Now we have a great chance to revisit what exactly we want to get from SM and when. Seems it might be 2 separate functions?

titusfortner:
> Streaming log messages is resolved (I am not care about memory leaking yet). Verified, works good.

Excellent!

> Can it be 2 separate functions?

I was going to say no, but we might be able to do it in 2 method calls... 
The main thing is that the driver method requires running the browser method, so they definitely can't be run async, and the rust functions don't maintain state between calls... But you could use the output from the browser method as an argument to the driver method. I'm just not sure how that makes the code better

> Structures

Eesh, I thought the point was to decrease the code we needed in the bindings? 😂 JSON would be a lot simpler, wouldn't require managing memory, and the overhead is going to be a lot less than what we have now. But I'm open to it.

nvborisenko:
Memory management is resolved. Rust part is here.

C# full example:
```csharp
using System;
using System.Runtime.InteropServices;

namespace ClassLibrary1
{
    public static partial class SeleniumManager
    {
        [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
        public delegate void LogCallback(int level, string message);

        private static readonly LogCallback _callback;

        static SeleniumManager()
        {
            // Define the callback logic
            _callback = (result, message) =>
            {
                Console.WriteLine($"Result: {result}, Message: {message}");
            };
        }

        public static string GetDummyDriverPath(string driverName)
        {
            // Convert the delegate to a function pointer
            IntPtr _callbackPtr = Marshal.GetFunctionPointerForDelegate(_callback);

            var result = get_dummy_webdriver_path(driverName, _callbackPtr);

            try
            {
                return Marshal.PtrToStringAnsi(result.DriverPath);
            }
            finally
            {
                free_webdriver_path_result(ref result);
            }
        }

        [StructLayout(LayoutKind.Sequential)]
        struct WebDriverPathResult
        {
            public bool IsSuccess;
            public IntPtr DriverPath;
            public IntPtr Error;
        }

        [DllImport("selenium_manager", CallingConvention = CallingConvention.Cdecl)]
        static extern WebDriverPathResult get_dummy_webdriver_path(string driverName, IntPtr callback);

        [DllImport("selenium_manager")]
        static extern void free_webdriver_path_result(ref WebDriverPathResult result);
    }
}
```

And usage:
```csharp
Console.WriteLine(SeleniumManager.GetDummyDriverPath("chrome").DriverPath)
```

Next step: resolve how rust will share errors (even unhandled).

rust/src/lib.rs
```diff
@@ -1674,3 +1674,80 @@ fn get_index_version(full_version: &str, index: usize) -> Result<String, Error>
...
+pub extern "C" fn get_dummy_webdriver_path(driver_name: *const std::ffi::c_char, log: LogCallback) -> WebDriverPathResult {
+    let result = std::panic::catch_unwind(|| {
+        
+        for i in 1..6 {
+            let message = std::ffi::CString::new("Hello, I am logging message").unwrap();
+            //let message = std::ffi::CString::new(String::from("A").repeat(10_000_000)).unwrap();
+            log(i, message.as_ptr());
+        }
+
+        //panic!("Intentional panic for testing");
```
nvborisenko: @bonigarcia can you help me here? We want to handle all panics in this block and propagate it in returning result. Seems I do it incorrectly.

nvborisenko:
@titusfortner about `ffi.rs` module. We can move all extern functions into this separate module, but we also should re-declare these functions in `lib.rs` anyway. As I understand correctly `lib.rs` is a compilation unit. BTW moving the implementations of functions to `ffi.rs` is a good idea, despite that we also need to duplicate function "signatures" in `lib.rs`. I think we still want doing it.

nvborisenko:
Moved to `ffi.rs`.

Now, as a .NET/C# consumer, I am ready to try out real function instead of "dummy".

How I see further steps:
- I can continue to adjust GitHub Workflow to publish native library
- We should finalize function signature (what to return and what to accept)
- @bonigarcia (or anybody else) will replace "dummy placeholder" with real implementation (in `ffi.rs`)

rust/src/lib.rs
nvborisenko: It has been moved to another file. I will request another review.

rust/src/ffi.rs
```diff
@@ -0,0 +1,91 @@
...
+pub extern "C" fn get_dummy_webdriver_path(driver_name: *const c_char, log: LogCallback) -> WebDriverPathResult {
+    let result = std::panic::catch_unwind(|| {
+        
+        for i in 1..6 {
+            let message = CString::new("Hello, I am logging message").unwrap();
+            //let message = CString::new(String::from("A").repeat(10_000_000)).unwrap();
+            log(i, message.as_ptr());
+        }
+
+        //panic!("Intentional panic for testing");
```
nvborisenko, line 41: @bonigarcia can you help me here? We want to handle all panics in this block and propagate it in returning result. Seems I do it incorrectly.

nvborisenko:
Earlier benchmark results, where `AsProcess` current behavior (cached), `AsLibrary` - "dummy" SM implementation.

| Method    | Mean             | Error           | StdDev          | Gen0   | Allocated |
|---------- |-----------------:|----------------:|----------------:|-------:|----------:|
| AsProcess | 266,338,900.0 ns | 4,967,127.39 ns | 4,878,383.27 ns |      - |   50608 B |
| AsLibrary |         992.8 ns |         8.97 ns |         7.95 ns | 0.0782 |     496 B |

  Mean      : Arithmetic mean of all measurements
  Error     : Half of 99.9% confidence interval
  StdDev    : Standard deviation of all measurements
  Gen0      : GC Generation 0 collects per 1000 operations
  Allocated : Allocated memory per single operation (managed only, inclusive, 1KB = 1024B)
  1 ns      : 1 Nanosecond (0.000000001 sec)

titusfortner:
So, for Windows you'd update the `ci-rust.yml` with something like this. It should all end up in selenium_manager_artifacts release as expected.

      - name: "Build shared library"
        run: cargo build --release --crate-type=cdylib
        working-directory: rust

      - name: "Rename binary"
        run: |
          mv rust/target/release/selenium-manager.exe selenium-manager-windows.exe
          mv rust/target/release/selenium_manager.dll selenium-manager-windows.dll

      - name: "Upload release binary"
        uses: actions/upload-artifact@v4
        with:
          name: selenium-manager-windows
          path: |
            selenium-manager-windows.exe
            selenium-manager-windows.dll
          retention-days: 6

In Release Job

          linux_lib_sha=$(shasum -a 256 artifacts/selenium-manager-linux/selenium-manager-linux.so | awk '{print $1}')
          macos_lib_sha=$(shasum -a 256 artifacts/selenium-manager-macos/selenium-manager-macos.dylib | awk '{print $1}')
          windows_lib_sha=$(shasum -a 256 artifacts/selenium-manager-windows/selenium-manager-windows.dll | awk '{print $1}')

          echo "{\"macos\": \"$macos_sha\", \"macos_lib\": \"$macos_lib_sha\", \"windows\": \"$windows_sha\", \"windows_lib\": \"$windows_lib_sha\", \"linux\": \"$linux_sha\", \"linux_lib\": \"$linux_lib_sha\"}" > latest.json


          files: |
            artifacts/selenium-manager-linux/selenium-manager-linux
            artifacts/selenium-manager-linux/selenium-manager-linux.so
            artifacts/selenium-manager-macos/selenium-manager-macos
            artifacts/selenium-manager-macos/selenium-manager-macos.dylib
            artifacts/selenium-manager-windows/selenium-manager-windows.exe
            artifacts/selenium-manager-windows/selenium-manager-windows.dll
            artifacts/selenium-manager-linux-debug/selenium-manager-linux-debug.tar
            artifacts/selenium-manager-macos-debug/selenium-manager-macos-debug.tar
            artifacts/selenium-manager-windows-debug/selenium-manager-windows-debug.exe

titusfortner:
Doing more research...

* `pub use ffi` --> this is only needed if we expect other rust applications to need access. We can add it later if we need it, but it isn't necessary so long as ffi file is compiled.

* Structs --> it looks like it is an especially good fit for .NET, but I'm not very elegant for other languages due to how optional values must be handled. Not important right now, but we will probably want a second rust method in ffi.rs that converts JSON to the struct when we implement in other languages and just take the small serialization performance hit.

* Configurations --> We probably need to make `ManagerConfig` the source of truth (merge it with Cli struct), so main is like: 

```
use clap::Parser;
use selenium_manager::{ManagerConfig, manage_selenium, SeleniumPaths};
use std::process::exit;

fn main() {
    let config = ManagerConfig::parse();

    match manage_paths(config) {
        Ok(SeleniumPaths { driver_path, browser_path }) => {
            exit(0);
        }
        Err(e) => {
            eprintln!("Error: {}", e);
            exit(1);
        }
    }
}
```

ffi is like:
```
use selenium_manager::{ManagerConfig, manage_selenium, SeleniumPaths};

#[no_mangle]
pub extern "C" fn run_selenium_manager(config: ManagerConfig) -> i32 {
    match manage_paths(config) {
        Ok(SeleniumPaths { driver_path, browser_path }) => {
            0
        }
        Err(_) => {
            1
        }
    }
}
```

Then everything that is in main now needs to get put into `manage_paths`.
We could break out logging and clear cache/metadata if we want to as well.

rust/src/ffi.rs
bonigarcia: Sorry, I don't have bandwidth for the following weeks. In any case, this deserves to be discussed in the following Selenium Dev Summit in Valencia (24-25 March).

nvborisenko:
What is plan? As usual just to forgive and forget as stale?

diemol:
I think we replied in Slack. 

But what we discussed in TLC is that this approach makes sense only for .NET, I beleve because it is the only where performance improvement could be seen. The other languages would need to do more to adapt to it and there won't be any performance improvements. 

The question is if we do this just for .NET or not. What do you think?

nvborisenko:
IMHO this story is not about .NET.

diemol:
Can you add arguments and say why you think that? Not only opinions.

nvborisenko:
> But it can be simplified:
>
> Share SM as native library (very easy to do!)
Expose rust extern function (easy to do)
Then I will be able to use Selenium Manager functionality without a hell of "OS process communication". I am from .NET world.
> Usage example

> In .NET it will be 20 lines of code instead of 200 (including realtime logging from SM!)
More performant! (and memory 50KB vs 200B)
Lightweight footprint for end users: like if my application is targeted to MacOS, then only MacOS native SM will be a part of my application (at this time all 3 targets)
I will be able to publish my application as single file. Currently SM binaries are placed besides of my application, I don't want to see SM.
Bindings will not include executable binaries which leads to be more friendly with Antiviruses.

PR #15395, opened 2025-03-09
[dotnet] [bidi] Align `Scipt.LocalValue.Map` with spec, enable negative zero

### LINKED ISSUE #15394: [🐛 Bug]: [dotnet] BiDi `LocalValue.Map` cannot be passed to the `Script.CallFunctionAsync` command
### What happened?

The BiDi type `Scipt.LocalValue.Map` should be passable to the `Script.CallFunctionAsync` method. However, it fails with a strange exception.

Per [the spec](https://w3c.github.io/webdriver-bidi/#type-script-LocalValue), the `Map` is serialized the same was an `Object` is:
Image

An Object is an `IEnumerable<IEnumerable<LocalValue>>`, and works fine in this scenario. For that reason, I am suspicious if `Map` should be changed into the same shape.

Same thing applies for the returned `RemoteValue.Map`.

```shell
#
var bidi = await driver.AsBiDiAsync();
var context = (await bidi.BrowsingContext.GetTreeAsync())[0].Context;
var map = new LocalValue.Map(new Dictionary<string, LocalValue>
{
    {"key", new LocalValue.String("value") }
});
var response = await context.Script.CallFunctionAsync($$"""
     (arg) => {
       return arg;
     }
     """, false, new() { Arguments = [map] });
```

### Relevant log output

```shell
Message: 
OpenQA.Selenium.BiDi.BiDiException : invalid argument: Invalid input in "arguments"/0.

  Stack Trace: 
Broker.ExecuteCommandCoreAsync[TCommand](TCommand command, CommandOptions options) line 216
Broker.ExecuteCommandAsync[TCommand,TResult](TCommand command, CommandOptions options) line 186
```

### Operating System

N/A

### Selenium version

trunk

chrome 133.0.6943.143

chromedriver 133.0.6943.141

_No response_

RenderMichael:
### **User description**
Fixes using the `Script.LocalValue.Map` type and the related return type. The spec describes it the same as `Object`:
image

So this is now a "clone" of `LocalValue.Object`.
 
Additionally, this PR deserializes negative zero from a string, in alignment with the spec:
image

### Motivation and Context

Fixes https://github.com/SeleniumHQ/selenium/issues/15394

### Types of changes

### Checklist

___

### **PR Type**
Bug fix, Tests

___

### **Description**
- Fixed `Script.LocalValue.Map` to align with the WebDriver BiDi spec.

- Updated `RemoteValue.Map` to use `IReadOnlyList<IReadOnlyList<RemoteValue>>`.

- Added comprehensive tests for `LocalValue` and `RemoteValue` roundtrip scenarios.

- Enhanced JSON serialization options to handle named floating-point literals.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>Broker.cs</strong><dd><code>Enhanced JSON serialization options.</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/BiDi/Communication/Broker.cs

<li>Added support for named floating-point literals in JSON serialization.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15395/files#diff-5f7fc0e0b816bad29734139ccd27ebb0ff8809c8f49de7827eab7c6c7e8f9669">+1/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Bug fix</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>LocalValue.cs</strong><dd><code>Updated `LocalValue.Map` structure.</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs

<li>Changed <code>Map</code> to use <code>IEnumerable<IEnumerable<LocalValue>></code>.<br> <li> Ensured alignment with <code>Object</code> serialization structure.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15395/files#diff-b2d3bd83230255db2e435ac34700a0af0595469211636058e6eb477107171b22">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>RemoteValue.cs</strong><dd><code>Modified `RemoteValue.Map` structure.</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/BiDi/Modules/Script/RemoteValue.cs

<li>Updated <code>Map</code> to use <code>IReadOnlyList<IReadOnlyList<RemoteValue>></code>.<br> <li> Ensured compatibility with the WebDriver BiDi spec.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15395/files#diff-0aa418714ecf5cc03a9f8342f491b0fb0fabcf2af051815e9bf56136834e31e5">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>CallFunctionParameterTest.cs</strong><dd><code>Added comprehensive tests for `LocalValue` and `RemoteValue`.</code></dd></summary>
<hr>

dotnet/test/common/BiDi/Script/CallFunctionParameterTest.cs

<li>Added new test cases for <code>LocalValue</code> and <code>RemoteValue</code> roundtrip <br>scenarios.<br> <li> Verified serialization and deserialization of various <code>LocalValue</code> <br>types.<br> <li> Tested <code>Map</code> and <code>Object</code> structures for compliance with the spec.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15395/files#diff-2211a55aa8f7fbbe821213f4f0a771f85a33b7b639fc7bbe2e4bd38a48c85608">+139/-0</a>&nbsp; </td>

</tr>

___


dotnet/src/webdriver/BiDi/Communication/Broker.cs
```diff
@@ -64,6 +64,7 @@ internal Broker(BiDi bidi, ITransport transport)
             PropertyNameCaseInsensitive = true,
             PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
             DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
+            NumberHandling = JsonNumberHandling.AllowNamedFloatingPointLiterals,
```
RenderMichael: Also fixes special numbers, which are allowed by the spec and are working in the tests
image

dotnet/test/common/BiDi/Script/CallFunctionParameterTest.cs
```diff
@@ -219,4 +222,140 @@ public async Task CanCallFunctionInARealm()
         Assert.That(res1, Is.EqualTo(3));
         Assert.That(res2, Is.EqualTo(5));
     }
+
+    [Test]
+    [TestCaseSource(nameof(RoundtripOptions))]
```
RenderMichael: Can `bazel` handle `TestCaseSource`?

dotnet/test/common/BiDi/Script/CallFunctionParameterTest.cs
```diff
@@ -219,4 +222,140 @@ public async Task CanCallFunctionInARealm()
...
+
+        yield return new TestCaseData(new LocalValue.Null(), new RemoteValue.Null(), "arg === null")
+        {
+            TestName = nameof(CanCallFunctionAndRoundtrip_Five) + "(Null)"
+        };
+        yield return new TestCaseData(new LocalValue.Undefined(), new RemoteValue.Undefined(), "typeof arg === 'undefined'")
+        {
+            TestName = nameof(CanCallFunctionAndRoundtrip_Five) + "(Undefined)",
+        };
+        //yield return new TestCaseData(new LocalValue.Boolean(true), new RemoteValue.Boolean(true), "typeof arg === true")
```
RenderMichael: I found this issue while testing the new `Boolean` local value. I will uncomment this when I have that working.

dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs
```diff
@@ -103,7 +103,7 @@ public record Array(IEnumerable<LocalValue> Value) : LocalValue;
 
     public record Date(string Value) : LocalValue;
 
-    public record Map(IDictionary<string, LocalValue> Value) : LocalValue; // seems to implement IDictionary
+    public record Map(IEnumerable<IEnumerable<LocalValue>> Value) : LocalValue;
```
RenderMichael, line 103: Breaking change, of course. It does not work without this, so hopefully this is acceptable.

dotnet/test/common/BiDi/Script/CallFunctionParameterTest.cs
```diff
@@ -219,4 +222,155 @@ public async Task CanCallFunctionInARealm()
...
+            Assert.That(actualMap.Value, Is.EqualTo(expectedMap.Value));
+        }
+        else if (response.Result is RemoteValue.Set actualSet && expected is RemoteValue.Set expectedSet)
+        {
+            Assert.That(actualSet.Value, Is.EqualTo(expectedSet.Value));
+        }
+        else if (response.Result is RemoteValue.Date actualDate && expected is RemoteValue.Date expectedDate)
+        {
+            var actualDateParsed = DateTime.SpecifyKind(DateTime.Parse(actualDate.Value, CultureInfo.InvariantCulture), DateTimeKind.Utc);
+            Assert.That(actualDateParsed.Kind, Is.EqualTo(DateTimeKind.Utc));
```
RenderMichael: No longer necessary
```suggestion
```

dotnet/test/common/BiDi/Script/CallFunctionParameterTest.cs
```diff
@@ -219,4 +222,154 @@ public async Task CanCallFunctionInARealm()
         Assert.That(res1, Is.EqualTo(3));
         Assert.That(res2, Is.EqualTo(5));
     }
+
+    [Test]
+    [TestCaseSource(nameof(RoundtripOptions))]
+    public async Task CanCallFunctionAndRoundtrip_Five(LocalValue local, RemoteValue expected, string javaScriptAssert)
```
RenderMichael: Better name

```suggestion
    public async Task CanCallFunctionAndRoundtripArguments(LocalValue local, RemoteValue expected, string javaScriptAssert)
```

dotnet/test/common/BiDi/Script/CallFunctionParameterTest.cs
nvborisenko: It supports, but I strongly recommend split this "big messing test" to many small tests. Your approach is elegant, I think having simple and straightforward test is better.

dotnet/src/webdriver/BiDi/Communication/Broker.cs
nvborisenko: I am glad it is built-in, thanks!

dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs
nvborisenko: Not a problem.

dotnet/src/webdriver/BiDi/Communication/Broker.cs
nvborisenko: In Firefox:
```json
{"type":"number","value":"-0"}
```

Number in quotes.

```suggestion
            NumberHandling = JsonNumberHandling.AllowNamedFloatingPointLiterals | JsonNumberHandling.AllowReadingFromString,
```

dotnet/test/common/BiDi/Script/CallFunctionParameterTest.cs
```diff
@@ -224,152 +224,329 @@ public async Task CanCallFunctionInARealm()
...
-        )
+        Assert.That(response.Result, Is.AssignableTo<RemoteValue.Null>());
+        Assert.That(response.Result, Is.EqualTo(new RemoteValue.Null()));
+    }
+
+    [Test]
+    public async Task CanCallFunctionAndRoundtripArguments_EmptyString()
+    {
+        var response = await context.Script.CallFunctionAsync($$"""
+            (arg) => {
```
nvborisenko: What if just to return `arg`? So much simplified I hope..

RenderMichael: We want to do assertions on the JavaScript side to make sure we're serializing to the real JS type

dotnet/test/common/BiDi/Script/CallFunctionParameterTest.cs
RenderMichael: Switched to individual test cases

nvborisenko: And let me be honest: `if` statement in tests is bad.

dotnet/test/common/BiDi/Script/CallFunctionParameterTest.cs
nvborisenko: Which leads to create separate tests? For `LocalValue` and for `RemoteValue`?..

RenderMichael: The tests here cover:

- Does `LocalValue` serialize to a JavaScript type?
- Does a JavaScript type deserialize to a `RemoteValue` type?

Also `Script.CallFunction` is tested, but that is covered by the rest of the tests in this class.

nvborisenko: I am pretty sure we should not test JS, we are more interested in communication level. I just executed
```csharp
        var response = await context.Script.CallFunctionAsync("arg => { return arg; }", false, new() { Arguments = [new LocalValue.Array([new LocalValue.String("hi")])] });

        var expectedArray = new RemoteValue.Array { Value = [new RemoteValue.String("hi")] };
        Assert.That(response.Result, Is.AssignableTo<RemoteValue.Array>());
```

And it works. Very simple as expected.

RenderMichael: How will we know if a `LocalValue` did not serialize correctly into native JS types?

nvborisenko: Split "Roudtrip" test to more concrete assertions. For `LocalValue` and for `RemoteValue`.

nvborisenko: I see. `RemoteValue` we can verify easily. But how to verify `LocalValue`..

RenderMichael: The JavaScript side assertion check that the LocalValue works correctly, mapping onto the right JS type

nvborisenko: Right, seems we should move some logic to JS part. Seems easy to do, but we should align expectations at communications level.

In any case let's split tests, `Local/Remote`

RenderMichael: I separated the `LocalValue` argument tests from the `RemoteValue` return tests ✅

dotnet/test/common/BiDi/Script/CallFunctionParameterTest.cs
```diff
@@ -219,4 +222,513 @@ public async Task CanCallFunctionInARealm()
...
+              }
+            }
+            """, false, new() { Arguments = [arg] });
+        }, Throws.Nothing);
+    }
+
+    [Test]
+    public void CanCallFunctionWithArgumentRecentDate()
+    {
+        const string PinnedDateTimeString = "2025-03-09T00:30:33.083Z";
```
RenderMichael: Small question here: .NET uses times with a precision of 7 decimal points on the second, eg.

```
"2025-03-09T00:30:33.0832532Z"
```

Whereas the JS transition only retains 3 decimal places
```
"2025-03-09T00:30:33.083Z"
```

So if you do:

```C#
var date = DateTime.Now;

var response = await context.Script.CallFunctionAsync<DateTime>("(arg) => { return arg; }", false, new()
{
    Arguments = [new LocalValue.Date(date.ToString("o"))]
});

Assert.That(response, Is.EqualTo(date));
```

This will fail:

```ini
  Expected: 2025-03-10 13:07:02.6991741
  But was:  2025-03-10 13:07:02.699
```

Is this expected? Is this a limitation of the BiDi server, or of JS?

dotnet/test/common/BiDi/Script/CallFunctionParameterTest.cs
nvborisenko, line 1: That's very nice testing coverage. Last breaking change ask: please put new tests to separate `CallFunctionLocalValueTest` and `CallFunctionRemoteValueTest` fixtures. Thanks.

dotnet/test/common/BiDi/Script/CallFunctionParameterTest.cs
nvborisenko: Thanks for understanding, now if test fails we can quickly understand whom to blame, local or remote value.

dotnet/test/common/BiDi/Script/CallFunctionParameterTest.cs
nvborisenko: In JS `Date` object handles only milliseconds. Thus it is expected.

dotnet/test/common/BiDi/Script/CallFunctionParameterTest.cs
RenderMichael: Good idea, this file is getting a little long.

dotnet/test/common/BiDi/Script/CallFunctionParameterTest.cs
nvborisenko: Seeing it:

```csharp
Arguments = [new LocalValue.Date(date.ToString("o"))]
```

And I really want `implicit cast`:
```csharp
Arguments = [DateTime.Now]
```

But it will be upcoming story! This thread is resolved.

RenderMichael: This seems like a good thing to document somewhere, maybe in the XML Docs of `LocalValue.Date` and `RemoteValue.Date`, once the API shape is finalized.

RenderMichael: I actually think this is better!! The issue I brought up makes it clear: `DateTime` != `Date`. BiDi should only handle the low-level `Date` functionality - users which use BiDi should be aware of the conversion, since it is lossy.

dotnet/test/common/BiDi/Script/CallFunctionLocalValueTest.cs
```diff
@@ -0,0 +1,331 @@
...
+    [Test]
+    public void CanCallFunctionWithArgumentNumberNegative0()
+    {
+        var arg = new LocalValue.Number(double.NegativeZero);
+
+        Assert.That(async () =>
+        {
+            await context.Script.CallFunctionAsync($$"""
+            (arg) => {
+              if (arg !== -0) {
```
RenderMichael: This isn't right, it would pass for `+0` too. Let me improve this.

RenderMichael: Hmm, I think there might be a bug here? Does `-0` need to be in quotes? It seems to pass in normal `0`

image

dotnet/src/webdriver/BiDi/Communication/Broker.cs
```diff
@@ -64,6 +64,7 @@ internal Broker(BiDi bidi, ITransport transport)
             PropertyNameCaseInsensitive = true,
             PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
             DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
+            NumberHandling = JsonNumberHandling.AllowNamedFloatingPointLiterals | JsonNumberHandling.AllowReadingFromString,
```
nvborisenko, line 70: I propose to leave a comment about why we use `AllowReadingFromString`: firefox returns `-0` as string. Probably FF will fix it, and it will be easier to understand why we use it.

dotnet/test/common/BiDi/Script/CallFunctionLocalValueTest.cs
```diff
@@ -0,0 +1,331 @@
...
+using OpenQA.Selenium.BiDi.Modules.Script;
+
+namespace OpenQA.Selenium.BiDi.Script;
+
+class CallFunctionLocalValueTest : BiDiTestFixture
+{
+    [Test]
+    public void CanCallFunctionWithArgumentUndefined()
+    {
+        var arg = new LocalValue.Undefined();
```
nvborisenko, line 30: What if we choose more specific name for local variable instead of just `arg`? Then it will be easier for us to understand what exact type is behind of it (and in others tests).

dotnet/test/common/BiDi/Script/CallFunctionLocalValueTest.cs
```diff
@@ -0,0 +1,331 @@
...
+              }
+            }
+            """, false, new() { Arguments = [arg] });
+        }, Throws.Nothing);
+    }
+
+    [Test]
+    public void CanCallFunctionWithArgumentSet()
+    {
+        var argument = new LocalValue.Set([new LocalValue.String("key")]);
```
nvborisenko: "key" string.. just small improvement: if we name it "city", or "weather". I mean just to use any fake domain data name to make it easier to read logs. The same situation for "value" string in logs, it is hard to read. If we use any other hardcoded string like "whoa" or "hi", it will allow us to be not crazy while debugging tests.

nvborisenko:
Very important PR, Thank you @RenderMichael , just small suggestions. And I am happy to see it landed.

dotnet/src/webdriver/BiDi/Communication/Broker.cs
nvborisenko: Just to highlight "exceptional" rather than "regular"

dotnet/test/common/BiDi/Script/CallFunctionLocalValueTest.cs
RenderMichael: Fixed with a custom double converter
image

dotnet/src/webdriver/BiDi/Communication/Broker.cs
RenderMichael: No longer necessary, with the double converter I just added

dotnet/test/common/BiDi/Script/CallFunctionLocalValueTest.cs
RenderMichael: Improved the variable names to something more searchable.

dotnet/test/common/BiDi/Script/CallFunctionLocalValueTest.cs
RenderMichael: Improved the variable names to something more searchable 👍

dotnet/src/webdriver/BiDi/Communication/Broker.cs
RenderMichael: Sending `"value": -0` didn't work on chrome, it needed to be serialized as a string `"value": "-0"`.

RenderMichael: spec says it should be a string

image

nvborisenko: It just worked with `JsonNumberHandling.AllowReadingFromString`, did it?

dotnet/test/common/BiDi/Script/CallFunctionLocalValueTest.cs
nvborisenko: Misunderstood: e.g. `stringLocalValue`, `numberLocalValue`.. No need to specify exact value, just type.

dotnet/src/webdriver/BiDi/Communication/Broker.cs
RenderMichael: No, that only affects Nan/Infinity/-Infinity. Negative zero is normally serialized as a number, but I guess BiDi spec wants to be extra safe, so -0 doesn't turn into 0 without realizing it.

dotnet/test/common/BiDi/Script/CallFunctionRemoteValueTest.cs
```diff
@@ -0,0 +1,231 @@
...
+        Assert.That(response.Result, Is.EqualTo(new RemoteValue.String(string.Empty)));
+    }
+
+    [Test]
+    public async Task CanCallFunctionAndReturnNonEmptyString()
+    {
+        var response = await context.Script.CallFunctionAsync("() => { return 'whoa'; }", false);
+
+        Assert.That(response.Result, Is.AssignableTo<RemoteValue.String>());
+        Assert.That(response.Result, Is.EqualTo(new RemoteValue.String("whoa")));
```
nvborisenko: Magic of `record` types. Let's us be more specific. I understand what I am requesting.

RenderMichael: I'm not sure what changes I should make for this.

dotnet/src/webdriver/BiDi/Communication/Broker.cs
nvborisenko: Your previous test works good in Chrome and FF without double converter:
```csharp
    [Test]
    public async Task CanCallFunctionAndReturnNumberNegative0()
    {
        var response = await context.Script.CallFunctionAsync("() => { return -0; }", false);

        Assert.That(response.Result, Is.AssignableTo<RemoteValue.Number>());

        var expectedNegativeZero = ((RemoteValue.Number)response.Result).Value;
        Assert.That(IsNegativeZero(expectedNegativeZero));

        static bool IsNegativeZero(double d)
        {
            // '== double.NegativeZero' does not work, -0 == 0 is considered true
            // Need special verification to tell if a double is -0
            return double.IsNegative(d) && d == 0;
        }
    }
```

RenderMichael: Yes the problem was not reading the remote value, it was writing the local value. This test failed before this PR (`-0` written as a number gets turned into regular `0`):

```C#
  [Test]
    public void CanCallFunctionWithArgumentNumberNegative0()
    {
        var minus0 = new LocalValue.Number(double.NegativeZero);

        Assert.That(async () =>
        {
            await context.Script.CallFunctionAsync($$"""
            (arg) => {
              if (arg !== 0 || arg.toLocaleString()[0] !== '-') {
                throw new Error("Assert failed: " + arg.toLocaleString());
              }
            }
            """, false, new() { Arguments = [minus0] });
        }, Throws.Nothing);
    }
```

RenderMichael: Bottom line: the special numbers are only defined in the `script.PrimitiveProtocolValue` type. This type is used by `script.LocalValue` and `script.RemoteValue`.

In fact, we can remove the special number handling completely, and add the `DoubleConverter` to `LocalValue.Number.Value` and `RemoteValue.Number.Value`. This way, we can contain the non-standard behavior.

dotnet/test/common/BiDi/Script/CallFunctionRemoteValueTest.cs
nvborisenko: Before:
```csharp
Assert.That(response.Result, Is.EqualTo(new RemoteValue.String("whoa")));
```
After:
```csharp
Assert.That((response.Result as RemoteValue.String).Value, Is.EqualTo("whoa"));
```
So just avoid "magic object" assertion. Or do you think we can rely on "magic of records" in unit tests? Hm, we can rely, why not...

dotnet/src/webdriver/BiDi/Communication/Json/Converters/DoubleConverter.cs
nvborisenko, line 1: It means we apply this logic for absolutely all `double`, even when it is not required. Let's apply it explicitly for known special cases.

nvborisenko (changes requested):
We need a good way how to handle "special" cases for BiDi Numbers.

dotnet/test/common/BiDi/Script/CallFunctionRemoteValueTest.cs
RenderMichael: Ah, I see what you mean.

The assertion we have now also makes sure things like `Handle` and `InternalId` are the same. `RemoteValue.String` doesn't have that, but other things like `RemoteValue.Array`, `Object`, etc. do.

`record` type + NUnit's `Is.EqualTo` is smart enough to do the correct assertion here. Maybe users will also do `myLocalValue == otherLocalValue` too? It's nice to know that it works.

dotnet/src/webdriver/BiDi/Communication/Json/Converters/DoubleConverter.cs
RenderMichael: Good point.

Two special cases: `RemoteValue.Number` and `LocalValue.Number`

I updated to only explicitly reference the converter
image

RenderMichael: > We need a good way how to handle "special" cases for BiDi Numbers.

Lucky: BiDi numbers are only in these two types.

dotnet/src/webdriver/BiDi/Communication/Broker.cs
nvborisenko: > spec says it should be a string

It is false. CDDL language just says it can be "-0", but how this value will be transferred via JSON is not defined (it is up to vendors, `-0` or `"-0"`). We have learned that different vendors implement it differently, the question is what it should be converted to .NET double? In .NET do we have something special for `negative zero` and `positive zero`?

nvborisenko: image

Hm, is it just an alias for regular `0`?

RenderMichael: Sending `-0` is broken and `"-0"` works. The test I pasted above shows that sending `-0` does not turn into -0 on the JavaScript side.

RenderMichael: > It is false.

Spec:
image

Did I misunderstand CDDL grammar? I thought a number would be `-0` and a string would be `"-0"`. `-0` is a valid JSON number, but it does not work with BiDi.

RenderMichael: > Hm, is it just an alias for regular 0?

No, it is a true negative zero.
[`Console.WriteLine(JsonSerializer.Serialize(double.NegativeZero));`](https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwLACgACAmAjBzABgAJNcA6AFQFMAPYcgKQGcB7KAbgwNwE4AKFuwDK1MAEsAhgBtxALzHlREmfOr84rCACNp1cgDlqAc0nBxAN2oAtMawCU9jkA===)

nvborisenko: > Hm, is it just an alias for regular 0?

No, it is absolutely different value. We have to handle it, so `-0` should be `double.NegativeZero`.

> `-0` is a valid JSON number, but it does not work with BiDi

Then we have 2 options: or satisfy all vendors (chrome/ff), or raise a bug for "bad" vendor. I vote for the 1st option - seems it is easy to do (but in right manner).

nvborisenko:
We learned:
- `double.NegativeZero` is a .NET type for `-0`
- Special converters should be applied for special cases only

I propose to gather how vendors serialize special numbers in JSON, then it will be easier for us to handle.

RenderMichael:
For de-serializing a `-0`:
`await context.Script.CallFunctionAsync("() => { return -0; }", false);`

- Firefox: `{"type":"number","value":"-0"}`
- Chrome: `{"type":"number","value":"-0"}`

For serializing a `-0`:
```C#
var minus0 = new LocalValue.Number(double.NegativeZero);

await context.Script.CallFunctionAsync($$"""
    (arg) => {
        if (arg !== 0 || arg.toLocaleString()[0] !== '-') {
        throw new Error("Assert failed: " + arg.toLocaleString());
        }
    }
    """, false, new() { Arguments = [minus0] });
```
When arguments are serialized as `"arguments":[{"type":"number","value":-0}]`

- On Firefox: `Error: Assert failed: 0` (in other words, it thinks the value is regular `0`)
- On Chrome: `Error: Assert failed: 0` (same as above)

When arguments are serialized as `"arguments":[{"type":"number","value":"-0"}]`

- On Firefox: No exception is thrown ("is negative zero" check passed)
- On Chrome: No exception is thrown

## Conclusion

Across the board, for serializing and deserializing, on all browsers, negative zero is expected to be a string `"-0"`

nvborisenko:
> When arguments are serialized as "arguments":[{"type":"number","value":-0}] On Firefox: Error: Assert failed: 0 (in other words, it thinks the value is regular 0)

On my Firefox your test passes when we `{"type":"number","value":-0}`

nvborisenko:
OK, so in Firefox everything works good without any custom double converters. Chromium cannot handle `-0` and everything works good.

I posted https://github.com/w3c/webdriver-bidi/issues/887 issue in BiDi spec. I propose to not introduce custom converter at this time, and move forward with ignored unit tests (for chromium only).

PS: I was wrong, spec requires `-0` to be string:
```
[Type](https://tc39.es/ecma262/#sec-ecmascript-data-types-and-values)(value) is Number
Switch on the value of value:

NaN
Let serialized be "NaN"
-0
Let serialized be "-0"
```

dotnet/src/webdriver/BiDi/Communication/Broker.cs
RenderMichael: > I propose to leave a comment about why we use `AllowReadingFromString`: firefox returns `-0` as string. Probably FF will fix it, and it will be easier to understand why we use it.

> Just to highlight "exceptional" rather than "regular"

Done, comment explains the reasoning.

dotnet/test/common/BiDi/Script/CallFunctionLocalValueTest.cs
nvborisenko: Given a knowledge we got, new round of review.

dotnet/src/webdriver/BiDi/Communication/Json/Converters/DoubleConverter.cs
nvborisenko: Given a knowledge we got, new round of review.

dotnet/test/common/BiDi/Script/CallFunctionRemoteValueTest.cs
nvborisenko: So, use "magic of records" in unit tests?

dotnet/src/webdriver/BiDi/Communication/Broker.cs
```diff
@@ -64,6 +64,10 @@ internal Broker(BiDi bidi, ITransport transport)
             PropertyNameCaseInsensitive = true,
             PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
             DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
+
+            // BiDi returns special numbers such as "NaN" as strings
+            // Additionally, -0 is returned as a string "-0"
+            NumberHandling = JsonNumberHandling.AllowNamedFloatingPointLiterals | JsonNumberHandling.AllowReadingFromString,
```
nvborisenko, line 70: Seems now `JsonNumberHandling.AllowReadingFromString` is not necessary?

nvborisenko: Ah, Firefox returns it as `"-0"`.

nvborisenko: Verified.

```
{"type":"success","id":2,"result":{"realm":"4406696d-94fa-4ab8-9f5b-690afd3a2756","type":"success","result":{"type":"number","value":"-0"}}}
```

dotnet/test/common/BiDi/Script/CallFunctionLocalValueTest.cs
```diff
@@ -0,0 +1,333 @@
...
+              if (arg !== -5) {
+                throw new Error("Assert failed: " + arg);
+              }
+            }
+            """, false, new() { Arguments = [numberMinus5] });
+        }, Throws.Nothing);
+    }
+
+    [Test]
+    public void CanCallFunctionWithArgumentNumber0()
```
nvborisenko:
```suggestion
    public void CanCallFunctionWithArgumentNumberZero()
```

dotnet/test/common/BiDi/Script/CallFunctionLocalValueTest.cs
```diff
@@ -0,0 +1,333 @@
...
+              }
+            }
+            """, false, new() { Arguments = [number0] });
+        }, Throws.Nothing);
+    }
+
+    [Test]
+    [IgnoreBrowser(Selenium.Browser.Edge, "Chromium can't handle -0 argument as a number: https://github.com/w3c/webdriver-bidi/issues/887")]
+    [IgnoreBrowser(Selenium.Browser.Chrome, "Chromium can't handle -0 argument as a number: https://github.com/w3c/webdriver-bidi/issues/887")]
+    public void CanCallFunctionWithArgumentNumberNegative0()
```
nvborisenko:
```suggestion
    public void CanCallFunctionWithArgumentNumberNegativeZero()
```

dotnet/test/common/BiDi/Script/CallFunctionLocalValueTest.cs
```diff
@@ -0,0 +1,333 @@
...
+    {
+        var minus0 = new LocalValue.Number(double.NegativeZero);
+
+        Assert.That(async () =>
+        {
+            await context.Script.CallFunctionAsync($$"""
+            (arg) => {
+              if (arg !== 0 || arg.toLocaleString()[0] !== '-') {
```
nvborisenko:
```suggestion
              if (!Object.is(x, -0)) {
```

dotnet/test/common/BiDi/Script/CallFunctionLocalValueTest.cs
```diff
@@ -0,0 +1,333 @@
...
+using OpenQA.Selenium.BiDi.Modules.Script;
+
+namespace OpenQA.Selenium.BiDi.Script;
+
+class CallFunctionLocalValueTest : BiDiTestFixture
+{
+    [Test]
+    public void CanCallFunctionWithArgumentUndefined()
+    {
+        var undefined = new LocalValue.Undefined();
```
nvborisenko:
```suggestion
        var undefinedLocalValue = new LocalValue.Undefined();
```

dotnet/test/common/BiDi/Script/CallFunctionLocalValueTest.cs
```diff
@@ -0,0 +1,333 @@
...
+              }
+            }
+            """, false, new() { Arguments = [undefined] });
+        }, Throws.Nothing);
+    }
+
+    [Test]
+    public void CanCallFunctionWithArgumentNull()
+    {
+        var @null = new LocalValue.Null();
```
nvborisenko:
```suggestion
        var nullLocalValue = new LocalValue.Null();
```

dotnet/test/common/BiDi/Script/CallFunctionLocalValueTest.cs
```diff
@@ -0,0 +1,333 @@
...
+              }
+            }
+            """, false, new() { Arguments = [epochDate] });
+        }, Throws.Nothing);
+    }
+
+    [Test]
+    public void CanCallFunctionWithArgumentNumberFive()
+    {
+        var number5 = new LocalValue.Number(5);
```
nvborisenko:
```suggestion
        var numberLocalValue = new LocalValue.Number(5);
```

nvborisenko: Or just `arg` given that we are already in `LocalValue` context?

dotnet/test/common/BiDi/Script/CallFunctionLocalValueTest.cs
nvborisenko: Obsolete, `arg` is good now.

dotnet/test/common/BiDi/Script/CallFunctionRemoteValueTest.cs
```diff
@@ -0,0 +1,231 @@
...
+
+    [Test]
+    public async Task CanCallFunctionAndReturnNumberNegativeZero()
+    {
+        var response = await context.Script.CallFunctionAsync("() => { return -0; }", false);
+
+        Assert.That(response.Result, Is.AssignableTo<RemoteValue.Number>());
+
+        var expectedNegativeZero = ((RemoteValue.Number)response.Result).Value;
+        Assert.That(IsNegativeZero(expectedNegativeZero));
```
nvborisenko:
```csharp
Assert.That(expectedNegativeZero, Is.EqualTo(0));
Assert.That(double.IsNegative(expectedNegativeZero), Is.True);
```

And please rename `expectedNegativeZero`  to `actualNumberValue`

RenderMichael: Fixed

nvborisenko: And please fix even further.

RenderMichael: That’s my bad, I didn’t read fully the first time 😀 

Fixed to your recommendation

dotnet/test/common/BiDi/Script/CallFunctionRemoteValueTest.cs
```diff
@@ -0,0 +1,225 @@
...
+        var response = await context.Script.CallFunctionAsync($$"""
+            () => {
+              const set = new Set();
+              set.add('setKey');
+              return set;
+            }
+            """, false);
+
+        Assert.That(response.Result, Is.AssignableTo<RemoteValue.Set>());
+        Assert.That(((RemoteValue.Set)response.Result).Value, Is.EqualTo(expected.Value));
```
nvborisenko, line 240: Here I don't trust to "magic of records". Let's be simple: assert this, assert that.

nvborisenko: Ah, again I am 50/50, "magic of records" works

RenderMichael: I removed the magic from the primitive types (undefined, null, bool, number, string).

However, I kept it for all the other types: they all have other properties. We can verify not just the value, but also the handle and internal ID. Records + NUnit = full validation.

RenderMichael:
Formatting build failure unrelated:

```diff
--- a/py/test/selenium/webdriver/common/selenium_manager_tests.py
+++ b/py/test/selenium/webdriver/common/selenium_manager_tests.py
```

Test failures not related:

```bazel
//py:common-chrome-bidi-test/selenium/webdriver/common/bidi_tests.py
//py:common-edge-bidi-test/selenium/webdriver/common/bidi_tests.py
//rb/spec/integration/selenium/webdriver:action_builder-chrome-remote
//rb/spec/integration/selenium/webdriver:driver-chrome
//rb/spec/integration/selenium/webdriver:driver-chrome-remote
//rb/spec/integration/selenium/webdriver:driver-firefox-beta
//rb/spec/integration/selenium/webdriver:manager-chrome
//rb/spec/integration/selenium/webdriver:select-chrome
//rb/spec/integration/selenium/webdriver:select-chrome-remote
```

nvborisenko (approved):
Good test coverage, thanks!

PR #15400, opened 2025-03-11
[dotnet] [bidi] Add remaining `PrimitiveProtocolValue`s, simplify conversation for primitives.

RenderMichael:
### Description

- Add remaining `PrimitiveProtocolValue`
  - `RemoteValue.BigInt`
  - `LocalValue.BigInt`
  - `LocalValue.Boolean`
- Add a number of helpers to convert between .NET types and BiDi types

### Motivation and Context

We were missing some `PrimitiveProtocolValue`s: BigInt and Boolean on the local side, and BigInt on the remote side.

### Types of changes

### Checklist

- [ ] I have added tests to cover my changes. // Will update after similar PR gets merged

### DIFF
--- a/dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs
+++ b/dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs
@@ -117,7 +117,8 @@ public abstract record LocalValue
                 return new Array(node.AsArray().Select(FromNode));
 
             case JsonValueKind.Object:
-                return new Map(node.AsObject().ToDictionary(m => m.Key, m => FromNode(m.Value)));
+                Dictionary<string, LocalValue> values = node.AsObject().ToDictionary(node => node.Key, node => FromNode(node.Value));
+                return Object.FromDictionary(values);
 
             default:
                 throw new InvalidCastException($"Could not convert node {node}");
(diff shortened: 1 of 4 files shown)


dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs
```diff
@@ -70,14 +79,85 @@ public static LocalValue ConvertFrom(object? value)
...
+        {
+            Value = value;
+        }
+
+        [JsonConstructor]
+        internal BigInt(string valueAsString)
+        {
+            Value = BigInteger.Parse(valueAsString);
+        }
+    }
```
RenderMichael, line 152: BigInt fits naturally into .NET's native `BigInteger` type, but it needs to get serialized as a string across BiDi

RenderMichael: Other option:
```C#
public record BigInt(string Value) : PrimitiveProtocolLocalValue
{
    public static BigInt FromBigInteger(BigInteger value)
    {
        return new BigInt(value.ToString());
    }
}
```

nvborisenko:
So many conceptual stuff here to be analyzed.

RenderMichael:
Still WIP 😄

nvborisenko:
As in related PR https://github.com/SeleniumHQ/selenium/pull/15395 I propose to gather json payloads to see how it is implemented by vendors to choose the best way how to handle it.

PR #15409, opened 2025-03-12
Add Logging and Debugging to Investigate RBE Failures

titusfortner:
~I thought most of the failures we have in trunk are related to CDP/browser/driver version mismatches, but that doesn't appear to be true~

~This is WIP trying to make changes to get tests passing before release. Making it a PR for visibility and so it is easier to track.~

I've merged CDP and browser updates to trunk, but there are still failures.

This PR is now focused on getting more information about the RBE results so I can better figure out how to address the Ruby failures.

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -61,11 +61,16 @@ on:
         required: false
         type: string
         default: ''
-      artifact-name:
+      upload-name:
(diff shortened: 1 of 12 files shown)


titusfortner:
I fixed the things that need to get fixed for next release, I'll do a clean PR with the logging and debugging code later

Issue #15329, opened 2025-02-24
[🚀 Feature]: [dotnet] Provide a possibility to implement BiDi Permissions module


nvborisenko:
### Feature and motivation

https://www.w3.org/TR/permissions/

The implementation should not rely on `internal` members. The goal is to expose minimal required to make it possible to implement new extendable by the third-partied modules. 

### Usage example

```csharp
var bidi = await driver.AsBiDiAsync(); // selenium responsibility

// straightforward API (not yet approved)
var permissionsModule = new PermissionsModule(bidi);

// or as extension to emphasize this is extension
var permissionsModule = bidi.AsPermissions();

await permissionsModule.QueryAsync();
```

nvborisenko:
I am thinking more about it and coming to a conclusion that "reusing" existing BiDi connection is "not possible". `JsonSerializerContext` is **immutable**, reading it as third-party module should initialize his own connection (`Broker`) providing described json context.

So, I think the following will be more appreciated:
```csharp
var bidi = await driver.AsBiDiAsync(); // W3C WebDriver implementation

var permissions = await driver.AsBiDiPermissionsAsync(); // It initializes new Broker/WebSocketConnection/JsonContext
```

Then each `bidi` and `permissions` are isolated, and both have own "context" including everything needed to deal with remote end (still immutable). The trick is that remote end send data (responses/events) per WebSocketConnection.

I think this is a good solution. But semantic of this solution is not related to extensibility.

nvborisenko:
> But semantic of this solution is not related to extensibility.

And I am entirely OK with it! `Permissions` module can be easily implemented (defining his own DTO types), `Bluetooth` module too.

Appium is a separate topic, but seems yes, we consider Appium as yet another BiDi module (with his own isolated "context").

nvborisenko:
```csharp
var bidi = await driver.AsBiDiAsync(opts => opts.WithPermissions(out var permissions));

await permissions.QueryAsync();
```

This way we inject JsonTypeInfoResolver, reusing entire core bidi infrastructure, and expose `permissions` object with methods related  to Permissions only. This way also guarantees there is no performance impact even if we will have 100 attached extensions.

Researching continued. @RenderMichael what do you think?

RenderMichael:
I think the configuration builder would work well, the main drawback is that it requires all of BiDi for an extension. If we want to have Permissions separate from BiDi, then we would need a separate flow.

nvborisenko:
[...]

PR #15420, opened 2025-03-14
[dotnet] [bidi] Provide extension points for custom BiDi modules

RenderMichael:
### Description

Enables the right extensibility for users to create their own Modules, and hook them into `BiDi`.

Example implementation of `Permissions` module:

```C#
public class PermissionsBiDi : BiDi
{
    private readonly Lazy<PermissionsModule> _permissionsModule;

    public PermissionsModule Permissions => _permissionsModule.Value;

    private PermissionsBiDi(string url)
        : base(url)
    {
        Broker.ProvideCustomSerializationContext(PermissionsModule.SerializerContext);
        _permissionsModule = new Lazy<PermissionsModule>(() => new PermissionsModule(Broker));
    }

    public static async Task<PermissionsBiDi> ConnectAsync(string webSocketUrl)
    {
        var bidi = new PermissionsBiDi(webSocketUrl);
        await bidi.Broker.ConnectAsync();
        return bidi;
    }
}

public class PermissionsModule(Broker broker) : Module(broker)
{
    public static JsonSerializerContext SerializerContext => PermissionsBiDiJsonSerializerContext.Default;

    public async Task SetPermissionAsync(string permissionName, PermissionState state, string origin, UserContext? userContext, SetPermissionOptions? options = null)
    {
        var @params = new SetPermissionCommandParameters(new PermissionDescriptor(permissionName), state, origin, userContext);

        await Broker.ExecuteCommandAsync(new SetPermissionCommand(@params), options).ConfigureAwait(false);
    }
}

public record SetPermissionOptions : CommandOptions;

internal class SetPermissionCommand(SetPermissionCommandParameters @params)
    : Command<SetPermissionCommandParameters>(@params, "permissions.setPermission") { }

internal record SetPermissionCommandParameters(PermissionDescriptor Descriptor, PermissionState State, string Origin, UserContext? UserContext) : CommandParameters { }

internal record PermissionDescriptor(string Name);

public enum PermissionState
{
    Granted,
    Denied,
    Prompt
}

[JsonSerializable(typeof(SetPermissionCommand), TypeInfoPropertyName = "Permissions_SetPermissionCommand")]
internal partial class PermissionsBiDiJsonSerializerContext : JsonSerializerContext;
```

Or, if user does not care about AOT-safe JSON:
<details>
<summary>Sample Permissions module</summary>

```C#
public class PermissionsBiDi : BiDi
{
    private readonly Lazy<PermissionsModule> _permissionsModule;

[...]

PR #15421, opened 2025-03-14
[dotnet] [bidi] Implement Permissions extension module

RenderMichael:
Implements the Permissions WebDriver BiDi module.

Reliant on https://github.com/SeleniumHQ/selenium/pull/15420

### Motivation and Context

Fixes https://github.com/SeleniumHQ/selenium/issues/15329

### Types of changes

### Checklist

### DIFF
--- a/dotnet/src/webdriver/BiDi/Modules/Script/ScriptEvaluateException.cs
+++ b/dotnet/src/webdriver/BiDi/Modules/Script/ScriptEvaluateException.cs
@@ -30,6 +30,4 @@ public class ScriptEvaluateException(EvaluateResult.Exception evaluateResultExce
     public long ColumNumber => _evaluateResultException.ExceptionDetails.ColumnNumber;
 
     public override string Message => $"{Text}{Environment.NewLine}{_evaluateResultException.ExceptionDetails.StackTrace}";
-
-    public override string? StackTrace => base.StackTrace;
 }


RenderMichael:
Aside from providing the `Permissions` module, this PR also provides a `PermissionsBiDi` which enables a "BiDi + permissions extension" built-in.

This fulfills the ask in the issue to do:

```C#
var permissions = await driver.AsBiDiPermissionsAsync(); 
```

However, anyone can extend the `BiDi` type and add the `PermissionsModule` + any other extensions they would like.

dotnet/src/webdriver/BiDi/Extensions/Permissions/PermissionsBiDi.cs
```diff
@@ -0,0 +1,45 @@
...
+// under the License.
+// </copyright>
+
+using OpenQA.Selenium.BiDi.Communication;
+using System;
+using System.Threading.Tasks;
+
+namespace OpenQA.Selenium.BiDi.Extensions.Permissions;
+
+public class PermissionsBiDi : BiDi
```
nvborisenko, line 26: It should not derive from `BiDi`.

nvborisenko (changes requested):
I guess this one we don't want conceptually.

PR #15442, opened 2025-03-17
[dotnet] [bidi] Expand usability of `CallFunctionAsync<TResult>` and `EvaluateAsync<TResult>`

RenderMichael:
### Description

Greatly expands usability of two methods which de-serialize BiDi types to .NET types

Test suite demonstrates the full extent, here is an example

```C#
var response = await context.Script.CallFunctionAsync<dynamic>("() => { return { objKey: 'objValue' }; }", false);

Assert.That(response.objKey, Is.EqualTo("objValue"));
```

```C#
DateTime before = DateTime.Now;
var response = await context.Script.CallFunctionAsync<DateTime?>("() => { return { new Date() }; }", false);
DateTime after = DateTime.Now;

Assert.That(response, Is.InRange(before, after));
```

### Motivation and Context

The two methods can be much more usable

### Types of changes

### Checklist


RenderMichael:
Per conversation on Slack:

> AOT-safe?

Yes, this is AOT-safe. It doesn't root any types that wouldn't certainly be rooted in else where, and patterns like `typeof(T) == typeof(int)` are special-cased in the compiler to trim away. So `ConvertTo<int>` will be very optimized.

PR #15499, opened 2025-03-24
[java] Use static Patterns for regex-matching

zodac-personal:
### Motivation and Context
There were several locations where `Pattern` objects were being used to perform regex matches. This is inefficient as the `Pattern` would be recompiled on each call. I've extracted these to constants in the appropriate class to ensure they are only compiled once. In one case there was a regex being used to check for digits, which I've replaced with a simple non-regex based function which should be less complex.

### Types of changes
- [x] Refactoring (no changes to public API)

### Checklist
- [x] All new and existing tests passed. (tested with `bazel test //java/... --test_size_filters=small`)


java/src/org/openqa/selenium/grid/commands/InfoCommand.java
```diff
@@ -109,7 +112,7 @@ public Executable configure(PrintStream out, PrintStream err, String... args) {
           break;
       }
 
-      String path = getClass().getPackage().getName().replaceAll("\\.", "/") + "/" + toDisplay;
+      String path = getClass().getPackage().getName().replace('.', '/') + "/" + toDisplay;
```
pujagani, line 115: Why is this changed? Just curious

java/src/org/openqa/selenium/grid/commands/InfoCommand.java
```diff
@@ -143,11 +146,11 @@ private String readContent(String path) throws IOException {
         } else if ("```".equals(line)) {
           inCode = !inCode;
         } else {
-          if (line.startsWith("=")) {
+          if (line.charAt(0) == '=') {
```
pujagani, line 149: Any reason for modifying this?

java/src/org/openqa/selenium/PersistentCapabilities.java
```diff
@@ -58,7 +56,7 @@ public PersistentCapabilities setCapability(String name, Object value) {
   @Override
   public Map<String, Object> asMap() {
     return getCapabilityNames().stream()
-        .collect(toUnmodifiableMap(Function.identity(), this::getCapability));
+        .collect(Collectors.toUnmodifiableMap(Function.identity(), this::getCapability));
```
pujagani: This change makes sense. But these are unrelated to changes made as part of the rest of this PR. Can we please separate this out into a separate PR?

pujagani (changes requested):

java/src/org/openqa/selenium/grid/commands/InfoCommand.java
zodac-personal: Similar to the other changes to make the `Pattern` instances static. The `#replaceAll(String, String)` method considered the first argument to be a regex, meaning it needs to be compiled for each call.

But since this call is just matching a literal dot, we can replace it with `#replace(char, char)` which will iterate through the characters in the String and replaces them.

java/src/org/openqa/selenium/grid/commands/InfoCommand.java
zodac-personal: Minor performance improvement, while I'm in the same class. The `#startsWith(String)` method can take a String of multiple characters, so it needs to search through the String. Whereas if you're only searching for a single character, `charAt(int)` directly accesses the char[] backing the String and returns the value in a single call.

Not a `Pattern`, so I can move it into another PR if you think that's appropriate.

java/src/org/openqa/selenium/PersistentCapabilities.java
zodac-personal: That's fine, I'll create a separate PR for these changes.

diemol (approved):
Thank you, @zodac

PR #15527, opened 2025-03-28
[dotnet] Fully annotate `Command` for AOT support

RenderMichael:
### **User description**
AOT warnings are user-facing, even if those warnings are within Selenium's own code. For that reason, we should ensure we are not emitting any warnings unless such code paths are annotated as AOT-unsafe.

### Motivation and Context

Contributes to https://github.com/SeleniumHQ/selenium/issues/14480

### Types of changes

### Checklist

___

### **PR Type**
Enhancement

___

### **Description**
- Added AOT-friendly annotations to the `Command` class.

- Refactored `JsonSerializerOptions` into a static nested class for safety.

- Introduced helper methods for parameter handling in `Command`.

- Updated `HttpCommandInfo` to use new parameter handling methods.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>Command.cs</strong><dd><code>Refactor `Command` class for AOT compatibility</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/Command.cs

<li>Added AOT and trimming annotations to methods and properties.<br> <li> Refactored <code>JsonSerializerOptions</code> into a static nested class.<br> <li> Introduced <code>HasParameters</code> and <code>TryGetValueAndRemoveIfNotNull</code> helper <br>methods.<br> <li> Updated parameter handling logic for better AOT support.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15527/files#diff-713ff04e841a84c186ba6acf6d405ad2b3eed93d2e850a3634cf54fb7edb3ae2">+50/-12</a>&nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>HttpCommandInfo.cs</strong><dd><code>Update parameter handling in `HttpCommandInfo`</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/HttpCommandInfo.cs

<li>Replaced direct parameter access with <code>HasParameters</code> and <br><code>TryGetValueAndRemoveIfNotNull</code>.<br> <li> Improved parameter extraction logic for AOT compatibility.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15527/files#diff-87be2d36d1439aa92c51841738358cfc6a1ad2b9d6e2e6d31ca1f6d31dcd8bd3">+3/-7</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

___


dotnet/src/webdriver/Command.cs
```diff
@@ -31,24 +32,33 @@ namespace OpenQA.Selenium
     /// </summary>
     public class Command
```
RenderMichael: Best viewed with [whitespace off](https://github.com/SeleniumHQ/selenium/pull/15527/files?w=1)

dotnet/src/webdriver/Command.cs
```diff
@@ -31,24 +32,40 @@ namespace OpenQA.Selenium;
 /// </summary>
 public class Command
 {
-    private readonly static JsonSerializerOptions s_jsonSerializerOptions = new()
+    [UnconditionalSuppressMessage("Trimming", "IL2026", Justification = $"All trimming-unsafe access points to {nameof(s_jsonSerializerOptions)} are annotated as such")]
+    [UnconditionalSuppressMessage("AOT", "IL3050", Justification = $"All AOT-unsafe access points to {nameof(s_jsonSerializerOptions)} are annotated as such")]
+    private static class JsonOptionsHolder
```
nvborisenko, line 37: Any reason to introduce dummy class? Just to hold some object used only in here? I think standard `field` is good.

dotnet/src/webdriver/WebDriver.cs
```diff
@@ -40,7 +40,7 @@ public class WebDriver : IWebDriver, ISearchContext, IJavaScriptExecutor, IFinds
     /// </summary>
     protected static readonly TimeSpan DefaultCommandTimeout = TimeSpan.FromSeconds(60);
     private IFileDetector fileDetector = new DefaultFileDetector();
-    private NetworkManager network;
+    private NetworkManager? network;
```
nvborisenko: Unintentional change?

dotnet/src/webdriver/Command.cs
```diff
@@ -81,18 +98,32 @@ public Command(SessionId? sessionId, string name, Dictionary<string, object?>? p
...
     /// <summary>
     /// Gets the parameters of the command as a JSON-encoded string.
     /// </summary>
     public string ParametersAsJsonString
     {
+        [UnconditionalSuppressMessage("Trimming", "IL2026", Justification = $"All trimming-unsafe access points to {nameof(_parameters)} are annotated as such")]
+        [UnconditionalSuppressMessage("AOT", "IL3050", Justification = $"All AOT-unsafe access points to {nameof(_parameters)} are annotated as such")]
         get
         {
-            if (this.Parameters != null && this.Parameters.Count > 0)
```
nvborisenko, line 93: `this.Parameters?.Count > 0` pretty simple to not introduce new method

dotnet/src/webdriver/Command.cs
```diff
@@ -81,18 +98,32 @@ public Command(SessionId? sessionId, string name, Dictionary<string, object?>? p
...
+            if (HasParameters())
             {
-                return JsonSerializer.Serialize(this.Parameters, s_jsonSerializerOptions);
+                try
+                {
+                    return JsonSerializer.Serialize(this._parameters, JsonOptionsHolder.s_jsonSerializerOptions);
+                }
+                catch (NotSupportedException ex)
+                {
+                    throw new WebDriverException("Attempted to serialize an unsupported type. Ensure you are using Selenium types, or well-known .NET types such as Dictionary<string, object> and object[]", ex);
```
nvborisenko, line 125: I thought we support any type, falling back to reflection based if not in the json context.

dotnet/src/webdriver/HttpCommandInfo.cs
```diff
@@ -119,17 +119,13 @@ private static string GetCommandPropertyValue(string propertyName, Command comma
                 propertyValue = commandToExecute.SessionId.ToString();
             }
         }
-        else if (commandToExecute.Parameters != null && commandToExecute.Parameters.Count > 0)
+        else if (commandToExecute.HasParameters())
         {
             // Extract the URL parameter, and remove it from the parameters dictionary
             // so it doesn't get transmitted as a JSON parameter.
-            if (commandToExecute.Parameters.TryGetValue(propertyName, out var propertyValueObject))
+            if (commandToExecute.TryGetValueAndRemoveIfNotNull(propertyName, out var propertyValueObject))
```
nvborisenko, line 126: Hm, interesting situation. Transmitter checks each time whether some "infected" parameter should be removed. What are these cases? I hope we can just fix it on consumer side. Based on comments, somebody tries to send `url`, thus we should fix this `somebody`.

dotnet/src/webdriver/Command.cs
RenderMichael: Fields cannot be AOT-annotated, only types and methods. Having a “holder” type is common when you want to annotate specific fields as AOT-unsafe but not everything in the type.

dotnet/src/webdriver/WebDriver.cs
RenderMichael: It’s intentional and correct (it is lazy-initialized) but not necessarily relevant to this PR. Just one more internal warning gone.

dotnet/src/webdriver/Command.cs
RenderMichael: `this.Parameters` is an AOT-unsafe member. We must handle the logic inside `Command` for the AOT annotations.

dotnet/src/webdriver/Command.cs
RenderMichael: Not if we are AOT-compiled. the reflection fallback will not work. That is the only case this `NotSupportedException` is thrown.

dotnet/src/webdriver/HttpCommandInfo.cs
RenderMichael: The case is for parameters in the URL. This extracts those values from the dictionary and uses them in the URL instead of the parameter body.

dotnet/src/webdriver/Command.cs
nvborisenko: Then we are not AOT-compatible. Users will don't see warnings at compile time, but will face issues at runtime. And what we?.. just suppress it. It is bad :(

I don't see good solution as a quick win, right solution would be avoid any `object` type and be strongly typed?

nvborisenko:
This is show stopper: https://github.com/SeleniumHQ/selenium/pull/15527#discussion_r2101241067

dotnet/src/webdriver/Command.cs
RenderMichael: No, this is the other way around!

We know that anyone could add any `object` to the command parameters. We’ve had all those problems until I made that PR with the reflection-based fallback. It was never AOT compatible.

This change will surface that AOT-incompatibility, and throw a good exception, but *only* when we are running in AOT.

The exception scenario: user compiles to AOT (not common), and adds a type to the command parameters that is not a common/known type (very, very uncommon).

The current state will silently work wrong (serialize to `{}` no matter what). With this change, the user sees a meaningful exception.

nvborisenko: That's true, the case is very rare (0.0001%). But previously user saw a warning something like "Selenium is not AOT compatible, you may face issues". And this warning is good for users. If we just suppress the warning, we lye to users.

PR #15532, opened 2025-03-28
[dotnet] [bidi] Add strongly-typed `LocalValue.ConvertFrom` overloads

RenderMichael:
### Description 
Add strongly-typed `LocalValue.ConvertFrom` overloads

### Motivation and Context

Prevents users from being forced to guess which types we happen to support.

### Types of changes

### Checklist

___

### **PR Type**
Enhancement

___

### **Description**
- Added strongly-typed `LocalValue.ConvertFrom` overloads for better type safety.

- Refactored `ConvertFrom` logic to delegate to specific type overloads.

- Introduced new methods for handling various data structures and primitives.

- Improved code maintainability and readability by reducing duplication.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>LocalValue.cs</strong><dd><code>Refactored and extended `LocalValue.ConvertFrom` functionality</code></dd></summary>
<hr>

dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs

<li>Added strongly-typed overloads for <code>ConvertFrom</code> method.<br> <li> Refactored implicit operators to use <code>ConvertFrom</code> methods.<br> <li> Introduced specialized methods for handling primitives and <br>collections.<br> <li> Enhanced maintainability by reducing redundant code and improving <br>structure.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15532/files#diff-b2d3bd83230255db2e435ac34700a0af0595469211636058e6eb477107171b22">+219/-63</a></td>

</tr>

___


dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs
```diff
@@ -41,10 +41,10 @@ namespace OpenQA.Selenium.BiDi.Modules.Script;
...
-    public static implicit operator LocalValue(int? value) { return value is int i ? new NumberLocalValue(i) : new NullLocalValue(); }
-    public static implicit operator LocalValue(double? value) { return value is double d ? new NumberLocalValue(d) : new NullLocalValue(); }
-    public static implicit operator LocalValue(string? value) { return value is null ? new NullLocalValue() : new StringLocalValue(value); }
+    public static implicit operator LocalValue(bool? value) { return ConvertFrom(value); }
+    public static implicit operator LocalValue(int? value) { return ConvertFrom(value); }
+    public static implicit operator LocalValue(double? value) { return ConvertFrom(value); }
+    public static implicit operator LocalValue(string? value) { return ConvertFrom(value); }
 
     // TODO: Extend converting from types
     public static LocalValue ConvertFrom(object? value)
```
nvborisenko, line 53: We already have unit tests for this method. I propose to create some universal test to demonstrate the power of conversion. Where user is able to properly convert any "free-style object" to `LocalValue` (including all supported complex types). The idea behind it is:
- double-check for us
- see all supported types in one place
- easily understand what we are missing

dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs
```diff
@@ -58,86 +58,210 @@ public static LocalValue ConvertFrom(object? value)
...
+    {
+        if (value is not null)
+        {
+            return new BigIntLocalValue(value.Value.ToString());
+        }
+
+        return new NullLocalValue();
+    }
+
+    public static LocalValue ConvertFrom(IEnumerable<object?>? value)
```
nvborisenko: Should be non-generic `IEnumerable`, then conversion will happen for `object` flow (via reflection). The same for `IDictionary` and `ISet`.

nvborisenko: Resolved, but there is no non-generic `ISet`. This is why we still see `public static LocalValue ConvertFrom<T>(ISet<T?>? value)`.

nvborisenko: It seems that this method is unnecessary since we have `LocalValue.ConvertFrom(object)`, but from user's point of view it is good to know that `LocalValue.ConvertFrom(new HashSet<string> { });` is supported.

dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs
nvborisenko: I fixed the conversion, including for `Set`. Now it is time to prepare tests.

dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs
```diff
@@ -58,86 +59,198 @@ public static LocalValue ConvertFrom(object? value)
...
+                return ConvertFrom(dictionary);
 
-                    return new MapLocalValue(bidiObject);
-                }
+            case IEnumerable enumerable:
+                return ConvertFrom(enumerable);
 
-            case IEnumerable<object?> list:
-                return new ArrayLocalValue(list.Select(ConvertFrom).ToList());
+            default:
```
nvborisenko, line 108: We are missing `RegEx` type, let's add it, seems not a big deal.

dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs
```diff
@@ -41,10 +42,10 @@ namespace OpenQA.Selenium.BiDi.Modules.Script;
 [JsonDerivedType(typeof(SetLocalValue), "set")]
 public abstract record LocalValue
```
nvborisenko, line 44: We are missing `script.RemoteReference` type (https://w3c.github.io/webdriver-bidi/#type-script-LocalValue)

dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs
nvborisenko: Added minimal tests, let's go.

dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs
```diff
@@ -58,86 +59,198 @@ public static LocalValue ConvertFrom(object? value)
...
+    {
+        if (value is not null)
+        {
+            return new StringLocalValue(value);
         }
+
+        return new NullLocalValue();
+    }
+
+    public static LocalValue ConvertFrom(DateTime? value)
```
nvborisenko: `DateTimeOffset`

dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs
nvborisenko: Will be resolved in https://github.com/SeleniumHQ/selenium/issues/15621

dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs
nvborisenko: Played around it and understood we have to know a conversion from `DateTime/DateTimeOffset` to JS string.

nvborisenko:
```csharp
await context.Script.CallFunctionAsync("() => { return Date.now(); }", false)
```

returns `EvaluateResultSuccess { Result = NumberRemoteValue { Value = 1744486840254 }, Realm = OpenQA.Selenium.BiDi.Modules.Script.Realm }`

unix epoch?

But `DateLocalValue` expects `string`. No ideas, we need help.

nvborisenko: https://tc39.es/ecma262/#sec-date-time-string-format

RenderMichael: `Date.now()` returns a number, you have to do `new Date()` to get the current date as a `Date` type.

https://github.com/SeleniumHQ/selenium/blob/12a1593042c3cd9c40de46cfdd1513ecfcfcd20a/dotnet/test/common/BiDi/Script/CallFunctionRemoteValueTest.cs#L81-L101

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#the_epoch_timestamps_and_invalid_date

RenderMichael: I prefer to avoid an implementation for `DateTimeOffset`, because I don't think we can fully represent it as `DateLocalValue`. For the same reason, casting `DateTimeOffset` -> `DateTime` requires an explicit cast, because you lose the offset/timezone info. Users should take care of that,

dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs
RenderMichael: Added `Regex` support. It is not trivial because of flags, I did the best I could mapping one-to-one.

dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs
nvborisenko: I understood we can use any .net type for `LocalValue`, even `DateOnly`. In future we can easily add supporting of all o them, just to serialize it to `string` properly. Let's land with `DateTimeOfffset` for now, because of I used it in another BiDi places, not only for `LocalValue`.

Fixed. We always can improve later.

dotnet/src/webdriver/BiDi/Modules/Script/LocalValue.cs
nvborisenko: Thank you, the conversion code looks hard, but I trust you. Minor requested feature - we will add tests later.

nvborisenko:
Our big concern here is whether we should support nullable primitive types. I guess it is more philosophic question, from implementation perspective for us it doesn't matter. For end users it is also kind of "doesn't matter".

nvborisenko (approved):

nvborisenko:
Thank you, Mike! Let's move on and improve later when necessary.

PR #15576, opened 2025-04-04
[py] Fixing flaky locator test in Firefox

shbenzer:
### 💥 What does this PR do?
Fixed flaky firefox test
added xfail for test that only fails in firefox

### 🔄 Types of changes


cgoldberg:
I made some adjustments to the wait times in `webdriverwait_tests.py` to reduce flakiness (see #15650), so we probably don't need this `xfail` marker anymore.

Do we still need the other XPATH change in this PR? (that test seems to work for me locally, but I did notice a failure from that test file in CI recently)

shbenzer:
Honestly the  XPATH fix only made it slightly less flaky (and tbf that could be confirmation bias) - I don't think the changes in this PR are worth merging yet

PR #15582, opened 2025-04-06
DRAFT [py]: increase fedcm tests timeout 

navin772:
### 💥 What does this PR do?

Debug FedCM timeout error on Linux

### 🔄 Types of changes


cgoldberg:
I can't get the dialog to open at all, even with `driver.fedcm_dialog(timeout=60)`

navin772:
@cgoldberg yeah, it's weird that its happening for those 2 tests only since the chrome update (and only on linux!)

cgoldberg:
@navin772 after more testing, I actually can get it to open.. it just takes a little bit.  I think raising the timeout to 15 secs should resolve this.

navin772:
I raised it to 20 sec for testing. Although, for me the dialog appears almost instantly in linux. Let's see if this passes else we might have to skip those 2 for now and test in later versions.

cgoldberg:
This PR shouldn't be needed now.. We can land #15583 instead.

PR #15625, opened 2025-04-13
[py] Skip bidi tests on browsers that don't support bidi

cgoldberg:
### 💥 What does this PR do?

This PR adjusts the internal Python PyTest configuration so tests are skipped when `--bidi` is enabled on browsers that don't support BiDi (`ie`, `safari`, `webkitgtk`, `wpewebkit`).

This removes the need to mark BiDi tests with `xfail` markers on browsers where they are expected to fail.

### 🔄 Types of changes

___

### **PR Type**
Bug fix

___

### DIFF
--- a/py/conftest.py
+++ b/py/conftest.py
@@ -126,6 +126,11 @@ def driver(request):
     if "WebKit" in driver_class and _platform == "Windows":
         pytest.skip("WebKit tests cannot be run on Windows")
 
+    # skip tests for drivers that don't support BiDi when --bidi is enabled
+    if request.config.option.bidi:
+        if driver_class in ("Ie", "Safari", "WebKitGTK", "WPEWebKit"):
+            pytest.skip(f"{driver_class} does not support BiDi")
+
     # conditionally mark tests as expected to fail based on driver
     marker = request.node.get_closest_marker(f"xfail_{driver_class.lower()}")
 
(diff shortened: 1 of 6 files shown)


navin772:
Can we have a common marker for Safari, WebKitGTK and WPEWebKit apart from their individual ones which we can use to skip tests? IMO, using that many xfail marker makes the test look not so elegant.

cgoldberg:
I can create one like `xfail_nobidi` or something. I'll update this PR later today.

cgoldberg:
@navin772 check out these changes... I updated the PR description also.

Now you don't need to use `xfail` markers at all on BiDi tests. Instead, tests will be skipped if you run them with `--bidi` on a browser/driver that doesn't support BiDi.

navin772:
This looks much better, can you also add a note in the testing guide for the same? It will help contributors understand which drivers are being skipped.

cgoldberg:
@navin772 Which testing guide do you mean? In the main `README.md` of the repo where the testing commands are described? There really isn't any Python-specific info about running tests in there (maybe there should be).
It's also pretty clear when and why the tests are skipped. You will see:
```
SKIPPED (WebKitGTK does not support BiDi) 
```
I am going to merge this PR, but I will update any documentation that you think this should be mentioned in.

PR #15628, opened 2025-04-14
[py]Fix: Timeout env variable GLOBAL_DEFAULT_TIMEOUT has no effect

### LINKED ISSUE #15604: [🐛 Bug]: Timeout env variable GLOBAL_DEFAULT_TIMEOUT has no effect
### Description

The timeout is always set to 120 seconds, the env variable has GLOBAL_DEFAULT_TIMEOUT has no effect. 

[https://github.com/SeleniumHQ/selenium/blob/ecb0dbf91415b10d718acf28afefc16654a9f13a/py/selenium/webdriver/chromium/remote_connection.py#L33](url)

`client_config = client_config or ClientConfig(
            remote_server_addr=remote_server_addr, keep_alive=keep_alive, timeout=120
        )`  

[https://github.com/SeleniumHQ/selenium/blob/ecb0dbf91415b10d718acf28afefc16654a9f13a/py/selenium/webdriver/remote/client_config.py#L107](url)
`self.timeout = (
            (
                float(os.getenv("GLOBAL_DEFAULT_TIMEOUT", str(socket.getdefaulttimeout())))
                if os.getenv("GLOBAL_DEFAULT_TIMEOUT") is not None
                else socket.getdefaulttimeout()
            )
            if timeout is None
            else timeout
        )`

Others:
[https://github.com/SeleniumHQ/selenium/blob/ecb0dbf91415b10d718acf28afefc16654a9f13a/py/selenium/webdriver/firefox/remote_connection.py#L35](url)
[https://github.com/SeleniumHQ/selenium/blob/ecb0dbf91415b10d718acf28afefc16654a9f13a/py/selenium/webdriver/safari/remote_connection.py#L35](url)

### Reproducible Code

```shell
driver.get("https://github.com/")
driver.implicitly_wait(130)
field = driver.find_element(By.ID, "qwertyzxcvb")
```

### Debugging Logs

```logs
/site-packages/urllib3/connectionpool.py", line 369, in _raise_timeout
    raise ReadTimeoutError(
urllib3.exceptions.ReadTimeoutError: HTTPConnectionPool(host='localhost', port=123): Read timed out. (read timeout=120)
```

XueSongTap:
### 🔗 Related Issues
Fixes #15604

### 💥 What does this PR do?
This PR fixes an issue where the `GLOBAL_DEFAULT_TIMEOUT` environment variable was being ignored when setting timeout values in Selenium's Python bindings. The current implementation always uses a hardcoded timeout value (120 seconds) in the remote connection classes, making the environment variable ineffective.

### 🔧 Implementation Notes
The implementation has been simplified to properly prioritize timeout sources:
1. Use `GLOBAL_DEFAULT_TIMEOUT` if it's set in the environment
2. Otherwise, use the explicitly provided timeout parameter
3. Fall back to `socket.getdefaulttimeout()` if neither is available

This same logic was also applied to the `reset_timeout()` method to ensure consistent behavior throughout the client lifecycle.

This fix restores expected behavior without changing the API. Users who set the `GLOBAL_DEFAULT_TIMEOUT` environment variable will now see it properly respected.

### 🔄 Types of changes

___

### **PR Type**
Bug fix

___

### **Description**
- Fixes the `GLOBAL_DEFAULT_TIMEOUT` environment variable being ignored.

- Adjusts timeout logic to prioritize environment variable correctly.

- Ensures consistent behavior in `reset_timeout` method.

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>client_config.py</strong><dd><code>Fix timeout logic to respect environment variable</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/selenium/webdriver/remote/client_config.py

<li>Updated timeout logic to prioritize <code>GLOBAL_DEFAULT_TIMEOUT</code>.<br> <li> Modified <code>reset_timeout</code> to respect <code>GLOBAL_DEFAULT_TIMEOUT</code>.<br> <li> Simplified conditional checks for timeout assignment.

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15628/files#diff-421b55d14e80d4473c60003a3f53c9fa819375d52cb45b6c40403f481a9b8925">+8/-8</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

___

### DIFF
--- a/py/selenium/webdriver/remote/client_config.py
+++ b/py/selenium/webdriver/remote/client_config.py
@@ -105,13 +105,9 @@ class ClientConfig:
         self.extra_headers = extra_headers
 
         self.timeout = (
-            (
-                float(os.getenv("GLOBAL_DEFAULT_TIMEOUT", str(socket.getdefaulttimeout())))
-                if os.getenv("GLOBAL_DEFAULT_TIMEOUT") is not None
-                else socket.getdefaulttimeout()
-            )
-            if timeout is None
-            else timeout
+            float(os.getenv("GLOBAL_DEFAULT_TIMEOUT"))
+            if os.getenv("GLOBAL_DEFAULT_TIMEOUT") is not None
+            else (timeout if timeout is not None else socket.getdefaulttimeout())
         )
 
         self.ca_certs = (
@@ -122,7 +118,11 @@ class ClientConfig:
 
     def reset_timeout(self) -> None:
         """Resets the timeout to the default value of socket."""
-        self._timeout = socket.getdefaulttimeout()
+        self._timeout = (
+            float(os.getenv("GLOBAL_DEFAULT_TIMEOUT"))
+            if os.getenv("GLOBAL_DEFAULT_TIMEOUT") is not None
+            else socket.getdefaulttimeout()
+        )
 
     def get_proxy_url(self) -> Optional[str]:
         """Returns the proxy URL to use for the connection."""
--- a/py/test/unit/selenium/webdriver/remote/client_config_tests.py
+++ b/py/test/unit/selenium/webdriver/remote/client_config_tests.py
@@ -1,0 +1,109 @@
+# Licensed to the Software Freedom Conservancy (SFC) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The SFC licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+import os
+import socket
+
+from selenium.webdriver.remote.client_config import ClientConfig
+
+
+def test_timeout_from_env_variable():
+    """Test that timeout is set from GLOBAL_DEFAULT_TIMEOUT environment variable."""
+    original_env = os.environ.get("GLOBAL_DEFAULT_TIMEOUT")
+    try:
+        os.environ["GLOBAL_DEFAULT_TIMEOUT"] = "45"
+        config = ClientConfig(remote_server_addr="http://localhost:4444")
+        assert config.timeout == 45
+    finally:
+        if original_env is None:
+            del os.environ["GLOBAL_DEFAULT_TIMEOUT"]
+        else:
+            os.environ["GLOBAL_DEFAULT_TIMEOUT"] = original_env
+
+
+def test_timeout_from_parameter():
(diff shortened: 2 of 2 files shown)


cgoldberg (changes requested):
There are 2 scenarios I can think of that we should also consider (and have tests for):

- if the environment variable is a number less than zero (i.e. `GLOBAL_DEFAULT_TIMEOUT=-10.0`), it will use this value, which will lead to an error.

- if the environment variable can't be converted to a float (i.e. `GLOBAL_DEFAULT_TIMEOUT=bad`), it doesn't raise a helpful error.

I think in both of these circumstances, we should just raise a `ValueError` with a useful message. Something like: `ValueError: GLOBAL_DEFAULT_TIMEOUT must be a number >= 0`

Also.. in the tests, you might want to mock `os.environ` rather than modifying it directly. It probably doesn't matter, because you are always setting it back to the original value in a `finally` block, but this is a nice pattern we use in other tests.  Here is an example:

https://github.com/SeleniumHQ/selenium/blob/f457f15ae63a4a6237f8463bd05c1ab7d0061188/py/test/selenium/webdriver/chrome/chrome_service_tests.py#L130

cgoldberg:
@XueSongTap Thanks for the PR!... this looks really good.

Please see my review comments, and also don't forget to sign the CLA as mentioned here:

https://github.com/SeleniumHQ/selenium/pull/15628#issuecomment-2802406152

Also, take a look at the code suggestions that the AI bot left in earlier comments.

XueSongTap:
Hi @Barancev, thank you for your helpful review!

You're absolutely right that reading the environment variable at module/class construction time makes it static and insensitive to any later changes to `GLOBAL_DEFAULT_TIMEOUT`.

To address this, I'm planning to refactor the timeout logic in the following way:

- Instead of evaluating the timeout once in `__init__`, I will expose it via a `@property` method, something like:

```python
@property
def timeout(self):
    env_value = os.getenv("GLOBAL_DEFAULT_TIMEOUT")
    if env_value is not None:
        return float(env_value)
    if self._explicit_timeout is not None:
        return self._explicit_timeout
    return socket.getdefaulttimeout()
```

- The constructor will keep `_explicit_timeout` to capture any user-specified value.
- Calling `reset_timeout()` will simply clear `_explicit_timeout`, allowing fallback to either the env variable or socket default.

This way, changes to `GLOBAL_DEFAULT_TIMEOUT` at runtime will be respected dynamically, which should make the behavior more predictable and flexible.

Would this approach align with your expectations? Let me know what you think before I proceed with the changes.

Thanks again!

titusfortner (changes requested):
Thanks for the PR, but let's wait on deciding that this is really what we want to do in #https://github.com/SeleniumHQ/selenium/issues/15604

cgoldberg (changes requested):
@XueSongTap We have decided to remove the use of the `GLOBAL_DEFAULT_TIMEOUT` environment variable (see discussion in #15604). Can you do that in your branch? We want to completely remove any reference to it, and keep the other logic as you have it.

cgoldberg:
@XueSongTap I didn't see any recent activity on this, so I submitted #15673 to address the decision in #15604. 

Thank you very much for your effort anyway.

PR #15660, opened 2025-04-23
[java][BiDi] implement web extensions

Delta456:
### 🔗 Related Issues

Implements https://github.com/SeleniumHQ/selenium/issues/15585

### 💥 What does this PR do?

Implements https://github.com/SeleniumHQ/selenium/issues/15585 for Java binding

### 🔧 Implementation Notes

The PR will fail on stable firefox and needs firefox to be 138 version as stated in https://github.com/SeleniumHQ/selenium/issues/15585#issuecomment-2782657812

### 🔄 Types of changes

- New feature (non-breaking change which adds functionality *and tests!*)

___

### **PR Type**
Enhancement, Tests

___

### **Description**
- Add BiDi web extension install/uninstall API for Java
  - Implements `webExtension.install` and `webExtension.uninstall` commands
  - Supports extension install via path, archive, or base64

- Introduce new classes for extension data handling

- Add comprehensive tests for extension installation methods

- Update Bazel build files for new module and test integration

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>7 files</summary><table>
<tr>
  <td><strong>ExtensionArchivePath.java</strong><dd><code>Add class for extension install via archive path</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15660/files#diff-3d3c3941a4332e633b21e3dbbf998257a45b2e5bf6f5c8b15c99ac200b2cef25">+34/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>ExtensionBase64Encoded.java</strong><dd><code>Add class for extension install via base64 encoding</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15660/files#diff-b81e32cc4dd6b8cfe7bfa17e99be45c07789552e7836b09dc182bdce45432972">+34/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
[...]

PR #15749, opened 2025-05-19
[py][bidi]: add bidi webExtension module

navin772:
### 💥 What does this PR do?

Adds the BiDi webExtension module to the python bindings - https://w3c.github.io/webdriver-bidi/#module-webExtension

### 🔄 Types of changes

___

### **PR Type**
Enhancement, Tests

___

### **Description**
- Add BiDi webExtension module to Python bindings

- Implement install and uninstall methods for web extensions

- Expose webextension property in Remote WebDriver

- Add comprehensive tests for webExtension functionality

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>webextension.py</strong><dd><code>Add BiDi WebExtension module with install/uninstall methods</code></dd></summary>
<hr>

py/selenium/webdriver/common/bidi/webextension.py

<li>Introduces new WebExtension class for BiDi protocol<br> <li> Implements install and uninstall methods for web extensions<br> <li> Provides parameter validation and command execution logic

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15749/files#diff-f6467ecd5e2e38f1ba13183bb5fae9f304b5edba8c9acb8855e5572ac5a6e71a">+74/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>webdriver.py</strong><dd><code>Integrate webExtension property into Remote WebDriver</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

<li>Imports and initializes the new WebExtension class<br> <li> Adds webextension property to expose BiDi webExtension commands<br> <li> Updates internal attributes to support webExtension<br> <li> Documents usage with examples in the property docstring

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15749/files#diff-c87c150ed2a08f5293db58a06da8bf7f14c3a5582586c6270c1be661dfdd0985">+24/-0</a>&nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>bidi_webextension_tests.py</strong><dd><code>Add tests for BiDi webExtension install/uninstall</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/test/selenium/webdriver/common/bidi_webextension_tests.py

<li>Adds tests for BiDi webExtension module functionality<br> <li> Tests install/uninstall via path, archive, and base64<br> <li> Includes fixture for locating extension files<br> <li> Marks tests as expected to fail on Chrome and Edge

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15749/files#diff-6e579bdd2c5a63ae03a0572188781b474610491e5b14c4ea1eaf65dc35750851">+98/-0</a>&nbsp; &nbsp; </td>

</tr>

___


cgoldberg (commented):
Nice. This looks really good.

- The tests should `xfail` on the browsers besides Chrome/Edge I think (safari, ie, webkitgtk, wpewebkit)

- Should the tests also load a page and verify the extension is enabled like this?:
https://github.com/SeleniumHQ/selenium/blob/4768f5a70d92e709f5a8816284154d3f1ac7a3b7/py/test/selenium/webdriver/firefox/ff_installs_addons_tests.py#L35

navin772:
@cgoldberg yes, we should verify after loading a page. I will update the tests. Also, found that using absolute path for the extensions is much better from the ff_addon test file.

Regarding xfail markers for safari and other non-bidi browsers, I think we already skip them here:

https://github.com/SeleniumHQ/selenium/blob/a413faa65ab11e77725886af23141e950d96d799/py/conftest.py#L137-L139

 Adding separate markers will make the tests unnecessarily lengthy. WDYT?

navin772:
The tests are failing due to extension path not found on engflow runner, locally they work fine. If someone can directly run on engflow and troubleshoot the path, it would be great. I cannot access it directly.

cgoldberg:
> Regarding xfail markers for safari and other non-bidi browsers, I think we already skip them

you're right.. I forgot about that.

> The tests are failing due to extension path not found on engflow runner

It is failing on the GitHub runner also... I'm not sure why.. nothing looks incorrect in the failing test.

other comments:

- You might want to create a function that installs/uninstalls the extension for the tests, to cut down on code repetition. The tests mostly do the same thing.
- The test for uninstalling looks redundant since all of the tests do an uninstall. You can probably remove that one, right?

navin772:
@cgoldberg thanks for the review. I will see if I can reduce the code duplication in the tests.

> The test for uninstalling looks redundant since all of the tests do an uninstall. You can probably remove that one, right?

There's a slight difference in the way it uninstalls in that test, it uses the extension id directly instead of the dict response we get from installing the extension.
Basically it covers the `else` statement in the `uninstall` method here:
```python
        if isinstance(extension_id_or_result, dict):
            extension_id = extension_id_or_result.get("extension")
        else:
            extension_id = extension_id_or_result
```

And for the test failures in the CI, I am still not sure why that's happening, any tips that will help in debugging?

cgoldberg:
> any tips that will help in debugging?

Not yet. I'll merge trunk so it runs again... maybe it will go away :)  I'll take another look later today to help figure it out.

navin772:
> It is failing on the GitHub runner also...

@cgoldberg I added the required directories in the bazel BUILD file and now the CI that runs on github runner ([CI / Python / Browser Tests (firefox, ubuntu)](https://github.com/SeleniumHQ/selenium/actions/runs/15156602915/job/42613317161?pr=15749))  passes but CI-RBE still fails :(

py/test/selenium/webdriver/common/bidi_webextension_tests.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def test_webextension_initialized(driver):
+    """Test that the webextension module is initialized properly."""
+    assert driver.webextension is not None
+
+
+@pytest.mark.xfail_chrome
+@pytest.mark.xfail_edge
+def test_install_extension_path(driver, pages):
```
oliverdunk, line 65: I want to clarify that `webExtension.install` is **supported** in Chrome (and I expect all Chromium based browsers). Currently we only allow the extension path case - you can see the implementation here: https://github.com/GoogleChromeLabs/chromium-bidi/pull/3012

Additionally, we only allow this if *both* of the following are true:

- The browser was started with the `--enable-unsafe-extension-debugging` flag
- The CDP connection between `chromedriver` and the browser is using a pipe, not a port. The easiest way to do this is using the `--remote-debugging-pipe` flag.

An example of doing both in Selenium is as follows:

```java
ChromeOptions options = new ChromeOptions()
        .enableBiDi()
        .addArguments("--enable-unsafe-extension-debugging", "--remote-debugging-pipe");
WebDriver driver = new ChromeDriver(options);
```

A couple of questions based on this:

- Have you tried the tests with Chrome when these prerequisites are met? It would be great to avoid marking them as failing if possible.
- Do we have any thoughts on what users should do? Is passing `--remote-debugging-pipe` this way a supported path?

You can see how we are supporting this in Puppeteer here: https://pptr.dev/guides/chrome-extensions

I really appreciate the work on this PR, and apologize for the drive-by comment, but I don't want to miss out on Chrome support unintentionally :)

cc/ @OrKoN

navin772: Hi @oliverdunk I did try with chrome but without the above 2 flags and it failed as expected. I will try again with those 2 args.

That said, I am trying to keep this PR short and for firefox only since the chrome implementation might be a bit different given the necessity of the 2 args. I will open another PR once this gets merged to support chrome.

A question - does chrome returns a valid error when using `webExtension.install` without the 2 args? For example if we miss any of the 2 args or both will it return an error saying the flag is required?



> Do we have any thoughts on what users should do? Is passing `--remote-debugging-pipe` this way a supported path?

I will have to check that.

oliverdunk: > A question - does chrome returns a valid error when using webExtension.install without the 2 args? For example if we miss any of the 2 args or both will it return an error saying the flag is required?

You should get an error, but unfortunately I don't think it references the flags specifically. It is something like "Method not allowed." which is the generic error used when a method is called from the wrong context.

navin772: Yeah, I was getting that, I think we should change that since its the same error we get when a browser doesn't supports a BiDi command. That initially confused me since I knew chrome supported `path` based installation as seen in WPT tests.

cgoldberg:
@oliverdunk 
Regarding those flags (`--remote-debugging-pipe`, `--enable-unsafe-extension-debugging`), we would either need to make this apparent to users through documentation (or possibly in an error message), or just add those to the default startup options. Would there be any danger or drawback of always running with those flags enabled? And will they remain available in future versions of Chrome? (I know flags often change).

navin772:
@cgoldberg all the tests are now passing. @shs96c helped in getting the files/dirs referenced correctly in bazel. 

Also, updated the tests to reduce duplication.

oliverdunk:
> Would there be any danger or drawback of always running with those flags enabled? And will they remain available in future versions of Chrome? (I know flags often change).

Great question! I spoke to the team about this:

- Enabling `--remote-debugging-pipe` is always safe, and in fact safer than the default where ChromeDriver uses a WebSocket to connect to Chrome (since that allows other processes to open parallel connections). However, it would likely be a breaking change, and also cause issues with the current implementation of the `WebDriver.createCDPConnection` method in Selenium.
- The `--enable-unsafe-extension-debugging` flag allows for installing persistent extensions which can be dangerous. Therefore, it should only be enabled if a user wants to test extensions.

I'm keen to see if we can loosen any of these requirements where the risk is lower enough, but I don't expect we will do that in the immediate future.

Side-note: It would be great to see https://github.com/SeleniumHQ/selenium/pull/15660 land but I haven't been able to reach anyone on the Selenium team about it. Is there any chance you would be able to help with finding a reviewer?

cgoldberg (approved):
@navin772 LGTM .. thanks
My only question is how do the tests in `ff_installs_addons_tests.py` possibly work without the same `runfiles` change? That's really weird. Should we make the same change in that file for consistency?

navin772:
@cgoldberg `ff_installs_addons_tests.py` test doesn't run on CI-RBE but only on the GitHub runner, the CI-RBE is a much more isolated environment and only files/dirs exposed are available. That's why they are passing, 

> I added the required directories in the bazel BUILD file and now the CI that runs on github runner ([CI / Python / Browser Tests (firefox, ubuntu)](https://github.com/SeleniumHQ/selenium/actions/runs/15156602915/job/42613317161?pr=15749)) passes but CI-RBE still fails :(

It happened with this test also in this PR.

cgoldberg:
> Is there any chance you would be able to help with finding a reviewer?

@oliverdunk I'll see if I can nudge someone to review it

cgoldberg:
@navin772 is this ready to merge? 4.33 release is going out soon and it would be nice to include this.

navin772:
@cgoldberg yes, it's ready!

cgoldberg:
@navin772 oops.. I just thought of one more thing. Add the module name to `py/docs/source/api.rst` so the API docs get generated for it!

Then go ahead and merge it.

rahulharpal1603:
Hi @navin772!

I have a quick question, is this feature supported for Google Chrome versions 135 and 136? 
Because I have used this module to fix this issue : https://github.com/seleniumbase/SeleniumBase/issues/3771 caused by versions 137 and above.

navin772:
Hi @rahulharpal1603, as per the [WPT](https://wpt.fyi/results/webdriver/tests/bidi/web_extension/install/install.py?label=master&label=stable&product=chrome-136.0.7103.94&product=edge&product=firefox&product=safari&aligned) tests, this should work with Chrome 136 but not 135.

rahulharpal1603:
> Hi @rahulharpal1603, as per the [WPT](https://wpt.fyi/results/webdriver/tests/bidi/web_extension/install/install.py?label=master&label=stable&product=chrome-136.0.7103.94&product=edge&product=firefox&product=safari&aligned) tests, this should work with Chrome 136 but not 135.

Okay, thank you so much.

rahulharpal1603:
@navin772 Can you tell if there is a way to load extensions in .crx format? Using this bidi module.

navin772:
@rahulharpal1603 No, currently chrome/edge doesn't supports loading packed extensions (`.crx`) via bidi.
Only Firefox supports it (via `.xpi` extensions).

rahulharpal1603:
> @rahulharpal1603 No, currently chrome/edge doesn't supports loading packed extensions (`.crx`) via bidi.
> Only Firefox supports it (via `.xpi` extensions).

Okay, thanks

PR #15794, opened 2025-05-25
[py][bidi]: add `enable_webextensions` option for chromium-based browsers

navin772:
### 🔗 Related Issues

More info - https://github.com/SeleniumHQ/selenium/pull/15749#discussion_r2099985400

### 💥 What does this PR do?

Chrome requires 2 flags for BiDi webextensions to work:
1. `--remote-debugging-pipe`
2. `--enable-unsafe-extension-debugging`

This PR adds a chrome option to enable these flags easily - `enable_webextensions`
User can either use the above chrome option or add the flags manually.

The `webExtension.install` method will return an exception if the user forgets to add these flags.

Usage:

```python
from selenium.webdriver.chrome.options import Options

options = Options()
options.enable_webextensions = True
driver = webdriver.Chrome(options=options)
```

OR directly pass the flags

```python
from selenium.webdriver.chrome.options import Options

options = Options()
options.add_argument("--remote-debugging-pipe")
options.add_argument("--enable-unsafe-extension-debugging")

driver = webdriver.Chrome(options=options)
```

### 🔄 Types of changes

___

### **PR Type**
Enhancement, Tests

___

### **Description**
- Adds `enable_webextensions` option to Chrome for BiDi webextensions
  - Automatically manages required Chrome flags for webextension support
  - Allows enabling/disabling via property or manual flags

- Improves error handling in `webExtension.install` for missing flags
  - Raises descriptive exception if required flags are missing

- Updates and documents BiDi webextension test for Chrome
  - Explains usage of new option and required flags

- Updates test configuration to enable webextensions for Chrome BiDi

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>options.py</strong><dd><code>Add `enable_webextensions` property to Chrome options</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/selenium/webdriver/chrome/options.py

<li>Adds <code>enable_webextensions</code> property with getter/setter<br> <li> Automatically adds/removes required Chrome flags for webextension <br>support<br> <li> Updates docstrings and internal logic for flag management

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15794/files#diff-9d0db654e45e56c900d79485b83227102777298cda5bc29b364184af88da9254">+37/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Bug fix</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>webextension.py</strong><dd><code>Improve error handling for Chrome webextension install</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/selenium/webdriver/common/bidi/webextension.py

<li>Adds error handling for missing Chrome webextension flags<br> <li> Raises descriptive exception if required flags are not set

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15794/files#diff-f6467ecd5e2e38f1ba13183bb5fae9f304b5edba8c9acb8855e5572ac5a6e71a">+12/-2</a>&nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>bidi_webextension_tests.py</strong><dd><code>Update and document BiDi webextension test for Chrome</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/test/selenium/webdriver/common/bidi_webextension_tests.py

<li>Updates test to document Chrome webextension enabling<br> <li> Adds logic to handle Chrome-specific extension id behavior<br> <li> Provides usage examples for new option and flags

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15794/files#diff-6e579bdd2c5a63ae03a0572188781b474610491e5b14c4ea1eaf65dc35750851">+26/-3</a>&nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>conftest.py</strong><dd><code>Enable Chrome webextensions in BiDi test configuration</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/conftest.py

<li>Enables webextensions for Chrome when BiDi is enabled in test config<br> <li> Checks for presence of new option before setting

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15794/files#diff-98e9e290ede5d0c7ac9e7df5b49edf63f5df5fcc946b7736e4a6139e4cda26e3">+4/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

___


navin772:
Is there a better way to directly use the `enable_webextensions` chrome option in the test instead of modifying the `conftest.py` file?

navin772:
@oliverdunk can you have a look at this PR?

cgoldberg:
> Is there a better way to directly use the enable_webextensions chrome option in the test instead of modifying the conftest.py file?

I guess the issue is that you don't want to add `enable_webextensions` to the `driver` fixture because it will affect most tests? 

You could create the driver in the test and not use the fixture. We do that in a few tests. You could add a new fixture similar to the `firefox_options` fixture in `conftest.py`. Then use this fixture in your test, and set `enable_webextensions` from inside the test.

It's not a great solution, but I think it will help.

(let me know if that's not a good explanation)

navin772:
@cgoldberg I have added a new pytest marker `webextension` and a check in `conftest.py` to set `enable_webextensions` to `True` only when this pytest marker is used (thus other tests won't be affected).

shbenzer:
> > Is there a better way to directly use the enable_webextensions chrome option in the test instead of modifying the conftest.py file?
> 
> I guess the issue is that you don't want to add `enable_webextensions` to the `driver` fixture because it will affect most tests?
> 
> You could create the driver in the test and not use the fixture. We do that in a few tests. You could add a new fixture similar to the `firefox_options` fixture in `conftest.py`. Then use this fixture in your test, and set `enable_webextensions` from inside the test.
> 
> It's not a great solution, but I think it will help.
> 
> (let me know if that's not a good explanation)

I agree with this - I don't think the new marker or conftest changes are necessary. We can create a driver specifically for the webextension tests, and then put a xfail_firefox marker (e.g. with reason "not implemented") on those tests for now. You can wrap the webextension tests in a test class to make this easier.

navin772:
> We can create a driver specifically for the webextension tests

Will the new driver be created inside the webextension test file or `conftest.py`? I am trying the former one with a class based approach (different class for Firefox and chrome/edge) and getting errors like the extension won't load (the install method won't fail but I can't see the extension being loaded in the browser)

Edit: Extension is loaded but `pages.load("blank.html")` is opening the page in the window that doesn't have the extension installed.

cgoldberg:
> Will the new driver be created inside the webextension test file or conftest.py

The solution I'm suggesting would create it within the test.

It's going to be difficult to have a common test that supports all browsers. I would write a test for Firefox (that's skipped by other browsers) and one for Chrome/Edge (that's skipped by other browsers). For an example, look at `./py/test/selenium/webdriver/firefox/firefox_sizing_tests.py`.

navin772:
@cgoldberg yes, I am trying it as you described above. I noticed one issue right now which results in my tests getting failures, while launching the driver, it opens 2 browser windows, extension installation is successful and when I do `pages.load("blank.html")`, it opens the blank.html page in the window that doesn't have the extension installed (the other tab has it installed).
If I do `driver.get("https://www.webpagetest.org/blank.html")` instead of `pages.load("blank.html")` it works fine.

I am trying to fix this.

navin772:
I think `pages.load` uses the normal driver and not the `chromium_driver` I added. For now, only this is working:

```python
chromium_driver.get("https://www.webpagetest.org/blank.html")
```

@cgoldberg @shbenzer can you please have a look and see if something can be done so that pages.load work as expected.

shbenzer:
> I think `pages.load` uses the normal driver and not the `chromium_driver` I added. For now, only this is working:
> 
> ```python
> chromium_driver.get("https://www.webpagetest.org/blank.html")
> ```

the pages fixture is defined in conftest.py as the following:

```python
@pytest.fixture
def pages(driver, webserver):
    class Pages:
        def url(self, name, localhost=False):
            return webserver.where_is(name, localhost)

        def load(self, name):
            driver.get(self.url(name))

    return Pages()
```

You're correct. The problem is this uses the initial driver - you'll need to create your own Pages class if you wish to use this functionality

navin772:
> you'll need to create your own Pages class

@shbenzer I guess that will be a lot of changes to do just for a single test file, also we have an issue in running this multi driver approach in [CI-RBE](https://github.com/SeleniumHQ/selenium/actions/runs/15303232919/job/43049130588?pr=15794#step:18:3842) which is giving a `user data dir` error (maybe due to these 2 drivers), although browser tests on github runner are working.

shbenzer:
> > you'll need to create your own Pages class
> 
> @shbenzer I guess that will be a lot of changes to do just for a single test file, also we have an issue in running this multi driver approach in [CI-RBE](https://github.com/SeleniumHQ/selenium/actions/runs/15303232919/job/43049130588?pr=15794#step:18:3842) which is giving a `user data dir` error (maybe due to these 2 drivers), although browser tests on github runner are working.

I don't think it's a big deal adding the class in if you'll use it, but do what you think is best. 

With regards to the error you're seeing in RBE:

> selenium.common.exceptions.SessionNotCreatedException: Message: session not created: probably user data directory is already in use, please specify a unique value for --user-data-dir argument, or don't use --user-data-dir

You gotta love when an error message uses "probably"... You could try creating a temporary directory then setting it in chromium_driver's options with something like `options.add_argument(f"--user-data-dir={temp_dir}")`?

cgoldberg:
@navin772 
few things... 

That `SessionNotCreatedException` is a super generic error that just means it couldn't launch the browser. It likely has nothing to do with `-user-data-dir` like the message says.

In your tests, you are instantiating `ChromeOptions` and `EdgeOptions`, but they don't include some of the default options we need to set from outside of the test (like the ability to add the `--headless` arg). Instead of creating them in the test file, you probably want to add a fixture to `conftest.py` for setting options, then create the driver in your test using the options fixture. Take a look at the `firefox_options` fixture. You could create a similar ones for chrome/edge, or combine them into one fixture that can handle options for any driver.

I don't think the `pages_chromium` fixture you added is going to work quite right, because it still uses the `pages` fixture which will launch a driver. You could use the `webserver` fixture instead, and then just navigate to the page in your test.

navin772:
> I don't think the `pages_chromium` fixture you added is going to work quite right, because it still uses the `pages` fixture which will launch a driver.

It does work because `chromium_driver.get((self.url(name))` will launch the page using the chromium_driver not the pages driver. But still it can be simplified more as you said.

I will try to move the fixture to `conftest.py`.

py/selenium/webdriver/chromium/options.py
```diff
@@ -126,6 +127,39 @@ def add_experimental_option(self, name: str, value: Union[str, int, dict, list[s
...
+        self._enable_webextensions = value
+        if value:
+            # Add required flags for Chromium webextension support
+            required_flags = ["--enable-unsafe-extension-debugging", "--remote-debugging-pipe"]
+            for flag in required_flags:
+                if flag not in self._arguments:
+                    self.add_argument(flag)
+        else:
+            # Remove webextension flags if disabling
+            flags_to_remove = ["--enable-unsafe-extension-debugging", "--remote-debugging-pipe"]
```
titusfortner, line 158: Should we error instead of silently removing these? In general if a user passes in capabilities and those capabilities can't be honored, we shouldn't ignore it.

navin772: If I understand correctly, we are referring to this scenario:
```python
options.add_argument("--enable-unsafe-extension-debugging")
options.enable_webextensions = False
```

And, the concern here is that setting `enable_webextensions` to False will silently remove the added flag?

diemol:
Posting the same comment I posted on Slack:

Actually, ChatGPT helped me understand why this happens.
Since we are setting the location of the binary in those tests, this is ChatGPT’s explanation:

> This test overrides the Chrome binary location, and probably (via getBinaryPaths) pulls a shared browserPath and driverPath that are used across tests.
Even if you’re not explicitly setting --user-data-dir, Chrome internally defaults to a shared user data dir unless told otherwise. And reusing a binary manually can cause odd behavior — particularly if Chrome launches before the profile lock is released.
That would explain:
Only this test failing (because it’s the only one using a custom binary),
Only on Ubuntu CI (Chrome on Linux is stricter about .lock files in the profile).

This is how I fixed it in our docs examples.
https://github.com/SeleniumHQ/seleniumhq.github.io/commit/ffbf252fb5e404269931f135761b9bb13b20a959

cgoldberg (approved):
LGTM 👍 

is this ready to be merged? Let me know and I'll pull the trigger.

cgoldberg:
Just a note on this:
Starting with Chrome 136, if you set `--remote-debugging-pipe`, you must also set a `--user-data-dir`:

https://developer.chrome.com/blog/remote-debugging-port

So `enable_extensions` will only work if that is also set.

Also note:
`--remote-debugging-pipe` makes the CDP connection between chromedriver and the browser use a pipe, not a port. So starting Chrome with this flag will make it so CDP things like `driver.start_devtools()` will not work and fail with a cryptic error.

shbenzer:
> Also note: `--remote-debugging-pipe` makes the CDP connection between chromedriver and the browser use a pipe, not a port. So starting Chrome with this flag will make it so CDP things like `driver.start_devtools()` will not work and fail with a cryptic error.

That’s a great note - we should definitely advise users about this in the docstring and the docs. I’ll make a pr real quick

py/selenium/webdriver/chromium/options.py
emanlove, line 158: Just wanted to note that from my reading of the `--remote-debugging-pipe` [description](https://peter.sh/experiments/chromium-command-line-switches/#remote-debugging-pipe) it is not exclusively a webextension command line switch. So if one adds it separately and chooses to disable web extensions via this method, as in,

```python
    options.add_argument("--remote-debugging-pipe")
    options.enable_webextensions = False
```

 it will negate that command line switch and might be tough to debug. Probably the method documentation should note that disabling removes both switches.

navin772: This was discussed but no conclusion was made - https://github.com/SeleniumHQ/selenium/pull/15794#discussion_r2113162844, what do you think would be the best way to resolve this? 

Adding docs is good.

PR #15811, opened 2025-05-28
[py] Lint Python with ruff

p0deje:
Use Ruff directly to format and lint Python source code files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,6 +75,7 @@ py/selenium/webdriver/remote/getAttribute.js
 py/selenium/webdriver/remote/isDisplayed.js
 py/docs/build/
 py/docs/source/**/*
+!py/docs/source/conf.py
 py/build/
 py/LICENSE
 py/pytestdebug.log
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,7 +1,6 @@
 module(name = "selenium")
 
 bazel_dep(name = "apple_rules_lint", version = "0.4.0")
-bazel_dep(name = "aspect_rules_lint", version = "1.4.2")
 bazel_dep(name = "aspect_bazel_lib", version = "2.13.0")
 bazel_dep(name = "aspect_rules_esbuild", version = "0.21.0")
 bazel_dep(name = "aspect_rules_js", version = "2.3.7")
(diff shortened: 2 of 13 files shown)


cgoldberg (commented):
Do you need to add `--exit-non-zero-on-fix` to `check`, and `--exit-non-zero-on-format` to `format`? We want the CI job to fail if it hits any violations.

Other than that, LGTM

p0deje:
> Do you need to add --exit-non-zero-on-fix to check, and --exit-non-zero-on-format to format

Our CI job runs the format and then checks for `git diff`, so it will fail.

cgoldberg (approved):

cgoldberg:
@p0deje There are some issues that ruff won't auto-fix (like lines that are too long)... so git diff wouldn't see those. You might want to add those args to make sure it fails.

p0deje:
@cgoldberg Hmm, I assumed that when Ruff cannot fix/format, it would fail with non-zero exit code. At least this is how I read  the flag name. Will update.

cgoldberg:
I'm not 100% sure what the exit code is when it can't fix... I'd have to try, but I suppose it doesn't hurt to add the flags.

p0deje:
I did some testing and `ruff` seems to exit non-zero when it cannot autofix:

```
$ ruff check --fix scripts/pinned_browsers.py; echo $status
scripts/pinned_browsers.py:208:5: E741 Ambiguous variable name: `l`
    |
206 |     mac = None
207 |     mac_hash = None
208 |     l = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA…
    |     ^ E741
209 |
210 |     for data in all_data:
    |

scripts/pinned_browsers.py:208:5: F841 Local variable `l` is assigned to but never used
    |
206 |     mac = None
207 |     mac_hash = None
208 |     l = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA…
    |     ^ F841
209 |
210 |     for data in all_data:
    |
    = help: Remove assignment to unused variable `l`

Found 2 errors.
No fixes available (1 hidden fix can be enabled with the `--unsafe-fixes` option).
1
```

For format, there seem to be no cases when it would exit non-zero and not format files.

I am going to merge this as-is, we should be safe even w/o the flags.

PR #15847, opened 2025-06-03
fixed errors in browser.py for 15697

rpallavisharma:
### 💥 What does this PR do?

fixed errors in browser.py for 15697
### 🔄 Types of changes

- Cleanup (formatting, renaming)
- Bug fix (backwards compatible)
- New feature (non-breaking change which adds functionality *and tests!*)
- Breaking change (fix or feature that would cause existing functionality to change)

___

### **PR Type**
Bug fix

___

### **Description**
- Added strict type validation to `ClientWindowInfo.from_dict`

- Raised `ValueError` for missing or invalid fields

- Improved error handling for dictionary parsing

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>browser.py</strong><dd><code>Add type checks and error handling to from_dict method</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/selenium/webdriver/common/bidi/browser.py

<li>Enhanced <code>from_dict</code> to validate field types strictly<br> <li> Now raises <code>ValueError</code> for missing or invalid fields<br> <li> Improved robustness of dictionary parsing for window info

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15847/files#diff-cd7018a350739e3db649c9e7f186036670f27a06f8df204de5909a03d7a04f5f">+44/-9</a>&nbsp; &nbsp; </td>

</tr>
</table></td></tr></tr></tbody></table>

___

> <details> <summary>  Need help?</summary><li>Type <code>/help how to ...</code> in the comments thread for any questions about Qodo Merge usage.</li><li>Check out the <a href="https://qodo-merge-docs.qodo.ai/usage-guide/">documentation</a> for more information.</li></details>

PR #15848, opened 2025-06-03
[py] Fix : Mypy type annotation errors - 2

ShauryaDusht:
### 🔧 What does this PR do?
This PR fixes several type annotation errors reported by Mypy in the following files:
- `selenium/webdriver/common/bidi/browser.py` - changes reverted to avoid conflict from another PR #15847 
- `selenium/webdriver/common/bidi/browsing_context.py`
- `selenium/webdriver/common/action_chains.py` - recently added
- `selenium/webdriver/common/utils.py` - recently added

---
### 💡 Additional Notes
- This is part of ongoing efforts to improve type annotations across the Selenium codebase.
- No behavior changes are introduced; only type hints are corrected.
- This is my second PR on the issue.
---
### 🔄 Types of changes
- Code cleanup (type annotation fixes, no functional changes)
___

### **PR Type**
Bug fix, Enhancement


ShauryaDusht:
Hi @cgoldberg, please review my PR (I have added a few more files in this PR)

navin772:
@ShauryaDusht another PR (https://github.com/SeleniumHQ/selenium/pull/15847) is resolving the type errors for `browser.py` file, can you pick another file to avoid conflict?

ShauryaDusht:
> @ShauryaDusht another PR (#15847) is resolving the type errors for `browser.py` file, can you pick another file to avoid conflict?

Will remove `browser.py` changes and update the PR with another file. Thanks!

navin772:
@ShauryaDusht I don't think silently defaulting to a value (like an empty string) when the value is not found is a good idea. This will introduce uncertainty and does not conform to the BiDi spec. 
For example, the changes made for `UserPromptOpenedParams`, `default_value` is set to an empty string if not found but the spec says:
`the defaultValue field set to default value if default value is not null or omitted otherwise`
which means it should either be omitted from the payload or should be `None`.

So, we should do something like this:
```python
		default_value = json.get("defaultValue")
        if default_value is not None and not isinstance(default_value, str):
           raise ValueError("defaultValue must be a string if provided")
```
Same with other classes and params.

ShauryaDusht:
Fixed this as per your suggestion — now handling `defaultValue` correctly without silent defaults. Let me know if anything else looks off.
Also, made changes suggested by bot.
Thanks!

ShauryaDusht:
Hi @navin772 
I have made the suggested changes. Let me know if anything else needs fixing!

ShauryaDusht:
Hi @navin772
Please review my changes

py/selenium/webdriver/common/utils.py
```diff
@@ -131,8 +131,7 @@ def keys_to_typing(value: Iterable[AnyKey]) -> list[str]:
     characters: list[str] = []
     for val in value:
         if isinstance(val, Keys):
-            # Todo: Does this even work?
```
navin772, line 134: Let's not remove the comment, maybe its a valid one!

py/selenium/webdriver/common/bidi/browsing_context.py
```diff
@@ -66,12 +67,23 @@ def from_json(cls, json: dict) -> "NavigationInfo":
...
+        if context is None or not isinstance(context, str):
+            raise ValueError("context is required and must be a string")
+
+        navigation = json.get("navigation")
+        if navigation is not None and not isinstance(navigation, str):
+            raise ValueError("navigation must be a string")
+
+        timestamp = json.get("timestamp")
+        if timestamp is None or not isinstance(timestamp, int):
+            raise ValueError("timestamp is required and must be an integer")
```
navin772:
```suggestion
        timestamp = json.get("timestamp")
        if timestamp is None or not isinstance(timestamp, int) or timestamp < 0:
            raise ValueError("timestamp is required and must be a non-negative integer")
```
`timestamp` is of js-uint type, change it in `DownloadWillBeginParams` also

py/selenium/webdriver/common/bidi/browsing_context.py
```diff
@@ -108,12 +120,25 @@ def from_json(cls, json: dict) -> "BrowsingContextInfo":
...
+                    children.append(BrowsingContextInfo.from_json(child))
+                else:
+                    warnings.warn(f"Unexpected child type in browsing context: {type(child)}")
+        context = json.get("context")
+        if context is None or not isinstance(context, str):
+            raise ValueError("context is required and must be a string")
+
+        url = json.get("url")
+        if url is None or not isinstance(url, str):
+            raise ValueError("url is required and must be a string")
```
navin772, line 140: Add validations for `parent`, `user_context`, `original_opener` and `client_window` params also.

navin772 (changes requested):

py/selenium/webdriver/common/utils.py
ShauryaDusht: I have reverted this one

py/selenium/webdriver/common/bidi/browsing_context.py
ShauryaDusht: I have made changes at both instances

py/selenium/webdriver/common/bidi/browsing_context.py
ShauryaDusht: I have added validation for all above params

navin772 (approved):
@ShauryaDusht I have made some changes as per the BiDi spec.

LGTM!

ShauryaDusht:
Thanks for the changes and review

Issue #15697, opened 2025-05-04
[🐛 Bug]: [py] Many type annotation errors need fixing


cgoldberg:
### Description

TLDR: Our use of type hints in the Python bindings needs some serious love.

----

Static typing is optional in Python, and we make use of it in many places throughout the Python bindings code. Static typing is not enforced at runtime, but can be checked with tooling like Mypy.

We currently have a Mypy environment and set of type checking rules declared in our `tox.ini` configuration. You can run Mypy locally with:
```
tox -e py/tox.ini -e mypy
```

Running this produces a screen full of type errors and:
```
Found 120 errors in 30 files (checked 114 source files)
```

We currently run Mypy in a CI workflow, but we don't display the errors or fail the job when it hits errors, so they just go unnoticed.

To resolve this issue, we should:
- go through the codebase and fix all of the errors that Mypy finds
- update the CI job to fail when Mypy encounters errors to enforce no future regressions  

### Reproducible Code

```shell
tox -e py/tox.ini -e mypy
```

### Debugging Logs

```logs
Found 120 errors in 30 files (checked 114 source files)
```

MohabASHRAF-byte:
Hello, can I work on this ?

navin772:
@MohabASHRAF-byte yes, go ahead. We appreciate contributions from your side.

MohabASHRAF-byte:
@navin772 Is there a communication channel other than GitHub?

navin772:
@MohabASHRAF-byte Yes, we use slack, you can join here - https://inviter.co/seleniumhq

cgoldberg:
@MohabASHRAF-byte thanks! Not only are there many Mypy errors, but many of the type annotations are using outdated classes and syntax. An easy way to update them is by using this:

https://github.com/asottile/pyupgrade

Run it across the whole codebase using Python 3.9 syntax.

Ping me here or on Slack if you have any questions.

MohabASHRAF-byte:
> [@MohabASHRAF-byte](https://github.com/MohabASHRAF-byte) thanks! Not only are there many Mypy errors, but many of the type annotations are using outdated classes and syntax. An easy way to update them is by using this:
> 
> https://github.com/asottile/pyupgrade
> 
> Run it across the whole codebase using Python 3.9 syntax.
> 
> Ping me here or on Slack if you have any questions.

ok, thank you, I will consider it, and I sent a request to join Slack.

MarcelWilson:
[...]

PR #15928, opened 2025-06-20
[py] Addressing mypy type annotation errors

bandophahita:
### 🔗 Related Issues
addresses part of #15697

### 💥 What does this PR do?
Updates annotations and sets proper default objects.

### 🔄 Types of changes

___

### **PR Type**
Other

___

### **Description**
- Fix mypy type annotation errors across webdriver services

- Change `Sequence[str]` to `list[str]` for service arguments

- Add explicit type annotations for class attributes

- Update ChromiumDriver constructor parameter types

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><details><summary>3 files</summary><table>
<tr>
  <td><strong>service.py</strong><dd><code>Update service_args type annotations to list[str]</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15928/files#diff-94e16ea2a7773a7ba8eeb142111c7956558fc0f90c08394bb1e99dd1fa7290b1">+4/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>service.py</strong><dd><code>Fix type annotations and add attribute declarations</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15928/files#diff-062c917a35158860c8065910187232dbab480ac982cc48bb293c8e808d29818e">+6/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>service.py</strong><dd><code>Change service_args type from Sequence to list</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15928/files#diff-fe6756f6489050e903be9db84475a54c6b90289dfc08db0aeb14d264a14c57d4">+3/-3</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Enhancement</strong></td><td><details><summary>4 files</summary><table>
<tr>
  <td><strong>webdriver.py</strong><dd><code>Add explicit type annotations for options and service</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15928/files#diff-5a60156e0c8a7c920aa161d26581ca3461625b9ea29c059bae02999fd388f22f">+3/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>webdriver.py</strong><dd><code>Update constructor parameter types and imports</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15928/files#diff-10fa59c3fd96a0c31762373ca8d7373c0ab669a437ae957b1a17ea1377b20608">+6/-7</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>service.py</strong><dd><code>Add log_output type annotation declaration</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15928/files#diff-646e4c91d7ef7fbd2b4c088347c86f5b2d2cab8121b14987f3cb783ad120634c">+2/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>webdriver.py</strong><dd><code>Add explicit type annotations for attributes</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15928/files#diff-7efa39061cd6f52f01b1996583fad8b66b0ddf7e8d41f5a03c910ac598dfe6dc">+3/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

___


navin772:
@bandophahita Thank you for the contribution! This looks good.
Could you also add mypy fixes for `wpewebkit/service.py` and `webkitgtk/service.py` in this PR, those are quite similar to the ones you updated above and will fix type errors for all `service.py` files.

Also, run the formatter - `./scripts/format.sh` to get the CI passing.

cgoldberg:
> Change Sequence[str] to list[str] for service arguments

Why do we need this? What's wrong with accepting a sequence?

bandophahita:
> > Change Sequence[str] to list[str] for service arguments
> 
> Why do we need this? What's wrong with accepting a sequence?

Need is probably too strong of a term. It made sense to me to keep it a list since that's what `_service_args` is and that's how all of the code internally passes arguments.  Nothing would prevent accepting a `Sequence` since it is converted to a list in the `__init__`.  

Short answer; keeping the types simple prevents confusion.  But in case that's an issue, I reverted the argument annotations.

cgoldberg:
> I reverted the argument annotations.

I still see a change from Sequence to list. Just leave them as a sequence unless there is a need to enforce a list.

bandophahita:
> I still see a change from Sequence to list. Just leave them as a sequence unless there is a need to enforce a list.

There are changes from Sequence to list but those are attributes.  I reverted the argument annotations.

bandophahita:
> I still see a change from Sequence to list. Just leave them as a sequence unless there is a need to enforce a list.

I think I should explain the change.  The problem was when referencing `service.service_args`  (which is a property returning `_service_args`) it was annotated as a Sequence; and yet internally there were actions that can only be performed as a list object.  Mypy was catching and reporting this inconsistancy.  Sequence is fine for the argument being passed in, but it was then being converted to a list.  The annotation was in incorrect for the attribute.

cgoldberg:
@bandophahita can you explain why we need the new class attributes?

bandophahita:
> @bandophahita can you explain why we need the new class attributes?

For sure.  These are technically not NEW attributes but simply annotations for the class attributes.  They dont do anything at runtime.  Mypy reads them as the annotations instead of autodetecting/using what is set in the `__init__`.  Mypy sometimes can't deterimne what a dynamic attribute is very well, these help it.  I can go through them and breakdown how, if you want.

py/selenium/webdriver/edge/webdriver.py
```diff
@@ -27,6 +27,8 @@
 class WebDriver(ChromiumDriver):
     """Controls the MSEdgeDriver and allows you to drive the browser."""
 
+    service: Service
```
bandophahita, line 30: This class annotation lets mypy understand that the webdriver has an attribute `service` that is an instance of `Service`.  The `Optional[Service]` argument annotation caused mypy to believe the attribute `webdriver.servce` could be `None`, when in fact it never is.  Mypy isn't smart enough to read the code that way.

py/selenium/webdriver/chromium/service.py
```diff
@@ -35,6 +35,9 @@ class ChromiumService(service.Service):
     :param driver_path_env_key: (Optional) Environment variable to use to get the path to the driver executable.
     """
 
+    _service_args: list[str]
+    log_output: Union[None, IOBase, int, SubprocessStdAlias]
```
bandophahita, line 39: Annotation to let mypy know what `log_output` attribute can be.  Autodetermination wasn't smart enough to see past the argument annotation which was fine for the argument but incomplete for the attribute.

py/selenium/webdriver/chromium/webdriver.py
```diff
@@ -31,10 +30,10 @@ class ChromiumDriver(RemoteWebDriver):
...
     def __init__(
         self,
-        browser_name: Optional[str] = None,
-        vendor_prefix: Optional[str] = None,
-        options: ArgOptions = ArgOptions(),
-        service: Optional[Service] = None,
+        browser_name: str,
+        vendor_prefix: str,
+        options: ChromiumOptions = ChromiumOptions(),
+        service: ChromiumService = ChromiumService(),
```
bandophahita, line 36: `options` and `service` are always supposed to be `ChromiumOptions` and `ChromiumService` respectively.  
In fact, they can never be `None` according to the code in the `__init__`.  

This doesn't directly impact `chrome.webdriver.WebDriver.__init__` nor `edge.webdriver.WebDriver.__init__` which both allow the arguments to be `None`; both create instances if the user doesn't pass it in.

py/selenium/webdriver/chromium/service.py
bandophahita: There is a real possibility that the argument annotation is incomplete here too.  I didn't address this.  Someone more familiar with the `log_output` argument should confirm it's annotation.

PR #15964, opened 2025-06-26
[py] Remove Exclusion of Chromium from Network BiDi tests

shbenzer:
### 💥 What does this PR do?
Remove Exclusion of Chromium from Network BiDi tests

### 🔄 Types of changes

### Notes
- Once this passes, Chromium browsers will have caught up to Firefox with respect to BiDi network functionality


cgoldberg:
These tests now pass in trunk and the `xfail` marker has already been removed.

PR #16003, opened 2025-07-06
[py] Allow free_port() to bind to IPv6 if IPv4 is unavailable

### LINKED ISSUE #14910: [🐛 Bug]: FindFreePort method hardcoded for IPv4 causing failure for pure IPv6 environments
### What happened?

When running Selenium on a "pure IPv6" system where all IPv4 interfaces have been removed, the FindFreePort method, which is used to find a free TCP port to use on the loopback interface for communication to the browser driver, fails since it is hard coded for IPv4 and there are no IPv4 interfaces.

(https://github.com/SeleniumHQ/selenium/blob/trunk/dotnet/src/webdriver/Internal/PortUtilities.cs) Line 40
`Socket portSocket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);`

AddressFamily.InterNetwork is IPv4.  AddressFamily.InterNetworkV6 is IPv6.

In Pure IPv6 environment, attempt create the Browser Driver object fails with exception: "An address incompatible with the requested protocol was used"

```shell
On windows host where selenium is being executed:
-Ensure IPv6 is working as expected
-Remove IPv4 by running:  netsh interface ipv4 uninstall
-Reboot for uninstall to complete
-Attempt to use Selenium

The below PowerShell code could be used for example. Just place the WebDriver.dll and selected browserdriver.exe in the same directory as the saved PowerShell script and execute:

Function Get-Browser {
    Param (
           [Parameter(Mandatory=$true)]
           [ValidateSet("Edge","Chrome")]
           [string]$type
          )
    
    Set-EnvironmentPath

    Add-Type -Path "$($PSScriptRoot)\WebDriver.dll"

    Switch ($type) {
          "edge" {
                  $Driver = New-Object OpenQA.Selenium.Edge.EdgeDriver
                 }
        "chrome" {
                  $Driver = New-Object OpenQA.Selenium.Chrome.ChromeDriver
                 }
    }
    $Driver
}

Function Set-EnvironmentPath {
    # Add the working directory of script to the environment path.
    # This is required for the WebDriver to work.
    if (($env:Path -split ';') -notcontains $PSScriptRoot)
    {
        Write-Verbose -Message "Enviornment path does not contain $PSScriptRoot.  Adding to Path"
        if ($env:Path.Substring($env:Path.Length-1) -eq ';')
        {
            $env:Path += "$PSScriptRoot"
        }
        else
        {
            $env:Path += ";$PSScriptRoot"
        }
    }
}

$browser = Get-Browser -type Edge
```

### Relevant log output

```shell
New-Object : Exception calling ".ctor" with "0" argument(s): "An address incompatible with the requested protocol was used"
At C:\WebDev\Start-Selenium.ps1:14 char:29
+ ...                  $Driver = New-Object OpenQA.Selenium.Edge.EdgeDriver
+                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [New-Object], MethodInvocationException
    + FullyQualifiedErrorId : ConstructorInvokedThrowException,Microsoft.PowerShell.Commands.NewObjectCommand
```

### Operating System

Windows 2019

### Selenium version

PowerShell

Edge  131.0.2903.86

Microsoft Edge WebDriver 131.0.2903.99 (f96a870b1316f430223b7df17df32918b222060e)

No

cgoldberg:
### 🔗 Related Issues

Fixes #14910 for Python

### 💥 What does this PR do?

This PR updates the `free_port()` function in `selenium.webdriver.common.utils` so it will bind to `localhost` using IPv6 (`::1`) if IPv4 (`127.0.0.1`) is not available.

Without this fix, Selenium can't be used on an IPv6-only system without some additional code.

For example, if you are on IPv6-only system, the following code will fail:

```
driver = webdriver.Chrome()
```

with `OSError: [Errno 99] Cannot assign requested address`

With the changes in the PR, it will work.

Note: `geckodriver` doesn't work on IPv6-only systems, so Firefox still won't work.

----

#### Notes on testing:

I didn't have an IPv6-only system to test on, so I used a a mixed environment and disabled IPv4 interfaces. Here are instructions for setting that up on Debian/Ubuntu Linux:

To see interface names and IP addresses, run `ip addr show`.

For example, I get the following output:

```
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
5: eth0@if6: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 00:16:3e:70:a6:12 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 100.115.92.196/28 brd 100.115.92.207 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 2601:189:8181:33b0:216:3eff:fe70:a612/64 scope global dynamic mngtmpaddr 
       valid_lft 345599sec preferred_lft 345599sec
    inet6 fe80::216:3eff:fe70:a612/64 scope link 
       valid_lft forever preferred_lft forever
```

You can see I have the following network interfaces with IPv4 addresses:

```
lo: 127.0.0.1/8
eth0: 100.115.92.196/28
```

I can disable these by running:

```
sudo ip addr del 100.115.92.196/28 dev eth0
sudo ip addr del 127.0.0.1/8 dev lo
```

Now I am running on a system that only allows IPv6 and can test.

(these changes will be reverted on reboot)

### 🔄 Types of changes

___

### **PR Type**
Enhancement

___

### **Description**
- Add IPv6 fallback support to `free_port()` function

- Enable Selenium to work on IPv6-only systems

- Maintain backward compatibility with IPv4 systems

___

### **Changes diagram**

```mermaid
flowchart LR
  A["free_port() called"] --> B["Try IPv4 bind"]
  B --> C{IPv4 available?}
  C -->|Yes| D["Bind to 127.0.0.1"]
  C -->|No| E["Fallback to IPv6"]
  E --> F["Bind to ::1"]
  D --> G["Return port"]
  F --> G
```

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>utils.py</strong><dd><code>Add IPv6 fallback to free_port function</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/selenium/webdriver/common/utils.py

<li>Modified <code>free_port()</code> to try IPv4 first, fallback to IPv6<br> <li> Added exception handling for OSError when IPv4 binding fails<br> <li> Updated function docstring to explain IPv6 fallback behavior

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16003/files#diff-b1516e150256fb4a5ef5144d5d8314fc45e92872415ba8900d4cd8d2cc9b9841">+12/-3</a>&nbsp; &nbsp; </td>

</tr>

___


nvborisenko:
@cgoldberg do you know how to easily test it?

cgoldberg:
@nvborisenko read my PR ... I wrote a whole section on how to test it :)

(for Linux at least)

nvborisenko:
I tried your steps to disable IPv4 interfaces on Ubuntu VM, and even cannot reproduce the issue when finding available port should fail :(

py/selenium/webdriver/common/utils.py
```diff
@@ -27,12 +27,31 @@
...
+        if free_socket:
+            free_socket.close()
+        # IPv6
+        try:
+            free_socket = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
+            free_socket.bind(("::1", 0))
+        except OSError:
+            raise RuntimeError("Can't find free port (Unable to bind to IPv4 or IPv6)")
+    try:
+        free_socket.listen(5)
```
navin772: Why do we need to listen when we are immediately closing the connection later? 
I know this wasn't added in this PR but I think we can remove this (and see if all tests pass)

cgoldberg: You're right... I think this will work fine without that. I'll remove it.

PR #16079, opened 2025-07-22
[py] Encapsulated Geo Location co-ordinates using `getters/setters`

sandeepsuryaprasad:
### 💥 What does this PR do?
Encapsulated Geo Location co-ordinates using `getters/setters`

### 🔧 Implementation Notes
Rather than having to validate the instance attributes, in this case Geo Location Co-ordinates inside constructor, It's a good idea and a standard practice to encapsulate all geo-location co-ordinates inside `getters/setters`, validate through setter methods. It is very clear from each `setter` method as to what is the attribute that we are setting and what are upper and lower limits to that attribute.

### 🔄 Types of changes

___

### **PR Type**
Other

___

### **Description**
- Refactored GeolocationCoordinates class to use property decorators

- Moved validation logic from constructor to setter methods

- Improved encapsulation with private attributes and getters/setters

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Constructor validation"] -- "refactored to" --> B["Property setters"]
  B --> C["Private attributes"]
  C --> D["Getter methods"]
```

<tr>
  <td>
    <details>
      <summary><strong>emulation.py</strong><dd><code>Property-based validation for GeolocationCoordinates</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/selenium/webdriver/common/bidi/emulation.py

<ul><li>Removed validation logic from <code>__init__</code> method<br> <li> Added property decorators for all coordinate attributes<br> <li> Implemented setter methods with validation for each property<br> <li> Changed attributes to private with underscore prefix</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16079/files#diff-18902e3ee516a208783ae0cb490460a1c9608548b8a3a24a15ed8d5a86ca71b1">+70/-15</a>&nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/py/selenium/webdriver/common/bidi/emulation.py
+++ b/py/selenium/webdriver/common/bidi/emulation.py
@@ -64,7 +64,7 @@ class GeolocationCoordinates:
     @latitude.setter
     def latitude(self, value):
         if not (-90.0 <= value <= 90.0):
-            raise ValueError("Latitude must be between -90.0 and 90.0")
+            raise ValueError("latitude must be between -90.0 and 90.0")
         self._latitude = value
 
     @property
@@ -74,7 +74,7 @@ class GeolocationCoordinates:
     @longitude.setter
     def longitude(self, value):
         if not (-180.0 <= value <= 180.0):
-            raise ValueError("Longitude must be between -180.0 and 180.0")
+            raise ValueError("longitude must be between -180.0 and 180.0")
         self._longitude = value
 
     @property
@@ -84,7 +84,7 @@ class GeolocationCoordinates:
     @accuracy.setter
     def accuracy(self, value):
         if value < 0.0:
-            raise ValueError("Accuracy must be >= 0.0")
+            raise ValueError("accuracy must be >= 0.0")
         self._accuracy = value
 
     @property
@@ -104,7 +104,7 @@ class GeolocationCoordinates:
         if value is not None and self.altitude is None:
             raise ValueError("altitude_accuracy cannot be set without altitude")
         if value is not None and value < 0.0:
-            raise ValueError("Altitude accuracy must be >= 0.0")
+            raise ValueError("altitude_accuracy must be >= 0.0")
         self._altitude_accuracy = value
 
     @property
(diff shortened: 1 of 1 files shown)


navin772 (commented):
I don't think this is required and feels a bit overkill for just validating the coordinates.

If you think of any specific case this would be required, let me know.

sandeepsuryaprasad:
@navin772 The co-ordinates gets validated only while creating an instance. But one can change the attributes even after creating instance.. e.g. 
```
coords = GeolocationCoordinates(45.5, -122.4194, accuracy=10.0)
coords.latitude = "45.5"  # bad value. this will not be caught in the current code on the trunk and this does't go through constructor.
```
So it's Safe and standard practice to have it inside a private attribute and provide access to the user via Getter and Setter interface. Definitely it's not a good practice to have all those attributes validated in constructor. By having separate set of `getters/setters` the code is more clean and readable.

sandeepsuryaprasad:
`setters` are meant for validating the values before attributes are being set. In this case we are actually validating the values of co-ordinates before setting it. So obvious choice is to use `setters`, which I don't feel it as overkill. why would someone use `setters` if it is not for validation? Let me know your thoughts on this. Thanks!

navin772:
> But one can change the attributes even after creating instance

This seems like a valid argument, what do you think @cgoldberg?

cgoldberg:
If we care about enforcing validation of coordinates after the class is instantiated, this PR is good. I don't have a strong preference, but would be fine merging this.

py/selenium/webdriver/common/bidi/emulation.py
```diff
@@ -72,6 +57,76 @@ def __init__(
...
+    @property
+    def altitude_accuracy(self):
+        return self._altitude_accuracy
+
+    @altitude_accuracy.setter
+    def altitude_accuracy(self, value):
+        if value is not None and self.altitude is None:
+            raise ValueError("altitude_accuracy cannot be set without altitude")
+        if value is not None and value < 0.0:
+            raise ValueError("Altitude accuracy must be >= 0.0")
```
cgoldberg: Change the message to "altitude_accuracy must be >= 0.0"

cgoldberg (changes requested):
We should remove the capitalization in all error messages when referring to an attribute name:

i.e. change:

"Longitude must be between -180.0 and 180.0"

to

"longitude must be between -180.0 and 180.0"

sandeepsuryaprasad:
@cgoldberg done! Removed capitalisation of all attributes in error messages.

cgoldberg (approved):
LGTM

Issue #15536, opened 2025-03-30
[🚀 Feature]: [dotnet] Revisit unit tests infrastructure


nvborisenko:
### Feature and motivation

Current situation is:

![Image](https://github.com/user-attachments/assets/97f1b7d2-180c-4a6e-9177-9731f39eb822)

### Usage example

Concerns:
- It could be one test project because Chrome/Edge/Firefox drivers live in one project supported by Selenium (like built-in classes).
- Default namespace in test project is `OpenQA.Selenium` when usually it should be `OpenQA.Selenium.Tests`. I don't understand why it is unusual.
- Avoid any `static` context (`EnvironmentManager` class) - it will allow to look at parallelization. My computer has 6/12 cores, I want to utilize all of them when running Selenium unit tests.
- Provide a possibility to override driver options, like required in https://github.com/SeleniumHQ/selenium/pull/15850

RenderMichael:
I don't have much of the historical context surrounding the design decisions of the test projects. There are some clear reasons I see for a heavily configuration-based `EnvironmentManager`. For example, the `EnvironmentManager.Instance` property is accessible in the `AssemblyFixture` types.

At the same time, I see some room for optimization.

Notably, the above `AssemblyFixture`'s one-time setup starts the web server and the remote server, but not the driver. The one-time teardown closes whatever driver may be opened, likely as a precaution. I fully believe we can move the "current driver" management to the test-side.

Rather than using `appconfig.json` (and all the `Runfiles.Create()` and `runfiles.Rlocation` ceremony surrounding that), we can give `DriverTestFixture` a method like so:

```C#
public abstract class DriverTestFixture
{
    public abstract IWebDriver CreateDriver();
}
```

I can experiment with some potential solutions.

nvborisenko:
Historical decision is very clear: reusing browser across 100 tests is faster than initiating new clean browser per test. I would agree at that long time ago, as a quick approach to run tests. But now I strongly disagree. Reusing shared context breaks main principle of testing: **isolation**. Tests should not be dependent on context, on browser state in our case. _(here I could put a link to ignored test which should not be ignored)._ This is why we currently have `NeedsFreshDriverAttribute`, meaning we have to keep in mind the state of browser, shared between tests. And it is bad in my opinion.

[...]

PR #16097, opened 2025-07-28
[dotnet] Enabling drivers to set log to console.

### LINKED ISSUE #15785: [🐛 Bug]: [dotnet] Selenium Library should never write to my console
### Description

I have my own application, let's say console application. I use many libraries to achieve my goal. No one is writing to console, except I explicitly allow to do it.

https://www.selenium.dev/documentation/webdriver/troubleshooting/logging/

The default logging level is `Info`.

### Reproducible Code

```shell
Selenium is a library! It performs low level operations. If this library wants to output something (for instance internal log messages), I should control it. But library should never output by default!
```

diemol:
### **User description**
This will be false by default, following what the .NET bindings have usually done.

### 🔗 Related Issues
Helps with #12273
Should fix #15785 

### 💥 What does this PR do?
It enables users to send driver logs to the console, disabled by default. 

I could only test it manually by creating a .NET app and using the build as a dependency.

@nvborisenko, could you please help me check this? 

### 🔄 Types of changes

___

### **PR Type**
Enhancement

___

### **Description**
- Add `LogToConsole` property to `DriverService` class

- Enable driver process console output logging (disabled by default)

- Handle mutual exclusivity between `LogToConsole` and `LogPath` in Firefox

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["DriverService"] --> B["LogToConsole Property"]
  B --> C["Process Start Configuration"]
  C --> D["Console Output Control"]
  E["FirefoxDriverService"] --> F["LogPath vs LogToConsole"]
  F --> G["Mutual Exclusivity Logic"]
```

<tr>
  <td>
    <details>
      <summary><strong>DriverService.cs</strong><dd><code>Add LogToConsole property and output redirection</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/DriverService.cs

<ul><li>Add <code>LogToConsole</code> boolean property with XML documentation<br> <li> Configure process to redirect standard output/error when <code>LogToConsole</code> <br>is false</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16097/files#diff-c38089621d41671b4719164b262303e275d4a4a442125fd9d4ea61eeca855c2b">+13/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>FirefoxDriverService.cs</strong><dd><code>Handle LogToConsole and LogPath mutual exclusivity</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/Firefox/FirefoxDriverService.cs

<ul><li>Add logic to handle mutual exclusivity between <code>LogToConsole</code> and <br><code>LogPath</code><br> <li> Disable <code>LogToConsole</code> when <code>LogPath</code> is specified</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16097/files#diff-8a02c4932a0154eb11aafcbf7e736968b70fe5d69dac3df52ac67fa8c8a7c965">+7/-3</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___


dotnet/src/webdriver/Firefox/FirefoxDriverService.cs
```diff
@@ -213,9 +213,13 @@ protected override void OnDriverProcessStarting(DriverProcessStartingEventArgs e
             // Initialize the log writer
             logWriter = new StreamWriter(this.LogPath, append: true) { AutoFlush = true };
 
-            // Configure process to redirect output
-            eventArgs.DriverServiceProcessStartInfo.RedirectStandardOutput = true;
-            eventArgs.DriverServiceProcessStartInfo.RedirectStandardError = true;
+            // LogToConsole and LogPath are mutually exclusive. LogPath takes precedence.
+            if (this.LogToConsole)
+            {
+                this.LogToConsole = false;
```
nvborisenko: Don't change initial state of the object, it is confusing.

dotnet/src/webdriver/DriverService.cs
```diff
@@ -151,6 +151,13 @@ public int ProcessId
     /// </summary>
     public string? DriverServicePath { get; set; }
 
+    /// <summary>
+    /// Gets or sets a value indicating whether the driver process console output should be logged
+    /// to the console. Defaults to <see langword="false"/>, meaning console output from the
+    /// driver will be suppressed.
+    /// </summary>
+    public bool LogToConsole { get; set; }
```
nvborisenko: I don't like both `LogToFile` and `LogToConsole` properties. In my opinion it could be one event, where user is able to get incoming log messages from service.

Usage example:
```
service.OnLogMessage += (sender, args) => { Console.WriteLine(args.Message); }
```

dotnet/src/webdriver/Firefox/FirefoxDriverService.cs
diemol: How would you do it? As both are exclusive.

dotnet/src/webdriver/DriverService.cs
diemol: `LogToFile`? Where is that being used?

dotnet/src/webdriver/Firefox/FirefoxDriverService.cs
diemol: Ah, that's okay. We just leave it as the user set it and document that `LogPath` has precedence.

dotnet/src/webdriver/DriverService.cs
nvborisenko: `LogPath` property, where naturally it is a command argument. Let's skip it, it just mimics command arguments, seems acceptable.

But let's think more about `LogToConsole` boolean property. It is unclear to what console, `stderr` or `stdout`? I believe it should be event, then user is able to do everything he wants.

diemol: The idea is to show what the browser driver process is outputting. This mimics what other language bindings are doing. Not more, not less.

nvborisenko: If other bindings have the same boolean flag with this name then it is OK, alignment is better, despite that the approach might be better. Thanks.

diemol: They don't have a boolean flag like this because their default is to show the logs from the browser driver process. The way they enable a user to turn it off is by setting the log level on each browser driver. The user can set it to off so they don't see output on their console. 

So you think that we should log the messages using `Console.WriteLine` instead of redirecting the output?

nvborisenko: I propose to be aligned. If they show driver's output by default, then let's do it. Meaning https://github.com/SeleniumHQ/selenium/issues/15785 is not resolved.

In any case `bool LogToConsole` property seems strange.

diemol: Yes and no. I think .NET users are used to not having any output on their consoles. So it is unlikely that one of them knows that this is the default behavior in other languages. I would say we need to leave this off by default to continue with the same idea .NET has had for a while.

nvborisenko: But the fact is users in .NET see driver service console output in default... And I agree, implicit output should be hidden.

diemol: They don't always see it. If they use MSTest, that output gets hidden. 

In any case, the output is now being sent to the console based on the flag and using the log you implemented.

diemol: Is there anything left or can this be resolved?

nvborisenko: I don't like `LogToConsole` property. But who cares..

nvborisenko: Literally meaning go ahead, as you usuauly do it.

diemol: What is your proposal? As you suggested, the user can now receive messages based on the log level they configure. 

Is it the flag name or the approach? I think the review you are giving is quite ambiguous.

nvborisenko: I am happy that now I will not see spam in my console by default. It is amazing for the library! As I understand this spam now is redirected to internal logging, it is also amazing!

Now lets think more about why I may see spam in my console. Obviously to collect the logs. And it is already possible via internal logging you implemented. I already have an option to reveal logs from driver service to my console.

Moreover, If I want to redirect driver service logs to another stream, I am also able to do it.

So the good question is do we really need unnecessary `LogToConsole` boolean flag? I don't think so.

diemol: Got it. We will let the user control the output through the log level, and we do not need to rely on the `LogToConsole` flag.

diemol: I ended up adding an internal variable: `WriteDriverLogToConsole`. With this, the `DriverService` won't consume the stream when the `FirefoxDriverService` sends the log stream to the file.

dotnet/src/webdriver/DriverService.cs
```diff
@@ -305,6 +317,16 @@ protected virtual void OnDriverProcessStarted(DriverProcessStartedEventArgs even
             throw new ArgumentNullException(nameof(eventArgs), "eventArgs must not be null");
         }
 
+        if (this.WriteDriverLogToConsole && eventArgs.StandardOutputStreamReader != null)
+        {
+            _ = Task.Run(() => ReadStreamAsync(eventArgs.StandardOutputStreamReader, "stdout"));
```
nvborisenko: Don't use async Tasks to capture log messages. By the nature async is unpredictable when exactly it will be executed, meaning for us we will see messed log messages.

dotnet/src/webdriver/DriverService.cs
```diff
@@ -386,4 +408,37 @@ private bool WaitForServiceInitialization()
...
+                        _logger.Error(line);
+                    }
+                }
+            }
+        }
+        catch (Exception ex)
+        {
+            if (_logger.IsEnabled(LogEventLevel.Error))
+            {
+                _logger.Error($"Error reading stream: {ex.Message}");
```
nvborisenko:
```suggestion
                _logger.Error($"Error reading stream: {ex}");
```

Reveal the entire exception with stacktrace.

dotnet/src/webdriver/DriverService.cs
```diff
@@ -151,6 +154,12 @@ public int ProcessId
     /// </summary>
     public string? DriverServicePath { get; set; }
 
+    /// <summary>
+    /// Collects the driver log output and writes it to the console. Defaults to <see langword="true"/>.
+    /// Internal variable to avoid using the stream when logs are sent to a file.
+    /// </summary>
+    protected bool WriteDriverLogToConsole { get; set; } = true;
```
nvborisenko: I am pretty sure we don't need this property, despite its name is not clear. Just don't introduce it.

diemol: In the case of sending GeckoDriver logs to a file, if we don't have a flag, even if it is internal, the streams get consumed by the console and the file. This ends up in an error because the console consumes the stream, and nothing gets sent to the file. That is why we need it.

dotnet/src/webdriver/DriverService.cs
diemol: Good suggestion. I wasn't aware of that in .NET.

diemol: Just pushed changes to do sync threads instead.

nvborisenko:
Please see minimal example how to capture output from process. Probably it helps.

```csharp
var service = new MyService();
service.Start();

class MyService
{
    public void Start()
    {
        var process = new Process();

        process.StartInfo.FileName = "ping";
        process.StartInfo.Arguments = "google.com";
        process.StartInfo.UseShellExecute = false;
        process.StartInfo.CreateNoWindow = true;

        process.StartInfo.RedirectStandardOutput = true;
        process.StartInfo.RedirectStandardError = true;

        process.OutputDataReceived += OnDataReceived;
        process.ErrorDataReceived += OnDataReceived;

        process.Start();

        process.BeginOutputReadLine();
        process.BeginErrorReadLine();

        process.WaitForExit();
    }

    protected virtual void OnDataReceived(object sender, DataReceivedEventArgs args)
    {
        if (string.IsNullOrEmpty(args.Data))
            return;

        // This method can be used to handle data received from the process
        Console.WriteLine($"Data received: {args.Data}");
    }
}
```

diemol:
Thanks. I did try the approach and made changes to use the idea you suggest, but then I end up with errors like this one:

```
1) Error : OpenQA.Selenium.Firefox.FirefoxDriverServiceTest.ShouldRedirectGeckoDriverLogsToConsole
System.InvalidOperationException : Cannot mix synchronous and asynchronous operation on process stream.
   at System.Diagnostics.Process.BeginOutputReadLine()
   at OpenQA.Selenium.DriverService.Start()
   at OpenQA.Selenium.Remote.DriverServiceCommandExecutor.ExecuteAsync(Command commandToExecute)
   at OpenQA.Selenium.WebDriver.ExecuteAsync(String driverCommandToExecute, Dictionary`2 parameters)
   at OpenQA.Selenium.WebDriver.StartSession(ICapabilities capabilities)
   at OpenQA.Selenium.Firefox.FirefoxDriver..ctor(FirefoxDriverService service, FirefoxOptions options, TimeSpan commandTimeout)
   at OpenQA.Selenium.Firefox.FirefoxDriver..ctor(FirefoxDriverService service, FirefoxOptions options)
   at OpenQA.Selenium.Firefox.FirefoxDriverServiceTest.ShouldRedirectGeckoDriverLogsToConsole() in ./dotnet/test/firefox/FirefoxDriverServiceTest.cs:line 88
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

Do you have any suggestions? I am only trying to log messages to console here, not even trying to log to a file yet.

diemol:
Please disregard the message above, I found out why this happens.

nvborisenko:
@diemol can I help here and implement it from scratch if you don't mind?

dotnet/src/webdriver/DriverProcessStartedEventArgs.cs
```diff
@@ -41,31 +40,10 @@ public DriverProcessStartedEventArgs(Process driverProcess)
         }
 
         this.ProcessId = driverProcess.Id;
-        if (driverProcess.StartInfo.RedirectStandardOutput && !driverProcess.StartInfo.UseShellExecute)
```
diemol, line 44: This might be to aggressive but since we don't want to do sync reads, we might delete it. Or deprecate first?

diemol: I did a quick search in GitHub and other places for `DriverProcessStartedEventArgs` and did not see mentions.

diemol:
@nvborisenko, I have not seen your comment before. I just pushed the code I had locally using the approach you recommended. 

Please feel free to create a PR from scratch or review this. Bear in mind that the changes to `FirefoxDriverService` fix a bug introduced in the PR that sent logs to a file. It also changes the approach that PR had to the one you suggest.

nvborisenko:
I quickly implemented these changes, and observed you did it already in pretty similar way. Let's continue here with your changes.

dotnet/src/webdriver/DriverService.cs
nvborisenko: Now, when we are listening logs as sync, finally we can remove this unnecessary property?

dotnet/src/webdriver/DriverService.cs
```diff
@@ -243,6 +252,12 @@ public void Start()
         this.driverServiceProcess.StartInfo.UseShellExecute = false;
         this.driverServiceProcess.StartInfo.CreateNoWindow = this.HideCommandPromptWindow;
 
+        this.driverServiceProcess.StartInfo.RedirectStandardOutput = true;
+        this.driverServiceProcess.StartInfo.RedirectStandardError = true;
+
+        this.driverServiceProcess.OutputDataReceived += (s, e) => this.OnDriverProcessDataReceived(s, e, isError: false);
```
nvborisenko: Please rename:
`s` => `sender`
`e` => `args`

dotnet/src/webdriver/DriverService.cs
```diff
@@ -308,6 +326,27 @@ protected virtual void OnDriverProcessStarted(DriverProcessStartedEventArgs even
...
+    /// <param name="args">The data received event arguments.</param>
+    /// <param name="isError">A value indicating whether the data received is from the error stream.</param>
+    protected virtual void OnDriverProcessDataReceived(object sender, DataReceivedEventArgs args, bool isError)
+    {
+        if (string.IsNullOrEmpty(args.Data))
+            return;
+
+        if (this.WriteDriverLogToConsole && !isError && _logger.IsEnabled(LogEventLevel.Info))
+        {
+            _logger.Info(args.Data);
```
nvborisenko: We have never used `Info` level yet, it was preserved for really high level operations: like "starting browser" or "strating driver". Now the default level is `Warn`, and we can start using `Info` level.

But I think logs from driver service is something low-level, I propose always put them to `Trace` level.

dotnet/src/webdriver/DriverService.cs
```diff
@@ -308,6 +326,27 @@ protected virtual void OnDriverProcessStarted(DriverProcessStartedEventArgs even
...
+        }
+        if (this.WriteDriverLogToConsole && isError && _logger.IsEnabled(LogEventLevel.Error))
+        {
+            _logger.Error(args.Data);
```
nvborisenko: From one point of view it might be reasonable. From another point of view `Error` level means "hey user, pay attention on this and fix it, or ask Selenium team to fix it!".

`Error` level is always revealed to end users, by default. I don't think user is able to fix errors came from driver services. I propose to redirect it as our `Trace`.

dotnet/src/webdriver/Firefox/FirefoxDriverService.cs
```diff
@@ -19,10 +19,11 @@
 
 using OpenQA.Selenium.Internal;
 using System;
+using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Text;
-using System.Threading.Tasks;
+using OpenQA.Selenium.Internal.Logging;
```
nvborisenko: Order namespaces (minor, up to you)

dotnet/src/webdriver/Firefox/FirefoxDriverService.cs
```diff
@@ -219,56 +221,55 @@ protected override string CommandLineArguments
...
+        if (string.IsNullOrEmpty(args.Data))
+            return;
+
+        if (!string.IsNullOrEmpty(this.LogPath))
         {
-            _ = Task.Run(() => ReadStreamAsync(eventArgs.StandardOutputStreamReader));
+            if (logWriter != null)
+            {
+                logWriter.WriteLine(args.Data);
+                logWriter.Flush();
```
nvborisenko: `Flush` can be removed, because `AutoFlush` is enabled.

nvborisenko: Just to improve performance.

dotnet/src/webdriver/DriverProcessStartedEventArgs.cs
nvborisenko: I was surprised here too. My initial thoughts were "no, don't expose it, it is so internal".

I propose to... not sure what. Let's leave this thread unresolved.

May be:
- deep dive and find correct combination of conditions when we are safe
- aj, remove it..

Let's wait little bit.

dotnet/test/firefox/FirefoxDriverServiceTest.cs
```diff
@@ -45,9 +72,44 @@ public void ShouldRedirectGeckoDriverLogsToFile()
         }
         finally
         {
-            driver2.Quit();
             File.Delete(logPath);
         }
     }
 
+    [Test]
+    public void ShouldRedirectGeckoDriverLogsToConsole()
```
nvborisenko, line 80: Specific driver tests are ignored. It will be nice to reincarnate them here: https://github.com/SeleniumHQ/selenium/issues/15536

Read and resolve this thread.

diemol: No, these tests are executed. I added a bazel file to this directory in a different PR.

dotnet/src/webdriver/DriverService.cs
diemol: Yeah, now with the approach it has now, this is unnecessary.

dotnet/src/webdriver/DriverProcessStartedEventArgs.cs
diemol: What do you mean by leaving this thread unresolved? What should we wait for?

dotnet/src/webdriver/DriverService.cs
diemol: Renamed.

dotnet/src/webdriver/DriverService.cs
diemol: Good point. Now all of it is at the trace level.

dotnet/src/webdriver/DriverService.cs
diemol: Good point. Now all of it is at the trace level.

dotnet/src/webdriver/Firefox/FirefoxDriverService.cs
diemol: This import was actually not needed.

dotnet/src/webdriver/Firefox/FirefoxDriverService.cs
diemol: Removed.

dotnet/src/webdriver/DriverProcessStartedEventArgs.cs
nvborisenko: I was thinking a little bit and came into conclusion there is no reason to expose `StandardOutputStreamReader` because it cannot be used anyway. So just removing it is good, even aggressively in my opinion.

nvborisenko (approved):
Thanks, long await feature/issue is now resolved! Good work.

PR #16131, opened 2025-08-05
[dotnet] Truncate log and set profile root in Firefox (GeckoDriver)

diemol:
### 🔗 Related Issues

Helps with #12273

### 💥 What does this PR do?
Allows Firefox to set truncated logs and set profile root

### 🔄 Types of changes

- Cleanup (formatting, renaming)
- Bug fix (backwards compatible)
- New feature (non-breaking change which adds functionality *and tests!*)
- Breaking change (fix or feature that would cause existing functionality to change)

___

### **PR Type**
Enhancement

___

### **Description**
- Add `LogTruncate` property to control GeckoDriver log truncation

- Add `ProfileRoot` property to specify profile directory

- Include directory validation for profile root path

- Update command line argument generation logic

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["FirefoxDriverService"] --> B["LogTruncate Property"]
  A --> C["ProfileRoot Property"]
  B --> D["--log-no-truncate Argument"]
  C --> E["--profile-root Argument"]
  C --> F["Directory Validation"]
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>FirefoxDriverService.cs</strong><dd><code>Add log truncation and profile root configuration</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/Firefox/FirefoxDriverService.cs

<ul><li>Added <code>LogTruncate</code> boolean property for controlling log truncation<br> <li> Added <code>ProfileRoot</code> string property for specifying profile directory<br> <li> Added directory existence validation for ProfileRoot<br> <li> Updated CommandLineArguments to include new command line flags</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16131/files#diff-8a02c4932a0154eb11aafcbf7e736968b70fe5d69dac3df52ac67fa8c8a7c965">+25/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></td></tr></tr></tbody></table>

</details>

___

PR #16138, opened 2025-08-06
Renaming variable and converting it to nullable bool?

diemol:
### 🔗 Related Issues

Addresses https://github.com/SeleniumHQ/selenium/pull/16131#discussion_r2256067508

### 🔄 Types of changes

### DIFF
--- a/dotnet/src/webdriver/Firefox/FirefoxDriverService.cs
+++ b/dotnet/src/webdriver/Firefox/FirefoxDriverService.cs
@@ -104,9 +104,13 @@ public sealed class FirefoxDriverService : DriverService
     public string? LogPath { get; set; }
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


dotnet/src/webdriver/Firefox/FirefoxDriverService.cs
```diff
@@ -199,7 +203,7 @@ protected override string CommandLineArguments
                 argsBuilder.Append(" --jsdebugger");
             }
 
-            if (this.LogNoTruncate)
+            if (this.LogTruncate.HasValue && this.LogTruncate.Value)
```
nvborisenko:
```suggestion
            if (this.LogTruncate is true)
```

nvborisenko (approved):
Thanks, lgtm with comment.

PR #15547, opened 2025-04-02
[grid] Migrate from Guava's CacheBuilder to Caffeine

VietND96:
### **User description**
**Thanks for contributing to Selenium!**
**A PR well described will help maintainers to quickly review and merge it**

Before submitting your PR, please check our [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) guidelines.
Avoid large PRs, help reviewers by making them as simple and short as possible.

### Motivation and Context
The javadoc of Guava's CacheBuilder recommends moving to Caffeine. So, besides concrete issues, it makes sense to switch.

https://github.com/google/guava/blob/master/guava/src/com/google/common/cache/package-info.java#L16-L17

More details here: https://github.com/google/guava/blob/master/guava/src/com/google/common/cache/CacheBuilder.java#L49

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist

- [ ] I have read the [contributing](https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

___

### **PR Type**
Enhancement

___

### **Description**
- Migrate from Guava's CacheBuilder to Caffeine for improved performance

- Add custom cache weigher with memory-aware sizing in GraphqlHandler

- Implement proper resource cleanup with AutoCloseable interface

- Update dependency versions (GraphQL, Netty, OpenTelemetry, Log4j)

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Guava CacheBuilder"] --> B["Caffeine Cache"]
  B --> C["Custom Weigher"]
  B --> D["Memory-aware Sizing"]
  B --> E["AutoCloseable Support"]
  F["Dependency Updates"] --> G["GraphQL 24.1"]
  F --> H["Netty 4.2.3"]
  F --> I["OpenTelemetry 1.52.0"]
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
[...]

PR #16154, opened 2025-08-10
[grid] Improve readTimeout and cache expiration of HttpClient between Router and Nodes

VietND96:
### 💥 What does this PR do?
- Add back Cache Builder for HttpClient in Router (use Caffeine Cache Builder, which was recently done via #15547)
- Remove the hardcoded value of 2 minutes for the timeout of stale connections.
- Implement dynamic expiry based on HttpClient read timeout (utilize `caffeine.cache.Expiry`). 
- Implement a private method to GET /status NodeStatus and get the node session timeout of the corresponding Node and use it for cache expiration (silent fallback default readTime of ClientConfig if node session timeout N/A)
- Use read timeout to set `expireAfterCreate` (the first time HttpClient is added to cache), `expireAfterRead` (when HttpClient is accessed from cache).
- HttpClient read timeout and cache expiration are more reliable now.

Here is few logs when enable log-level `FINE`

> 14:50:16.912 DEBUG [HandleSession.lambda$loadSessionId$2] - Creating new HTTP client for http://172.18.0.13:5555
14:50:16.927 DEBUG [JdkHttpClient.execute0] - Executing request: (GET) /status
14:50:16.987 DEBUG [HandleSession.lambda$loadSessionId$2] - Created connection for http://172.18.0.13:5555 (read timeout: 300 seconds)
14:50:16.988 DEBUG [HandleSession$1.expireAfterCreate] - Set (read timeout: 300 seconds) as initial expiration for http://172.18.0.13:5555 in cache
14:50:16.989 DEBUG [HandleSession$1.expireAfterRead] - Set (read timeout: 300 seconds) as expiration after read for http://172.18.0.13:5555 in cache

### 🔄 Types of changes

___

### **PR Type**
Enhancement

___

### **Description**
- Replace manual HTTP client cache with Caffeine cache library

- Simplify resource management with automatic cleanup

- Add proper executor service shutdown in GridStatusHandler

- Remove complex manual cache eviction logic

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Manual Cache Management"] --> B["Caffeine Cache"]
  C["Custom CacheEntry Class"] --> D["Direct HttpClient Storage"]
  E["Manual Cleanup Thread"] --> F["Automatic Expiry & Removal"]
  G["Complex Usage Counting"] --> H["Simple Cache Access"]
```

<tr>
  <td>
    <details>
      <summary><strong>HandleSession.java</strong><dd><code>Replace manual cache with Caffeine implementation</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

java/src/org/openqa/selenium/grid/router/HandleSession.java

<ul><li>Replace ConcurrentHashMap with Caffeine cache for HTTP clients<br> <li> Remove manual cleanup thread and usage counting logic<br> <li> Simplify cache entry management with automatic expiry<br> <li> Add removal listener for proper HTTP client cleanup</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16154/files#diff-3f58425b642a490159052d431ed753fa75cf79a4d5e756ed37dc26039983c59a">+42/-102</a></td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>GridStatusHandler.java</strong><dd><code>Add proper resource cleanup</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

java/src/org/openqa/selenium/grid/router/GridStatusHandler.java

<ul><li>Implement AutoCloseable interface<br> <li> Add graceful executor service shutdown<br> <li> Import ExecutorServices utility for cleanup</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16154/files#diff-93e80d00adc4680d45cd2a87730f14dc6ab5f9d40dd0440a3f3741c8ae1cb453">+10/-1</a>&nbsp; &nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Dependencies</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>BUILD.bazel</strong><dd><code>Add Caffeine dependency</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

java/src/org/openqa/selenium/grid/router/BUILD.bazel

- Add Caffeine cache library dependency

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16154/files#diff-0c31cf4a0201d2360ad41433f0462f18ad406d6fd724116f6ea3f7ebbf62932b">+1/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___


java/src/org/openqa/selenium/grid/router/GridStatusHandler.java
```diff
@@ -165,4 +168,10 @@ public HttpResponse execute(HttpRequest req) {
       return res;
     }
   }
+
+  @Override
+  public void close() {
+    LOG.info("Shutting down GridStatusHandler executor service");
+    ExecutorServices.shutdownGracefully("Grid status executor", EXECUTOR_SERVICE);
```
joerg1985: Closing the static field will break all other instances of `GridStatusHandler`

joerg1985 (changes requested):
@VietND96  Switching back to a standard cache will bring back the issue descriped in #12978. Therefore this PR shouldn't be merged in my mind.

VietND96:
@joerg1985 Ok I see. Since I just tried this after moving on to Caffeine cache builder recently.
Cache builder only care access timeout, while the connection uses read timeout. I missed this session handle includes the connection from client also.

VietND96:
@joerg1985, can you review again? I spent hours reading Caffeine support, now the cache expiration is aligned with read timeout and dynamically based on the destination node session timeout (I also handle resilient fallback to get default ClientConfig if without /status endpoint).

java/src/org/openqa/selenium/grid/router/GridStatusHandler.java
VietND96: I reverted this point

joerg1985:
@VietND96 i don't think it is enought to use a different timeout, as the connection timeout and delays in processing (e.g. gc pauses) must be respected too, to ensure the client is not closed before the response is processed. Additionally the timeout handling inside the JDK is not perfect (there is an open issue regarding this in the JDK Jira), so the client might get a worse response as the client is closed than from the real timeout. 


The current timeout handling will calculate the inactivity from the last time the client did return a result and not from the last time the client was used. So a client might get closed to fast and the next request to the session has to start a new client first.

VietND96:
Hmm, so for example, read timeout is 300s, at the second 100th, HttpClient is loaded from cache to execute a request, expiration immediately reset back to 300s, also not a safe enough protocol?

joerg1985:
I think usage counting is the correct pattern to ensure the client is not in use.

Perhaps @ben-manes can tell if there is a matching pattern inside Coffein or it might be interresting to have such a mechanism integrated to Caffeine. The basic problem is a long running operation on the cached object and it should not be disposed while this operation is running. This does not sound to exotic to handle it inside a library to me.

java/src/org/openqa/selenium/grid/router/HandleSession.java
```diff
@@ -60,92 +60,150 @@
...
+    }
+
+    /**
+     * Check if this cache entry has expired based on the HttpClient's read timeout. Method is used
+     * by Cache builder to determine if an entry should be evicted.
+     *
+     * @param lastAccessTime the last time this entry was accessed
+     * @return true if the entry should be evicted
+     */
+    boolean isExpired(long lastAccessTime) {
```
joerg1985: I could not find any call to this method, is guess this can be removed or called 😀

VietND96: Yes, I am reworking this based on your feedback. Actually, when using Expiry (by create, read, update Cache), we could not overwrite the expired logic, so this isExpired is redundant and useless.
You're correct, client is not in use is still a safe protocol. I am thinking how can implement cache with this pattern.

java/src/org/openqa/selenium/grid/router/HandleSession.java
```diff
@@ -60,92 +60,150 @@
...
-              thread.setName("HandleSession - Clean up http clients cache");
-              return thread;
-            });
-    cleanUpHttpClientsCacheService.scheduleAtFixedRate(
-        GuardedRunnable.guard(cleanUpHttpClients), 1, 1, TimeUnit.MINUTES);
+    // Create Cache with dynamic expiry based on HttpClient read timeout
+    // and a removal listener to close HTTP clients
+    this.httpClientCache =
+        Caffeine.newBuilder()
+            .expireAfter(
```
ben-manes: If you don't need to log based on the operation type, you could use
```java
Expiry.accessing((key, value) -> {
  LOG.fine("Set (read timeout: {} seconds) for {} in cache",
      cacheEntry.getConfig().readTimeout().toSeconds(), uri);
  return cacheEntry.getConfig().readTimeout()
});
```

ben-manes:
>  The basic problem is a long running operation on the cached object and it should not be disposed while this operation is running. This does not sound to exotic to handle it inside a library to me.

You can use [pinning](https://github.com/ben-manes/caffeine/wiki/Faq#pinning-entries) if you inform the cache of the long-running operation to exclude the entry from consideration. That requires calling into the cache to modify the entry's metadata. That can be difficult to retrofit and there isn't an entry callback for in-use since that would degrade eviction to O(n) and unbounded usage. You might consider using a two-level cache with the primary as weak reference for reference counting that loads and evicts from an expiring cache.

java/src/org/openqa/selenium/grid/router/HandleSession.java
```diff
@@ -60,150 +67,254 @@
...
+      entry.inUse.decrementAndGet();
+    }
+  }
+
+  /**
+   * Custom weigher that implements "pinning" by assigning very high weight to entries with inUse >
+   * 0. This prevents Caffeine from evicting active connections during size-based eviction.
+   *
+   * <p>Weight Strategy: - inUse == 0: Weight = 1 (normal, can be evicted) - inUse > 0: Weight =
+   * Integer.MAX_VALUE (effectively pinned, won't be evicted)
```
ben-manes, line 196: pinning has a weight of zero. It represents how much space the entry takes, so MAX_VALUE is well everything.

VietND96:
@joerg1985 I added back `inUse` and `lastUse` were introduced by you before, and implemented cache around it
I also tried to use pinning entries. Can you also review the usage is correct @ben-manes ?

java/src/org/openqa/selenium/grid/router/HandleSession.java
```diff
@@ -220,42 +389,104 @@ private Callable<UsageCountingReverseProxyHandler> loadSessionId(
...
+                      + sessionTimeout.toSeconds()
+                      + " seconds)");
+            } else {
+              in.skipValue();
+            }
+          }
+          in.endObject();
+        } else {
+          in.skipValue();
+        }
```
joerg1985, line 461: We should call /se/grid/node/status to get only the status this will allow to use org.openqa.selenium.grid.web.Values.get(...); to read the NodeStatus response.

java/src/org/openqa/selenium/grid/router/HandleSession.java
```diff
@@ -220,42 +389,104 @@ private Callable<UsageCountingReverseProxyHandler> loadSessionId(
...
+            newEntry.inUse.decrementAndGet();
             throw t;
           }
         });
   }
 
+  private ClientConfig fetchNodeSessionTimeout(URI uri) {
+    ClientConfig config = ClientConfig.defaultConfig().baseUri(uri).withRetries();
+    Duration sessionTimeout = config.readTimeout();
+    HttpClient httpClient = httpClientFactory.createClient(config);
```
joerg1985, line 434: close the httpClient after use

java/src/org/openqa/selenium/grid/router/HandleSession.java
```diff
@@ -220,42 +389,104 @@ private Callable<UsageCountingReverseProxyHandler> loadSessionId(
...
+          in.skipValue();
+        }
+      }
+    } catch (Exception e) {
+      LOG.fine(
+          "Use default from ClientConfig (read timeout: "
+              + config.readTimeout().toSeconds()
+              + " seconds)");
+    }
+    config = config.readTimeout(sessionTimeout);
```
joerg1985, line 469: add the default connection timeout to the sessionTimeout, as the node might need to connect e.g. a relay / docker / what ever node

java/src/org/openqa/selenium/grid/router/HandleSession.java
```diff
@@ -220,42 +389,104 @@ private Callable<UsageCountingReverseProxyHandler> loadSessionId(
...
-            cacheEntry.inUse.decrementAndGet();
+            // Ensure we don't keep the http client when an unexpected throwable is raised
+            newEntry.inUse.decrementAndGet();
             throw t;
           }
         });
   }
 
+  private ClientConfig fetchNodeSessionTimeout(URI uri) {
+    ClientConfig config = ClientConfig.defaultConfig().baseUri(uri).withRetries();
```
joerg1985, line 432: I would remove withRetries and exception handling in case we are unable to fetch the status the request should fail with a matching error (502 Bad Gateway).

joerg1985 (changes requested):
Currently the read timeout is 3 min, so a not responding request will block until the timeout is reached. In case the client does catch the exception and will try to run another command, the other command might pass (e.g. the network is up again).
As soon as the session timeout is used as read timeout the client will be blocked until the session is dead. Now the client is not able to catch the exception an try to execute another command, as the session has be terminated due to the timeout.

But we detect the client does go away (aka. does hit his read timeout) and interrupt the upstream (https://github.com/SeleniumHQ/selenium/commit/71753492d070eb910dafbacfb78e73c0f556e39e). So this should save us from the descriped issue above, as long as the client timeout is lower than the session timeout. 
We should definitely add tests to ensure we have this 'interrupt the upstream' behaviour permanent in the future too, there might be already tests for this i am not aware of.

PR #16163, opened 2025-08-12
[grid] Improve readTimeout in handle session between Router and Node

VietND96:
### 💥 What does this PR do?
HttpClient cache issue described in https://github.com/SeleniumHQ/selenium/pull/12978.
However, the HttpClient created still uses the default ClientConfig
In order to improve the connection between Router and Node for routing commands, Router will fetch the session timeout that Node exposed to set readTimeout in ClientConfig before creating HttpClient (if no session timeout found, silent fallback to default of ClientConfig).

FINE logs as below

```
16:25:36.597 DEBUG [JdkHttpClient.execute0] - Executing request: (GET) /se/grid/node/status

16:25:36.647 DEBUG [JdkHttpClient.execute0] - Ending request (GET) /se/grid/node/status in 140ms

16:25:36.649 DEBUG [HandleSession.fetchNodeSessionTimeout] - Fetched session timeout from node status (read timeout: 300 seconds) for http://172.18.0.12:5555

16:25:36.661 DEBUG [JdkHttpClient.execute0] - Ending request (GET) /se/grid/node/status in 64ms
```

### 🔄 Types of changes

___

### **PR Type**
Enhancement

___

### **Description**
- Improve Router-Node connection timeout handling

- Fetch session timeout from Node status endpoint

- Configure HttpClient readTimeout dynamically

- Add error handling for timeout configuration

___

### Diagram Walkthrough

```mermaid
flowchart LR
  Router["Router"] -- "GET /status" --> Node["Node"]
  Node -- "NodeStatus with sessionTimeout" --> Router
  Router -- "Configure HttpClient with timeout" --> HttpClient["HttpClient"]
  HttpClient -- "Execute session commands" --> Node
```

<tr>
  <td>
    <details>
      <summary><strong>HandleSession.java</strong><dd><code>Dynamic timeout configuration for Router-Node communication</code></dd></summary>
<hr>

java/src/org/openqa/selenium/grid/router/HandleSession.java

<ul><li>Add method to fetch session timeout from Node status endpoint<br> <li> Replace hardcoded ClientConfig with dynamic timeout configuration<br> <li> Add JSON parsing for NodeStatus response<br> <li> Include error handling with fallback to default timeout</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16163/files#diff-3f58425b642a490159052d431ed753fa75cf79a4d5e756ed37dc26039983c59a">+51/-2</a>&nbsp; &nbsp; </td>

</tr>

</details>

___


VietND96:
@joerg1985, ignore the change to bring back Cache builder in session handler. Here I just add to fetch session timeout and set to ClientConfig. Can you review it makes sense?
@diemol, Can you also review this, it would be better if we can deliver this release.

joerg1985:
@VietND96 i did not look too deep into the code, but my first approach would be returning the timeout (and other config) to the response the node does receive when registering it self to the grid. In this case we should have no need for fallbacks or additional execption handling.

What happens in case someone did set a different timeout to the node via the cli? Do we need to handle this? I am not sure ...

VietND96:
@joerg1985, actually Distributor also gets the node session timeout to set session timeout in GridModel.
There is an implement in NodeStatus itself, it always has session timeout (set by user via CLI option or default 300s). So I can remove the additional exception handling here.

> What happens in case someone did set a different timeout to the node via the cli

As I mentioned above on NodeStatus, timeout would be different between Node via its config. By implementing this, we will have a unified timeout between sessions, the connection of the router to a particular Node.

VietND96:
@joerg1985 I fixed one of your review in another PR
> We should call /se/grid/node/status to get only the status this will allow to use org.openqa.selenium.grid.web.Values.get(...); to read the NodeStatus response.

The usage of `org.openqa.selenium.grid.web.Values.get(...);` require response a Map start with key `value`. However, `/se/grid/node/status` does not respond the same contract, so I updated NodeStatus under key `value`. I searched around, and it looks like nowhere uses `/se/grid/node/status`, so it would be fine for backward compatibility.

diemol (approved):

joerg1985 (approved):
I still think the try catch is not needed, but this should not hurt.

ps: The test i mentioned we should add is allready in place `NettyServerTest.doesInterruptPending`.

VietND96:
@joerg1985 after checking again, I removed the catch block, and added `null` check since `Values.get()` might return null when not 200 or error response.

VietND96:
> The usage of org.openqa.selenium.grid.web.Values.get(...); require response a Map start with key value. However, /se/grid/node/status does not respond the same contract, so I updated NodeStatus under key value. I searched around, and it looks like nowhere uses /se/grid/node/status, so it would be fine for backward compatibility.

Actually, it breaks backward compatibility, where a Node with the previous release in grid. The error was early seen in docker-selenium tests. Reverted key `value` in Node response.
I also realized that `/se/grid/node/status` already returned a contract of NodeStatus, just use `Json.toType(string(res), NodeStatus.class)` instead of `Values.get()`.
For a safe protocol, I also added back the `catch` block.
All were done in commit https://github.com/SeleniumHQ/selenium/commit/5d281ef05306e449774fca55a91dfffe95219075

PR #16170, opened 2025-08-13
[py][docs]: update dead API docs link to API reference in `index.rst`

navin772:
### 💥 What does this PR do?

Removes the dead `api.html` link from `index.rst`. This page is rendered in [pypi](https://pypi.org/project/selenium/) and [docs](https://selenium-python-api-docs.readthedocs.io/en/latest/) site, having a dead link at the project docs first page is not ideal.

Removing it and renaming `Docs` to `API Docs` should be good enough.

### 🔄 Types of changes

- Documentation

___

### **PR Type**
Documentation

___

### **Description**
- Remove dead `api.html` link from Python documentation index

- Consolidate documentation links by renaming "Docs" to "API Docs"

- Fix broken link displayed on PyPI and ReadTheDocs sites

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["index.rst table"] --> B["Remove dead api.html link"]
  B --> C["Rename 'Docs' to 'API Docs'"]
  C --> D["Clean documentation links"]
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Documentation</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>index.rst</strong><dd><code>Fix dead link in documentation table</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/docs/source/index.rst

<ul><li>Remove dead <code>api.html</code> link from documentation table<br> <li> Rename "Docs" entry to "API Docs" <br> <li> Consolidate two documentation links into one</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16170/files#diff-2c5a26dfc3d3ae29a17ca6d93cc1397eda3b360017822160ed46218cfefb5651">+1/-3</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/py/docs/source/index.rst
+++ b/py/docs/source/index.rst
@@ -9,17 +9,19 @@ Python language bindings for Selenium WebDriver.
 
 The `selenium` package is used to automate web browser interaction from Python.
 
-+-------------------+------------------------------------------------+
-| **Home**:         | https://selenium.dev                           |
-+-------------------+------------------------------------------------+
-| **GitHub**:       | https://github.com/SeleniumHQ/Selenium         |
-+-------------------+------------------------------------------------+
-| **PyPI**:         | https://pypi.org/project/selenium              |
-+-------------------+------------------------------------------------+
-| **IRC/Slack**:    | https://www.selenium.dev/support/#ChatRoom     |
-+-------------------+------------------------------------------------+
-| **API Docs**:     | https://www.selenium.dev/selenium/docs/api/py  |
-+-------------------+------------------------------------------------+
++-------------------+--------------------------------------------------------+
+| **Home**:         | https://selenium.dev                                   |
++-------------------+--------------------------------------------------------+
+| **GitHub**:       | https://github.com/SeleniumHQ/Selenium                 |
++-------------------+--------------------------------------------------------+
+| **PyPI**:         | https://pypi.org/project/selenium                      |
++-------------------+--------------------------------------------------------+
+| **IRC/Slack**:    | https://www.selenium.dev/support/#ChatRoom             |
++-------------------+--------------------------------------------------------+
+| **Docs**:         | https://www.selenium.dev/selenium/docs/api/py          |
++-------------------+--------------------------------------------------------+
+| **API Reference**:| https://www.selenium.dev/selenium/docs/api/py/api.html |
++-------------------+--------------------------------------------------------+
 
 Updated documentation published with each commit is available at: `readthedocs.io <https://selenium-python-api-docs.readthedocs.io/en/latest>`_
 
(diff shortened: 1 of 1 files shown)


cgoldberg:
I knew this was going to be a problem when I added it :)

The only reason I used a local link for the API Docs was because if you are viewing the nightly docs (https://selenium-python-api-docs.readthedocs.io/) or a local version, and you click that link, you jump to the version of the documentation on selenium.dev.

Also, I think it's worthwhile keeping a separate link to `api.html`, because it is not linked anywhere else in the documentation and is infuriatingly difficult to find via Google search.

I think the best possible solution so we don't have a dead link on PyPI, is to have the 2 links as:

```
**Docs**:     | https://www.selenium.dev/selenium/docs/api/py
**API Docs**: | https://www.selenium.dev/selenium/docs/api/py/api.html
```
(you probably have to widen the table to fit that in)

Maybe we should call it "API Reference" instead?

navin772:
I didn't knew that `api.html` linked to that page (I clicked on pypi page only). We should keep it. `API Reference` sounds better.

But if we update this and when we are on the page - https://selenium-python-api-docs.readthedocs.io/en/selenium-4.35.0/ and then click on newly added `API Reference` link it will take us to https://www.selenium.dev/selenium/docs/api/py/api.html and not https://selenium-python-api-docs.readthedocs.io/en/selenium-4.35.0/api.html.

That might not be ideal since clicking on the links will take to different version.

cgoldberg:
> clicking on the links will take to different version.

Right, that's why I had it as a local link previously... but I think linking to the selenium.dev version is better than having a dead link on PyPI or wherever else displays it.

The issue is that `README.rst` is a symlink to `index.rst`, but we want different links in each version :)

navin772:
> The issue is that `README.rst` is a symlink to `index.rst`, but we want different links in each version :)

Which one is rendered by pypi? Can we remove the symlink and update in the pypi one only? It will definitely increase maintainability but should solve this problem.

cgoldberg:
> Which one is rendered by pypi?

`README.rst` is what is what you see on PyPI or when visiting https://github.com/SeleniumHQ/selenium/tree/trunk/py

`docs/source/index.rst` is what is used in the Sphinx generated docs.

> Can we remove the symlink and update in the pypi one only?

That would solve the problem, but it's kind of nice keeping them in sync as a symlink. We don't have to remember to keep them independently updated.

I think it's best to leave the symlink and update the "API Reference" link to the selenium.dev one. If someone is motivated enough to read our nightly docs, they can probably figure out where the nightly published API reference is.

cgoldberg (approved):
LGTM ... I think that's the best solution.

PR #15916, opened 2025-06-19
[dotnet] [bidi] Add OnHistoryUpdated event

nvborisenko:
### **User description**
https://w3c.github.io/webdriver-bidi/#event-browsingContext-historyUpdated

### 💥 What does this PR do?
Implement the event.

### 🔄 Types of changes

- Cleanup (formatting, renaming)
- Bug fix (backwards compatible)
- New feature (non-breaking change which adds functionality *and tests!*)
- Breaking change (fix or feature that would cause existing functionality to change)

___

### **PR Type**
Enhancement

___

### **Description**
• Add OnHistoryUpdated event for BiDi browsing context
• Implement async handlers for history navigation tracking
• Create HistoryUpdatedEventArgs record class

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>BrowsingContext.cs</strong><dd><code>Add history updated event handlers</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/BiDi/BrowsingContext/BrowsingContext.cs

• Add OnHistoryUpdatedAsync method with Func<Task> handler<br> • Add <br>OnHistoryUpdatedAsync method with Action handler

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/15916/files#diff-e72b719bd0bbb2a35be2b9acaaff6e7d9ad914658bd3069ac10a90e0f1057e39">+10/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>HistoryUpdatedEventArgs.cs</strong><dd><code>Create HistoryUpdatedEventArgs record class</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

[...]

PR #16209, opened 2025-08-19
[py][bidi]: add `downloadEnd` event for browsing context

navin772:
### 💥 What does this PR do?

Adds support for the `downloadEnd` event in bidi browsing context module - https://w3c.github.io/webdriver-bidi/#event-browsingContext-downloadEnd 

1. This event works in Chrome 141 (currently in canary), so the test will be added later.
2. I have also updated `DownloadWillBeginParams`'s `from_json` method to be better as done for other Download classes.

### 🔄 Types of changes

___

### **PR Type**
Enhancement

___

### **Description**
- Add `downloadEnd` event support for BiDi browsing context

- Implement `DownloadCanceledParams` and `DownloadCompleteParams` classes

- Refactor `DownloadWillBeginParams.from_json` method for consistency

- Register new event in BrowsingContext EVENT_CONFIGS

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["NavigationInfo base class"] --> B["DownloadCanceledParams"]
  A --> C["DownloadCompleteParams"]
  B --> D["DownloadEndParams"]
  C --> D
  D --> E["DownloadEnd event class"]
  E --> F["BrowsingContext EVENT_CONFIGS"]
```

<tr>
  <td>
    <details>
      <summary><strong>browsing_context.py</strong><dd><code>Add downloadEnd event classes and BiDi integration</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/selenium/webdriver/common/bidi/browsing_context.py

<ul><li>Add <code>DownloadCanceledParams</code> and <code>DownloadCompleteParams</code> classes<br> <li> Implement <code>DownloadEndParams</code> wrapper class with status-based factory<br> <li> Create <code>DownloadEnd</code> event class for BiDi integration<br> <li> Refactor <code>DownloadWillBeginParams.from_json</code> to use <br><code>NavigationInfo.from_json</code></ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16209/files#diff-a65d02c951aeb477672aa52e5eb0106645ba869aa186c0af6d2672c42cac95c6">+101/-29</a></td>

</tr>

</details>

___


shbenzer:
LGTM - once you’ve got a test or 2 for this I think you’re good to go

navin772:
Current stable browsers are not supporting this event, so can't test it :(

shbenzer:
Hm, that’s an interesting case then. We might want to put this pr on hold… what do y’all think @cgoldberg @diemol ?

cgoldberg:
> Current stable browsers are not supporting this event
> what do y’all think

hmm... that's a tough call. If we release it now, it's not fully tested and not usable by most people... which might cause confusion. If we wait for Chrome/Edge/Firefox to all support it in stable versions, it's safer but we are late to the party and might be holding back useful functionality for a subset of users.

I think either way has a valid argument, so I'm not sure :)

Whatever we decide, we should make it a project policy for timing the release of new BiDi functionality. Have other bindings hit this issue?

navin772:
> Have other bindings hit this issue?

Dotnet adds some new events without tests (like https://github.com/SeleniumHQ/selenium/pull/15916), there are some events in Java which don't have tests - `onDownloadWillBegin`, `onNavigationFailed` which I am planning to add.

diemol:
I think we can add this, and perhaps write in the equivalent of JavaDocs for Python that this is supported on X version of the browser. Otherwise this PR might become stale, like many others we have.

cgoldberg:
> Dotnet adds some new events without tests 

I more meant the issue of releasing features that all browsers don't yet support in stable versions.

I like the idea from @diemol of just adding a note in docstrings that it requires the browser to support this. My only concern is how nicely this fails when run on a browser that doesn't support it. Do we give some cryptic error? Other than that, this seems fine to ship.

shbenzer:
My biggest concern would be that we don’t know if this code truly works (though I’m sure it does, @navin772 is awesome) until at least one browser supports it - I am uncomfortable w/ the precedent of merging untested code.

navin772:
> equivalent of JavaDocs for Python that this is supported on X version of the browser.

Chrome 140 supports this (next stable version) and there is no info regarding edge and firefox, they are currently failing the WPT tests. What should the docstring include in this case?


> My biggest concern would be that we don’t know if this code truly works

@shbenzer This is a valid concern, I just tested this code against Chrome 140 which is currently in beta and it works as expected. We recently started testing python tests against chrome beta in CI-RBE so this test should pass there too.

But the issue will be that we need to selectively skip the chrome stable test in CI (so CI stays green) for which we don't have a xfail marker as of now. Should we add it? Is this a better approach or just wait for 140 to release?

cgoldberg:
> What should the docstring include in this case?

I think we should add something vague in a module-level docstring about BiDi functionality only working in supported browsers. I don't think we should get into detail there about which browsers and versions currently work with each feature, or do this for every method. We don't want to constantly update the individual docstrings and keep an internal version of BiDi feature support. Maybe all of the BiDi modules should have a similar docstring.

I think more important is the behavior and error message when someone tries to use some BiDi functionality on a browser that doesn't support it. It would be nice to fail with an error like "sorry, your browser doesn't support this yet" instead of some cryptic exception (I don't know what currently happens).

navin772:
We get a pretty good error message from the browser which we pass on to the user's stdout. For example, firefox throws this error:
```
selenium.common.exceptions.WebDriverException: Message: invalid argument: browsingContext.downloadEnd is not a valid event name
```
But chrome isn't throwing any error and would fail when no event is received.

Should we modify this behaviour?

shbenzer:
> We get a pretty good error message from the browser which we pass on to the user's stdout. For example, firefox throws this error:
> 
> ```
> selenium.common.exceptions.WebDriverException: Message: invalid argument: browsingContext.downloadEnd is not a valid event name
> ```
> 
> But chrome isn't throwing any error and would fail when no event is received.
> 
> Should we modify this behaviour?

We could append to it w/ some clarifying details

nvborisenko:
My 50 cents: Selenium should not wait until all vendors implement BiDi spec. We can implement the specification in advance. But the question is are we sure we implement it correctly? My personal plan in .NET area is easy: implement method/event, add small subset of tests which works at least with one browser, we are happy - release to end users.

Mentioning something in DocStrings is not necessary, because browser can be updated at some time with supporting of new method/event, thus DoCStrings become invalid. Note: Users might easily request to use `beta/dev/preview` browsers.

cgoldberg:
> Mentioning something in DocStrings is not necessary, because browser can be updated at some time with supporting of new method/event, thus DoCStrings become invalid

That's why I recommended not mentioning specific browsers or versions. I think a note in doctrings that says "some of these classes/methods may not be supported yet in all browsers" is useful until BiDi is fully supported... Users will see it in the API documentation.

nvborisenko:
> I think a note in doctrings that says "some of these classes/methods may not be supported yet in all browsers" is useful until BiDi is fully supported...

It is good to mention it globally, not for particular method/event.

cgoldberg:
> It is good to mention it globally, not for particular method/event

In Python we have module docstrings, which is where I suggested it go.

navin772:
> In Python we have module docstrings, which is where I suggested it go.

@cgoldberg where should I put the docstring? Should it be at the top of the `bidi/browsing_context.py` file? The above docstring you mentioned is applicable for all modules, so is there a better place to put it so it applies to all modules?

cgoldberg:
> where should I put the docstring?

I just looked at the API docs and noticed module level docstrings show up in the Table of Contents next to each module name, so we probably can't put a big block of text. I guess we should hold off on adding this until we figure out a good place. That shouldn't hold up this PR.

cgoldberg (approved):

PR #16250, opened 2025-08-24
[py] Fix type annotation errors in service.py

NoStory-py:
### **User description**
Fixed the instance variable log_output by properly annotating it as int | IOBase | None. Removed unnecessary #type: ignore comments where mypy requires them.

### 🔗 Related Issues
Partially addresses #15697 

### 💥 What does this PR do?
Fixes mypy type annotation errors in 
`selenium/webdriver/common/service.py`:
- Annotated `self.log_output` as `int | IOBase | None`.
- Used `cast()` only where mypy requires it, removed unnecessary `cast()`.
- Removed unnecessary `#type: ignore` comments.

### 🔧 Implementation Notes
- Annotated `self.log_output` as `Optional[Union[int, IOBase]]`.
- Alternative consideration was to simply use the annotation that the parameter `log_output` uses but that also includes `str` which the instance variable will never be `str`.
- added `cast()` on `self.log_output = log_output` because parameter `log_output` is annotated with `IO[Any]` instead of `IOBase`.

- No tests are needed I already tested with mypy and `ruff check`
- This is my first PR if more changes are needed please let me know.

### 🔄 Types of changes

___

### **PR Type**
Other

___

### **Description**
- Fix type annotation errors in service.py

- Remove unnecessary type ignore comments

- Add proper type annotation for log_output instance variable

- Clean up cast() usage where not needed

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Type Annotation Issues"] --> B["Add log_output Type Annotation"]
  B --> C["Remove Unnecessary Casts"]
  C --> D["Remove Type Ignore Comments"]
  D --> E["Clean Python Code"]
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Miscellaneous</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>service.py</strong><dd><code>Fix type annotations and remove unnecessary casts</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/selenium/webdriver/common/service.py

<ul><li>Added explicit type annotation for <code>log_output</code> instance variable<br> <li> Removed unnecessary <code>cast()</code> calls in conditional branches<br> <li> Removed <code># type: ignore</code> comments for Windows subprocess attributes<br> <li> Added single <code>cast()</code> where mypy requires it for type compatibility</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16250/files#diff-646e4c91d7ef7fbd2b4c088347c86f5b2d2cab8121b14987f3cb783ad120634c">+7/-6</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___


py/selenium/webdriver/common/service.py
```diff
@@ -208,9 +209,9 @@ def _start_process(self, path: str) -> None:
         try:
             start_info = None
             if system() == "Windows":
-                start_info = subprocess.STARTUPINFO()  # type: ignore[attr-defined]
-                start_info.dwFlags = subprocess.CREATE_NEW_CONSOLE | subprocess.STARTF_USESHOWWINDOW  # type: ignore[attr-defined]
-                start_info.wShowWindow = subprocess.SW_HIDE  # type: ignore[attr-defined]
+                start_info = subprocess.STARTUPINFO()
+                start_info.dwFlags = subprocess.CREATE_NEW_CONSOLE | subprocess.STARTF_USESHOWWINDOW
+                start_info.wShowWindow = subprocess.SW_HIDE
```
navin772: You need to add back `type: ignore` else mypy will raise type errors. I think mypy is not smart enough here since the subprocess package does contains the `STARTUPINFO()` and other classes.

navin772 (changes requested):

py/selenium/webdriver/common/service.py
NoStory-py: I see ok I will add them back, although for me mypy says `error: Unused "type: ignore" comment  [unused-ignore]` about these comments and yea after removing them no new error appears. I used mypy by doing `tox -e py/tox.ini -e mypy`

NoStory-py: Done I added the comments back

navin772: Weird, I was getting below errors when type ignore was removed:
```
selenium/webdriver/common/service.py:212: error: Module has no attribute "STARTUPINFO"  [attr-defined]
selenium/webdriver/common/service.py:213: error: Module has no attribute "CREATE_NEW_CONSOLE"  [attr-defined]
selenium/webdriver/common/service.py:213: error: Module has no attribute "STARTF_USESHOWWINDOW"  [attr-defined]
selenium/webdriver/common/service.py:214: error: Module has no attribute "SW_HIDE"  [attr-defined]
```

NoStory-py: Yea I figured you were getting errors. 

It's probably because I am using windows so mypy recognises the 'type' of this windows specific code on my end. You are right its better to keep this comments to maintain consistency over all os

cgoldberg:
> It's probably because I am using windows so mypy recognises the 'type' of this windows specific code on my end.

This is exactly the issue. We are using `platform.system()` for detecting Windows, but mypy doesn't respect that:
 
https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks

I wonder if we should change our platform detection across the entire codebase to use `sys.platform` instead? I thought `platform.system()` was the newer/better way, but it's causing issues here.

I don't really care if we leave the comments in this PR (we're going to get an error on one platform or the other either way). I'll open a new issue to investigate switching to `sys.platform` to avoid this.

navin772 (approved):
LGTM! We can take care of windows error in the other issue created. Currently, `mypy` on CI runs on Ubuntu so it should not raise any type errors.

PR #16273, opened 2025-08-29
[py] Allow `Service` as `command_executor` in `Remote.__init__`

### LINKED ISSUE #14464: [🚀 Feature]: `Service` shutting down via `__del__` is unintuitive (as `Remote` only consumes the service url)
### Feature and motivation

I recently faced a weird bug where returning from a function would cause the browser to close and Selenium to complain they couldn't reach the driver. Turned out, my `Service` was garbage collected (and quitting it via `__del__`) after I had used its `service_url`.
```py
from selenium.webdriver.chrome import service


def get_driver():
    webdriver_service = service.Service("operadriver")
    webdriver_service.start()

    options.add_experimental_option('w3c', True)

    driver = webdriver.Remote(webdriver_service.service_url, options=options)

    return driver


driver = get_driver()
driver.get("https://example.com")
```
While in hindsight it is expected behavior based on the code (I did not pass `webdriver_service`, I passed a string I got from it), it's not very weird (especially if you don't know the internals of how webdrivers work) to overlook the `.service_url` and assume that `Remote` would set `webdriver_service` as an instance attribute, keeping it from being garbage collected.
This behavior is in my opinion very unintuitive. A possible solution would be to only allow passing the whole `Service` into `Remote` instead of just the `service_url` (and setting it as an instance attribute), but this would be backwards-incompatible. Maybe add another method for passing the whole `Service` (or overload the original method), and encourage using that (for example by raising a warning when the `service_url` parameter is accessed) instead of the `service_url`? I don't have a good proposal, but I thought I would submit an issue anyway, since it cost me a lot of time to debug.

After writing this, it came to mind that using `Service` and `Remote` directly might not have been intended anyway. The readme of `webdriver-manager` told me to do it this way, probably because Opera was deprecated. I'll submit it anyway, but if this is the case, it's probably a lot less important.

### Usage example

Not applicable

GideonBear:
### 🔗 Related Issues

Closes #14464
Ping @cgoldberg

### 💥 What does this PR do?

Adds `Remote(service)` as an alternative for `Remote(service.service_url)`. It makes sure the `Service` doesn't get garbage collected.

Example:
```diff
 from selenium import webdriver
 from selenium.webdriver.chrome import service

 def get_driver():
-    global webdriver_service  # This is necessary to avoid dropping the `Service`
-
     webdriver_service = service.Service("operadriver")
     webdriver_service.start()
 
     options = webdriver.ChromeOptions()
     options.add_experimental_option('w3c', True)
 
-    driver = webdriver.Remote(webdriver_service.service_url, options=options)
+    driver = webdriver.Remote(webdriver_service, options=options)
 
     return driver
 
 driver = get_driver()
 driver.get("https://example.com")
 print("Success!")
```

### 🔧 Implementation Notes

Added a test that confirms the time at which the `Service` is dropped, and that any arguments are passed through correctly.

Ideally, all code currently using `Remote(service.service_url)` should change to `Remote(service)`. This could in theory be warned for by Selenium, but that would involve making `service_url` some kind of `str` wrapper.

### 🔄 Types of changes

___

### **PR Type**
Enhancement

___

### **Description**
- Add `Remote.from_service` class method for better service lifecycle management

- Prevent service garbage collection by maintaining reference in Remote instance

- Include comprehensive test for service lifecycle and argument passing

___

### Diagram Walkthrough

```mermaid
flowchart LR
  Service["Service Instance"] -- "from_service()" --> Remote["Remote WebDriver"]
  Remote -- "maintains reference" --> Service
  Service -- "prevents GC" --> Lifecycle["Service Lifecycle"]
```

<tr>
  <td>
    <details>
      <summary><strong>webdriver.py</strong><dd><code>Add from_service class method</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

<ul><li>Add <code>from_service</code> class method to Remote class<br> <li> Import <code>Self</code> type and <code>Service</code> class<br> <li> Store service reference to prevent garbage collection</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16273/files#diff-c87c150ed2a08f5293db58a06da8bf7f14c3a5582586c6270c1be661dfdd0985">+11/-0</a>&nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>from_server_tests.py</strong><dd><code>Add comprehensive from_service tests</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/test/unit/selenium/webdriver/remote/from_server_tests.py

<ul><li>Create test service class with deletion tracking<br> <li> Test Remote subclass for argument validation<br> <li> Verify service lifecycle management and argument passing</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16273/files#diff-1949674fea9018ae6ae7c488335e7b645064580294cbd7965e513b93ddefe3fd">+48/-0</a>&nbsp; &nbsp; </td>

</tr>

</details>

___


cgoldberg:
Thanks.. I re-opened the original issue.

hmm... what if we added an additional kwarg to `Service`, that would control if the service gets stopped when garbage collected?

The default would be `True`, but you could create a service like:

```python
service.Service("operadriver", auto_stop=False)
```

Then inside `__del__` on the `Service` class, we don't stop the service when it's created with this flag:

```python
def __del__(self) -> None:
    try:
        if self.auto_stop:
            self.stop()
    except Exception:
        pass
```

Would that solve the original problem?

GideonBear:
> Would that solve the original problem?

Then the `Service` would have to be manually stopped, right? That seems like a bad idea.
Besides that, my solution solves the issue implicitly, without the user having to be aware of stopping, quitting, etc. `.from_service` can be used *anywhere* even when the `Service` stays in scope. Your idea would not have solved my original problem, which came from a code snippet (copied from `webdriver_manager`'s README to be specific) being copied into a function. Using `.from_service` in that code snippet (which I will make sure to PR if this is merged :)) would have prevented this; but there would be no reason for the code snippet to use `auto_stop=False`.

GideonBear:
> That seems like a bad idea.

Small clarification; I don't mean this is a bad idea always, but having to use this to avoid GC is unnecessary IMO. I think my solution results in much cleaner user code, and I would still prefer to see a solution that (as I said) can be applied *everywhere*; but if that isn't possible that's the next best solution.

cgoldberg:
Yes, my suggestion would require you to explicitly stop the service.

Your solution adds a classmethod to the base class of all webdrivers that directly instantiates the `Service` class (an abstract base class). This wouldn't work for any webdriver except `Remote`.

There might be a nice implicit solution to this, but the current code in your PR isn't it.

GideonBear:
> Your solution adds a classmethod to the base class of all webdrivers that directly instantiates the Service class (an abstract base class). This wouldn't work for any webdriver except Remote.

It does not??
This is `selenium.webdriver.remote.webdriver.WebDriver`, which is re-exported as `selenium.webdriver.Remote`.

GideonBear:
It also does *not* instantiate a `Service`. It accepts one as an argument.

cgoldberg:
> > It does not??

It does.

`selenium.webdriver.remote.webdriver.Webdriver` is the base class for all webdrivers. Look at the code for any other webdriver class.

GideonBear:
You're right, apologies. I expected that to be `BaseWebdriver`... Sorry, the class hierarchy in Selenium is complexer than I thought.

But that means there is currently no isolated subclass for a `Remote`, because it is also used a superclass, right? Could a solution be to create a subclass of `Remote` that is to be used by users, where we replace `selenium.webdriver.remote.webdriver.WebDriver` with `NewRemote` as `selenium.webdriver.Remote`?

GideonBear:
Then `selenium.webdriver.remote.webdriver.WebDriver` can also be made an `ABC` and all users should instead use the `NewRemote`.

cgoldberg:
I mean, yea we could reorganize the class hierarchy to make this work. It's just not a trivial change and I don't want to break any existing code relying on the current API. I'm not sure it's worth it just to address a somewhat uncommon use case that we have never really had complaints about before.

GideonBear:
Yep I get it! I was under the impression that `Remote` was on the same level as `Chrome` or `Firefox` (both in class hierarchy and in uses) but that's not true. It's more of an internal class.

I was on the verge of closing this but one more idea came to me. What if `selenium.webdriver.remote.webdriver.WebDriver.__init__`'s `command_executor` argument just takes a `Union[str, RemoteConnection, Service]`? (previously `Union[str, RemoteConnection]`)
This:
1. Won't break any existing code; there is no overlap between the three types
2. Won't have any impact on subclasses, since they override `__init__` and don't get expect a `command_executor` at all
3. Results in clean user code (`Remote(service, options=options)`; previously `Remote(service.service_url, options=options)`)

GideonBear:
Implemented, the documentation might need to be improved. I don't know what the standards for types and such are.

GideonBear:
Rebased. Most previous comments can be ignored, as I pivoted in the implementation.

cgoldberg:
> After writing this, it came to mind that using Service and Remote directly might not have been intended anyway.

Normally when using `webdriver.Remote`, you are not running a local webdriver that requires a Service. `Service` classes are used for local webdrivers.

Since your case is kind of different, I would suggest just structuring your code so the `Service` class doesn't go out of scope and get garbage collected.

I am going to close this PR.

PR #16283, opened 2025-09-02
[py] Fix mypy errors

rpallavisharma:
### 🔗 Related Issues

Fixes the mypy error from #15697 - interaction file and related files
### 💥 What does this PR do?

the mypy listed out error for the wheel action python file and other related file. it is because of type mismatch.
changed interaction to expect InputDevice as source passed, instead of str data type. 
changed other files related to this. 

ran the format script
and the pytest . 
### 🔄 Types of changes

___

### **PR Type**
Bug fix

___

### **Description**
- Fixed mypy type error in Interaction and other related files.

- modified the interaction to expect source as InputDevice and not string. 

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["WheelActions.__init__"] --> B["super().__init__(source.type)"]
  A --> C["self._source = source"]
  D["pause() method"] --> E["self._source.create_pause()"]
  F["scroll() method"] --> G["self._source.create_scroll()"]
```

<tr>
  <td>
    <details>
      <summary><strong>wheel_actions.py</strong><dd><code>Fix mypy type error in WheelActions</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/selenium/webdriver/common/actions/wheel_actions.py

<ul><li>Modified constructor to pass <code>source.type</code> to parent class instead of <br><code>source</code> object<br> <li> Added <code>self._source</code> attribute to store the wheel input source<br> <li> Updated <code>pause()</code> and <code>scroll()</code> methods to use <code>self._source</code> instead of <br><code>self.source</code></ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16283/files#diff-4d0e09189e3c76803916008f8f60fb343728187528e4599f4b7a08cef34eafba">+4/-3</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___


navin772:
The changes look good, could you also fix mypy errors in `pointer_actions.py`, it's quite similar and relevant to add in this PR.

rpallavisharma:
I am under assumption we keep PRs small . I'm going by fixing mypy errors per file I see them which aren't fix yet. 
If it's okay this fix to be merged please kindly merge it for this file.

cgoldberg:
That's really confusing that the base `Interaction` class requires a string for `source`, but subclasses use a source object?

Instead of the changes in this PR, would it make sense to change the type annotation on the `Interaction.__init__` method to not require a string? This seems like we are making changes to satisfy the type checker for an incorrect type annotation.

rpallavisharma:
I have another question to this, before mypy errors were generated how did this code previously function? I noticed what you mentioned. Interaction is a parent class to more than one child classes. I am asking to understand how did this work before mypy errors? Sorry if this question sounds basic for python I'm trying to make sense of this.

cgoldberg:
> how did this work before mypy errors?

In the `WheelActions` class, `source` is an instance of `WheelInput`, which is a subclass of 
`InputDevice`. I think in the base `Interaction` class, the type annotation for `source` should be `InputDevice`, not string.

The type hint in the base class looks incorrect to me, but it works fine because type hints aren't enforced at runtime.. they just cause errors in mypy.

rpallavisharma:
Thanks for the explanation Corey. 

Made changes, kindly review 
@navin772 / @cgoldberg

rpallavisharma:
@cgoldberg , requesting review and merge of the PR. if any more changes aren't required.

py/selenium/webdriver/common/actions/key_actions.py
```diff
@@ -28,19 +28,7 @@ class KeyActions(Interaction):
     def __init__(self, source: KeyInput | PointerInput | WheelInput | None = None) -> None:
         if source is None:
             source = KeyInput(KEY)
-        self.input_source = source
```
cgoldberg, line 31: Can you leave this line in? We don't use it anywhere else in codebase, but it's not a private attribute and could possibly break some user code that might be accessing it.

cgoldberg (changes requested):
@pallavigitwork sorry for the delay. I am requesting one tiny change.. besides that it LGTM. I'll merge it as soon as you make the change from my comment.

cgoldberg (approved):
Thanks! 👍

rpallavisharma:
Thank you .

PR #16300, opened 2025-09-08
[py] Fix mypy errors in by file and exceptions file

rpallavisharma:
### 🔗 Related Issues
#15697 

### 💥 What does this PR do?

Fixes type annotations

### 🔄 Types of changes

___

### **PR Type**
Bug fix

___

### **Description**
- Fix mypy type errors


py/selenium/webdriver/remote/switch_to.py
```diff
@@ -87,7 +87,7 @@ def frame(self, frame_reference: Union[str, int, WebElement]) -> None:
                 try:
                     frame_reference = self._driver.find_element(By.NAME, frame_reference)
                 except NoSuchElementException as exc:
-                    raise NoSuchFrameException(frame_reference) from exc
+                    raise NoSuchFrameException(str(frame_reference)) from exc
```
cgoldberg: I don't think this is needed. This already gets implicitly converted to string.

py/selenium/webdriver/support/relative_locator.py
```diff
@@ -48,7 +48,7 @@ def with_tag_name(tag_name: str) -> "RelativeBy":
     warnings.warn("This method is deprecated and may be removed in future versions. Please use `locate_with` instead.")
     if not tag_name:
         raise WebDriverException("tag_name can not be null")
-    return RelativeBy({By.CSS_SELECTOR: tag_name})
+    return RelativeBy({"css selector": tag_name})
```
cgoldberg: Why do we need this? The original is the preferred way.

cgoldberg (changes requested):

py/selenium/webdriver/remote/switch_to.py
rpallavisharma: this resolved the mypy error .

py/selenium/webdriver/support/relative_locator.py
rpallavisharma: mypy error is being generated if we keep the original. 
Appears it's expecting strings like css selector , id, name... instead of By. 
That's what I understood.

rpallavisharma:
@cgoldberg kindly let me know if any more change is required?

py/selenium/webdriver/support/relative_locator.py
cgoldberg: So what is happening here is that `RelativeBy` is expecting the arg to be `dict[ByType, str]`. `"css selector"` is a valid Literal for `ByType`, but `By.CSS_SELECTOR` doesn't have a type annotation, so mypy thinks it is a string rather than a valid Literal.

I think a better fix is to leave this code the way it was (`return RelativeBy({By.CSS_SELECTOR: tag_name})`), and then add type annotations to the class attributes in the `By` class (in `py/selenium/webdriver/common/by.py`).

So the attributes should look like:

`CSS_SELECTOR: ByType = "css selector"`

note: you will also have to move the `ByType` declaration to the top of the file before the `By` class.

Let me know if this isn't clear and I can make a commit to your branch with what I mean.

Doing it this way should not only fix this mypy error, but also add missing type hints we should have.

cgoldberg (changes requested):

py/selenium/webdriver/support/relative_locator.py
rpallavisharma: Ok, I will try this based on what you explained, thank you.

py/selenium/webdriver/remote/switch_to.py
```diff
@@ -80,13 +80,14 @@ def frame(self, frame_reference: Union[str, int, WebElement]) -> None:
...
         if isinstance(frame_reference, str):
+            frame_name = frame_reference
             try:
                 frame_reference = self._driver.find_element(By.ID, frame_reference)
             except NoSuchElementException:
                 try:
                     frame_reference = self._driver.find_element(By.NAME, frame_reference)
                 except NoSuchElementException as exc:
-                    raise NoSuchFrameException(frame_reference) from exc
+                    raise NoSuchFrameException(frame_name) from exc
```
cgoldberg: All this does is suppress the mypy error by passing an argument with no type. I don't think we should be making changes like this just to keep the type checker happy. Instead, let's fix the root cause.

The reason mypy is complaining is that all of our custom exceptions accept a `msg` argument that currently has the following type annotation:

`msg: Optional[str] = None`

However, there's no need to restrict it to being a string. Any object that gets passed as the exception message is fine (its `__str__()` method will get called). I think we should change the type annotations on all exceptions to:

`msg: Optional[Any] = None`

This should be done for all exceptions in `py/selenium/common/exceptions.py`. After that, you should be able to revert the changes in this file and not get any type errors.

cgoldberg (changes requested):

py/selenium/webdriver/remote/switch_to.py
rpallavisharma: Ok, i will try and fix it this way.

rpallavisharma:
@cgoldberg kindly check, if any other changes required?

rpallavisharma:
Hi @cgoldberg is there any other change required, kindly let me know. thanks.

cgoldberg (approved):
LGTM now. Thanks for making those changes.

Issue #10132, opened 2021-12-10
[🐛 Bug]:  org.openqa.selenium.remote.RemoteWebDriver$ByteBuddy$RwtMl7Wf cannot be cast to org.openqa.selenium.devtools.HasDevTools


jantekb:
### What happened?

I am running latest selenium-docker image:

    sudo docker run -d -p 4444:4444 --shm-size="2g" -e SCREEN_WIDTH=1280 -e SCREEN_HEIGHT=720 selenium/standalone-chrome:4.1.0-20211209

In my Java application I have the following code snippet that throws ClassCastException:

```
ChromeOptions options = new ChromeOptions();
String driverUrl = "http://127.0.0.1:4444/wd/hub";
WebDriver driver = new RemoteWebDriver(new URL(driverUrl), options);
driver = new Augmenter().augment(driver);
DevTools devTools = ((HasDevTools) driver).getDevTools();
devTools.createSession();
```

As soon as the `HasDevTools` cast is attempted, I am getting:

`org.openqa.selenium.remote.RemoteWebDriver$ByteBuddy$RwtMl7Wf cannot be cast to org.openqa.selenium.devtools.HasDevTools: java.lang.ClassCastException: org.openqa.selenium.remote.RemoteWebDriver$ByteBuddy$RwtMl7Wf cannot be cast to org.openqa.selenium.devtools.HasDevTools
`

I have carefully reviewed all the similar issues, but none of them seems to have concluded on this particular bug.

My relevant grade dependency is:

    implementation 'org.seleniumhq.selenium:selenium-java:4.1.0'
    implementation 'org.seleniumhq.selenium:selenium-devtools-v94:4.1.0'

My JDK is:

OS Version: 4.14.252-195.483.amzn2.x86_64
Java Name: OpenJDK 64-Bit Server VM
Java Vendor: Oracle Corporation
Java Version: 9.0.4
Java VM Version: 9.0.4+11

### How can we reproduce the issue?

```shell
ChromeOptions options = new ChromeOptions();
String driverUrl = "http://127.0.0.1:4444/wd/hub";
WebDriver driver = new RemoteWebDriver(new URL(driverUrl), options);
driver = new Augmenter().augment(driver);
DevTools devTools = ((HasDevTools) driver).getDevTools();
devTools.createSession();
```

### Relevant log output

```shell
[...]

Issue #9803, opened 2021-09-08
RemoteWebDriver cannot be cast to HasDevTools


boris779:
## 🐛 Bug Report

It's not possible to use the DevTools with the RemoteDriver. I was told to use the HasDevTools Interface but it still not working when using a RemoteWebDriver (with a Grid)

```
java.lang.ClassCastException: class org.openqa.selenium.remote.RemoteWebDriver cannot be cast to class org.openqa.selenium.devtools.HasDevTools (org.openqa.selenium.remote.RemoteWebDriver and org.openqa.selenium.devtools.HasDevTools are in unnamed module of loader 'app')
```

Detailed steps to reproduce the behavior:

webDriver is RemoteWebDriver which is used in a SeleniumGrid Environment.

Java:
 `DevTools devTools = ((HasDevTools) webDriver).getDevTools();`

Kotlin:
 `val devTools: DevTools = (webDriver as HasDevTools).devTools`

Used Selenium Version: 4.0.0-rc-1 (also tested with beta-3 and beta-4)

## Environment

OS: Ubuntu 21
Browser: Chrome 
Browser version: > 87
Browser Driver version: >87
Language Bindings version: Java

pujagani:
Thank you for providing the details. I am sharing a working code snippet below that uses Augmenter. Essentially, Augmenter enhances the RemoteWebDriver to allow the use for DevTools apis. 

```
public class ChromeDevToolsAugmenter {

  public static void main(String[] args) throws MalformedURLException {

    ImmutableCapabilities capabilities = new ImmutableCapabilities("browserName", "chrome");
    WebDriver driver = new RemoteWebDriver(new URL("http://localhost:4444"), capabilities);

    driver = new Augmenter().augment(driver);

    driver.get("https://www.duckduckgo.com");
    DevTools devTools = ((HasDevTools) driver).getDevTools();
    devTools.createSession();

    devTools.send(Log.enable());
    devTools.send(Browser.setWindowBounds(new WindowID(1), new Bounds(
      Optional.of(20),
      Optional.of(20),
      Optional.of(20),
      Optional.of(20),
      Optional.of(WindowState.NORMAL))));

    driver.quit();
  }
}
```
Please try it out with Augmenter and provide feedback if the error still persists. Will be happy to dig into it.

boris779:
Hi!. 
No sorry, I tried some code examples but still get the same error.
[...]

PR #16338, opened 2025-09-20
make augmentation of HasBiDi/HasDevTools lazy-loaded

asolntsev:
### **User description**
When CDP url is not accessible, the command `new Augmenter().augment(remoteWebDriver)` fails immediately - even if I **don't want to use CDP or BiDi**.

### 🔗 Related Issues
This PR fixes the problem described in https://github.com/SeleniumHQ/selenium/issues/9803, https://github.com/SeleniumHQ/selenium/issues/10132, https://github.com/selenide/selenide/issues/2797, https://github.com/selenide/selenide/issues/3107 etc.

### 💥 What does this PR do?
This PR fixes `DevToolsProvider` and `BiDiProvider`, so that augmentation of `HasBiDi` and `HasDevTools` interfaces is now lazy-loaded. 

#### Context:
1. We run a remote Chrome driver in TestContainer / Selenoid / Moon / Selenium Grid
2. The CDP url is not accessible (because it contains the port which is accessible only inside the docker)
3. BUT I don't want to use CDP or BiDi.
4. I use Augmenter to cast my webdriver, say, to `JavascriptExecutor`:

```java
WebDriver wd = new Augmenter().augment(remoteWebDriver);  // FAILS
JavascriptExecutor js = (JavascriptExecutor) wd;
```

This code fails because of CDP connectivity issue during augmentation:

```java
org.openqa.selenium.remote.http.ConnectionFailedException: JdkWebSocket initial request execution error
Build info: version: '4.35.0', revision: '1c58e5028b'
System info: os.name: 'Mac OS X', os.arch: 'aarch64', os.version: '26.0', java.version: '17.0.14'
Driver info: driver.version: unknown
	at org.openqa.selenium.remote.http.jdk.JdkHttpClient.openSocket(JdkHttpClient.java:253)
	at org.openqa.selenium.devtools.Connection.<init>(Connection.java:89)
	at org.openqa.selenium.devtools.SeleniumCdpConnection.<init>(SeleniumCdpConnection.java:36)
	at org.openqa.selenium.devtools.SeleniumCdpConnection.lambda$create$2(SeleniumCdpConnection.java:103)
	at java.base/java.util.Optional.map(Optional.java:260)
	at org.openqa.selenium.devtools.SeleniumCdpConnection.create(SeleniumCdpConnection.java:103)
	at org.openqa.selenium.devtools.SeleniumCdpConnection.create(SeleniumCdpConnection.java:49)
	at org.openqa.selenium.devtools.DevToolsProvider.getImplementation(DevToolsProvider.java:50)
	at org.openqa.selenium.devtools.DevToolsProvider.getImplementation(DevToolsProvider.java:29)
	at org.openqa.selenium.remote.Augmenter.augment(Augmenter.java:202)
	at org.openqa.selenium.remote.Augmenter.augment(Augmenter.java:173)
```

### 🔄 Types of changes

___

### **PR Type**
Bug fix

___

### **Description**
- Make `HasBiDi` and `HasDevTools` augmentation lazy-loaded

- Fix connection failures when CDP/BiDi URLs are inaccessible

- Enable augmentation to succeed even without CDP/BiDi access

- Implement double-checked locking pattern for thread safety

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Augmenter.augment()"] --> B["DevToolsProvider/BiDiProvider"]
  B --> C["Lazy Implementation"]
  C --> D["Connection Created Only When Accessed"]
  D --> E["No Immediate Connection Failure"]
```

<tr>
  <td>
    <details>
      <summary><strong>BiDiProvider.java</strong><dd><code>Implement lazy BiDi connection initialization</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

java/src/org/openqa/selenium/bidi/BiDiProvider.java

<ul><li>Replace immediate BiDi connection creation with lazy-loaded <br>implementation<br> <li> Add double-checked locking pattern for thread-safe initialization<br> <li> Move connection logic inside <code>maybeGetBiDi()</code> method<br> <li> Make <code>getBiDiUrl()</code> method static</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16338/files#diff-aacb9c581a9640b47869934bf789a9e8334d50eb1439ad9336e52e87bcded434">+20/-7</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>DevToolsProvider.java</strong><dd><code>Implement lazy DevTools connection initialization</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

java/src/org/openqa/selenium/devtools/DevToolsProvider.java

<ul><li>Replace immediate DevTools connection creation with lazy-loaded <br>implementation<br> <li> Add double-checked locking pattern for thread-safe initialization<br> <li> Move CDP connection logic inside <code>maybeGetDevTools()</code> method<br> <li> Defer version detection and connection creation until first access</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16338/files#diff-687c3e2ecb48dffb059f8adf771223a0a7d1624646f4726427d27477f311d916">+18/-6</a>&nbsp; &nbsp; </td>

</tr>

</details>

___


joerg1985:
I have some concerns about this PR:
- At least for BDI the URI should only be inside the capabilities when requested by the user, so for BiDi this should not needed.
  In case a feature the client requested is not working, the augmentation should fail in my mind.
- I guess most people have a retry on startup error implemented, so the retry coukd have handled this in the past and now the exception happens later and might not handled by this any more.

asolntsev:
> * At least for BDI the URI should only be inside the capabilities when requested by the user, so for BiDi this should not needed. 

@joerg1985 
Maybe for BiDi, yes. But for `HasDevTools`, it's automatically enabled for all Chromium browsers (at least). Users cannot avoid it (even if they don't need to use DevTools). That's the root problem. 

>   In case a feature the client requested is not working, the augmentation should fail in my mind.

No, it should not. The goal of augmentation is NOT to check all the connectivity issues. The goal is to CAST `RemoteWebDriver` to the needed interface. Only to CAST, **not to call its methods**.

> * I guess most people have a retry on startup error implemented, so the retry coukd have handled this in the past and now the exception happens later and might not handled by this any more.

But the augmentation is NOT automatically called on startup.
The augmentation happens only user explicitly calls `new Augmenter().augment(remoteWebDriver)`. So there is no automated retry out-of-the-box anyway.

AND if I really want to check BiDi connectivity, I would explicitly call some BiDi method on startup. 

One argument for this PR:
1. All other AugmenterProviders ARE lazy. See `AddHasPermissions`, `AddHasCdp`, `AddHasExtensions`, `AddHasLogEvents`, `AddHasContext`, `AddHasNetworkConditions`  - they all have `getImplementation` of form:

```java
  public HasDebugger getImplementation(Capabilities capabilities, ExecuteMethod executeMethod) {
    return () -> { ... ALL THE WORK ONLY HERE ...}
  }

  // or

  public HasPermissions getImplementation(Capabilities capabilities, ExecuteMethod executeMethod) {
    return new HasPermissions() {
          ... ALL THE WORK ONLY HERE ...
    };
  }
```

2. Those people need to check connectivity on startup, they have an option: call any BiDi method. 
But those people who don't need BIdi or DevTools (like in my case), just cannot avoid it. They have **no way to avoid** this connectivity issue.

joerg1985:
@asolntsev i only wanted to point to this change in behaviour, i think having the same behaviour for all augmentations is a good idea. So i only have one comment left, all other areas rely on external synchronizations in case multiple threads are involved.
Is there a specific need to sychronize here or could this be removed?

asolntsev:
> Is there a specific need to sychronize here or could this be removed?

If I understand your question correctly, then yes, the synchronization is needed in this case.
This is what I mean:

```java
@Test void augmenterThreadSafety() {
  RemoteWebDriver driver = new RemoteWebDriver(gridUrl(), new ChromeOptions());
  WebDriver webDriver = new Augmenter().augment(driver);
  HasDevTools devTools = (HasDevTools) webDriver;

    for (int i = 0; i < 10; i++) {
      new Thread(() -> {
        devTools.getDevTools().getDomains(); // would create multiple instances of `DevTools` without synchronization
      }).start();
   }
}
```

joerg1985:
@asolntsev in my mind the client code has to take care about synchronization, in your example:

```@Test void augmenterThreadSafety() {
  RemoteWebDriver driver = new RemoteWebDriver(gridUrl(), new ChromeOptions());
  WebDriver webDriver = new Augmenter().augment(driver);
  HasDevTools devTools = (HasDevTools) webDriver;

    for (int i = 0; i < 10; i++) {
      new Thread(() -> {
	synchronized (devTools) {
        	devTools.getDevTools().getDomains(); 
	}
      }).start();
   }
}
```

Other areas of the selenium code do not expect the client code will use different threads.
So why expecting it inside the HasBiDi / HasDevTools implementation?

e.g. the code below will end in chaos without synchronization in the client code:

```@Test void threadSafety() {
  RemoteWebDriver driver = new RemoteWebDriver(gridUrl(), new ChromeOptions());

    for (int i = 0; i < 10; i++) {
      new Thread(() -> {
	driver.findElement(By.id("x")).click();
      }).start();
   }
}
```

asolntsev:
> in my mind the client code has to take care about synchronization

1. You cannot expect from all Selenium users that they can do synchronization properly. 
2. I sincerely don't understand why should we intentionally make Selenium less secure. This synchronization is very easy to do, it's cheap, it's safe. I see zero problems with it. 


> the code below will end in chaos without synchronization in the client code:

Yes, maybe this code wouldn't do anything reasonable, but it's _safe_. I mean, it doesn't cause connection leak etc.
While [the code above](https://github.com/SeleniumHQ/selenium/pull/16338#issuecomment-3319822105) _will cause connection leak_.

joerg1985:
@asolntsev lets see what others think about this

nvborisenko:
Easy, should be thread-safe.

asolntsev:
> Easy, should be thread-safe.

@nvborisenko Great! Then can we already merge this PR?

nvborisenko:
Sorry, I am not from Java world.

How I see the situation how it should be:
```
var driver = new AnyDriver(); // it is still NOT initializing DevTools/BiDi connection

var devtools = driver.getDevTools(); // only here new DevTools object is allocated, which initializes underlying network connection

var devtools2 = driver.getDevTools(); // it should return previously initialized object

var devtoolsN = Parallel.10Times(() => driver.getDevTools()); // it should be thread-safe and initialize one single object
```

If this is what you propose, then I approve PR verbally :)

asolntsev:
> If this is what you propose, then I approve PR verbally :)

Yes, this is exactly my proposal. :)

nvborisenko (approved):
Thanks, this approach also avoids resource leakage. Sorry, I approved approach, but not code.. let's wait experts.

titusfortner:
I don't have time to dig into the details here, but the current mechanism for avoiding problems with inaccessible devtools is to use `"se:cdpEnabled": false` in capabilities so HasDevTools won't be added. HasBiDi won't be added unless websocketUrl is true, which should be false by default.

I'd say more magic is bad, but Augmenter is already a lot of magic, so... :)

asolntsev:
> avoid inaccessible devtools by `"se:cdpEnabled": false` in capability.

Good point. Indeed, adding `se:cdpEnabled=false` would fix the initial problem. :)

But still, I believe this PR is not about magic at all. It just makes two tiny object lazy-loaded. Absolutely reasonable. 
It doesn't even deserve all this long discussion above. :(

P.S. Ironically, capability named "cdp" ("se:cdpEnabled") does NOT affect CDP.

java/src/org/openqa/selenium/bidi/BiDiProvider.java
```diff
@@ -44,18 +44,32 @@ public Class<HasBiDi> getDescribedInterface() {
...
+            }
+          }
+        }
+        return biDi;
+      }
+    };
   }
 
-  private Optional<URI> getBiDiUrl(Capabilities caps) {
+  private static Optional<URI> getBiDiUrl(Capabilities caps) {
```
diemol: Why is this static? Am I missing something?

java/src/org/openqa/selenium/bidi/BiDiProvider.java
```diff
@@ -44,18 +44,32 @@ public Class<HasBiDi> getDescribedInterface() {
 
   @Override
   public HasBiDi getImplementation(Capabilities caps, ExecuteMethod executeMethod) {
+    return new HasBiDi() {
```
diemol, line 60: I am not 100% sure about the BiDi part because the user is expressing the intention to use BiDi by setting `webSocketUrl: true`. I think it makes sense for CDP because we are doing that implicitly.

diemol (changes requested):

java/src/org/openqa/selenium/bidi/BiDiProvider.java
asolntsev: Yes, that's right: users can enable or disable BiDi.
But it doesn't matter. Anyway, **it's not needed to establish BiDi connection while no BiDi methods are called**.

java/src/org/openqa/selenium/bidi/BiDiProvider.java
asolntsev: It doesn't really matter. It's a private method, and it doesn't depend on any fields. 
It can be either static or non-static.

java/src/org/openqa/selenium/bidi/BiDiProvider.java
diemol: I get that, but if the user has set the capability, then it means they want to use BiDi methods. I would prefer to avoid this change. The CDP one seems fine.

java/src/org/openqa/selenium/bidi/BiDiProvider.java
diemol: Then let's remove the `static` keyword.

java/src/org/openqa/selenium/bidi/BiDiProvider.java
asolntsev: Wait, why? If users want to use BiDi methods, they will use BiDi methods. 
I still don't see any reasons to perform side effects during augmentation. 
* Augmenter should only augment (create instance of interfaces). 
* Augmenter should not perform any other actions (establish CDP connection, establish BiDi connection etc.)

java/src/org/openqa/selenium/bidi/BiDiProvider.java
asolntsev: no problems, I've removed the `static` keyword.

java/src/org/openqa/selenium/bidi/BiDiProvider.java
nvborisenko: > if the user has set the capability, then it means they want to use BiDi methods

This is big mistake. If I set some capability, it means I set capability. I don't know who is BiDi.

asolntsev: @nvborisenko Diego meant a very specific capability "se:cdpEnabled" which allows to disable the BiDi connection.

asolntsev: > I am not 100% sure about the BiDi part because the user is expressing the intention to use BiDi by setting `webSocketUrl: true`. I think it makes sense for CDP because we are doing that implicitly.

@diemol What can I do to get this PR merged?
At least you agree that it makes sense for CDP, right? 
Do you want me to revert the BiDiProvider part and keep only DevToolsProvider change? **If I do so, will you merge the PR**?

(just be worried that then BiDiProvider and DevToolsProvider will behave differently: one would be lazy-loaded and the other eager. One has side effect, and all the other providers don't.)

asolntsev:
Please.
  Code.
    Review.

Please?
  Code??
    Review??

Please!
  Code!!
    Review!!

@mkurz @diemol @iampopovich @VietND96 @cgoldberg @joerg1985 @navin772 @vicky-iv @giulong @pujagani

titusfortner (changes requested):
If you don't have socket support...
By default `webSocketUrl` is false, so BiDi is not augmented
CDP is on by default, so you can toggle it off with `"se:cdpEnabled": false`

Please explain why a user should not have to set the capabilities to match the behavior they get?

As for the code itself, I'm a little concerned about doing synchronization things inside this method, but not in other methods. I'd have to think through the implications. If Diego or Joerg say it is fine, then I'll go with their expertise, but in general I'd say not to add the extra code if it isn't needed.

asolntsev:
> If you don't have socket support... By default `webSocketUrl` is false, so BiDi is not augmented CDP is on by default, so you can toggle it off with `"se:cdpEnabled": false`
> 
> Please explain why a user should not have to set the capabilities to match the behavior they get?

Look, we care about simplicity for our users, right?
We want Selenium to be simple to use as possible, right?
For user, the most simple is just to call `new Augmenter().augment(webdriver)` **without setting any capabilities**.
Users who doesn't need CDP should be able to just start browser without setting any CDP capabilities. Right?

**Honestly, I don't understand why are you trying to make your users more work than needed.**

> I'm a little concerned about doing synchronization things inside this method.

I don't see problems here. Synchronized are exactly those methods which establish a new connection. Exactly as it should be.

nvborisenko:
.NET binding never establishes new network connection if it is not required.

java/src/org/openqa/selenium/devtools/DevToolsProvider.java
```diff
@@ -42,14 +43,35 @@ public Class<HasDevTools> getDescribedInterface() {
...
+        if (devTools == null) {
+          synchronized (lock) {
+            if (devTools == null) {
+              Object cdpVersion = caps.getCapability("se:cdpVersion");
+              String version =
+                cdpVersion instanceof String ? (String) cdpVersion : caps.getBrowserVersion();
 
-    return () -> devTools;
+              CdpInfo info = new CdpVersionFinder().match(version).orElseGet(NoOpCdpInfo::new);
+              this.devTools =
```
titusfortner: catch & close connection if this has runtime issue?

java/src/org/openqa/selenium/devtools/DevToolsProvider.java
```diff
@@ -42,14 +43,35 @@ public Class<HasDevTools> getDescribedInterface() {
...
+                SeleniumCdpConnection.create(caps)
```
titusfortner: Is it better for this to be the only thing inside the synchronized block? Caps won't change, so no need to lock when checking them

java/src/org/openqa/selenium/bidi/BiDiProvider.java
```diff
@@ -45,22 +46,28 @@ public Class<HasBiDi> getDescribedInterface() {
   @Override
   public HasBiDi getImplementation(Capabilities caps, ExecuteMethod executeMethod) {
     return new HasBiDi() {
-      private volatile Optional<BiDi> biDi;
+      private volatile BiDi biDi;
+      private final Object lock = new Object();
 
       @Override
       public Optional<BiDi> maybeGetBiDi() {
```
titusfortner, line 62: This isn't backwards compatible; users could be doing `(maybeGetBiDi().isPresent()) { … }` which would now break

java/src/org/openqa/selenium/devtools/DevToolsProvider.java
```diff
@@ -43,21 +44,28 @@ public Class<HasDevTools> getDescribedInterface() {
   @Override
   public HasDevTools getImplementation(Capabilities caps, ExecuteMethod executeMethod) {
     return new HasDevTools() {
-      private volatile Optional<DevTools> devTools;
+      private volatile DevTools devTools;
+      private final Object lock = new Object();
 
       @Override
       public Optional<DevTools> maybeGetDevTools() {
```
titusfortner, line 60: This isn't backwards compatible; users could be doing `(maybeGetDevTools().isPresent()) { … }` which would now break

java/src/org/openqa/selenium/bidi/BiDiProvider.java
```diff
@@ -45,22 +46,28 @@ public Class<HasBiDi> getDescribedInterface() {
...
-                  getBiDiUrl(caps).orElseThrow(() -> new BiDiException("BiDi not supported"));
+                getBiDiUrl(caps).orElseThrow(() -> new BiDiException("BiDi not supported"));
 
               HttpClient.Factory clientFactory = HttpClient.Factory.createDefault();
               ClientConfig wsConfig = ClientConfig.defaultConfig().baseUri(wsUri);
               HttpClient wsClient = clientFactory.createClient(wsConfig);
               Connection connection = new Connection(wsClient, wsUri.toString());
 
-              biDi = Optional.of(new BiDi(connection));
+              biDi = new BiDi(connection);
```
titusfortner: we should make sure we close the connection if there's a problem in the constructor

java/src/org/openqa/selenium/devtools/DevToolsProvider.java
```diff
@@ -43,21 +44,28 @@ public Class<HasDevTools> getDescribedInterface() {
...
-                  cdpVersion instanceof String ? (String) cdpVersion : caps.getBrowserVersion();
+                cdpVersion instanceof String ? (String) cdpVersion : caps.getBrowserVersion();
 
               CdpInfo info = new CdpVersionFinder().match(version).orElseGet(NoOpCdpInfo::new);
               this.devTools =
-                  SeleniumCdpConnection.create(caps)
-                      .map(conn -> new DevTools(info::getDomains, conn));
+                SeleniumCdpConnection.create(caps)
+                  .map(conn -> new DevTools(info::getDomains, conn))
+                  .orElseThrow(() -> new DevToolsException("Unable to create DevTools connection"));;
```
titusfortner: nit: 2 semicolons; our linter should have caught this?

titusfortner: Also this throws, but doesn't close the connection if it was created and then failure in the map

titusfortner (changes requested):
I think if we decide to lazy load DevTools we should do the same for BiDi. Eventually that's going to be default. 

I took at deeper look at code and noted some things.
Maybe we can log more info when getDevTools / getBiDi fail
Let's at least log that we're lazy loading since it's a change.
Can we add a test for these?

I'm still deferring to @diemol if he says no. 😄

java/src/org/openqa/selenium/bidi/BiDiProvider.java
asolntsev: @titusfortner This would not fail, this would just return `false`.
And this is totally reasonable.

Think what happened before: expression `maybeGetBiDi().isPresent()` always opened a BiDi connection, and **always returned true**. Wasn't it nonce? ;)

P.S. In reality, this method is always used only when closing the webdriver:
```java
  if (maybeGetBiDi().isPresent()) bidi.close();
```

java/src/org/openqa/selenium/bidi/BiDiProvider.java
asolntsev: @titusfortner Do you mean exception in this constructor?
```java
new BiDi(connection)
```

Then no problems, this constructor cannot throw exceptions. It's a trivial one-liner. It only throws NPE if there is no connection (thus, nothing to close).

java/src/org/openqa/selenium/devtools/DevToolsProvider.java
asolntsev: same as before: `isPresent` would not fail, it would just return `false` (if no connection is open). Used in `WebDriver.close()` to close the connection if it had been opened.

java/src/org/openqa/selenium/devtools/DevToolsProvider.java
asolntsev: I've analyzed this code, and came to a conclusion that the only case when this code can throw an exception is when there is no CDP connection. This, there is nothing to close.

Once CDP connection gets established, the following code is just primitive constructor which creates a bunch of simple objets like `new v139Events()`, `new v139Javascript()` etc.

P.S. Anyway, the current code in `trunk` behaves the same.

java/src/org/openqa/selenium/devtools/DevToolsProvider.java
asolntsev: If we move other lines outside of "synchronized", the only thing that changes is:

if multiple parallel threads call `getDevTools()` at the same moment, they will execute these lines multiple times. I don't see any benefits from it.

java/src/org/openqa/selenium/devtools/DevToolsProvider.java
asolntsev: * 2 semicolons: ops, fixed.   (about linter: maybe it doesn't fail because it's not a critical problem?)
* this throws: again, this code can throw an exception only when CDP connection could not be opened. This, it's notting to close.

asolntsev:
@titusfortner I've extracted the "synchronization" code to a separate class `Lazy`.
I hope it make `DevToolsProvider` and `BiDiProvider` much simpler, and not-so-scary.
Now they use `lazy` function (similar to Kotlin [lazy idiom](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/lazy.html)):

```java
  @Override
  public HasBiDi getImplementation(Capabilities caps, ExecuteMethod executeMethod) {
    return new HasBiDi() {
      private final Lazy<BiDi> biDi = lazy(() -> establishBiDiConnection(caps));

      @Override
      public Optional<BiDi> maybeGetBiDi() {
        return biDi.getIfInitialized();
      }

      @Override
      public BiDi getBiDi() {
        return biDi.get();
      }
    };
  }
```

java/src/org/openqa/selenium/concurrent/Lazy.java
```diff
@@ -0,0 +1,37 @@
...
+public class Lazy<T> {
+
+  @Nullable private volatile T value;
+  private final Supplier<T> supplier;
+
+  private Lazy(Supplier<T> supplier) {
+    this.supplier = supplier;
+  }
+
+  @Nullable
```
titusfortner: Do we need this now that it is an optional object?

java/src/org/openqa/selenium/concurrent/Lazy.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.function.Supplier;
+import org.jspecify.annotations.Nullable;
+
+public class Lazy<T> {
+
+  @Nullable private volatile T value;
+  private final Supplier<T> supplier;
+
+  private Lazy(Supplier<T> supplier) {
+    this.supplier = supplier;
```
titusfortner, line 32: Should we require this to be non-null at the beginning?

asolntsev: Do you mean `supplier`? Yes, `supplier` should be non-nullable.
Thanks to annotation `@NullMarked` in `package-info.java`, all parameters, vars and return values are by default non-nullable in this package.

titusfortner:
I know it is just an abstraction, but the new code feels so much better to me. Thanks also for the tests.
I still think we should add logging that we're doing it lazy now instead of on startup (maybe in the `getImplementation()` methods?). We can also add a warning entry for URISyntaxException, which Chrome & Firefox implementations do.

To clarify for @diemol the use case that made me think we should do this now is that a user might need to run their code on a remote grid that doesn't support sockets, the grid config would need to change not just the client code. So the user can't just toggle this from their capabilities and ensure it works as I previously suggested.
Also, if this is the right approach for CDP, I think it should also be used for BiDi, because at some point we're going to be turning on BiDi by default, and this would make that transition easier (refer to previous conversations with Simon).

java/src/org/openqa/selenium/concurrent/Lazy.java
asolntsev: @titusfortner Oops, you are correct: `@Nullable` is not needed because the returned object is `Optional`, not null. :) Fixed.

asolntsev:
> I still think we should add logging that we're doing it lazy now instead of on startup

Sure, we can add the logging. 
Just to clarify: it was never done on "startup", it was done when user called `new Augmenter().augment(webdriver)`. 

So you are suggesting that during `new Augmenter().augment(webdriver)` call, Selenium will write done two new logs:

1. `INFO  [BiDiProvider] Add BiDi implementation - but BiDi connection will be establish only on a first usage`
2. `INFO  [DevToolsProvider] Add DevTools implementation - but DevTools connection will be establish only on a first usage`

Right? Do the texts sound good? Should it be "INFO" or "WARNING" or "DEBUG"?

P.S. If you really assume some users want to check-up all connections during augmentation, we could add an optional parameter to `augment` method:

```java
Augmenter augmenter = new Augmenter();
WebDriver wd1 = augmenter.augment(webdriver); // lazy-initialized
WebDriver wd1 = augmenter.augment(webdriver, LAZY); // lazy-initialized
WebDriver wd2 = augmenter.augment(webdriver, EAGER); // checks all connections right now

// where the second parameter is:
enum InitializationMode {LAZY, EAGER}
```

titusfortner:
I was thinking of when you use webdriver builder, it automatically augments and that's on startup.

The get implementation method is only called during augmentation and we already know it matched on is applicable. Info is good for this now since this is a behavior change. I only do warnings if I think it's something the user needs to change. 

I would say something like "driver augmented with bidi/devtools interface, but the connection is not verified until first use." 

If someone really needs to check right away, they can cast and call the get bidi command, right? No need to make the implementation more complicated with an optional parameter.

asolntsev:
@titusfortner Ok, added the log:

```java
LOG.log(INFO, "WebDriver augmented with DevTools interface; connection will not be verified until first use.");
LOG.log(INFO, "WebDriver augmented with BiDi interface; connection will not be verified until first use.");
```

java/src/org/openqa/selenium/bidi/BiDiProvider.java
diemol: Let's go with it and see if someone complains.

diemol (approved):
Please run ` ./scripts/format.sh` so we can get this merged. Thanks for the effort to explain your approach.

asolntsev:
> Please run ` ./scripts/format.sh` so we can get this merged

Strange... I am sure I already executed `./scripts/format.sh` before...

UPD Now I see the problem: this script always fails on my machine:
```
INFO: Running command line: bazel-bin/buildifier.bash
./.bazelbsp/aspects/rules/android/android_info.bzl:3:30: unexpected input character '$'
./.bazelbsp/aspects/rules/android/android_info.bzl # reformat
```

titusfortner (approved):
Looks good, thank you for the updates and your patience 😄

PR #16386, opened 2025-10-05
[dotnet] [bidi] AOT safe enums serialization

nvborisenko:
### **User description**
Contributes to https://github.com/SeleniumHQ/selenium/issues/14480

### 💥 What does this PR do?
Generated json serialization now treats enums? Meaning AOT safe?

### 🔄 Types of changes

___

### **PR Type**
Bug fix

___

### **Description**
- Enable AOT-safe enum serialization in BiDi JSON context

- Add UseStringEnumConverter option for string-based enum handling

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["BiDi JSON Context"] --> B["Add UseStringEnumConverter"] --> C["AOT Safe Enums"]
```

<tr>
  <td>
    <details>
      <summary><strong>BiDiJsonSerializerContext.cs</strong><dd><code>Add AOT-safe enum serialization configuration</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/BiDi/Communication/Json/BiDiJsonSerializerContext.cs

<ul><li>Added JsonSourceGenerationOptions attribute with <br>UseStringEnumConverter = true<br> <li> Enables AOT-compatible enum serialization for BiDi communication</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16386/files#diff-940ec174f427c0fe37ed4a09cd37840b888f4a379f42a49d8877b75460cd4048">+2/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___


nvborisenko:
@RenderMichael do you know? (hard to debug)

nvborisenko:
Oh, it is still not correct. We have added global converter:

```csharp
new JsonStringEnumConverter(JsonNamingPolicy.CamelCase)
```

As soon as I remove this converter, then enums are converted as PascalCase. Will do CamelCase custom converter.

nvborisenko:
I added `CamelCaseEnumConverter<TEnum>` - looks good. Now I need to add attribute for all known enums, will do. But the question is: will it be AOT safe? Or I still need to add all known enums to `SerializerContext`?

RenderMichael:
> will it be AOT safe?

Yes, this is AOT-safe and the preferred way of enabling this behavior.

>  I still need to add all known enums to `SerializerContext`?

These enums are already known to the serializer context because they are properties of types which are known to the serializer context.

dotnet/src/webdriver/BiDi/Communication/Broker.cs
```diff
@@ -83,8 +83,6 @@ internal Broker(BiDi bidi, Uri url)
                 new InternalIdConverter(_bidi),
                 new PreloadScriptConverter(_bidi),
                 new RealmConverter(_bidi),
-                new RealmTypeConverter(),
```
RenderMichael, line 86: These two types can also be removed ✅

nvborisenko: I think no, strange string names:

```csharp
"dedicated-worker" => RealmType.DedicatedWorker
```

where `dedicated-worker` is not CamelCase.

RenderMichael: We can use `KebabCase` here instead, since all values follow this pattern. Will need another attribute, `KebabCaseEnumConverter` alongside `CamelCaseEnumConverter` 

https://github.com/SeleniumHQ/selenium/blob/64778708e346783be3dc108a9cb4da12250f3eb4/dotnet/src/webdriver/BiDi/Communication/Json/Converters/RealmTypeConverter.cs#L33-L44

nvborisenko: Agree, nice proposal. Will do.

RenderMichael:
> Now I need to add attribute for all known enums, will do

I much prefer this approach for all converter attributes, because we are safer to use multiple serializer contexts, and let users serialize the types with their own serializer. What do you think @nvborisenko ? Not in this PR, either way, but want to know your opinion.

nvborisenko:
Absolutely, we already discussed that DTO objects might be reused across modules, even across extension modules (like https://github.com/SeleniumHQ/selenium/issues/15329). Seems we are on the way to apply converters via attributes. Not in this PR, but yes: design approach is approved. Thanks for your help.

RenderMichael (approved):
Nice update!

PR #16397, opened 2025-10-07
[dotnet] [bidi] Provide type info immediately when serializing

nvborisenko:
### **User description**
Big performance improvement. This is a continuation of https://github.com/SeleniumHQ/selenium/pull/16392 - so merge that before it.

### 💥 What does this PR do?
Each module knows what it sends and what it expects to get, just be more precise.

___

### **PR Type**
Enhancement

___

### **Description**
- Refactor BiDi module system for better performance

- Replace generic type parameters with explicit type info

- Centralize JSON serialization configuration

- Implement module factory pattern with shared context

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["BiDi Class"] --> B["Module Factory"]
  B --> C["Shared JSON Context"]
  C --> D["Type-Safe Serialization"]
  E["Broker"] --> F["Direct Type Info"]
  F --> G["Performance Improvement"]
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>13 files</summary><table>
<tr>
  <td><strong>BiDi.cs</strong><dd><code>Refactor module management with factory pattern</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16397/files#diff-3f77d71f94e2a47ea90b74f6d37b5a5b94dc83b2f558b30ab070588b9028c2f6">+46/-129</a></td>

</tr>

<tr>
  <td><strong>BrowserModule.cs</strong><dd><code>Update command execution with type info</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16397/files#diff-8e4bcc00ef63cf737c6e26397ac40f1e18d960e4ff47589fee1bae1b9fe8cdc7">+11/-10</a>&nbsp; </td>

</tr>

<tr>
[...]

PR #16392, opened 2025-10-06
[dotnet] [bidi] Modules as extensions

nvborisenko:
### **User description**
Contributes to https://github.com/SeleniumHQ/selenium/issues/15329

### 🔄 Types of changes

- Cleanup (formatting, renaming)
- Bug fix (backwards compatible)
- New feature (non-breaking change which adds functionality *and tests!*)
- Breaking change (fix or feature that would cause existing functionality to change)

___

### **PR Type**
Enhancement

___

### **Description**
- Refactor BiDi modules to use extension pattern

- Centralize JSON serialization configuration in BiDi class

- Replace individual module properties with generic AsModule method

- Update all module constructors to parameterless design

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["BiDi Class"] --> B["AsModule<T>() Method"]
  B --> C["ConcurrentDictionary<Type, Module>"]
  C --> D["Module.Create<T>()"]
  D --> E["Individual Modules"]
  A --> F["Centralized JsonSerializerOptions"]
  F --> G["BiDiJsonSerializerContext"]
  G --> E
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>13 files</summary><table>
<tr>
  <td><strong>BiDi.cs</strong><dd><code>Centralize JSON context and implement module extensions</code>&nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16392/files#diff-3f77d71f94e2a47ea90b74f6d37b5a5b94dc83b2f558b30ab070588b9028c2f6">+46/-129</a></td>

</tr>

<tr>
  <td><strong>Module.cs</strong><dd><code>Add factory method and JSON context properties</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16392/files#diff-b45f63b7ea084aed81f19bba490ac80626f908cecea9448d35db3d4fbc663ddc">+22/-2</a>&nbsp; &nbsp; </td>

</tr>

<tr>
[...]

PR #16414, opened 2025-10-11
[dotnet] Implement third-party Permissions module

### LINKED ISSUE #15329: [🚀 Feature]: [dotnet] Provide a possibility to implement BiDi Permissions module

RenderMichael:
### 🔗 Related Issues
Fixes https://github.com/SeleniumHQ/selenium/issues/15329

New API surface:

```C#
var permissions = bidi.AsPermissions();
await permissions.SetPermissionAsync("geolocation", PermissionState.Denied, newPage, userContext.UserContext);
```

___

### **PR Type**
Enhancement

___

### **Description**
- Enable external BiDi module creation and extension

- Refactor module architecture with InternalModule base class

- Expose Broker and JsonOptions for external access

- Simplify module creation with public static method

___

### Diagram Walkthrough

```mermaid
flowchart LR
  BiDi["BiDi Class"] -- "exposes" --> Broker["Internal Broker"]
  BiDi -- "exposes" --> JsonOptions["JSON Options"]
  Module["Base Module"] -- "extends" --> InternalModule["InternalModule"]
  InternalModule -- "used by" --> AllModules["All Built-in Modules"]
  Module -- "public creation" --> ExternalModules["External Modules"]
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>13 files</summary><table>
<tr>
  <td><strong>BiDi.cs</strong><dd><code>Expose internal properties and refactor module creation</code>&nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16414/files#diff-3f77d71f94e2a47ea90b74f6d37b5a5b94dc83b2f558b30ab070588b9028c2f6">+26/-21</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>InternalModule.cs</strong><dd><code>Add new base class for internal modules</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16414/files#diff-41fbd3302ee5fd4b867a3bb57f38e3fafd2e63be6a71d7a2c307fd37d7fdab7d">+34/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>Module.cs</strong><dd><code>Refactor to support external module creation</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16414/files#diff-b45f63b7ea084aed81f19bba490ac80626f908cecea9448d35db3d4fbc663ddc">+6/-7</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>BrowserModule.cs</strong><dd><code>Change inheritance to InternalModule</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16414/files#diff-8e4bcc00ef63cf737c6e26397ac40f1e18d960e4ff47589fee1bae1b9fe8cdc7">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>BrowsingContextModule.cs</strong><dd><code>Change inheritance to InternalModule</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16414/files#diff-1f04192c50706a21f9dfdc1b79916855ccf2aba6c2c895811f158eeac915680c">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>EmulationModule.cs</strong><dd><code>Change inheritance to InternalModule</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16414/files#diff-326495cc635fdb9686d337228f031a251b2aaed55960d7096b1345e0f1c267ff">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>InputModule.cs</strong><dd><code>Change inheritance to InternalModule</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16414/files#diff-f98f384929541b30f9b87fb08f96ead9181d6b14a2353f5485ba985ed7cbd672">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>LogModule.cs</strong><dd><code>Change inheritance to InternalModule</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16414/files#diff-fdb6bcf97d10b7240dada7ced38218b43fcb0b8ba716b273dd9c60e17c286e3f">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>NetworkModule.cs</strong><dd><code>Change inheritance to InternalModule</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16414/files#diff-ddd0c9f9565ffffc6a3ce489431b76319cc38c727006605f7dc6345b2c0f7c81">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>ScriptModule.cs</strong><dd><code>Change inheritance to InternalModule</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16414/files#diff-da8cf630954e0732aa2b7f257aca8b49117eeceae343e373e2ba5914e639b6d2">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>SessionModule.cs</strong><dd><code>Change inheritance to InternalModule</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16414/files#diff-31b1c3c5b298bcc68ac630c0af2fe9ecd548a737be8b5fa14574317f6eff3ff3">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>StorageModule.cs</strong><dd><code>Change inheritance to InternalModule</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16414/files#diff-c2832a3991518e29f9a8f706cbe7e7c1d5ec4a007fbb57da280bb1be9fc7b2c2">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>WebExtensionModule.cs</strong><dd><code>Change inheritance to InternalModule</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16414/files#diff-0eaa7ea3946fcd852a8db4e125d1bd98c674f8ce8489cc0279f452dca8f30a25">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___


dotnet/src/webdriver/BiDi/Browser/BrowserModule.cs
```diff
@@ -22,7 +22,7 @@
 
 namespace OpenQA.Selenium.BiDi.Browser;
 
-public sealed class BrowserModule : Module
+public sealed class BrowserModule : CoreModule
```
nvborisenko: But why `CoreModule`?

dotnet/src/webdriver/BiDi/Module.cs
```diff
@@ -18,29 +18,28 @@
...
+using System.Text.Json.Serialization;
 
 namespace OpenQA.Selenium.BiDi;
 
 public abstract class Module
 {
     protected Broker Broker { get; private set; }
 
-    internal BiDiJsonSerializerContext JsonContext { get; private set; }
+    internal JsonSerializerContext JsonContext { get; private set; }
```
nvborisenko, line 29: I am not ready to generalize it in this PR. Still not clear how to move on.

dotnet/src/webdriver/BiDi/Browser/BrowserModule.cs
nvborisenko: Let me remind: we made `Module` class `ctor` parameterless, especially for external modules.

dotnet/src/webdriver/BiDi/Module.cs
RenderMichael: I have an example in this PR’s description, of how to move forward with these changes.

nvborisenko: I though making `Module.Create(...)` to public would be enough, please give me time to review.

dotnet/src/webdriver/BiDi/Browser/BrowserModule.cs
RenderMichael: We have 2 requirements:

1) Each module should maintain its own serialization
2) We want the first-party modules to share the same serialization

For that reason, they need to have some shared logic.

dotnet/src/webdriver/BiDi/Module.cs
nvborisenko: Observed it is theoretical changes. Let be practicable.

RenderMichael: It is, except for the shared JSON context introduced in https://github.com/SeleniumHQ/selenium/pull/16402

It is not theoretical changes.

nvborisenko (changes requested):
No value,. In any case thank you.

dotnet/src/webdriver/BiDi/Module.cs
nvborisenko: I came to the following:

1. BiDi has new method `AsModule<T>()`
2. PermissionsModule creates its own context based on options (and keep it as private field)

Then changes looks simple.

nvborisenko: I feel introducing new `AsModule<T>()` method will be good for us:
- we may introduce cached modules
- it negotiates use of internal properties

nvborisenko: All right, `AsModule<T>` is required. Built-in modules are cached, third-party modules also want to be cached per `bidi` instance. Let's do it (here or in separate PR).

dotnet/src/webdriver/BiDi/Browser/BrowserModule.cs
```diff
@@ -18,12 +18,17 @@
...
+using OpenQA.Selenium.BiDi.Communication.Json;
+using System.Text.Json;
+using System.Text.Json.Serialization;
 using System.Threading.Tasks;
 
 namespace OpenQA.Selenium.BiDi.Browser;
 
 public sealed class BrowserModule : Module
 {
+    internal new BiDiJsonSerializerContext JsonContext => (BiDiJsonSerializerContext)base.JsonContext;
```
nvborisenko, line 29: @RenderMichael this piece is strange, seems there is no reason to keep JsonContext in base class. If you have the same opinion, I can remake it to private field.

dotnet/src/webdriver/BiDi/Extensions/Permissions/PermissionDescriptor.cs
```diff
@@ -0,0 +1,22 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+// </copyright>
+
+namespace OpenQA.Selenium.BiDi.Extensions.Permissions;
```
nvborisenko: I propose to simplify namespace to be "regular" module, but with `bidi.AsPermissions()` extention method to emphasize this is still extension.

nvborisenko: And given that we will have public `bidi.AsModule<PermissionsModule>()`, I think it would look really simple.

dotnet/src/webdriver/BiDi/Module.cs
```diff
@@ -18,29 +18,28 @@
...
 
-    protected virtual void Initialize(JsonSerializerOptions options) { }
+    protected abstract JsonSerializerContext Initialize(JsonSerializerOptions options);
```
nvborisenko: We should choose between `void Initialize(...)` and `JsonSerializerContext ConfigureJsonContext(...)`. I vote for `void`, letting each module to decide whatever it wants. Input parameters might be changed in future (breaking change), but I think it will be another story.

dotnet/src/webdriver/BiDi/Extensions/Permissions/PermissionDescriptor.cs
RenderMichael: That's a good point, I'll move the namespace

dotnet/src/webdriver/BiDi/Module.cs
RenderMichael: I don't see an issue with this, especially because we have specific requirements

1. Each Module controls its own JSON context
2. Modules must be able to re-use contexts from a cache

Maybe we can rename this to `CreateJsonContext`, and if necessary, introduce a separate `Initialize` in the future?

dotnet/src/webdriver/BiDi/Browser/BrowserModule.cs
RenderMichael: The base class is responsible for creating the JSON context. This is the only way to bridge the gap between "strongly-typed JSON context" and the logic in Module.Create

RenderMichael: Another way to get around this is by making the Json Context a generic parameter, eg. `BrowserModule : Module<BiDiJsonSerializerContext>`. That requires making BiDiJsonSerializerContext public, and more publicly exposed. I don't think that's desirable.

nvborisenko: I also was thinking about `BrowserModule : Module<BiDiJsonSerializerContext>`, but it seems even stranger. Module of JsonContext... I am reading it hard, given that `BiDiJsonSerializerContext` cannot be inherited.

Ideally we should get rid of `BiDiJsonSerializerContext`, we are moving in this direction.

About `internal new BiDiJsonSerializerContext JsonContext => (BiDiJsonSerializerContext)base.JsonContext;`: For me it is risky, at least not obvious...
- is it one time initialization? (question from interview)
- what is executed first? our `void Initialize(...)` or this `implicit initialization`
- does it mean that we cast each time when we invoke `JsonContext` getter?

Private field looks really simple.

dotnet/src/webdriver/BiDi/Browser/BrowserModule.cs
nvborisenko: Obsoleted.

dotnet/src/webdriver/BiDi/Browser/BrowserModule.cs
RenderMichael: If we want to go that route, then we will have to undo optimizations such as this one: https://github.com/SeleniumHQ/selenium/pull/16397 

Otherwise, we will need a separate JSON context for each module

> Ideally we should get rid of `BiDiJsonSerializerContext`, we are moving in this direction.

What do you have in mind?

RenderMichael: Second optimization that would need to be de-optimized: https://github.com/SeleniumHQ/selenium/pull/16392

nvborisenko: Internal modules will still use global JsonContext for a while, because of f**g cross-referencing (stateful converters, polymorphic dto). In future, each module wants to:

```csharp
void Initialize(BiDi bidi, JsonOptions options) // just common options like Naming policy
{
  var moduleOptions = new JsonOptions(options) // make a copy
  {
     Converters = { new MySpecificConverter(bidi) } // bad example, just to highlight possibilities
  }

  _myContext = new MyCoontext(moduleOptions);
}
```

dotnet/src/webdriver/BiDi/WebDriver.Extensions.cs
```diff
@@ -41,4 +42,9 @@ public static async Task<BiDi> AsBiDiAsync(this IWebDriver webDriver, BiDiOption
 
         return bidi;
     }
+
+    public static PermissionsModule AsPermissions(this BiDi bidi)
```
nvborisenko: This is definitely should be in another class. The reason is we are CLS compatible (I don't still understand why, but anyway), meaning some languages cannot deal with `extension methods`. They will use static method: `WebDriverExtensions.AsPermissions(bidi)`. But it is not extension of `WebDriver`.

We should introduce good name for static class, and its namespace.

`BiDiExtensions.AsPermissions(bidi)`?

dotnet/src/webdriver/BiDi/Browser/BrowserModule.cs
RenderMichael: Since we do cannot have a global JsonContext for third-party extensions, it looks like we will need 2 different kinds of modules: extensions and non-extension (core?).

> because of f**g cross-referencing (stateful converters, polymorphic dto)

That shouldn't be a problem! We can get rid of stateful converters (very slightly less pleasant usability, I think it is no problem) and polymorphic DTOs got addressed by moving the converters out to attributes.

The only remaining converter is `DateTimeOffsetConverter`. We *could* go through to each instance of `DateTimeOffset` and add that converter as an attribute. Alternatively, we could change the type to `long` in the DTO, since it is technically magic. If we want to address that, then we will have no converters left!

dotnet/src/webdriver/BiDi/WebDriver.Extensions.cs
nvborisenko: When `Bluetooth` module comes to under umbrella, then what? Should we have single static class for all them, or they define own static class in theirs namespace?

I vote for "everything in their namespace". It would be good pattern to demonstrate good practices how to develop module. Separate namespace is not a problem, IDE is still able to suggest `extension method`.

RenderMichael: extension methods do not affect CLS compliance. Users can call `WebDriverExtensions.AsPermissions(bidi)` even on extension methods.

But I agree that extensions should be in separate files. I will make those changes.

dotnet/src/webdriver/BiDi/Browser/BrowserModule.cs
nvborisenko: `DateTimeOffsetConverter` is not a problem, it is kind of common JsonOptions. it is OK to have global well-known converters.

I don't want to block this PR, we can improve later. But seems `private MyJsonContext` should be here. `Broker` requires `JsonTypeInfo`.

nvborisenko: > Since we do cannot have a global JsonContext for third-party extensions, it looks like we will need 2 different kinds of modules: extensions and non-extension (core?).

Each module should have own JsonContext, ideally. And it is out of scope of this PR.

dotnet/src/webdriver/BiDi/WebDriver.Extensions.cs
nvborisenko: My concern is about naming: `WebDriverExtensions.AsPermissions(bidi)` - it is not extension for `WebDriver`, it is for `BiDi`.

RenderMichael: I followed the two suggestions: `WebDriverExtensions` -> `BiDiExtensions`, and permissions in its own area

nvborisenko: Not sure it is good naming. I thought we leave `WebDriverExtensions.AsBiDiAsync()` in `BiDi` namespace, and introduce `BiDiExtensions.AsPermissions()` in `BiDi.Permissions` namespace. But even not sure about this. I think of we may find a good reference in popular project how usually it can be named.

nvborisenko: ### 1. Isolated namespace
What if:

image

And usage:

image

For IDE and compiler it is not a problem. For CLS they will specify full path to `BiDiExtensions` static class (I suppose).

### 2. Simplified

Extension methods leaves in root `BiDi` namespace, because modules are under our umbrella... why not to simplify...

@RenderMichael @YevgeniyShunevych Let's vote.

dotnet/src/webdriver/BiDi/Browser/BrowserModule.cs
nvborisenko: Resolving this thread to just unblock.

dotnet/src/webdriver/BiDi/WebDriver.Extensions.cs
YevgeniyShunevych: I definitely vote for # 2 Simplified. Less namespaces.

RenderMichael: The common pattern I have seen is either.

1. For type `MyType`, the extension class is called `MyTypeExtensions`
2. More topic-related. For example, the many Span<T> extension methods are in `MemoryExtensions`. Additionally, Linq methods are on `Enumerable`.

For the case of Permissions, I prefer the topic-related name of `PermissionsExtensions`, but I don’t have strong opinions here.

RenderMichael: I like the motif of “everything Permissions in a separate permissions area/namespace”. That way, users will have to import Permissions to use that extension method. I like the clarity it provides.

dotnet/src/webdriver/BiDi/Module.cs
nvborisenko: Let's move on as is, improve later.

dotnet/src/webdriver/BiDi/Module.cs
nvborisenko: > Modules must be able to re-use contexts from a cache

This is obsolete requirement, we already learnt JsonContext cannot be extended while the original one is already in use. I see only one way: each module will create a copy of JsonOptions (if necessary), and create own `new JsonContext(jsonOptions)`. For now seems only JsonOptions are required by modules to be initialized. But what if in future some more parameters may be required. This is why preserving `void Initialize(....)` is better.

Let's think more... what if `void Configure(Action<Options> act)` where `Options` is a holder for required parameters? Just to avoid breaking changes in future.

**BUT**
I don't expect custom 3rd party modules. We know only 3 modules, and we want to put it under Selenium umbrella. So let's move on as is.

RenderMichael: The solution I have implemented here fulfills that requirement while still being extensible while other modules are already in use. Let me know what you think we can improve!

dotnet/src/webdriver/BiDi/WebDriver.Extensions.cs
nvborisenko: If we add all extension methods in one `BiDiExtensions` class under root `BiDi` namespace. And a developer wants to create his own extension under `BiDi` namespace. Is it criminal? Does he break end-users? I am thinking over naming conflicts. (UPD: I mean class with the same FQN but in different assemblies)

Personally I vote for # 2, but let's analyze impact. # 1 is very straightforward, even if end-user imports namespace.

RenderMichael: There are some libraries that wrap/extend Selenium that use the `OpenQA.Selenium` namespace segment, so it’s not a marginal concern.

nvborisenko: My final confidence: # 1 - 65%, # 2 - 35%;

Please think and share your confidence (don't cheat).

RenderMichael: For me: # 1 - 75%, # 2 - 25%

nvborisenko: @YevgeniyShunevych please just vote blindly.

YevgeniyShunevych: From my point of view, extra namespaces within the same assembly, which do not have strong practical need, often provide only complexity without benefits. I'm for more clear type names, rather then giving short type names, which increases name conflict chance, and putting such types in different namespaces.

For example, Visual Studio 2022 IntelliSense, will not suggest you `AsPermissions()` extension method for `bidi` object unless you explicitly add `using OpenQA.Selenium.BiDi.Permissions;` to a `.cs` file before that. But I would like to see all the available "As..." methods in IDE without knowing and adding exact namespace before that.

I generally do support project folder structuring, similar as currently Selenium.WebDriver project is structured. It is also fine that we have all functionality under `OpenQA.Selenium.BiDi` namespace. But I would stop with sub-namespaces at this point. I don't see a strong need having 10+ sub-namespaces, one for each submodule. I am talking not only about extension classes, but all classes inside `/dotnet/src/webdriver/BiDi` folder.

Despite of namespace topic, the current BiDi folder is good. I'm ok with keeping methods like `AsPermissions` in separate extension classes inside corresponding folder ("Permissions"). If I had to choose a name for such extension class having method `public static PermissionsModule AsPermissions(this BiDi bidi)`, I would name it `PermissionsBiDiExtensions`, not `PermissionsExtensions`, because the class's extension methods extend `BiDi` type. So just `{Module}BiDiExtensions`.

If you are not ready to apply single common `OpenQA.Selenium.BiDi` namespace for all types of `/dotnet/src/webdriver/BiDi`, it's understandable. But at least, it would be great to have `{Module}BiDiExtensions` in `OpenQA.Selenium.BiDi` namespace.

nvborisenko: Thanks for your valuable input.

> If you are not ready to apply single common OpenQA.Selenium.BiDi namespace for all types of /dotnet/src/webdriver/BiDi, it's understandable.

We will have hundreds files in one folder, with naming collision (despite that one file may contain several classes). I really like how we split it by modules, absolutely the same as specification dictates. It is low-level representation, so I think we are OK with it.

> For example, Visual Studio 2022 IntelliSense, will not suggest you AsPermissions() extension method for bidi object unless you explicitly add using OpenQA.Selenium.BiDi.Permissions;

We should double check it. I already provided screenshot how VS suggests `AsPermissions()` extension method even if I didn't import that namespace.

> I would name it PermissionsBiDiExtensions, not PermissionsExtensions, because the class's extension methods extend BiDi type. So just {Module}BiDiExtensions.

Absolutely, agree.

**To sum up:** now we need to decide in which namespace `PermissionsBiDiExtensions` should live, in root or in `Permissions`. I really don't have strong opinion. But if we put `PermissionsBiDiExtensions` into root namespace, it might be precedent to put all possible extensions for all modules in root.

RenderMichael: > Visual Studio 2022 IntelliSense, will not suggest you AsPermissions() extension method for bidi object unless you explicitly add using OpenQA.Selenium.BiDi.Permissions;

For me, modern VS and Rider suggest `AsPermissions()` even without the import statement. Like in the screenshot, the suggestion shows that it is in a new namespace, which signals "extension" to users.

I also agree with naming the extension provider type `PermissionsBiDiExtensions`. I will push those changes now.

RenderMichael: > now we need to decide in which namespace `PermissionsBiDiExtensions` should live, in root or in `Permissions`.

If we have `AsPermissions()` in a separate "permissions extensions" type, then it seems strange not to have that in the `OpenQA.Selenium.BiDi.Permissions` namespace. I don't have a strong opinion on moving `AsPermissions()` back to `BiDiExtensions`, though.

nvborisenko: No consensus :( Let me decide based on `external modules should be emphasized`. I propose to finish this thread with:
- all DTO types in `Permissions` namespace
- any external module might provide extension method in `{Module}BiDiExtensions` static class under `BiDi.{Module}` namespace

nvborisenko:
I didn't verify whether models follow patterns. Since we are OK with the approach, please give me a time to focus on DTO.

dotnet/src/webdriver/BiDi/Permissions/PermissionsModule.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Text.Json.Serialization;
+using System.Threading.Tasks;
+
+namespace OpenQA.Selenium.BiDi.Extensions.Permissions;
+
+public class PermissionsModule : Module
+{
+    private PermissionsJsonSerializerContext JsonContext => (PermissionsJsonSerializerContext)base.JsonContext;
+
+    public async Task SetPermissionAsync(string permissionName, PermissionState state, string origin, UserContext? userContext, SetPermissionOptions? options = null)
```
nvborisenko: It doesn't follow patter:

```
permissions.SetPermissionParameters = {
  descriptor: permissions.PermissionDescriptor,
  state: permissions.PermissionState,
  origin: text,
  ? embeddedOrigin: text,
  ? userContext: text,
}
```

where `descriptor`, `state`, `origin` is mandatory.

nvborisenko: It should accept `PermissionDescriptor`.

Syntax sugar? `PermissionDescriptor` might be implicitly converted from `string`.

nvborisenko: Please avoid optional parameters. We are library, introducing one more option al parameter is risky, literally is a behavior breaking change.

Use `Options` instead.

dotnet/src/webdriver/BiDi/Permissions/PermissionsModule.cs
```diff
@@ -0,0 +1,48 @@
...
+        await Broker.ExecuteCommandAsync(new SetPermissionCommand(@params), options, JsonContext.Permissions_SetPermissionCommand, JsonContext.Permissions_SetPermissionResult).ConfigureAwait(false);
+    }
+
+    protected override JsonSerializerContext CreateJsonContext(JsonSerializerOptions options)
+    {
+        return new PermissionsJsonSerializerContext(options);
+    }
+}
+
+[JsonSerializable(typeof(SetPermissionCommand), TypeInfoPropertyName = "Permissions_SetPermissionCommand")]
```
nvborisenko:
```suggestion
[JsonSerializable(typeof(SetPermissionCommand))]
```

It is not necessary, and hope naming conflicts will never be inside module (we are moving to JsonContext per Module)

nvborisenko:
We are on the finish line. @RenderMichael will you be able to pick up it?

dotnet/src/webdriver/BiDi/Permissions/PermissionsModule.cs
RenderMichael: I am seeing the pattern of `XYZOptions? options = null` as the last parameter across many modules.

I'm not sure what you mean by using `Options` instead. Could you paste the code for what you are picturing?

dotnet/src/webdriver/BiDi/Permissions/PermissionsModule.cs
RenderMichael: Done.

dotnet/src/webdriver/BiDi/Permissions/PermissionsModule.cs
RenderMichael: Not sure how I missed `embeddedOrigin`. Maybe the spec changed without me noticing.

Updated to expose all parameters directly.

RenderMichael: > Syntax sugar? `PermissionDescriptor` might be implicitly converted from `string`.

Ever since the release of C# 14's extension functionality, my appetite for syntax sugar has plummeted. Users can now define this implicit conversion if they desire.

nvborisenko: Sorry, was not clear. Preferred way:

```csharp
SetPermissionAsync(string permissionName, object anotherRequiredArgument, new() { UserContext = ..., AnotherOptionalArgument = ... })
```

dotnet/src/webdriver/BiDi/Permissions/PermissionsModule.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Text.Json.Serialization;
+using System.Threading.Tasks;
+
+namespace OpenQA.Selenium.BiDi.Extensions.Permissions;
+
+public class PermissionsModule : Module
+{
+    private PermissionsJsonSerializerContext JsonContext => (PermissionsJsonSerializerContext)base.JsonContext;
+
+    public async Task SetPermissionAsync(PermissionDescriptor permissionName, PermissionState state, string origin, string? embeddedOrigin, UserContext? userContext, SetPermissionOptions? options = null)
```
nvborisenko: `permissionName` -> `descriptor` ?

nvborisenko: Return `Task<SetPermissionResult>`

dotnet/src/webdriver/BiDi/BiDiExtensions.cs
```diff
@@ -22,7 +22,7 @@
 
 namespace OpenQA.Selenium.BiDi;
 
-public static class WebDriverExtensions
+public static class BiDiExtensions
```
nvborisenko: It is still `WebDriverExtentions`, extension methods for `WebDriver`. Previously it was correct.

nvborisenko: Done.

dotnet/src/webdriver/BiDi/Permissions/PermissionsModule.cs
nvborisenko: Done.

nvborisenko: Done.

dotnet/src/webdriver/BiDi/Permissions/PermissionsModule.cs
nvborisenko: Done.

nvborisenko: Done.

dotnet/src/webdriver/BiDi/WebDriver.Extensions.cs
nvborisenko: Done, thanks all!

nvborisenko (approved):
It was amazing collaboration, thanks @RenderMichael and @YevgeniyShunevych!

PR #16442, opened 2025-10-15
[rb] Fix Network issue by removing nil values on network requests

aguspe:
### 💥 What does this PR do?
This PR resolves a bug where optional, unspecified parameters were being sent as `null` in the JSON payload for network modification commands (`network.continueRequest`, `network.continueResponse`, `network.provideResponse`). 

This was causing errors, particularly with the Firefox BiDi implementation, which expects optional fields to be nil.

The fix uses the Ruby method `.compact` to remove any key-value pairs with a `nil` value from the command arguments hash before serialization.

***

### 🔧 Implementation Notes
The core of the issue was that parameters like `cookies`, `headers`, or `body`, when left unprovided by the user, defaulted to `nil`. 

When constructing the command hash (e.g., in `#continue_request`), these `nil` values were present, and upon JSON serialization, they became JSON `null` values.

We called **`.compact`** on the hash arguments to remove nil values:

```ruby
          args = {
            request: args[:id],
            body: args[:body],
            cookies: args[:cookies],
            # ...
          }.compact
```

### 🔄 Types of changes


rb/lib/selenium/webdriver/bidi/network.rb
```diff
@@ -82,15 +82,16 @@ def cancel_auth(request_id)
...
-            'network.continueRequest',
+          args = {
             request: args[:id],
             body: args[:body],
             cookies: args[:cookies],
             headers: args[:headers],
             method: args[:method],
             url: args[:url]
-          )
+          }.compact
```
titusfortner, line 92: The issue is that headers and cookies aren't nil by default, they are empty objects that get serialized with empty parameters.
https://github.com/SeleniumHQ/selenium/blob/92be65e3ebbad3801e7791fdbf528c0e54dd82bd/rb/lib/selenium/webdriver/bidi/network/intercepted_request.rb#L59-L65

titusfortner: So I think these should just be nil by default not empty hash or empty object

aguspe: Great thank you Titus, I updated them now and I added more tests to coverwere the headers and cookies are nil

titusfortner (approved):

diemol (approved):

diemol:
A few tests are consistently failing. Could you please have a look, @aguspe?

titusfortner:
well, I don't think that's what I wanted to do. It doesn't want me to push from command line, so I tried to edit on Github, don't think it did what I wanted it to do...

titusfortner:
I accidentally copied the integration tests onto the unit tests.
I've put them back, but between github complaining about my authentication from CLI and inability to edit a file from the browser in a PR that doesn't have that file edited.... I can't fix this one. Easy enough to do, just use the header and cookie setters with a Hash.

aguspe:
@titusfortner @diemol I fixed the unit and integration tests, please titus let me know if any changes are needed after the review, thanks!

PR #16516, opened 2025-10-27
[Java] Fix array copy length in PrintOptions.setPageRanges method

manuelsblanco:
###  Related Issues
Fixes a bug in `PrintOptions#setPageRanges` where the last page range argument was not being copied due to an off-by-one error in the `System.arraycopy` call.

### What does this PR do?
This PR fixes a small but critical logic error in the `setPageRanges(String firstRange, String... ranges)` method inside `org.openqa.selenium.print.PrintOptions`.

Previously, only `ranges.length - 1` elements were copied from the `ranges` array into `this.pageRanges`.  
As a result, the **last page range was silently omitted** whenever multiple ranges were passed to the method.

#### Example of the previous behavior:
```java
setPageRanges("1-3", "5", "8-10");
// Result before fix:
["1-3", "5", null]
setPageRanges("1-3", "5", "8-10");
// Correct result:
["1-3", "5", "8-10"]
```


asolntsev (approved):
@diemol I recommend to merge it.

PR #16557, opened 2025-11-06
[dotnet] Support `UnhandledPromptBehavior` option as string and map (breaking change)

### LINKED ISSUE #16102: [🐛 Bug]: [dotnet] Unable to set dictionary to the UnhandledPromptBehavior option
### Description

Starting from Chrome 137 it's unable to manage browser prompts (alerts) using BiDi driver. Issues on ChromeDriver were created, but they were closed as intendent behavior:
https://issues.chromium.org/issues/425390962
https://issues.chromium.org/issues/428153658

Based on the [specification ](https://www.w3.org/TR/webdriver/#dfn-get-the-prompt-handler) we should provide unhandledPromptBehavior as a dictionary. Based on the comment in [the issue](https://issues.chromium.org/issues/428153658) it's possible to do it for java, using MutableCapabilities, but I didn't find any solution for c#. 

In my project I use Selenium 4.23 + .net 4.7.2. I tried the following code:

```
var promptBehavior = new Dictionary<string, string>
{
				{"alert", "ignore"},
				{"confirm", "ignore"},
				{"prompt", "ignore"},
				{"beforeUnload", "ignore"},
				{"default", "ignore"}
};
options.AddAdditionalOption("unhandledPromptBehavior", promptBehavior);
```
But exception is thrown : _System.ArgumentException: 'There is already an option for the unhandledPromptBehavior capability. Please use the UnhandledPromptBehavior property instead.
Parameter name: capabilityName'_

I can't set dictionary to the options.UnhandledPromptBehavior either, because it's waiting for the UnhandledPromptBehavior map (for example, UnhandledPromptBehavior.Ignore).

Before Selenium 4 we were able to set capabilities via DesiredCapabilities, but in Selenium 4.23 this class is internal. 

So, it seems, that there are no possibility to set dictionary to the unhandledPromptBehavior option when starting Chrome Driver. 

Note: everything worked fine till Chrome 137. 

### Reproducible Code

```shell
1. Start Chrome 137+ using BiDi ChromeDriver and use the following Chrome options:
options.UseWebSocketUrl = true;
options.UnhandledPromptBehavior = UnhandledPromptBehavior.Ignore;

Actual result: browser alert is not shown
Expected result: alert is shown, because UnhandledPromptBehavior was set to Ignore. 

Possible reason: value for UnhandledPromptBehavior is not a dictionary.
```

nvborisenko:
### 🔗 Related Issues
Fixes https://github.com/SeleniumHQ/selenium/issues/16102

### 💥 What does this PR do?
This PR changes `UnhandledPromptBehavior` property return type to custom type, supporting both `string` and `map` representation at serialization phase.

```csharp
// Same handler for every prompt
options.UnhandledPromptBehavior = UnhandledPromptBehavior.Accept;

// Per prompt type (only set what you need)
options.UnhandledPromptBehavior = new UserPromptHandler.PerPromptType
{
    Alert = UnhandledPromptBehavior.Accept,
    Confirm = UnhandledPromptBehavior.Dismiss,
    File = UnhandledPromptBehavior.Dismiss,    // BiDi only
    Default = UnhandledPromptBehavior.DismissAndNotify, // fallback
};

// Leave unset (driver picks spec default = "dismiss and notify")
options.UnhandledPromptBehavior = null;
options.UnhandledPromptBehavior = default;
```

### 🔧 Implementation Notes
It is not a breaking change at compilation level! One file change to emphasize that users are not affected.

I intentionally didn't introduce new property in Options class, it would lead to a mess in my opinion.

### 🔄 Types of changes

- New feature (non-breaking change for 99% users)
- Breaking change, binary

___

### **PR Type**
Enhancement

___

### **Description**
- Support `UnhandledPromptBehavior` as both string and dictionary representations

- Introduce abstract record hierarchy for flexible prompt behavior configuration

- Enable per-prompt-type behavior specification via `UnhandledPromptBehaviorMultiOption`

- Maintain backward compatibility with existing enum-based API

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["UnhandledPromptBehavior enum"] -->|implicit conversion| B["UnhandledPromptBehaviorOption"]
  B -->|Single| C["UnhandledPromptBehaviorSingleOption"]
  B -->|Multi| D["UnhandledPromptBehaviorMultiOption"]
  C -->|serializes to| E["string capability"]
  D -->|serializes to| F["dictionary capability"]
```

<tr>
  <td>
    <details>
      <summary><strong>DriverOptions.cs</strong><dd><code>Add flexible prompt behavior option types with dual serialization</code></dd></summary>
<hr>

dotnet/src/webdriver/DriverOptions.cs

<ul><li>Introduced abstract record <code>UnhandledPromptBehaviorOption</code> with two <br>sealed implementations: <code>UnhandledPromptBehaviorSingleOption</code> (for <br>string representation) and <code>UnhandledPromptBehaviorMultiOption</code> (for <br>dictionary representation with per-prompt-type settings)<br> <li> Added implicit operator and factory methods (<code>Single()</code>, <code>Multi()</code>) for <br>convenient instantiation<br> <li> Changed <code>UnhandledPromptBehavior</code> property type from enum to nullable <br><code>UnhandledPromptBehaviorOption</code><br> <li> Refactored serialization logic in <code>GenerateDesiredCapabilities()</code> to <br>handle both single and multi-option cases, converting to appropriate <br>capability format<br> <li> Updated merge conflict detection to use null checks instead of enum <br>comparison<br> <li> Extracted enum-to-string conversion into local helper function <br><code>UnhandledPromptBehaviorToString()</code></ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16557/files#diff-3a23deddc7c714feb41423be7a50f6e63df3541f4a50e21c28286d52fa3584fa">+79/-18</a>&nbsp; </td>

</tr>

</details>

___


dotnet/src/webdriver/DriverOptions.cs
```diff
@@ -26,6 +26,37 @@
 
 namespace OpenQA.Selenium;
 
+public abstract record UnhandledPromptBehaviorOption
```
nvborisenko: I will add xml docs as soon as we approve the approach.

dotnet/src/webdriver/DriverOptions.cs
```diff
@@ -26,6 +26,37 @@
...
+{
+    public static implicit operator UnhandledPromptBehaviorOption(UnhandledPromptBehavior? value)
+        => Single(value);
+
+    public static UnhandledPromptBehaviorOption Single(UnhandledPromptBehavior? value)
+        => new UnhandledPromptBehaviorSingleOption(value);
+
+    public static UnhandledPromptBehaviorOption Multi(UnhandledPromptBehavior? alert = null,
```
nvborisenko: Do we really need static factory? Even with optional method arguments? We can, but optional arguments are risky in the library (breaking change if we add new arg)

dotnet/src/webdriver/DriverOptions.cs
nvborisenko: Move out from this file to dedicated (with "nested/related" types).

dotnet/src/webdriver/DriverOptions.cs
```diff
@@ -508,29 +539,59 @@ protected IWritableCapabilities GenerateDesiredCapabilities(bool isSpecification
...
+        static string? UnhandledPromptBehaviorToString(UnhandledPromptBehavior? behavior) => behavior switch
         {
-            string unhandledPropmtBehaviorSetting = "ignore";
-            switch (this.UnhandledPromptBehavior)
+            Selenium.UnhandledPromptBehavior.Ignore => "ignore",
+            Selenium.UnhandledPromptBehavior.Accept => "accept",
+            Selenium.UnhandledPromptBehavior.Dismiss => "dismiss",
+            Selenium.UnhandledPromptBehavior.AcceptAndNotify => "accept and notify",
+            Selenium.UnhandledPromptBehavior.DismissAndNotify => "dismiss and notify",
+            _ => null
```
nvborisenko: We should raise exception by default. Otherwise there is a chance to miss branch with unrecognized enum value.

nvborisenko: Fixed.

dotnet/src/webdriver/DriverOptions.cs
```diff
@@ -303,7 +334,7 @@ public virtual DriverOptionsMergeResult GetMergeResult(DriverOptions other)
             return result;
         }
 
-        if (this.UnhandledPromptBehavior != UnhandledPromptBehavior.Default && other.UnhandledPromptBehavior != UnhandledPromptBehavior.Default)
+        if (this.UnhandledPromptBehavior is not null && other.UnhandledPromptBehavior is not null)
```
nvborisenko: Hm, `!=` operator for records?

RenderMichael: It does feel like this method should tolerate it when both values are the same.

However, we should investigate the history of this method before making those decisions.

RenderMichael (changes requested):
A breaking change like this should be well-tested. Could we add tests here?

dotnet/src/webdriver/DriverOptions.cs
nvborisenko: I understand it as don't send anything over wire if default.

nvborisenko:
> A breaking change like this should be well-tested. Could we add tests here?

I have no capacity to test the functionality which was not indented to be tested. You can "request changes", please propose.

RenderMichael:
> > A breaking change like this should be well-tested. Could we add tests here?
> 
> I have no capacity to test the functionality which was not indented to be tested. You can "request changes", please propose.

I propose new tests for this functionality.

nvborisenko:
Existing tests don't satisfy you? You ask me to increase the coverage. But what is coverage?

nvborisenko:
@RenderMichael welcome to https://github.com/SeleniumHQ/selenium/issues/15536

UPD: is it only one concern from your side?

YevgeniyShunevych:
Generally, looks good, @nvborisenko. What I don't like is a nullability of properties and parameters (`UnhandledPromptBehaviorOption?`, `UnhandledPromptBehavior?`). If I understand correctly, for `UnhandledPromptBehavior?` `null` and `Default` values are equivalent. So 2 states basically indicate the same state. So why not to make it non-nullable and use `Default` as a default. I would also rename `UnhandledPromptBehavior.Default` to `None`, as it is not clear what "Default" means, but "None" means no behavior is specified.

dotnet/src/webdriver/DriverOptions.cs
```diff
@@ -26,6 +26,37 @@
...
+
+    public static UnhandledPromptBehaviorOption Multi(UnhandledPromptBehavior? alert = null,
+        UnhandledPromptBehavior? confirm = null,
+        UnhandledPromptBehavior? prompt = null,
+        UnhandledPromptBehavior? beforeUnload = null,
+        UnhandledPromptBehavior? @default = null)
+        => new UnhandledPromptBehaviorMultiOption() with { Alert = alert, Confirm = confirm, Prompt = prompt, BeforeUnload = beforeUnload, Default = @default };
+}
+
+public sealed record UnhandledPromptBehaviorSingleOption(UnhandledPromptBehavior? Value) : UnhandledPromptBehaviorOption;
```
nvborisenko: Since we have `UnhandledPromptBehavior.Default` enum value, than making nullable enum is strange? Let's be aligned with another similar properties. So everything non nullable.

dotnet/src/webdriver/DriverOptions.cs
```diff
@@ -26,6 +26,37 @@
...
+
+public sealed record UnhandledPromptBehaviorMultiOption : UnhandledPromptBehaviorOption
```
nvborisenko: @RenderMichael any alternative names for these records? I like `UnhandledPromptBehavior*Option` pattern.

dotnet/src/webdriver/DriverOptions.cs
nvborisenko: Fixed.

dotnet/src/webdriver/DriverOptions.cs
nvborisenko: > However, we should investigate the history of this method before making those decisions.

Oh yes, we should understand the purpose of this method, not obvious to me at glance.

nvborisenko: @jimevans do you know?

dotnet/src/webdriver/DriverOptions.cs
nvborisenko: No optional method args.

dotnet/src/webdriver/UnhandledPromptBehaviorOption.cs
```diff
@@ -0,0 +1,150 @@
...
+/// </summary>
+/// <remarks>Use this class to configure distinct behaviors for alert, confirm, prompt, and beforeunload dialogs
+/// encountered during browser automation. Each property allows you to control the response to a specific type of
+/// unhandled prompt, enabling fine-grained handling beyond a single global setting.</remarks>
+public sealed record UnhandledPromptBehaviorMultiOption : UnhandledPromptBehaviorOption
+{
+    /// <summary>
+    /// Gets or sets the behavior to use when an unexpected alert is encountered during automation.
+    /// </summary>
+    public UnhandledPromptBehavior Alert { get; set; } = UnhandledPromptBehavior.Default;
```
RenderMichael, line 108: Is it correct that

1) All of these are set to Default
2) They have setters?

nvborisenko: Please help me if you see hidden stones.

RenderMichael: I have no idea if these work because there are no tests yet, but this looks wrong.

nvborisenko: True, it looks strange because historically we have `Default` in enum. It should be nullable, but it is as is. Please read comments in this PR.

How to move on...

dotnet/src/webdriver/DriverOptions.cs
```diff
@@ -508,29 +470,55 @@ protected IWritableCapabilities GenerateDesiredCapabilities(bool isSpecification
...
+            _ => throw new ArgumentOutOfRangeException(nameof(behavior), $"UnhandledPromptBehavior value '{behavior}' is not recognized."),
+        };
+
+        if (this.UnhandledPromptBehavior is UnhandledPromptBehaviorSingleOption singleOption && singleOption.Value != Selenium.UnhandledPromptBehavior.Default)
+        {
+            var stringValue = UnhandledPromptBehaviorToString(singleOption.Value);
+
+            capabilities.SetCapability(CapabilityType.UnhandledPromptBehavior, stringValue);
+        }
+        else if (this.UnhandledPromptBehavior is UnhandledPromptBehaviorMultiOption multiOption)
```
RenderMichael: We should follow our usual pattern here to have a `ToDictionary()` method directly on the `UnhandledPromptBehaviorMultiOption` type:

```C#
public Dictionary<string, string> ToDictionary()
{
    Dictionary<string, string> multiOptionDictionary = [];

    if (Alert is not Selenium.UnhandledPromptBehavior.Default)
    {
        multiOptionDictionary["alert"] = UnhandledPromptBehaviorToString(Alert);
    }

    if (Confirm is not Selenium.UnhandledPromptBehavior.Default)
    {
        multiOptionDictionary["confirm"] = UnhandledPromptBehaviorToString(Confirm);
    }

    if (Prompt is not Selenium.UnhandledPromptBehavior.Default)
    {
        multiOptionDictionary["prompt"] = UnhandledPromptBehaviorToString(Prompt);
    }

    if (BeforeUnload is not Selenium.UnhandledPromptBehavior.Default)
    {
        multiOptionDictionary["beforeUnload"] = UnhandledPromptBehaviorToString(BeforeUnload);
    }

    if (Default is not Selenium.UnhandledPromptBehavior.Default)
    {
        multiOptionDictionary["default"] = UnhandledPromptBehaviorToString(Default);
    }

    return multiOptionDictionary;
}
```

and then to call that method here

```C#
else if (this.UnhandledPromptBehavior is UnhandledPromptBehaviorMultiOption multiOption)
{
    multiOptionDictionary = multiOption.ToDictionary();
    if (multiOptionDictionary.Count != 0)
    {
        capabilities.SetCapability(CapabilityType.UnhandledPromptBehavior, multiOptionDictionary);
    }
}
```

dotnet/src/webdriver/DriverOptions.cs
```diff
@@ -508,29 +470,55 @@ protected IWritableCapabilities GenerateDesiredCapabilities(bool isSpecification
...
+                multiOptionDictionary["beforeUnload"] = UnhandledPromptBehaviorToString(multiOption.BeforeUnload);
             }
 
-            capabilities.SetCapability(CapabilityType.UnhandledPromptBehavior, unhandledPropmtBehaviorSetting);
+            if (multiOption.Default is not Selenium.UnhandledPromptBehavior.Default)
+            {
+                multiOptionDictionary["default"] = UnhandledPromptBehaviorToString(multiOption.Default);
+            }
+
+            if (multiOptionDictionary.Count != 0)
```
RenderMichael: Do we need to guard against `{}`? It feels like we should transparently sent this to the server.

What happens if we do this, anyway?

dotnet/src/webdriver/DriverOptions.cs
RenderMichael: This also allows us to move `UnhandledPromptBehaviorToString` to the `UnhandledPromptBehaviorOption`, cleaning up this 150-line method.

dotnet/src/webdriver/UnhandledPromptBehaviorOption.cs
```diff
@@ -0,0 +1,150 @@
...
+    /// <param name="value">The <see cref="UnhandledPromptBehavior"/> to apply for all prompt types.</param>
+    /// <returns>An <see cref="UnhandledPromptBehaviorOption"/> wrapping the provided behavior.</returns>
+    public static UnhandledPromptBehaviorOption Single(UnhandledPromptBehavior value)
+        => new UnhandledPromptBehaviorSingleOption(value);
+
+    /// <summary>
+    /// Creates an <see cref="UnhandledPromptBehaviorOption"/> allowing individual <see cref="UnhandledPromptBehavior"/> values per prompt type.
+    /// </summary>
+    /// <returns>An <see cref="UnhandledPromptBehaviorOption"/> with per-prompt configurable behaviors.</returns>
+    public static UnhandledPromptBehaviorOption Multi()
```
RenderMichael: These semantics are a little unwieldy:

Image

Maybe we could have a nested class here?

Image

RenderMichael: Since we are changing the type name, I think it would align better with the spec to call it `UnhandledPromptHandlers`

dotnet/src/webdriver/UnhandledPromptBehaviorOption.cs
```diff
@@ -0,0 +1,150 @@
...
+    /// </summary>
+    /// <remarks>Use this property to specify how the application should respond to beforeunload dialogs that
+    /// appear unexpectedly during automated browser interactions. This setting determines whether such dialogs are
+    /// automatically accepted, dismissed, or cause an error.</remarks>
+    public UnhandledPromptBehavior BeforeUnload { get; set; } = UnhandledPromptBehavior.Default;
+
+    /// <summary>
+    /// Gets or sets the default behavior to use when an unexpected browser prompt is encountered.
+    /// </summary>
+    public UnhandledPromptBehavior Default { get; set; } = UnhandledPromptBehavior.Default;
```
RenderMichael, line 138: Missing `file` https://w3c.github.io/webdriver/#dfn-valid-prompt-types

RenderMichael: Never mind, seeing

> The "file" prompt type is respected only in [[WebDriver-BiDi](https://w3c.github.io/webdriver/#bib-webdriver-bidi)] sessions.

dotnet/src/webdriver/DriverOptions.cs
nvborisenko: Applied `!=` operator from records. Theoretically this method compares two values, and if they are not equal mark the merge process as conflicted. I guess `!=` does the job for us.

dotnet/src/webdriver/DriverOptions.cs
nvborisenko: Good idea to introduce `ToCapabilities()` internal method.

dotnet/src/webdriver/DriverOptions.cs
nvborisenko: Originally we sent nothing if the value is `default`, meaning we don't spam wire if user don't specifies non-default. Anyway need to double check whether we really avoid spamming.

dotnet/src/webdriver/UnhandledPromptBehaviorOption.cs
nvborisenko: Fixed static factory, now it is possible to:

```csharp
UnhandledPromptBehaviorOption.Multi() with { BeforeUnload = UnhandledPromptBehavior.Accept }
```

Interesting point about `UnhandledPromptHandlers`.

dotnet/src/webdriver/UnhandledPromptBehaviorOption.cs
nvborisenko: It should be here as well.

dotnet/src/webdriver/UnhandledPromptBehaviorOption.cs
nvborisenko: 1. We already discussed nested types are bad (in BiDi context), preventing us introducing static factories.
2. About `UnhandledPromptHandlers` type name.. Could you please provide examples (`Single`/`Multi`)? I think of `UnhandledPromptHandlers` is more `IEnumerarable<>` rather than regular class.

dotnet/src/webdriver/UnhandledPromptBehaviorOption.cs
nvborisenko: To use `bidi` we should initiate the first browser (`userContext/browsingContext`). It is done only via selenium classic command, and here user is able to specify initial behavior how to deal with prompts. BTW it is impossible to change this later. So yes, we need `file` property with good documentation.

JuliaNaumetsss:
@nvborisenko , are there any updates for this PR?
This issue is still blocker for us, and we are waiting for this fix.

nvborisenko:
Changes:
- `UserPromptHandler` class name according spec
- Applied nested types as suggested (static factory is useless here)
- Use `Nullable<>` instead of `Enum.Default` (`Default` is now obsoleted) - still no breaking change (compilation). It is better start using it instead break again in the future.

dotnet/src/webdriver/DriverOptions.cs
nvborisenko: Outdated.

dotnet/src/webdriver/UnhandledPromptBehaviorOption.cs
nvborisenko: `UserPromptHandler` spec defined... though

dotnet/src/webdriver/DriverOptions.cs
nvborisenko: Outdated.

dotnet/src/webdriver/DriverOptions.cs
```diff
@@ -162,7 +124,7 @@ protected DriverOptions()
     /// Gets or sets the value for describing how unexpected alerts are to be handled in the browser.
     /// Defaults to <see cref="UnhandledPromptBehavior.Default"/>.
     /// </summary>
-    public UnhandledPromptBehavior UnhandledPromptBehavior { get; set; } = UnhandledPromptBehavior.Default;
+    public UserPromptHandler? UnhandledPromptBehavior { get; set; } = Selenium.UnhandledPromptBehavior.Default;
```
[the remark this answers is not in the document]
nvborisenko: Good suggestion, but supporting 2 branches which additional branch will have bad name..

dotnet/src/webdriver/UserPromptHandler.cs
```diff
@@ -0,0 +1,195 @@
...
+{
+    private UserPromptHandler() { }
+
+    /// <summary>
+    /// Converts a nullable <see cref="UnhandledPromptBehavior"/> to a <see cref="UserPromptHandler"/> instance,
+    /// or <see langword="null"/> when <paramref name="value"/> is <see langword="null"/>.
+    /// </summary>
+    /// <param name="value">The <see cref="UnhandledPromptBehavior"/> value to convert.</param>
+    public static implicit operator UserPromptHandler?(UnhandledPromptBehavior? value)
+        => value is { } v ? new Uniform(v) : null;
```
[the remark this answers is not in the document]
nvborisenko: Hopefully no merge logic upfront.

dotnet/src/webdriver/UserPromptHandler.cs
```diff
@@ -0,0 +1,195 @@
...
+/// Available variants:
+/// <list type="bullet">
+/// <item><description><see cref="Uniform"/> - Wraps a single <see cref="UnhandledPromptBehavior"/> value applied to all prompt types. Create via the implicit conversion from <see cref="UnhandledPromptBehavior"/> or by constructing <see cref="Uniform"/> directly.</description></item>
+/// <item><description><see cref="PerPromptType"/> - Allows configuring per-prompt behaviors (Alert, Confirm, Prompt, BeforeUnload, File, Default).</description></item>
+/// </list>
+/// </para>
+/// </remarks>
+public abstract record UserPromptHandler
+{
+    private UserPromptHandler() { }
```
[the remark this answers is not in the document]
nvborisenko: I have no phantasy to distinguish everything. `LogLevel` is nice example :D

nvborisenko:
@YevgeniyShunevych @RenderMichael please review again, significantly simplified.

dotnet/src/webdriver/UserPromptHandler.cs
```diff
@@ -0,0 +1,195 @@
...
+        internal override object? ToCapabilities()
+        {
+            Dictionary<string, string> capabilities = [];
+            AddIfSet(capabilities, "alert", Alert);
+            AddIfSet(capabilities, "confirm", Confirm);
+            AddIfSet(capabilities, "prompt", Prompt);
+            AddIfSet(capabilities, "beforeUnload", BeforeUnload);
+            AddIfSet(capabilities, "file", File);
+            AddIfSet(capabilities, "default", Default);
+            return capabilities.Count == 0 ? null : capabilities;
```
RenderMichael: If we have a `PerPromptType` with no values set, I think we should still send an empty object `{ }` instead of `null`. We should not add this logic client-side.

dotnet/test/webdriver/AlertsTests.cs
```diff
@@ -34,6 +34,22 @@ public void ShouldBeAbleToOverrideTheWindowAlertMethod()
...
+        {
+            UnhandledPromptBehavior = new UserPromptHandler.PerPromptType { Default = UnhandledPromptBehavior.Dismiss }
+        });
+
+        driver.Url = CreateAlertPage("cheese");
+
+        driver.FindElement(By.Id("alert")).Click();
+
+        // If we can perform any action again, we're good to go
+        driver.FindElement(By.Id("alert")).Click();
```
RenderMichael: Could we `Assert.That(() => { ... }, Throws.Nothing, "Alert should have been auto-dismissed");` just so the assertion is explicit?

dotnet/src/webdriver/UserPromptHandler.cs
nvborisenko: Agree, `new object()` should be `{ }`

nvborisenko: Resolved.

dotnet/test/webdriver/AlertsTests.cs
nvborisenko: I found proper fixture which verifies unhandled prompts, moved test there.

dotnet/test/webdriver/UnexpectedAlertBehaviorTests.cs
```diff
@@ -71,56 +71,49 @@ public void CanSilentlyDismissUnhandledAlert()
...
+    private void ExecuteTestWithUnhandledPrompt(UserPromptHandler behavior, string expectedAlertText)
     {
-        bool silentlyHandlePrompt = behavior == UnhandledPromptBehavior.Accept || behavior == UnhandledPromptBehavior.Dismiss;
-        UnhandledPromptBehaviorOptions options = new UnhandledPromptBehaviorOptions();
-        if (behavior != UnhandledPromptBehavior.Default)
+        UnhandledPromptBehaviorOptions options = new()
         {
-            options.UnhandledPromptBehavior = behavior;
-        }
+            UnhandledPromptBehavior = behavior,
```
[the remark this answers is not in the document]
nvborisenko: non-nullable context

dotnet/src/webdriver/UserPromptHandler.cs
```diff
@@ -0,0 +1,195 @@
...
+        internal override object? ToCapabilities()
+        {
+            Dictionary<string, string> capabilities = [];
+            AddIfSet(capabilities, "alert", Alert);
+            AddIfSet(capabilities, "confirm", Confirm);
+            AddIfSet(capabilities, "prompt", Prompt);
+            AddIfSet(capabilities, "beforeUnload", BeforeUnload);
+            AddIfSet(capabilities, "file", File);
+            AddIfSet(capabilities, "default", Default);
+            return capabilities;
```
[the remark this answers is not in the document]
nvborisenko: Intentional.

dotnet/test/webdriver/UnexpectedAlertBehaviorTests.cs
```diff
@@ -71,56 +71,49 @@ public void CanSilentlyDismissUnhandledAlert()
...
+        };
 
         localDriver = EnvironmentManager.Instance.CreateDriverInstance(options);
         localDriver.Url = alertsPage;
         IWebElement resultElement = localDriver.FindElement(By.Id("text"));
         localDriver.FindElement(By.Id("prompt-with-default")).Click();
 
-        WaitFor(ElementTextToBeEqual(resultElement, expectedAlertText, silentlyHandlePrompt), "Did not find text");
+        WaitFor(() => resultElement.Text == expectedAlertText, "Did not find text");
     }
```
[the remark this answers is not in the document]
nvborisenko: Not in this PR.

RenderMichael (approved):

PR #16601, opened 2025-11-15
[dotnet] [bidi] Possibility to reset viewport

nvborisenko:
### **User description**
New pattern detected.

https://w3c.github.io/webdriver-bidi/#command-browsingContext-setViewport

```
browsingContext.SetViewportParameters = {
  ? context: browsingContext.BrowsingContext,
  ? viewport: browsingContext.Viewport / null,
  ? devicePixelRatio: float / null,
  ? userContexts: [+browser.UserContext],
}
```

Here `viewport` property is nullable. But if remote-end faces real `null` in `json`, then it behaves differently. So we just always send `null` if it is `null`.

### 🔧 Implementation Notes
Introducing special type `Optional<T>` with represents 3-states (has value, null, has no value):

### 🔄 Types of changes

___

### **PR Type**
Bug fix, Enhancement

___

### **Description**
- Add JsonIgnore attribute to Viewport property for null serialization

- Enable viewport reset capability via BiDi setViewport command

- Ensure null viewport values are properly sent to remote endpoint

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["SetViewportParameters"] -->|"Add JsonIgnore attribute"| B["Viewport property"]
  B -->|"WhenWritingNull condition"| C["Null values serialized correctly"]
  C -->|"Enables"| D["Viewport reset capability"]
```

<tr>
  <td>
    <details>
      <summary><strong>SetViewportCommand.cs</strong><dd><code>Add JsonIgnore attribute for null viewport serialization</code>&nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/BiDi/BrowsingContext/SetViewportCommand.cs

<ul><li>Added <code>System.Text.Json.Serialization</code> using statement<br> <li> Applied <code>[property: JsonIgnore(Condition = </code><br><code>JsonIgnoreCondition.WhenWritingNull)]</code> attribute to <code>Viewport?</code> property <br>in <code>SetViewportParameters</code> record<br> <li> Enables proper serialization of null viewport values to reset viewport <br>via BiDi protocol</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16601/files#diff-1646883d8a24b438f48ee1e9bb297a303054bb16790e3d4a4a8aef3870d15be6">+2/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___


dotnet/src/webdriver/BiDi/BrowsingContext/SetViewportCommand.cs
```diff
@@ -18,13 +18,14 @@
...
 using OpenQA.Selenium.BiDi.Communication;
+using System.Text.Json.Serialization;
 
 namespace OpenQA.Selenium.BiDi.BrowsingContext;
 
 internal sealed class SetViewportCommand(SetViewportParameters @params)
     : Command<SetViewportParameters, SetViewportResult>(@params, "browsingContext.setViewport");
 
-internal sealed record SetViewportParameters(BrowsingContext Context, Viewport? Viewport, double? DevicePixelRatio) : Parameters;
+internal sealed record SetViewportParameters(BrowsingContext Context, [property: JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)] Viewport? Viewport, double? DevicePixelRatio) : Parameters;
```
nvborisenko: Nice catch, we already started doing it in another places. Let's find all occurences of this type (`property: type / null`) and mark all with this attribute.

nvborisenko: Ah, interesting: `WhenWritingNull` or `WhenWritingDefault`. We want to send json `null` obly if user explicitly wrote `SomeNullableProperty = null`.

nvborisenko: Verified, in both cases `WhenWritingNull` and `WhenWritingDefault` it doesn't serializes to json `null`.

```csharp
await context.SetViewportAsync(new() { Viewport = null });
await context.SetViewportAsync();
```

```
    19:40:02.793 TRACE WebSocketTransport: BiDi SND --> {"id":3,"method":"browsingContext.setViewport","params":{"context":"D0E4F713FD009541B7FADB9D86BC5D6C"}}
19:40:02.795 TRACE WebSocketTransport: BiDi RCV <-- {"id":3,"result":{},"type":"success"}
19:40:02.796 TRACE WebSocketTransport: BiDi SND --> {"id":4,"method":"browsingContext.setViewport","params":{"context":"D0E4F713FD009541B7FADB9D86BC5D6C"}}
19:40:02.798 TRACE WebSocketTransport: BiDi RCV <-- {"id":4,"result":{},"type":"success"}
```

nvborisenko: LOL, we have  used `JsonIgnoreCondition.Never` for these cases :D

nvborisenko: Yes, it always sends `null`:
```
19:46:18.576 TRACE WebSocketTransport: BiDi SND --> {"id":4,"method":"browsingContext.setViewport","params":{"context":"EAEFDFDBD3A2386DD196A128BE21FF1E","viewport":null}}
```

But I am more interesting whether it's possibly to send `null` only and only when user explicitly set it to `null`. Is it poossible?

nvborisenko: Bad news: if we send `null` then browser resets viewport, but user didn't want to do it.

So we really want to understand how to:

> send null only and only when user explicitly set it to null

nvborisenko: Filed this interesting case to spec: https://github.com/w3c/webdriver-bidi/issues/1039

nvborisenko: So we should migrate from `[JsonIgnoreCondition.WhenWritingNull]` to custom `Optional<T>`

nvborisenko: @RenderMichael I tried to do something, kind of mess. Please review, I believe we can simplify somewhere.

nvborisenko: Resolving this thread, let's start new discussions.

dotnet/src/webdriver/BiDi/Optional.cs
```diff
@@ -0,0 +1,47 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+// </copyright>
+
+using System;
+
+namespace OpenQA.Selenium.BiDi;
+
+public readonly record struct Optional<T>
```
nvborisenko, line 24: `readonly struct`, do we see issues? There are limitations of its use. I propose to be safe and don't use it.

nvborisenko, line 24: In `emulation` module most of commands behaves the same: https://github.com/SeleniumHQ/selenium/blob/c064dcfaa47c292e06be1ace0b2bae929ee03c06/dotnet/src/webdriver/BiDi/Emulation/EmulationModule.cs#L26 where `timezone` is three-states value.

Does it mean we should use `Optional<T>` here as well? I think so.

nvborisenko: Or not...

Spec:
```
timezone: text / null
```

It is not optional.

nvborisenko: Seems it is another issue for BiDi spec. Before posing it  let's prove how it works.

RenderMichael: This type should definitely be a `struct`, otherwise usability tanks. It will create an additional state otherwise, where `value == null` is different from `value.HasValue == false`. In other words, `default(Optional<T>)` should be the same as `HasValue == false`

I think `T` should be constrained to `: class?`, in order to have clear usage compared to Nullable<T>.

nvborisenko: @RenderMichael I wanted to double-check how remote-end behaves. Is it really ``three-states value`?

RenderMichael: Your converter writes `null` when it has no value, so it might not display that behavior.

To test is quickly, you can add `[JsonIgnore(WhenWritingDefault)]` to the optional property. We can fresh out a general-case converter after that.

RenderMichael: Sorry, I posted this under the wrong comment.

Looks like property is not optional, so we don’t need it here.

nvborisenko: Ah, it is not ref struct. https://github.com/SeleniumHQ/selenium/pull/16601#discussion_r2561396609 - so resolved.

nvborisenko: Verified, I didn't sent argument:
```
BiDi SND --> {"id":2,"method":"emulation.setTimezoneOverride","params":{"contexts":["737CC07752DA1C5ED0128E60DF1DA3B7"]}}
BiDi RCV <-- {"error":"invalid argument","id":2,"message":"Invalid input in \"timezone\".","type":"error"}
```

So these arguments are really not optional.

nvborisenko: So many topics in this thread. Let's separate. I am resolving this thread.

nvborisenko, line 24: https://github.com/SeleniumHQ/selenium/pull/16601/files#r2561396609

> This type should definitely be a struct, otherwise usability tanks. It will create an additional state otherwise, where value == null is different from value.HasValue == false. In other words, default(Optional<T>) should be the same as HasValue == false

> I think T should be constrained to : class?, in order to have clear usage compared to Nullable.

Seems reasonable. It also means DTO cannot be `struct`? BTW `Nullable<double>` is `struct`. I think we really want to see `Optional<double?>`.

nvborisenko: Current behavior:

```csharp
await context.SetViewportAsync(new() { Viewport = default(Viewport?) }); // Explicitly sends "null", resetting to default
```

Seems strange?

RenderMichael: Good point, I am in agreement that `Optional<T>` should not be constrained.

nvborisenko: I guess we implemented required behavior. In any case this hidden feature will be documented. In future we can improve.

nvborisenko:
@RenderMichael I am OK with these changes, even with `Optional<Viewport?>?` (nullable of nullable). I would like to simplify this nullable of nullable somehow, but I don't see ways how it could be done.

PR #16420, opened 2025-10-13
[dotnet] Introduce global editroconfig

nvborisenko:
### **User description**
Code style is opiniated. We want to introduce some rules. But which rules - it is opiniated again.

### 💥 What does this PR do?
I took rules from dotnet itself: https://github.com/dotnet/runtime/blob/main/.editorconfig

### 🔧 Implementation Notes
And executed `dotnet format`

### 🔄 Types of changes

- Cleanup (formatting, renaming)

___

### **PR Type**
Other

___

### **Description**
- Introduce global `.editorconfig` with comprehensive C# formatting rules

- Reorder `using` directives to place `System` namespaces first

- Adjust code formatting: braces, spacing, and indentation per new rules

- Apply consistent code style across entire .NET codebase

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Add .editorconfig"] --> B["Define C# style rules"]
  B --> C["Run dotnet format"]
  C --> D["Reorder using statements"]
  C --> E["Adjust brace placement"]
  C --> F["Fix spacing/indentation"]
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Configuration changes</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>.editorconfig</strong><dd><code>Add comprehensive editorconfig with C# formatting rules</code>&nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16420/files#diff-cfb867d5de3aaa02df169c9665b9d05d4f96b045d92063289b1e1e9b2d88b2cf">+200/-1</a>&nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Formatting</strong></td><td><details><summary>18 files</summary><table>
<tr>
  <td><strong>ResponseValueJsonConverter.cs</strong><dd><code>Adjust brace placement in switch cases</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
[...]

PR #16603, opened 2025-11-16
[dotnet] [bidi] Fix events subscription possibilities

### LINKED ISSUE #16582: [🚀 Feature]: [dotnet] [bidi] Global subscription doesn't allow `context/UserContext`
### Description

```csharp
bidi.Network.OnBeforeRequestSent(...)
```

There is not support to specify `contexts`/`userContexts`.

https://w3c.github.io/webdriver-bidi/#cddl-type-sessionsubscriptionrequest

Previously we agreed low-level api should mimic spec, that's it.

_No response_

nvborisenko:
### 🔗 Related Issues

I am not able to scope global events to `context`. Fixes https://github.com/SeleniumHQ/selenium/issues/16582

```csharp
bidi.Network.OnBeforeRequestAsync(new SubscriptionOptions { Contexts = [ ...??? ]});
```

It is design issue. When I am in bidi global context, then I should be able to specify `Contexts` or `UserContexts`.

### 💥 What does this PR do?

Now it is clear. When I use `bidi` global object, then I can do everything as specification describes. But if I am in `browsingContext`, then events are automatically scoped to this BrowsingContext (implicitly provided when subscribing).

### 🔄 Types of changes


nvborisenko:
@RenderMichael @YevgeniyShunevych now public API for end-user looks clear, and provides all required capabilities.

But internally the implementation blows my mind. Can we optimize it somehow to avoid types messing?

dotnet/src/webdriver/BiDi/BrowsingContext/BrowsingContextNetworkModule.cs
```diff
@@ -17,9 +17,9 @@
 // under the License.
 // </copyright>
 
-using System.Threading.Tasks;
-using System;
 using OpenQA.Selenium.BiDi.Network;
+using System;
```
RenderMichael, line 21: There are some conventions on the `using` statements which say it should be in alphabetical order; there are others that say alphabetical *except* ones that start with `System`.

We should pick one and set it in the editorconfig, to avoid going back and forth forever.

dotnet/src/webdriver/BiDi/Subscription.cs
```diff
@@ -51,15 +51,23 @@ public async ValueTask DisposeAsync()
...
-    {
-        Timeout = options?.Timeout;
-    }
+    public TimeSpan? Timeout { get; set; }
+}
 
-    public IEnumerable<BrowsingContext.BrowsingContext>? Contexts { get; set; }
+internal static class ContextSubscriptionOptionsExtensions
+{
+    public static SubscriptionOptions WithContext(this ContextSubscriptionOptions? options, BrowsingContext.BrowsingContext context) => new()
```
RenderMichael, line 70: It is strange to have an extension method with a reciever that can be `null`. Do we ever have a null `ContextSubscriptionOptions options`?

RenderMichael (approved):
I like this approach.

dotnet/src/webdriver/BiDi/BrowsingContext/BrowsingContextNetworkModule.cs
nvborisenko: Current "convention" is alphabetical. I tried to apply editorconfig from dotnet/runtime: https://github.com/SeleniumHQ/selenium/pull/16420 - it is big change. Will be glad to return to this discussion later.

dotnet/src/webdriver/BiDi/Subscription.cs
nvborisenko: Yes, usually user may not provide non-null options to subscribe.

nvborisenko: Left a comment for clarity.

dotnet/src/webdriver/BiDi/BrowsingContext/BrowsingContextNetworkModule.cs
RenderMichael: Good idea, separate discussion. Just bringing it up because of the code change.

dotnet/src/webdriver/BiDi/Subscription.cs
RenderMichael: Since this method is not public API, it is not a big deal.

RenderMichael (approved):

PR #16623, opened 2025-11-21
[py][bidi] Allow resetting viewport

### LINKED ISSUE #16622: [🐛 Bug]: [py] Allow resetting viewport in BiDi
### Description

related to: #16601 and https://github.com/w3c/webdriver-bidi/issues/1039

see: https://w3c.github.io/webdriver-bidi/#command-browsingContext-setViewport

The BiDi spec allows you to reset the viewport size and device pixel ratio by sending `"viewport": null` or `"devicePixelRatio": null`. Currently, there is no way to do this in Python.

To handle this case, we should update the `set_viewport` method in `py/selenium/webdriver/common/bidi/browsing_context.py`

If the method is called with no `viewport` argument, the behavior should remain the same (don't set the `viewport` key). However, if it is called with viewport explicitly set to `None` (i.e. `set_viewport(viewport=None)` we should set the dictionary entry to `params["viewport"] = None`, so when it gets serialized, it sends `"viewport": null` in the JSON payload.

If the method is called with no `device_pixel_ratio` argument, the behavior should remain the same (don't set the `viewport` key). However, if it is called with device pixel ratio explicitly set to `None` (i.e. `set_viewport(device_pixel_ratio=None)` we should set the dictionary entry to `params["devicePixelRatio"] = None`, so when it gets serialized, it sends `"devicePixelRatio": null` in the JSON payload.

Note: make sure to document this behavior in the docstring and add tests.

### Reproducible Code

```shell
set_viewport(viewport=None, device_pixel_ratio=None)
```

cgoldberg:
### 🔗 Related Issues

Fixes: #16622
Also see: ##16601 and https://github.com/w3c/webdriver-bidi/issues/1039

### 💥 What does this PR do?

The BiDi spec allows you to reset the viewport size and device pixel ratio by sending `"viewport": null` or `"devicePixelRatio": null`:

https://w3c.github.io/webdriver-bidi/#command-browsingContext-setViewport

This PR accomplishes that by sending `"viewport": null` or `"devicePixelRatio": null`, when the arguments to `set_viewport()` are explicitly set to `None`.

For example:

```
set_viewport(viewport=None)
set_viewport(device_pixel_ratio=None)
set_viewport(viewport=None, device_pixel_ratio=None)
```

If either argument is not passed, the behavior remains the same as before (`viewport` and/or `devicePixelRatio` is not included in the JSON command payload).

This PR also includes a test for this and updates the other viewport tests to reset the viewport and device pixel ratio to default values in their teardown so they don't affect other tests.

### 🔧 Implementation Notes

To differentiate between being invoked with the arguments set to `None` and not being provided, I used a sentinel value as the default for the keyword args. Python doesn't have a built-in utility for defining sentinel values (it may someday, see PEP661: https://peps.python.org/pep-0661/), so rather than writing my own, I used 
the `Sentinel` class from `typing_extensions` (we already include this as a dependency): https://typing-extensions.readthedocs.io/en/latest/#sentinel-objects

### 🔄 Types of changes

___

### **PR Type**
Enhancement, Bug fix

___

### **Description**
- Allow resetting viewport and device pixel ratio to defaults via `None`

- Use sentinel value to differentiate explicit `None` from unspecified arguments

- Add comprehensive tests for viewport reset functionality

- Update existing tests with teardown to reset viewport state

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["set_viewport arguments"] -->|UNDEFINED sentinel| B["Skip parameter"]
  A -->|None value| C["Send null to BiDi"]
  A -->|dict/float value| D["Send value to BiDi"]
  C --> E["Reset to defaults"]
  D --> F["Apply custom settings"]
```

<tr>
  <td>
    <details>
      <summary><strong>browsing_context.py</strong><dd><code>Implement viewport reset with sentinel value pattern</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/selenium/webdriver/common/bidi/browsing_context.py

<ul><li>Import <code>Sentinel</code> from <code>typing_extensions</code> and define <code>UNDEFINED</code> sentinel <br>value<br> <li> Update <code>set_viewport()</code> method signature to use sentinel defaults for <br><code>viewport</code> and <code>device_pixel_ratio</code> parameters<br> <li> Implement conditional logic to differentiate between unspecified <br>arguments (UNDEFINED), explicit <code>None</code> (reset), and actual values<br> <li> Update docstrings to clarify that <code>None</code> resets parameters to defaults</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16623/files#diff-a65d02c951aeb477672aa52e5eb0106645ba869aa186c0af6d2672c42cac95c6">+19/-7</a>&nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>bidi_browsing_context_tests.py</strong><dd><code>Add viewport reset tests and cleanup test teardowns</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/test/selenium/webdriver/common/bidi_browsing_context_tests.py

<ul><li>Add try-finally blocks to existing viewport tests to reset viewport <br>state in teardown<br> <li> Update test values to be unique (251x301, 252x302) to distinguish test <br>cases<br> <li> Add new <code>test_set_viewport_back_to_default()</code> test to verify reset <br>functionality<br> <li> Verify that passing <code>None</code> restores default viewport and device pixel <br>ratio values</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16623/files#diff-cea00ba1c71fd6f4ca93d87f14bcc291e1b87383cd61cd1bf1eec678e8d9efbf">+43/-12</a>&nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/py/selenium/webdriver/common/bidi/browsing_context.py
+++ b/py/selenium/webdriver/common/bidi/browsing_context.py
@@ -1002,15 +1002,15 @@ class BrowsingContext:
             ValueError: If both `contexts` and `user_contexts` are used
         """
 
-        if contexts is not None and user_contexts is not None:
-            raise ValueError("Cannot specify both contexts and user_contexts")
+        if context is not None and user_contexts is not None:
+            raise ValueError("Cannot specify both context and user_contexts")
 
-        if contexts is None and user_contexts is None:
-            raise ValueError("Must specify either contexts or user_contexts")
+        if context is None and user_contexts is None:
+            raise ValueError("Must specify either context or user_contexts")
 
         params: dict[str, Any] = {}
-        if contexts is not None:
-            params["contexts"] = contexts
+        if context is not None:
+            params["context"] = context
         elif user_contexts is not None:
             params["userContexts"] = user_contexts
         if viewport is UNDEFINED:
--- a/py/selenium/webdriver/common/bidi/browsing_context.py
+++ b/py/selenium/webdriver/common/bidi/browsing_context.py
@@ -992,14 +992,14 @@ class BrowsingContext:
 
         Args:
             context: The browsing context ID.
-            viewport: The viewport parameters (`None` resets to default).
-            device_pixel_ratio: The device pixel ratio (`None` resets default).
+            viewport: The viewport parameters - {"width": <int>, "height": <int>} (`None` resets to default).
+            device_pixel_ratio: The device pixel ratio (`None` resets to default).
             user_contexts: The user context IDs.
 
         Raises:
             Exception: If the browsing context is not a top-level traversable
-            ValueError: If neither `contexts` or `user_contexts` are used
-            ValueError: If both `contexts` and `user_contexts` are used
+            ValueError: If neither `context` nor `user_contexts` is provided
+            ValueError: If both `context` and `user_contexts` are provided
         """
         if context is not None and user_contexts is not None:
             raise ValueError("Cannot specify both context and user_contexts")
--- a/py/selenium/webdriver/common/bidi/browsing_context.py
+++ b/py/selenium/webdriver/common/bidi/browsing_context.py
@@ -1012,17 +1012,9 @@ class BrowsingContext:
             params["context"] = context
         elif user_contexts is not None:
             params["userContexts"] = user_contexts
-        if viewport is UNDEFINED:
-            pass
-        elif viewport is None:
-            params["viewport"] = None
-        else:
+        if viewport is not UNDEFINED:
             params["viewport"] = viewport
-        if device_pixel_ratio is UNDEFINED:
-            pass
-        elif device_pixel_ratio is None:
-            params["devicePixelRatio"] = None
-        else:
+        if device_pixel_ratio is not UNDEFINED:
             params["devicePixelRatio"] = device_pixel_ratio
 
         self.conn.execute(command_builder("browsingContext.setViewport", params))
--- a/py/selenium/webdriver/common/bidi/browsing_context.py
+++ b/py/selenium/webdriver/common/bidi/browsing_context.py
@@ -20,8 +20,12 @@ from collections.abc import Callable
 from dataclasses import dataclass
 from typing import Any
 
+from typing_extensions import Sentinel
+
 from selenium.webdriver.common.bidi.common import command_builder
 from selenium.webdriver.common.bidi.session import Session
+
+UNDEFINED = Sentinel("UNDEFINED")
 
 
 class ReadinessState:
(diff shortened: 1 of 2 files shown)


cgoldberg:
I haven't tested this yet, so we'll see if browsers actually support it in CI.

This issue was discussed at length in Slack today, and there might be other BiDi functionality that needs to follow this same pattern when `null` is an accepted value in the BiDi spec.

This is kind of awkward in Python, but defaulting to a sentinel is the best approach I could think of.

navin772 (commented):
Should we also test a flow where we don't pass the `viewport` param (the undefined case)?

It could be: 
`set_viewport` with {width, height} -> assert viewport -> `set_viewport` with no viewport param(undefined) -> assert still same height, width -> `set_viewport` as None -> assert resets to default

We could add it to a previous test.

py/selenium/webdriver/common/bidi/browsing_context.py
```diff
@@ -998,9 +1002,17 @@ def set_viewport(
         params: dict[str, Any] = {}
         if context is not None:
```
navin772, line 1011: Also, could you update the validations for `contexts` and `userContexts`, as per the spec we cannot pass both, and need to pass atleast one of them.

So, update to something like this:
```python
if contexts is not None and user_contexts is not None:
    raise ValueError("Cannot specify both contexts and user_contexts")

if contexts is None and user_contexts is None:
    raise ValueError("Must specify either contexts or user_contexts")


if contexts is not None:
    params["contexts"] = contexts
elif user_contexts is not None:
    params["userContexts"] = user_contexts
``` 

This is not directly related to this PR, but lmk if I should update it in a diff PR once this is merged

cgoldberg: Thanks. I added a test for the undefined case, and added the validation you suggested for contexts/user_contexts.

navin772: Should be `context` instead of `contexts`

navin772: I updated in your branch

navin772:
@cgoldberg I think using `Ellipsis` here would be elegant than Sentinel and would not require the if-else checks and the new import, here is how we can use it:

```python
    def set_viewport(
      self,
      context: str | None = None,
      viewport: dict | None | type(Ellipsis) = ...,
      device_pixel_ratio: float | None | type(Ellipsis) = ...,
      user_contexts: list[str] | None = None,
    ) -> None:

        if context is not None and user_contexts is not None:
            raise ValueError("Cannot specify both context and user_contexts")

        if context is None and user_contexts is None:
            raise ValueError("Must specify either context or user_contexts")

        params: dict[str, Any] = {}

        if context is not None:
            params["context"] = context
        else:
            params["userContexts"] = user_contexts

        if viewport is not ...:
            params["viewport"] = viewport  # handles all 3 cases

        if device_pixel_ratio is not ...:
            params["devicePixelRatio"] = device_pixel_ratio

        self.conn.execute(command_builder("browsingContext.setViewport", params))
```

WDYT about this approach?

cgoldberg:
> I think using Ellipsis here would be elegant than Sentinel and would not require the if-else checks

That's interesting. Your code made me realize I can remove the if-else checks to simplify it even if I use the sentinel. I just updated the code.

I guess it just comes down to whether it's less confusing to use the Ellipses or the Sentinel.

They discuss it here:

https://peps.python.org/pep-0661/#use-the-existing-ellipsis-sentinel-value

I don't know what they mean by Ellipses "can’t be as confidently used in all cases".

shbenzer:
LGTM. I personally prefer Sentinel as “…” can represent a few things depending on the use case (e.g. in arrays) @navin772 @cgoldberg

shbenzer (approved):

navin772 (approved):
LGTM! Either of them looks good now.

PR #16734, opened 2025-12-15
Add guidance for testing conventions in each language

titusfortner:
### Reason for PR
Each language has its own tribal knowledge right now for how it works, from annotations to helper methods to common locations for things. This is intended to document those things to make it easier to work in each language.

### 💥 What does this PR do?
Removes language specific testing details from primary README and moves everything to separate `TESTING.md` files

This is in draft because they were auto-generated as a place to start. I updated some of them more than others, but they can all use a lot of work. Please assist, thanks.


README.md
Delta456, line 1: Hyperlinking to each binding readme would also be beneficial.

titusfortner: Yes, this is a good idea. I was going to do that, but our READMEs aren't consistent right now with what they do and don't contain, and I was thinking it might make sense to update them to mostly point to our website docs, and right now the coding agents we're targeting can't access the internet.

dotnet/TESTING.md
```diff
@@ -0,0 +1,134 @@
...
+    public void ShouldDoSomething()
+    {
+        // Skipped on Safari
+    }
+}
+```
+
+## Running Tests
+
+Bazel creates test targets for each browser. Always use `--pin_browsers`.
```
RenderMichael, line 35: This is relevant for CI, but for day to day inner-loop testing, I use the Visual Studio integration comes in through NUnit. Rider supports it too. It gives the same results, modulo any build differences that comes in when `bazel build` and `dotnet build` accidentally diverge.

RenderMichael: (comment aimed at the entire Running Tests section)

py/TESTING.md
```diff
@@ -0,0 +1,127 @@
...
+    ├── common/              # Cross-browser tests
+    ├── chrome/
+    ├── firefox/
+    ├── safari/
+    └── remote/
+```
+
+Test files end in `_tests.py` (e.g., `visibility_tests.py`).
+
+## Running Tests
```
navin772, line 41: We can also add section to run tests directly with pytest, and the flags we require.

I will add them.

diemol (approved):

bonigarcia (approved):

dotnet/TESTING.md
shs96c: I set `//common:pin_browsers` in my `.bazelrc.local` to ensure that I'm always using pinned browsers.

java/TESTING.md
```diff
@@ -0,0 +1,98 @@
...
+
+| Annotation | When to Use                                                                                                                                              |
+|------------|----------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `@NeedsFreshDriver` | Test needs clean browser state with default capabilities. Driver is restarted before the test.                                                           |
+| `@NoDriverBeforeTest` | Test needs custom capabilities or tests driver creation itself. Driver is destroyed, must use `createNewDriver(capabilities)` in the test to create one. |
+| `@NoDriverAfterTest` | Test leaves browser in a bad state. Driver is restarted after the test. Also accepts `failedOnly`.                                                       |
+
+For tests needing two browsers simultaneously (e.g., multi-user scenarios), create a second instance with `localDriver = new ChromeDriver()`. This driver is automatically quit after the test.
```
shs96c, line 82: Nope! This will (at best) create a Chrome instance instead of the browser being used for the tests. The only reason why the browser is closed after tests is because the entire process tree is killed. 

The correct way to start a new browser instance is via something like:

```java
RemoteWebDriver.builder().oneOf(Browser.detect()).build();
```

titusfortner: Thank you! This is exactly why we need this file. (this is much more helpful than the generic warning I made about being careful hard-coding creation of new drivers)

Also, if you name it `localDriver`, it is automatically checked in after hook: https://github.com/SeleniumHQ/selenium/blob/trunk/java/test/org/openqa/selenium/testing/JupiterTestBase.java#L90

dotnet/TESTING.md
titusfortner: @RenderMichael - yes, we should document working in IDEs without Bazel as well. We have a section in the current README for how to run the ruby code with RubyMine (by pointing to the ruby version generated by bazel). I agree Bazel isn't as good for interactively debugging code as VSCode/Rider is.

Bazel should be essentially calling the dotnet command under the hood, just in a hermetic way, so the results shouldn't diverge. Bazel is invaluable if you want to run tests or do builds without needing a local .NET dev environment. If all you're doing is local .NET development, I understand it can be an extra hurdle.

README.md
titusfortner: I added an issue to track this - https://github.com/SeleniumHQ/selenium/issues/16740

dotnet/TESTING.md
RenderMichael: Bazel is growing on me the more I understand it, it’s like Docker with a slightly different use case.

My question is, who is the target audience for each language’s respective TESTING.md file? If eg. this file is for contributors to the .NET bindings, the dev environment would be relevant. Otherwise, who would need to run a language binding’s tests? Someone who’s made changes to the grid? I feel like the 99% scenario is people who have a local env and want to test their changes.

titusfortner: The `dotnet/TESTING.md` is specifically for contributors to .NET bindings
Essentially what things would have made it easier for you to start contributing to this code base when you first pulled it up?
It should include: here are the conventions we use, here's how to use it with your IDE, here's how to verify it builds/tests with Bazel so there aren't problems with releases, etc.
You can probably assume contributors have a local dev environment for it, and if they don't, they'll know their way around the bazel pieces well enough anyway.

rb/TESTING.md
```diff
@@ -0,0 +1,121 @@
...
+        ├── bidi/
+        └── spec_support/      # Test helpers
+```
+
+Test files end in `_spec.rb` (e.g., `driver_spec.rb`).
+
+## Build Files
+
+* Adding tests shouldn't require Bazel changes—`rb_integration_test` uses glob patterns.
+* Make sure `*_spec.rb` files are in a directory with a `BUILD.bazel` containing `rb_integration_test`.
```
aguspe, line 121: I think we should add a section on how to use environment variables especially for BiDi testing @titusfortner :


## Environment Variables

Environment variables control test execution behavior and enable specific features.

### BiDi Testing

To run tests with BiDi (Bidirectional) protocol enabled:

```shell
# Enable BiDi for all tests
WD_REMOTE_BROWSER=chrome BIDI=true bazel test //rb/spec/integration/...

# Run BiDi-specific tests
bazel test //rb/spec/integration/selenium/webdriver/bidi/...
```

### Available Variables

| Variable | Purpose | Values | Example |
|----------|---------|--------|---------|
| `BIDI` | Enable BiDi protocol | `true`, `false` | `BIDI=true` |
| `WD_REMOTE_BROWSER` | Specify browser for remote tests | `chrome`, `firefox`, `edge`, `safari` | `WD_REMOTE_BROWSER=firefox` |
| `HEADLESS` | Run tests in headless mode | `true`, `false` | `HEADLESS=true` |
| `DEBUG` | Enable debug logging | `true`, `false` | `DEBUG=true` |

### Examples

```shell
# Run Chrome tests with BiDi enabled
BIDI=true bazel test //rb/spec/integration/... --test_tag_filters=chrome

# Run headless Firefox tests
HEADLESS=true bazel test //rb/spec/integration/... --test_tag_filters=firefox

# Run remote tests on Edge with BiDi
WD_REMOTE_BROWSER=edge BIDI=true bazel test //rb/spec/integration/... --test_tag_filters=remote

# Combine multiple variables
BIDI=true HEADLESS=true DEBUG=true bazel test //rb/spec/integration/selenium/webdriver/bidi/...
```

### Testing Guard Behavior

Environment variables interact with test guards. For example:

```ruby
# This test only runs when BiDi is enabled
it 'uses BiDi feature', only: {bidi: true} do
  # Test code
end

# This test is excluded when BiDi is enabled
it 'classic WebDriver only', exclusive: {bidi: false} do
  # Test code
end
```

Run with `BIDI=true` to see these guards in action.

dotnet/TESTING.md
nvborisenko: Looks like we are writing this on stones... Meaning it will be harder to "improve" it: https://github.com/SeleniumHQ/selenium/issues/15536

Final word: I cloned https://github.com/microsoft/playwright-dotnet, opened in any IDE, and successfully executed ALL tests without any instructions.

cgoldberg (approved):

dotnet/TESTING.md
titusfortner: @nvborisenko this is guidance for writing tests not running them.

navin772 (approved):

dotnet/TESTING.md
RenderMichael: > I cloned https://github.com/microsoft/playwright-dotnet

@nvborisenko Did you also clone https://github.com/microsoft/playwright and build the bits locally? Otherwise, you are using the pre-bundled Playwright bits and only the .NET side gets compiled with minimal cross-platform aspect to the build .

nvborisenko: > > I cloned https://github.com/microsoft/playwright-dotnet
> 
> @nvborisenko Did you also clone https://github.com/microsoft/playwright and build the bits locally? Otherwise, you are using the pre-bundled Playwright bits and only the .NET side gets compiled with minimal cross-platform aspect to the build .

Ideal!

Anyway, if it helps humans to be involved in - please move forward (**human**).

rb/TESTING.md
titusfortner: Please add away, you should be able to commit to this PR directly. Everything here - https://github.com/SeleniumHQ/selenium/blob/trunk/README.md#ruby-1 should probably be included. 
Thanks!

dotnet/TESTING.md
```diff
@@ -0,0 +1,134 @@
...
+bazel test //dotnet/test/common:AllTests --pin_browsers=true  # Default browser (Firefox)
+bazel test //dotnet/test/common:AllTests-chrome --pin_browsers=true
+bazel test //dotnet/test/common:AllTests-firefox --pin_browsers=true
+bazel test //dotnet/test/common:AllTests-edge --pin_browsers=true
+
+# Additional Arguments
+bazel test //dotnet/... --flaky_test_attempts=3 --pin_browsers=true
+bazel test //dotnet/... --test_output=all --pin_browsers=true
+```
+
```
RenderMichael, line 47: A few lines about running tests in the IDE.

```suggestion

### IDE Support

The test infrastructure supports Visual Studio's and Rider's IDE, which can be more convenient in the inner loop when developing features and writing tests. Simply open the solution after building with `bazel` and the IDE should work like normal.

Note that any bugs involving a discrepancy between the `.csproj` and `BUILD.bazel` files would manifest in running the tests through the IDE rather than with `bazel test`; any changes that affect either build systems should be tested with `bazel test` to ensure proper functionality.
```

dotnet/TESTING.md
RenderMichael: My 2 cents are covered in https://github.com/SeleniumHQ/selenium/pull/16734/files#r2637559076. Resolving this thread unless @nvborisenko has further concerns here.

cgoldberg:
@titusfortner I just updated this branch and fixed the merge conflicts. Is there any more work that needs to be done, or is this ready to merge?

dotnet/TESTING.md
```diff
@@ -0,0 +1,134 @@
...
+    }
+}
+```
+
+## Running Tests
+
+Bazel creates test targets for each browser. Always use `--pin_browsers`.
+
+```shell
+bazel test //dotnet/test/common:AllTests --pin_browsers=true  # Default browser (Firefox)
```
nvborisenko: Recently we removed `common`

dotnet/TESTING.md
```diff
@@ -0,0 +1,134 @@
...
+}
+```
+
+Browser values: `Browser.Chrome`, `Browser.Firefox`, `Browser.Edge`, `Browser.Safari`, `Browser.IE`, `Browser.Remote`, `Browser.All`
+
+### Driver Lifecycle
+
+| Attribute | When to Use |
+|-----------|-------------|
+| `[NeedsFreshDriver(IsCreatedBeforeTest = true)]` | Fresh driver before test |
```
nvborisenko, line 95: I am going to improve it in https://github.com/SeleniumHQ/selenium/issues/15536

diemol:
Do we need to update this, or should we discard it? All this work and time need to have a conclusion eventually.

aguspe:
> Do we need to update this, or should we discard it? All this work and time need to have a conclusion eventually.

  ---
 Hey @diemol, I think we should update this, I see value on the individual testing.md files and on making a lightweight README because usually when you want to collaborate with the project you stay in one binding, the rb one looks pretty good but if there is anything else I can do to help let me know

diemol:
@aguspe thanks, that makes sense. Why don't we split this PR per language then? I dislike it when we have PRs that try to cover everything across all bindings, and they cannot land until a large group of people agrees.

aguspe:
> @aguspe thanks, that makes sense. Why don't we split this PR per language then? I dislike it when we have PRs that try to cover everything across all bindings, and they cannot land until a large group of people agrees. 

That makes a lot of sense to me, I could split the rb one from here if everyone is onboard

titusfortner:
@aguspe & @diemol I did a refresh of these addressing the comments above.

@nvborisenko this isn't designed as something to be set in stone, it is for committers to better understand the "tribal knowledge" in bindings they aren't actively familiar with. Useful for both devs and agents. Yes it can diverge, but I think maintenance costs for this is better than nothing. Also, we already have a lot of this documented in README right now, and part of the intent is to pull it out and put it where it belongs.

We can split it out or merge if it is good enough. I think this is better than what we have for sure.

diemol:
Thanks, @titusfortner! 

Why don't we merge it now, and if a modification or update is needed, we create a new PR?

dotnet/TESTING.md
titusfortner: @RenderMichael please update anything you want to see improved after this gets merged, this is intended as a baseline and I'm not going to have everything correct in it.

PR #16764, opened 2025-12-20
[py] load bazel files first for local dev work

titusfortner:
### **User description**
The goal is to be able to run normal pytest in your IDE of choice and have it "just work"
But you should also be able to run bazel commands directly.
The current implementation causes problems with having two sources of truth, this implementation should let you switch between debugging with PyTest and a local Python environment and running bazel commands without errors.

___

### **PR Type**
Enhancement

___

### **Description**
- Dynamically load Bazel-generated files for local development

- Append Bazel package path to selenium module path

- Simplify local dev setup by removing manual file copying

- Allow seamless switching between pytest and bazel workflows

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["conftest.py imports selenium"] --> B["Check for bazel-bin/py/selenium"]
  B --> C["Append to selenium.__path__"]
  C --> D["Local sources override Bazel files"]
  E["Rakefile local_dev task"] --> F["Remove manual file copying"]
  F --> G["Simplified setup process"]
```

<tr>
  <td>
    <details>
      <summary><strong>conftest.py</strong><dd><code>Add dynamic Bazel package path resolution</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/conftest.py

<ul><li>Import selenium module at top level<br> <li> Add logic to detect and append Bazel-generated package path to <br><code>selenium.__path__</code><br> <li> Allow local sources to override Bazel files when both exist<br> <li> Enable fallback to Bazel-generated files when local sources <br>unavailable</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16764/files#diff-98e9e290ede5d0c7ac9e7df5b49edf63f5df5fcc946b7736e4a6139e4cda26e3">+7/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>Rakefile</strong><dd><code>Simplify local dev task setup</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

Rakefile

<ul><li>Remove manual file copying from <code>bazel-bin/py/selenium/webdriver</code> to <br><code>py/selenium/webdriver</code><br> <li> Remove devtools directory cleanup step<br> <li> Simplify local_dev task by relying on dynamic path loading</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16764/files#diff-ee98e028c59b193d58fde56ab4daf54d43c486ae674e63d50ddf300b07943e0f">+0/-3</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -579,9 +579,6 @@ namespace :py do
   task :local_dev do
     Bazel.execute('build', [], '//py:selenium')
     Rake::Task['grid'].invoke
-
-    FileUtils.rm_rf('py/selenium/webdriver/common/devtools/')
-    FileUtils.cp_r('bazel-bin/py/selenium/webdriver/.', 'py/selenium/webdriver', remove_destination: true)
   end
 
   desc 'Update generated Python files for local development'
--- a/py/conftest.py
+++ b/py/conftest.py
@@ -22,11 +22,18 @@ from pathlib import Path
 
 import pytest
 
+import selenium
 from selenium import webdriver
 from selenium.common.exceptions import WebDriverException
 from selenium.webdriver.remote.server import Server
 from test.selenium.webdriver.common.network import get_lan_ip
 from test.selenium.webdriver.common.webserver import SimpleWebServer
+
+# Allow local sources to override, and fall back to Bazel-generated files when present.
+_repo_root = Path(__file__).resolve().parents[1]
+_bazel_pkg = _repo_root / "bazel-bin" / "py" / "selenium"
+if _bazel_pkg.is_dir() and str(_bazel_pkg) not in selenium.__path__:
+    selenium.__path__.append(str(_bazel_pkg))
 
 drivers = (
     "chrome",


cgoldberg (commented):
The problem with this change is that it no longer copies the bazel-generated files (like CDP files and selenium manager) into the expected paths in the local source tree. Right now, I can run the `local_dev` task, and then I have a working selenium module I can import. I very often write small scripts that use the local module or open a REPL to test changes. This breaks that ability and I would have to do something similar to what you added to `./conftest.py` so it knows to look at the `bazel-bin` directory for the generated files.

I don't understand why the current setup doesn't work? (I run both pytest and bazel commands). Is there a way to fix whatever problem you are having, while still copying the bazel-generated files so it works for normal package use and not just with pytest or bazel?

titusfortner:
For Ruby I used to copy files like Python was doing. Then we made the equivalent change and it was much easier (it's not like you are going to edit the generated files, you just need them there to be referenced). Does Python not actually reference all the files correctly with this code?

cgoldberg:
> Does Python not actually reference all the files correctly with this code

I think it will when you run tests because of the change you made to `selenium.__path__` in `conftest.py`, but if I just write a script or open a REPL and `import selenium`, it won't be able to find any of the generated files because they only exist under `bazel-bin/`.

How about using symlinks instead of copying the files or messing with the module `__path__`? I think if you remove the changes you made to `conftest.py`, and have the `local_dev` task add symlinks, it would work. That way, we would have a single source of truth, but the references to the files would exist once you run the `local_dev` Rake task.

I think you would just need add the Bazel equivalent of this to `local_dev`:

```
ln -s bazel-bin/py/selenium/webdriver/devtools py/selenium/webdriver/devtools
ln -s bazel-bin/py/selenium/webdriver/linux py/selenium/webdriver/linux
ln -s bazel-bin/py/selenium/webdriver/macos py/selenium/webdriver/macos
ln -s bazel-bin/py/selenium/webdriver/windows py/selenium/webdriver/windows
```

titusfortner:
I ran the local_dev command, then tried to run a bazel test and got a bunch of errors that went away once I deleted the copied directory, so I thought we needed to find a way to get both to work together by looking at the same spot.

If this doesn't address your use case, then I'll close, I thought it would be an obvious improvement over just copying files. There are more things to symlink than just that, and it assumes we know ahead of time what all needs to be added. I'll explore more next time I dig into Python I guess.

PR #16771, opened 2025-12-21
Fix bug in FileHandler: it always failed on MacOS

asolntsev:
### **User description**
The `alternatives` list is immutable, so trying to add `"mac/" + name` to it always failed.

P.S. I suspect this class is not used anymore, and should be removed.

### 🔄 Types of changes

___

### **PR Type**
Bug fix

___

### **Description**
- Fix immutable list error in FileHandler on macOS

- Convert immutable Arrays.asList to mutable ArrayList

- Simplify imports and use static constants

- Enable dynamic addition of macOS-specific resource paths

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Immutable Arrays.asList"] -->|"Replace with"| B["Mutable ArrayList"]
  B -->|"Enables"| C["Add macOS paths dynamically"]
  D["Locale.ENGLISH"] -->|"Replace with"| E["Locale.ROOT"]
  F["Objects.requireNonNull"] -->|"Replace with"| G["Static import requireNonNull"]
```

<tr>
  <td>
    <details>
      <summary><strong>FileHandler.java</strong><dd><code>Fix macOS resource path handling with mutable list</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

java/src/org/openqa/selenium/io/FileHandler.java

<ul><li>Changed <code>alternatives</code> from immutable <code>Arrays.asList()</code> to mutable <br><code>ArrayList</code> to allow adding macOS-specific paths<br> <li> Replaced <code>Locale.ENGLISH</code> with static import of <code>Locale.ROOT</code> for <br>consistency<br> <li> Replaced <code>Objects.requireNonNull()</code> with static import for cleaner code<br> <li> Consolidated wildcard import for <code>java.util.*</code> to reduce import clutter</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16771/files#diff-6b9adce3a2137262e499f93062848cf9fb1962a076dc153500ecebcb33a132bf">+8/-8</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/java/src/org/openqa/selenium/io/FileHandler.java
+++ b/java/src/org/openqa/selenium/io/FileHandler.java
@@ -17,16 +17,16 @@
 
 package org.openqa.selenium.io;
 
+import static java.util.Locale.ROOT;
+import static java.util.Objects.requireNonNull;
+
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.file.Files;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Locale;
-import java.util.Objects;
+import java.util.*;
 
 /** Utility methods for common filesystem activities */
 public class FileHandler {
@@ -42,10 +42,10 @@ public class FileHandler {
 
   private static InputStream locateResource(Class<?> forClassLoader, String name)
       throws IOException {
-    String arch =
-        Objects.requireNonNull(System.getProperty("os.arch")).toLowerCase(Locale.ENGLISH) + "/";
-    List<String> alternatives = Arrays.asList(name, "/" + name, arch + name, "/" + arch + name);
-    if (System.getProperty("os.name").toLowerCase(Locale.ENGLISH).contains("mac")) {
+    String arch = requireNonNull(System.getProperty("os.arch")).toLowerCase(ROOT) + "/";
+    List<String> alternatives =
+        new ArrayList<>(List.of(name, "/" + name, arch + name, "/" + arch + name));
+    if (System.getProperty("os.name").toLowerCase(ROOT).contains("mac")) {
       alternatives.add("mac/" + name);
       alternatives.add("/mac/" + name);
     }


cgoldberg:
@asolntsev Sorry to be a pest, but can you start adding "[java]" to the beginning of your PR titles, so they appear in the commit messages? It also ensures CI for Java gets run:
 https://github.com/SeleniumHQ/selenium/blob/ac59dd85c34ec1490a0d4ca5449e9106448a5238/.github/workflows/ci.yml#L69-L70

asolntsev:
No problems,  I will add [java] to the title.

But I hope this system doesn't hope only on the title, and there is some robust detection of java PRs.

Maybe this one?

```
contains(needs.check.outputs.targets, '//java') ||
```

cgoldberg:
> I hope this system doesn't hope only on the title, and there is some robust detection of java PRs.

It also checks for changes in the bazel `//java` target, so if you make any changes under the `./java/`, directory it will run.

titusfortner:
Can you add unit tests for this class? I'm planning to add running unit tests in windows & mac in CI-Java.
They are public, and low impact so I don't think they need to be removed right now. But unit tests would have caught this issue. 🙁

asolntsev:
> It also checks for changes in the bazel `//java` target, so if you make any changes under the `./java/`, directory it will run.

Great!
Then I don't understand why these lines are needed at all?
```
       contains(join(github.event.commits.*.message), '[java]') || 
       contains(github.event.pull_request.title, '[java]') 
```

Why do we need this additional complexity?
Why it's not enough to just check "if there were changes in Java files"?  @cgoldberg

asolntsev:
> Can you add unit tests for this class?

Sure, I can add unit-tests.
I was thinking about it. But it seemed weird to write tests for the code that I didn't write, and that is not used in Selenium project. Isn't it more reasonable to mark it as `@Deprecated` and avoid wasting time on it? @titusfortner

cgoldberg:
> Why do we need this additional complexity?
> Why it's not enough to just check "if there were changes in Java files"?

You could make changes outside of the Java tree that might affect Java (like a CI workflow or something under ./common) and trigger Java tests by tagging your PR "[java]". I don't think it's extremely useful, but it really doesn't add any additional complexity.

titusfortner:
Well, I found out today that `contains(join(github.event.commits.*.message), '[java]')` doesn't work in PRs, because I did make a change in a workflow that can affect languages that aren't being tested. 🤷

diemol:
They used to work. Those were created around 3 years ago. Probably GitHub made changes.

PR #16775, opened 2025-12-22
[dotnet] test Selenium Manager

titusfortner:
### 🔗 Related Issues

.NET equivalent of #16756 

### 💥 What does this PR do?

* Testing selenium manager in all three OS with .NET
* Adding shared cache per OS for github
* Not sure why the integration test suite was changed to "small" putting it back to large

### 🔧 Implementation Notes

This could be split out in each of the browser directories instead of all handled in one test

___

### **PR Type**
Tests, Enhancement

___

### **Description**
- Add Selenium Manager tests for .NET across Windows, macOS, and Ubuntu

- Verify driver and browser path discovery via DriverFinder

- Validate driver service startup with discovered paths

- Configure CI workflow with OS-specific caching and test matrix

- Skip Safari and IE browsers as they don't require Selenium Manager

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["SeleniumManagerTest.cs"] -->|Tests| B["DriverFinder paths"]
  A -->|Tests| C["DriverService startup"]
  D["ci-dotnet.yml"] -->|Adds| E["Manager Tests job"]
  E -->|Runs on| F["Windows, macOS, Ubuntu"]
  E -->|Tests| G["Chrome, Firefox, Edge"]
  H["BUILD.bazel"] -->|Updates| I["Test suite size"]
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Tests</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>SeleniumManagerTest.cs</strong><dd><code>New Selenium Manager test suite for .NET</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/test/common/SeleniumManagerTest.cs

<ul><li>New test class with 97 lines testing Selenium Manager functionality<br> <li> Tests <code>DriverFinder.GetDriverPath()</code> and <code>GetBrowserPath()</code> methods<br> <li> Validates driver and browser paths exist and contain cache directory<br> <li> Tests <code>DriverService.Start()</code> with discovered driver paths<br> <li> Supports Chrome, Firefox, and Edge browsers with factory methods<br> <li> Skips IE, Safari, and Remote browsers via <code>IgnoreBrowser</code> attributes</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16775/files#diff-22f0b3f5b49032bf49ed44835bd02268e3eff2c82e3f06e7efd8b74a8df3d858">+97/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>B-build</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>ci-dotnet.yml</strong><dd><code>Add CI workflow for Selenium Manager tests</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

.github/workflows/ci-dotnet.yml

<ul><li>Add new <code>manager-tests</code> job running on Windows, macOS, and Ubuntu<br> <li> Configure test matrix with OS-specific cache keys<br> <li> Run Selenium Manager tests for Chrome, Firefox, and Edge browsers<br> <li> Set test environment variables for browser download and driver path <br>handling<br> <li> Update build job with matrix strategy for Windows and macOS<br> <li> Simplify integration tests job by removing 8dot3name configuration<br> <li> Change cache-key strategy from <code>false</code> to OS-specific keys</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16775/files#diff-9f761b9e8edbe9c2db37b8ef3863ecc6318bc1930386c72585733ae09ae9016f">+38/-7</a>&nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>BUILD.bazel</strong><dd><code>Update test suite configuration</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/test/common/BUILD.bazel

<ul><li>Change AllTests suite size from "small" to "large"<br> <li> Remove commented-out copy_file rule for manager-macos</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16775/files#diff-42c4bbda3788fb91b43ddd7ed9dce8876478e3d5dac2eb88274d6bdc184ff625">+1/-7</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___


RenderMichael (approved):

dotnet/test/common/SeleniumManagerTest.cs
```diff
@@ -0,0 +1,97 @@
...
+        var options = CreateOptionsForCurrentBrowser();
+        var driverFinder = new DriverFinder(options);
+
+        string driverPath = driverFinder.GetDriverPath();
+        string browserPath = driverFinder.GetBrowserPath();
+
+        Assert.That(File.Exists(driverPath), Is.True, $"Driver path should exist: {driverPath}");
+        Assert.That(File.Exists(browserPath), Is.True, $"Browser path should exist: {browserPath}");
+        Assert.That(driverPath, Does.Contain(CacheDirectory), $"Driver path should contain cache directory: {driverPath}");
+        Assert.That(browserPath, Does.Contain(CacheDirectory), $"Browser path should contain cache directory: {browserPath}");
```
RenderMichael: Seems like a product bug

```c#
Failed : OpenQA.Selenium.SeleniumManagerTest.ShouldGetDriverAndBrowserPaths
  Browser path should contain cache directory: C:\Program Files (x86)\Microsoft\Edge\Application\143.0.3650.96\msedge.exe
  Expected: String containing ".cache\selenium"
  But was:  "C:\Program Files (x86)\Microsoft\Edge\Application\143.0.3650.96\msedge.exe"
   at OpenQA.Selenium.SeleniumManagerTest.ShouldGetDriverAndBrowserPaths() in .\dotnet\test\common\SeleniumManagerTest.cs:line 74
```

titusfortner: Well, the fact that it is looking for .cache\selenium instead of .cache/selenium needs to be fixed, but even with that...

I wonder if this means that .NET bazel doesn't accept --test_env values, I'll investigate.

RenderMichael: NUnit has the `Is.SubPathOf` constraint that should handle all the slashes for us.

titusfortner: @RenderMichael can you put a code suggestion comment on this? https://github.com/SeleniumHQ/selenium/pull/16775/changes#diff-22f0b3f5b49032bf49ed44835bd02268e3eff2c82e3f06e7efd8b74a8df3d858R38

(or is it more complicated)

RenderMichael: The parent directory seems correct, we would just need to add a different assertion constraint. 

```suggestion
        Assert.That(browserPath, Is.SubPathOf(CacheDirectory), $"Browser path should be nested inside the cache directory: {browserPath}");
```

dotnet/test/common/SeleniumManagerTest.cs
```diff
@@ -0,0 +1,97 @@
...
+    {
+        var options = CreateOptionsForCurrentBrowser();
+        var driverFinder = new DriverFinder(options);
+
+        string driverPath = driverFinder.GetDriverPath();
+        string browserPath = driverFinder.GetBrowserPath();
+
+        Assert.That(File.Exists(driverPath), Is.True, $"Driver path should exist: {driverPath}");
+        Assert.That(File.Exists(browserPath), Is.True, $"Browser path should exist: {browserPath}");
+        Assert.That(driverPath, Does.Contain(CacheDirectory), $"Driver path should contain cache directory: {driverPath}");
```
RenderMichael: Same as below assert

```suggestion
        Assert.That(driverPath, Is.SubPathOf(CacheDirectory), $"Driver path should be nested under the cache directory: {driverPath}");
```

dotnet/test/common/SeleniumManagerTest.cs
titusfortner: Hmm, it seems to be pulling it properly in Mac at least.
@bonigarcia does SM print out in debug or trace which environment variables/config settings it is using?

@RenderMichael can you pull this branch locally and see if the values work for you on windows? (I need to create a Windows VM)

```
bazel test //dotnet/test/common:SeleniumManagerTest-chrome --test_env=SE_FORCE_BROWSER_DOWNLOAD=true --test_env=SE_SKIP_DRIVER_IN_PATH=true
```

bonigarcia: As a general rule, SM does not log the configuration values overridden with envs/config. But the debug/trace logs should be descriptive enough to follow what SM is doing.

titusfortner: Would it make sense to log those, though? Perhaps someone added a config file and someone else can't figure out why it isn't behaving the way they think it should, etc?

bonigarcia: I believe the existing logs should be already descriptive enough. But if there is some specific configuration which is not clear, we can always include new logs

titusfortner: I figured out what the issue was for this at least.

nvborisenko:
In my opinion this is bad precedent. What we test... We should test that binding "invokes external dependency with proper parameters" and properly parse response. That's it.

Testing... how to test... i don't see real unit tests. No clue.

titusfortner:
Well, needing unit tests in dotnet is a completely separate issue. 😂

Even then, this should be part of a slightly bigger change, and applied across the bindings.

titusfortner:
But also, something is weird on Windows... SM is not seeing the environment variables I'm passing in. Not sure if it is how the test is wired up, or an issue with how .NET is calling it. Once I figure out why, then we'll know if a unit test would have caught it. 😄

nvborisenko:
> Even then, this should be part of a slightly bigger change, and applied across the bindings.

I don't mind having one integration test to invoke SM executing on GHA. +1

dotnet/test/common/SeleniumManagerTest.cs
```diff
@@ -0,0 +1,119 @@
...
+        try
+        {
+            service.Start();
+            Assert.That(service.ServiceUrl, Is.Not.Null);
+            Assert.That(service.ServiceUrl.IsAbsoluteUri, Is.True);
+        }
+        finally
+        {
+            service.Dispose();
+        }
```
nvborisenko, line 117: `using` statement

dotnet/test/common/SeleniumManagerTest.cs
```diff
@@ -0,0 +1,119 @@
...
+        if (!string.IsNullOrWhiteSpace(cacheDirectory))
+        {
+            return cacheDirectory;
+        }
+
+        string homeDirectory = System.Environment.GetFolderPath(System.Environment.SpecialFolder.UserProfile);
+        if (!string.IsNullOrWhiteSpace(homeDirectory))
+        {
+            return Path.Combine(homeDirectory, ".cache", "selenium");
+        }
```
nvborisenko, line 53: This is responsibility of SM, dotnet should not know about this folder.

titusfortner: This is a way for us to see that it has in fact been downloaded and not just using the current installation. The contract is SE_CACHE_PATH which we check first. This is currently failing in windows, and again, depending on why would validate the need for this test. 😄

We could also set SE_CACHE_PATH as part of the test which would exercise the contract that way. That might be the right way to do this.

Issue #16809, opened 2025-12-29
[🚀 Feature]: Selenium Testing Plan


titusfortner:
### Problems
- Many things tested on RBE are being duplicated in GitHub Actions 
- Our **GHA logic is too coarse**: check-bazel-targets code often triggers full binding CI workflow even when only a subset of targets (e.g., `*-remote`) is affected.
- Selenium Manager is not currently tested across bindings (can't be run on RBE)
- Inconsistent GHA testing across bindings (Java, .NET & JS have almost nothing, while Ruby's take ~6 hours of runtime)
- Also inconsistent RBE tests in general (beta versions of browsers, testing against grid, etc)
- Pinned Browser Cache on GitHub Actions is prohibitive

### Goals
- Keep **per-PR signal fast** and scoped to the change.
- Ensure **consistent target coverage across bindings** (same categories exist everywhere even if implementations differ).
- Add **upstream browser regression detection** (betas).
- Keep results **reported by binding** (maintainer ownership).

### Per PR (always filtered by affected targets)
1. Linux (RBE)
   - Standardize (or add missing) targets per binding: Beta versions, Remote execution
   - Track skip-rbe and .skipped-tests
2. Windows (GHA) - filter for chrome/firefox, use Selenium Manager, include unit tests (Ruby/Python/JS)
3. Mac (GHA) - filter for safari, use Selenium Manager, include unit tests (Ruby/Python/JS)
4. Linux (GHA) - Alternate Interpreter unit tests (Ruby/Python/JS) & Selenium Manager tests

Note: if no affected targets found, but the binding is indicated in title or commit, run all targets for that binding

### Scheduled (GHA only)
- Windows: All targets, no browser filters
- macOS: filter for safari, chrome
- Linux: Alternate Interpreter unit tests (Ruby/Python/JS) & Selenium Manager

### Have you considered any alternatives or workarounds?

1. We could just add "remote tests" job for each binding and try to run remote vs local depending on what the check targets finds, but that adds more complexity
2. We could increase coverage to match what Ruby currently does on all the bindings, but that'll take forever.
3. We could run browser tests with pinned browsers, and run Selenium Manager tests separately, but the cache limitations, and might just be duplication.

diemol:
The initial implementation of the Bazel targets checks file names from the diff and, based on that, queries Bazel to retrieve the list of targets to run. I don't remember why we changed this. I guess we assumed the cache was going to help us, but now that the size is restricted, we have trouble.

[...]

PR #16817, opened 2025-12-30
[java] update what tests are run on Github

titusfortner:
### 🔗 Related Issues

Java implementation of https://github.com/SeleniumHQ/selenium/issues/16809

We are mostly relying on RBE to tell us if there is a problem, this is supposed to be extra information that strikes a good balance between information and execution time.

### 💥 What does this PR do?

* `ci.yml` kicks off on every commit. It runs the new and improved `check-bazel-targets.sh`
* Instead of just checking to see if there are any `java` targets and running the whole ci-java workflow, the unique set of applicable test targets are passed from the script to `ci-java.yml` and browser tests are only run on those
* Instead of hard coded targets in the yml file, this is relying on tags to determine what to run. So, added a "smoke" tag to run the things. 

### 🔧 Implementation Notes

Unlike Ruby/Python/JS this is not running a full test suite in Windows, just smoke tests; keeps it comparable to what we're doing currently, we can re-evaluation.
Added`ElementFindingTests` to Smoke, might want something more/different that exercises more windows-specific-concerning behavior?

Need to figure out how to test what is in skipped-tests, or why things are in skipped-tests

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -83,6 +83,7 @@ jobs:
       TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
       TWINE_USERNAME: ${{ secrets.TWINE_USERNAME }}
       NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      SE_CACHE_PATH: $RUNNER_TEMP/selenium
     steps:
       - name: Checkout source tree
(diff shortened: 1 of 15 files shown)


diemol (commented):
Can you update the PR so we can review only the changes to the workflows? There is noise right now that prevents me to review properly the PR.

titusfortner:
I'll convert it to a draft while I'm working on it.

PR #16818, opened 2025-12-30
[dotnet] update what tests are run on GitHub Actions

titusfortner:
### 🔗 Related Issues

.NET implementation of #16809 

### 💥 What does this PR do?

* `ci.yml` kicks off on every commit. It runs the new and improved `check-bazel-targets.sh`
* Instead of just checking to see if there are any `dotnet` targets and running the whole ci-dotnet workflow, the unique set of applicable test targets are passed from the script to `ci-dotnet.yml` and browser tests are only run on those

To get all the tests passing I did:
* Changed InstallCommand extension handling to support Windows better
* Truncated the logs to make results easier to read (100 char to match Java), I didn't make it configurable, but we could
* Submit() to use GetDomAttribute instead of GetAttribute (since we know it is an attribute)
* A couple test fixes and guards

### 🔧 Implementation Notes

* Current implementation is only 2 smoke tests on Windows, this is going to run all the tests for 3 browsers right now; I want to see what is passing right now before restricting it more.

* Total execution time for all of this is: [3h 38m 59s 
](https://github.com/SeleniumHQ/selenium/actions/runs/20648912306/usage)
* We need define "smoke tests" for these and limit tests
* We should also create Unit Tests
* Hmm, and should try to run Safari just to make sure it works.

___

### **PR Type**
Enhancement

___

### **Description**
- Add workflow inputs to `ci-dotnet.yml` for selective test execution

- Implement target filtering logic to run only affected .NET tests

- Expand browser test matrix to cover Chrome, Firefox, and Edge

- Pass filtered targets from `ci.yml` to `ci-dotnet.yml` workflow

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["ci.yml check job"] -->|"targets output"| B["ci-dotnet.yml"]
  B -->|"filter-targets job"| C["Filter .NET targets"]
  C -->|"filtered targets"| D["browser-tests matrix"]
  D -->|"Chrome/Firefox/Edge"| E["Run bazel tests"]
```

<tr>
  <td>
    <details>
      <summary><strong>ci-dotnet.yml</strong><dd><code>Add selective test execution and browser matrix</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

.github/workflows/ci-dotnet.yml

<ul><li>Added workflow inputs <code>targets</code> and <code>run-full-suite</code> to support selective <br>test execution<br> <li> Introduced <code>filter-targets</code> job to filter applicable .NET targets from <br>upstream workflow<br> <li> Replaced hardcoded <code>integration-tests</code> job with parameterized <br><code>browser-tests</code> job using matrix strategy<br> <li> Added browser matrix for Chrome, Firefox, and Edge with enhanced bazel <br>test flags</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16818/files#diff-9f761b9e8edbe9c2db37b8ef3863ecc6318bc1930386c72585733ae09ae9016f">+54/-6</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>ci.yml</strong><dd><code>Pass filtered targets to .NET workflow</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

.github/workflows/ci.yml

<ul><li>Pass <code>targets</code> output from check job to <code>ci-dotnet.yml</code> workflow<br> <li> Add <code>run-full-suite</code> input to control full test suite execution based on <br>event type<br> <li> Removed commit message check from conditional, relying on PR title <br>check instead</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16818/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f">+9/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/dotnet/private/dotnet_nunit_test_suite.bzl
+++ b/dotnet/private/dotnet_nunit_test_suite.bzl
@@ -101,11 +101,22 @@ _HEADLESS_ARGS = select({
     "//conditions:default": [],
 })
 
+_BIDI_BROWSERS = ["chrome", "edge", "firefox"]
+
 def _is_test(src, test_suffixes):
     for suffix in test_suffixes:
         if src.endswith(suffix):
             return True
     return False
+
+def _browser_variant_name(test_name, browser):
+    return "%s-%s" % (test_name, browser)
+
+def _browser_variant_tags(browser, is_bidi):
+    tags = [browser] + COMMON_TAGS + _BROWSERS[browser]["tags"]
+    if is_bidi:
+        tags.append("bidi")
+    return tags
 
 _NUNIT_ARGS = [
     "--workers=1",  # Bazel tests share a single driver instance; prevent NUnit parallelism
@@ -139,6 +150,8 @@ def dotnet_nunit_test_suite(
         suffix = src.rfind(".")
         test_name = src[:suffix]
 
+        is_bidi = "BiDi/" in src
+
         if not browsers or not len(browsers):
             csharp_test(
                 name = test_name,
@@ -146,14 +159,16 @@ def dotnet_nunit_test_suite(
                 deps = deps + extra_deps,
                 target_frameworks = target_frameworks,
                 data = data,
-                tags = tags,
+                tags = tags + (["bidi"] if is_bidi else []),
                 size = size,
                 **kwargs
             )
             tests.append(test_name)
         else:
             for browser in browsers:
-                browser_test_name = "%s-%s" % (test_name, browser)
+                if is_bidi and browser not in _BIDI_BROWSERS:
+                    continue
+                browser_test_name = _browser_variant_name(test_name, browser)
 
                 if browser == default_browser:
                     native.test_suite(
@@ -168,7 +183,7 @@ def dotnet_nunit_test_suite(
                     target_frameworks = target_frameworks,
                     args = _NUNIT_ARGS + _BROWSERS[browser]["args"] + _HEADLESS_ARGS,
                     data = data + _BROWSERS[browser]["data"],
-                    tags = tags + [browser] + COMMON_TAGS + _BROWSERS[browser]["tags"],
+                    tags = tags + _browser_variant_tags(browser, is_bidi),
                     size = size,
                     **kwargs
                 )
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -83,6 +83,7 @@ jobs:
       TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
       TWINE_USERNAME: ${{ secrets.TWINE_USERNAME }}
       NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      SE_CACHE_PATH: ${{ runner.temp }}/selenium
     steps:
       - name: Checkout source tree
         uses: actions/checkout@v4
--- a/.github/workflows/ci-dotnet.yml
+++ b/.github/workflows/ci-dotnet.yml
@@ -58,8 +58,9 @@ jobs:
       matrix:
         browser: [chrome, firefox, edge]
         os: [windows]
+        protocol: [classic, bidi]
     with:
-      name: Browser Tests - ${{ matrix.browser }}, ${{ matrix.os }}
+      name: Browser Tests - ${{ matrix.browser }} (${{ matrix.protocol }}), ${{ matrix.os }}
       os: ${{ matrix.os }}
       browser: ${{ matrix.browser }}
       run: >
@@ -68,7 +69,7 @@ jobs:
         --build_tests_only
         --flaky_test_attempts 3
         --local_test_jobs 1
-        --test_tag_filters=${{ matrix.browser }}
+        "--test_tag_filters=${{ matrix.browser }},${{ matrix.protocol == 'bidi' && 'bidi' || '-bidi' }}"
         --test_size_filters=large
         --pin_browsers=false
         --test_env=SE_FORCE_BROWSER_DOWNLOAD=true
--- a/.github/workflows/ci-dotnet.yml
+++ b/.github/workflows/ci-dotnet.yml
@@ -63,6 +63,7 @@ jobs:
       name: Browser Tests - ${{ matrix.browser }} (${{ matrix.protocol }}), ${{ matrix.os }}
       os: ${{ matrix.os }}
       browser: ${{ matrix.browser }}
+      rerun-with-debug: true
       run: >
         bazel test
         --keep_going
--- a/.github/workflows/ci-dotnet.yml
+++ b/.github/workflows/ci-dotnet.yml
@@ -2,6 +2,15 @@ name: CI - .NET
 
 on:
   workflow_call:
+    inputs:
+      targets:
+        required: false
+        type: string
+        default: ''
+      run-full-suite:
+        required: false
+        type: boolean
+        default: true
   workflow_dispatch:
 
 permissions:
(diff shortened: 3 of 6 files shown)


dotnet/src/webdriver/BiDi/WebExtension/InstallCommand.cs
```diff
@@ -33,11 +33,11 @@ internal sealed record InstallParameters(ExtensionData ExtensionData) : Paramete
 [JsonDerivedType(typeof(ExtensionPath), "path")]
 public abstract record ExtensionData;
 
-public sealed record ExtensionArchivePath(string Path) : ExtensionData;
+public sealed record ExtensionArchivePath([property: JsonPropertyName("path")] string Path) : ExtensionData;
```
nvborisenko: Don't introduce explicit property name, we are following spec definition.

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
```diff
@@ -452,7 +452,7 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
 #else
                 var requestContent = await request.Content.ReadAsStringAsync().ConfigureAwait(false);
 #endif
-                requestLogMessageBuilder.AppendFormat("{0}{1}", Environment.NewLine, requestContent);
+                requestLogMessageBuilder.AppendFormat("{0}{1}", Environment.NewLine, Truncate(requestContent));
```
nvborisenko: Behaviour change, out of of scope of this PR.

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
```diff
@@ -463,20 +463,27 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
 
             responseLogMessageBuilder.AppendFormat("<< {0} {1}", (int)response.StatusCode, response.ReasonPhrase);
 
-            if (!response.IsSuccessStatusCode && response.Content != null)
+            if (response.Content != null)
```
nvborisenko: Return the changes back!

dotnet/src/webdriver/Remote/HttpCommandExecutor.cs
```diff
@@ -463,20 +463,27 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
...
             {
 #if NET8_0_OR_GREATER
                 var responseContent = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
 #else
                 var responseContent = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
 #endif
 
-                responseLogMessageBuilder.AppendFormat("{0}{1}", Environment.NewLine, responseContent);
+                responseLogMessageBuilder.AppendFormat("{0}{1}", Environment.NewLine, Truncate(responseContent));
```
nvborisenko: Behavior.

dotnet/src/webdriver/WebElement.cs
```diff
@@ -612,7 +612,7 @@ public virtual void SendKeys(string text)
     /// <exception cref="StaleElementReferenceException">Thrown when the target element is no longer valid in the document DOM.</exception>
     public virtual void Submit()
     {
-        string? elementType = this.GetAttribute("type");
+        string? elementType = this.GetDomAttribute("type");
```
nvborisenko: STOP! Recently we understood that `GetAttribute()` is legible method.

nvborisenko (changes requested):

dotnet/src/webdriver/WebElement.cs
titusfortner: we want to keep it for users as a convenience because they don't know the difference between a property and an attribute. We do. This is more performant.

dotnet/src/webdriver/BiDi/WebExtension/InstallCommand.cs
titusfortner: what we have now doesn't work on Windows. Please suggest an alternate fix.

titusfortner: I'll remove it and we can look at the failure

titusfortner:
@nvborisenko you were right, I didn't need that code. I must have fixed the issue with a different setting.

dotnet/test/common/BiDi/WebExtension/WebExtensionTest.cs
```diff
@@ -88,7 +88,16 @@ private static string LocateRelativePath(string path)
...
         {
-            return Bazel.Runfiles.Create().Rlocation($"_main/{path}");
+            var runfiles = Bazel.Runfiles.Create();
+            string resolved = runfiles.Rlocation($"_main/{path}");
+            if (!string.IsNullOrEmpty(resolved))
+            {
+                return resolved;
+            }
+
+            // For directories, locate a file inside and get parent (runfiles manifest only lists files)
```
nvborisenko, line 98: Not sure it is needed, at least it worked to me in IDE and bazel environments. Up to you, I prefer keeping as minimal as possible.

nvborisenko (approved):

dotnet/test/common/BiDi/WebExtension/WebExtensionTest.cs
titusfortner: It's a bazel thing. Manifest technically only uses files not directories. But because of how Linux does symlinks differently it picks up the directories, but windows does not.

nvborisenko: Good to know (Linux is untested on my end). Seems this method is a great candidate to be global static helper, just a remark (out of scope of this PR).

titusfortner: we want to code it so it works for bazel in all OS. If there's a better place to put it, we can move it.

dotnet/private/dotnet_nunit_test_suite.bzl
```diff
@@ -107,6 +107,19 @@ def _is_test(src, test_suffixes):
...
+        return "%s-%s-bidi" % (test_name, browser)
+    return "%s-%s" % (test_name, browser)
+
+def _browser_variant_tags(browser, is_bidi):
+    if is_bidi:
+        return ["bidi"] + COMMON_TAGS + _BROWSERS[browser]["tags"] + ["%s-bidi" % browser]
+    return [browser] + COMMON_TAGS + _BROWSERS[browser]["tags"]
+
 _NUNIT_ARGS = [
     "--workers=1",  # Bazel tests share a single driver instance; prevent NUnit parallelism
```
nvborisenko, line 122: Hm, can we temporary remove this arg? Classic tests cannot be executed in parallel, but BiDi tests are designed to be parallelizable. AFAIK NUnit doesn't utilize parallelization if not requested explicitly.

titusfortner: it's 4 by default. Let me see if we can set it so it only applies to RBE.

Issue #15596, opened 2025-04-08
[🚀 Feature]: Rewrite Atoms in TypeScript


titusfortner:
### Description

We've discussed this idea [for years](https://github.com/SeleniumHQ/selenium/issues/9668#issuecomment-1072652972).

At this point we know that Closure is not going to play nicely with Bazel 9, and we really need to update this code to use Typescript.

joerg1985:
Is this the right time to ask for improvements to the current atoms?

e.g. a element with width 0px is considered as displayed as soon as there is a text node inside, there is no check for the content of this text node, so a `&nbsp;` will make the element "visible". 

I could provide an example html of this if needed.

AutomatedTester:
I have moved the following to TypeScript

* Get Attribute #17370 
* Relative locators #17458 
* isShown #17316

PR #16829, opened 2026-01-02
Closure to js

shs96c:
This migrates the existing use of the Closure Library and Compiler to use TypeScript and `esbuild`. This should allow us to continue working on the atoms in a more modern way, with a toolchain that is supported.


titusfortner:
For tracking purposes the issue this was attached to https://github.com/SeleniumHQ/selenium/issues/15596

I've updated that issue with current status based on work that @automatedtester has since done

PR #16627, opened 2025-11-22
16612 Download large files from Grid

asolntsev:
### 🔗 Related Issues
Fixes #16612

### 💥 What does this PR do?
This PR allows downloading very large files (I tried up to 4GB) from Selenium Grid, while consuming very low memory.

### 🔄 Types of changes

- Performance improvement (backwards compatible)

___

### **PR Type**
Enhancement, Performance

___

### **Description**
- Add new Grid endpoint `/se/files/:name` for direct file downloads without Base64 encoding

- Optimize `RemoteWebDriver.downloadFile()` to stream files directly instead of loading into memory

- Extract anonymous `Contents.Supplier` implementations into separate named classes for better debugging

- Improve HTTP response handling to support binary streams and large file transfers

- Fix flaky tests by adding proper wait conditions and increased timeouts

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Grid Node"] -->|"GET /se/files/:name"| B["New Direct Download Endpoint"]
  B -->|"Stream file"| C["FileContentSupplier"]
  C -->|"InputStream"| D["RemoteWebDriver"]
  D -->|"Files.copy"| E["Target Location"]
  F["Old Method"] -->|"Base64 + JSON"| G["Memory Issues"]
  style B fill:#90EE90
  style C fill:#90EE90
  style E fill:#87CEEB
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>20 files</summary><table>
<tr>
  <td><strong>Node.java</strong><dd><code>Add new GET endpoint for direct file downloads</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16627/files#diff-9724970402f0d888442caa09bf30d6f8074edde726dae3e98a52eda4f2ed8462">+3/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>LocalNode.java</strong><dd><code>Implement direct file streaming and URL parsing logic</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
[...]

Issue #16612, opened 2025-11-19
[🐛 Bug]: Critically slow file transfer from Node to host for large files


ilya-corp:
### Description

Downloading of large files (approx. 500 Mb) from Selenium Grid Node is critically slow.
To reproduce you need to open browser at node and try to download an large file from server. In my case downloading of 500Mb from server to node takes approx. 10 seconds and uploading file from node to host - 250 seconds.

Potentially problem placed in the following code: https://github.com/SeleniumHQ/selenium/blob/da9dfdcf92dd62b58dd5d8c053749febe484613f/java/src/org/openqa/selenium/remote/RemoteWebDriver.java#L727

```
Response response = execute(DriverCommand.DOWNLOAD_FILE, Map.of("name", fileName));
    String contents = ((Map<String, String>) response.getValue()).get("contents");
    Zip.unzip(contents, targetLocation.toFile());
```

Encoding of 500Mb to Base64 takes a lot of time and born approx. X1.5 volume size (750Mb). Additionally, processing os large GSON also requires many processor resources and memory. Reverse encoding from Base64 to bytes also requires time.
The idea is to use raw bytes transferring without encoding.

Selenium 4.38

### Reproducible Code

```shell
WebElement webElement = ...;
webElement.download();
```

### Debugging Logs

```logs
no
```

PR #16839, opened 2026-01-04
[java] Allow downloading files from old Grid server

asolntsev:
### The problem
Users with
* Selenium client 4.39 or newer
* Selenium Grid 4.38 or older

cannot download files from Grid. 

### 🔧 Implementation Notes
Endpoint "/se/files/:name" was added in Selenium 4.39.0. If user has upgraded only Selenium client version (but not the server!) then method `remoteDriver.downloadFile()` fails with "unsupported command" exception.

Now the download method will work again, falling back to the old endpoint `POST /se/files` (which is slow for large files).

### 🔗 Related Issues
See issue #16612 and PR #16627.

After some time, we can remove this fallback. Probably starting from Selenium 5.0.0

### 🔄 Types of changes

___

### **PR Type**
Bug fix

___

### **Description**
- Add fallback to old Grid endpoint for file downloads

- Support clients with newer Selenium but older Grid versions

- Gracefully handle UnsupportedCommandException with warning log

- Maintain backwards compatibility for file download functionality

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["downloadFile request"] --> B{"Try new endpoint<br/>/se/files/:name"}
  B -->|Success| C["Stream file content"]
  B -->|UnsupportedCommandException| D["Log warning about<br/>old Grid version"]
  D --> E["Fallback to old endpoint<br/>POST /se/files"]
  E --> F["Unzip file contents"]
  C --> G["Save to target location"]
  F --> G
```

<tr>
  <td>
    <details>
      <summary><strong>RemoteWebDriver.java</strong><dd><code>Add fallback for legacy Grid file download endpoint</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

java/src/org/openqa/selenium/remote/RemoteWebDriver.java

<ul><li>Added try-catch block to handle UnsupportedCommandException when new <br>endpoint fails<br> <li> Implemented fallback to legacy POST /se/files endpoint for older Grid <br>servers<br> <li> Added warning log message when old Grid version is detected<br> <li> Imported UnsupportedCommandException and Zip utility classes<br> <li> Refactored file download logic to support both new and old Grid <br>endpoints</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16839/files#diff-0cf141b01e48a733ec9037c51d8e610e7727d05a6b79aa0176def7c3a0fc311b">+20/-6</a>&nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/java/src/org/openqa/selenium/remote/RemoteWebDriver.java
+++ b/java/src/org/openqa/selenium/remote/RemoteWebDriver.java
@@ -18,9 +18,9 @@
 package org.openqa.selenium.remote;
 
 import static java.util.Collections.singleton;
+import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static java.util.logging.Level.SEVERE;
-import static org.openqa.selenium.HasDownloads.DownloadedFile;
 import static org.openqa.selenium.remote.CapabilityType.PLATFORM_NAME;
 
 import java.io.BufferedInputStream;
@@ -69,6 +69,7 @@ import org.openqa.selenium.PrintsPage;
 import org.openqa.selenium.SearchContext;
 import org.openqa.selenium.SessionNotCreatedException;
 import org.openqa.selenium.TakesScreenshot;
+import org.openqa.selenium.UnsupportedCommandException;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.WebElement;
@@ -83,6 +84,7 @@ import org.openqa.selenium.interactions.Interactive;
 import org.openqa.selenium.interactions.Sequence;
 import org.openqa.selenium.internal.Debug;
 import org.openqa.selenium.internal.Require;
+import org.openqa.selenium.io.Zip;
 import org.openqa.selenium.logging.LocalLogs;
 import org.openqa.selenium.logging.LoggingHandler;
 import org.openqa.selenium.logging.Logs;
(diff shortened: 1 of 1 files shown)


titusfortner:
Wait, why did we add a new endpoint for this? It makes a lot more sense for file name to be in the payload instead of in the endpoint.

asolntsev:
> Wait, why did we add a new endpoint for this? It makes a lot more sense for file name to be in the payload instead of in the endpoint.

Why do you prefer to put file name in the payload instead of URL? I don't understand.

But the new payload was needed not because of file name.
It was needed because we need to use different response format. Instead of putting the file content into ZIP (which is very slow and memory-consuming in case of large files), we want to print out the raw file content right to the response. This is much faster. 

P.S. Just in case, it was already done in Selenium 4.39.0.

titusfortner:
because the name can have spaces? Plus introducing new endpoints affects all the languages, not just Java. And this isn't the convention used everywhere else?

titusfortner:
At least for this, it is an implementation detail. I think it would make more sense to provide a toggle in existing endpoint if we need different behavior? But I didn't really read through everything you did in the PR. Just introducing a new endpoint should be a big deal, though. 😂

asolntsev:
> because the name can have spaces?
Usually it's not a problem. File name just should be encoded. E.g. space replaced by '+' etc.

> introducing new endpoints affects all the languages, not just Java.

No matter if it was new endpoint or new parameter/header/whatever else, it doesn't immediately affect other languages.
Other languages can continue using the old endpoint, everything will work as previously.
At any moment, anyone can update other bindings to use the new endpoint/parameter/header.

> And this isn't the convention used everywhere else?

What convention do you mean? @titusfortner 

> it would make more sense to provide a toggle in existing endpoint if we need different behaviour?

It's doable. We can replace new "new" endpoint with a parameter/header in the old endpoint (e.g. http header `Accept: binary/octet-stream`).

asolntsev:
Anyway, we can merge this PR and then discuss how to improve the mechanism for downloading files.

Issue #14474, opened 2024-09-04
[🐛 Bug]: [java] PageFactory.initElements overrides final fields


mk868:
### What happened?

The `PageFactory.initElements(element, this)` method overrides all object's fields, including assigned final fields.
I think this behavior is not correct - final fields should not be changed by reflection.

Even IntelliJ hints that the assigned final fields should not change:
![image](https://github.com/user-attachments/assets/a55f64a1-c347-4dbd-b5ae-31aeaf17562c)

**The problem occurs in the `selenium-support` artifact**

- It is not related to the browser vendor or browser version

**Proposed solution:**

- The `DefaultFieldDecorator` should ignore **final** fields when decorating objects.
- Additionally: the `DefaultFieldDecorator` should also ignore **static** fields when decorating objects.

### How can we reproduce the issue?

_index.html_
```
<div class="card">
    <h2>Title 1</h2>
    <p>Description 1</p>
</div>
<div class="card">
    <h2>Title 2</h2>
    <p>Description 2</p>
</div>
<div class="card">
    <h2>Title 3</h2>
    <p>Description 3</p>
</div>
```

_MainApp.java_
```java
import org.openqa.selenium.By;
import org.openqa.selenium.firefox.FirefoxDriver;
import org.openqa.selenium.firefox.FirefoxOptions;

import java.io.File;

public class MainApp {

    public static void main(String[] args) throws InterruptedException {
        var options = new FirefoxOptions();
        options.setBinary("/snap/firefox/current/usr/lib/firefox/firefox");
        var webDriver = new FirefoxDriver(options);

        webDriver.get(new File("index.html").toURI().toString());

        Thread.sleep(1000);

        var cards = webDriver.findElements(By.cssSelector(".card")).stream()
                .map(CardComponent::new)
                .toList();

        for (var card : cards) {
            System.out.println("Card title: '" + card.getTitle() + "' description: '" + card.getDescription() + "'");
        }

        webDriver.close();
    }
}
```

_CardComponent.java_
```java
import org.openqa.selenium.WebElement;
import org.openqa.selenium.support.FindBy;

public class CardComponent extends Component {

    @FindBy(css = "h2")
    private WebElement title;
    @FindBy(css = "p")
    private WebElement description;

    protected CardComponent(WebElement element) {
        super(element);
    }

    public String getTitle() {
        return title.getText();
    }

    public String getDescription() {
        return description.getText();
    }
}
```

_Component.java_
```java
[...]

PR #16849, opened 2026-01-05
[java] Migrate PageFactory code to be self-contained

shs96c:
This change deprecates the existing classes, but delegates the implementations to the copies made in the `pagefactory` package. This means that a later PR can be made which removes the original classes in `support` once we're ready.

### DIFF
--- a/java/src/org/openqa/selenium/support/pagefactory/BUILD.bazel
+++ b/java/src/org/openqa/selenium/support/pagefactory/BUILD.bazel
@@ -1,0 +1,17 @@
+load("//java:defs.bzl", "java_library")
+
+java_library(
+    name = "pagefactory",
+    srcs = glob([
+        "*.java",
+        "internal/*.java",
+    ]),
+    visibility = [
+        "//java/src/org/openqa/selenium/support:__pkg__",
+        "//java/test/org/openqa/selenium/support:__subpackages__",
+    ],
+    deps = [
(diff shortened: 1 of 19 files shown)


java/src/org/openqa/selenium/support/pagefactory/BUILD.bazel
```diff
@@ -0,0 +1,17 @@
+load("//java:defs.bzl", "java_library")
+
+java_library(
```
shs96c, line 3: We can make this a `java_export` to make this its own maven package. When we're ready to remove PageFactory from the main `support` package, we can then remove this as a dependency of `support` and we'll be done. At that point, we can extract the whole directory to a separate project, or just keep it around.

mk868:
Can we also address this issue https://github.com/SeleniumHQ/selenium/issues/14474 before publishing this change?
The transition to the `.pagefactory.` package will require some changes in the users' code anyway, so they could adapt their code to the changes related to https://github.com/SeleniumHQ/selenium/issues/14474 .

Motivation: https://openjdk.org/jeps/500

What do you think?

diemol:
@mk868, the idea is to move it to a different package when we release Selenium 5 because we aren't going to work more on this code, so we want to move it to a different place so people notice it. We are also considering ways to offer a path that allows people to move out of this pattern.

shs96c:
Not needed

PR #16877, opened 2026-01-09
[rb] check driver status endpoint rather than socket connection

titusfortner:
### **User description**

I'm getting issues in tests where the driver is hanging after starting. Right now we check the socket is connected and then send a session request where it hangs and dies. This code matches what Java, JS & .NET do by checking the status endpoint before continuing. Hopefully if there is a problem we find it here with a shorter timeout.

### 💥 What does this PR do?

* starting a driver checks status endpoint instead of using socketpoller

### 🔧 Implementation Notes

Considered actually looking at ready status, but other bindings don't do that, and probably not what we need here
Shouldn't be a breaking change, but need to test it thoroughly.

### 🔄 Types of changes

___

### **PR Type**
Bug fix

___

### **Description**
- Replace socket polling with HTTP status endpoint check

- Align Ruby driver with Java, JS, and .NET implementations

- Add proper error handling and timeout management

- Improve connection diagnostics with detailed error messages

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Start Driver"] -->|"Old: SocketPoller"| B["Check Socket Connected"]
  A -->|"New: HTTP GET /status"| C["connection_error?"]
  C -->|"Success"| D["Return - Driver Ready"]
  C -->|"Error"| E["Retry with 0.1s delay"]
  E -->|"Within timeout"| C
  E -->|"Timeout exceeded"| F["Raise WebDriverError"]
```

<tr>
  <td>
    <details>
      <summary><strong>service_manager.rb</strong><dd><code>Replace socket polling with HTTP status endpoint</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

rb/lib/selenium/webdriver/common/service_manager.rb

<ul><li>Replaced <code>SocketPoller</code> with HTTP GET request to <code>/status</code> endpoint<br> <li> Implemented <code>connection_error?</code> method to handle HTTP requests and <br>connection exceptions<br> <li> Added <code>current_time</code> helper using <code>Process.clock_gettime</code> for monotonic <br>timing<br> <li> Enhanced error messages to include response codes and exception <br>details<br> <li> Changed polling loop to retry every 0.1 seconds until timeout</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16877/files#diff-91325f9a98b4d10ef61117b88a24a3467a433e1a27a9a16d2b9d4046b1d92c4a">+25/-3</a>&nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/rb/lib/selenium/webdriver/common/service_manager.rb
+++ b/rb/lib/selenium/webdriver/common/service_manager.rb
@@ -130,7 +130,7 @@ module Selenium
         deadline = current_time + START_TIMEOUT
 
         loop do
-          error = connection_error?
+          error = check_connection_error
           return unless error
 
           raise Error::WebDriverError, "#{cannot_connect_error_text}: #{error}" if current_time > deadline
@@ -139,7 +139,7 @@ module Selenium
         end
       end
 
-      def connection_error?
+      def check_connection_error
         response = Net::HTTP.start(@host, @port, open_timeout: 0.5, read_timeout: 1) do |http|
           http.get('/status', {'Connection' => 'close'})
         end
--- a/rb/lib/selenium/webdriver/common/service_manager.rb
+++ b/rb/lib/selenium/webdriver/common/service_manager.rb
@@ -144,10 +144,14 @@ module Selenium
           http.get('/status', {'Connection' => 'close'})
         end
 
-        "status returned #{response.code}\n#{response.body}" unless response.is_a?(Net::HTTPSuccess)
+        return "status returned #{response.code}\n#{response.body}" unless response.is_a?(Net::HTTPSuccess)
+
+        status = JSON.parse(response.body)
+        ready = status['ready'] || status.dig('value', 'ready')
+        "driver not ready: #{response.body}" unless ready
       rescue Errno::ECONNREFUSED, Errno::ECONNRESET, Errno::EPIPE, Errno::ETIMEDOUT,
              Errno::EADDRNOTAVAIL, Errno::EHOSTUNREACH, Net::OpenTimeout, Net::ReadTimeout,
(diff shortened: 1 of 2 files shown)


titusfortner:
@cgoldberg / @navin772 not sure if you've had any weird issues in Python, but Python also does not check status endpoint, and just checks for the socket connection.

navin772:
> Python also does not check status endpoint, and just checks for the socket connection.

I haven't seen any issues, maybe some flaky tests, but I agree we should check for the `/status` endpoint, its a better way.

In python, we already have a helper that checks the `/status` endpoint - `is_url_connectable` but we aren't using it, I will create a PR to update it.

Also, the AI bot's `Validate JSON ready status` looks good to have.

titusfortner:
@navin772 Java & .NET don't check ready value, and I'm not sure we do want that. At least my LLM told me not to when I implemented it first, I don't recall why. 😄

navin772:
Can there be a case when the `/status` endpoint returns `ready` as False? 
In that case our check wouldn't work. The spec has this field, so I think we can use it.

Are there any cons for this approach? 
Also, LLMs tend to follow patterns, so it may have copied the Java/.net logic 🙂

titusfortner:
@navin772 I pinged harder on this, and I think the concerns are ones that don't apply to this situation - grid/remote/insufficient capacity. It makes sense for Java to be more lenient since it is the same code for the nodes.

navin772:
> It makes sense for Java to be more lenient since it is the same code for the nodes.

Sure, lets check how it behaves with python first.

titusfortner:
I updated to match Python, we'll see what it looks like

PR #16881, opened 2026-01-10
[py]: use `is_url_connectable` for driver status check

navin772:
### 💥 What does this PR do?

Python implementation for https://github.com/SeleniumHQ/selenium/pull/16877.

### 🔧 Implementation Notes

It also checks for the `ready` value as per the spec - https://www.w3.org/TR/webdriver/#dfn-status to be True. It may be on the top level or inside the `value` field,

### 🔄 Types of changes

___

### **PR Type**
Bug fix, Enhancement

___

### **Description**
- Implement W3C WebDriver /status endpoint check for driver readiness

- Parse response body to verify `ready` field is True

- Support both top-level and nested `value.ready` response formats

- Add timeout parameter to HTTP request for reliability

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["is_connectable method"] -->|calls| B["is_url_connectable"]
  B -->|makes HTTP request| C["/status endpoint"]
  C -->|returns JSON| D["Parse response body"]
  D -->|check ready field| E["ready is True"]
  E -->|return| F["Boolean result"]
```

<tr>
  <td>
    <details>
      <summary><strong>service.py</strong><dd><code>Replace socket check with /status endpoint verification</code>&nbsp; &nbsp; </dd></summary>
<hr>

py/selenium/webdriver/common/service.py

<ul><li>Updated <code>is_connectable()</code> method to use <code>is_url_connectable()</code> instead of <br>socket connection check<br> <li> Changed implementation to verify W3C WebDriver /status endpoint <br>readiness<br> <li> Updated docstring to reflect HTTP-based status check instead of socket <br>connectivity</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16881/files#diff-646e4c91d7ef7fbd2b4c088347c86f5b2d2cab8121b14987f3cb783ad120634c">+4/-3</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>utils.py</strong><dd><code>Add ready status validation to URL connectivity check</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

py/selenium/webdriver/common/utils.py

<ul><li>Added <code>json</code> import for parsing response body<br> <li> Enhanced <code>is_url_connectable()</code> to parse JSON response and check <code>ready</code> <br>field<br> <li> Added timeout parameter (1 second) to HTTP request<br> <li> Support both top-level <code>ready</code> and nested <code>value.ready</code> response formats <br>per W3C spec<br> <li> Updated docstring with Returns section</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16881/files#diff-b1516e150256fb4a5ef5144d5d8314fc45e92872415ba8900d4cd8d2cc9b9841">+14/-2</a>&nbsp; &nbsp; </td>

</tr>

</details>

___


titusfortner (dismissed):
I don't think we actually want to verify the ready status itself. Java & .NET do not. Let's get more clarification on pros/cons before committing this.

titusfortner:
Actually, if it passes the tests, let's give it a try with the more restrictive behavior. We can add a fallback later or change it to just checking status 200 if there's an issue.

navin772:
The `proxy_tests.py` test is failing and it is related to this PR, I will push a fix, we need to disable proxy for local conn.

navin772:
There seems to be an issue with the linter, the CI fails saying it needs a blank line after the function comment, but running `scripts/format.sh` deletes that blank newline which I added manually.
Could you check @titusfortner @cgoldberg , maybe there's a conflict between a ruff rule and CI linting

titusfortner:
CI linting rule is just running `scripts/format.sh` there shouldn't be any difference between them.

titusfortner:
Oh, yeah, it's the ruff job itself. It "fixes" the extra line then deletes it, lol.

titusfortner:
Oh, it needs an empty line on 127 inside the docstring and no empty line after the docstring.

This passes:
```
    def is_connectable(self) -> bool:
        """Check if the service is ready via the W3C WebDriver /status endpoint.

        This makes an HTTP request to the /status endpoint and verifies if it is ready to accept new sessions.

        Returns:
            True if the service is ready to accept new sessions, False otherwise.
        """
        return utils.is_url_connectable(self.port, host="localhost", scheme="http")
```

cgoldberg:
I think we can add that rule to the ignore list in `pyproject.toml`... or you can shut it off for that file with `# ruff: noqa: XXXX`.

cgoldberg:
I think we can add that rule to the ignore list in `pyproject.toml`... or you can shut it off for that file with `# ruff: noqa: XXXX`.

cgoldberg:
Oh.. looks like you got it to pass

titusfortner:
The error message it gave was definitely confusing, but the LLM understood it 😂

titusfortner:
The code lgtm; I'm not sure how Python does the proxy part of this.

navin772:
Hi @cgoldberg, could you review this?

py/selenium/webdriver/common/utils.py
```diff
@@ -133,19 +134,37 @@ def is_connectable(port: int, host: str | None = "localhost") -> bool:
 
 def is_url_connectable(
     port: int | str,
-    host: str | None = "127.0.0.1",
-    scheme: str | None = "http",
+    host: str = "127.0.0.1",
```
cgoldberg: I think we might want to change this to `"localhost"` so it works by default on IPv6-only systems. Then in `service.py` you can just can just call it with defaults: `return utils.is_url_connectable(self.port)`

cgoldberg (approved):
Minor fix we might want to make.. besides that. LGTM!

cgoldberg (approved):

navin772:
@titusfortner could you approve this, not able to merge with requested changes.

cgoldberg:
@navin772 if you click the 3 dots to the right of his name under "requested change", you can "Dismiss review" and merge ;)

PR #16879, opened 2026-01-10
[rb][py][dotnet][java][js] move logic for reruns and collecting logs into separate scripts

titusfortner:
### 🔗 Related Issues

This is what https://github.com/SeleniumHQ/selenium/pull/16840 was supposed to be.
I separated it from https://github.com/SeleniumHQ/selenium/pull/16858 

* Reruns are disabled by default; running all the tests to show that it does not break existing setup
* Currently we upload all logs on failure, not just failing logs; and they are all nested deep inside directories, this changes that
* Note that with debugging enabled, the output is likely too large for the console, which is why figuring out the uploads is important

### 💥 What does this PR do?

1. Run Bazel step:
    * now executes in a subshell so it still outputs to the file even when it fails, but still fails the step
1. Rerun failures with debug step:
    * Executes rerun-failures.sh with the run inputs and whether or not to rerun
    * Parses failing targets from console log generated in Run Bazel step into `build/failures/_run1.txt`
    * Exits if not rerun
    * Rerun failing targets only with SE_DEBUG=true and flakiness=1; tee console log to `build/bazel-console2.log`
    * Parses failing targets from rerun console log into failures/_run2.txt
3. Collect Failed logs step
    * Iterate over targets in `_run2.txt` if rerun happened, or `_run1.txt` if not
    * copy log files into `build/failures/` and rename with the name of the path so everything that failed is in one directory 
4. Upload Logs step
    * Uploaded file is a zip of the `failures` directory, with a `_run1.txt`, optional `run2.txt` and one file for each of the failures.

### 🔧 Implementation Notes

Putting things in build directory since it is gitignored and we don't want to disrupt the step that generates a git patch with changes. Could put it somewhere else or have a failures directory be in gitignore

### 💡 Additional Considerations

You can see the failing logs here: https://github.com/SeleniumHQ/selenium/actions/runs/20857910080/job/59929556499#step:23:30 

### 🔄 Types of changes

- New feature (non-breaking change which adds functionality *and tests!*)

___

### **PR Type**
Enhancement

___

### **Description**
- Extract test failure handling into separate reusable scripts

- Parse failures from Bazel console logs and optionally rerun with debug

- Collect failed test logs from testlogs directory into single location

- Upload organized failure logs as artifacts with improved structure

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Run Bazel<br/>output to console.log"] --> B["Parse failures<br/>from console.log"]
  B --> C{Rerun<br/>enabled?}
[...]

PR #16894, opened 2026-01-11
[dotnet] rerun failing tests with debugging enabled

titusfortner:
### 🔗 Related Issues

Implements https://github.com/SeleniumHQ/selenium/pull/16879 for Python

### 💥 What does this PR do?

* Turns off default debugging since it is expensive
* Turns on debugging for reruns

We should consider logging in our tests the same way Java does by storing it in a buffer and only writing them to console on failure.


titusfortner:
Turns out .NET filters out these tests since there are no pinned browsers, so nothing is actually being run on GitHub right now, just builds. I'll fix that in another PR, this is noop.

PR #16900, opened 2026-01-13
[java] SE_DEBUG dumps all log output to stderr

titusfortner:
### 🔗 Related Issues

Existing `Debug.java` implementation uses a clever trick to provide additional debugging info via system property by logging things that are normally "FINE" level at "INFO" so they are seen by the user without messing with their other logging setup. (See #12892)

So when I implemented #16816 I did the easy thing ad used it to toggle the clever behavior. Except this usage isn't consistent in the codebase, so it doesn't actually do that much.

### 💥 What does this PR do?

* Keeps existing released behavior for the 2 system properties
* Firehoses all debugging info to stderr when `SE_DEBUG` is set.
* Overrides Grid & Driver logging settings to use level `FINE`/`DEBUG` sent to stderr
* Adds a handler for Selenium JUL, so it is additive to user's other logging solutions unless they explicitly remove selenium handlers elsewhere.

### 🔧 Implementation Notes

* This is a "firehose to stderr" solution
* If a user has `SE_DEBUG` set for Selenium manager this will add a lot of extra output, so could be `SE_DEBUG_ALL`?

### 💡 Additional Considerations

Other languages next

### 🔄 Types of changes

- New feature (non-breaking change which adds functionality *and tests!*)

___

### **PR Type**
Enhancement

___

### **Description**
- Adds `SE_DEBUG` environment variable support for comprehensive debug logging

- Routes all Selenium debug output to stderr when `SE_DEBUG` is enabled

- Overrides driver and grid logging to use `FINE`/`DEBUG` level with stderr output

- Configures JUL handler for Selenium package to capture all debug information

___

### Diagram Walkthrough

```mermaid
flowchart LR
  SE_DEBUG["SE_DEBUG env var"]
  isDebugAll["isDebugAll() check"]
  configLogger["configureLogger()"]
  driverService["Driver Services"]
  gridLogging["Grid Logging"]
  julHandler["JUL Handler to stderr"]
  
  SE_DEBUG --> isDebugAll
  isDebugAll --> configLogger
  configLogger --> julHandler
  isDebugAll --> driverService
  isDebugAll --> gridLogging
  driverService --> julHandler
  gridLogging --> julHandler
```

<details><summary><h3>File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
[...]

PR #16903, opened 2026-01-13
[dotnet] have SE_DEBUG output driver logs to stderr

titusfortner:
### 🔗 Related Issues

.NET implementation of #16900

### 💥 What does this PR do?

* When `SE_DEBUG` is set, driver logs are output to stderr just like selenium logs
* ~Silent/Verbose/Log Level are mutually exclusive, so updated all driver classes to use conditionals for these 3.~
*  Update: keeping original logic where users who pass in more than one setting get failing response
* `SuppressInitialDiagnosticInformation` was in superclass but Firefox was ignoring it, now it won't 

### 🔧 Implementation Notes

~Could change the order of verbose then silent, then log level but that seems best.~

### 🔄 Types of changes

___

### **PR Type**
Enhancement

___

### **Description**
- Implement SE_DEBUG environment variable to output driver logs to stderr

- Make verbose, silent, and log level settings mutually exclusive

- Ensure Firefox respects SuppressInitialDiagnosticInformation setting

- Enable process redirection when SE_DEBUG is set across all drivers

___

### Diagram Walkthrough

```mermaid
flowchart LR
  SE_DEBUG["SE_DEBUG env var set"]
  SE_DEBUG -->|"Check in all drivers"| Verbose["Enable verbose/debug logging"]
  SE_DEBUG -->|"Enable redirection"| Redirect["Redirect process output to stderr"]
  Verbose -->|"Mutually exclusive"| Silent["Silent mode"]
  Silent -->|"Fallback to"| LogLevel["Custom log level"]
  Redirect -->|"Output via"| Stderr["Console.Error.WriteLine"]
```

<tr>
  <td>
    <details>
      <summary><strong>ChromiumDriverService.cs</strong><dd><code>Implement SE_DEBUG and mutually exclusive logging options</code></dd></summary>
<hr>

dotnet/src/webdriver/Chromium/ChromiumDriverService.cs

<ul><li>Check SE_DEBUG environment variable to enable verbose logging<br> <li> Restructure logging options as mutually exclusive (verbose > silent > <br>log level)<br> <li> Move log level argument handling into conditional chain<br> <li> Modernize string formatting from string.Format to string interpolation</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16903/files#diff-7b7a876beec5c5bbf8892cfef44f5d3cc18a1ced78506de3f69ea119d6b80b50">+10/-12</a>&nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>DriverService.cs</strong><dd><code>Enable process redirection and stderr logging for SE_DEBUG</code></dd></summary>
<hr>

dotnet/src/webdriver/DriverService.cs

<ul><li>Change EnableProcessRedirection to check SE_DEBUG environment variable<br> <li> Implement OnDriverProcessDataReceived to output logs to stderr when <br>SE_DEBUG is set<br> <li> Convert property from auto-property to computed property</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16903/files#diff-c38089621d41671b4719164b262303e275d4a4a442125fd9d4ea61eeca855c2b">+6/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>FirefoxDriverService.cs</strong><dd><code>Add SE_DEBUG support and respect suppress flag</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/Firefox/FirefoxDriverService.cs

<ul><li>Check SE_DEBUG to set log level to debug<br> <li> Implement SuppressInitialDiagnosticInformation by setting log level to <br>fatal<br> <li> Restructure logging as mutually exclusive chain (SE_DEBUG > suppress > <br>log level)<br> <li> Update EnableProcessRedirection to include base class check</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16903/files#diff-8a02c4932a0154eb11aafcbf7e736968b70fe5d69dac3df52ac67fa8c8a7c965">+11/-3</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>InternetExplorerDriverService.cs</strong><dd><code>Add SE_DEBUG support and mutually exclusive logging</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/IE/InternetExplorerDriverService.cs

<ul><li>Add System namespace import<br> <li> Check SE_DEBUG to set log level to DEBUG<br> <li> Implement SuppressInitialDiagnosticInformation with silent flag<br> <li> Restructure logging as mutually exclusive chain (SE_DEBUG > suppress > <br>log level)<br> <li> Modernize string formatting to string interpolation</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16903/files#diff-b5bc14c6fe80476d83f529175715431b5a04ac8b0bdffd4d39f1d32f9c865615">+11/-7</a>&nbsp; &nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>I-cleanup</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>LogContextManager.cs</strong><dd><code>Fix variable naming typo</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

dotnet/src/webdriver/Internal/Logging/LogContextManager.cs

<ul><li>Fix typo in variable name from defaulLogHandler to defaultLogHandler</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16903/files#diff-259c2c4f3567044d233aa46147d492b89964cbe4bc5e45fd976a03c88674007f">+2/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/dotnet/src/webdriver/Chromium/ChromiumDriverService.cs
+++ b/dotnet/src/webdriver/Chromium/ChromiumDriverService.cs
@@ -120,23 +120,29 @@ public abstract class ChromiumDriverService : DriverService
 
             if (Environment.GetEnvironmentVariable("SE_DEBUG") is not null)
             {
-                if (this.SuppressInitialDiagnosticInformation || this.LogLevel != ChromiumDriverLogLevel.Default)
+                if (this.SuppressInitialDiagnosticInformation ||
+                    (this.LogLevel != ChromiumDriverLogLevel.Default && this.LogLevel != ChromiumDriverLogLevel.All))
                 {
                     Console.Error.WriteLine("WARNING: Environment Variable `SE_DEBUG` is set; forcing ChromiumDriver --verbose and overriding --silent/--log-level settings.");
                 }
                 argsBuilder.Append(" --verbose");
             }
-            else if (this.EnableVerboseLogging)
+            else
             {
-                argsBuilder.Append(" --verbose");
-            }
-            else if (this.SuppressInitialDiagnosticInformation)
-            {
-                argsBuilder.Append(" --silent");
-            }
-            else if (this.LogLevel != ChromiumDriverLogLevel.Default)
-            {
-                argsBuilder.Append($" --log-level={this.LogLevel.ToString().ToUpperInvariant()}");
+                if (this.EnableVerboseLogging)
+                {
+                    argsBuilder.Append(" --verbose");
+                }
+
+                if (this.SuppressInitialDiagnosticInformation)
+                {
+                    argsBuilder.Append(" --silent");
+                }
+
+                if (this.LogLevel != ChromiumDriverLogLevel.Default)
+                {
+                    argsBuilder.Append($" --log-level={this.LogLevel.ToString().ToUpperInvariant()}");
+                }
             }
 
             if (this.EnableAppendLog)
--- a/dotnet/src/webdriver/Chromium/ChromiumDriverService.cs
+++ b/dotnet/src/webdriver/Chromium/ChromiumDriverService.cs
@@ -113,19 +113,30 @@ public abstract class ChromiumDriverService : DriverService
                 argsBuilder.AppendFormat(CultureInfo.InvariantCulture, " --adb-port={0}", adb);
             }
 
-            if (this.SuppressInitialDiagnosticInformation)
-            {
-                argsBuilder.Append(" --silent");
-            }
-
             if (this.DisableBuildCheck)
             {
                 argsBuilder.Append(" --disable-build-check");
             }
 
-            if (this.EnableVerboseLogging)
+            if (Environment.GetEnvironmentVariable("SE_DEBUG") is not null)
+            {
+                if (this.SuppressInitialDiagnosticInformation || this.LogLevel != ChromiumDriverLogLevel.Default)
+                {
+                    Console.Error.WriteLine("WARNING: Environment Variable `SE_DEBUG` is set; forcing ChromiumDriver --verbose and overriding --silent/--log-level settings.");
+                }
+            }
+
+            if (Environment.GetEnvironmentVariable("SE_DEBUG") is not null || this.EnableVerboseLogging)
             {
                 argsBuilder.Append(" --verbose");
+            }
+            else if (this.SuppressInitialDiagnosticInformation)
+            {
+                argsBuilder.Append(" --silent");
+            }
+            else if (this.LogLevel != ChromiumDriverLogLevel.Default)
+            {
+                argsBuilder.Append($" --log-level={this.LogLevel.ToString().ToUpperInvariant()}");
             }
 
             if (this.EnableAppendLog)
@@ -155,12 +166,7 @@ public abstract class ChromiumDriverService : DriverService
 
             if (!string.IsNullOrEmpty(this.AllowedIPAddresses))
             {
-                argsBuilder.Append(string.Format(CultureInfo.InvariantCulture, " -allowed-ips={0}", this.AllowedIPAddresses));
-            }
-
-            if (this.LogLevel != ChromiumDriverLogLevel.Default)
-            {
-                argsBuilder.Append(string.Format(CultureInfo.InvariantCulture, " --log-level={0}", this.LogLevel.ToString().ToUpperInvariant()));
+                argsBuilder.Append($" -allowed-ips={this.AllowedIPAddresses}");
             }
 
             // Unconditionally redirect browser logs to the same log as the driver
--- a/dotnet/src/webdriver/Chromium/ChromiumDriverService.cs
+++ b/dotnet/src/webdriver/Chromium/ChromiumDriverService.cs
@@ -124,9 +124,9 @@ public abstract class ChromiumDriverService : DriverService
                 {
                     Console.Error.WriteLine("WARNING: Environment Variable `SE_DEBUG` is set; forcing ChromiumDriver --verbose and overriding --silent/--log-level settings.");
                 }
+                argsBuilder.Append(" --verbose");
             }
-
-            if (Environment.GetEnvironmentVariable("SE_DEBUG") is not null || this.EnableVerboseLogging)
+            else if (this.EnableVerboseLogging)
             {
                 argsBuilder.Append(" --verbose");
             }
--- a/dotnet/src/webdriver/DriverService.cs
+++ b/dotnet/src/webdriver/DriverService.cs
@@ -176,7 +176,8 @@ public abstract class DriverService : ICommandServer
     /// <remarks>Set this property to <see langword="true"/> to force all process output and error streams to
     /// be redirected, even if redirection is not required by default behavior. This can be useful in scenarios where
     /// capturing process output is necessary for logging or analysis.</remarks>
-    protected virtual internal bool EnableProcessRedirection { get; } = false;
+    protected virtual internal bool EnableProcessRedirection =>
+        Environment.GetEnvironmentVariable("SE_DEBUG") is not null;
 
     /// <summary>
     /// Gets a value indicating whether the service is responding to HTTP requests.
@@ -347,7 +348,10 @@ public abstract class DriverService : ICommandServer
     /// <param name="args">The data received event arguments.</param>
     protected virtual void OnDriverProcessDataReceived(object sender, DataReceivedEventArgs args)
     {
-
+        if (Environment.GetEnvironmentVariable("SE_DEBUG") is not null && !string.IsNullOrEmpty(args.Data))
+        {
+            Console.Error.WriteLine(args.Data);
+        }
     }
 
     /// <summary>
--- a/dotnet/src/webdriver/Internal/Logging/LogContextManager.cs
+++ b/dotnet/src/webdriver/Internal/Logging/LogContextManager.cs
@@ -19,24 +19,29 @@
 
 using System;
 using System.Diagnostics.CodeAnalysis;
-using System.Threading;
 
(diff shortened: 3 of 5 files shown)


nvborisenko:
I believe `SE_DEBUG` env variable should control internal logging level. And then each component can rely on the current level to understand what command arg can be included additionally. Internal logger already can be configured via env variables (?), and already writes to stderr. I think it is good idea just to **implicitly** add process args depending on internal log level. In future we can easily fine tune how to configure it via env vars, rather than fix all components.

titusfortner:
The goal is to have an easy toggle to get driver output in CI logs. Just making sure this implementation makes sense for .NET. Plus I wanted to highlight the slightly changed service logic.

nvborisenko:
This toggle already exists: https://github.com/SeleniumHQ/selenium/blob/1a8200fb3859d5cbf2710e880aeb7f105fddd418/dotnet/src/webdriver/Internal/Logging/LogContextManager.cs#L34-L37

dotnet/src/webdriver/Chromium/ChromiumDriverService.cs
```diff
@@ -113,20 +113,23 @@ protected override string CommandLineArguments
...
             {
                 argsBuilder.Append(" --disable-build-check");
             }
 
-            if (this.EnableVerboseLogging)
+            if (Environment.GetEnvironmentVariable("SE_DEBUG") is not null || this.EnableVerboseLogging)
             {
                 argsBuilder.Append(" --verbose");
             }
+            else if (this.SuppressInitialDiagnosticInformation)
```
RenderMichael: I'm not certain the history behind why we have separate configuration knobs for `EnableVerboseLogging`, `SuppressInitialDiagnosticInformation`, and `LogLevel`, but it feels like too much magic to only one of these threee configurations.

If we consider something like `EnableVerboseLogging = true, SuppressInitialDiagnosticInformation = true` to be a misconfiguration, then I still feel like we should just transparently pass these values along here and let happen what may.

RenderMichael: My only concern is about some of the existing configuration knobs, the actual `SE_DEBUG` implementation seems sound.

The knobs `SuppressInitialDiagnosticInformation` and `EnableVerboseLogging` look like they are C#-specific (the latter even chromium-specific). People do seem to be using `SuppressInitialDiagnosticInformation`, at least according to [SuppressInitialDiagnosticInformation | Grep by Vercel](https://grep.app/search?f.lang=C%23&q=SuppressInitialDiagnosticInformation) and `EnableVerboseLogging` is impossible to really grep for (hard to say what lies in the test suites of massive companies of course). It feels too big of a silent breaking change for gains that I'm not necessarily seeing.

titusfortner: > This toggle already exists:

@nvborisenko it doesn't apply to driver logs, which is the point of this PR

@RenderMichael It literally breaks if you use more than one of those:
```
titus@Tituss-MacBook-Pro selenium % ~/.cache/selenium/chromedriver/mac-arm64/144.0.7559.31/chromedriver --verbose --log-level=INFO
Starting ChromeDriver 144.0.7559.31 (18537d992890dc68b11c8a4f748cad3c7df165c4-refs/branch-heads/7559@{#1444}) on port 0
Please see https://chromedriver.chromium.org/security-considerations for suggestions on keeping ChromeDriver safe.
Only one of --log-level, --verbose, or --silent is allowed.
Unable to initialize logging. Exiting...
```

So if we are going to force override DEBUG for the purposes of getting the logs with a single toggle, we need to make sure the other values aren't set.

titusfortner: We could keep the current logic and do what we're doing in Ruby/Python where we only ignore things if SE_DEBUG is set, but this won't break existing code, it might only confuse future code if people try to set both silent and log level info expecting info and getting silent.

RenderMichael: I guess this is coming from the experience of forgetting that various environment variables are set before trying to run some code; the following code would produce unintuitive results without a clear culprit:

```C#
// Say SE_DEBUG is set
var service = ChromeDriverService.CreateDefaultService();
service.SuppressInitialDiagnosticInformation = true;
var driver = new ChromeDriver(service); // why isn't this silent?
```

In my opinion, users would benefit from us adding a trace log in the contradictory branches explaining that the 
environment variable overrides this config.

Per [the docs on log level](https://www.selenium.dev/documentation/webdriver/browsers/edge/#log-level):

> There are 6 available log levels: `ALL`, `DEBUG`, `INFO`, `WARNING`, `SEVERE`, and `OFF`. Note that `--verbose` is equivalent to `--log-level=ALL` and `--silent` is equivalent to `--log-level=OFF`

The two scenarios I see are:
- `this.SuppressInitialDiagnosticInformation && Environment.GetEnvironmentVariable("SE_DEBUG") is not null` 
-  `this.LogLevel != ChromiumDriverLogLevel.Default && this.LogLevel != ChromiumDriverLogLevel.All && Environment.GetEnvironmentVariable("SE_DEBUG") is not null`

titusfortner: @RenderMichael good point. I've updated the code to put a warning on first use, and then any time there is a conflict between what was asked for and what SE_DEBUG is forcing.

I'll add this to the other languages as well. Thanks.

dotnet/src/webdriver/Chromium/ChromiumDriverService.cs
```diff
@@ -113,20 +113,31 @@ protected override string CommandLineArguments
...
             {
-                argsBuilder.Append(" --silent");
+                argsBuilder.Append(" --disable-build-check");
             }
 
-            if (this.DisableBuildCheck)
+            if (Environment.GetEnvironmentVariable("SE_DEBUG") is not null)
             {
-                argsBuilder.Append(" --disable-build-check");
+                if (this.SuppressInitialDiagnosticInformation || this.LogLevel != ChromiumDriverLogLevel.Default)
```
RenderMichael: Somewhat of a nitpick: f `this.LogLevel` is set to `All`, then there is no conflict here and a warning seems unnecessary.

dotnet/src/webdriver/Chromium/ChromiumDriverService.cs
RenderMichael: These changes definitely alleviate my concerns in the general case, thanks.

I’m still a little hesitant with the new ‘else’-ing, where previously conflicting arguments would surface an error like you mentioned earlier; with these changes, we now silently prioritize `verbose` over `silent` over `log-level`.

dotnet/src/webdriver/DriverService.cs
```diff
@@ -176,7 +176,8 @@ public int ProcessId
     /// <remarks>Set this property to <see langword="true"/> to force all process output and error streams to
     /// be redirected, even if redirection is not required by default behavior. This can be useful in scenarios where
     /// capturing process output is necessary for logging or analysis.</remarks>
-    protected virtual internal bool EnableProcessRedirection { get; } = false;
+    protected virtual internal bool EnableProcessRedirection =>
+        Environment.GetEnvironmentVariable("SE_DEBUG") is not null;
 
     /// <summary>
```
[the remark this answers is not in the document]
titusfortner: I'm not opposed to it, but I think the pattern is clear the way it is.

dotnet/src/webdriver/Internal/Logging/LogContextManager.cs
```diff
@@ -19,24 +19,29 @@
 
 using System;
 using System.Diagnostics.CodeAnalysis;
-using System.Threading;
 
 namespace OpenQA.Selenium.Internal.Logging;
 
 internal class LogContextManager
 {
+    private static bool _seDebugWarned;
```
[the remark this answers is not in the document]
titusfortner: so... it prints twice? the horror.

dotnet/src/webdriver/Chromium/ChromiumDriverService.cs
titusfortner: @RenderMichael ok, you convinced me. I put it back.

titusfortner: I think this is good to go?

RenderMichael (approved):

dotnet/src/webdriver/Chromium/ChromiumDriverService.cs
RenderMichael: Yes, thank you for seeing where I’m coming from!

PR #16917, opened 2026-01-16
[build] Use symlinks in local_dev

cgoldberg:
### 💥 What does this PR do?

This PR removes the build step from the `local_dev` rake task in the `go` wrapper.

When you invoke bazel through a rake task and build python, it creates bytecode files in some of the output directories when running the `generate.py` script to generate CDP files, and bazel doesn't like running after that.

To do local Python development, NEVER build using `go`.

instead, do:

```
bazel build //py/...
./go py:local_dev
```

This PR also changes the task to symlink the necessary files into your local source tree instead of copying them.. this gives us a single source of truth for all files.

### 🔄 Types of changes

- Dev/Build

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -577,8 +577,6 @@ namespace :py do
 
   desc 'generate and copy files required for local development'
   task :local_dev do
-    Bazel.execute('build', [], '//py:selenium')
-    Rake::Task['grid'].invoke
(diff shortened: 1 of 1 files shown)


cgoldberg:
Bazel config was fixed.. we don't need this

PR #16961, opened 2026-01-20
[java] fix JSON parsing of numbers with exponent

joerg1985:
### 🔗 Related Issues
I noticed this when reading the z-index of an element via Javascript.
All numbers with exponent and no fraction part are broken.
e.g. `42e-1` should be parsed as `double 4.2` but instead was parsed as `long 4`

### 💥 What does this PR do?
Fix the JSON parsing of numbers with exponent and no fraction part.

### 🔄 Types of changes

___

### **PR Type**
Bug fix

___

### **Description**
- Fix JSON parsing of numbers with exponent notation

- Correctly parse exponents without fractional parts as doubles

- Refactor number parsing logic using switch statement

- Add test cases for exponent notation variants

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["JSON Number Input"] --> B["Check for decimal point or exponent"]
  B --> C{Contains decimal or exponent?}
  C -->|Yes| D["Parse as Double via BigDecimal"]
  C -->|No| E["Parse as Long"]
  D --> F["Return Number"]
  E --> F
```

<tr>
  <td>
    <details>
      <summary><strong>JsonInput.java</strong><dd><code>Refactor number parsing with exponent support</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

java/src/org/openqa/selenium/json/JsonInput.java

<ul><li>Refactored number parsing logic from character-by-character checks to <br>switch statement<br> <li> Changed tracking from <code>fractionalPart</code> boolean to <code>decimal</code> boolean to <br>detect both decimal points and exponents<br> <li> Fixed bug where numbers with exponents but no fractional part were <br>incorrectly parsed as Long instead of Double<br> <li> Simplified parsing logic to return Long for integers and Double for <br>numbers with decimal or exponent notation</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16961/files#diff-162d1e84f8475ebea31dd037e38d700e07beacf189da303f454b1b62ad7b8483">+30/-18</a>&nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>JsonTest.java</strong><dd><code>Add exponent notation test cases</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

java/test/org/openqa/selenium/json/JsonTest.java

<ul><li>Added test case for exponent with decimal fraction: <code>4.2e+1</code> should <br>parse as <code>42.0</code><br> <li> Added test case for positive exponent without fraction: <code>42e+1</code> should <br>parse as <code>420.0</code><br> <li> Added test case for negative exponent without fraction: <code>42e-1</code> should <br>parse as <code>4.2</code><br> <li> Added test case for negative exponent with fraction: <code>4.2e-1</code> should <br>parse as <code>0.42</code></ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16961/files#diff-ca8b46815d2ed84d1c82776cf430ab11d8b9cdb91a4c904338c96dc482147623">+4/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/java/src/org/openqa/selenium/json/JsonInput.java
+++ b/java/src/org/openqa/selenium/json/JsonInput.java
@@ -220,31 +220,43 @@ public class JsonInput implements Closeable {
    */
   public Number nextNumber() {
     expect(JsonType.NUMBER);
+    boolean decimal = false;
     StringBuilder builder = new StringBuilder();
     // We know it's safe to use a do/while loop since the first character was a number
-    boolean fractionalPart = false;
+    boolean read = true;
     do {
-      char read = input.peek();
-      if (Character.isDigit(read)
-          || read == '+'
-          || read == '-'
-          || read == 'e'
-          || read == 'E'
-          || read == '.') {
-        builder.append(input.read());
-      } else {
-        break;
+      switch (input.peek()) {
+        case '-':
+        case '+':
+        case '0':
+        case '1':
+        case '2':
+        case '3':
+        case '4':
+        case '5':
+        case '6':
+        case '7':
+        case '8':
+        case '9':
+          builder.append(input.read());
+          break;
+        case '.':
+        case 'e':
+        case 'E':
+          decimal = true;
+          builder.append(input.read());
+          break;
+        default:
+          read = false;
       }
-
-      fractionalPart |= (read == '.');
-    } while (true);
+    } while (read);
 
     try {
-      Number number = new BigDecimal(builder.toString());
-      if (fractionalPart) {
-        return number.doubleValue();
+      if (!decimal) {
+        return Long.valueOf(builder.toString());
       }
-      return number.longValue();
+
+      return new BigDecimal(builder.toString()).doubleValue();
     } catch (NumberFormatException e) {
       throw new JsonException("Unable to parse to a number: " + builder + ". " + input);
     }
--- a/java/test/org/openqa/selenium/json/JsonTest.java
+++ b/java/test/org/openqa/selenium/json/JsonTest.java
@@ -63,6 +63,10 @@ class JsonTest {
     assertThat((Number) new Json().toType("42", Number.class)).isEqualTo(42L);
     assertThat((Integer) new Json().toType("42", Integer.class)).isEqualTo(42);
     assertThat((Double) new Json().toType("42", Double.class)).isEqualTo(42.0);
+    assertThat((Double) new Json().toType("4.2e+1", Double.class)).isEqualTo(42.0);
+    assertThat((Double) new Json().toType("42e+1", Double.class)).isEqualTo(420.0);
+    assertThat((Double) new Json().toType("42e-1", Double.class)).isEqualTo(4.2);
+    assertThat((Double) new Json().toType("4.2e-1", Double.class)).isEqualTo(0.42);
   }
 
   @Test


VietND96:
This is fine to me, can you also check the code suggestion `Implement robust stateful number parsing` is valid or not

VietND96 (approved):

asolntsev (changes requested):
I found a bug:

```java
assertThat((Double) new Json().toType("42e+1", Integer.class)).isEqualTo(420);
```

throws `java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.Double` :)

joerg1985:
> I found a bug:
> 
> ```java
> assertThat((Double) new Json().toType("42e+1", Integer.class)).isEqualTo(420);
> ```
> 
> throws `java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.Double` :)

The line above does try to cast an `Integer` to a `Double`, there is no valid change to the Json parser that could fix this.

asolntsev:
@joerg1985 My bad, sorry! :)
LOL I was too rushing with findings, like a young cowboy... :) 

Anyway, I have one small suggestion how to simplify the code.

java/src/org/openqa/selenium/json/JsonInput.java
```diff
@@ -220,31 +220,43 @@ public String nextName() {
...
     try {
-      Number number = new BigDecimal(builder.toString());
-      if (fractionalPart) {
-        return number.doubleValue();
+      if (!decimal) {
+        return Long.valueOf(builder.toString());
       }
-      return number.longValue();
+
+      return new BigDecimal(builder.toString()).doubleValue();
```
asolntsev, line 259: Seems like we could simplify this part and always return `BigDecimal` (instead of Long/Double).

Instead of 
```java
if (!decimal) {
        return Long.valueOf(builder.toString());
      }

      return new BigDecimal(builder.toString()).doubleValue();
```

We can just leave one line:

```java
return new BigDecimal(builder.toString());
```

asolntsev (changes requested):

java/src/org/openqa/selenium/json/JsonInput.java
joerg1985: This will lead to alot of `ClassCastException`s, the client code might expect either a `Long` or a `Double`.
And yes this is propably an issue in the client code, the `Number` interface should be used there.

By the way, the Selenium code is also not prepared for such a change:
 - the `RemoteWebDriver` does cast to `Long`
 - the `Credential` does cast to `Long`
 - the `AddHasNetworkConditions` does cast to `Long`
 - the `RemoteLogs` does cast to `Long`

asolntsev: @joerg1985  No, it will not cause a problem.

This method returns `java.lang.Number` (and `BigDecimal` implements `Number`).
At the next step, during casting one of methods `number.longValue()` or `number.intValue()` or `number.doubleValue()` will be called.

At least unit-tests stay green.

joerg1985: I have just created the `json-fix-exponent-bigdecimal` branch and did debug into into it to see the estimation is true.
And `JsonInputTest.shouldParseDecimalNumbersAsDoubles` does instantly [fail](https://github.com/SeleniumHQ/selenium/actions/runs/21223441265/job/61063972893)

But even worse the `NumberCoercer` does get active and [guess the type](https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/json/JsonTypeCoercer.java#L82-L88) of the `BigDecimal` to `Long` or `Double` in the other places. So there is no `ClassCastException`, as all `Numbers` without fraction are guessed to `Long` and luckly we never cast blind `Double`.

So i would suggest to play safe and continue with the fix without the risk to break something.

asolntsev: Sure, my ideas should not tackle this PR :)
I can play with the simplification later.

asolntsev (approved):
A general question is: why Selenium re-invented the wheel instead of using some read-made JSON parser, e.g. Jackson?

This bug shows why it was a bad idea...

asolntsev (dismissed):
@joerg1985 Wait, this time I found a bug. I think. Maybe. Probably. :)

```java
  @Test
  void shouldParseNonDecimalNumbersAsLongs_e() {
    try (JsonInput input = newInput("42e+1")) {
      assertThat(input.peek()).isEqualTo(NUMBER);
      assertThat(input.nextNumber()).isEqualTo(420L);
    }
  }

expected: 420L
 but was: 420.0
```

joerg1985:
> A general question is: why Selenium re-invented the wheel instead of using some read-made JSON parser, e.g. Jackson?
> 
> This bug shows why it was a bad idea...

I don't know, but in the past GSON has been used and i guess they had good reasons for this.

Into the exponent without fraction i have to look closer the next days.

asolntsev:
Gson is not maintained anymore. So yes, it was a good idea to stop using Gson. But it doesn't mean we had to re-invent the wheel. ;)

This bug is not critical at all, indeed. I think we can merge this PR.

asolntsev:
@joerg1985 I found another problem. At least inconsistency.

The case is when JsonInput is parsing a number followed by text. For two different texts, it shows a different behaviour:

```java
  @Test
  void bug() {
    try (JsonInput input = newInput("12345altruistic")) {
      assertThat(input.peek()).isEqualTo(NUMBER);
      assertThat(input.nextNumber()).isEqualTo(12345L);
    }
    try (JsonInput input = newInput("12345egoistic")) { // throws JsonException: Unable to parse to a number: 12345e
      assertThat(input.peek()).isEqualTo(NUMBER);
      assertThat(input.nextNumber()).isEqualTo(12345L);
    }
  }
```

asolntsev:
@joerg1985 I tried a bit different fix using `BigDecimal.getScale()`.
Perhaps a bit simpler in the end.

https://github.com/SeleniumHQ/selenium/pull/16988

joerg1985:
> @joerg1985 I found another problem. At least inconsistency.
> 
> The case is when JsonInput is parsing a number followed by text. For two different texts, it shows a different behaviour:
> 
> ```java
>   @Test
>   void bug() {
>     try (JsonInput input = newInput("12345altruistic")) {
>       assertThat(input.peek()).isEqualTo(NUMBER);
>       assertThat(input.nextNumber()).isEqualTo(12345L);
>     }
>     try (JsonInput input = newInput("12345egoistic")) { // throws JsonException: Unable to parse to a number: 12345e
>       assertThat(input.peek()).isEqualTo(NUMBER);
>       assertThat(input.nextNumber()).isEqualTo(12345L);
>     }
>   }
> ```

The `JsonInput` does kind of stream the data, so the stream must be read to the end to be sure the complete input was valid.
So this is more or less by design and usually there should be a following `next...`, `begin...` or `end...` call, which must throw in this case.

PR #17006, opened 2026-01-25
[rb] Add rubocop binary target for flexible linting

titusfortner:
### 💥 What does this PR do?

Adds a flexible `//rb:rubocop` binary target that accepts any rubocop flags, 
and removes the now-redundant `//rb:lint-unsafe` target.

### 🔧 Implementation Notes

- `//rb:rubocop` includes only the config and file paths, letting you pass any flags via `--`
- Examples:
  - `bazel run //rb:rubocop -- --autocorrect-all` (replaces `//rb:lint-unsafe`)
  - `bazel run //rb:rubocop -- --only Metrics/AbcSize`
  - `bazel run //rb:rubocop -- --auto-gen-config`
- `//rb:lint` unchanged - still runs as a test with `--autocorrect`

This matches the pattern I've been using with ruff and format

### 🔄 Types of changes

### DIFF
--- a/rb/BUILD.bazel
+++ b/rb/BUILD.bazel
@@ -211,6 +211,15 @@ _LINT_DEPS = [
     "@bundle",
 ]
 
+rb_binary(
+    name = "rubocop",
+    testonly = True,
+    args = _LINT_ARGS,
+    data = _LINT_DATA,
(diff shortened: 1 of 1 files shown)


titusfortner:
fixed and merged into trunk.

PR #17011, opened 2026-01-26
[dotnet] run dotnet-format adopting all suggestions

titusfortner:
*Update*: In draft until we figure out the right approach. See comments below.

I did not understand what I was doing in #16999 

This turns on formatting for suggestion/info level, and this is what it spits out.

I have no idea if we actually want this, but I'd really like to have some kind of enforceable style for dotnet. If this is not the one you want, I think we can tweak .editor.config values to get what we want? I think?

Run this from trunk to see what the rules are it is flagging on:
```
bazel run //dotnet:format -- style --severity info --verify-no-changes
```

@nvborisenko & @RenderMichael which of these do we want to enforce vs ignore?

```
dotnet/src/support/Events/EventFiringWebDriver.cs(632,13): info IDE1005: Delegate invocation can be simplified. [dotnet/src/support/Selenium.WebDriver.Support.csproj]
dotnet/src/support/Events/EventFiringWebDriver.cs(668,13): info IDE1005: Delegate invocation can be simplified. [dotnet/src/support/Selenium.WebDriver.Support.csproj]
dotnet/src/support/Events/EventFiringWebDriver.cs(644,13): info IDE1005: Delegate invocation can be simplified. [dotnet/src/support/Selenium.WebDriver.Support.csproj]
dotnet/src/support/Events/EventFiringWebDriver.cs(728,13): info IDE1005: Delegate invocation can be simplified. [dotnet/src/support/Selenium.WebDriver.Support.csproj]
dotnet/src/support/Events/EventFiringWebDriver.cs(656,13): info IDE1005: Delegate invocation can be simplified. [dotnet/src/support/Selenium.WebDriver.Support.csproj]
dotnet/src/support/Events/EventFiringWebDriver.cs(680,13): info IDE1005: Delegate invocation can be simplified. [dotnet/src/support/Selenium.WebDriver.Support.csproj]
dotnet/src/support/Events/EventFiringWebDriver.cs(608,13): info IDE1005: Delegate invocation can be simplified. [dotnet/src/support/Selenium.WebDriver.Support.csproj]
dotnet/src/support/Events/EventFiringWebDriver.cs(716,13): info IDE1005: Delegate invocation can be simplified. [dotnet/src/support/Selenium.WebDriver.Support.csproj]
dotnet/src/support/Events/EventFiringWebDriver.cs(692,13): info IDE1005: Delegate invocation can be simplified. [dotnet/src/support/Selenium.WebDriver.Support.csproj]
[...]

PR #16999, opened 2026-01-25
[dotnet] Add linting support with configurable dotnet format

titusfortner:
### **User description**
Figuring out how to get linting and formatting working across project

### 💥 What does this PR do?

In #16986 I added support for dotnet's format feature, but it hard coded one setting.

This PR allows the user to pass in parameters just like to native format:
```bash
bazel run //dotnet:format                  # everything (format + style + analyzers)
bazel run //dotnet:format -- whitespace    # whitespace only
bazel run //dotnet:format -- style         # style rules only
bazel run //dotnet:format -- analyzers     # analyzers only
```

The previous also added dotnet to format.sh which adds about 100 seconds to execution.

This PR only runs: `bazel run //dotnet:format -- whitespace` which is about 30 seconds.

This PR then wraps the typical features with:
- `./go dotnet:format` - runs whitespace-only formatting
- `./go dotnet:lint` - runs full format + style + analyzers

Finally, @nvborisenko created `dotnet/.editorconfig` last year, I set it to error so it would fail the lint job.  

### 🔧 Implementation Notes

Whitespace-only mode is ~3x faster than full linting, making it suitable for the general `format.sh` script while full linting can be run separately.

### 💡 Additional Considerations

This is a precursor to a final lint PR that makes sure everything is handled consistently across bindings and properly run by CI.

More things can be enforced by `.editorconfig` now if you want (@RenderMichael / @nvborisenko)

### 🔄 Types of changes

- Cleanup (formatting, renaming)
- New feature (non-breaking change which adds functionality)

___

### **PR Type**
Enhancement, Tests

___

### **Description**
- Add configurable dotnet format with argument support

- Implement separate format (whitespace) and lint (full) tasks

- Apply linting results to codebase with namespace declarations

- Enforce file-scoped namespace style via .editorconfig

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["dotnet_format.bzl"] -- "accept arguments" --> B["format target"]
  B -- "whitespace mode" --> C["format.sh"]
  B -- "full mode" --> D["dotnet.rake"]
  D -- "format task" --> E["whitespace only"]
  D -- "lint task" --> F["format + style + analyzers"]
  G[".editorconfig"] -- "enforce error level" --> F
  H["WebElementWrapper.cs"] -- "apply linting" --> I["file-scoped namespace"]
```

<details><summary><h3>File Walkthrough</h3></summary>

[...]

PR #17019, opened 2026-01-28
[dotnet] update formatting rules

titusfortner:
@RenderMichael / @nvborisenko 
Please look at the individual commits. If there is a style change that you don't think we should make, we need to change the rule in editorconfig. 

We aren't hand-editing anything in this PR except the `.editorconfig`

### 🔗 Related Issues

Replaces #17011 (and a follow-on to #16999)

The previous PR was too much and too opaque

### 💥 What does this PR do?

* Updates editorconfig file in dotnet to be explicit about expected behavior
* I picked a handful of things to explicitly update to demonstrate the pattern
    * csharp_using_directive_placement = outside_namespace:warning (IDE1005)
    * Import ordering
    * dotnet_diagnostic.IDE1006.severity = suggestion  # Naming rule violation
    * dotnet_diagnostic.IDE0059.severity = warning  # Unnecessary assignment
    * dotnet_diagnostic.IDE0044.severity = warning  # Make field readonly
    * dotnet_diagnostic.IDE0076.severity = warning  # Invalid suppression
    * dotnet_diagnostic.IDE1005.severity = warning  # Delegate invocation

### 🔧 Implementation Notes

* Most of these are the defaults, but, again, want to be explicit about what is allowed in case someone does something weird
* We want to run default `//bazel:format -- style` as part of linting process for things that can be automatically fixed, but IDE1006 cannot, so while the code has been updated to pass it, it has been marked suggestion instead of warning, so we can enforce it in a separate CI run.

* This is the last PR to set things up for the formatting / linting PR

### 🔄 Types of changes


dotnet/src/webdriver/Interactions/Actions.cs
```diff
@@ -262,7 +262,7 @@ public Actions KeyUp(IWebElement? element, string theKey)
             throw new ArgumentException("The key value must not be null or empty", nameof(theKey));
         }
 
-        ILocatable? target = GetLocatableFromElement(element);
```
nvborisenko, line 265: Self double check the logic, most likely hidden bug

dotnet/src/webdriver/Interactions/Actions.cs
```diff
@@ -662,7 +662,7 @@ public void Reset()
                 driverAsType = wrapper.WrappedDriver as T;
                 if (driverAsType != null)
                 {
-                    driver = wrapper.WrappedDriver;
```
nvborisenko, line 665: Self double check

nvborisenko (approved):
Left minor comments to double check. If somebody can verify there is no hidden bugs, let's proceed further. Thanks.

dotnet/test/common/DriverTestFixture.cs
```diff
@@ -17,14 +17,15 @@
...
+using System;
 using NUnit.Framework;
 using OpenQA.Selenium.Environment;
 using OpenQA.Selenium.Support.UI;
-using System;
 using static NUnit.Framework.Interfaces.ResultState;
 
 namespace OpenQA.Selenium;
 
+#pragma warning disable IDE1006 // Legacy public member names are used widely across tests.
```
RenderMichael: This is about `driver` right? We can just make that a field.

dotnet/src/webdriver/Interactions/Actions.cs
RenderMichael: Agree, I'm apprehensive about the `_ =` changes. They might codify or legitimize a method call that doesn't need to exist.

dotnet/test/common/DriverTestFixture.cs
titusfortner: No, this wants all the page names to be Pascal instead of camel

titusfortner: and I didn't want to change everything in the test files, though we could...

titusfortner:
@RenderMichael & @nvborisenko I just reverted the 2 commits that were concerning. Ideally one or both of you can go through the TODO section in editorconfig and add the things you want to see. I think the more consistent we are with common formatting practices across the project the better.

Another thing I didn't address here, but our test class names are all over the place. Should the bee `Something`, `SomethingTest` or `SomethingTests`. Ideally it's consistent, but out of scope for this.

dotnet/test/common/DriverTestFixture.cs
titusfortner: but. I'll revert it, you can decide how you want to do it in a follow-up 😁

dotnet/src/webdriver/Interactions/Actions.cs
titusfortner: reverted

dotnet/src/webdriver/Interactions/Actions.cs
titusfortner: reverted

Issue #13989, opened 2024-05-20
[🚀 Feature]: Leverage streaming output in the bindings


titusfortner:
### Status
* **Ruby** —
* **Python** —
* **Java** —
* **.NET** — https://github.com/SeleniumHQ/selenium/pull/17024
* **JS** —

### Feature and motivation

Now that #13260 has been implemented with #13414 we need each of the bindings to
1. switch from `--output json` to `--output mixed`
2. leverage #13437 and use `--log-level` rather than `--debug`
3. switch from printing logged output at the end of the selenium manager execution to using the streamed output

### Usage example

```
./selenium-manager --browser chrome --log-level debug --output mixed
```

aguspe:
@titusfortner is there help wanted for the ruby bindings for this feature?

titusfortner:
Sure!

nvborisenko:
Executed:

```
selenium-manager.exe --browser chrome --log-level trace --output mixed
```

Here `trace` is switch for logging. ~**TBD**: what are all possible values for the switch~ `trace`/`debug` is enough. Then we will put this value depending on current log level set by user.

STDOUT:
```
{
  "driver_path": "C:\\Users\\Nick\\.cache\\selenium\\chromedriver\\win64\\143.0.7499.169\\chromedriver.exe",
  "browser_path": "C:\\Program Files\\Google\\Chrome\\Application\\chrome.exe",
  "error": ""
}
```

STDERR:
```
[2025-12-27T12:05:28.806Z TRACE] Reading metadata from C:\Users\Nick\.cache\selenium\se-metadata.json
[2025-12-27T12:05:28.807Z TRACE] Checking chromedriver in PATH
[2025-12-27T12:05:28.831Z DEBUG] chromedriver not found in PATH
[2025-12-27T12:05:28.831Z DEBUG] chrome detected at C:\Program Files\Google\Chrome\Application\chrome.exe
[2025-12-27T12:05:28.831Z TRACE] Path C:\Program Files\Google\Chrome\Application\chrome.exe has been escaped to C:\\Program Files\\Google\\Chrome\\Application\\chrome.exe
[2025-12-27T12:05:28.831Z DEBUG] Detected browser: chrome 143.0.7499.170
[2025-12-27T12:05:28.832Z TRACE] Reading metadata from C:\Users\Nick\.cache\selenium\se-metadata.json
[...]

PR #17024, opened 2026-01-29
[dotnet] Stream Selenium Manager output to internal logging

nvborisenko:
The changes modernize how browser and driver paths are discovered, enhance logging capabilities, and simplify the interface for driver discovery.

### 🔗 Related Issues
Contributes to https://github.com/SeleniumHQ/selenium/issues/13989

### 💥 What does this PR do?
* Replaced the old `BinaryPaths` method and argument-string construction with a new `DiscoverBrowser` API that takes structured options, builds arguments internally, and returns a `DiscoveryResult` object. This makes the API clearer, more robust, and easier to extend.
* Removed the manual argument-building logic from `DriverFinder` and now use the new `SeleniumManager.DiscoverBrowser` method for driver and browser path resolution.
* Updated references throughout `DriverFinder` to use local constants for key names instead of referencing them from `SeleniumManager`.
* Enhanced the process execution logic for Selenium Manager: now supports timeouts, improved error and output logging, and more robust error messages in case of process failure or timeout.
* Extended the logging interfaces (`ILogger`, `ILogContext`) to support emitting log messages with explicit timestamps, and updated implementations accordingly.

### 🔄 Types of changes


dotnet/src/webdriver/Manager/SeleniumManager.cs
```diff
@@ -268,53 +301,127 @@ private static ResultResponse RunCommand(string arguments)
...
+    /// <summary>
+    /// Contains the paths to the discovered browser driver and browser executable.
+    /// </summary>
+    /// <param name="DriverPath">The absolute path to the browser driver executable.</param>
+    /// <param name="BrowserPath">The absolute path to the browser executable.</param>
+    public record DiscoveryResult(
+        [property: JsonPropertyName("driver_path")] string DriverPath,
+        [property: JsonPropertyName("browser_path")] string BrowserPath);
+
+    const string LogMessageRegexPattern = @"^\[(.*) (INFO|WARN|ERROR|DEBUG|TRACE)\t?\] (.*)$";
```
nvborisenko, line 388: Double check possible values.

dotnet/src/webdriver/SeleniumManager.cs
```diff
@@ -268,53 +309,127 @@ private static ResultResponse RunCommand(string arguments)
...
-            if (_logger.IsEnabled(LogEventLevel.Trace))
+            stdOutputBuilder.AppendLine(e.Data);
+        }
+
+        void HandleErrorOutput(object sender, DataReceivedEventArgs e)
+        {
+            if (e.Data is not null)
             {
-                foreach (var entry in jsonResponse.Logs)
+#if NET8_0_OR_GREATER
```
RenderMichael: nit: we should provide `LogMessageRegex()` on <8 and the `#if` should be there. Something like

```C#

# if NET8_0_OR_GREATER
    [GeneratedRegex(LogMessageRegexPattern)]
    private static partial Regex LogMessageRegexCore();

    private static Regex LogMessageRegex => LogMessageRegexCore();
#else
    private static Regex LogMessageRegex { get; } = new Regex(LogMessageRegexPattern);
#endif
```

and here we just call `var match = LogMessageRegex.Match(e.Data);`

dotnet/src/webdriver/Manager/SeleniumManager.cs
```diff
@@ -168,49 +187,71 @@ public static class SeleniumManager
...
 
-    /// <summary>
-    /// Executes a process with the given arguments.
-    /// </summary>
-    /// <param name="arguments">The switches to be used by Selenium Manager.</param>
-    /// <returns>
-    /// the standard output of the execution.
-    /// </returns>
-    private static ResultResponse RunCommand(string arguments)
+    private static TResult RunCommand<TResult>(string arguments, JsonTypeInfo<TResult> jsonResultTypeInfo, TimeSpan? timeout = null)
```
RenderMichael, line 246: This is the only callsite for this method, unless I'm missing one. We can hard-code `DiscoveryResult` instead of using a generic parameter.

dotnet/src/webdriver/Manager/SeleniumManager.cs
nvborisenko: Seems correct.

dotnet/src/webdriver/Manager/SeleniumManager.cs
nvborisenko: Intentionally remained this method as `private`, most likely in future seleniummanager will support more commands, meaning `RunCommand` method will be reusable.

dotnet/src/webdriver/SeleniumManager.cs
nvborisenko: Done.

dotnet/src/webdriver/SeleniumManager.cs
```diff
@@ -168,49 +187,71 @@ public static class SeleniumManager
...
+            argsBuilder.AppendFormat(CultureInfo.InvariantCulture, " --proxy \"{0}\"", options?.Proxy);
+        }
+
         argsBuilder.Append(" --language-binding csharp");
-        argsBuilder.Append(" --output json");
+        argsBuilder.Append(" --output mixed");
+
         if (_logger.IsEnabled(LogEventLevel.Debug))
         {
             argsBuilder.Append(" --debug");
```
nvborisenko: Use `--log-level` switch. Not sure why, it was mentioned in the original requirements.

nvborisenko: Fixed.

dotnet/src/webdriver/SeleniumManager.cs
```diff
@@ -268,53 +308,127 @@ private static ResultResponse RunCommand(string arguments)
...
+                    var dateTime = DateTimeOffset.Parse(match.Groups[1].Value);
+                    var logLevel = match.Groups[2].Value;
+                    var message = match.Groups[3].Value;
+
+                    switch (logLevel)
+                    {
+                        case "INFO":
+                            _logger.LogMessage(dateTime, LogEventLevel.Info, message);
+                            break;
+                        case "WARNING":
```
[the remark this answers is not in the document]
nvborisenko: Good catch. Resolved.

dotnet/src/webdriver/Manager/SeleniumManager.cs
```diff
@@ -168,49 +187,73 @@ public static class SeleniumManager
...
     {
-        Process process = new Process();
-        process.StartInfo.FileName = _lazyBinaryFullPath.Value;
+        string smBinaryPath = _lazyBinaryFullPath.Value;
+
+        if (_logger.IsEnabled(LogEventLevel.Info))
+        {
+            _logger.Info($"Starting Selenium Manager process: {Path.GetFileName(smBinaryPath)} {arguments}");
```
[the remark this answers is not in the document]
nvborisenko: `Info` level is not enabled by default. Not an issue at all.

dotnet/src/webdriver/SeleniumManager.cs
```diff
@@ -168,49 +187,73 @@ public static class SeleniumManager
...
+            argsBuilder.Append(" --log-level trace");
+        }
+        else if (_logger.IsEnabled(LogEventLevel.Debug))
+        {
+            argsBuilder.Append(" --log-level debug");
         }
 
-        return binaryPaths;
+        return RunCommand(argsBuilder.ToString(), SeleniumManagerSerializerContext.Default.DiscoveryResult, options?.Timeout);
     }
```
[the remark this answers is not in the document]
nvborisenko: New method is nice!

dotnet/src/webdriver/Internal/Logging/Logger.cs
```diff
@@ -67,8 +67,13 @@ public bool IsEnabled(LogEventLevel level)
...
+    public void LogMessage(DateTimeOffset timestamp, LogEventLevel level, string message)
+    {
+        Log.CurrentContext.EmitMessage(this, timestamp.ToLocalTime(), level, message);
+    }
+
     private void LogMessage(LogEventLevel level, string message)
     {
-        Log.CurrentContext.EmitMessage(this, level, message);
+        LogMessage(DateTimeOffset.Now, level, message);
     }
```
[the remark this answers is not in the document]
nvborisenko: New api is covered implicitly. It is bad that good project is not able to use good native tools because of `b**l`.

dotnet/src/webdriver/SeleniumManager.cs
```diff
@@ -268,53 +311,127 @@ private static ResultResponse RunCommand(string arguments)
...
                 }
             }
         }
+    }
 
-        return jsonResponse.Result;
+    /// <summary>
+    /// Provides optional configuration for browser and driver discovery.
+    /// </summary>
+    public record DiscoveryOptions
```
nvborisenko: `DiscoveryOptions` class is nested here, what is not very good for public API. At the same time `SeleniumManager` class lives in the root namespace.

Options to make it better:
1. Move everything selenium manager related into new namespace
2. Move `DiscoveryOptions` into root namespace, but rename it to something more specific

nvborisenko: Renamed to `BrowserDiscoveryOptions` and `BrowserDiscoveryResult`, which is more specific and not nested.

dotnet/src/webdriver/Manager/SeleniumManager.cs
nvborisenko, line 1: I really propose to move Selenium Manager into separate namespace. In Java it is even separate package! I also see benefits having this as separate package, but for now let's shift it to namespace. Meaning in future it will be easier to isolate it as package.

nvborisenko:
```csahrp
namespace OpenQA.Selenium.Manager; // potential "Selenium.WebDriver.Manager" package

class SeleniumManager;
```

nvborisenko: Done.

dotnet/src/webdriver/Internal/Logging/LogContext.cs
```diff
@@ -97,14 +97,14 @@ public bool IsEnabled(ILogger logger, LogEventLevel level)
...
     }
 
-    public void EmitMessage(ILogger logger, LogEventLevel level, string message)
+    public void EmitMessage(ILogger logger, DateTimeOffset timestamp, LogEventLevel level, string message)
     {
         if (IsEnabled(logger, level))
         {
+            var logEvent = new LogEvent(logger.Issuer, timestamp, level, message);
+            
             message = TruncateMessage(message, _truncationLength);
```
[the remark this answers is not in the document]
nvborisenko: Thanks, good catch of incorrect merge.

dotnet/src/webdriver/Manager/SeleniumManager.cs
```diff
@@ -268,53 +311,127 @@ private static ResultResponse RunCommand(string arguments)
...
+                            _logger.LogMessage(dateTime, LogEventLevel.Trace, message);
+                            break;
+                    }
+                }
+                else
+                {
+                    errOutputBuilder.AppendLine(e.Data);
                 }
             }
         }
```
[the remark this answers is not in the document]
nvborisenko: Already tested implicitly.

dotnet/src/webdriver/Manager/SeleniumManager.cs
```diff
@@ -0,0 +1,443 @@
...
+                    exceptionMessageBuilder.AppendLine("--- End Error Output ---");
+                }
+
+                throw new WebDriverException(exceptionMessageBuilder.ToString());
+            }
+        }
+        catch (Exception ex)
+        {
+            throw new WebDriverException($"Error starting process: {process.StartInfo.FileName} {arguments}", ex);
+        }
```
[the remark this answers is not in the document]
nvborisenko: we put inner exception

dotnet/src/webdriver/Manager/SeleniumManager.cs
```diff
@@ -0,0 +1,443 @@
...
+/// <item><description>Caching drivers for subsequent use</description></item>
+/// <item><description>Providing paths to both driver and browser executables</description></item>
+/// </list>
+/// <para>
+/// The Selenium Manager binary is automatically included with the Selenium package.
+/// Set the SE_MANAGER_PATH environment variable to use a custom binary location.
+/// </para>
+/// </remarks>
+public static partial class SeleniumManager
+{
```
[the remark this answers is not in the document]
nvborisenko: normal

dotnet/src/webdriver/DriverFinder.cs
```diff
@@ -102,18 +106,29 @@ public bool TryGetBrowserPath([NotNullWhen(true)] out string? browserPath)
     /// <exception cref="NoSuchDriverException">If one of the paths does not exist.</exception>
     private Dictionary<string, string> BinaryPaths()
     {
-        if (paths.ContainsKey(SeleniumManager.DriverPathKey) && !string.IsNullOrWhiteSpace(paths[SeleniumManager.DriverPathKey]))
+        if (paths.TryGetValue(DriverPathKey, out string? cachedDriverPath) && !string.IsNullOrWhiteSpace(cachedDriverPath))
```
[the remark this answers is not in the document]
nvborisenko: I don't want to change the logic in scope of this PR

dotnet/src/webdriver/DriverFinder.cs
```diff
@@ -23,6 +23,7 @@
 using System.Globalization;
 using System.IO;
 using System.Text;
```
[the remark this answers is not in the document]
nvborisenko: Done

dotnet/src/webdriver/Manager/SeleniumManager.cs
```diff
@@ -0,0 +1,443 @@
...
+                        case "ERROR":
+                            _logger.LogMessage(dateTime, LogEventLevel.Error, message);
+                            break;
+                        case "DEBUG":
+                            _logger.LogMessage(dateTime, LogEventLevel.Debug, message);
+                            break;
+                        case "TRACE":
+                        default:
+                            _logger.LogMessage(dateTime, LogEventLevel.Trace, message);
+                            break;
```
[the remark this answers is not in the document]
nvborisenko: Done.

PR #17026, opened 2026-01-30
[py] Make extensions tests work when not running bazel

cgoldberg:
### 💥 What does this PR do?

This PR allows extensions tests to work when not running bazel.

Currently, there are 2 test files that use bazel Runfiles to locate web extensions when running tests. If you try to run these directly with pytest (without bazel), they will fail because Runfiles doesn't exist.

This PR adds a function to locate extensions in the local source tree when Runfiles isn't installed, and updates tests to use it for locating extensions.

This also fixes a subtle bug where the new `ruff-check.py` was giving incorrect results when finding import ordering issues.

It also skips `devtools` when running `ruff-check.py` and `ruff-format.py` from bazel.

### 🔄 Types of changes

- Dev/Test

### DIFF
--- a/py/private/ruff_check.py
+++ b/py/private/ruff_check.py
@@ -28,7 +28,7 @@ import sys
 from python.runfiles import Runfiles
 
 ALL_DIRS = ["py", "scripts", "common", "dotnet", "java", "javascript", "rb"]
-EXCLUDES = ["**/node_modules/**", "**/.bundle/**"]
+EXCLUDES = ["**/node_modules/**", "**/.bundle/**", "**/devtools/**"]
 
 
 def run_check(ruff, exclude_args, dirs, extra_args):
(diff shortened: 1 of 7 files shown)


py/private/ruff_check.py
```diff
@@ -28,7 +28,7 @@
 from python.runfiles import Runfiles
 
 ALL_DIRS = ["py", "scripts", "common", "dotnet", "java", "javascript", "rb"]
-EXCLUDES = ["**/node_modules/**", "**/.bundle/**"]
+EXCLUDES = ["**/node_modules/**", "**/.bundle/**", "**/devtools/**"]
```
navin772, line 31: Let's exclude `devtools` in `py/private/ruff_format.py` also.

navin772 (approved):
LGTM! 
Nice work

PR #17028, opened 2026-01-30
[java][BiDi] implement `emulation.setScreenSettingsOverride`

Delta456:
### 💥 What does this PR do?

Implements `emulation.setScreenSettingsOverride` in Java from the [BiDi W3C spec](https://w3c.github.io/webdriver-bidi/#command-emulation-setScreenSettingsOverride)

### 🔄 Types of changes


java/src/org/openqa/selenium/bidi/emulation/SetScreenSettingsOverrideParameters.java
```diff
@@ -0,0 +1,57 @@
...
+
+package org.openqa.selenium.bidi.emulation;
+
+import java.util.List;
+import java.util.Map;
+
+public class SetScreenSettingsOverrideParameters extends AbstractOverrideParameters {
+  int height, width;
+
+  public SetScreenSettingsOverrideParameters(Map<String, Integer> screenArea) {
```
asolntsev: Sorry, I didn't understand why we pass width+height as a map?
Why don't we just have `int width` and `int height` parameters?

```java
  public SetScreenSettingsOverrideParameters(int width, int height) {
      this.width = width;
      this.height = height;
  }
```

asolntsev (changes requested):

java/src/org/openqa/selenium/bidi/emulation/SetScreenSettingsOverrideParameters.java
Delta456: Yes, that is indeed better. I was manuallly testing the code and forgot to change it :sweat_smile:

Delta456: As per W3C Spec, `screenArea: emulation.ScreenArea / null`. `screenArea` can have `null` values. 

For convenience, we can another constructor so people can provide via map/null and width/height separately.

asolntsev: If you want to allow null values, just mark them as nullable:

```java
SetScreenSettingsOverrideParameters(@Nullable Integer width, @Nullable Integer height) {
```

This is the natural way. 

Constructor with **Map is not needed** in any case.

asolntsev (changes requested):

java/src/org/openqa/selenium/bidi/emulation/SetScreenSettingsOverrideParameters.java
Delta456: would `SetScreenSettingsOverrideParameters(null, null)` be a good idea? I think not. In emulation, we already do `SetXXXOverrideParameters(null)`; for this case, it would be consistent. If you have a better idea, let me know.

asolntsev: Ok, now I understand. You didn't mean that `width` or `height` can be null. You meant that the Map can be null.

Then we just need to create a separate object `ScreenArea` instead of Map:

```java
public class ScreenArea {
  public final int width;
  public final int height;
}

public class SetScreenSettingsOverrideParameters extends AbstractOverrideParameters {
  @Nullable
  private final ScreenArea screenArea;

  public SetScreenSettingsOverrideParameters(@Nullable ScreenArea screenArea) { ... }
```


Then it will be consistent with other similar classes:

```java
public SetGeolocationOverrideParameters(GeolocationCoordinates coordinates) { ... }
public SetScreenOrientationOverrideParameters(ScreenOrientation screenOrientation) { ... }
```

java/src/org/openqa/selenium/bidi/emulation/Emulation.java
```diff
@@ -73,4 +73,10 @@ public void setNetworkConditions(SetNetworkConditionsParameters parameters) {
 
     bidi.send(new Command<>("emulation.setNetworkConditions", parameters.toMap(), Map.class));
   }
+
+  public void setScreenSettingsOverride(SetScreenSettingsOverrideParameters parameters) {
+    Require.nonNull("SetScreenSettingsOverride parameters", parameters);
+
+    bidi.send(new Command<>("emulation.setScreenSettingsOverride", parameters.toMap(), Map.class));
```
asolntsev, line 80: I think `Map.class` is not needed here.
This command returns empty result, the method return type is void.

asolntsev (approved):
@Delta456  all good except one small suggestion about unneeded `Map.class` parameter.

java/src/org/openqa/selenium/bidi/emulation/Emulation.java
Delta456: Absolutely. I feel its better done in another PR as another emulation commands have the same repetition.

PR #17032, opened 2026-01-31
[py]: enable edge browser tests for bidi

navin772:
### 💥 What does this PR do?

Currently we are not running the bidi tests on edge browser in any of the CI for python. Sometimes there are differences between edge and chrome in some bidi modules, so its important to run the bidi tests.

This does runs other non-bidi tests under the bidi target like - `//py:test/selenium/webdriver/common/window_tests-edge-bidi` so there is repeitition as with other browsers. We should probably fix it so these tests run only once per browser.

This will run these tests in CI-RBE only, we don't want to run them as integration/browsers tests CI to avoid unnecessary duplication.

### 🔄 Types of changes

- Tests


cgoldberg (approved):
LGTM

Issue #15265, opened 2025-02-10
[🚀 Feature]: Unifying Select Class Across All Bindings


shbenzer:
### Feature and motivation

User @FFederi noticed that the `Select` class is not consistent across language bindings in #15135. 

I'm creating this ticket so we can discuss and track making this consistent

### Usage example

In Java:

`selectByVisibleText()` and `selectByContainsVisibleText()` should _both_ check properties "visibility", "display", and "opacity", for values of "hidden", "none", "0", "0.0".

### Tracking

- [x] Python #15135 
- [x] Java #16220
- [ ] Dotnet
- [ ] Ruby
- [ ] Javascript

VertexXX6:
### I've created a small test with Java and HTML to demonstrate the visibility issue.
The test checks if selectByVisibleText() skips options hidden using CSS (display: none, visibility: hidden, opacity: 0).
Below is my test code that shows the issue. I hope it can help in fixing the problem.

****
```html
<!DOCTYPE html>
<html>
<head>
    <style>
        #hiddenDisplay { display: none; }
        #hiddenVisibility { visibility: hidden; }
        #hiddenOpacity { opacity: 0; }
    </style>
</head>
<body>
    <select id="mySelect">
        <option value="1">Visible Option</option>
        <option id="hiddenDisplay" value="2">Hidden Option (display:none)</option>
        <option id="hiddenVisibility" value="3">Hidden Option (visibility:hidden)</option>
        <option id="hiddenOpacity" value="4">Hidden Option (opacity:0)</option>
    </select>
</body>
</html>
```

**java code**




```java
package org.example;

import io.github.bonigarcia.wdm.WebDriverManager;
import org.openqa.selenium.By;
import org.openqa.selenium.NoSuchElementException;
import org.openqa.selenium.WebDriver;
import org.openqa.selenium.WebElement;
import org.openqa.selenium.chrome.ChromeDriver;
import org.openqa.selenium.chrome.ChromeOptions;
import org.openqa.selenium.support.ui.Select;
import java.util.List;
import java.util.Scanner;

public class SelectHiddenOptionsTest {
    public static void main(String[] args) {
        WebDriverManager.chromedriver().setup();
        ChromeOptions options = new ChromeOptions();
        options.addArguments("--remote-allow-origins=*");
        WebDriver driver = new ChromeDriver(options);
        try {
            driver.get("file:///C:/Users/dell/OneDrive/Desktop/test_select.html");
[...]

PR #16220, opened 2025-08-20
[java] Unifying select class

vicky-iv:
### 🔗 Related Issues
[🚀 Feature]: Unifying Select Class Across All Bindings #15265 

### 💥 What does this PR do?
I implemented unifying the Select class across all bindings for the Java package.
I also extended the unit tests to cover this functionality by shadowing PR for python #15135 

### 🔧 Implementation Notes
The `selectByContainsVisibleText` method already had the necessary implementation, so I just added the missing exception throwing for the `selectByVisibleText` method.

All tests from the following packages are passed locally: 
` //java/test/org/openqa/selenium/support/ui/...  `

### 💡 Additional Considerations
No

### 🔄 Types of changes
Bug fix (backwards compatible)

___

### **PR Type**
Bug fix

___

### **Description**
- Fix `selectByVisibleText` to throw exception for hidden options

- Add visibility checks to prevent selecting invisible elements

- Enhance test coverage for hidden option scenarios

- Update Java language level to JDK 17

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["selectByVisibleText method"] --> B["Add visibility check"]
  B --> C["Throw NoSuchElementException"]
  C --> D["Enhanced test coverage"]
  E["Java language level"] --> F["Update to JDK 17"]
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>Select.java</strong><dd><code>Add visibility validation to selectByVisibleText</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

java/src/org/openqa/selenium/support/ui/Select.java

<ul><li>Add <code>assertSelectIsVisible()</code> call to <code>selectByVisibleText</code> method<br> <li> Add visibility checks using <code>hasCssPropertyAndVisible()</code> for options<br> <li> Throw <code>NoSuchElementException</code> when trying to select invisible options</ul>

</details>

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/16220/files#diff-255eaa03e269acb38bc85bd146a129dd90e1ae3e20d94fc6de98def15cc4e59f">+5/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
[...]

PR #17037, opened 2026-02-01
[rb] fix select being able to select options hidden by css rules

FFederi:
### 🔗 Related Issues
Related to issue: [[🚀 Feature]: Unifying Select Class Across All Bindings](https://github.com/SeleniumHQ/selenium/issues/15265)

### 💥 What does this PR do?
Modifies select behaviour to make it the same as python and Java bindings.
In particular, makes it so selecting an option hidden by CSS rules raises an exception.

### 🔧 Implementation Notes
I couldn't find (or understand) a general way to check visibility of an element, so I added a simple check on CSS values like it's been done with Java and python.

### 🔄 Types of changes

### DIFF
--- a/rb/lib/selenium/webdriver/support/select.rb
+++ b/rb/lib/selenium/webdriver/support/select.rb
@@ -164,6 +164,9 @@ module Selenium
         private
 
         def select_by_text(text)
+          assert_select_enabled
+          assert_select_visible
+
           opts = find_by_text text
 
           raise Error::NoSuchElementError, "cannot locate element with text: #{text.inspect}" if opts.empty?
@@ -174,6 +177,20 @@ module Selenium
             select_option(opt)
             break unless multiple?
           end
+        end
+
+        def assert_select_enabled
+          return if @element.enabled?
+
+          raise Error::UnsupportedOperationError,
+                'You may not select an option in a disabled select'
+        end
+
+        def assert_select_visible
+          return if option_visible?(@element)
+
+          raise Error::UnsupportedOperationError,
+                'You may not select an option in an invisible select'
         end
 
         def select_by_index(index)
--- a/rb/spec/unit/selenium/webdriver/support/select_spec.rb
+++ b/rb/spec/unit/selenium/webdriver/support/select_spec.rb
@@ -24,14 +24,16 @@ module Selenium
     module Support
       describe Select do
         let(:select) do
-          select_element = instance_double(Element, tag_name: 'select')
+          select_element = instance_double(Element, tag_name: 'select', enabled?: true)
           allow(select_element).to receive(:dom_attribute).with(:multiple)
+          allow(select_element).to receive(:css_value).and_return('')
           select_element
         end
 
         let(:multi_select) do
-          select_element = instance_double(Element, tag_name: 'select')
+          select_element = instance_double(Element, tag_name: 'select', enabled?: true)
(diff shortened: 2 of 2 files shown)


rb/lib/selenium/webdriver/support/select.rb
```diff
@@ -217,10 +217,20 @@ def deselect_by_index(index)
         def select_option(option)
           raise Error::UnsupportedOperationError, 'You may not select a disabled option' unless option.enabled?
 
+          unless css_property_and_visible?(option)
+            raise Error::ElementNotInteractableError,
```
aguspe, line 221: Java's #16220 raises `NoSuchElementException` here, I think we should make it the same as Java or put it in the PR just for visibility

rb/lib/selenium/webdriver/support/select.rb
```diff
@@ -266,6 +276,15 @@ def find_by_index(index)
         def find_by_value(value)
           @element.find_elements(xpath: ".//option[@value = #{Escaper.escape value}]")
         end
+
+        def css_property_and_visible?(element)
+          css_value_candidates = %w[hidden none 0 0.0].to_set
```
aguspe, line 281: the name hides that it returns `true` when visible, rename to `option_visible?` maybe? Btw could we lift the candidate lists to frozen constants so they're not rebuilt each call

rb/lib/selenium/webdriver/support/select.rb
```diff
@@ -217,10 +217,20 @@ def deselect_by_index(index)
...
+          unless css_property_and_visible?(option)
+            raise Error::ElementNotInteractableError,
+                  'You may not select an invisible option'
+          end
+
           option.click unless option.selected?
         end
 
         def deselect_option(option)
+          unless css_property_and_visible?(option)
```
aguspe, line 229: `select_option` checks `enabled?` but `deselect_option` doesn't, pre-existing, just flagging also do we actually need the invisibility check on deselect, or keep it select-only like Java

rb/spec/integration/selenium/webdriver/select_spec.rb
```diff
@@ -120,6 +121,14 @@ module Support
               it 'errors when not found' do
                 expect { multi_select.select_by(:text, 'invalid') }.to raise_exception(Error::NoSuchElementError)
               end
+
+              it 'errors when option is invisible', :aggregate_failures do
```
aguspe, line 125: Since the check now lives in `select_option`/`deselect_option`, want to add `:value` and `:index` cases too?

FFederi:
@aguspe 
Sorry for the delay, thanks for your review!
I tried addressing the points you raised trying to prioritizing making the logic similar to Java's, what do you think?

aguspe:
> @aguspe Sorry for the delay, thanks for your review! I tried addressing the points you raised trying to prioritizing making the logic similar to Java's, what do you think?

@FFederi thank you for the awesome work! I just have one small comment in Java selectByVisibleText also calls assertSelectIsEnabled() and assertSelectIsVisible() on the parent select before iterating options, want to add the same here?

FFederi:
@aguspe 
> I just have one small comment in Java selectByVisibleText also calls assertSelectIsEnabled() and assertSelectIsVisible() on the parent select before iterating options, want to add the same here?

Sorry for the back and forth, I addressed this on [b0a4c10](https://github.com/SeleniumHQ/selenium/pull/17037/commits/b0a4c10527901850d0ffe6d4f942459366cc3037) !

PR #17058, opened 2026-02-07
[java] specify nullability in package `org.openqa.selenium.bidi.module`

asolntsev:
### 💥 What does this PR do?
Marks the whole package `org.openqa.selenium.bidi.module` as non-nullable by default.

### 🔄 Types of changes

### DIFF
--- a/java/src/org/openqa/selenium/bidi/module/BUILD.bazel
+++ b/java/src/org/openqa/selenium/bidi/module/BUILD.bazel
@@ -30,5 +30,6 @@ java_library(
         "//java/src/org/openqa/selenium/json",
         "//java/src/org/openqa/selenium/remote/http",
         artifact("com.google.auto.service:auto-service-annotations"),
+        "@maven//:org_jspecify_jspecify",
     ],
 )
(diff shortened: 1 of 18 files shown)


cgoldberg:
@asolntsev try to remember to add "[java]" to your PR titles :)

asolntsev:
> @asolntsev try to remember to add "[java]" to your PR titles :)

Actually, last time we found out that it doesn't affect anything.
As you see, GA added label "c-java" without having this prefix:

image

cgoldberg:
It just a convention we use that makes it easier to scan the list of PRs and see which they belong to (you don't see labels in GitHub notifications). It also ends up in the commit message when you squash if you don't change it. Every single contributor adds them except the PRs coming from you.

asolntsev:
Yes, I agree with this argument:
> It also ends up in the commit message when you squash if you don't change it

(though, I shudder at the thought that there might be developers in the world who don't edit their commit message when squashing).


But this:
> Every single contributor adds them except the PRs coming from you.

sounds a bit harsh to me. I accept logical arguments. I don't want to do things without knowing the reason just because everyone else does it.

PR #17079, opened 2026-02-11
[javascript] Add headless flag to internal Chrome tests

ElliotNewtonMiller:
### 💥 What does this PR do?

Make internal javascript testing on Chrome headless because Chromedriver will otherwise not connect with Chrome.
This change was implimented in the internal testing `Environment` object.

### 🔧 Implementation Notes

I used the Environment object in the "testing/index.js" file because it seemed to be solely used for internal testing (in Bazel), which is the only place this bug exists.

I considered trying to add environment flags via Bazel commands.

The tests don't all pass. Hopefully fixing them can be my focus in future PRs.

### 🔄 Types of changes

### DIFF
--- a/javascript/selenium-webdriver/testing/index.js
+++ b/javascript/selenium-webdriver/testing/index.js
@@ -343,6 +343,16 @@ class Environment {
         builder.setCapability('moz:debuggerAddress', true)
       }
 
+      // Necessary to connect the driver to Chrome browser from within Bazel.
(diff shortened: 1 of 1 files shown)


cgoldberg:
Why is this needed? Isn't there already a `--headless` option when calling bazel that works with all browsers?

cgoldberg (changes requested):
There is an option when running bazel that sets headless for any browser (`--headless`). If that doesn't work in JS, it should be fixed. This PR forces headless (for Chrome only), which is not what we want.

PR #17094, opened 2026-02-17
[dotnet] Implicit global usings

nvborisenko:
Simplify codebase via importing namespaces globally.

### 💥 What does this PR do?
- Introduced new rule: Remove unnecessary using directives - `warning` level

### 🔄 Types of changes


dotnet/src/support/Properties/GlobalUsings.cs
```diff
@@ -0,0 +1,24 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+// </copyright>
+
+global using System;
```
nvborisenko: `global::System`

And in the second file too

RenderMichael (approved):

PR #17117, opened 2026-02-20
[py] Add return type annotation to execute_script and execute_async_script

adamtheturtle:
## Summary
- Add `-> Any` return type annotation to `WebDriver.execute_script`
- Add `-> Any` return type annotation to `WebDriver.execute_async_script`

## Motivation
`execute_script` returns the result of the executed JavaScript, which can be any type (string, number, bool, list, dict, None, WebElement, etc.). Adding an `Any` return type annotation improves IDE support and documents this behavior.

## Test plan
- No behavioral changes; annotation-only change

### DIFF
--- a/py/selenium/webdriver/remote/webdriver.py
+++ b/py/selenium/webdriver/remote/webdriver.py
@@ -506,7 +506,7 @@ class WebDriver(BaseWebDriver):
         """
         return list(self.pinned_scripts)
 
-    def execute_script(self, script: str, *args) -> dict[str, Any]:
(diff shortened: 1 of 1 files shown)


cgoldberg:
Thanks.. I just pushed a change to your branch that also fixes the return type for `execute_async_script`.. it was also incorrect.

I'll merge this after CI runs.

cgoldberg:
CI is failing for an unrelated reason.. I'll update this branch and run it again in a little while once we get it fixed.

PR #17130, opened 2026-02-23
[java][BiDi] implement `speculation` module

Delta456:
### 💥 What does this PR do?

This PR implements an external BiDi [nav speculation](https://wicg.github.io/nav-speculation/prefetch.html#speculation-module) for Java binding and adds [speculation.PrefetchStatusUpdated](https://wicg.github.io/nav-speculation/prefetch.html#cddl-type-speculationprefetchstatusupdated).

The speculation module contains commands for managing the remote end behavior for prefetches, prerenders, and speculation rules.

### 🔧 Implementation Notes

Took tests from https://wpt.fyi/results/webdriver/tests/bidi/external/speculation?label=master&label=stable&aligned

### 🔄 Types of changes


java/src/org/openqa/selenium/bidi/module/SpeculationInspector.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+
+  public void removeListener(long subscriptionId) {
+    this.bidi.removeListener(subscriptionId);
+  }
+
+  @Override
+  public void close() {
+    this.bidi.clearListener(Speculation.prefetchStatusUpdated());
```
asolntsev, line 74: @Delta456 Two questions:
1. Don't we want to use variable `this.prefetchStatusUpdatedEvent ` instead of calling `Speculation.prefetchStatusUpdated()`?
2. Is it ok that we add listener with a specific `browsingContextId`, but clear listeners without specifying `browsingContextId`?

asolntsev (changes requested):
@Delta456 Looks good, only few questions in https://github.com/SeleniumHQ/selenium/pull/17130/changes#r2847114237

java/src/org/openqa/selenium/bidi/module/SpeculationInspector.java
Delta456: 1. `Speculation.prefetchStatusUpdated()` sounds more name scoped and function oriented
2. We should definitely specific `browserContextId` for clearing listeners. I will add that.

Delta456: Just checked, `LogInspector` works the same way. It closes without specifying `browserContextId`.

For consistency, we should keep it that way.

asolntsev: > Speculation.prefetchStatusUpdated() sounds more name scoped and function oriented
Sorry, I don't understand this argument. 
Then why do we have the field `this.prefetchStatusUpdatedEvent` at all?
Why we use field in some cases, but new instance in other cases?

> For consistency, we should keep it that way.

Sounds strange for me... Like "we had one bug - no we have two bugs _for consistency_".
Well, maybe it's not urgent. Seems like nobody uses `LogInspector` with multiple contexts...

Delta456: We can resolve this bug in a later PR with `LogInspector`. That should work.

asolntsev (approved):

PR #17133, opened 2026-02-23
[py] Create module for importing latest devtools

cgoldberg:
### 💥 What does this PR do?

This PR creates a `selenium.webdriver.common.devtools.latest` module that points to the latest devtools module.

It is generated by bazel during build and simply consists of a an `__init.__.py` that does a wildcard import of the latest devtools.

i.e.:

```
from ..v145 import *
```

This allows users to import the latest devtools modules with:

```
from selenium.webdriver.common.devtools.latest import network
```

instead of always specifying the version number in imports:

```
from selenium.webdriver.common.devtools.v145 import network
```

### 🔄 Types of changes


AutomatedTester:
We need to be aware that using latest may lead to breaking changes as CDP is not stable

cgoldberg:
> We need to be aware that using latest may lead to breaking changes as CDP is not stable

yea.. using "latest" could lead to code breakage if there are changes in CDP.

However, if you pin your imports to a specific version of devtools and upgrade selenium, you could also have issues with having an incompatible CDP version.

Users would have to be aware of the tradeoffs either way... I think adding this just gives some flexibility. Specifying a version in imports still works as it did previously.

navin772 (approved):

PR #17142, opened 2026-02-26
[dotnet] [bidi] Wait until events are dispatched when unsubscribing

### LINKED ISSUE #16574: [🐛 Bug]: [dotnet] BiDi logs potentially missing with race condition
### Description

It appears that not all console logs are sent over the BiDi web socket and there is some kind of race condition when closing down the connection.

If I want to make sure that I receive all the log entries before the browser closes, it appears that I need to add a time based delay, which results in flaky and/or slow tests.

I've attached a simple example that demonstrates the logs are not send to `OnEntryAddedAsync` unless I add a time-based delay.

Am I doing something wrong here?

### Reproducible Code

```shell
//     <PackageReference Include="Selenium.WebDriver" Version="4.38.0" />

using System.Collections.Concurrent;
using OpenQA.Selenium.BiDi;
using OpenQA.Selenium.Chrome;

var options = new ChromeOptions { UseWebSocketUrl = true };
using var driver = new ChromeDriver(options);
await using var biDi = await driver.AsBiDiAsync();
var logs = new ConcurrentBag<string?>();
await using var logSubscription = await biDi.Log.OnEntryAddedAsync(entry => logs.Add(entry.Text));

driver.Navigate().GoToUrl("about:blank");
driver.ExecuteScript("console.log('Test log.');");

//Uncommenting this line will "fix" the issue.
//await Task.Delay(1000);

if(!logs.SequenceEqual(["Test log."]))
    throw new Exception("Logs missing.");
Console.WriteLine("Logs present.");
```

nvborisenko:
### 🔗 Related Issues
Fix https://github.com/SeleniumHQ/selenium/issues/16574

### 💥 What does this PR do?
Refactors the event dispatching mechanism in the `EventDispatcher` class to improve thread safety, reliability, and event handler management. The changes introduce a new `EventItem` structure, enhance handler registration/removal, and ensure that all pending events are processed before unsubscribing handlers. Error handling and concurrency controls have also been improved.

### 🔄 Types of changes

### DIFF
--- a/dotnet/src/webdriver/BiDi/EventDispatcher.cs
+++ b/dotnet/src/webdriver/BiDi/EventDispatcher.cs
@@ -42,12 +42,10 @@ internal sealed class EventDispatcher : IAsyncDisposable
 
     private readonly Task _eventEmitterTask;
 
-    private static readonly TaskFactory _myTaskFactory = new(CancellationToken.None, TaskCreationOptions.DenyChildAttach, TaskContinuationOptions.None, TaskScheduler.Default);
-
     public EventDispatcher(Func<ISessionModule> sessionProvider)
     {
         _sessionProvider = sessionProvider;
-        _eventEmitterTask = _myTaskFactory.StartNew(ProcessEventsAwaiterAsync).Unwrap();
+        _eventEmitterTask = Task.Run(ProcessEventsAwaiterAsync);
     }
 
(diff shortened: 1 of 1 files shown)


nvborisenko:
Seems overcomplicated, but it resolves simple behaviour:

```csharp
await subscription.DisposeAsync(); // waiting until events in the queue are dispatched
```

Can be simplified if https://github.com/w3c/webdriver-bidi/issues/1032 is resolved via `subscriptionId` in all events.

PR #17146, opened 2026-02-27
[grid] Router bypass WebSocket data path via transparent TCP tunnel

VietND96:
### 🔗 Related Issues
The Grid Router was proxying all WebSocket (e.g. BiDi) connections at the application layer: it deserialized every frame through the full Netty HTTP/WS pipeline, passed it through the Selenium handler chain,
  and re-serialized it on the way out. This adds CPU overhead and latency on every message for the entire lifetime of a session - a significant cost for real-time bidirectional protocols.

### 💥 What does this PR do?
Add a transparent TCP tunnel in the Netty pipeline. When the Router receives a WebSocket upgrade for a known session, instead of going through `ProxyWebsocketsIntoGrid`, it:

  1. Resolves the target Node URI from the session map using the session ID in the request path
  2. Opens a raw TCP connection to the Node on the same Netty event loop (no extra threads)
  3. Forwards the original HTTP upgrade request to the Node
  4. On HTTP 101 from the Node, sends the 101 back to the client, rewires both channels into `TcpTunnelHandler` (raw byte forwarder), and strips all HTTP/WS codec handlers from the client pipeline
  5. After setup the Router is completely out of the data path — bytes flow directly between client and Node

  Robustness for complex deployments

  Two additional fixes ensure the tunnel degrades gracefully in complex network topologies (Docker, Kubernetes, cloud VPC):

  Graceful fallback on connect failure - In the original implementation, if the raw TCP connect to the Node failed (e.g. Node is unreachable due to K8s port-forward topology, split-network/NAT, or a transient failure), the client channel was simply closed. Now the original `HttpRequest` is fired back through the pipeline via `ctx.fireChannelRead(req)`, allowing `ProxyWebsocketsIntoGrid` to attempt the connection via the Grid's `HttpClient.Factory` (which respects configured HTTP proxies and other NetworkOptions settings). The default port for https:// Node URIs is also fixed from 80 to 443.

  TLS support for HTTPS nodes — The original bootstrap used a plain `NioSocketChannel` with no SSL handler, so any Node registered with an https:// URI would fail. Now when the Node URI scheme is https, a
  lazily-initialised client SslContext (built with `InsecureTrustManagerFactory`, appropriate for internal Grid traffic where nodes commonly use self-signed certificates) is added as the first handler in the
  node-side bootstrap pipeline. The SSL handler is intentionally left in place after the tunnel is established — it transparently handles TLS framing for the raw byte stream.

  Fallback behaviour summary

  The tunnel is bypassed (falls through to `ProxyWebsocketsIntoGrid`) when any of the following are true:
  - Request is not a WebSocket upgrade
  - URI contains no extractable Selenium session ID
  - Session is not found in the SessionMap
  - TCP connect to the Node fails (unreachable network, K8s port-forward, etc.)
  - SSL context initialisation fails (rare JVM security policy issue)

### 🔄 Types of changes


java/src/org/openqa/selenium/netty/server/TcpUpgradeTunnelHandler.java
```diff
@@ -0,0 +1,330 @@
...
+        tunnelEstablished = true;
+        Channel nodeChannel = ctx.channel();
+
+        // Build a proper Netty HTTP 101 response, copying all headers from the Node's response.
+        // Writing a DefaultFullHttpResponse goes through HttpResponseEncoder, which correctly
+        // encodes it, and the HttpServerKeepAliveHandler does not close the channel for 101.
+        DefaultFullHttpResponse clientResponse =
+            new DefaultFullHttpResponse(
+                HttpVersion.HTTP_1_1,
+                HttpResponseStatus.SWITCHING_PROTOCOLS,
```
vuh71029-web: ###Fixes #17146

PR #17147, opened 2026-02-27
[java] Deduplicate Unicode PUA mappings in Keys; make OPTION an alias of ALT and deprecate FN

seethinajayadileep:
## 💥 What does this PR do?

Removes duplicate Unicode Private Use Area (PUA) mappings in `org.openqa.selenium.Keys`.

- Updates `OPTION` to be an alias of `ALT` using `OPTION(Keys.ALT)` instead of duplicating `RIGHT_ALT` (`\uE052`).
- Converts `FN` into a deprecated alias of `RIGHT_CONTROL` using `FN(Keys.RIGHT_CONTROL)` instead of directly mapping `\uE051`.
- Updates Javadoc to reflect the change.

---

## 🔧 Implementation Notes

`Keys.getKeyFromUnicode()` returns the first matching enum value. When multiple enum constants share the same Unicode PUA value:

- The lookup becomes ambiguous.
- One constant becomes unreachable.
- Behavior becomes order-dependent.

This change ensures that each Unicode PUA value maps to a single canonical key.

`OPTION` remains a platform-friendly alias without introducing a duplicate Unicode value.

`FN` is retained for backward compatibility but deprecated. It now delegates to `RIGHT_CONTROL` instead of defining its own duplicate Unicode value.

---

## 💡 Additional Considerations

- No breaking API changes.
- No duplicate Unicode values remain.
- Backward compatibility preserved.

---

## 🔄 Types of changes

### DIFF
--- a/common/repositories.bzl
+++ b/common/repositories.bzl
@@ -50,8 +50,8 @@ js_library(
 
     http_archive(
         name = "linux_beta_firefox",
-        url = "https://ftp.mozilla.org/pub/firefox/releases/148.0b15/linux-x86_64/en-US/firefox-148.0b15.tar.xz",
-        sha256 = "23621cf9537fd8d52d3a4e83bba48984f705facd4c10819aa07ae2531e11e2e5",
+        url = "https://ftp.mozilla.org/pub/firefox/releases/149.0b2/linux-x86_64/en-US/firefox-149.0b2.tar.xz",
+        sha256 = "14057fe24b65ef64125d04bde3507f48e489464b9f22cda832d34db92dede817",
         build_file_content = """
 load("@aspect_rules_js//js:defs.bzl", "js_library")
 package(default_visibility = ["//visibility:public"])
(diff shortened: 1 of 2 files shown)


asolntsev (approved):
Good for me.

@seethinajayadileep 
Isn't there a risk that somebody is already using constant `Keys.FN`? 
What would be a workaround for them?

seethinajayadileep:
> Good for me.
> 
> @seethinajayadileep Isn't there a risk that somebody is already using constant `Keys.FN`? What would be a workaround for them?

You’re right — removing Keys.FN would be a source-breaking change for anyone already referencing it.

Even though FN is not part of the W3C WebDriver specification and doesn’t have reliable driver support, existing code may still compile against it.

A safer approach would be to keep FN for backward compatibility and mark it as @Deprecated, documenting that it is symbolic only and not part of the standardized key set.

That way existing users are not broken, but future usage is discouraged. I can update the PR accordingly.

asolntsev (approved):

seethinajayadileep:
The current failures appear to be timeout/browser-related (RBE + Ruby Windows) and unrelated to the Keys changes in this PR.
I don’t have permissions to rerun the jobs.
Could  please rerun the failed checks? If they pass, this should be ready to merge

seethinajayadileep:
The rerun is now failing across multiple Firefox beta integration targets (Java AlertsTest-firefox-beta and Ruby target_locator-firefox-beta / BiDi).

These failures span both Java and Ruby suites and don’t relate to the Keys.java change in this PR.

This looks like a Firefox beta CI issue. Could a maintainer please confirm or advise how to proceed?

asolntsev:
@seethinajayadileep Yes, it seems to be some generic issue, it's failing in all PRs. 
So, I've decided to ignore the failures and merge the PRs.

PR #17149, opened 2026-02-28
[java] remove `@Nullable` from return value for `ExpectedConditions` that never return null

### LINKED ISSUE #17122: [🐛 Bug]: [java] `FluentWait.until(ExpectedConditions.…(…))` cannot be called from Kotlin
### Description

Image

This used to compile in 4.39.0, then in 4.40.0 the annotations were added and it triggered this problem: https://github.com/SeleniumHQ/selenium/issues/16970, the fix to that problem looks like it caused this one. Note: I acknowledge this might be a Kotlin compiler issue, I can't decide. Happy to forward the issue if you can help me figure out what's causing it that you can't help with some more/less annotations.

The problem on `.until` can be "fixed" by `.until<WebElement>(…)`, but that shouldn't be needed either. For the parameter to `until` I found no workaround other than lying about the type:
`.until(ExpectedConditions.…(…) as ExpectedCondition<WebElement>)` + `@Suppress("UNCHECKED_CAST")`.

### Reproducible Code

```shell
fun WebDriver.wait(): WebDriverWait =
	WebDriverWait(this, 30.seconds.toJavaDuration())

fun WebDriver.delayedExecute(locator: By, action: (WebElement) -> Unit) {
	val element = this.wait().until(ExpectedConditions.presenceOfElementLocated(locator))
	this.wait().until(ExpectedConditions.visibilityOf(element))
	action(element)
}
```

### Debugging Logs

```logs
e: file:///.../wait.kt:38:28 [CANNOT_INFER_PARAMETER_TYPE] Cannot infer type for type parameter 'V'. Specify it explicitly.
e: file:///.../wait.kt:38:34 [ARGUMENT_TYPE_MISMATCH] Argument type mismatch: actual type is 'ExpectedCondition<WebElement?>', but 'Function<in WebDriver, uninferred V! (of fun <V : Any> until)>' was expected.
e: file:///.../wait.kt:39:14 [CANNOT_INFER_PARAMETER_TYPE] Cannot infer type for type parameter 'V'. Specify it explicitly.
e: file:///.../wait.kt:39:20 [ARGUMENT_TYPE_MISMATCH] Argument type mismatch: actual type is 'ExpectedCondition<WebElement?>', but 'Function<in WebDriver, uninferred V! (of fun <V : Any> until)>' was expected.
```

---
ℹ️ **Last known working version:** `4.40.0`

asolntsev:
### 🔗 Related Issues
Fixes #17122

### 💥 What does this PR do?
1. Removes `@Nullable` annotation from several return value for `ExpectedConditions` that never return null
2. Marks <V> type of Function both nullable and non-nullable (using a crazy trick `<V extends @Nullable Object>`).

### 🔧 Implementation Notes
1. Some implementations of `ExpectedConditions` can return null. 
3. But expression `wait.until(ExpectedConditions.visibilityOf(...))` never returns null. It either returns Object or true, or throws `TimeoutException`.

Probably we should create a test suite in Kotlin, so that nullability problems arise earlier. 

### 🔄 Types of changes
- Bug fix

### Backward incompatible change #1

Signature of public method `Wait.until` and `FluentWait.until` was changed.

Before:
```java
  @Override
  public <V> V until(Function<? super T, @Nullable V> isTrue) { ... }
```

After:
```java
  @Override
  public <V extends @Nullable Object> @NonNull V until(Function<? super T, ? extends V> isTrue) { ... }
```

### Backward incompatible change #2

Built-in conditions `ExpectedConditions.*` now return non-nullable value.

### DIFF
--- a/java/src/org/openqa/selenium/support/ui/ExpectedConditions.java
+++ b/java/src/org/openqa/selenium/support/ui/ExpectedConditions.java
@@ -247,8 +247,8 @@ public class ExpectedConditions {
    */
   public static ExpectedCondition<@Nullable List<WebElement>> visibilityOfAllElementsLocatedBy(
       final By locator) {
-    return new ExpectedCondition<@Nullable List<WebElement>>() {
-      private int indexOfInvisibleElement;
+    return new ExpectedCondition<>() {
+      private int indexOfInvisibleElement = -1;
       private @Nullable WebElement invisibleElement;
 
       @Override
@@ -304,8 +304,8 @@ public class ExpectedConditions {
    */
   public static ExpectedCondition<@Nullable List<WebElement>> visibilityOfAllElements(
       final List<WebElement> elements) {
-    return new ExpectedCondition<@Nullable List<WebElement>>() {
-      private int indexOfInvisibleElement;
+    return new ExpectedCondition<>() {
+      private int indexOfInvisibleElement = -1;
       private @Nullable WebElement invisibleElement;
 
       @Override
@@ -345,7 +345,7 @@ public class ExpectedConditions {
    * @return the (same) WebElement once it is visible
    */
   public static ExpectedCondition<@Nullable WebElement> visibilityOf(final WebElement element) {
-    return new ExpectedCondition<@Nullable WebElement>() {
+    return new ExpectedCondition<>() {
       @Override
       public @Nullable WebElement apply(WebDriver driver) {
         return elementIfVisible(element);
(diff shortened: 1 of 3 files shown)


iampopovich:
changes came from #16489 .
It added JSpecify null-safety annotations to the expected conditions classes. conditions do return null as a "not yet ready" signal. However, the annotations accidentally leaked into the public contract of the until() method, even though it never returns null to the caller. it either returns a value or throws TimeoutException. seems like Java silently accepts this, Kotlin does not: the code stopped compiling.

PR #17151, opened 2026-03-01
[java] Guard against NPE in Platform.extractFromSysProperty

seethinajayadileep:
### 💥 What does this PR do?

Prevents a possible `NullPointerException` in  
`Platform.extractFromSysProperty(String osName, String osVersion)`.

Previously:

- `osName` and `osVersion` were assumed to be non-null
- Calls such as `osVersion.equals("6.2")` could throw `NullPointerException`
- The public 2-argument method could receive `null` directly from callers

This PR makes the method defensive against null inputs.

---

### 🔧 Implementation Notes

The following changes were made:

1. Normalize `osName` and `osVersion` using `Objects.requireNonNullElse`:

   ```java
   osName = requireNonNullElse(osName, "");
   osVersion = requireNonNullElse(osVersion, "");
   osName = osName.toLowerCase(Locale.ENGLISH);
   ```

2. Provide a safe default when reading the system property:

   ```java
   System.getProperty("os.version", "")
   ```

These changes ensure:

- No NPE when null is passed to the public 2-argument method
- No NPE during version comparisons
- Existing platform detection logic remains unchanged

---

- Backwards compatible change
- No modification to detection heuristics
- Default behavior continues to fall back to `UNIX` when no match is found
- Improves defensive robustness for external callers

---

### 🔄 Types of changes

### DIFF
--- a/java/src/org/openqa/selenium/Platform.java
+++ b/java/src/org/openqa/selenium/Platform.java
@@ -428,9 +428,10 @@ public enum Platform {
    * @return the most likely platform based on given operating system name and version
    */
   public static Platform extractFromSysProperty(String osName, String osVersion) {
-    osName = osName == null ? "" : osName;
+    osName = requireNonNullOrElse(osName, "");
+    osVersion = requireNonNullOrElse(osVersion, "");
     osName = osName.toLowerCase(Locale.ENGLISH);
-    osVersion = osVersion == null ? "" : osVersion;
+
     // os.name for android is linux
     if ("dalvik".equalsIgnoreCase(System.getProperty("java.vm.name"))) {
       return Platform.ANDROID;
--- a/java/src/org/openqa/selenium/Platform.java
+++ b/java/src/org/openqa/selenium/Platform.java
@@ -415,7 +415,7 @@ public enum Platform {
    * @return the most likely platform based on given operating system name
    */
   public static Platform extractFromSysProperty(String osName) {
-    return extractFromSysProperty(osName, System.getProperty("os.version"));
+    return extractFromSysProperty(osName, System.getProperty("os.version", ""));
   }
 
   /**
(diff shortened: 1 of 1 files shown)


java/src/org/openqa/selenium/Platform.java
```diff
@@ -415,7 +415,7 @@ public static Platform getCurrent() {
    * @return the most likely platform based on given operating system name
    */
   public static Platform extractFromSysProperty(String osName) {
-    return extractFromSysProperty(osName, System.getProperty("os.version"));
+    return extractFromSysProperty(osName, System.getProperty("os.version", ""));
```
asolntsev, line 420: I think it's not needed because "os.version" is a required system property. It's always present in any Java process according to the spec:

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/System.html#getProperties()

But ok, we can keep this safeguard - just in case.

java/src/org/openqa/selenium/Platform.java
```diff
@@ -428,7 +428,9 @@ public static Platform extractFromSysProperty(String osName) {
    * @return the most likely platform based on given operating system name and version
    */
   public static Platform extractFromSysProperty(String osName, String osVersion) {
+    osName = osName == null ? "" : osName;
     osName = osName.toLowerCase(Locale.ENGLISH);
+    osVersion = osVersion == null ? "" : osVersion;
```
asolntsev: Don't you want to use shorthand?

```java
osName = requireNonNullOrElse(osName, "");
osVersion = requireNonNullOrElse(osVersion, "");
```

asolntsev (approved):
+ few minor optional suggestions

java/src/org/openqa/selenium/Platform.java
seethinajayadileep: Good point — I can switch to Objects.requireNonNullElse(...) for conciseness if preferred. Happy to update the PR.

java/src/org/openqa/selenium/Platform.java
seethinajayadileep: You're right — per the Java spec os.version should always be present in a standard JVM.

I kept the safeguard mainly for robustness in restricted or managed JVM environments (e.g., custom launchers or security-constrained setups)

asolntsev (approved):

PR #17174, opened 2026-03-04
[java] Add serialVersionUID to TimeoutException to avoid serialization warnings

seethinajayadileep:
### 💥 What does this PR do?

Adds an explicit `serialVersionUID` to the `UrlChecker.TimeoutException` class.

Since `TimeoutException` extends `Exception`, it is implicitly `Serializable`. Declaring a `serialVersionUID` prevents compiler/IDE serialization warnings and ensures a stable serialization identifier if the class structure changes in the future.

### 🔧 Implementation Notes

A `serialVersionUID` field with value `1L` was added to the `TimeoutException` class.

```java
private static final long serialVersionUID = 1L;
```

This follows common Java best practices for classes that implement `Serializable` (directly or indirectly through inheritance).

No behavioral changes were introduced.

This change is purely a code hygiene improvement and does not modify runtime behavior. No additional tests or documentation changes are required.

### 🔄 Types of changes

* Cleanup (formatting, renaming)

### DIFF
--- a/common/repositories.bzl
+++ b/common/repositories.bzl
@@ -50,8 +50,8 @@ js_library(
 
     http_archive(
         name = "linux_beta_firefox",
-        url = "https://ftp.mozilla.org/pub/firefox/releases/149.0b2/linux-x86_64/en-US/firefox-149.0b2.tar.xz",
-        sha256 = "14057fe24b65ef64125d04bde3507f48e489464b9f22cda832d34db92dede817",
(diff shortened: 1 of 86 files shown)


asolntsev:
@seethinajayadileep I think it's better to disable such warnings. Nobody is ever going to serialize TimeoutException - or any other exceptions - in Selenium project.

seethinajayadileep:
Thanks for the clarification. My intention was to address the serialization warning since `TimeoutException` extends `Exception`. But I agree that if Selenium doesn't rely on Java serialization for exceptions, this change may not be necessary. I'm happy to close the PR if preferred.

PR #17177, opened 2026-03-05
[java] Fix silent overflow when coercing JSON numbers to integral types

seethinajayadileep:
### 💥 What does this PR do?

This PR fixes a silent overflow issue when coercing JSON numbers into Java integral types (`Byte`, `Short`, `Integer`, `Long`).

Previously, `NumberCoercer` relied on `Number#byteValue`, `shortValue`, `intValue`, and `longValue` through the mapper. These methods silently overflow when the numeric value is outside the target type range.

Examples of the previous incorrect behavior:

* `"2147483648"` coerced to `Integer` resulted in `-2147483648`
* `"100000"` coerced to `Short` resulted in an incorrect wrapped value

This change adds validation before applying the mapper:

* Convert the parsed number to `BigDecimal`
* Use `toBigIntegerExact()` to ensure the value is an exact integer
* Validate that the value is within the bounds of the target type
* Throw a `JsonException` if the value is fractional or out of range

This prevents silent numeric corruption during JSON deserialization.

---

### 🔧 Implementation Notes

The validation logic was added inside `NumberCoercer` before applying the numeric mapper.

Implementation steps:

1. Convert the parsed `Number` to `BigDecimal`
2. Use `toBigIntegerExact()` to reject fractional values
3. Check the numeric range against the target type (`Byte`, `Short`, `Integer`, `Long`)
4. Throw `JsonException` if the value is outside the allowed bounds

---

### 🧪 Tests

Unit tests were added to verify the new behavior:

* Out-of-range values for integral types throw `JsonException`
* Fractional values cannot be coerced to integral types

These tests ensure the overflow validation remains enforced.

---
### 🔄 Compatibility

JsonInput#nextNumber() behavior was preserved to continue returning Double values,
avoiding changes to existing callers that rely on the current runtime type.

This change applies to the Java binding. Other Selenium bindings may handle
numeric coercion differently and can be aligned in follow-up changes if needed.

---

This change only affects invalid numeric inputs that previously produced incorrect wrapped values. Valid numeric inputs continue to behave the same.

---

### 🔄 Types of changes

* Bug fix (backwards compatible)


seethinajayadileep:
This is still valid. Java CI was green on the overflow fix. Ready for review.

PR #17184, opened 2026-03-07
Revert "[java] make the signature change in `ExecuteMethod` backward compatible"

vaibhav09mbm:
Reverts SeleniumHQ/selenium#17183


VietND96:
@vaibhav09mbm What is the reason for this revert?
Cc: @asolntsev

asolntsev:
@vaibhav09mbm I don't think we need to revert https://github.com/SeleniumHQ/selenium/pull/17183. It was useful.
Closing the PR until we get more info.

PR #17187, opened 2026-03-07
[java] Fix duplicate field name handling in InstanceCoercer#getFieldWriters

seethinajayadileep:
### 💥 What does this PR do?
This PR fixes duplicate field name handling in `InstanceCoercer#getFieldWriters`.

Previously, field writers were collected using `Collectors.toMap(...)` without an explicit merge strategy. When duplicate field names were encountered (for example through field hiding in inheritance), the collector would fail with an implicit `IllegalStateException` caused by duplicate keys.

This change makes duplicate handling explicit by detecting duplicates during map collection and throwing a `JsonException` with a clear error message.

This prevents ambiguous field mappings and replaces the implicit collector failure with a controlled and descriptive exception.

---

### 🔧 Implementation Notes

The change modifies the `Collectors.toMap(...)` call in `getFieldWriters`.

**Previous behavior**
- `Collectors.toMap` was used without a merge function.
- Duplicate field names caused an implicit `IllegalStateException` from the collector.

**New behavior**
- A merge function is provided that throws a `JsonException` when duplicate field names are encountered.
- This makes the failure explicit and easier to diagnose.

Example scenario that previously caused an implicit collector failure:

```java
class Parent {
  String value;
}

class Child extends Parent {
  String value;
} 

Both fields resolve to the same JSON key (value), producing a duplicate during field writer collection.

With this change, such cases now fail with a clear JsonException instead of a generic collector error.

💡 Additional Considerations

This change only affects cases where duplicate serialized field names exist.

Valid classes with unique field names continue to behave exactly as before.

A regression test was added to verify that duplicate field names detected during BY_FIELD deserialization result in a JsonException.

🔄 Types of changes

Bug fix (backwards compatible)

### CODE REVIEW

**Comment by seethinajayadileep** [comment]
Hi maintainers, could someone please assign a reviewer for this PR? Thank you.

**Review by asolntsev — APPROVED** [review]
It's slightly better than previously, so we can merge this PR. Thank you. 

Before:
```java
java.lang.IllegalStateException: Duplicate key value (attempted merging values org.openqa.selenium.json.InstanceCoercer$TypeAndWriter@4e70a728 and org.openqa.selenium.json.InstanceCoercer$TypeAndWriter@b7838a9)
```

After:
```java
org.openqa.selenium.json.JsonException: Duplicate JSON field name detected while collecting field writers
```

P.S. But the name of the field is not still shown in the error message. Ideally, we should show the name as well. But I can it in a separate PR if you wish.

PR #17199, opened 2026-03-10
[rb] Fix failing firefox beta tests for browser context

aguspe:
### 🔗 Related Issues
I made this pr #17188, and I keep getting Firefox Beta errors

### 💥 What does this PR do?
I'm trying to use event subscription that is natively BiDi for the test instead of alert and see if this fixes the errors

### 🔧 Implementation Notes
I think using the dismiss alert action was messing with the handle user prompt from bidi

### 🔄 Types of changes

### DIFF
--- a/rb/spec/integration/selenium/webdriver/bidi/browsing_context_spec.rb
+++ b/rb/spec/integration/selenium/webdriver/bidi/browsing_context_spec.rb
@@ -87,46 +87,49 @@ module Selenium
             reset_driver!(unhandled_prompt_behavior: :ignore)
           end
 
-          it 'accepts users prompts without text' do
+          it 'accepts user prompts without text' do
             browsing_context = described_class.new(bridge)
 
             driver.navigate.to url_for('alerts.html')
             window = driver.window_handle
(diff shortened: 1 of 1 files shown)


cgoldberg:
@aguspe I'm not totally sure, but I think the failures in Firefox Beta are related to: https://github.com/mozilla/geckodriver/issues/2241 
Once our CI pulls the updated version of Firefox Beta with this fix, it might just work again

aguspe:
> @aguspe I'm not totally sure, but I think the failures in Firefox Beta are related to: [mozilla/geckodriver#2241](https://github.com/mozilla/geckodriver/issues/2241)
> Once our CI pulls the updated version of Firefox Beta with this fix, it might just work again

@cgoldberg thank you so much for the link! I'm trying to give it a try to only use BiDi events and fix these tests to make it more resilient in the future too, and make it less flaky. Maybe it's not the best approach but if it works we will not have to wait until the 24th to try it out :)

aguspe:
I will close this one since we will wait until firefox fixes their bug

PR #17219, opened 2026-03-13
[rb] Patch devtools to permit up to 3 versions back

luke-hill:
Whilst there is an issue with devtools being released to rubygems, this is a short "quick" temporary fix to unblock users who now are on chrome v146 and cannot use selenium w/ devtools

### 🔄 Types of changes


diemol:
I just pushed devtools 144 and 145. This is not needed.

PR #17227, opened 2026-03-14
[Javascript] Delete dead atoms that won't work in browsers

AutomatedTester:
This deletes code that have been removed from the web platform. These dont work anywhere so removing them as part of the plan to move the atoms to typescript

### 🔄 Types of changes


AutomatedTester:
closing this as I am taking a different route

PR #17263, opened 2026-03-27
[py] fix selenium-manager binary selection on FreeBSD

obiwac:
Previously, sys.platform was checked to match "freebsd". However,
Py_GetPlatform on FreeBSD returns the version number along with it (e.g.
"freebsd15") [1]. So check sys.platform starts with "freebsd" instead.

Also, platform.machine() gives "amd64" on FreeBSD, not "x86_64", so fix
that.

Do the same with OpenBSD, and only allow Windows binary to be selected
if arch is correct ("AMD64" on Windows).

Update warning about Selenium Manager binary on FreeBSD to hint at user
what they need to do to get the Linux binary working on FreeBSD.

Fixes: 5a7a2ac896d6 ("[py] allow bsd operating systems to use the linux binary")

[1] <https://python.readthedocs.io/fr/stable/c-api/init.html#c.Py_GetPlatform>

OpenBSD probably needs a similar change, but I have not investigated that.

This patch is a little ugly, but I'm not sure how to make this more elegant.

### 🔄 Types of changes

### DIFF
--- a/py/selenium/webdriver/common/selenium_manager.py
+++ b/py/selenium/webdriver/common/selenium_manager.py
@@ -87,21 +87,24 @@ class SeleniumManager:
         elif compiled_path.is_file():
             path = compiled_path
         else:
-            allowed = {
-                ("darwin", "any"): "macos/selenium-manager",
-                ("win32", "any"): "windows/selenium-manager.exe",
-                ("cygwin", "any"): "windows/selenium-manager.exe",
-                ("linux", "x86_64"): "linux/selenium-manager",
-                ("freebsd", "x86_64"): "linux/selenium-manager",
-                ("openbsd", "x86_64"): "linux/selenium-manager",
-            }
+            arch = platform.machine()
 
-            arch = platform.machine() if sys.platform in ("linux", "freebsd", "openbsd") else "any"
-            if sys.platform in ["freebsd", "openbsd"]:
-                logger.warning(f"Selenium Manager binary may not be compatible with {sys.platform}; verify settings")
-
-            location = allowed.get((sys.platform, arch))
-            if location is None:
+            if sys.platform == "darwin":
+                location = "macos/selenium-manager"
+            elif sys.platform in ("win32", "cygwin") and arch == "AMD64":
+                location = "windows/selenium-manager.exe"
+            elif sys.platform == "linux" and arch == "x86_64":
+                location = "linux/selenium-manager"
+            elif sys.platform.startswith("freebsd") and arch == "amd64":
+                logger.warning(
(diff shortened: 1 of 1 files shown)


cgoldberg (changes requested):
Thanks. We don't have many FreeBSD users...

In Python 3.14+, they fixed this, so `sys.platform` won't include version... but I guess this is useful for older versions.

There is also another subtle bug in the current code. We only ship a Windows version for x86-64, but the code allows `"any"`, so that is being skipped. Also, Windows reports `platform.machine()` as `AMD64`.

What do you think of this small tweak instead... it should handle the `amd64/AMD64` cases correctly for all platforms:


```
            allowed = {
                ("darwin", "any"): "macos/selenium-manager",
                ("win32", "x86_64"): "windows/selenium-manager.exe",
                ("cygwin", "x86_64"): "windows/selenium-manager.exe",
                ("linux", "x86_64"): "linux/selenium-manager",
                ("freebsd", "x86_64"): "linux/selenium-manager",
                ("openbsd", "x86_64"): "linux/selenium-manager",
            }

            platform_name = sys.platform
            arch = "any" if platform_name == "darwin" else platform.machine().lower()
            arch = "x86_64" if arch == "amd64" else arch

            if platform_name.startswith("freebsd"):
                logger.warning(
                    "Selenium Manager binary may not be compatible with FreeBSD; you may need to run "
                    "'brandelf -t linux' on it and load linux64.ko"
                )
                platform_name = "freebsd"
            elif platform_name.startswith("openbsd"):
                logger.warning("Selenium Manager binary may not be compatible with OpenBSD; verify settings")
                platform_name = "openbsd"

            location = allowed.get((platform_name, arch))
```

obiwac:
@cgoldberg I think at that point the logic is complex enough that we shouldn't try to be smart with it. See my updated diff - what do you think of it? I can also confirm OpenBSD returns amd64, not x86_64, too. Also included the Windows thing but maybe that should be its own commit.

cgoldberg:
> See my updated diff 

We still want the dictionary of allowed platforms because we will soon be shipping ARM64 binaries and we need to select the correct binary on those platforms.

I just submitted #17271 instead of this PR.

obiwac:
> We still want the dictionary of allowed platforms because we will soon be shipping ARM64 binaries and we need to select the correct binary on those platforms.

still think my solution was better

cgoldberg:
> still think my solution was better

Yours would have to be refactored soon if I used it. There are already Pull Requests open that add more platforms/architectures to the dictionary, and the dictionary dispatch is nicer (IMO) than a long if/else chain... sorry if you don't agree.

PR #17266, opened 2026-03-27
[py] Use generated Bidi files instead of hand curated ones

AutomatedTester:
### 💥 What does this PR do?

Deletes all the hand curated bidi code in the python tree and uses the generated code created in #16914 . Once that PR is approved we can merge this into there or do it into trunk as a 2nd PR. #16914 MUST BE FIRST INTO TRUNK

### 🔄 Types of changes


rake_tasks/python.rake
```diff
@@ -80,6 +80,19 @@ task :local_dev, [:all] do |_task, arguments|
...
+    bidi_dest = "#{lib_path}/common/bidi"
+    if Dir.exist?(bidi_src)
+      FileUtils.mkdir_p(bidi_dest)
+      Dir.children(bidi_src).sort.each do |entry|
+        src = File.join(bidi_src, entry)
+        next unless File.file?(src) || File.symlink?(src)
+
+        resolved_src = File.symlink?(src) ? File.realpath(src) : src
+        FileUtils.cp(resolved_src, File.join(bidi_dest, entry))
+      end
```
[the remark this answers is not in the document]
cgoldberg: > Consider deleting bidi_dest (or at least removing *.py in it)

yea, it will also fail if you this multiple times because the files already exist and are read-only. Deleting them first should work

rake_tasks/python.rake
```diff
@@ -80,6 +80,19 @@ task :local_dev, [:all] do |_task, arguments|
     FileUtils.rm_rf("#{lib_path}/common/devtools")
```
cgoldberg, line 60: I don't think this line is necessary. the next line removes everything under `py/selenium/webdriver` and copies it

rake_tasks/python.rake
```diff
@@ -80,6 +80,19 @@ task :local_dev, [:all] do |_task, arguments|
...
+      Dir.children(bidi_src).sort.each do |entry|
+        src = File.join(bidi_src, entry)
+        next unless File.file?(src) || File.symlink?(src)
+
+        resolved_src = File.symlink?(src) ? File.realpath(src) : src
+        FileUtils.cp(resolved_src, File.join(bidi_dest, entry))
+      end
+    end
+
     %w[common/devtools common/linux common/mac common/windows].each do |dir|
```
cgoldberg: can you also fix this? `common/mac` isn't correct. the directory is named `common/macos`

cgoldberg (changes requested):

cgoldberg:
We need to add the directory to ruff config in `pyproject.toml` or else the linter will fail if you run it with the files in your local source tree.

You can add it to:

```
[tool.ruff]
extend-exclude = [
    "selenium/webdriver/common/devtools/",
    "selenium/webdriver/common/bidi/",
]
```

cgoldberg:
@AutomatedTester I pushed a fix to your branch that resolves all the comments I made and also adds the generated files to `.gitignore`.

cgoldberg (approved):

cgoldberg:
there are still some linting/formatting errors in the new generation code

py/generate_bidi.py
```diff
@@ -0,0 +1,1824 @@
+#!/usr/bin/env python3.10
```
cgoldberg: this can be removed?

rake_tasks/python.rake
cgoldberg: this should be `common/macos`

rake_tasks/python.rake
cgoldberg, line 75: do `FileUtils.rm_f()` on each destination file before copying, or this will fail when you run it a 2nd time.

cgoldberg (changes requested):
I fixed `python.rake` previously, but you force pushed over my commit :/  I just added comments again.

We should add the generated files to `.gitignore`

Also, run the formatter so it passes the lint job in CI.

cgoldberg (approved):

PR #17271, opened 2026-03-29
[py] Improve Selenium Manager platform/architecture detection

cgoldberg:
### 💥 What does this PR do?

Supersedes: #17263

This PR improves platform/architecture detection for Selenium Manager.

On versions of Python older than 3.14. `sys.platform` on FreeBSD/OpenBSD includes the version number (i.e. `freebsd15`). This handles those cases by explicitly setting a normalized platform name. It also updates the warning about Selenium Manager binary on FreeBSD to hint at user what they need to do to get the Linux binary working on FreeBSD.

Also, some platforms return `amd64` or `AMD64` for architecture. This normalizes them to `X86_64` so they are handled the same.

Previously, we were allowing "any" architecture on Windows, so the proper exception was not raised and Selenium Manager would fail to run instead.

This PR also adds unit tests for these.

### 🔄 Types of changes

### DIFF
--- a/py/selenium/webdriver/common/selenium_manager.py
+++ b/py/selenium/webdriver/common/selenium_manager.py
@@ -89,18 +89,30 @@ class SeleniumManager:
         else:
             allowed = {
                 ("darwin", "any"): "macos/selenium-manager",
-                ("win32", "any"): "windows/selenium-manager.exe",
-                ("cygwin", "any"): "windows/selenium-manager.exe",
+                ("win32", "x86_64"): "windows/selenium-manager.exe",
+                ("cygwin", "x86_64"): "windows/selenium-manager.exe",
                 ("linux", "x86_64"): "linux/selenium-manager",
                 ("freebsd", "x86_64"): "linux/selenium-manager",
                 ("openbsd", "x86_64"): "linux/selenium-manager",
             }
 
-            arch = platform.machine() if sys.platform in ("linux", "freebsd", "openbsd") else "any"
-            if sys.platform in ["freebsd", "openbsd"]:
-                logger.warning(f"Selenium Manager binary may not be compatible with {sys.platform}; verify settings")
+            # some operating systems report x86-64 architecture as amd64/AMD64
+            platform_name = sys.platform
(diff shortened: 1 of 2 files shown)


py/test/unit/selenium/webdriver/common/selenium_manager_tests.py
```diff
@@ -46,24 +46,27 @@ def test_uses_environment_variable(monkeypatch):
...
-
     assert str(binary) == sm_path
 
 
 def test_uses_windows(monkeypatch):
     monkeypatch.setattr(sys, "platform", "win32")
     binary = SeleniumManager()._get_binary()
-
     project_root = Path(selenium.__file__).parent.parent
     assert binary == project_root.joinpath("selenium/webdriver/common/windows/selenium-manager.exe")
```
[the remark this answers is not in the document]
cgoldberg: fixed.

cgoldberg:
The RBE failures are unrelated.. all python tests pass

py/selenium/webdriver/common/selenium_manager.py
```diff
@@ -89,18 +89,30 @@ def _get_binary() -> Path:
...
 
-            arch = platform.machine() if sys.platform in ("linux", "freebsd", "openbsd") else "any"
-            if sys.platform in ["freebsd", "openbsd"]:
-                logger.warning(f"Selenium Manager binary may not be compatible with {sys.platform}; verify settings")
-
-            location = allowed.get((sys.platform, arch))
+            # some operating systems report x86-64 architecture as amd64/AMD64
+            platform_name = sys.platform
+            arch = "any" if platform_name == "darwin" else platform.machine().lower()
+            arch = "x86_64" if arch == "amd64" else arch
```
obiwac, line 102: This is wrong. You meant to do `arch.lower()` and have the comment down there

cgoldberg: The code is correct as-is. Maybe the comment should be a line lower, but it does the right thing

obiwac: ah, indeed. Mixed up the `platform.machine()` being lowered with the `platform_name`

Issue #17264, opened 2026-03-27
[🐛 Bug]: Ruby: WebSocket hangs when receiving payload larger than WebSocket.max_frame_size


alpaca-tc:
### Description

When using Selenium with Ruby, the WebSocket connection used for DevTools (CDP) can hang if it receives a payload larger than `WebSocket.max_frame_size` (default: 20MB).

This issue is difficult to detect because, by default, no error is raised and the process appears to hang silently.

## Root Cause

The issue is caused by the behavior of `websocket-ruby` and how it is used in Selenium.

- When the payload size exceeds `WebSocket.max_frame_size`, `websocket-ruby` raises a `TooLong` exception during frame decoding:
  - https://github.com/imanel/websocket-ruby/blob/8f9f4c12b60934a476b6c571aab33a84309a3c68/lib/websocket/frame/handler/handler03.rb#L127
- However, unless `WebSocket.should_raise = true` is set, this exception is rescued internally and `nil` is returned:
  - https://github.com/imanel/websocket-ruby/blob/8f9f4c12b60934a476b6c571aab33a84309a3c68/lib/websocket/exception_handler.rb#L24
- In Selenium, `incoming_frame.next` is used to read messages:
  - https://github.com/SeleniumHQ/selenium/blob/fda2c56e5894b6ac68c6912a441b5810c4fd6ce0/rb/lib/selenium/webdriver/common/websocket_connection.rb#L135
- Once a frame exceeds `WebSocket.max_frame_size`, decoding will always fail, but the loop continues calling `next`, which keeps returning `nil`.

As a result, the WebSocket reading loop effectively becomes stuck, and the process appears to hang.

## Reproduction

This issue occurs when a large payload is sent over the DevTools WebSocket connection.

The easiest way to reproduce is to reduce the frame size limit:

```ruby
WebSocket.max_frame_size = 1
```

Then trigger any CDP message.

In my case, the issue occurred under the following conditions:
1. Use DevTools request interception (e.g., `driver.intercept`)
  - This internally enables [ `Network.enable` ](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-enable):
2. Display a large preview image in the browser using a `data:image/...` URL
3. CDP emits events such as:
  - `Network.responseReceived`
  - `Network.requestWillBeSent`
  - These include the large data: URL in the payload.
4. The Ruby process hangs

## Expected Behavior

When a frame exceeds `WebSocket.max_frame_size`, one of the following should happen:
[...]

PR #17286, opened 2026-04-01
[rb] fix silent hang on large WebSocket frame

Chandan25sharma:
- Sets `WebSocket.should_raise = true` to prevent `websocket-ruby` from 
  returning `nil` frames on failure, which previously caused infinite read 
  loops and silent hangs.
- Increases the default `WebSocket.max_frame_size` to 100MB (from 20MB) 
  to handle large CDP payloads such as Data URLs and screenshots.
- Updates `WebSocketConnection#send_cmd` to fail immediately if the 
  listener thread dies, avoiding a 30-second timeout wait when the 
  connection is already broken.
- Changes WebSocket error logging from `debug` to `warn` to make failures 
  visible to users by default.
- Adds `track_cancelled` option to `Driver#intercept` (default: `true`), 
  enabling users to opt-out of the `Network` domain to reduce bandwidth 
  from chatty events.

### DIFF
--- a/rb/lib/selenium/webdriver/common/driver_extensions/has_network_interception.rb
+++ b/rb/lib/selenium/webdriver/common/driver_extensions/has_network_interception.rb
@@ -59,7 +59,7 @@ module Selenium
         # @yieldparam [Proc] continue block which proceeds with the request and optionally yields response
         #
 
-        def intercept(&block)
+        def intercept(**options, &block)
           if browser == :firefox
             WebDriver.logger.deprecate(
               'Driver#intercept on Firefox',
@@ -68,7 +68,7 @@ module Selenium
             )
           end
           @interceptor ||= DevTools::NetworkInterceptor.new(devtools)
-          @interceptor.intercept(&block)
+          @interceptor.intercept(**options, &block)
         end
       end # HasNetworkInterception
     end # DriverExtensions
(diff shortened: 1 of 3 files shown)


aguspe (changes requested):
hey thanks for looking into this, the websocket hang is a real pain

the global WebSocket config setting `should_raise` and `max_frame_size` at require time will affect any other websocket-ruby usage in the same process which is not great, maybe we can scope this to the connection instance instead?

also when `track_cancelled: false` the `cancelled_requests` never gets populated but `with_cancellable_request` still checks `cancelled?`, so browser cancellations will raise instead of being swallowed, that needs to be handled

this PR is doing a lot of things at once though, websocket config, dead thread detection, logging changes and the track_cancelled feature, wouldnt it be easier to review if we split into smaller PRs? maybe start with the websocket hang fix first and then the track_cancelled feature separately, we'd also need tests for the new behavior

aguspe:
Thanks again for digging into this @Chandan25sharma. I've opened #17655 which fixes the WebSocket-hang part of #17264 in a focused way (lazy + conditional 100MB limit, plus surfacing the dropped frame as a logged error instead of relying on global `WebSocket.should_raise`). The `track_cancelled` feature from this PR is a separate concern and isn't included there — happy to look at it separately if you'd like to split it out.

PR #17295, opened 2026-04-02
[dotnet] [build] Don't skip tests

nvborisenko:
Remove some tests from the skipped lists.

### 💥 What does this PR do?
This pull request updates the `.skipped-tests` file to remove several test exclusions, allowing these tests to run again. This suggests that the previously skipped tests are now expected to pass.


nvborisenko:
Passes.

PR #17319, opened 2026-04-08
fix(update_cdp): update file paths for ChromeDriver configuration

diemol:
Fixing this path to do a release.

### DIFF
--- a/scripts/update_cdp.py
+++ b/scripts/update_cdp.py
@@ -182,8 +182,8 @@ def update_dotnet(chrome_milestone):
     for file in files:
         replace_in_file(file, old_chrome(chrome_milestone), new_chrome(chrome_milestone))
 
-    files = [root_dir / "dotnet/test/common/CustomDriverConfigs/StableChannelChromeDriver.cs"]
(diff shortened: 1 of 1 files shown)


scripts/update_cdp.py
```diff
@@ -182,8 +182,8 @@ def update_dotnet(chrome_milestone):
     for file in files:
         replace_in_file(file, old_chrome(chrome_milestone), new_chrome(chrome_milestone))
 
-    files = [root_dir / "dotnet/test/common/CustomDriverConfigs/StableChannelChromeDriver.cs"]
-    dir_path = root_dir / "dotnet/test/common/DevTools"
+    files = [root_dir / "dotnet/test/webdriver/Infrastructure/DriverConfigs/StableChannelChromeDriver.cs"]
```
nvborisenko, line 185: My bad. Would be nice to keep binding specific stuff per directory (e.g. `dotnet`). And top-level script will delegate work. So it would be easier to understand what is impacted. Just observation.

diemol: IDEs help with this type of things usually, unless you tell them to not check strings.

PR #17342, opened 2026-04-13
Fix py:local_dev rake task

cgoldberg:
### 💥 What does this PR do?

This PR fixes the `py:local_dev` rake task so it deletes BiDi files from the source tree before copying them from bazel output. Without this, the task will fail since the files are read-only.

It also fixes the `macos` directory name so this works on Mac.

### 🔄 Types of changes
- Development

### DIFF
--- a/rake_tasks/python.rake
+++ b/rake_tasks/python.rake
@@ -57,6 +57,7 @@ task :local_dev, [:all] do |_task, arguments|
 
   copy_all = arguments[:all] == 'all'
   if copy_all
+    FileUtils.rm_rf("#{lib_path}/common/devtools")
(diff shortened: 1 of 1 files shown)


cgoldberg:
The Formatter CI job failed because of:

```
Error in download: java.io.IOException: Error downloading [https://github.com/aspect-build/bsdtar-prebuilt/releases/download/v3.8.1/tar_linux_amd64] to /home/runner/.bazel/external/aspect_bazel_lib++toolchains+bsd_tar_linux_amd64/tar: GET returned 504 Gateway Time-out
```

I ran formatting locally and everything passed

PR #17345, opened 2026-04-14
Improve error message when driver is not found

Rohini2222:
This PR improves the error message when WebDriver is not found by providing clearer guidance to the user.


cgoldberg:
This isn't an improvement and also has formatting errors and breaks functionality.

PR #17346, opened 2026-04-14
Improve error message for invalid page load strategy

Rohini2222:
This PR improves the error message for invalid page load strategy by making it clearer and more user-friendly. It now provides better guidance on the expected valid values (normal, eager, none).


cgoldberg:
This isn't an improvement and also has formatting errors and breaks functionality.

PR #17347, opened 2026-04-14
Add test for invalid page load strategy validation

Rohini2222:
This PR adds a test case to validate that an invalid page load strategy raises a ValueError with an appropriate error message.


cgoldberg:
This PR has formatting errors, breaks functionality, contains commits from your other PR's, and unnecessarily adds a new test file that doesn't follow our conventions.

PR #17363, opened 2026-04-19
[java] remove field `ChromiumDriver.capabilities`

asolntsev:
... that duplicated superclass' field `RemoteWebDriver.capabilities`.

### 💥 What does this PR do?

+ extract method `resolvePlatform` and cover with tests
+ remove unneeded safeguard `if (capabilities == null)` - the `capabilities` is never null (validated in constructor).
+ added @NullMarked for few test-implementations of RemoteWebDriver

### 🔄 Types of changes

### DIFF
--- a/java/src/org/openqa/selenium/chromium/ChromiumDriver.java
+++ b/java/src/org/openqa/selenium/chromium/ChromiumDriver.java
@@ -85,7 +85,6 @@ public class ChromiumDriver extends RemoteWebDriver
       name -> CHROME.is(name) || EDGE.is(name) || OPERA.is(name);
   private static final Logger LOG = Logger.getLogger(ChromiumDriver.class.getName());
 
-  private final Capabilities capabilities;
   private final HasNetworkConditions networkConditions;
   private final HasPermissions permissions;
   private final HasLaunchApp launch;
--- a/java/src/org/openqa/selenium/remote/RemoteWebDriver.java
+++ b/java/src/org/openqa/selenium/remote/RemoteWebDriver.java
@@ -19,6 +19,7 @@ package org.openqa.selenium.remote;
 
 import static java.util.Collections.singleton;
 import static java.util.Objects.requireNonNull;
+import static java.util.Objects.requireNonNullElseGet;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static java.util.logging.Level.SEVERE;
 import static org.openqa.selenium.remote.CapabilityType.PLATFORM_NAME;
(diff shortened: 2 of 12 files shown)


diemol (commented):
What would happen if someone were using this class directly in a custom Chromium-based browser?

asolntsev:
@diemol Which class do you mean? ChromeDriver? Then nothing would change because it was a private field.

java/src/org/openqa/selenium/chromium/ChromiumDriver.java
```diff
@@ -85,7 +85,6 @@ public class ChromiumDriver extends RemoteWebDriver
       name -> CHROME.is(name) || EDGE.is(name) || OPERA.is(name);
   private static final Logger LOG = Logger.getLogger(ChromiumDriver.class.getName());
 
-  private final Capabilities capabilities;
```
diemol, line 88: This class. What would happen if someone were to extend this class on their own?

asolntsev: Nothing happens. This was a private final field. No one could use it directly. 

Just in case: there is a field with exactly the same name in superclass (RemoteWebDriver).

pzygielo:
Not sure yet this is related, but after upgrading to 4.44.0 in

- https://github.com/pzygielo/jasmine-maven-plugin/pull/678

I'm getting:

```
  java.lang.NullPointerException: Cannot invoke "org.openqa.selenium.Capabilities.toString()" because the return value of "org.openqa.selenium.remote.RemoteWebDriver.getCapabilities()" is null
        at org.openqa.selenium.support.ui.WebDriverWait.timeoutException(WebDriverWait.java:95)
        at org.openqa.selenium.support.ui.FluentWait.until(FluentWait.java:234)
```

:eyes:

diemol:
@asolntsev can you please check?

asolntsev:
@pzygielo Hi. That's why I recommend to avoid mocks as much as possible. :)

In your tests, the mocked webdriver's method `getCapabilities()` returns `null` which caused NullPointerException. While in reality, method `getCapabilities()` never returns `null`.

This is the fix that changes the mock so that it behaves like real object:
https://github.com/pzygielo/jasmine-maven-plugin/compare/4.x-pz...asolntsev:fix/tests?expand=1

pzygielo:
> In your tests, the mocked webdriver's method `getCapabilities()` returns `null` which caused NullPointerException. While in reality, method `getCapabilities()` never returns `null`.

Thank you for checking.

(Don't want to excuse myself, but this is forked plugin which I just try to keep a bit more up to date than abandoned upstream. I admit - I don't know a lot of codebase there... Simply other bumps, including Selenium, went more or less clearly so far.)

> This is the fix that changes the mock so that it behaves like real object: https://github.com/pzygielo/jasmine-maven-plugin/compare/4.x-pz...asolntsev:fix/tests?expand=1

Appreciate that!

Issue #17355, opened 2026-04-16
[🐛 Bug]: Selenium should depend on net.bytebuddy:byte-buddy 1.18.8 instead of 1.18.8-jdk5


tmortagne:
### Description

https://github.com/SeleniumHQ/selenium/commit/74c7f0249ab2a3013bcb06b8695122274872be7f upgraded `net.bytebuddy:byte-buddy` dependency from 1.18.5 to 1.18.8-jdk5, but my understanding is that the point of the `-jdk5` flavor is to support Java versions lower than 8, which is not the case of Selenium (unless I missed something).

Selenium is of course (probably) working fine with `1.18.8-jdk5` too, but we have a check to make sure all transitive dependencies are compatible, and Maven sees `1.18.8-jdk5` as greater than our managed dependency `1.18.8` which makes the enforcer plugin complain.

### Reproducible Code

```shell
Not much to reproduce here, it's strictly a dependency problem.
```

---
ℹ️ **Last known working version:** `4.41.0`

StefanPenndorf:
We ran into this problem as well.

Since Byte Buddy will continue publishing the `-jdk5` variant alongside regular releases, this issue could reappear with the next automated dependency update. The version `1.18.8-jdk5` was introduced as part of a release preparation commit: https://github.com/SeleniumHQ/selenium/pull/17322

Do you have any plans or thoughts about avoiding the reintroduction of the `-jdk5` suffix in the future?

diemol:
This happened again, https://github.com/SeleniumHQ/selenium/blob/trunk/MODULE.bazel#L229, we need to add a rule to the script that updates the dependencies.

PR #17382, opened 2026-04-24
[java] Use byte-buddy 1.18.8 instead of 1.18.8-jdk5

pierluigilenoci:
### 🔗 Related Issues

Fixes https://github.com/SeleniumHQ/selenium/issues/17355

### 💥 What does this PR do?

Changes the `net.bytebuddy:byte-buddy` dependency from `1.18.8-jdk5` to `1.18.8` in `MODULE.bazel`.

The `-jdk5` flavor of byte-buddy is intended for Java versions lower than 8, which Selenium does not target. Using the `-jdk5` variant causes Maven enforcer plugin failures in downstream projects because Maven considers `1.18.8-jdk5` as a higher version than the managed dependency `1.18.8`, triggering convergence errors.

**Note:** The `java/maven_install.json` lock file will need to be regenerated after this change. A maintainer with Bazel set up can run:
```
REPIN=1 bazel run @unpinned_maven//:pin
```

### 🤖 AI assistance
    - Tool(s): Claude Code (Anthropic)
    - What was generated: The commit and PR were created with AI assistance. The fix itself is a single-line version change in MODULE.bazel.

### 🔄 Types of changes
- Bug fix


pierluigilenoci:
Regenerated `java/maven_install.json` to fix the lockfile signature mismatch after the byte-buddy version change.

**Changes in `java/maven_install.json`:**
- Updated `__INPUT_ARTIFACTS_HASH` for `net.bytebuddy:byte-buddy` to match version `1.18.8`
- Updated artifact SHA256 checksums (jar + sources) from Maven Central
- Updated version from `1.18.8-jdk5` to `1.18.8`
- Recomputed `__RESOLVED_ARTIFACTS_HASH` for 10 affected transitive dependents (mockito-core, assertj-core, selenium-remote-driver, selenium-support, etc.)

This is equivalent to running `REPIN=1 bazel run @maven//:pin` — the hash computation was done programmatically using the same algorithm as `rules_jvm_external` v3 lockfile format.

Issue #12315, opened 2023-07-05
[🚀 Feature]: Publish `StrongNamed` builds to nuget feed 


harunurhan:
### Feature and motivation

It would make it easier to consume if strong name signed versions of the packages, such as `Selenium.WebDriver.StrongNamed` were published to nuget.org feed, currently it's only possible to download from github.

### Usage example

N/A

nvborisenko:
I don't think we will publish it in nuget feed. More realistic scenario is making Selenium.WebDriver with strong name.

Mek7:
OK, it would be very useful also for us. Any news on this?

superbull:
Hi, when will the strong named version published in nuget feed?

nvborisenko:
Only with Selenium 5 as a major breaking change, and if core team doesn't mind.

Mek7:
Major breaking change?
Strongly naming a library doesn't influence existing consumers, only broadens its usability for environments where everything is strongly named (and thus, non-strongly named libraries cannot be referenced).
Am I missing something?

tbremard:
Hello,
I can see that this ticket is open since 1 year (july 2023 when we are in june 2024.
I did not see it and created this ticket (which has been closed because of 'duplication'):
https://github.com/SeleniumHQ/selenium/issues/14115

let me copy paste my last comment:

Strong-naming an assembly introduces a small initial overhead, as it requires generating a key pair and integrating the signing process into the build pipeline. However, once this key pair is generated, it can be reused for all subsequent releases, making the process transparent and straightforward for future updates.

For a widely used library like Selenium, providing strong-named assemblies in the NuGet package would enable developers to use these assemblies in environments that require strong-naming, such as in large enterprises with strict security policies. It allows these developers to maintain strong-naming across their entire application stack, ensuring compatibility and compliance with their security requirements.

Moreover, it is worth noting that most NuGet packages available today include a public key, which means they are already strong-named. This has become a standard practice, reflecting the industry's emphasis on security and integrity. By aligning with this standard, the Selenium library will not only enhance its security posture but also meet the expectations of its user base, facilitating smoother integration and wider adoption.

The benefits of strong-named assemblies in promoting security, integrity, and trustworthiness far outweigh the small initial overhead. Incorporating strong-named assemblies in the next minor release would significantly enhance the utility of the Selenium library, making it more suitable for a wider range of applications and increasing its adoption in secure and enterprise environments.
[...]

PR #17397, opened 2026-04-26
[dotnet] Strongly signed assemblies

nvborisenko:
Finally getting strongly named assemblies.

### 🔗 Related Issues
So many through the history. Fixes https://github.com/SeleniumHQ/selenium/issues/12315 - don't wait v5, we don't follow SemVer.

We should be loud, most likely blog post saying "Now we are strongly signed, that's it."

### 🔄 Types of changes


dotnet/src/webdriver/BUILD.bazel
nvborisenko, line 1: Good chance to rename assembly file: `Selenium.WebDriver.dll`

dotnet/src/support/BUILD.bazel
nvborisenko, line 1: Good chance to rename assembly file: `Selenium.WebDriver.Support.dll`

nvborisenko: `Selenium.Support.dll` to be aligned with package name.

nvborisenko:
Fall back: https://www.selenium.dev/blog/2026/dotnet-strong-name-signing/

PR #17398, opened 2026-04-27
[bazel] Swap to rules_rs for Rust build rules

dzbarsky:
### 💥 What does this PR do?
This PR migrates to using rules_rs instead of rules_rust for our Cargo lockfile integration and toolchains. rules_rs is a redesign ruleset that can operate directly on Cargo.toml/Cargo.lock without the need for additional bazel lockfiles or a bazel-specific repin operation. The toolchains are also optimized to minimize downloads of duplicative bytes compared to the default rules_rust ones.

### 🔧 Implementation Notes
This was straightforward.

### 🤖 AI assistance
Codex 0-shot :) It followed the instructions in rules_rs README perfectly.

N/a

### 🔄 Types of changes
Cleanup / build ergonomics improvement

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -70,20 +70,15 @@ common --host_per_file_copt=external/.*grpc.*@--GRPC_WAS_NOT_SUPPOSED_TO_BE_BUIL
 # Required to get `protobuf` compiling, which is required for `rules_closure`
 common --incompatible_enable_proto_toolchain_resolution
 build --incompatible_enable_cc_toolchain_resolution
-build:linux --cxxopt=-std=c++17
-build:linux --host_cxxopt=-std=c++17
-build:macos --cxxopt=-std=c++17
-build:macos --host_cxxopt=-std=c++17
-build:windows --cxxopt=/std:c++17
-build:windows --host_cxxopt=/std:c++17
+build --cxxopt=-std=c++17
+build --host_cxxopt=-std=c++17
 build:windows --per_file_copt=external/protobuf\\+.*@/w
 build:windows --host_per_file_copt=external/protobuf\\+.*@/w
-common --define=protobuf_allow_msvc=true
 
 # For build stamping
 common --enable_platform_specific_config
 common:linux --host_platform=//:local_linux_gnu
-common:windows --host_platform=//:local_windows_msvc
+common:windows --host_platform=//:local_windows_gnullvm
 build:linux --workspace_status_command=scripts/build-info.sh
 build:macos --workspace_status_command=scripts/build-info.sh
 build:windows --workspace_status_command="powershell.exe scripts/build-info.ps1"
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -10,7 +10,7 @@ bazel_dep(name = "bazel_features", version = "1.45.0")
 bazel_dep(name = "bazel_skylib", version = "1.9.0")
 bazel_dep(name = "buildifier_prebuilt", version = "8.5.1")
 bazel_dep(name = "contrib_rules_jvm", version = "0.32.0")
-bazel_dep(name = "llvm", version = "0.7.3")
+bazel_dep(name = "llvm", version = "0.7.9")
 bazel_dep(name = "package_metadata", version = "0.0.7")
 bazel_dep(name = "platforms", version = "1.0.0")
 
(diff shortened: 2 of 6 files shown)


rake_tasks/rust.rake
```diff
@@ -12,22 +12,6 @@ task :build do |_task, arguments|
   Bazel.execute('build', args, '//rust:selenium-manager')
 end
 
-desc 'Update the rust lock files'
```
dzbarsky, line 15: i'm not familiar with rake, is there anything additional that needs to be done here?

MODULE.bazel
```diff
@@ -386,15 +391,44 @@ use_repo(ruby, "bundle", "ruby", "ruby_toolchains")
...
+        "aarch64-apple-darwin",
+        "aarch64-pc-windows-msvc",
+        "aarch64-unknown-linux-gnu",
+        "x86_64-apple-darwin",
+        "x86_64-pc-windows-msvc",
+        "x86_64-unknown-linux-gnu",
+    ],
+    use_experimental_platforms = True,
+)
+crate.annotation(
```
dzbarsky, line 436: rules_rs has recommended annotations for these to swap to rules_cc based builds, but I kept things as they were for this change

rake_tasks/rust.rake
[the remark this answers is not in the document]
dzbarsky: Removing the `rust:update` call, do we need to keep `rust:update` as a no-op to avoid breaking workflows or is it fine to remove?

AutomatedTester:
@dzbarsky could you sign the CLA before we can look to merge it?

dzbarsky:
> @dzbarsky could you sign the CLA before we can look to merge it?

yep, all good!

rake_tasks/rust.rake
shs96c: This should be fine, I think.

.bazelrc
```diff
@@ -81,7 +81,9 @@ build:windows --host_per_file_copt=external/protobuf\\+.*@/w
 common --define=protobuf_allow_msvc=true
 
 # For build stamping
-build --enable_platform_specific_config
+common --enable_platform_specific_config
+common:linux --host_platform=//:local_linux_gnu
```
shs96c, line 80: Why do we need these changes? The local cc_toolchain setup should already have set these to the right values. I also notice that `macos` is skipped.

dzbarsky: it's because rules_rs, unlike rules_rust, supports multiple target triples per OS (*-linux-gnu and *-linux-musl, *-windows-gnullvm, *-windows-gnu, *-windows-msvc). It was not possible to target musl and non-msvc windows in rules_rust. However, this means there's an extra constraint needed to distinguish these targets, and since we need to also build targeting the host platform (for proc macros and cfg=exec targets like build scripts), we need to specify the constraint on the host as well.

Macos doesn't have these variants, hence it doesn't need the extra configuration! (and yes this was tested on macos)

rake_tasks/rust.rake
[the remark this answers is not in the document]
cgoldberg: `rust:update` is still called in `release.yml` and `pre-release.yml` workflows. It would be better to update those instead of keeping it as a no-op.

dzbarsky:
@cgoldberg thanks for pointing out the `rust:update` bit, I *think* I updated the callers but I'm not sure if I fully understand the templating that was going out there, mind sanity checking? I pushed those bits as a separate commit in this branch to make it easy!

cgoldberg:
@dzbarsky yea, I think your workflow changes look correct. thanks.

.bazelrc
shs96c: OK. Thanks for clarifying.

dzbarsky:
@cgoldberg I had a CI failure with the remote build using a platform without the constraint and thus failing toolchain resolution. Updated and repushed, could you help trigger CI? Would love a bit of help getting this in!

cgoldberg:
> could you help trigger CI?

It's running now

dzbarsky:
> > could you help trigger CI?
> 
> It's running now

hmm i'm not sure how target selection is implemented but Rust job was skipped, that seems dangerous :)

cgoldberg:
> Rust job was skipped

I'm not sure how the target selection works either, and I can't figure out how to trigger the Rust job manually on your branch :/

cgoldberg:
CI is failing all over the place because GitHub Actions is having issues right now. I'll try to figure it out once they get things stabilized.

cgoldberg:
build is failing on Windows: https://github.com/SeleniumHQ/selenium/actions/runs/25440138960/job/74647344814?pr=17398#step:23:196

dzbarsky:
> build is failing on Windows: https://github.com/SeleniumHQ/selenium/actions/runs/25440138960/job/74647344814?pr=17398#step:23:196

Yep I need to fix @llvm module. Mentioned to Simon on slack but should have posted here as well to keep you in the loop. My bad!

PR #17412, opened 2026-05-01
[rb] Modern Firefox does not like both the `-v` and `--log` flags at the same time

shs96c:


rb/lib/selenium/webdriver/firefox/service.rb
```diff
@@ -33,25 +33,26 @@ def initialize(args: nil, **)
...
-          if (index = args.index('--log'))
-            args.delete_at(index) # delete '--log'
-            args.delete_at(index) if args[index] && !args[index].start_with?('-') # delete value if present
-            warn_driver_log_override
-          elsif (index = args.index { |arg| arg.start_with?('--log=') })
-            args.delete_at(index)
-            warn_driver_log_override
+        def configure_debug_args(args)
+          if args.any? { |arg| arg.start_with?('--log') }
+            args.reject! { |arg| /\A-v+\z/.match?(arg) }
```
aguspe: so SE_DEBUG does not overrides the user's --log and the warning's gone, is that what we want? and why?

shs96c: Good point. I've amended the PR.

luke-hill: Quick question / Minor nit. Looking at the regex here we're checking absolute starts. This can be more optimally done with `#start_with?` unless I'm misreading.

So you could write.

```
args.reject! { |arg| arg.start_with?('-v') }
```

PR #17425, opened 2026-05-08
[java] add test to prevent updating pinned chrome version before cdp

titusfortner:
### 💥 What does this PR do?

I don't think we should be testing later versions of Chrome than we support with CDP for a few reasons.

### 🔧 Implementation Notes
Add a python test to scripts so bazel should pick it up when the change is made

### 🤖 AI assistance

    - Tool(s): Codex
    - What was generated: regex/boilerplate


titusfortner:
yeah, the right fix is just to release faster.

PR #17427, opened 2026-05-08
[rust] Allow cross-compilation of `selenium-manager` on all platforms

shs96c:


MODULE.bazel
```diff
@@ -405,7 +405,9 @@ use_repo(rust_toolchains, "default_rust_toolchains")
 
 register_toolchains("@default_rust_toolchains//:all")
 
-# TODO: Register only Windows exec for now, full @llvm migration can be done as followup.
+# Register Windows + macOS + Linux execs so the per-platform Selenium
+# Manager variants in //rust can be cross-compiled on any developer host
+# and on RBE.
 llvm_toolchains = use_extension("@llvm//extensions:toolchain.bzl", "toolchain")
```
dzbarsky, line 424: you can actually snip this entire extension and its configuration, just do `register_toolchains("@llvm//toolchain:all")`. I only did this to scope down what i was registering to avoid messing with the other platforms :)

MODULE.bazel
```diff
@@ -445,6 +463,25 @@ crate.annotation(
...
+crate.annotation(
+    crate = "windows-link",
+    patch_args = ["-p1"],
+    patches = ["//third_party/bazel:windows_link_readme_case.patch"],
+)
+crate.annotation(
+    crate = "windows-targets",
+    patch_args = ["-p1"],
+    patches = ["//third_party/bazel:windows_targets_readme_case.patch"],
+    # Only the 0.53 line has this case mismatch; 0.52.x ships a lowercase
```
dzbarsky, line 482: oh, they regressed it? wild

rust/BUILD.bazel
```diff
@@ -7,79 +10,128 @@ rustfmt_config(
...
-        "manual",
-    ],
-    visibility = [
-        "//common/manager:__pkg__",
-    ],
-)
+# Each (OS, arch) variant cross-compiles directly via rules_rs
+#
+# Linux targets use musl rather than glibc because Rust's prebuilt glibc
+# stdlib hard-references `-lgcc_s`, which the @llvm toolchain doesn't ship.
```
dzbarsky, line 21: btw this can be made to work by passing `--@llvm//config:experimental_stub_libgcc_s=True` flag (which we will actually enable by default I think) but it's probably still nicer to ship a statically-linked musl variant and not worry about glibc versioning

rust/BUILD.bazel
```diff
@@ -7,79 +10,128 @@ rustfmt_config(
...
+# musl produces a fully self-contained binary that runs on any Linux
+# distribution.
+#
+# Windows targets use the LLVM-based MinGW (gnullvm) ABI rather than MSVC
```
dzbarsky, line 25: we also have in-flight work to allow shipping fully-hermetic MSVC variants as well, it requires patching rules_cc and llvm though. Just mentioning in case you prefer that route, see https://github.com/hermeticbuild/toml2json/blob/master/MODULE.bazel#L8-L76

(we're going to make those hashes part of defaults/etc when this is ready to ship for real)

rust/BUILD.bazel
```diff
@@ -1,3 +1,6 @@
+load("@aspect_bazel_lib//lib:copy_to_directory.bzl", "copy_to_directory")
```
dzbarsky, line 1: nit: this module got renamed to bazel_lib as part of 3.0. Not sure if yall are using both already or still need to migrate

rb/BUILD.bazel
```diff
@@ -12,22 +13,10 @@ package(default_visibility = ["//:__subpackages__"])
...
-    name = "manager-macos",
-    src = "//common/manager:selenium-manager-macos",
-    out = "bin/macos/selenium-manager",
-)
-
-copy_file(
-    name = "manager-windows",
-    src = "//common/manager:selenium-manager-windows",
-    out = "bin/windows/selenium-manager.exe",
+copy_directory(
```
dzbarsky: i wonder if it might make sense to turn that bundle rule into a macro that can be invoked in each language package, that might prevent some directory copies?

PR #17435, opened 2026-05-11
[grid] Close pre-handshake race in WebSocket proxy

shs96c:
## Summary

Frames received from the upstream WebSocket before the client-side handshake completes used to travel through `MessageOutboundConverter` via a fallback `Consumer<Message>`. After `onUpgradeComplete` rewires the pipeline (removing the Message-layer handlers), a frame in flight on the upstream listener thread could land in a pipeline that no longer had its converter and silently drop.

Move the client `Channel` reference and a pre-handshake buffer into `DirectForwardingListener` itself:

- Pre-handshake `onText` / `onBinary` calls queue a `WebSocketFrame` onto a per-listener deque under a lock.
- `FrameProxyConsumer.onUpgradeComplete` invokes `listener.onUpgrade(channel)`, which drains the buffer in arrival order and *then* publishes the channel as `volatile`.
- Post-handshake calls do a lock-free volatile read of the channel and write directly — the fast path is unchanged.
- `onClose` / `onError` pre-handshake release buffered frames and latch a `closed` flag so any in-flight call doesn't leak ref-counted frames.

A focused `EmbeddedChannel`-based test pins the buffer-then-drain order; the existing medium `ProxyWebsocketTest` continues to cover the end-to-end Router→Node path.

## Test plan

- [x] \`bazel test --config=rbe //java/test/org/openqa/selenium/grid/router:small-tests\`
- [x] \`bazel test --config=rbe //java/test/org/openqa/selenium/grid/router:ProxyWebsocketTest\`
- [x] \`bazel test --config=rbe //java/test/org/openqa/selenium/grid/router:TunnelWebsocketTest\`

### DIFF
--- a/java/src/org/openqa/selenium/grid/router/ProxyWebsocketsIntoGrid.java
+++ b/java/src/org/openqa/selenium/grid/router/ProxyWebsocketsIntoGrid.java
@@ -337,9 +337,11 @@ public class ProxyWebsocketsIntoGrid
         synchronized (lock) {
           ch = clientChannel;
           if (ch == null) {
-            // Pre-handshake close: record the close so onUpgrade can surface it to the client.
-            // Pending frames are kept so the client receives any data the upstream queued before
-            // closing.
+            // Pre-handshake close: drop the buffer so ref-counted frames are released
+            // immediately even if the client-side handshake never lands (for instance because
+            // the client disconnected mid-handshake or the upgrade itself failed). Record the
+            // close so onUpgrade can still surface it to the client if the handshake does fire.
+            discardPendingLocked();
             closed = true;
             closeCode = code;
             closeReason = reason;
--- a/java/test/org/openqa/selenium/grid/router/DirectForwardingListenerTest.java
+++ b/java/test/org/openqa/selenium/grid/router/DirectForwardingListenerTest.java
@@ -60,30 +60,27 @@ class DirectForwardingListenerTest {
   }
 
   @Test
-  void preHandshakeCloseIsSurfacedAndClosesChannelOnUpgrade() {
+  void preHandshakeCloseSurfacesCloseAndReleasesBufferedFrames() {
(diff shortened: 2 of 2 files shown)


shs96c:
Pushed `bf52afc` addressing both qodo-bot findings:

1. **Pre-upgrade close not forwarded** — `DirectForwardingListener` now records the close code + reason when `onClose`/`onError` fires pre-handshake. On `onUpgrade` it drains any frames the upstream had queued, then writes a `CloseWebSocketFrame` with the recorded details and closes the channel, instead of publishing it for normal forwarding. New test `preHandshakeCloseIsSurfacedAndClosesChannelOnUpgrade` pins the behaviour.

2. **Unbounded frame buffer** — capped `pending` at 128 frames. On overflow the buffer is released, the listener latches a 1009 ("buffer overflow") terminal state, and the upstream client is closed; the eventual `onUpgrade` sees the terminal state and surfaces a clean 1009 close to the client. New test `overflowOfPreHandshakeBufferArmsCloseAndDrains`.

The Ruby Windows test failure is unrelated — `gpg: error retrieving '<key>' via WKD: Try again later` during MSYS2 toolchain setup. No Java tests ran on that job before it failed.

shs96c:
Good catch — pushed `0cd57dd` addressing it.

`onClose` pre-handshake used to keep the buffered ref-counted frames "for a future `onUpgrade` drain". If the client-side handshake never lands (client disconnected mid-handshake, or the upgrade itself failed and triggered the upstream close), those frames are never released.

Made `onClose` symmetric with `onError` — drop the buffer immediately. The code/reason are still recorded so a late `onUpgrade` can write a proper close to the client, just without the trailing data. The trade-off is small: a client that has just received the 101 cannot usefully consume frames if we are about to close them immediately afterward.

Existing test updated to reflect the new behaviour (only the close frame on the wire after pre-handshake close + upgrade).

PR #17441, opened 2026-05-12
[grid] Bundle Redis-backed SessionMap by default

VietND96:
### 💥 What does this PR do?
- Redis-backed SessionMap is available out of the box with no extra configuration. `selenium-session-map-redis` was not bundled in selenium-server.jar by default, requiring users to manually manage it on the classpath - unlike distributor/redis, which has been built-in since #17396. Generic is Redis-backed support as default cross components.
- Bug fix: null startKey in RedisBackedSessionMap.get()
- Increase test coverage (RedisBackedSessionMapTest.java)

### 🤖 AI assistance

    - Tool(s):
    - What was generated:

### 🔄 Types of changes

### DIFF
--- a/.github/workflows/commit-changes.yml
+++ b/.github/workflows/commit-changes.yml
@@ -17,7 +17,7 @@ on:
         type: string
         default: ''
       push-branch:
-        description: Branch to push to (defaults to current branch); always force-pushes
+        description: Branch to force push to; if unset, pushes to the checked-out branch without force
         required: false
         type: string
         default: ''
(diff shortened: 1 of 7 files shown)


diemol:
Does this mean we're increasing the size of the jar file? What is the size difference since we have added this to the distributor as well? We should pay attention to that so we don't bloat the server.

VietND96:
@diemol. Yes, the diff is 42.3 MB...51M. The same size applies to cross components (Distributor, SessionMap, and later SessionQueue). Do you think it's worth it?

diemol:
Should be OK.

PR #17467, opened 2026-05-15
[dotnet] include snupkg files when packaging things up and allow the use of sourcelink

AutomatedTester:
### 🤖 AI assistance

    - Tool(s):
    - What was generated:

### 🔄 Types of changes

### DIFF
--- a/dotnet/private/sourcelink.bzl
+++ b/dotnet/private/sourcelink.bzl
@@ -1,0 +1,93 @@
+"""C# library rule with SourceLink metadata embedded in PDB files."""
+
+load("@bazel_skylib//rules:common_settings.bzl", "BuildSettingInfo")
+load("@rules_dotnet//dotnet/private:common.bzl", "is_debug")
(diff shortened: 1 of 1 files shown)


dotnet/private/nuget_push.bzl
```diff
@@ -38,6 +38,8 @@ def _create_unix_script(ctx, dotnet, nupkg_files):
     """Create bash script for Unix/macOS/Linux."""
     push_commands = []
     for nupkg in nupkg_files:
+        if nupkg.basename.endswith(".snupkg"):
```
shs96c, line 41: It looks like we're deliberately not uploading the source packages. Is there a reason? Do we need to push the source and binary package at the same time?

AutomatedTester: the skip is intentional because dotnet nuget push on the .nupkg auto-detects the co-located .snupkg and pushes both together. The snupkg still ends up on NuGet, it just gets pushed implicitly rather than explicitly

PR #17471, opened 2026-05-15
[dotnet] include .snupkg in nuget_push runfiles so symbols get published

titusfortner:
When I was looking at SourceLink things, the Agent said we were releasing wrong for the new snupkg (or at least not ideally). This is the suggested fix. @nvborisenko can you verify?

### 💥 What does this PR do?

nuget_pack produces both a .nupkg (main package) and a .snupkg (symbol/debug package) per target. Previously, allow_files = [".nupkg"] filtered the .snupkg out when nuget_push collected files via ctx.files.packages. That meant the symbol package never made it into nupkg_files, and therefore never made it into the runfiles tree built on line 16 (ctx.runfiles(files = nupkg_files + [dotnet])).

snupkg is now in the iteration list (because of change [#1](https://github.com/SeleniumHQ/selenium/issues/1)), but we don't want a separate push command for it. We let dotnet nuget push handle symbols via its built-in auto-detection

dotnet nuget push foo.nupkg automatically looks for foo.snupkg in the same directory and uploads it as the symbol package. nuget_pack declares both files with the same stem (e.g. Selenium.WebDriver.4.27.0.nupkg and Selenium.WebDriver.4.27.0.snupkg), so once the .snupkg is in runfiles, it lands right next to its .nupkg and dotnet finds it on its own.

### 🤖 AI assistance

    - Tool(s): Claude
    - What was generated: this PR

### DIFF
--- a/dotnet/private/nuget_push.bzl
+++ b/dotnet/private/nuget_push.bzl
@@ -38,6 +38,8 @@ def _create_unix_script(ctx, dotnet, nupkg_files):
     """Create bash script for Unix/macOS/Linux."""
     push_commands = []
     for nupkg in nupkg_files:
+        if not nupkg.basename.endswith(".nupkg"):
+            continue
         nupkg_runfiles_path = _to_runfiles_path(nupkg.short_path)
         push_commands.append(
             '"$DOTNET" nuget push "$RUNFILES_DIR/{nupkg}" --api-key "$NUGET_API_KEY" --source "$NUGET_SOURCE" --skip-duplicate'.format(nupkg = nupkg_runfiles_path),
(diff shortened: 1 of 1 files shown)


AutomatedTester:
I think #17467 should be used instead of this PR

nvborisenko:
`dotnet nuget push *.nupkg` auto-detects `*.snupk`. Currently it works as expected, symbols are pushed to the remote.
Covered by https://github.com/SeleniumHQ/selenium/pull/17467 - we can merge this one as minimal changes. Before merging we should "test" whether `nupkg` and `snupkg` are located in the same directory.

PR #17477, opened 2026-05-16
[dotnet] update system.text.json to 8.0.6

titusfortner:
User facing change but want to use the latest, should be no issues merging after tests


nvborisenko:
User facing dependency, unnecessary upgrade. In .NET we should target to minimal version as we can.

PR #17480, opened 2026-05-16
[build] bump bazel version to 9.1

titusfortner:
Shouldn't be any issues in this update, though also don't see any features we can use, so bumping to stay up to date


titusfortner:
Running https://github.com/SeleniumHQ/selenium/actions/runs/25963122292 as well

PR #17457, opened 2026-05-14
[dotnet] [build] Format solution instead of projects

nvborisenko:
2x faster

### 💥 What does this PR do?
Use SDK with setting up workspace once, instead calling it one by one per project.

### 🤖 AI assistance

- [ ] No substantial AI assistance used
- [x] AI assisted (complete below)
    - Tool(s):
    - What was generated:
    - [x] I reviewed all AI output and can explain the change

### 🔄 Types of changes

- Cleanup (formatting, renaming)

PR #17483, opened 2026-05-16
[dotnet] run format against slnx instead of looping csproj

titusfortner:
### 🔗 Related Issues
Implements #17457

### 💥 What does this PR do?
Uses slnx instead of looping over csproj. Format commit time is cut in half.

### 🤖 AI assistance

    - Tool(s): Codex
    - What was generated: everything


titusfortner:
Interesting, this works locally on my mac, even when I delete cache and things, adding diagnostics to debug issue on windows.

titusfortner:
There we go, it was erroring based on Pruned data on Windows, these settings fixed it. verified that time on CI is much less as well as local

PR #17492, opened 2026-05-17
[java] publish selenium-devtools-latest alias artifact

asolntsev:
Instead of writing 
```java
import org.openqa.selenium.devtools.v148.page.Page;
```

and updating "v148" part with every new release, now users can just write forever:

```java
import org.openqa.selenium.devtools.latest.page.Page;
```

Similar to PR https://github.com/SeleniumHQ/selenium/pull/17133, but for Java bindings.

## Summary

- Adds a new Maven artifact `org.seleniumhq.selenium:selenium-devtools-latest` that republishes the most recent supported CDP version under the stable `org.openqa.selenium.devtools.latest` package. Downstream consumers no longer have to bump a version-suffixed dependency (`selenium-devtools-vNNN`) each time Chrome ships a new major release.
- Generated CDP code is produced by running `CdpClientGenerator` with `"latest"` against the protocol files of `LATEST_CDP_VERSION`; the hand-written shim classes are derived by sed-rewriting the corresponding files from the latest `vNNN` package at build time, so nothing is duplicated on disk.
- `LATEST_CDP_VERSION` is computed as `max(CDP_VERSIONS, key = lambda v: int(v[1:]))` so the alias keeps pointing at the highest-numbered entry even after `scripts/update_cdp.py` rewrites `versions.bzl` in place (which can leave the list unsorted).
- No source files inside the new `latest/` directory — only `BUILD.bazel`. Adding a new CDP version (and removing the oldest) via `./go update_cdp` continues to "just work"; the `shim-sources` filegroup added to `v148/BUILD.bazel` propagates forward through `update_cdp.py`'s copy step.

## Test plan

- [x] `bazelisk build //java/src/org/openqa/selenium/devtools/latest:latest`
- [x] Inspected `latest-project.jar`: contains both shim classes (`latestCdpInfo`, `latestDomains`, …) and all generated CDP domain classes (`Fetch`, `Network`, `Target`, …) under `org/openqa/selenium/devtools/latest/`, plus a `META-INF/services/org.openqa.selenium.devtools.CdpInfo` entry registering `latestCdpInfo`.
- [x] `bazelisk build //java/src/org/openqa/selenium:client-combined //java/src/org/openqa/selenium/grid:grid` — existing consumers of `CDP_DEPS` still build.
- [x] `./go format`

### DIFF
--- a/java/spotbugs-excludes.xml
+++ b/java/spotbugs-excludes.xml
@@ -220,6 +220,11 @@
   </Match>
 
   <Match>
+    <Class name="~org.openqa.selenium.devtools.latest\.latest\w+" />
+    <Bug pattern="NM_CLASS_NAMING_CONVENTION"/>
+  </Match>
+
+  <Match>
     <Class name="org.openqa.selenium.manager.SeleniumManager"/>
     <Bug pattern="NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE,IS2_INCONSISTENT_SYNC"/>
   </Match>
--- a/java/src/org/openqa/selenium/devtools/README.md
+++ b/java/src/org/openqa/selenium/devtools/README.md
@@ -7,7 +7,11 @@ They are typically downloaded from the
 [devtools source](https://github.com/ChromeDevTools/devtools-protocol/tree/master/json)
 
 * Edit `//java/src/org/openqa/selenium/devtools:versions.bzl` to include the new version number (and
-  possibly delete the old one, we tend to support only the last 3 versions)
+  possibly delete the old one, we tend to support only the last 3 versions). The last entry of
+  `CDP_VERSIONS` is automatically picked up as `LATEST_CDP_VERSION` and republished under the
+  `org.openqa.selenium.devtools.latest` package by
+  `//java/src/org/openqa/selenium/devtools/latest`, so users get the new CDP version under the
+  stable `selenium-devtools-latest` Maven coordinate without any extra work.
 * Copy the most recent `//java/src/org/openqa/selenium/devtools/vXX` to
   `//java/src/org/openqa/selenium/devtools/vXX+1`.
 * Do a search and replace in that directory, converting `XX` to `XX+1`
--- a/java/src/org/openqa/selenium/devtools/latest/BUILD.bazel
+++ b/java/src/org/openqa/selenium/devtools/latest/BUILD.bazel
@@ -2,22 +2,17 @@ load("@rules_jvm_external//:defs.bzl", "artifact")
 load("//common:defs.bzl", "copy_file")
 load("//java:defs.bzl", "java_export", "java_library")
 load("//java:version.bzl", "SE_VERSION")
+load("//java/src/org/openqa/selenium/devtools:versions.bzl", "LATEST_CDP_VERSION")
 
-cdp_version = "v148"
-
-# Hand-written shim sources reused by //java/src/org/openqa/selenium/devtools/latest
-# to produce a copy of this package under org.openqa.selenium.devtools.latest.
-filegroup(
-    name = "shim-sources",
-    srcs = glob(["*.java"]),
-    visibility = [
-        "//java/src/org/openqa/selenium/devtools/latest:__pkg__",
-    ],
-)
+# Publishes the most recent supported CDP version under the stable
+# `org.openqa.selenium.devtools.latest` package, so downstream consumers can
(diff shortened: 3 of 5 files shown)


titusfortner:
Except this will end up being the default people use even if things get out of sync. Maybe we label it @Beta or call it `unstable` instead of `latest` as a warning to users?
One of the reasons we supported it this way was to highlight the potential issues with CDP, and to encourage the switch to BiDi (which obviously is a better argument when we actually support BiDi)

We probably need to delete `@AutoService` annotation for it as well

asolntsev:
> Except this will end up being the default people use even if things get out of sync. Maybe we label it @beta or call it `unstable` instead of `latest` as a warning to users? One of the reasons we supported it this way was to highlight the potential issues with CDP, and to encourage the switch to BiDi (which obviously is a better argument when we actually support BiDi)
> 
> We probably need to delete `@AutoService` annotation for it as well

Users are very different.
I assume most users (including me) use only a very small set of CDP functions that almost never change.
For such users, "latest" is just perfect. 
We don't want to change these versions with every browser update. It's a waste of time. 
And yes, we are always ready for the risk that something might go wrong after updating browser/selenium/selenide/java/gradle etc. It's not a big deal, it's just our everyday job.

asolntsev:
> Do we still show a warning when the browser version is higher than the ones the Selenium version the user has is offering?

Yes, just checked with Selenium 4.43.0:
```
May 18, 2026 7:48:54 PM org.openqa.selenium.devtools.CdpVersionFinder findNearestMatch
WARNING: Unable to find an exact match for CDP version 148, returning the closest version; found: 147; Please update to a Selenium version that supports CDP version 148
```

titusfortner:
Implementation issues, though. What if we make selenium-devtools-latest a thin alias over the current vXXX artifact rather than a self-contained copy.
  
With this PR code as implemented, the new artifact ships its own sed-rewritten shim classes (latestCdpInfo, latestDomains, etc.) with @AutoService(CdpInfo.class). That creates a duplicate CdpInfo registration for the same major version when both selenium-devtools-latest and selenium-devtools-v<LATEST> are on the classpath (which is the default for selenium-java consumers,  since CDP_DEPS now contains both). ServiceLoader returns both, and CdpVersionFinder.findNearestMatch exact-matches on whichever is iterated first — functionally equivalent, but  observably non-deterministic and a noticeable jar-size duplication.
  
An alternative design that I think is cleaner:

  - selenium-devtools-latest contains only the generated CDP protocol types under org.openqa.selenium.devtools.latest.* (Page, Network, Fetch, etc. — the things users actually  import).
  - Drop the shim entirely from this artifact (no latestCdpInfo, no latestDomains, no @AutoService).
  - Declare a Maven dependency on selenium-devtools-v<LATEST_CDP_VERSION> so the actual CdpInfo registration is provided transitively by the versioned artifact.

  Trade-offs this resolves:

  1. No duplicate registration, ever. v148CdpInfo is the only registered impl for major 148; latest is purely an alias namespace.
  2. No shim file maintenance. The hardcoded outs list in create-shim-srcs, the shim-sources filegroup in v148/BUILD.bazel, and the spotbugs NM_CLASS_NAMING_CONVENTION exemption all
  go away. No risk of a future v149Foo.java being silently dropped from the alias.
  3. mvn dependency:tree is self-documenting — users see selenium-devtools-latest → selenium-devtools-v148 and know which CDP version they're actually running.
  4. No behavioral drift risk between v148Domains and a sed-copy latestDomains, because the sed-copy doesn't exist.

  User-facing experience is identical: imports like org.openqa.selenium.devtools.latest.page.Page still work, Selenium upgrades pick up the new CDP under the same coordinate, Chrome
  upgrades degrade gracefully via the existing fudge-factor logic.

asolntsev:
@titusfortner Sorry, I didn't understand what means "thin alias".

For the end user, the only important thing that I don't want to change this import every time:
```java
import org.openqa.selenium.devtools.v148.page.Page;
```

Instead, I want to have the constant import:
```java
import org.openqa.selenium.devtools.latest.page.Page;
```

How exactly it's achieved - it doesn't matter. 
But essentially, this "thin alilas" needs to contain classes like `Page` in a different packages ("latest").

titusfortner:
@asolntsev I was also making it too complicated. We can just let the generator do all of the work and avoid all the problems: #17562 I built it and everything looks correct.

asolntsev:
Seems that https://github.com/SeleniumHQ/selenium/pull/17562 works and fully solves the problem. 
Thank you @titusfortner!

PR #17403, opened 2026-04-28
[py][build] Update bazel rules_python

cgoldberg:
### 💥 What does this PR do?

Bump bazel rules_python to 2.0.1 (no toolchain version upgrades in this release)

### 🤖 AI assistance

- [x] No substantial AI assistance used
- [ ] AI assisted (complete below)
    - Tool(s):
    - What was generated:
    - [ ] I reviewed all AI output and can explain the change

### 🔄 Types of changes

- Build

PR #17502, opened 2026-05-18
[build][py] bump rules_python to v2

titusfortner:
### 🔗 Related Issues
Follow on to #17461 

### 💥 What does this PR do?
Update to rules_python v2


cgoldberg:
this is the same as #17403 ... I'll close mine

titusfortner:
We can do a workaround by replacing all "/" with "_" but I think we wait until fix is made upstream.

titusfortner:
rules_ruby 2.0.2 is out, trying again.

titusfortner:
Need to wait for rules_python 2.0.3 to be released

titusfortner:
rules_python 2.x uses much longer canonical repo names, so Windows runfiles paths now exceed MAX_PATH (260) and tests fail.

The failure is specifically in DLL loading: Windows' loader (LoadLibrary) ignores long-path support, so a .pyd over 260 chars can't be imported. `.py` files are fine — file I/O honors long paths; only the loader is capped. This is a known Windows OS limitation that can't be avoided.

Here's an example:
```
D:\b\execroot\_main\bazel-out\x64_windows-fastbuild\bin\
py\test\unit\selenium\webdriver\virtual_authenticator\virtual_authenticator_options_tests-unit.exe.runfiles\  
rules_python++pip+pydev_310_cffi_cp310_cp310_win_amd64_b18a3ed7\ 
site-packages\_cffi_backend.cp310-win_amd64.pyd
```

The best fix is not to require the DLL loading when not necessary, which would mean getting rid of cffi dependencies everywhere we can, which is only required by trio, which is only required by CDP. This should be a straightforward fix.

Other tweaks we can make:
* In `MODULE.bazel` update pip `hub_name` from `"py_dev_requirements"` to shorter `"pydev"`
* de-stutter test paths (`virtual_authenticator\virtual_authenticator_options_tests` --> `virtual_authenticator\options_tests`)
* consider removing webkit tests since not much value/usage from them

PR #17523, opened 2026-05-19
[java] Prevent byte-buddy -jdkN variant from being picked by dependency updates

### LINKED ISSUE #17355: [🐛 Bug]: Selenium should depend on net.bytebuddy:byte-buddy 1.18.8 instead of 1.18.8-jdk5

pierluigilenoci:
## Description

Byte Buddy publishes `-jdkN` compatibility variants alongside its regular releases. Maven Central indexes those variants as newer than the regular release (e.g. `1.18.8-jdk5` sorts higher than `1.18.8`), so the automated dependency update task ([`rake_tasks/java.rake`](https://github.com/SeleniumHQ/selenium/blob/trunk/rake_tasks/java.rake) — `java:update`, called from `./go all:update`) keeps picking the `-jdkN` flavor even though Selenium targets Java 8+ and only needs the regular release.

The Maven enforcer plugin then complains because `1.18.8-jdk5` sorts as greater than the managed dependency `1.18.8`.

This was first fixed in #17382 by reverting `MODULE.bazel` to `1.18.8`. The next automated dependency update silently reintroduced `1.18.8-jdk5`, which is what #17355 reports.

## Fix

Add a guard in the `java:update` Rake task that detects `net.bytebuddy:*` artifacts whose proposed version carries a `-jdkN` suffix and falls back to `maven_stable_release(artifact)`, which already filters to versions matching `\A\d+\.\d+(\.\d+)*\z` and therefore excludes any `-jdkN` variant.

The guard mirrors the existing `graphql` exception in the same loop.

Also revert `MODULE.bazel` from `1.18.8-jdk5` back to `1.18.8` and re-pin `java/maven_install.json` so the repo state is consistent again.

## Changes

- `rake_tasks/java.rake` — add `net.bytebuddy:*` `-jdkN` guard in `java:update` (calls `maven_stable_release` instead of using the version returned by `@maven//:outdated`)
- `MODULE.bazel` — `1.18.8-jdk5` → `1.18.8`
- `java/maven_install.json` — regenerated via `RULES_JVM_EXTERNAL_REPIN=1 bazel run @maven//:pin`

## Verification

- Re-pinned successfully: `byte-buddy:jar` lockfile metadata now reads `"version": "1.18.8"`.
- The new branch in `versions.each` matches the format of the existing `graphql` exception, so it is covered by the same `next if version.nil?` safety net when `maven_stable_release` cannot resolve.

Fixes #17355

Status: draft until CI is green.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -224,7 +224,7 @@ maven.install(
         "io.opentelemetry:opentelemetry-sdk-trace",
         "org.testcontainers:testcontainers:2.0.5",
         "javax.inject:javax.inject:1",
-        "net.bytebuddy:byte-buddy:1.18.8-jdk5",
+        "net.bytebuddy:byte-buddy:1.18.8",
         "org.htmlunit:htmlunit-core-js:4.21.0",
         "org.apache.logging.log4j:log4j-core:2.26.0",
         "org.assertj:assertj-core:3.27.7",
--- a/java/maven_install.json
+++ b/java/maven_install.json
@@ -52,7 +52,7 @@
     "io.opentelemetry:opentelemetry-sdk-trace": 1165401435,
     "javax.inject:javax.inject": -297932879,
     "junit:junit": -744267592,
-    "net.bytebuddy:byte-buddy": 1750588375,
+    "net.bytebuddy:byte-buddy": 731630558,
     "net.java.dev.javacc:javacc": 520686333,
     "org.apache.logging.log4j:log4j-core": 2067797603,
     "org.assertj:assertj-core": -1145412507,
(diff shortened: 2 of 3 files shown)


titusfortner:
CI "failure" is unrelated, I'm going to merge it.

Issue #16095, opened 2025-07-27
[🚀 Feature]: [dotnet] [bidi] Finalize public API to prevent further breaking changes


nvborisenko:
### Description

Let's see how confident we are about BiDi API. In general I like the applied approach, but with some concerns.

- [ ] `driver.AsBiDiAsync()` extension method should be finally live in `AnyDriver` class. Thus `AnyDriver` object is an owner of the `BiDi` instance, and dispose `BiDi` as soon as owner is disposed. Not sure, but I think `IWebDriver` interface is a great candidate to have this new `AsBiDiAsync()` method.
It is important. Finally we will have straightforward dependency: `Selenium.Webdriver` -> `Selenium.WebDriver.BiDi`.
- [x] Really expose all hidden methods from users, such as `AddIntercept()`. So just make it `public` instead of `internal`, that's it. https://github.com/SeleniumHQ/selenium/issues/15612
- [x] Revisit `InterceptRequestAsync()` method (and neighbors: response, auth). This method(s) provides more convenient API to intercept network. Literally saying it should be extensions built on top of already publicly available methods.
UPD: .NET 10 provides new extensions capabilities, where we can create properties. Amazing, so the decision here would be move all helpers/extensions to dedicated namespace.
RESOLUTION: Removed them at all to be low-level, extension methods are still in mind.
- [x] `EmptyResult` should be valid return type for all methods instead of `void`. https://github.com/SeleniumHQ/selenium/issues/15562
- [x] Normal types as event args (inheritance topic) https://github.com/SeleniumHQ/selenium/issues/15546 **Solution**: introduce non-serializable top-level `*EventArgs`, and construct it from spec defined type (including `IBiDi` hydration!) https://github.com/SeleniumHQ/selenium/pull/17304
- [x] We have more and more strange syntax sugar, like `Result` class implements `IReadOnlyList<T>` - https://github.com/SeleniumHQ/selenium/blob/55f02a983fb3061ff84020ffe945ed60ee948e33/dotnet/src/webdriver/BiDi/BrowsingContext/GetTreeCommand.cs#L50. It is useful definitely, but seems it is overkill. We can create `extensions`, like `.GetTreeAsync().AsContexts()`. In any case I propose to remove `IReadOnlyList<T>`, and more important remove strange json converters for these types. - https://github.com/SeleniumHQ/selenium/pull/16219
[...]

PR #17533, opened 2026-05-20
[dotnet] [bidi] Immutable user facing collections

nvborisenko:
Use `ImmutableArray<T>` instead of `IEnumerable<T>` and `IReadOnlyList<T>`.

Now it is immutable everywhere by contract. Eliminating the following cases at compilation level:

Before:
```csharp
List<string> args = ["a", "b"];
bidi.CallFunctionAsync(.... , Args = args);
args.Clear(); // BUG, collection was modified during serialization
```

After:
```csharp
List<string> args = ["a", "b"];
bidi.CallFunctionAsync(.... , Args = [..args]); // It is always a copy
args.Clear(); // OK, it it not that collection you put into command
```

### 🔗 Related Issues
Contributes to https://github.com/SeleniumHQ/selenium/issues/16095

### 💥 What does this PR do?
Strict intention of contract.

### 🤖 ~~AI assistance~~ Thanks to @RenderMichael and @YevgeniyShunevych 

    - Tool(s):
    - What was generated:

### 🔄 Types of changes

### DIFF
--- a/dotnet/src/webdriver/BiDi/Script/LocalValue.cs
+++ b/dotnet/src/webdriver/BiDi/Script/LocalValue.cs
@@ -89,14 +89,7 @@ public abstract record LocalValue
                 {
                     IEnumerable set = (IEnumerable)value;
 
-                    var setValues = ImmutableArray.CreateBuilder<LocalValue>();
-
-                    foreach (var obj in set)
-                    {
-                        setValues.Add(ConvertFrom(obj));
-                    }
-
-                    return new SetLocalValue(setValues.ToImmutable());
+                    return new SetLocalValue([.. set.Cast<object?>().Select(ConvertFrom)]);
                 }
 
             case IDictionary dictionary:
@@ -208,14 +201,7 @@ public abstract record LocalValue
             return new NullLocalValue();
         }
 
-        List<LocalValue> list = [];
-
-        foreach (var element in value)
-        {
-            list.Add(ConvertFrom(element));
-        }
-
-        return new ArrayLocalValue([.. list]);
+        return new ArrayLocalValue([.. value.Cast<object?>().Select(ConvertFrom)]);
     }
 
     public static LocalValue ConvertFrom(IDictionary? value)


dotnet/src/webdriver/BiDi/Script/LocalValue.cs
```diff
@@ -232,7 +232,7 @@ public static LocalValue ConvertFrom(IDictionary? value)
             bidiObject.Add([ConvertFrom(key), ConvertFrom(value[key])]);
         }
 
-        return new MapLocalValue(bidiObject);
+        return new MapLocalValue([.. bidiObject.Select(p => (ImmutableArray<LocalValue>)[.. p])]);
```
RenderMichael: Can we make `bidiObject` a `List<ImmutableArray<LocalValue>>` to simplify this? The inner list is always two elements anyways.

dotnet/src/webdriver/BiDi/Script/LocalValue.cs
```diff
@@ -244,7 +244,7 @@ public static LocalValue ConvertFrom<T>(ISet<T?>? value)
 
         LocalValue[] convertedValues = [.. value.Select(x => ConvertFrom(x))];
 
-        return new SetLocalValue(convertedValues);
+        return new SetLocalValue([.. convertedValues]);
```
RenderMichael: `convertedValues` can just be an immutable array

dotnet/src/webdriver/BiDi/Script/LocalValue.cs
```diff
@@ -276,7 +276,7 @@ private static LocalValue ReflectionBasedConvertFrom(object? value)
             values.Add([property.Name, ConvertFrom(propertyValue)]);
         }
 
-        return new ObjectLocalValue(values);
+        return new ObjectLocalValue([.. values.Select(p => (ImmutableArray<LocalValue>)[.. p])]);
```
RenderMichael: Same here as with `bidiObject`

dotnet/src/webdriver/BiDi/Script/LocalValue.cs
nvborisenko: Even better, immutable builder. Thank you, nice optimization.

dotnet/src/webdriver/BiDi/Script/LocalValue.cs
nvborisenko: Same ^

dotnet/src/webdriver/BiDi/Script/LocalValue.cs
nvborisenko: Yes, thx.

Issue #17326, opened 2026-04-09
Selenium CLI tool


AutomatedTester:
# **Product Requirements Document: Sel-CLI (v1.2 \- 2026 Edition)**

**Project Focus: Hybrid Rust/TypeScript CLI for Agentic & Industrial-Grade Browser Automation.**

---

## **1\. Competitive Analysis & Gap Summary**

* **Playwright (2026): Dominates with the Speedboard Timeline (visualizing slow test bottlenecks) and AI Test Agents (autonomous healer loops).**
* **Cypress (2026): Leading in Quality Intelligence (AI identifying untested code gaps) and simplified Studio recording.**
* **Selenium (2026): Moving to WebDriver BiDi, allowing two-way communication (streaming console/network events natively).**
* **Sel-CLI Advantage: Uses Rust (Selenium Manager core) for raw speed and TypeScript (Oclif) for the best AI-agent integration (MCP support), while supporting legacy and cross-browser grids that competitors ignore.**

---

### **Module 1: Infrastructure & Performance (Rust Core)**

| ID | Feature | Description | Priority |
| :---- | :---- | :---- | :---- |
| **FR-1.1** | **Universal Binary** | **Single Rust binary that replaces the need for local Java or Python environments.** | **P0** |
| **FR-1.2** | **Install Browsers and Grid** | **Download browsers and selenium grid and get them ready to run.** | **P0** |
| **FR-1.3** | **Speedboard Logs** | **Native profiling of every Selenium command to generate a Timeline of "Time to Interactive" (Playwright parity).** | **P1** |
| **FR-1.4** | **BiDi Streamer** | **Direct pipe to browser console and network events using WebDriver BiDi for real-time debugging.** | **P1** |

### **Module 2: Agentic Automation (AI-First)**

| ID | Feature | Description | Priority |
| :---- | :---- | :---- | :---- |
| **FR-2.1** | **MCP Server Support** | **Integration with Model Context Protocol (MCP) to allow Cursor/Claude to control browsers natively.** | **P1** |
| **FR-2.2** | **Healer Loop** | **If a test fails due to a changed locator, the CLI autonomously attempts to find a new locator and suggests a PR.** | **P1** |
| **FR-2.3** | **YAML Snapshots** | **Converts DOM to ultra-compact YAML. Crucial for reducing LLM token costs by 70% (Firecrawl/Agent-native).** | **P1** |

### **Module 3: Advanced Observability**

| ID | Feature | Description | Priority |
[...]

PR #17542, opened 2026-05-22
[py][js] Initial Creation of a external shippable selenium manager

AutomatedTester:
this adds the ability to call npx or uvx to be able to call selenium manager and it handles the scenarios around picking the correct version for your computer

### 🔗 Related Issues
This is creating the shippable parts of #17326 

### 🤖 AI assistance

    - Tool(s): Claude
    - What was generated: 

### 🔄 Types of changes

### DIFF
--- a/py/selenium-manager-linux-x86-64/selenium_manager_linux_x86_64/__init__.py
+++ b/py/selenium-manager-linux-x86-64/selenium_manager_linux_x86_64/__init__.py
@@ -22,9 +22,7 @@ import sys
 
 
 def main() -> None:
-    with importlib.resources.as_file(
-        importlib.resources.files(__package__) / "bin" / "selenium-manager"
-    ) as binary:
+    with importlib.resources.as_file(importlib.resources.files(__package__) / "bin" / "selenium-manager") as binary:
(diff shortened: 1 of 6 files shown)


nvborisenko:
Would be amazing to ship it for dotnet too, `dnx` command. Some info: https://andrewlock.net/exploring-dotnet-10-preview-features-5-running-one-off-dotnet-tools-with-dnx/

cgoldberg:
Should we add the architecture name to the Windows packages? (like they are for Linux)
i.e.: `selenium-manager-windows-x86-64` instead of `selenium-manager-windows`

That will give us the ability to ship ARM64 Windows builds in the future and differentiate them easily.

Also, is that naming convention forced by the package repos, or can we unify them between JS and Python?

i.e. publish both packages as `selenium-manager-macos`, instead of `selenium-manager-darwin` and `selenium-manager-macos`?

titusfortner:
I have a number of concerns that I'll enumerate on Slack.

The first one - if we are releasing this independently we can (and should) make the plausible tracking opt-in instead of opt-out.

PR #17543, opened 2026-05-22
[grid] Apply TCP backpressure across the WebSocket tunnel handler

shs96c:
## Summary

The optional transparent TCP tunnel introduced in #17146 forwarded inbound bytes from one channel to the other with no flow control. A slow drain on one leg let the opposite leg keep shipping bytes, which Netty queued in the outbound buffer indefinitely. A session with sustained bidirectional traffic and a slow client could grow the Router's heap until the per-channel outbound queue was many megabytes.

Hook `channelWritabilityChanged` on each tunnel handler and mirror the state onto the peer's read side via `setAutoRead`. When this channel's outbound buffer crosses the high-water mark Netty fires the event and the peer stops reading; when the buffer drains below the low-water mark the peer resumes. The two legs of the tunnel each watch their own writability and toggle the other's read side, so backpressure propagates in both directions.

The thresholds come from the channel's existing `WriteBufferWaterMark` (Netty's default of 32 KiB low / 64 KiB high, or whatever an operator configures). No new numeric cap is introduced.

### DIFF
--- a/java/src/org/openqa/selenium/netty/server/TcpTunnelHandler.java
+++ b/java/src/org/openqa/selenium/netty/server/TcpTunnelHandler.java
@@ -26,6 +26,10 @@ import java.util.logging.Logger;
 /**
  * Forwards every inbound {@link io.netty.buffer.ByteBuf} to a target {@link Channel}. Used on both
  * ends of a transparent TCP tunnel once the WebSocket upgrade handshake has been proxied.
+ *
+ * <p>Backpressure is mirrored across the tunnel: when {@code ctx.channel()}'s outbound buffer
+ * passes its high-water mark, {@code target}'s read side is paused so the peer stops shipping bytes
+ * the kernel cannot drain. When the buffer drains below the low-water mark the peer is resumed.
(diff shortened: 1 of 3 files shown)


shs96c:
Addressed both qodo findings in 88b59df:

- **Writability event swallowed.** Added `ctx.fireChannelWritabilityChanged()` so the event continues down the pipeline. The handler is currently the last in its pipeline, but propagation is correct Netty hygiene and lets future handlers (metrics, throttling, debugging) observe the transition.
- **Cross-event-loop `setAutoRead`.** Verified this is not a real issue today: `TcpUpgradeTunnelHandler` puts both legs on the same event loop (`Bootstrap().group(clientChannel.eventLoop())`), and `DefaultChannelConfig.setAutoRead` is documented thread-safe in any case — the flag is mutated via an atomic updater and the `channel.read()` triggered on a false→true transition marshalls itself onto the channel's event loop. Added a comment at the call site explaining why direct invocation is safe, in case a future reader has the same concern.

PR #17544, opened 2026-05-22
[grid] Add BinaryMessage.wrap for transferring an owned byte array

shs96c:
## Summary

`BinaryMessage`'s two public constructors both copy their input defensively: the `byte[]` constructor calls `System.arraycopy`, and the `ByteBuffer` constructor copies into a freshly allocated array. That is the right default when the caller cannot promise ownership, but several call sites already produce a fresh array via `ByteArrayOutputStream.toByteArray()` and then immediately have it copied a second time by the constructor. For a multi-megabyte payload the second copy is wasted work and a wasted allocation.

Add a static `BinaryMessage.wrap(byte[])` factory that takes ownership of the array without copying. Its javadoc explicitly states that the caller must not mutate the array afterwards. The two public constructors keep their copy semantics, so existing callers that pass in a borrowed array are unaffected.

Use the new factory at two slow-path reassemblers that already produce a fresh array we own:

- `MessageInboundConverter`, when a fragmented binary message finishes via `buffer.toByteArray()` before being forwarded to `WebSocketMessageHandler`.
- `WebSocketFrameProxy`, when the inbound frame proxy reassembles a fragmented message for an upstream WebSocket that does not support per-frame sends.

The hot path through the Router (#17435) and Node proxies no longer touches `BinaryMessage` at all — frames travel end-to-end as Netty `WebSocketFrame` objects — so the saving here is on the fallback paths that handle fragmented binary messages.

### DIFF
--- a/java/src/org/openqa/selenium/netty/server/MessageInboundConverter.java
+++ b/java/src/org/openqa/selenium/netty/server/MessageInboundConverter.java
@@ -71,7 +71,8 @@ class MessageInboundConverter extends SimpleChannelInboundHandler<WebSocketFrame
           }
 
           if (finalFragment) {
-            message = new BinaryMessage(buffer.toByteArray());
+            // toByteArray() returns a fresh copy we own; transfer it without re-copying.
+            message = BinaryMessage.wrap(buffer.toByteArray());
             buffer.reset();
             next = Continuation.None;
           } else {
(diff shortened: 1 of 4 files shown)


shs96c:
Addressed in 2c15d04: the private constructor now calls `Require.nonNull("Ownership", ownership)` on the sentinel parameter so the argument is actually used and the contract is checked at runtime. Comment moved to the enum to explain that it exists purely to give the no-copy constructor a distinct signature from the defensive-copy public one.

Issue #13954, opened 2024-05-16
[🚀 Feature]: Support Electron


diemol:
### Feature and motivation

Right now, to run tests on electron apps, it pretty much requires a JavaScript dev environment. Electron versions are defined in the development directory, and npm is used to install the correct version of chromedriver via electron-chromedriver.npm

So the proposal is:
1. Create an `ElectronOptions` class that is a subclass of `ChromiumOptions`, where the browser name is `electron`.
2. Create an `ElectronDriver` class that requires an ElectronOptions argument, and errors if the binary has not been set
3. Create an `ElectronService` in case the user wants to specify the location of the chromedriver (maybe?)
4. Selenium Manager downloads the driver from https://github.com/electron/electron/releases based on the specified version, or gets the latest. I think we should not  try to automatically determine the version of the electron app; the app is bundled with a specific runtime version but it appears to be quite challenging to actually surface the version of it.

### Usage example

```
ElectronOptions options = new ElectronOptions();
options.browserVersion = "29.0";
options.binary = "/path/to/electron/app";

ElectronDriver driver = new ElectronDriver(options);
```

titusfortner:
I updated the description here. 
An example of how users are currently working with electron can be found in the [Sauce docs](https://docs.saucelabs.com/web-apps/automated-testing/electron/)

The bindings work should be easy, not certain about the selenium manager piece with the naming conventions

Delta456:
I can implement this in Python.

3255834892:
> ElectronOptions

How should I introduce ElectronOptions? I can't find it when I create new ElectronOptions. Do I need to introduce other dependencies to use it?

cgoldberg:
> How should I introduce ElectronOptions? 

This is a proposal. It doesn't exist yet.

titusfortner:
Do we actually need `ElectronDriver` and `ElectronService`?

Could we just create an `ElectronOptions that sets the right browser version and then error if `browserVersion` and `binary` are not specified?
```
ChromeDriver driver = new ChromeDriver(electronOptions)
```

PR #17559, opened 2026-05-23
[java] support driving Electron apps with ElectronOptions and ElectronDriver

titusfortner:
### 🔗 Related Issues
Implements #13954 for Java

### 💥 What does this PR do?
* Creates an ElectronOptions class that requires the binary in the constructor
* Creates an ElectronDriver class 

### 🔧 Implementation Notes
* My first implementation was just ElectronOptions inheriting from ChromeOptions which was the most simple because there was only one change needed to driver class (generateExecutor), but that ends up with a lot of chrome specific things included and no way to verify a binary is passed in
* Next implementation was extending ChromiumOptions but still using ChromeDriver, but that's 4 new constructors and a number of things that don't apply to Electron
* ~The DriverFinder conditional is hacky, but the altnerative would be an interface and extra methods that seemed excessive.~ Ok, I'm going to redo this with a custom capability, because yes, parsing class name is hacky.
* This validates that the location passed into the constructor is valid which prevents grid support. But should be easy to remove once we figure out how to get it working with grid

### 🤖 AI assistance

    - Tool(s):
    - What was generated:

This approach would need to be applied across all the bindings

### 🔄 Types of changes


java/src/org/openqa/selenium/chrome/ElectronDriver.java
```diff
@@ -0,0 +1,105 @@
...
+    Require.nonNull("Driver options", options);
+    Require.nonNull("Driver clientConfig", clientConfig);
+    DriverFinder finder = new DriverFinder(service, options);
+    service.setExecutable(finder.getDriverPath());
+    return new ChromiumDriverCommandExecutor(service, Map.of(), clientConfig);
+  }
+
+  @Override
+  public List<Map<String, String>> getCastSinks() {
+    throw new UnsupportedOperationException("Casting is not supported on Electron applications");
```
Delta456, line 72: `"Casting is not supported on Electron applications"` should be constant as it is used in several places.

titusfortner: You're right, that is a rule of DRY, but I personally prefer to avoid DRY when I think it reduces clarity. In this case it is better to have descriptions or messages inline than create a lookup. The message is unlikely to change and doesn't represent a concept or a unit of logic.

java/src/org/openqa/selenium/chrome/ElectronDriver.java
```diff
@@ -0,0 +1,105 @@
...
+
+  private static ChromiumDriverCommandExecutor generateExecutor(
+      ChromeDriverService service, ElectronOptions options, ClientConfig clientConfig) {
+    Require.nonNull("Driver service", service);
+    Require.nonNull("Driver options", options);
+    Require.nonNull("Driver clientConfig", clientConfig);
+    DriverFinder finder = new DriverFinder(service, options);
+    service.setExecutable(finder.getDriverPath());
+    return new ChromiumDriverCommandExecutor(service, Map.of(), clientConfig);
+  }
```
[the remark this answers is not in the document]
titusfortner: Selenium Manager doesn't discover the electron binary path, that is required to be in the options. The generateExecutor() method needs to know which driver to start, the driver needs to know which binary to drive and it will get that from the options

PR #17563, opened 2026-05-24
[dotnet] implement Central Package Management

titusfortner:
While working with Renovate I realized that we have dotnet versions specified in 6 different places, and Renovate can't help with all of them, and versions are different between them, especially transitive versions.

### 💥 What does this PR do?
1. Use CPM for webdriver csproj files
2. Dynamically generate nuspec versions for users

### 🤖 AI assistance

    - Tool(s):
    - What was generated:

I have a few follow-on PRs for dynamically generating paket.dependencies from CPM, but this seems like the easiest first step.

### 🔄 Types of changes


nvborisenko:
I suggest we hold on this. It adds maintenance overhead without fully covering CPM features, complicating our already difficult Selenium build process.

PR #17567, opened 2026-05-24
[rust] honor --browser-version in Selenium Manager Electron driver resolution

### LINKED ISSUE #17549: [🐛 Bug]: SM Electron does not honor --browser-version when fetching driver
### Description

Selenium Manager's Electron driver fetch ignores `--browser-version`. The URL is always built from the `/releases/latest` redirect, so the user-requested Electron version never selects the matching release's chromedriver asset.

The version is only used as a metadata cache key — not for selecting which release to download.

### Relevant Code

- [rust/src/electron.rs:124-130](https://github.com/SeleniumHQ/selenium/blob/trunk/rust/src/electron.rs#L124-L130) — `request_driver_version()` always reads `read_redirect_from_link(latest_url, ...)`, discarding any requested version.
- [rust/src/electron.rs:150-163](https://github.com/SeleniumHQ/selenium/blob/trunk/rust/src/electron.rs#L150-L163) — `get_driver_url()` then formats the URL using whatever `get_driver_version()` returned (always latest).

Expected: when `--browser-version 29.1.4` (or major like `29`) is supplied, fetch the matching release's chromedriver asset from `https://github.com/electron/electron/releases/download/v<version>/chromedriver-v<version>-<platform>.zip`.

### Reproducible Code

```
./selenium-manager --browser electron --browser-version 29.0.0 --debug
```

Observe that the driver downloaded is from the latest Electron release, not 29.0.0.

gaurav0107:
### Description

`ElectronManager::request_driver_version()` always read the GitHub redirect from `electron/electron/releases/latest` and used the resolved tag as the driver version. The user-supplied `--browser-version` was only consulted as a metadata cache key (via `get_major_browser_version()`); it never affected the actual URL resolution. As a result, `get_driver_url()` formatted the chromedriver download URL using whatever version came back from `/releases/latest` rather than the version the user requested.

For Electron, the chromedriver shipped inside each release is tagged identically to the Electron release itself (`https://github.com/electron/electron/releases/download/v{VERSION}/chromedriver-v{VERSION}-{platform}.zip`), so when the caller already supplies a concrete version, we can use it directly and avoid the network roundtrip altogether.

The fix:

- Inside the `_ =>` arm of `request_driver_version`, if `get_browser_version()` is a concrete version (non-empty, not `stable`, not an unstable channel like `dev`/`beta`/`nightly`), return it directly as the driver version.
- Otherwise, keep the existing `/releases/latest` redirect path unchanged.
- Gate the metadata-cache write on a non-empty `major_browser_version` (mirrors the existing guard in `chrome.rs:338`), so we no longer persist empty-keyed driver entries.

A regression test (`electron_browser_version_test`) was added in `rust/tests/electron_tests.rs` that pins `--browser-version 36.2.1` and asserts the manager succeeds. Locally:

```
cargo test --manifest-path rust/Cargo.toml --test electron_tests electron_browser_version_test
running 1 test
test electron_browser_version_test::case_1 ... ok
```

The pre-existing `electron_latest_test` (which exercises the unchanged latest-fallback path) also still passes.

### Motivation and Context

Closes #17549.

When Selenium Manager is asked to provision the Electron driver against a specific Electron version, the user expects the chromedriver from that release. Today, every request - regardless of `--browser-version` - silently downloads whatever is at `/releases/latest`, so cross-version testing of Electron apps via SM is effectively broken.

### Types of changes

### Checklist

- [x] My change requires a change to the documentation. - N/A; no user-facing CLI changes, only behaviour fix.
- [x] My change is covered by tests. - `electron_browser_version_test` added, exercises the new specific-version path; `electron_latest_test` continues to cover the unchanged latest-fallback path.
- [x] All new and existing tests passed locally. - `cargo build`, `cargo fmt --check`, `cargo test --lib`, and both `electron_*` integration tests pass.

### AI assistance disclosure

Per the CONTRIBUTING.md AI Tool Use section: AI assistance was used to draft the patch and the regression test. I read, reviewed, and ran the changes locally, and I am the responsible author. No `Co-Authored-By` trailer for AI tooling per project policy.

### DIFF
--- a/rust/src/electron.rs
+++ b/rust/src/electron.rs
@@ -25,7 +25,8 @@ use crate::metadata::{
     create_driver_metadata, get_driver_version_from_metadata, get_metadata, write_metadata,
 };
 use crate::{
-    LATEST_RELEASE, Logger, OFFLINE_REQUEST_ERR_MSG, SeleniumManager, WINDOWS, create_http_client,
+    BETA, CANARY, DEV, ESR, LATEST_RELEASE, Logger, NIGHTLY, OFFLINE_REQUEST_ERR_MSG, STABLE,
+    SeleniumManager, WINDOWS, create_http_client,
 };
 use anyhow::Error;
 use reqwest::Client;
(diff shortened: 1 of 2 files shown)


titusfortner:
@gaurav0107:
Thanks for working on this.
Qodo's three flags are all valid. 

Two notes:
* For the "unnormalized version" one: fix the vv36.2.1 by stripping only a leading v, not with parse_version — that would mangle prerelease tags like 36.0.0-beta.1.
* The "major as tag" one is actually a regression: today 36 falls back to /releases/latest, but this builds a 404 v36 URL. Only treat the pin as an exact tag when it's a full x.y.z; let 36 keep falling back. A test for that case would be good as well.

gaurav0107:
@titusfortner thanks for the review. All three Qodo flags are addressed in 1f74218:

- **Major-as-tag regression**: `request_driver_version` now only short-circuits when the pin is a *specific* version (contains a dot). `--browser-version 36` keeps falling back to `/releases/latest`. Added `electron_major_only_browser_version_falls_back_test` and unit-level coverage.
- **Unnormalized `vv` prefix**: only a leading `v`/`V` is stripped — no `parse_version()` — so prerelease tags like `36.0.0-beta.1` are preserved. Added `v36.2.1` to the parameterized integration test plus unit tests.
- **Cache-overrides-pin**: the pinned-version short-circuit now runs before the metadata-cache lookup, so a cached entry keyed on major `36` cannot override an explicit `36.2.1` request.

`cargo test --lib electron` (6 new unit tests) and `cargo test --test electron_tests` (5 tests, including the new cases) pass locally.

Note on the failing CI checks: `config_test::case_2` (firefox) is a pre-existing CI-runner flake — geckodriver 0.36.0 in PATH vs firefox 151.0.1 emits a WARN that the test asserts against; not related to this diff. Validate-workflows is failing on `.github/workflows/bazel.yml` (`property "browser" is not defined`), also pre-existing on trunk.

diemol:
@bonigarcia could you please have a look and see if this makes sense?

gaurav0107:
Thanks again for the review and approval, @bonigarcia! Gentle nudge that this is still green and ready to merge whenever it's convenient for the team. Happy to rebase if it's drifted behind `trunk` — just let me know.

Issue #17487, opened 2026-05-16
[🚀 Feature]: generate bidi code for JS from CDDL


AutomatedTester:
### Description

Take the CDDL code and generate code so we can include new specs quickly

PR #17574, opened 2026-05-26
[js] Add Javascript/Typescript CDDL code generator for WebDriver BiDi

pujagani:
### 🔗 Related Issues
Related to #17487 
Part of the CDDL WebDriver BiDi API generator. Implemented as defined in the spec https://gist.github.com/AutomatedTester/148437f846a3759f7678c7ff6debb210

### 💥 What does this PR do?

Adds a code generator (`generate_bidi.mjs`) that reads the merged WebDriver BiDi CDDL
specification and emits one typed TypeScript module per domain into `bidi/generated/`.
The generator is wired into the Bazel build so the output is compiled and included in
the npm package automatically.

Each generated domain class (`Browser`, `BrowsingContext`, `Input`, `Log`, `Network`,
`Script`, `Storage`, etc.) exposes typed command methods and event subscription helpers
via a `static async create(driver)` factory. An enhancements manifest
(`private/bidi_enhancements_manifest.json`) allows hand-written additions (deprecated
compatibility shims, convenience wrappers, and method overloads) to be injected at
build time without modifying generated files.

### 🔧 Implementation Notes

**Why a two-pass generator?**
`cddl2ts` only produces type declarations; it has no concept of the BiDi command/event
distinction. A second pass walks the raw CDDL AST to identify commands (groups with a
`method` literal and `params`) and events (same structure plus `type: "event"`) and
emits the implementation class from that.

**Why throw on error responses?**
`bidi/index.js` always resolves, it never rejects on BiDi error responses. Both void
and non-void generated methods inspect the response payload and throw if
`response.type === 'error'`, so unhandled server errors surface as test failures rather
than being silently swallowed.

**Why an enhancements manifest?**
The existing hand-crafted modules use a builder pattern
(`getBrowserInstance(driver)`, `AddInterceptParameters`). The generated API uses typed
params objects. Replacing the old exports directly would be a breaking change. The
manifest lets deprecated compatibility wrappers live alongside generated code until a
major version bump removes them.

### 🤖 AI assistance
    - Tool(s): Claude (Anthropic)
    - What was generated: generator implementation, Bazel macro, enhancements manifest,
      integration tests under `test/bidi/generated/`

**Breaking change:** The generated API shape (`Domain.create(driver)`, typed params
objects) differs from the existing hand-crafted modules (positional args, builder
classes). Deprecated compatibility shims are in place for the most-used modules but a
full audit is needed before the old modules can be removed. See the Phase 7 audit in
the PR description or tracking issue for the complete list.

**Follow-on work required:**
- Add missing convenience wrappers to the enhancements manifest:
  `captureBoxScreenshot`, `captureElementScreenshot`, `locateNode`, `locateElement`,
  `locateElements` (BrowsingContext); filtered log level variants (Log). 
- Decide if deprecation is required for `browsingContextInspector.js` and `networkInspector.js` — superseded by
  event methods on the generated `BrowsingContext` and `Network` classes
- Decide fate of the builder/DTO classes (`AddInterceptParameters`,
  `ContinueRequestParameters`, `CookieFilter`, etc.).

### 🔄 Types of changes
- New feature (non-breaking change which adds functionality and tests)

### DIFF
--- a/javascript/selenium-webdriver/generate_bidi.mjs
+++ b/javascript/selenium-webdriver/generate_bidi.mjs
@@ -257,7 +257,9 @@ function loadEnhancements(manifestPath) {
     throw new Error(`Failed to parse enhancements manifest at ${fullPath}: ${err.message}`)
   }
   if (typeof parsed !== 'object' || parsed === null || Array.isArray(parsed)) {
-    throw new Error(`Enhancements manifest at ${fullPath} must be a JSON object, got ${Array.isArray(parsed) ? 'array' : typeof parsed}`)
+    throw new Error(
+      `Enhancements manifest at ${fullPath} must be a JSON object, got ${Array.isArray(parsed) ? 'array' : typeof parsed}`,
+    )
   }
   return parsed
 }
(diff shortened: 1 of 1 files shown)


AutomatedTester:
@pujagani This looks good, can we make sure the AI reviews are correct and resolve accordingly?

pujagani:
Thank you for reviewing @AutomatedTester. Will do.

PR #17585, opened 2026-05-28
[SM] Automatically prune cache entries older than 30 days

### LINKED ISSUE #17196: [🚀 Feature]: Automatically prune Selenium Manager cache
### Description

The Selenium cache (`~/.cache/selenium`) grows without bounds: I just have found and deleted 19 GB of old `chrome` and `chromedriver` versions :)

I would like to impose a limit here, and automatically prune this directory by deleting everything but the last N versions.

This limit could be opt-in, and should be evaluated/imposed whenever a new version is downloaded.

- `selenium-manager --clear-cache` (and `--clear-metadata`) exists, but that's a nuclear option similar to `rm -rf ~/.cache/selenium`. I do usually need the latest chrome/chromedriver version, so removing the cache directory on a regular basis means (unnecessarily re-) downloading those programs.
- manually running a `find | sort | tail -n+10 | xargs rm -rf` command is tedious, error prone (at least due to version number sorting) and dangerous (I'm never comfortable with piping into `rm -rf`).

I'm primarily using Selenium Manager through the `selenium-webdriver` Rubygem. Accessing the `selenium-manager` binary is a bit complicated, because the binary lives somewhere in `$GEM_HOME/gems/selenium-webriver-$version/bin/linux/`

AutomatedTester:
Selenium Manager now removes driver/browser version directories from the cache (~/.cache/selenium) that have not been modified in over 30 days. The prune runs on every invocation after clear-cache/clear-metadata, so the cache size is bounded without requiring manual intervention.

Fixes #17196

### 🤖 AI assistance

    - Tool(s):
    - What was generated:

### 🔄 Types of changes

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -1778,6 +1778,22 @@ fn delete_cached_asset(log: &Logger, cache: &Path, asset_name: &str, asset_versi
                 err
             ))
         });
+        // Remove any ancestor directories that are now empty, up to (not including) the cache root
+        let mut parent = path.parent();
+        while let Some(dir) = parent {
+            if dir == cache {
+                break;
+            }
+            if fs::read_dir(dir)
+                .map(|mut d| d.next().is_none())
+                .unwrap_or(false)
+            {
+                let _ = fs::remove_dir(dir);
+            } else {
+                break;
+            }
+            parent = dir.parent();
+        }
     }
 }
 
@@ -1846,6 +1862,10 @@ mod tests {
         assert!(
             !version_dir.exists(),
             "Version directory should be removed when last_used is older than ttl_days"
+        );
+        assert!(
+            !cache.join("chromedriver").exists(),
+            "Empty parent directories should be cleaned up"
         );
         let updated = get_metadata(&log, &Some(cache.to_path_buf()));
         assert!(


AutomatedTester:
I forgot about how infrequent geckodriver releases are... I will update accordingly

rust/src/metadata.rs
```diff
@@ -52,11 +52,24 @@ pub struct Driver {
...
+    pub last_used: u64,
+}
+
 #[derive(Serialize, Deserialize)]
 pub struct Metadata {
     pub browsers: Vec<Browser>,
     pub drivers: Vec<Driver>,
     pub stats: Vec<Stats>,
+    #[serde(default)]
+    pub cached_assets: Vec<CachedAsset>,
```
bonigarcia, line 72: I am not convinced of this part. What is `asset_name` in `CachedAsset`?
My idea was to add the new field (`last_used`) for browser and driver metadata (i.e., in `Browser` and `Driver` structs).

AutomatedTester: This gets around the issue where `get_metadata` can filter out items during the `retain` which can lead to the following scenario

1. chromedriver v130 is used today → last_used = now
2. One hour passes → its driver_ttl expires
3. User asks for geckodriver (unrelated)
4. `get_metadata()` is called → chromedriver v130 entry is filtered out by `retain`
5. `write_metadata()` writes metadata with geckodriver updated → chromedriver v130 is permanently gone
6. The 30-day pruner now has no record that chromedriver v130 was used today

I have tested this out manually and the code currently works well

bonigarcia: Yes, that also works. In my head, I imagined removing the filtering (`retain`) on reading and using the TTL_driver condition later to check staleness. That will avoid having an additional entry in the metadata for the cache assets. But it also works that way. Thanks!

diemol:
Are the docs up to date with this feature?

bonigarcia:
I created a PR in the Selenium doc about it: https://github.com/SeleniumHQ/seleniumhq.github.io/pull/2673

PR #17586, opened 2026-05-28
[rust] Allow cross-compilation of selenium-manager on all platforms

AutomatedTester:
## Summary

Taking over #17427 from @shs96c who was unable to complete it.

This PR enables cross-compilation of `selenium-manager` from any host platform to all target platforms (Linux x86_64/aarch64, macOS aarch64, Windows x86_64/aarch64) using the hermetic LLVM toolchain.

### Key changes (from Simon's original work)
- Upgrades `rules_rs` 0.0.61 → 0.0.76 and `llvm` 0.7.9 → 0.8.0
- Registers LLVM exec toolchains for Linux (x86_64, aarch64) and macOS (aarch64)
- Adds cross-compile targets for all five platforms in `rust/BUILD.bazel`
- Restructures language binding (py, rb, dotnet, java) BUILD files to consume a per-arch bundle
- On Windows, uses native MSVC build for the host arch and stubs the rest (cross-compile from Linux for releases)
- Adds musl targets for Linux (no glibc dependency) and gnullvm targets for Windows

### Fixes applied in this PR (on top of Simon's work)
- **`windows-link`/`windows-targets` patches**: The original patches changed `include_str!("../readme.md")` → `include_str!("../README.md")` but the crate only ships `readme.md` and Bazel's Rust sandbox doesn't include non-source files anyway. Fixed by removing the doc attribute entirely.
- **Windows MSVC linker**: The hermetic LLVM package for Windows provides only a GNU-mode `clang++.exe` which can't handle MSVC-style linker flags (`/NOLOGO`, `.lib` files). Re-enabled MSVC CC auto-detection on Windows so Visual Studio's `link.exe` is used for MSVC targets.

### Known remaining issues
- **macOS Rust tests**: `aws-lc-sys` (pulled in transitively through `reqwest` → `rustls` → `rustls-platform-verifier`) requires `CoreServices.framework` which is not in the hermetic LLVM macOS SDK sysroot. Adding the aarch64 macOS exec LLVM toolchain (new in this PR) causes this to surface. Fixing requires either a different TLS cert backend or a richer macOS sysroot — tracked as a follow-up.

Closes #17427


titusfortner:
It's suboptimal, but I'm ok with Windows or Mac not being able to cross compile everything so long as Linux can do it all.

The primary missing piece as I see it is the lack of x86 support for Mac. Ideally we can continue to release a fat binary instead of requiring the bindings to pick between them, but I guess the bindings fix is easy enough if we need to go that route.

AutomatedTester:
https://bazelbuild.slack.com/archives/CSV56UT0F/p1781613082861409

Raised the above discussion as I think there is a bug in the way transitive deps are handled

AutomatedTester:
> https://bazelbuild.slack.com/archives/CSV56UT0F/p1781613082861409
> 
> Raised the above discussion as I think there is a bug in the way transitive deps are handled

I need to come back to this next week and watch updates to `rules_rs`

MugundanMCW:
Hi @AutomatedTester 
Any updates on this PR?

PR #16733, opened 2025-12-15
report language version to plausible

titusfortner:
### **User description**

*Note*: this PR includes the code from #16736 to get the tests to pass

### 🔗 Related Issues

Implements #16731 

### 💥 What does this PR do?

Each language obtains version to send to selenium manager
Selenium Manager reports to Plausible

### 🔧 Implementation Notes

The only nonstandard one is that .NET reports both the target framework and the Runtime Version:

  | Scenario            | lang_version     |
  |---------------------|------------------|
  | .NET 8 user         | net8.0/8         |
  | .NET 10 user         | net8.0/10         |
  | .NET 6 user         | netstandard2.0/6 |
  | .NET Framework user | net462           |

### 🔄 Types of changes

- New feature (non-breaking change which adds functionality *and tests!*)

___

### **PR Type**
Enhancement

___

### **Description**
- Add language version reporting to Selenium Manager across all bindings

- Java reports runtime feature version via Runtime.version().feature()

- .NET reports target framework and runtime version with conditional compilation

- JavaScript reports Node.js version via process.versions.node

- Python reports major.minor version via sys.version_info

- Ruby reports version via RUBY_VERSION constant

- Rust infrastructure updated to parse and forward language-version argument

___

### Diagram Walkthrough

```mermaid
flowchart LR
  Java["Java<br/>Runtime.version()"] --> SM["Selenium Manager<br/>--language-version"]
  DotNet[".NET<br/>Target Framework"] --> SM
  JS["JavaScript<br/>Node.js version"] --> SM
  Python["Python<br/>sys.version_info"] --> SM
  Ruby["Ruby<br/>RUBY_VERSION"] --> SM
  SM --> Rust["Rust Config<br/>language_binding_version"]
  Rust --> Stats["Stats/Metadata<br/>lang_version field"]
  Stats --> Plausible["Plausible Analytics"]
```

<details><summary><h3>File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>10 files</summary><table>
<tr>
[...]

PR #17591, opened 2026-05-29
[py][SM] Collect language version data in Selenium Manager

### LINKED ISSUE #16731: [🚀 Feature]: Collect language version data in Selenium Manager
### Description

It would be helpful to know how many people using the latest version of Selenium are using which languages to better inform when we can deprecate previous languages.

_No response_

AutomatedTester:
## Summary

- Adds `--language-version` CLI flag to Selenium Manager (Rust) to accept the runtime version of the invoking language binding
- Wires the value through `ManagerConfig`, the `SeleniumManager` trait, and into the `Props` struct sent to Plausible.io telemetry
- Updates the metadata dedup check so `lang_version` is part of the cache key (avoiding duplicate submissions)

Closes #16731

## Deferred: Python binding arg

The Python `--language-version` arg is **not** included in this PR. The pre-compiled pinned binary in `common/selenium_manager.bzl` (`selenium-manager-f6318a4`) predates this change and rejects the flag with `error: unexpected argument '--language-version' found`. Passing it would break all RBE tests.

**Follow-up steps:**
1. Merge this PR (Rust plumbing only)
2. Run `bazel run scripts:selenium_manager` to rebuild and re-pin the binary in `common/selenium_manager.bzl`
3. Add `--language-version {major}.{minor}` to `py/selenium/webdriver/common/selenium_manager.py` (and other bindings)

## Test plan

- [x] `bazel build //rust:selenium-manager` — passes
- [x] `bazel test //rust:unit` — 14/14 pass
- [x] `bazel test //py:unit` — 27/27 pass
- [ ] After re-pinning binary: verify Plausible.io dashboard shows `lang_version` property after a real browser session
- [ ] Other bindings (Java, Ruby, JS, .NET) can add `--language-version` in follow-up PRs

## Notes

The flag is optional in Rust — if not supplied, `lang_version` defaults to an empty string, preserving backwards compatibility.

### DIFF
--- a/py/selenium/webdriver/common/selenium_manager.py
+++ b/py/selenium/webdriver/common/selenium_manager.py
@@ -49,8 +49,6 @@ class SeleniumManager:
             args.append("--debug")
         args.append("--language-binding")
         args.append("python")
-        args.append("--language-version")
-        args.append(f"{sys.version_info.major}.{sys.version_info.minor}")
         args.append("--output")
         args.append("json")
 
--- a/py/test/unit/selenium/webdriver/common/selenium_manager_tests.py
+++ b/py/test/unit/selenium/webdriver/common/selenium_manager_tests.py
@@ -38,15 +38,7 @@ def test_gets_results(monkeypatch):
     ):
         SeleniumManager().binary_paths([])
         mock_get_binary.assert_called_once()
-        expected_run_args = [
-            "/path/to/sm",
-            "--language-binding",
-            "python",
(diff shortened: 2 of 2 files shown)


titusfortner:
This PR will not close 16731, that is supposed to track all bindings, not just rust.

Also, this is a duplicate of #16733

nvborisenko:
50 cents: we will start to collect some data (in .NET it is even not related to language version). Then what?

PR #17606, opened 2026-06-01
[py] Extract feature-specific modules from //py:common 

AutomatedTester:
## Summary

- Five modules were bundled in `//py:common` but are only imported by a small subset of tests. Changing any of them triggered a full rebuild of all ~700 browser test targets.
- This PR extracts them into dedicated libraries and wires per-feature test sub-suites so that the dep graph is precise.

## New libraries

| Target | Source |
|---|---|
| `//py:common_fedcm` | `fedcm/**/*.py` |
| `//py:common_timeouts` | `timeouts.py` |
| `//py:common_virtual_authenticator` | `virtual_authenticator.py` |
| `//py:common_api_request_context` | `api_request_context.py` |
| `//py:common_print_page_options` | `print_page_options.py` |

## Why lazy imports in `remote/webdriver.py`

`remote/webdriver.py` previously imported all five modules at the top level, which would have re-added them to every test's dep graph via `//py:remote → //py:support → all tests`. Three changes remove this:

1. `from __future__ import annotations` makes all type annotations lazy strings — no runtime import needed for parameter/return types like `def print_page(self, print_options: PrintOptions | None = None)`.
2. The four methods that actually instantiate these types (`timeouts` property getter, `request` property, `dialog` property, `get_credentials`) now lazy-import inside the method body.
3. `required_virtual_authenticator` and `required_chromium_based_browser` are inlined as private functions in `remote/webdriver.py` (they were only used there; 10 lines removes the top-level dep).

## Test suite structure

Each browser now has per-feature sub-suites (e.g. `test-chrome-timeouts`, `test-chrome-fedcm`) alongside the existing `-common` and `-actions` sub-suites. The top-level aggregate targets (`test-chrome`, `test-chrome-bidi`, `test-chrome-remote`) are unchanged — CI jobs continue to run the same targets.

## Impact

Changing `timeouts.py` now rebuilds ~11 targets (`timeout_tests-*` variants) instead of ~700.

## Test plan

- [ ] `bazel query 'rdeps(//py:*, //py:common_timeouts)' | grep py_test | grep -v timeouts` returns empty
- [ ] `bazel query 'kind(py_test, //py:*)' | grep timeout_tests` shows `timeout_tests-chrome-timeouts` etc.
- [ ] Existing unit and browser test suites pass

### DIFF
--- a/py/selenium/webdriver/remote/webdriver.py
+++ b/py/selenium/webdriver/remote/webdriver.py
@@ -1668,6 +1668,7 @@ class WebDriver(BaseWebDriver):
             WebDriverException: If FedCM not supported.
         """
         from selenium.common.exceptions import NoAlertPresentException
+        from selenium.webdriver.common.fedcm.dialog import Dialog
         from selenium.webdriver.support.wait import WebDriverWait
 
         self._require_fedcm_support()


AutomatedTester:
The build graph moves from

Screenshot 2026-06-01 at 14 42 53

to this

Screenshot 2026-06-01 at 14 43 00

PR #17622, opened 2026-06-03
[bazel] Use MRI on RBE

p0deje:
Updates the Ruby ruleset for Bazel, which now supports MRI (CRuby) toolchains on RBE. We no longer need to use JRuby for the RBE job and can stick to the default Ruby interpreter.

### DIFF
--- a/common/remote-build/BUILD.bazel
+++ b/common/remote-build/BUILD.bazel
@@ -41,7 +41,7 @@ platform(
         "@llvm//constraints/libc:gnu.2.28",
     ],
     exec_properties = {
-        "container-image": "docker://docker.io/p0deje/selenium-remote-build@sha256:7f917a409785aceff81e72358222c6d1bdc24d0b7df2d02f0f46e7392d5c8871",
+        "container-image": "docker://docker.io/selenium/selenium-remote-build@sha256:7f917a409785aceff81e72358222c6d1bdc24d0b7df2d02f0f46e7392d5c8871",
         "OSFamily": "Linux",
         "dockerNetwork": "standard",  # Allow internet access for the Rust tests
         "dockerReuse": "True",


common/remote-build/cc/BUILD
```diff
@@ -86,31 +121,38 @@ cc_toolchain_config(
...
         "/usr/include",
-        "/usr/include/c++/9",
-        "/usr/include/x86_64-linux-gnu/c++/9",
-        "/usr/include/c++/9/backward",
+        "/usr/include/c++/11",
+        "/usr/include/x86_64-linux-gnu/c++/11",
+        "/usr/include/c++/11/backward",
     ],
-    cxx_flags = ["-std=c++0x"],
+    cxx_flags = ["-std=c++17"],
```
shs96c, line 137: If we're setting this here, I think we can remove the equivalent entries in the `.bazelrc`. Can be done in a follow up PR.

common/remote-build/BUILD.bazel
```diff
@@ -41,7 +41,7 @@ platform(
         "@llvm//constraints/libc:gnu.2.28",
     ],
     exec_properties = {
-        "container-image": "docker://docker.io/selenium/selenium-remote-build@sha256:ca164352826812f02eaa1b4dab89adcf5b3e57deb6798dd9f15d004099db59fe",
+        "container-image": "docker://docker.io/p0deje/selenium-remote-build@sha256:7f917a409785aceff81e72358222c6d1bdc24d0b7df2d02f0f46e7392d5c8871",
```
shs96c: It should be the `selenium` org hosting this image

p0deje: Yes, this is just for testing.

common/remote-build/cc/BUILD
p0deje: Hmm, isn't this file only enabled on RBE, while .bazelrc is enabled for everything?

Issue #16741, opened 2025-12-16
[🚀 Feature]: Testing Selenium Manager


titusfortner:
### Description

We made a change here to always use pin_browsers in CI: https://github.com/SeleniumHQ/selenium/pull/15987

~This removed all testing of Selenium Manager.~ This removed testing our suites with Selenium Manager.

Ideally each binding has Service class tests that exercise Selenium Manager (some of them seem to do this already), and these tests will only work on GitHub Actions and not on RBE.

### Have you considered any alternatives or workarounds?

I think we may want to make pin_browsers the default for Bazel, and explicitly set it to false for Selenium Manager tests:
implemented the first part here: #16743

PR #17645, opened 2026-06-05
[py] add Selenium Manager integration tests

titusfortner:
### 🔗 Related Issues

Implements #16741 for Python
Part of #17539

### 💥 What does this PR do?

* Adds integration test coverage of Selenium Manager via Driver Finder class

### 🔧 Implementation Notes

* Because it goes through Driver Finder, the tests pass with and without pin_browsers being set by taking different paths (or skipping the cache location validation)
- Selected by tag (`--test_tag_filters=se-manager`) rather than per-browser target names, so all applicable browsers run in a single invocation per OS (3 jobs, not browser×OS).

### 🤖 AI assistance
    - Tool(s): Claude Code
    - What was generated: the test, BUILD targets, and CI job

- Will need to be deconflicted with #17643 depending on which merges first

### 🔄 Types of changes

### DIFF
--- a/py/test/selenium/webdriver/common/driver_finder_tests.py
+++ b/py/test/selenium/webdriver/common/driver_finder_tests.py
@@ -48,7 +48,7 @@ def driver_finder(browser_name, request):
         options.binary_location = _resolve_bazel_path(request.config.option.binary).strip("'")
 
     executable = request.config.option.executable
-    service = module.service.Service(executable_path=_resolve_bazel_path(executable) if executable else None)
+    service = module.service.Service(executable_path=_resolve_bazel_path(executable).strip("'") if executable else None)
 
     return DriverFinder(service, options)
 


py/test/selenium/webdriver/common/driver_finder_tests.py
```diff
@@ -0,0 +1,103 @@
...
+    monkeypatch.setenv("SE_SKIP_DRIVER_IN_PATH", "true")
+    # Match by basename so 8.3 short names on Windows don't fail the comparison.
+    assert tmp_path.name in driver_finder.get_driver_path()
+
+
+def test_downloads_the_browser_into_the_selenium_cache(driver_finder, browser_name, pinned, tmp_path, monkeypatch):
+    if pinned:
+        pytest.skip("Pinned runs do not use Selenium Manager")
+    if browser_name == "safari":
+        pytest.skip("Safari ships with the OS")
```
AutomatedTester, line 85: Why is the skips written differently between this test and the next test?

titusfortner: just to distinguish the error messages

titusfortner: Oh, and Edge on Windows downloads the driver to the cache, but not the browser, so the behavior is different for the 2 tests.

PR #17655, opened 2026-06-06
[rb] fix silent hang on oversized WebSocket frames

aguspe:
💥 What does this PR do?

 The Ruby devtools/bidi websocket listener can hang when it gets a frame bigger than WebSocket.max_frame_size (websocket-ruby defaults to 20MB).
 
 Websocket-ruby raises TooLong internally but rescues it unless you set should_raise, so incoming_frame.next just returns nil, the oversized bytes stay buffered, and the loop spins forever at ~100% cpu.

 builds on #17286, thanks @Chandan25sharma. left the track_cancelled part out, that's separate.

  🔧 Implementation Notes

  the bump happens in WebSocketConnection#initialize, not at load time, and only when the current limit is lower, so non-devtools users or anyone who already set a bigger value aren't affected. websocket-ruby only
  exposes max_frame_size globally so there's no per-instance option. didn't touch should_raise, detection uses the per-instance incoming_frame.error? instead.

  🤖 AI assistance

    - Tool(s): Claude Code
    - What was generated: Partially the description description and RBS signatures

  🔄 Types of changes

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -233,6 +233,7 @@ jobs:
       - name: Setup Safari Technology Preview
         if: inputs.safari-preview && inputs.os == 'macos'
         run: |
+          brew update --quiet
           brew install --cask safari-technology-preview
           sudo "/Applications/Safari Technology Preview.app/Contents/MacOS/safaridriver" --enable
       - name: Import GPG key
--- a/.github/workflows/mirror-selenium-releases.yml
+++ b/.github/workflows/mirror-selenium-releases.yml
@@ -14,6 +14,7 @@ jobs:
     steps:
     - uses: actions/checkout@v6
       with:
+        ref: trunk
         persist-credentials: false
     - name: Read api.github.com and filter response
       run: |
@@ -54,4 +55,4 @@ jobs:
       uses: ad-m/github-push-action@master
       with:
         github_token: ${{ secrets.SELENIUM_CI_TOKEN }}
-        branch: ${{ github.ref }}
+        branch: trunk
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -68,12 +68,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language:
-          - java
-          # - python
-          - ruby
-          - dotnet
-          - javascript
+        language: [java, python, ruby, dotnet, javascript]
     uses: ./.github/workflows/bazel.yml
     with:
       name: Nightly ${{ matrix.language }} Release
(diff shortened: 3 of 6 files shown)


Chandan25sharma:
Two complementary fixes:

apply_frame_size_limit — raises WebSocket.max_frame_size to 100 MB inside WebSocketConnection#initialize, not at load time, and only when the current limit is lower. Non-devtools users or anyone who already configured a larger value are unaffected. websocket-ruby exposes max_frame_size as a global-only setting; there is no per-instance option.

frame_dropped? — checks incoming_frame.error? (per-instance flag, not should_raise) after each read. When set, logs an actionable error and closes the connection cleanly instead of leaving a dead listener spinning on a stale buffer.

should_raise was left untouched — using incoming_frame.error? is the per-instance equivalent and avoids affecting unrelated connections.


Tested the fix locally. The two-part approach is correct:

apply_frame_size_limit in initialize prevents the hang for the common case (DevTools heap snapshots, large screenshots, etc.) without affecting users who already set a higher limit
frame_dropped? catches the remaining edge case cleanly — logs an actionable error and closes the connection instead of leaving the listener spinning at 100% CPU
The choice to use incoming_frame.error? over should_raise is the right call — should_raise is global and would affect all connections, error? is per-instance.

Unit tests cover both paths: the limit bump (raises when lower, leaves alone when higher) and the frame_dropped? shutdown. Builds cleanly on #17286.

aguspe:
> Two complementary fixes:
> 
> apply_frame_size_limit — raises WebSocket.max_frame_size to 100 MB inside WebSocketConnection#initialize, not at load time, and only when the current limit is lower. Non-devtools users or anyone who already configured a larger value are unaffected. websocket-ruby exposes max_frame_size as a global-only setting; there is no per-instance option.
> 
> frame_dropped? — checks incoming_frame.error? (per-instance flag, not should_raise) after each read. When set, logs an actionable error and closes the connection cleanly instead of leaving a dead listener spinning on a stale buffer.
> 
> should_raise was left untouched — using incoming_frame.error? is the per-instance equivalent and avoids affecting unrelated connections.
> 
> Tested the fix locally. The two-part approach is correct:
> 
> apply_frame_size_limit in initialize prevents the hang for the common case (DevTools heap snapshots, large screenshots, etc.) without affecting users who already set a higher limit frame_dropped? catches the remaining edge case cleanly — logs an actionable error and closes the connection instead of leaving the listener spinning at 100% CPU The choice to use incoming_frame.error? over should_raise is the right call — should_raise is global and would affect all connections, error? is per-instance.
> 
> Unit tests cover both paths: the limit bump (raises when lower, leaves alone when higher) and the frame_dropped? shutdown. Builds cleanly on #17286.

Thank you so much for the comment!, so far the global bump is the best we can do because the gem only gives us max_frame_size

Issue #17654, opened 2026-06-06
[🚀 Feature]: Expose AST and BiDi Model from JS for cross-binding usage


titusfortner:
## Description

Expose the parsed WebDriver BiDi specification as shared build artifacts so every binding's code generator can use them without needing to reparse CDDL and re-derive the BiDi command/event structure.

PR #17574 shares `//py/private:merge_cddl`, but the data valuable for cross-binding usage is parsed in memory inside `create-bidi-src_ts` and then discarded.

Each binding's generator has two distinct requirements:

### 1. Language-specific types — requires the AST

Every binding generates its own param/result/object/enum types. This cannot be shared: the CDDL→type mapping (`int → number` vs `long`) and the structural decomposition differ per language. What can be shared is the parsed input, so no binding has to write its own CDDL parser.

AST artifact (`create-bidi-src_ast`) — generic `cddl.parse → JSON`, single file, protocol-agnostic. Each binding walks it with its own type emitter (JS uses `cddl2ts`).

### 2. Generic command/event structure

Every binding needs the same BiDi command/event surface: domains → commands (`method`, params type, result type) and events (`method`, params type). This is shareable — it is pure protocol data, identical across languages, and it is the hard-to-derive piece (traversing the `CommandData`/`EventData` unions, including extension specs). Derive it once.

Model artifact (`create-bidi-src_json`) — BiDi extraction → JSON, single file, with type references by neutral CDDL name (e.g. `browser.CreateUserContextParameters`) so each binding applies its own naming.

### Stitching stays per-binding

Each generator needs to join its own types to the shared command/event classes, by neutral type name. The layout is language-specific, so this is not shared — it lives inside each binding's emit target.

## Flow

Now — `create-bidi-src_ts` does everything internally and discards the intermediate data:

```
create-bidi-src_merged_cddl
  └─► create-bidi-src_ts   (generate_bidi.mjs: cddl.parse → ast,
                            then cddl2ts(ast)→types + extract(ast)→model + stitch → .ts;
                            ast and model discarded)
        └─► create-bidi-src   (tsc: .ts → .js + .d.ts)
```

Desired — split the parse and the extraction into their own targets so the AST and Model become artifacts every binding consumes:

```
create-bidi-src_merged_cddl
  └─► create-bidi-src_ast          (cddl.parse → ast.json)
        └─► create-bidi-src_json       (extract(ast.json) → model.json — no re-parse)
[...]

PR #17657, opened 2026-06-08
[js] Expose BiDi CDDL ast and model as shared artifacts

titusfortner:
### 🔗 Related Issues

Addresses #17654. In draft until #17574 is merged, since it is based on that code.

### 💥 What does this PR do?

Parses the CDDL to expose two reusable artifacts, instead of being built in memory inside `create-bidi-src_ts` and discarded. Other bindings (Java/Ruby/Python) can then consume one shared description of the BiDi command/event surface rather than each re-parsing the spec. The CDDL is parsed exactly once, and the shipped JS package is unchanged — the generated TypeScript is byte-identical to #17574.

| Target | Status    | Consumes | Produces |
| --- |-----------| --- | --- |
| `create-bidi-src_merged_cddl` | unchanged | CDDL spec files | `create-bidi-src_merged.cddl` |
| `create-bidi-src_ts` (old) | *before*  | merged CDDL | one `.ts` module per BiDi domain |
| `create-bidi-src_ast` | **new**   | merged CDDL | `create-bidi-src_ast.json` — full parsed AST |
| `create-bidi-src_json` | **new**   | `create-bidi-src_ast.json` | `create-bidi-src_model.json` — binding-neutral command/event model |
| `create-bidi-src_ts` (new) | *after*   | `create-bidi-src_ast.json` + `create-bidi-src_model.json` | one `.ts` module per BiDi domain |
| `create-bidi-src` | unchanged | the `.ts` modules | `.js` + `.d.ts` (tsc) |

### 🔧 Implementation Notes

- **Binding-neutral by design:** the model references types by their CDDL name (e.g. `browser.CreateUserContextParameters`), not cddl2ts's PascalCase, so each binding applies its own naming. Shape per domain: `commands: [{method, name, params, result}]`, `events: [{method, name, params}]`.
- **Honest null fields:** `params`/`result` are `null` when a command takes no params / returns no value. Void is decided from the AST at model-build time (absent result type, or one aliased to `EmptyResult`) rather than in the JS emitter, so a consumer can trust the field without re-deriving it. Verified this matches the existing JS void detection across all 77 commands.
- `create-bidi-src_ast` and `create-bidi-src_json` are visible to `//java`, `//py`, `//rb` so each binding depends on only the artifact it needs.

### 🤖 AI assistance
    - Tool(s): Claude Code
    - What was generated: the staged-pipeline refactor of `generate_bidi.mjs`, the binding-neutral model extractor, and the Bazel target wiring.

Deviations from #17654 (intentional refinements):
- In the model a command's `result` is `null` when it returns nothing and `params` is `null` when it takes no arguments, so a consumer can spot void/argument-less commands from the model alone (the issue listed those fields without the empty case). The model also adds a `name` field — the bare operation name (`createUserContext`) next to the full `method` (`browser.createUserContext`).

Follow-ups / things noticed (left as-is here):
- Java/Ruby generators consuming `create-bidi-src_model.json` in future PRs.
- Recommend that Python change implementation to consume this as well to remove the regex work
- There is no mechanism to validate the model's completeness against the spec. The extractor finds commands/events by traversing the CDDL `CommandData`/`EventData` (and extension) unions, so anything the spec defines but never wires into a union is silently absent. This is a limitation of the CDDL structure, not the extractor. Example: `bluetooth.characteristicEventGenerated` and `bluetooth.descriptorEventGenerated` are defined but added to no `BluetoothEvent` union, so they don't appear in the model (not needed for Selenium, so not chasing this one down right now).

### 🔄 Types of changes
- New feature (non-breaking change which adds functionality)

### DIFF
--- a/javascript/selenium-webdriver/private/generate_bidi.bzl
+++ b/javascript/selenium-webdriver/private/generate_bidi.bzl
@@ -157,7 +157,7 @@ def generate_bidi_library(
     # Step 2: parse the merged CDDL once into the reusable AST artifact.
     # Exposed to the other bindings so they can consume it directly.
     ast_target = name + "_ast"
-    ast_out = "bidi-ast.json"
+    ast_out = name + "_ast.json"
     js_run_binary(
         name = ast_target,
         srcs = [":" + merged_name],
@@ -175,7 +175,7 @@ def generate_bidi_library(
     # Step 3: extract the binding-neutral command/event model from the AST.
     # Exposed to the other bindings so they can consume it directly.
     json_target = name + "_json"
-    model_out = "bidi-model.json"
+    model_out = name + "_model.json"
     js_run_binary(
         name = json_target,
         srcs = [":" + ast_target],


pujagani:
This looks good to me. Thank you, Titus!

Issue #17628, opened 2026-06-04
[🚀 Feature]: [bidi] Mark BiDi implementation classes and methods as internal across bindings


titusfortner:
### Description

See: https://github.com/SeleniumHQ/seleniumhq.github.io/pull/2659

Some users are treating Selenium's current BiDi implementation classes/methods as the stable, primary interface to that functionality. The intent is the opposite: we want to wrap everything users need in higher-level APIs so they never need the lower-level BiDi access directly. These BiDi implementation types are subject to change and are not the primary interface we want users to build on.

This proposes marking BiDi implementation classes and methods as **internal** in each binding, prioritizing **IDE visibility** (strikethrough, hover warnings) **without runtime warnings or deprecation notices** (we want to signal "internal / subject to change," not removal — so `@Deprecated`/`[Obsolete]` are deliberately avoided). The core problem is that **each language surfaces "internal" differently**, so we need a per-language mechanism.

Proposed per-language mechanism:

| Language | Mechanism | Strikethrough | Hover warning | Compile warning |
|----------|-----------|:---:|:---:|:---:|
| .NET | `[Experimental("SELENIUM001", UrlFormat=…)]` | ✓ | ✓ | ✓ (suppressible) |
| Java | JetBrains `@ApiStatus.Internal` + Javadoc | ✓ (IntelliJ) | ✓ | ✗ |
| Python | Docstring (`## ⚠️ Internal API — subject to change`) | ✗ | ✓ (varies) | ✗ |
| Ruby | YARD `@api private` + `@see` | ✗ | ✓ (RubyMine) | ✗ |

(.NET has no public-but-internal marker, so `[Experimental]` is the mechanism there; the tooltip text still reads "Internal API.")

Supporting work:
- Add a `/documentation/warnings/bidi-implementation` page: what these APIs are, why they're internal, what higher-level API to use instead, and the stabilization roadmap (if known).
- Identify all BiDi implementation classes/methods per binding and annotate them.
- .NET: suppress the new diagnostic ID internally in Selenium's own project files so we don't warn on our own usage.

### Have you considered any alternatives or workarounds?

- `@Deprecated`/`[Obsolete]` — rejected, implies removal rather than instability.
- Runtime warnings — rejected, too noisy for an API we ship and use internally.
- Documentation-only — insufficient on its own; IDE strikethrough/hover signals are the primary deterrent where the language supports them.

nvborisenko:
[...]

Issue #13991, opened 2024-05-20
[🚀 Feature]: Convenience methods for BiDi


titusfortner:
### Feature and motivation

Java & Python are missing convenience methods for enabling BiDi.

Right now we have....
.NET — options.UseWebSocketUrl = true
Ruby — options.web_socket_url = true
Python — options.enable_bidi = True
JS — enableBidi()
Java - enableBidi()

Should we update everything to use `enableBiDi()` name?

### Usage example

```
options.enableBiDi(true)
```

titusfortner:
@pujagani I'm guessing we don't have a good reason for not adding it to java, yet?

pujagani:
Thank you for bringing this up. Actually we can add it.

pujagani:
"options.enableBiDi(true)" - @titusfortner  Should the user need to pass in true? Isn't `enableBidi()` sufficient?

titusfortner:
@pujagani  I thought the other methods in Java Options that are boolean require passing in true/false.
So, we should do whatever matches the rest of our code.

pujagani:
That is a fair point. Though the naming, in that case, is like ` setEnableDownloads(boolean enableDownloads)`. I was thinking if the name is `enableBiDi`, then that means just enable it. But if we want to set a value then the name should be such. What do you think?

titusfortner:
If setXX(boolean) is the pattern, then we should follow that, though I agree it seems a bit much. @diemol any opinions on our conventions here?

diemol:
I am not sure. If we want to set a property or a value, I like the `setXYZ` pattern, but if we want to enable a feature, I prefer the `enableXYZ` pattern.

This is to be said, I saw `setEnableDownloads` too late; it was already released. I would have done `enableDownloads`.

pujagani:
Thank you folks. I think I will go with `enableBiDi`.

pujagani:
Present in python https://github.com/SeleniumHQ/selenium/commit/a51ddee3b10e4e0708f6f936668420bfccc72275

PR #17670, opened 2026-06-11
[adr] BiDi implementation boundaries

titusfortner:
📄 The decision, its rationale, considered options, and consequences are in the record file this
PR adds (`docs/decisions/17670-bidi-implementation-boundaries.md`); read it there. The
sections below are proposal notes and review logistics.

### 🔗 Related

- #17628 — per-language mechanism for marking BiDi implementation internal.
- #16095 — .NET's effort to finalize BiDi as a public, supported API; this record sets the opposite default (BiDi internal), so it redirects that work.
- #13991 — earlier decision that standardized the `enableBiDi` toggle (a quasi-ADR from before we had ADRs); this record reverses it, renaming the toggle off the protocol.
- #17717 — Selenium 5 release charter; this record settles its **BiDi support boundary** requirement.
- #17709 — three-layer proposal being closed in favour of this record (see Discussion).
- #17701 — how the internal generated layer is produced and structured; this record defers that to it.
- Prior discussion: SeleniumHQ/seleniumhq.github.io#2659.

### 📝 Proposal notes

- **This record settles the BiDi support boundary in #17717** and, on acceptance, the charter links back to it. It is not itself a milestone marker.
- **This answer places the BiDi module-ADR series** (17671–17677, 17681) as internal implementation, not supported surface — consistent with the charter deferring those subjects.

### 🗣 Discussion

- From Slack and the #17709 / #17701 threads: #17709 proposed an alternative — an explicitly supported BiDi surface with convenience methods on top. Those convenience methods aren't needed to give users access to the BiDi spec: the generated protocol code is itself directly callable ([#17701](https://github.com/SeleniumHQ/selenium/pull/17701)), now demonstrated end-to-end in Ruby by the merged [#17731](https://github.com/SeleniumHQ/selenium/pull/17731) (the full spec reachable as `BiDi::Protocol::BrowsingContext.new(driver).create(...)`). So native access comes through the generated code as an internal, unsupported surface, and the supported surface stays the protocol-neutral APIs of #17717. On that basis #17709 is closed in favour of this record. The conversation didn't settle what is public / private / internal or what those terms mean, so this record defines them.

### 📌 Tracking


docs/decisions/0001-bidi-is-an-implementation-mechanism.md
```diff
@@ -0,0 +1,56 @@
...
+
+Bindings use WebDriver BiDi to implement and extend the Selenium API. This record
+settles BiDi's relationship to the public API: what users program against, and where
+protocol code lives. CDP was managed differently across the bindings, but all of them
+resulted in users needing to understand CDP implementation details to work with it 
+to one extent or another. This is happening in BiDi the same way.
+
+## Decision
+
+BiDi is an implementation mechanism, not a public API. New capabilities are exposed as
```
nvborisenko: Low-level API should be publicly available, it is power. But without warranties, spec is living doc. Not clear how to emphasize it per language. What I see: spec is trying to be not breaking changeable.

In .NET/C# we already funded a plan how to react on breaking changes. I really don't understand why low-level implementation should be `internal`.. .NET is ready for changes.

diemol: This low-level api will be accessible through the `internal` namespace, as mentioned below:

```
Everything in a BiDi namespace is internal: still reachable for 
anything not yet wrapped, but not documented and subject to change without warning
```

pujagani:
I agree with the rationale explained in this ADR.

titusfortner:
Implementation tracking for this accepted record: #17814 — one checkbox per binding, linking each implementing PR as it lands.

PR #17665, opened 2026-06-10
[docs] add design decision record process and template

titusfortner:
### 💥 What does this PR do?

Adds `docs/decisions/` — a lightweight process for recording design decisions that apply across bindings (API shape, cross-binding semantics, deprecation commitments), plus the template for writing one. Decisions are proposed as PRs, discussed on the PR thread and at TLC meetings, and accepted by TLC consensus; the merged file is the canonical record, so settled questions get answered with a link instead of being re-litigated.

### 🔧 Implementation Notes

- Format is based on MADR/Nygard ADRs with one Selenium-specific addition: a per-binding status table so convergence across the five bindings is tracked in the record itself.
- These live in trunk rather than the website repo because the audience is contributors and the merge-is-acceptance process depends on the code-review flow. They can be rendered to selenium.dev later without moving the source of truth.

### 🤖 AI assistance

- [x] AI assisted (complete below)
    - Tool(s): Claude Code
    - What was generated: Initial drafts of the README and template, revised through discussion and review
    - [x] I reviewed all AI output and can explain the change

### 💡 Additional Considerations

- This PR follows the process it introduces: feedback is encouraged from all contributors; it merges by TLC consensus — a majority of TLC members responding with no unresolved objections, after at least one week open and discussion at a TLC meeting — with the Selenium Project Lead performing the merge. Merging adopts the process.

### 🔄 Types of changes

- Documentation (contributor process)

PR #17671, opened 2026-06-11
[docs] decision: BiDi events are awaited with expect_* context managers

AutomatedTester:
### 💥 What does this PR do?

Proposes a design decision record: **BiDi events are awaited with `expect_*` context managers**.

Today the only way to react to a BiDi event is a fire-and-forget callback. To *wait for* an event triggered by a user action, users must subscribe, act, then wait — which races (the event can fire before the wait starts) and forces everyone to re-implement capture-and-filter boilerplate. This record proposes a race-free, predicate-filtered `expect_*` family (arm-before-action context managers) over a small reusable `Subscription` primitive, consistent across all five bindings, as the foundation the other event-driven decisions (navigation, downloads, user prompts) build on.

### 🔧 Implementation Notes

- Follows the `docs/decisions/` process and template introduced in #17665 — **depends on #17665 landing** (this PR adds only `docs/decisions/0001-*.md`; no conflict with the template/README files).
- The decision is language-neutral and normative across bindings; Python is the reference implementation (the `expect_*` + `Subscription` work is implemented and a code PR is pending).
- No wire-protocol change: the backing BiDi events already exist and are emitted by current browsers.

### 🤖 AI assistance

    - Tool(s): Claude Code
    - What was generated: Initial draft of the decision record, derived from a CDDL-validated comparison of the Python BiDi surface against Playwright; revised through discussion and review

Part of a set of BiDi ergonomics decision records proposed together (events, user prompts, storage/cookies, network body+timing, emulation, navigation). This one should be considered first since the others build on the `expect_*` primitive it defines. Cross-binding convergence is tracked in the record's binding-status table.

### 🔄 Types of changes

### DIFF
--- a/docs/decisions/17671-bidi-events-awaited-with-expect-context-managers.md
+++ b/docs/decisions/17671-bidi-events-awaited-with-expect-context-managers.md
@@ -1,4 +1,4 @@
-# 0001. BiDi events are awaited with `expect_*` context managers
+# 17671. BiDi events are awaited with `expect_*` context managers
 
 - Status: Proposed
 - Date: 2026-06-11
@@ -53,11 +53,11 @@ Normative requirements for all bindings:
 
 Concrete shorthands each binding SHOULD provide (built on the generic primitive):
 `expect_request`, `expect_response`, `expect_console_message`, `expect_navigation`
-(see [0006](0006-navigation-awaited-with-expect-helpers.md)), `expect_user_prompt`
-(see [0002](0002-user-prompts-handled-through-typed-handler-api.md)), `expect_download`,
+(see [17676](17676-navigation-awaited-with-expect-helpers.md)), `expect_user_prompt`
+(see [17672](17672-user-prompts-handled-through-typed-handler-api.md)), `expect_download`,
 and — for tabs/windows the page opens itself — `expect_page` / `expect_popup` over
 `browsingContext.contextCreated`, returning the new context as a handle object
-(see [0008](0008-browsing-contexts-exposed-as-handle-objects.md)).
+(see [17681](17681-browsing-contexts-exposed-as-handle-objects.md)).
 
 Code sketch — Python (reference implementation):
 
@@ -118,11 +118,11 @@ const response = await driver.network().expectResponse('**/api/**', async () => 
   subscribe/unsubscribe I/O must not be held under a dispatch lock).
 - The other `expect_*`-based decisions (navigation, downloads, user prompts, and the
   `expect_page`/`expect_popup` handles in
-  [0008](0008-browsing-contexts-exposed-as-handle-objects.md)) build on this primitive, so
+  [17681](17681-browsing-contexts-exposed-as-handle-objects.md)) build on this primitive, so
   this record should land first.
 - The thread-safety fixes this surfaces (lock-guarded callback maps, non-busy-wait command
   completion, bounded event dispatch) are also the prerequisite for the multi-thread
-  concurrency contract in [0008](0008-browsing-contexts-exposed-as-handle-objects.md).
+  concurrency contract in [17681](17681-browsing-contexts-exposed-as-handle-objects.md).
 - No deprecations. Existing callback APIs are unaffected.
 
 ## Binding status


docs/decisions/17671-bidi-events-awaited-with-expect-context-managers.md
```diff
@@ -0,0 +1,144 @@
+# 17671. BiDi events are awaited with `expect_*` context managers
```
diemol, line 1: This ARD is very Python-idiomatic, since `expect_*` is what Playwright is doing for Python. Could we name this ADR so that the concept applies to all language bindings and avoid keywords specific to a particular programming language?

docs/decisions/17671-bidi-events-awaited-with-expect-context-managers.md
```diff
@@ -0,0 +1,144 @@
...
+1. **A time-of-check/time-of-use race.** If the event fires between "perform the action"
+   and "start waiting", it is lost. Correctly ordered code must subscribe *before* the
+   action — which the callback pattern does not make natural or obvious.
+2. **No predicate.** Users hand-write loops that inspect each event to find the one they
+   care about (a specific URL, a console error), reinventing the same filter every time.
+
+Playwright solved this with `with page.expect_event(...) as info: action()`, where the
+listener is armed on `__enter__` (before the action) and `info.value` blocks on
+`__exit__` until a matching event arrives or the timeout elapses. This is the single most
+common asynchronous pattern in browser automation, and Selenium has no equivalent.
```
diemol, line 27: Is there a way to make this section more general? This is specific to Python. Below, in the section of the options considered, the pattern `waitFor*` is discarded, but that is the convention Playwright uses for this concept. I would like to see, if possible, all the patterns we could use for all bindings. I don't mind resembling the ones from Playwright.

docs/decisions/17671-bidi-events-awaited-with-expect-context-managers.md
```diff
@@ -0,0 +1,144 @@
...
+
+The forces: the BiDi event surface already exists in every binding; the wire protocol
+needs nothing new; the only question is the *shape* of the waiting API and whether it is
+consistent across bindings.
+
+## Decision
+
+Every binding exposes an **`expect_*` family of context-manager (or block) helpers** that
+arm a one-shot, predicate-filtered subscription before the user action and resolve to the
+captured event after it.
```
diemol, line 37: The [comment](https://github.com/SeleniumHQ/selenium/pull/17671/changes#r3445703593) I made before applies to this section.

docs/decisions/17671-bidi-events-awaited-with-expect-context-managers.md
```diff
@@ -0,0 +1,144 @@
...
+- Existing callback registration (`add_event_handler`/`addListener`/`.on`) stays; this is
+  additive.
+
+Concrete shorthands each binding SHOULD provide (built on the generic primitive):
+`expect_request`, `expect_response`, `expect_console_message`, `expect_navigation`
+(see [17676](17676-navigation-awaited-with-expect-helpers.md)), `expect_user_prompt`
+(see [17672](17672-user-prompts-handled-through-typed-handler-api.md)), `expect_download`,
+and — for tabs/windows the page opens itself — `expect_page` / `expect_popup` over
+`browsingContext.contextCreated`, returning the new context as a handle object
+(see [17681](17681-browsing-contexts-exposed-as-handle-objects.md)).
```
diemol, line 60: I would prefer to define this naming for all bindings as part of this ADR. This would mean we are all on the same page. As I mentioned before, I do not mind using the same naming Playwright uses.

docs/decisions/17671-bidi-events-awaited-with-expect-context-managers.md
```diff
@@ -0,0 +1,144 @@
...
+
+- **`expect_*` context managers (chosen)** — arms before the action by construction;
+  matches the dominant Playwright idiom users already know; predicate + timeout built in.
+- **Document "subscribe first, then act" with the existing callbacks** — no new API, but
+  leaves the race as a footgun, provides no predicate or timeout, and every user
+  re-implements the capture-and-wait boilerplate. Rejected: solves nothing structurally.
+- **A future/promise returned from a `waitForEvent(...)` call** — viable in async
+  bindings, but in synchronous bindings it reintroduces the race (the call to start
+  waiting happens after the action) unless wrapped in a block anyway. Rejected as the
+  primary shape; bindings MAY offer it as a secondary convenience where idiomatic.
```
diemol, line 110: This ADR's section is the confusing part for me. My understanding is to follow the idea from Playwright, specifically the [naming](https://playwright.dev/docs/events#waiting-for-event) they use.

docs/decisions/17671-bidi-events-awaited-with-expect-context-managers.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Consequences
+
+- Users get race-free, predicate-filtered event waiting with no boilerplate; the most
+  common async pattern becomes a one-liner.
+- Each binding gains a small reusable `Subscription` primitive; implementing it surfaces
+  and forces fixes to event-dispatch thread-safety (callback maps must be lock-guarded,
+  subscribe/unsubscribe I/O must not be held under a dispatch lock).
```
diemol, line 118: Would it be necessary to specify how concurrency is handled? What happens if two events that match enter at the same time? If we don't specify it, we could end up with inconsistent behavior across language binding implementations.

docs/decisions/17671-bidi-events-awaited-with-expect-context-managers.md
```diff
@@ -0,0 +1,144 @@
...
+WebDriver BiDi is event-driven: navigation, network traffic, console output, user
+prompts, and downloads all surface as asynchronous events. Today the only way a binding
+exposes them is a fire-and-forget callback registration — e.g. Python's
+`add_event_handler(event, callback)`, Java's `addListener`, JavaScript's `.on(...)`.
+
+To *wait for* an event that a user action triggers (click a button, wait for the matching
+network response), users must register a callback, stash the event into a shared variable
+or queue, perform the action, then poll/wait. This has two problems:
+
+1. **A time-of-check/time-of-use race.** If the event fires between "perform the action"
```
titusfortner, line 18: Are you saying there's a race condition based on implementation or that users need to remember to add the handler before the action they want to handle? Requiring users to do the latter does seems natural and obvious for the callback pattern we are using. But if it is the former, I want to make sure I understand how that would happen.

docs/decisions/17671-bidi-events-awaited-with-expect-context-managers.md
nvborisenko: Agree. I think it is good that execution the same code repeatedly should produce the same results (execution path). Events come to us ordered (via websocket), we can/should deliver to the end accordingly.

docs/decisions/17671-bidi-events-awaited-with-expect-context-managers.md
```diff
@@ -0,0 +1,144 @@
...
+   and "start waiting", it is lost. Correctly ordered code must subscribe *before* the
+   action — which the callback pattern does not make natural or obvious.
+2. **No predicate.** Users hand-write loops that inspect each event to find the one they
```
nvborisenko, line 21: For 99% use cases awaiting a single event is good. What if I want to await 2 events?.. Not sure, may be we should think about it in more generic way. Challenge ourself.
cases:
- wait with timeout
- wait while some action, then return fast? or wait the action and only then return?
- wait different event types? like I want to await until new `context/tab` is created and then some http request completes at that `context/tab`

docs/decisions/17671-bidi-events-awaited-with-expect-context-managers.md
nvborisenko: expect/wait/assert/verify/until

Different semantics - different naming.

docs/decisions/17671-bidi-events-awaited-with-expect-context-managers.md
pujagani: I think identifying how handling of two events that match the same filter is important. I will help design and implement that internal working of the proposed API.

PR #17673, opened 2026-06-11
[docs] decision: storage exposes a typed cookie API and storage_state save/restore

AutomatedTester:
### 💥 What does this PR do?

Proposes a design decision record: **storage exposes a typed cookie API and `storage_state` save/restore**.

BiDi cookies use an awkward wire shape (`value` is a `BytesValue` descriptor, partitioning via a descriptor), so the common "set a session cookie" case carries needless ceremony. And the most-requested workflow — log in once, capture state, restore it to skip login in later sessions (Playwright's `storage_state`) — has no Selenium equivalent. This record proposes a typed `Cookie` value for `set_cookie`/`get_cookies` plus a `save_state`/`load_state` pair, while being explicit that cookies come from the BiDi protocol whereas `localStorage` is captured via `script.evaluate` and IndexedDB is out of scope.

### 🔧 Implementation Notes

- Follows the `docs/decisions/` process from #17665 — **depends on #17665 landing** (adds only `docs/decisions/0003-*.md`).
- Cookies are fully BiDi; web storage is **not** in the spec, so `storage_state` captures `localStorage` by evaluating script per origin — the record documents this split deliberately rather than implying uniform protocol support.
- No wire-protocol change.

### 🤖 AI assistance

    - Tool(s): Claude Code
    - What was generated: Initial draft of the decision record, derived from a CDDL-validated comparison of the Python BiDi surface against Playwright; revised through discussion and review

One of a set of BiDi ergonomics decision records proposed together. The cookie convenience and `storage_state` are kept in one record as a single cohesive storage-ergonomics decision; reviewers who prefer them split can request that. Cross-binding convergence is tracked in the binding-status table.

### 🔄 Types of changes

### DIFF
--- a/docs/decisions/17673-storage-typed-cookies-and-storage-state.md
+++ b/docs/decisions/17673-storage-typed-cookies-and-storage-state.md
@@ -24,11 +24,14 @@ Two forces are in tension. Cookies are fully covered by the BiDi protocol. But
 requires running script (`script.evaluate`) in each origin. A storage-state feature must
 be honest about that split rather than implying protocol support it does not have.
 
-In the layering of [17709](17709-bidi-api-layering.md) this is a **Layer 2** decision: an
-explicit BiDi module offered *beside* the Classic cookie API (`driver.manage().addCookie()`,
-which stays the Layer 1 surface and may itself become BiDi-backed), not a replacement for it.
-It earns first-class surface because it does things the Classic API cannot express —
-`storage_state` save/restore and partition-scoped cookies (a specific user context).
+Under [17670](17670-bidi-implementation-boundaries.md) this record specifies **supported,
+protocol-neutral API**, not a BiDi-shaped module offered beside it: the surface named here
+must not reference BiDi or hand the user a BiDi type, and the internal BiDi implementation
+stays reachable only by composing it with the driver. It is additive to the Classic cookie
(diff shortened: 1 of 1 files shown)


AutomatedTester:
You've put your finger on the cross-cutting question exactly, and I think it deserved a decision of its own rather than being answered ad-hoc in each module ADR — so I've written up #17709 ("BiDi is exposed in three API layers") to settle it for the whole series.

Short version: it's both A and B, layered.

- **Layer 1 — the unified high-level API.** `driver.manage().addCookie()` stays the primary surface and, when a BiDi session is active, becomes BiDi-backed under the hood (your Option B). Users don't change their code or learn a namespace.
- **Layer 2 — explicit BiDi modules.** `driver.storage` (this ADR) is the power-user tool *beside* the Classic cookie API, not a replacement — your "then `driver.storage` should be a middle-layer API." It's where things with no Classic shape live: `storage_state` save/restore, and partition-scoped cookies (targeting a specific user context, which `manage().addCookie()` can't express).

I've added a line to this ADR's Context declaring it a Layer 2 decision. Could you take a look at #17709? If we agree on the layering there, this one (and navigation/network/emulation) stop competing with the Classic surface — which I think is what made them feel like duplicate namespaces.

PR #17631, opened 2026-06-04
[py] Add high-level BiDi permissions API

AutomatedTester:
Introduces a `PermissionsManager` helper (`py/private/_permissions_handlers.py`)
that layers convenience methods on top of the existing `permissions.setPermission`
command, following the same handler-module pattern used for network and script.

## Changes

- New `_permissions_handlers.py` staged into the bidi package by Bazel
- `Permissions.__init__` now accepts an optional driver argument (same as `Script`)
- `bidi_enhancements_manifest.py` wires up `PermissionsManager` and exposes the
  convenience methods on the generated `Permissions` class
- `BUILD.bazel` adds `_permissions_handlers.py` to `create-bidi-src` `extra_srcs`
- `generate_bidi.py` extends the driver-arg `__init__` to the permissions module
- Unit tests and browser tests for all new methods

## New API

```python
# Grant one or more permissions
driver.permissions.grant("geolocation", origin=origin)
driver.permissions.grant(["geolocation", "camera"], origin=origin)

# Deny a permission
driver.permissions.deny("geolocation", origin=origin)

# Reset one, several, or all tracked overrides
driver.permissions.reset("geolocation", origin=origin)
driver.permissions.reset(["geolocation", "camera"], origin=origin)
driver.permissions.reset()  # clears all overrides applied via grant/deny/override

# Temporary override — resets to prompt on exit, even if an exception is raised
with driver.permissions.override("geolocation", "granted", origin=origin):
    ...
```

## API comparison: Selenium BiDi vs. Playwright

| Feature | **Playwright** (`BrowserContext`) | **Selenium BiDi** (`driver.permissions`) |
|---|---|---|
| Grant one | `grant_permissions(["geolocation"], origin=…)` | `grant("geolocation", origin=…)` |
| Grant many | `grant_permissions(["geo", "camera"], origin=…)` | `grant(["geo", "camera"], origin=…)` |
| Deny | ❌ not supported — denied is the browser default | ✅ `deny("geolocation", origin=…)` |
| Reset one | ❌ only bulk via `clear_permissions()` | ✅ `reset("geolocation", origin=…)` |
| Reset many | ❌ only bulk | ✅ `reset(["geolocation", "camera"], origin=…)` |
| Reset all | `clear_permissions()` | `reset()` — clears tracked overrides |
| Temporary override | ❌ no context manager | ✅ `override("geolocation", "granted", origin=…)` |
| Origin scoping | ✅ optional `origin=` | ✅ optional `origin=` |
[...]

PR #17677, opened 2026-06-12
[docs] decision: browser permissions are managed through grant/deny/reset/override helpers

AutomatedTester:
### 💥 What does this PR do?

Proposes a design decision record: **browser permissions are managed through grant / deny / reset / override helpers**.

BiDi exposes one permissions primitive (`permissions.setPermission`), already surfaced raw in the bindings. This record makes the high-level convenience layer a recorded **cross-binding** decision — `grant` (single or list), `deny`, `reset` (one/many/all via client-side tracking), an exception-safe `override(...)` block, plus the low-level `set_permission` escape hatch — so Java/Ruby/.NET/JavaScript converge on the same names and semantics. It is intentionally a superset of Playwright (which has only `grant_permissions` + bulk `clear_permissions`).

There is a working reference implementation reviewers can see **in action**: **#17631** (the Python high-level permissions API). This ADR links to it from the binding-status table so the exact shape is visible.

### 🔧 Implementation Notes

- Follows the `docs/decisions/` process from #17665 — **depends on #17665 landing** (adds only `docs/decisions/0007-*.md`).
- Records one **open cross-binding question** for review: the CDDL marks `origin` as required, but the Python reference (#17631) currently defaults it to `None` and omits it when absent (spec-non-conformant, browser-dependent). The record's normative rule is that every binding must require `origin` or adopt one documented default; the resolution should flow back into #17631.
- No wire-protocol change.

### 🤖 AI assistance

    - Tool(s): Claude Code
    - What was generated: Initial draft of the decision record, derived from a CDDL-validated review of the Python permissions implementation in #17631 against the W3C Permissions BiDi extension and Playwright; revised through discussion and review

Completes the set of BiDi ergonomics decision records proposed together (#17671–#17676). Unlike the others, this one already has a code PR (#17631) to anchor it. Cross-binding convergence is tracked in the binding-status table.

### 🔄 Types of changes

### DIFF
--- a/docs/decisions/17677-permissions-managed-through-grant-deny-reset-override.md
+++ b/docs/decisions/17677-permissions-managed-through-grant-deny-reset-override.md
@@ -25,14 +25,17 @@ For comparison, Playwright offers `grant_permissions([...], origin=...)` and a b
 `clear_permissions()` — no `deny` (denied is the browser default), no per-permission
 reset, and no scoped temporary override. The proposal below is intentionally a superset.
 
-In the layering of [17709](17709-bidi-api-layering.md) this is a **Layer 2** decision: an
-explicit BiDi module. Permissions have no Classic equivalent, so there is no Layer 1 surface to
-sit beside or route through — it is purely additive BiDi surface for users who reach for it.
+Under [17670](17670-bidi-implementation-boundaries.md) this record specifies **supported,
+protocol-neutral API**, not a BiDi-shaped module users reach into: the surface named here must
+not reference BiDi or hand the user a BiDi type, and the internal implementation stays
+reachable only by composing it with the driver. Permissions have no Classic equivalent, so
+there is no existing surface to sit beside or route through — this is purely additive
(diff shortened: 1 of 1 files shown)


AutomatedTester:
> Shouldn't we agree on #17681 and #17671 before reviewing this?

Probably but the idea of sharing all my ideas is to show how I have thought through all the different layers

AutomatedTester:
Following up now that the layering is written down: I've opened #17709 ("BiDi is exposed in three API layers"), the foundational decision this and the rest of the series sit under. As I said above, I shared everything at once to show the layered thinking — #17709 makes that explicit, with #17681 (browsing-context handles) and #17671 (events) being the other two foundational ones.

In that model this permissions ADR is a **Layer 2** module decision (I've noted that in its Context), so I'm happy to defer detailed review here until #17709 / #17681 / #17671 settle. One tie-in already in place: the `user_context` argument here cross-references #17681, where the isolation unit is defined.

PR #17681, opened 2026-06-15
[docs] decision: browsing contexts are exposed as handle objects

AutomatedTester:
### 💥 What does this PR do?

Proposes a design decision record: **browsing contexts are exposed as handle objects**.

Working with multiple tabs/windows over BiDi is awkward because there is no object representing a single browsing context — every operation is called on one shared module instance and takes the context id explicitly (`browsing_context.navigate(context=id, ...)`). For parallel work this means each worker repeats `context=` bookkeeping against shared state, with no per-tab unit to own. This record proposes a per-context **handle object** (returned by `create`/`get_tree`/`expect_page`) that carries the per-context operations as methods and scopes events to that context — matching the Playwright model where operations live on the object, which is exactly what makes one-context-per-worker parallelism clean. The flat module API stays as the compatibility surface.

### 🔧 Implementation Notes

- Follows the `docs/decisions/` process from #17665 — **depends on #17665 landing** (adds only `docs/decisions/0008-*.md`).
- Builds on the `expect_*` primitive (#17671) for `expect_page`/`expect_popup`.
- States a **concurrency contract** (one driver drivable from multiple threads, one context per thread) but leaves the enabling transport correctness as a per-binding internal with its own tests — not decided here.
- Records two follow-ups: exposing the **user context** as a grouping object (≈ Playwright `BrowserContext`), and fixing per-context event-subscription scoping.
- A full asyncio-native `Page` model is explicitly out of scope (its own RFC). No wire-protocol change.

### 🤖 AI assistance

    - Tool(s): Claude Code
    - What was generated: Initial draft of the decision record, derived from a review of the current `browsing_context` surface and BiDi transport against Playwright's object/concurrency model; revised through discussion and review

Extends the set of BiDi ergonomics decision records (#17671–#17677) with the parallelisation/object-model concept. The cross-binding name of the handle (`Page`-like / `Tab` / `BrowsingContextHandle`) is part of what's being decided. Cross-binding convergence is tracked in the binding-status table.

### 🔄 Types of changes

### DIFF
--- a/docs/decisions/17681-browsing-contexts-exposed-as-handle-objects.md
+++ b/docs/decisions/17681-browsing-contexts-exposed-as-handle-objects.md
@@ -103,6 +103,12 @@ await tab.navigate('https://example.com');
 await Promise.all(tabs.map(t => t.navigate(url)));
 ```
 
+```java
+// Java — same semantics, idiomatic shape
+BrowsingContext tab = driver.browsingContext().create(WindowType.TAB);  // -> handle
+tab.navigate("https://example.com", ReadinessState.COMPLETE);
+```
+
 ### User contexts (the isolation unit)
 
 Because a context's user context is fixed at creation (see Context), the isolation unit is the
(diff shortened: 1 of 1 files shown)


docs/decisions/17681-browsing-contexts-exposed-as-handle-objects.md
```diff
@@ -0,0 +1,290 @@
...
+- **The isolation types are binding-internal.** The user-context object and the handle types
+  are private/implementation structures — returned and usable, but not a prominent public class
+  to learn. Bindings keep the surface minimal (id access, `remove`, the factory method), since
+  the overwhelming majority of use is `isolated=True`.
+- **The default user context** is reachable through the same model, so ordinary (non-isolated)
+  tabs are not a special case.
+
+```python
+# 80% — isolation on/off, zero config, returns a normal tab handle
+tab = driver.browsing_context.create(type=WindowTypes.TAB, isolated=True)
```
nvborisenko, line 142: `browsing_context` here is still verbose, simple accessor for couple of methods even without internal state... Can be simplified:

```
tab = driver.new_tab(...)
ctx = driver.new_context(...)
ctx.new_tab(...)
```

`tab`, `context` here is an example of alternate names, honestly for me as for end user it is not obvious what is BiDi's `browsingContext` and `userContext`. `Tab` or `Page` is cleaner.

AutomatedTester: For the most part, I don't think users should need to understand contexts which is why I call it "isolation" in terms of browser context we already have the `switchTo` APIs which we can extend to understand isolation. 

I don't really want too much on driver as we can head into the space of creating a god object which is worse.

AutomatedTester: I think for the most part the `UserContext` doesn't need to be surfaced to the user. I would for the verbose case have this and for a more succinct API that allows people to do whatever they want

PR #17685, opened 2026-06-16
[adr] Network async/event API proposal

titusfortner:
📄 The decision, its rationale, considered options, and consequences are in the record file this PR
adds ([`docs/decisions/17685-network-handler-behavior.md`](docs/decisions/17685-network-handler-behavior.md));
read it there. This body is context for the decision, not a restatement of it.

### 🔗 Related

- #13993 — the handler surface (add / remove / clear for requests, responses, and authentication,
  under `driver.network`) agreed at the Dev Summit; this record consolidates it, and the per-binding
  implementation history is tracked there.
- #17717 — the Selenium 5 release charter.
- #17670 — BiDi is an internal implementation mechanism; `driver.network` is the protocol-neutral
  surface consistent with that boundary.
- @p0deje's [BiDi Protocol Design Document](https://docs.google.com/document/d/1icS4UmN6-x6LMENux7d9MDVpD30OQ9ZcjNIYLL94vFo/edit),
  where @nvborisenko and I raised behavior concerns that were never resolved before meetings stopped —
  this record settles them.

### 📝 Proposal notes

- **Satisfies a charter requirement.** This is the ADR the charter (#17717) names for the Network
  async/event API bucket; on acceptance the charter links back to it.
- **Already agreed — this record consolidates.** From the summit (#13993): the `driver.network`
  accessor, and add / remove / clear handlers for requests, responses, and authentication. The rest
  of the record settles how those handlers behave.
- **Non-blocking handlers are out of scope.** Handlers block by default, and that is all this record
  settles. Whether to add non-blocking observation — and, if so, whether as a mode on these methods
  or its own surface — is a separate ADR; the last TLC discussion leaned toward blocking everything
  for now, and @nvborisenko and @diemol have argued observation is distinct enough to stand on its
  own. Nothing here forecloses that: it arrives later as an opt-in, leaving the blocking default
  intact.
- **Implementation is unspecified,** but the behaviors must compose into one dispatch. For
  illustration, this Ruby sketch — each applicable handler run last-registered-first, the first to settle winning, and a raise failing the event — satisfies them:

```ruby
def process_request(request)
  @handlers.reverse_each do |h|
    next unless h.applies?(request)
    begin
      h.call(request)
    rescue
      fail_request(request)
      raise
    end
    return fail_request(request) if request.failed?
    return provide_response(request) if request.response?
    return continue_request(request) if request.submit?
  end
  continue_request(request)
end
```

### 🗣 Discussion

- [2026-06-18](https://www.selenium.dev/meetings/2026/tlc-06-18/#network-interception-adr-and-python-websocket-transport) — the
  handler manages data collection, not the user.
- [2026-07-16](https://www.selenium.dev/meetings/2026/tlc-07-16/#network-asyncevent-api-adr) — walked through point by point: handler
  references become objects, URL filtering belongs on the `add` methods with the argument type left
  to each binding, and the default disposition is to process other handlers. `submit` and
  observation-vs-interception left to revise.
- [2026-07-23](https://www.selenium.dev/meetings/2026/tlc-07-23/#network-asyncevent-api-adr) — URLs pass through as strings rather than
  being decomposed or rejected; wildcards error for now; authentication gets a convenience method
  alongside the callable handler.
- [2026-08-06](https://www.selenium.dev/meetings/2026/tlc-08-06/#network-asyncevent-api-adr) — observation vs interception dropped from
  the record, intercept by default; valid URL patterns pass through with an optional per-binding
  warning; scoping to user and browser context added.
- [2026-08-13](https://www.selenium.dev/meetings/2026/tlc-08-13/#network-asyncevent-api-adr) — an uncaught handler exception skips
  later handlers and submits the pre-handler state.
- [2026-09-03](https://www.selenium.dev/meetings/2026/tlc-09-03/#network-and-script-asyncevent-api-adrs) — an uncaught
  handler exception fails the event rather than sending the handler's partial state, the Rack-style middleware
  behavior @p0deje suggested (strict for now, relaxable later); the context-scoping text (one window handle or
  one user context) confirmed.

### 📌 Tracking

Tracking issue: #18019

### DIFF
--- a/docs/decisions/17685-network-handler-behavior.md
+++ b/docs/decisions/17685-network-handler-behavior.md
@@ -134,8 +134,9 @@ network.addAuthentication(UsernameAndPassword.of("user", "pass"),
 
 4. **When a handler settles a disposition, the first to do so resolves the event and
    stops the chain.** An event's chain is the registered handlers whose URL patterns (decision 2) and
-   scope (decision 11) match it; a handler outside the event's scope is not consulted, even when a
-   broader handler is what caused the event to be intercepted. The user settles the event by acting on
+   scope (decision 11) match it. Interception blocks the event as a whole, not per handler, so one
+   matching handler is enough to block it; a handler the event does not match is not consulted, even
+   though it was blocked on another handler's behalf. The user settles the event by acting on
    the object the callable receives. A handler that only stages mutations does not settle; it passes
    the event to the next handler (decision 5).
    * A request has three: `fail` (BiDi's `FailRequest`) ends it with an error;
@@ -277,7 +278,7 @@ network.addRequestHandler(new BodyCollection(), r -> log(r.body()));
     to one belongs in the callable.
 
     To scope a handler elsewhere the user passes either a window handle or a user context, never both.
-    A window handle targets that one tab, including a background tab that does not have focus. A user
+    A window handle targets that one window or tab, including one in the background without focus. A user
     context targets every window handle it contains, including ones opened later, so it scopes
     interception to a whole user context rather than a single known tab. The two are mutually
     exclusive: a handler is scoped by one or the other, and a binding rejects being given both. A


docs/decisions/17685-network-handler-behavior.md
```diff
@@ -0,0 +1,230 @@
...
+   * The alternative is being stuck with the top-level behavior everywhere, or not being able to
+     set top-level defaults at all.
+
+```ruby
+# Header will be there because removal is attempted before it is added
+network.add_request_handler { |r| r.add_header("X-Test", true) }
+network.add_request_handler { |r| r.remove_header("X-Test") }
+```
+
+4. **Handlers with uncaught exceptions are not processed.** Handling proceeds as if the handler
```
titusfortner: Actually, maybe the right answer here is to make this user-toggled with the default being to raise. I didn't consider this when suggesting alternatives below. Perhaps we do want to error-by-default but provide a global escape hatch while things are still in draft.

AutomatedTester: I feel fail fast should always be the right approach for people

titusfortner: First, to clarify, Python as written does the error behavior in this ADR (except for it doesn't discard staged changes).

As for fast-fail, that's correct for synchronous code evaluating the behavior. In network handlers, the failures will come from things the author didn't write like third-party scripts, analytics beacons, favicon retries, an auth challenge from an unrelated CDN. We don't want to necessarily couple the test's pass/fail to the open internet via a still-under-development BiDi protocol.

The reason I don't love this proposal as written is because when a user wants to mutate a specific thing as the intent of the test, silently ignoring it is a real problem, but I also don't want to de facto require users to always wrap everything in a try/catch which seems like a bad experience.

Honestly the desired default behavior is going to be different if a user wants to observe vs mutate, which might be a good reason to support the differentiation earlier. The alternative would be an error handling toggle for propagate vs log

docs/decisions/17685-network-handler-behavior.md
```diff
@@ -0,0 +1,230 @@
...
+## Context
+
+A user can register more than one handler for the same network phase, and matching handlers
+can disagree: the company framework always adds a test header, the local suite stubs all calls
+to a domain, and one test aborts a single call. Selenium must resolve that and provide a single
+response to the browser in a consistent and obvious way.
+
+The TLC discussed these ideas in a design document (by @p0deje). That document included
+prescribed implementation details that this ADR is avoiding, to focus on the user-facing
+behaviors we want rather than what needs to be implemented to achieve them.
```
diemol: I don't think this part is necessary unless you want to give more context by linking to the document. Which I think is not really necessary.

docs/decisions/17685-network-handler-behavior.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Decision
+
+This applies to request and response handlers, but not authentication handlers, since
+authentication should not use a callable.
```
diemol: I am confused. What does `this` mean? Is this "A user can register more than one handler for the same network phase, and matching handlers can disagree", the decision?

I am just looking for a paragraph that describes the decision.

docs/decisions/17685-network-handler-behavior.md
```diff
@@ -0,0 +1,230 @@
...
+     so it does not need to address this.
+
+```ruby
+# Nothing gets raised
+network.add_request_handler { |r| raise unless r.headers.include?("X-Test") }
+network.add_request_handler { |r| raise if r.request.headers.include?("X-Test") }
+network.add_request_handler { |r| r.add_header("X-Test", true) }
+```
+
+7. **A handler can set a complete status.** Allow the user to specify how the handler is disposed
```
diemol: What happens if the condition to set a complete status is never true? Will there be a timeout, or should we think of a different way to unblock the flow?

docs/decisions/17685-network-handler-behavior.md
nvborisenko: +1 to fail fast, notifying user your handler is bad. In low-level BiDi/.NET the default behavior is to break entire pipeline:

Example (pseudo code):
```
var handler = using network.intercept(r => raise exception)
network.navigate("selenium.dev")
```
`navigate` fails, swallowed. User sees exception at disposal. Dear user, investigate.

AutomatedTester: I wasn't suggesting my way was correct, I followed what other docs suggested which is why I did it.

titusfortner: As the ADR says though, I'd prefer if we could limit propagating failures to the tests/intentions not any potential malformed network call that the code or BiDi spec doesn't properly parse.

@nvborisenko yeah intercepts should propagate errors because they likely get into the intent of the test and are likely filtered in one fashion or another. My concern is when you create a handler to "log everything" that you're going to end up with a ton of unexpected errors because the internet is complicated and our spec is new. My concern goes away if we differentiate behavior between observers and intercepts, but if we continue with the "everything is an intercept for now" I see potential problems.

docs/decisions/17685-network-handler-behavior.md
titusfortner: Yes, this more accurately belongs in the PR body not the ADR itself.

After a few ADRs I think I better understand what the split should look like and I tried to clarify here: https://github.com/SeleniumHQ/selenium/pull/17697

docs/decisions/17685-network-handler-behavior.md
titusfortner: "This" == "This ADR"

Yeah, let me update this ADR

docs/decisions/17685-network-handler-behavior.md
titusfortner: This would never block automatically. It's just available to allow a user to block if they want. The code example below should actually look like:

```
handle = network.add_request_handler { |r| r.complete if condition }
do_the_thing_that_completes
wait.until { !network.get_completed_request(handle).nil? }
completed_request = network.get_completed_request(handle)

assert_on_request(completed_request)
```

If it isn't complete and you don't wait, then the getter just returns nil

titusfortner:
@diemol 
> Whereas https://github.com/SeleniumHQ/selenium/pull/17671 is about observation, which should not block, and only subscribes to events

In the last TLC meeting people were pushing for the idea that for now everything that can block will block in our implementations and we won't handle it any differently.

I'm saying they are in opposition because if you adopt this ADR, you don't need the extra methods proposed by the other ADR.

Also, I think it is wrong to characterize these as "middle-layer API." Anything that affects the code the user needs to write is user-facing and needs to be part of the high-level API contract we are creating that we want to maintain.

docs/decisions/17685-network-handler-behavior.md
diemol: But if the handler never resolves or reaches a true state, what happens? That is my question. 

Also, if there is a handler, it will eventually block. Otherwise, what is the purpose of it?

diemol:
> @diemol
> 
> > Whereas #17671 is about observation, which should not block, and only subscribes to events
> 
> In the last TLC meeting people were pushing for the idea that for now everything that can block will block in our implementations and we won't handle it any differently.
> 
> I'm saying they are in opposition because if you adopt this ADR, you don't need the extra methods proposed by the other ADR.
> 
> Also, I think it is wrong to characterize these as "middle-layer API." Anything that affects the code the user needs to write is user-facing and needs to be part of the high-level API contract we are creating that we want to maintain.

The way I see these two ADRs:
- This one is to actually manipulate the network call.
- The other one is for observation, because if you want to do observation and interception, then this could actually slow down tests, because you're intercepting and manipulating the call. I don't think that's what we want to do. 

That's why I think there are two other options: either observe or intercept here.

I still think that this is a middle-layer API. I don't want the user to have to handle all these callbacks and other things to manipulate their code in BiDi. I want this API to land and be used in a higher-level API to do network routing, mocking, and method interception, covering all of this implementation. 

In my head, we should have:
1. The CDDL parsing to generate the low-level stuff
2. The part where we map commands and do the instantiation of the driver
3. This current ADR, which is like the middle level
4. Some high-level features that make it very easy for users to actually interact with BiDi

If this current ADR is what we're doing at the high level, it is way too complex.

docs/decisions/17685-network-handler-behavior.md
nvborisenko: Intercepting and just listening are different cases. BiDi provides different capabilities. I feel it should be different api methods. Argue: to listen to network traffic, we just subscribe. To intercept it: we have to add interception and react accordingly. Seems unnecessary client/server roundtrip just for listening.

diemol: Yes, I see it how @nvborisenko sees it. That is why I think this ADR and #17671 complement each other.

docs/decisions/17685-network-handler-behavior.md
titusfortner: That's not how it was working, but I've talked myself out of even doing this one right now. So I'm removing it from the list.

titusfortner:
1. I removed the requirement for a "complete" resolution method because it was confusing things and also because I no longer think it is necessary right now
2. I added the requirement for toggling observation vs interception to minimize the confusion around this one vs #17671. Our normal handler method needs to support observation directly either now or in the future, we don't need (and shouldn't have) a separate method just because of that.
3. I want to push back on calling this middle layer. This is what we agreed to as the high layer at the TLC summit in SF 2 years ago. To clarify, what I think you are suggesting is methods that wrap the handler so users don't need to work with lambdas for common cases? If so, that represents a major expansion of the API  across all the bindings for no functional gain when we don't even have the underlying behavior we want sorted out.

Can we get the functionality implemented across the bindings then figure out which wrappers and helpers and fixtures to add later? 

I feel like we may need to step back from the ADRs and agree to the milestones and requirements for Selenium 5 again.

diemol:
>I removed the requirement for a "complete" resolution method because it was confusing things and also because I no longer think it is necessary right now

This is confusing because my interpretation is that the `complete` resolution is one of the core items of this ADR. Specially for network interception.

>Our normal handler method needs to support observation directly either now or in the future, we don't need (and shouldn't have) a separate method just because of that.

I think this bit is what #17671 and this ADR need to agree on. That is the foundation for both. 

>I want to push back on calling this middle layer. This is what we agreed to as the high layer at the TLC summit in SF 2 years ago. To clarify, what I think you are suggesting is methods that wrap the handler so users don't need to work with lambdas for common cases? If so, that represents a major expansion of the API across all the bindings for no functional gain when we don't even have the underlying behavior we want sorted out.

I am calling it a middle layer, which is core to what we want to achieve, because when I check Playwright's API, I see a high-level API that lets users handle less. But I think you have a point: landing all this "first and later" about major abstractions.

docs/decisions/17685-network-handler-behavior.md
titusfortner: I updated this to include requiring separation between Observation mode and Interception mode, so I split out the default behavior of the Error propagation to match.

titusfortner:
> my interpretation is that the complete resolution is one of the core items of this ADR.

Based on your questions it didn't seem like you understood my intention for including it or how it worked (it was never blocking anything, there was nothing in the lambda that needed resolving or a true state), so it seemed distracting from what I consider to be the primary things I want us to agree on.

Essentially 17671 and the "complete" method are 2 different ways to manage a wrapper around behavior the user can already easily do themselves for network calls. 
```
desired_request = nil
handler = driver.network.add_request_handler { |req| desired_request = req if matching_condition(req) }
do_something
wait.until { !desired_request.nil? }
driver.network.remove_request_handler(handler)
```
Is it actually a priority to provide convenience methods to replaces those 4 lines?

When I started thinking in those terms I decided that we don't need wrapper behavior right now, so if I want to reject 17671 on that basis, I should also reject my "complete" method. If we decide we want wrapper behavior, then I would still prefer "complete" over "expect_*" 

But also it feels like we're arguing over implementation details of a concept we haven't even agreed is behavior we need right now, so I also want to be able to step back one level on it.

For instance, I don't think "how to differentiate between observation and interception" is actually fundamental to any of our other conversations. The TLC meeting 2 weeks ago the general consensus from everyone else was to defer this decision until later because "intercept everything for now" works well enough for most things. I'd rather hammer it out now, but I'm fine with pushing it off. I only added it here to show that supporting it isn't a sufficient reason on its own to expand the API and accept 17671.

docs/decisions/17685-network-handler-behavior.md
```diff
@@ -0,0 +1,220 @@
...
+```ruby
+# Same method, two modes; the event object handed to the block differs
+network.add_request_handler { |r| r.fail if something }         # intercept: mutable, blocking
+network.add_request_handler(observe: true) { |r| log(r.url) }   # observe: read-only, non-blocking
+
+# An observed event object has no mutation methods, so trying to mutate raises
+network.add_request_handler(observe: true) { |r| r.fail }       # raises: observed events are read-only
+```
+
+2. **An intercept handler can specify event disposition.** Allow the user to specify how the event is disposed of
```
nvborisenko: In the end we will have a chain of delegates (methods/functions), it is kind of pipeline. And we should decide how this pipeline should work. Seems some hidden magic we should agree.

What if we review the concept of pipeline used in [ASP.NET](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/middleware/?view=aspnetcore-10.0):

```csharp
await using var _ = await bidi.InterceptAsync("**/*.png", (req, next) =>
{
  Console.WriteLine("I am starting to change request");
  await next(); // this is a game changer
  Console.WriteLine("I am done");
});
```

Here `next` is really next delegate in the pipeline. Thus user is fully responsible to decide whether continue pipeline, or halt it. I feel it could be powerful mechanism from implementation point of view, but don't see yet whether it is ergonomic way for users.

titusfortner: That would require a lot more orchestration work and couple all the handlers. We wouldn't be able to provide implicit resolution, and it looks like it could cause some weird behavior if not very careful with implementation. 
Even Playwright doesn't go here: fallback() is a flat one-way handoff, not a nested wrapper. Do you have a specific use case in mind that only the wrapped form supports, to make the overhead worth it?

nvborisenko: > We wouldn't be able to provide implicit resolution

We would, register default initial handler (first or last in the chain).

> Even Playwright

:)

I see only one practical use case within this implementation: exception handler and loggers. Within proposed "implicit calling chain" it is a magic what happens with "unresolved" request in the chain.. Or bad handler in the chain. Selenium should apply some behavior by default and document it. With `next()` approach this comes to user explicitly via contract. User is able to break chain fast (don't see use cases, he might resolve some conflicts).

I agree this is overengineering, just highlighted somebody might find it useful.

titusfortner: It's an interesting idea, but I think keeping the handlers independent will be better overall.

diemol: I will resolve this since it seems you are on the same page.

docs/decisions/17685-network-handler-behavior.md
```diff
@@ -0,0 +1,314 @@
...
+     fallback (process other handlers, if any).
+   * Selenium supports:
+     * Request: `fail` (Playwright's `abort`, BiDi's `FailRequest`), `respond` (Playwright's `fulfill`, BiDi's `ProvideResponse`), and `submit` (Playwright's `continue`, BiDi's `ContinueRequest`).
+     * Response: `fail` (BiDi's `FailRequest`), and `submit`: note that since we don't need to prevent a round trip from a request, whether this is a BiDi `ContinueResponse` or `ProvideResponse` can be an implementation detail based on whether a replacement body value is provided.
+
+```ruby
+# Names and params can match spec detail; response verbs mirror request (fail, submit)
+network.add_request_handler { |r| r.fail if something }
+network.add_request_handler { |r| r.respond(content: mocked_response) if something }
+network.add_request_handler { |r| r.add_header("X-Test", true) && r.submit if something }
```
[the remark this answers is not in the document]
titusfortner: this is true, but current code is fine for illustrative purposes.

docs/decisions/17685-network-handler-behavior.md
```diff
@@ -0,0 +1,384 @@
...
+
+11. **Body data is collected only when the handler opts in at registration.** A body is not available
+    by default; the handler declares that it needs the body when it is registered — not from inside
+    the callback, since the collector must be in place before the event — and Selenium then owns the
+    collector's lifecycle, size cap, and browser-support quirks. The body is readable on the event
+    inside that handler.
+    * The user never calls `addDataCollector` / `getData` or tears a collector down.
+    * There is no way to collect or read body data outside a handler; collection happens only through
+      the `add_request_handler` / `add_response_handler` registration.
+
```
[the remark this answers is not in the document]
titusfortner: Actually you can't access collect body from an intercepted response, so it would be hard to diverge here. I'll add where applicable.

docs/decisions/17685-network-handler-behavior.md
```diff
@@ -0,0 +1,385 @@
...
+   * How a binding lets the user pick the mode — a keyword argument, an options object, an overload —
+     is its own idiom; what is fixed is that it is the same method, not a separate observe one.
+
+```ruby
+# Same method, two modes; the event object handed to the block differs
+network.add_request_handler { |r| r.fail if something }         # intercept: mutable, blocking
+network.add_request_handler(observe: true) { |r| log(r.url) }   # observe: read-only, non-blocking
+
+# An observed event object has no mutation methods, so trying to mutate raises
+network.add_request_handler(observe: true) { |r| r.fail }       # raises: observed events are read-only
```
nvborisenko: This is big mistake. API allows users to write bad code! I strongly disagree.

titusfortner: You're right, thankfully this is just a bad code example. The reality is that the "r" is going to be a different class that won't even have those methods. The error for dynamic languages will be "no such method", not, "haha you can't do that, fooled you" 😂

nvborisenko: > How a binding lets the user pick the mode — a keyword argument, an options object, an overload —
     is its own idiom; **what is fixed is that it is the same method**, not a separate observe one.

One method for different purposes - bad.

titusfortner: We've had this conversation before and I've legitimately gone back and forth on which way I prefer.

The first thing is to acknowledge that these will technically be different methods in a static language because the object in the callable must be different. The more correct way to say this is that the method name should be the same even if the parameters are different. I'm updating the ADR to reflect that.

This really comes down to 2 questions:
1. how many methods do we need and what should they be named
2. are observation and interception too different to share a name.

I just don't see an easy way to properly name the methods in a reasonable way that fits with the naming scheme we've chosen, and moving to a different naming scheme would significantly increase the number of methods we would need to remove and clear everything.

As to how different these are:

* They aren't different at the protocol layer. If you have one "observation callable" and one "interception callable" for the same event, they both get the exact same payload, including "isBlocked" being true. 
* They aren't different at the point of registration, as they both must subscribe to the same event the same way with the same parameters. The only difference is that an interception must also call addIntercept.

docs/decisions/17685-network-handler-behavior.md
```diff
@@ -0,0 +1,387 @@
...
+
+```java
+network.addRequestHandler(r -> r.addHeader("X-Test", "true"));
+```
+
+7. **Later-registered handlers are consulted first.** This applies to every family — request,
+   response, and authentication. Registering an additional handler can mutate the state used by
+   previously registered ones.
+   * Ordering is a property of the chain, so it covers intercepting handlers; observing handlers
+     have no defined order, including relative to one another.
```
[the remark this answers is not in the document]
titusfortner: I'm not sure this matters, though. I think we can leave it as an implementation detail and not a behavior that needs to be defined.

docs/decisions/17685-network-handler-behavior.md
diemol: Maybe, for simplicity and to avoid confusing the user, we should use different names: one for observation, one for interception. We know how easily users can get confused.

nvborisenko: Method names and methods count is not a problem at all:
- naming can be any we like (language idiomatic but at least similar)
- we already chose a way to split functionality per domain (`driver.network.do_something`); how many this `do_something` we expect?.. Playwright chose flatten structure, looks simple. And, maybe, more convenient: we type "reques" and IntelliSense shows what we can do with "request". In domains world we should remember domain, so should type "network.reques".

Observation vs interception.. let's review simple code:
```
driver.network.add_request_handler("https://*.example.com/api/v1/**", req -> print(req.url))
```

Who can understand whether it is interception or just events listening? In any case selenium must _always_ enable interception and continue requests silently. This is unnecessary roundtrip, performance penalty, risks to make mistake in implementation (vendor issues) - cost just for events listening.

Interception at spec level brings more possibilities: we are able to intercept specific urls, filtered server side https://github.com/w3c/webdriver-bidi/issues/919. If we want to use one god method, then we should support the same possibility at client side (filtering via UrlPattern).

docs/decisions/17685-network-handler-behavior.md
```diff
@@ -0,0 +1,388 @@
...
+   event object: it can stage changes and settle the event, and traffic is paused until handling
+   resolves it.
+   * How a binding lets the user pick the mode — a keyword argument, an options object, an overload —
+     is its own idiom; what is fixed is that it is the same method name, not an
+     `add_request_observer` alongside an `add_request_intercept`.
+
+```ruby
+# Same method, two modes; the event object handed to the block differs
+network.add_request_handler { |r| r.fail if something }         # intercept: mutable, in the chain
+network.add_request_handler(observe: true) { |r| log(r.url) }   # observe: read-only, outside the chain
```
nvborisenko: I prefer `observe` by default (90%).
`network.add_request_handler(intercept: true)`
Only if TLC chooses one god method for observation and interception.

docs/decisions/17685-network-handler-behavior.md
titusfortner: > Maybe, for simplicity and to avoid confusing the user, we should use different names: one for observation, one for interception. We know how easily users can get confused.

The point is that there shouldn't be enough difference to require confusion. This is in reality mostly a performance optimization that 99% of users won't care about.

titusfortner: There also isn't a simple option for "use different names" without confusion or completely changing our naming scheme. Maybe I need to cover this more thoroughly in the considered options section.

titusfortner: I'm removing conversation of interception versus observation from the ADR, there was consensus at the TLC meeting a month ago that we are fine to implement intercept only to start with. I wanted to to at least define what observation would look like here, but I don't have energy for this particular fight right now. This ADR now covers handler implementation with all callables blocking. We can sort out the shape of observation later.

docs/decisions/17685-network-handler-behavior.md
titusfortner: We'll sort observation in a later ADR now.

docs/decisions/17685-network-handler-behavior.md
```diff
@@ -0,0 +1,343 @@
...
+```
+
+10. **Body data is collected only when the handler opts in at registration.** A body is not available
+    by default; the handler declares that it needs the body when it is registered — not from inside
+    the callback, since the collector must be in place before the event — and Selenium then owns the
+    collector's lifecycle, size cap, and browser-support quirks. The body is readable on the event
+    inside that handler.
+    * The user never calls `addDataCollector` / `getData` or tears a collector down.
+    * There is no way to collect or read body data outside a handler; collection happens only through
+      the `addRequestHandler` / `addResponseHandler` registration.
```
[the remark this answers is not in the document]
titusfortner: The requirement is that it's available if users requested it when adding the handler. I don't think we need to find the negative case either way we can leave it as an implementation detail.

docs/decisions/17685-network-handler-behavior.md
```diff
@@ -0,0 +1,351 @@
...
+     `respond` (`ProvideResponse`) replies with a mock, so nothing reaches the server; `submit`
+     (`ContinueRequest`) sends it on, with any staged mutations, and consults no further handler.
+   * `submit` is never required — a handler that settles nothing lets the event continue anyway
+     (decision 5) — and because it short-circuits the chain it can override what a shared handler
+     installed. That is occasionally necessary and easy to invoke by accident, so its name should read
+     as a deliberate, terminal override.
+   * A response has `fail` and `submit`. It has already round-tripped, so whether `submit` maps to
+     `ContinueResponse` or `ProvideResponse` follows from whether a replacement body was given.
+   * Within one handler, settling more than once is an error — after it settles, a further
+     disposition call raises rather than overriding the first.
```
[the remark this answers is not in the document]
titusfortner: raises means uncaught exception which means decision 7  applies.

docs/decisions/17685-network-handler-behavior.md
```diff
@@ -0,0 +1,351 @@
...
+Handlers are reached through `driver.network`, the supported protocol-neutral API established by the
+BiDi implementation boundaries decision ([#17670](https://github.com/SeleniumHQ/selenium/pull/17670));
+nothing here exposes a protocol type.
+
+## Decision
+
+By default, a handler blocks the event until it has run. The decisions below can be implemented in
+more than one way; the Ruby and Java examples show the user-facing shape, not a prescribed API.
+
+1. **Handlers can be added, removed, and cleared.** Each family — request,
```
nvborisenko, line 38: This is **push based** mechanism. I agree it should be blocked/ordered. Not yet clear if handler fails occasionally, then what (in .NET I fail fast, stop delivering, reveal at the end)?

Additionally I propose **pull based** events. Already in .NET:
```csharp
await using var stream = await bidi.Network.BeforeRequestSent.StreamAsync();

// do something
// stop stream (via dispose)

// magic I am sure available in any binding
var allNotFoundImageUrls = await stream.Where(r => r.Status == 404 && r.Url.EndsWith("png")).Select(r => r.Url).ToListAsync();
```

This is amazing API for free - **async enumeration**.

titusfortner: On failure: Good question. What's in the ADR isn't exactly what I thought. User must handle exceptions themselves. Unhandled exceptions are propagated to the user, but do not automatically stop the session. It should be possible for a user to have code that can catch an exception from the main thread still. The ADR currently says other handlers will process the event, but I think we need to change it so that the event stops being processed if there is an error, and the event is submitted. I'll bring this up at the next TLC meeting.

I like it pull-based streaming but, but it fits observation approach, and doesn't really work for interception, and async-enumeration only works with an async/await solution which isn't available for several of the bindings, so it can't be part of the shared contract.

docs/decisions/17685-network-handler-behavior.md
```diff
@@ -0,0 +1,385 @@
...
+```java
+network.addRequestHandler(r -> { if (!r.headers().containsKey("X-Test")) throw new AssertionError(); });
+network.addRequestHandler(r -> { if (r.request().headers().containsKey("X-Test")) throw new AssertionError(); });
+network.addRequestHandler(r -> r.addHeader("X-Test", "true"));
+```
+
+10. **Body data is collected only when a request handler opts in at registration.** A body is not
+    available by default; the handler declares that it needs the body when it is registered — not from
+    inside the callback, since the collector must be in place before the event — and Selenium then owns
+    the collector's lifecycle, size cap, and browser-support quirks. The body is readable on the event
```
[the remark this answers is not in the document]
titusfortner: I think this can be an implementation detail and doesn't have to be consistent across bindings. We can work out what makes sense.

docs/decisions/17685-network-handler-behavior.md
```diff
@@ -0,0 +1,390 @@
...
+   than deconstructing it to an object. Predicates are not accepted; a user who
+   wants one may write it inside the callable.
+
+   Everything specified by a url pattern argument must be resolvable by the
+   remote end. A binding may serialize a supported input into the remote's
+   pattern form, but it does no URL matching or pattern expansion of its own;
+   patterns are forwarded to the remote for evaluation, and input that is not a
+   valid pattern errors locally before anything is sent. A binding may log a
+   warning when a value looks like a glob, to flag that Selenium forwards it
+   rather than expanding it; that detection is optional and left to the binding
```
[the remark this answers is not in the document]
titusfortner: I'm ok with this being an implementation detail. Anything that fails remotely is ideally errored locally, but this may be easier to capture in some languages than others, worst case is an unnecessary round trip when the handler is registered, and I think that's fine.

docs/decisions/17685-network-handler-behavior.md
```diff
@@ -0,0 +1,386 @@
...
+   than deconstructing it to an object. Predicates are not accepted; a user who
+   wants one may write it inside the callable.
+
+   Everything specified by a url pattern argument must be resolvable by the
+   remote end. A binding may serialize a supported input into the remote's
+   pattern form, but it does no URL matching or pattern expansion of its own;
+   patterns are forwarded to the remote for evaluation, and input that is not a
+   valid pattern errors. A binding may log a warning when a value looks like a
+   glob, to flag that Selenium forwards it rather than expanding it; that
+   detection is optional and left to the binding rather than specified here.
```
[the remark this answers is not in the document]
titusfortner: It's just a warning, it's fine if not every binding warns for the same things, the behavior the user gets from the remote end will be the same. This is optional user assistance.

docs/decisions/17685-network-handler-behavior.md
```diff
@@ -0,0 +1,385 @@
...
+```
+
+```java
+network.addRequestHandler(new BodyCollection(), r -> log(r.body()));
+```
+
+11. **Handlers are scoped to the current browsing context by default.** A handler applies to the
+    browsing context that is active when it is registered, resolved from the current window; the user
+    may pass a browsing context, a user context, or both to scope it elsewhere. This lets a handler
+    apply to a context that is not the active one, such as a background tab that does not currently
```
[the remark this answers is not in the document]
titusfortner: addressed in commit 2d97033

docs/decisions/17685-network-handler-behavior.md
titusfortner: To come back to this, unhandled errors we are going to fail the event instead of submitting something.

diemol (approved):
This looks good to me, should we merge it?

PR #17689, opened 2026-06-18
[build] document Bazel worktree output-base cleanup

titusfortner:
### 💥 What does this PR do?
Documents that Bazel creates a per-checkout **output base** that, unlike the shared
`--disk_cache`/`--repository_cache`, is never garbage-collected when a worktree is deleted. In
worktree-heavy workflows these orphaned output bases accumulate and leak gigabytes. The README
"Using Worktrees" section and AGENTS.md "Agent workspace" now explain how to make a worktree
self-cleaning by pointing its output base inside the worktree (`startup --output_base=.local/bazel-out`),
with a Windows caveat preserving the existing `--output_user_root` guidance.

### 🤖 AI assistance
    - Tool(s): Claude Code
    - What was generated: Documentation wording in README.md and AGENTS.md

### 🔄 Types of changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,25 @@ dependencies (e.g. `http_archive` tarballs). Both directories grow unbounded ove
 periodically if disk space matters. Keep the cache on the same filesystem as your checkouts so Bazel
 can hardlink instead of copy.
 
+Bazel also creates a separate **output base** (compiled outputs, analysis cache, and the Bazel
+server) per checkout path. Unlike the caches above, it is **not** removed when you delete a worktree —
+so frequently created and discarded worktrees leak gigabytes of stale output.
+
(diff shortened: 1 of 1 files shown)


AutomatedTester:
This would be better in the `bazelrc` file so that everyone benefits

titusfortner:
I like the idea, but I don't think we want to force this on everyone all the time. The output-base leak is specific to temporary worktrees on Mac/Linux. It would break things for Windows, and using a relative path can cause weird behaviors unless you always run bazel from top level directory.

titusfortner:
merged part of this with https://github.com/SeleniumHQ/selenium/pull/17756 because my memory is bad, this becomes just a readme update.

PR #17691, opened 2026-06-19
[rust] Support pbzx-compressed macOS .pkg payloads

AutomatedTester:
## Description

Selenium Manager fails to install newer Firefox builds on macOS with:

```
cpio archive error: bad magic value encountered
  3: selenium_manager::files::uncompress_pkg
```

### Root cause

Mozilla changed the Firefox macOS `.pkg` Payload compression from **gzip-cpio** to Apple's **`pbzx`** (LZMA-block) format, starting on the **beta** channel (`153.0b1`). The `apple-flat-package` crate's payload decoder only sniffs the gzip header — for anything else it passes the bytes through untouched — so the `pbzx`/LZMA bytes were fed straight into the cpio reader, which rejected them as bad magic.

This was verified end-to-end: the downloaded `.pkg` is byte-for-byte valid (SHA256 matches Mozilla's published `SHA256SUMS`); only extraction was broken. The Payload begins with `70 62 7a 78` (`pbzx`) on `153.0b1` versus `1f 8b` (gzip) on `152.0`/ESR.

Today this affects the beta channel; once `153` promotes to stable it would break **all** macOS Firefox installs via Selenium Manager.

### Fix

`uncompress_pkg` now reads the `Payload` via the XAR reader, detects the `pbzx` magic, and decodes it (pbzx → xz blocks → cpio) before extraction. gzip and uncompressed payloads are handled exactly as before. No new dependencies (`xz2`/`flate2` were already present).

## Verification

- New unit tests cover stored, xz, multi-chunk, and non-pbzx inputs (`bazel test //rust:unit`).
- End-to-end on macOS (arm64):
  - `firefox --browser-version beta` (`153.0b1`, pbzx) — now extracts and resolves the browser path. ✅
  - `firefox --browser-version 121.0.1` (gzip) — still works (no regression). ✅

## Types of changes

## Checklist

### DIFF
--- a/rust/src/files.rs
+++ b/rust/src/files.rs
@@ -20,7 +20,7 @@ use crate::config::OS::WINDOWS;
 use crate::{Command, Logger, MACOS, format_one_arg, run_shell_command};
 use anyhow::Error;
 use anyhow::anyhow;
-use apple_flat_package::PkgReader;
+use apple_flat_package::{ComponentPackageReader, PkgReader};
 use bzip2::read::BzDecoder;
 use directories::BaseDirs;
 use flate2::read::GzDecoder;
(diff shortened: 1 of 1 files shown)


diemol:
I made this [change](https://github.com/SeleniumHQ/selenium/pull/17687/changes#diff-a223ca8815bf06c61bd96d36c5ebae1789d76dbc4d91ae6aa82d47f6d087f611L566) in the browser update PR. It seems we should be downloading the dmg instead of pkg. 

The dmg is what we are actually using in pinned browsers. 

Edit: I made means (Claudio and me).

AutomatedTester:
`pkg` files are better for enterprise environments especially MDM software, I think we may want to revert your change @diemol  once this change has landed

PR #17692, opened 2026-06-19
[adr] Java CDDL generator proposal

pujagani:
### 💥 What does this PR do?

Adds ADR recording the decision to generate Java WebDriver BiDi bindings from the CDDL spec.

### 🔧 Implementation Notes

The three cross-binding decisions recorded here : factory construction, subscription cancellation handles, and unified modules. These have user-visible consequences that needed to be settled before implementation begins. Python and JavaScript already generate from the same CDDL source. This record closes the loop on the Java binding's approach and captures the rejected alternatives (static utility pattern, raw listener IDs, split command/event classes) so they don't resurface in review.

Full generator design proposal: https://gist.github.com/pujagani/0a62197fd7c9b6d120e3a658fd7381d0 

### 🤖 AI assistance

- [X] AI assisted (complete below)
    - Tool(s): Claude
    - What was generated: The spec and the ADR
    - [X] I reviewed all AI output and can explain the change

The @Beta annotation on all existing hand-written BiDi classes ships in the next release, giving users one release of warning before the cutover PR removes them.

### 🔄 Types of changes

### DIFF
--- a/docs/decisions/17692-java-bidi-cddl-generator.md
+++ b/docs/decisions/17692-java-bidi-cddl-generator.md
@@ -1,8 +1,8 @@
-# 0001. Java WebDriver BiDi bindings are generated from the CDDL spec and expose stateful module classes with subscription-based event cancellation
+# Java WebDriver BiDi CDDL generator proposal
 
 - Status: Proposed
 - Date: 2026-06-19
-- Discussion: <!-- link to this record's PR once opened -->
+- Discussion: https://github.com/SeleniumHQ/selenium/pull/17692
 
 ## Context
 


titusfortner:
I think the manifest is a bad idea and easily avoidable. We don't need to store orchestration logic in the generated module classes. Keep the generated low-level in its own namespace and let a hand-written, checked-in coordination layer import it rather than be injected into it. If the generated code lands in a new namespace, we don't need to ensure parity with existing code. The low level things should all be private API, and any high level implementations can move to the new implementation. The CDDL spec should be the oracle, not the existing implementation.

titusfortner:
The Type Generation section covers the common cases well (enums from string-literal unions, POJOs from named groups, param builders), but a handful of CDDL shapes don't fit those patterns and tend to fail silently and need to be accounted for.

* Inline string-literal unions on a field. The plan generates enums from named string-literal unions, but some fields have the union inline (e.g. cacheBehavior, orientation, Initiator.type, ClientWindowInfo.state). These should still become enums — easy to special-case only the single-literal form and let multi-literal fields fall back to String, which drops the allowed values.
* Commands whose params are a union, not a record. A few commands don't have a flat parameter object: either a variant group nested inside the params (continueWithAuth, setClientWindowState, setGeolocationOverride, handleRequestDevicePrompt) or params that are a top-level union (session.unsubscribe). A POJO/builder pass that expects named members walks right past these and emits a command with missing or empty parameters.
* Cross-field rules the builder can't express. Some commands have constraints beyond per-field types: a field required only for a given discriminator value (continueWithAuth: credentials required when action=provideCredentials) and exactly-one-of (session.unsubscribe: events xor subscriptions). Worth deciding up front whether these are enforced by overloads, sealed types, or a runtime check — a plain fluent builder allows the invalid combinations.
* Enum wire values must come from the spec, not the constant name. Some values are hyphenated (powered-off, subscribe-to-notifications) and can't be reconstructed from a SCREAMING_SNAKE constant by case conversion. The toString() → wire-value mapping needs the literal copied from the AST.
* Degenerate result unions. Many command results are a union with a single member or just an empty result. Worth collapsing these rather than generating a one-variant type + dispatch for each.
* Open-map / extension construct (* text => any). Beyond the {* text => T} → Map mapping, CDDL allows arbitrary extra keys on some types. Decide whether the generator drops unknown keys or round-trips them.
* A coverage assertion in the generator. Given several of the above fail silently, I'd add a check that errors if any command, event, or field in the model isn't emitted. The Phase 8 diff against the existing hand-written classes won't catch these, since a couple of these commands aren't implemented there either.

titusfortner:
Actually I think this belongs in the bidi-model generation, let me whip up a PR.

titusfortner:
The PR that addresses my previous comments is here: #17700

pujagani:
Thank you for reviewing! I think we should go ahead with the shared model. Once the shape of that is good, the generator can build on top of that. @titusfortner

pujagani:
I am closing this one, since this is not a valid ADR. What Titus has proposed and existing ADR supersedes this one. For Java CDDL generator, I am not sure an ADR is required if the TCL is aligned on underlying design decisions.  The Java CDDL generator work can be done as part of code review.

PR #16486, opened 2025-10-22
[rb] implement client config class

titusfortner:
### 🔗 Related Issues

Addresses Ruby portion of #16269 

### 💥 What does this PR do?

* Creates config class for customizing everything currently customizable about Default http client
* Allows passing in the client_config into the driver constructors which passes it to the Bridge constructor
* Bridge Constructor has to manage all the ways the parameters can be in conflict

### 🔧 Implementation Notes

Went with a `Data` instance based on its goal of immutability. I overrode the constructor so that it only allows keywords and they are all optional
PORO seemed overkill and Struct is too loose

### 💡 Additional Considerations

1. My first thought was that we would be deprecating the existing http_client approach, but that only makes sense if net-http can do everything everyone needs it to do, and apparently there is a lot it can't do, so I guess we continue to allow people to also pass in an http_client? Or we figure out some kind of factory to use this client config with the various options?
2. There are several additional things that can be modified here; we could add more to the client config:
    * max redirects
    * retries
    * keep-alive timeout
    * http version
3. Finally, update #16472 to use this approach to support websocket values as well

### 🔄 Types of changes

- New feature (non-breaking change which adds functionality *and tests!*)

___

### **PR Type**
Enhancement

___

### **Description**
- Introduces `ClientConfig` immutable data class for HTTP client customization

- Allows passing `client_config` to driver constructors and bridge initialization

- Implements validation to prevent conflicting `http_client` and `client_config` parameters

- Refactors bridge initialization with URL normalization and server URL handling

- Adds comprehensive test coverage for new client config functionality

___

### Diagram Walkthrough

```mermaid
flowchart LR
  Driver["Driver Constructor"]
  Bridge["Bridge Initialize"]
  ClientConfig["ClientConfig Data Class"]
  HttpClient["Http::Default Client"]
  
  Driver -->|"client_config param"| Bridge
  Bridge -->|"validates params"| ClientConfig
  ClientConfig -->|"builds from config"| HttpClient
  Bridge -->|"uses http_client"| HttpClient
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>4 files</summary><table>
<tr>
[...]

Issue #12368, opened 2023-07-14
[🚀 Feature]: Configurable HTTP Client settings across bindings


titusfortner:
### Status
Python: https://github.com/SeleniumHQ/selenium/pull/13286 (partial)
Java:
Ruby:
JS:
.NET:

### Feature and motivation

*Update*: Here's the example page describing client settings and what all examples need to be added and what features still need to be implemented: [WIP - HTTP Client Documentation](https://deploy-preview-1433--selenium-dev.netlify.app/documentation/webdriver/drivers/http_client/)

I went down a rabbit hole of timeout settings in Selenium between the different bindings, and what the defaults are and what can be configured. *Edit: added WebDriverIO for reference*

Language  | Max Redirects | Read Timeout | Configurable
-- | -- | -- | --
Python | 3 | ∞ | No
Ruby | 20 | 60 | Timeout, not redirects
Java | 100 | 180 | Timeout, not Redirects
.NET | 50 | 60 | No
JS | ∞ | ∞ | No
*(WDIO)* | 3 | 120 | Yes
**recommended** | 20 | 120 | Yes!

According to Jari, Max Redirect should be 20.(https://github.com/SeleniumHQ/selenium/commit/93eee69ac30c0aabe34c092d0e14bf19e37c0a30)

It is also problematic that the default page load timeout is 300 when the read timeout in most languages is so much less (the driver will wait for the page long after the code errors and can't send a quit command). I'm not sure why it was set that high in the first place? I think we should change the defaults across the board to be 120 second read timeout and 115 second page load timeout. 

### Usage example

```java
ClientConfig config = ClientConfig.defaultConfig().readTimeout(Duration.ofSeconds(300)).redirects(25).baseUrl(url)
```

```rb
http_client = Selenium::WebDriver::Remote::Http::Default.new(timeout: 300, redirects: 25)
```

```py
client_config = ClientConfig()
client_config.timeout = 300
client_config.redirects = 25
```

titusfortner:
@christian-bromann does JS itself have defaults here, or did you completely add these?

christian-bromann:
> does JS itself have defaults here, or did you completely add these?

We are using a request library which I believe has some defaults in that regards. These particular timeouts are set by me. If you have better suggestions for them, let me know.

titusfortner:
[...]

PR #17699, opened 2026-06-21
[rb] add ClientConfig for HTTP client customization

titusfortner:
### 🔗 Related Issues
Implements Ruby portion of #12368 and replaces previous implementation #16486

### 💥 What does this PR do?
* Adds a `ClientConfig` for Ruby that centralizes HTTP transport settings for: open/read timeouts, max redirects, proxy, extra headers, user-agent, and server URL
* Users can pass this or an HTTP Client instance to any driver constructor, but not both 
* `ClientConfig` owns the defaults

### 🔧 Implementation Notes
- Using the read-timeout default change recommended in #12368 (60s → 120s).
- Decouples http client creation from the Bridge; the Bridge only sees the http_client instance
- The HTTP Client instance stores the client config and makes it accessible to the Bridge for future compatibility (socket information)
- The implementation in #16486 was an immutable `Data` class, but that's the wrong approach; this should take the same approach as Options allowing setting values after creation before usage

### 🤖 AI assistance
    - Tool(s): Claude Code
    - What was generated: Implementation, tests, and type signatures

- Consider deprecating http_client as a parameter the user can pass in and require everyone to use ClientConfig
- Deprecate the legacy `Http::Common` class methods in favor of ClientConfig class methods
- `ClientConfig` is positioned to carry the WebSocket settings for the BiDi follow-up.

### 🔄 Types of changes
- Breaking change (default read timeout 60s → 120s)

### DIFF
--- a/rb/lib/selenium/webdriver/common/client_config.rb
+++ b/rb/lib/selenium/webdriver/common/client_config.rb
@@ -43,7 +43,7 @@ module Selenium
       # @param [Numeric] open_timeout Seconds to wait for the connection to open.
       # @param [Numeric] read_timeout Seconds to wait for a response.
       # @param [Integer] max_redirects Maximum number of redirects to follow.
-      # @param [Proxy, Hash] proxy Proxy to use for the connection.
+      # @param [Proxy] proxy Proxy to use for the connection.
       # @param [Hash] extra_headers Additional headers to send with each request.
       # @param [String] user_agent Value to send as the User-Agent header.
       # @param [String, URI] server_url URL of the server to connect to.
@@ -89,7 +89,7 @@ module Selenium
         return unless proxy
 
         no_proxy = ENV.fetch('no_proxy', nil) || ENV.fetch('NO_PROXY', nil)
-        proxy = "http://#{proxy}" unless proxy.start_with?('http://')
+        proxy = "http://#{proxy}" unless proxy.match?(%r{\Ahttps?://})
         Proxy.new(http: proxy, no_proxy: no_proxy)
       end
     end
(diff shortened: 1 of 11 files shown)


rb/lib/selenium/webdriver/remote/http/default.rb
```diff
@@ -94,16 +116,20 @@ def request(verb, url, headers, payload, redirects = 0)
...
+              follow_redirect(response, redirects)
             else
               WebDriver.logger.debug("   <<<  #{response.instance_variable_get(:@header).inspect}", id: :header)
               create_response response.code, response.body, response.content_type
             end
           end
 
+          def follow_redirect(response, redirects)
+            WebDriver.logger.debug("Redirect to #{response['Location']}; times: #{redirects}")
+            raise Error::WebDriverError, 'too many redirects' if redirects >= client_config.max_redirects
```
aguspe, line 128: @titusfortner how does this affect the curb.rb client? I remember working on it a while ago to add timeouts and it has a MAX_REDIRECTS constant https://github.com/SeleniumHQ/selenium/blob/bd1d9c7ca84930d2574ab88b6001bf3b2f98aeeb/rb/lib/selenium/webdriver/remote/http/curb.rb#L88

titusfortner: I figured since the curb client is deprecated we should just leave it alone. So you can't use client config instance to construct curb client at all.

Curb is subclassing Common, not Default, so all the custom behavior in Default won't apply. The changes in Common should be completely backwards compatible for existing curb users, though.

rb/lib/selenium/webdriver/common/client_config.rb
```diff
@@ -0,0 +1,88 @@
...
+
+      def server_url=(url)
+        if url.nil?
+          @server_url = nil
+        else
+          url = url.to_s
+          url += '/' unless url.end_with?('/')
+          @server_url = URI.parse(url)
+        end
+      end
```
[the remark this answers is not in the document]
titusfortner: Only matters if we supported query/fragment here

PR #17701, opened 2026-06-21
[adr] BiDi low-level definitions are generated from a shared spec model without orchestration

titusfortner:
📄 The decision, its rationale, considered options, and consequences are in the record file
this PR adds; read it there. The sections below are proposal notes and review logistics.

### 🔗 Related

- Builds on #17670 (BiDi is an internal implementation mechanism); assumes it is accepted.
- Alternative to #17692 (Java BiDi generator proposal).
- Related: #17709 (user-facing API layering) — a separate axis from this record's internal structure.
- Shared-model reference implementation: #17700.

### 📝 Proposal notes

- One record because the three decisions are facets of one design for the generated definitions layer — one context, one rationale.
- Out of scope here: the high-level API, the transport substrate, and orchestration internals (left to each binding).

### 🗣 TLC discussion

_Pending — not yet discussed._

### 📌 Tracking

### DIFF
--- a/docs/decisions/17701-bidi-generated-protocol-layer.md
+++ b/docs/decisions/17701-bidi-generated-protocol-layer.md
@@ -10,13 +10,12 @@ A binding's BiDi support spans four layers, from the wire up:
 
 - **Transport / session substrate** — the connection, sending each command and correlating its
   response by the envelope id, and delivering inbound event frames upward. Domain-blind.
-- **Low-level definitions** — the types, command shapes, and event shapes the spec defines,
-  including the id types (a subscription, an intercept) and the commands that produce and consume
-  them.
-- **Orchestration** — the code that composes the definitions into capabilities and manages their
-  lifecycle: storing a subscription to unsubscribe later, mapping an intercept id to a handler,
-  matching an event to its registered callback, and wrapping events into the objects a handler
-  receives.
+- **Low-level definitions** — the types, command shapes, and event shapes the spec defines
+  (including the id types — a subscription, an intercept — and the commands that produce and consume
+  them), exposed by generated modules that can execute a single command.
+- **Orchestration** — the stateful coordination across those commands and events: storing a
+  subscription to unsubscribe later, mapping an intercept id to a handler, matching an event to its
+  registered callback, and wrapping events into the objects a handler receives.
 - **High-level API** — the protocol-neutral, idiomatic capabilities users program against.
 
 Bindings differ today in how they produce the definitions and how cleanly they separate them from
@@ -46,17 +45,19 @@ defines the shape, the generated definitions need not match the existing impleme
 shape or byte-for-byte. The shared model is the one place the spec is interpreted and normalized, so
 bindings stay consistent with each other; each still emits its own language-idiomatic code from it.
 
-**2. The generated definitions are read-only data, not the actionable surface.** Where a generated
-object reaches a user — for example, the request object inside a network handler — it is immutable,
-informational data; users act through the orchestration wrapper, not on the generated object.
+**2. Generated data objects are immutable; generated modules may call single commands.** The
+generated protocol data objects — command parameters and results, event payloads — are immutable
+and information-only. Generated domain modules may expose directly callable single-command methods
+(e.g. a generated `BrowsingContext.navigate(...)` that executes one command) as internal, unsupported
+implementation APIs — not the supported, user-facing surface, which is the high-level API.
 
-**3. Orchestration stays out of the low-level definitions.** The definitions carry the spec — types,
-commands, and events — and nothing that coordinates them: subscription lifecycle, event dispatch,
-mapping ids to handlers, and the objects handed to a handler live in a separate layer that imports
-the definitions, not spliced into the generated classes (e.g. through an enhancements manifest). A
-thin, stateless convenience over a single command is a lesser matter; what must not land here is
-coordination. The definitions stay a projection of the spec, so regenerating them never disturbs the
-layer that imports them, and they depend only on the transport's send-and-deliver interface.
+**3. Stateful orchestration stays out of the generated layer.** Generated modules execute individual
+commands, but own no stateful coordination: subscriptions, event dispatch, handler and intercept-id
+mappings, body collectors, and lifecycle management live in the orchestration layer, which imports
+the generated definitions rather than being spliced into them (e.g. through an enhancements manifest).
+A thin convenience over a single command may be generated; what must not land here is coordination.
+The generated definitions stay a projection of the spec, so regenerating them never disturbs the
+orchestration that imports them, and they depend only on the transport's send-and-deliver interface.
 
 ## Considered options
 
@@ -80,18 +81,17 @@ layer that imports them, and they depend only on the transport's send-and-delive
     inconsistencies forward.
   - Nothing supported depends on the generated definitions, so they need not match it.
 
-- **Generate each binding independently from CDDL** — every binding interprets the spec on its own
-  (or hand-writes), with no shared model.
-  - Each binding normalizes types, names, and gap handling separately, so the bindings drift apart
-    over time — the cross-binding inconsistency this record exists to prevent.
-  - A shared, binding-neutral model avoids the drift while still letting each binding emit
-    language-idiomatic code from it: the source model is shared, the generated definitions are
-    per-language.
+- **Generate each binding independently from CDDL** — every binding walks the CDDL/AST and builds
+  its own model with its own logic, rather than consuming one shared model.
+  - The grouping of modules, commands, and events follows the spec, but the normalization, naming,
+    and gap handling are re-derived per binding, so they drift apart over time.
+  - A shared, binding-neutral model makes those decisions once; each binding still emits
+    language-idiomatic code from it.
 
 ## Consequences
 
-- The supported API depends only on the wrapper, not on the generated definitions'
-  shape, so regenerating from a changed spec does not change what users program against.
+- What users program against is the high-level API, not the generated definitions, so regenerating
+  from a changed spec does not change it.
 - The generated definitions can live in their own namespace that the higher layers migrate onto,
   and the existing implementation is retired.
 - Orchestration and the high-level API are checked-in source, navigable and reviewable, and


docs/decisions/17701-bidi-generated-protocol-layer.md
```diff
@@ -0,0 +1,101 @@
...
+They also differ on source of truth — the CDDL spec or the existing implementation — and on whether
+they share one model or each interpret CDDL independently.
+
+This record decides the low-level definitions layer — what it is, where it comes from, and how the
+layers around it relate to it. This ADR assumes #17670 is accepted and that BiDi is an internal
+implementation.
+
+## Decision
+
+**1. The spec is the oracle, through one shared model.** The definitions are generated from a
```
diemol, line 41: Where is this "shared model"? What form does it have? Who owns it?

docs/decisions/17701-bidi-generated-protocol-layer.md
```diff
@@ -0,0 +1,101 @@
...
+
+**2. The generated definitions are read-only data, not the actionable surface.** Where a generated
+object reaches a user — for example, the request object inside a network handler — it is immutable,
+informational data; users act through the orchestration wrapper, not on the generated object.
+
+**3. Orchestration stays out of the low-level definitions.** The definitions carry the spec — types,
+commands, and events — and nothing that coordinates them: subscription lifecycle, event dispatch,
+mapping ids to handlers, and the objects handed to a handler live in a separate layer that imports
+the definitions, not spliced into the generated classes (e.g. through an enhancements manifest). A
+thin, stateless convenience over a single command is a lesser matter; what must not land here is
```
diemol: What does a thin, stateless convenience mean? Should we be more explicit here so that PRs implementing this understand what belongs in the low-level definition and what goes into a different layer?

docs/decisions/17701-bidi-generated-protocol-layer.md
```diff
@@ -0,0 +1,101 @@
...
+
+**1. The spec is the oracle, through one shared model.** The definitions are generated from a
+single shared, binding-neutral projection of the spec — not reconstructed from the existing
+implementation, and not parsed independently per binding. Because the spec, not the existing code,
+defines the shape, the generated definitions need not match the existing implementation — in API
+shape or byte-for-byte. The shared model is the one place the spec is interpreted and normalized, so
+bindings stay consistent with each other; each still emits its own language-idiomatic code from it.
+
+**2. The generated definitions are read-only data, not the actionable surface.** Where a generated
+object reaches a user — for example, the request object inside a network handler — it is immutable,
```
pujagani: This is exactly how all CDDL generator should be doing it. This was true for CDP generator too

docs/decisions/17701-bidi-generated-protocol-layer.md
```diff
@@ -0,0 +1,101 @@
...
+  - Coordination is harder to find, review, and type-check when it lives inside generated output.
+  - Thin conveniences are not the concern — the objection is coordination in the low-level layer.
+
+- **Derive the definitions from the existing implementation** — generate to
+  reproduce the current shape.
+  - Treats the existing implementation as the source of truth instead of the spec, carrying its
+    inconsistencies forward.
+  - Nothing supported depends on the generated definitions, so they need not match it.
+
+- **Generate each binding independently from CDDL** — every binding interprets the spec on its own
```
pujagani: I am curious to understand, in what cases will every binding interpret spec separately while generating from CDDL? They will follow the language specific practices but grouping of modules, events and commands will be same the spec right. Am I missing something?

pujagani:
I think to understand this better, I might beed an example of what the proposed shared spec model looks like.

docs/decisions/17701-bidi-generated-protocol-layer.md
```diff
@@ -0,0 +1,101 @@
...
+- Discussion: https://github.com/SeleniumHQ/selenium/pull/17701
+
+## Context
+
+The WebDriver BiDi specification is defined in CDDL (Concise Data Definition Language).
+A binding's BiDi support spans four layers, from the wire up:
+
+- **Transport / session substrate** — the connection, sending each command and correlating its
+  response by the envelope id, and delivering inbound event frames upward. Domain-blind.
+- **Low-level definitions** — the types, command shapes, and event shapes the spec defines,
```
AutomatedTester: Just for clarification, this means if we add a new CDDL for a different part of the web platform that we can't offer anything callable until we have the orchestration/high level agreed. The current python code at least makes the low level stuff callable and gives us an opportunity for later  high level/orchestration

AutomatedTester: if we can have a meaningful way to make things callable then we can close #17709  as I think we can assume the layer 1 is a given in that ADR, Layer 2 is the orchestration part, Layer 3 is callable, it doesn't need to in those modules but it shouldn't be overly hard to use. E.g. we have permissions CDDL in the tree, we're working on the ADR for high level/orchestration but the primitives are already being built. 

The callable primitives gives us an edge over competing tools since we're allowing people to keep up with the web platform as it is being written.

docs/decisions/17701-bidi-generated-protocol-layer.md
titusfortner: Do we need to define it for the ADR or leave it as an implementation detail?
What I meant is what we're doing in https://github.com/SeleniumHQ/selenium/pull/17700
It's taking the cddl npm to generate an AST, walks it to generate a bidi-model, takes the model and the ast and generates a schema that can be directly used by the bindings without additional walking/parsing.

This is currently done with bazel targets in javascript/selenium-webdriver that are marked visible to other bindings

docs/decisions/17701-bidi-generated-protocol-layer.md
titusfortner: I mean a wrapper that makes one protocol command easier to call; it does not store IDs, manage events, or control anything after that command finishes.

docs/decisions/17701-bidi-generated-protocol-layer.md
titusfortner: by "interpret the spec" I just meant that it would walk the ast and create the model with its own logic. This is essentially saying we could do what David did in Python for Java and Ruby.

titusfortner:
@pujagani 
```
bazel build //javascript/selenium-webdriver:create-bidi-src_schema
```
produces:
https://gist.github.com/titusfortner/2c35676da0d05f82195cec60acba8064

docs/decisions/17701-bidi-generated-protocol-layer.md
titusfortner: I think they've been immutable only in Java, Ruby & Python CDP was mutable, and for the current Python implementation of BiDi they have to be mutable.

docs/decisions/17701-bidi-generated-protocol-layer.md
titusfortner: Ok, yes, I wrote this in a confusing way. It should say that it generates immutable data objects for working with protocol responses, not that the generated code shouldn't generate callable command methods with parameters.

The boundary this ADR is proposing is that generated modules may execute individual commands, but they do not own stateful coordination such as subscriptions, event dispatch, handler mappings, intercept IDs, collectors, or lifecycle management.

I’ll update the decision to say:

Generated protocol data objects are immutable. Generated domain modules may expose directly callable single-command methods as unsupported implementation APIs. Stateful orchestration and lifecycle management remain in separate handwritten code.

So permissions primitives would be callable as soon as the CDDL is generated; a supported permissions API can be designed later. Does that address the callability concern behind #17709?


For instance, the Ruby implementation I'm working on supports this in our Bridge:

```ruby
def get(url)
  BiDi::Protocol::BrowsingContext.new(self).navigate(context: window_handle, url: url, wait: readiness)
end
```
Where BrowsingContext class is auto-generated from cddl schema and looks something like:

```ruby
def navigate(context:, url:, wait: UNSET)
  @transport.execute(cmd: 'browsingContext.navigate', params: NavigateParameters.new(context: context, url: url, wait: wait), result: BrowsingContext::NavigateResult)
end

```

docs/decisions/17701-bidi-generated-protocol-layer.md
pujagani: Understood. Thank you!

pujagani:
After the update you made, I now understand what this ADR is trying to address. I think it makes sense. For most part of JS CDDL generation , it follows this ADR. The manifest only has some convenience methods that hand-written code created, mainly to maintain backward compatibility (which we can remove to align with this ADR once the TLC agrees on it).

pujagani:
I only wonder where the access to unsubscribing to an event should live.

titusfortner:
The subscribe/unsubscribe commands are in the spec, so they're generated like any other command and directly callable on the generated module. What lives outside the generated code is the management: holding the subscription id, mapping it to a handler, and the cancellation behavior we want. The shape of that orchestration is decided elsewhere.

titusfortner:
As far as the JS implementation, the schema from #17700 would be a replacement for cddl2ts and require less follow-up work. Haven't looked into it in much detail, but that's the idea is that it is better shaped for what we need.

pujagani:
> As far as the JS implementation, the schema from #17700 would be a replacement for cddl2ts and require less follow-up work. Haven't looked into it in much detail, but that's the idea is that it is better shaped for what we need.

Thank you! I will look into JS and update things for JS accordingly.

docs/decisions/17701-bidi-generated-protocol-layer.md
pujagani: The more I think about this, I am not very clear where the module specific events live. And what does that look like?

docs/decisions/17701-bidi-generated-protocol-layer.md
diemol: Maybe I would just link it to where it can be found. No more than that.

titusfortner:
I'm gonna close this in favor of #17786
 I'll address any outstanding comments that are still relevant

docs/decisions/17701-bidi-generated-protocol-layer.md
titusfortner: Note that this specific item is no longer relevant with the new ADR

docs/decisions/17701-bidi-generated-protocol-layer.md
titusfortner: I think we got this one sorted properly in the new ADR we've defined the behavior required for the low-level implementation that requires callable primitives, regardless of whether it is automatically generated

Issue #5678, opened 2018-03-26
Instance ChroneDriver Error: ConnectFailure


Subhacse17:
## Meta -
OS:  
Ubuntu 16.04.4
Linux 4.10.0
Selenium Version:  
3.9.0
Browser:  
Chrome

Browser Version:  
65.0.3325.181
ChromeDriver version 2.35

## Steps to reproduce -
After instanciating a chrome Driver as in the screenshot1 we get a list of error written "Error: ConnectFailure (Connection refused)" Screenshot2

## Screenshot1
![image](https://user-images.githubusercontent.com/33542007/37905970-4185bfc2-311e-11e8-9272-5f99a8735828.png)

## Screenshot2
![image](https://user-images.githubusercontent.com/33542007/37906043-82a65cbe-311e-11e8-869c-08cd47144deb.png)

PS : for the first instance of chromedriver we don't get this error in console but for all the rest of instanciation we ll get it.

Help as soon as possible.

cgoldberg:
> Browser Version: 65.0.3325.181
> ChromeDriver version 2.35

try a more recent version of Chromedriver.  Version 2.35 only supports Chrome v62-64.
https://chromedriver.storage.googleapis.com/2.35/notes.txt

testphreak:
@cgoldberg even with `chromedriver` v2.36 or v2.37, the issue persists. It appears the issue has been [fixed](https://bugs.chromium.org/p/chromedriver/issues/detail?id=2340) but not released. Even tried their canary [build](https://console.cloud.google.com/storage/browser/chromedriver-data/continuous/?prefix=chromedriver_mac64_2.37) from an hour back.

The issue is being seen on all platforms.

Subhacse17:
@cgoldberg  I am getting the same with new driver also and in firefox also.

Subhacse17:
@testphreak I am not trying to signing into any google product still getting. Can you suggest me any Soln? On windows I am not getting this error.Only with Linux I have this, and more interesting things this error did not cause any block to the code. All tests are running as expectation. This error coming n numbers of time and just increasing my log file length of Jenkins.

lmtierney:
I do not have this issue on Linux, check your environment and provide [concise reproducible test case](http://sscce.org/) so that we can act on this issue. See [CONTRIBUTING.md](https://github.com/SeleniumHQ/selenium/blob/master/CONTRIBUTING.md#issue-contributions)

barancev:
[...]

Issue #1234, opened 2015-11-10
2.48 doesn't trigger javascript in link's href on click()


merike:
Please see https://drive.google.com/file/d/0Bzx3gCDqfzVdQUFoWHdzdWlFYVU/view?usp=sharing and https://drive.google.com/file/d/0Bzx3gCDqfzVdUmhhckFkWVFlbTQ/view?usp=sharing

The test case there alerts with 2.47.1 but not with 2.48.0 or 2.48.2. This is with Firefox 42.0 32bit on 64bit machine.

doberkofler:
+1

barancev:
Can't reproduce the issue with Selenium 2.49.1 and Firefox 43.0.4, please retest with the latest Selenium version.

merike:
Strange, for me 43.0.4 and 2.50.1 still don't trigger the alert. Wait.. except when Firefox is in the background. Why would it work differently when Firefox is the foreground application? Unless it's a reverse bug of #825?

barancev:
Closing the issue as there is no reproduction case provided.

PR #17706, opened 2026-06-22
[grid] Fix classpath packaging Redis-backed for SessionQueue

VietND96:
<!-- Thanks for Contributing to Selenium! -->
<!-- Please read our contribution guidelines: https://github.com/SeleniumHQ/selenium/blob/trunk/CONTRIBUTING.md -->

### 🔗 Related Issues
<!-- Example: Fixes #1234 or Closes #5678 -->
<!-- If the reason for this PR is not obvious, consider creating an issue for it first -->

### 💥 What does this PR do?
  Two related bugs prevented using a Redis-backed SessionQueue via CLI:                                                                                                                                                                                    
                                                                                                                                                                                                                                                           
  1. `--sessionqueue-implementation` was not a recognised flag. Passing it produced: Was passed main parameter '--sessionqueue-implementation' but no main parameter was defined in your arg class. The @Parameter annotation was simply absent from NewSessionQueueFlags, unlike the equivalent flags in DistributorFlags and SessionMapFlags.                                                                                                                                                               
  2. `RedisBackedNewSessionQueue` was not bundled into the sessionqueue server jar. The class is loaded via Class.forName() at runtime, so it must be an explicit Bazel dep of the httpd target even though nothing imports it at compile time. Without it, startup failed with ClassNotFoundException.

### 🔧 Implementation Notes
<!--- Why did you implement it this way? -->
<!--- What alternatives to this approach did you consider? -->

### 🤖 AI assistance
<!-- Check one. See CONTRIBUTING.md for policy details. -->
    - Tool(s):
    - What was generated:

<!--- Are there any decisions that need to be made before accepting this PR? -->
<!--- Is there any follow-on work that needs to be done? (e.g., docs, tests, etc.) -->

### 🔄 Types of changes
<!-- ✂️ Please delete anything that doesn't apply -->

### DIFF
--- a/java/src/org/openqa/selenium/grid/sessionqueue/httpd/BUILD.bazel
+++ b/java/src/org/openqa/selenium/grid/sessionqueue/httpd/BUILD.bazel
@@ -7,6 +7,7 @@ java_library(
     visibility = [
         "//java/src/org/openqa/selenium/grid:__pkg__",
         "//java/test/org/openqa/selenium/grid/router:__pkg__",
+        "//java/test/org/openqa/selenium/grid/sessionqueue/httpd:__pkg__",
     ],
     deps = [
         "//java:auto-service",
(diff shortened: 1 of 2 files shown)


java/src/org/openqa/selenium/grid/sessionqueue/httpd/BUILD.bazel
```diff
@@ -17,7 +17,10 @@ java_library(
         "//java/src/org/openqa/selenium/grid/server",
         "//java/src/org/openqa/selenium/grid/sessionqueue",
         "//java/src/org/openqa/selenium/grid/sessionqueue/config",
+        # Loaded via reflection when the default local implementation is used.
```
diemol: Why are these comments needed?

VietND96: Okay, I removed redundant comment lines.

Issue #17696, opened 2026-06-20
[🚀 Feature]: [dotnet][bidi] `IEventStream<>.ConfigureAwait()` call is ambiguous


kevinoid:
### Description

I noticed an ergonomic issue with the event streams added in https://github.com/SeleniumHQ/selenium/pull/17349 (which are fantastic, by the way!) that seemed worth noting:  When `ConfigureAwait(bool)` is called on `IEventStream<>` to control capturing context during disposal, it causes compile error [`CS0121`](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/overload-resolution) due to ambiguity between [`TaskAsyncEnumerableExtensions.ConfigureAwait(IAsyncDisposable, bool)`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskasyncenumerableextensions.configureawait?view=net-10.0#system-threading-tasks-taskasyncenumerableextensions-configureawait(system-iasyncdisposable-system-boolean)) and [`TaskAsyncEnumerableExtensions.ConfigureAwait<T>(IAsyncEnumerable<T>, bool)`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskasyncenumerableextensions.configureawait?view=net-10.0#system-threading-tasks-taskasyncenumerableextensions-configureawait-1(system-collections-generic-iasyncenumerable((-0))-system-boolean)).

This isn't a showstopper, since a cast to either type can be added to resolve the ambiguity.  However, this seems less than ideal as it clutters the code for a common operation, and steepens the learning code as developers need to understand how to resolve the error initially and determine which cast to make.

Perhaps it would make sense to provide a `.ConfigureAwait<T>(IEventStream<T>, bool)` extension method which forwards to [`TaskAsyncEnumerableExtensions.ConfigureAwait<T>(IAsyncEnumerable<T>, bool)`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskasyncenumerableextensions.configureawait?view=net-10.0#system-threading-tasks-taskasyncenumerableextensions-configureawait-1(system-collections-generic-iasyncenumerable((-0))-system-boolean)) to avoid the ambiguity?

kevinoid:
[...]

PR #17711, opened 2026-06-23
[dotnet][bidi] fix CS0121 ambiguous ConfigureAwait on IEventStream<T>

Chandan25sharma:
Problem
-------
IEventStream<T> inherits both IAsyncEnumerable<T> and IAsyncDisposable. System.Linq.Async ships a ConfigureAwait overload for each of those interfaces in TaskAsyncEnumerableExtensions:

  ConfigureAwait<T>(IAsyncEnumerable<T>, bool)   -> ConfiguredCancelableAsyncEnumerable<T>
  ConfigureAwait(IAsyncDisposable, bool)          -> ConfiguredAsyncDisposable

Both overloads match when the receiver is an IEventStream<T>, so the compiler emits CS0121 ("The call is ambiguous between ...").  Users had to add an explicit cast to either interface to work around this, which is both surprising and hard to discover.

Fix
---
Add EventStreamExtensions.ConfigureAwait<T>(this IEventStream<T>, bool) that explicitly delegates to the IAsyncEnumerable<T> overload.  Because a more-derived receiver type takes priority over the two BCL overloads, the compiler now resolves the call unambiguously without any cast at the call site.  The IAsyncEnumerable<T> variant is the correct choice here since callers use ConfigureAwait to control context capture inside `await foreach` loops, not during disposal.

Testing
-------
EventStreamExtensionsTests verifies at the type level that the return type is ConfiguredCancelableAsyncEnumerable<T> (not ConfiguredAsyncDisposable) and at the behavioural level that events flow through a ConfigureAwait(false) enumeration correctly.

Fixes: #17696

### DIFF
--- a/dotnet/src/webdriver/BiDi/EventStreamExtensions.cs
+++ b/dotnet/src/webdriver/BiDi/EventStreamExtensions.cs
@@ -1,0 +1,51 @@
+// <copyright file="EventStreamExtensions.cs" company="Selenium Committers">
+// Licensed to the Software Freedom Conservancy (SFC) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The SFC licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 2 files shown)


nvborisenko:
On hold. This extension method doesn't resolve the issue. Why `ConfigureAwait()` over iteration instead of disposal, hidden magic.

Potentially would be resolved by https://github.com/SeleniumHQ/selenium/pull/17705

nvborisenko:
Already fixed conceptually via design.

Thanks for your contribution!

PR #17714, opened 2026-06-24
Make Either.stream() and iterator() empty for a left value

vasiliy-mikhailov:
## Problem

`Either.stream()` returned `Stream.of(right())` and `Either.iterator()` returned `Collections.singleton(right()).iterator()` unconditionally, regardless of which side the `Either` holds. For a left-valued `Either`, `right()` is `null`, so:

- `stream()` yields a single `null` element (and throws `NullPointerException` in downstream operations such as `toList()` on some pipelines), and
- `iterator()` yields one `null` element instead of being empty.

A left-valued `Either` has no right value, so both should be empty — consistent with treating `Either` as a (possibly empty) collection of the right value.

## Fix

Return an empty stream / iterator when the `Either` holds a left value.

## Test

Adds `EitherTest` covering `stream()` and `iterator()` for both left and right values. The left-value cases fail on `trunk` (non-empty, containing `null`) and pass with this change.

### DIFF
--- a/common/repositories.bzl
+++ b/common/repositories.bzl
@@ -50,8 +50,8 @@ js_library(
 
     http_archive(
         name = "linux_beta_firefox",
-        url = "https://ftp.mozilla.org/pub/firefox/releases/153.0b3/linux-x86_64/en-US/firefox-153.0b3.tar.xz",
-        sha256 = "99d48331fa82fb354d655fc04b2d175558c569673d9c5c6a443971e32c3b81a0",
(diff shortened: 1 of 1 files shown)


vasiliy-mikhailov:
Thanks @diemol! Added `java/test/org/openqa/selenium/internal/BUILD.bazel` — a small `java_test_suite` that globs the `*Test.java` files in that package. That wires up `EitherTest`, and it turns out `MapsTest`/`MultimapTest`/`SetsTest` weren't running either (the directory had no `BUILD.bazel`), so this picks those up too. I mirrored the pattern used in sibling test packages; happy to move it into the parent `BUILD.bazel` instead if you'd prefer.

PR #14094, opened 2024-06-06
[rb] implement navigation commands with BiDi

titusfortner:
### Description
* creates context manager class, constructed with bridge instance
* pulls page load strategy from capabilities and uses in required methods
* we eventually want to figure out how to store current browsing context as state on the bridge (I think)? For now this always gets the current window handle first
* deprecates `BrowsingContext`

### Considerations
* navigate and reload return `NavigateResult` objects; should I return it as a Struct or just ignore it since we aren't doing anything with it and not promising a return value? 
* Should these BiDi classes be marked private api?
* I want to implement this as `BrowsingContext` instead of creating a new class and deprecating that one; do we need to be backwards compatible with this since we didn't mark it private api?

### Motivation and Context
Ruby implementation of #13995

___

### **PR Type**
Enhancement, Tests

___

### **Description**
- Added `ContextManager` class to handle navigation commands using BiDi.
- Deprecated `BrowsingContext` class in favor of `ContextManager`.
- Enhanced `BiDiBridge` with new navigation methods: `get`, `go_back`, `go_forward`, and `refresh`.
- Added integration tests for navigation methods.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement
</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>bidi.rb</strong><dd><code>Add `ContextManager` to autoload list in BiDi module</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
rb/lib/selenium/webdriver/bidi.rb

- Added `ContextManager` to autoload list.

</details>
    

  </td>
  <td><a href="https://github.com/SeleniumHQ/selenium/pull/14094/files#diff-b7660c7b53fec20ce95452b2afdbf306e9c36363e5bf8812863375b7bc96e42e">+1/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>browsing_context.rb</strong><dd><code>Deprecate `BrowsingContext` class in favor of `ContextManager`</code></dd></summary>
<hr>
      
rb/lib/selenium/webdriver/bidi/browsing_context.rb

- Added deprecation warning for `BrowsingContext` class.

</details>
    

  </td>
[...]

PR #17715, opened 2026-06-24
[py][bidi] Route navigation commands through BiDi

shossain786:
### 🔗 Related Issues
Part of #13995

### 💥 What does this PR do?
Implements the **Python** binding for routing the Classic navigation commands through WebDriver BiDi when BiDi is enabled, following the pattern of the merged Ruby PR (#14094) and the in-review Java PR (#17232).

When the session has BiDi enabled (the remote end advertised a `webSocketUrl`), `WebDriver` now routes:

| Classic | BiDi |
|---|---|
| `get(url)` | `browsing_context.navigate(context, url, wait=<readiness>)` |
| `back()` | `browsing_context.traverse_history(context, delta=-1)` |
| `forward()` | `browsing_context.traverse_history(context, delta=1)` |
| `refresh()` | `browsing_context.reload(context, wait=<readiness>)` |

When BiDi is not enabled, the existing Classic HTTP commands are used unchanged.

The `pageLoadStrategy` capability is mapped to a BiDi readiness state: `normal → COMPLETE`, `eager → INTERACTIVE`, `none → NONE` (defaulting to `COMPLETE` to preserve Classic `get()` blocking semantics). The current window handle is used as the browsing-context id.

### 🔧 Implementation Notes
The generated `BrowsingContext` module already exposes `navigate` / `traverse_history` / `reload`, so no generator changes were needed — the change is contained in `webdriver.py` plus a new test module.

Unlike Ruby (which introduced a dedicated `BiDiBridge`), the Python binding has a single `WebDriver` class, so this uses an inline `if self._is_bidi_enabled()` branch in the four navigation methods plus two small helpers (`_is_bidi_enabled`, `_page_load_readiness`). If maintainers prefer a separate bridge/mixin to keep the Classic and BiDi paths cleanly split, I'm happy to refactor — see Additional Considerations.

### 🤖 AI assistance
    - Tool(s): Claude Code (Claude Opus)
    - What was generated: the navigation routing in `webdriver.py`, the two helper methods, and the new `bidi_navigation_tests.py`

- **Design question:** inline branching in `WebDriver` vs. a dedicated BiDi bridge/mixin. Open to either — guidance welcome.
- The new tests run under the `--bidi` flag (auto-globbed into `BIDI_TESTS` via the `*bidi*_tests.py` pattern).

### 🔄 Types of changes

### DIFF
--- a/py/selenium/webdriver/remote/webdriver.py
+++ b/py/selenium/webdriver/remote/webdriver.py
@@ -736,10 +736,12 @@ class WebDriver(BaseWebDriver):
     def _is_bidi_enabled(self) -> bool:
         """Returns True if WebDriver BiDi is enabled for this session.
 
-        BiDi is enabled when the remote end advertised a ``webSocketUrl`` in
-        the session capabilities (i.e. the user requested it via options).
+        BiDi is enabled when the remote end advertised a ``webSocketUrl`` URL
+        string in the session capabilities. A remote end may echo the requested
+        capability back as a boolean, so only a string value (the actual socket
+        URL) counts as enabled; otherwise navigation falls back to Classic.
         """
-        return bool(self.caps.get("webSocketUrl"))
+        return isinstance(self.caps.get("webSocketUrl"), str)
 
     def _page_load_readiness(self) -> ReadinessState:
         """Maps the session's ``pageLoadStrategy`` capability to a BiDi readiness state.


py/selenium/webdriver/remote/webdriver.py
```diff
@@ -708,15 +714,44 @@ def switch_to(self) -> SwitchTo:
...
 
     def refresh(self) -> None:
         """Refreshes the current page."""
-        self.execute(Command.REFRESH)
+        if self._is_bidi_enabled():
+            self.browsing_context.reload(context=self.current_window_handle, wait=self._page_load_readiness())
+        else:
+            self.execute(Command.REFRESH)
+
+    def _is_bidi_enabled(self) -> bool:
```
cgoldberg, line 736: I'm not sure this method is needed. Elsewhere in this class, we already use `self._websocket_connection` to check for BiDi (it is `None` when BiDi is not enabled).

shossain786: Good catch on the consistency. The reason I checked the `webSocketUrl` capability rather than `self._websocket_connection` is that `_websocket_connection` is initialized lazily — it stays `None` until the first BiDi feature is accessed (each BiDi property calls `_start_bidi()` on demand). Since navigation is usually one of the first things a user does, gating on `_websocket_connection` would route the initial `get()`/`back()`/etc. through Classic and only switch to BiDi *after* the user independently touched some other BiDi API — so it can't gate the very navigation that's meant to use BiDi.

Checking the `webSocketUrl` capability instead means "this session was created with BiDi enabled," which is the condition we actually want here.

Happy to adjust though — I can rename/inline the helper, or eagerly start the websocket when BiDi is enabled so `_websocket_connection` becomes a reliable signal to gate on — whichever you prefer. Worth noting this also overlaps with the still-open "how is BiDi support detected" ADR referenced in #17717.

cgoldberg: that makes sense... it's probably fine as you have it then

Issue #17554, opened 2026-05-23
[🚀 Feature]: Change Selenium Manager PATH behavior


titusfortner:
### Description

The original plan for the selenium manager contract was for it to be opt-in while in beta and then the full release would be opt-out

When we moved PATH detection from bindings to Selenium Manager, we broke this contract a little bit for convenience sake when there weren't side-effects. But, for instance, when we added plausible integration, we didn't exempt drivers found on path from that flow, which further eroded the original contract. 

I think there is only one place remaining where we make a distinction, and I think it has become a distinction without a difference.

Right now if SM finds a driver on PATH it will use it unless the user has specified a version in their options class. 
If the driver version does not match the browser SM detects on the system it doesn't download a new driver, it just logs a warning, even though we know that it will error. 

This is intentional because of the contract we agreed to at the beginning:  if user is managing the drivers themselves, SM must respect it.

The actually desired implementation is in order:
* use browser with specified version
* use browser installed on system
* use latest stable browser
Then download browser if necessary and always ensure the most correct driver for that browser we can provide (lots of edge cases here but we handle these pretty well right now)

I can't think of a scenario where making this change now negatively impacts users. The biggest change is that many users with slightly out of sync driver/browser versions will have more correct driver versions downloaded now, and fixing it will be a good thing.

PR #17717, opened 2026-06-25
[docs] Selenium 5 release charter

titusfortner:
📄 The agreed scope is in `docs/plans/selenium-5.md` — read it there. This body is the context:
why we need it, the required decisions, and the possible additions to consider.

### Why we need this

We have dozens of ADR PRs in flight, and many assume different answers to "what is Selenium 5."
Binding-level efforts in .NET and Python are each moving on their own reading of the goal. We need
one agreed boundary for what blocks the release before arguing implementation details. This
charter is that boundary; the cross-binding mechanics are left to the ADRs it lists.

### How the scope has moved

- **~2 years ago, at the TLC Summit in San Francisco**, the plan
  ([`Selenium 5.0` milestone](https://github.com/SeleniumHQ/selenium/milestone/16)) was: implement
  the BiDi API, take Selenium Manager out of beta, deprecate CDP, with big breaking changes — and a
  Classic fallback so enabling BiDi never changed existing behavior.
- **Last year, at the TLC Summit in Valencia**, the agreement was to implement every existing
  method with BiDi, each with a Classic fallback.
- **December 2025** I consolidated to a minimal, focused release in
  [seleniumhq.github.io#2556](https://github.com/SeleniumHQ/seleniumhq.github.io/pull/2556).

This charter takes the #2556 scope. The Classic-fallback mechanism, in the plan since the start,
was removed from 5.0 scope at the 6/25 TLC meeting and folded into the classic-over-BiDi migration.
Where an older milestone or issue conflicts with the charter, the charter governs 5.0 scope.

### Required decisions

Each must be settled by an accepted ADR before 5.0 ships; **none is accepted yet**. Each ADR has to
consolidate **both** what we agreed to implement — the surface, much of it settled at the summits
before we used ADRs — **and** how it behaves. The existing PRs cover only parts of that.

- **BiDi support boundary** — is BiDi internal-only, or is some surface supported, and where is the
  line? [#17670](https://github.com/SeleniumHQ/selenium/pull/17670) proposes internal-only.
  [#17709](https://github.com/SeleniumHQ/selenium/pull/17709) describes three layers but does not
  say which is supported, why, or how they differ — so it does not yet define the boundary. The ADR
  has to pin where supported ends and internal begins, and on what basis.
- **Network async/event API** — the handler surface (add / remove / clear for requests, responses,
  and authentication; [#13993](https://github.com/SeleniumHQ/selenium/issues/13993)) and how
  handlers behave (observe vs. intercept, disposition, ordering;
  [#17685](https://github.com/SeleniumHQ/selenium/pull/17685)). #17685 covers only the behavior; the
  agreed surface still needs to be folded into the same decision.
- **Script and logging async/event API** — the surface (pinned scripts; console, JS-error, and
  DOM-mutation handlers; [#13992](https://github.com/SeleniumHQ/selenium/issues/13992)) and how it
  behaves. No ADR yet.
- **Selenium Manager released API** — formalize the interface and behaviors needed to drop the Beta
  label and release independently of the bindings (e.g. how an unrecognized command is handled).
  Proposed in [#17741](https://github.com/SeleniumHQ/selenium/pull/17741); issues
  [#11694](https://github.com/SeleniumHQ/selenium/issues/11694),
  [#15754](https://github.com/SeleniumHQ/selenium/issues/15754),
  [#17554](https://github.com/SeleniumHQ/selenium/issues/17554).

### Possible additions to consider

Subjects the charter defers; this PR is where we decide whether any should be required for 5.0. The
charter decides only whether a **subject** is in scope — each linked ADR is the **design** proposal
if it is. Each has an in-flight proposal:

- **Full classic-over-BiDi migration** — routing classic commands through BiDi now, and the
  partial-BiDi support / Classic fallback that goes with it (navigation: [#17715](https://github.com/SeleniumHQ/selenium/pull/17715),
  [#17232](https://github.com/SeleniumHQ/selenium/pull/17232), POC [#13190](https://github.com/SeleniumHQ/selenium/pull/13190)).
- **DevTools deprecation** — [#13990](https://github.com/SeleniumHQ/selenium/issues/13990).
- **Convenience layers** — `expect_*` event waiting ([#17671](https://github.com/SeleniumHQ/selenium/pull/17671)),
  response body & request timing ([#17674](https://github.com/SeleniumHQ/selenium/pull/17674)),
  navigation waiting ([#17676](https://github.com/SeleniumHQ/selenium/pull/17676)).
- **Browser context API** — browsing contexts as handle objects
  ([#17681](https://github.com/SeleniumHQ/selenium/pull/17681)).
- **Capability mapping** — exposing BiDi capability modules as high-level APIs: permissions
  ([#17677](https://github.com/SeleniumHQ/selenium/pull/17677)), storage / cookies
  ([#17673](https://github.com/SeleniumHQ/selenium/pull/17673)), emulation
  ([#17675](https://github.com/SeleniumHQ/selenium/pull/17675)), user prompts
  ([#17672](https://github.com/SeleniumHQ/selenium/pull/17672)).

### DIFF
--- a/docs/plans/selenium-5.md
+++ b/docs/plans/selenium-5.md
@@ -1,8 +1,8 @@
 # Selenium 5 Release Charter
 
-- Status: Proposed <!-- Proposed | Accepted | Released | Superseded -->
+- Status: Proposed
 - Owner: Selenium Technical Leadership Committee (TLC)
-- Discussion: <!-- link to this charter's PR -->
+- Discussion: https://github.com/SeleniumHQ/selenium/pull/17717
 
 ## Purpose
 


docs/plans/selenium-5.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Required for release
+
+### BiDi support boundary — _ADR pending_
+
+How the WebDriver BiDi protocol is exposed to users across the bindings. The ADR sets where the
+boundary sits between supported Selenium API and internal implementation, and how each binding
+marks it.
+
+### Classic fallback for BiDi-backed commands — _ADR pending_
```
diemol: This could be an implementation detail of the migration to Selenium 5 ADR.

docs/plans/selenium-5.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Script and logging async/event API — _ADR pending_
+
+The cross-binding API for pinned scripts (pin / unpin / execute) and for console-message,
+JavaScript-error, and DOM-mutation handlers.
+
+### Selenium Manager finalized API — _ADR pending_
+
+A finalized, documented API the bindings invoke, versioned independently of the client.
+
```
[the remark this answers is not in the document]
titusfortner: The intent was to make this part of the ADR for "Classic fallback for BiDi-backed commands" since it is kind of an implementation detail. But agreed, especially if we are going to move that bullet to out of scope (not a requirement for Selenium 5) we need to decide separately if we want to define the mechanism for deciding BiDi vs Classic path for sessions and implement something like Navigation, which Ruby already does.

docs/plans/selenium-5.md
```diff
@@ -0,0 +1,73 @@
...
+### Full classic-over-BiDi migration
+
+Routing every classic command through BiDi. Selenium 5 requires the fallback mechanism, not
+coverage of every command.
+
+### DevTools deprecation
+
+Deprecating or removing DevTools (CDP) support.
+
```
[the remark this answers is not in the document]
titusfortner: Yes, this is the Out of Scope section.

docs/plans/selenium-5.md
titusfortner: 1. This is too specific of subject. It should be "Partial BiDi implementation support" where we decide what it means to support a remote end that doesn't implement every feature.
2. One specific issue with the "per command" fallback is that not all commands can switch between classic and bidi depending on what previous commands were sent
3. This issue is not just about "what if Safaridriver implements BiDi support but not for everything at first" it can also be about supporting older browsers with inicomplete BiDi implementations 

The consensus at the 6/25 TLC meeting was that this probably doesn't belong as a blocker or a prerequisite for Selenium 5, and it can be part of the ADR for the Classic Migration.

p0deje: We're taking that out per 7/2 TLC meeting.

docs/plans/selenium-5.md
```diff
@@ -0,0 +1,73 @@
...
+### Full classic-over-BiDi migration
+
+Routing every classic command through BiDi.
+
+### Partial BiDi implementation support
+
+What it means to support a remote end that does not implement every BiDi feature — choosing the
+BiDi or Classic path per session (not every command can switch mid-session), for older browsers or
+drivers with incomplete BiDi. Part of the Classic-migration ADR.
```
gsnedders: Given there continue to be additions to BiDi, e.g., https://github.com/w3c/webdriver-bidi/commit/54175aaf, it's probably worthwhile defining _what_ counts as partial/incomplete support, because otherwise [both Chrome and Firefox are out-of-scope](https://wpt.fyi/results/webdriver/tests/bidi/browsing_context/start_screencast/start_screencast.py?run_id=5957828169957376&run_id=5148960267436032). Even going back to spec changes around a year ago, you'll run into Firefox [not supporting the `emulation.setScriptingEnabled` command](https://wpt.fyi/results/webdriver/tests/bidi/emulation/set_scripting_enabled/enabled.py?run_id=5957828169957376&run_id=5148960267436032), in case that makes it out-of-scope.

That said, I think it should be obvious that support of `browsingContext.StartScreencast` and `emulation.setScriptingEnabled` are not really comparable to support of `browsingContext.create`.

More interestingly would be cases like supporting everything _except_ the `emulation` module, or supporting only the commands that correspond to Classic ones. Are those out-of-scope?

docs/plans/selenium-5.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Script and logging async/event API — _ADR pending_
+
+The cross-binding API for pinned scripts (pin / unpin / execute) and for console-message,
+JavaScript-error, and DOM-mutation handlers.
+
+### Selenium Manager released API — _ADR pending_
+
+Formalize the interface and behaviors required to drop the Beta label and release independently of
+the bindings.
```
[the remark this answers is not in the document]
titusfortner: The charter defines scope, not the versioning scheme. "Release independently of the bindings" is the requirement; the concrete scheme is settled in the Selenium Manager ADR not here.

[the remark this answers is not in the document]
titusfortner: specific CLI flag is implementation detail for the Selenium Manager ADR

docs/plans/selenium-5.md
titusfortner: Agreed it needs a real definition, and you're right that by a strict reading every implementation is effectively partial — Ruby already routes some methods (navigation) over BiDi today, so 5.0 ships a classic/BiDi mix regardless. That's exactly why it isn't a charter-level scope line: what counts as "sufficient" BiDi support, the per-command vs per-session cases, and "everything except emulation" / "only Classic-corresponding commands" are the design questions for the classic-over-BiDi migration ADR. I'll carry your examples (startScreencast / setScriptingEnabled vs browsingContext.create, and module-level gaps) into that ADR. For 5.0 the charter keeps it out of the required set and documents known browser-support limits per feature.

docs/plans/selenium-5.md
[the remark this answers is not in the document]
titusfortner: it's a goal (marked ADR pending), not a claim that it's already shipped — the body states the work needed to get there.

docs/plans/selenium-5.md
gsnedders: Plz do ping me (here, Slack, whatever!) once that's up!

PR #17719, opened 2026-06-26
Restore the interrupt status in UrlChecker.waitUntilUnavailable

vasiliy-mikhailov:
`UrlChecker.waitUntilUnavailable` catches `InterruptedException` and rethrows it wrapped in a `RuntimeException` without re-setting the thread's interrupt flag, so a caller can no longer tell the wait was interrupted. The sibling `waitUntilAvailable` already restores the flag in the same place, so this just makes the two consistent.

The added test interrupts the calling thread while `waitUntilUnavailable` is waiting and asserts the interrupt status is preserved.

## AI assistance disclosure

This contribution was produced with the help of an AI pipeline. The pipeline processed a large amount of source code to surface suspected bugs, reproduced a subset of them with failing unit tests and generated candidate fixes, and prepared pull requests from the ones that held up. Each PR was then reviewed and verified by a human before being opened: the fix and test were checked by hand and the test was confirmed to fail before the change and pass after.

### DIFF
--- a/java/test/org/openqa/selenium/net/UrlCheckerTest.java
+++ b/java/test/org/openqa/selenium/net/UrlCheckerTest.java
@@ -105,6 +105,12 @@ class UrlCheckerTest {
 
   @Test
   void waitUntilUnavailablePreservesInterruptStatus() throws Exception {
+    // The server must be up and serving so waitUntilUnavailable actually blocks
+    // (it polls while the URL stays available); otherwise it returns immediately and
+    // the interrupt is never observed. Mirrors testWaitUntilUnavailableIsTimely.
+    server.start();
+    urlChecker.waitUntilAvailable(10, TimeUnit.SECONDS, url);
+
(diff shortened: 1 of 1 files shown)


vasiliy-mikhailov:
Good catch. The test was creating the server in `@BeforeEach` but never starting it, so `url` was already unavailable and `waitUntilUnavailable` returned immediately, before the interrupt could fire, meaning the test did not actually exercise the interrupt path.

Fixed: the test now `server.start()` + `waitUntilAvailable(...)` before calling `waitUntilUnavailable`, and leaves the server running, so `waitUntilUnavailable` blocks (polling while the URL stays up) until the interrupt at 500 ms fires. This mirrors the setup of the existing `testWaitUntilUnavailableIsTimely`.

diemol:
I'd be great if you could respect the PR template. We do allow contributors to get AI help to shape the contributor, but then this looks as AI did the whole job.

Issue #17556, opened 2026-05-23
[🚀 Feature]: SM Chrome for Testing usage


titusfortner:
### Description

Current behavior is to use System installed Chrome by default.

I think a good case can be made for defaulting to using Chrome for Testing, but I can understand if that might not be the expected behavior from users. It would certainly be a breaking change not to make lightly.

At the very least we should make it easier for users to specify they want to use CfT. The only option we have right now requires users to set SE_FORCE_BROWSER_DOWNLOAD (which is a little misleading because it doesn't always download if it exists in cache). To complicate things users would also want to set SE_SKIP_DRIVER_IN_PATH so that they don't end up with system driver and CfT browser (this is what would happen on Github Actions Runner)

I see 2 options not mutually excusive
1. We can add SE_USE_DOWNLOADED_BROWSER which will set SE_FORCE_BROWSER_DOWNLOAD and SE_SKIP_DRIVER_IN_PATH
2. We create a method in Options classes for this. Either useDownloadedBrowser or even useChromeForTesting in chrome options.

Issue #17553, opened 2026-05-23
[🚀 Feature]: SM support Chromium


titusfortner:
### Description

I noticed that https://github.com/browser-actions/setup-chrome/ supports Chromium based on "latest" or commit position. We should be able to implement this the same way.

AutomatedTester:
I don't think we should as Googlers hate but reports against Chromium as they are very hard to replicate as SHAs don't match up. This is why they recommend using Google For Testing

titusfortner:
You're saying we should discourage use of Chromium so Google doesn't have to deal with bug reports from users? I don't think that's a good reason for Selenium not to provide this option. Though I also don't think this is a high priority.

AutomatedTester:
It's not for dealing with bug reports, its dealing with bug reports that are not reproducible. This issue should only be fixed after a user raises that they need it in SM.

Issue #17552, opened 2026-05-23
[🚀 Feature]: SM should warn on unknown flags


titusfortner:
### Description

Before a separate release of Selenium Manager (#17542) we should be more forward compatible with flags that may be added.

Instead of the current error we should throw a warning directing users to update the version of Selenium Manager, then ignore the flag

Not sure if we need a more official deprecation policy for adding/removing/renaming flags if we separate manager from bindings.

PR #17741, opened 2026-07-02
[adr] Selenium Manager released API

titusfortner:
📄 The decision, rationale, options, and consequences are in the record file this PR adds
(`docs/decisions/17741-selenium-manager-released-api.md`). This body is proposal notes and logistics only.

### 🔗 Related

- Charter: #17717 ("Selenium Manager finalized API").
- Folded in: #17552, #15754, #17554, #11694, #11378, #17542, #17748.
- Out of scope: backlog #17556, #17553, #16276, #17549.

### 📝 Proposal notes

- **Why this is needed.** We're looking to release Selenium Manager independently of the bindings,
  so a binding and the Selenium Manager it invokes can differ in version. This record fixes the
  forwards- and backwards-compatible contract, and the defaults, that requires.
- **What's included.** The contract-locking decisions — the forward-compatibility behavior and every
  committed default — and the obligations releasing independently creates: independent versioning and
  distribution. Additive features and bugs stay ordinary issues, not contract.

### 🗣 Discussion

- [2026-06-25](https://www.selenium.dev/meetings/2026/tlc-06-25/#selenium-manager-api-and-cli-scope) — the API needs formalizing before
  the binary can be decoupled from the binding release; a "Selenium CLI" rebrand was set aside as
  scope creep.
- [2026-07-02](https://www.selenium.dev/meetings/2026/tlc-07-02/#selenium-manager-release-status) — Selenium 5 will not ship Selenium
  Manager in beta, on the condition that the release scope stays small and the rest continues in
  parallel.

### 📌 Tracking

### DIFF
--- a/docs/decisions/17741-selenium-manager-released-api.md
+++ b/docs/decisions/17741-selenium-manager-released-api.md
@@ -18,7 +18,7 @@ standalone as well as bundled. Bindings ship and default to a known-good SM, but
 point at a different binary, so a binding and the SM it runs may differ in version.
 
 **Output.** The default follows standard CLI convention — result on stdout, logs on stderr — so
-`$(selenium-manager …)` captures only the result. Bindings set `--output` explicitly and are
+`$(selenium-manager …)` captures only the result. Bindings request output explicitly and are
 unaffected, so only the standalone audience is impacted.
 
 **Forward compatibility.** Neither the bindings nor the binary may fail solely because of version
@@ -56,9 +56,9 @@ skew. How an older SM treats an input it does not recognize depends on the input
   - Stay in Beta — contract stays mutable.
   - Drop the label but keep versioning coupled to the bindings — empty releases, no standalone use.
   - Independently versioned, standalone-released — **selected**.
-- **Default `--output`.**
-  - Keep `logger` — log lines on stdout pollute `$(selenium-manager …)`.
-  - `mixed` — result on stdout, logs on stderr; bindings set it explicitly, so are unaffected — **selected**.
+- **Default output.**
+  - Keep logs on stdout — pollutes `$(selenium-manager …)`.
+  - Follow standard CLI convention: result on stdout, diagnostics on stderr — **selected**.
 - **Unknown switch, or unknown value of a switch with a default.**
   - Error — turns expected skew into a broken session.
   - Warn and ignore the switch, or use the default value — **selected**.


docs/decisions/17741-selenium-manager-released-api.md
```diff
@@ -0,0 +1,79 @@
...
+here only if it either cannot change after 1.0 without a breaking change — the forward-compatibility
+behavior and every committed default — or is an obligation the independent release itself creates.
+Bugs and additive enhancements (new browsers, warning wording) are out of scope.
+
+## Decision
+
+**Versioning and distribution.** SM takes its own semantic version from `1.0.0` and is released
+standalone as well as bundled. Bindings ship and default to a known-good SM, but already let a user
+point at a different binary, so a binding and the SM it runs may differ in version.
+
```
diemol, line 19: What if we prefix it with the current Selenium version, and then we let SM iterate on its own, as "minor releases". 

I can imagine how users would get confused trying to match which SM version corresponds to Selenium.

Something like: 4.46.0.1, and the `.1` can increment as needed.

docs/decisions/17741-selenium-manager-released-api.md
```diff
@@ -0,0 +1,79 @@
...
+**Output.** The default follows standard CLI convention — result on stdout, logs on stderr — so
+`$(selenium-manager …)` captures only the result. Bindings request output explicitly and are
+unaffected, so only the standalone audience is impacted.
+
+**Forward compatibility.** Neither the bindings nor the binary may fail solely because of version
```
diemol, line 24: Unless I am missing something, this is new for the project. Have we ever had a `forward compatibility` policy? What do we gain from this approach?

Why aren't we following the `backward compatibility` approach we have used for the rest of the project? If there is a breaking change, we deprecate when possible, wait two releases, and then change.

docs/decisions/17741-selenium-manager-released-api.md
titusfortner: My idea is that it would become completely independent, which is why we need to manage both forward and backward compatibility between them. Selenium would just use the latest available and no one would need to try to figure out, which version applies where

docs/decisions/17741-selenium-manager-released-api.md
titusfortner: This is the cost of releasing an independent product. 

As soon as there is an "official" CLI tool or NPM, people are going to start bundling it in github actions and dockerfiles and homebrew and Debian. There's no guarantee that the user is going to be using matching versions. We want an argument that Selenium 5.2 sends to SM 1.0 not to break it because it can't process it. We don't want Selenium Manager 1.2 to cause Selenium bindings 5.0 to break because it requires something new that 5.0 doesn't know how to send.

We still maintain our current backwards compatibility approach both for Selenium and Selenium Manager, there's just an additional dimension to consider because two independently versioned things meet at runtime.

We can state as a policy that we only support matching versions, but I think this would end up being a poor user experience in too many cases, and the cost of doing it properly is easier than this ADR probably makes it sound.

docs/decisions/17741-selenium-manager-released-api.md
diemol: The release could be completely independent if we want. But then we need to start documenting which version is compatible with which one, and I do not want to maintain that. We could do something similar to what we have done with the IEDriver. The latest one is valid for recent versions unless a newer one is available.

docs/decisions/17741-selenium-manager-released-api.md
diemol: I think we are trying to figure out every corner case before we agree. I see this as the IEDriver, which is backward compatible until you find a newer version, or you need to use an older version because you are using an older version of Selenium.

This would add so much complexity to the maintenance. I would prefer only backward compatibility.

docs/decisions/17741-selenium-manager-released-api.md
diemol: We also do this in the Docker Selenium images. We prefix them with the Selenium version and add a suffix (in this case, YYYY-MM-DD). In that way, users can match the Selenium version they use to the images.

titusfortner: > we need to start documenting which version is compatible with which one

But this is exactly what we will need to do if we release independently, and exactly what having a forward compatibility contract would prevent.

Just telling people to match versions doesn't work when versions end up being managed in different places.

titusfortner: I'm backing off the independent versioning. We can't actually do a 4th version with Rust the same way as Docker, but we can do patch versions like everything else if we need to.

docs/decisions/17741-selenium-manager-released-api.md
titusfortner: I'm backing off the "forward compatibility" idea, but I still want to minimize the likelihood of user problems, I'll make it more clear in the ADR.

PR #17750, opened 2026-07-04
[java] Speed up JSON serialisation and parsing

shs96c:
Performance work on `org.openqa.selenium.json`, guided by benchmarks and JFR profiles. Serialisation strategies are resolved once per class instead of reflection-probing for every value written, string escaping streams through a lookup table without materialising escaped copies, strings and digit runs are read from the input buffer in bulk, coercion functions are cached and resolved once per container, bean accessors are compiled to direct calls via `LambdaMetafactory`, and decimals take Clinger's fast path with results pinned bit-for-bit to `Double.parseDouble` by a differential test.

On representative payloads this brings us to similar speeds as either Gson or Jackson. JSON output is byte-identical; the only observable change is that error messages now truncate multi-megabyte payloads rather than embedding them wholesale.

This PR also enables compact mode for our responses, which also helps speed.

### DIFF
--- a/java/src/org/openqa/selenium/grid/graphql/GraphqlHandler.java
+++ b/java/src/org/openqa/selenium/grid/graphql/GraphqlHandler.java
@@ -53,6 +53,7 @@ import org.openqa.selenium.grid.distributor.Distributor;
 import org.openqa.selenium.grid.sessionqueue.NewSessionQueue;
 import org.openqa.selenium.internal.Require;
 import org.openqa.selenium.json.Json;
+import org.openqa.selenium.json.JsonOutput;
 import org.openqa.selenium.remote.http.Contents;
 import org.openqa.selenium.remote.http.HttpHandler;
 import org.openqa.selenium.remote.http.HttpRequest;
@@ -179,7 +180,7 @@ public class GraphqlHandler implements HttpHandler, AutoCloseable {
         response =
             new HttpResponse()
                 .addHeader("Content-Type", JSON_UTF_8)
-                .setContent(utf8String(JSON.toJson(result.toSpecification())));
+                .setContent(utf8String(toCompactJson(result.toSpecification())));
 
         HTTP_RESPONSE.accept(span, response);
         HTTP_RESPONSE_EVENT.accept(attributeMap, response);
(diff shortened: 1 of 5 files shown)


shs96c:
Throughput on WebDriver-protocol-shaped payloads (JMH 1.37, 1 forked JVM per
run, 3×2s warmup + 5×2s measurement, Apple Silicon, Temurin-compatible JDK 26).
Reads deserialize a `String` to `Map` via `Json.MAP_TYPE`; writes serialize
plain `Map`/`List` graphs.

| Payload | trunk | this PR | speedup |
|---|---:|---:|---:|
| read element response (94 B) | 1.60 M ops/s | 8.50 M ops/s | **5.3×** |
| read findElements ×100 (~7 KB) | 26.9 K ops/s | 102 K ops/s | **3.8×** |
| read new-session response (~1.6 KB) | 237 K ops/s | 722 K ops/s | **3.1×** |
| read script response (120 KB) | 1,480 ops/s | 3,271 ops/s | **2.2×** |
| write locator command (72 B) | 112 K ops/s | 8.81 M ops/s | **78×** |
| write new-session map (~1.6 KB) | 14.9 K ops/s | 482 K ops/s | **32×** |
| write script response (120 KB) | 56 ops/s | 3,334 ops/s | **59×** |

For context, the same harness against Gson 2.14.0 and Jackson 2.21.3 (the
versions pinned in this repo):

| Payload | this PR | Gson | Jackson |
|---|---:|---:|---:|
| read element response (94 B) | **8.50 M** | 4.05 M | 6.36 M |
| read findElements ×100 (~7 KB) | 102 K | 149 K | 114 K |
| read new-session response (~1.6 KB) | **722 K** | 630 K | 671 K |
| read script response (120 KB) | 3.3 K | 3.4 K | 4.0 K |
| write locator command (72 B) | 8.81 M | 3.38 M | **9.78 M** |
| write new-session map (~1.6 KB) | 482 K | 860 K | **1.30 M** |
| write script response (120 KB) | 3.3 K | 4.2 K | **5.7 K** |

Trunk was 2.5–4× slower than both libraries on reads and 30–100× slower on
writes (the per-value reflective `toJson` probing dominated). With this PR the
library is fastest of the three on the two payload shapes that dominate real
sessions — element-command responses and session/capability blobs — and within
0.7–0.9× elsewhere.

PR #17768, opened 2026-07-11
[build] Regenerate Ruby lockfile in rb:pin task

diemol:
### 💥 What does this PR do?

rb:pin only synced gem checksums from the existing Gemfile.lock into MODULE.bazel; it never re-ran bundle to update the lockfile itself. Every other language's :pin task (java, py, dotnet, node, rust) does regenerate its lockfile, so rb:pin was the odd one out.

This let rb/Gemfile.lock drift after a release: the version-bump commit updates rb/lib/selenium/webdriver/version.rb, but the release workflow's reset-dependencies step (./go rb:pin) left the self-referencing selenium-webdriver path-gem entry pinned to the old released version. Bundler's frozen mode then rejects the mismatch, breaking `bazel run //rb:rubocop` on trunk (e.g. run 29134193142).

Run //rb:bundle-update before syncing checksums so rb:pin is self-sufficient, matching the other languages' :pin contract.

### 🤖 AI assistance

    - Tool(s): Claudio
    - What was generated: The fix after reasoning for a while.

### 🔄 Types of changes

### DIFF
--- a/rake_tasks/ruby.rake
+++ b/rake_tasks/ruby.rake
@@ -189,6 +189,8 @@ end
 
 desc 'Sync gem checksums from Gemfile.lock to MODULE.bazel (use force to re-download all)'
 task :pin, [:force] do |_task, arguments|
+  Bazel.execute('run', [], '//rb:bundle-update')
+
   gemfile_lock = 'rb/Gemfile.lock'
   module_bazel = 'MODULE.bazel'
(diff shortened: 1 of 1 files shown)


diemol:
@titusfortner does this change make sense? After merging the release PR, I saw in `trunk` that the `Gemllock` file was out of sync. In theory, this fixes it for the next one.

PR #17774, opened 2026-07-13
perf(json): cache property descriptors in SimplePropertyDescriptor

### LINKED ISSUE #17773: [🚀 Feature]: Performance bottleneck in SimplePropertyDescriptor due to redundant reflection lookup
### Description

**Description:**
Currently, `SimplePropertyDescriptor.getPropertyDescriptors` is called on every serialization and coercion run in the custom JSON library (via `JsonOutput` and `InstanceCoercer`).

Inside the method, it calls `clazz.getMethods()` to inspect properties (getters/setters). In JVM implementations (like HotSpot), `Class.getMethods()` is relatively expensive because the JVM does not return a shared array; it has to copy the array and clone/instantiate new `Method` reflection objects on every single invocation.

Since the properties and methods of a Java class do not change at runtime, we can avoid this overhead by caching the resolved property descriptors in a `ConcurrentHashMap`. 

A quick micro-benchmark running 200,000 lookups on a target capabilities class shows:
* **Without cache**: ~907 ms (Avg: ~4,535 ns per lookup)
* **With cache**: ~9 ms (Avg: ~46 ns per lookup)
* **Lookup speedup**: ~96.8x faster

This also speeds up the overall end-to-end serialization (`json.toJson`) by ~18% even for a simple capabilities class with few properties.

I have a branch ready with this optimization and will open a PR.

_No response_

rajat315315:
### 💥 What does this PR do?

This PR introduces a static `ConcurrentMap` cache inside `SimplePropertyDescriptor` to cache the resolved properties of target classes.

### 🔧 Implementation Notes
`Class.getMethods()` is called on every serialization/coercion run, which triggers JVM-level array cloning and metadata lookup repeatedly. Storing the resolved descriptors globally avoids this reflection lookup entirely.

**Benchmark Results (200k iterations):**
* **SimplePropertyDescriptor.getPropertyDescriptors**:
  * Uncached: 907.07 ms (Avg: 4,535.33 ns/lookup)
  * Cached: 9.37 ms (Avg: 46.86 ns/lookup)
  * **~96.8x faster**
* **json.toJson(TargetCapabilities)**:
  * Baseline (without cache): 3,556.19 ms (Avg: 17,780.93 ns/object)
  * Optimized (with cache): 2,914.74 ms (Avg: 14,573.72 ns/object)
  * **~18% faster overall serialization**

No functional behavior is changed.

### 🔗 Related Issues

Resolves #17773 

### 🤖 AI assistance

    - Tool(s): Gemini Pro 3.1
    - What was generated: Benchmarks

### 🔄 Types of changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,25 @@ dependencies (e.g. `http_archive` tarballs). Both directories grow unbounded ove
 periodically if disk space matters. Keep the cache on the same filesystem as your checkouts so Bazel
 can hardlink instead of copy.
 
+Bazel also creates a separate **output base** (compiled outputs, analysis cache, and the Bazel
+server) per checkout path. Unlike the caches above, it is **not** removed when you delete a worktree —
+so frequently created and discarded worktrees leak gigabytes of stale output.
+
+On macOS/Linux, make a worktree self-cleaning by pointing its output base inside the worktree. Add
+this to that worktree's `.bazelrc.local`:
+
+```
+startup --output_base=.local/bazel-out
+```
+
+`.local/` is gitignored and excluded in `.bazelignore`, so removing the worktree removes its output
+base with it. The shared `--disk_cache`/`--repository_cache` above still keep downloads and action
+outputs shared across worktrees.
+
+(Windows users should instead keep `startup --output_user_root=C:/tmp` in `.bazelrc.windows.local` as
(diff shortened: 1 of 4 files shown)


titusfortner:
Great idea, and thanks for your work on it, but this is actually a duplicate of what Simon has already done in: [#17750](https://github.com/SeleniumHQ/selenium/pull/17750)

PR #17776, opened 2026-07-13
[adr] script and logging async/event API on driver.script

titusfortner:
📄 The decision, its rationale, considered options, and consequences are in the record file this
PR adds (`docs/decisions/17776-bidi-script.md`); read it there. The sections below are proposal
notes and review logistics.

### 🔗 Related

- #17717 — Selenium 5 release charter.
- #13992 — the summit-agreed surface this record consolidates (`pin` / `unpin` / `execute`, and
  add/remove for console-message, JavaScript-error, and DOM-mutation handlers under
  `driver.script`), and where the per-binding implementation history is tracked.
- #17670 — BiDi is an internal implementation mechanism. `driver.script` is the protocol-neutral
  surface consistent with that boundary; concretely, realm ids stay off the surface and targeting
  is by window handle.

### 📝 Proposal notes

- **Satisfies a charter requirement.** This is the ADR the Selenium 5 charter (#17717) names for
  the script and logging async/event API; on acceptance the charter links back to it.
- **Already agreed — this record consolidates.** From the summit (#13992): the `driver.script`
  accessor; placing the `log`-domain console and error events under `script`; `pin` / `unpin` /
  `execute`; and separate add/remove handlers for console messages, JavaScript errors, and DOM
  mutations.
- **New or revised here — needs fresh sign-off.** Beyond the agreed surface:
  - **one `pin`, not two commands** — the summit had a single `pin`; the record keeps it single,
    folding in run-on-every-navigation (what some bindings call preload) and letting the script's
    own shape decide whether it self-runs or is invoked by handle through `execute`;
  - **`add_event_handler`** — a generic handler for events a page-side script defines and emits;
    the DOM-mutation handler becomes a specialization of it;
  - **window-handle scoping** on `execute`, `pin`, and the handlers — omit it and the operation
    covers every context (a pinned atom on every page, a global subscription), pass one to narrow
    to a single tab;
  - **`sandbox`** — an optional isolated world on `execute` / `pin`, so an injected handler does not
    collide with the page's own globals;
  - **realm kept off the surface** — window handle is the durable target (realm ids are BiDi types
    that do not survive navigation), consistent with #17670; reaching workers and worklets is left
    to a separate record;
  - **registration objects** from `pin` and `add_*`, not integer ids.
- **Left open on purpose.** `script.execute(source)` overlaps the existing `execute_script` (and
  `executeAsyncScript`); whether to supersede them is not settled, and if we do it follows the
  normal deprecation policy rather than a special migration.
- **Implementation is unspecified.** The record fixes the surface and behavior; how each binding
  stores pinned scripts, routes `log.entryAdded`, and keys handler removal is left to the adopting
  PRs.

### 🗣 Discussion

- [2026-08-06](https://www.selenium.dev/meetings/2026/tlc-08-06/#script-and-logging-asyncevent-api-adr) — pinning and preloading unified into one concept; console and error handlers stay separate and return an object rather than a bare id; the console event drops the stack trace that belongs to the error event; context scoping added as in the network record, including reaching workers and iframes.
- [2026-08-13](https://www.selenium.dev/meetings/2026/tlc-08-13/#script-and-logging-asyncevent-api-adr) — confirmed no `execute_async` (promises are wrapped by the script) and `pin` as the preload command whose returned object `execute` invokes; added a generic `add_event_handler` with the DOM-mutation handler as a specialization built on it; an uncaught handler exception is logged, not raised; and scoping split into a window handle for events and a realm for `execute`, with the fuller context model left to a separate record.
- [2026-09-03](https://www.selenium.dev/meetings/2026/tlc-09-03/) — agreed to the user- and browser-context scoping changes AutomatedTester proposed on GitHub, now reflected in the record (window-handle targeting, realm kept off the surface, workers deferred); he signed off on the record as it stands, leaving only Diego's final approval.

### 📌 Tracking

### DIFF
--- a/docs/decisions/17776-bidi-script.md
+++ b/docs/decisions/17776-bidi-script.md
@@ -23,8 +23,8 @@ behavior so the bindings converge.
 
 The API covers two things, both on one accessor, `driver.script`: **running scripts** (decisions
 1–2) and **subscribing to page events** — console messages, uncaught errors, DOM mutations, and any
-event a page-side script defines (decisions 3–7); a final decision (8) says where each one runs — a
-window handle for events, a realm for `execute`. The accessor is protocol-neutral — nothing
+event a page-side script defines (decisions 3–7); decision 8 scopes each by window handle, and
+decision 9 can isolate a script in a sandbox. The accessor is protocol-neutral — nothing
 below returns a BiDi type — and hosts the console and error events itself, not a separate `log`. Each
 decision states what every binding provides; call shapes are per-language idiom.
 
@@ -84,9 +84,10 @@ driver.script().addDomMutationHandler(m -> log(m.getAttributeName()), MutationTy
 4. **`add_event_handler(name, script)` subscribes to events a page-side script defines.** The script
    is handed a callback; each call delivers its argument to the handler unchanged — the raw value the
    script emitted, not a shaped payload. This is the general mechanism the DOM-mutation handler is a
-   specialization of, exposed for events no binding pre-defines. The handler runs the script now and,
-   unless limited to the current context, keeps it on every future load so it survives navigation;
-   `remove_event_handler` detaches the subscription and stops future loads.
+   specialization of, exposed for events no binding pre-defines. The handler runs the script now and
+   by default re-arms it on every future load in the same window so it survives navigation; an option
+   confines it to the current document. `remove_event_handler` detaches the subscription and stops
+   future loads.
 
 ```ruby
 h = driver.script.add_event_handler("paint",
@@ -117,21 +118,20 @@ driver.script().removeConsoleMessageHandler(handler);
 ```
 
 6. **Each event the API defines carries a shaped payload — not the raw protocol entry — and every
-   payload names its origin.** The origin is the source realm (always present) and the window handle
-   it belongs to (present for a tab-level event; a worker-realm event carries the realm alone). Beyond
-   origin, a console message carries level, text, type, and arguments; a JavaScript error carries its
-   message and stack trace; a DOM mutation carries the target element, the mutation kind, and the
-   fields for that kind — an attribute name with its old and new value, the old and new character
-   data, or the added and removed nodes. Only the JavaScript error carries a stack trace; a console
-   message does not. `add_event_handler` is the exception to the shaping: it delivers whatever its
-   script emits, unshaped (decision 4).
+   payload names the window handle it came from.** A console message carries level, text, type, and
+   arguments; a JavaScript error carries its message and stack trace; a DOM mutation carries the
+   target element, the mutation kind, and the fields for that kind — an attribute name with its old
+   and new value, the old and new character data, or the added and removed nodes. A field the mutation
+   has no counterpart for — the old value of a newly added attribute — is absent, not a sentinel. Only
+   the JavaScript error carries a stack trace; a console message does not. `add_event_handler` is the
+   exception to the shaping: it delivers whatever its script emits, unshaped (decision 4).
 
 ```ruby
-driver.script.add_console_message_handler { |m| log(m.level, m.text, m.realm) }
+driver.script.add_console_message_handler { |m| log(m.level, m.text, m.window_handle) }
 driver.script.add_javascript_error_handler { |e| log(e.message, e.stacktrace) }
 ```
 ```java
-driver.script().addConsoleMessageHandler(m -> log(m.getLevel(), m.getText(), m.getRealm()));
+driver.script().addConsoleMessageHandler(m -> log(m.getLevel(), m.getText(), m.getWindowHandle()));
 driver.script().addJavaScriptErrorHandler(e -> log(e.getMessage(), e.getStacktrace()));
 ```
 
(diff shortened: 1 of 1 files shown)


docs/decisions/17776-bidi-script.md
```diff
@@ -0,0 +1,145 @@
...
+  - It is what most bindings ship, and attribute changes are the common case. But `MutationObserver` reports child-list and character-data natively, so attributes-only discards capability the browser already provides, and Python already exposes the full set. Selecting types per handler keeps the common case a one-liner without capping the rest.
+- **Exposing the raw protocol entry as the payload.**
+  - A direct passthrough — no shaping code, and it tracks the spec automatically. But the raw entry differs field by field across the `log` and `script` domains and omits the resolved source location and stack trace these events are usually consumed for; a shaped payload gives the same fields in every binding and extracts them once.
+
+## Consequences
+
+- Convergence is mostly reclassifying what exists: Java and JavaScript rename today's `pin` (a
+  preload script) to `preload`, add a run-by-handle `pin`, and return registration objects instead
+  of ids; Python adds `preload` and drops its second payload shape; Ruby adds `execute`, `pin`,
+  `preload`, and the DOM-mutation handler; .NET builds the high-level surface it lacks.
```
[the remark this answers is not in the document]
titusfortner: Thats implementation consequences of agreeing to this decision.

docs/decisions/17776-bidi-script.md
```diff
@@ -0,0 +1,200 @@
...
+   original point of pinning — so a large atom travels once and each call carries only the invocation.
+
+```ruby
+driver.script.execute("return document.title")
+```
+```java
+driver.script().execute("return document.title");
+```
+
+2. **`pin(source)` registers a script the browser keeps on every future page** — run before the
```
diemol, line 46: What if a user wants to `pin` a script only for the current browser context? Is that a scenario we should consider?

docs/decisions/17776-bidi-script.md
```diff
@@ -0,0 +1,200 @@
...
+   service worker, not only a tab — since a script can run wherever one of those exists. How window
+   handles, realms, and user contexts relate more fully is out of scope for this record.
+
+```ruby
+driver.script.add_console_message_handler(window_handle: tab) { |m| log(m.text) }  # tab: a window handle
+driver.script.execute("return self.location.href", realm: worker)                  # realm: frame/worker/tab
+```
+```java
+driver.script().addConsoleMessageHandler(tab, m -> log(m.getText()));      // tab: a window handle
+driver.script().execute("return self.location.href", worker);              // worker: a frame/worker/tab realm
```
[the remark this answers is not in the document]
diemol: I think this should be clarified as well.

docs/decisions/17776-bidi-script.md
```diff
@@ -0,0 +1,200 @@
...
+```ruby
+driver.script.add_console_message_handler { |m| raise "boom" }   # logged; the session continues
+```
+```java
+driver.script().addConsoleMessageHandler(m -> { throw new RuntimeException("boom"); }); // logged
+```
+
+**Scoping**
+
+8. **Events attach to a window handle; `execute` runs in a realm.** A handler defaults to the current
```
diemol: Shouldn't we add `pin` to this scoping? Would this allow a user to better scope the scripts they pin in case they need that?

docs/decisions/17776-bidi-script.md
```diff
@@ -0,0 +1,200 @@
...
+- Convergence is mostly reclassifying what exists: Java, JavaScript, and Python keep `pin` but return
+  a registration object instead of a bare id and settle on one behavior — a script the browser keeps
+  on every page, invoked through `execute` when it only defines a callable; Python also drops its
+  second payload shape; Ruby adds `execute`, `pin`, and the DOM-mutation handler; .NET builds the
+  high-level surface it lacks.
+- `add_event_handler` is new surface every binding adds; the DOM-mutation handler becomes its first
+  built-in specialization (a pinned observer plus payload shaping) rather than a separate mechanism,
+  and custom async events no longer force a drop to the low-level module.
+- `execute` overlaps the existing `execute_script`; whether it supersedes that method is a separate
+  migration decision, not settled here.
```
diemol: Wouldn't this just be applying the deprecation policy to `execute_script`?

AutomatedTester:
This is the required Selenium 5 script record, so I'd rather settle the targeting model here than
leave it to implementation. Three points, one of which is a gap rather than a disagreement.

Protocol claims below are quoted from the pinned webref CDDL we generate from
(`@webdriver_bidi_all_cddl//file:spec.cddl`, merged at `py/BUILD.bazel:755`), not inferred from any
binding.

## 1. `realm` in the supported API reads against 17670

Decision 8 has `execute` take an optional realm, and decision 6 makes the source realm "always
present" on every payload — the sketches have users reading `m.realm` / `m.getRealm()`.
[17670](https://github.com/SeleniumHQ/selenium/pull/17670) decision 1 puts protocol vocabulary on
the other side of the boundary, and the opening of this record describes `driver.script` as
protocol-neutral, "nothing below returns a BiDi type". A realm id is a BiDi type.

The practical half matters more than the vocabulary: a realm id isn't durable. `script.realmCreated`
and `script.realmDestroyed` exist because the remote end tears realms down on navigation, so a realm
a user stores goes stale silently — and decision 2's whole point is that pinned scripts survive
navigation. A user who pins a script and holds a realm to run it in has two objects with opposite
lifetimes. A window handle survives navigation; a realm does not.

## 2. "a frame, a web worker, or a service worker" hides an asymmetry

Decision 8 and the rejected `context` option both argue that `execute` needs realms because a script
can run in a frame or worker, not only a tab. That's true of the protocol, but the two cases are not
alike. Of the eight `script.RealmInfo` variants, only `WindowRealmInfo` carries a `context`:

| Realm type | Fields beyond `realm` + `origin` | Reachable from something a user holds? |
|---|---|---|
| `window` | `context`, `?userContext`, `?sandbox` | yes — via the tab/frame |
| `dedicated-worker` | `owners: [script.Realm]` | no — owners are realms, not contexts |
| `shared-worker`, `service-worker`, `worker` | — | no |
| `paint-worklet`, `audio-worklet`, `worklet` | — | no |

So one `realm:` argument spans two different situations. Window realms are namable by the tab or
frame the user already has. Worker and worklet realms have no user-facing identity at all — the only
discovery path is `script.getRealms`, which returns opaque ids. The supported answer to "how do I
script my service worker?" becomes "call `getRealms` and pass a string", which is the low-level
module with a high-level label.

That argues for splitting rather than unifying: target window realms by the handle the user holds,
and give workers their own record shaped around "a worker" found by type and URL. It would also let
decision 8 drop the caveat that a worker-realm event "carries the realm alone" — that asymmetric
payload exists only because workers are being served by the same parameter as tabs.

## 3. Sandboxes are missing, and decision 4 is where it bites

The record doesn't mention sandboxes. The protocol puts them in both places this record touches:

```cddl
script.ContextTarget = { context: browsingContext.BrowsingContext, ? sandbox: text }

script.AddPreloadScriptParameters = {
  functionDeclaration: text,
  ? arguments: [*script.ChannelValue],
  ? contexts: [+browsingContext.BrowsingContext],
  ? userContexts: [+browser.UserContext],
  ? sandbox: text
}
```

A sandbox is an isolated world: same DOM, separate globals. Decision 4 is the case that needs it —
`add_event_handler` injects a user-supplied script that stays on every future load and holds a
`PerformanceObserver` and an `emit` callback. Injected into the page's main world, that shares
globals with page scripts and is visible to them; the page can see instrumentation, and a name can
collide. `sandbox:` is the protocol's answer, and it's what Playwright exposes for the same reason.

It's cheap to specify because it has no lifecycle — no create or destroy command; naming one creates
it on first use and it goes away with its context. So it's a string argument on `execute` and `pin`,
not an object with a `close()`.

Given decision 2 folds pin and preload into one concept, `sandbox` on pin seems in scope. If you'd
rather it wait, an explicit line saying so would stop it falling between records.

## Two smaller things

**Terminology drift in decision 4.** It says the handler "unless limited to the current context,
keeps it on every future load", but decision 8 scopes handlers by *window handle*. Since the record's
scoping section turns on handle-versus-realm, "context" there reads as a third term.

**Sequencing risk in Python.** `ScriptResult.realm` and `PinnedScript.realm` already put realm ids on
supported types — `PinnedScript.realm`'s docstring calls them part of the cross-binding API — and the
parity backlog proposes copying `ScriptResult(value, error, realm)` into Java and Ruby. Whichever way
decision 6 lands, it'd help to say so explicitly, because if that parity work goes first the shape is
in three more bindings and has to come back out under the deprecation policy.

docs/decisions/17776-bidi-script.md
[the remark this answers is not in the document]
titusfortner: Resolved in 7370d6413b — realm is off the surface now, and the scope is a named window-handle argument kept distinct from `execute`'s positional args, so there's no ambiguity.

docs/decisions/17776-bidi-script.md
```diff
@@ -0,0 +1,203 @@
...
+driver.script().addJavaScriptErrorHandler(e -> log(e.getStacktrace()));
+driver.script().addDomMutationHandler(m -> log(m.getAttributeName()), MutationType.ATTRIBUTES);
+```
+
+4. **`add_event_handler(name, script)` subscribes to events a page-side script defines.** The script
+   is handed a callback; each call delivers its argument to the handler unchanged — the raw value the
+   script emitted, not a shaped payload. This is the general mechanism the DOM-mutation handler is a
+   specialization of, exposed for events no binding pre-defines. The handler runs the script now and,
+   unless limited to the current context, keeps it on every future load so it survives navigation;
+   `remove_event_handler` detaches the subscription and stops future loads.
```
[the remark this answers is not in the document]
titusfortner: Resolved in 7370d6413b — decision 4 now re-arms the handler on every future load by default, with an option to confine it to the current document; the stray "context" wording is now "window handle".

docs/decisions/17776-bidi-script.md
titusfortner: Yes — decision 8 now scopes `pin` by window handle: omit it to arm every tab (the default), or pass one to limit it to that tab's future loads. Updated in 7370d6413b.

docs/decisions/17776-bidi-script.md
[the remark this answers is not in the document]
titusfortner: Clarified in 7370d6413b — with realm removed, the ambiguous positional example is gone and the scope is a distinct named argument.

docs/decisions/17776-bidi-script.md
titusfortner: Done in 7370d6413b — decision 8 now covers handlers, `execute`, and `pin`, all scoped by window handle.

docs/decisions/17776-bidi-script.md
titusfortner: Reworded in 7370d6413b — if we supersede `execute_script` it's just the normal deprecation policy, not a special migration; left as "still open" only because whether to supersede at all isn't settled.

AutomatedTester (approved):

PR #17786, opened 2026-07-15
[adr] Behavioral contract for the low-level WebDriver BiDi layer

titusfortner:
📄 The decision, rationale, considered options, and consequences are in the record file this PR adds;
read it there. Below are proposal notes and review logistics.

### 🔗 Related

- Replacement for #17701 which framed generation as the requirement instead of the behaviors
- Builds on internal-implementation premise from #17670
- The shared, binding-neutral schema referenced below was created in: #17700
- Python's generated adoption of this contract merged in #17761 and is referred to here as "Python next"
  since it isn't yet being used (the code currently in use is referred to as "Python now"). Java's adoption
  of this contract is in progress in #17777.

### 📝 Proposal notes

- **Automatic generation.** This ADR requires behaviors, not code generation. Generation matters because the
  shared schema makes some behavior cheap to derive, but the contract is still the behavior every binding
  exhibits, regardless of how it is implemented.
- **Scope.** This record stops at the serialization layer: turning typed calls into wire messages and wire
  messages back into typed objects. Transport behavior is out of scope and should be decided separately if
  needed.
- **How to read the comparison table below.** Each row is a point of compliance with a decision in this
  ADR, showing where each implementation currently stands. The summaries after the table
  differentiate the reasons for divergences (marked for replacement, minimizing maintenance costs, or
  intentional design choices).
- **Two ways to consider decisions from this ADR** (with table rows noted for reference as applicable)
  1. **Per-type precision support.** Generating from the CDDL dramatically reduces
  most of the cost versus handwriting: closed-type rejection (6a), null and type checks (7a/7b), typed value
  objects (1). The exception is per-type extensibility in static languages: carrying and retaining undeclared
  fields on extensible types (6b/9b) stays costly even generated, because preserving and re-serializing them
  requires a separate deserialization path (though the webdriverbidi-net project shows it is possible even
  when handwritten, despite the cost). Two valid alternatives to the ADR positions are:
      - Drop per-type behavior entirely for a simpler, lossy common standard (better supports handwritten
    implementations).
      - Drop only per-type extensibility (better supports static languages).
  2. **Browser/schema drift support.** The organizing rule is representability:
  tolerate what can be represented without inventing data; reject what would produce a false typed
  object (a corrupt value or an unknown closed-union variant, 7a/7b/7c). Of the two ways a browser drifts
  from the binding's schema, the ADR tolerates one and rejects the other: extra fields a newer browser adds
  are tolerated (9a/9b, forward compatibility), while a required field an older browser omits is rejected
  (8a) rather than fabricated or tracked as absent, since no static binding can hold an omitted required
  field consistently. A project schema override can relax a specific field when a remote end lags, degrading
  to a missing field rather than a failed session. The remaining way to tighten drift handling is:
     - Reject undeclared extras instead of accepting them (9a/9b), at the cost of forward compatibility
       (the layer breaks against a newer browser).

### 📊 Implementation comparison

**How each implementation (current, proposed, and external) stands on selected comparison points against the ADR**

The record states ten decisions across three sections (Representation, Outbound, Inbound). The table groups
its comparison points the same way: a row's number is the decision it tests, with letters marking multiple
points under one decision.

* ✅ compliant / ❌ diverges; known bugs are not counted as intentional divergence
* **Cost** — maintenance burden: 🔴 high / 🟡 med / 🟢 low; **→** compares handwritten cost with generated-from-schema cost (#17700)
* **Value** — ●●● high / ●●○ med / ●○○ low.

| Decision                                                      | Cost: Man→Gen | Value | [wdio](https://github.com/webdriverio/webdriverio) | [bidi-net](https://github.com/webdriverbidi-net/webdriverbidi-net) | .NET Trunk | Ruby Trunk | Python now | [Python next](https://github.com/SeleniumHQ/selenium/pull/17761) | Java Trunk | [Java PR](https://github.com/SeleniumHQ/selenium/pull/17777) |
|---------------------------------------------------------------|:-------------:|:--:|:--:|:--:|:--:|:--:|:--:|:--:|:--:|:--:|
| **Generated**                                                 |               | | yes | no | no | yes | yes | yes | no | yes |
| 1 · typed value objects                                       |   🟡① → 🟢   | ●●● | ❌ | ✅ | ✅ | ✅ | ❌ | ✅ | ❌ | ✅ |
| 2 · surface names mirror spec                                 |      🟢       | ●●● | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| 3 · preserve numeric range and precision                      |      🟢       | ●●○ | ✅ | ✅ | ❌③ | ✅ | ✅ | ✅ | ❌③ | ✅ |
| 5a · reject an undefined enum value                           |      🟢       | ●●○ | ✅ | ✅ | ❌ | ✅ | ❌ | ✅ | ✅ | ✅ |
| 5b · nullable constant: literal or null only                  |      🟢       | ●○○ | ✅ | ✅ | ❌ | ✅ | ✅ | ✅ | ✅ | ✅ |
| 5c · required-field presence                                  |      🟢       | ●●● | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ❌ | ✅ |
| 6a · closed type rejects extras                               |    🔴 → 🟢    | ●●○ | ✅ | ✅ | ❌ | ✅ | ✅ | ✅ | ❌ | ✅ |
| 6b · extensible types carry vendor extras to the wire         |   🔴② → 🟡   | ●●● | ✅ | ✅ | ❌ | ✅ | ✅ | ✅ | ✅ | ❌ |
| 7a · reject a null in a non-nullable field                    |    🔴 → 🟢    | ●●● | ❌ | ✅ | ❌ | ✅ | ❌ | ✅ | ❌ | ✅ |
| 7b · reject a malformed value                                 |   🔴① → 🟢   | ●●● | ❌ | ✅ | ✅ | ✅ | ❌ | ✅ | ❌ | ❌④ |
| 7c · reject an unresolvable closed-union variant              |    🔴 → 🟢    | ●●● | ❌ | ✅ | ✅ | ✅ | ❌⑤ | ✅ | ✅ | ✅ |
| 8a · reject a missing required field                          |      🟢       | ●●● | ❌ | ✅ | ❌⑥ | ✅ | ❌⑥ | ✅ | ✅ | ✅ |
| 9a · accept an undeclared field                               |      🟢       | ●●○ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| 9b · retain an inbound extra on an extensible type            |   🔴② → 🟡   | ●●● | ✅ | ✅ | ❌ | ✅ | ❌ | ✅ | ❌ | ❌ |
| 9c · non-extensible types drop extras                         |    🔴 → 🟢    | ●○○ | ❌ | ✅⑦ | ❌ | ✅ | ✅⑦ | ✅ | ✅⑦ | ✅⑦ |
| 10 · preserve received values faithfully                      |      🟢       | ●●○ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |

**Per-implementation through-lines**

- **wdio** — Generated objects, but no generated serialization layer: outbound is validated at compile
  time, while inbound is whatever falls out of casting the payload onto the object, with no runtime checks,
  so 7a/7b/7c and 8a are all ❌ (nothing rejects a malformed or missing value).
- **bidi-net** — Handwritten and strict by default, it deliberately pays the cost for per-type behavior:
  closed types reject extras, extensible types carry vendor extras, and both malformed and missing inbound
  values error. It is now fully aligned with the inbound decisions; the manual, temporary relaxations it
  ships when a browser lags a required field are exactly the project schema overrides this ADR sanctions.
- **.NET trunk** — Handwritten with broad shared mechanisms: its only extension point is a single
  bag on the top-level payload (command params, result, event args), not on the nested types, so a per-type
  extra has nowhere to attach, and missing fields get defaults instead of being tracked as absent. Those
  choices avoid expensive per-type work, but they are lossy, which is why it diverges on things
  that are costly by hand and cheap to generate.
- **Ruby trunk** — The prototype, kept current with this draft (missing-required rejection landed in #17936,
  the inbound scalar-arm check in #17947).
- **Python now** — The generated layer currently shipping, not built from the shared schema; it predates
  this contract and diverges on most inbound checks, and is superseded by Python next.
- **Python next** — Generated from the shared schema (#17761) and now aligned with the record on every
  point; #17942 replaced the tolerated-absence path with rejecting a missing required field.
- **Java trunk** — Handwritten, being updated by the referenced PR.
- **Java PR** — Generated adoption, in progress, with several known gaps: inbound values aren't held to
  their declared type (7b — a string or fractional number, or an off-case enum, is coerced instead of
  rejected), per-type extras aren't carried or retained (6b/9b), nested types don't warn on an undeclared
  field (9c), and scalar-arm unions aren't generated from the schema's `scalarValues` signal (`input.Origin`).

**Notes**

① Cheaper for **static** implementations: the type system gives it with little extra code, even
un-generated.

② Cheaper for **dynamic** implementations: a plain runtime value does the job, with no static wrapper or
typed map.

③ Narrows a small, bounded `js-uint` field (a status code, a line or column number) to a 32-bit `int`; the
genuinely wide fields (byte sizes, counts) stay `long`, so the miss is limited to these small slots.

④ Selenium's shared `NumberCoercer` is too lenient to enforce this (it truncates fractional integers and
accepts float- or string-encoded numbers); the BiDi layer likely needs a stricter, BiDi-scoped coercer.

⑤ Performs no runtime union dispatch, so an undeclared closed-union variant is not rejected.

⑥ Fills a missing required field with a default (`0`/`null`/`false`) a caller can't distinguish from a real
value, so downstream code acts on fabricated data with no signal the field never arrived.

⑦ Compliant, but does not log the warning the ADR requires here.

### 🗣 Discussion

- [2026-08-06](https://www.selenium.dev/meetings/2026/tlc-08-06/) — compared the ADR point by point with the webdriverbidi-net implementation; the two agreed on desired behavior except on tolerating a missing required inbound field, and pujagani took an action to check whether it was feasible in Java.
- [2026-08-13](https://www.selenium.dev/meetings/2026/tlc-08-13/) — pujagani reported that allowing a missing required field would be challenging to implement properly in Java, so the point was dropped, resolving the divergence and leaving titusfortner, pujagani, and jimevans in agreement.
- [2026-09-03](https://www.selenium.dev/meetings/2026/tlc-09-03/) — clarified that validation is done against a generated schema (published specs plus supported vendor extensions), where issues like a missing required field can be managed with temporary overrides during generation; AutomatedTester questioned the growing number of CDDL files and checking the schema into the repo, and was satisfied by the versioning rationale.

### 📌 Tracking

### DIFF
--- a/docs/decisions/17786-bidi-low-level-behavioral-contract.md
+++ b/docs/decisions/17786-bidi-low-level-behavioral-contract.md
@@ -121,14 +121,12 @@ unrecognized error code included, even if that reason would otherwise fail one o
   still layer one on top. This covers only a carrier for a discriminator the spec does not declare; a shared
   carrier for a declared variant a binding has not modeled distinctly is the decision-1 representation
   choice, not this behavior.
-- **Tolerate a missing required field inbound** (decision 8). Represent an absent required field as
-  *omitted* and warn, rather than erroring, so a remote end lagging a newly-required field does not cost the
-  caller the message. Rejected: a static, generated binding cannot hold a required field as *omitted*
-  without typing it away from its declared shape (a required-nullable field then needs an extra
-  omitted-vs-null state), and no clean option exists for this in Java, so the tolerance is not implementable
-  consistently across bindings. Erroring keeps required-ness symmetric and matches every binding, including
-  the webdriverbidi-net reference; a lagging field is handled by a project schema override, not by runtime
-  tolerance.
+- **Tolerate a missing required field inbound** (decision 8). Represent it as *omitted* and warn instead of
+  erroring, so a lagging remote end does not cost the caller the message. Rejected: an absent required field
+  is then null at runtime, corrupting the typed contract in every statically-typed binding — held null under
+  a non-null type, or widened to nullable so every consumer must check a field the spec guarantees. Only
+  untyped dynamic bindings tolerate freely. Erroring keeps the guarantee and one behavior across bindings; a
+  lagging field is relaxed in the schema (a project override) instead.
 - **Surface message-level extras in this layer** (decision 9) rather than leaving them to the transport.
   Rejected: the envelope is the transport's; this layer governs per-type extensibility only.
 - **Retain extras only where they can be sent back** (decision 9), narrower than every extensible type.
@@ -144,9 +142,7 @@ unrecognized error code included, even if that reason would otherwise fail one o
 - Outbound validity differs in cost by binding: a static binding gets it from construction, while a
   dynamic binding must enforce it with an explicit runtime check. The contract requires the behavior from
   both; where a dynamic binding does not yet check, that is a gap to close, not an exemption.
-- Required-ness is symmetric: a required field must be present and valid in both directions (decisions 5
-  and 8). A static binding enforces it on the type itself, typing each field by its declared shape with no
-  omitted-vs-null representation, which is what makes the contract implementable in a generated static
-  binding like Java. The cost is backward compatibility: a remote end that omits a newly-required field
-  errors the whole message; relaxing that field with a project override degrades it to a missing field
-  rather than a failed session.
+- Required-ness is symmetric (decisions 5 and 8): a required field is validated, not represented as absent,
+  so it is never null at runtime and no binding widens it to nullable. The cost is backward compatibility —
+  a remote end omitting a newly-required field errors the whole message until that field is relaxed in the
+  schema (a project override), a reactive fix (notice, override, ship) during which the message fails.


docs/decisions/17786-bidi-low-level-behavioral-contract.md
```diff
@@ -0,0 +1,190 @@
...
+### Inbound (receiving and parsing)
+
+5. **Variants and vocabulary are resolved by the spec's declared rule.** *(Compliance.)* A union resolves
+   to a variant by its declared rule (discriminator value, presence of required keys, or declared
+   default), never a re-derived structural guess, so a valid payload always resolves to the same variant;
+   resolving it wrong is a parsing bug. Raising on an *unrecognized* enum token or union variant is the
+   strict default (a Decision, item 9), not part of this compliance floor; for a union whose arms are
+   all objects, a non-object payload has no variant to select and raises.
+
+6. **Inbound fields are validated strictly against the resolved type.** *(Decision.)* Once item 5 has
```
nvborisenko: This may lead to bad situation. Let's image one required (non-nullable) field is added into the spec. Selenium added it accordingly. Users will get serialization error because this field is missing in payload... because... some vendor still didn't implement it.
IMHO serialization process should be relaxed.
Real example: https://github.com/w3c/webdriver-bidi/pull/1117

titusfortner: So you know, this comment resulted in my spending the past 5 days overhauling the ADR. Not sure it was worth that much time investment, but I'm much happier with it now. The TL/DR is that I now think inbound should be strictly validated *except for missing required parameters.*

titusfortner:
I just pushed a substantial rewrite of both the ADR and updated the PR description to match it.

The main behavioral change is inbound required-field absence: the proposal now leaves a missing required inbound field omitted and warns, rather than rejecting the whole message by default. That will likely end up being the main policy question for review: should Selenium tolerate that incorrectness in order to be more flexible to browser/schema drift.

The ADR is now framed around representability: outbound stays strict because Selenium controls what it sends; inbound tolerates only departures that can be represented without inventing data; corrupt present values and unknown closed-vocabulary members still error.

The implementation comparison in the PR body is there to show which current differences are implementation cost/adoption gaps and which ones are real policy choices.

docs/decisions/17786-bidi-low-level-behavioral-contract.md
```diff
@@ -0,0 +1,207 @@
...
+   and failed field checks (item 6) raise; a strict contract is cheaper to loosen than a lenient one is
+   to tighten. A specific field may be relaxed where a real remote end sends it off-spec, and any such
+   relaxation is bounded:
+   - **Only absence is relaxable** (item 6): never corruption, and never a field item 5 dispatches on.
+   - **It removes a check; it never exposes a type, variant, or enum value the spec does not define.**
+     Inventing one to absorb a malformed payload models a browser's bug as protocol, and it outlives
+     the bug.
+   - **The tolerated absence is still reported**, never silently absorbed. A binding *may* additionally
+     let a caller admit an uncatalogued absence at runtime, so a user is not blocked until the next
+     release; that is a per-binding convenience, not required here, and it too reports.
```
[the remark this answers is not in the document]
titusfortner: I've clarified this to the extent I think it's necessary

docs/decisions/17786-bidi-low-level-behavioral-contract.md
```diff
@@ -0,0 +1,137 @@
...
+2. **Outbound, a caller's own missing or extra field is rejected.** For any command defined in the spec, a
+   missing required field, or an unknown property on a non-extensible type, is a local error rather than a
+   remote round-trip — strict, because the payload is the caller's to get right.
+
+3. **Inbound, those same two are tolerated.** Process error responses first: a remote error surfaces as an
+   error even if its payload fails validation — an unrecognized error code included — and the checks here
+   must not turn it into a local serialization failure; the error's contents are otherwise out of scope.
+   Otherwise these rules govern every typed payload the layer parses — a command's result and a
+   server-initiated event's parameters alike, with response correlation and event routing out of scope:
+   - **A missing required field warns, and is left *omitted*** — not an explicit *null*, and not a generic
```
nvborisenko: There are shared types which are involved in both inbound and outbound messages. [network.header](https://w3c.github.io/webdriver-bidi/#cddl-type-networkheader) is good example.

nvborisenko: Resolving it from implementation perspective... if we really differentiate rules for inbound/outbound then apply it  at caller side:

Outbound:
```
DoSomething(arg)
{
  // validate arg
}
```

titusfortner:
Reorganized and clarified since the last push — the four-decision record (D1–D4) is now three sections (Representation, Outbound, Inbound) with ten numbered decisions, so a fresh read top to bottom helps. Substantive changes:

* Numeric: BiDi integers are bounded to the JS safe-integer range (±(2^53−1)), so a 64-bit int or a double is exact; dropped the "int64/bigint, never a double" framing.
* Extensibility: stated in decision 1, including that a declared key never appears in the extras map — closes a value-shadowing footgun found implementing Ruby.
* Unions/`RemoteValue`: "unresolvable" now means a discriminator the spec doesn't declare; structurally-identical variants may share one type, so a shared carrier is conformant and only a genuinely unknown/newer type is rejected.
* Fidelity: it's fidelity of the value, not its byte-form — native types are fine as long as nothing is lost.
* Type validity: decision 4 now reads as a shared definition that decisions 5 (outbound) and 7 (inbound) enforce, not a behavior of its own — so it has no standalone compliance point.

docs/decisions/17786-bidi-low-level-behavioral-contract.md
nvborisenko: And inbound:

```
DoSomething(arg)
{
  // validate arg
  var result = DelegateToWire(arg...);
  // validate result (inbound)
}
```

The idea is simple: don't include validation into DTO types, caller should be aware of correctness of representation layer. Leave DTO as is.

titusfortner: The ADR added the proposal not to error for missing required fields after your previous comment. This doesn't mean there can't be additional validation in DTO types, though (some of which static languages get for free). 

If we decide to forego the added flexibility and align completely with the webdriverbidi-net implementation by more correctly erroring on missing required fields, then almost everything can be validated on the DTO type for static languages. 

Most importantly, if the validations aren't being done by the DTO type, they must be done by the  serialization implementation at the boundary.

docs/decisions/17786-bidi-low-level-behavioral-contract.md
```diff
@@ -0,0 +1,157 @@
...
+6. **Send an extra field only on an extensible type.** An extensible type carries it in the map (decision 1)
+   and serializes it; a non-extensible type has no such map, so an extra field cannot be represented.
+
+### Inbound
+
+An inbound payload is the remote end's responsibility; the layer must tolerate what it can accurately
+represent and reject only what it cannot, providing flexibility for a remote end on a spec version newer or
+older than the one the binding validates against.
+
+7. **Reject an invalid value.** A value that is present but not a valid instance (decision 4)
```
nvborisenko, line 77: I remember a case where Chromium sent `float` instead of `integer`: https://github.com/SeleniumHQ/selenium/blob/076d428580b7b101c62722cef10be995ffd2e44c/dotnet/src/webdriver/BiDi/Json/Converters/DateTimeOffsetConverter.cs#L29

One more case to be relaxed?

nvborisenko: Honestly, I am providing real cases just to challenge. Is it show-stopper for end users?

I still believe we want to send/receive messages to/from remote end as is. At very higher level we might validate messages.

nvborisenko: I could understand that Jim's implementation is north star for you, but current Selenium.NET is the best, I can explain any line in bidi namespace.

docs/decisions/17786-bidi-low-level-behavioral-contract.md
```diff
@@ -0,0 +1,159 @@
...
+The transport layer (the connection that carries messages and matches responses to their commands) and
+the orchestration layer (session lifecycle, event subscription and routing, and the high-level API these
+feed) are out of scope, as is any ergonomic sugar over these types that a higher public layer adds.
+
+## Decision
+
+In this record, "the spec" is the declared schema the layer validates against, which includes all relevant
+published WebDriver BiDi specifications and any vendor-defined extensions. The spec fixes the layer's
+baseline: spec strings reach the wire verbatim, an *omitted* field stays distinct from an explicit `null`,
+and a union resolves by the rule the spec declares. What it leaves open,
```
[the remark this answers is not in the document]
titusfortner: true, but not worth the churn to clarify in the framing clause. It is correct in the decision.

docs/decisions/17786-bidi-low-level-behavioral-contract.md
titusfortner: A whole number sent as a float (5.0) is already accepted under decision 4. If the browser sends a fractional value, we don't want to change it to a different number just to fit our typed representation, that's the opposite of receiving it "as is".

docs/decisions/17786-bidi-low-level-behavioral-contract.md
```diff
@@ -0,0 +1,152 @@
...
+remote end on a newer spec version does not break the binding; it rejects a malformed value (decision 7) or
+a missing required field (decision 8), neither of which yields a valid typed object.
+
+7. **Reject an invalid value.** A value that is present but not a valid instance (decision 4)
+   cannot yield a valid object, so the layer must raise an error.
+8. **Reject a missing required field.** A required field absent from an inbound payload is an error: the
+   layer must raise rather than substitute a placeholder or represent the field as *omitted*.
+9. **Tolerate an undeclared field.** If the type is declared extensible, the layer must preserve the
+   field in the type's map (decision 1). If it is not, the layer must log a warning that an undeclared field
+   was received, and drop it.
```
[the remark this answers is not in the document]
titusfortner: it's just a logged warning.

docs/decisions/17786-bidi-low-level-behavioral-contract.md
titusfortner: And it turns out that strictly typed languages can't easily tolerate missing required parameters without inserting placeholder data, so the ADR is back to proposing it be validated, but to have a way to relax the requirements from the schema generation if necessary.

nvborisenko: Two types of parameters: reference and value.

1. Reference type: can be easily be fulfilled by `null`. Convinient for end users? I think yes, not bad. User gets "truly" response from remote end. Nice to NOT fail if user even don't access this "required" parameter.
2. Value type... integer/bool. If selenium fullfill it by default, then selenium lies! It is bad. But what if user even doesn't access it?.. Fail the entire train is also bad.

Our train is so complicated.. ff, chromium, safari. I am dreaming we will work with all.. how.. be relaxed.

docs/decisions/17786-bidi-low-level-behavioral-contract.md
nvborisenko: Browser is doing it.. it sends `123.321` while in spec it is `integer`.

titusfortner: Agreed a browser can send 123.321 for a field the spec types integer. The options are reject it (surfaces the mismatch), silently truncate to 123 (invents a value the browser didn't send — the opposite of "as is"), or keep 123.321 (then it isn't an integer, and the typed object breaks). We reject. If a field legitimately carries fractional values the spec should type it number; meanwhile a project override can relax that specific field rather than loosening the integer rule everywhere.

docs/decisions/17786-bidi-low-level-behavioral-contract.md
```diff
@@ -0,0 +1,152 @@
...
+10. **Preserve received values faithfully.** Fidelity is of the value, not its byte-form: a binding may hold
+    any value it parses, a declared field or a retained extra alike, in an ergonomic native type (a 64-bit
+    integer for a `js-int`, a date object for a date), provided it loses nothing and can reproduce what the
+    wire carried. It must not truncate, round, re-case, or otherwise normalize a value beyond
```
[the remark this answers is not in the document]
titusfortner: fidelity of the value, not the byte-form

PR #17801, opened 2026-07-20
Add high-level BiDi script pin and unpin commands

mayank-at-sauce:
**Part of #13992** — implements the pin / unpin cells for the Ruby binding.

**What does this PR do?**
Adds driver.script.pin and driver.script.unpin to the Ruby bindings, bringing the high-level BiDi Script API closer to parity with Python/Java/JS (which already have these).

pin takes a function declaration and registers it as a BiDi preload script, so it runs on every new browsing context before the page's own scripts — handy for injecting helpers, polyfills, or instrumentation. It returns an id that you pass to unpin to remove it later:

id = driver.script.pin('() => { window.helper = () => 42; }')
driver.navigate.to("https://example.com")
# window.helper is available on the page
driver.script.unpin(id)

**Implementation Notes**
Under the hood these are thin wrappers over the existing script.addPreloadScript / script.removePreloadScript BiDi commands, sent via @bidi.send_cmd — the same pattern the network handlers already use. I kept the surface deliberately minimal (return the raw preload-script id rather than a wrapper object) to match how the JS binding exposes it. Script#initialize now keeps a reference to the bidi object so both the log handlers and these commands can share it; behaviour when BiDi isn't enabled is unchanged (it still raises on driver.script). Added RBS signatures and integration specs covering both the pin path (script runs on a freshly navigated document) and the unpin path (script no longer runs).

### DIFF
--- a/common/repositories.bzl
+++ b/common/repositories.bzl
@@ -182,8 +182,8 @@ js_library(
 
     http_archive(
         name = "mac_edgedriver",
-        url = "https://msedgedriver.microsoft.com/150.0.4078.65/edgedriver_mac64_m1.zip",
-        sha256 = "0a3edefa906c7d492e7f1104c6875222228049a1fe37d6ad779b6d84df7fe076",
+        url = "https://msedgedriver.microsoft.com/150.0.4078.83/edgedriver_mac64_m1.zip",
+        sha256 = "431c67a8a15b677ff96d208f983a9c02f9be9a26e07a0a0eac6b12bd1cafba90",
         build_file_content = """
 load("@aspect_rules_js//js:defs.bzl", "js_library")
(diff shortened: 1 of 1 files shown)


titusfortner:
We need to clarify the behavior we expect from this, since it is inconsistent between bindings right now.
That means deciding this first https://github.com/SeleniumHQ/selenium/pull/17776

titusfortner:
Also, we're migrating things to implement off of the newly generated `BiDi::Protocol` classes. I haven't dug too deep into your implementation, but I'd like to get rid of the BiDi class entirely in the near future. Are you on Slack? Can we discuss it there?

mayank-at-sauce:
Hi @titusfortner sure, let me connect with you on slack. Thanks!

PR #17812, opened 2026-07-22
[build] generate and ship cargo-free SBOM and license notices with Selenium Manager

### LINKED ISSUE #17808: [🚀 Feature]: Include SBOMs with Selenium Manager binaries
### Description

The Python packages for *selenium* currently bundle Rust-based binaries in their wheels:

* `selenium/webdriver/common/linux/selenium-manager`
* `selenium/webdriver/common/macos/selenium-manager`
* `selenium/webdriver/common/windows/selenium-manager.exe`

For these binaries, the license data is incomplete (especially regarding the used dependencies) and thus possibly violates the corresponding licenses. For this reason, I propose to at least provide corresponding SBOM files with the distributions for the bundled dependencies and possibly generate a third-party license/notice file as well.

Use the corresponding `Cargo.lock` file from Git to generate the corresponding files myself.

titusfortner:
### 🔗 Related Issues

Fixes #17808

### 💥 What does this PR do?

- Generates two license-compliance artifacts inside every package that bundles the Selenium Manager binary — Python wheel, Java jar, Ruby gem, npm package, NuGet package.
- `selenium-manager.cdx.json` — a CycloneDX SBOM inventorying the ~300 statically-linked Rust crates (name, version, checksum, dependency graph, SPDX license) for vulnerability and license scanning.
- `selenium-manager-THIRD-PARTY-NOTICES.txt` — the third-party attribution file reproducing each bundled crate's copyright notice and full license text, which MIT/BSD/ISC/Apache-2.0 require in binary distributions.

### 🔧 Implementation Notes

- Generated entirely within Bazel with no cargo dependency. A Bazel aspect walks the crate dependency graph and harvests the `Cargo.toml` manifests and `LICENSE`/`NOTICE` files crate_universe has already vendored; the SBOM component graph comes from `rust/Cargo.lock`.
- Each package bundles all three OS binaries, which link different crate subsets (e.g. `winapi` only on Windows). A split transition seeds the aspect under every shipped target triple and unions the results, so both artifacts cover all platforms from a single build without cross-compiling — only the crates' source files are read.
- SBOM component list is intentionally the full `Cargo.lock` graph — industry-standard, where over-inclusion is safe and omission is a coverage gap. Crates not linked on any platform (build-only, test-only) resolve to `NOASSERTION`. Three crates declare an SPDX license but bundle no license file and are listed as such.
- Python sdist ships Rust source (including `Cargo.lock`) rather than the compiled binary so is excluded.

### 🤖 AI assistance

    - Tool(s): Claude Code
    - What was generated: SBOM + NOTICE generators, Bazel aspect/rules, and per-binding packaging wiring

- Both files sit next to the binary (present and valid) rather than in the ecosystems' emerging reserved SBOM locations — `.dist-info/sboms/` ([PEP 770](https://peps.python.org/pep-0770/)) for wheels and `META-INF/sbom/` for jars — because the pinned `rules_python` `py_wheel` cannot write into `.dist-info/` and the Java move would disturb the manager binary's resource-loading path. Deferred as a follow-up (gated on `rules_python` PEP 770 support) so a generic scanner can auto-discover the SBOM.

### 🔄 Types of changes

### DIFF
--- a/common/manager/BUILD.bazel
+++ b/common/manager/BUILD.bazel
@@ -1,7 +1,60 @@
+load("@rules_python//python:defs.bzl", "py_binary")
+load(":crate_metadata.bzl", "selenium_manager_notice", "selenium_manager_sbom")
+
 package(
     default_visibility = [
         "//dotnet/src/webdriver:__pkg__",
         "//dotnet/test/common:__pkg__",
+        "//java/src/org/openqa/selenium/manager:__pkg__",
+        "//javascript/selenium-webdriver:__pkg__",
+        "//py:__pkg__",
+        "//rb:__pkg__",
+    ],
+)
+
+py_binary(
+    name = "sbom_generator",
+    srcs = ["generate_sbom.py"],
+    main = "generate_sbom.py",
+    visibility = ["//visibility:private"],
+)
+
+py_binary(
+    name = "notice_generator",
+    srcs = ["generate_notice.py"],
+    main = "generate_notice.py",
+    visibility = ["//visibility:private"],
+)
+
+# CycloneDX SBOM for the bundled Selenium Manager binary. Fully cargo-free: the
+# component list and dependency graph come from rust/Cargo.lock, and licenses
+# are harvested off the crate dependency graph via crate_manifests_aspect (the
+# vendored Cargo.toml manifests crate_universe already fetched). Every binding
+# copies this artifact next to the binary it ships.
+selenium_manager_sbom(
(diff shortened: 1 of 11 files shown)


titusfortner:
Unfortunately, I think the right way to do this is actually to extend our current release assets hack rather than try to get it working cross platform with bazel.

titusfortner:
Closing in favor of #17820 
The bazel-only approach will need to be sorted in #17586

PR #17875, opened 2026-08-05
[dotnet] support SE_*DRIVER environment variables to set driver locations

titusfortner:
### 🔗 Related Issues

Part of #14045

### 💥 What does this PR do?

- Adds .NET support for setting driver locations via environment variables, bringing it in line with Ruby, Python, and Java.
- When set, the matching variable is used to locate the driver and Selenium Manager is skipped: `SE_CHROMEDRIVER`, `SE_EDGEDRIVER`, `SE_GECKODRIVER`, `SE_IEDRIVER`, `SE_SAFARIDRIVER`.
- An explicitly configured driver path still takes precedence over the environment variable.

### 🔧 Implementation Notes

- Resolution order matches Java/Ruby: configured path → environment variable → Selenium Manager.

### 🤖 AI assistance
    - Tool(s): Claude Code
    - What was generated: implementation and tests

### 🔄 Types of changes

### DIFF
--- a/dotnet/test/webdriver/DriverServiceTests.cs
+++ b/dotnet/test/webdriver/DriverServiceTests.cs
@@ -17,6 +17,7 @@
 // under the License.
 // </copyright>
 
+using System.ComponentModel;
 using System.IO;
 using OpenQA.Selenium.Chrome;
 using OpenQA.Selenium.Edge;
@@ -30,8 +31,6 @@ namespace OpenQA.Selenium.Tests;
 [NonParallelizable]
 public class DriverServiceTests
 {
-    private sealed class StopStartupException : Exception;
-
     private static IEnumerable<TestCaseData> DriverServices()
     {
         yield return new TestCaseData((Func<DriverService>)ChromeDriverService.CreateDefaultService, "SE_CHROMEDRIVER").SetName("Chrome");
@@ -49,19 +48,10 @@ public class DriverServiceTests
         try
         {
             Environment.SetEnvironmentVariable(environmentVariable, expectedPath);
-            DriverService service = createService();
 
-            string actualFileName = null;
-            service.DriverProcessStarting += (_, e) =>
-            {
-                actualFileName = e.DriverServiceProcessStartInfo.FileName;
-
-                // Stop before the (nonexistent) driver is actually launched.
-                throw new StopStartupException();
-            };
-
-            Assert.ThrowsAsync<StopStartupException>(async () => await service.StartAsync());
-            Assert.That(actualFileName, Is.EqualTo(expectedPath));
+            Assert.That(
+                async () => await createService().StartAsync(),
+                Throws.InstanceOf<Win32Exception>().With.Message.Contains(expectedPath));
         }
         finally
         {


dotnet/src/webdriver/DriverService.cs
```diff
@@ -147,6 +147,23 @@ public int ProcessId
...
+
+    /// <summary>
+    /// Gets the driver executable path from <see cref="DriverServiceEnvironmentVariable"/>, or
+    /// <see langword="null"/> if it is unset. When set, Selenium Manager is not invoked.
+    /// </summary>
+    public string? DriverPathFromEnvironment =>
+        this.DriverServiceEnvironmentVariable is string name
+        && Environment.GetEnvironmentVariable(name) is string path
+        && path.Length > 0
+            ? path
```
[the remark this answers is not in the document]
titusfortner: unnecessary / overly defensive

dotnet/test/webdriver/DriverServiceTests.cs
```diff
@@ -0,0 +1,57 @@
...
+        yield return new TestCaseData((Func<DriverService>)FirefoxDriverService.CreateDefaultService, "SE_GECKODRIVER").SetName("Firefox");
+        yield return new TestCaseData((Func<DriverService>)InternetExplorerDriverService.CreateDefaultService, "SE_IEDRIVER").SetName("InternetExplorer");
+        yield return new TestCaseData((Func<DriverService>)SafariDriverService.CreateDefaultService, "SE_SAFARIDRIVER").SetName("Safari");
+    }
+
+    [TestCaseSource(nameof(DriverServices))]
+    public void ReadsDriverPathFromEnvironmentVariable(Func<DriverService> createService, string environmentVariable)
+    {
+        string original = Environment.GetEnvironmentVariable(environmentVariable);
+        string expectedPath = Path.Combine("path", "to", "driver");
```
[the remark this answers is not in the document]
titusfortner: The test project isn't compiled in a nullable-enabled context

dotnet/src/webdriver/DriverService.cs
```diff
@@ -147,6 +147,23 @@ public int ProcessId
     /// </summary>
     public string? DriverServicePath { get; set; }
 
+    /// <summary>
+    /// Gets the name of the environment variable used to specify the driver executable location,
+    /// or <see langword="null"/> if the service does not support one.
+    /// </summary>
+    protected virtual string? DriverServiceEnvironmentVariable => null;
```
nvborisenko: Not clear whether it is a value or key.. Better: `DriverServiceEnvironmentVariableName`

dotnet/src/webdriver/DriverService.cs
```diff
@@ -147,6 +147,23 @@ public int ProcessId
...
+
+    /// <summary>
+    /// Gets the driver executable path from <see cref="DriverServiceEnvironmentVariable"/>, or
+    /// <see langword="null"/> if it is unset. When set, Selenium Manager is not invoked.
+    /// </summary>
+    public string? DriverPathFromEnvironment =>
+        this.DriverServiceEnvironmentVariable is string name
+        && Environment.GetEnvironmentVariable(name) is string path
+        && path.Length > 0
```
nvborisenko: Please use `string.IsNullOrWhitespace(path)` to be more friendly for users. Or at least `string.IsNullOrEmpty(path)`.
Then can be simplified: `string.IsNullOrWhitespace(Environment.GetEnvironmentVariable(name))`

dotnet/src/webdriver/DriverService.cs
```diff
@@ -147,6 +147,23 @@ public int ProcessId
...
+    /// Gets the name of the environment variable used to specify the driver executable location,
+    /// or <see langword="null"/> if the service does not support one.
+    /// </summary>
+    protected virtual string? DriverServiceEnvironmentVariable => null;
+
+    /// <summary>
+    /// Gets the driver executable path from <see cref="DriverServiceEnvironmentVariable"/>, or
+    /// <see langword="null"/> if it is unset. When set, Selenium Manager is not invoked.
+    /// </summary>
+    public string? DriverPathFromEnvironment =>
```
nvborisenko: New public API surface.. We already have properties:
- `DriverServiceExecutableName`
- `DriverServicePath`

Introducing one more `DriverPathFromEnvironment` property makes it unclear.

Not yet dived deeper, but my first think is to not expose it.

titusfortner: Yeah, the Agent wanted it for testing, which was excessive. Removing both.

nvborisenko: For testing purpose we can test behavior: let driver service to start, and see whether it throws exception. I believe it should throw exception.
If not, then we can subscribe on `BeforeDriverServiceProcessStarting` event and verify actual file name.

If we can test it easily, I prefer having test rather than nothing.

titusfortner: Not sure if what I added is what you're looking for, either, though.

nvborisenko: I pushed simplified version of the test right here (sorry).

dotnet/test/webdriver/DriverServiceTests.cs
```diff
@@ -0,0 +1,61 @@
...
+        try
+        {
+            Environment.SetEnvironmentVariable(environmentVariable, expectedPath);
+
+            Assert.That(
+                async () => await createService().StartAsync(),
+                Throws.InstanceOf<Win32Exception>().With.Message.Contains(expectedPath));
```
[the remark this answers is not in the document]
nvborisenko: That's fine at this moment. Eventually we will raise our better exception, but this is different story.

titusfortner:
Thanks for fixing, much easier than me trying to figure it out. :)

PR #17886, opened 2026-08-06
[docs] ADR: Selenium waits for interaction readiness before it interacts

AutomatedTester:
📄 The decision, its rationale, considered options and consequences are in the record file this PR
adds; read it there. The sections below are proposal notes and review logistics.

**Revised after review** ([a9f895c](https://github.com/SeleniumHQ/selenium/commit/a9f895cac7be5498992cdaa1221033db4b53c256)):
an earlier revision proposed that readiness waiting become the default for every BiDi session. It no
longer does. Readiness is enabled by a new `se:interactionReadiness` capability, off by default, and
when enabled it **replaces the implicit wait** rather than adding a second timeout beside it. Making
it the default is now a follow-up decision, probably release-gated.

### 🔗 Related

- Builds on [17670 — BiDi implementation boundaries](https://github.com/SeleniumHQ/selenium/blob/trunk/docs/decisions/17670-bidi-implementation-boundaries.md):
  this is high-level, protocol-neutral API with BiDi as an implementation mechanism that must not
  leak into signatures. No method signature changes and no BiDi type is named; the behavior is
  capability-gated.
- Reference implementation (Python): [`py-quiescence-bidi-preload`](https://github.com/SeleniumHQ/selenium/tree/py-quiescence-bidi-preload)
  — the readiness oracle as a JavaScript atom registered as a BiDi preload script, with ~1,500 lines
  of behavioral tests. Evidence that the semantics are implementable and testable; **not** the
  proposed API shape (see divergences below).
- Intended to become a proposed WebDriver BiDi `quiescence` module; no working-group issue filed yet.

### 📝 Proposal notes

**A correction the record now carries.** The first revision claimed WebDriver does not wait before
interacting. That is wrong, and the corrected version is a stronger argument:

| Command | Waiting behavior in the spec |
|---|---|
| Element Clear | Steps 6–10: takes the implicit wait timeout, starts a timer, waits for the element to become *interactable* |
| Element Send Keys | Steps 7.2–7.6: the same pattern, waits for *keyboard-interactable* |
| Element Click | Steps 5–7: scroll, then fail. No wait |

*keyboard-interactable* is "any element that has a focusable area, is a `body` element, or is the
document element" — invisible, covered, mid-animation and `aria-disabled` elements all satisfy it,
and *interactable* is satisfied the same way. The timeout is initially `0`, and when set it is also
spent on element retrieval. So the gap is not that WebDriver never waits: it is that the predicate is
too weak to prevent the failures users hit, it is absent for `click`, it cannot say why it expired,
it shares one budget with element location, and remote ends improvise past it (chromedriver waits for
displayed before clicking).

**Why the capability replaces the implicit wait instead of adding a timeout.** For `send_keys` and
`clear` the remote end already runs its own interactability wait on the implicit wait timeout. A
separate readiness timeout would stack with it and hand a user who set an implicit wait a total they
never asked for. So when the capability is on, Selenium stops forwarding a non-zero implicit wait and
spends that budget client-side on a predicate that is actually useful.

**The consequence that needs a decision.** Because the implicit wait is no longer forwarded,
`findElement` loses its remote-end retry, so the record proposes Selenium take over element-location
waiting on the client for these sessions. That trades local spinning at the remote end for polling
round-trips. It is decision 2 and considered option 8, and it is the part of the design I would most
like confirmed.

**What the wait covers.** Actionability only — visible, enabled (including `aria-disabled` and the
`fieldset`/`legend` exception), editable where the interaction writes, in the viewport, unobstructed,
not moving. It deliberately does *not* require page settledness: an application that long-polls or
animates continuously would pay a settle timeout on every click. The low-level Actions API does not
auto-wait either.

**How users reach the rest.** There is no new wait method on the driver or on the element. The wait
is inside the interaction commands; the underlying state is reachable as a snapshot (composable with
an existing `WebDriverWait`) and as a settledness handler following the existing
`driver.script.add_dom_mutation_handler` pattern — preload script, `ChannelValue`, `script.onMessage`
— so no client library is forced into Selenium's waiting strategy to use Selenium's readiness data.
The record fixes the arguments and return shapes of these
accessors; only the per-binding spelling of their names is deliberately left open.

**Also proposed:** a `settled` page load strategy, managed locally because `pageLoadStrategy` is a
validated `none`/`eager`/`normal` capability, plus a per-action strategy override.

**Compatibility.** Nothing changes for a session that does not set the capability. Where it is set,
an expired wait raises a readiness-specific *subclass* of what the command raises today
(`ElementClickInterceptedException`, `ElementNotInteractableException`), so existing
`catch`/`except`/`rescue` blocks keep matching while the readiness case stays distinguishable.

**Deliberately out of scope,** as follow-up decisions: making readiness the default for BiDi sessions
and in which release; a variant for classic sessions (considered option 10 argues it cannot be done
at full fidelity without a preload); deprecating `ExpectedConditions.elementToBeClickable` and
equivalents; the per-binding spelling of the accessor names; the BiDi module proposal itself.

**Divergences from the reference implementation.** It exposes two `driver.*` wait methods, which
decisions 4 and 5 replace, and a single atom behind one combined entry point, which decision 8 keeps
as one artifact but re-shapes into narrow named exports. The capability, the implicit-wait replacement, the
client-side element-location retry, the error subclasses and the `settled` strategy are proposed here
and not yet built.

**Cross-binding impact.** Every binding gains the capability, the interaction-path change, the
snapshot and handler surface, the error subclasses, the page load strategy value, and the packaging
wiring to ship a JavaScript resource; bindings that already ship atoms have most of the last part.
Because the capability is off by default, existing tests are unaffected — each binding instead needs
new coverage for on, off, and preload registration failing.

### 🗣 Discussion

Questions I would most like input on:

1. **Element location.** Is Selenium taking over `findElement` retry on the client the intended
   reading of "readiness replaces implicit wait", or should this be scoped to interaction only and
   the implicit wait still forwarded?
2. **The budget when no implicit wait is set.** The record proposes 10s. Playwright uses 30s. A high
   value makes failing suites slow; a low one makes the feature look unreliable on slow CI.
3. **Naming and placement of the snapshot and handler surface** — left open here on purpose, for a
   separate discussion.
4. **Scope of the guarantee.** The heuristics will be wrong sometimes; a false "not ready" turns a
   passing test into a timeout. Is "best-effort, documented limits, off unless asked for" enough?
5. **The path to default-on.** Off by default is shippable but reaches only users who know the
   capability exists — which is not the population that flakes. What would make default-on
   acceptable: a major release, a release cycle of feedback, something else?

- [2026-08-13](https://www.selenium.dev/meetings/2026/tlc-08-13/#interaction-readiness-adr) — stays a single preload script with
  separate exposed entry points rather than separate atoms, split along stateful vs stateless lines,
  with the interactability path fixed to use the readiness state it currently ignores.

### 📌 Tracking

### DIFF
--- a/docs/decisions/17886-interaction-readiness-waits.md
+++ b/docs/decisions/17886-interaction-readiness-waits.md
@@ -132,8 +132,7 @@ reach the functionality. The wait lives inside the interaction commands, where t
 have to know it is there; everything else is reached through decision 5.
 
 **5. Readiness state is exposed as snapshots and as events, not only as waits.** Both signals are
-readable from the `script` domain, so users can compose them with the waiting strategy they already
-have:
+readable without waiting, so users can compose them with the waiting strategy they already have:
 
 - A snapshot call returns the current record — pending work classes, still-mutating regions, or for
   an element the per-check result and the obstructing element — without waiting. This is what makes
@@ -142,6 +141,29 @@ have:
   (`java/src/org/openqa/selenium/remote/RemoteScript.java:85`): preload script, `ChannelValue`,
   `script.onMessage`. This is the same machinery the readiness observer already needs, so a
   settledness event is a payload change rather than new plumbing.
+
+Each accessor is placed by the scope of the state it describes: document state on the `script`
+domain, element state on the element itself. Nothing takes an element as an argument to a call that
+does not belong to it, which is what decision 4 objects to. The mechanism is the `script` domain in
+both cases.
+
+```python
+# document state
+driver.script.pending_work()                                  -> PendingWork
+driver.script.settledness(root=None, settle_window=None)      -> Settledness
+driver.script.add_settled_handler(callback)                   -> int   # id
+driver.script.remove_settled_handler(id)                      -> None
+
+# element state
+element.actionability(interaction="click")                    -> Actionability
+```
+
+Names are spelled here in Python and are illustrative: each binding uses its own casing and its own
+optional-argument idiom, and the names themselves are the follow-up in the Consequences. What the
+record fixes is the placement, the arguments and the return shape, so that five bindings expose the
+same thing under five spellings. `settle_window` defaults to the session's configured window and
+`root` to the document, so the common call takes no arguments; `interaction` defaults to `click` and
+accepts the same values the atom accepts.
 
 This shape does not assume an asynchronous script API. It fits the script module's planned
 synchronous execution with handlers for asynchronous behavior, and it means no client library is
@@ -165,11 +187,49 @@ exposes a small set of named exports rather than one combined entry point. Each 
 the arguments belonging to it and returns only its own result — pending work for a document,
 settledness given a root and a settle window, actionability given an element and an interaction — so
 a binding that wants an actionability snapshot neither constructs arguments for settledness nor
-interprets a record containing it. Bindings pass arguments and marshal results; no binding adds
-logic of its own. One artifact is what holds the heuristics together: what counts as meaningful
-mutation, when a spinner is inert and how obstruction is hit-tested share definitions, will not stay
-identical across five hand-written ports, and are one thing to hand to the BiDi working group. The
-exports are what keep a caller from taking on the parts it did not ask for.
+interprets a record containing it:
+
+```ts
+export function pendingWork(doc: Document): PendingWork;
+export function settledness(root: Node, settleWindowMs: number): Settledness;
+export function actionability(element: Element, interaction: Interaction): Actionability;
+
+type Interaction = 'click' | 'type' | 'clear' | 'submit' | 'hover' | 'drop' | 'screenshot';
+
+type PendingWork = {
+  quiet: boolean;
+  observed: boolean;   // false when no preload script ran: history is unavailable (decision 9)
+  active: { timers: number; intervals: number; animationFrames: number;
+            requests: number; sockets: number; animations: number };
+  inert:  { timers: number; intervals: number; animationFrames: number };
+};
+
+type Settledness = {
+  settled: boolean;
+  quietForMs: number;
+  pendingWork: PendingWork;
+  mutating: Array<{ path: string; sinceMs: number;
+                    reason: 'mutation' | 'animation' | 'transition' }>;
+};
+
+type Actionability = {
+  actionable: boolean;
+  checks: Record<'visible' | 'enabled' | 'editable' | 'inViewport' | 'unobstructed' | 'stable',
+                 boolean | null>;   // null where the check does not apply to this interaction
+  interactionPoint: { x: number; y: number } | null;
+  obstructedBy: Element | null;
+};
+```
+
+`Settledness` nests `PendingWork` because the ledger is its input (decision 1); `Actionability`
+nests neither, which is what makes the interaction path cheap. Every export is synchronous and
+returns a decided answer for the instant it is called — the waiting is the caller's, so the same
+export serves a binding's internal wait, a user's snapshot and a `WebDriverWait` predicate. Bindings
+pass arguments and marshal results; no binding adds logic of its own. One artifact is what holds the
+heuristics together: what counts as meaningful mutation, when a spinner is inert and how obstruction
+is hit-tested share definitions, will not stay identical across five hand-written ports, and are one
+thing to hand to the BiDi working group. The exports are what keep a caller from taking on the parts
+it did not ask for.
 
 **9. The atom is installed as a preload script, and degrades explicitly.** Registration is
 per-binding plumbing that must never affect navigation: if it fails, the failure is logged once and
@@ -184,6 +244,65 @@ asynchronously by a periodic callback are misattributed, and workers, `MessageCh
 server-sent events, WebTransport, IndexedDB and canvas-only animation frames are not tracked.
 Selenium takes this to the WebDriver BiDi working group; if it lands, the bindings keep their API,
 and drop the polyfill.
+
+### Illustrative usage
+
+Names and casing are per-binding and not fixed by this record; the calls below show the shape of
+each access route.
+
+The primary route is that there is no call. A session asks for readiness once, and interaction code
+is unchanged:
+
+```python
+options = ChromeOptions()
+options.enable_bidi = True
+options.interaction_readiness = True      # se:interactionReadiness
+driver = webdriver.Chrome(options=options)
+
+driver.get("https://example.com/app")
+driver.find_element(By.ID, "submit").click()   # waits for actionability, then dispatches
+```
+
+When it expires, the error is the one the command raises today, and the subclass carries the reason:
+
+```python
+try:
+    driver.find_element(By.ID, "submit").click()
+except ElementClickInterceptedException as e:
+    # ReadinessTimeoutError subclass: "not actionable after 10s: unobstructed=False,
+    # obstructed by div.modal-backdrop at (412, 260)"
+    print(e)
+```
+
+A snapshot composes with the waiting strategy the user already has, and answers "what is it
+waiting for" without one:
+
+```python
+WebDriverWait(driver, 10).until(
+    lambda d: d.find_element(By.ID, "submit").actionability("click").actionable
+)
+
+state = driver.script.settledness()
+if not state.settled:
+    print(state.pending_work.active, [m.path for m in state.mutating])
+```
+
+A handler reports settling as it happens, on the `add_dom_mutation_handler` pattern:
+
+```java
+driver.script().addSettledHandler(settled ->
+    log.fine("settled after " + settled.getQuietForMs() + "ms"));
+```
+
+Settledness as a page load strategy, per session and overridden for one action:
+
+```python
+options.page_load_strategy = "settled"     # consumed locally; "normal" goes on the wire
+driver.get("https://example.com/dashboard")
+
+driver.get("https://example.com/stream", wait="none")   # never settles; do not wait for it
+driver.find_element(By.ID, "refresh").click(wait="settled")
+```
 
 ## Considered options
 
@@ -303,9 +422,9 @@ and drop the polyfill.
   capability off, and preload registration failing.
 - Follow-up decisions this makes necessary: whether readiness becomes the default for BiDi sessions
   and in which release; whether `ExpectedConditions.elementToBeClickable` and its equivalents are
-  deprecated; whether classic sessions ever get a reduced-fidelity variant; the concrete names and
-  signatures for the snapshot and handler surface; and the shape of the BiDi `quiescence` module
-  proposal.
+  deprecated; whether classic sessions ever get a reduced-fidelity variant; the per-binding
+  spelling of the snapshot and handler surface, whose placement, arguments and return shapes this
+  record fixes but whose names it does not; and the shape of the BiDi `quiescence` module proposal.
 
 ## Appendix
 


docs/decisions/17886-interaction-readiness-waits.md
```diff
@@ -0,0 +1,234 @@
...
+identical in five bindings, or the wait becomes another source of cross-binding divergence rather
+than a fix for one.
+
+## Decision
+
+Selenium waits for interaction readiness before it acts. On a BiDi-enabled session this is the
+default behavior of the ordinary interaction commands; the semantics are defined once, implemented
+once, and also exposed as explicit waits.
+
+**1. Readiness is modelled in three composable layers.**
```
titusfortner: These aren't three composable layers: settled builds on pending work, but actionable is completely independent (which is what decision 2 depends on). Conflating these makes some of the other decisions less clear.

docs/decisions/17886-interaction-readiness-waits.md
```diff
@@ -0,0 +1,234 @@
...
+- Status: Proposed
+- Discussion: https://github.com/SeleniumHQ/selenium/pull/17886
+
+## Context
+
+The single most common cause of flaky Selenium tests is acting on an element that is present but
+not yet ready: the page is still hydrating, a spinner is still animating, a modal is fading in over
+the button, a list is re-rendering under the pointer. WebDriver's own actionability checks run at
+the moment of the command — click scrolls into view and verifies the in-view centre point is
+pointer-interactable — and then either succeeds or throws. They do not wait, and they do not
```
titusfortner: Couple important clarifications here. The spec does say to wait for interactability on send keys and clear, using the implicit wait timeout. Additionally, Chrome waits for the element to be displayed before clicking, so we can't assume that nothing is happening currently with an implicit wait value set.

docs/decisions/17886-interaction-readiness-waits.md
```diff
@@ -0,0 +1,234 @@
...
+- *DOM settledness* — no *meaningful* mutation for a settle window. Meaningful excludes attribute
+  churn that cannot affect layout, and excludes mutation confined to a region annotated inert or
+  detected as periodic noise. Running CSS animations and transitions count as activity. Waiting can
+  be scoped to a subtree.
+- *Element actionability* — visible, enabled (including `aria-disabled` and the `fieldset`/`legend`
+  exception), editable when the planned interaction writes, in the viewport, not obstructed at its
+  interaction point, and not moving. Which checks apply depends on the interaction: `type` and
+  `clear` additionally require editable; `drop` and `screenshot` do not require enabled.
+
+**2. Interaction commands wait for actionability by default when the session has BiDi enabled.**
```
titusfortner: Waiting for actionable will stack with the interactability waits remote ends already run under the implicit wait timeout as I mentioned above. I don't think we can make this default behavior without causing problems for users. As such, I think it makes the most sense to replace implicit wait setting with this behavior because we don't want to  send an implicit wait value to the remote end if we are also executing this actionable wait. But we'll still need to keep implicit wait value at 0 by default.

docs/decisions/17886-interaction-readiness-waits.md
```diff
@@ -0,0 +1,234 @@
...
+explicit, and inform the diagnosis when a wait fails.
+
+The low-level Actions API does not auto-wait. It exists to dispatch an exact input sequence, and
+inserting waits between its ticks would change what the sequence means.
+
+This is capability-gated behavior, not a protocol-shaped API. No signature changes, no BiDi type is
+named, nothing new is reachable off the driver — 17670 constrains what the API *says*, and it stays
+silent about BiDi here.
+
+**3. A new `readiness` session timeout bounds the implicit wait, and failures keep their current
```
titusfortner: Agree failures should keep their current type ,and since the timeout raises client-side we could even use per-type subclasses so readiness timeouts are distinguishable without breaking existing rescues. As per decision 2 comment I don't think we want a new toggle or timeout value.

docs/decisions/17886-interaction-readiness-waits.md
```diff
@@ -0,0 +1,234 @@
...
+type.** Default 10 seconds; `0` disables implicit waiting for that session, as does the
+session-level toggle. When the wait expires the command raises what it raises today —
+`ElementClickInterceptedException` when the element is obstructed,
+`ElementNotInteractableException` when it is not visible, enabled or editable — with the diagnosis
+appended to the message. Existing `catch`/`except` blocks keep working; a test that fails today
+fails the same way, later and with a better message.
+
+**4. Two explicit entry points remain, both on the driver, both protocol-neutral.** Names follow
```
titusfortner: The pattern of driver.method(element) breaks object orientation, and would work better either acting on the element object, or in the driver.script domain. The latter seems the easier change, especially since I don't think we want either of these methods to be the primary way users access this functionality. I'd much rather see us integrate them better into our existing API, and then provide more flexibility beyond just the remote-end controlled wait.  Maybe we can discuss what these might look like on Slack.

docs/decisions/17886-interaction-readiness-waits.md
```diff
@@ -0,0 +1,234 @@
...
+driver.waitUntilActionable(element, interaction, timeout)              → readiness result
+```
+
+These are the way to reach layers 1 and 2, the way to wait on a classic session, and the way to
+wait for readiness without interacting. `waitUntilActionable` composes the layers: settle the
+document, then wait for the element, scrolling it into view once if it starts outside the viewport.
+The pending-work ledger is an input to the other two (`requirePendingQuiet`), not a third public
+entry point — "are there pending timers?" is not a question users should have to ask.
+
+**5. An explicit wait that times out raises the binding's existing timeout error, and the error
```
titusfortner: I don't think we want to limit exposure to this functionality to two remote-end managed wait methods. The plan for script module is to have a synchronous script execution method and then support async behavior with handlers. So the examples in your python repo won't fit what we're planning to support. 

I think we want to provide more powerful access to settledness via a callable method similar to add_dom_mutation_handler, and there is a lot of value to providing access to a snapshot of the record for any client library to manage with their own waiting strategies instead of forcing a single option.

docs/decisions/17886-interaction-readiness-waits.md
```diff
@@ -0,0 +1,234 @@
...
+carries the diagnosis.** Consistency with `WebDriverWait` matters more than the convenience of a
+boolean. The error message names what was still active: the blocking work classes, the
+still-mutating regions, or for an element the specific failing check and the obstructing element.
+"Timed out after 10s" is what makes current waits unhelpful; this is the main thing the feature has
+to fix. A non-blocking inspection call returning the same snapshot without waiting is available for
+diagnostics and for users building their own conditions.
+
+**6. One implementation, shared by all bindings.** The readiness oracle is a single JavaScript
```
titusfortner: This is conflating a shared implementation with a single atom combining all the functionality described here. We can share the functionality with multiple atoms that perform specific functions with specific arguments that are directly applicable so the bindings don't need to add additional logic beyond passing arguments to the atom.

AutomatedTester: Agreed, and updated in 5a48bdf98c after talking this through offline.

The record now proposes **one shared atom that exposes narrow named exports** rather than several separate atoms. One artifact keeps the heuristics that share definitions together — what counts as meaningful mutation, when a spinner is inert, how obstruction is hit-tested — and stays a single thing to hand to the BiDi working group. The per-signal exports are what keep a caller from taking on the parts it did not ask for: pending work for a document, settledness given a root and a settle window, actionability given an element and an interaction, each taking only its own arguments and returning only its own result. A binding that wants an actionability snapshot neither constructs arguments for settledness nor interprets a record containing it, so there is still no place for per-binding logic to creep back in.

Decision 8 is rewritten and the considered options now carry all three shapes explicitly — combined entry point (rejected), separate atoms per signal (rejected, since artifacts that build and version apart are the divergence this record exists to prevent), named exports (accepted).

docs/decisions/17886-interaction-readiness-waits.md
AutomatedTester: You are right, and the corrected version is a stronger Context than the claim I had. Rewritten in a9f895c.

I checked the spec text rather than take it from memory, and the detail matters for the argument:

| Command | Waiting behavior |
|---|---|
| Element Clear | steps 6–10: take the implicit wait timeout, start a timer, "wait for element to become *interactable*" |
| Element Send Keys | steps 7.2–7.6: the same pattern, "wait for element to become *keyboard-interactable*" |
| Element Click | steps 5–7: scroll into view, then fail. No wait step |

So "they do not wait" was wrong for clear and send keys. What survives is the predicate. *Keyboard-interactable* is "any element that has a focusable area, is a `body` element, or is the document element" — invisible, covered by a modal, mid-animation and `aria-disabled` all satisfy it. *Interactable* is "pointer-interactable or keyboard-interactable", so clear's wait is satisfied the same way.

The Context now argues the gap as: the predicate is too weak to prevent the failures users hit, it is absent for click, it says nothing about why it expired, it shares one budget with element location, and what happens beyond it differs per remote end. Your chromedriver point is in there too, as evidence that remote ends already improvise here — which reads as an argument for defining it once rather than against defining it at all.

docs/decisions/17886-interaction-readiness-waits.md
AutomatedTester: Conceded — and you were right that it was leaking into the other decisions. Fixed in a9f895c.

Decision 1 is now **two independent signals**, with the pending-work ledger demoted to an *input* of settledness rather than a peer:

- *Page settledness* — no meaningful DOM mutation for a settle window, with the ledger (timers, intervals, animation frames, `fetch`, `XHR`, WebSocket) as its input.
- *Element actionability* — visible, enabled, editable when the interaction writes, in the viewport, not obstructed at its interaction point, not moving.

The record now states outright that actionability does not depend on settledness and is not derived from it, and that this independence is what makes decision 2 affordable: the interaction path consults actionability only. An app that long-polls or animates continuously would otherwise pay a settle timeout on every click.

docs/decisions/17886-interaction-readiness-waits.md
AutomatedTester: Both halves taken, in a9f895c.

Decision 3 now keeps the existing types with readiness-specific **subclasses**: obstruction raises a subclass of `ElementClickInterceptedException`, not-visible/not-enabled/not-editable a subclass of `ElementNotInteractableException`, each carrying the diagnosis in its message. Existing `catch`/`except`/`rescue` keeps matching; anyone who wants to tell "still not ready after the budget" from "not ready at the instant of the command" can catch the subclass. Raising client-side is what makes the subclass possible at all, which is worth stating since it is the only reason this is available to us.

The new `readiness` session timeout and the session-level toggle are both deleted. Also dropped a divergence I had recorded against the prototype: blocking waits raise, snapshots return records, so there is no "returns a result object instead of raising" inconsistency left to reconcile.

docs/decisions/17886-interaction-readiness-waits.md
AutomatedTester: Agreed, and we went further than moving it. Landed in a9f895c, with signatures added in 2fd6078.

There is now **no new wait method on the driver or on the element**. The wait lives inside `click`, `send_keys`, `clear` and `submit`, where the user does not have to know it is there — which is your "integrate it into the existing API" point rather than a second blessed entry point.

Placement of the state follows the object model, with one rule stated explicitly so it cannot drift back: document state on the `script` domain, element state on the element.

```python
driver.script.pending_work()          # document
driver.script.settledness()           # document
element.actionability("click")        # element
```

The record fixes arguments and return shapes but deliberately leaves the per-binding spelling of the names open — happy to take that to Slack, along with the per-action override shape in decision 7, which currently reads as a keyword argument and may not survive contact with Java and .NET.

docs/decisions/17886-interaction-readiness-waits.md
AutomatedTester: Agreed on both counts, and it turns out to be cheap because the pattern already exists.

Decision 5 now exposes the state as **snapshots and events, not only as waits**:

- a snapshot returning the current record without waiting — pending work classes, still-mutating regions, or for an element the per-check result and the obstructing element — so it composes with whatever waiting strategy a client already has, including inside an existing `WebDriverWait`;
- a settledness handler following the existing `driver.script.add_dom_mutation_handler` path (`java/src/org/openqa/selenium/remote/RemoteScript.java:85`) — preload script, `ChannelValue`, `script.onMessage`. That is the same wiring the prototype's observer already uses, with a different payload, so this is an existing pattern rather than new machinery.

Signatures went in with 2fd6078. On the synchronous script module: nothing here needs an async evaluate — the atom exports are synchronous functions, so a snapshot is one synchronous call and the async side is handlers only, which lines up with where you are taking that module. The prototype's Python examples do not fit that plan and are evidence only; the record now says so in the appendix.

docs/decisions/17886-interaction-readiness-waits.md
AutomatedTester: The stacking defect is real and it changed the design. Landed in a9f895c: readiness is off by default behind an `se:interactionReadiness` capability, Selenium stops forwarding a non-zero implicit wait to the remote end when it is on, and the budget is the implicit wait value when set with a documented 10s default otherwise — so there is no second timeout and no new toggle. The default-on proposal is recorded as retreated from, and re-filed as a follow-up decision, probably release-gated as you suggested.

**On your later point about `locateNodes` — the premise checks out, and it is worse than you put it.** From the spec source:

```cddl
browsingContext.LocateNodesParameters = {
   context: browsingContext.BrowsingContext,
   locator: browsingContext.Locator,
   ? maxNodeCount: (js-uint .ge 1),
   ? serializationOptions: script.SerializationOptions,
   ? startNodes: [ + script.SharedReference ]
}
```

No wait, no timeout, no retry. Beyond that, the string `timeout` appears **twice in the whole of webdriver-bidi**, both TODO notes about a possible future `script.evaluate` timeout. BiDi has no timeout concept at all. So there is nothing for the remote end to drive an implicit wait with, and I agree that kills "actionability *is* the BiDi implicit wait" — they do different jobs and my current decision 2 is wrong on its own terms.

**Where I would push back is on the conclusion that implicit wait must therefore be unsupported.** The protocol cannot provide it; that does not stop us, and the machinery this record already proposes is a better vehicle than the remote end ever was: subscribe to the mutation signal the preload observer already emits, and re-run `locateNodes` only when the DOM actually changed, instead of polling on a fixed interval. That is a handful of round trips over a socket that is already open rather than N per 100ms, and it fails with a real diagnosis instead of a bare `NoSuchElementException`. The alternative — reimplementing locators inside the atom so the retry is fully client-side — I would reject, since XPath, inner text and accessibility locators would drift from the remote end's implementations.

Dropping implicit wait instead is a defensible strategic call, and I would rather it were chosen as a deprecation than accepted as a protocol casualty, because the migration cost lands on the wrong people: someone who enables BiDi for network interception and happens to have `implicitly_wait(10)` set silently loses element-location retry. That is a worse break than the stacking it fixes.

**One consequence worth naming before you finish thinking it through:** if implicit wait is no longer the budget, readiness needs its own timeout value — the knob you asked me to delete on the decision 3 comment. Not a contradiction, just a consequence, and there are three ways out: repurpose the now-free implicit wait setting as the readiness budget (I would reject that — same name, changed meaning), a new readiness timeout with implicit wait deprecated on BiDi, or a new readiness timeout with implicit wait implemented client-side as above. I would take the third. Which do you want?

Edge cases I would want on the list, since you said you were still working through them:

- **The runtime setter.** `timeouts().implicitlyWait()` on a BiDi session: raise, warn, no-op, or silently become the readiness budget? Silent is the worst option and the easiest to ship by accident.
- **A non-zero default versus negative tests.** Your "maybe a major release?" concern survives the stacking fix — suites asserting an element is *not* interactable now wait N seconds before failing. A timing regression rather than a correctness one, but real.
- **Stale references.** Ordering between the readiness wait and `StaleElementReferenceException` for an element found before a navigation.
- **Partial BiDi.** `webSocketUrl` present but some commands still routed over classic, or Grid in between — does the capability alone gate the behavior, or per-command transport?
- **The Actions API.** The record currently says it does not auto-wait, on the grounds that it exists to dispatch an exact input sequence. Worth confirming you agree, because it is where Playwright puts real effort and it is the obvious "why not here too" question from users.

I am deliberately not rewriting decision 2 a third time until you land, since you flagged the position as unfinished.

PR #17893, opened 2026-08-09
[rb] Lock the starting port in a file rather than on the port below it

ikraamg:
### 🔗 Related Issues

Picks up #10176, which was closed with "if anyone wants to PR a better solution, we can do that".

### 💥 What does this PR do?

ServiceManager takes its startup lock by binding port - 1, so starting a driver on port N needs N-1 to be free as well. When an unrelated service already listens there the lock can never be taken: startup spins for 45 seconds and then raises, even though the requested port is free. This is easy to hit once you pick driver ports yourself, for example a pool assigning a port per process that lands next to a running Redis. I found it running the Firefox WebDriver BiDi render pipeline for trmnl.com in production.

With a listener on 9998 and a driver requested on 9999:

    before   FAILED after 45.1s: unable to bind to locking port 9998 within 45 seconds
    after    started in 0.0s on port 9999

The lock now lives in a file under Dir.tmpdir named after the starting port, so it needs no port of its own.

- replaces SocketLock with PortLock, which flocks tmpdir/selenium-port-<port>.lock
- locks on the requested port instead of the one below it, so the name matches what is being guarded
- renames SOCKET_LOCK_TIMEOUT to PORT_LOCK_TIMEOUT and ServiceManager#socket_lock to #port_lock, both @api private
- drops the Steepfile ignore that existed only for socket_lock.rb
- creates the lock file 0600, since nothing but the creating process ever opens it
- adds rb/spec/unit/selenium/webdriver/common/port_lock_spec.rb

Cross-process exclusion is unchanged, covered by a spec that holds the lock and asserts a second lock on the same port is refused while a lock on a different port goes through.

### 🔧 Implementation Notes

The lock only ever needed to be a name two processes could agree on, and binding a TCP port to get one is what drags an unrelated port into the requirements. A file in tmpdir gives the same mutual exclusion without consuming anything.

Keeping the socket lock and binding the requested port instead cannot work, since the lock is held across find_free_port and the port being probed has to stay free. Picking a lock port further away only moves the collision somewhere less predictable, and skipping the lock when a port was passed explicitly drops the protection for the case it exists for, several processes starting drivers at once.

The retry came out of CI. My first version opened the file read-only when it could not be opened for writing, which is fine on POSIX, but on Windows File.open raises Errno::EACCES while another process holds the lock, and an exclusive lock cannot be taken on a read-only handle there. That turned ordinary contention into a permanent spin and failed all ten Windows targets. EACCES is now treated as the lock not being available yet, and retried.

EROFS is not, since a read-only temp directory never becomes writable. Retrying it spun for the full 45 seconds and then blamed the lock, where the socket lock this replaces did not touch the filesystem at all, so it raises straight away and names the path. The spec for it takes 0.01s with the raise and 2.0s without, which is the timeout it used to wait out.

### 🤖 AI assistance

    - Tool(s): Claude Code
    - What was generated: a first draft of PortLock and its spec. The problem, the design, the measurement and the Windows fix are mine.

- No cross-user exclusion. If the lock file belongs to another user this treats it as contention until the timeout. Trunk has the same gap, TCPServer.new on a port another user holds fails the same way.
- The lock file stays in tmpdir after release. It is empty and reused.
- PortLock is @api private, as SocketLock was, so nothing public changes.

### 🔄 Types of changes

### DIFF
--- a/rb/spec/unit/selenium/webdriver/common/port_lock_spec.rb
+++ b/rb/spec/unit/selenium/webdriver/common/port_lock_spec.rb
@@ -24,7 +24,9 @@ module Selenium
     describe PortLock do
       subject(:port_lock) { described_class.new(port, 2) }
 
-      let(:port) { 4444 }
+      # The lock file is named after the port, so a fixed one would collide with a real
+      # driver on 4444 and with any other run of this spec on the same machine.
+      let(:port) { 40_000 + (Process.pid % 10_000) }
 
       it 'yields to the block' do
         expect { |block| port_lock.locked(&block) }.to yield_control


diemol:
Can you use the PR template?

PR #17894, opened 2026-08-09
[rb] Release a stopped service instead of holding it until the process exits

ikraamg:
### 🔗 Related Issues

None open that I could find.

### 💥 What does this PR do?

`ServiceManager#start` registers an `at_exit` block per service, and the block captures the service, so every service ever started stays reachable for the life of the process along with its `ChildProcess`. Stopping the service does not release it, so the growth is unbounded in a process that starts many drivers: a suite that starts one per spec file, or a pool that recycles browsers.

I found it running the Firefox WebDriver BiDi render pipeline for trmnl.com in production.

Starting 500 services and stopping every one of them, then running GC:

    before   live ServiceManagers: 500
    after    live ServiceManagers: 1

One retained `ServiceManager` holds 720 bytes across 12 objects once you follow the whole reachable graph, with a real `ChildProcess` attached, after a normal `stop`.

Services are now tracked in one list with a single exit hook per process, and `#stop` removes the service from it. A service that is still running is still held, which is what lets the exit hook stop it.

- adds `ServiceManager.track`, `.untrack` and `.stop_running`
- registers the exit hook on first `start` rather than at load, so a service started in a forked child is still stopped when that child exits
- clears the list when the pid changes, so a child does not stop services belonging to its parent
- adds `rb/spec/unit/selenium/webdriver/common/service_manager_spec.rb`

### 🔧 Implementation Notes

The per-service `at_exit` was doing two jobs: making sure a running service gets stopped, and, as a side effect, keeping the service alive to do it. Only the first is wanted. A single class-level list gives the exit hook everything it needs to stop, and `#stop` removing itself from that list is what lets a stopped service be collected.

Alternatives I considered:

- a `WeakRef` or an `ObjectSpace::WeakMap` of services. A stopped service would be collected, but so could a running one before the process exits, which is the case the hook exists for.
- unregistering the individual `at_exit` block on `stop`. Ruby has no API for that.
- leaving `start` alone and having `stop` clear the service's own state so the retained object is small. It shrinks the leak instead of removing it, and still grows without bound.

Registering the hook lazily on the first `start` is the part I would look at hardest. Registering at load would mean a process that forks after loading selenium-webdriver has the hook only in the parent, so a driver started in the child is never stopped. Arming it inside `claim_for_this_process`, which also clears the list when the pid has changed, gives the child its own hook and its own list. I verified this with a real `fork`: with a service started in the parent and another in a forked child, the child tracks only its own and the parent is unaffected.

The list is guarded by a mutex because drivers are commonly started from several threads, and `stop_running` iterates a copy so a service stopping itself during the walk cannot mutate the list underneath it.

### 🤖 AI assistance

    - Tool(s): Claude Code (Claude Opus)
    - What was generated: a first draft of the tracking methods and the spec. The leak, the retained-size measurement and the fork behaviour are mine, and the fork case was found by testing rather than by the draft.

- `stop_running` is public because the exit hook calls it, and it is useful to a pool that wants to shut everything down; happy to make it private if you would rather not add surface.
- No behaviour change for the ordinary case: services still get stopped at exit, and a service that is already stopped is simply no longer stopped twice.

### 🔄 Types of changes

### DIFF
--- a/rb/sig/lib/selenium/webdriver/common/service_manager.rbs
+++ b/rb/sig/lib/selenium/webdriver/common/service_manager.rbs
@@ -55,7 +55,7 @@ module Selenium
 
       def self.untrack: (ServiceManager manager) -> void
 
-      def self.claim_for_this_process: () -> void
+      private def self.claim_for_this_process: () -> void
 
       def initialize: (untyped config) -> void
 


diemol:
Can you use the PR template and explain what is being done here? Letting AI do the work is not enough for us to accept a PR.

titusfortner:
Ah, why did you close this? I prepared some feedback on it if you still want to pursue it, it's a needed improvement.

ikraamg:
Hey @titusfortner, I am happy to continue as you think its needed. I thought it was a minor leak and did not want to waste any repo maintainer time on it.

titusfortner:
Well, maybe I'm getting wires crossed on issues, leave it open until I can review it tomorrow at least.

PR #17895, opened 2026-08-10
[grid] Embed Kubernetes node in server jar without fabric8

VietND96:
## Motivation

Dynamic Grid on Kubernetes ships as a separate `--ext` extension jar (`selenium-node-kubernetes`) to keep the fabric8 client tree out of the lightweight server jar. That makes it hard to **early-test the feature against the nightly `selenium-server.jar`**, because no matching nightly extension artifact is published (and one cannot be added to the nightly release artifacts).

This change lets the K8s node be exercised from the nightly server jar directly, while still keeping fabric8 out of the jar.

## What this does

- Embeds the `selenium-node-kubernetes` **classes** (and their `@AutoService` `NodeSessionFactoryProvider` registration) into `selenium-server.jar`.
- Keeps the **fabric8** client libraries out of the jar by marking them compile-only (`neverlink`). fabric8 is supplied at runtime via `--ext`.
- When Kubernetes is enabled without fabric8 on `--ext`, the provider now throws a clear `ConfigException` pointing at `--ext` instead of an opaque `NoClassDefFoundError`. fabric8 is probed by name so the provider stays fabric8-free and `ServiceLoader`-instantiable when fabric8 is absent.

The `KubernetesSessionFactoryProvider` is lazy — `isEnabled()` only reads config and fabric8 is not touched until `loadFactories()` runs (i.e. only when `[kubernetes] configs` is set). So nodes that don't use Kubernetes start unaffected even without fabric8 on the classpath.

The existing `kubernetes` `java_export` (real fabric8, published maven artifact + docker `--ext` jar) is unchanged; only the server binary's `runtime_deps` gain the new compile-only-fabric8 variant. The published `selenium-grid` artifact is unchanged.

## How to test against a server jar

```
java -jar selenium-server-<version>.jar \
  --ext $(coursier fetch -p \
    io.fabric8:kubernetes-client:7.8.0 \
    io.fabric8:kubernetes-client-api:7.8.0 \
    io.fabric8:kubernetes-model-batch:7.8.0 \
    io.fabric8:kubernetes-model-core:7.8.0) \
  node -K selenium/standalone-chrome:latest '{"browserName":"chrome"}'
```

Only public Maven Central fabric8 is fetched — no separate Selenium extension artifact required.

## Jar package: before vs after

Comparing the official `selenium-server-4.46.0.jar` release asset with the `selenium_server_deploy.jar` built from this branch:

| | Last release — 4.46.0 | This PR — 4.47.0-SNAPSHOT |
|---|---|---|
| Jar size | 51 MB | 52 MB |
| k8s node classes (`grid/node/kubernetes/*`) | 0 — absent | 8 classes (+ package dir) |
| fabric8 classes (`io/fabric8/*`) | 0 — absent | **0 — absent** |
| `NodeSessionFactoryProvider` SPI registration | none | `…kubernetes.KubernetesSessionFactoryProvider` |

Net package delta: **+8 k8s class entries + 1 SPI registration, +0 fabric8** (≈ +1 MB).

## Verification

Inspected `selenium_server_deploy.jar`:

- ✅ All k8s classes present (`KubernetesSessionFactoryProvider`, `KubernetesOptions`, `KubernetesSession`, …)
- ✅ **0 fabric8 entries** in the jar (`neverlink` excluded them)
- ✅ `META-INF/services/…NodeSessionFactoryProvider` registered → `KubernetesSessionFactoryProvider`
- ✅ Provider instantiates via `ServiceLoader` with **no fabric8 on the classpath** (nodes boot fine when K8s is unused)
- ✅ Enabling Kubernetes with no fabric8 throws `ConfigException` (not `NoClassDefFoundError`), with an actionable `--ext` message — covered by a new mockless unit test (`KubernetesSessionFactoryProviderTest`)

### Test coverage for the embedded mode

`KubernetesEmbeddedWithoutFabric8Test` runs against `:kubernetes-embedded` with **no fabric8 on the runtime classpath** — the actual "embedded classes + external fabric8 via `--ext`" deployment mode (existing k8s tests run *with* fabric8). It asserts:

- fabric8 is genuinely off the runtime classpath (premise guard — fails if fabric8 ever leaks into that target's deps)
- the provider is `ServiceLoader`-discoverable and instantiable without fabric8 (guards against the provider touching fabric8 during discovery)
- `isEnabled` works without fabric8
- `loadFactories` fails with the actionable `ConfigException`, not a linkage error

## Notes

- **Packaging change**: this puts the k8s *classes* (not fabric8) into `selenium-server.jar` — a deliberate shift from the "fully external extension" model. No binding changes: all bindings consume the same server jar unchanged.

### DIFF
--- a/java/src/org/openqa/selenium/grid/node/kubernetes/KubernetesSession.java
+++ b/java/src/org/openqa/selenium/grid/node/kubernetes/KubernetesSession.java
@@ -17,6 +17,7 @@
 
 package org.openqa.selenium.grid.node.kubernetes;
 
+import io.fabric8.kubernetes.api.model.DeletionPropagation;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.LocalPortForward;
@@ -98,7 +99,18 @@ public class KubernetesSession extends DefaultActiveSession {
 
   private void deleteJob() {
     try {
-      kubeClient.batch().v1().jobs().inNamespace(namespace).withName(jobName).delete();
+      // Background propagation: return as soon as the Job is marked for deletion and let K8s reap
+      // the Pod (and its containers) asynchronously. Stopping a session must not block the Node
+      // waiting on Pod teardown. Set explicitly so the cascade is documented rather than relying on
+      // the client default, matching the intent of the factory's own Job deletion.
+      kubeClient
+          .batch()
+          .v1()
+          .jobs()
+          .inNamespace(namespace)
+          .withName(jobName)
+          .withPropagationPolicy(DeletionPropagation.BACKGROUND)
+          .delete();
     } catch (KubernetesClientException e) {
       LOG.log(
           Level.WARNING,
--- a/java/test/org/openqa/selenium/grid/node/kubernetes/KubernetesSessionFactoryProviderTest.java
+++ b/java/test/org/openqa/selenium/grid/node/kubernetes/KubernetesSessionFactoryProviderTest.java
@@ -29,9 +29,12 @@ class KubernetesSessionFactoryProviderTest {
 
   @Test
   void requireFabric8PassesWhenClientIsOnClasspath() {
-    // The test classpath includes the fabric8 client, so the probe resolves.
+    // The test classpath includes the fabric8 client, so the probe resolves. Use the same
+    // classloader expression as the production call site in loadFactories().
     assertThatCode(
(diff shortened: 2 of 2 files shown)


VietND96:
@diemol, maybe you're confused. 1MB here is +8 k8s class entries of K8s Dynamic Grid + 1 SPI registration, which is included in jar now (instead of loading via --ext module `selenium-node-kubernetes` downloaded from maven central using `coursier`).
However, the K8s client library is still excluded from the server JAR, users need to load it via --ext. We will not package this, as the server JAR size will double.

PR #17931, opened 2026-08-20
Honour the `DO_NOT_TRACK` env var

shs96c:
Rather than having multiple different mechanisms per app for handling opting out of tracking, there's a move to standardise on this env var (https://donottrack.sh/) Selenium should honour that.


AutomatedTester (approved):

cgoldberg:
nice.. I already have that set in my shell config :)

Issue #14694, opened 2024-10-31
[🐛 Bug]: AttributeError: 'AppiumConnection' object has no attribute '_ca_certs'


Trouler:
### What happened?

We are automating actions on both an Android phone on Windows, and on a Iphone on macOS. The macOS automation stopped working in the latest update, with this error:
AttributeError: 'AppiumConnection' object has no attribute '_ca_certs'

It's happening when attempting to initiate a Remote WebDriver

### How can we reproduce the issue?

```shell
Will be removing quite a lot of internal code here, but hopefully relevant code is enough to figure out why it happens.
---------------------------------------------------------

from appium.webdriver import Remote
from appium.options.ios import XCUITestOptions
from appium.webdriver.appium_service import AppiumService

options = XCUITestOptions()
options.set_capability("newCommandTimeout", 3000)
options.device_name = "SECRET"
options.platform_version = "16.4"
options.automation_name = "XCUITest"
options.show_xcode_log = True
options.xcode_org_id = "SECRET"
options.xcode_signing_id = "iPhone Developer"
options.udid = "SECRET"

host = "localhost"
port = "4723"

appium_service = AppiumService()
appium_service.start(args=["--address", host, "--port", port], main_script="/usr/local/bin/appium")
assert appium_service.is_running
driver = Remote(f"http://{host}:{port}", options=options)
```

### Relevant log output

```shell
self.driver = Remote(f"http://{appium_host}:{appium_port}", options=options)
.venv/lib/python3.12/site-packages/appium/webdriver/webdriver.py:227: in __init__
    command_executor = AppiumConnection(command_executor, keep_alive=keep_alive)
.venv/lib/python3.12/site-packages/appium/webdriver/appium_connection.py:43: in __init__
    super().__init__(remote_server_addr, keep_alive=keep_alive, ignore_proxy=ignore_proxy)
.venv/lib/python3.12/site-packages/selenium/webdriver/remote/remote_connection.py:340: in __init__
    self._conn = self._get_connection_manager()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <appium.webdriver.appium_connection.AppiumConnection object at 0x1101eedb0>
    def _get_connection_manager(self) -> Union[urllib3.PoolManager, urllib3.ProxyManager]:
[...]

PR #17944, opened 2026-08-25
[py]: Apply ClientConfig `user_agent` and `extra_headers` per connection

navin772:
### 💥 What does this PR do?

`RemoteConnection.__init__` copied the connection's `user_agent` and `extra_headers` onto the **class**, not the instance:

```python
RemoteConnection.extra_headers = self._client_config.extra_headers or RemoteConnection.extra_headers
RemoteConnection.user_agent = self._client_config.user_agent or RemoteConnection.user_agent
```

Every `RemoteConnection` in the process then shared them, and they outlived `driver.quit()`. A driver pointed at an authenticated Grid followed by a local driver meant the local driver sent the Grid's `Authorization` header:

```python
a = webdriver.Remote(client_config=ClientConfig(url_a, extra_headers={"Authorization": "Bearer SECRET"}))
b = webdriver.Remote(client_config=ClientConfig(url_b))  # also sends Bearer SECRET
```

Now they're applied per-instance in `_request()` from `self._client_config`, so one connection's headers can't reach another. Class-level `RemoteConnection.extra_headers` / `user_agent` still work as process-wide defaults.

### 🤖 AI assistance

    - Tool(s): Claude code

### 🔄 Types of changes


Delta456 (approved):

navin772:
@cgoldberg could you review this PR?

cgoldberg (approved):
nice. LGTM

py/selenium/webdriver/remote/remote_connection.py
```diff
@@ -419,6 +419,13 @@ def _request(self, method, url, body=None) -> dict:
         """
         parsed_url = parse.urlparse(url)
         headers = self.get_remote_connection_headers(parsed_url, self._client_config.keep_alive)
+        # Apply this connection's own user-agent and extra headers from its client_config so
+        # per-connection configuration is not shared across RemoteConnection instances via
+        # class attributes.
+        if self._client_config.user_agent:
+            headers["User-Agent"] = self._client_config.user_agent
+        if self._client_config.extra_headers:
+            headers.update(self._client_config.extra_headers)
```
AutomatedTester: **Blocking — this is where the leak reverses direction rather than closing.**

The write moved per-instance but the read didn't. `get_remote_connection_headers` is a `@classmethod` still reading `cls.extra_headers` / `cls.user_agent`, and these lines then layer the instance config *on top of* its output. Trunk had one slot with replace semantics (`cfg.extra_headers or RemoteConnection.extra_headers`); this is two layers with union semantics. A process-wide class-level header now reaches every connection that has its own config — which trunk specifically prevented:

```python
RemoteConnection.extra_headers = {"Authorization": "Bearer GLOBAL-SECRET"}
b = RemoteConnection(client_config=ClientConfig(url_b, extra_headers={"X-Tenant": "A"}))
b._request("GET", url_b + "/status")
```

|  | headers on the wire |
| --- | --- |
| trunk | `X-Tenant: A` |
| this branch | `X-Tenant: A`, **`Authorization: Bearer GLOBAL-SECRET`** |

The description says class-level attributes "still work as process-wide defaults", but a default is what applies when nothing else is set — this makes them an unconditional floor.

Two further consequences of the same root cause, resolving *after* the hook instead of *inside* it:

**1. `get_remote_connection_headers()` is public and no longer reports what goes on the wire.** With `ClientConfig(user_agent="my-ua/1", extra_headers={"X-Tenant": "A"})` the hook returns `User-Agent: selenium/4.49.0…` and no `X-Tenant`, while the request sends `my-ua/1` and `X-Tenant: A`. On trunk the two agreed.

**2. It breaks the downstream that the comment a few lines above names (#14694).** appium-python-client's `AppiumConnection` sets `user_agent = f'appium/{library_version()} ({RemoteConnection.user_agent})'` and merges `cls.extra_headers` inside its override. Running that class verbatim against `ClientConfig(user_agent="my-app/1.0", extra_headers={"X-Idempotency-Key": "user"})`:

|  | User-Agent sent | X-Idempotency-Key |
| --- | --- | --- |
| trunk | `appium/5.2.6 (selenium/4.49.0 (python mac))` | Appium's uuid |
| this branch | `my-app/1.0` | `user` |

That's a public behaviour change in a PR labelled "Bug fix (backwards compatible)". Whether subclass class attributes *should* outrank `ClientConfig` is a fair question, but it should be a deliberate decision rather than a consequence of where the `headers.update()` landed.

**Smallest fix** — resolve once, with `or` (replace, matching trunk), into instance-owned state, and read it during header construction instead of patching over the result:

```python
# in __init__, in place of the two removed lines
self.user_agent = self._client_config.user_agent or type(self).user_agent
self.extra_headers = dict(self._client_config.extra_headers or type(self).extra_headers or {}) or None
```

The obstacle is that `get_remote_connection_headers` is a `@classmethod`, so `cls.*` cannot see instance state — which I assume is why the merge ended up in `_request`. A small class-or-instance descriptor solves it, and usefully keeps Appium working unchanged: their override is a `@classmethod`, so its zero-arg `super()` is class-bound and still reads its own subclass attributes exactly as on trunk.

```python
class _classorinstancemethod:
    def __init__(self, func):
        self.func = func
        functools.update_wrapper(self, func)

    def __get__(self, obj, cls=None):
        return functools.partial(self.func, obj if obj is not None else cls)
```

I prototyped that and re-ran every probe: the original two-driver leak stays fixed, the global-header leak above is gone, the hook matches the wire again, Appium's prefix survives, and `RemoteConnection.get_remote_connection_headers(url)` is still callable on the class. Adding keyword arguments to the classmethod instead is worse — it breaks Appium's override signature outright.

py/selenium/webdriver/remote/remote_connection.py
```diff
@@ -331,8 +331,8 @@ def __init__(
         RemoteConnection._timeout = self._client_config.timeout
         RemoteConnection._ca_certs = self._client_config.ca_certs
         RemoteConnection._client_config = self._client_config
-        RemoteConnection.extra_headers = self._client_config.extra_headers or RemoteConnection.extra_headers
-        RemoteConnection.user_agent = self._client_config.user_agent or RemoteConnection.user_agent
+        # user_agent and extra_headers are not mirrored onto the class: that leaked one
```
AutomatedTester: **The comment overclaims — the three lines above it still mirror onto the class.**

`_timeout`, `_ca_certs` and especially `_client_config` are still copied to the class, and `_client_config` carries `username` / `password` / `token` / `auth_type` *as well as* `extra_headers` / `user_agent`. So the description's "they outlived `driver.quit()`" is narrowed rather than fixed — the class still holds a live reference to the last-constructed connection's credentials after quit.

The deprecated class-level setters also still mutate whichever connection happened to be built last:

```python
a = RemoteConnection(client_config=cfg_a)   # timeout 111
b = RemoteConnection(client_config=cfg_b)   # timeout 222
RemoteConnection.set_timeout(9)             # -> cfg_b.timeout == 9, cfg_a untouched
```

Confirmed on both trunk and this branch. Same bug class, same root cause. Either drop those three as well (with a deprecation pointer, since #14694 is why they exist) or keep them and reword this comment — as written it reads as though the whole block had been cleaned up.

AutomatedTester (commented):
Thanks for digging into this — the leak is real and the two-driver repro in the description is exactly right. But I don't think this closes it, I think it reverses its direction.

I checked the claims below by running trunk, this branch (`83bd329`) and a prototype side by side rather than by reading the diff, so the tables are measured rather than inferred.

The framing I'd suggest: `RemoteConnection.extra_headers` / `user_agent` want to be **class-level templates**, handled the way capabilities already are in this codebase — `copy.deepcopy(caps)` in `_create_caps`, `DesiredCapabilities.CHROME.copy()` in `ChromiumOptions.default_capabilities`. A template gets copied by every consumer at construction and is never written back to. Trunk writes back to the template — that's the leak you found. This PR stops writing back, but leaves the template as a live layer underneath every request. Copying at construction takes it out of the request path altogether.

The copy is missing in a second place too, and it isn't in this diff — `ClientConfig.__init__` stores the caller's dicts by reference:

```python
self.init_args_for_pool_manager = init_args_for_pool_manager or {}
self.extra_headers = extra_headers
```

So two connections handed the same `ClientConfig` share one `extra_headers` dict, and anything that mutates it in place reaches back into the user's own dict. Appium does precisely that: `cls.extra_headers[HEADER_IDEMOTENCY_KEY] = str(uuid.uuid4())`. `dict(extra_headers) if extra_headers is not None else None` plus `dict(init_args_for_pool_manager or {})` covers it.

Two things I'd file separately rather than grow this PR:

- `ClientConfig.__init__` has `proxy: Proxy | None = Proxy(raw={"proxyType": ProxyType.SYSTEM})` — one instance, evaluated at import, shared by every `ClientConfig` ever constructed.
- `_request` follows a 3xx `Location` to an arbitrary host and re-sends `extra_headers` plus `get_auth_header()`. I confirmed `Authorization: Bearer SECRET` is sent on to the redirect target, on both trunk and this branch.

py/test/unit/selenium/webdriver/remote/remote_connection_tests.py
```diff
@@ -674,3 +683,36 @@ def test_get_remote_connection_selects_browser_specific_handler(
         ignore_local_proxy=True,
     )
     assert type(conn) is expected_handler
+
+
+def test_client_config_headers_do_not_leak_across_connections(monkeypatch):
+    """Per-connection user_agent/extra_headers must not be shared across RemoteConnection instances."""
+    # Reset any class-level pollution left by other tests so the default is deterministic.
+    monkeypatch.setattr(RemoteConnection, "extra_headers", None)
```
AutomatedTester, line 698: **This line nulls out the one interaction that is now broken.**

Setting `extra_headers` to `None` and `user_agent` to a fixed default is precisely the configuration in which the layering problem cannot show up, so the regression is invisible to the suite and CI stays green.

The case worth adding: a class-level `extra_headers` **and** a per-connection `ClientConfig(extra_headers=...)` both set, asserting the class-level entry is *not* on the wire.

```python
monkeypatch.setattr(RemoteConnection, "extra_headers", {"Authorization": "Bearer GLOBAL"})
conn = RemoteConnection(client_config=ClientConfig("http://localhost:4444", extra_headers={"X-Tenant": "A"}))
sent = _capture_sent_headers(conn)
conn._request("GET", "http://localhost:4444/status")
assert sent[0]["X-Tenant"] == "A"
assert "Authorization" not in sent[0]
```

One other thing worth guarding: `_capture_sent_headers` only replaces `self._conn`, which exists only when `keep_alive=True`. A config with `keep_alive=False` falls through to `_get_connection_manager()` and would attempt a real request instead of failing loudly.

cgoldberg (commented):
erm.. yeah what he said ^^ :)

AutomatedTester:
Thanks — this addresses all three findings from my earlier review. I re-ran the same probes
against trunk and this head rather than reading the diff, plus the module's own unit tests.

| behaviour | trunk | previous head | this head |
|---|---|---|---|
| conn A's `Authorization` reaches conn B | **yes** ✗ | no | **no** ✓ |
| process-wide class header forced onto a conn with its own config | no | **yes** ✗ | **no** ✓ |
| class-level header still applies when a conn sets none | yes | yes | **yes** ✓ |
| `conn.get_remote_connection_headers()` matches the wire | yes | **no** ✗ | **yes** ✓ |
| real `AppiumConnection`: `appium/` UA prefix + its idempotency key | yes | **no** ✗ | **yes** ✓ |
| caller's `extra_headers` / `init_args_for_pool_manager` aliased | **yes** ✗ | **yes** ✗ | **no** ✓ |
| two conns sharing one `ClientConfig` share one headers dict | **yes** ✗ | n/a | **no** ✓ |
| `mock.patch.object` on the hook still intercepts | yes | yes | **yes** ✓ |

I checked the Appium row against the real `appium-python-client` source rather than the
synthetic subclass in the test: `AppiumConnection` declares its own `extra_headers = {}`, so
its classmethod `super()` stays class-bound and reads its own attributes — trunk parity exactly.

Test suite: 94/94 pass on this head. Of the nine added/rewritten tests, four fail against
trunk (`test_client_config_headers_do_not_leak_across_connections`,
`test_class_level_extra_headers_are_a_default_not_a_floor`, and both `client_config` copy
tests), so they are genuine regression coverage; the other five pass on trunk as parity locks
against the previous approach. Nothing tautological.

Six small things below. None of them are blocking.

## 1. `functools.partial` loses five bits of the method's API surface

`_classorinstancemethod.__get__` returns a `functools.partial`, which is not a bound method:

| | trunk | this head |
|---|---|---|
| `type(...)` | `method` | `partial` |
| `inspect.ismethod(...)` | `True` | `False` |
| `__func__` / `__self__` | present | **absent** |
| `__name__` | `get_remote_connection_headers` | **absent** |
| `__doc__` | `"Get headers for remote request."` | **`"partial(func, *args, **keywords) - new function with partial application..."`** |

The `__doc__` one is user-visible: `help(RemoteConnection.get_remote_connection_headers)` now
prints `functools.partial`'s docstring instead of the method's. `functools.update_wrapper` on
the descriptor doesn't help, because the wrapper lives on the descriptor and the partial is
what callers actually receive.

One-line fix — `types.MethodType` is what `classmethod` itself hands back, so all five rows
return to the trunk column:

```diff
+import types
@@
     def __get__(self, obj, cls=None):
-        return functools.partial(self.func, obj if obj is not None else cls)
+        return types.MethodType(self.func, obj if obj is not None else cls)
```

I ran this variant: every behaviour the new tests assert still holds, including the
class-bound `super()` path for the Appium-shaped subclass, and `inspect.signature` stays
`(parsed_url, keep_alive=False)`.

## 2. Class-level headers set *after* a connection exists no longer reach it

`__init__` now always writes `self.extra_headers` / `self.user_agent`, so an instance
attribute exists from construction and shadows the class attribute for that connection's
lifetime:

```python
conn = RemoteConnection(client_config=ClientConfig(url))
RemoteConnection.extra_headers = {"Foo": "bar"}
conn._request("GET", url + "/status")   # trunk: sends Foo. this head: does not.
```

Set-before-construct still works, which is the documented pattern, so this may well be fine —
but it silently breaks refreshing a class-level token mid-session. Worth either a deliberate
note in the changelog/docstring or a test pinning whichever semantics you want. Right now
nothing covers it.

## 3. The `ClientConfig` copy has a second door

`__init__` copies, but assignment through `_ClientConfigDescriptor` does not, so the invariant
the two new tests assert only holds via the constructor:

```python
caller = {"X-Api-Key": "secret"}
cfg = ClientConfig(url)
cfg.extra_headers = caller
assert cfg.extra_headers is not caller   # fails, on trunk and on this head
```

Not a regression, and arguably out of scope — but if the goal is "caller-owned dicts can't be
aliased in", the setter is the other way in.

## 4. Stale comment from the previous approach

`py/test/unit/selenium/webdriver/remote/remote_connection_tests.py:475-477` still says
extra_headers "are applied per-connection in `_request`". They're now resolved in `__init__`
and read through the descriptor; `_request` no longer touches them.

## 5. Needs a rebase — one trivial conflict, and the description now describes the old approach

The branch is `CONFLICTING` against trunk. Both source files auto-merge; the only conflict is
in `client_config_tests.py`, where this PR's new tests and trunk's
`test_default_proxy_is_not_shared_between_instances` (`56e5c616d8`) were added at the same
spot. Keeping both blocks is the whole resolution — I merged `origin/trunk` locally that way
and the module is **95/95 green**, with trunk's `proxy is None` change and this PR's
`init_args_for_pool_manager` copy sitting happily in the same `__init__`.

Two things worth knowing about the rebase:

- Trunk also enabled `disallow_untyped_decorators` in `py/pyproject.toml` (`34c744284a`), which
  landed after this branch's base. I checked whether the new `_classorinstancemethod` decorator
  trips it: it doesn't — mypy reports zero errors in `remote_connection.py` and
  `client_config.py` under the stricter config (the only `untyped-decorator` hits are the
  pre-existing `webdriver.py` ones that `34c744284a` already fixed). So the green mypy check
  here holds after the rebase.
- The PR description still describes the previous implementation — "applied per-instance in
  `_request()`" and "class-level `RemoteConnection.extra_headers` / `user_agent` still work as
  process-wide defaults". Neither is true of this head any more: they're resolved once in
  `__init__`, and class-level values are now genuine defaults rather than a floor merged under
  every request. Worth rewriting before merge, since the description is what ends up in the
  squashed commit message.

## 6. Scope note on the `__init__` comment

The new comment is accurate and correctly scoped to headers now. The three lines directly
above it still mirror onto the class, though:

```python
RemoteConnection._timeout = self._client_config.timeout
RemoteConnection._ca_certs = self._client_config.ca_certs
RemoteConnection._client_config = self._client_config
```

So `RemoteConnection._client_config.password` still returns the last-constructed connection's
credentials after `quit()`, and `RemoteConnection.set_timeout(9)` still mutates the
last-built config rather than the class. Both unchanged from trunk and reasonable to leave
here — but it does mean the description's "they outlived `driver.quit()`" only holds for the
header half now, which folds into the rewrite in (5). A follow-up issue for the
`_client_config` mirroring would be worth filing.

## One separate issue, pre-existing

`_request` follows a 3xx `Location` to an arbitrary host and re-sends `extra_headers` plus
`get_auth_header()`, so `Authorization` can reach a redirect target. Unchanged on both trees
and out of scope here, but same "who is allowed to see this header" question.

(The other one I mentioned last time — the shared `Proxy()` default argument — is already
fixed on trunk by `56e5c616d8`, so ignore that.)

AutomatedTester (approved):
Looks good now, please rebase/fix merge conflicts so we can merge this

PR #17957, opened 2026-08-28
[java] shrink published javadoc jars

titusfortner:
### 🔗 Related Issues

While investigating why the 4.48.0 release failed package verification, I found `org.seleniumhq` is over Sonatype's monthly publishing limit with a single release.

### 💥 What does this PR do?

- Published javadoc jars no longer bundle the DejaVu web fonts, which were about 62% of a Selenium release.
- The generated CDP binding artifacts publish a stub javadoc jar instead of ~10 MB of generated pages each.

These changes will take the release size (including htmlunit driver) from ~165 MB to ~55 MB.
@rbri if we release htmlunit3-driver without bundling the dependencies we can drop that to ~28 MB per release.

### 🔧 Implementation Notes

- Javadoc bundles 3.85 MB of DejaVu fonts into every jar it generates and a release ships 22 of them; the stylesheet's font stacks fall back to Arial/Georgia/monospace, so dropping them only changes typography on javadoc.io.
- The Java API docs published to gh-pages do not rely on javadoc jars and are completely unaffected by these changes.
- The Java API docs we publish already exclude the CDP classes, so javadoc.io was the only place they were published and stubbing removes them there too.
- Stubbing CDP has no impact on IDEs, which render documentation from the sources jar.

### 🤖 AI assistance
    - Tool(s): Claude Code (Opus 5)
    - What was generated: the build changes, and the size analysis that traced the release payload to bundled fonts

Options for saving space. Rows marked \* are what this PR does.

| Options                                                 | space saved |
|---------------------------------------------------------|---|
| \* fonts dropped from javadoc jars                      | 84.8 MB |
| htmlunit3-driver does not include jar-with-dependencies | 27.2 MB |
| \* CDP javadoc stubbed                                  | 24.6 MB |
| CDP javadoc options reduced instead of stubbed          | 11.6 MB |
| CDP sources stubbed                                     | 4.8 MB |
| Package two CDP versions instead of three               | 1.7 MB |

Note: #17586 will probably add back another ~4 MB

### 🔄 Types of changes


rbri:
@titusfortner will have a look

rbri:
@titusfortner had a first look at this, but the deploy plugins exclude feature seem not to work. Have opened a ticket, will inform you about the progress. For the weekend i plan an HtmlUnit/driver release, hope to have at least a workaround in place until then.

Thanks for the headsup....

rbri:
@titusfortner found a solution the next release (this weekend) will no longer deploy the file jar-with-dependencies (will only make it available from the github release page.

Again thanks....

Issue #17955, opened 2026-08-27
[🐛 Bug]: Grid 4.46 API /se/files/$file return error 500 with file containing spaces


emerisier:
### Description

Hi
The API returns a 500 error as soon as a file contains spaces.
Tested with "%20", "%2520", and "+".
Bet regards

### Reproducible Code

```shell
/wd/hub/session/$id/se/files

{
  "value": {
    "names": [
      "xxxxx_Attestation pour votre Employeur.pdf"
    ],
    "files": [
      {
        "lastModifiedTime": 1787827948467,
        "creationTime": 1787827948462,
        "size": 123390,
        "name": "xxxxx_Attestation pour votre Employeur.pdf"
      }
    ]
  }
}

wd/hub/session/$id/se/files/xxxxx_Attestation%20pour%20votre%20Employeur.pdf
```

### Debugging Logs

```logs
{
  "value": {
[...]

PR #16844, opened 2026-01-04
[java] rework downloading files from Grid

asolntsev:
### **User description**
* remove the "new" endpoint `GET /se/files/:name` (that was recently added in 4.39.0)
* use the "old" endpoint `POST /se/files`, but with optional "format" parameter in payload. Possible values: "application/octet-stream" or "application/json" (default).

Endpoint `POST /se/files` can respond either in JSON or binary format depending on the "format" parameter.

* Currently only Java binding sends "format=application/octet-stream"
* Other bindings don't yet send "format" parameter, thus downloading file using the old method (which encodes file content into Base64 and Json).
  * Other bindings can gradually migrate to "format=application/octet-stream" as well in following releases.

As a side effect, this PR fixes downloading files with space in name. :)

### 🔄 Types of changes
- Bug fix (backwards compatible)
- Breaking change (fix or feature that would cause existing functionality to change)

___

### **PR Type**
Bug fix, Enhancement

___

### **Description**
- Remove deprecated `GET /se/files/:name` endpoint, use `POST /se/files` with Accept header

- Support binary file downloads via Accept: application/octet-stream header

- Fix downloading files with special characters (spaces, unicode, etc.)

- Add backward compatibility for older Grid versions (4.38.0 and earlier)

- Expand test coverage with parameterized tests for various filename types

___

### Diagram Walkthrough

```mermaid
flowchart LR
  Client["Client<br/>RemoteWebDriver"]
  OldEndpoint["Old Endpoint<br/>GET /se/files/:name"]
  NewEndpoint["New Endpoint<br/>POST /se/files<br/>+ Accept Header"]
  BinaryResponse["Binary Response<br/>application/octet-stream"]
  JsonResponse["JSON Response<br/>for older Grid"]
  
  Client -->|Remove| OldEndpoint
  Client -->|Use| NewEndpoint
  NewEndpoint -->|Accept: octet-stream| BinaryResponse
  NewEndpoint -->|Fallback| JsonResponse
```

<details><summary><h3>File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><details><summary>2 files</summary><table>
<tr>
  <td><strong>LocalNode.java</strong><dd><code>Remove GET endpoint, add binary response support</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
[...]

PR #17968, opened 2026-08-28
[grid] Fix 500 when downloading a file whose name contains spaces

ashrafiucse:
### 🔗 Related Issues                                                                                                                                                              
                                                                                                                                                                                      
   Fixes https://github.com/SeleniumHQ/selenium/issues/17955                                                                                                                          
                                                                                                                                                                                      
   ### 💥 What does this PR do?                                                                                                                                                       
                                                                                                                                                                                      
   `GET /session/{sessionId}/se/files/{fileName}` returns 500 whenever the file                                                                                                       
   name contains a space (reported on 4.46, still present in 4.48):                                                                                                                   
                                                                                                                                                                                      
       java.lang.RuntimeException: Unable to execute request for an existing session:                                                                                                 
       Illegal character in path at index 91: http://x.x.x.x:5555/session/.../se/files/xxx aaa.pdf                                                                                    
       Caused by: java.net.URISyntaxException                                                                                                                                         
                                                                                                                                                                                      
   The root cause is a decode/re-encode asymmetry between the server and the                                                                                                          
   HTTP client:                                                                                                                                                                       
                                                                                                                                                                                      
   1. The Netty server decodes the path of every incoming request                                                                                                                     
      (`QueryStringDecoder.path()` in `RequestConverter`), so `%20` becomes a                                                                                                         
      literal space in the `HttpRequest` URI.                                                                                                                                         
   2. When the router proxies the request to the node, `JdkHttpMessages#getRawUri`                                                                                                    
      builds the target URI with `URI#create`, which rejects characters like a                                                                                                        
      literal space, and the whole request fails with a 500.                                                                                                                          
                                                                                                                                                                                      
   This PR makes the proxy quote such characters, and stops the node from                                                                                                             
   decoding the file name a second time.                                                                                                                                              
                                                                                                                                                                                      
   What changed:                                                                                                                                                                      
                                                                                                                                                                                      
   * `JdkHttpMessages` — characters that are not allowed in a URI are                                                                                                                 
     percent-encoded when building the URI of an outgoing request. Every                                                                                                              
     character that `URI` accepts is left untouched, so URIs that were valid                                                                                                          
     before are sent byte-for-byte exactly as they were.                                                                                                                              
   * `LocalNode#extractFileName` — no longer calls                                                                                                                                    
     `urlDecode(...).replace(' ', '+')`. The path of the request has already                                                                                                          
     been decoded by the server, so the extra decode turned a real file name                                                                                                          
     into `name+with+spaces.pdf`, which never matched the file on disk.                                                                                                               
     The substring after `/se/files/` is now used as-is.                                                                                                                              
   * `HttpClientTestBase` — new round-trip test: a request whose path holds a                                                                                                         
     space (and a non-ASCII character) is sent through a real Netty server, and                                                                                                       
     the handler must receive the same path decoded back.                                                                                                                             
   * `LocalNodeTest` — `extractsFileNameFromRequestUri` now also covers a name                                                                                                        
     with spaces; the existing assertions are unchanged.                                                                                                                              
                                                                                                                                                                                      
   ### 🔧 Implementation Notes                                                                                                                                                        
                                                                                                                                                                                      
   An alternative would be to keep the raw, undecoded path in                                                                                                                         
   `RequestConverter`, but every route (`UrlTemplate`, `getSessionId`, …)                                                                                                             
   matches against the decoded value, so that would be a much larger change.                                                                                                          
   Quoting at the outgoing edge instead is local and provably safe: the                                                                                                               
   transformation is the identity function for every URI that was already                                                                                                             
   valid, so only requests that previously failed with `URISyntaxException`                                                                                                           
   behave differently.                                                                                                                                                                
                                                                                                                                                                                      
   The deprecated `GET /se/files/{name}` endpoint is kept working rather than                                                                                                         
   removed (it was marked for removal in #16844). The `POST /se/files` endpoint,                                                                                                      
   which all bindings use, is unaffected.                                                                                                                                             
                                                                                                                                                                                      
   ### 🤖 AI assistance                                                                                                                                                               
                                                                                                                                                                                      
       - Tool(s): GLM (Z.ai)                                                                                                                                                          
       - What was generated: Root-cause analysis of the reported issue, the fix                                                                                                       
         in `JdkHttpMessages` and `LocalNode#extractFileName`, and the two test                                                                                                       
         additions. I reviewed the final diff, ran the tests locally                                                                                                                  
         (`//java/test/org/openqa/selenium/remote/http:small-tests`,                                                                                                                  
         `//java/test/org/openqa/selenium/grid/node/local:LocalNodeTest`)                                                                                                             
         and verified the fix end-to-end against a running Grid server.                                                                                                               
                                                                                                                                                                                      
   * File names containing a literal `?` still cannot be referenced by this                                                                                                           
     endpoint (the character separates the query string) — same as before this                                                                                                        
     PR.                                                                                                                                                                              
   * A file name containing a literal `%` is ambiguous after decoding, and is                                                                                                         
     likewise out of scope here.                                                                                                                                                      
                                                                                                                                                                                      
   ### 🔄 Types of changes                                                                                                                                                            
                                                                                                                                                                                      
### DIFF
--- a/java/test/org/openqa/selenium/remote/internal/HttpClientTestBase.java
+++ b/java/test/org/openqa/selenium/remote/internal/HttpClientTestBase.java
@@ -336,6 +336,21 @@ public abstract class HttpClientTestBase {
     assertThat(response.contentAsString()).isEqualTo(path);
   }
 
+  @Test
+  void shouldSendRequestWithPathThatNeedsQuotingForNonAsciiCharacters() {
+    // Same scenario as above, but for a file name that also holds non-ASCII characters,
+    // including one outside the Basic Multilingual Plane (a surrogate pair), to exercise the
+    // UTF-8 percent-encoding of a multi-byte code point rather than just a single ASCII byte.
+    String path = "/session/772f83578930be5dce8f626ecf8ea935/se/files/файл tähtedega 😀.pdf";
+
+    HttpResponse response =
+        executeWithinServer(
+            new HttpRequest(GET, path),
+            req -> new HttpResponse().setContent(Contents.utf8String(req.getUri())));
+
+    assertThat(response.contentAsString()).isEqualTo(path);
+  }
+
   private HttpResponse getResponseWithHeaders(final Multimap<String, String> headers) {
     return executeWithinServer(
         new HttpRequest(GET, "/foo"),


joerg1985:
did not look too deep into this but:
1. `%` is uri not safe, i guess this is to hide issues with allready encoded URIs.
2. the `URLEncoder` must know the charset to encode a URI, but this implementation not? suspicious.
3. the `JdkHttpMessages` is a very low level class, high risk of breaking things.


PS: this is a fix to a deprecated method to download the file:
https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/grid/node/local/LocalNode.java#L936-L938
```
      if (req.getMethod().equals(HttpMethod.GET)) {
        // Left here for backward compatibility.
        // Remove this IF in Selenium 4.41, 4.42 or 4.43
        return getDownloadedFile(downloadsDirectory, extractFileName(req));
      }
```

ashrafiucse:
Thanks for taking a look, @joerg1985! Some answers, hopefully they address the concerns:

**1. On `%`** — that one is deliberate, and it is what keeps the change safe for URIs that are *already* encoded. The proxy regularly forwards paths holding legitimate `%XX` sequences, and re-encoding `%` would corrupt them — guava's own `urlFragmentEscaper` turns `a b%20c.pdf` into `a%20b%2520c.pdf`, for example. So `%` passes through untouched, which makes the transformation the identity for every URI that was valid before this change; only characters `URI.create` actually rejects (a literal space, non-ASCII, …) get percent-encoded. The residual ambiguity for a file literally named `50%.pdf` is the same as before the PR — such requests already failed — and is called out under "Additional Considerations".

**2. On the charset** — it is pinned to UTF-8: each disallowed code point is turned into its UTF-8 bytes and percent-encoded per RFC 3986. I avoided `URLEncoder` on purpose, because it is form encoding (`application/x-www-form-urlencoded`) and writes `+` for spaces — exactly the historical bug in `extractFileName` this PR removes. Guava's escaper cannot be dropped in either, for the `%2520` reason above.

**3. On the risk in `JdkHttpMessages`** — agreed that it is low level, which is why the quoter is conservative: identity for everything that already worked, and only requests that previously died with `URISyntaxException` behave differently. The round-trip test in `HttpClientTestBase` pushes a request with a space and a non-ASCII character through a real Netty server and asserts the handler sees the same decoded path back, for both client implementations. If there is a specific scenario you are worried about, I am happy to add a test for it.

**PS. on the deprecated endpoint** — fair point. The `JdkHttpMessages` part is endpoint-agnostic and fixes proxying for *any* request whose decoded path holds characters `URI` refuses. The `LocalNode` hunk only keeps the legacy GET from mangling file names until its scheduled removal; if you would rather drop that hunk and pull the removal forward, that works for me too.

diemol (approved):

